Issue #4, opened 2023-11-29
Create a set of issue labels to support efficient task management


Scorpil:
I propose a following set of github issue labels to organize community's and core team's work:

* `pending decision`: every newly created issue needs to be reviewed/verified by at least one core team member
* `bug`: bug report
* `question`: usage question
* `feature`: feature request
* `improvement`: general optimizations and improvement proposals
* `rfc`: Request for Comments, structured change request for public discussion, required for large scope changes
* `docs`: documentation improvements
* `help wanted`: open for community to work on
* `good first issue`: simple issue, good for beginner contributors

Thoughts?

Issue #3, opened 2023-11-29
Adjust issue template and PR template to fit OpenBao


Scorpil:
Issue template and PR template should fit OpenBao workflow.

joewxboy:
Can we close this issue, or are there more PRs needed to completely address?

Scorpil:
Right, this one's done 👍

PR #5, opened 2023-11-29
Adjust issue template and PR template to fit OpenBao

Scorpil:
Solves #3 

Removing plugin submission template for now.
Added RFC template based on OpenTofu's with slight modifications.

Labels are configured based on #4


jbutlerdev:
Personally I like bao. It looks like opentofu uses tofu, so it would make sense to follow their lead

Scorpil:
Agree @jbutlerdev. I don't believe there are any popular `bao` packages that could cause a clash on some systems, and being in sync with OpenTofu is a plus.

naphelps:
Make sure all commits use `-s` or `--signoff` for sign off.

.github/ISSUE_TEMPLATE/rfc.yml
```diff
@@ -0,0 +1,145 @@
...
+        - Are there any disadvantages of implementing this?
+      placeholder:
+      value:
+    validations:
+      required: false
+
+  - type: textarea
+    id: security
+    attributes:
+      label: Security Implications
```
cipherboy, line 98: I don't think its worth highlighting _everything_ in the RFC design, but perhaps adding user experience here would be a good catch-all bucket: e.g., developer experience integrating with this proposed change, operator experience trying to manage/configure it, ...

Just to make sure it gets thought about before dropping a (complex but useful) feature in...

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -2,16 +2,16 @@
...
 
 ---
 
 <!-- Please reserve GitHub issues for bug reports and feature requests.
 
-For questions, the best place to get answers is on our [discussion forum](https://discuss.hashicorp.com/c/vault), as they will get more visibility from experienced users than the issue tracker.
+For questions, the best place to get answers is on our [mailing list](https://lists.lfedge.org/g/openbao), as they will get more visibility from experienced users than the issue tracker.
 
-Please note: We take Vault's security and our users' trust very seriously. If you believe you have found a security issue in Vault, please responsibly disclose by contacting us at security@hashicorp.com. Our PGP key is available at [our security page](https://www.hashicorp.com/security/).
+Please note: We take OpenBao's security and our users' trust very seriously. If you believe you have found a security issue in OpenBao, please responsibly disclose by contacting us at TODO: security email. Our PGP key is available at [TODO: our security page](TODO).
```
joewxboy, line 14: Security email is openbao-security@lists.lfedge.org

joewxboy (changes requested):

.github/ISSUE_TEMPLATE/rfc.yml
Scorpil: Fair enough, added ✅

naphelps (approved):
Looks good to me!

PR #18, opened 2023-12-11
import from github.com/openbao/openbao throughout project

alrs:
This replaces every mention of `github.com/hashicorp/vault` in the go source code with `github.com/openbao/openbao` to avoid importing non-MPL code.


alrs:
I suspect that the `go.mod` fix here should be merged first https://github.com/openbao/openbao-plugin-secrets-kv/pull/7

naphelps:
Merged in a different pull request.

PR #26, opened 2023-12-18
chore: small format changes in markdown files

yriveiro:
# Description

This PR performs formatting changes on top of some markdown files to improve standardization.

As part of this PR two new workflows were added:
- One to linting markdown files using `markdownlint` as added as well as suggested in the comments
- A [problem-matcher](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) workflow

The sets of rules are the default ones defined in markdownlint [rules](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md), if the community decide to drift from the current default, markdownlint will look `.markdownlint-cli2.jsonc` for custom configurations.


yriveiro:
Do I need to add the `changelog/PR.txt` file as well? From what I saw the PR counter was reset in Openbao and the ones in the folder are 24XXX (I assuming, they are PR number from the backports)

naphelps:
@yriveiro The community has not formally decided what the changelog format and structure will be under Openbao. I am not sure documentation adjustments warrant a standalone line item in the changelog.

cipherboy:
>  I am not sure documentation adjustments warrant a standalone line item in the changelog.

I concur. I've filed #31 for discussing the changelog process in general. :-)

yriveiro:
Changes in markdown files should not trigger go linter, at the end of the day, you are consuming action minutes in files that are not in the scope.

cipherboy:
@yriveiro Sorry, let me rephrase. If we're going to accept a PR doing _markdown_ linting, we should add a markdown linting workflow so all future PRs will comply, rather than periodically drifting and re-syncing with the linter as someone runs it manually. :-) Otherwise, there's no value in accepting this PR, as it will just continue to drift as not everyone naturally writes markdown that compiles with this particular linter.

In general, there's probably an argument to be said about compatibility and preventing obvious non-markdown, so I think there's probably value in adding the linter to the workflows. My 2c!

yriveiro:
:smile: OK got it, I will work in the workflow linting as part of this PR.

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,32 @@
+name: Markdownlint
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "**/*.md"
```
cipherboy, line 7: @yriveiro Not to cause more questions for you... But do you know if it supports [`.mdx` files](https://github.com/openbao/openbao/blob/development/website/content/api-docs/secret/pki.mdx)? All of our API/usage docs are currently using a combination of Markdown and JSX used by Vercel, that could be linted as well. :-)

yriveiro: Unfortunately, `markdownlint` doesn't support mdx files https://github.com/DavidAnson/markdownlint/issues/723#issuecomment-1434017183

One way to overcome this could be adding a new step to the workflow that renders mdx files with something like `@mdx-js/mdx`, once we get the output in a temporary folder we run the linting.

yriveiro: It seems that what I said in the previous comment is not accurate, indeed `markdownlint` doesn't support mdx files, nevertheless, the action uses `markdownlint-cli2` and this one supports mdx.

The folder `website` will probably throw a lot of warnings as most of the code is generated and that implies that the generator needs to implement the defined rules the community decides first.

cipherboy: Ah cool!

From what I recall, I don't think many of the `.mdx` files under `website/` are auto-generated. I know a lot of them were manually maintained, so I don't think we're liable to overwrite them any time soon. That said, I definitely agree, there's a lot of stuff there that is not well formatted and it would be time consuming to fix...

It doesn't look like there's an easy changed-files-only version of this workflow... I think we'd have to look at `GITHUB_BASE_REF` and `GITHUB_SHA` to see which files changed and then look for `**/*.mdx` files (like we do [for the changelog checker](https://github.com/openbao/openbao/blob/development/.github/workflows/changelog-checker.yml)). Perhaps best to ignore them for now then given there are 1k+ of them...

Sorry for the dead-end!

JanMa:
Closing this because we have entirely rewritten the website and docs in the meantime

PR #28, opened 2023-12-21
gofmt entire project

alrs:
I'm doing some exploratory work to deal with the circular imports between this repo and https://github.com/openbao/openbao-plugin-secrets-kv, and I noticed that my diffs were being muddied by `gofmt` tweaks affecting source files throughout the project.

This commit runs `gofmt` over every `.go` file in the repo to do this as a big bang and keep spurious changes out of subsequent diffs.


cipherboy (approved):
Looks like you need DCO on this one @alrs, otherwise LGTM. 

Maybe a name that begins with "va" would've been better... :D

cipherboy:
@naphelps this should be good to merge. :-) 

I'll file an issue about making gofmt be a required gating test.

naphelps (approved):

PR #2, opened 2023-11-29
Update go mod ref

jbutlerdev:
Update `hashicorp/vault` references to point to `lf-edge/openbao`

This will currently fail to build as the `github.com/lf-edge/openbao/vault/hcp_link/proto` needs to exist with the updated `go.mod` file before the `go.sum` file can be properly updated

PR #30, opened 2023-12-22
Merge github.com/openbao/openbao-plugin/secrets/kv Into Main Repo

alrs:
This repo currently imports code from https://github.com/hashicorp/vault-plugin-secrets-kv. The openbao version of that repo imports code from the legacy Hashicorp repo.

Here:
`github.com/hashicorp/vault-plugin-secrets-kv v0.15.0`

There:
`github.com/hashicorp/vault/api v1.3.0            `                                             
`github.com/hashicorp/vault/sdk v0.3.1-0.20220112143259-b48602fdb885`                    

 his PR resolves this circular import by moving `openbao-plugin-secrets-kv` into the local package `plugins/secrets/kv`. Git history for the old repo was retained in this merge.

The README in the new `plugins/secrets/kv` package is out-of-date, but it is also out of date in https://github.com/openbao/openbao-plugin-secrets-kv, so I didn't address it here.

Should this PR land, next steps would be to continue fixing up the `kv` package here, and removing the https://github.com/openbao/openbao-plugin-secrets-kv repo.


cipherboy:
@alrs I'm curious what motivated you to do this approach?

I personally might've removed the dependency on the other way, i.e., by [using our forked version](https://github.com/openbao/openbao-plugin-secrets-kv) and now that our `development` branch here builds, [substituting `hashicorp/vault` with `openbao/openbao`](https://github.com/openbao/openbao-plugin-secrets-kv/blob/master/go.mod) instead.

Then here, point `hashicorp/vault-plugin-secrets-kv` at `openbao/openbao-plugin-secrets/kv`.

This keeps the circular dependency (like upstream has) but points everything at our own code and avoids dragging in upstream as a dependency. 

 (I believe that's why we didn't do that earlier with that plugin -- we hadn't yet merged #2 here and so neither were buildable. While we don't have a tag in place yet, we could point it at a commit for the time being.)


One of the complaints about upstream is the large binary size: if we go the opposite direction (and take plugins like Transit and PKI and make them external -- of which the first step would be separate repository like K/V, but then would also require a further step to build them as an external plugin binary) -- we could potentially reduce the dependencies in the main `bao` binary and drop its size significantly (especially once we get to the Azure plugin -- the Azure SDK is used for minimal functionality but IIRC contributes greatly to binary size). We could do the same with `api` and `sdk`: while the core repo and all plugins would still depend on it, this code is meant to be used externally (via go imports -- unlike e.g., `github.com/openbao/openbao/vault`)  and we could benefit from making that clearer, at the expense of potentially making PRs harder (in that changes might require updates in two or more repos).

On the other hand, there is a good argument to be made that this simplifies development and maintenance since it is all in the same repository... And K/V in particular _is_ a central use case of a secrets manager, so maybe it should be treated differently... But IMO external plugins bring potential benefits (like running them in containers like upstream has started to introduce).

But curious your thoughts :-)

alrs:
@cipherboy: I did use the forked version, and merged it into this repo. `plugins/secrets/kv` came directly from https://github.com/openbao/openbao-plugin-secrets-kv.

Moving `kv` into this repo does nothing to change the binary sizes.

cipherboy:
@alrs Hmmm on second though, playing around with this more, I kinda like this. @jbutlerdev I know this breaks with what we discussed on Matrix (and not deviating from upstream's processes too much), but I don't think this would prevent us from splitting K/V out as a separate binary later (from the same repo). And it makes it clear that the plugin needs to be versioned with the binary (how it is today, upstream). 

Do you have any thoughts on this matter, @scorpil?

plugins/secrets/kv/README.md
```diff
@@ -0,0 +1,125 @@
+# Vault Plugin: Key-Value Secrets Backend [![Build Status](https://travis-ci.org/hashicorp/vault-plugin-secrets-kv.svg?branch=master)](https://travis-ci.org/hashicorp/vault-plugin-secrets-kv)
```
cipherboy, line 1: @alrs Not sure which file to comment on, but could we move this tree to `builtin/logical/kv` rather than `plugins/secrets/kv`? That's keeping in line with the other built-in plugin locations (PKI, ...).

alrs: I'm happy to move it to wherever there is consensus.

Scorpil:
I like the idea of having this particular plugin locally, I think it's a right place for it in the long run. However moving the code now will make it much harder incorporating new commits from upstream. I'd keep the repo external until when we drop api-compatibility.

@cipherboy @alrs

cipherboy:
@Scorpil Hmm, I think we could rewrite upstream patches with `git apply` to be relative to the new directories... Take a little experimentation but it shouldn't be too difficult, and would be easier than recreating the patch from scratch.

plugins/secrets/kv/README.md
cipherboy: @alrs This does need to be at `builtin/logical/kv`, with `cmd/kv` as a subpackage under that.

See:

https://github.com/openbao/openbao/blob/04f1f227b9e65575385218a9d5d883ad13bd2be7/vault/testing.go#L1686

which in turn expects stuff under `builtin/logical`, e.g.:

https://github.com/openbao/openbao/blob/04f1f227b9e65575385218a9d5d883ad13bd2be7/helper/testhelpers/pluginhelpers/pluginhelpers.go#L45-L50

So swapping `consul` to `kv` won't work (if we remove `consul`), with `kv` here.

naphelps:
Merged in a different pull request.

cipherboy:
Thank you @alrs for the suggestion :-) credit should be preserved in #109.

Issue #35, opened 2024-01-06
RFE - XChaCha20 support


cipherboy:
Copied from `https://github.com/hashicorp/vault/issues/13243`, reported by @DemiMarie:

> **Is your feature request related to a problem? Please describe.** Currently, `transit` keys need to be rotated after 2³² uses, due to the algorithm employed (AES-GCM with a random nonce).
> 
> **Describe the solution you'd like** Use a long, random nonce with an algorithm such as XChaCha20-Poly1305. For algorithms such as AES-GCM that don’t support this natively, one can use any KDF with a root key and any random number, and include the number in the nonce.
> 
> **Describe alternatives you've considered** Maintain a stateful counter on each node, together with a consensus algorithm to ensure that each node has a unique node number. This is not particularly attractive, since it means a consensus problem turns into a cryptographic disaster.
> 
> **Explain any additional use-cases** Operations are dramatically simplified: users do not need to worry about key rotation. The overhead of the KDF should be dwarfed by the overhead of the Vault RPC itself, unless I am missing something.
> 
> **Additional context** Most people do not deal with key rotation (`https://github.com/hashicorp/vault/issues/13025#issuecomment-961889260`), nor should they need to.

PR #36, opened 2024-01-06
Add support to Transit, keysutil for XChaCha20-Poly1305

cipherboy:
This adds support for XChaCha20-Poly1305, a variant of ChaCha20-Poly1305 with double the nonce size (12->24). Due to the construction of most AEAD cipher modes which lack a synthetic IV (mostly this and GCM), nonce reuse becomes problematic. A small default nonce size (96 bits) results in only 96/2 = 2^48 message encryptions before reuse becomes likely due to the birthday paradox. As a result, NIST has mandated 2^32 as the upper limit on AES-GCM encryption operations in FIPS to keep the threshold sufficiently small. This necessitates key rotation when the limit is reached.

By using a larger nonce (24-bytes), key rotation can be avoided.

Resolves: https://github.com/openbao/openbao/issues/35

cc: @DemiMarie

---

This can be tested by rebasing on top of my `prune-extra-plugins` branch (to fix the build & most tests) and running `go test github.com/openbao/openbao/builtin/logical/transit` to avoid hitting issues discussed [on matrix](https://matrix.to/#/!iNNKBYObdlLtUybZCN:chat.lfx.linuxfoundation.org/$SDTWkFO82Vt2bvlcOSygaa7noCqblF8gU433l1XDXFE?via=chat.lfx.linuxfoundation.org&via=matrix.org&via=chat.ucml.dev).


website/content/api-docs/secret/transit.mdx
```diff
@@ -722,8 +728,9 @@ will be returned.
 - `nonce` `(string: "")` – Specifies the **base64 encoded** nonce value. This
   must be provided if convergent encryption is enabled for this key and the key
   was generated with Vault 0.6.1. Not required for keys created in 0.6.2+. The
-  value must be exactly 96 bits (12 bytes) long and the user must ensure that
-  for any given context (and thus, any given encryption key) this nonce value is
+  value must be exactly 96 bits (12 bytes) long for AES and ChaCha20 or 192
+  bits (24 bytes) long for XChaCha20 and the user must ensure that for any
+  given context (and thus, any given encryption key) this nonce value is
```
DemiMarie, line 733: Does this need to be supported for XChaCha20?  I don’t think it is of any value there, since Vault 0.6.1 didn’t support XChaCha20.  Therefore, it just adds a way for users to do something insecure.

cipherboy: I do wonder if we can just remove it. Upstream wasn't too happy about that, but given how old `v0.6.x` is, it might be fine to do now for the fork.

I think this accurately reflects the code though. While there is no way to do this via API, if you overwrite storage and fudge the key version there, I think you can force the backend into this state, and so this documentation would be correct if you do that.

cipherboy: Filed https://github.com/openbao/openbao/issues/37.

JanMa (approved):
Lgtm 👍

naphelps (approved):

PR #44, opened 2024-01-12
replace Vault with OpenBao in documentation

JanMa:
This replaces `vault` with `openbao` in the documentation where I think it makes sense. As with the API docs, I left references to older Vault versions and the Enterprise offering in tact. I also did not change any API responses or `X-Vault` HTTP headers.

Sorry in advance for the huge PR :-)


cipherboy:
Will take a look at this when I get back!

website/content/docs/agent-and-proxy/agent/apiproxy.mdx
```diff
@@ -1,62 +1,62 @@
...
+# OpenBao agent API proxy
 
-Vault Agent's API Proxy functionality allows you to use Vault Agent's API as a proxy
-for Vault's API.
+OpenBao Agent's API Proxy functionality allows you to use OpenBao Agent's API as a proxy
+for OpenBao's API.
 
 ~> Note: This functionality will be deprecated in a future release. Please
-switch to using [Vault Proxy](/vault/docs/agent-and-proxy/proxy) for API proxying purposes, instead.
+switch to using [OpenBao Proxy](/vault/docs/agent-and-proxy/proxy) for API proxying purposes, instead.
```
joewxboy, line 15: I don't know if/when we'll also want to change the path here and in subsequent locations.

website/content/docs/agent-and-proxy/agent/caching/index.mdx
```diff
@@ -232,13 +232,13 @@ Here is an example of a cache configuration with the optional `persist` block,
...
 ```hcl
-# Other Vault agent configuration blocks
+# Other OpenBao agent configuration blocks
 # ...
 
 cache {
 	persist = {
 		type = "kubernetes"
-		path = "/vault/agent-cache/"
+		path = "/openbao/agent-cache/"
```
joewxboy, line 241: Let's confirm the cache path change here and in subsequent locations.

website/content/docs/agent-and-proxy/agent/caching/persistent-caches/kubernetes.mdx
```diff
@@ -1,21 +1,21 @@
...
 
-The Vault Agent persistent cache file in Kubernetes should only be used for handing off Vault tokens
-and leases between initialization and sidecar Vault Agent containers. This cache file should be shared
-using a memory volume between the Vault Agent containers.
+The OpenBao Agent persistent cache file in Kubernetes should only be used for handing off OpenBao tokens
+and leases between initialization and sidecar OpenBao Agent containers. This cache file should be shared
+using a memory volume between the OpenBao Agent containers.
 
-If the Vault Agent Injector for Kubernetes is being used, the persistent cache is automatically configured
+If the OpenBao Agent Injector for Kubernetes is being used, the persistent cache is automatically configured
```
joewxboy, line 18: The ID target in the URL (text after the hash sign) is likely based on a section title, and has likely already been updated.  Please check

website/content/docs/agent-and-proxy/agent/generate-config/index.mdx
```diff
@@ -40,13 +40,13 @@ unless a path is specified as an argument.
 Before generating a configuration file, let's insert a secret `foo`:
 
 ```shell-session
-$ vault kv put -mount=secret foo user="admin" password="s3cr3t"
+$ openbao kv put -mount=secret foo user="admin" password="s3cr3t"
```
joewxboy, line 43: The binary/CLI for `openbao` is just `bao`, I believe.  We'll need to change these on all the pages.

website/content/docs/agent-and-proxy/agent/generate-config/index.mdx
```diff
@@ -75,7 +75,7 @@ auto_auth {
     type = "token_file"
 
     config {
-      token_file_path = "/Users/avean/.vault-token"
+      token_file_path = "/Users/avean/.openbao-token"
```
joewxboy, line 78: Let's confirm that the token name has changed, and if it is "openbao" or just "bao"

website/content/docs/agent-and-proxy/agent/index.mdx
```diff
@@ -131,25 +131,25 @@ These are the currently-available general configuration options:
...
   identical in function to `caching` for historical reasons. Can also be configured by setting the `VAULT_AGENT_DISABLE_KEEP_ALIVES`
   environment variable as a comma separated string. This environment variable will override any values found in a configuration file.
 
-- `template` <code>([template][template]: <optional\>)</code> - Specifies options used for templating Vault secrets to files.
+- `template` <code>([template][template]: <optional\>)</code> - Specifies options used for templating OpenBao secrets to files.
 
 - `template_config` <code>([template_config][template-config]: <optional\>)</code> - Specifies templating engine behavior.
 
-- `exec` <code>([exec][process-supervisor]: <optional\>)</code> - Specifies options for vault agent to run a child process
+- `exec` <code>([exec][process-supervisor]: <optional\>)</code> - Specifies options for openbao agent to run a child process
```
joewxboy, line 148: Looks like a proper name, so "OpenBao" instead of "openbao"

website/content/docs/agent-and-proxy/agent/index.mdx
```diff
@@ -158,7 +158,7 @@ These are the currently-available general configuration options:
 
 - `log_level` - Equivalent to the [`-log-level` command-line flag](#_log_level).
 
-  ~> **Note:** On `SIGHUP` (`kill -SIGHUP $(pidof vault)`), Vault Agent will update the log level to the value
+  ~> **Note:** On `SIGHUP` (`kill -SIGHUP $(pidof openbao)`), OpenBao Agent will update the log level to the value
```
joewxboy, line 161: "pidof bao" I think

website/content/docs/agent-and-proxy/agent/index.mdx
```diff
@@ -176,22 +176,22 @@ These are the currently-available general configuration options:
 There can at most be one top level `vault` block, and it has the following
 configuration entries:
 
-- `address` `(string: <optional>)` - The address of the Vault server to
+- `address` `(string: <optional>)` - The address of the OpenBao server to
   connect to. This should be a Fully Qualified Domain Name (FQDN) or IP
-  such as `https://vault-fqdn:8200` or `https://172.16.9.8:8200`.
+  such as `https://openbao-fqdn:8200` or `https://172.16.9.8:8200`.
```
joewxboy, line 181: Let's confirm that hostname.

website/content/docs/agent-and-proxy/agent/index.mdx
```diff
@@ -272,44 +272,44 @@ listener configuration, an Agent's listener configuration also supports the foll
...
    Configuration](#example-configuration) section for an example configuration.)
 
-1. Start a Vault Agent with the configuration file.
+1. Start a OpenBao Agent with the configuration file.
 
    **Example:**
 
    ```shell-session
-   $ vault agent -config=/etc/vault/agent-config.hcl
+   $ openbao agent -config=/etc/openbao/agent-config.hcl
```
joewxboy, line 303: likewise, while the CLI is "bao" instead of "openbao", let's also check to see if the config path is "bao" or "openbao"

joewxboy (changes requested):
In summary ...
1) Filenames and paths likely use either "openbao" or just "bao" but let's confirrm with the development team.  These include cacert, binpath, config, token, and logfiles.
2) The hostname is either "openbao-fqdn" or just "bao-fqdn" but I don't know which.  Please confirm.
3) The CLI (or shell) is definitely just "bao" and not "openbao".  Please change all instances.
4) Let's NOT change the documentation page filenames and paths just yet.  Let's handle that in a separate change.  This would also include the section IDs, which appear in the URL past the hash sign (#).
5) Let's confirm the service name, display name, and registry key, both to ensure it is "openbao" and not "bao" but also to confirm the case.

website/content/docs/agent-and-proxy/agent/apiproxy.mdx
JanMa: Until now I left the path names intact on purpose. This allows us to still build the documentation until we integrated it into the new website.

website/content/docs/agent-and-proxy/agent/caching/index.mdx
JanMa: Those path-names are arbitrary hand have no default value. So I think using `/openbao` here makes sense.
See: https://developer.hashicorp.com/vault/docs/agent-and-proxy/agent/caching#path

website/content/docs/agent-and-proxy/agent/generate-config/index.mdx
cipherboy: So far this hasn't changed, we can use what we want:

https://github.com/cipherboy/openbao/blob/8dead566a8ab017c97a6417df33d326167eb5f57/command/token/helper_internal.go#L35-L39

Using openbao might be better for clarity, IMO?

website/content/docs/agent-and-proxy/agent/index.mdx
cipherboy: Hostname isn't defined by Bao, it is just used for a placeholder.

website/content/docs/agent-and-proxy/agent/index.mdx
cipherboy: So far we haven't updated the configuration file yet either:

https://github.com/openbao/openbao/blob/8dead566a8ab017c97a6417df33d326167eb5f57/.release/linux/package/usr/lib/systemd/system/vault.service#L6 


Interestingly, the systemd service uses `vault.d` whereas all of the docs appear to refer to `vault` sans `.d`. 

Maybe we should standardize on `openbao.d` and use that everywhere? Thoughts @JanMa ?

website/content/docs/agent-and-proxy/agent/generate-config/index.mdx
cipherboy: :+1: and I can confirm this change has been made to the build system already :-)

website/content/docs/agent-and-proxy/agent/apiproxy.mdx
cipherboy: We may want to consider adding a redirect: 

https://github.com/openbao/openbao/blob/8dead566a8ab017c97a6417df33d326167eb5f57/website/redirects.js#L88-L92

That way anyone wanting to replace the URLs and compare between upstream and us can do so easily.

website/content/docs/agent-and-proxy/agent/index.mdx
chicks-net: The `.d` convention is for cases where you read in all of the config files in a directory and then process them together.  Since we are reading single config files, the `openbao` convention, leaving off the `.d` seems like the better one for this use case.

cipherboy: @chicks-net Ah, good point. I don't think it likely that we'll update to support a configuration directory, as I'm not sure if it's possible without language work. We'll see. :-)

Let's do `openbao` sans `.d` then.

website/content/docs/agent-and-proxy/agent/generate-config/index.mdx
JanMa: That's good to know 👍 ,
I'll change the name of the CLI accordingly

website/content/docs/agent-and-proxy/agent/generate-config/index.mdx
JanMa: I'd be in favor of `.openbao-token` even though the CLI is named `bao`. But I think for now I need to revert that change because the token path does not seem configurable in many cases. Also there's the question if we want to keep supporting `.vault-token` files at least for some time to ease adoption. Same goes for `X-Vault` HTTP headers, but that's a discussion for another time :-)

website/content/docs/agent-and-proxy/agent/index.mdx
JanMa: I am also in favor of omitting the `.d` suffix. Not sure why upstream chose it only for the systemd unit, maybe because this is the standard naming convention in systemd. 🤔

website/content/docs/agent-and-proxy/agent/caching/persistent-caches/kubernetes.mdx
JanMa: This ID was not renamed. See: https://github.com/JanMa/openbao/blob/sanitize-docs/website/content/docs/platform/k8s/injector/annotations.mdx#L199

website/content/docs/auth/jwt/oidc-providers/index.mdx
```diff
@@ -8,12 +8,12 @@ description: OIDC provider configuration quick starts
...
 This page collects high-level setup steps on how to configure an OIDC
 application for various providers. For more general usage and operation
-information, see the [Vault JWT/OIDC method documentation](/vault/docs/auth/jwt).
+information, see the [OpenBao JWT/OIDC method documentation](/vault/docs/auth/jwt).
 
 OIDC providers are often highly configurable, and you should become familiar with
 their recommended settings and best practices. The guides listed below are
 largely community-driven and intended to help you get started. Corrections
-and additions may be submitted via the [Vault Github repository](https://github.com/hashicorp/vault).
+and additions may be submitted via the [OpenBao Github repository](https://github.com/hashicorp/openbao).
```
joewxboy, line 16: This repo path is incorrect.  It should be openbao/openbao instead of hashicorp/openbao

joewxboy (approved):
Two items to address in a separate (future) GitHub issue:

1. Style guide to address whether "server", "token", "agent", "proxy" should be capitalized when used as a proper name, such as "OpenBao Agent".  The documentation is currently inconsistent in usage.

2. Ensure all GitHub repo path references are correct.

website/content/docs/auth/jwt/oidc-providers/index.mdx
JanMa: Good catch! I'll make sure to fix the repo references. 👍

JanMa: But since this PR is already approved, I'll open a new one for it once this is merged

naphelps (approved):

cipherboy (approved):
Wow, a lot of really good content here! Thank you @JanMa for your attention to detail :-)

PR #56, opened 2024-01-21
Draft: Plugin removal proposal

cipherboy:
This proposes to remove all plugins with non-OSI integrations (storage, secrets, auth, and database). 

See https://gist.github.com/cipherboy/2217ee39fbdf240cdeac3b282db2b326#auth for context on this proposal. Replaces #54.

In the context of #30, this merges various external plugins in-tree.


cipherboy:
This has been merged and split as #99 to #109.

PR #57, opened 2024-01-23
Re-add stepwise to sdk

cipherboy:
Note that this package is in the SDK so it remains importable by third-party plugins wishing to test against our instance.
    
However, this allows it to be maintained alongside our distribution and is one fewer external dependency to maintain.

This was removed in 4b96eaf53fb92140e84cf8ad44afe1921526b136, though under a different path; this aligns more with the new `sdk/helper/testcluster` package naming that Nick added.  

---

Without this change (or, forking `vault-testing-stepwise` into our namespace), we get failures like:

```
[cipherboy@lp10 openbao]$ go test github.com/openbao/openbao/builtin/credential/userpass
/home/cipherboy/GitHub/openbao/openbao
# github.com/openbao/openbao/builtin/credential/userpass [github.com/openbao/openbao/builtin/credential/userpass.test]
builtin/credential/userpass/stepwise_test.go:22:20: cannot use api.PluginTypeCredential (constant 1 of type "github.com/openbao/openbao/api".PluginType) as "github.com/hashicorp/vault/api".PluginType value in struct literal
builtin/credential/userpass/stepwise_test.go:61:11: cannot use func(resp *api.Secret, err error) error {…} (value of type func(resp *"github.com/openbao/openbao/api".Secret, err error) error) as stepwise.AssertionFunc value in struct literal
FAIL	github.com/openbao/openbao/builtin/credential/userpass [build failed]
```


cipherboy:
Note that failing DCO relates to the existing commits from `vault-testing-stepwise` and were copied verbatim here by `git subtree add` (and thus, have the same commit hashes). Adding artificial DCO to these would change the commit hashes.

naphelps (approved):

PR #63, opened 2024-01-25
Remove HCP Link from OpenBao

cipherboy:
HCP Link is a service to link local Vault Enterprise instances to HashiCorp's HCP cloud offering. This isn't useful for consumers of OpenBao and we do not wish to have to maintain it. Thus, removal is most prudent.


cipherboy:
Note: I tested this by applying it on top of my `pruned-with-stepwise` branch and running various tests. The core set of tests are still failing, but AFAICT, no new tests fail.

naphelps (approved):

Issue #23, opened 2023-12-15
Remove Enterprise documentation


JanMa:
The existing documentation is full of references to the Enterprise Version of Vault. Those should be removed because OpenBao is not likely to support any of the former Enterprise features in it's initial releases.

JanMa:
Closed in #65

PR #65, opened 2024-01-25
Remove Enterprise documentation

JanMa:
This PR removes all documentation of Vault Enterprise features. Note that I have not yet cleaned all mentions of Vault namespaces, since many APIs of the OSS version return stub namespace information. We need to have a seperate discussion how we want to handle those.

Closes: #23


website/content/docs/concepts/storage.mdx
```diff
@@ -15,7 +15,7 @@ storage backend is untrusted storage used purely to keep encrypted information.
 
 @include 'ent-supported-storage.mdx'
 
-Many other options for storage are available with community support for OpenBao - see our
+Many other options for storage are available with for OpenBao - see our
```
joewxboy, line 18: Remove the word "with" to make the sentence grammatically correct.

joewxboy (changes requested):
One small change noted.

cipherboy (commented):
A few more things to remove I think:

```
[cipherboy@lp10 website]$ gif 'Performance Secondary'
content/docs/release-notes/1.5.0.mdx:25:- A new DR secondary replication dashboard, and an updated performance secondary replication dashboard
content/docs/release-notes/1.7.0.mdx:46:- **Client Controlled Consistency (Enterprise):** New options to use configurable headers to control the consistency of reads after writes to performance secondary clusters and performance standby nodes.
content/docs/upgrading/upgrade-to-1.1.1.mdx:67:possible from a performance secondary. These have been resolved, and these
content/docs/upgrading/upgrade-to-1.1.1.mdx:68:operations may now be run from a performance secondary.
content/docs/upgrading/upgrade-to-1.13.x.mdx:104:If a revocation request comes in to a standby or performance secondary node,
content/docs/upgrading/upgrade-to-1.2.0.mdx:118:possible from a performance secondary. These have been resolved, and these
content/docs/upgrading/upgrade-to-1.2.0.mdx:119:operations may now be run from a performance secondary.
content/docs/upgrading/upgrade-to-1.2.4.mdx:25:after a mount filter has been created on an upstream Performance secondary
content/partials/known-issues/update-primary-data-loss.mdx:19:  Calling `update-primary` on a performance secondary with local data in shared
content/partials/known-issues/update-primary-data-loss.mdx:51:Reindex the performance secondary to update the merkle tree with the missing
content/partials/pki-forwarding-bug.mdx:6:intended.  Furthermore, if a certificate is subsequently revoked on a performance secondary node, the secondary's 
content/partials/pki-forwarding-bug.mdx:9:Certificates issued after the fix are correctly stored locally to the performance secondary.
[cipherboy@lp10 website]$ gif 'Performance Replicat'
content/docs/release-notes/1.10.0.mdx:73:Vault’s [eventual consistency](/vault/docs/enterprise/consistency) model precludes read-after-write guarantees when clients interact with performance standbys or performance replication clusters. The [Client Controlled Consistency](/vault/docs/enterprise/consistency#vault-1-7-mitigations) mitigations supported with Vault 1.7 provide ways to achieve consistency through client modifications or by using the agent for proxied requests, which is not possible in all cases. The Server Side Consistent Tokens feature provides an implicit way to achieve consistency by embedding the minimum Write-Ahead-Log state information in the Service tokens returned from logins or token-create requests. This feature introduces changes in the token format and the new tokesn will be the default tokens starting in Vault 1.10.0. Vault 1.10.0 is backwards compatible with old tokens.
[cipherboy@lp10 website]$ gif 'performance primary'
content/docs/release-notes/1.5.0.mdx:26:- Updated DR and performance primary dashboards that include a list of known secondaries and their corresponding UI URLs (for easier context-switching between the UIs)
```

JanMa:
Thanks for double-checking @cipherboy. So far I left all upgrade docs and release notes untouched because i plan to remove them in a future PR. But I did miss the `managed keys` sections in some docs and removed them accordingly :+1:

joewxboy:
> A few more things to remove I think:
> 
> ```

Are we keeping a comprehensive list of what has been removed?  I suspect we'll need it.

cipherboy:
> > A few more things to remove I think:
> 
> Are we keeping a comprehensive list of what has been removed? I suspect we'll need it.

Note that these are removals by virtue of forking sans Enterprise. So mostly reducing references to Enterprise in the documentation, plus features which we know to be Enterprise only already. 

The Enterprise code base lives in a separate, internal repo and thus we never had that code and needed to sync up code+docs removals. Upstream used the OSS repo's docs for both OSS + Enterprise. 

I don't know that I see a quick, concise list of Enterprise-only features on HashiCorp's website though. For the most part, its things like Performance Replication, Disaster Recovery Replication, Seal Wrapping, Managed Keys, and various other features which layer on top of the former (e.g., PKI has a cross-cluster CRL mentioned in these docs, necessitated by having performance replication -- the code is present under the MPL in OSS, but won't work and may not be necessary unless upstream's Enterprise clustering semantics are followed exactly)...

What we'll do with those stub features, I don't know.

naphelps (approved):

PR #68, opened 2024-01-27
DRAFT: Buildable & Testable Base

cipherboy:
Based on the proposal in #64 and previous code in #56, this builds towards passing test cases post-fork. 

Note that this is rather large and includes commits currently in `development` due to the inability to rebase `git subtree` (likely a user problem).

This PR and #56 will be split up into separate PRs, one per plugin, plus various miscellaneous small changes on top. 

If #64 is changed significantly enough, this PR may need to be reworked. However, if the external->internal merge is not viewed favorably, this probably can be elided and not impact the results too much.


cipherboy:
This has now been split properly into Parts 1...17 as linked above, and thus can be closed.

Issue #37, opened 2024-01-06
Transit - drop support for pre-v0.6.2 keys


cipherboy:
In https://github.com/openbao/openbao/pull/36#discussion_r1443820358, it was mentioned again that Transit <v0.6.2 convergent encryption keys are weak (and the v2 convergent encryption scheme was as well, though less so). We should consider fully removing support for these operations so we can avoid a repeat of [the vulnerability that occurred](https://discuss.hashicorp.com/t/hcsec-2023-28-vault-s-transit-secrets-engine-allowed-nonce-specified-without-convergent-encryption/58249) as a result of this old code paths being supported.

This probably warrants a bigger discussion about deprecation paths. This version is very old, upstream has not supported it for years, but yet the one-shot upgrade path (and necessity to decrypt existing data and prevent workflow breakage) has caused support to remain.

PR #85, opened 2024-02-01
Forbid v1 & v2 convergent keys in Transit

cipherboy:
This prohibits the usage of the nonce field entirely, removing it as an API option from Transit. Similarly in the keysutil backend, v1 & v2 keys are forbidden.

Because this completely removes nonce from the APIs, this results in OpenBao issuing a warning about the unknown parameter via the CLI, but does not fail requests. A nil value for nonce will be passed into the SDK, which will then generate a fresh value (for encryption) or use the value inferred from the ciphertext (for decryption).

---

Resolves: #37

### DIFF
--- a/builtin/logical/transit/backend_test.go
+++ b/builtin/logical/transit/backend_test.go
@@ -1026,11 +1026,6 @@ func testDerivedKeyUpgrade(t *testing.T, keyType keysutil.KeyType) {
 }
 
 func TestConvergentEncryption(t *testing.T) {
-	testConvergentEncryptionCommon(t, 0, keysutil.KeyType_AES256_GCM96)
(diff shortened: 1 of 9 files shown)


JanMa (approved):
Lgtm 👍

Gabrielopesantos (approved):
Looks good. Curious, what makes it v1 and v2 convergent keys aren't safe but v3 is?

cipherboy:
> Looks good. Curious, what makes it v1 and v2 convergent keys aren't safe but v3 is? The fact that `nonce`s aren't under user control?

Correct. Check this out for details: https://cryptopals.com/sets/8/challenges/63.txt

naphelps (approved):

PR #102, opened 2024-02-02
Part 4/n - Remove secrets plugins

cipherboy:
In alignment with the plugin policy decisions of #64, this removes all non-OSI secrets plugins, preferring to make them external plugins.

To ease maintenance burden, only integrations licensed under OSI licenses will be included in the main repository; other plugins will be community supported and not coupled with the binary or the release process.

---

This is part of #68, broken up to make review easier.


cipherboy:
@naphelps Updated :-)

naphelps (approved):

PR #103, opened 2024-02-02
Part 5/n - Add jwt plugin

cipherboy:
This imports the JWT plugin as discussed in #64, bringing it in-tree with history.

---

This is part of #68, broken up to make review easier.

@naphelps When it comes time for merge, I'd suggest making this one a rebase merge if you can to preserve history. Thanks!


cipherboy:
@naphelps Updated :-)

naphelps (approved):

PR #110, opened 2024-02-03
Part 12/n - Remove non-OSI go-kms-wrapping backends

cipherboy:
This mostly removes the non-OSI go-kms-wrapping backends from being configurable via on-disk configuration. However, note that one proprietary seal method remains, awskms, used in vault/core.go: understanding this more is required before removing it (as, it doesn't look configurable).

Cherry-pick of 26bf915b75c6883d2c6984b671138d690fd32e67.

---

Part of #68; broken up for reviewability.


JanMa (changes requested):
I don't think it's a good idea to drop this at the moment. All the plugins we drop can be compiled independently and still be used via the custom plugin mechanism. But the KMS keys you drop can't be used via custom plugins. So this is essentially a breaking change that forces almost everyone to change their seal mechanism since only Vault Transit seals are left.

If we have an alternative way for people who want to use, for example, an GCP seal to do so, I'm fine with dropping support from the core binary

cipherboy:
I think its fine to drop this one and not merge it.

PR #117, opened 2024-02-03
Use formal volume bind mount specification for website build

cipherboy:
With rootless podman, `chown=true` and `relabel=shared` is required to allow these mounts to succeed. Without `chown=true`, the container lacks permissions to read or stat the given files, and without `relabel=shared`, SELinux will not allow watching the directories for changes.

If someone could verify this works on regular Docker prior to merging, I'd appreciate it. :-)

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -22,12 +22,11 @@ DOCKER_RUN_FLAGS=-it \
 		--publish "3000:3000" \
 		--rm \
 		--tty \
-		--volume "$(PWD)/content:/app/content" \
-		--volume "$(PWD)/public:/app/public" \
-		--volume "$(PWD)/data:/app/data" \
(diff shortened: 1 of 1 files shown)


website/Makefile
```diff
@@ -22,12 +22,11 @@ DOCKER_RUN_FLAGS=-it \
 		--publish "3000:3000" \
 		--rm \
 		--tty \
-		--volume "$(PWD)/content:/app/content" \
-		--volume "$(PWD)/public:/app/public" \
-		--volume "$(PWD)/data:/app/data" \
-		--volume "$(PWD)/redirects.js:/app/redirects.js" \
-		--volume "next-dir:/app/website-preview/.next" \
```
cipherboy, line 29: This directory doesn't exist for me.

cipherboy:
This is no longer necessary after #156. Thanks @JanMa!

PR #119, opened 2024-02-03
Remove patch for SHA-1 certificates

cipherboy:
Vault < 1.12 supported a patch to allow certificates with SHA-1 hashes to continue to function after Go removed support in Go 1.18 (or, rather, gated it behind a [feature flag](https://go.dev/doc/godebug#go-118)).

As our version number remains unclear, we'll want to remove this patch in case we choose a lower version number for our releases.

Note: this will not affect users' abilities to manually override this according to Go's settings to do so; this only removes dead code that did so automatically on older versions.


JanMa (approved):

naphelps (approved):

cipherboy:
Thank you @naphelps!

Issue #120, opened 2024-02-05
Fork - Update Go version


cipherboy:
Updating the Go version in `/.go-version` should be done prior to release so that we use the latest Go toolchain. Looking at prior commits to that file should give an indication of how to do that.

PR #124, opened 2024-02-06
Move to Go 1.22.0

cipherboy:
This moves from Go 1.20 to 1.21 to follow the latest toolchain release.

Resolves: #120


Gabrielopesantos:
What about updating the go version in `go.mod`, is it not necessary?

cipherboy:
Nah, this is a good point, the comment in `go.mod` does suggest we should probably bump it. I'm going to rework this PR for 1.22 since that was just released as well.

Gabrielopesantos (approved):

cipherboy:
@naphelps I'm happy with this one based on more testing I've done locally. I've switched to it as my primary toolchain now.

PR #136, opened 2024-02-10
Transit: Fix bug where locks were not being freed on specific operations

Gabrielopesantos:
This PR includes the fix to a bug happening in certain transit operations, sign and verify, where locks weren't being freed as expected which lead to goroutine deadlocks (operations hitting their timeout).

The deferred call of `Unlock` has been included in the `if` block so that it is only called if `Lock` has been initially called.
EDIT: After doing some additional tests (on other files), it seems there are cases where makes difference between leaving the deferred `Unlock` call to be in the `if` block or outside. Are there any known cases where `Unlock` calls are expected even if `Lock` isn't called?

Resolves `https://github.com/hashicorp/vault/issues/25325`.

### DIFF
--- a/builtin/logical/transit/path_sign_verify.go
+++ b/builtin/logical/transit/path_sign_verify.go
@@ -366,10 +366,10 @@ func (b *backend) pathSignWrite(ctx context.Context, req *logical.Request, d *fr
 	}
 	if !b.System().CachingDisabled() {
 		p.Lock(false)
+		defer p.Unlock()
 	}
 
 	if !p.Type.SigningSupported() {
-		p.Unlock()
 		return logical.ErrorResponse(fmt.Sprintf("key type %v does not support signing", p.Type)), logical.ErrInvalidRequest
 	}
 
@@ -385,12 +385,10 @@ func (b *backend) pathSignWrite(ctx context.Context, req *logical.Request, d *fr
 	if batchInputRaw != nil {
 		err = mapstructure.Decode(batchInputRaw, &batchInputItems)
 		if err != nil {
-			p.Unlock()
 			return nil, fmt.Errorf("failed to parse batch input: %w", err)
 		}
 
 		if len(batchInputItems) == 0 {
-			p.Unlock()
 			return logical.ErrorResponse("missing batch input to process"), logical.ErrInvalidRequest
 		}
 	} else {
@@ -491,7 +489,6 @@ func (b *backend) pathSignWrite(ctx context.Context, req *logical.Request, d *fr
 		}
 	} else {
 		if response[0].Error != "" || response[0].err != nil {
-			p.Unlock()
 			if response[0].Error != "" {
 				return logical.ErrorResponse(response[0].Error), response[0].err
 			}
@@ -509,7 +506,6 @@ func (b *backend) pathSignWrite(ctx context.Context, req *logical.Request, d *fr
 		}
 	}
 
-	p.Unlock()
 	return resp, nil
 }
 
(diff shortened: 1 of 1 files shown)


cipherboy:
@Gabrielopesantos said:

> The deferred call of `Unlock` has been included in the `if` block so that it is only called if `Lock` has been initially called. 
>
> EDIT: After doing some additional tests (on other files), it seems there are cases where makes difference between leaving the deferred `Unlock` call to be in the `if` block or outside. Are there any known cases where `Unlock` calls are expected even if `Lock` isn't called?

Ahh yes, I remember now.

Carefully read the code within `keysutil.LockManager.GetPolicy(...)`: if we're _not_ using a cache, and we found a policy to return (`retP != nil`), we do not trigger the unlock. This means this non-cached policy holds a lock, and thus needs to be unlocked by someone at a later time.

Remember, the conditional `if !b.System().CachingDisabled() { ... }` really means `if b.System().CachingEnabled() { ... }`!

Combining the code, we thus get an order of operations like:

 - Someone hands us a policy via `GetPolicy(...)`...
 - If caching is enabled, the policy is currently unlocked, so _we_ need to lock it to prevent concurrent modification (but allow parallel use as it is a RW lock).
 - In either case _because policy fetch was successful_, we need to defer an unlock: either we acquired the lock ourselves (cache enabled), or, Get Policy acquired the lock.

The calling semantics are rather complicated, though, I imagine they might've been done for a reason...? I'd be curious if the commit history has anything to say about this.

--

I think this means, for this PR, like you see elsewhere, the `defer p.Unlock()` needs to be done either way, i.e., outside of the if statement? My 2c...?

builtin/logical/transit/path_sign_verify.go
```diff
@@ -366,10 +366,10 @@ func (b *backend) pathSignWrite(ctx context.Context, req *logical.Request, d *fr
 	}
 	if !b.System().CachingDisabled() {
 		p.Lock(false)
+		defer p.Unlock()
```
cipherboy, line 369: Aha! Yes, play around with the tests in my branch here: https://github.com/openbao/openbao/pull/136#issuecomment-1937357440

If you rebase them (`git rebase -i main`), you can edit them to be like so:

```
pick e37f70887 Defer policy Unlock call when Lock is called  ## < your commit
pick e97f53909 Add test cases for Transit sign/verify deadlock ## < test cases
break ## < break to let you run tests
pick 6769b473e Defer cleanup in keysutil.LockManager.GetPolicy() ## < unrelated cleanup in keysutil
pick 82fb4d715 Simplify unlock behavior in Transit with defers ## < moving all `p.Unlock()` to the consistent format
```

You can then run tests like so:

```bash
$ reset ; cd builtin/logical/transit && go test -count=1 -timeout=25s -run=TestTransit_NoDeadlock_SignVerify .
```

and see a failure if you search for this:

```
panic: test timed out after 25s
running tests:
	TestTransit_NoDeadlock_SignVerify/rsa-2048/direct (25s)
	TestTransit_NoDeadlock_SignVerify/rsa-3072/direct (25s)
	TestTransit_NoDeadlock_SignVerify/rsa-4096/direct (25s)
```

This shows that with caching disabled (not taking this if branch), we need to call unlock as well. :-)  One set of tests run using a cluster with caching enabled (default), the other runs with caching disabled. 

(you can then `--continue` the rebase to exit)

cipherboy: Feel free to grab what you want from that branch and add it to this PR :-) If you want to do the move to `defer`s yourself, happy to let you do that. But if you get a moment, perhaps expanding the deadlock cases to cover the other endpoints (verify, decrypt, encrypt, hmac, ...) might be good just to make sure we didn't miss anything... Now that I remember how to test it in short time. :-D

Gabrielopesantos: Yes, thanks for the explanation, in that case it completely makes sense to have the `defer`s outside of the `if` blocks.
As most of the work has already been done in your branch I'm ok with closing this one. As for the testing you are suggesting and missing `defer` in some Unlock calls I can commit those into your branch (if that's possible).

PR #137, opened 2024-02-11
Remove old version docs

JanMa:
This PR removes all documentation which mentions older Vault versions. It also cleans up no longer references `partials`.


website/content/docs/get-started/developer-qs.mdx
```diff
@@ -72,153 +46,7 @@ Now, let's add the import statements for the client library to the top of the fi
 <CodeBlockConfig heading="import statements for client library" lineNumbers>
 
 ```go
-import vault "github.com/openbao/openbao/api"
```
cipherboy, line 75: Curious, why doesn't this work? It might not work right now, but it should definitely work in general.

cipherboy: Ah I see, you left it at the end, its no longer `vault` :-D 

But in general, I think we do want to mention that we're still planning on being compatible with Vault client libraries.

website/content/partials/faq/client-count/computing-clients.mdx
```diff
@@ -1,43 +1,10 @@
...
-
-**Yes**.
-
-Use the [Vault auditor tool](/vault/tutorials/monitoring/usage-metrics#vault-auditor-tool)
-to compute and display client count data for Vault v1.3 &ndash; v1.5 using the
-client compute logic available in Vault 1.7. Auditor use with Vault versions
-older than 1.3 is untested.
-
-### Can i compute KMIP clients for Vault? ((#kmip))
+### Can i compute KMIP clients for Openbao? ((#kmip))
```
cipherboy, line 1: KMIP can be removed entirely since it is Enterprise only btw.

website/content/partials/ui-pki-control-groups-known-issue.mdx
```diff
@@ -1,17 +0,0 @@
-### Users limited by control groups can only access issuer detail from PKI overview page ((#ui-pki-control-groups))
-
-#### Affected versions
-
-- Vault 1.14.x
```
cipherboy, line 5: I do wonder if this still impacts us.

cipherboy (approved):
Wow, a lot of great changes here @JanMa! Thank you!!

naphelps (approved):

PR #138, opened 2024-02-12
fork: Rename Environment Variables

### LINKED ISSUE #66: Fork: Rename Environment Variables
A lot of environment variables and headers use `VAULT_` as a prefix: as we rename Vault->Bao, we should support both `VAULT_` and `BAO_` as prefixes, warning (once) on the former being accessed, though we probably want to prefer the latter for users running both Vault and Bao. 

This would be a good first issue for someone looking to see how the broader project's code base was put together.

Mab879:
This PR will allow users to configure OpenBAO with both `VAULT_` and `BAO_` prefixes for environment variables. 

Resolves #66

## Target Release
GA

### DIFF
--- a/api/env_test.go
+++ b/api/env_test.go
@@ -1,0 +1,44 @@
+package api
+
+import (
+	"os"
+	"testing"
+)
+
+func TestReadBaoVariable_Vault(t *testing.T) {
+	actual := "example_value"
+	os.Setenv("VAULT_TEST", actual)
+	expected := ReadBaoVariable("BAO_TEST")
+	if actual != expected {
+		t.Fatalf("bad: Failed to Read Enviroment Variable actual: %s expected: %s", actual, expected)
+	}
+}
+
+func TestReadBaoVariable_Bao(t *testing.T) {
+	actual := "example_value"
+	os.Setenv("BAO_TEST", actual)
+	expected := ReadBaoVariable("BAO_TEST")
+	if actual != expected {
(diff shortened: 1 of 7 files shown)


sdk/helper/envutil/envutil.go
```diff
@@ -0,0 +1,19 @@
+package envutil
```
cipherboy, line 1: So I think I gave you bad advice. Because `api/` uses environment variables and `sdk/` imports `api/`, we've gotta move this.

cipherboy (approved):
Looks great, thanks @Mab879!

api/env_test.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"os"
+	"testing"
+)
+
+func TestReadBaoVariable_Vault(t *testing.T) {
+	actual := "example_value"
+	os.Setenv("VAULT_TEST", actual)
+	expected := ReadBaoVariable("BAO_TEST")
+	if actual != expected {
```
cipherboy, line 12: For the future, the project is really inconsistent about using it, but there's a library [called `require`](https://pkg.go.dev/github.com/stretchr/testify/require) that lets you do one-line assertions common in other languages. Three lines here isn't bad, so I'm happy without it but if you find yourself with more unit tests, don't feel you need to write a bunch of if statements. :-)

naphelps (approved):

PR #141, opened 2024-02-14
Cleanup more obsolete docs

JanMa:
This removes some non-working code examples in the docs as well as references to third-party Vault tooling


website/content/api-docs/libraries.mdx
```diff
@@ -26,186 +26,3 @@ These libraries are officially maintained by HashiCorp.
 $ go get github.com/openbao/openbao/api
 ```
```
cipherboy, line 28: I think this goes to my comment on #137 -- we want a message _somewhere_ that states we still intend to remain compatible with the broader ecosystem of things that integrate with HashiCorp Vualt.

JanMa: I'll add a message linking back to the relevant upstream docs 👍

naphelps (approved):

PR #142, opened 2024-02-14
Fork: Remove 'Enterprise' and related references from user-visible codebase

### LINKED ISSUE #121: Fork - Remove references to Enterprise from user-visible code base
Searching for `Enterprise` in the Go code base makes numerous references. We should probably audit and trim these down as much as possible, especially where e.g., they are returned in user-facing strings.

This likely means removing references to managed keys presently (from Transit and PKI) as that code is not present in OpenBao. 

#126 applied to the CLI, but I think we're probably better off keeping it, IMO.

Gabrielopesantos:
Resolves #121 

## Description

Removes references to upstream Enterprise features;
- User visible references in CLI, API;
- Transit and PKI mentions to managed keys;
- Enterprise features such as licensing, entropy configuration;

## Target Release

Alpha


command/server/config.go
```diff
@@ -745,7 +745,7 @@ func ParseConfig(d, source string) (*Config, error) {
 	}
 
 	if err := result.parseConfig(list); err != nil {
-		return nil, fmt.Errorf("error parsing enterprise config: %w", err)
```
cipherboy, line 748: I think both `command/server/config_util.go` and `internalshared/configutil/config_util.go` could be removed. :-) Will cause some build failures that you'll need to address.

Gabrielopesantos: Only saw this comment today. Didn't think anyone would look into the PR while in draft.

Gabrielopesantos: Do you think makes sense to also try to exclude all license related code in this PR?

Gabrielopesantos: Do you think makes sense to also try to exclude all license related code in this PR?

vault/core.go
```diff
@@ -156,28 +156,6 @@ var (
...
-	PerformanceMerkleRoot        = merkleRootImpl
-	DRMerkleRoot                 = merkleRootImpl
-	LastRemoteWAL                = lastRemoteWALImpl
-	LastRemoteUpstreamWAL        = lastRemoteUpstreamWALImpl
-	WaitUntilWALShipped          = waitUntilWALShippedImpl
-	storedLicenseCheck           = func(c *Core, conf *CoreConfig) error { return nil }
-	LicenseAutoloaded            = func(*Core) bool { return false }
-	LicenseInitCheck             = func(*Core) error { return nil }
-	LicenseSummary               = func(*Core) (*LicenseState, error) { return nil, nil }
-	LicenseReload                = func(*Core) error { return nil }
```
Gabrielopesantos, line 180: There are still some references to these functions that I did not consider.

command/server/config.go
Gabrielopesantos: UPDATE: Looks like the build is now failing after removing the licensing references. Still not sure why but would like to understand. :thinking:

cipherboy: Yeah, I think removing the licensing references is fair game. :-) I expected that to break the build... It looks like you might just have a few more references to licensing running around still. :-)

Gabrielopesantos: Yes, there we still some licensing references around but I believe to have removed them all now. The problem seems to be in the `core` `systemBarrierView` which is nil.

vault/core.go
Gabrielopesantos: Ended up keeping the WAL related functions even though they are an enterprise feature.

Gabrielopesantos:
There seem to be a few tests broken, under builtin/logical/database and /vault/external_tests/raft but I don't think they are related to these changes.

command/operator_diagnose.go
```diff
@@ -588,17 +589,7 @@ SEALFAIL:
 	licenseCtx, licenseSpan := diagnose.StartSpan(ctx, "Check For Autoloaded License")
 	// If we are not in enterprise, return from the check
 	if !constants.IsEnterprise {
```
cipherboy, line 591: I like the idea of leaving `constants.IsEnterprise` alone for now, but we might think about removing it later...

command/operator_diagnose.go
```diff
@@ -14,6 +14,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/openbao/openbao/helper/constants"
```
cipherboy, line 17: I'd run `make fmt` if you don't mind!

cipherboy: Oh interesting, I guess you did run `make fmt` but that didn't reorganize the inputs. Huh, TIL!

command/operator_diagnose.go
cipherboy: Mostly its used in PKI and in Core, which isn't all that surprising. I think perhaps I'd suggest we remove it now.

cipherboy (approved):
Broadly I think this looks fairly good and am happy to see this merged, with or without the `constants.IsEnterprise` removal! 

Long term, I think we should think about removing all the stubs from files that are marked `//go:build !enterprise`: we'll need to remove a lot of call sites, but these should be unnecessary to keep given we don't have Vault Enterprise code.

Gabrielopesantos:
> Broadly I think this looks fairly good and am happy to see this merged, with or without the `constants.IsEnterprise` removal!
> 
> Long term, I think we should think about removing all the stubs from files that are marked `//go:build !enterprise`: we'll need to remove a lot of call sites, but these should be unnecessary to keep given we don't have Vault Enterprise code.

Just committed the removal of all "enterprise" stubs in the files. With your previous comment did you mean just that or the files that had that stub and all code references/calls to that code?

cipherboy:
> With your previous comment did you mean just that or the files that had that stub and all code references/calls to that code?

Ah, I was thinking that you'd remove the functions themselves, and maybe the files too, to that where you can... E.g., `builtin/logical/pki/managed_key_util.go` only has no-op functions, so if you want to make the more invasive change, you can remove that file, and then remove all the function calls throughout the code base to those functions.

Its more work, but reduces our code size a fair bit :-)

That's why I thought that perhaps a separate PR (maybe logically grouped -- one for managed keys, one for clustering hooks, &c) could be better. :-)

Gabrielopesantos:
> > With your previous comment did you mean just that or the files that had that stub and all code references/calls to that code?
> 
> Ah, I was thinking that you'd remove the functions themselves, and maybe the files too, to that where you can... E.g., `builtin/logical/pki/managed_key_util.go` only has no-op functions, so if you want to make the more invasive change, you can remove that file, and then remove all the function calls throughout the code base to those functions.
> 
> Its more work, but reduces our code size a fair bit :-)
> 
> That's why I thought that perhaps a separate PR (maybe logically grouped -- one for managed keys, one for clustering hooks, &c) could be better. :-)

Sure, I'm ok with leaving that for a separate PR(s) as you are suggesting.

cipherboy:
@Gabrielopesantos No, I think this is fine as-is, just need to remember which files you need to trim up later.

cipherboy (approved):
Looks very nice @Gabrielopesantos!

naphelps (approved):

naphelps (changes requested):
@Gabrielopesantos Your commits need to be signed before I can merge your pull request.

Gabrielopesantos:
Hey @naphelps, do you mean GPG-signed? If it's just the "sign-off" tag they are already signed and the DCO check is passing.

naphelps (approved):

cipherboy (changes requested):
Hmm, wonder if something got messed up in a rebase:

```
--- FAIL: TestUnknownFieldValidation (0.00s)
    config_test_helpers.go:453: found unexpected error: unknown or unsupported field license_path found in configuration at ./test-fixtures/config.hcl:54:1
FAIL
FAIL	github.com/openbao/openbao/command/server	15.330s
```

Gabrielopesantos:
> Hmm, wonder if something got messed up in a rebase:
> 
> ```
> # github.com/openbao/openbao/command [github.com/openbao/openbao/command.test]
> command/operator_diagnose.go:29:2: could not import github.com/openbao/openbao/helper/constants (open : no such file or directory)
> command/operator_diagnose.go:598:15: coreConfig.LicensePath undefined (type "github.com/openbao/openbao/vault".CoreConfig has no field or method LicensePath)
> command/operator_diagnose.go:601:15: coreConfig.License undefined (type "github.com/openbao/openbao/vault".CoreConfig has no field or method License)
> command/server.go:436:29: config.Entropy undefined (type *server.Config has no field or method Entropy)
> command/server.go:438:10: config.Entropy undefined (type *server.Config has no field or method Entropy)
> command/server.go:1138:10: config.LicensePath undefined (type *server.Config has no field or method LicensePath)
> command/server.go:1141:10: config.License undefined (type *server.Config has no field or method License)
> command/server.go:1656:19: undefined: vault.LicenseReload
> command/server.go:2751:3: unknown field License in struct literal of type "github.com/openbao/openbao/vault".CoreConfig
> command/server.go:2751:42: config.License undefined (type *server.Config has no field or method License)
> command/server.go:2751:3: too many errors
> ?   	github.com/openbao/openbao/builtin/plugin/v5	[no test files]
> FAIL	github.com/openbao/openbao/command [build failed]
> 
> --- FAIL: TestUnknownFieldValidation (0.00s)
>     config_test_helpers.go:453: found unexpected error: unknown or unsupported field license_path found in configuration at ./test-fixtures/config.hcl:54:1
> FAIL
> FAIL	github.com/openbao/openbao/command/server	15.330s
> ```

This error was fairly simple to fix but it seems I broke some additional tests. Having a look.

cipherboy (approved):
This looks much better, thanks @Gabrielopesantos ! Tests pass again :-D

cipherboy:
@Gabrielopesantos said:

> This error was fairly simple to fix but it seems I broke some additional tests. Having a look.
> UPDATE: I was wrong, it was the tests were timing out due to the value I had set in the `timeout` flag.

Yeah, if you look at the `Makefile`, it sets a really long (45min?) timeout... Certain packages take nearly forever to run.

PR #167, opened 2024-03-02
Remove Consul service registration

cipherboy:
This removes the non-BUSL Consul service registration; the only remaining service registration is thus Kubernetes.


JanMa (changes requested):
@cipherboy please also remove the relevant documentation if we're going to drop the feature

cipherboy:
@JanMa ah yes, good reminder!

JanMa (approved):

naphelps (approved):

Issue #152, opened 2024-02-21
Fork - GitHub Actions CI Pipelines


cipherboy:
Many CI pipelines were forked; some of these may not be relevant (if they mention Enterprise), others of these just aren't showing up at all (e.g., Go test pipelines), and yet others aren't passing (e.g., format checks or semgrep).

This is a requirement to fix before Alpha release.

grzegorz-sz:
Please, assign to me - as discussed

Issue #42, opened 2024-01-10
Add Support for riscv64


liberodark:
Hi,

I know it's going to take you a while before you make a release.
But I would like you to consider releasing support for riscv64.
Also I propose to provide riscv64 servers if necessary.

Best Regards

cipherboy:
It looks like Go toolchain already supports riscv64; it'd be interesting to see what fails if you attempt a build @liberodark. I don't presently have access to riscv hardware ATM, though could try playing with emulation sometime.

liberodark:
Hi, if you want i can give you access on riscv64 hardware for free.
Also have build apps in go on riscv64 hardware with success.
My Goal is to open cluster riscv64 server for developpers want to access on riscv64 hardware and that for free.

liberodark:
My build : 

```
apt install -y golang-go rsync gpg git nodejs npm
git clone https://github.com/openbao/openbao
cd openbao
git checkout development
go build
```

cipherboy:
@liberodark This looks good :) I think then, when we get more mature and have CI up and running, we'll want to sync on getting access to these RISC-V build machines and setting up a CI/testing pass on them, but in the mean time, I think we can call this good to start!

liberodark:
Im adding link of website if you want access to RISC-V hardware : https://open-riscv-initiative.com/

cipherboy:
@liberodark I finally got around to fixing the build pipelines, #168 should add RISC-V support for you :-)

Issue #31, opened 2023-12-22
Changelog process - enforce GH workflow


cipherboy:
Currently the changelog process uses [`go-changelog`](https://github.com/hashicorp/go-changelog) for generation. 

It seems like some changelog entries were removed (which IMO, is good as we don't want to maintain it retroactively for prior upstream versions) -- but we likely [want to remove the rest](https://github.com/openbao/openbao/tree/development/changelog) which snuck in. 

In particular, changelog entries are numbered off of PR numbers (and the [README](https://github.com/openbao/openbao/tree/development/changelog#changelog) is generally helpful), so we'll have a ways to go before they'll conflict :-) 

There should be a label ([`pr/no-changelog`](https://github.com/openbao/openbao/blob/development/.github/workflows/changelog-checker.yml)) which avoids the checker in the GH action; it might be nice to make this a required workflow. 

Are there any thoughts on the existing process (doc'd in that README)? Should we choose something else or is this generally amenable to everyone?

joewxboy:
Thanks, @cipherboy .  I need to look at this.

cipherboy:
#70 should remove the unnecessary changelog entries.

Issue #153, opened 2024-02-21
Fork - Release Pipeline


cipherboy:
GitHub Actions should also be used to build release (and perhaps nightly or per-commit) artifacts. Ideally this would be a purely automated workflow (involving some manual trigger in the form of pushing a formatted tag) and end up with a release created on the page, containing an updated changelog.

grzegorz-sz:
Please, assign to - as discussed

cipherboy:
@grzegorz-sz See #168 for where CI pipelines are currently -- I think `scripts/dist.sh` will need to change and we'll need to remove dependence on `hashicorp/actions-packaging-linux` perhaps.

PR #168, opened 2024-03-02
Fix CI Pipelines, remove references to HashiCorp Infrastructure

cipherboy:
This fixes many pipelines and removes others that aren't necessary any more. In doing so, we had to:

 - Fix the go modules to refer to the correct API versions,
 - Move kubernetes integration tests into the main package,
 - Remove all references to HashiCorp runners,
 - Remove references to internal HashiCorp credential stores,
 - Remove many references to Vault Enterprise,
 - and then remove unnecessary actions.

I also added RISC-V 64-bit support while I was here. :-) 

This pipeline can be seen in action on my fork: https://github.com/cipherboy/openbao/pull/1 --> https://github.com/cipherboy/openbao/pull/1/checks

TODO:

 - [x] ~Check that builds work~ --> ~Builds do not work as they rely on `dist/vault` and `hashicorp/actions-packaging-linux` -- as part of #153 we should replace those.~ I've fixed the build pipelines as part of this. Releasing is a different story; these builds aren't signed (for obvious reasons); I'm not sure if the Linux Foundation handles this for us or not.
 - [x] Ensure that CI executes for both Go and UI components.

Resolves: #31
Resolves: #42
Resolves: #152 
Related: #153


cipherboy:
@naphelps This one should probably be squash merged. :-)

cipherboy:
@naphelps When you merge this, I think you can configure the branch protection rules on the repo to require that the `format` job of `Run linters` (defined in `workflows/code-checker.yml`) passes, resolving https://github.com/openbao/openbao/issues/29.

Gabrielopesantos:
Are we going to continue with a single file per change approach instead of a single CHANGELOG file?

cipherboy:
@Gabrielopesantos If I recall correctly, GH actions doesn't provision a privileged token for pull requests from forks... Wonder how other projects handle this?

Upstream mostly had contributions from inside the repo, so it wasn't as much of a problem... Hmmm...

Gabrielopesantos:
> @Gabrielopesantos If I recall correctly, GH actions doesn't provision a privileged token for pull requests from forks... Wonder how other projects handle this?
> 
> Upstream mostly had contributions from inside the repo, so it wasn't as much of a problem... Hmmm...
> 
> And to answer your question, yes, I think we're planning on using the changelog builder still.

Makes sense, I didn't consider that. Something like [this](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) might be what we are looking for. As far as I understood, it is possible to set workflows to only run once approved by someone with write-permission.

cipherboy:
@Gabrielopesantos said:
> Makes sense, I didn't consider that. Something like [this](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) might be what we are looking for. As far as I understood, it is possible to set workflows to only run once approved by someone with write permission.

I think this page: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

states that the token for pull requests should have write permission. But its not clear to me why it failed:

```
Run ./.github/scripts/report_failed_builds.sh
gh: Resource not accessible by integration (HTTP 403)
{"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/issues/comments#create-an-issue-comment"}
Error: Process completed with exit code 1.
```

Unless perhaps `REPO` is not set correctly? but this is set like:

```
build.yml:238:          REPO: ${{ github.event.repository.name }}
```

so I'm still suspicious of write permissions, IMO.

naphelps (approved):

Issue #140, opened 2024-02-12
Support Paginated Lists in Storage, APIs


cipherboy:
### Summary

OpenBao's `LIST` Plugin APIs presently only supports returning all keys within a tree, resulting in many performance issues (memory and compute). By implementing pagination, the caller can limit the number of responses and thus reduce both the client's and server's instantaneous resource consumption.

### Problem Statement

`LIST` APIs in plugins historically have been forced to be authenticated, to mitigate DoS likelihood. This is because many secrets engines (and other areas of OpenBao) have unbounded creation, and thus, unbounded lists. For example, [K/V allows](https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#create-update-secret) an arbitrary number of keys within a path, PKI places all [(valid or revoked) certificates](https://developer.hashicorp.com/vault/api-docs/secret/pki#list-certificates) in a single path by serial number, and Transit allows [arbitrary key creation](https://developer.hashicorp.com/vault/api-docs/secret/transit#create-key) (again, placed within a single path in storage). Browsing these potentially large collections (either via UI or an API client) is unwieldy and causes resource consumption on both the client and server. This often makes it hard for users to find what they're looking for.

Furthermore, certain operations (like PKI's tidy operation) requires `List(...)`ing the entire collection in order to prune them; this requires the collection of items fits in memory of the server. However, this operation can be interrupted, whether through explicit pauses and releasing and re-acquiring locks. Keeping this entire collection in memory thus reduces the amount of resources for regular operations to continue. In extreme cases, thousands of mounts with 100k-1m leaf certificates per mount have been observed; this leads to around 100MB+ of memory just for the list, per mount, potentially requiring 100GB+ of memory for all mounts to be able to tidy in parallel.  

These limitations are discussed in various upstream issues, e.g., `https://github.com/hashicorp/vault/issues/21164` or `https://github.com/hashicorp/vault/issues/21041`.

### User-facing description

Efficient pagination allows incremental discovery over large data sets. For instance, the web UI will load the first subset of data much more quickly and subsequent navigation to later pages will happen much more quickly. Similarly, periodic scripts that need to iterate over a large quantity of data can do so more efficiently than requesting an entire list up front.

However, note that this pagination is not intended to be bound to a transaction: entries may not be present if they were created after a given page range was iterated over.  

### Examples

Given a storage prefix with keys like:

 - `abcd`
 - `efgh`
 - `ijkl`
 - `mnop`
 - `qrst`
[...]

PR #170, opened 2024-03-03
Implement paginated lists in CLI, APIs, and storage

cipherboy:
This implements list paginations using an `after`/`limit` style system, discussed in #140.

In particular, `LIST` API operations are `GET` requests with a `?list=true` parameter to an endpoint. This adds the ability (in the CLI, API, and underlying storage mechanism) to specify two additional attributes, a key, `after`, to show results whose key is strictly greater than (alphabetically) the specified value, and a number, `limit`, of results to show.

This allows better user interface design: pagination can be done on the server (in addition to the UI), which allows the client to consume less resources (both locally and on the remote server), when lots of entries are present. Endpoints like PKI's revoked or valid certificates, Transit's lists of keys, or entries in K/V v2 are all common endpoints upgraded with pagination that benefit from this improvement.

This is an example of a change made possible by dropping support for alternative storage backends. 

Resolves: #140


schultz-is:
This is great! I can attest to this being a highly desired feature. Notably, KV and PKI backend deployments can get large fast, so the ability to page over all data within them is super handy. The fact that this approach is also backwards compatible (doesn't add any output parameters like cursor paging,) is a nice bonus; that means nobody has to update any existing scripts. Passing the current `List` operations through the `ListPage` operation will also net free test coverage.

website/content/docs/internals/telemetry/metrics/all.mdx
JanMa, line 1: Why are you removing these partials here?

builtin/credential/approle/path_role.go
```diff
@@ -308,6 +308,18 @@ can only be set during role creation and once set, it can't be reset later.`,
 				OperationPrefix: operationPrefixAppRole,
 				OperationSuffix: "roles",
 			},
+			Fields: map[string]*framework.FieldSchema{
+				"after": {
+					Type: framework.TypeString,
+					Description: `Optional entry to list begin listing after,
+not required to exist.`,
```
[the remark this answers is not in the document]
cipherboy: Hmm, thanks. I've updated them, but do note they're inconsistent between plugins (whether or not line wrapping is used). The path-help itself removes them:

https://github.com/openbao/openbao/blob/2c27dde78465c3a633976350b13a8520c27fc657/sdk/framework/path.go#L348

so it ultimately doesn't matter and is mostly a matter of style, if we want a fixed-ish width or want to rely on editor's wrapping :-)

website/content/docs/internals/telemetry/metrics/all.mdx
cipherboy: Partials were actually removed in #194, but it turns out I didn't remove the references to them :facepalm: This time, when adding the RFC, I apparently _actually_ did the `cd website && make`  I thought I did in that PR, but didn't. :-D 

So this is just to fix the website build, as noted in the PR.

JanMa: I see, thanks :-)

JanMa (approved):

Issue #174, opened 2024-03-04
Add Policy To Contributing For Source-Available Licensed Sources


naphelps:
Per feedback from the LF Edge subgroup reviewing OpenBao's proposal, policy/guidelines need to be added to the project's contributors file to warn new contributing not to make use of source-available licensed works with specific mentions to Hashicorp.

PR #175, opened 2024-03-04
Issue 174: Adding source-available licensing policy to contributing

naphelps:
References: #174

Changes:
- Addition of a source-available license policy to contributing.
- Moved DCO section higher in contributing. Added additional language to the section.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,46 @@ talk to us! You can find out our thoughts and ensure that your contribution
 won't clash or be obviated by OpenBao's normal direction. A great way to do this
 is via the [GitHub Discussions][2].
 
+## Source-Available License Policy
+
+This policy prohibits the utilization or copy from any source that is protected 
+by a source-available license agreement including Business Source License 1.1. 
+This specifically includes all source repository members under the Hashicorp GitHub 
+organization. It is the intention of this policy that when developing for feature 
+parity, bug fixes, CVEs, etc. in comparison to Hashicorp Vault that a clean room 
+methodology be used and all deliverables made original.
+
+## Developer Certificate of Origin (DCO) Sign Off Policy
+
+OpenBao adheres to the Linux Foundation's [DCO](https://developercertificate.org/).
+All contributors are required to confirm their legal rights to the source they 
+contribute. This is done through a "DCO Sign Off". Note that this is different from 
+commit signing, such as using `PGP` or `gitsign`.
+
+To add your sign-off, include the --signoff option in your `git commit` or `git rebase` 
+commands:
+
+```
+# Sign off a commit
+git commit --signoff -m"my commit"
+
+# Add a signoff to the last commit you made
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -15,6 +15,46 @@ talk to us! You can find out our thoughts and ensure that your contribution
 won't clash or be obviated by OpenBao's normal direction. A great way to do this
 is via the [GitHub Discussions][2].
 
+## Source-Available License Policy
+
+This policy prohibits the utilization or copy from any source that is protected 
+by a source-available license agreement including Business Source License 1.1. 
+This specifically includes all source repository members under the Hashicorp GitHub
```
cipherboy, line 22: Including or excluding the MPL ones? We still depend on a number of HashiCorp libraries; OpenTofu has opted not to fork these MPL licensed projects until upstream is unwilling to take changes, and IMO, it seems like a sane course of action.

naphelps: Just the BSL ones, MPL is not source-available its open-source. Though its easier to say for the common person to just not use the Hashicorp source. This tries to kind of cut both ways. To cover source-available licenses, but to also acknowledge OpenBao's upstream.

naphelps: It is also fitting that the DCO section immediately follow this, as both sections are complementary.

cipherboy: Ah, I think you might've dropped the "available" from this sentence? I see source repositories, which I take to be all code repos?

naphelps: I was using source to mean more than just code it could be documentation, media, images, etc. It might be better just to drop the word "source" from that sentence specifically for clarity to refer to all repos generally.

cipherboy: Hmmm, I think it'd still read clearer like:

```suggestion
This specifically includes all source-available repositories under the HashiCorp GitHub 
```

as otherwise, a statement like:

> This specifically includes **all** repositories under the HashiCorp GitHub organization.

reads like excluding the MPL-licensed repositories (raft, ... &c) as well (I think, for me, its the _all_ that does it). Which, isn't what you're trying to say :-) 


My 2c.

naphelps: I disagree, its more that we are explicitly including _all_ Hashicorp repos in the policy which is the point and LF point. This policy would then apply at scale large/small depending on how a BUSL license was deployed in a specific repo, while also covering all repos should they change broadly over time.

naphelps: The MPL licensed repos like raft would be fine to consume under this policy. It would only warn the contributor to stop and check licenses first.

naphelps: That language comes from LF, hands are tied on that one.

naphelps: Outside of additional maintainership burdens its probably in our combined best interests to do so long term.

cipherboy (approved):

Issue #149, opened 2024-02-20
docs: recreate images and diagrams


JanMa:
Many images and diagrams in the docs show the Vault logo and we need to recreate those with OpenBao branding.
I am currently unsure if we should recreate/edit the images or if it might be even better to replace them with [Mermaid Diagrams](https://mermaid.js.org/) which are natively supported in the Markdown engine we use to render the docs.

joewxboy:
I'm a big fan of diagrams as code, but state of the art a year ago had limitations on spacing and unpredictable layout. We should probably pick a few samples and see if Mermaid can achieve the spirit of the desired output. Otherwise I'd recommend draw.io, and also version the draw.io sources.

David-Langan:
@naphelps - making a comment here so that you can assign this to me.

Issue #40, opened 2024-01-09
Fork projects for first class Kubernetes support


JanMa:
Vault has first class Kubernetes support by offering a HELM chart, CSI driver and Agent injector. If OpenBao should also make Vault on Kubernetes a first class citizen, then we need to fork the relevant upstream projects. Otherwise we should remove the relevant Kubernetes docs from the OpenBao documentation.

djds:
Also, https://github.com/hashicorp/vault-secrets-operator is pretty cool but switched to BSL with https://github.com/hashicorp/vault-secrets-operator/commit/cc1b024fc3496e07253ad2d3f039d8d87ffec279

yuval9313:
I rely on vault secrets operator and would like to see an open source version of it as well, I wish that first-clas k8s support would be maintained

Thank you for all the job you have made this far :)

jessebot:
> Also, https://github.com/hashicorp/vault-secrets-operator is pretty cool but switched to BSL with [hashicorp/vault-secrets-operator@cc1b024](https://github.com/hashicorp/vault-secrets-operator/commit/cc1b024fc3496e07253ad2d3f039d8d87ffec279)

I think this means that the fork could start from here though right? (the commit before the license change):
https://github.com/hashicorp/vault-secrets-operator/tree/06bc1738beb805455cc875f1b41e96ec35ecb514

If I'm not wrong, this is the helm chart: https://github.com/hashicorp/vault-helm
The helm chart has a "Mozilla Public License 2.0" license. Can that be forked safely?

JanMa:
Hey @jessebot you're right regarding both points. The fork can start at the last commit that is still MPL licensed, and those projects are in general save to fork.

jessebot:
@JanMa awesome, thanks for the quick response! If those repos are forked under the openbao org, I would be happy to pinged for help on the k8s side of things. I maintain a lot of helm charts, so if there's something specific to those, I can help :)

JanMa:
That's great, thanks a lot @jessebot. I'll let you know once we have those repos available in our org

AdrianAbraham:
Please also include support for https://github.com/hashicorp/vault-k8s . Our projects rely on the older "Agent Injector" instead of the newer "Vault Secrets Operator" or "Vault CSI Provider."

JanMa:
@jessebot @AdrianAbraham we have now forked the relevant repos into our org:
- https://github.com/openbao/openbao-secrets-operator
[...]

PR #180, opened 2024-03-06
Update images used in website documentation

David-Langan:
As part of issue [149](https://github.com/openbao/openbao/issues/149), update the images in the website documentation to reference OpenBao instead of Vault in their text/logos.


website/public/img/OB-comparison-table.png
cipherboy, line 1: I'd probably convert this to a table.

Some commentary here is that we haven't forked this yet, but would if there's community interest, see #40. 

The other is references to Consul storage backend should be removed, see #64 and #99. (raft is sometimes called "Integrated storage" below).

naphelps:
@David-Langan Do not forget to add the `-s` or `--signoff` flag to your git commits. This needs to be done for all of your commits.

cipherboy:
Thanks for the PR @David-Langan! @JanMa I think has addressed this everywhere in #397 so I'm going to go ahead and close this.

PR #181, opened 2024-03-06
Add Shamir's to shared sdk/, fix titles in API & SDK READMEs

cipherboy:
This moves Shamir's to `sdk/`, as it is a generally useful package for others. Additionally, we fix the titles of the API & SDK while we're here.


JanMa (approved):

Gabrielopesantos (approved):
Just a question out of curiosity, if Shamir is usually useful, why doesn't Upstream also have it under  `/sdk`

cipherboy:
@Gabrielopesantos said:
> Just a question out of curiosity, if Shamir is usually useful, why doesn't upstream also have it under `/sdk`?

Honestly, I don't recall! I recall it being discussed, and [is used by external parties](https://github.com/incipher/shamir). Though, it doesn't appear on my [favorite list](https://github.com/dsprenkels/sss)... I suspect people (like the first) have pinned an old version of the library that could be imported and haven't upgraded since. My 2c.?

naphelps (approved):

Issue #87, opened 2024-02-01
Fork - Review dependencies for CVEs, update


cipherboy:
Once #64 has been decided on, the remaining dependencies of the core binary should be evaluated for vulnerabilities and supportability, and updated if necessary. 

Some examples are:

 - `https://github.com/hashicorp/vault/issues/25131`
 - `https://github.com/hashicorp/vault/issues/23912`

And perhaps others.

PR #183, opened 2024-03-06
Fix vulnerable dependencies in core, api/, and sdk/

cipherboy:
This uses `govulncheck` to fix all vulnerabilities in all dependent modules, which requires us to update to Go 1.22.1 to fix vulnerabilities in `net/http` and `crypto/x509` among others.

We introduce a new, informational CI pipeline stage to warn us about issues in the future.

Resolves: #87


cipherboy:
@naphelps The last [job passed](https://github.com/openbao/openbao/actions/runs/8189662304/job/22394887802?pr=183) but I realized it had a copy/paste error (`"Check deprecation"` as the step description), so I rebased it locally & updated. Sorry!

naphelps (approved):

PR #185, opened 2024-03-07
Fix OpenBao OIDC followed by Google OIDC not working on Android

siepkes:
This fixes the issue when having an Android app authenticate with the OpenBoa OIDC identity provider and having OpenBao in turn authenticate the user with Google OIDC the redirect back to the Android app from OpenBao doesn't work. So after logging in with Google you get correctly redirected to OpenBao but then OpenBAO doesn't redirect back to the app. Chrome (104) on Android would show the logo loading animation forever.

This code was originally developed by @prajnamohan1 while working at Service Planet Rotterdam B.V.. I, Jasper Siepkes, in my capacity of IT Manager at Service Planet Rotterdam B.V. herby grant permission to use this code under the MPL-2.0 license.

Resolves: openbao#184

### DIFF
--- a/changelog/184.txt
+++ b/changelog/184.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: The latest versions of Chrome do not automatically redirect back to an Android app after multiple redirects during an OIDC authentication flow. A link was added to allow the user to manually redirect back to the app.
+```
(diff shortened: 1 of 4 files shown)


cipherboy (approved):
This looks very useful, many thanks for the contribution @siepkes and @prajnamohan1! 

I do not have an Android app handy to test with, but the code looks straightforward and makes sense to me.

naphelps (approved):

PR #196, opened 2024-03-10
Transit: Remove enterprise no-ops references from API and SDK

Gabrielopesantos:
Description
This PR is a follow-up to https://github.com/openbao/openbao/pull/142, it removes the managed_key_util.go file from the Transit engine, with no-ops enterprise functions and all of these functions.
Besides that, it also removes all references and usages of managed keys in the engine and SDK.


builtin/logical/transit/path_sign_verify.go
```diff
@@ -373,13 +372,6 @@ func (b *backend) pathSignWrite(ctx context.Context, req *logical.Request, d *fr
 		return logical.ErrorResponse(fmt.Sprintf("key type %v does not support signing", p.Type)), logical.ErrInvalidRequest
 	}
 
-	// Allow managed keys to specify no hash algo without additional conditions.
-	if hashAlgorithm == keysutil.HashTypeNone && p.Type != keysutil.KeyType_MANAGED_KEY {
-		if !prehashed || sigAlgorithm != "pkcs1v15" {
```
cipherboy, line 378: This inner if branch should be retained I believe.

builtin/logical/transit/path_sign_verify.go
```diff
@@ -627,13 +604,6 @@ func (b *backend) pathVerifyWrite(ctx context.Context, req *logical.Request, d *
 		return logical.ErrorResponse(fmt.Sprintf("key type %v does not support verification", p.Type)), logical.ErrInvalidRequest
 	}
 
-	// Allow managed keys to specify no hash algo without additional conditions.
-	if hashAlgorithm == keysutil.HashTypeNone && p.Type != keysutil.KeyType_MANAGED_KEY {
-		if !prehashed || sigAlgorithm != "pkcs1v15" {
```
cipherboy, line 632: As above.

builtin/logical/transit/path_sign_verify.go
Gabrielopesantos: Thanks for the early feedback. Going to have a look. :eyes:

builtin/logical/transit/path_sign_verify.go
Gabrielopesantos: Same as above.

builtin/logical/transit/path_sign_verify.go
Gabrielopesantos: Having the inner loop seems to be breaking the transit tests.

cipherboy: To be clear, I think the full conditional should be:

```go
if hashAlgorithm == keysutil.HashTypeNone && (!prehashed || sigAlgorithm != "pkcs1v15") { ... }
```

right? The top conditional applied if the key type was _not_ a managed key, and thus should be retained with the assertion around None hash type.

Gabrielopesantos: Of course 🤦.
With this change, the tests pass again.

cipherboy (approved):
Changes look good to me! Many thanks @Gabrielopesantos!

naphelps (approved):

naphelps (approved):

Issue #32, opened 2024-01-04
Bind tokens to an IP address on issuance 


thequailman:
It would be nice if OpenBao had a configurable token policy to bind a token to the generator's source IP on issuance.  This would help limit token leakage damage, and force roaming clients (like employees or contractors) to reauthenticate if they switch networks.

OpenBao can currently use CIDR ranges to restrict token issuance, but I'd like to see this applied during usage too: a token can only be used from the IP it was issued to.

I think this setting could be added to the general token policies as a bool, maybe `token_bind_source_ip`.

cipherboy:
@thequailman This one should be possible if you're interested in providing a PR. The on-disk definition of the token policy will need to change, as will many of the auth plugins to support this new policy boolean... I _think_, perhaps, we can get away with reusing `tokenutil.TokenParams.TokenBoundCIDRs`, setting the value at token creation time to the actual request source IP address. This gets passed via the adjacent `PopulateTokenAuth` to the auth entry, which should prevent usage from a different IP address. 

Some effort will probably be necessary to handle `X-forwarded-for`, which is already partially supported-ish (e.g., cert auth lacks it), but that can be a separate issue/PR. 

I think this info on the request should be sufficient to allow the backend plugin to decide the source IP without needing to modify Core to do it:

```sdk
    // Headers will contain the http headers from the request. This value will
    // be used in the audit broker to ensure we are auditing only the allowed
    // headers.
    Headers map[string][]string `json:"headers" structs:"headers" mapstructure:"headers" sentinel:""`

    // Connection will be non-nil only for credential providers to
    // inspect the connection information and potentially use it for
    // authentication/protection.
    Connection *Connection `json:"connection" structs:"connection" mapstructure:"connection"`
```

I think though, each backend might need to do it as there's no common helper called by them all -- `PopulateTokenAuth` doesn't take enough parameters (namely the request) to do it itself and updating that parameter would break external auth plugins. 

My 2c. :-)

PR #202, opened 2024-03-11
Allow strictly binding tokens to source IP address

cipherboy:
This adds the option, `token_strictly_bind_ip`, to strictly bind the
token's approved CIDRs to the source IP address a login request was
associated with. This means if the IP address were ever ported to a
system with a new IP address, the login workflow would have to be
restarted to request a fresh token.

This binding is done in accordance with the `X-Forwarded-For` header,
allowing the CIDR value to be populated and validated against the
correct end-client's IP address (when sent via trusted `X-Forwarded-For`
header).

---

This also fixes the rejection of invalid `X-Forwarded-For` headers:

When `x_forwarded_for_reject_not_authorized=false` is explicitly set in
the listener configuration, we need to remove the header before
processing the request, otherwise downstream consumers of the request
will not know if the header is authorized or not, and presume that,
because we passed it along, we trusted it.

---

This is missing tests; I might try using the container based testing infrastructure to do this; I think I can run OpenBao within the test suite and use a curl container to check if the binding works. Though, I don't think I'll have the capabilities to test `X-Forwarded-For` there. 

Curious to get your thoughts on this one @thequailman! 

Resolves: #32


cipherboy:
I see, I'll need to address how the number of auth plugins are computed:

```

2024-03-16T22:56:36.3123147Z === [31mFailed[0m
2024-03-16T22:56:36.3123917Z === [31mFAIL[0m: command TestAuthEnableCommand_Run/mount_all (2.01s)
2024-03-16T22:56:36.3124851Z     auth_enable_test.go:208: expected 9 credential backends, got 10
2024-03-16T22:56:36.3125964Z         	expected: []string{"ldap", "oidc", "radius", "kerberos", "kubernetes", "userpass", "approle", "cert", "jwt"}
2024-03-16T22:56:36.3127811Z         	backends: []string{"approle", "cert", "jwt", "kerberos", "kubernetes", "ldap", "radius", "token", "userpass", "userpass_binary"}
```

Perhaps I'll move `backend/credential/userpass_binary` to `vault/external_tests` for now.

sdk/helper/tokenutil/tokenutil.go
```diff
@@ -274,6 +296,23 @@ func (t *TokenParams) PopulateTokenAuth(auth *logical.Auth) {
 	auth.TokenType = t.TokenType
 	auth.TTL = t.TokenTTL
 	auth.NumUses = t.TokenNumUses
+
+	if t.TokenStrictlyBindIP {
+		addrs := []string{req.Connection.RemoteAddr}
+		values, present := req.Headers[textproto.CanonicalMIMEHeaderKey("X-Forwarded-For")]
+		if present && len(values) > 0 {
+			addrs = values
```
JanMa, line 304: Are you sure this has the effect you're intending it to have? All IPs which are part of the `X-Forwarded-For` header will be added to the token's allowed CIDR list. So if OpenBao is behind some L7 load-balancer, it's IP will be added to the token and thus everyone with that token can access the cluster.

By convention the first IP in the list of IPs of the `X-Forwarded-For` header is the one of the original client. 
_If_ you want to rely on the header to get the client's source IP, I think it would be good to only add the first value of the header to the token CIDR list.
See the relevant [MDN page ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax) for more information about this.

JanMa (changes requested):

sdk/helper/tokenutil/tokenutil.go
cipherboy: @JanMa Aha!

Initially I thought `X-Forwarded-For` would need to be manually added as a [`passthrough_request_header`](https://janma.github.io/openbao/api-docs/system/mounts/#tune-mount-configuration), but even then, we should probably respect [`x_forwarded_for_hop_skips`](https://janma.github.io/openbao/docs/configuration/listener/tcp/) in the listener config...


...but now that I read `http.WrapForwardedForHandler` more closely (see the set of `r.RemoteAddr` just before  the last `ServeHTTP`), I realize that I can just take `req.RemoteAddr` verbatim and not deal with this, so we can remove this entirely. Thanks!

naphelps (approved):

PR #204, opened 2024-03-12
Remove Enterprise features from api/, sdk/

cipherboy:
This removes various references of WALs, Enterprise-only Auto Pilot
configuration, Performance Secondaries, and Disaster Recovery from
`api/` and `sdk/`, including dependent pieces in command, agent, and
proxy (around enforcing consistency, FIPS status, and other related
topics).

This also fixes the README to update references to `DockerTestHelper` to
remove Vault Enterprise support.

---

Will be rebased once #203 is merged.


naphelps (approved):

cipherboy:
The CI failure:

> 2024-03-15T19:41:15.0104714Z     agent_test.go:874: timed out waiting for templates to render, last error: expected (13) templates, got (12)

is a test I've seen locally fail a bit when I'm stressing the system too much and happens independent of this PR. I've not really found a way to make it more reliable, so wonder if perhaps something is racing in a weird way. If anyone's interested in investigating it more, feel free!

command/base_predict_test.go
```diff
@@ -538,6 +539,8 @@ func TestPredict_PathsKVv2(t *testing.T) {
 	client, closer := testVaultServerWithKVVersion(t, "2")
 	defer closer()
 
+	time.Sleep(250 * time.Millisecond)
```
Gabrielopesantos, line 542: Was this test flaky or something?

cipherboy: Yeah, thought I added a comment, but perhaps not.

The test flakes when I run on power saving mode on my laptop, because it is running so many tests in parallel and the KVv1->KVv2 auto-migration that is done on mounting isn't complete by the time the subsequent write (below) is performed. Cleaner would be to add a retry loop to `testVaultServerWithKVVersion` to wait until the KVv2 migration has been performed, but I'm (a) lazy and (b) there's plenty of other command tests that an extra 250ms among friends won't be missed. :-)

You're welcome to tackle that if you want, I think there's other examples of retry loops (e.g., in PKI tests) that you could use if you want. :D

PR #205, opened 2024-03-12
Fix Solaris and remove i386 builds

cipherboy:
Red Hat all but has dropped support for 32-bit hardware back with
RHEL 8, with many other distros following suit.

We fixed support for Solaris temporarily by removing the dependency of
the final binary on Docker. This occurred as imports previously in a
test-only package landed in a non-test package (despite only being used
in tests), causing the unnecessary import.

See also: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/8.4_release_notes/deprecated_functionality


siepkes:
Objection, Your honor! :wink: 

In all seriousness, I use the "Solaris" build on illumos. illumos is the opensource continuation of OpenSolaris after Oracle re-closed the source of Solaris. illumos is maintained by a lot of old Sun folk who previously worked on Solaris but also by a lot of new blood. illumos is the basis for example for [Triton Datacenter](https://www.tritondatacenter.com/) and [Oxide's product](https://oxide.computer/).

Go also [supports illumos as a goos build tag](https://go.dev/doc/go1.13#illumos). While I can understand not wanting to invest time in Oracle's proprietary Solaris version I would like to make a case for keeping illumos included.

If you need any help on illumos specific issues you can also reach out to me.

cipherboy:
> Objection, Your honor! 😉

I think this is as valid of a reason to keep it as any ;-)

But the reason for removal wasn't that it was working -- it wasn't! :D

Check out the [latest build failures](https://github.com/openbao/openbao/actions/runs/8273453914/job/22637298152):

> ```
> **Run make ci-build**
> --> Building OpenBao revision 8f3383735c9db25688c42d51a917967b48029c2a, built 2024-03-13T20:21:54Z
> ~/work/openbao/openbao ~/work/openbao/openbao
> + go build -v -tags '' -ldflags ' -X github.com/openbao/openbao/version.GitCommit=8f3383735c9db25688c42d51a917967b48029c2a -X github.com/openbao/openbao/version.BuildDate=2024-03-13T20:21:54Z' -o dist/bao
> internal/goarch
> internal/unsafeheader
> internal/coverage/rtcov
> internal/godebugs
> internal/goexperiment
> ... SNIP ...
> github.com/docker/docker/pkg/pools
> github.com/docker/go-connections/sockets
> github.com/docker/docker/pkg/meminfo
> github.com/docker/docker/pkg/system
> text/scanner
> # github.com/docker/docker/pkg/system
> Error: ../../../go/pkg/mod/github.com/docker/docker@v24.0.7+incompatible/pkg/system/lstat_unix.go:20:9: undefined: fromStatT
> Error: ../../../go/pkg/mod/github.com/docker/docker@v24.0.7+incompatible/pkg/system/stat_unix.go:66:9: undefined: fromStatT
> github.com/moby/sys/sequential
> github.com/docker/docker/pkg/stdcopy
> github.com/michaelklishin/rabbit-hole/v2
> github.com/docker/docker/client
> github.com/moby/patternmatcher
> github.com/openbao/openbao/builtin/logical/totp
> ```

This only appears to occur on Solaris, because it lacks a `fromStatT` definition in Docker.

Now, what is weird is back on 50d31a1a5edefc7480d73d8bbbd81f56cd808196, the pipeline passed and Docker built fine. In 5fa3d9a8847373bc4b553642cbf82b952d20c4b3  was where the first build failure was with Solaris. While admittedly, we could use a `go mod tidy` on main, I don't see what in this commit could've caused that -- it is strictly a bunch of removals. 

Anyhow, I'll mark this one as draft and if you can fix Solaris, we'll keep it :D  If you get ambitious, a past project of mine used MacOS runners to get access to virtualization to be able to run arbitrary guest OSes, so you could install Solaris and run the test suite against it for us...

Let me know!

cipherboy:
@siepkes Yes, that sounds good to me!

siepkes:
Great! I'll get started on it!

naphelps (approved):

Issue #47, opened 2024-01-17
AWS Lambda Extension - keep or remove?


JanMa:
Upstream Vault maintains an AWS Lambda Extension which allows Lambda functions to read secrets from a Vault cluster. This Extension is explicitly mentioned in the [official documentation](https://developer.hashicorp.com/vault/docs/platform/aws/lambda-extension) which begs the question: Are we going to fork and keep it or remove the relevant docs?

**Vote**:
:+1:  to keep it
:-1:  to remove it.

cipherboy:
Like our discussion about the SSH helper, I imagine that this exclusively uses public APIs and can be left unforked.

naphelps:
Support of any feature or function is organic if there is community willing to support it.

cipherboy:
@naphelps agreed, but note that in this case, we're unlikely to see many benefits of forking rather than directly contributing upstream here, for this particular package.

naphelps:
@cipherboy Agreed.

vamsinm:
we modeled off of [extension](https://aws.amazon.com/blogs/compute/caching-data-and-configuration-settings-with-aws-lambda-extensions/) from aws and updated to use for vault. advantage of this extension is you dont need to load vault specifi apis and secrets are stored in memory instead of disk.

JanMa:
Closing as we chose to remove the docs in 27f0d4d7f213ca9ad57b97f56d6e184c4fc8cda8

PR #207, opened 2024-03-14
Server: Update references to server environment variables to reflect new name

Gabrielopesantos:
Updates the references to server environment variables, from prefix with "VAULT" to "BAO", in the documentation.
These variables were initially updated in the PR linked below.

Follow up to https://github.com/openbao/openbao/pull/157

## Target Release

Alpha


website/content/docs/platform/aws/lambda-extension.mdx
cipherboy, line 1: @JanMa I don't think we forked the AWS lambda extension, did we? I think that's what #47 talks about, which was inconclusive as we didn't have someone step forward and want to maintain it?

website/content/docs/platform/aws/lambda-extension.mdx
```diff
@@ -206,7 +206,7 @@ initialization timeout imposed by the Extensions API when writing files to disk.
...
 | `VAULT_DEFAULT_CACHE_ENABLED` | Enable caching for all requests, without needing to set the X-Vault-Cache-Control header for each request. Must be set to a boolean value.                                      | No       | `true`              |
 | `VAULT_ASSUMED_ROLE_ARN`      | Valid ARN of an IAM role that can be assumed by the execution role assigned to your Lambda function.                                                                            | No       | `arn:aws:iam::123456789012:role/xaccounts3access`
-| `VAULT_LOG_LEVEL`             | Log verbosity level, one of TRACE, DEBUG, INFO, WARN, ERROR, OFF. Defaults to INFO.                                                                                             | No       | `DEBUG`
+| `BAO_LOG_LEVEL`               | Log verbosity level, one of TRACE, DEBUG, INFO, WARN, ERROR, OFF. Defaults to INFO.                                                                                             | No       | `DEBUG`
```
cipherboy, line 209: Technically per above, we didn't fork this and so it would still use the `VAULT_LOG_LEVEL`, but if Jan agrees, we might remove this instead. But I wouldn't hold up merge on this, my 2c.,

cipherboy (approved):
Looks great, thanks @Gabrielopesantos!

website/content/docs/platform/aws/lambda-extension.mdx
Gabrielopesantos: Oh, this is docs from an extension. I used a command to automatically replace everything at once and then went through each change to understand if they made change, this one might have escaped.

Gabrielopesantos: So if we don't fork the aws extension, which I agree we shouldn't (at least for now), we keep this documentation file as it was originally?

cipherboy: Either we keep it or we remove it entirely. I'd probably just delete it entirely for now, tbh. My 2c.

website/content/docs/platform/aws/lambda-extension.mdx
JanMa: So far we did not touch it. Therefore I also left the docs mostly unchanged for now. IMO if there's no-one willing to support it before GA, we should remove the docs.

website/content/docs/platform/aws/lambda-extension.mdx
Gabrielopesantos: Done :+1:

cipherboy (approved):
Looks good, thank you @Gabrielopesantos !

naphelps:
@Gabrielopesantos There is a single commit that requires a --signoff for the DCO check to pass. https://github.com/openbao/openbao/pull/207/checks?check_run_id=23078806963

naphelps (approved):

Issue #220, opened 2024-03-18
`-non-interactive` flag still creates interactive command


cipherboy:
As reported by @bmendric on https://github.com/hashicorp/vault/issues/25987:

> **Describe the bug** Running `vault operator unseal -non-interactive` creates a prompt asking for an unseal key
> 
> **To Reproduce** Steps to reproduce the behavior:
> 
>     1. Run `vault server -dev`
> 
>     2. Run `vault operator unseal -address=http://localhost:8200 -non-interactive`
> 
> 
> **Expected behavior** The Vault CLI to exit with a non-zero exit code stating an unseal key was not provided.
> 
> **Environment:**
> 
>     * Vault Server Version (retrieve with `vault status`): N/A
> 
>     * Vault CLI Version (retrieve with `vault version`): v1.15.4
> 
>     * Server Operating System/Architecture: Ubuntu 20.04 amd64
> 
> 
> Vault server configuration file(s): N/A -- issue with the CLI
> 
> **Additional context** For context, because I know this looks like a crazy thing to do, I have some scripts to help setup dev environments where I run `vault operator unseal -non-interactive $UNSEAL_KEY`. Today I did that in a way where `$UNSEAL_KEY` was an empty string, which resulted in the "rendered" command line above.

---

This looks like it was because the `-non-interactive` flag was added later around the time of MFA, explaining why it was only used for disabling MFA. However, the description of the flag is broader, which makes callers think it can be applied more generally. It makes sense (especially for scripting purposes) to properly support this flag so it is worth fixing this bug.

This is a little entertaining though: for e.g., `bao login`, backends existing in the API an don't (presently) take a configurable `os.Stdin` or base CLI instance and so can't (automatically) detect the `flagNonInteractive`; we'll need to extend that interface.

PR #221, opened 2024-03-18
Add global handling of -non-interactive

cipherboy:
This adds global handling of the `-non-interactive` flag introduced originally for non-interactive MFA support. Most commands with interactive use of `stdin` will either get an empty reader (presumably allowing it to err later) or explicitly err about not being passed input that would otherwise be requested via stdin (e.g., `operator unseal` or `login`).

Resolves: #220

### DIFF
--- a/builtin/credential/cert/cli.go
+++ b/builtin/credential/cert/cli.go
@@ -13,7 +13,7 @@ import (
 
 type CLIHandler struct{}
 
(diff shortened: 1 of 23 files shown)


naphelps (approved):

cipherboy:
@naphelps looks like this one still fails tests, I'll address that when I get time later. Thought I had fixed that.

cipherboy:
@naphelps This should be fixed. Not sure why I added it to unseal, tbh, meant the test to be for `init`.

PR #223, opened 2024-03-19
Fix anchor generation in website partials

JanMa:
We use a custom plugin from upstream to include partials in our MDX docs
which can also contain MDX syntax that needs to be evaluated. The
inclusion needs to happen before the default docusaurus remark plugins
run for them to be able to properly evaluate the contained MDX syntax.

One example where this currently does not work is when we overwrite the
generated anchors of a heading.

**Current status:**
image

**Fixed status:**
image

### DIFF
--- a/website/docusaurus.config.ts
+++ b/website/docusaurus.config.ts
@@ -44,7 +44,7 @@ const config: Config = {
           // Please change this to your repo.
           // Remove this to remove the "edit this page" links.
           editUrl: "https://github.com/openbao/openbao/tree/main/website/",
-          remarkPlugins: [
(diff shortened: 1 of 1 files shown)


cipherboy (approved):
Assuming I'm looking at the right page (/docs/internals/telemetry/metrics/authn/#vault-expire-revoke) I think this looks good!

JanMa:
> Assuming I'm looking at the right page (/docs/internals/telemetry/metrics/authn/#vault-expire-revoke) I think this looks good!

I've added some screenshots to the PR body to show the difference

naphelps (approved):

PR #233, opened 2024-03-26
mark unfinished docs as draft

JanMa:
This has the effect that these docs are only visible when you start a
local development server. Once we decided hat to do regarding the
support of the platforms mentioned in these docs, we can remove the
draft tag.


cipherboy (approved):
Makes sense; I didn't look to see if there's more docs to mark as such.

naphelps (approved):

Issue #234, opened 2024-03-26
Unable to import api / sdk go modules


ollevche:
Hey there!

**Describe the bug**
Running `go get github.com/openbao/openbao/api` results in the following error:

> go: github.com/openbao/openbao/api@upgrade (v1.11.0) requires github.com/openbao/openbao/api@v1.11.0: parsing go.mod:
>         module declares its path as: github.com/hashicorp/vault/api
>                 but was required as: github.com/openbao/openbao/api

**To Reproduce**
1. Init a go module: `go mod init example.test`
2. Run `go get github.com/openbao/openbao/api`
3. See error

**Expected behavior**
OpenBao API go module has been added to project's go.mod as a dependency for further use.

**Environment:**
Clean go module.

**Additional context**
It's possible that Go module proxy server cached the old version of OpenBao API's go.mod (with `github.com/hashicorp/vault/api` specified). It can be resolved by publishing a new go mod version.

ollevche:
To ensure this is a cache problem, you can disable proxy by updating `GOPROXY` via the command: `go env -w GOPROXY=direct`. After that, it imports successfully.

cipherboy:
@ollevche yes, sadly this is a cache issue. Is why I want @naphelps to tag the next alpha as above the past releases. 

When forking, all tags were pushed over and subsequently tagged prior to removal, which has caused issues as we've not yet released. I think perhaps you can do what I did in https://github.com/cipherboy/devbao and pin to a hash in the interim.

ollevche:
thanks for the reply @cipherboy!

Bypassing the go proxy worked for me, but I believe CI could fail later: `GOPROXY="direct" go get github.com/openbao/openbao/api`.

Ideally, you would want to retract older versions, yes. But I think it's fine to simply publish a new one.

BTW when do you plan to release? I'm working on OpenTofu <> OpenBao integration (https://github.com/opentofu/opentofu/issues/1174) and this issue could affect our build process so that's why I'm curious.

nerijus:
Please see https://github.com/openbao/openbao/issues/231#issuecomment-2018939620 about the release.

ollevche:
@nerijus thanks!

naphelps:
@cipherboy I was going to tag the release `v2.20240329.0-alpha`. I am taking other suggestions as well.

cipherboy:
@naphelps 2>1, so that works for me!

naphelps:
[...]

PR #238, opened 2024-03-29
Issue 234: Updating OpenBao Version For Alpha

naphelps:
References: #234

Changes:
- Updates OpenBao to `v2.0.0-alpha20240329`.
- Retracts all prior Vault versions.
- Retracts all prior Vault API versions.
- Retracts all prior Vault SDK versions.
- Added changelog entry.


go.mod
```diff
@@ -354,3 +354,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+retract [v0.1.0, v1.14.9]
```
cipherboy, line 358: Do we want to retract priror API & SDK versions as well, in the respective modules?

naphelps: I will retract those as well.

cipherboy (approved):
CI looks good, but I don't have a laptop to test on at the moment!

PR #239, opened 2024-03-29
Issue-153: Adding a github workflow for handling releases.

naphelps:
References: #153

Changes:
- Added condition check to code-checker workflow to fix merge queues.
- Created a release workflow utilizing GoReleaser.
- Added GPG and Cosign signing support to release artifacts.
- Added alpine image to docker file for riscv64 architecture support.
- Bumped alpine image version to 3.19 from 3.18.
- Bumped UBI version to 9.3 from 8.8.


.github/workflows/code-checker.yml
```diff
@@ -74,6 +74,8 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: returntocorp/semgrep@sha256:ffc6f3567654f9431456d49fd059dfe548f007c494a7eb6cd5a1a3e50d813fb3
+    # Skip any PR created by dependabot and any check triggered by merge group
+    if: (github.actor != 'dependabot[bot]') && (github.event != 'merge_group')
```
cipherboy, line 78: Not sure I understand the latter half of this. Running on dependabot does seem wasteful (except that we'd want to bump the semgrep hash automagically, perhaps, and would prefer it runs then?) but would we ever have a scenario where PR A passes semgrep and PR B passes semgrep but the combination A+B doesn't? I guess it probably depends on the rules? My 2c., but I'd perhaps leave off the latter.

Dockerfile.release
cipherboy, line 1: This file is unnecessary; `Dockerfile` is already used for release builds by the `build` CI stage. `scripts/docker/Dockerfile` is the build variant.

.goreleaser.yaml
```diff
@@ -0,0 +1,494 @@
...
+    # You may remove this if you don't use go modules.
+    - go mod tidy
+    # you may remove this if you don't need go generate
+    - go generate ./...
+    - make bootstrap
+
+builds:
+  - id: builds-linux
+    ldflags:
+      - -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{ .CommitDate }}
```
cipherboy, line 22: This generate stripped builds which is incompatible with upstream's binaries which ship debug information. As I commented in https://github.com/hashicorp/vault/issues/21069, several people rely on the debug symbols; I can't remember the name of the product but there's a (popular among a subset of customers) product that instruments binaries in production and aggregates performance and other information.

.goreleaser.yaml
```diff
@@ -0,0 +1,494 @@
...
+    goarch:
+      - amd64
+      - arm
+      - arm64
+      - riscv64
+    goarm:
+      - "6"
+      - "7"
+    ignore:
+      - goos: darwin
```
cipherboy, line 36: `darwin` shouldn't be possible under a linux target unless I'm missing something?

.goreleaser.yaml
```diff
@@ -0,0 +1,494 @@
...
+    mod_timestamp: "{{ .CommitTimestamp }}"
+    skip: false
+  - id: builds-other
+    ldflags:
+      - -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{ .CommitDate }}
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - darwin
+      #- illumos
```
cipherboy, line 63: `illumos` is commented out, but @siepkes had asked for it. :-)

cipherboy (commented):
Couple of inline comments... I do think the `build` workflow could be updated with riscv support if we want, that was missed in #205.

.goreleaser.yaml
```diff
@@ -0,0 +1,494 @@
...
+      - src: ./.release/linux/package/etc/openbao/openbao.env
+        dst: /etc/openbao/openbao.env
+        file_info:
+          mode: 0644
+      - src: ./.release/linux/package/etc/openbao/openbao.hcl
+        dst: /etc/openbao/openbao.hcl
+        file_info:
+          mode: 0644
+      - src: ./.release/linux/package/usr/lib/systemd/system/openbao.service
+        dst: /usr/systemd/system/openbao.service
```
cipherboy, line 136: Based on `openbao-2.0.0~alpha20240329-1.x86_64.rpm` from https://github.com/openbao/openbao/actions/runs/8482716317, this should be at `/usr/lib/systemd/system/openbao.service`; the **`/lib`** is missing.

naphelps: Good catch, thanks. I will get this fixed.

.goreleaser.yaml
naphelps: Space constraints on the runner. I love to get more in but the runner runs out of disk space currently.

.goreleaser.yaml
naphelps: I copy-pasted the two build ids. I was tinkering around with space available and seeing feasibility of releasing the build paths sequentially.

.goreleaser.yaml
naphelps: Stripping those out then!

.github/workflows/code-checker.yml
naphelps: Per the semgrep project, merge queues are not supported at the moment. This is their workaround to allow the group to function.

cipherboy: Ah, this is sad!

.goreleaser.yaml
naphelps: resolved.

Dockerfile.release
naphelps: I will have to tinker to goreleaser to see if I can get it to separate out the linux binaries like `COPY dist/$TARGETOS/$TARGETARCH/$BIN_NAME /bin/`.

cipherboy: Well, if we're using goreleaser, we should just drop the old build system. No point wasting CI cycles. My 2c.

naphelps: I managed to alter the binary build output from goreleaser to match what the Dockerfile expects for the COPY. Updated the alpine image used with the Makefile to 3.19 as well.

.goreleaser.yaml
naphelps: Made a self-hosted runner. With the additional headroom I was able to add in Illumos, Solaris, BSD, and Arch Linux support.

naphelps:
@cipherboy I think I got all of your in-lines.

.goreleaser.yaml
```diff
@@ -0,0 +1,522 @@
...
+      - goos: windows
+        goarch: riscv64
+      - goos: windows
+        goarch: wasm
+    mod_timestamp: "{{ .CommitTimestamp }}"
+    skip: false
+
+report_sizes: true
+
+nfpms:
```
JanMa, line 119: Something causes the binary paths of the created packages to be weird. For example the `amd64` Debian package has the `bao` binary placed in `/usr/bin/dist/linux/amd64/bao` which is likely to cause issues. I think the binary should just be placed in `/usr/bin/bao`

JanMa (changes requested):

.goreleaser.yaml
naphelps: Ah so that is the side effect. I will need to replace our dockerfile then which will break or other ci.

Dockerfile.release
naphelps: From Jan, this has the side effect of storing the binaries in the packages in the wrong directory paths. I will overwrite the existing copy lines in the dockerfile for goreleaser, this will break the existing workflows.

.goreleaser.yaml
cipherboy: @naphelps yeah, my intent was to remove the build CI if we're happy with this.

naphelps: Reverted the directory changes in the build process. Adjusted the dockerfile to match. This will break the existing CI.

PR #264, opened 2024-04-03
Start a contributing guide with code layout

cipherboy:
This should help to familiarize new contributors to the layout of the repository and what its contents are. Individual areas are not covered in great detail, but instead a high-level overview of areas are present.

---

@naphelps @JanMa -- My general thought was perhaps topical coverage might be interesting? E.g., talking about how OpenBao starts up and unseals, walking through that with links to the code at a point in time? This grows out of date if non-pinned hashes are used (so I'd use pinned ones and drift from `main` over time)... Perhaps when changes are made, the docs could be periodically refreshed and updated?

Thoughts?

### DIFF
--- a/website/content/docs/contributing/code-organization.mdx
+++ b/website/content/docs/contributing/code-organization.mdx
@@ -1,0 +1,180 @@
+---
+description: |-
+  Documentation
+displayed_sidebar: docs
+---
+
+# Code organization
+
+OpenBao repository is organized into five main modules:
+
+ - An API client ([in `api/`](https://github.com/openbao/openbao/tree/main/api))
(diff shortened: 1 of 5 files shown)


naphelps:
@cipherboy I would say that a formal documentation review issue should be created for each GA release at the minimum. Point releases on an as needed basis. Creating the issue at the start of the process and linking to the milestone will prevent it from being lost or forgotten. As features/enhancements are developed over the course of the release, update the doc issue with the area of the codebase touched. That keeps a loose running list of action items to be reviewed prior to the release.

naphelps (approved):

Issue #274, opened 2024-04-07
Leftover: change from /etc/vault.d/ to /etc/openbao not fully done in .release/linux/package/usr/lib/systemd/system/openbao.service


johanneskastl:
The systemd service package still looks for its environment file in `/etc/openbao.d/`, while everything else is in `/etc/openbao/` (without the `.d`)

https://github.com/openbao/openbao/blob/a3e93c860dc58e06cab540066be8f15dc93be8df/.release/linux/package/usr/lib/systemd/system/openbao.service#L12

For comparison:
https://github.com/openbao/openbao/blob/a3e93c860dc58e06cab540066be8f15dc93be8df/.release/linux/package/usr/lib/systemd/system/openbao.service#L6
https://github.com/openbao/openbao/blob/a3e93c860dc58e06cab540066be8f15dc93be8df/.release/linux/package/usr/lib/systemd/system/openbao.service#L23

cipherboy:
Thanks @johanneskastl! I've opened #275 to fix this.

PR #275, opened 2024-04-07
Fix path to systemd environment

cipherboy:
This was missed in the fixes in
50d31a1a5edefc7480d73d8bbbd81f56cd808196.

Credit goes to @johanneskastl for identifying this.

Resolves: #274


johanneskastl (commented):
Thanks, looks correct!

JanMa (approved):

PR #290, opened 2024-04-11
Update dependency to github.com/go-viper/mapstructure/v2

mikelolasagasti:
Module `github.com/go-viper/mapstructure/v2` is the blessed fork for module `github.com/mitchellh/mapstructure` as annoucned by Mitchell Hashimoto in
https://gist.github.com/mitchellh/90029601268e59a29e64e55bab1c5bdc

Alpha version is just the module rename
https://github.com/go-viper/mapstructure/compare/v1.6.0...v2.0.0-alpha.1


cipherboy:
@mikelolasagasti what's your thoughts on removing this dependency?

mikelolasagasti:
As module `github.com/mitchellh/mapstructure` won't be maintained anymore and there is a blessed fork, I though it would be a good idea to adopt it for better future maintenance.  

The 2.0.0-alpha1 version of the new module is just the rename of the original module. 

OpenTofu accepted the PR I sent a few weeks ago.

cipherboy:
@mikelolasagasti Right, but I'm wondering if we'd be better off removing this dependency anyways.

I think we mostly considered its use as deprecated; there were a few places where we'd use it for compatibility, but in general, new uses should not have used it. This is especially true in the public facing APIs: we preferred explicit serialization over automatic serialization of the underlying datastructure, as it makes incremental breaking changes harder (e.g., if we change from an `int` -> `time.Duration` in the underlying datastructure, it is easier to provide request/response compatibility if we own the serialization ourselves). 

This isn't about whether it'd be good to adopt this, but whether we can simplify and remove it altogether. :-) But perhaps best we adopt this in the interim and remove usage of this and `github.com/fatih/structs` long-term.

cipherboy (approved):

cipherboy:
Hmm @mikelolasagasti -- now v2.0.0 is out of alpha, I went and updated to the latest version (`v2.2.1`) and it seems like `mapstructure.Error` has disappeared. Are you interested in fixing this? Sorry about the delay!

mikelolasagasti:
Yes, it seems a breaking change was introduced in 2.0.0 GA https://github.com/go-viper/mapstructure/releases/tag/v2.0.0

I changed `builtin/logical/transit/path_encrypt.go` to use Go's default error types.

cipherboy:
@mikelolasagasti Ignoring the conflict for a moment, it seems like `mapstructure` was also used in API request data parsing:

https://github.com/openbao/openbao/actions/runs/12785858995/job/35641849461?pr=290#step:7:13475

Somewhere along the way, the behavior of comma int slices has changed, and now the empty string gets parsed as a `[]int{0}` value... and `1,2,3` also doesn't get parsed correctly any more. 

This is I think the same underlying problem as https://github.com/openbao/openbao/actions/runs/12785858995/job/35641850227?pr=290. 

In the past, it looks like maybe `mapstructure` also did promotion from `string->[]string`, based on `command/agent/config/test-fixtures/config-template-full.hcl`.

Joy :-D

cipherboy:
Dropping from v2.2.0 due to https://github.com/go-viper/mapstructure/issues/64.

cipherboy:
https://github.com/openbao/openbao/pull/1459 has merged, so I'm going ahead and closing this. Thank you @mikelolasagasti for getting this started and @KrzysztofKornalewski-Reply for picking it up!

PR #292, opened 2024-04-13
Transactional storage - Tests & RFC

cipherboy:
This adds support for interactive transactions across the backend, physical, and plugin storage levels. This will ultimately allow for greater assurance in secrets management operations built on top of OpenBao.

See the RFC in #296 for more detailed information about the design of this change.

Resolves: #296

----

TODO:

 - [x] Resolve unresolved questions. 
 - [x] Benchmark subsystem performance when using transactions.
 - [x] Implement support for GRPC connected plugins.
 - [x] Expand randomized testing to `logical.Storage`, finish adding remaining layers to the cross-test. 
 - [x] Obey error handling discussed in RFC. 
 - [x] Add transaction concurrency limiting to InMem as well. 
 - [x] Expand `physical`'s `testing.go` to support basic testing of transactions. 
 - [x] ~Add support for disabling transactional storage from the configuration.~
 - [x] ~Update inmem-ha, `sdk/physical/latency.go`, and `sdk/physical/errors.go` to be transactional.~
 - [x] Think about separate raft limit size of transactions. 
 - [x] Rebase after GA and split PRs land.
    - [x] Split off removal of one-shot transactional interface into separate PR. -- #422 
    - [x] Split seal unwrapper removal into separate PR.  -- #423
    - [x] Split off core transaction definitions -- #433
    - [x] Split off each implementation -- #436, #437, #438, #439, #441, #442, and #443.
    - [x] Split off second level implementations (`BarrierView` and `AES-GCM Barrier View`) -- #440, #495
    - [x] Split off GRPC implementation -- #497
    - [x] Split off PKI root CA creation updates -- #498
    - [x] Add RFC text in this PR, merge.

<details>

<summary> Benchmarks </summary>

Without transactions (b56dc9c110b7b4f9170a1148d9febcc9a35cef38):

```
$ vault-benchmark run -config=docs/examples/pki/stored/pki-ed25519.hcl 
2024-07-22T12:28:57.564-0400 [INFO]  vault-benchmark: setting up targets
2024-07-22T12:29:01.596-0400 [INFO]  vault-benchmark: starting benchmarks: duration=30s
2024-07-22T12:29:31.604-0400 [INFO]  vault-benchmark: benchmark complete
Target: http://127.0.0.1:8200
op         count   rate         throughput   mean        95th%       99th%       successRatio
pki_issue  216959  7231.820942  7230.238976  1.363455ms  3.034576ms  5.149475ms  100.00%
```

With transactions (unused):

```
$ vault-benchmark run -config=docs/examples/pki/stored/pki-ed25519.hcl 
2024-07-22T12:27:22.730-0400 [INFO]  vault-benchmark: setting up targets
2024-07-22T12:27:26.775-0400 [INFO]  vault-benchmark: starting benchmarks: duration=30s
2024-07-22T12:27:56.779-0400 [INFO]  vault-benchmark: benchmark complete
Target: http://127.0.0.1:8200
op         count   rate         throughput   mean        95th%       99th%       successRatio
pki_issue  218338  7277.948872  7277.202214  1.356165ms  3.004142ms  4.872949ms  100.00%
```

With transactions (used for PKI issuance):

```
$ vault-benchmark run -config=docs/examples/pki/stored/pki-ed25519.hcl 
2024-07-22T12:33:15.869-0400 [INFO]  vault-benchmark: setting up targets
2024-07-22T12:33:19.896-0400 [INFO]  vault-benchmark: starting benchmarks: duration=30s
2024-07-22T12:33:49.898-0400 [INFO]  vault-benchmark: benchmark complete
Target: http://127.0.0.1:8200
op         count   rate         throughput   mean        95th%       99th%       successRatio
pki_issue  224037  7467.877003  7467.576805  1.318337ms  3.120374ms  5.828247ms  100.00%
```

<details>

<summary> Diff for PKI issuance transactions </summary>

```diff
$ git diff
diff --git a/builtin/logical/pki/path_issue_sign.go b/builtin/logical/pki/path_issue_sign.go
index 8c1a01451..7e554d129 100644
--- a/builtin/logical/pki/path_issue_sign.go
+++ b/builtin/logical/pki/path_issue_sign.go
@@ -388,6 +388,18 @@ func (b *backend) pathIssueSignCert(ctx context.Context, req *logical.Request, d
                return nil, logical.ErrReadOnly
        }
 
+       // If we have a transactional storage backend, let's use it.
+       if txnStorage, ok := req.Storage.(logical.TransactionalStorage); ok {
+               b.Logger().Debug("using transaction for root CA generation")
+               txn, err := txnStorage.BeginTx(ctx)
+               if err != nil {
+                       return nil, err
+               }
+
+               defer txn.Rollback(ctx)
+               req.Storage = txn
+       }
+
        // We prefer the issuer from the role in two cases:
        //
        // 1. On the legacy sign-verbatim paths, as we always provision an issuer
@@ -561,6 +573,14 @@ func (b *backend) pathIssueSignCert(ctx context.Context, req *logical.Request, d
                }
        }
 
+       // Finally, commit our transaction if we have one!
+       if txn, ok := req.Storage.(logical.Transaction); ok {
+               if err := txn.Commit(ctx); err != nil {
+                       return nil, err
+               }
+               resp.AddWarning("cert issued with transactions")
+       }
+
        resp = addWarnings(resp, warnings)
 
        return resp, nil
```

</details>

</details>

### DIFF
--- a/changelog/292.txt
+++ b/changelog/292.txt
@@ -1,0 +1,10 @@
+```release-note:feature
+**Transactional Storage**: Plugin developers can now take advantage of safe
+  storage modification APIs when the underlying physical storage supports
+  them. The `physical.TransactionalBackend` and `logical.TransactionalStorage`
+  types allow developers to begin read-only and writable transactions,
+  committing or rolling back the desired changes.
+```
+```release-note:improvement
+storage/raft: Add support for transactional storage semantics.
+```
--- a/physical/crosstest/cross_test.go
+++ b/physical/crosstest/cross_test.go
@@ -1,0 +1,1255 @@
+// We wish to build a backend cross-testing framework which allows us to find
+// differences in the way various storage interfaces behave. This will also
+// ultimately allow us to compose and stack various layers to ensure we have
+// correct semantics across composition as well.
+//
+// In OpenBao, the following storage layers are used, from lowest to highest:
+//
+//	[disk]
+//	1. A low-level backend, such as raft, file, or inmem. Only the former is
+//	   supported for production use, but inmem is frequently used for dev mode
+//	   and as a part of our test suite.
+//	2. A cache; this caches read operations, avoiding having to round-trip to
+//	   a potentially slow backend over a slow network for repeated reads to
+//	   the same entry.
+//	3. An error validation layer (encoding); this prevents writes with invalid
+//	   keys (non-utf-8 or non-printable characters).
+//	4. Our barrier encryption; this handles all encryption of storage entries.
+//	   At this point, we convert a physical.Backend into a logical.Storage
+//	   interface.
+//	5. A storage view; this restricts the access of the above barrier
+//	   encryption layer to a prefix.
+//	6. A barrier view; this is what is ultimately given to the backends; it
+//	   has a limited interface.
+//	[plugin]
+//
+// In the event of a GRPC-attached plugin, another layer appears as the GRPC
+// client stubs out support to transit storage operations through to the
+// GRPC server, though this isn't tested here.
+package crosstest
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/armon/go-metrics"
+	"github.com/go-test/deep"
+	log "github.com/hashicorp/go-hclog"
+	"github.com/stretchr/testify/require"
+
+	"github.com/openbao/openbao/physical/postgresql"
+	"github.com/openbao/openbao/physical/raft"
+	"github.com/openbao/openbao/sdk/v2/helper/logging"
+	"github.com/openbao/openbao/sdk/v2/physical"
+	"github.com/openbao/openbao/sdk/v2/physical/file"
+	"github.com/openbao/openbao/sdk/v2/physical/inmem"
+)
+
+const (
+	opsLogFile   = "/tmp/openbao-physical-random-ops.json"
+	txOpsLogFile = "/tmp/openbao-physical-random-tx-ops.json"
+	numOps       = 1000
+	numTxOps     = numOps
+)
+
+func Test_ExerciseBackends(t *testing.T) {
+	t.Parallel()
+
+	backends, cleanup := allPhysical(t)
+	defer cleanup()
+
+	exerciseBackends(t, backends)
+
+	// If any were transactions, let's rollback. We can't commit them as
+	// we wrote to the same area of storage in lots of places.
+	for name, backend := range backends {
+		if txn, ok := backend.(physical.Transaction); ok {
+			err := txn.Rollback(context.Background())
+			require.NoError(t, err, "failed to rollback transaction: %v", name)
+		}
+	}
+}
+
+func Test_RandomOpsBackends(t *testing.T) {
+	t.Parallel()
+
+	backends, cleanup := allPhysical(t)
+	defer cleanup()
+
+	ops := getRandomOps(t, numOps, false, 0)
+	// ops := replayOps(t, opsLogFile)
+	executeRandomOps(t, backends, ops)
+}
+
+func Test_RandomOpsTransactionalBackends(t *testing.T) {
+	t.Parallel()
+
+	backends, cleanup := allTransactionalPhysical(t)
+	defer cleanup()
+
+	txLimit := 10
+	ops := getRandomOps(t, numTxOps, true, txLimit)
+	executeRandomTransactionalOps(t, backends, ops, txLimit)
+}
+
+func replayOps(t *testing.T, file string) []*inmem.InmemOp {
+	data, err := os.ReadFile(file)
+	require.NoError(t, err, "error reading operations file")
+
+	var results []*inmem.InmemOp
+	err = json.Unmarshal(data, &results)
+	require.NoError(t, err, "error unmarshaling operations json")
+
+	return results
+}
+
+func Test_ExerciseTransactionalBackends(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+
+	backends, cleanup := allTransactionalPhysical(t)
+	defer cleanup()
+
+	// Create transactions and exercise the backend, rolling them back.
+	txns := make(map[string]physical.Backend, 2*len(backends))
+	for name, backend := range backends {
+		txn, err := backend.BeginTx(ctx)
+		require.NoError(t, err, "failed to create write transaction: %v", name)
+		txns[fmt.Sprintf("%v-tx-rw", name)] = txn
+
+		ro_txn, err := backend.BeginTx(ctx)
+		require.NoError(t, err, "failed to create read-only transaction: %v", name)
+		txns[fmt.Sprintf("%v-tx-ro", name)] = ro_txn
+	}
+
+	exerciseBackends(t, txns)
+
+	for name, txn := range txns {
+		err := txn.(physical.Transaction).Rollback(ctx)
+		require.NoError(t, err, "failed to rollback transaction: %v", name)
+	}
+
+	// Ensure we can do a single read/write transactions and commit them.
+	// This will leave us with an empty state, but potentially entries in
+	// any transaction logs.
+	txns = make(map[string]physical.Backend, len(backends))
+	for name, backend := range backends {
+		txn, err := backend.BeginTx(ctx)
+		require.NoError(t, err, "failed to create write transaction: %v", name)
+		txns[fmt.Sprintf("%v-tx-rw", name)] = txn
+	}
+
+	exerciseBackends(t, txns)
+
+	for name, txn := range txns {
+		err := txn.(physical.Transaction).Commit(ctx)
+		require.NoError(t, err, "failed to commit transaction: %v", name)
+	}
+
+	// Finally, exercise transactions.
+	exerciseTransactions(t, backends)
+}
+
+func getFile(t *testing.T, logger log.Logger) (physical.Backend, func()) {
+	backendPath, err := ioutil.TempDir("", "vault")
+	require.NoError(t, err, "error while creating file storage")
+
+	b, err := file.NewFileBackend(map[string]string{
+		"path": backendPath,
+	}, logger)
+	require.NoError(t, err, "error while initializing file backend")
+
+	return b, func() {
+		os.RemoveAll(backendPath)
+	}
+}
+
+func allPhysical(t *testing.T) (map[string]physical.Backend, func()) {
+	ctx := context.Background()
+	logger := logging.NewVaultLogger(log.Debug)
+	disableTxConf := map[string]string{"disable_transactions": "true"}
+
+	// Basic storage backends.
+
+	// raft, no transaction called on it.
+	prb, raftPureDir := raft.GetRaft(t, true, true)
+
+	// raft
+	rb, raftDir := raft.GetRaft(t, true, true)
+
+	// raft-in-tx
+	//
+	// Inside a raft transaction should behave the same as outside
+	// if it is writable. We are fine to reuse the same raft instance
+	// here as the transaction should not see stuff created after it.
+	rt, err := rb.BeginTx(ctx)
+	require.NoError(t, err, "failed to start raft transaction")
+
+	// file
+	fb, fileCleanup := getFile(t, logger)
+
+	// inmem
+	inm, err := inmem.NewInmem(disableTxConf, logger)
+	require.NoError(t, err, "failed to create in-memory backend")
+
+	// txinmem
+	_txinm, err := inmem.NewInmem(nil, logger)
+	require.NoError(t, err, "failed to create transactional in-memory backend")
+	txinm := _txinm.(physical.TransactionalBackend)
+
+	// txinmem-in-tx
+	txinmtx, err := txinm.BeginTx(ctx)
+	require.NoError(t, err, "failed to start transaction from transactional in-memory backend")
+
+	// postgresql
+	psql, psqlCleanup := postgresql.GetTestPostgreSQLBackend(t, logger)
+
+	// Now compose multiple storage backends together!
+
+	// raft + cache
+	rbc, raftCacheDir := raft.GetRaft(t, true, true)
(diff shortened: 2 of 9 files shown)


command/commands.go
```diff
@@ -145,13 +145,10 @@ var (
 	}
 
 	physicalBackends = map[string]physical.Factory{
-		"file_transactional":     physFile.NewTransactionalFileBackend,
-		"file":                   physFile.NewFileBackend,
-		"inmem_ha":               physInmem.NewInmemHA,
-		"inmem_transactional_ha": physInmem.NewTransactionalInmemHA,
-		"inmem_transactional":    physInmem.NewTransactionalInmem,
```
cipherboy, line 152: Probably need to add this back, now that I added a transactional InMem data store to compare the raft one against.

physical/raft/raft.go
```diff
@@ -1709,21 +1649,41 @@ func (b *RaftBackend) applyLog(ctx context.Context, command *LogData) error {
 		return errors.New("entries on FSM response were empty")
 	}
 
-	for i, logOp := range command.Operations {
-		if logOp.OpType == getOp {
-			fsmEntry := fsmar.EntrySlice[i]
+	if len(command.Operations) < 1 {
```
cipherboy, line 1652: We should expand the tests to handle `raftchunking.ChunkingApply` + transaction interaction.

command/server.go
```diff
@@ -136,12 +135,10 @@ type ServerCommand struct {
 	flagDevSkipInit        bool
 	flagDevThreeNode       bool
 	flagDevFourCluster     bool
-	flagDevTransactional   bool
```
cipherboy, line 139: We should probably add this back, to allow the dev server to use the transactional in-mem backend.

helper/testhelpers/teststorage/teststorage.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 func MakeInmemBackend(t testing.T, logger hclog.Logger) *vault.PhysicalBackendBundle {
-	inm, err := inmem.NewTransactionalInmem(nil, logger)
```
cipherboy, line 34: I'd probably revert this again and make transactional the default. We could potentially make this a random switch, to test most backends with either, randomly (emitting a log message to identify it)... But this seems unnecessary, given we can show sufficient equivalence with the randomized testing below.

physical/raft/fsm.go
```diff
@@ -36,6 +36,10 @@ const (
 	putOp
 	restoreCallbackOp
 	getOp
+	verifyReadOp
+	verifyListOp
+	beginTxOp
+	commitTxOp
```
cipherboy, line 42: `beginTxOp` / `commitTxOp` are mostly transaction sentinels for the time being. However, when implementing the single hash method (number 3), we can send the hash type in `beginTxOp` and the final hash in `commitTxOp`. We'll need `verifyReadOp` because while we can re-order any un-modified reads to occur at the end (sending them along in the `commitTxOp` if we wanted fewer operations in the log), we can't avoid sending reads before writes. We'll need `verifyListOp`, because its results can be impacted by writes and we won't necessarily know to compact (or not compact) them, so we'd prefer to keep them in execution order.

physical/raft/fsm.go
```diff
@@ -684,15 +697,18 @@ func (f *FSM) ApplyBatch(logs []*raft.Log) []interface{} {
...
 	}
 
-	err = f.db.Update(func(tx *bolt.Tx) error {
-		b := tx.Bucket(dataBucketName)
-		for _, commandRaw := range commands {
-			entrySlice := make([]*FSMEntry, 0)
+	for _, commandRaw := range commands {
+		inTx := false
+		entrySlice := make([]*FSMEntry, 0)
+		err = f.db.Update(func(tx *bolt.Tx) error {
```
cipherboy, line 703: Worth adding a comment around this re-ordering.

physical/raft/fsm.go
```diff
@@ -684,15 +697,18 @@ func (f *FSM) ApplyBatch(logs []*raft.Log) []interface{} {
...
+			b := tx.Bucket(dataBucketName)
 			switch command := commandRaw.(type) {
 			case *LogData:
 				for _, op := range command.Operations {
 					var err error
 					switch op.OpType {
+					case beginTxOp, commitTxOp:
```
cipherboy, line 710: Probably worth sanity-checking that the first operation is a `beginTxOp` and the last is a `commitTxOp` and we never see these appear anywhere else.

sdk/physical/transactions.go
```diff
@@ -5,149 +5,50 @@ package physical
```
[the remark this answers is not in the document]
cipherboy: I suppose it'd be cleaner if I just deleted the file entirely in that former commit, perhaps creating it as a separate PR, and then start with a clean copyright header here. I'll do that.

physical/raft/transaction.go
```diff
@@ -0,0 +1,605 @@
...
+
+func (b *RaftBackend) newTransaction(ctx context.Context, writable bool) (*RaftTransaction, error) {
+	// Grab a transaction permit pool entry so that we can limit the number of
+	// concurrent transactions. Also grab a read lock in the underlying FSM
+	// to prevent key changes from occurring while a transaction is ongoing.
+	// These will be released when we finish this transaction.
+	b.txnPermitPool.Acquire()
+	b.fsm.l.RLock()
+
+	// All transactions are read-only; this gives us a consistent view of
```
cipherboy, line 166: underlying

physical/raft/transaction.go
```diff
@@ -0,0 +1,605 @@
...
+					OpType: beginTxOp,
+				},
+			},
+		},
+		writable: writable,
+	}, nil
+}
+
+func (t *RaftTransaction) Put(ctx context.Context, entry *physical.Entry) error {
+	if !t.writable {
```
cipherboy, line 189: I guess this doesn't strictly need to be an atomic since it can't be modified, but it'd be good practice to make this consistent.

physical/raft/transaction.go
```diff
@@ -0,0 +1,605 @@
...
+		OpType: putOp,
+		Key:    entry.Key,
+		Value:  entry.Value,
+	})
+
+	return nil
+}
+
+func (t *RaftTransaction) Get(ctx context.Context, key string) (*physical.Entry, error) {
+	t.l.RLock()
```
cipherboy, line 239: This should probably be dropped to a regular mutex, as even for list/read operations, we append to `t.log.Operations`, so we can't necessarily do those in parallel.

physical/raft/transaction.go
```diff
@@ -0,0 +1,605 @@
...
+
+	return nil
+}
+
+func (t *RaftTransaction) List(ctx context.Context, prefix string) ([]string, error) {
+	return t.ListPage(ctx, prefix, "", -1)
+}
+
+// returns (entryName, isFolder, shouldVisit)
+func shouldIncludeEntry(prefix string, after string, key string) (string, bool, bool) {
```
cipherboy, line 338: Probably should include `list` in the name.

physical/raft/transaction.go
```diff
@@ -0,0 +1,605 @@
...
+	// put or delete has occurred in the transaction, it supersedes the
+	// value we would've gotten from the underlying data store. Here however,
+	// we always want to execute the list and remove entries if there have
+	// been writes that affect it.
+	//
+	// This is complex to do efficiently. We might have deleted an entire
+	// subtree that might show up in a list. We could've also added more
+	// entries, such that the list is unnecessary.
+	//
+	// We do this in two steps: perform the underlying list, ignoring results
```
cipherboy, line 370: Comment needs to be updated.

physical/raft/transaction.go
```diff
@@ -0,0 +1,605 @@
...
+	}
+
+	// Now that we have our result, save the operation in the log
+	// for verification. To do so, we hash the results.
+	listParams, contentsHash, err := createListVerificationEntry(prefix, after, limit, keys)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add the read to the verification log, to ensure nobody else has written
```
cipherboy, line 487: read->list

physical/raft/transaction.go
```diff
@@ -0,0 +1,605 @@
...
+		return err
+	}
+
+	// If no writes have occurred, we don't need to send a log to Raft. We
+	// might have conflicted on a verification, but we won't negatively impact
+	// an other writer (due to not causing a conflict ourselves). Our state of
+	// reads were guaranteed to be consistent, so it would be no different than
+	// having executed in a read-only transaction.
+	if !t.writable || !t.haveWritten {
+		// Clear our log and updates, restoring us to an empty state.
```
cipherboy, line 527: This should probably go in the defer; I don't think we'd want to potentially leave this dangling in the event of an error.

physical/raft/transaction.go
```diff
@@ -0,0 +1,605 @@
...
+					OpType: beginTxOp,
+				},
+			},
+		}
+
+		return nil
+	}
+
+	// Append the last log entry, if necessary.
+	if t.log.Operations[len(t.log.Operations)-1].OpType != commitTxOp {
```
cipherboy, line 541: This should never be false based on the one-commit/rollback assumption.

physical/raft/transaction.go
```diff
@@ -0,0 +1,605 @@
...
+
+	return keys, nil
+}
+
+func (t *RaftTransaction) Commit(ctx context.Context) error {
+	t.l.Lock()
+	defer t.l.Unlock()
+
+	if t.haveFinishedTx {
+		return nil
```
cipherboy, line 503: `nil` is not correct, we'd want to make this a transaction reuse error.

sdk/helper/keysutil/encrypted_key_storage.go
```diff
@@ -307,3 +321,169 @@ func (s *encryptedKeyStorage) encryptPath(path string) (string, error) {
...
+}
+
+// List implements the logical.Storage List method, and decrypts all the items
+// in a path prefix. This can only operate on full folder structures so the
+// prefix should end in a "/".
+func (e *eksTransaction) List(ctx context.Context, prefix string) ([]string, error) {
+	return e.ListPage(ctx, prefix, "", -1)
+}
+
+// ListPage implements the logical.Storage List method, and decrypts a subset
```
cipherboy, line 365: Is reusing the LRU bad here? I don't actually think it is: all it does is help with decrypting keys in the underlying list, which means we can drop this implementation and reuse the underlying one instead.

sdk/logical/storage_view.go
```diff
@@ -9,37 +9,74 @@ import (
...
+type StorageViewTransaction interface {
+	Transaction
+	StorageViewCore
 }
 
 var ErrRelativePath = errors.New("relative paths not supported")
 
-func NewStorageView(storage Storage, prefix string) *StorageView {
-	return &StorageView{
+func NewStorageView(storage Storage, prefix string) StorageView {
```
cipherboy, line 39: Since this is called from `SubView`, we need to make sure it also handles `Transaction`, creating a `StorageViewTransaction` instead.

vault/barrier.go
```diff
@@ -161,27 +155,41 @@ type SecurityBarrier interface {
...
 
-	// SecurityBarrier must provide the storage APIs
-	logical.Storage
-
 	// SecurityBarrier must provide the encryption APIs
 	BarrierEncryptor
 }
 
-// BarrierStorage is the storage only interface required for a Barrier.
-type BarrierStorage interface {
```
cipherboy, line 172: `BarrierStorage` was unused and conformed to `logical.Storage`.

vault/barrier_aes_gcm.go
```diff
@@ -93,6 +99,15 @@ type AESGCMBarrier struct {
 	totalLocalEncryptions *atomic.Int64
 }
 
+type TransactionalAESGCMBarrier struct {
+	*AESGCMBarrier
+}
+
+type AESGCMBarrierTransaction struct {
```
cipherboy, line 106: Probably worth a comment here why we deviate from the pattern elsewhere.

vault/sealunwrapper.go
```diff
@@ -1,188 +0,0 @@
-// Copyright (c) HashiCorp, Inc.
-// SPDX-License-Identifier: MPL-2.0
-
-package vault
```
cipherboy, line 4: This removal could be a separate PR as well.

sdk/physical/transactions.go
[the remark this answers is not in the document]
naphelps: That would be what I would do. Keeps things as clean and separate as possible.

command/commands.go
cipherboy: I think I'll leave it off: we probably want inmem to be transactional by default.

physical/raft/raft.go
cipherboy: This works fine, but we'll probably want to run it separately as it blows up the tests a lot to run it frequently.

command/server.go
cipherboy: As above, let's leave it enabled by default instead for dev mode.

physical/raft/transaction.go
cipherboy: I moved these checks inside the lock. We'll need the lock anyways, so there's no point in using an atomic for this as erring early doesn't buy us much.

sdk/logical/storage_view.go
cipherboy: `SubView` more generally is also one of the few storage layers that we'd expect to occur within a transaction. Things like the encrypted storage view (used by K/V plugin) are less likely to occur (as you'd probably wrap the entire outer storage entry). Thus, exposing `Commit`/`Rollback` on the `SubView` directly might be nicer for callers.

cipherboy:
@DanGhita @JanMa I'm inclined to land this with just the tests and the RFCs now and then rework both as necessary, when we're landing PostgreSQL transactions.

However, I still think PostgreSQL transactions should be landed before we do 2.1.0 release.

Thoughts?

DanGhita (approved):

cipherboy:
Thank you @DanGhita and @JanMa for all your reviews!! Merging...

PR #294, opened 2024-04-14
Correctly seek in raft's ListPage when after=.

cipherboy:
When a non-empty path is given that ends in a slash and after=., the resulting seek prefix from joining the path and the period has the slash trimmed off. This results in no list results, because this seek prefix lacks the trailing slash and is thus not prefixed by the expected prefix. E.g., for foo/ and after=., we end up with seekPrefix=foo, which doesn't start with our prefix (foo/).

Correctly handle this edge case. This was caught by randomized testing across multiple storage backends.

### DIFF
--- a/changelog/294.txt
+++ b/changelog/294.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+physical/raft: fix ListPage calls when after=. resulting in an empty list
+```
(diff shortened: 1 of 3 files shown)


Gabrielopesantos (commented):
I wasn't familiar with this behaviour from joining paths, interesting edge case. In which case an `after` be a "." for this to happen?
P.S. CI is failing.

Gabrielopesantos (approved):

cipherboy:
@Gabrielopesantos Yeah, I think the build pipelines were broken by the new release tooling, I was suggesting it could be removed but no dice. CI proper (test execution) seems fine.

This was mostly detected by the randomized testing of #292, it doesn't really impact much in practice (unless `.` was used as a prefix in some storage entries), but still good to fix for consistency.

naphelps (approved):

PR #313, opened 2024-04-29
Avoid invoking token helper on login

### LINKED ISSUE #312: `bao login` calls the token helper with `get` and expects that to succeed
**Describe the bug**

When I run `bao login`, and a [token helper](https://openbao.org/docs/commands/token-helper/) is configured, Bao will call the token helper with `get`. This `get` might fail, the token helper might not have a token. The reason for running `bao login` in the first place, is to obtain a token and store it using the token helper. But even if the token helper has no token yet, Bao forces it to exit with exit code 0. If it exits with a nonzero exit code, then `bao login` fails, and we never get to the `store` step.

We can of course make the token helper exit with exit code 0, but then if you run e.g. `bao kv get`, and the token helper has no token, then the token helper no longer has the ability to make the `kv get` fail. Instead the `kv get` will fail with a misleading “permission denied” error. Permission is denied, because no `X-Vault-Token` was included in the request. Bao did not include a token, because the token helper did not return 1 but signalled success either way.

**To Reproduce**
1. Configure a token helper in `~/.bao` by putting e.g. `token_helper = "/tmp/token_helper.sh"` there.
2. In that file, put the following and make the file executable:
   ```bash
   #!/usr/bin/bash
   echo "$@" >&2
   exit 1
   ```
4. Run `bao login`.
5. Observe how the script got called with `get`.
6. Observe how returning 1 from that `get` prevented `bao login` from doing any login.

**Expected behavior**
I would expect `bao login` to not call the token helper with `get`. It should only call the token helper with `store`.

Once `bao login` does not call the token helper with `get`, it becomes possible to return 1 when a get lookup fails (because the token does not exist), and that enables the token helper to print a helpful error message in that case, instead of leaving the user with a misleading permission denied error.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): irrelevant, we don’t even get tot he point where `bao` talks to a server
* OpenBao CLI Version (retrieve with `bao version`): `OpenBao v2.0.0-alpha20240329 ('74c2dddb0612b9a3da79384c20638266aa7de407'), built 2024-04-26T10:19:19Z`
* Server Operating System/Architecture: Linux 6.8.7 / x86-64

OpenBao server configuration file(s): irrelevant

**Additional context**

This bug report is an adaptation of https://github.com/hashicorp/vault/issues/23194.

ruuda:
Resolves #312.

### Problem

`bao login` used to call the token helper in `get` mode, because it constructs an HTTP client, and the client automatically loads the token. For almost everything that is the right thing to do, but for login, that is the thing that is supposed to retrieve the token, and login itself does not require the token. In fact, `vault login` would erase the token on the client later on.

Calling the token helper in `get` mode is a problem, because if the token helper fails, that blocks the login. But the token helper might fail because it doesn't have a token yet.

The call to `Get` originates here, in the construction of the http client: https://github.com/openbao/openbao/blob/6ef2a6031453b8bdc62fe9cb9511c223323a9e66/command/base.go#L145

That is being called from here: https://github.com/openbao/openbao/blob/6ef2a6031453b8bdc62fe9cb9511c223323a9e66/command/login.go#L215

### Solution

Split `Client` into two parts: a `ClientWithoutToken` that does most of what `Client` did previously, except for setting a token. And `Client` that also sets the token. This does change _when_ the token gets set on the client for calls to `Client`, but as far as I can tell, the remainder of the former `Client`, now `ClientWithToken` method does not rely on the token being set.

For all existing code except `login`, `Client` still behaves the way it did and nothing changes. For `login` in particular, we can now call `ClientWithoutToken`. This ensures that the token helper does not get called when the http client is initialized. The token helper still gets called to store the token later.

### Open questions, testing

There is still this snippet that happens after constructing the client: https://github.com/openbao/openbao/blob/6ef2a6031453b8bdc62fe9cb9511c223323a9e66/command/login.go#L221-L225

I _think_ this is now useless, the client will not have a token anyway so we don’t need to reset it. But it makes me wonder — if `authMethod == "token"`, does my change break anything?

Also, I am not sure what the best way is to add tests for this, if somebody can point me in the right direction, I would appreciate.

## Note about Source-Available License Policy

I adapted this patch from one I originally submitted to Hashicorp Vault in https://github.com/hashicorp/vault/pull/23209. However, the change was not accepted there and never included into a Hashicorp release, so I think it’s fine to submit it here?

The pull request was closed there with a message that’s unclear to me, but given that it took more than half a year and me posting and asking in multiple places before a person even looked at it, I don’t have high hopes for getting a clarification there. I hope OpenBao is more open to community contributions. If the approach in this pull request is not the right one, I’m happy to change it if somebody can explain what the correct approach would be.

## Target Release

No particular target, by default this was pre-filled to 1.14.7.

### DIFF
--- a/command/login.go
+++ b/command/login.go
@@ -216,7 +216,7 @@ func (c *LoginCommand) Run(args []string) int {
 	}
 
 	// Create the client
-	client, err := c.Client()
+	client, err := c.ClientWithoutToken()
 	if err != nil {
 		c.UI.Error(err.Error())
 		return 2
--- a/changelog/313.txt
+++ b/changelog/313.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli/login: Avoid calling the token helper in `get` mode.
+```
--- a/command/base.go
+++ b/command/base.go
@@ -73,9 +73,10 @@ type BaseCommand struct {
 	client *api.Client
 }
 
-// Client returns the HTTP API client. The client is cached on the command to
-// save performance on future calls.
-func (c *BaseCommand) Client() (*api.Client, error) {
+// Construct the HTTP API client, but do not set the token on it yet. This is to
+// avoid invoking the token helper for calls that do not need a token, such as
+// `vault login`.
+func (c *BaseCommand) ClientWithoutToken() (*api.Client, error) {
 	// Read the test client if present
 	if c.client != nil {
 		return c.client, nil
@@ -133,26 +134,6 @@ func (c *BaseCommand) Client() (*api.Client, error) {
 	// Set the wrapping function
 	client.SetWrappingLookupFunc(c.DefaultWrappingLookupFunc)
 
-	// Get the token if it came in from the environment
-	token := client.Token()
-
-	// If we don't have a token, check the token helper
-	if token == "" {
-		helper, err := c.TokenHelper()
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to get token helper")
-		}
-		token, err = helper.Get()
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to get token from token helper")
-		}
-	}
-
-	// Set the token
-	if token != "" {
-		client.SetToken(token)
-	}
-
 	client.SetMFACreds(c.flagMFA)
 
 	// flagNS takes precedence over flagNamespace. After resolution, point both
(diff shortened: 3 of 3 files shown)


cipherboy:
@ruuda in this case, since it was your own PR that you own the rights to, it is fine. In general, taking PRs from upstream from others isn't accepted. :-)

I'll have to take a closer look at this, but did you happen to handle @divyaac's comment? At first glance it looks valid; a token helper could be used in conjunction with `bao login -method=token`, which would expect the helper to already have set said token. Thoughts? :-)

command/login.go
```diff
@@ -216,7 +216,7 @@ func (c *LoginCommand) Run(args []string) int {
 	}
 
 	// Create the client
-	client, err := c.Client()
+	client, err := c.ClientWithoutToken()
```
cipherboy, line 219: In particular, down below where `token` is set, I'd add an else condition to re-trigger the helper logic on (new) lines 189...196 above in `command/base.go`. 

Other than that, LGTM!

ruuda:
> A user want to log in (`bao login`). They want a token (and may not have one set). If they do, use that one. If they don't, consult the token helper and use that one. Now if the token helper fails, they're explicitly logging in, so... they wouldn't have a token. And so they'd want to know that and fail bao login, IMO?

Yes it does make sense to me now! I was confused about the order of things, if you provide a token in the environment or on the command line, but the token helper doesn’t have it yet, then we should not check the token helper, and the login should succeed. I’ll update the PR accordingly.

ruuda:
I’m testing this further, and the current behavior of `bao login -method=token`, even if you pass a `VAULT_TOKEN` in the environment, is to ask the user interactively. The prompt is only skipped if you explicitly pass the token as a CLI argument. Even if there is a valid token in `~/.vault-token` (which is where the default token helper stores it), `bao login -method=token` does not use it.

If we make `bao login -method=token` consult the token helper now, that would change that behavior. In particular, it would change the behavior of `bao login` without additional arguments, which defaults to `-method=token`, in the case where you have an old expired token in `~/.vault-token`. (Which for me is very common, it happens every time I start my working day.) In the past it would then have prompted for a token interactively, but if we consult the token helper, it would print a permission denied error.

I think it is probably best not to change that behavior? I will make the case where the token is provided as a CLI arg store it in the token helper then, but not change how it works without additional args. @cipherboy what do you think?

cipherboy:
Huh, TIL!

Then I guess we're good. Thank you @ruuda!

cipherboy (approved):
LGTM! Thanks @ruuda. Mind merging this one @naphelps?

PR #319, opened 2024-05-03
Add oidc device flow after direct flow

DrDaveD:
This adds support for OIDC device flow on top of pr #318.  #318 has to be committed first and all its changes are included here.  If you'd like to see just the changes compared to that pr, see [my own pr 1](https://github.com/DrDaveD/openbao/pull/1).

Device flow has several advantages over direct callback mode:
1. There's no need to configure allowed redirect uris for the client.
2. There's no need to configure firewalls to allow the Authorization Server to call back to Vault.
3. There's no need for the Authorization Server to recognize the CA cert for Vault.
4. The URL that the user sees is simpler.

So it's worth having device flow even compared to direct callback mode, although direct callback mode is good when Authorization Servers don't support device flow.

Device flow is enabled with this implementation by setting the role configuration `callback_mode=device`.  The device authorization endpoint is auto-discovered.   This also adds an additional optional role configuration option `poll_interval` which defaults to 5.

The client API is slightly extended, to add an optional `user_code` option in the auth response, and to add a `slow_down` reply to a poll request.  A `redirect_uri` passed in to the auth API is ignored in device flow.

This is essentially the same PR as hashicorp/vault-plugin-auth-jwt#131 which many people have expressed an interest in but has been sitting unmerged for a few years.

### DIFF
--- a/builtin/credential/jwt/path_config.go
+++ b/builtin/credential/jwt/path_config.go
@@ -8,8 +8,12 @@ import (
 	"crypto"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"errors"
+	"fmt"
+	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/hashicorp/cap/jwt"
@@ -161,6 +165,91 @@ func (b *jwtAuthBackend) config(ctx context.Context, s logical.Storage) (*jwtCon
 	b.cachedConfig = config
 
 	return config, nil
+}
+
+func contactIssuer(ctx context.Context, uri string, data *url.Values, ignoreBad bool) ([]byte, error) {
+	var req *http.Request
+	var err error
+	if data == nil {
+		req, err = http.NewRequest("GET", uri, nil)
+	} else {
+		req, err = http.NewRequest("POST", uri, strings.NewReader(data.Encode()))
+	}
+	if err != nil {
+		return nil, err
+	}
+	if data != nil {
+		req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	}
+
+	client, ok := ctx.Value(oauth2.HTTPClient).(*http.Client)
+	if !ok {
+		client = http.DefaultClient
+	}
+	resp, err := client.Do(req.WithContext(ctx))
+	if err != nil {
+		return nil, nil
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, nil
+	}
+
+	if resp.StatusCode != http.StatusOK && (!ignoreBad || resp.StatusCode != http.StatusBadRequest) {
+		return nil, fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	return body, nil
+}
+
+// Discover the device_authorization_endpoint URL and store it in the config
+// This should be in coreos/go-oidc but they don't yet support device flow
+// At the same time, look up token_endpoint and store it as well
+// Returns nil on success, otherwise returns an error
+func (b *jwtAuthBackend) configDeviceAuthURL(ctx context.Context, s logical.Storage) error {
+	config, err := b.config(ctx, s)
+	if err != nil {
+		return err
+	}
+
+	b.l.Lock()
+	defer b.l.Unlock()
+
+	if config.OIDCDeviceAuthURL != "" {
+		if config.OIDCDeviceAuthURL == "N/A" {
+			return fmt.Errorf("no device auth endpoint url discovered")
+		}
+		return nil
+	}
+
+	caCtx, err := b.createCAContext(b.providerCtx, config.OIDCDiscoveryCAPEM)
+	if err != nil {
+		return errwrap.Wrapf("error creating context for device auth: {{err}}", err)
+	}
+
+	issuer := config.OIDCDiscoveryURL
+
+	wellKnown := strings.TrimSuffix(issuer, "/") + "/.well-known/openid-configuration"
+	body, err := contactIssuer(caCtx, wellKnown, nil, false)
+	if err != nil {
+		return errwrap.Wrapf("error reading issuer config: {{err}}", err)
+	}
+
+	var daj struct {
+		DeviceAuthURL string `json:"device_authorization_endpoint"`
+		TokenURL      string `json:"token_endpoint"`
+	}
+	err = json.Unmarshal(body, &daj)
+	if err != nil || daj.DeviceAuthURL == "" {
+		b.cachedConfig.OIDCDeviceAuthURL = "N/A"
+		return fmt.Errorf("no device auth endpoint url discovered")
+	}
+
+	b.cachedConfig.OIDCDeviceAuthURL = daj.DeviceAuthURL
+	b.cachedConfig.OIDCTokenURL = daj.TokenURL
+	return nil
 }
 
 func (b *jwtAuthBackend) pathConfigRead(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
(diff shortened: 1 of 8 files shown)


DrDaveD:
This now includes docs for direct callback mode but not yet device callback mode.

DrDaveD:
@cipherboy This now includes the needed documentation changes and I have removed the Draft status

builtin/credential/jwt/path_config.go
```diff
@@ -163,6 +167,91 @@ func (b *jwtAuthBackend) config(ctx context.Context, s logical.Storage) (*jwtCon
...
+func contactIssuer(ctx context.Context, uri string, data *url.Values, ignoreBad bool) ([]byte, error) {
+	var req *http.Request
+	var err error
+	if data == nil {
+		req, err = http.NewRequest("GET", uri, nil)
+	} else {
+		req, err = http.NewRequest("POST", uri, strings.NewReader(data.Encode()))
+	}
+	if err != nil {
+		return nil, nil
```
cipherboy, line 179: I'm curious why we're throwing away all of these error codes?

builtin/credential/jwt/path_config.go
```diff
@@ -163,6 +167,91 @@ func (b *jwtAuthBackend) config(ctx context.Context, s logical.Storage) (*jwtCon
...
+
+	if resp.StatusCode != http.StatusOK && (!ignoreBad || resp.StatusCode != http.StatusBadRequest) {
+		return nil, fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	return body, nil
+}
+
+// Discover the device_authorization_endpoint URL and store it in the config
+// This should be in coreos/go-oidc but they don't yet support device flow
```
cipherboy, line 208: Does https://github.com/coreos/go-oidc/pull/365 do what you want? I'm not familiar enough with the ecosystem to know, but my quick read is that it does. Perhaps we just need to bump the `go-oidc` version in OpenBao?

builtin/credential/jwt/path_oidc.go
```diff
@@ -240,23 +243,28 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
 		return logical.ErrorResponse(errLoginFailed + " Expired or missing OAuth state."), nil
 	}
 
+	deleteRequest := true
+	defer func() {
```
cipherboy, line 247: Nice!

Should we have timeout/expiry on these and clean them as part of the [periodic function](https://github.com/openbao/openbao/blob/b7526c7f719b0dec0cc7c9cc9b12fbc211fe57fa/sdk/framework/backend.go#L63-L71)? Or is there some other check we need to do to ensure the OIDC request is still valid / isn't resuming an old request that should've expired?

builtin/credential/jwt/path_oidc.go
```diff
@@ -288,8 +296,13 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
 		return nil, errwrap.Wrapf("error getting provider for login operation: {{err}}", err)
 	}
 
-	var rawToken oidc.IDToken
+	oidcCtx, err := b.createCAContext(ctx, config.OIDCDiscoveryCAPEM)
```
cipherboy, line 299: Out of curiosity, given we keep the same timeouts &c, why not reuse `ctx` here and simply add the value to the existing context we use elsewhere?

Is this an existing pattern that I'm missing?

cipherboy: I guess the issue is what happens when we cancel a context? Would we want to cancel through to `caCtx` (or if it was cancelled separately, would we want to cancel the parent too)?

cipherboy (commented):
Sorry about the delay in reviewing -- some comments for you!

builtin/credential/jwt/path_config.go
DrDaveD: Good catch, that looks like a bug.

builtin/credential/jwt/path_config.go
DrDaveD: That does indeed look promising, a feature added pretty recently. I will look into it.

builtin/credential/jwt/path_oidc.go
DrDaveD: That's a good observation.  I had to dig into it.  It turns out that the requests are stored as a cache implemented by https://github.com/patrickmn/go-cache which automatically cleans items out after they expire.  If a poll is successfully completed, they get deleted earlier in the poll function.  So I think this is OK.

builtin/credential/jwt/path_oidc.go
DrDaveD: I think it could be done that way but on the other hand it's probably a small amount of overhead to create and tear down a context, especially compared with everything else that's happening, so it's probably not worth it to cache the context.

builtin/credential/jwt/path_config.go
DrDaveD: So go-oidc is used only from https://github.com/hashicorp/cap.  The latest tagged version of that does use a new enough version of go-oidc, but it does not expose the DeviceAuthURL information to the caller of that library.  I'm trying to test to see if the addition of an `Endpoint()` function that calls the go-oidc [`Endpoint()`](https://github.com/coreos/go-oidc/blob/v3/oidc/oidc.go#L282) function will be sufficient, and if so I plan to try to submit that as a PR.  Perhaps openbao should be forking that module as well.

DrDaveD: That works, so I made hashicorp/cap#139

cipherboy: Thanks @DrDaveD! If they don't merge it, we can think about forking. But I'd like to give them the chance to keep the dependencies the same :-)

builtin/credential/jwt/path_oidc.go
cipherboy: Yeah, fair.

builtin/credential/jwt/path_oidc.go
cipherboy: Ah, smart. Sorry, I had thought this was backed by storage but it isn't.

That is curious, do you run a HA cluster and have you noticed any OIDC auth failures when the cluster changes leadership at all?

builtin/credential/jwt/path_config.go
DrDaveD: Meanwhile how about if we use the current implementation and optimize it later after cap is updated and has a new tag?

builtin/credential/jwt/path_oidc.go
DrDaveD: Another very good observation.  I do run HA clusters but I don't think I explicitly tried switching leadership in the middle of an OIDC authentication.  I tried it now and indeed it gets a failure of `Vault login failed. Expired or missing OAuth state.`

But since oidc authentications are manual operations and HA cluster leadership switches are pretty rare, I think it's an acceptable behavior.  People will just retry if it fails.  I did not introduce the use of the go-cache, that has been the implementation at least before I started looking at it..

cipherboy: Right, perfect.

Thanks @DrDaveD!

builtin/credential/jwt/path_config.go
cipherboy: @DrDaveD We can pull the updated cap without waiting for a new tag. Let's give it a bit, while it'd be nice to merge this, I tagged it for 2.1.0 and we'll still need to merge transactional storage before 2.1.0 will be release ready...

DrDaveD: hashicorp/cap#139 has been replaced by hashicorp/cap#140 after feedback from the developers.

DrDaveD:
@cipherboy Is there anything remaining for met to do on this one?

cipherboy (approved):
Sorry it took so long for me to review it! Merging. :-)

PR #320, opened 2024-05-03
Add oauth2_metadata config option

DrDaveD:
This PR adds a configuration option `oauth2_metadata` which is a list of token types that can be returned in metadata.  The list can be any of `access_token`, `id_token`, and `refresh_token`, and when selected they are returned in metadata names `oauth2_access_token`, `oauth2_id_token`, and `oauth2_refresh_token`, respectively.

This option makes it possible for a user to authenticate with bao once and also use the tokens directly from the oauth2 issuer for another purpose.   For example, with some help from a client application to store the refresh token back into a vault secrets plugin like the [Puppet labs oauthapp plugin](https://github.com/puppetlabs/vault-plugin-secrets-oauthapp), from then on access tokens can be read from the secrets plugin and renewed when they expire. 

The new option simply puts the selected tokens into the returned metadata in a similarly to the way that the `claims_mapping` option returns things in metadata.

This PR was proposed for vault several years ago in hashicorp/vault-plugin-auth-jwt#119 but has been sitting unmerged since then.  It is an essential component of the https://github.com/fermitools/htvault-config package and has been in production use for a couple of years.

### DIFF
--- a/builtin/credential/jwt/path_role.go
+++ b/builtin/credential/jwt/path_role.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
@@ -148,6 +149,10 @@ Defaults to 60 (1 minute) if set to 0 and can be disabled if set to -1.`,
 				Type:        framework.TypeKVPairs,
 				Description: `Mappings of claims (key) that will be copied to a metadata field (value)`,
 			},
+			"oauth2_metadata": {
+				Type:        framework.TypeCommaStringSlice,
+				Description: `Comma-separated list of one or more of access_token, id_token, refresh_token to return in metadata`,
+			},
 			"user_claim": {
 				Type:        framework.TypeString,
 				Description: `The claim to use for the Identity entity alias name`,
@@ -238,6 +243,7 @@ type jwtRole struct {
 	BoundClaimsType      string                 `json:"bound_claims_type"`
 	BoundClaims          map[string]interface{} `json:"bound_claims"`
 	ClaimMappings        map[string]string      `json:"claim_mappings"`
+	Oauth2Metadata       []string               `json:"oauth2_metadata"`
 	UserClaim            string                 `json:"user_claim"`
 	GroupsClaim          string                 `json:"groups_claim"`
 	OIDCScopes           []string               `json:"oidc_scopes"`
@@ -354,6 +360,7 @@ func (b *jwtAuthBackend) pathRoleRead(ctx context.Context, req *logical.Request,
 		"bound_claims_type":       role.BoundClaimsType,
 		"bound_claims":            role.BoundClaims,
 		"claim_mappings":          role.ClaimMappings,
+		"oauth2_metadata":         role.Oauth2Metadata,
 		"user_claim":              role.UserClaim,
 		"user_claim_json_pointer": role.UserClaimJSONPointer,
 		"groups_claim":            role.GroupsClaim,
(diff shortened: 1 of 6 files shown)


DrDaveD:
@cipherboy If this can be merged along with #318 and #319 that would be awesome.

cipherboy:
@DrDaveD Are these two (#319) independent?

DrDaveD:
Yes they are independent.  It's just that my application needs both.

changelog/320.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
cipherboy, line 1:
```suggestion
```release-note:improvement
```

I don't know this is big enough to be considered a top-level feature (unlike the other two, which I think qualify better), but let me know what you think.

builtin/credential/jwt/path_role.go
```diff
@@ -561,6 +567,10 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
 		role.ClaimMappings = claimMappings
 	}
 
+	if oauth2Metadata, ok := data.GetOk("oauth2_metadata"); ok {
+		role.Oauth2Metadata = oauth2Metadata.([]string)
```
cipherboy, line 571: We should probably validate values here and prevent duplicates.

cipherboy (commented):
Two minor comments.

I think I'm happy with this change, but I'd like to see it called out better in the docs as something to be mindful of: previously an OpenBao OIDC-bound token could not be used to pivot to other systems natively (if a token was compromised), but with this in the metadata, it could potentially be used to pivot now (depending on claims?).

(That is, assuming I'm correct about the former, which I _believe_ is correct, but might be wrong about the existing client-bound OIDC workflow that existed in upstream -- I don't believe it currently sees either of these tokens and they're internal to upstream/OpenBao even in that workflow, but my understanding could be wrong.)

changelog/320.txt
DrDaveD: I'm fine with that.

builtin/credential/jwt/path_role.go
DrDaveD: The names are validated and duplicates get removed in [path_oidc.go](https://github.com/openbao/openbao/blob/85655b829596ca8d2ab7c0fa33fc4a1b6c5bec1e/builtin/credential/jwt/path_oidc.go#L384) so I think that's fine.  Do you agree?

DrDaveD:
I added a caution to the web documentation.

builtin/credential/jwt/path_role.go
cipherboy: I agree, but doesn't that validate it at the time of use, potentially causing login to fail even though the role config would've been misleadingly successful?

Whereas if we validate it here, we prevent login from failing due to bad values and would fail the role config update, making it clear earlier? 

My 2c!

DrDaveD: Good point.  I put in an extra check.  Duplicates are not a problem.

cipherboy (approved):
Looks great, thanks @DrDaveD for the changes!

cipherboy:
Looks like there's a flake in a different test with the data race detector:

```go
2024-08-13T22:35:33.2621062Z === [31mFailed[0m
2024-08-13T22:35:33.2622304Z === [31mFAIL[0m: vault TestOIDC_PeriodicFunc/test-key-nil-next-signing-key (4.62s)
...
2024-08-13T22:35:33.3387975Z     identity_store_oidc_test.go:1165: For key: test-key-nil-next-signing-key at cycle: 2 expected namedKey's KeyRing to be at least of length 3 but was: 2
2024-08-13T22:35:33.3389873Z     identity_store_oidc_test.go:1176: For key: test-key-nil-next-signing-key at cycle: 2 expected public keys to be at least of length 3 but was: 2
```

PR #321, opened 2024-05-04
Fix compatibility with Vault plugins

cipherboy:
This adds Upstream (`VAULT_` prefixed) variants of all OpenBao-set variables during plugin startup, allowing plugins compiled for upstream to work with OpenBao.

This should resolve #317 (a discussion).

@DrDaveD Let me know if this works for you!

### DIFF
--- a/api/env.go
+++ b/api/env.go
@@ -5,8 +5,22 @@ import (
 	"strings"
 )
 
+const (
+	OpenBaoEnvPrefix  = "BAO_"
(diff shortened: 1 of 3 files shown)


DrDaveD:
Yes it does

JanMa (approved):

cipherboy:
This should be ready now that I added a changelog entry @naphelps :-)

naphelps (approved):

PR #333, opened 2024-05-08
Complete the conversion of jwt plugin html responses to OpenBao

DrDaveD:
I noticed the Vault logo and other references to Vault and HashiCorp were still in the jwt plugin's cli responses.  This replaces those with OpenBao.


JanMa:
Hey @DrDaveD,
could you post a screenshot of the difference or some steps on how to produce the cli responses?
Thanks in advance :-)

cipherboy:
> could you post a screenshot of the difference or some steps on how to produce the cli responses?

Maybe my copy/paste was bad, but my SVG viewer couldn't render the SVG when I copied it out-of-tree. Hmmm

DrDaveD:
I tested it by taking the svg code and putting it into a .html file surrounded by `<body> ... </body>`.   Then I read it into my web browser (firefox) using a `file:///` url. 

It's somewhat hard to reproduce with the vault client api.  You need to configure the jwt plugin with a token issuer and make sure it allows a callback URL to the vault client including the port that it chooses.  I tested it in combination with my pr #318 and got the following screenshot:
Screenshot 2024-05-09 at 1 10 16 PM

Oh now I see that both the links are to the same thing.  I should probably reduce the redundancy there.

DrDaveD:
It works well to copy each of the 3 entire strings into .html files and reading them into your browser.   For instance here's now a screenshot of the success string: 
Screenshot 2024-05-09 at 1 40 17 PM

JanMa (approved):
Awesome. Thanks a lot @DrDaveD

naphelps (approved):

PR #348, opened 2024-05-29
Pass `BAO_ADDR` to the token helper

### LINKED ISSUE #314: `BAO_ADDR` is not propagated to the token helper when provided as `-address`
**Describe the bug**

The token helper might inherit `BAO_ADDR` from the environment, but if the address is configured through the `-address` flag instead, then the token helper has no way of knowing about the address.

**To Reproduce**

1. Set a program that prints `BAO_ADDR` and then fails as the token helper in `~/.bao`.
2. Run `BAO_ADDR=foo bao login`, observe it gets the `BAO_ADDR`.
3. Run `bao login -address foo`, observe it does not get `BAO_ADDR`.

**Expected behavior**
The token helper needs to know the address of the instance it’s connecting to, so it can store a token per instance, that is the point of the token helper. Therefore, `bao` needs to set `BAO_ADDR` in the environment of the token helper it executes, and not rely on it being present implicitly.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): irrelevant
* OpenBao CLI Version (retrieve with `bao version`): `OpenBao v2.0.0-alpha20240329 ('74c2dddb0612b9a3da79384c20638266aa7de407'), built 2024-04-26T10:19:19Z`
* Server Operating System/Architecture: Linux 6.8.7 / x86-64

OpenBao server configuration file(s): irrelevant

**Additional context**

This bug report is an adaptation of https://github.com/hashicorp/vault/issues/22257.

ruuda:
Previously the token helper might inherit `BAO_ADDR` from the process, but if the address was specified through an `-address` command-line flag, then the token helper would not know the address, or it would use the wrong one. Fix that by propagating the address everywhere, and then setting `BAO_ADDR` explicitly in the token helper's environment.

Resolves #314.

* [x] One thing I’m not sure about: should we set `VAULT_ADDR` in addition to `BAO_ADDR`? I see there are still many uses of `VAULT_ADDR` in the codebase. ⇒ Yes, set both.

## Target Release

It filled out this by default: 1.14.7

### DIFF
--- a/command/config/util.go
+++ b/command/config/util.go
@@ -9,7 +9,7 @@ import (
 
 // DefaultTokenHelper returns the token helper that is configured for Vault.
 // This helper should only be used for non-server CLI commands.
-func DefaultTokenHelper() (token.TokenHelper, error) {
+func DefaultTokenHelper(vaultAddr string) (token.TokenHelper, error) {
 	config, err := LoadConfig("")
 	if err != nil {
 		return nil, err
@@ -24,5 +24,12 @@ func DefaultTokenHelper() (token.TokenHelper, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &token.ExternalTokenHelper{BinaryPath: path}, nil
+
+	// If the user specifed the address to connect to on the command line instead
+	// of through an environment variable, we propagate the address to the token
+	// helper through an environment variable. Otherwise the token helper may
+	// read BAO_ADDR and assume a different address than the one we are using.
+	env := []string{"BAO_ADDR=" + vaultAddr}
+
+	return &token.ExternalTokenHelper{BinaryPath: path, Env: env}, nil
 }
--- a/command/config/util.go
+++ b/command/config/util.go
@@ -29,7 +29,7 @@ func DefaultTokenHelper(vaultAddr string) (token.TokenHelper, error) {
 	// of through an environment variable, we propagate the address to the token
 	// helper through an environment variable. Otherwise the token helper may
 	// read BAO_ADDR and assume a different address than the one we are using.
-	env := []string{"BAO_ADDR=" + vaultAddr}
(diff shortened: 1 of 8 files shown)


command/config/util.go
```diff
@@ -24,5 +24,12 @@ func DefaultTokenHelper() (token.TokenHelper, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &token.ExternalTokenHelper{BinaryPath: path}, nil
+
+	// If the user specifed the address to connect to on the command line instead
+	// of through an environment variable, we propagate the address to the token
+	// helper through an environment variable. Otherwise the token helper may
+	// read BAO_ADDR and assume a different address than the one we are using.
+	env := []string{"BAO_ADDR=" + vaultAddr}
```
cipherboy, line 32: I'd copy from #321 and set `VAULT_ADDR` as well. :-)

cipherboy:
\o hello @ruuda -- I think there's some test failures introduced by this PR. If I rebuild this PR on top of my #362, I think in `command/login_test.go` there's several places where `cmd.TokenHelper()` is called without the `vaultAddr` string. 

Sorry for the noisy test failures, I hadn't realized some of my PRs had been merged without passing tests :o

cipherboy:
\o hey @ruuda! Now that GA has been released and we're looking at merging more things, do you mind rebasing this and taking a look at the test failures?

If you can't, just let me know and I'm happy to pick it up.

ruuda:
Thanks for the reminder! I rebased this and I updated the tests to pass in a dummy address. As far as I can tell none of the tests are exercising the process spawning, so this seems like the easiest way to fix them, though it feels a bit ugly.

cipherboy (approved):
This looks like a great improvement, thanks @ruuda! 

In the future, it could be cool to expose multi-node token management in the internal helper as well :-) But an improvement for later!

PR #351, opened 2024-06-01
Clarify Integrated Storage is only backend

IohannesArnold:
The Internals / Integrated Storage page contained legacy Vault information about supporting "several storage options". This commit makes clear that Integrated Storage is the only storage backend in OpenBao.


cipherboy (approved):

JanMa (approved):
LGTM, thanks @IohannesArnold

Issue #354, opened 2024-06-04
RFC - Changing OpenBao's usage of mlock


IohannesArnold:
### Summary

To prevent writing sensitive information to disk through swap, Vault used the mlock syscall to prevent any process memory from being swapped out to disk. This never worked perfectly, and while OpenBao has so far retained Vault's usage of mlock, other design decisions made since the fork call the entire mlock design into question.

### Problem Statement

[`mlock` is a Posix syscall that will lock some or all of a process's virtual memory in physical memory and prevent it from being paged out to disk](https://www.man7.org/linux/man-pages/man2/mlock.2.html). This provides a security advantage for applications like OpenBao, which deal with sensitive information. By using `mlock`, OpenBao can ensure that sensitive information stays in memory, instead requiring the deploying sysadmin to securely overwrite, encrypt, or disable swap. For this reason, Vault used `mlock` where possible, and OpenBao has inherited this code so far.

However, for various reasons, discussed in more detail later, `mlock` can not be used in all situations. [It can only be used on Linux and the BSDs](https://github.com/openbao/go-secure-stdlib/blob/main/mlock/mlock_unix.go), [not Windows, macOS, or other platforms](https://github.com/openbao/go-secure-stdlib/blob/main/mlock/mlock_unavail.go). It will also [lead to a bug if it is used with Integrated Storage](https://developer.hashicorp.com/vault/tutorials/operations/performance-tuning#mlock), so while enabling `mlock` usage on Linux with Integrated Storage is _possible_, [it is discouraged by Vault](https://developer.hashicorp.com/vault/docs/configuration#disable_mlock), and so far [this has been retained by OpenBao](https://openbao.org/docs/configuration/#parameters).

As part of the fork, [all external storage options were removed from OpenBao](https://github.com/orgs/openbao/discussions/64), leaving the Integrated Storage as the sole production-quality storage backend. As a result, in the current state of the codebase, there are **zero** production-level configurations where it is advisable to have `mlock` enabled. Nevertheless, `mlock` is _enabled_ by default on those platforms which support it. This seems less than ideal, and the purpose of this RFC is to discuss how OpenBao's usage of `mlock` might evolve in a more elegant direction.

### User-facing description

Configuring `mlock` is only relevant to sysadmins maintaining the OpenBao server process. Since it is enabled on Linux by default, they must read the docs (currently rather unclear on this point) to learn that they should disable it. Disabling it requires a single line in the server config.

[...]

PR #362, opened 2024-06-13
Fix go tests

cipherboy:
This fixes Go tests broken by #321 that were not caught prior to merging (plugin compatibility, external binary testing panic caused by reverting SSCT removal, an issue with the Caddy EAB test, and a stray log statement post-test that Go now panics on).

Also removes the obsolete upstream build process and fixes the missing UI page to mention OpenBao.

PR #363, opened 2024-06-14
Remove mlock and replace with cgroups

IohannesArnold:
In [RFC #354](https://github.com/openbao/openbao/issues/354), the mlock implementation inherited from Vault was deemed buggy. Here it is ripped out of all core OpenBao code. A few stubs are retained for compatibility's sake:

1. The config file parser will still parse the setting "disable_mlock". It will do nothing when set to true, and it will immediately error if set to false (i.e. the user is explicitly expecting mlock to be enabled).
2. The dynamicSystemView struct has a MlockEnabled method so it can still implement pluginutil.RunnerUtil. This method now just returns false.

All mlock code is RETAINED in all ./sdk files, because the question is not yet settled whether plugins built against the *OpenBao SDK* should be binary-compatible with Vault. If this is eventually resolved in the negative, most of the mlock related code in the SDK can be subbed out.

As mlock is no longer used, Docker-related scripts have also had setcap calls removed.

In place of mlock, documentation has been added to draw attention to the danger of sensitive information leaking through swap space and stress the importance of disabling or encrypting swap on any platform, or on Linux, changing the cgroupv2 setting memory.swap.max to 0. This last option has also been included in the example systemd service file.

### DIFF
--- a/.release/linux/package/usr/lib/systemd/system/openbao.service
+++ b/.release/linux/package/usr/lib/systemd/system/openbao.service
@@ -17,8 +17,7 @@ ProtectHome=read-only
 PrivateTmp=yes
 PrivateDevices=yes
 SecureBits=keep-caps
-AmbientCapabilities=CAP_IPC_LOCK
-CapabilityBoundingSet=CAP_SYSLOG CAP_IPC_LOCK
+CapabilityBoundingSet=CAP_SYSLOG
 NoNewPrivileges=yes
 ExecStart=/usr/bin/bao server -config=/etc/openbao/openbao.hcl
 ExecReload=/bin/kill --signal HUP $MAINPID
@@ -28,7 +27,7 @@ Restart=on-failure
 RestartSec=5
 TimeoutStopSec=30
 LimitNOFILE=65536
-LimitMEMLOCK=infinity
+MemorySwapMax=0
 
 [Install]
 WantedBy=multi-user.target
--- a/changelog/363.txt
+++ b/changelog/363.txt
@@ -1,0 +1,3 @@
+```release-note:change
+core: Remove mlock functionality from OpenBao and make the "disable_mlock" config option obsolete.
+```
--- a/command/server/test-fixtures/config_custom_response_headers_multiple_listeners.hcl
+++ b/command/server/test-fixtures/config_custom_response_headers_multiple_listeners.hcl
@@ -56,4 +56,3 @@ listener "tcp" {
 }
 
 
-disable_mlock = true
--- a/internalshared/configutil/config.go
+++ b/internalshared/configutil/config.go
@@ -13,6 +13,11 @@ import (
 	"github.com/hashicorp/hcl/hcl/token"
 )
 
+const mlockMsg = "OpenBao has dropped support for mlock. Please remove\n" +
+	"the line \"disable_mlock\" = false from your config and disable\n" +
+	"or encrypt swap instead. For more information, see:\n" +
+	"https://openbao.org/docs/install/#Post-installation-hardening"
+
 // SharedConfig contains some shared values
 type SharedConfig struct {
 	FoundKeys  []string     `hcl:",decodedFields"`
@@ -25,7 +30,8 @@ type SharedConfig struct {
 
 	Seals []*KMS `hcl:"-"`
 
-	DisableMlock    bool        `hcl:"-"`
+	// mlock is no longer used by OpenBao, but this is kept as a config option for
+	// compatibility's sake and to give a warning for those expecting it.
 	DisableMlockRaw interface{} `hcl:"disable_mlock"`
 
 	Telemetry *Telemetry `hcl:"telemetry"`
@@ -75,10 +81,12 @@ func ParseConfig(d string) (*SharedConfig, error) {
 	}
 
 	if result.DisableMlockRaw != nil {
-		if result.DisableMlock, err = parseutil.ParseBool(result.DisableMlockRaw); err != nil {
-			return nil, err
+		isDisabled, err := parseutil.ParseBool(result.DisableMlockRaw)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing 'disable_mlock': %w", err)
+		} else if !isDisabled {
+			return nil, fmt.Errorf(mlockMsg)
 		}
-		result.FoundKeys = append(result.FoundKeys, "DisableMlock")
 		result.DisableMlockRaw = nil
 	}
 
@@ -146,7 +154,6 @@ func (c *SharedConfig) Sanitized() map[string]interface{} {
 
 	result := map[string]interface{}{
 		"default_max_request_duration":  c.DefaultMaxRequestDuration,
-		"disable_mlock":                 c.DisableMlock,
 		"log_level":                     c.LogLevel,
 		"log_format":                    c.LogFormat,
 		"pid_file":                      c.PidFile,
--- a/vault/dynamic_system_view.go
+++ b/vault/dynamic_system_view.go
@@ -274,9 +274,9 @@ func (d dynamicSystemView) ListVersionedPlugins(ctx context.Context, pluginType 
 	return d.core.pluginCatalog.ListVersionedPlugins(ctx, pluginType)
 }
 
-// MlockEnabled returns the configuration setting for enabling mlock on plugins.
+// OpenBao no longer uses mlock but MlockEnabled is retained for plugin compatibility.
 func (d dynamicSystemView) MlockEnabled() bool {
-	return d.core.enableMlock
+	return false
 }
 
 func (d dynamicSystemView) EntityInfo(entityID string) (*logical.Entity, error) {
--- a/website/content/docs/install.mdx
+++ b/website/content/docs/install.mdx
@@ -82,3 +82,56 @@ on your `PATH` to avoid receiving an error that OpenBao is not found.
 ```shell-session
 $ bao -h
 ```
+
+## Post-installation hardening
+
+After installing OpenBao, you may want to take additional steps to secure it
+against leaking your secrets. OpenBao normally does this very well, but there
+is an operating system feature that undermines OpenBao's protection. This is
+[memory paging (aka swap)](https://en.wikipedia.org/wiki/Memory_paging). To
+provide extra security, you will want to make sure that your OS has swap
+disabled or that its swap space is encrypted.
+
+### Linux
+
+The example systemd service file provided with the OpenBao source code comes
+configured to disable swap for the OpenBao process. To verify that swap is
+disabled, run `systemctl cat openbao` and check for the line `MemorySwapMax=0`.
+Alternatively, to allow the openbao process to swap out, make sure that line is
+deleted.
+
+If you are not using systemd, you can achieve the same effect by setting the 
+cgroupv2 value `memory.swap.max` to `0` using your tool of choice. You can
+disable swap for the entire OS by running `swapoff` (this is not recommended).
+Encrypting swap space in Linux is possible, but as usual, there are many
+options, and a guide is outside the scope of these docs. Consult your distro's
+documentation.
+
+### BSDs and other Unix-like
+
+It is recommended to confirm that swap is encrypted. This can be done on all the
+major BSDs.
+ * [FreeBSD guide to encrypted swap](https://docs.freebsd.org/en/books/handbook/disks/#swap-encrypting)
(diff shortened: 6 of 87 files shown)


IohannesArnold:
Okay, I'm _now_ ready for review and merging. The bao binary still builds; since not all the tests passed before, I didn't learn how to get the whole test suite running, and I'm not sure if I broke anything more.

Many of these deleted lines are trivial; for review, the more substantial edits (if you trust me) are:
* [.release/linux/package/usr/lib/systemd/system/openbao.service](https://github.com/openbao/openbao/pull/363/files#diff-373a0c654a97c4ae9d1a586fa0ee9145a171c90e3231f48432eb4668c2d2ae66)
* [command/server.go](https://github.com/openbao/openbao/pull/363/files#diff-de23c936a9ff46fc796a96721e088734a0d6281da4f7c0fe19c3ca6fc86c5dc4)
* [internalshared/configutil/config.go](https://github.com/openbao/openbao/pull/363/files#diff-68a76ba561b15a5db5b5e1d16e31883452494648f61ad152322ad190ba933b11)
* [vault/core.go](https://github.com/openbao/openbao/pull/363/files#diff-a0a40b411f03b3c927d6a2eafaf8a17693cc05a44e81bfe1362045a9372b9cfe)
* [vault/dynamic_system_view.go](https://github.com/openbao/openbao/pull/363/files#diff-321fc062fce4a794ac87052c0863ebd4177ac8e21d9cb1b718bf21c1851b5feb)
* [website/content/docs/install.mdx](https://github.com/openbao/openbao/pull/363/files#diff-4687f91265f93fe46f2b0041ca5c6a9feee41c051a70d10de17a5cdd829c997a)
---
For the last of these, the website docs addition to discussion swap disabling or encryption on various OS's, I added a "Post-installation hardening" section to the "Installing OpenBao" page. I think this leaves something to be desired, but I think trying to do any better would lead to a more general reorganization of the docs. So I include my less-than-fully-satisfactory docs in this PR. Bigger changes to the docs are now on my mind for another one. :wink:

cipherboy:
@IohannesArnold I think #362 should fix the tests; I'm tracking down the last issue with the race tests, but having a hard time reproducing it locally... If you'll rebase after that lands we can probably get passing tests here :-)

.release/linux/package/usr/lib/systemd/system/openbao.service
```diff
@@ -28,7 +27,7 @@ Restart=on-failure
 RestartSec=5
 TimeoutStopSec=30
 LimitNOFILE=65536
-LimitMEMLOCK=infinity
+MemorySwapMax=0
```
cipherboy, line 30: So this would only work for native (non-containerized) installation... For the latter, would we recommend running with `--memory-swappiness=0` on the container execution CLI? [IIUC](https://docs.docker.com/config/containers/resource_constraints/#--memory-swappiness-details) that seems to do what we want? 

If so, we'd probably want to mention it in the various places where `mlock` was previously mentioned.

command/server/test-fixtures/config_custom_response_headers_multiple_listeners.hcl
```diff
@@ -56,4 +56,3 @@ listener "tcp" {
 }
 
 
-disable_mlock = true
```
cipherboy, line 59: Neither here nor there, just a note mostly to myself... We should probably split a bunch of these tests to use `disable_cache=false` and make sure we exercise both paths for tests which do benefit from testing and not testing caching.

internalshared/configutil/config.go
```diff
@@ -75,10 +76,12 @@ func ParseConfig(d string) (*SharedConfig, error) {
 	}
 
 	if result.DisableMlockRaw != nil {
-		if result.DisableMlock, err = parseutil.ParseBool(result.DisableMlockRaw); err != nil {
-			return nil, err
+		isDisabled, err := parseutil.ParseBool(result.DisableMlockRaw)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing 'disable_mlock': %w", err)
+		} else if !isDisabled {
+			return nil, fmt.Errorf("OpenBao does not support mlock; please disable and use CGroups instead")
```
cipherboy, line 83: Probably here as well I'd recommend the option.

Do we want to just always err in this block? I don't _really_ see the value in even attempting to parse the boolean? My 2c.?

vault/dynamic_system_view.go
```diff
@@ -274,9 +274,9 @@ func (d dynamicSystemView) ListVersionedPlugins(ctx context.Context, pluginType
 	return d.core.pluginCatalog.ListVersionedPlugins(ctx, pluginType)
 }
 
-// MlockEnabled returns the configuration setting for enabling mlock on plugins.
+// OpenBao no longer uses mlock but MlockEnabled is retained for plugin compatibility.
```
cipherboy, line 277: I think this is the right approach; I don't see any callers of this function but since it is defined in the interfaces, we can't remove it.

cipherboy (commented):
This looks fairly good, thank you @IohannesArnold!

Some comments inline. What do you think about controlling cgroups directly in the binary? Would this work inside a container or not easily? Thinking something like https://pkg.go.dev/github.com/containerd/cgroups or https://pkg.go.dev/github.com/opencontainers/runc/libcontainer/cgroups, though I've not looked into how to use either. Was wondering if there might be more useful options to constrain external plugins with...

vault/dynamic_system_view.go
IohannesArnold: Yes, I tried ripping it out and OpenBao then failed to compile because `dynamicSystemView` wasn't implementing a part of the `pluginutil.RunnerUtil` interface.

internalshared/configutil/config.go
IohannesArnold: IMO this really depends on how much OpenBao wants to be a (near) drop-in replacement for Vault 1.14. My thinking was that if `disable_mlock` was absent in a config, that meant the user didn't really consider the issue, so we're fine to do something different. And if mlock was explicitly disabled, we're also fine. But if a user explicitly added a configuration option to say that mlock was enabled, which was the default in Vault anyway, then OpenBao should error out and bring this change in behavior to this user's attention.

cipherboy: Ah I see, this is an err. I'd probably warn always (if not erring), and if this is enabled, then error.

However, note that [by the RFC](https://openbao.org/docs/policies/migration/#proposal), we're only aiming for API compatibility.

.release/linux/package/usr/lib/systemd/system/openbao.service
IohannesArnold: Oh yeah, I didn't think about the Docker case... That flag looks right. I'll add it to the "Post-installation hardening" section. If you're talking about the other places in the docs where mlock was mentioned, I don't want to replace them in this commit, because frankly I thought they were a bit random. I will work on a bigger docs reorg, and maybe add more warnings about swap then.

cipherboy: Ah, I was thinking about perhaps in the container entrypoint scripts/Dockerfiles?

IohannesArnold:
> controlling cgroups directly in the binary

I don't have any experience with this either, so I can't say much.

My gut reaction is that swap is really an OS feature, so considering how to secure it should be an OS concern, not the concern of a process. I was surprised how easy it was to find ways of encrypting swap for Windows and macOS.

.release/linux/package/usr/lib/systemd/system/openbao.service
IohannesArnold: I don't think you can specify this on the Dockerfile, nor in the entrypoint script _within the container_. Do we provide a docker-compose template? If we did we could do it there...

internalshared/configutil/config.go
IohannesArnold: What would be the idiomatic way to warn in `internalshared/configutil`? I don't see any other warnings currently in that file, or in the files that call `configutil.ParseConfig`.

.release/linux/package/usr/lib/systemd/system/openbao.service
cipherboy: Oh, sorry, no, but we had commentary e.g. here:

https://github.com/openbao/openbao/blob/ee45b4d893832f0184c669f65ae8eed125548f4f/Dockerfile#L63-L64

that mentions mlock, so I was thinking perhaps suggesting that the caller use these options outside of the containerfile?

internalshared/configutil/config.go
cipherboy: Well, I guess the logger isn't setup at this point, so the short answer is there is none :D 

I think that means it our options are to treat it like an error or ignore it... Thoughts? I guess erring only if set to true is probably the kinder thing to do, so let's leave this as is.

website/content/docs/install.mdx
JanMa, line 1: We provide OpenBao binaries for FreeBSD, OpenBSD and NetBSD who all support the relevant `mlockall` syscall. It would be nice if we also provide subsections on how to disable swap for the various BSDs here

internalshared/configutil/config.go
JanMa, line 83: Small nitpick with the error message: CGroups are a Linux only feature, but mlock is also supported by the various BSDs we build binaries for. In light of this I'd rather ask the user to disable swap on their system

JanMa (changes requested):
Looking very good overall, thanks @IohannesArnold. I have two smaller remarks wrt. the BSD targets we support. Also I see this as a potential breaking change. Please provide a `changelog` entry and mark it as such

IohannesArnold:
Thanks @JanMa -- I've rebased against main HEAD (although it's since drifted again), got all the tests to pass, and believe I have addressed the issues that you and Alex have raised. I think this should be good to go for final approval.

changelog/363.txt
```diff
@@ -0,0 +1,3 @@
+```release-node:change
```
cipherboy, line 1:
```suggestion
```release-note:change
```

cipherboy:
@IohannesArnold Looks like `go.mod` changed as `mlock` is no longer a direct dependency. Do you mind running `make tidy-all`?

cipherboy (approved):
Looks good from my side, thanks @IohannesArnold!

internalshared/configutil/config.go
```diff
@@ -13,6 +13,11 @@ import (
 	"github.com/hashicorp/hcl/hcl/token"
 )
 
+const mlockMsg = "OpenBao has dropped support for mlock. Please remove\n" +
+	"the line \"disable_mlock\" = false from your and disable or\n" +
```
JanMa, line 17:
```suggestion
	"the line \"disable_mlock\" = false from your config and disable or\n" +
```

website/content/docs/install.mdx
```diff
@@ -82,3 +82,60 @@ on your `PATH` to avoid receiving an error that OpenBao is not found.
...
+disable swap for the entire OS by running `swapoff` (this is not recommended).
+Encrypting swap space in Linux is possible, but as usual, there are many
+options, and a guide is outside the scope of these docs. Consult your distro's
+documentation.
+
+### BSDs and other Unix-like
+
+It is recommended to confirm that swap is encrypted. This can be done on all the
+major BSDs.
+ * [FreeBSD guide to encrypted swap](https://docs.freebsd.org/en/books/handbook
```
JanMa, line 114: Markdown does not support linebreaks in a link target. If you take a look at the page preview in this PR, the links here are not properly rendered.

website/content/docs/install.mdx
```diff
@@ -82,3 +82,60 @@ on your `PATH` to avoid receiving an error that OpenBao is not found.
...
+```
+Then reboot.
+
+### Docker
+
+When running the Docker image, include the flag `--memory-swappiness=0`.
+
+### macOS
+
+[The swap space on macOS is always encrypted](https://support.apple.com/guide
```
JanMa, line 140: Here as well

JanMa (changes requested):
A few small fixes, but then you have my approval :-)

IohannesArnold:
@JanMa Commit-final-FINAL-v2-for-real :stuck_out_tongue_closed_eyes:

website/content/docs/install.mdx
JanMa: @IohannesArnold you missed this one, sorry :sweat_smile:

JanMa (changes requested):

IohannesArnold:
:melting_face:

JanMa (approved):
LGTM :heavy_check_mark: Thanks a lot @IohannesArnold

Issue #272, opened 2024-04-06
Remove last vestiges of Unified CRL


cipherboy:
See e.g., https://github.com/openbao/openbao/blob/main/builtin/logical/pki/path_fetch_issuers.go#L1214-L1223 -- and similar locations.

There's many places where cross-cluster Unified CRLs still exist in the code base, despite being a Vault Enterprise (and in particular, Performance Secondary) scenario. We should remove these, prior to GA.

PR #365, opened 2024-06-15
Remove unified CRLs, cross-cluster revocations

cipherboy:
In Vault Enterprise's split-storage clustering (Performance Secondary), the PKI engine needed a way of maintaining a single, unified CRL across all Perf Secondary clusters and accepting revocations given to any cluster for certs stored on any other cluster. The combination of Unified CRLs and Revocation Queues solved this problem for Vault Enterprise, but the code largely lived in Vault Community.
    
Since OpenBao lacks these clustering semantics and is unlikely to add them, remove this code.

Resolves: #272

---

Based on #362; will be rebased once that merges.


cipherboy:
Thanks @DanGhita -- there are extensive tests still left testing CRLs in `crl_test.go` and `backend_test.go`. :-)

PR #375, opened 2024-06-18
Fix artifact signing, use default runner

cipherboy:
GPG signing was broken as `--detach-sign` does not itself take an argument to a file to sign; instead this should be a separate positional argument to the CLI as a whole. This means that stdin was signed instead of the specified file, resulting in bogus signatures.

While the existing cosign signatures work, they require additional calls to rekor to fetch the corresponding certificate used to sign. Mirroring with what OpenTofu does, we can save the certificates directly so that users can verify without additional calls to the rekor network.

Lastly, switch to GitHub-hosted runners to avoid needing to use a self-hosted runner for this release stage.

Thanks to @JanMa and @janosdebugs for their help.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,7 @@ permissions:
 
 jobs:
   release:
-    runs-on: self-hosted
+    runs-on: ubuntu-latest
     env:
       DOCKER_CLI_EXPERIMENTAL: "enabled"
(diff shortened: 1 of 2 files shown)


JanMa (approved):

naphelps (approved):

DanGhita (commented):
OK for me

PR #378, opened 2024-06-19
fix release asset parsing for download page

JanMa:
We started to generate SBOM and signature files for our release assets
which are currently shown in our download page. This updates our parsing
logic to exclude them and to also handle the new uppercase naming
scheme.


cipherboy (approved):
Looks much better, thanks @JanMa!

PR #381, opened 2024-06-25
Downgrade test-summary/action from 2.3 to 2.2

JanMa:
We're running into this open Issue which seems to have been introduced
in version 2.3:
https://github.com/test-summary/action/issues/44


cipherboy (approved):
Thanks @JanMa!

PR #391, opened 2024-06-28
Update docs to include mlock removal RFC

IohannesArnold:
Adds the RFC to the docs and fixes a few related typos.

As per @cipherboy 's [request in the RFC issue](https://github.com/openbao/openbao/issues/354#issuecomment-2194753640)


cipherboy (commented):
Do you mind also adding this to the side bar? `/website/sidebars.ts` :-)

Thanks!

IohannesArnold:
Eventually I'll submit a PR that's right on the first try...

JanMa (approved):

JanMa:
Thanks @IohannesArnold

JanMa (changes requested):
I am very sorry @IohannesArnold but could you please sign your commit? :see_no_evil: 
We require this in our branch protection rules, so I can't merge your PR.

IohannesArnold:
Isn't it?
https://github.com/openbao/openbao/pull/391/commits/7ef7ac2c9d9998d647b9a37d2ac658c30f0de0f5
```
Adds the RFC to the docs and fixes a few related typos.

Signed-off-by: John Arnold <code@iohannes.us>
```

JanMa:
Sorry for the confusion, what I meant was that the commit must be signed with your GPG or SSH key, so it is shown as "verified".
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

cipherboy:
Yeah, apologies, this was my mistake when merging the last PR; I hadn't seen the missing commit signatures and we hadn't yet had the community call two weeks ago wherein Dan expressed interest in keeping signatures as a requirement.

IohannesArnold:
I'm not a major contributor here (yet*), so I don't expect that this opinion counts for much, but I'm not sure this policy does anything worthwhile. First, it looks like most of the commits going into main are via the "Squash and merge" functionality, which destroys the commit you got from me and replaces it with a new one created via Github, which is signed _by Github_. You can see that [my prior commit implementing the mlock RFC is "verified"](https://github.com/openbao/openbao/commit/ac477241a6497faf1a11e9926eb265456bce3c5f), although I didn't sign it, because Github itself signed the commit it created after "Squash and merge".

But even for the commit signed by my SSH key, the semantics of a verified sign are atemporal. [This gets into what I mean in more detail](https://lobi.to/writes/wacksigning/#gpg-keys), but in brief, even if this commit is verified now, if I ever delete my SSH key from Github's UI, my previously "verified" commit (which, just to repeat, will not be in the main branch but only in the history of the PR, assuming you continue to squash and merge) will become "unverified". I was in the habit of deleting old unused SSH keys from Github; now I'll have to remember not to delete this one ever -- unless I think it's been compromised, in which case I'll have to delete it, and this "verified" commit become "unverified".

But don't take me the wrong way, I'm not getting angry about this. I've gone ahead and signed it and pushed it as you've asked. I'm just expressing my doubt that it means much. Take this as you will.

*maybe

JanMa (approved):

cipherboy:
> I'm not sure this policy does anything worthwhile.

FWIW, I'm inclined to believe. The closest I've found to supporting this would be that if you reach critical mass of signed commits, you can then start excluding signatures from invalid keys (e.g., if someone pushed a commit under my name but not using a key I'd been known to use, it could be worthwhile). However, without a true WoT verification, which is rather hard to maintain in this distributed of a project, it could be less useful to have.

And yes, entirely concur about GitHub rewriting the commit with its own keys... Sadly rebase and merge is even worse, as it won't do verification and thus also won't re-sign the commits with its own keys.

I'll keep looking for feedback and see if we can reach a more informed consensus as a community. If you want to start a discussion thread, that'd be appreciated.

Issue #400, opened 2024-07-11
OpenBao Beta shows wrong version


JanMa:
**Describe the bug**
The OpenBao Beta version still reports itself as the Alpha version when running `bao version`

**To Reproduce**
1. Download and extract the OpenBao Beta binary
2. Run `bao version`

The output will show `OpenBao v2.0.0-alpha20240329`

**Expected behavior**
The binary should report the version as `OpenBao v2.0.0-beta20240618`

cipherboy:
@JanMa `version/VERSION` will need to be updated I think.

PR #401, opened 2024-07-12
set bao binary version info with Goreleaser

JanMa:
This adjusts the Goreleaser config to fill the required veriables which
are used when reporting the binary version with `bao version`.

Closes: #400 

## Target Release

2.0.0


version/version_base.go
```diff
@@ -19,9 +18,7 @@ var (
 	// Whether cgo is enabled or not; set at build time
 	CgoEnabled bool
 
-	// Version and VersionPrerelease info are now being embedded directly from the VERSION file.
-	// VersionMetadata is being passed in via ldflags in CI, otherwise the default set here is used.
-	//go:embed VERSION
+	// Filled by Goreleaser
```
cipherboy, line 21: Just a note, but I think this will mean we'll break local builds:

```
$ bao version
OpenBao v ('f87b36cbe71689b205147b3c91f560318916357c'), built 2024-07-12T22:18:27Z
```

Perhaps we want to do `local` or `dev` or something in the `fullVersion`? Though we can lack commit info (e.g., from forks), so we don't know exactly where this commit was branched off of... Thoughts?

cipherboy (approved):
Thanks, @JanMa!

version/version_base.go
cipherboy: @JanMa Ah, I see, I gave bad advice. `fullVersion` needs to be parseable by [go-version](https://github.com/hashicorp/go-version), so it needs to be semver... Hmm

JanMa: Leaving the default empty also broke a whole lot of tests. I went with `0.0.0-HEAD`

cipherboy: Ah, but it needs to be at least `v1` I think. So maybe it is better to leave `VERSION` in the source?

JanMa: Interesting :thinking: Is that the reason some of the token related tests suddenly start failing?
https://github.com/openbao/openbao/actions/runs/9914044176/job/27392390116?pr=401#step:7:5976

cipherboy: @JanMa Yeah, I believe so. We ran into that earlier when we were trying to choose a version number.

Issue #407, opened 2024-07-15
Generating a ruby client from the Open API spec will result to multiple `initialize` methods in System API


fcbajao:
**Describe the bug**

The `OperationVerb` for [`POST /sys/init`](https://github.com/openbao/openbao/blob/main/website/content/api-docs/system/init.mdx#start-initialization) is [currently set to `initialize`](https://github.com/openbao/openbao/blob/22c93f779e57183bb87ca9d5719f317accc4498c/vault/logical_system_paths.go#L584-L590). This becomes problematic when generating a ruby client from the OpenAPI spec because it will result to 2 copies of the `initialize` method for the `SystemApi` class, for example:

```ruby
module OpenbaoClient
  class SystemApi
    # This is the actual constructor that we want
    def initialize(api_client = ApiClient.default)
      @api_client = api_client
    end
   
    # ...

    # This would unintentionally override the constructor
    def initialize(initialize_request, opts = {})
      initialize_with_http_info(initialize_request, opts)
      nil
    end
  end
end
```

**To Reproduce**
1. Run [`./scripts/gen_openapi.sh`](https://github.com/openbao/openbao/blob/main/scripts/gen_openapi.sh)
2. Generate a ruby client through https://openapi-generator.tech:

    ```shell
    openapi-generator generate -g ruby -i scripts/openapi.json -o openbao_client --additional-properties=gemName=openbao_client
    ```
4. See the generated `lib/openbao_client/api/system_api.rb` for the double `initialize` methods.

**Expected behavior**

We should add `system` as the `OperationSuffix` so that it would result to `initialize_system` which would avoid the conflict.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`):
* OpenBao CLI Version (retrieve with `bao version`):
* Server Operating System/Architecture:

OpenBao server configuration file(s):

```hcl
# Paste your OpenBao config here.
# Be sure to scrub any sensitive values
```

**Additional context**

PR #408, opened 2024-07-15
Add system as OperationSuffix to initialize

fcbajao:
Resolves https://github.com/openbao/openbao/issues/407

## Target Release

1.14.7

### DIFF
--- a/vault/logical_system_paths.go
+++ b/vault/logical_system_paths.go
@@ -583,7 +583,8 @@ func (b *SystemBackend) configPaths() []*framework.Path {
 				},
 				logical.UpdateOperation: &framework.PathOperation{
 					DisplayAttrs: &framework.DisplayAttributes{
-						OperationVerb: "initialize",
+						OperationVerb:   "initialize",
+						OperationSuffix: "system",
 					},
 					Summary:     "Initialize a new OpenBao instance.",
 					Description: "The OpenBao instance must not have been previously initialized. The recovery options, as well as the stored shares option, are only available when using OpenBao HSM.",


fcbajao:
cc @cipherboy

vault/logical_system_paths.go
```diff
@@ -584,6 +584,7 @@ func (b *SystemBackend) configPaths() []*framework.Path {
 				logical.UpdateOperation: &framework.PathOperation{
 					DisplayAttrs: &framework.DisplayAttributes{
 						OperationVerb: "initialize",
```
cipherboy, line 586: Just a note for others, but the subsequent place this is done (line 817 originally prior to this change) isn't affected as the entire group of operations has `OperationPrefix: "rekey-attempt"`, making the full function call `rekey_attempt_initialize(...)` in the Ruby client. 

Further this is the [only keyword](https://stackoverflow.com/questions/16099073/ruby-reserved-class-method-names) that appears to be reserved:

```
$ comm -12 /tmp/ruby-reserved-keywords /tmp/operations 
initialize
```

cipherboy (approved):

fcbajao:
@cipherboy there are failing checks. Is there anything needed to be done on my end?

cipherboy:
@fcbajao Ah, do you mind running `make fmt`? It [looks like](https://github.com/openbao/openbao/actions/runs/9944103236/job/27469857104?pr=408) formatting differs. A [changelog entry](https://github.com/openbao/openbao/blob/main/changelog/README.md) might be nice as well just for consistency, but I don't know that we consider OAS production ready yet, given we don't have any official client libraries built off of it.

fcbajao:
> but I don't know that we consider OAS production ready yet, given we don't have any official client libraries built off of it.

@cipherboy if this is the case, maybe we can skip the changelog entry for now.

fcbajao:
> Ah, do you mind running `make fmt`? It [looks like](https://github.com/openbao/openbao/actions/runs/9944103236/job/27469857104?pr=408) formatting differs

@cipherboy fixed the formatting. Please let me know if there's anything else needed on my end.

cipherboy:
Given the above discussion around changelog, I'm going ahead and merging. Thanks @fcbajao!

PR #409, opened 2024-07-16
Fix OpenAPI spec operation descriptions

cipherboy:
Many entries were missing summary and descriptions because while they used the new format, they didn't add separate summary and descriptions for each operation separately from the path. This meant our OAS generator was leaving them empty.

In the future, we should probably audit these cases and see if the global synopsis/description are good or if we'd prefer per-operation descriptions instead.

### DIFF
--- a/sdk/framework/openapi.go
+++ b/sdk/framework/openapi.go
@@ -340,8 +340,14 @@ func documentPath(p *Path, specialPaths *logical.Paths, requestResponsePrefix st
 				requestResponsePrefix,
 			)
 
-			op.Summary = props.Summary
-			op.Description = props.Description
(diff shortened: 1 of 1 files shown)


driif:
Hey @cipherboy, looks like minimal changes, that solve a problem without overengineering. We also need to update tests reflcting new behavior, TestOpenAPI_Paths - fails.

cipherboy:
@driif Do you want to pick up the change?

driif:
@cipherboy sure. I'll take a look to tests.

cipherboy:
Closing as @driif was potentially interested in picking up the change.

PR #410, opened 2024-07-16
Update Go version, changelog, modules for GA

cipherboy:
This introduces a new Makefile target, release-changelog, which requires to environment variables (LAST_RELEASE, THIS_RELEASE) to generate a changelog difference between. This can be revised and placed in CHANGELOG.md and on the website.


cipherboy:
@JanMa I forgot the `sdk/go.mod` for g-k-w updates. Tricky!

JanMa (approved):

cipherboy:
@JanMa I've pushed a new commit for fixing the vulnerable Go dependencies.

cipherboy:
@JanMa This should remove the offending test case; as explained in the comment, I thought it was better to remove than update it. Also shows the difference between running go tests from the root versus within the `api/` subfolder. :-D

PR #411, opened 2024-07-17
Add GOOS matrix to release workflow

cipherboy:
This splits the release workflow into a matrix, using matrix parallelism to avoid exceeding default disk space (14GB) per OS. If we have too many Linux architectures in the future, we could split by architecture as well.

This should mean that each workflow executes a separate operating system but contributes to a single release draft. We'll need to remember to do manual clean up of releases in the event that a pipeline fails, as currently we'll default to appending to the existing one due to the parallelism.

---

We'll need to recreate the tag after this in order to use this workflow, since `tag` isn't an option in the inputs, only the source branch. We could change that if we desire.

In particular, the issue with this approach is that we have to create a non-draft release right away, however, we get the benefit of being able to rerun failed jobs. If we want to rerun everything, we could drop the release and re-do the pipeline from the same commit.

v6 is require to "support" this type of custom matrix to reuse the same release; it looks like this code wasn't in the version of the tool that v5 uses. `--debug` was removed as an option from that tool version too. 

We keep the two underlying build jobs (`build-linux` and `build-other`); in the future we'll probably have a `build-hsm` and potentially other options; using the sed approach seems best from the templating PoV. `nfpms` cannot exist in a `build-other` non-linux list so we have to exclude it manually. `checksums.txt` is the only artifact that is the same between runs in the matrix, so we've manually created a OS variant of it since we're already invoking `sed` on the file.

A sample release without signatures or containers is available here: https://github.com/cipherboy/openbao/releases/tag/v2.0.0-testing12

PR #412, opened 2024-07-17
Gate Docker steps behind GOOS

cipherboy:
In #411, we added the `#LINUXONLY#` prefix to the `nfpms` section as it triggered on non-Linux builds; Docker was not thought about as I had removed it for my local repository testing.

This shows the need for a proper staging release area, complete with registries and signing keys.

### DIFF
--- a/.goreleaser-template.yaml
+++ b/.goreleaser-template.yaml
@@ -226,576 +226,554 @@ checksum:
   disable: false
   # split: false
 
-#snapcrafts:
-#  - name: openbao
-#    title: OpenBao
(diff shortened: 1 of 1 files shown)


JanMa (approved):

.goreleaser-template.yaml
```diff
@@ -226,576 +226,554 @@ checksum:
   disable: false
   # split: false
 
-#snapcrafts:
```
cipherboy, line 229: To avoid confusion, this unused config segment was removed.

PR #414, opened 2024-07-17
update download page logic for new release assets

JanMa:
Ignore some more metadata files and add a section for Arch Packages


cipherboy (approved):
Slick, thanks @JanMa!

image

PR #416, opened 2024-07-19
use openbao instead of vault prefix for kubernetes labels

finkandreas:
The helm chart creates the service looking for the pod labels `openbao-active=true`. This PR fixes that the label is applied to this name instead of `vault-active`. Additionally it also puts all other labels into `openbao` prefix instead of `vault`.

This is the line in the helm chart has uses as only valid name `openbao-active`: https://github.com/openbao/openbao-helm/blob/main/charts/openbao/templates/server-ha-active-service.yaml#L60

### DIFF
--- a/changelog/416.txt
+++ b/changelog/416.txt
@@ -1,0 +1,4 @@
+```release-note:change
+serviceregistration/kubernetes: labels use `openbao` as prefix instead of `vault`.
+```
+
--- a/serviceregistration/kubernetes/service_registration.go
+++ b/serviceregistration/kubernetes/service_registration.go
@@ -17,11 +17,11 @@ import (
 
 const (
 	// Labels are placed in a pod's metadata.
-	labelVaultVersion = "vault-version"
-	labelActive       = "vault-active"
-	labelSealed       = "vault-sealed"
-	labelPerfStandby  = "vault-perf-standby"
-	labelInitialized  = "vault-initialized"
+	labelVaultVersion = "openbao-version"
+	labelActive       = "openbao-active"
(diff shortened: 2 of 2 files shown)


JanMa:
Thanks a lot for opening this PR @finkandreas. Can you please amend your commit to [sign it off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s) and add a change log entry to the `changelog` folder. Then I'll happily merge it :-)

finkandreas:
Both points addressed. Thanks for merging :)

JanMa (approved):

cipherboy (approved):
LGTM; bypassing signature requirement for trivial commit.

kim0:
@JanMa @cipherboy Can we please do a release with this fix in
I'm trying to stand an openbao cluster in ha mode and hitting https://github.com/openbao/openbao-helm/issues/12 with no workaround available!
I was really hoping to get this deployed tomorrow, so it would be really appreciated if we can get this done asap 🙏

JanMa:
Hello @kim0, we're a bit short staffed this week because of the current vacation season. We've already had a few fixes contributed since 2.0.0 and I agree with you that we should release a patch soon, but I can't make any promises as to when this will happen.

In the meantime, you can always build OpenBao and the Docker image yourself from the `main` branch :-)

cipherboy:
@kim0 @JanMa I think personally, we've merged part of the transactional storage PR series, so it'd be good to finish that up for the next release. My 2c -- I'm happy doing reviews of OIDC and PKCS#11 here soon too.

Issue #419, opened 2024-07-20
Failed to set process secure bits


in0rdr:
After the upgrade from v2.0.0-beta20240618 to v2.0.0, I was unable to start bao server.

The systemd service unit fails with the following message:
```
Jul 19 21:03:18 systemd[1]: Starting "OpenBao - A tool for managing secrets"...
Jul 19 21:03:18 systemd[1820]: openbao.service: Failed to set process secure bits: Operation not permitted
Jul 19 21:03:18 systemd[1820]: openbao.service: Failed at step SECUREBITS spawning /usr/bin/bao: Operation not permitted
```

**Environment:**
* OpenBao Server Version 2.0.0
* OpenBao CLI Version 2.0.0
* Ubuntu 20.04.6 LTS (x86_64)

OpenBao server configuration file(s):

```hcl
ui = true

storage "file" {
  path = "/opt/openbao/data"
}

listener "tcp" {
  address       = "0.0.0.0:8200"
  tls_cert_file = "/opt/openbao/tls/tls.crt"
  tls_key_file  = "/opt/openbao/tls/tls.key"
}
```

cipherboy:
@in0rdr Interesting! I'm able to reproduce this on 20.04, but not on Fedora 40. It looks like, for some reason, 20.04 still requires:

```
AmbientCapabilities=CAP_IPC_LOCK
CapabilityBoundingSet=CAP_IPC_LOCK
```

But I'm not quite sure why. Do you perhaps have ideas?

In #363 we've [removed mlock](https://openbao.org/docs/rfcs/mlock-removal/), so I'm surprised that it is still conditionally required.

Perhaps you have thoughts @IohannesArnold?

in0rdr:
Thank you @cipherboy for connecting it to the change. The suggested fix also works on my (old) machine. Using following variables in the service unit now:
```
CapabilityBoundingSet=CAP_SYSLOG CAP_IPC_LOCK
AmbientCapabilities=CAP_IPC_LOCK
```

I already assumed it could be related to mlock, but only quickly skimmed through the docs in https://openbao.org/docs/rfcs/mlock-removal.

For now, I will probably update my old OS :smile: and read a bit more about the Systemd configuration variables and how they relate to mlock. Also, I'm going to rollout the new bao 2.0.0 on some more recent 12 (bookworm) machines to see the difference.

Thanks again, much appreciated.

cipherboy:
@in0rdr No problem, let us know what you find out! 

[...]

PR #420, opened 2024-07-21
Add known issue about SecureBits on Ubuntu 20.04

cipherboy:
Resolves: https://github.com/openbao/openbao/issues/419

// cc: @in0rdr


JanMa (approved):
LGTM

PR #422, opened 2024-07-22
Remove one-shot transactional interface

cipherboy:
This removes the legacy one-shot transaction interface and its various implementers, including in the File, InMem, and Raft backends. In the former's case (file), this might not have truly had the strict consistency guarantees one would expect from a transaction implementation.

---

This PR has been split from #292 and needs to merge prior to that one.


cipherboy:
@DanGhita @JanMa @schultz-is Any chance I could get a review on this one? Thanks :-)

DanGhita (approved):
@cipherboy : I think the next step, concerning the actual implementation of the new transactional storage that you have proposed, it is very important and it will require a deep analyze, work and review from the community.

cipherboy:
@DanGhita Agreed. Thank you!

PR #425, opened 2024-07-23
Fix api, sdk modules with v2.0.1

cipherboy:
This retracts the` v2.0.0` module version that we pushed and finishes the Go module at `/v2`. In the future, the following module tagging order will need to be used:

 - api
 - api/auth/*
 - sdk
 - Any external dependencies (go-kms-wrapping, openbao-template, &c)
 - Finally, the main module

In this case, the api/auth/* modules aren't currently buildable until the v2.0.1 tag is pushed for api. However, due to replace statements, the main binary should build just fine.

PR #427, opened 2024-07-24
Use quay.io/openbao/openbao in docker tests

cipherboy:
Now that OpenBao has GA'd with v2.0.0, we should update the SDK's docker testing framework to pull that instead of upstream Vault images.

---

This should land ahead of #425 so that consumers of the SDK have the right repository. This ended up being slightly larger than I expected as `sdk/helper/stepwise/environments/docker/environment.go` wasn't updated, but I didn't know how much it was actually used.


JanMa (approved):
LGTM

Issue #296, opened 2024-04-15
RFC - Transactional Storage for Plugins & Core


cipherboy:
### Summary

OpenBao and its upstream lacks transactional storage semantics. This means all storage operations are logically separate: between two put operations, the contents of storage may differ or the system may crash. Adding support for transactions gives resilience and better fault tolerance to OpenBao.

### Problem Statement

Transactions are a well-utilized concept in databases. OpenBao's [storage interface](https://github.com/openbao/openbao/blob/main/sdk/logical/storage.go#L31-L38) remains primitive and does not support transactions. The underlying [physical layer](https://github.com/openbao/openbao/blob/main/sdk/physical/physical.go) supports a [one-shot transaction interface](https://github.com/openbao/openbao/blob/main/sdk/physical/transactions.go) but is used only [in a few places](https://github.com/openbao/openbao/blob/124a5f10bc234a68c1c954a2ff47f519b22d1d33/vault/sealunwrapper.go#L158-L175) due to its inflexibility.

As [mentioned recently](https://github.com/openbao/openbao/issues/270), fixing this is important for hitting organization's internal SLA's and offering requirements. OpenBao fundamentally pairs a K/V storage engine with a plugin architecture; managing secrets requires strict storage safety guarantees.

Extending OpenBao's storage model to include [more standard](https://pkg.go.dev/database/sql#DB.BeginTx) transaction support (with the ability to interleave storage operations with other code) would give the plugin programmer more standard but expressive safety mechanisms. This would resolve broad categories of faults and bring OpenBao in line with industry at large relying directly on transactional storage semantics.

### User-facing description

To the end-consumer of OpenBao, few changes are expected to occur: at most, some write operations may fail due to transactions failing. This will necessitate additional client side retries for a new category of errors: transaction commit failures. This is discussed more [below](#survey-of-requirements).

This will not be client-driven transactions (i.e., long-lived transactions that the client can use to ensure consistency across multiple API calls and which could potentially starve other clients); these will be similar to database transactions that the plugins themselves can implement.

To the programmer of plugins, this will conditionally open up new transaction APIs (if the backend supports it), allowing them to opt into safer storage semantics. This will be detectable at runtime through type assertions on the `logical.Storage` interface passed in to the plugin.

### Technical Description

This section is broken up as follows:

 1. [Survey of Requirements](#survey-of-requirements)
 2. [Prior Art](#prior-art)
 3. [Implementation in Raft](#implementation-in-raft)
[...]

PR #433, opened 2024-07-27
Add new transaction interfaces to sdk/

cipherboy:
This adds interfaces for incremental transactions in sdk/, including to logical.Storage and physical.Backend, eventually allowing support throughout the storage stack.

---

This was split from #292 and was the next required commit to review. 

/cc: @schultz-is

### DIFF
--- a/sdk/logical/storage_transactions.go
+++ b/sdk/logical/storage_transactions.go
@@ -1,0 +1,48 @@
+// Copyright (c) 2024 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package logical
+
+import (
+	"context"
+)
+
+// Transactional is an optional interface for backends that support
+// interactive (mixed code & statement) transactions in a similar
+// style as Go's Database paradigm. This is equivalent to
+// physical.Transactional, not the earlier, one-shot version of the
+// interface.
+type Transactional interface {
+	// This function allows the creation of a new interactive transaction
+	// handle, only supporting read operations. Attempts to perform write
+	// operations (
+	BeginReadOnlyTx(context.Context) (Transaction, error)
+
+	// This function allows the creation of a new interactive transaction
+	// handle, supporting read/write transactions. In some cases, the
+	// underlying physical storage backend cannot handle parallel read/write
+	// transactions.
+	BeginTx(context.Context) (Transaction, error)
+}
+
+// Transaction is an interactive transactional interface: backend storage
+// operations can be performed, and when finished, Commit or Rollback can
+// be called. When a read-only transaction is created, write calls (Put(...)
+// and Delete(...)) will err out.
+type Transaction interface {
+	Storage
+
+	// Commit a transaction; this is equivalent to Rollback on a read-only
+	// transaction.
+	Commit(context.Context) error
+
+	// Rollback a transaction, preventing any changes from being persisted.
+	Rollback(context.Context) error
+}
+
+// TransactionalStorage is implemented if a storage backend implements
+// Transactional as well.
+type TransactionalStorage interface {
+	Storage
+	Transactional
+}
--- a/sdk/physical/physical.go
+++ b/sdk/physical/physical.go
@@ -10,7 +10,10 @@ import (
 	log "github.com/hashicorp/go-hclog"
 )
 
-const DefaultParallelOperations = 128
+const (
+	DefaultParallelOperations   = 128
+	DefaultParallelTransactions = 64
+)
 
 // The operation type
 type Operation string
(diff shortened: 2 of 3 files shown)


sdk/logical/storage_transactions.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) OpenBao, Inc.
```
JanMa, line 1: Are you sure that this is correct? For the website I was asked to use the following:
> Copyright © 2024 OpenBao a Series of LF Projects, LLC

JanMa (approved):
LGTM but see the comment about the copyright header

sdk/logical/storage_transactions.go
cipherboy: Interesting, I didn't know that we had gotten clarity about that. Happy to update that.

JanMa: Interestingly OpenTofu uses
> Copyright (c) The OpenTofu Authors
According to https://github.com/opentofu/opentofu/issues/454#issuecomment-1904213569 this seems to have been discussed with their lawyers. So maybe we could also adopt a similar wording

cipherboy: I'm going to with:

`Copyright (c) 2024 OpenBao a Series of LF Projects, LLC`

since I don't want to copy/paste the special © every time.

cipherboy:
Going to go ahead and merge this one since it holds up the implementations and those are what need the most review. Semantics are fairly well set and tested by #292 and defined in #296.

Thanks @JanMa!

sdk/logical/storage_transactions.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Transactional is an optional interface for backends that support
+// interactive (mixed code & statement) transactions in a similar
+// style as Go's Database paradigm. This is equivalent to
+// physical.Transactional, not the earlier, one-shot version of the
+// interface.
+type Transactional interface {
+	// This function allows the creation of a new interactive transaction
+	// handle, only supporting read operations. Attempts to perform write
+	// operations (
```
DanGhita, line 18: The explanation sentence seems to be unfinished.

sdk/logical/storage_transactions.go
```diff
@@ -0,0 +1,48 @@
...
+// Transaction is an interactive transactional interface: backend storage
+// operations can be performed, and when finished, Commit or Rollback can
+// be called. When a read-only transaction is created, write calls (Put(...)
+// and Delete(...)) will err out.
+type Transaction interface {
+	Storage
+
+	// Commit a transaction; this is equivalent to Rollback on a read-only
+	// transaction.
+	Commit(context.Context) error
```
DanGhita, line 37: I'm suggesting the following comments: "Commit a write transaction.  It has no effect for a read-only transaction."

sdk/logical/storage_transactions.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// Rollback a transaction, preventing any changes from being persisted.
+	Rollback(context.Context) error
```
DanGhita, line 40: I'm suggesting the following comments: "Rollback a write transaction, preventing any changes from being persisted. It has no effect for a read-only transaction."

cipherboy: \o hey @DanGhita -- strictly this is untrue. Transactions (whether here in OpenBao or more generally into databases and other things) can consume resources and need to be closed. See e.g., https://go.dev/doc/database/execute-transactions#example --- `defer tx.Rollback(...)` is a common pattern in Go.

"It has no effect for a read-only transaction" is thus incorrect: it has the potential to release memory and thus should be called. Giving the opposite impression (i.e., read-only transactions do not have any resources associated with them and thus Rollback would not need to be called on it) would cause leakage in practice. 

This is similar to https://pkg.go.dev/go.etcd.io/bbolt#Tx.Rollback -- but note that we have more forgiving semantics (allowing commit on read-only transactions to gracefully ignore the transaction).

sdk/logical/storage_transactions.go
cipherboy: See below.

sdk/logical/storage_transactions.go
cipherboy: Good catch, this was meant to read:

Attempts to perform write operations (Put, Delete) will err.

sdk/logical/storage_transactions.go
DanGhita: OK, I understand, but the current comments do not clearly reflect the fact that the caller must call either Commit or Rollback even for read-only transactions.

PR #435, opened 2024-07-27
Fix OIDC challenge_code regex in test

cipherboy:
The given authURL is validated against various components that are expected to exist. Per RFC however, the challenge code may have non-word characters (`-` and `_`) when using the alternative URL-safe Base64 encoding mechanism. Expand the regex to handle this case.

See also: https://datatracker.ietf.org/doc/html/rfc7636#section-4.2


JanMa (approved):
LGTM

cipherboy:
Thanks, @JanMa. Since this is a simple test fix, going ahead and meeting.

PR #436, opened 2024-07-28
Add transactions to Raft backend

cipherboy:
This implements transactions in the Raft backend, using an augmented ApplyLog operation entry. By adding list and get operations (hashing their output state) to the operation log, we can verify state hasn't been modified in parallel by another writer, failing the transaction otherwise.

The current implementation is perhaps inefficient at times: multiple get or list operations over the same path could potentially be collapsed, and all get operations moved to the start of transaction (combing the hashing of multiple entries into a single value, reducing the size of the log on the wire), but this needs to be done carefully. Thus, this work is deferred until later when the current implementation is deemed insufficient.

### DIFF
--- a/physical/raft/fsm.go
+++ b/physical/raft/fsm.go
@@ -36,6 +36,10 @@ const (
 	putOp
 	restoreCallbackOp
 	getOp
+	verifyReadOp
+	verifyListOp
+	beginTxOp
+	commitTxOp
 
 	chunkingPrefix   = "raftchunking/"
 	databaseFilename = "vault.db"
@@ -59,6 +63,12 @@ var (
 
 type restoreCallback func(context.Context) error
 
+// fsmEntryTxErrorKey is the value for a FSMEntry to signal that it is
+// not the result of a regular Get operation (which cannot occur in a
+// transaction) but is instead contains the response value from failing to
+// apply this transaction.
+const fsmEntryTxErrorKey = "[\ttransaction-commit-failure\t]"
+
 type FSMEntry struct {
 	Key   string
 	Value []byte
@@ -66,6 +76,22 @@ type FSMEntry struct {
 
 func (f *FSMEntry) String() string {
 	return fmt.Sprintf("Key: %s. Value: %s", f.Key, hex.EncodeToString(f.Value))
+}
+
+func (f *FSMEntry) IsTxError() bool {
+	return f.Key == fsmEntryTxErrorKey
+}
+
+func (f *FSMEntry) AsTxError() error {
+	str := string(f.Value)
+	commitErr := physical.ErrTransactionCommitFailure.Error()
+
+	split := strings.SplitN(str, commitErr, 1)
+	if len(split) != 2 {
+		return errors.New(str)
+	}
+
+	return fmt.Errorf("%v%w%v", split[0], physical.ErrTransactionCommitFailure, split[1])
 }
 
 // FSMApplyResponse is returned from an FSM apply. It indicates if the apply was
@@ -538,6 +564,23 @@ func (f *FSM) ListPage(ctx context.Context, prefix string, after string, limit i
 	f.l.RLock()
 	defer f.l.RUnlock()
 
+	var err error
+	var keys []string
+	err = f.db.View(func(tx *bolt.Tx) error {
+		keys, err = listPageInner(ctx, tx, prefix, after, limit)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+
+	return keys, err
+}
+
+func listPageInner(ctx context.Context, tx *bolt.Tx, prefix string, after string, limit int) ([]string, error) {
+	var keys []string
+
 	prefixBytes := []byte(prefix)
 	seekPrefix := []byte(filepath.Join(prefix, after))
 	if after == "" {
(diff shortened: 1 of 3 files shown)


schultz-is (approved):
I like the approach here. Feels like a good starting point for general transaction support across the project with limited overhead required.

JanMa (approved):
Thanks for the review @schultz-is

physical/raft/raft.go
```diff
@@ -510,7 +531,9 @@ func NewRaftBackend(conf map[string]string, logger log.Logger) (physical.Backend
 		dataDir:                    path,
 		localID:                    localID,
 		permitPool:                 physical.NewPermitPool(physical.DefaultParallelOperations),
+		txnPermitPool:              physical.NewPermitPool(physical.DefaultParallelOperations),
```
cipherboy, line 534: This is wrong, it needs to be `physical.DefaultParallelTransactions`

physical/raft/transaction.go
```diff
@@ -0,0 +1,611 @@
...
+
+	hashType := expected[0]
+
+	actual, err := createVerificationEntryOfType(hashType, key, value)
+	if err != nil {
+		return err
+	}
+
+	if subtle.ConstantTimeCompare(actual, expected) == 0 {
+		return fmt.Errorf("for get %v (actual=%v vs expected=%v - actual value `%v`): %w", key, hex.EncodeToString(actual), hex.EncodeToString(expected), hex.EncodeToString(value), physical.ErrTransactionCommitFailure)
```
cipherboy, line 100: Should remove this debugging information.

physical/raft/transaction.go
```diff
@@ -0,0 +1,611 @@
...
+
+	sItems := strings.Join(items, "\n")
+
+	actual, err := createVerificationEntryOfType(hashType, repr, []byte(sItems))
+	if err != nil {
+		return err
+	}
+
+	if subtle.ConstantTimeCompare(actual, expected) == 0 {
+		return fmt.Errorf("for list %v: %w", repr, physical.ErrTransactionCommitFailure)
```
cipherboy, line 127: As above.

cipherboy:
@DanGhita Do you want me to wait for you to merge?

DanGhita:
@cipherboy : please go ahead, don't wait for me, I'm reviewing the code in parallel and I will let you know if I have remarks.

cipherboy:
Another flaky data-race test:

```go

=== Failed
=== FAIL: builtin/logical/openldap TestAutoRotate/auto_rotate_role (6.00s)
2024-08-14T10:18:07.996Z [INFO]  initializing database rotation queue
2024-08-14T10:18:07.996Z [INFO]  populating role rotation queue
2024-08-14T10:18:07.996Z [DEBUG] no WAL entries found
2024-08-14T10:18:07.996Z [INFO]  starting periodic ticker
2024-08-14T10:18:07.997Z [INFO]  stopping periodic ticker
2024-08-14T10:18:08.000Z [DEBUG] wrote WAL: role=hashicorp WAL ID=389fd06c-3349-60ad-e332-212542867180
2024-08-14T10:18:08.000Z [DEBUG] deleted WAL: WAL ID=389fd06c-3349-60ad-e332-212542867180
    rotation_test.go:109: expected last_password to be set after auto rotation, it wasn't
2024-08-14T10:18:14.000Z [INFO]  stopping periodic ticker

=== FAIL: builtin/logical/openldap TestAutoRotate (6.00s)

DONE 358 tests, 1 skipped, 2 failures in 209.942s
```

physical/raft/fsm.go
```diff
@@ -68,6 +78,22 @@ func (f *FSMEntry) String() string {
 	return fmt.Sprintf("Key: %s. Value: %s", f.Key, hex.EncodeToString(f.Value))
 }
 
+func (f *FSMEntry) IsTxError() bool {
```
DanGhita, line 81: @cipherboy : I'm not sure about the pattern here in Go, but are you sure there is no need to check if "f" is a  nil pointer or not before using it in the function ("class method" )  ?

cipherboy: Technically, yes: https://go.dev/play/p/sFM5BnM8vDz 

However, checking for a nil value in a method call isn't common practice in Go: as far as I can tell with semgrep, this is never done across the entire code base:

```yml
rules:
  - id: pointer-checked-nil
    patterns:
      - pattern-inside: |
          func ($FOO *$TYPE) $FUNC(...) {
             ...
          }
      - pattern:
          if $FOO == nil { ... }
    message: "Checking pointer receiver against nil"
    languages: [go]
    severity: ERROR
```

I have not seen it done either in other code bases. TBH, I'm not sure how you'd handle this, return an error? Now every method needs to return an error where one didn't exist, rather than panicking. 

In our case, we know from construction that this will never be nil.

physical/raft/fsm.go
```diff
@@ -68,6 +78,22 @@ func (f *FSMEntry) String() string {
...
+	return f.Key == fsmEntryTxErrorKey
+}
+
+func (f *FSMEntry) AsTxError() error {
+	str := string(f.Value)
+	commitErr := physical.ErrTransactionCommitFailure.Error()
+
+	split := strings.SplitN(str, commitErr, 1)
+	if len(split) != 2 {
```
DanGhita, line 90: @cipherboy : I'm not sure about this: we should rather have strings.SplitN(str, commitErr, **2**) ?

cipherboy: Ah, good catch, playground confirms you are right: https://go.dev/play/p/d3_hn453lQN

Fortunately this just means `errors.Is(...)` isn't true and so is relatively minor, but I will address it in a follow-up PR.

cipherboy: https://github.com/openbao/openbao/pull/466

PR #437, opened 2024-07-28
Add transaction support to encrypted_key_storage

cipherboy:
`EncryptedKeyStorage` is a wrapper used in K/V to ensure that paths (which in the case of K/V may be sensitive) are protected and not disclosed if the underlying (encrypted) storage is compromised. Adding transactions here means that we can modify entries with encrypted path names safely as well.


cipherboy:
Thanks @JanMa!

PR #438, opened 2024-07-28
Add transaction support to LogicalStorage

cipherboy:
`LogicalStorage` is a direct wrapper over a physical backend, only useful for testing when barrier encryption is not desirable. Otherwise, barrier encryption would handle this. Since we want transactions in everything (if the underlying storage mechanism supports it), we want to add it here as well.


JanMa (approved):
LGTM

PR #439, opened 2024-07-28
Support transactions in StorageView

cipherboy:
This converts `StorageView` to an interface, aligning with the `Storage` interface. The only consumer, as far as I can tell, is the `BarrierView`, which was easy enough to convert to an interface user, and will itself need to be made transactional in a future commit.

### DIFF
--- a/sdk/logical/storage_view.go
+++ b/sdk/logical/storage_view.go
@@ -9,37 +9,80 @@ import (
 	"strings"
 )
 
-type StorageView struct {
-	storage Storage
-	prefix  string
+type StorageViewCore interface {
+	Prefix() string
+	SubView(prefix string) StorageView
+	SanityCheck(key string) error
+	ExpandKey(suffix string) string
+	TruncateKey(full string) string
+}
+
+type StorageView interface {
+	Storage
+	StorageViewCore
+}
+
+type TransactionalStorageView interface {
+	TransactionalStorage
+	StorageViewCore
+}
+
+// Note that, within a transaction, creating and committing from a
+// SubView commits the entire transaction.
+type StorageViewTransaction interface {
+	Transaction
+	StorageViewCore
 }
 
 var ErrRelativePath = errors.New("relative paths not supported")
 
-func NewStorageView(storage Storage, prefix string) *StorageView {
-	return &StorageView{
+func NewStorageView(storage Storage, prefix string) StorageView {
+	sv := &storageView{
 		storage: storage,
 		prefix:  prefix,
 	}
+
+	if _, ok := storage.(TransactionalStorage); ok {
+		return &transactionalStorageView{
+			*sv,
+		}
+	}
+
+	if _, ok := storage.(Transaction); ok {
+		return &storageViewTransaction{
+			*sv,
+		}
+	}
+
+	return sv
 }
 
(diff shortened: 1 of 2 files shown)


cipherboy:
@JanMa @DanGhita I think this is the next easiest one :-)

vault/barrier_view.go
```diff
@@ -19,7 +19,7 @@ import (
 // BarrierView implements logical.Storage so it can be passed in as the
 // durable storage mechanism for logical views.
 type BarrierView struct {
-	storage         *logical.StorageView
+	storage         logical.StorageView
```
DanGhita, line 22: @cipherboy : I didn't understand why we have to replace the pointer by the structure itself.

sdk/logical/storage_view.go
```diff
@@ -9,37 +9,80 @@ import (
...
 }
 
+var (
+	_ Storage     = &storageView{}
+	_ StorageView = &storageView{}
+)
+
 // logical.Storage impl.
-func (s *StorageView) List(ctx context.Context, prefix string) ([]string, error) {
+func (s *storageView) List(ctx context.Context, prefix string) ([]string, error) {
```
DanGhita, line 71: @cipherboy : is this a kind of Go encapsulation mechanism ? The public method is defined with a private (package) receiver.

DanGhita (approved):

vault/barrier_view.go
cipherboy: We didn't :-) 

In `sdk/logical/storage_view.go` we renamed `StorageView` (the struct) to `storageView` (note lowercase S, so it is not exported). `StorageView` is now an interface (in a two parts -- the `View` portion called `StorageViewCore` and the `Storage` portion), implemented by `storageView` with pointer receivers.

So, this change replaces a pointer-to-struct with an interface, which is mostly equivalent to pointer-to-struct, except that it hides the internal implementation details (which weren't directly used anyways).

sdk/logical/storage_view.go
cipherboy: Not really.

Go has the concept of an interface, which is an arbitrary grouping of methods. You can type assert between various interfaces or implementations there-of (so `var x logical.StorageView` could be type asserted to `x.(*storageView)` if you were in `logical`. 

The interface itself (capital S `StorageView`) cannot have default implementations attached to it (unless they added this and I missed it).

At least in this code base, and in a few others I've seen, it is relatively common to hide the implementation of a public interface behind a private struct implementing it.

So, by making the rename change above (`type StorageView struct {}` --> `type storageView struct {}` and adding `type StorageView interface{}` in its place), we get a way of hiding the implementation details behind an interface (collection of methods) and can update the only callers that bothered to use it directly (in this case, `vault/barrier_view.go`).

vault/barrier_view.go
DanGhita: @cipherboy : OK, thank you!

sdk/logical/storage_view.go
DanGhita: @cipherboy : Yes, thank you, that's what I thought, but I wasn't sure: it allows to hide implementation details for a given interface.

PR #440, opened 2024-07-28
Support transactions in AES-GCM Barrier

cipherboy:
This extends `SecurityBarrier` into a `TransactionalSecurityBarrier` variant, in which access to the underlying storage (but not the barrier keyring itself!) can be modified in a transaction-aware way.

Note that modifying the barrier keyring within a transaction makes little sense: keyring operations are globally locked and we don't want to rotate to a key used only within a transaction and forked into a different series later. Thus we only support modifications to the barrier keyring outside of the scope of a transaction.

---

This one depends on #443 as we wrap an inmem physical backend for testing, which must be transactional for AES GCM Barrier to become transactional.

### DIFF
--- a/vault/barrier.go
+++ b/vault/barrier.go
@@ -70,13 +70,7 @@ const (
 	shamirKekPath = "core/shamir-kek"
 )
 
-// SecurityBarrier is a critical component of Vault. It is used to wrap
-// an untrusted physical backend and provide a single point of encryption,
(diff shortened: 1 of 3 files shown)


cipherboy:
@JanMa @DanGhita This one and #442 should both be reviewable in parallel. Thanks!

cipherboy:
Thanks @DanGhita!

PR #442, opened 2024-07-28
Add transaction support to cache

cipherboy:
Cache is a layer over the top of the physical storage backend (but below barrier encryption) to reduce strain of repeated read operations on the backend.

We implement transactions by wrapping the underlying physical transaction with a new (empty) cache. This means that we may not have the data locality in oft-read paths, but means that the cache is internally consistent (and that the transaction's cache doesn't need to be merged back into the primary cache). In the future, this could be improved.

### DIFF
--- a/sdk/physical/cache.go
+++ b/sdk/physical/cache.go
@@ -59,8 +59,19 @@ func cacheRefreshFromContext(ctx context.Context) bool {
 // and provide an LRU cache layer on top. Most of the reads done by
 // Vault are for policy objects so there is a large read reduction
 // by using a simple write-through cache.
-type Cache struct {
+type Cache interface {
+	ToggleablePurgemonster
+	Backend
+}
+
+type TransactionalCache interface {
+	ToggleablePurgemonster
+	TransactionalBackend
+}
+
+type cache struct {
 	backend         Backend
+	size            int
 	lru             *lru.TwoQueueCache
 	locks           []*locksutil.LockEntry
 	logger          log.Logger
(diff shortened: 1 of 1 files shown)


JanMa (approved):
LGTM

sdk/physical/cache.go
```diff
@@ -197,14 +230,69 @@ func (c *Cache) Delete(ctx context.Context, key string) error {
...
+		c,
+	}, nil
+}
+
+func (c *cacheTransaction) Commit(ctx context.Context) error {
+	if err := c.cache.backend.(Transaction).Commit(ctx); err != nil {
+		return err
+	}
+
+	c.parent.Purge(ctx)
```
cipherboy, line 287: I think this warrants a comment: we need to purge the cache when we commit the transaction as otherwise we'd need to do the handling in `Put(...)` or `Delete(...)` and replay write operations and invalidate their entries in the parent.

Technically this isn't necessary for `Rollback` since no data was modified so I'll remove this below.

@JanMa do you agree?

cipherboy: I ended up strictly doing invalidations of modified entries. This has also been pushed to #292 so that we can ensure it passes the new transactional test suite :)

cipherboy:
@JanMa @DanGhita This is ready for re-review. :-)

DanGhita (approved):
LGTM

cipherboy:
Thank you, @DanGhita!!

PR #443, opened 2024-07-28
Add in-memory transactional database

cipherboy:
This modifies the in-memory physical backend to support transactions in an inefficient manner: clone the entire tree, build an action log, and reply the log against the main tree, validating the state of any touched entries.

A copy-on-write version of go-radix would improve the performance substantially (letting the old version be maintained in the transaction), but since this backend isn't persistent across restarts of OpenBao, it likely is not worth the effort.

### DIFF
--- a/sdk/physical/inmem/inmem.go
+++ b/sdk/physical/inmem/inmem.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


cipherboy:
@JanMa @schultz-is @DanGhita I think this is the next one blocking merge of the other PRs if someone minds reviewing it? :)

DanGhita (approved):
It seems good to me.

sdk/physical/inmem/inmem.go
```diff
@@ -26,13 +27,15 @@ var (
...
-	ListDisabledError     = errors.New("list operations disabled in inmem backend")
-	GetInTxnDisabledError = errors.New("get operations inside transactions are disabled in inmem backend")
+	PutDisabledError    = errors.New("put operations disabled in inmem backend")
+	GetDisabledError    = errors.New("get operations disabled in inmem backend")
+	DeleteDisabledError = errors.New("delete operations disabled in inmem backend")
+	ListDisabledError   = errors.New("list operations disabled in inmem backend")
 )
 
+// Regular usage with physical.Encoding don't allow non-printable space.
+const missingEntrySentinel = "[\tmissing-entry-sentinel\t]"
```
cipherboy, line 37: This constant appears unused both in here and in #292.

cipherboy:
Thanks @DanGhita!

Issue #399, opened 2024-07-11
Discontinue Enforcing Signed Commits via Branch Protection Rules


JanMa:
### Summary

This RFC proposes discontinuing the enforcement of signed commits via branch
protection rules in our GitHub repositories. The current practice requires every
commit to be signed, with the intention of ensuring authenticity and authorship
verification. However, this enforcement introduces complexities and barriers
that may outweigh its perceived security benefits. The primary goal of this
proposal is to streamline the contribution process, making it more accessible to
all contributors while maintaining adequate security practices through other
means. This document outlines the rationale, potential impacts, and alternative
solutions to ensure the integrity and security of our codebase.

### Problem Statement

The requirement for signed commits is intended to enhance the security of our
codebase by verifying the identity of contributors. However, this policy has
introduced several challenges:

- **Onboarding Difficulty**: New contributors often face a steep learning curve
  to configure commit signing correctly.
- **Increased Overhead**: Regular contributors spend additional time ensuring
  that their commits are signed, which can slow down the development process.
- **Exclusion of Contributors**: Contributors without the technical know-how or
  access to necessary tools are discouraged from contributing, which reduces
  community engagement and diversity.
- **Temporal Validity of Verification**: The validity of commit signatures is
  temporal. If a contributor deletes their SSH key from GitHub's UI, previously
  "verified" commits may become "unverified," compromising the long-term
  reliability of signature verification.
- **Rebase Merge Issues**: Ideally, we would like to rebase pull requests when
  merging to maintain the original commit history of the PR. However, due to a
  long-standing bug in GitHub, all signatures in individual commits become
  invalidated during a rebase, causing them to appear as unverified.
- **Lack of Proper Trust Verification**: We currently have no means to verify
  the Web of Trust of signed commits, which means we cannot reliably detect if a
  contributor's GitHub account has been compromised.

### User-facing description

We are proposing to stop requiring signed commits for contributions to our
GitHub repositories. This change aims to simplify the contribution process,
making it easier for both new and existing contributors to participate. We will
continue to use other security measures to maintain the integrity of our
codebase.

### Technical Description

To implement this change, we will:

1. Update the branch protection rules in GitHub to remove the requirement for
   signed commits.
2. Communicate the change to all contributors through appropriate channels
   (e.g., mailing lists, GitHub discussions).
3. Update our contribution guidelines to reflect the new policy.

### Rationale

The primary rationale for this change is to reduce the barriers to contribution
[...]

PR #447, opened 2024-08-02
add workflow to verify maintainer commits

JanMa:
When a Pr is opened, this workflow checks if all commits made by any of the core
maintainers of OpenBao are verified. This way we can stop enforcing commits via
GitHub branch protection rules for all committers while still ensuring that
maintainer commits need to be signed.

### DIFF
--- a/.github/scripts/verify-commits.jq
+++ b/.github/scripts/verify-commits.jq
@@ -1,0 +1,23 @@
+# Expects the list of commits on a pull requests as input.
+# See: https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-commits-on-a-pull-request
+[
+  # Iterate over all commits
+  .[] |
+  # select maintainer commits
+  select(
+    .author.login == "JanMa" or
+    .author.login == "cipherboy" or
+    .author.login == "DanGhita" or
+    .author.login == "naphelps"
+  ) |
+  # select any unsigned commits
+  select(.commit.verification.verified == false)
+] |
+# check if there are unsigned commits
+if (. | length) != 0 then
+  # return error
+  ("Pr contains unsigned maintainer commits!\n" | halt_error(1))
+else
+  # return success
+  "All maintainer commits are signed"
+end
--- a/.github/workflows/verify-commits.yml
+++ b/.github/workflows/verify-commits.yml
@@ -1,0 +1,22 @@
+name: Ensure Verified Commits
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  verify_commits:
+    name: Verify Commit Signatures
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check commit signatures
+        run: |
+          gh api \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            /repos/openbao/openbao/pulls/${{ github.event.pull_request.number }}/commits | \
+          jq -f .github/scripts/verify-commits.jq
+        env:
+          GH_TOKEN: ${{ github.token }}


edomaur:
I like this :-)  I don't know Github Action very well but this looks fine. Also, this could later evolve with other code signing schemes, or with a separate test against an external signature depot. More flexible, then good :+1:

DanGhita (approved):
I really don't know how exactly all those scripts work and how they can interact with GitHub, but I was wondering if there is a way for having the list of the maintainers in one single place  (a file) that would  be shared with GitHub roles allowing sensitive operations (PR merges, etc.)

.github/scripts/verify-commits.jq
```diff
@@ -0,0 +1,23 @@
+# Expects the list of commits on a pull requests as input.
+# See: https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-commits-on-a-pull-request
+[
+  # Iterate over all commits
+  .[] |
+  # select maintainer commits
+  select(
+    .author.login == "JanMa" or
```
cipherboy, line 8: I think this doesn't solve this concern: https://github.com/openbao/openbao/issues/399#issuecomment-2267192422

In particular, I think we could perhaps more strongly enforce a three-fold condition here:

1. PRs should only have one login for a series of commits (additional authors should be moved to `Co-authored-by` tags, though note a particular user may have multiple emails associated with their GH account).
2. A PR's author should match the author of the commits.
3. All commits from maintainers should be signed.

This prevents the concern articulated there, but uniformly, for everyone. :-)

cipherboy: I think my jq-foo is weak, but perhaps:

```jq
# Expects the list of commits on a pull requests as input.
# See: https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-commits-on-a-pull-request
[
 # Iterate over all commits
 .[] |
 # Select commit authors
 .author.login
] |
# Check if there are multiple authors
if (. | length) != 1 then
 # Return error
 ("PR contains commits from multiple GitHub users!\n" | halt_error(1))
elif .[0] != "GITHUB_PR_AUTHOR" then
 # Return error
 ("PR contains commits from other than the PR author!\n" | halt_error(1))
else
 # Return success
 "All maintainer commits are signed"
end
```

(and then, add another call to `gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" /repos/openbao/openbao/pulls/447 | jq -r .user.login` and `sed` `GITHUB_PR_AUTHOR` to the desired value from this call?)

This would be in addition to the current script and enforce 1 & 2

JanMa: I think I disagree that points 1 and 2 should be implemented. It is common enough that two people collaborate on a PR and both push commits to the branch it's being worked on and we shouldn't prohibit this.

I've also seen maintainers support community members with their PRs by pushing commits to the PR branch, and when opening a PR you can explicitly opt-in to allowing this. By enforcing those rules we block ourselves from doing so in the future, and helping community members get their changes merged.

JanMa:
> I really don't know how exactly all those scripts work and how they can interact with GitHub, but I was wondering if there is a way for having the list of the maintainers in one single place (a file) that would be shared with GitHub roles allowing sensitive operations (PR merges, etc.)

Can you elaborate on what you mean by that? We could read the list of maintainers from a `maintainers.yaml` file in the project root, this just needs some massaging of the jq filter. But I am not sure what you mean with the second part of your answer?

.github/scripts/verify-commits.jq
cipherboy: Hmm, if the maintainer or anyone else is pushing commits, the result will be squashed anyways so there'll still only be one author with other commit authors moved to `Co-authored-by`. So having them set an explicit `--author=<PR Owner>` and then add a `Co-authored-by <themselves>` would end up being equivalent tbh, though admittedly more work.

But I do agree it is useful; usually I'd just take over community PRs at HashiCorp if the author had disappeared (but signed the CLA) and we wanted to merge things in  a more timely manner. 

We can skip this, but I'd still like to see some guard preventing others from pushing (even signed) commits to PRs. Otherwise, someone could take signed commits from our forks and push them in a PR (these would need to be taken verbatim, and their modifications added in subsequent commits).

cipherboy (approved):

PR #448, opened 2024-08-04
Correct annotations in injector documentation

Forgenn:
The [OpenBao documentation](https://openbao.org/docs/platform/k8s/injector/annotations/) for the injector annotations doesn't match how they are named in [openbao-k8s](https://github.com/openbao/openbao-k8s/blob/main/agent-inject/agent/annotations.go), the TLD should be .org.

Resolves -

## Target Release

1.14.7


JanMa (approved):
LGTM, thanks @Forgenn!

PR #453, opened 2024-08-07
Clarify Rollback and Commit semantics

cipherboy:
Also fixes typo in read-only transaction definition.

### DIFF
--- a/sdk/logical/storage_transactions.go
+++ b/sdk/logical/storage_transactions.go
@@ -15,7 +15,7 @@ import (
 type Transactional interface {
 	// This function allows the creation of a new interactive transaction
 	// handle, only supporting read operations. Attempts to perform write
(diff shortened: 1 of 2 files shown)


DanGhita (approved):
Great, thank you Alex!

cipherboy:
> One more remark: I suppose we should have similar explanations in the comments from the new interfaces defined in "sdk/physical/transactions.go". Could you please check the comments are all aligned throughout the layers (logical, physical etc.).

Thanks, added. FWIW, only logical will be interacted with externally; physical is only accessible within a very narrow scope of Core, before barrier encryption is applied. Most everyone, even inside Core, will interact with barrier and thus the logical definitions. Updated!

JanMa (approved):

PR #454, opened 2024-08-08
Begin drafting release checklist

cipherboy:
@JanMa @DanGhita Let me know if I'm missing anything.

### DIFF
--- a/website/content/docs/policies/release.mdx
+++ b/website/content/docs/policies/release.mdx
@@ -1,0 +1,71 @@
+---
+layout: docs
+page_title: OpenBao Release Checklist
+sidebar_label: Release Checklist
+description: |-
+  The OpenBao checklist for releases
+---
+
+# Release checklist
+
+This checklist was created after the v2.0.0 release in order to facilitate
+a better release process going forward.
+
+## Pre-release checklist
+
+ - [ ] Clone this template into a GitHub release issue.
+ - [ ] Identify target release version (major, minor, or patch) and milestones
+       (alpha, beta, ...).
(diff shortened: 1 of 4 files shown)


website/content/docs/policies/release.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+ - [ ] Tag release commit
+ - [ ] Start [release workflow](https://github.com/openbao/openbao/actions/workflows/release.yml):
+       select target tag, and set applicable pre-release/latest set
+       (pre-releases should not be marked latest on GitHub).
+ - [ ] Spot-check release artifacts:
+      - [ ] Container images work.
+      - [ ] Linux binaries work.
+      - [ ] Website download page.
+ - [ ] Open PR to update `CHANGELOG.md`.
```
JanMa, line 56: I am a fan of using the `CHANGELOG` update commit as the new version Tag, so you can directly see what changed when pulling the tag. Otherwise you need to fetch the up to date changelog from `main`.

JanMa (approved):
Small remark, but otherwise looks good :+1:

website/content/docs/policies/release.mdx
cipherboy: Ah, good idea. Then the auto-generated source code from GitHub is correct. Added :-)

cipherboy:
@DanGhita Ah, good idea. Added.

JanMa (approved):

cipherboy:
Thanks all! We can keep updating this as we have more ideas.

Issue #315, opened 2024-05-02
Remove pre-1.0 migration paths


cipherboy:
OpenBao contains a number of migration paths from pre-1.0. It is rather difficult to find all references, but some:

 - https://github.com/openbao/openbao/blob/74c2dddb0612b9a3da79384c20638266aa7de407/vault/expiration.go#L2152-L2159
 - https://github.com/openbao/openbao/blob/74c2dddb0612b9a3da79384c20638266aa7de407/command/ssh.go#L291-L315
 - https://github.com/openbao/openbao/blob/74c2dddb0612b9a3da79384c20638266aa7de407/sdk/helper/pluginutil/env.go#L65-L70

&c.

More formally, we should consider policy on what the minimum migration version is; we could say a user must come from Vault 1.14.0 series prior to moving to OpenBao (if it is not a fresh install). 

/cc @naphelps This would be a good agenda item for next week.

PR #457, opened 2024-08-09
Remove deprecated code paths

cipherboy:
This removes several code paths which should not be triggered (or supported) on OpenBao v2.0.0 as they were deprecated or removed prior to v1.14:

1. `-stored-shares` flag,
2. Legacy password handling in userpass,
3. LeaseEntry namespace handling (which also does not exist due to it being an Enterprise-only feature)
4. Auto-detection in SSH CLI
5. Legacy NetRPC support detection
6. Version check around SSCT token generation.

Resolves: #315

### DIFF
--- a/builtin/credential/userpass/path_login.go
+++ b/builtin/credential/userpass/path_login.go
@@ -5,7 +5,6 @@ package userpass
 
 import (
 	"context"
-	"crypto/subtle"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


cipherboy:
#458 will need to be merged before this one will pass tests cleanly.

JanMa:
@cipherboy could you add a changelog entry for this please? Otherwise LGTM :+1:

JanMa (approved):

cipherboy:
Ah, my bad, I had missed this. Added and rebased. Thanks @JanMa!

Issue #270, opened 2024-04-06
Rely on database layer for transaction, replication, and durability


hixichen:
Claim: The statement is my own and does not represent my company

As someone who has operated Vault for a few years, I feel that Vault's design has become trapped in a cocoon of its own making.

In its approach to modeling a generic product with an API - Controller - DB framework, Vault places itself partially within the DB's responsibilities. This includes investing significantly in features like Raft/replication and transactions, which, in my opinion, add an unnecessary burden.

Taking a quick look at the database requirements for secrets, keys, and certificates. 

- Durability: Guarantees that keys and secrets are preserved without loss, using advanced data persistence methods to withstand failures. For example, S3 offers an SLA of 11 nines, while Spanner provides a five-nine SLA.
- Reliability: Ensures continuous service availability, engineered to be always accessible to meet the demands of critical operations.
- Read Efficiency: Facilitates the swift retrieval of secrets and decryption of keys, minimizing latency to enable immediate access and use of secure data.
- Write/Update Consistency: Prioritizes strong consistency for write and update operations. Although these processes may be slower, they must ensure complete data integrity and consistency across the system.
- Multi-Tenancy and Data Isolation: Supports access for multiple users or tenants while ensuring strict data isolation, providing secure and separate environments for each tenant’s data.

it's clear that there are numerous databases on the market that already support these features, like spanner, foundationDB. 

Vault's built-in Raft and its approach to locking its controller layer and database hinder the foundational database's ability to perform cross-region replication for reading. This seems to be a strategy to push for a business license that includes cross-region replication.

I hold high hopes for this project, especially because it is truly open source. I wish it would rely more on databases for durability, reliability, and replication, rather than on Vault itself.

I suggest pushing Vault towards functioning mainly as a controller layer, where each node can handle reads and all nodes can write, assuming the database supports transactions. 

This could eliminate the need for a leader election model, the fake high availability(why you even need standby) that doesn't necessarily contribute to the system's robustness.

cipherboy:
\o Hello @hixichen, 

Thanks for the issue! This came at a good time, as we are considering a GA roadmap as a community; we welcome any additional thoughts you might have in that direction (perhaps in a separate issue?).

This reply also got rather long, so apologies in advance. :-) 

---

This has definitely been on my mind a lot lately. There's three sets of issues that I had in mind when removing all of the storage backends (see #64):

[...]

PR #467, opened 2024-08-15
Re-add Postgres storage engine

cipherboy:
This starts to re-adds the Postgres storage engine, albeit without transaction support.

We still ~need to add it to the backend binary itself (see e.g. [here](https://github.com/openbao/openbao/blob/0dcb5778fce69c6948d84d2982c3d6aa698ae425/command/server.go#L458-L464) and [here](https://github.com/openbao/openbao/blob/0dcb5778fce69c6948d84d2982c3d6aa698ae425/command/commands.go#L147-L152) and more)~ and, once the testing portion of #292 merges, we'll want to add this to the relevant test suite.

I'm reopening this in case someone else wants to pick up the work to make this Postgres backend transactional while I focus on other portions of the Transactional storage PR series.

---

Related: #270

### DIFF
--- a/changelog/467.txt
+++ b/changelog/467.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+physical/postgres: Reintroduce Postgres database for OpenBao storage, implementing paginated list support. This feature is currently in **preview** and breaking changes may occur.
+```
--- a/command/commands.go
+++ b/command/commands.go
@@ -36,6 +36,7 @@ import (
 	logicalDb "github.com/openbao/openbao/builtin/logical/database"
 	logicalKv "github.com/openbao/openbao/builtin/logical/kv"
 
+	physPostgresql "github.com/openbao/openbao/physical/postgresql"
 	physRaft "github.com/openbao/openbao/physical/raft"
 	physFile "github.com/openbao/openbao/sdk/v2/physical/file"
 	physInmem "github.com/openbao/openbao/sdk/v2/physical/inmem"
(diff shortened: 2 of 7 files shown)


cipherboy:
@JanMa @DanGhita This should be ready for review. I've tested it with the cross-testing in #292, with a small patch:

```diff
diff --git a/physical/crosstest/cross_test.go b/physical/crosstest/cross_test.go
index c542904a7..ecfa73871 100644
--- a/physical/crosstest/cross_test.go
+++ b/physical/crosstest/cross_test.go
@@ -44,6 +44,7 @@ import (
        log "github.com/hashicorp/go-hclog"
        "github.com/stretchr/testify/require"
 
+       "github.com/openbao/openbao/physical/postgresql"
        "github.com/openbao/openbao/physical/raft"
        "github.com/openbao/openbao/sdk/v2/helper/logging"
        "github.com/openbao/openbao/sdk/v2/physical"
@@ -207,6 +208,9 @@ func allPhysical(t *testing.T) (map[string]physical.Backend, func()) {
        txinmtx, err := txinm.BeginTx(ctx)
        require.NoError(t, err, "failed to start transaction from transactional in-memory backend")
 
+       // postgres
+       psql, psqlCleanup := postgresql.GetTestPostgreSQLBackend(t, logger)
+
        // Now compose multiple storage backends together!
 
        // raft + cache
@@ -299,6 +303,8 @@ func allPhysical(t *testing.T) (map[string]physical.Backend, func()) {
                        "txinmem": txinm,
 
                        "txinmtx": txinmtx,
+
+                       "psql": psql,
                }, func() {
                        os.RemoveAll(raftPureDir)
                        os.RemoveAll(raftDir)
@@ -309,6 +315,7 @@ func allPhysical(t *testing.T) (map[string]physical.Backend, func()) {
                        fileEncodingCleanup()
                        os.RemoveAll(raftCacheEncodingDir)
                        fileCacheEncodingCleanup()
+                       psqlCleanup()
                }
 }
 
```

Otherwise, the primary changes are that this is now enabled for server testing and we've marked it experimental for the time being. This will give us time to consider non-backwards-compatible changes and other extensions to the engine.

DanGhita (approved):
@cipherboy : if you are OK and it's not very urgent, I could try  to add the transaction support in a dedicated PR.

cipherboy:
> @cipherboy : if you are OK and it's not very urgent, I could try to add the transaction support in a dedicated PR.

Yes, separate PR is what I was hoping for, regardless of who does it. :-)

And thank you!

PR #469, opened 2024-08-18
Add missing spaces around link in FAQ section.

jimender2:
Simple fix here to add some missing spaces.  I think I followed the contributing guide correctly but if I did not, please let me know.

image

Resolves #

## Target Release

1.14.7

### DIFF
--- a/website/src/components/FAQSection/index.tsx
+++ b/website/src/components/FAQSection/index.tsx
@@ -43,7 +43,7 @@ const accordionData: AccordionItem[] = [
         but prior to 1.14.9 being cut.
 
         From here forward, we'll be adhering to our{" "}
-        <a href="/docs/policies/migration/#proposal">{" "}compatibility policy</a>
+        <a href="/docs/policies/migration/#proposal">compatibility policy</a>{" "}
         to provide API-compatibility with upstream when possible.
       </>
     ),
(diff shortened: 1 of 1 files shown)


JanMa (approved):
Lgtm, thanks @jimender2

website/src/components/FAQSection/index.tsx
```diff
@@ -42,8 +42,8 @@ const accordionData: AccordionItem[] = [
         prior to the BUSL. This corresponds to a few commits after 1.14.8,
         but prior to 1.14.9 being cut.
 
-        From here forward, we'll be adhering to our
-        <a href="/docs/policies/migration/#proposal">compatibility policy</a>
+        From here forward, we'll be adhering to our{" "}
+        <a href="/docs/policies/migration/#proposal">{" "}compatibility policy</a>
```
cipherboy, line 46:
```suggestion
        From here forward, we'll be adhering to our{" "}
        <a href="/docs/policies/migration/#proposal">compatibility policy</a>{" "}
```

Doesn't the current version mean we'll end up with two spaces in front and none behind?

JanMa: It does, you're right.

jimender2:
Thanks @cipherboy for that catch.

PR #475, opened 2024-08-19
Allow `secret tune` of `sys` and `identity` mounts

### LINKED ISSUE #474: Allow non-hmac audit messages tuning of `sys` and `identity`

As SecOps, I want to be able to easily parse and analyze audit logs from Vault. With hmac-encrypted messages this is rather complicated. We have compiled a list of non-sensitive fields that we do not want to be hmac'd in the audit logs but we can't tune either `sys` or `identity` mountpoints where most of these log messages are coming from.

I want to be able to set `audit_non_hmac_request_keys` and `audit_non_hmac_request_keys` for `sys` and `identity` mountpoints, ideally through a pre-existing facilites, such as _secrets tuning_.

A simplified approach might be to just remove both mountpoints from the [untunableMounts](https://github.com/openbao/openbao/blob/cf24f420ae9ac90c592ba6185e5333c700eed788/vault/mount.go#L102) though I'm not sure if there might be any undesirable side effects.

I can't think of any other use cases for _non-hmacs_ but I think other use cases might exist for tuning the mountpoints in general.

**Additional context**
I'm not mentioning how, or even if, any competing products have tried to solve a similar issue - not sure how the contributing guidelines would apply in such a case.

jficz:
_note: this PR does not (yet) fully adhere to the contribution guidelines because, in its current state, it does not have the ambitions to be merged. It is primarily meant to serve as a starting point of a discussion to solve a specific issue by a more wider approach_

`sys` and `identity` mountpoints don't allow tuning which in turn makes setting `audit_non_hmac_request_keys` and `audit_non_hmac_response_keys` impossible. Use cases [exist](https://github.com/openbao/openbao/issues/474) which may require these tunables to be available for these mountpoints.

Resolves #474 

This change is simple and naive and is (at this point) intended to be a preview or a discussion starter.
I don't know what the possible side effects might, hence the discussion mentioned above.

Naive tests:

```
bin/bao server -dev start
```

and

```
❯ ./bao secrets tune -audit-non-hmac-request-keys=name sys
Success! Tuned the secrets engine at: sys/

❯ bin/bao read sys/mounts/sys/tune
Key                            Value
---                            -----
audit_non_hmac_request_keys    [name]
...
```

## Target Release

TBA

### DIFF
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -101,9 +101,7 @@ var (
 
 	untunableMounts = []string{
 		mountPathCubbyhole,
-		mountPathSystem,
 		"audit/",
-		mountPathIdentity,
 	}
 
 	// singletonMounts can only exist in one location and are


vault/mount.go
```diff
@@ -101,9 +101,7 @@ var (
 
 	untunableMounts = []string{
 		mountPathCubbyhole,
-		mountPathSystem,
 		"audit/",
```
cipherboy, line 104: If we're being thorough... I wonder if we just change the definition of `untunable`? 

`audit/` likely suffers from the same issue: should audit path be audited (probably, yes)? Same probably holds for `mountPathCubbyhole` -- it seems like perhaps we should leave these two paths _in_, but change the usage to disallow certain tunable options?

https://openbao.org/api-docs/system/mounts/#tune-mount-configuration

Of these fields:

 - `audit_non_hmac_request_keys`
 - `audit_non_hmac_response_keys`

both make sense and none of the others do (TTL not used or would be a breaking change, description is system-set, `listing_visibility` likewise, headers don't matter as we already override that, and plugin_version is the system version).

So I think I'd probably prefer the other PR, in `vault/logical_system.go`:

https://github.com/openbao/openbao/blob/cf24f420ae9ac90c592ba6185e5333c700eed788/vault/logical_system.go#L1565-L1571

and only tune these two options, returning early if it was handled (so that other values aren't affected). My 2c!

jficz: Agree completely, this makes more sense in the context of the issue I'm trying to fix.

Speaking about being thorough though - I would even argue that setting non-hmac audit keys doesn't logically belong to `tune` at all and should be set independently as  `audit_settings` or something along those lines. But that would be a rather breaking change and I don't think such a change would be warranted on its own (it could be entertained if a larger rework of auditing is done in the future for example).

cipherboy: Hmmm, yeah, auditing is one of those weird things where its cross-role really, tbh. A separate set of paths might make some sense; perhaps we could introduce a path like `sys/mounts/:path/audit-tune` which just has those two keys? Tuning I do agree, seems to mostly be the scope of a mount-level admin, but auditing seems to more broadly require collaboration between instance/auditing admins and the admins of individual plugins being SMEs and knowing what is safe to HMAC/non-HMAC.

jficz: agreed but I think it's off-topic in this MR, created a new feature request https://github.com/openbao/openbao/issues/476

cipherboy: @jficz Just so I make sure we're on the same page :D 

You're going to update this to make sure only the hmac fields are modified by the actual tuning code and the other fields aren't, when tuning the sys/ and audit/ paths? And then we'll merge this and let the #476 be addressed later, perhaps by someone else?

jficz: I can try but it's probably not gonna be as simple as this so I can't really guarantee either code quality or an ETA :) give me a few days to familiarize myself with the codebase and I'll see what I can come up with

cipherboy: \o hey @jficz! Still interested in contributing this? Got any questions? Happy to talk through anything if you're stuck. :-)

jficz: hey @cipherboy! I am, but the project was sidelined for a while and I got other priorities atm. But I still want to return to this either in this Q or in Q1/2025. I don't want to block the work tho so if anyone else is interested in working on this, we can close the draft to not block anyone.

cipherboy: \o hey @jficz, happy Q1! :-) Are you interested in continuing this? Otherwise I might pick it up.

jficz: hi @cipherboy , _interested_ yes but it's unfortunately still quite low on my other priorities :( So feel free to pick it up, I'd be happy to look at the PR or test or just watch over your shoulder to learn, should you opt for smaller commits ;)

cipherboy: @jficz What do you say about finding an hour to meet and we'll pair program and I'll post the recording somewhere for the benefit of others as well (perhaps on YouTube for lack of anywhere better)?

jficz: @cipherboy sure, sounds good :) I'll send you an e-mail

cipherboy:
Replaced by: https://github.com/openbao/openbao/pull/921

Thanks @jficz!

PR #479, opened 2024-08-22
Disable UI tests

cipherboy:
This disables the UI tests as the present in-tree UI does not pass tests and significant work is required to fix and update it.


JanMa (approved):

cipherboy:
Nice, the pipeline works :D 

image

PR #480, opened 2024-08-22
add signed commits RFC to docs

JanMa:

Closes: #399

### DIFF
--- a/website/content/docs/rfcs/signed-commits.mdx
+++ b/website/content/docs/rfcs/signed-commits.mdx
@@ -1,0 +1,100 @@
+---
+sidebar_label: Signed commits
+description: |-
+  An OpenBao RFC for allowing users to contribute without GPG/SSH signing their commits.
+---
+
+# Discontinue enforcing signed commits
+
+**Status**: this was agreed upon in [Issue #399](https://github.com/openbao/openbao/issues/399)
+
+## Summary
+This RFC proposes discontinuing the enforcement of signed commits via branch
+protection rules in our GitHub repositories. The current practice requires every
+commit to be signed, with the intention of ensuring authenticity and authorship
+verification. However, this enforcement introduces complexities and barriers
+that may outweigh its perceived security benefits. The primary goal of this
+proposal is to streamline the contribution process, making it more accessible to
+casual contributors while maintaining adequate security practices through other
+means. This document outlines the rationale, potential impacts, and alternative
+solutions to ensure the integrity and security of our codebase.
+
+## Problem Statement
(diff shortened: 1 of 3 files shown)


website/content/docs/rfcs/signed-commits.mdx
```diff
@@ -0,0 +1,100 @@
...
+  access to necessary tools are discouraged from contributing, which reduces
+  community engagement and diversity.
+* **Temporal Validity of Verification**: The validity of commit signatures is
+  temporal. If a contributor deletes their SSH key from GitHub's UI, previously
+  "verified" commits may become "unverified," compromising the long-term
+  reliability of signature verification.
+* **Rebase Merge Issues**: Ideally, we would like to rebase pull requests when
+  merging to maintain the original commit history of the PR. However, due to a
+  long-standing bug in GitHub, all signatures in individual commits become
+  invalidated during a rebase, causing them to appear as unverified.
```
cipherboy, line 41: One minor additional nit is that we use squash merges, which re-signs the squashed version of the original commits in the series with GitHub's key, effectively deleting the contributor's own attestation.

website/content/docs/rfcs/signed-commits.mdx
```diff
@@ -0,0 +1,100 @@
...
+## Technical Description
+To implement this change, we will:
+
+1. Update the branch protection rules in GitHub to remove the requirement for
+   signed commits.
+2. Add a GitHub Actions Workflow to ensure Maintainer commits are always
+   verified.
+2. Communicate the change to all contributors through appropriate channels
+   (e.g., mailing lists, GitHub discussions).
+3. Update our contribution guidelines to reflect the new policy.
```
cipherboy, line 61: FWIW, our existing [CONTRIBUTING.md](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco-sign-off-policy) only mentions DCO and not commit GPG signing, so formalizing this policy actually brings us in line with the existing guidelines. :-)

cipherboy (approved):
Much appreciated, thanks @JanMa! I don't think the discussion on squash merges is worth holding up the merge over, but is otherwise useful to point out.

website/content/docs/rfcs/signed-commits.mdx
JanMa: It is mentioned a bit below in the `Rationale` section that we use Squash Merge

cipherboy: Oh, d'oh. ;-) Thanks! :D

PR #483, opened 2024-08-23
Ensure modifying entries do not affect cache

cipherboy:
Previously, modifications to the storage entry (after performing a put call) could impact the value saved in cache. While likely not common in actual plugins (unless a large number of writes are performed), this pattern is common in tests, to reuse a single Entry for multiple direct write operations.

Switch to creating a deep copied entry to prevent future modifications to the entry from impacting the cache.


JanMa (approved):

cipherboy:
Thanks @JanMa!

PR #487, opened 2024-08-24
Add precise control of ssh certificate validity using absolute time.

vhespanha:
This PR introduces a new `not_before_absolute` field for SSH certificate roles, allowing users to specify the exact timestamp for when an SSH certificate should become valid. It should not be merged yet, as tests and documentation are lacking.

Key changes:
1. Added the `not_before_absolute` field to the SSH role configuration.
2. Modified the certificate signing process to use the `not_before_absolute` time when set.
3. Included logic to give precedence to `not_before_absolute` over `not_before_duration` when both are present.
4. Updated relevant structs and functions in the SSH package to accommodate the new field.
5. Added error handling for invalid `not_before_absolute` time formats.

Rationale:
To quote @mehmetakbulut on [#24084](https://github.com/hashicorp/vault/issues/24084):

> 1. Many air-gapped systems without GPS and/or RTC boot up with their clocks set to the Unix epoch. These cannot accept a validity window starting at utc_now - not_before_duration.
> 2. Certificates could be generated ahead of time for key rotation or other similar exercises with a future not_before date and remain invalid until rotation occurs.

The implementation maintains backward compatibility with the existing `not_before_duration` field, ensuring that current configurations continue to work as expected. I also deliberately chose to use the name `not_before_absolute` instead of just `not_before` as suggested in [#485](https://github.com/openbao/openbao/issues/485) and [#24084](https://github.com/hashicorp/vault/issues/24084). In my eyes, this seems a little less prone to confusion, but I can always change it if needed.

Next steps:
- [ ] Include the new field in tests across the project and ensure code coverage.
- [ ] Include usage examples in the project's documentation.

I decided to open this PR as is so I can check with the repo maintainers if everything is good so far with my commits since this is my first PR here. Work on tests and documentation will begin as soon as possible.

Resolves [#485](https://github.com/openbao/openbao/issues/485), [#24084](https://github.com/hashicorp/vault/issues/24084)

1.14.7


cipherboy (commented):
Ah, interesting, I hadn't remembered correctly :D I thought PKI already had a field for this, but I guess not! Interested in opening up the corresponding (`not_before`) field on PKI as well? 

Also I think I'll have another proposed change to the PKI engine.

builtin/logical/ssh/path_roles.go
```diff
@@ -66,6 +66,7 @@ type sshRole struct {
 	AlgorithmSigner            string            `mapstructure:"algorithm_signer" json:"algorithm_signer"`
 	Version                    int               `mapstructure:"role_version" json:"role_version"`
 	NotBeforeDuration          time.Duration     `mapstructure:"not_before_duration" json:"not_before_duration"`
+	NotBeforeAbsolute          time.Time         `mapstructure:"not_before_absolute" json:"not_before_absolute"`
```
cipherboy, line 69: So I think `not_before` would be the best field name here:

1. PKI calls the field `not_before_duration`, mirroring what we have here for the other field. 
2. It doesn't have `not_before` at all (and as I mentioned above, should probably be added :D)
3. We have `not_after` and `ttl` in PKI, which to me implies the bare field should be the timestamp variant.

So, let's choose to follow the latter and call this `not_before` in both places.

builtin/logical/ssh/path_roles.go
```diff
@@ -374,6 +375,18 @@ func pathRoles(b *backend) *framework.Path {
 					Value: 30,
 				},
 			},
+
+			"not_before_absolute": {
+				Type: framework.TypeTime,
```
cipherboy, line 380: Interesting, I had not used `TypeTime` before; we had went with `TypeString` for PKI.

I think this should be compatible both ways, but I don't think you'll [ever get a string as output](https://github.com/openbao/openbao/blob/1044cd544c9f5011c16510fe13d15061132224ed/sdk/framework/field_data.go#L268-L279) below. :-)

builtin/logical/ssh/path_roles.go
```diff
@@ -512,6 +525,22 @@ func (b *backend) createCARole(allowedUsers, defaultUser, signer string, data *f
 		NotBeforeDuration:         time.Duration(data.Get("not_before_duration").(int)) * time.Second,
 	}
 
+	// HACK: I don't like this check, refactor this later
+	if notBeforeAbsoluteRaw, ok := data.GetOk("not_before_absolute"); ok {
+		switch v := notBeforeAbsoluteRaw.(type) {
```
cipherboy, line 530: I think you can just do:

```
role.NotBeforeAbsolute = notBeforeAbsoluteRaw.(time.Time)
```

and it will be safe to type assert based on the above.

cipherboy: So, `GetOk(...)` has different semantics than `Get(...)` in that if the value isn't provided, `GetOk` will return `false`, so we'll skip the assignment. We don't have a default value, which is fine, but if we call `Get(...)` and assign to `time.Time` or `*time.Time`, we'll fail: https://go.dev/play/p/TGU0JR5GIRn 

So, I think what this needs to be is, outside of the above role entry, just like you've done here, we'd do:

```go
notBeforeRaw := data.Get("not_before")
if notBeforeRaw != nil {
  role.NotBefore = notBeforeRaw.(time.Time)
}
```

(and because we're creating `role` fresh above, we'll have it set to the default zero value)

cipherboy (commented):
Looks fairly good @vhespanha! One minor naming nit and then I think the behavior isn't quite correct for the role path.

We have two categories of paths in OpenBao:

1. Mostly configuration paths (`pki/config/crl` for instance) which are PATCH-as-POST, i.e., only modify the sent fields.
2. What role paths fall into, which are complete overwrite paths coupled with optional PATCH handler implementations (I think PKI has one, but SSH doesn't). This resets to default values if a field isn't present in the request body. 

SSH's handlers fall into the second category, so this should be a regular POST/reset-to-default-if-not-present handler behavior :-) `GetOk` does the PATCH behavior.  

---

Two other small things:

1. As you mentioned in your PR: it'd be good to get a test case for this, I'd suggest `builtin/logical/ssh/backend_test.go`. The docs can be found in `website/content/api-docs/secret/ssh.mdx`. :-) 
2. We should probably add a [changelog entry](https://github.com/openbao/openbao/blob/main/changelog/README.md). 

Thanks!

mehmetakbulut:
When you add unit tests, it would be great if you could add one for Unix epoch `not_before="1970-01-01T00:00:00+00:00"`.

vhespanha:
Hey, sorry for the delay!!

I just reimplemented it from the ground up with your suggestions in mind, I'm now trying to implement the tests but I have almost no experience with that, so I'll make sure to push it all so anyone can help me if they spot any code smell.

cipherboy:
\o hey @vhespanha! Interested in finishing this up? Looks like our changes to `ssh` have caused this to conflict, sadly.

vhespanha:
Hey!
Yeah i'm really interested in finishing it, I haven't been able to do much for some time since I had some problems with my machine, but I plan on going back to this feature soon, thanks for asking.

PR #494, opened 2024-08-28
Add docker-based website viewer

cipherboy:
This adds a new `Dockerfile` and command to enable users to run the website without installing any dependencies locally besides Docker.

In my case, my work laptop lacks `npm` and I've gotten good mileage off of this instead. :-) 

It looks like between v1 and v2 of Docusaurus, they've quit provisioning the `Dockerfile` automatically.

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -1,4 +1,8 @@
 NPM := $(shell which npm)
+DOCKER?=docker
+
+# For Podman, invoke via `make docker-start DOCKER=podman DOCKER_MOUNT_ARGS=,chown=true,relabel=shared`
+DOCKER_MOUNT_ARGS?=
 
 .PHONY: start, clean, deploy, serve, install, build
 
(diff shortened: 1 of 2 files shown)


JanMa (approved):

website/Makefile
```diff
@@ -21,3 +22,7 @@ node_modules: package.json
 	@$(NPM) install
 
 install: node_modules
+
+docker-start:
+	$(DOCKER) build -t openbao-docs .
+	$(DOCKER) run -p 3000:3000 --mount "type=bind,source=$(CURDIR),destination=/opt/docusaurus,chown=true,relabel=shared" -it openbao-docs
```
cipherboy, line 28: Apparently the `chown` and `relabel` attributes only work and are required for rootless Podman.. they don't work for Docker.

cipherboy: Fixed.

cipherboy:
Thanks! Merging...

PR #495, opened 2024-08-30
Support transactions in BarrierView

cipherboy:
This converts `BarrierView` to an interface, updating definitions
everywhere to use it as such. This lets us replace it with two different
structs (`barrierView` and `transacitonalBarrierView`) depending on whether
the underlying physical storage supports transactions.

Note that `BarrierView` is a translation layer, like
`logical.LogicalStorage`, which converts from a `physical.Backend` to a
`logical.Storage` instance. This means `BarrierView` is not itself
stackable.

`BarrierView` is used to ensure mounts only have local access to storage
and cannot impact other plugins' data.


cipherboy:
@DanGhita This is also ready for review :-)

DanGhita (approved):
LGTM

PR #497, opened 2024-08-30
Add transactions to plugin GRPC layer

cipherboy:
This adds support for transactions to the plugin GRPC layer, opportunistically upgrading the client storage implementation to a transactional aware one if the underlying storage implementation on the server supports it. This retains bidirectional interoperability with upstream, allowing plugins compiled with OpenBao's SDK to continue to work with upstream server implementations and visa-versa.

---

Split from #292.


cipherboy:
@DanGhita This would be ready for review if you want :-)

cipherboy:
I was testing this with the KV plugin as an external plugin and it doesn't seem to be detecting it as transactional, so I'll have to revise this.

cipherboy:
Ok, you need to make this change to KV to test:

```diff
diff --git a/builtin/logical/kv/backend.go b/builtin/logical/kv/backend.go
index e29a87ef95..125d258b8b 100644
--- a/builtin/logical/kv/backend.go
+++ b/builtin/logical/kv/backend.go
@@ -81,9 +81,9 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
        var b logical.Backend
        var err error
        switch version {
-       case "1", "":
+       case "1":
                return LeaseSwitchedPassthroughBackend(ctx, conf, conf.Config["leased_passthrough"] == "true")
-       case "2":
+       case "2", "":
                b, err = VersionedKVFactory(ctx, conf)
        }
        if err != nil {
```

You can then build OpenBao with this patch and explicitly build the plugin, load it into OpenBao, and execute it.

```
$ mkdir -p bin/plugins/secret
$ go build -o bin/plugins/secret/kv -v github.com/openbao/openbao/builtin/logical/kv/cmd/kv
$ devbao node stop prod
$ killall bao
$ sleep 1
$ devbao node start --initialize --unseal --audit --force

$ sleep 1

$ . <(devbao node env prod)

$ cp bin/plugins/secret/kv /home/cipherboy/.local/share/devbao/nodes/prod/plugins/bao-secret-kv-custom

$ sleep 1

$ bao plugin register -sha256=$(sha256sum bin/plugins/secret/kv | awk '{print $1}') secret bao-secret-kv-custom
$ bao secrets enable -path=grpc-txn-kv bao-secret-kv-custom

$ echo '{"data":{"key": "value"}}' | bao write grpc-txn-kv/data/asdf -
$ bao read grpc-txn-kv/data/asdf
```

If you modify the KV plugin to log when it has a TXNable storage, you'll see this uses that and otherwise behaves correctly.

DanGhita (approved):
LGTM

PR #498, opened 2024-08-30
Add transactions to PKI

cipherboy:
This adds transactions to PKI in two places:

1. Root generation
2. Intermediate CA import

Notably, because the CRL rebuild isn't strictly important (though, if it fails, we'd prefer to avoid writing the CAs to storage), we do this in a non-transaction backed storage (if available).


DanGhita (approved):
LGTM

Issue #496, opened 2024-08-30
RFC - Post-Quantum Cryptography Migration Roadmap


cipherboy:
### Summary

In [August 2024](https://www.nist.gov/news-events/news/2024/08/nist-releases-first-3-finalized-post-quantum-encryption-standards), NIST published the results of its [Post-Quantum Cryptographic Standardization competition](https://csrc.nist.gov/Projects/post-quantum-cryptography/post-quantum-cryptography-standardization). This started the process for industry applications (and no longer just early adopters and cryptographic library authors) to begin migrating their applications to PQC now that presumably secure algorithms are finalized.

While OpenBao's critical storage largely remains secure (depending on unseal mode: anything using purely symmetric ciphers like Shamir's will be fine), we have numerous places where a harvest-now-decrypt-later (HNDL) attack could impact us or places where more time is beneficial in quantum rotation.

### Problem Statement

OpenBao needs to be hardened against quantum adversaries: even if one doesn't materialize in a reasonable time frame, there is still risk to classical algorithms as both RSA and EC have been shown not to rely on as hard of problems as once thought. While in no means immediately breakable, we nonetheless wish to aid in organizations migrations to PQC.

[Mosca's theorem](https://utimaco.com/service/knowledge-base/post-quantum-cryptography/what-mosca-theorem) attempts to allow organizations to quantity their time horizon for a quantum migration with the equation `(X+Y)>Z`:

 - `X`, the time you need the encryption to be secure for,
 - `Y`, the time the migration will take,
 - `Z`, the time at which large-scale quantum adversaries (or, equivalently, attacks in classical algorithms) will be available.

For an organization to remain secure, `(X+Y)` must be strictly less than `Z`. Since `Z` is unknown, it is important to aid organizations wishing to minimize the overall transition time (`Y`, the most direct variable they control). Thus, we should start considering our own quantum roadmap.

### User-facing description

Some changes will be immediately available to end-users: different encryption algorithms will be selectable as candidates in various points of the process. For instance, [Transit](https://openbao.org/docs/secrets/transit/#key-types) and [PKI](https://openbao.org/api-docs/secret/pki/#generate-root) might allow pure post-quantum algorithms or various hybrid constructions. Other changes might be strictly internal, like selection of [key type](https://github.com/openbao/openbao/blob/57669cff41f5b105bede654347f7038667eac7a1/vault/cluster.go#L211-L220) for cross-node identity in Seal HA.

[...]

PR #502, opened 2024-09-03
Update Go version to 1.23.2

cipherboy:
Go 1.23.0 was released on August 13th.

Notably, related to #496, it includes:

> The experimental post-quantum key exchange mechanism X25519Kyber768Draft00 is now enabled by default when [Config.CurvePreferences](https://go.dev/pkg/crypto/tls#Config.CurvePreferences) is nil. The default can be reverted by adding tlskyber=0 to the GODEBUG environment variable.

However, this is implemented in an internal package (`crypto/internal/mlkem768`) presently and not exported for general use.


cipherboy:
OK, this should be fixed and ready for review now.

cipherboy:
@DanGhita I think you approved this one before I had it working (CI was broken). Mind taking a second look? :-)

DanGhita (approved):

PR #504, opened 2024-09-03
Update to Go 1.22.6 toolchain for v2.0.1

cipherboy:
It looks like from #502 that Go 1.23.0 might have some changes that will impact us; update only to v1.22.6 for this release.

This fixes go-releaser to use the project-specified one (that the Makefile and other GitHub actions uses) and sets the new version to the latest 1.22.x series toolchain, 1.22.6.


cipherboy:
Build passed, but adding changelog: 

image

PR #505, opened 2024-09-03
Update Docker dependency

cipherboy:
Docker is not used in the release binary:

    $ go tool nm bao | grep -i '\(docker\|moboy\)'
    $

But it is still worthwhile to update the version pinned in our release go.mod file anyways.

mholt/archiver and aws/aws-sdk-go are both still vulnerable without fixes. The former is limited to scenarios when running against a server (with authentication) to build a debug package (which occurs locally on the client).


JanMa (approved):

cipherboy:
@siepkes This broke Illumos / Solaris support. We can upgrade to v26/v27 with a fair bit of work (`types.ContainerListOptions` seems to have disappeared, will have to figure out how to replace that since we use it), but without Illumos/Solaris CI to tell us, we end up breaking releases.

Is there a Illumos CI runner you'd be willing to contribute? Otherwise, there are [very few downloads](https://tooomm.github.io/github-release-stats/?username=openbao&repository=openbao) (`<10`) for Solaris or Illumos so we might be more inclined to download it.

cc: @JanMa / @DanGhita

siepkes:
@cipherboy Thanks for the headsup! I don't have access to a Github illumos CI runner unfortunately. I'm also not really familiar with them. I guess it's manual building on illumos for now. I will try to keep an eye on the build-ability and submit patches when possible.

cipherboy:
@siepkes Yeah, I'm partial to GH Actions myself, but we could always use something like Jenkins or the MacOS runner + virtualization to achieve this as well. See #711 for some info.

But, thank you!

Issue #510, opened 2024-09-05
`output-curl-string` flag returns a string that depends on the `vault` client instead of the `bao` client


Btijmen:
**Describe the bug**
When using the flag `-output-curl-string` with any `bao` command, it returns the curl string that would correspond to the command. This is useful in troubleshooting. However, the string includes a function to get the token, and that function references the `vault` client as opposed to the `bao` client. Clearly, that doesn't work.

**To Reproduce**

Run `bao status -output-curl-string`

observe the following output:

`curl -H "X-Vault-Request: true" -H "X-Vault-Token: $(vault print token)" https://127.0.0.1:8200/v1/sys/seal-status`

the `$(vault print token)` bit is problematic and should be `$(bao print token)`

**Expected behavior**
I expect the `bao` client to depend on `bao` and not `vault`.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.0.0
* OpenBao CLI Version (retrieve with `bao version`): OpenBao v2.0.0 (700fe3f27ab1f0ec39ce20c36f6d9d97c9fe6ac3), built 2024-07-17T22:05:40Z
* Server Operating System/Architecture: server: linux; workstation: mac

OpenBao server configuration file(s):

```hcl
disable_cache = false

cluster_addr = "<scrubbed>"
api_addr = "<scrubbed>"

ui = true

listener "tcp" {
  address         = "[::]:8200"
  cluster_address = "[::]:8201"
  tls_cert_file = "<scrubbed>"
  tls_key_file = "<scrubbed>"
}


storage "raft" {
  path = "/data/raft"
  node_id = "raft_node_1"
}

max_lease_ttl = "10h"
default_lease_ttl = "10h"
cluster_name = "testcluster"
pid_file = "./pidfile"
raw_storage_endpoint = true
introspection_endpoint = true
disable_sealwrap = true
disable_printable_check = true
enable_response_header_hostname = true
enable_response_header_raft_node_id = true
```

**Additional context**
This fix is trivial and I have a [commit](https://github.com/openbao/openbao/compare/main...Btijmen:openbao:main) in my own fork, but to open a PR I needed to create an issue first.

cipherboy:
Fixed by #511. Thanks @Btijmen!

PR #511, opened 2024-09-05
fix command `vault print token` to `bao print token` in curl string generated by `buildCurlString()` 

Btijmen:
This is a trivial fix to ensure the curl string generated by `buildCurlString()`, invoked by the `-output-curl-string` flag, actually uses the `bao` client instead of attempting to use the `vault` client.

## Resolves

https://github.com/openbao/openbao/issues/510

## Target Release

Next minor release

### DIFF
--- a/api/output_string.go
+++ b/api/output_string.go
@@ -81,7 +81,7 @@ func (d *OutputStringError) buildCurlString() (string, error) {
 	for k, v := range d.Request.Header {
 		for _, h := range v {
 			if strings.ToLower(k) == "x-vault-token" {
-				h = `$(vault print token)`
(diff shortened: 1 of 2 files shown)


Btijmen:
This has to be the shortest PR I ever filed anywhere :D

Btijmen:
Apologies for the force push, forgot to sign off the commit at first.

cipherboy (approved):
Ah, another missing place. Thanks @Btijmen!

And no worries, force pushes are all good.

cipherboy:
@Btijmen Actually, since this is a bug fix, do you mind adding a changelog entry before I merge?

Issue #509, opened 2024-09-05
"invalid memory address or nil pointer dereference" panic on malformed AppRole login


mossblaser:
I have stumbled upon a way to trigger a panic using the /auth/approle/login API endpoint of OpenBao using a simple malformed request.

**Describe the bug**

In a request to `/auth/approle/login`, setting either the `role_id` or `secret_id` parameters to an object (i.e. `{}`) rather than a string triggers the issue. This results in the connection being closed to the client without a response being sent. On the server side, a log message beginning with the following (followed by a stack trace) is seen:

     [INFO]  http: panic serving 127.0.0.1:54504: runtime error: invalid memory address or nil pointer dereference
    <...stack trace...>

<details>
<summary>Full output with stack trace (click to expand)</summary>

```
2024-09-05T08:12:48.249+0100 [INFO]  http: panic serving 127.0.0.1:54504: runtime error: invalid memory address or nil pointer dereference
goroutine 945 [running]:
net/http.(*conn).serve.func1()
        /opt/hostedtoolcache/go/1.22.6/x64/src/net/http/server.go:1903 +0xbe
panic({0x3428060?, 0x649f2d0?})
        /opt/hostedtoolcache/go/1.22.6/x64/src/runtime/panic.go:770 +0x132
github.com/openbao/openbao/vault.(*Core).aliasNameFromLoginRequest(0xc000806e08, {0x45622f8, 0xc000d6f530}, 0xc00080c1a0)
        /home/runner/work/openbao/openbao/vault/core.go:3692 +0x37a
github.com/openbao/openbao/vault.(*Core).getLoginUserInfoKey(0xc001104540?, {0x45622f8?, 0xc000d6f530?}, 0xc00059b180, 0xc001104540?)
        /home/runner/work/openbao/openbao/vault/request_handling.go:1710 +0x29
github.com/openbao/openbao/vault.(*Core).isUserLocked(0xc000806e08, {0x45622f8, 0xc000d6f530}, 0xc00059b180, 0xc000b265d0?)
        /home/runner/work/openbao/openbao/vault/request_handling.go:1763 +0x3f
github.com/openbao/openbao/vault.(*Core).handleLoginRequest(0xc000806e08, {0x45622f8, 0xc000d6f530}, 0xc00080c1a0)
        /home/runner/work/openbao/openbao/vault/request_handling.go:1290 +0xae6
[...]

PR #512, opened 2024-09-06
Fix aliasNameFromLoginRequest panic

cipherboy:
In `aliasNameFromLoginRequest`, a malformed request could be routed to the plugin, which resulted in a `logical.ErrorResponse(...)` being returned. This is a "secret" typed response and not an auth response like we're expecting. We'd thus panic:

> 2024-09-05T08:12:48.249+0100 [INFO]  http: panic serving 127.0.0.1:54504: runtime error: invalid memory address or nil pointer dereference
> goroutine 945 [running]:
> net/http.(*conn).serve.func1()
>         /opt/hostedtoolcache/go/1.22.6/x64/src/net/http/server.go:1903 +0xbe
> panic({0x3428060?, 0x649f2d0?})
>         /opt/hostedtoolcache/go/1.22.6/x64/src/runtime/panic.go:770 +0x132
> github.com/openbao/openbao/vault.(*Core).aliasNameFromLoginRequest(0xc000806e08, {0x45622f8, 0xc000d6f530}, 0xc00080c1a0)
>         /home/runner/work/openbao/openbao/vault/core.go:3692 +0x37a
> github.com/openbao/openbao/vault.(*Core).getLoginUserInfoKey(0xc001104540?, {0x45622f8?, 0xc000d6f530?}, 0xc00059b180, 0xc001104540?)
>         /home/runner/work/openbao/openbao/vault/request_handling.go:1710 +0x29

Instead, update this handler to properly return both error cases: an error from the request handling and a `logical.ErrorResponse(...)`, converting the latter into a regular error again.

This also addresses a case in `doResolveRoleLocked` where a similar panic could potentially occur, though without the error resolution.

Resolves: #509

### DIFF
--- a/changelog/512.txt
+++ b/changelog/512.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Fix server panic on AppRole login requests with invalid parameter typing
+```
--- a/vault/core.go
+++ b/vault/core.go
@@ -3643,7 +3643,7 @@ func (c *Core) doResolveRoleLocked(ctx context.Context, mountPoint string, match
 		Data:       data,
 		Storage:    c.router.MatchingStorageByAPIPath(ctx, mountPoint+"login"),
 	})
-	if err != nil || resp.Data["role"] == nil {
+	if err != nil || resp == nil || resp.Data == nil || resp.Data["role"] == nil {
 		return ""
 	}
 
@@ -3686,9 +3686,20 @@ func (c *Core) aliasNameFromLoginRequest(ctx context.Context, req *logical.Reque
 		Data:       req.Data,
 		Storage:    c.router.MatchingStorageByAPIPath(ctx, req.Path),
 	})
-	if err != nil || resp.Auth.Alias == nil {
+
+	// Certain errors (like field validation errors) are returned as
+	// logical.ErrorResponse(...)s, which means Auth will not be set,
+	// only resp.Data["error"]. Surface the errors appropriately.
+	if err != nil || resp == nil {
+		return "", err
+	}
+	if resp.Data != nil && len(resp.Data["error"].(string)) > 0 {
+		return "", errors.New(resp.Data["error"].(string))
+	}
+	if resp.Auth == nil || resp.Auth.Alias == nil {
 		return "", nil
 	}
+
 	return resp.Auth.Alias.Name, nil
 }
 
--- a/vault/external_tests/approle/alias_name_panic_test.go
+++ b/vault/external_tests/approle/alias_name_panic_test.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2024 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package approle
+
+import (
+	"testing"
+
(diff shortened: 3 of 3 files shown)


changelog/512.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
```
JanMa, line 1: Could you describe the effect of the fix rather than what you changed in which function? Something in the likes of 

> fixed panic on invalid AppRole login request

JanMa (changes requested):

changelog/512.txt
cipherboy: I'm not quite sure the extent of the implications of this, hence why I worded it like I did :-) 

I _think_, theoretically, this could occur on any login method. However, the `userpass` method doesn't work as `username` is forced to be a string by virtue of it being part of the URL. CertAuth likewise doesn't use any request data. [Radius is maybe applicable](https://openbao.org/api-docs/auth/radius/#login) as `username` in the URL is optional.... so it is probably on a case-by-case basis, but definitely broader than AppRole.

However, on the same type of request that panics AppRole, applied to Radius, I don't get a panic:

```
$ curl     --request POST     --header "Content-Type: application/json"     --data '{"username": {}, "password": ""}'     'http://localhost:8200/v1/auth/radius/login'
{"errors":["Field validation failed: error converting input map[] for field \"username\": '' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[]'"]}
```

(despite AppRole still panicing).

The same also holds for Kubernetes, which might theoretically also be applicable:

```
$ bao auth enable kubernetes
Success! Enabled kubernetes auth method at: kubernetes/
$ curl     --request POST     --header "Content-Type: application/json"     --data '{"jwt": {}, "role": {}}'     'http://localhost:8200/v1/auth/kubernetes/login'
{"errors":["Field validation failed: error converting input map[] for field \"jwt\": '' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[]'"]}
```

But AppRole on this server definitely crashes:

```
$ curl     --request POST     --header "Content-Type: application/json"     --data '{"role_id": {}, "secret_id": ""}'     http://localhost:8200/v1/auth/approle/login
curl: (52) Empty reply from server
```

So I think perhaps we're at:

```
fix server panic on login requests with invalid parameter typing
```

in the assumption it is broader than just AppRole, but without having concrete confirmation?

cipherboy: Ah, I understand now.

`c.isUserLocked` calls `c.getLoginUserInfoKey` which calls the affected `c.aliasNameFromLoginRequest`, but the first method is only called when user lockout is not disabled. We define lockout-able mount types in `configutil.GetSupportedUserLockoutsAuthMethods()` which currently is defined as `"userpass", "approle", "ldap"`. Both userpass and LDAP have the username in the request path, which forces it to be a string and thus match the type of the field data.

Other mounts don't panic, as this alias handler isn't called at all, and thus they return a regular error when the actual request handler is called.

Thus only AppRole is affected.

JanMa (approved):

PR #515, opened 2024-09-08
(secrets/pki): add not_before parameter

### LINKED ISSUE #490: PKI - Add not_before to align with SSH

In #485, we discuss adding `not_before` to the SSH API. We should add this as well to the PKI API, so that callers who need precise control over `not_before` can have it. Like the existing `not_after` in PKI (and recognizing #489), we should probably implement it similarly, where the field exists both on the role and on the issue endpoints.

n/a

n/a

**Additional context**

n/a

Nerkho:
I went ahead and created a first draft PR.

I let the existing `not_before_duration` but made it so that `not_before` take precedence over it. As with `not_before_duration`, `not_before` has currently no impact on TTL. I am not sure if we should change that. 
Maybe we should even get rid of `not_before_duration` all together?

I noticed there was no existing test in for `getCertificateNotAfter`. Anyway, I added a very simple one for `getCertificateNotBefore`. I need to double-check if we need some other test somewhere else.

Resolves #490 

(I also noticed my editor made some linting on save, hope it's fine; otherwise I can revert it).

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -226,6 +226,7 @@ func TestPKI_DeviceCert(t *testing.T) {
 		"allow_bare_domains": true,
 		"allow_subdomains":   true,
 		"not_after":          "9999-12-31T23:59:59Z",
+		"not_before":         "1900-01-01T00:00:00Z",
 	})
 	if err != nil {
 		t.Fatal(err)
@@ -252,6 +253,10 @@ func TestPKI_DeviceCert(t *testing.T) {
 	notAfter = cert.NotAfter.Format(time.RFC3339)
 	if notAfter != "9999-12-31T23:59:59Z" {
 		t.Fatal(fmt.Errorf("not after from certificate  is not matching with input parameter"))
+	}
+	notBefore := cert.NotBefore.Format(time.RFC3339)
+	if notBefore != "1900-01-01T00:00:00Z" {
+		t.Fatal(fmt.Errorf("not before from certificate  is not matching with input parameter"))
 	}
 }
 
@@ -3868,6 +3873,7 @@ func TestReadWriteDeleteRoles(t *testing.T) {
 		"allowed_domains_template":           false,
 		"allow_token_displayname":            false,
 		"country":                            []interface{}{},
+		"not_before":                         "",
 		"not_after":                          "",
 		"postal_code":                        []interface{}{},
 		"use_csr_common_name":                true,
--- a/builtin/logical/pki/cert_util.go
+++ b/builtin/logical/pki/cert_util.go
@@ -1434,12 +1434,32 @@ func generateCreationBundle(b *backend, data *inputBundle, caSign *certutil.CAIn
 		PostalCode:         strutil.RemoveDuplicatesStable(data.role.PostalCode, false),
 	}
 
+	// Get certificate's Not Before
+	notBefore, err := getCertificateNotBefore(data)
+	if err != nil {
+		return nil, warnings, err
+	}
+
 	// Get the TTL and verify it against the max allowed
 	notAfter, ttlWarnings, err := getCertificateNotAfter(b, data, caSign)
 	if err != nil {
 		return nil, warnings, err
 	}
 	warnings = append(warnings, ttlWarnings...)
+
+	// Verify that notBefore is older than notAfter.
+	if notBefore.After(notAfter) {
+		return nil, nil, errutil.UserError{
+			Err: fmt.Sprintf("The certificate's Not Before (%v) is later than the certificate's Not After (%v)", notBefore.String(), notAfter.String()),
+		}
+	}
+
+	// Disallow zero duration certificate.
+	if notBefore.Equal(notAfter) {
+		return nil, nil, errutil.UserError{
+			Err: fmt.Sprintf("The certificate's Not Before (%v) is equal to the certificate's Not After (%v)", notBefore.String(), notAfter.String()),
+		}
+	}
 
 	// Parse SKID from the request for cross-signing.
 	var skid []byte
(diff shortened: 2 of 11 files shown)


cipherboy (commented):
> (I also noticed my editor made some linting on save, hope it's fine; otherwise I can revert it).

There is a Makefile target, `make fmt`, which will apply the project's linting. We use `gofumpt`, a variant of the default `go fmt`.

builtin/logical/pki/path_issue_sign.go
```diff
@@ -77,6 +77,11 @@ func buildPathIssue(b *backend, pattern string, displayAttrs *framework.DisplayA
 								Description: `Serial Number`,
 								Required:    true,
 							},
+							"not_before": {
```
cipherboy, line 80: I hadn't thought about adding it to the output during issuance, I suppose that's fine :-)

sdk/helper/certutil/helpers.go
```diff
@@ -887,9 +887,13 @@ func createCertificate(data *CreationBundle, randReader io.Reader, privateKeyGen
 		return nil, errutil.InternalError{Err: fmt.Sprintf("error getting subject key ID: %s", err)}
 	}
 
+	if data.Params.NotBeforeDuration > 0 && data.Params.NotBefore.IsZero() {
+		data.Params.NotBefore = time.Now().Add(-1 * data.Params.NotBeforeDuration)
+	}
```
cipherboy, line 892: If I'm not mistaken, if `NotBefore.IsZero() && NotBeforeDuration == 0`, we end up with `NotBefore = 0`, unlike the current behavior of `NotBefore = -30s`? 

I'd probably prefer to keep existing behavior, and let the caller manually decide on a `NotBefore` that closely aligns with zero seconds for this case. In general, I'd expect there to be some amount of clock skew.

website/content/api-docs/secret/pki.mdx
```diff
@@ -1736,7 +1751,7 @@ corresponding full CRL when it has been regenerated; otherwise, some serial
 numbers may not appear in the local copy of the full CRL if the remote
 complete and delta CRLs has been regenerated.
 
-Endpoints with source `local` only include cluster-local revocations.
```
cipherboy, line 1739: This whitespace I'm fine removing :-D

cipherboy (commented):
Looks fairly good, @Nerkho! 

Some minor commentary.

changelog/512.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+secrets/pki: add `not_before` parameter to precisely define a certificate's "not before" field.
```
cipherboy, line 2: I think this needs to be in file `515.txt`, sorry :-)

builtin/logical/pki/cert_util.go
```diff
@@ -1434,13 +1434,26 @@ func generateCreationBundle(b *backend, data *inputBundle, caSign *certutil.CAIn
...
+
 	// Get the TTL and verify it against the max allowed
 	notAfter, ttlWarnings, err := getCertificateNotAfter(b, data, caSign)
 	if err != nil {
 		return nil, warnings, err
 	}
 	warnings = append(warnings, ttlWarnings...)
 
+	// Verify that notBefore is older than notAfter.
+	if notBefore.After(notAfter) {
```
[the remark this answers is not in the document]
Nerkho: I now added that as well :)

builtin/logical/pki/fields.go
```diff
@@ -481,16 +493,16 @@ in most operational scenarios).`,
 	fields["tidy_acme"] = &framework.FieldSchema{
 		Type: framework.TypeBool,
 		Description: `Set to true to enable tidying ACME accounts,
-orders and authorizations.  ACME orders are tidied (deleted) 
+orders and authorizations.  ACME orders are tidied (deleted)
```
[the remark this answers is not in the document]
Nerkho: I've been trying out Zed the last few weeks. Needs some adjustments :D

I've cleaned up the mess now.

sdk/helper/certutil/helpers.go
Nerkho: You're right. I've adjusted this bit so that `NotBefore` defaults to `-30s`. 

If `not_before` is set, it takes precedence over `not_before_duration`.

cipherboy: If I were to nitpick, I would ask for the opposite conditional ordering (`if !notBefore.isZero()`) to mirror the precedence, but I don't think it matters too much :-)

cipherboy (approved):
Looks good from my side if tests pass! Thanks @Nerkho!

cipherboy:
@Nerkho ah yes, `TestReadWriteDeleteRoles` needs to be updated. This was added I think to catch a nasty serialization bug.

builtin/logical/pki/backend_test.go
```diff
@@ -226,6 +226,7 @@ func TestPKI_DeviceCert(t *testing.T) {
 		"allow_bare_domains": true,
 		"allow_subdomains":   true,
 		"not_after":          "9999-12-31T23:59:59Z",
+		"not_before":         "9998-12-31T23:59:59Z",
```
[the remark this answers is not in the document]
Nerkho: Replaced it with `1900-01-01T00:00:00Z`. 

With `0001-01-01T00:00:00Z` the test would fail. I believe that's not a valid date.

Nerkho:
> @Nerkho ah yes, `TestReadWriteDeleteRoles` needs to be updated. This was added I think to catch a nasty serialization bug.

Just added that as well :)

cipherboy:
>  Maybe we should even get rid of `not_before_duration` all together?

No, I think `not_before_duration` is valid. IMHO, if I were to be overly pedantic, I'd say that `ttl` is equivalent to `not_before_duration` but from a server time PoV: `not_before_duration`/`not_before` is a grace period for which certificates can be backdated (or, I suppose, forward dated with an explicit `not_before`), and `ttl`/`not_after` is the expiration, relative to OpenBao server time, of the certificate. In an ideal world with no forward-dating and no clock skew, `not_before_duration=0s` and `ttl=<actual validity period>`, but in a less-than-ideal world with skew, we need some minimal backdating to Do The Right Thing (TM). 

_Presumably_, if there's clock skew in the system, it would be from various PKI consumers deviating from the OpenBao server and this OpenBao server would be synced close to "real" time (for some definition of real). So, ttl as an idealized notion of "now until expiry" on the server makes sense and IMO, doesn't make sense to decrease TTL by the backdating except in certain circumstances (like WebPKI's 90-day strict validity limits, but here you can decrease TTL by the backdate window). 

My 2c!

cipherboy (approved):
Looks good to me, thanks @Nerkho!

Issue #189, opened 2024-03-08
Fork - Remove UI for Alpha Release


cipherboy:
Per discussion in https://github.com/openbao/openbao/issues/74#issuecomment-1983946480:

@naphelps said:
> Per 2024-03-07 community call, the UI will not be shipped as an Alpha release deliverable. This will give the community more time ahead of the future GA release to remediate the remaining issues.

We need to update the `build.yml` pipeline to remove the UI from the build.

PR #517, opened 2024-09-09
Add Login MFA support for Okta TOTP

cognifloyd:
This PR adds support to Login MFA for Okta TOTP. I had to restore most of a test file deleted in dc92bbceb2c6561a182479e6c547ca928ea6620a in which only one test used the auth/okta plugin; the rest of the tests were for the Login MFA bits which were not removed from openbao.

I asked about porting my vault PR to openbao here: https://github.com/orgs/openbao/discussions/64#discussioncomment-9756535 (response here: https://github.com/orgs/openbao/discussions/64#discussioncomment-9765668)

Login MFA is a feature that has been part of open source vault (and now "community" vault) since version 1.10. But, Login MFA providers cannot be defined in an external plugin right now. So, any changes to Login MFA have to be in core. Eventually, it would be nice to split the okta/duo/pingid bits out of core openbao, but that will require creating or extending a plugin type and the plugin type will need a way to inject javascript logic in the web UI.

This rebases this backport branch of my vault PR: [hashicorp/vault#24563](https://www.github.com/hashicorp/vault/pull/24563) (_link has `www.` to prevent github's auto-backref from spamming that PR_)
- `release/1.14.x`: [backport/loginmfa-okta-totp/1.14.x](https://github.com/cognifloyd/vault/tree/backport/loginmfa-okta-totp/1.14.x) (MPL 2.0)
To be clear, I always intended this code to be available under MPL 2.0: https://www.github.com/hashicorp/vault/pull/24563#issuecomment-1858705603

The vault issue that this addresses is [hashicorp/vault#24562](https://www.github.com/hashicorp/vault/issue/24562).

## Target Release

2.1.0

### DIFF
--- a/vault/external_tests/identity/login_mfa_duo_test.go
+++ b/vault/external_tests/identity/login_mfa_duo_test.go
@@ -6,7 +6,6 @@ package identity
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"testing"
 
 	"github.com/openbao/openbao/api/v2"


cognifloyd:
This is a draft until I have time to actually run the tests `vault/external_tests/identity/login_mfa_okta_test.go` (which I had to reintroduce in this PR, except for the auth/okta plugin test). I ran them with the vault PR when I wrote this, but I have not run the tests with openbao yet.

TODO:
- [x] port the PR to openbao
- [ ] Actually test the PR with openbao

vault/external_tests/identity/login_mfa_okta_test.go
```diff
@@ -0,0 +1,333 @@
...
+	"github.com/openbao/openbao/helper/testhelpers"
+	logicaltest "github.com/openbao/openbao/helper/testhelpers/logical"
+	vaulthttp "github.com/openbao/openbao/http"
+	"github.com/openbao/openbao/sdk/logical"
+	"github.com/openbao/openbao/vault"
+)
+
+// To run these tests, set the following env variables:
+// VAULT_ACC=1
+// OKTA_ORG=dev-219337
```
cipherboy, line 27: This is one of the reasons I was originally pushing for open-only tests :-) It is hard for those of us without Okta org accounts to run this integration test suite. It doesn't look like there's a mock Okta-for-testing container officially published, but putting this in an external plugin makes more sense to me (as it can then be updated and fixed independently of the core binary).

But I do also understand that MFA hasn't typically been plugin-ized, so it is probably hard to do. :-)

cipherboy (commented):
@cognifloyd Wow! This is rather comprehensive.

Some initial thoughts:

I don't know if you've seen the [draft roadmap](https://gist.github.com/cipherboy/6654ad34eb07e2ff820fcb26bb137ef3) that I've proposed to our TSC, but one of the big items is a plugin registry. I had initially forgotten about MFA plugins (but covered auth, secret, audit, and go-kms-wrapper plugins!), but what do you think about working on these in parallel?

The [UI was removed](https://github.com/openbao/openbao/issues/189) prior to Alpha; so the work to do the plugin framework (+ allow plugin-indicated JS) wouldn't be all that bad, insofar as, if the plugin can indicate JS and we expose that _somehow_, we can likely punt on the actual integration of that until later. As you can tell, it exists, but doesn't run (well). 

As I mentioned there, I'm still a little hesitant to add more proprietary providers... But given this is a feature for an existing provider, I'm OK to add it.

cognifloyd:
> And it looks like the test removal was accidental, my bad!

At a glance, that file DID use the auth/okta plugin, so it made sense to drop it given what you knew. You would have had to read the tests to see that only the first test in the file actually used it. That test, it turns out, wasn't even a test for Login MFA, so it really didn't belong in that file any way. This commit drops that test: https://github.com/openbao/openbao/pull/517/commits/d4053684eca6e90b6af4b3ead3a386f56fb36cd6

helper/identity/mfa/types.proto
```diff
@@ -85,6 +85,8 @@ message OktaConfig {
 	string base_url = 4;
 	// @inject_tag: sentinel:"-"
 	bool primary_email = 5;
+	// @inject_tag: sentinel:"-"
+	bool use_passcode = 6;
```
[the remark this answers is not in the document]
cognifloyd: Oh. Hmm. I'm a little fuzzy on protobuf and backwards compatibility guarantees.

What happens if upstream ever gets around to merging my PR? Then the `use_passcode` field would have both `6` upstream and `666` (or whatever) in openbao. Would moving from `666` to `6` at some point be a problem? Or is there a way in protobuf to say that this tag is an alias?

vault/external_tests/identity/login_mfa_okta_test.go
cognifloyd: Yeah. This is an ugly test setup.

Creating a dev org/account with okta is free (free as in "no charge"). Maybe there's a way to create an org just for openbao's usage in CI, and put the relevant secrets in GHA? The test in question could also run only on a schedule (once a week maybe) to avoid hitting API usage limits.

The biggest gotcha, though, is having a device with the okta app installed. I don't think there's an android runner in GHA, so I'm not sure how that would work.

cipherboy: My favorite hack is that the MacOS runner supports virtualization, so if you can fully script the setup, you could put Android in a VM... :laughing: I don' recommend that idea though.

cipherboy:
> This commit drops that test: [d405368](https://github.com/openbao/openbao/commit/d4053684eca6e90b6af4b3ead3a386f56fb36cd6)

If we get around to pushing the Okta auth method to another repo, we should remember to shuffle that into the correct place. :-)

helper/identity/mfa/types.proto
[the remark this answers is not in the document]
cipherboy: I think users would have to re-compile to use the new version if we blanket update the server. But if the field means and does the same thing, I think we might be able just to write it into both places (server->client) or accept either place (client->server), preferencing the higher valued variant (and calling the other field something different, perhaps calling ours `vault_use_passcode`). IOW, I think it is possible, just not easy? My 2c!

cognifloyd:
PolicyMFA is an enterprise only feature and the test was protected by `constants.IsEnterprise`. Since `constants.IsEnterprise` was removed in 7b8e5f464598f04d0e4bdb826bd941e19a3362b7, I removed the PolicyMFA tests for both duo and okta (removing the okta test also drops the `constants.IsEnterprise` usage which should allow CI to complete.

Removing the PolicyMFA test makes me happy! It was a thorn in my side because I could never actually run the test. PolicyMFA is actually the sticking point that prevented my PR from being merged upstream--no devs were assigned to complete the PolicyMFA counterpart to the LoginMFA bits I wrote.

cipherboy:
Out of curiosity, what _is_ the difference between CE and EE MFA? I never really understood that, perhaps because I didn't take the time to look. :-) If you want to, happy to accept PRs if they are of general interest.

cognifloyd:
> Out of curiosity, what _is_ the difference between CE and EE MFA? I never really understood that, perhaps because I didn't take the time to look. :-) If you want to, happy to accept PRs if they are of general interest.

CE: only has Login MFA.
EE: also has Policy MFA (which was originally known simply as "MFA" before "Login MFA" was released in 1.10).

Login MFA is MFA that kicks in at the end of authentication before issuing the token.
Policy MFA extends standard vault policies with the ability to specify MFA requirements for certain paths, effectively allowing you to require MFA each time someone accesses a particular secret (or set of secrets), or even any sensitive sys operations. So, Login MFA (in CE and EE) only happens before a token is issued while Policy MFA (only EE) kicks in to require extra MFA from an actor that already has a valid token.

helper/identity/mfa/types.proto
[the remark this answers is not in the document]
cognifloyd: Go figure: there _are_ ways to run an android emulator in GHA.
https://www.atkinsondev.com/post/android-emulator-tests-github-actions/
https://github.com/ReactiveCircus/android-emulator-runner

So, then we'd have to figure out how to install and automate the okta app.

vault/external_tests/identity/login_mfa_okta_test.go
cognifloyd: Go figure: there _are_ ways to run an android emulator in GHA.
https://www.atkinsondev.com/post/android-emulator-tests-github-actions/
https://github.com/ReactiveCircus/android-emulator-runner

So, then we'd have to figure out how to install and automate the okta app.

cipherboy:
\o hey @cognifloyd -- do you know what's left? It looks like you wanted to test with OpenBao, but just checking you're not waiting on more reviews from me right now or something? :-)

cognifloyd:
Not waiting. I haven't had time to test yet. I looked into the MFA plugins to see about what the future might hold, and how the protobuf changes would impact it.

cipherboy:
Hey @cognifloyd! I'm going ahead and closing that if its alright :-) We've made progress creating config-drive, declarative plugin registration and externalizing KMS seal mechanisms using it, so I think Login MFA is another logical step. If you're interested in that, do let us know, we'll happily take a short RFC and an update of this, but given that it is the direction we're going, I think I'm going to opt to close this approach for the time being.

Thank you!

PR #520, opened 2024-09-11
Improve download page

cipherboy:
This adds both GPG and Cosign signatures to our download page, improves visibility of our container registries, and mentions where to find our GPG key and SBOMs.

----

<details>

<summary> Full-page screenshot </summary>

Screenshot 2024-09-11 at 17-22-19 Downloads OpenBao

</details>

Docker switcher:

image

New download card:

image

Both the "Download" and label up top ("AMD64 - Debian Package") take you to download the package.

### DIFF
--- a/website/src/pages/downloads.tsx
+++ b/website/src/pages/downloads.tsx
@@ -9,6 +9,7 @@ import {
     GetReleases,
     AssetArchitecture,
     OsPrettyPrint,
+    ArchPackageMapApply,
 } from "@site/src/components/Releases";
 
 // Create a context
@@ -24,7 +25,7 @@ const OptionsProvider = ({ children }) => {
         const fetchOptions = async () => {
             try {
                 // Check if options are cached in localStorage and not expired
-                const cachedOptions = localStorage.getItem("options");
+                const cachedOptions = localStorage.getItem("gh-releases");
                 if (cachedOptions) {
                     const { data, timestamp } = JSON.parse(cachedOptions);
                     if (new Date().getTime() - timestamp < 600000) {
@@ -41,14 +42,14 @@ const OptionsProvider = ({ children }) => {
                     "https://api.github.com/repos/openbao/openbao/releases",
                 );
                 if (!response.ok) {
-                    throw new Error("Failed to fetch options");
+                    throw new Error("Failed to fetch gh-releases");
                 }
                 const ghReleases = await response.json();
                 const releases = GetReleases(ghReleases);
                 const versions = Object.keys(releases);
                 setOptions(releases);
                 localStorage.setItem(
-                    "options",
+                    "gh-releases",
                     JSON.stringify({
                         data: releases,
                         timestamp: new Date().getTime(),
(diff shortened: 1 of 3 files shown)


website/src/pages/downloads.tsx
```diff
@@ -103,18 +104,76 @@ const VersionSelect = () => {
...
+        return;
+    }
+
     const { selectedItem } = useOptions();
     return (
         <div className="pagination-nav__item">
-            <a className="pagination-nav__link" href={url}>
-                <div className="pagination-nav__label">
-                    {AssetArchitecture(url).toUpperCase()}
+            <div className="card">
```
JanMa, line 146: Can you please give the cards a stronger boarder? They are very difficult to keep apart, especially in dark mode.
**New**:
image

**Old**:
image

website/src/pages/downloads.tsx
```diff
@@ -103,18 +104,76 @@ const VersionSelect = () => {
...
+                            coSig &&
+                            <a class="button button--secondary" href={coSig}>Cosign Signature</a>
+                        }
+                        {
+                            coCert &&
+                            <a class="button button--secondary" href={coCert}>Cosign Certificate</a>
+                        }
+                    </div>
+                </div>
+            </div>
```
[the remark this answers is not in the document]
JanMa: It might also be that the whole card element is not sized properly

JanMa (changes requested):
Some smaller CSS issues but I like the general approach :+1:

website/src/pages/downloads.tsx
```diff
@@ -103,18 +104,76 @@ const VersionSelect = () => {
...
+    let coSig: string;
+    let name: string = "Binary";
+
+    for (let url of urls) {
+        if (url.toLowerCase().includes(".deb")) {
+            name = "Debian Package";
+        } else if (url.toLowerCase().includes(".rpm")) {
+            name = "RPM Package";
+        } else if (url.toLowerCase().includes(".pkg") && url.toLowerCase().includes("linux")) {
+            name = "Alpine Package";
```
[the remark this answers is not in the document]
cipherboy: My bad!

website/src/pages/downloads.tsx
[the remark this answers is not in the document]
cipherboy: Hmmm, so it looks better if I combine the `navitem` and `card` thingy, but not really sure how to turn this into a more flexible flexbox :-D It already should be a flexbox, but I don't understand why it truncates it versus reflowing it...

website/src/pages/downloads.tsx
cipherboy: Interesting, yeah, I see the borders aren't good. I suspect it is because it is a nested card (the outer border is also a card) so perhaps it truncated one of the borders as a result, I'm not sure. I'll play with it!

cipherboy: image

I think I fixed this by changing the background color and fixing the "light" shadow color.

website/src/pages/downloads.tsx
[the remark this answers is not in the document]
cipherboy: @JanMa I _think_ I've fixed this, but give it a try!

website/src/pages/downloads.tsx
```diff
@@ -103,18 +104,76 @@ const VersionSelect = () => {
...
                 </div>
-                <div className="pagination-nav__sublabel">
-                    Version: {selectedItem}
+                <div className="card__body">
+                    <p className="text--center">
+                        Version: {selectedItem}
+                    </p>
                 </div>
-            </a>
+                <div className="card__footer">
```
JanMa, line 159: Now the buttons are cut off in all cards for me:
image

JanMa (changes requested):
Sorry, still one CSS issue left

website/src/pages/downloads.tsx
cipherboy: @JanMa this should be fixed, let me know!

JanMa: Your latest changes seem to fix this :+1:

cipherboy: Ok, I'll try again lol. I'm not sure how much we can do, given the buttons are from the framework and we probably just need bigger changes at some point...

JanMa: I've played around with it a bit and what works very well for smaller screen sizes is to just have one row of cards instead of two. Not sure yet what CSS magic one needs to use so that we'll still have two rows for bigger screens 😄

cipherboy: @JanMa Can you stick it in a media query for like 720px screens and below?

JanMa: Yes, much better :-) One last small improvement suggestion:
image
I think it looks much nicer if we force all corners of the buttons to be round

website/src/pages/downloads.tsx
```diff
@@ -103,18 +104,75 @@ const VersionSelect = () => {
...
+                        }
+                        {
+                            coSig &&
+                            <a class="button button--secondary" href={coSig}><span>Cosign Signature</span></a>
+                        }
+                        {
+                            coCert &&
+                            <a class="button button--secondary" href={coCert}><span>Cosign Certificate</span></a>
+                        }
+                    </div>
```
JanMa, line 173: This changes the buttons to have round corners everywhere, which looks much nicer if the cards are resized
```suggestion
                    <div className="button-group button-group--block">
                        <a
                            className="button button--primary"
                            href={asset}
                            style={{
                                borderRadius: "var(--ifm-button-border-radius)",
                            }}
                        >
                            <span>Download</span>
                        </a>
                        {gpgSig && (
                            <a
                                className="button button--secondary"
                                href={gpgSig}
                                style={{
                                    borderRadius:
                                        "var(--ifm-button-border-radius)",
                                }}
                            >
                                <span>GPG Signature</span>
                            </a>
                        )}
                        {coSig && (
                            <a
                                className="button button--secondary"
                                href={coSig}
                                style={{
                                    borderRadius:
                                        "var(--ifm-button-border-radius)",
                                }}
                            >
                                <span>Cosign Signature</span>
                            </a>
                        )}
                        {coCert && (
                            <a
                                className="button button--secondary"
                                href={coCert}
                                style={{
                                    borderRadius:
                                        "var(--ifm-button-border-radius)",
                                }}
                            >
                                <span>Cosign Certificate</span>
                            </a>
                        )}
                    </div>
```

JanMa (approved):
LGTM, one small suggestion

website/src/pages/downloads.tsx
cipherboy: Ahh, yes, I like.

website/src/pages/downloads.tsx
cipherboy: Cool, I added a property to the existing CSS rule for these to do so. Thanks @JanMa!!

PR #523, opened 2024-09-12
Fix kubernetes Service Registration Guide

ajaygk95:
I was trying out the Kubernetes service registration in openbao and noticed that, setting up,

`VAULT_K8S_NAMESPACE` and `VAULT_K8S_POD_NAME` env's as described in the documentation does not work. We get the below error,
`[WARN]  service_registration.kubernetes: unable to set initial state due to \"namespace\" is unset, will retry`

Fix - These env's have to be replaced with `BAO_K8S_NAMESPACE` and `BAO_K8S_POD_NAME`

### DIFF
--- a/website/content/docs/configuration/service-registration/kubernetes.mdx
+++ b/website/content/docs/configuration/service-registration/kubernetes.mdx
@@ -26,8 +26,8 @@ service_registration "kubernetes" {
 Alternatively, the namespace and pod name can be set through the following
 environment variables:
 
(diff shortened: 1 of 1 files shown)


cipherboy (approved):
It looks like there's a separate bug that causes the `VAULT_` variants not to work; I'll go ahead and open a PR to fix that.

cipherboy:
Thank you!

PR #524, opened 2024-09-12
Document the existing package manager for MacOS homebrew

norman-abramovitz:
Resolves #

### DIFF
--- a/website/content/docs/install.mdx
+++ b/website/content/docs/install.mdx
@@ -27,6 +27,8 @@ $ brew info openbao
 $ brew install openbao
 ```
 
+### Linux
+
 :::info
 
 
--- a/website/content/docs/install.mdx
+++ b/website/content/docs/install.mdx
@@ -20,7 +20,15 @@ There are several options to install OpenBao:
 
 ## Package manager
 
+### Homebrew - MacOS
+
+```shell-session
+$ brew info openbao
+$ brew install openbao
+```
+
 :::info
+
 
 OpenBao does not yet have a package repository. For now you need to download and
 install packages manually.


website/content/docs/install.mdx
```diff
@@ -20,8 +20,16 @@ There are several options to install OpenBao:
 
 ## Package manager
 
+### Homebrew - MacOS
+
+```shell-session
+$ brew info openbao
+$ brew install openbao
+```
+
```
cipherboy, line 29: Not to nitpick too terribly much, but do you mind adding a heading so it is clear the below is for Linux? Thanks!

```suggestion

### Linux

```

cipherboy:
Thanks!

cipherboy (approved):

norman-abramovitz:
I just saw the request to add a header for Linux, and I am glad I double-checked to make sure the PR was still open.   

The Apple Branding convention is some version of macOS. 
https://en.wikipedia.org/wiki/MacOS#:~:text=Apple%20shortened%20the%20name%20to,iOS%2C%20watchOS%2C%20and%20tvOS.

I am not sure calling  "macOS" "Linux" is appropriate.   I would do whatever the OpenBao maintainers want.  WIth the PR being merged, there is nothing to do.   No response is need if my assumption is correct.

cipherboy:
@norman-abramovitz Sorry, the header was below your addition of MacOS instructions -- the text under it was talking about Linux distribution packaging, so the MacOS header would be confusing. 

Let me know if you think something still needs to be changed:

image

norman-abramovitz:
I am good with it.  Another reasonable title I can think of is  "Other Package Managers" to be inclusive for Windows.  ROTFL.

PR #525, opened 2024-09-12
Remove Zededa from TSC

cipherboy:
Per clarification, Michael Maxey intends to resign from the TSC on behalf of Zededa without replacement by Zededa.

See also: https://lists.lfedge.org/g/OpenBao-TSC/topic/zededa_and_open_bao/108134993

---

The change in PCC has also been made.


JanMa (changes requested):
Can you please also remove them here:
https://github.com/openbao/openbao/blob/main/website/src/components/Supporters/index.tsx#L29-L32

cipherboy:
Updated, thanks @JanMa!

JanMa (approved):

PR #529, opened 2024-09-14
Add repositories to website

cipherboy:
@JanMa This should work in theory, but might require a few attempts to figure out: I just added GPG key signing based on the release workflow but I haven't tested it yet. 

---

This adds two repositories to the website: RPM and Debian. These are built from our latest release (while `createrepo_c` supports multi-release repos to allow downgrading, `reprepro` doesn't). 

TODO:

- [ ] Find and rename architectures on RPM to align with the distros so that `$basearch` works. 
- [ ] Validate Debian signing actually works.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -39,6 +39,22 @@ jobs:
         working-directory: ./website
         run: |
           make build
+      - id: cache-repos
+        name: cache-repos
+        uses: actions/cache@ab5e6d0c87105b4c9c2047343972218f562e4319 # v4.0.1
+        with:
+          path: website/build/repos
+          key: docs-cache-repo-build
+      - id: gpg-import
+        name: GPG Import
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-deploy.yml
```diff
@@ -39,6 +39,22 @@ jobs:
         working-directory: ./website
         run: |
           make build
+      - id: cache-repos
+        name: cache-repos
+        uses: actions/cache@ab5e6d0c87105b4c9c2047343972218f562e4319 # v4.0.1
+        with:
+          path: website/build/repos
+          key: docs-cache-repo-build
+      - id: gpg-import
```
cipherboy, line 48: @JanMa I'd comment out this section of the workflow and also in `genrepos.sh`, drop the line:

```
SignWith: E617DCD4065C2AFC0B2CF7A7BA8BC08C0F691F94
```

from the `conf/distributions` file and this should let you run it in your own fork (if you don't want to use `cipherboy.github.io/openbao`) without the GPG key.

Or, you could set up some secrets in your own repo and modify the signing subkey identifier to your own.

cipherboy:
Closing in favor of #1386, thanks @Hoeze!

PR #548, opened 2024-09-23
add vault symlink to Docker images

JanMa:
This adds a symlink to our Docker images which links `/bin/vault` to
`/bin/bao`. It will fix the issue of a not working Vault/OpenBao agent
injector in our HELM chart.


DanGhita (approved):
LGTM

cipherboy (approved):
I don't think this is quite the approach we want for RPM packaging (since it would cause `bao` to conflict with `vault`), but I agree this works well for containers.

Thanks, @JanMa!

PR #550, opened 2024-09-24
Add blog post on profiles

cipherboy:
This blog suggests a new profile-based request system, simplifying application developer's experience by giving them a single endpoint to fetch all secrets configured by their administrator.

---

I'm thinking this might be a good Friday or next Monday topic!

### DIFF
--- a/website/content/blog/2024-09-27-profiles.md
+++ b/website/content/blog/2024-09-27-profiles.md
@@ -1,0 +1,89 @@
+---
+title: Profiles for Cross-Plugin Communication
+description: A server-side request framework might alleviate some need for a new ACL system.
+slug: profiles
+authors: cipherboy
+tags: [technical, plugins, core]
+image: https://openbao.org/img/how-openbao-works.png
+---
+
+OpenBao and upstream [lack server-side cross-pluign communication](https://github.com/hashicorp/vault/issues/21023#issuecomment-1579232437).
+
+As recently seen with an [OIDC feature](https://github.com/openbao/openbao/pull/320), this shortcoming often needs to be worked around on the client side, potentially exposing sensitive information.
+
+There's usually two paths discussed for cross-plugin communication:
+
+1. Requests bound under the context of the original user token.
+2. Designing some other authorization system or an internal API based design.
+
+<!-- truncate -->
+
+Both have some trade-offs: the security model of the first makes more intuitive sense to reason about the complexity of its implementation. There's no additional design necessary for restricting access to other plugins. However, it has some undesirable side-effects: it prevents a trusted (server-bound) channel between two plugins. As we later decided to revisit the OIDC changes, we opted to allow auth plugins to write to an [internal data field](https://github.com/openbao/openbao/issues/519) that becomes accessible on the entity. This prevented the need for the client (an partially trusted party) from needing to hold the sensitive OIDC tokens but allowed any plugin to be able to access its value. However, plugins would still need to be updated to read this field; thus it has limited utility.
+
+While the need for a trusted channel still exists, I'd like to talk about a more flexible idea for improving the developer experience of OpenBao, that could potentially address ACL-bound requests. A **profile system**.
+
+One of the major learnings of [ACME](https://datatracker.ietf.org/doc/html/rfc8555) is that simplicity of automation and integration greatly improve the developer experience of a protocol. OpenBao, from a developer's perspective, is rather hard to understand: there's all sorts of auth and secrets engines that one can use and it is down to an organization's policies and technical decisions which ones they'll use. Unless they're interacting with a KMS, however, many applications' usage of secrets is transactional: single requests to many engines to collect all the secrets they need to operate.
+
+Let's introduce an **application profile**: a way for operators to define a single endpoint which collects a series of requests into a single place on behalf of a caller. Perhaps this includes four major components: `input` from the request, an optional `auth` step (to transparently handle authentication on behalf of the caller), one or more `request` blocks to collect the necessary secrets, and a `output` block to format the response.
+
+This looks a lot like the ACL policy system with some extra directives:
+
+```hcl
+input {
+  authenticated = true
+  data = {
+    "common_name": "string"
+  }
+}
+
+request "cert" {
+  path = "/v1/pki/issue/example"
+  operation = "read"
+  err = fatal
+  data = {
+    "common_name": input.data.common_name
+  }
+}
+
+request "dbpass" {
+  path = "/v1/database/creds/example"
+  operation = "read"
+  err = fatal
+}
+
+output {
+  data = {
+    "listener": {
+      "certificate": request.cert.data.certificate,
+      "private_key": request.cert.data.private_key,
+      "issuer": request.cert.data.issuing_ca
+    },
+    "database": {
+      "username": request.dbpass.data.username,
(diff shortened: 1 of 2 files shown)


cipherboy:
>  What we want to achieve is not very different from the OIDC authorization delegation: plugin#1 authenticates with OpenBao saying it will use resources from plugin#2, ... , plugin#N, and after successful authentication then plugin#1 can consume resources from the other plugins.

I think with `deny-direct`, you can also _effectively_ do the second form ("internal API"), while leaving it open to admins to allow that direct access if they want. However, you could also see (once the profile engine is built) allowing an internal API set (perhaps under `internal/v1`) that plugins can expose as an entirely separately ACLed format of secrets.

Not sure. :-)

JanMa:
Interesting idea, feels a bit like you're reinventing the[ OpenBao Agent](https://openbao.org/docs/agent-and-proxy/agent/) just on the server side though :smile: But regardless of that, I'm not sure I like writing a blog post about a potential feature which hasn't even been discussed in an RFC. What happens if we end up not doing it, or end up doing it very differently from the proposed solution in the post? Will we delete the post, update it or just leave it as is?

cipherboy:
\o hey @JanMa! 

> Interesting idea, feels a bit like you're reinventing the[ OpenBao Agent](https://openbao.org/docs/agent-and-proxy/agent/) just on the server side though 😄

Yes, that'd be a way of describing it. :-) 

The difference is that this would execute within/next to a trusted context and would reduce the latency of request chaining, so the client incurs a single round-trip time. It might (with the `deny-direct`, that OpenBao Agent can't enforce by virtue of not being in the server context) also allow us to do server-side authorization as well. Or get us closer to that.

Just a thought. :-)

---

> What happens if we end up not doing it, or end up doing it very differently from the proposed solution in the post? Will we delete the post, update it or just leave it as is?

I think this post serves to do something similar to posts in the Rust community: start inspiring people to think about an area we'd like to consider. Maybe they'll write their own blog post, too, with a different design. Or point out why ours has flaws.

In our case, I think our goal is that we'd like to encourage thinking about a simpler Developer UX for secrets management and what that might look like. If it is good (and not too specific to OpenBao), we can potentially think about trying to see if others are interested in standardizing this across secrets managers. Maybe working with another SM partner first, and then maybe thinking about taking that to a standard body.

(With a different path structure, I think we could make this rather like ACME. We could keep the current path structure or have a config flag to expose this as that.)

This also signals an amount of being receptive to open dialogue on topics. I think this topic in particular is sufficiently complex that we won't be wanting a single implementation to just show up, and we may not even want a single RFC just to show up -- we probably want careful back and forth on different goals and ideas and designs before starting an RFC. Blog posts are (or can be) a way of doing that. We don't delete our GitHub issues and mailing list threads when we decide designs aren't good (or at least, I'm not planning to!). :-D 

So I think we'd probably just update it to say, hey, this was implemented differently as XYZ, check it out!

DanGhita:
Hi Alex and JanMa,

 

In some way, this topic remembers me the other ones related to the OIDC flows improvement and the inter-plugins communication.

 

I have the feeling that OpenBao (inherited from upstream) exposes extremely flexible public APIs. But as always, this flexibility comes with a drawback: it’s up to the client to understand and implement the logic of the work-flows.

 

Maybe, we should try to have the best of the two worlds: for very “standard” flows, having some “macro” / “helper” APIs for making client’s life easier could be beneficial. Additionally, this could  come with  performance and some  security advantages (the server knows better the  right flow to be used and it is supposed to run in  a more secure context).

 

So, for coming back to the initial topic, I think it is an interesting subject and the solution you are proposing, Alex, could be the right one (but unfortunately, I don’t have a full knowledge of  the Bao ACL/policies/groups concepts…) .

 

BR,

Dan

 

 

 

De : Alexander Scheel ***@***.***> 
Envoyé : mercredi 25 septembre 2024 00:06
À : openbao/openbao ***@***.***>
Cc : GHITA Dan ***@***.***>; Review requested ***@***.***>
Objet : Re: [openbao/openbao] Add blog post on profiles (PR #550)

 

\o hey @JanMa <https://eur02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FJanMa&data=05%7C02%7Cdan.ghita%40viaccess-orca.com%7C950b2f7bd20f4261b10208dcdce5222d%7C6a6c76397a264a3bb832507d1afc7b2c%7C0%7C0%7C638628123906747484%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=6sm5n%2BHDdpbiNBizjC4SReH5tofq3ta85FsAd0Tu9d4%3D&reserved=0> !

	Interesting idea, feels a bit like you're reinventing the OpenBao Agent <https://eur02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fopenbao.org%2Fdocs%2Fagent-and-proxy%2Fagent%2F&data=05%7C02%7Cdan.ghita%40viaccess-orca.com%7C950b2f7bd20f4261b10208dcdce5222d%7C6a6c76397a264a3bb832507d1afc7b2c%7C0%7C0%7C638628123906781579%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=2OXpefnH%2B2eNS6HmUJYscmvcPBESjvahMJOkS1B83aQ%3D&reserved=0>  just on the server side though 😄

Yes, that'd be a way of describing it. :-)

The difference is that this would execute within/next to a trusted context and would reduce the latency of request chaining, so the client incurs a single round-trip time. It might (with the deny-direct, that OpenBao Agent can't enforce by virtue of not being in the server context) also allow us to do server-side authorization as well. Or get us closer to that.

Just a thought. :-)

________________________________

	What happens if we end up not doing it, or end up doing it very differently from the proposed solution in the post? Will we delete the post, update it or just leave it as is?

I think this post serves to do something similar to posts in the Rust community: start inspiring people to think about an area we'd like to consider. Maybe they'll write their own blog post, too, with a different design. Or point out why ours has flaws.

In our case, I think our goal is that we'd like to encourage thinking about a simpler Developer UX for secrets management and what that might look like. If it is good (and not too specific to OpenBao), we can potentially think about trying to see if others are interested in standardizing this across secrets managers. Maybe working with another SM partner first, and then maybe thinking about taking that to a standard body.

(With a different path structure, I think we could make this rather like ACME. We could keep the current path structure or have a config flag to expose this as that.)

This also signals an amount of being receptive to open dialogue on topics. I think this topic in particular is sufficiently complex that we won't be wanting a single implementation to just show up, and we may not even want a single RFC just to show up -- we probably want careful back and forth on different goals and ideas and designs before starting an RFC. Blog posts are (or can be) a way of doing that. We don't delete our GitHub issues and mailing list threads when we decide designs aren't good (or at least, I'm not planning to!). :-D

So I think we'd probably just update it to say, hey, this was implemented differently as XYZ, check it out!

cipherboy:
With out-of-band ascent from @JanMa, going ahead and merging.

Thanks all!

PR #558, opened 2024-09-28
Fix transactions and PKI interactions

cipherboy:
This should fix `main` and was caused by the incremental, not-up-to-date merging of the transactional PR series (#292). 

---

There were two separate bugs:

1. Cache invalidation was not occurring correctly. There were two separate bugs here, one which I observed (not invalidating the cache) and the other (race condition on `cacheTransaction.modified`) was only theoretically possible. I'm not sure if concurrent access to the same transaction is something we should deal with, given some backends like PostgreSQL may not support it.
2. Fixing the transaction commit ordering w.r.t. CRL rebuild in PKI. I had forgotten common entries were modified in both, so I've since fixed that. Also, that order may not have worked on PostgreSQL with transactions.

### DIFF
--- a/builtin/logical/pki/path_manage_issuers.go
+++ b/builtin/logical/pki/path_manage_issuers.go
@@ -422,25 +422,6 @@ func (b *backend) pathImportIssuers(ctx context.Context, req *logical.Request, d
 	}
 
 	if len(createdIssuers) > 0 {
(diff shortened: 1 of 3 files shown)


JanMa (approved):

cipherboy:
Thanks, @JanMa!

Issue #482, opened 2024-08-23
KV -  Canceled secret metadata deletes might corrupt state (Vault #16488)


cipherboy:
As mentioned by @johanneswuerbach on [#16488](https://github.com/hashicorp/vault/issues/16488)

> **Describe the bug**
> 
> Canceling a `DELETE secret/metadata/xyz` operation can cause the respective secret to become permanently broken and vault to only return:
> 
> ```
> Code: 500. Errors:
> 
> * 1 error occurred:
> 	* could not find version data
> ```
> 
> **To Reproduce**
> 
> The issue seems to be timing related, but running the script you can find here https://github.com/johanneswuerbach/vault-repro is reliably reproducing this issue for me.
> 
> **Expected behavior** A clear and concise description of what you expected to happen.
> 
> **Environment:**
> 
>     * Vault Server Version (retrieve with `vault status`): 1.11.1 (but also tested 1.10.4 and 1.2.2)
> 
>     * Vault CLI Version (retrieve with `vault version`): `github.com/hashicorp/vault/api v1.7.2`
> 
>     * Server Operating System/Architecture: Linux amd64 / arm64
> 
> 
> Vault server configuration file(s):
> 
> ```hcl
> disable_mlock = true
> ui = true
> 
> listener "tcp" {
>   tls_disable = 1
>   address = "[::]:8200"
>   cluster_address = "[::]:8201"
> }
> storage "file" {
>   path = "/vault/data"
> }
> ```
> 
> But also with the config in https://github.com/johanneswuerbach/vault-repro/blob/main/compose.yaml#L8
> 
> **Additional context**
> 
> We saw this issue the first time a couple of days after raising the `max_versions` from the default 10 to 100, but I'm not sure whether this is related or just coincidence.

---

As transactional storage is landing in #296 and #292, it would be great to use transactions inside KVv2 operations.

cipherboy:
Hello @johanneswuerbach -- many thanks for the reproducer! I can confirm this does indeed affect OpenBao as well.

I've opened a PR with changes (#560) in case you or @astromechza wish to try it out on a test cluster.

PR #560, opened 2024-09-28
Add transactions to K/V secrets engine

cipherboy:
This adds transaction support to the K/V secrets engine, enabling safe modifications to entries stored in the engine, even when operations are canceled partway through.

Notably, this requires careful for ordering: locks should be grabbed before beginning the transaction, otherwise we risk parallel operations causing issues when modifying the same entries, even when we previously would've succeeded.

Resolves: #482

### DIFF
--- a/builtin/logical/kv/path_config.go
+++ b/builtin/logical/kv/path_config.go
@@ -83,6 +83,18 @@ func (b *versionedKVBackend) pathConfigRead() framework.OperationFunc {
 // pathConfigWrite handles create and update commands to the config
 func (b *versionedKVBackend) pathConfigWrite() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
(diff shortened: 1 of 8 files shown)


cipherboy:
> Some Read functions implement now R/O transactions, while some others are unchanged (pathDataRead, for ex.): is there a reason for this?

Sorry, can you expand on this @DanGhita ? In my code on this PR, `pathDataRead` uses a R/O TXN. Do you have another path you're thinking of that uses a R/W TXN or no TXN?

DanGhita:
@cipherboy : You are right about the other Read functions, , the "unchanged" function is pathConfigRead.

cipherboy:
Ah, I didn't think reading the configuration was important to be tracked in the transaction. The issue is the lock acquisition occurs _after_ config read (and we need to open the write transaction after the lock), so it didn't make sense to me to put it in a separate read transaction given it was a single storage entry and thus already atomic.

PR #561, opened 2024-09-29
Deny empty valid_principals during SSH issuance

cipherboy:
HashiCorp Vault recently released a security vulnerability identifying global issuance without `valid_principals`. They [introduced a new role option](https://developer.hashicorp.com/vault/api-docs/secret/ssh#allow_empty_principals), `allow_empty_principals`, breaking existing users (as it defaults to `false`), but allowing operators to override it on a per-role level.

As evidenced by the extent of the test changes, this is definitely considered a breaking change.

See also: https://discuss.hashicorp.com/t/hcsec-2024-20-vault-ssh-secrets-engine-configuration-did-not-restrict-valid-principals-by-default/70251
See also: https://groups.google.com/g/opensshunixdev/c/RFKeIwNvtn8
See also: https://github.com/openssh/openssh-portable/commit/0a80ca190a39943029719facf7edb990def7ae62
See also: https://github.com/openssh/openssh-portable/blob/67a115e7a56dbdc3f5a58c64b29231151f3670f5/regress/cert-userkey.sh#L341-L343
See also: https://github.com/openssh/openssh-portable/blob/67a115e7a56dbdc3f5a58c64b29231151f3670f5/regress/cert-hostkey.sh#L247

### DIFF
--- a/builtin/logical/ssh/path_roles.go
+++ b/builtin/logical/ssh/path_roles.go
@@ -60,6 +60,7 @@ type sshRole struct {
 	AllowBareDomains           bool              `mapstructure:"allow_bare_domains" json:"allow_bare_domains"`
 	AllowSubdomains            bool              `mapstructure:"allow_subdomains" json:"allow_subdomains"`
 	AllowUserKeyIDs            bool              `mapstructure:"allow_user_key_ids" json:"allow_user_key_ids"`
+	AllowEmptyPrincipals       bool              `mapstructure:"allow_empty_principals" json:"allow_empty_principals"`
 	KeyIDFormat                string            `mapstructure:"key_id_format" json:"key_id_format"`
 	OldAllowedUserKeyLengths   map[string]int    `mapstructure:"allowed_user_key_lengths" json:"allowed_user_key_lengths,omitempty"`
 	AllowedUserKeyTypesLengths map[string][]int  `mapstructure:"allowed_user_key_types_lengths" json:"allowed_user_key_types_lengths"`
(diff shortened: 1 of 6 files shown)


builtin/logical/ssh/path_roles.go
```diff
@@ -60,6 +60,7 @@ type sshRole struct {
 	AllowBareDomains           bool              `mapstructure:"allow_bare_domains" json:"allow_bare_domains"`
 	AllowSubdomains            bool              `mapstructure:"allow_subdomains" json:"allow_subdomains"`
 	AllowUserKeyIDs            bool              `mapstructure:"allow_user_key_ids" json:"allow_user_key_ids"`
+	AllowEmptyPrincipals       bool              `mapstructure:"allow_empty_principals" json:"allow_empty_principals"`
```
cipherboy, line 63: I wonder if it makes sense to make this a `*bool` and default to some global value (on `ssh/config/certs` or something). I'm not sure how you'd unset the value in the Role though, tbh; you'd probably have to delete the role and re-create it, or accept a string (`true`, `false`, or `default`).

I'm kinda in favor of keeping parity with HashiCorp here, though, I think that's a lot of work for something few will need unless they've got bad workloads already.

cipherboy: I took a slightly different approach: given `allowed_users=*` would let you get a certificate for _any_ concrete user anyways (or similarly for `allowed_hosts=*`), it isn't an issue to bypass `allow_empty_principals=false` here. This should reduce the compatibility breakage.

JanMa (approved):
LGTM but please mark this as a potentially breaking change in the release notes

changelog/561.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:security
+secrets/ssh: Deny globally valid certificate issuance without valid_principals or allow_empty_principals override. HCSEC-2024-20 / CVE-2024-7594.
```
cipherboy, line 2:
```suggestion
secrets/ssh: Deny globally valid certificate issuance without valid_principals or allow_empty_principals override. HCSEC-2024-20 / CVE-2024-7594. (**potentially breaking**)
```

cipherboy:
Thanks @JanMa and thanks @DanGhita for your thoughts earlier.

Merging!

PR #574, opened 2024-10-01
Removed rootcerts module and refactored code accordingly

### LINKED ISSUE #484: api: drop dependency on go-rootcerts (Vault #26136)
@dolmen writes on [#26136](https://github.com/hashicorp/vault/issues/26136)

> The [`github.com/hashicorp/go-rootcerts`](https://pkg.go.dev/github.com/hashicorp/go-rootcerts) module doesn't seem necessary anymore with recent Go versions:
> 
>     * The package was initially created to workaround a _Darwin-only_ [bug](https://github.com/golang/go/issues/14514) that has been fixed in Go 4 years ago
> 
>     * I filled a question about the relevance of that module ([Is this module still needed? go-rootcerts#21](https://github.com/hashicorp/go-rootcerts/issues/21)) almost one year ago, and nobody replied. Obviously nobody is maintaining that module.
> 
>     * Dropping this dependency will also remove a dependency on [`github.com/mitchellh/go-homedir`](https://pkg.go.dev/github.com/mitchellh/go-homedir) (see also [Use os.UserHomeDir() (darwin) and drop dependency on go-homedir (all) go-rootcerts#22](https://github.com/hashicorp/go-rootcerts/pull/22) about another way to drop that dependency)
> 
> 
> ```
> $ cd api
> $ go mod graph | grep go-rootcerts          
> github.com/hashicorp/vault/api github.com/hashicorp/go-rootcerts@v1.0.2
> github.com/hashicorp/go-rootcerts@v1.0.2 github.com/mitchellh/go-homedir@v1.1.0
> github.com/hashicorp/vault/api@v1.0.5-0.20200519221902-385fac77e20f github.com/hashicorp/go-rootcerts@v1.0.1
> github.com/hashicorp/go-rootcerts@v1.0.1 github.com/mitchellh/go-homedir@v1.1.0
> $ go mod graph | grep go-homedir  
> github.com/hashicorp/go-rootcerts@v1.0.2 github.com/mitchellh/go-homedir@v1.1.0
> github.com/hashicorp/go-rootcerts@v1.0.1 github.com/mitchellh/go-homedir@v1.1.0 
> ```

---

> There are only [a few references](https://github.com/search?q=repo%3Ahashicorp%2Fvault%20rootcerts.Config&type=code) to [`rootcerts.Config`](https://pkg.go.dev/github.com/hashicorp/go-rootcerts#Config) in this repo:
> 
>     * api/client.go
> 
>     * api/ssh_agent.go
> 
>     * builtin/credential/cert/backend_test.go
> 
> 
> If I write the patch myself, is there a chance for it to be accepted?

---

I concur that these APIs can be handled using the standard library and would welcome such a PR here. :-)

fatima2003:
This PR removes the direct dependency on [github.com/hashicorp/go-rootcerts](https://github.com/hashicorp/go-rootcerts) and refactors the related code in `api/client.go`, `api/ssh_agent.go`, and` builtin/credential/cert/backend_test.go` to use Go’s standard library for certificate loading and handling. 

This module was originally required to work around a macOS-specific issue, which has since been resolved, rendering the module unnecessary. The removal also eliminates `github.com/mitchellh/go-homedir` as a transitive dependency.

I ran `go mod tidy` in the appropriate directories, and as a result, [github.com/hashicorp/go-rootcerts](https://github.com/hashicorp/go-rootcerts) is no longer a direct or indirect dependency in `sdk` and is removed as a direct dependency in `api`. However, it remains as an indirect dependency in a few other areas:

```
github.com/google/tink/go@v1.7.0
github.com/openbao/go-kms-wrapping/wrappers/transit/v2@v2.1.0
github.com/openbao/openbao-template@v1.0.0
github.com/openbao/openbao/api@v1.9.2
```

Upon further investigation, I found that compared to March 25, 2024 (referencing the [Vault Issue 26136](https://github.com/hashicorp/vault/issues/26136)), additional components now depend on [github.com/hashicorp/go-rootcerts](https://github.com/hashicorp/go-rootcerts). For instance, github.com/google/tink/go is a read-only repo and has a dependency on go-rootcerts. Therefore, removing this indirect dependency entirely may not be possible without addressing those upstream dependencies.

## Resolves

Resolves #484 
Resolves Vault #26136

## Target Release

1.14.7


builtin/credential/cert/backend_test.go
```diff
@@ -169,10 +168,10 @@ func connectionState(serverCAPath, serverCertPath, serverKeyPath, clientCertPath
 		return tls.ConnectionState{}, err
 	}
 	// Load the CA cert required by the client to authenticate the server.
-	rootConfig := &rootcerts.Config{
+	rootConfig := &api.CertConfig{
```
cipherboy, line 171: On second thought... I think we can keep `api.certConfig` and `api.loadCACerts` private if we handle this one case (which doesn't need priority, just pool building) separately. 

e.g., something like:

```go
    pemData, err := os.ReadFile(serverCAPath)
    require.NoError(t, err)

    pool := x509.NewCertPool()
    pool.AppendCertsFromPEM(pemData)
```

Then we're not in the same scenario as `go-rootcerts` in a few years and don't have to provide a consistent API to external callers (and nobody will depend on us for this). :-) 

Thoughts?

api/rootcerts.go
```diff
@@ -0,0 +1,126 @@
...
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// CertConfig determines where LoadCACerts will load certificates from.
+type CertConfig struct {
+	// CAFile is a path to a PEM-encoded certificate file or bundle. Takes
```
cipherboy, line 14: You know, I wonder if we can do something better (from a UX perspective) and just greedily take all three of them. :-) 

The pool itself does deduplication: https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/crypto/x509/cert_pool.go;l=199-202

So we could do something like:

```go
pool, err := x509.SystemCertPool()
...
if CAFile != "" {
  pool.AddCerts(...)
}

&c
```

and get more CAs, rather than fewer. I think this would _technically_ break anyone who expected only a single CA to be provisioned, but I don't know how often that really occurs versus just wanting all certs to be taken. My 2c!

api/rootcerts.go
```diff
@@ -0,0 +1,126 @@
+package api
```
cipherboy, line 1: Since this file is very similar to https://github.com/hashicorp/go-rootcerts/blob/master/rootcerts.go, which is under the MPL, I believe we need a copyright header like:

```
// Copyright (c) 2024 OpenBao a Series of LF Projects, LLC
// Copyright (c) HashiCorp, Inc.
// SPDX-License-Identifier: MPL-2.0
//
// See https://github.com/hashicorp/go-rootcerts's rootcerts.go as of bb0b55efd5a1de16cb10686096b9596cfbab1fad.
```

to reflect the dual-source nature of this new file.

cipherboy (commented):
Looks good @fatima2003! Just a minor comment, I think we can make our new implementation private (which will be good to prevent people from depending on it: they should depend on `go-rootcerts` instead). 

I don't worry too much about the indirect dependency from `go-tink` :-)

api/rootcerts.go
fatima2003: I think this is a good idea in the long run. I guess that would mean there is no more priority among sources since all of them are loaded if provided. For users who expect a single CA or depend on the priority stuff, do you think it would make sense to add an optional parameter for prioritizing a specific source? If it's left empty, all sources would be loaded, but if set, only that particular certificate source would be used.

fatima2003: @cipherboy I found a edge case not being accounted for in the previous rootcerts. If in `CAPath`, there was a non-certificate PEM, the tests would fail with an error: `client_test.go:400: error reading environment: Error loading CA Path: Couldn't parse PEM in: /Users/fatima/openbao/api/test-fixtures/keys/key.pem` I fixed this in commit 66a85bbb0c5ed0e49add3f0b9da746fa799087fa by adding a check for contents of the PEM file. 

The reason the test wasn't failing previously was because CAPath has the lowest priority and wasn't tested independently before:
https://github.com/openbao/openbao/blob/e6ed492e570d7783c02ae7556e3f8fdba9f96ac5/api/client.go#L297-L301

builtin/credential/cert/backend_test.go
fatima2003: Yes, I agree that this is a positive change. Keeping `api.CertConfig` and `api.LoadCACerts` private, while handling this particular case separately for internal use would help avoid the complexity of maintaining a consistent API for external users. 

Updated in fe7e7dec44749f3ce4778bd768ea90262179e480

api/rootcerts.go
```diff
@@ -0,0 +1,157 @@
...
+		rest := pemData
+		for {
+			block, rest = pem.Decode(rest)
+			if block == nil {
+				break
+			}
+
+			// Ensure the PEM block is a certificate
+			if block.Type != "CERTIFICATE" {
+				fmt.Printf("Ignoring non-certificate PEM block found in %s\n", path)
```
cipherboy, line 128: Hmmm.... How I dislike PEM some days :-) 

The use of `fmt.Printf` I'd remove: (semgrep fails on it -- I think you can run it with `make docker-semgrep-ci`) https://github.com/openbao/openbao/actions/runs/11141741694/job/30963245478?pr=574 -- philosophically, `api/` is a library which may be called in various contexts, so I wouldn't always expect this to be seen, but I agree it isn't an error; I'd just ignore it.

That said...

The `block.Type != "CERTIFICATE"` is also worth changing. There's [the `ParsePEMBundle`](https://github.com/openbao/openbao/blob/main/sdk/helper/certutil/helpers.go#L252-L309) function in `sdk/helper/certutil` that handles this, IMO, correctly with opportunistic parsing:

https://github.com/openbao/openbao/blob/e6ed492e570d7783c02ae7556e3f8fdba9f96ac5/sdk/helper/certutil/helpers.go#L285-L289

(In particular, there's a bunch of legacy header types you can sometimes run into... Like `X509 CERTIFICATE` and others).

api/rootcerts.go
cipherboy: Nice catch! :D 

I agree, that's a good thing to avoid as some systems often have non-cert files in their CA directory. E.g., `/etc/pki/ca-trust/extracted/pem/` on my Fedora box has a `README` in it, which is neither cert or key. 

---

Here is good because its local to where the code and context is :-)

api/rootcerts.go
```diff
@@ -0,0 +1,157 @@
...
+	t.RootCAs = pool
+	return nil
+}
+
+// LoadCACerts loads a CertPool from all available CA sources in CertConfig.
+func loadCACerts(c *certConfig) (*x509.CertPool, error) {
+	// Start with system CAs.
+	pool, err := loadSystemCAs()
+	if err != nil {
+		pool = x509.NewCertPool()
```
cipherboy, line 51: Sorry to faff here, but I realized an implementation detail when responding to https://github.com/openbao/openbao/issues/575#issuecomment-2388451822: 

[In `x509.Certificate.Verify(...)`](https://pkg.go.dev/crypto/x509#Certificate.Verify), it says:

> If `opts.Roots` is `nil`, the platform verifier might be used, and verification details might differ from what is described below. If system roots are unavailable the returned error will be of type `SystemRootsError`. 

So we probably want to verify if we've _added_ anything successfully to the pool, and if so, return `pool`. But if we didn't have anything to add to the pool, and we fell through to this line because the system pool failed... We'd now be returning a valid-but-empty pool on line 75 which means publicly trusted certs would fail to verify.

(If, for instance, there was some issue with System CAs from our call but the next call in Verify worked. idk.)

I think it'd be good to avoid that scenario and let `pool` be `nil` in that case... That does mean keeping track of an `added` bool or something and making all of the `if err := append....` calls into a `added, err := ...` format.

Does that sound right? Now I'm not sure if I like the return-the-system-pool-by-default behavior after all, given that it is implied... Hm! 

Thoughts?

api/rootcerts.go
fatima2003: Having the `Printf` statement in production code didn't feel right, and I appreciate the test that checks for it and the suggestion to remove it. I’ve also updated the certificate parsing to use `x509.ParseCertificates` as suggested. I'm not too familiar with all the helper functions yet, but I'll take some time to look through them and get more acquainted.

api/rootcerts.go
```diff
@@ -16,7 +16,7 @@ import (
 	"path/filepath"
 )
 
-// CertConfig determines where LoadCACerts will load certificates from.
+// certConfig determines wherelLoadCACerts will load certificates from.
```
cipherboy, line 19:
```suggestion
// certConfig determines where loadCACerts will load certificates from.
```

cipherboy:
Test failure looks unrelated:

```

2024-10-03T18:49:04.4239870Z === [31mFailed[0m
2024-10-03T18:49:04.4240858Z === [31mFAIL[0m: vault TestOIDC_PeriodicFunc/test-key-nil-next-signing-key (4.35s)
2024-10-03T18:49:04.4243295Z     identity_store_oidc_test.go:1165: For key: test-key-nil-next-signing-key at cycle: 2 expected namedKey's KeyRing to be at least of length 3 but was: 2
2024-10-03T18:49:04.4246055Z     identity_store_oidc_test.go:1176: For key: test-key-nil-next-signing-key at cycle: 2 expected public keys to be at least of length 3 but was: 2
2024-10-03T18:49:04.4248262Z 2024-10-03T18:46:30.477Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.core: shutdown called
2024-10-03T18:49:04.4250257Z 2024-10-03T18:46:30.477Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.core: marked as sealed
2024-10-03T18:49:04.4252179Z 2024-10-03T18:46:30.477Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.core: clearing forwarding clients
2024-10-03T18:49:04.4254482Z 2024-10-03T18:46:30.477Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.core: done clearing forwarding clients
2024-10-03T18:49:04.4256843Z 2024-10-03T18:46:30.477Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.core: pre-seal teardown starting
2024-10-03T18:49:04.4258762Z 2024-10-03T18:46:30.477Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.expiration: stop triggered
2024-10-03T18:49:04.4260881Z 2024-10-03T18:46:30.477Z [TRACE] TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.expiration.job-manager: terminating job manager...
2024-10-03T18:49:04.4263397Z 2024-10-03T18:46:30.477Z [TRACE] TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.expiration.job-manager: terminating dispatcher
2024-10-03T18:49:04.4265520Z 2024-10-03T18:46:30.477Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.expiration: finished stopping
2024-10-03T18:49:04.4267493Z 2024-10-03T18:46:30.477Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.rollback: stopping rollback manager
2024-10-03T18:49:04.4269259Z 2024-10-03T18:46:30.478Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.core: pre-seal teardown complete
2024-10-03T18:49:04.4270651Z 2024-10-03T18:46:30.478Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.core: clustering disabled, not stopping listeners
2024-10-03T18:49:04.4272268Z 2024-10-03T18:46:30.478Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.core: sealing barrier
2024-10-03T18:49:04.4274142Z 2024-10-03T18:46:30.478Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-next-signing-key.core: vault is sealed
2024-10-03T18:49:04.4275872Z 2024-10-03T18:46:30.482Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-signing-key.core: shutdown called
2024-10-03T18:49:04.4277537Z 2024-10-03T18:46:30.482Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-signing-key.core: marked as sealed
2024-10-03T18:49:04.4279296Z 2024-10-03T18:46:30.482Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-signing-key.core: clearing forwarding clients
2024-10-03T18:49:04.4281360Z 2024-10-03T18:46:30.482Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-signing-key.core: done clearing forwarding clients
2024-10-03T18:49:04.4283502Z 2024-10-03T18:46:30.482Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-signing-key.core: pre-seal teardown starting
2024-10-03T18:49:04.4285305Z 2024-10-03T18:46:30.482Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-signing-key.expiration: stop triggered
2024-10-03T18:49:04.4287333Z 2024-10-03T18:46:30.482Z [TRACE] TestOIDC_PeriodicFunc/test-key-nil-signing-key.expiration.job-manager: terminating job manager...
2024-10-03T18:49:04.4289505Z 2024-10-03T18:46:30.482Z [TRACE] TestOIDC_PeriodicFunc/test-key-nil-signing-key.expiration.job-manager: terminating dispatcher
2024-10-03T18:49:04.4291453Z 2024-10-03T18:46:30.482Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-signing-key.expiration: finished stopping
2024-10-03T18:49:04.4293814Z 2024-10-03T18:46:30.482Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-signing-key.rollback: stopping rollback manager
2024-10-03T18:49:04.4295684Z 2024-10-03T18:46:30.483Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-signing-key.core: pre-seal teardown complete
2024-10-03T18:49:04.4297686Z 2024-10-03T18:46:30.483Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-signing-key.core: clustering disabled, not stopping listeners
2024-10-03T18:49:04.4299538Z 2024-10-03T18:46:30.483Z [DEBUG] TestOIDC_PeriodicFunc/test-key-nil-signing-key.core: sealing barrier
2024-10-03T18:49:04.4301176Z 2024-10-03T18:46:30.483Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-signing-key.core: vault is sealed
2024-10-03T18:49:04.4302020Z 
2024-10-03T18:49:04.4302392Z === [31mFAIL[0m: vault TestOIDC_PeriodicFunc (0.00s)```

File: api/rootcerts.go (hunk 5 above)
Comment (cipherboy) [reply api/rootcerts.go:54]: I like that idea!

File: api/rootcerts.go [hunk 7]
```diff
@@ -26,6 +26,10 @@ type certConfig struct {
 
 	// CAPath is a path to a directory populated with PEM-encoded certificates.
 	CAPath string
+
+	// SystemCerts determines if system certificates should be included in the pool.
+	// Defaults to true
+	SystemCerts *bool
```
cipherboy, line 32:
```suggestion
	SkipSystemCerts bool
```

I think we get the same semantics if we invert and simplify the logic. This saves callers from having to deal with the allocation here which is sometimes yucky. 

My 2c, but it really is a nitpick at this point.

fatima2003: Aah I thought they would be a better way... I updated the logic :)

cipherboy (approved):
One minor thing to add before we merge: [a changelog entry](https://github.com/openbao/openbao/blob/main/changelog/README.md).

I think, because we now load all certs, an operator could have an invalid cert in some directory and end up with an error now, if they specified multiple envvars -- earlier the highest priority var would win.

So maybe something like:


    ```release-note:change
    api: Load all CA certificates specified in environment variables.
    ```

cipherboy:
Cool, looks like this is working as intended as far as I can tell. Thank you!

PR #603, opened 2024-10-09
Add CodeQL security analysis tooling

cipherboy:
This should help with [OpenSSF](https://www.bestpractices.dev/en/projects/9126) as it is a static analysis tool, e.g., towards:

>  At least one static code analysis tool (beyond compiler warnings and "safe" language modes) MUST be applied to any proposed major production release of the software before its release, if there is at least one FLOSS tool that implements this criterion in the selected language. [static_analysis]

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,51 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches: [ "main", "release/*" ]
+  pull_request:
+    branches: [ "main", "release/*" ]
+  schedule:
+    - cron: '34 23 * * 4'
+
+jobs:
+  analyze:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,51 @@
...
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches: [ "main", "release/*" ]
+  pull_request:
+    branches: [ "main", "release/*" ]
```
Gabrielopesantos, line 12: If I'm not mistaken, with these trigger configurations, a commit pushed to an open pull request will trigger two pipelines, one from `push` and the other from `pull_request`. Or is this to accommodate for PR's from forks as they can't push directly to the repository?

cipherboy: Won't both only be triggered if a `pull_request` originates from a `push`-allowed branch?

Gabrielopesantos: Yes, I believe so. Only pushes from the repo will trigger the `push` workflow, otherwise only `pull_request`.

JanMa (approved):

PR #608, opened 2024-10-10
Add transactional storage to PostgreSQL

cipherboy:
Title says it all :-)

This updates our test suite to test transactions in parallel go routines so that it can conform to PostgreSQL's semantics. Then we add transactions using PostgreSQL's native transactions. I've done some testing with K/V and things seem to behave nicely with the reproducer from #482. 

Resolves: #270

### DIFF
--- a/physical/postgresql/transaction.go
+++ b/physical/postgresql/transaction.go
@@ -1,0 +1,221 @@
+// Copyright (c) 2024 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package postgresql
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"sync"
+
+	"github.com/openbao/openbao/sdk/v2/physical"
+)
+
+type PostgreSQLBackendTransaction struct {
+	l  sync.Mutex
+	b  *PostgreSQLBackend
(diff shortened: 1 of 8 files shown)


cipherboy:
@genelet, @hixichen and  @alberk8 -- welcome your thoughts and testing!

schultz-is (approved):
Lgtm!

physical/postgresql/transaction.go
```diff
@@ -0,0 +1,221 @@
...
+	haveFinishedTx bool
+}
+
+func (b *PostgreSQLBackend) newTransaction(ctx context.Context, readOnly bool) (physical.Transaction, error) {
+	// Grab a transaction permit pool entry so that we can limit the number of
+	// concurrent transactions.
+	b.txnPermitPool.Acquire()
+
+	tx, err := b.client.BeginTx(ctx, &sql.TxOptions{
+		Isolation: sql.LevelRepeatableRead,
```
cipherboy, line 30: The only real surprise was needing to use `LevelRepeatableRead` isolation: I'd get errors while reading entries inside my read-only transactions, that were deleted by another write transaction if I used the default `LevelReadCommitted`.

DanGhita (approved):

Issue #612, opened 2024-10-10
storage/postgresql: Automatically add tables when selecting postgresql as secret storage (vault#28545)


cipherboy:
@FunkForever writes on https://github.com/hashicorp/vault/issues/28545: 

> **Is your feature request related to a problem? Please describe.**

According to current documentation (https://github.com/hashicorp/vault/blob/main/website/content/docs/configuration/storage/postgresql.mdx vault cannot create tables automatically. I use helm chart from bank vault to deploy vault (https://github.com/bank-vaults/vault-helm-chart). At the moment I am unfolding the tables manually. When deploying manually, the first installation always fails, because vault initially creates roles, users and a vault database in PostgreSQL, and then creates tables on top of that. Pre-installation in PostgreSQL is difficult because user credentials created before helm deployment will have different passwords and the application will not be able to get into the database (to create tables, the vault_owner_user user is used, which is automatically created when helm is deployed, and the password is put in secrets).
> 
> **Describe the solution you'd like** Automatically add tables (vault_kv_store, vault_ha_locks) and function (vault_kv_put) when selecting postgresql as secret storage.

---

This makes sense and should be fairly easy to add: a new parameter (perhaps `skip_table_creation=false`) could be added to the configuration. When not skipping, we could detect that [tables do not exist](https://github.com/openbao/openbao/blob/main/physical/postgresql/postgresql.go#L197) and then attempt to create them.

FunkForever:
Great idea! I fully support it.

cipherboy:
@FunkForever Cool! Are you interested in implementing it? Happy to review a PR if so. :-)

If you build OpenBao from `main` currently, you can also try out our new support for transactions with the Postgres backend; see #608.

FunkForever:
Thanks, but I can't help you with the implementation. I am a devops engineer and rarely do development in practice.

cipherboy:
\o hey @FunkForever -- this was in our v2.1.0 release that was released on Friday :-) Let us know if you end up using it or run into any problems!

PR #614, opened 2024-10-10
Allow PostgreSQL backend to create tables

cipherboy:
When running in certain environments such as Kubernetes via Helm chart, an operator may not necessarily have direct access to the database to create tables manually. Allow OpenBao to create the tables, which is enabled by default if it detects it does not have its tables present.

In the future, this sets us up to handle potential schema migrations, potentially allowing us to scale across multiple tables based on namespaces or other logical shards.

Resolves: #612

### DIFF
--- a/physical/postgresql/postgresql.go
+++ b/physical/postgresql/postgresql.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/armon/go-metrics"
 	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-secure-stdlib/parseutil"
 	"github.com/hashicorp/go-uuid"
 	_ "github.com/jackc/pgx/v4/stdlib"
 	"github.com/openbao/openbao/api/v2"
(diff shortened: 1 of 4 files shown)


physical/postgresql/postgresql.go
```diff
@@ -236,6 +237,57 @@ func NewPostgreSQLBackend(conf map[string]string, logger log.Logger) (physical.B
...
+		return nil, fmt.Errorf("failed to parse value for `skip_create_table`: %w", err)
+	}
+	if !skip_create_table {
+		// Try a query and see if there's a table not found error for both
+		// tables: data and ha locks.
+		var path string
+		row := db.QueryRow("SELECT path FROM " + quoted_table + " LIMIT 1")
+		if err := row.Scan(&path); err != nil && strings.Contains(err.Error(), `relation `+quoted_table+` does not exist`) {
+			// Successfully detected that there is no table; create it.
+			createTableQuery := "CREATE TABLE " + quoted_table + " (" +
```
cipherboy, line 256: Thoughts on if this should just be a `CREATE TABLE IF NOT EXISTS` instead?

cipherboy: Technically I think if we wrap these in a single transaction, we could unconditionally issue them and better handle multiple nodes starting at the same time with the same database.

physical/postgresql/postgresql_test.go
```diff
@@ -22,8 +23,9 @@ func TestPostgreSQLBackend(t *testing.T) {
 	// Use docker as pg backend if no url is provided via environment variables
 	connURL := os.Getenv("PGURL")
 	if connURL == "" {
-		_, u := postgresql.PrepareTestContainer(t, "11.1")
+		cleanup, u := postgresql.PrepareTestContainer(t, "11.1")
```
cipherboy, line 26: We were leaking Postgres containers before.

physical/postgresql/postgresql.go
```diff
@@ -236,6 +237,56 @@ func NewPostgreSQLBackend(conf map[string]string, logger log.Logger) (physical.B
...
+		defer txn.Rollback()
+
+		createTableQuery := "CREATE TABLE IF NOT EXISTS " + quoted_table + " (" +
+			`parent_path TEXT COLLATE "C" NOT NULL,` +
+			`  path        TEXT COLLATE "C",` +
+			`  key         TEXT COLLATE "C",` +
+			`  value       BYTEA,` +
+			`  CONSTRAINT pkey PRIMARY KEY (path, key)` +
+			`);`
+		if _, err := db.Exec(createTableQuery); err != nil {
```
cipherboy, line 266: The one issue with this is that, if we don't have table creation permissions, the query will fail even if the tables already exist. I think this could be a problem in some deployments, but PostgreSQL was not part of 2.0.0's GA compatibility guarantee so I think we can assume fresh deployments from here on out. They'll need to manually set `skip_create_tables=false` in their configuration.

website/content/docs/configuration/storage/postgresql.mdx
```diff
@@ -40,8 +40,68 @@ to disable SSL.
...
+
+- `connection_url` `(string: <required>)` – Specifies the connection string to
+  use to authenticate and connect to PostgreSQL. The connection URL can also be
+  set using the `BAO_PG_CONNECTION_URL` environment variable. A full list of supported
+  parameters can be found in the [pgx library][pgxlib] and [PostgreSQL connection string][pg_conn_docs]
+  documentation. For example connection string URLs, see the examples section below.
+
+- `table` `(string: "openbao_kv_store")` – Specifies the name of the table in
+  which to write OpenBao data. This table must already exist (OpenBao will not
+  attempt to create it).
```
DanGhita, line 53: I'm a little but confused: some tables of the DB will be automatically created by OpenBao but not all of them ?
In this case, we should probably mention the tables that are automatically created ?
Furthermore, if the user enables the KV engine, maybe we should automatically add the table for this ?

physical/postgresql/postgresql.go
DanGhita: Is there a secure way for providing OpenBao with  PostgreSQL credentials (including for allowing table creation) without exposing those credentials in the configuration files ?
Generally spiking, this seems to me another "chicken and egg problem" that we will not solve here, but we could maybe consider in the future the possibility to have in OpenBao a kind of initial secrets (bootstrap secrets) that would be available just after the unseal but before setting up the storage (maybe provided by a KMS, or by doing a ECDH ...) .
In any case, it seems to be a non-trivial problem.

cipherboy: The issue with the PostgreSQL engine is that it stores all data in a single table (`table`, usually set to `openbao_kv_data`). If we want to change that (to improve scalability), we need clients to grant table creation to OpenBao's credentials, e.g., to create separate tables for each namespace or something, that are used online and not one-off at the first startup.

The user can also always start OpenBao, have it create tables, and then restart it with more limited credentials as well.

website/content/docs/configuration/storage/postgresql.mdx
cipherboy: @DanGhita Pushed :-)

DanGhita (approved):

Issue #607, opened 2024-10-09
Broader usage of Transactions


cipherboy:
**Is your feature request related to a problem? Please describe.**

In #296, #292, #498, and #560 , initial support for transactions were added to the underlying storage backend and to the PKI and K/V secrets engines. However, there's many more places this can be used. 

**Describe the solution you'd like**

Secrets: 

- [ ] More PKI operations use transactions
- [ ] Transit is evaluated for transaction usage
- [ ] SSH is evaluated for transaction usage
- [ ] Database is evaluated for transaction usage
- [ ] RabbitMQ is evaluated for transaction usage
- [ ] totp is evaluated for transaction usage
- [ ] OpenLDAP is evaluated for transaction usage
- [ ] KVv1 is evaluated for transaction usage

Auth:

- [ ] AppRole is evaluated for transaction usage
- [ ] Cert is evaluated for transaction usage
- [ ] JWT is evaluated for transaction usage
- [ ] Kerberos is evaluated for transaction usage
- [ ] Kubernetes is evaluated for transaction usage
- [ ] LDAP is evaluated for transaction usage
- [ ] Radius is evaluated for transaction usage
- [ ] Token is evaluated for transaction usage
- [ ] Userpass is evaluated for transaction usage

Core:

- [ ] Token store
- [ ] Policy store
- [ ] ... more?

**Describe alternatives you've considered**

Some places such as identity, mfa, and others may rely on MemDB and thus don't immediately need transactions. 

**Explain any additional use-cases**

n/a

**Additional context**

v0nNemizez:
I can help with some of these :)

cipherboy:
@v0nNemizez Cool! Let me know if you have any questions. I'd suggest tackling each in separate issues so we get proper changelog entries for them all. :-)

suprjinx:
It seems like there'd be a way to open/commit/rollback Transactions in a global way -- eg, via middleware applied to all endpoints.

suprjinx:
> 
> All this to say, I think these transactions are a little different than most are used to, so we might want to handle them individually or have an escape hatch.

Could we check for the existence of a transaction (perhaps in Context) and use it automatically unless we decided to explicitly forego? So, eg, PKI CRL could open a new transaction-less connection for reading entries, but will otherwise read/write with the active transaction where available?

cipherboy:
[...]

PR #617, opened 2024-10-11
Adding transactions to AppRole authentication

v0nNemizez:
In  light of issue #607,  Im looking to add this to the AppRole auth engine.

Resolves parts of #607

## Target Release

1.14.7

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,7 +67,8 @@ that lacks a sign-off to this agreement will not be accepted by the OpenBao proj
 | James Butcher     | james@iotechsys.com                 |                                            | Brad Corrion (brad@iotechsys.com)       | IOTech Systems            | Member             |
 | Alain Nochimowski | alain.Nochimowski@viaccess-orca.com |                                            | Dan Ghita (dan.ghita@viaccess-orca.com) | Viaccess-Orca             | Member             |
 | Julian Cassignol  | jcassignol@wallix.com               |                                            | Cristian Popi (cpopi@wallix.com)        | Wallix                    | Member             |
-| Alexander Scheel  | ascheel@gitlab.com                  | [@cipherboy](https://github.com/cipherboy) | Jocelyn Eillis (jeillis@gitlab.com)     | GitLab                    | Chair              |
+| Alexander Scheel  | ascheel@gitlab.com                  | [@cipherboy](https://github.com/cipherboy) |                                         | GitLab                    | Chair (non-voting) |
+
 
 ### OpenBao Working Groups
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,8 +67,7 @@ that lacks a sign-off to this agreement will not be accepted by the OpenBao proj
 | James Butcher     | james@iotechsys.com                 |                                            | Brad Corrion (brad@iotechsys.com)       | IOTech Systems            | Member             |
 | Alain Nochimowski | alain.Nochimowski@viaccess-orca.com |                                            | Dan Ghita (dan.ghita@viaccess-orca.com) | Viaccess-Orca             | Member             |
 | Julian Cassignol  | jcassignol@wallix.com               |                                            | Cristian Popi (cpopi@wallix.com)        | Wallix                    | Member             |
-| Alexander Scheel  | ascheel@gitlab.com                  | [@cipherboy](https://github.com/cipherboy) |                                         | GitLab                    | Chair (non-voting) |
-
+| Alexander Scheel  | ascheel@gitlab.com                  | [@cipherboy](https://github.com/cipherboy) | Jocelyn Eillis (jeillis@gitlab.com)     | GitLab                    | Chair              |
 
 ### OpenBao Working Groups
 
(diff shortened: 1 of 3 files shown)


v0nNemizez:
@cipherboy open to suggestions :)

CONTRIBUTING.md
```diff
@@ -67,8 +67,7 @@ that lacks a sign-off to this agreement will not be accepted by the OpenBao proj
 | James Butcher     | james@iotechsys.com                 |                                            | Brad Corrion (brad@iotechsys.com)       | IOTech Systems            | Member             |
 | Alain Nochimowski | alain.Nochimowski@viaccess-orca.com |                                            | Dan Ghita (dan.ghita@viaccess-orca.com) | Viaccess-Orca             | Member             |
 | Julian Cassignol  | jcassignol@wallix.com               |                                            | Cristian Popi (cpopi@wallix.com)        | Wallix                    | Member             |
-| Alexander Scheel  | ascheel@gitlab.com                  | [@cipherboy](https://github.com/cipherboy) |                                         | GitLab                    | Chair (non-voting) |
-
+| Alexander Scheel  | ascheel@gitlab.com                  | [@cipherboy](https://github.com/cipherboy) | Jocelyn Eillis (jeillis@gitlab.com)     | GitLab                    | Chair              |
```
cipherboy, line 70: Hmm, I think you'll need to fix your rebase here. We have a [maintainer commit signature requirement](https://openbao.org/docs/rfcs/signed-commits/) and [workflow](https://github.com/openbao/openbao/blob/main/.github/workflows/verify-commits.yml) that means others can't merge unsigned commits from maintainers.

builtin/credential/approle/path_login.go
```diff
@@ -389,16 +389,29 @@ func (b *backend) pathLoginUpdate(ctx context.Context, req *logical.Request, dat
 
 // Invoked when the token issued by this backend is attempting a renewal.
 func (b *backend) pathLoginRenew(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-	roleName := req.Auth.InternalData["role_name"].(string)
-	if roleName == "" {
+	// Fetch the role_name from internal data
+	roleName, ok := req.Auth.InternalData["role_name"].(string)
+	if !ok || roleName == "" {
```
cipherboy, line 394: Given `InternalData` is a `map[string]string`, I think this is equivalent to the [existing logic](https://go.dev/play/p/AemIH7kDzt5)?

builtin/credential/approle/path_login.go
```diff
@@ -389,16 +389,29 @@ func (b *backend) pathLoginUpdate(ctx context.Context, req *logical.Request, dat
...
 		return nil, fmt.Errorf("failed to fetch role_name during renewal")
 	}
 
-	lock := b.roleLock(roleName)
```
cipherboy, line 397: So I think we're going to need to leave the lock and start the transaction afterwards; see https://github.com/openbao/openbao/pull/558 for info. Essentially, Raft transactions aren't blocking in the same way that PostgreSQL are, so if there's no lock, we'll cause conflicts when they otherwise would've succeeded due to this lock.

builtin/credential/approle/path_login.go
```diff
@@ -389,16 +389,29 @@ func (b *backend) pathLoginUpdate(ctx context.Context, req *logical.Request, dat
...
+		txn, err = txnStorage.BeginTx(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to begin transaction: %w", err)
+		}
+		defer txn.Rollback(ctx) // Ensure rollback if anything goes wrong
+		req.Storage = txn
+	}
+
+	// Fetch the role entry from storage
 	role, err := b.roleEntry(ctx, req.Storage, roleName)
```
cipherboy, line 415: Given we only read a single entry, I'm not convinced this is the best place to use a transaction. I avoided using one in the case of KV's config read for instance; see https://github.com/openbao/openbao/pull/560#issuecomment-2389461300 for instance.

cipherboy: I think `path_role.go` has more entries, though I've not finished auditing everything in `path_login.go` to confidently say nothing here doesn't need roles.

cipherboy (commented):
@v0nNemizez Some thoughts! :D

builtin/credential/approle/path_login.go
v0nNemizez: Okay! Thanks for The input :)

v0nNemizez: so all Create/update/delete operations shoud use transactions in your opinion?

cipherboy: @v0nNemizez I think so. Create/Update especially as you want to make sure it hasn't changed between when you read it (to modify the values) and when you write it back out.

cipherboy: I take that back. Anything that deleted a single entry without reading it would be atomic. But deleting a role wouldn't be as you also have to delete all secret IDs.

cipherboy:
@v0nNemizez hmm it looks like there's still old commits: https://github.com/openbao/openbao/pull/617/commits

v0nNemizez:
this bacame a mess... i will open a new one :)

v0nNemizez:
making a new pr

PR #618, opened 2024-10-12
Allow JWT claims to template ACL policies

cipherboy:
In partial solution to [glob based claim matching](https://lists.lfedge.org/g/openbao-dev-wg/topic/glob_based_group_alias/108635285) discussed on the Dev WG, this implements `sdk/helper/template`-based templating of ACL policies onto a token, from the claims on a JWT or OIDC token.

In particular, `text/template` allows complex expressions such as `if` and `range` expressions, allowing us to restrict policy attachment: while we don't allow creating more than one policy per template, we do remove empty policies from the final list, which allows us to over-provision templates to policies if we need to sometimes allow multiple policies per matching claim.

We enforce that missing claims are fatal. However, callers can work around this with some conditional templating logic: 

```ruby
{{ if ne nil (index . "optional_claim") }}{{ .optional_claim }}{{ end }}
```

----

cc @DrDaveD @fcbajao


sdk/helper/template/template.go
```diff
@@ -73,6 +84,10 @@ func Function(name string, f interface{}) Opt {
 //   - base64 encodes the previous value.
 //     Example: {{ .DisplayName | base64 }}
 //
+// - unbase64
```
JanMa, line 87: nitpic: something like `base64dec` would be a more familiar naming

JanMa (approved):

sdk/helper/template/template.go
cipherboy: Yeah, the existing one isn't `base64enc` tho, so I feel weird naming it `base64dec`... Maybe `decbase64`?

JanMa: Another function is called `truncate_sha256, so how about `decode_base64` ?

cipherboy: Nice, I like that one!

DanGhita (approved):
This seems to me a sensitive feature and, at the moment, I'm not in a position to know whether there might be side effects (including security flaws).
Maybe we should have a talk about this during  the next WG meeting.
Many thanks!

website/content/api-docs/auth/jwt.mdx
```diff
@@ -177,6 +177,7 @@ entities attempting to login. At least one of the bound values must be set.
   time the user was actively authenticated with the OIDC provider. If set, the `max_age` request parameter
   will be included in the authentication request. See [AuthRequest](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)
   for additional details. Accepts an integer number of seconds, or a Go duration format string.
+- `token_policies_template_claims` `(bool: false)` - When enabled, allows token_policies to represent templates which are computed with all claims on the underlying JWT or OIDC token. This uses [`sdk/helper/template`](https://pkg.go.dev/github.com/openbao/openbao/sdk/v2@v2.0.1/helper/template) which is based on [`text/template`](https://pkg.go.dev/text/template).
```
DrDaveD, line 180: This explanation is quite hard to follow.  Can it (or somewhere else in the docs) be explained by example?

When you say `OIDC token` here, are you talking in particular about the OIDC id_token?

DrDaveD (changes requested):
I would like to participate in the discussion of this, although unfortunately I will not be able to attend this week's meeting.

website/content/api-docs/auth/jwt.mdx
cipherboy: Yes, this would be from the id_token if you're doing auth with an OIDC type role and not JWT:

https://github.com/openbao/openbao/blob/d393d9dae54dcb5234f654a3374dd9ec6e0d0be5/builtin/credential/jwt/path_oidc.go#L291

https://github.com/openbao/openbao/blob/d393d9dae54dcb5234f654a3374dd9ec6e0d0be5/builtin/credential/jwt/path_oidc.go#L332-L337

https://github.com/openbao/openbao/blob/d393d9dae54dcb5234f654a3374dd9ec6e0d0be5/builtin/credential/jwt/path_oidc.go#L436-L438

I'll work to revise the docs with some examples.

cipherboy: @DrDaveD Pushed; I've also sent an email to schedule a meeting with everyone.

cipherboy:
Thanks all! I rebased on top of #319 and will merge :-)

PR #622, opened 2024-10-13
Split mount tables with transactions

cipherboy:
OpenBao has an implicit limit on the number of mounts much lower than can theoretically fit in memory: the maximum Raft storage entry size. In practice, the use of compression (after applying JSON encoding to the table) lets us achieve a higher than expected limit. Large deployments can still hit this limit.

By relying on the newly introduced transactional storage, we can split the mount table into separate entries (one per mount), which is performant due to reuse of the bbolt transaction. This lets us avoid the lower entry limit at the risk of hitting the higher transaction limit in the future.

However, this also makes updates scoped: only reading or writing a large number of entries (combining path + hash) and then committing will hit this limit; individual mount additions or modifications will not trigger re-writing all entries and thus becomes safer.

We introduce this change for both auth and secret mounts, but not for audit mounts as audit mounts are not used as extensively as auth and secrets engines in practice. This lets us avoid unnecessary complexity there.

Note that this change cannot be retroactively reverted: users must take a snapshot before upgrading if they wish to later downgrade to the old (non-transactional) storage layout.

---

TODO:

 - [x] Add changelog entry
 - [x] Add RFC
 - [x] Add tests for the migration in particular

### DIFF
--- a/sdk/physical/error.go
+++ b/sdk/physical/error.go
@@ -19,38 +19,69 @@ const (
 )
 
 // ErrorInjector is used to add errors into underlying physical requests
-type ErrorInjector struct {
+type ErrorInjector interface {
+	Backend
+	SetErrorPercentage(int)
+}
+
+type errorInjector struct {
 	backend      Backend
 	errorPercent int
 	randomLock   *sync.Mutex
 	random       *rand.Rand
 }
 
-// Verify ErrorInjector satisfies the correct interfaces
+var _ ErrorInjector = &errorInjector{}
+
+type transactionalErrorInjector struct {
+	*errorInjector
+}
+
 var (
-	_ Backend = (*ErrorInjector)(nil)
+	_ ErrorInjector        = &transactionalErrorInjector{}
+	_ TransactionalBackend = &transactionalErrorInjector{}
+)
+
+type errorInjectorTransaction struct {
+	*errorInjector
+}
+
+var (
+	_ ErrorInjector = &errorInjectorTransaction{}
+	_ Transaction   = &errorInjectorTransaction{}
 )
 
 // NewErrorInjector returns a wrapped physical backend to inject error
-func NewErrorInjector(b Backend, errorPercent int, logger log.Logger) *ErrorInjector {
+func NewErrorInjector(b Backend, errorPercent int, logger log.Logger) ErrorInjector {
 	if errorPercent < 0 || errorPercent > 100 {
 		errorPercent = DefaultErrorPercent
 	}
-	logger.Info("creating error injector")
 
-	return &ErrorInjector{
+	if logger != nil {
+		logger.Info("creating error injector")
+	}
+
+	e := &errorInjector{
 		backend:      b,
 		errorPercent: errorPercent,
 		randomLock:   new(sync.Mutex),
 		random:       rand.New(rand.NewSource(int64(time.Now().Nanosecond()))),
 	}
+
+	if _, ok := b.(TransactionalBackend); ok {
+		return &transactionalErrorInjector{
+			e,
+		}
+	}
+
+	return e
 }
 
-func (e *ErrorInjector) SetErrorPercentage(p int) {
+func (e *errorInjector) SetErrorPercentage(p int) {
 	e.errorPercent = p
 }
 
-func (e *ErrorInjector) addError() error {
+func (e *errorInjector) addError() error {
 	e.randomLock.Lock()
 	roll := e.random.Intn(100)
 	e.randomLock.Unlock()
@@ -61,37 +92,65 @@ func (e *ErrorInjector) addError() error {
 	return nil
 }
 
-func (e *ErrorInjector) Put(ctx context.Context, entry *Entry) error {
+func (e *errorInjector) Put(ctx context.Context, entry *Entry) error {
 	if err := e.addError(); err != nil {
 		return err
 	}
 	return e.backend.Put(ctx, entry)
 }
 
-func (e *ErrorInjector) Get(ctx context.Context, key string) (*Entry, error) {
+func (e *errorInjector) Get(ctx context.Context, key string) (*Entry, error) {
 	if err := e.addError(); err != nil {
 		return nil, err
 	}
 	return e.backend.Get(ctx, key)
 }
 
-func (e *ErrorInjector) Delete(ctx context.Context, key string) error {
+func (e *errorInjector) Delete(ctx context.Context, key string) error {
 	if err := e.addError(); err != nil {
 		return err
 	}
 	return e.backend.Delete(ctx, key)
 }
 
-func (e *ErrorInjector) List(ctx context.Context, prefix string) ([]string, error) {
+func (e *errorInjector) List(ctx context.Context, prefix string) ([]string, error) {
 	if err := e.addError(); err != nil {
 		return nil, err
 	}
 	return e.backend.List(ctx, prefix)
 }
 
-func (e *ErrorInjector) ListPage(ctx context.Context, prefix string, after string, limit int) ([]string, error) {
+func (e *errorInjector) ListPage(ctx context.Context, prefix string, after string, limit int) ([]string, error) {
 	if err := e.addError(); err != nil {
 		return nil, err
 	}
 	return e.backend.ListPage(ctx, prefix, after, limit)
+}
+
+func (e *transactionalErrorInjector) BeginReadOnlyTx(ctx context.Context) (Transaction, error) {
+	txn, err := e.errorInjector.backend.(TransactionalBackend).BeginReadOnlyTx(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	ret := NewErrorInjector(txn, e.errorInjector.errorPercent, nil)
+	return &errorInjectorTransaction{ret.(*errorInjector)}, nil
+}
+
+func (e *transactionalErrorInjector) BeginTx(ctx context.Context) (Transaction, error) {
+	txn, err := e.errorInjector.backend.(TransactionalBackend).BeginTx(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	ret := NewErrorInjector(txn, e.errorInjector.errorPercent, nil)
+	return &errorInjectorTransaction{ret.(*errorInjector)}, nil
+}
+
+func (e *errorInjectorTransaction) Commit(ctx context.Context) error {
+	return e.errorInjector.backend.(Transaction).Commit(ctx)
+}
+
+func (e *errorInjectorTransaction) Rollback(ctx context.Context) error {
+	return e.errorInjector.backend.(Transaction).Rollback(ctx)
 }
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -8,6 +8,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/hashicorp/go-secure-stdlib/strutil"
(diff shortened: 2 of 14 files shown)


vault/mount_test.go
```diff
@@ -808,6 +829,23 @@ func testCore_MountTable_UpgradeToTyped_Common(
...
+
+		for _, txnEntry := range postTxnEntries {
+			t.Logf("removing entry: %v", path+"/"+txnEntry)
+			if err := c.barrier.Delete(context.Background(), path+"/"+txnEntry); err != nil {
+				t.Fatal(err)
+			}
+		}
+	}
+
+	// Write the pre-typed version
```
cipherboy, line 848: @JanMa @DanGhita This existing test is sufficient to handle the migration case, IMO, from a single-entry table to a multi-entry table.

website/content/docs/rfcs/split-mount-tables.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+This change will apply for both auth and secret engine mounts.
+
+OpenBao's storage semantics mean that entries can both be a directory and a file at the same time. This means we can write individual mount entries from `coreMountConfigPath` into `coreMountConfigPath/{uuid}` and treat the latter (`coreMountConfigPath/`) as a listable directory of mounts. Any modification to individual mounts can be done without impacting other mounts once the UUID is known. By wrapping all updates in a transaction, we can ensure modifications to the mount table are consistent. Further, because the mount table is kept in memory, we will rarely need to load the entire mount table from disk except when invalidating the legacy single-entry table.
+
+More precisely, we have the following places where the mount table is potentially adjusted:
+
+ - When starting up and loading the mount table. This will handle the migration from legacy to standard if necessary.
+ - When an invalidation occurs and we trigger a reload of the core (e.g., leadership changed).
+ - When [mounting](https://openbao.org/api-docs/system/mounts/#enable-secrets-engine) a new secrets engine.
```
DanGhita, line 37: The same thing is true for auth mounts too ?

DanGhita (approved):
It looks good to me,  but I'm not sure I've understood all the details.

vault/logical_system_test.go
```diff
@@ -2911,7 +2912,7 @@ func TestSystemBackend_rawReadWrite_Compressed(t *testing.T) {
 		)
 
 		// Read back and check gzip was applied by looking for prefix byte
-		req = logical.TestRequest(t, logical.ReadOperation, "raw/core/mounts")
+		req = logical.TestRequest(t, logical.ReadOperation, "raw/core/audit")
 		req.Data = map[string]interface{}{
 			"compressed": false,
 			"encoding":   "base64",
```
cipherboy, line 2918: @fatima2003 I forgot to mention yesterday, but remember how on last week's call we couldn't figure out how to read the certs entry with `sys/raw`? I had forgotten about `encoding=base64` as a parameter to the request. :D

fatima2003: Ohh, I will be trying this out :)) Thank you!

vault/auth.go
```diff
@@ -499,50 +500,210 @@ func (c *Core) taintCredEntry(ctx context.Context, nsID, path string, updateStor
...
+		c.logger.Info("migrating legacy mount table to transactional layout")
+		needPersist = true
 	}
 	if rawLocal != nil {
 		localAuthTable, err := c.decodeMountTable(ctx, rawLocal.Value)
 		if err != nil {
-			c.logger.Error("failed to decompress and/or decode the local mount table", "error", err)
-			return err
+			c.logger.Error("failed to decompress and/or decode the legacy local auth mount table", "error", err)
+			return true, err
```
codergs, line 678: this true is very silent. it can use a comment - it counted as legacy as we found some legacy non-local mounts.

vault/auth.go
```diff
@@ -608,7 +769,31 @@ func (c *Core) loadCredentials(ctx context.Context) error {
...
+		barrier = c.barrier
+	}
+
+	// Gracefully handle a transaction-aware backend, if a transaction
+	// wasn't created for us. This is safe as we do not support nested
+	// transactions.
+	needTxnCommit := false
+	if _, ok := barrier.(logical.TransactionalStorage); ok {
+		var err error
+		barrier, err = barrier.(logical.TransactionalStorage).BeginTx(ctx)
```
codergs, line 785: nit: defend against assertion safety

vault/mount.go
```diff
@@ -1286,15 +1469,39 @@ func (c *Core) runMountUpdates(ctx context.Context, needPersist bool) error {
...
+		barrier = c.barrier
+	}
+
+	// Gracefully handle a transaction-aware backend, if a transaction
+	// wasn't created for us. This is safe as we do not support nested
+	// transactions.
+	needTxnCommit := false
+	if _, ok := barrier.(logical.TransactionalStorage); ok {
+		var err error
+		barrier, err = barrier.(logical.TransactionalStorage).BeginTx(ctx)
```
codergs, line 1493: nit: assertion safety

vault/auth.go
```diff
@@ -499,50 +500,210 @@ func (c *Core) taintCredEntry(ctx context.Context, nsID, path string, updateStor
...
+		if _, ok := barrier.(logical.Transaction); ok {
+			return false, nil
+		}
+		c.logger.Info("no mounts in legacy auth table; adding default mount table")
 		c.auth = c.defaultAuthTable()
 		needPersist = true
-	} else {
-		// only record tableMetrics if we have loaded something from storge
-		c.tableMetrics(len(c.auth.Entries), false, true, raw.Value)
+	} else if _, ok := barrier.(logical.Transaction); ok {
```
codergs, line 669: the legacy path should be similar to what it was before tx aware backend - is the else dropped here necessary?

codergs (commented):
Looks good overall. left some remarks!

sdk/physical/error.go
```diff
@@ -19,38 +19,69 @@ const (
...
 	backend      Backend
 	errorPercent int
 	randomLock   *sync.Mutex
 	random       *rand.Rand
 }
 
-// Verify ErrorInjector satisfies the correct interfaces
+var _ ErrorInjector = &errorInjector{}
+
+type transactionalErrorInjector struct {
```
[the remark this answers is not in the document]
cipherboy: @codergs Hmm, for a lot of the simple implementations, especially where there's reuse of non-transactional methods (e.g., `errorInjectorTransaction.Get(...)` calls `errorInjector.Get(...)` implicitly), I've opted to leave the methods in the same place so its easier to visually see and jump between them. Ideally, everything would have a transactional counterpart.

From a naming standpoint, maybe this file should be called `error_injector.go`; if we were to split it, probably `error_injector_transactional.go`? 

Thoughts?

sdk/physical/error.go
```diff
@@ -61,37 +92,65 @@ func (e *ErrorInjector) addError() error {
...
-func (e *ErrorInjector) ListPage(ctx context.Context, prefix string, after string, limit int) ([]string, error) {
+func (e *errorInjector) ListPage(ctx context.Context, prefix string, after string, limit int) ([]string, error) {
 	if err := e.addError(); err != nil {
 		return nil, err
 	}
 	return e.backend.ListPage(ctx, prefix, after, limit)
 }
+
+func (e *transactionalErrorInjector) BeginReadOnlyTx(ctx context.Context) (Transaction, error) {
+	txn, err := e.errorInjector.backend.(TransactionalBackend).BeginReadOnlyTx(ctx)
```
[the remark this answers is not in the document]
cipherboy: `transactionalErrorInjector` is private, so there'd be no way to construct a bad instance outside of this package, except via `NewErrorInjector(...)` which handles the casting differences for us and creates either an `errorInjector` or `transactionalErrorInjector` based on the underlying backend's type.

vault/auth.go
```diff
@@ -499,50 +500,210 @@ func (c *Core) taintCredEntry(ctx context.Context, nsID, path string, updateStor
 
 // loadCredentials is invoked as part of postUnseal to load the auth table
 func (c *Core) loadCredentials(ctx context.Context) error {
-	// Load the existing mount table
-	raw, err := c.barrier.Get(ctx, coreAuthConfigPath)
+	// Previously, this lock would be held after attempting to read the
```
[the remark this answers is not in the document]
cipherboy: Hmm, `loadCredentials` is only called during unseal, which is strictly linear anyways and does not allow handling of requests yet, so I actually don't think this lock is necessary, much less performance sensitive. But while I'm inclined to leave it in case we later restructure unseal... I don't think the refactor into a `WithLock`/`WithoutLock` makes sense to do? 

Thoughts?

vault/auth.go
```diff
@@ -499,50 +500,210 @@ func (c *Core) taintCredEntry(ctx context.Context, nsID, path string, updateStor
...
+		c.logger.Info("reading transactional auth mount table")
+		if err := c.loadTransactionalCredentials(ctx, txn); err != nil {
+			return fmt.Errorf("failed to load transactional auth mount table: %w", err)
+		}
+	}
+
+	// Update entries in memory: we don't sync the root namespace ID to save
+	// storage space, but we do want to validate all referenced namespaces
+	// exist.
+	for _, entry := range c.mounts.Entries {
```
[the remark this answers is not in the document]
cipherboy: @codergs Ah, this is actually redundant. `runCredentialUpdates` handles this for us. I'll remove!

vault/auth.go
cipherboy: Ah, we should probably make sure we emit metrics, good catch.

Otherwise, the else-if is necessary to ensure we write the legacy entries to the new format.

vault/auth.go
cipherboy: I think this true should actually be `false` fwiw. We err here, so the transaction will be aborted and unseal will ultimately fail fwiw.

vault/auth.go
cipherboy: That is done two lines above. :-) If barrier changes between now and then, we've got bigger problems :-D

vault/mount.go
cipherboy: As above, this is guarded by two lines above if I'm not mistaken. :-)

sdk/physical/error.go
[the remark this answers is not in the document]
codergs: error_injector_transactional.go <- works

vault/auth.go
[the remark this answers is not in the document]
codergs: Actually, with the above arguments around this only happening at unseal, I see the linearity case your are making. In that case, this is ok.

vault/auth.go
codergs: yeah, with current setup true or false yield the same result. so I was not sure why we treated this as true - only thing I could think of was we read some global mounts (may be)

sdk/physical/error.go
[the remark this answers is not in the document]
codergs: ack - sound argument. keep as is.

vault/auth.go
codergs: can't see. maybe make use of val, ok - as you are already doing casting on a previous step.

sdk/physical/error.go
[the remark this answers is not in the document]
codergs: I usually feel splitting is better UX here for readers as transactional stuff came later. But totally up to you here as these are just simple breakouts.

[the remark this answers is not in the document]
cipherboy: Cool, I'll go ahead and resolve this in a follow-up PR; other prior changes similar to this have already merged so it'd be good to be consistent across all of them.

codergs (approved):

codergs:
LGTM, CI is unhappy though.

cipherboy:
@codergs Ah yeah, I hadn't correctly handled metrics.

cipherboy:
Thanks all!

PR #624, opened 2024-10-13
Fix cache spamming logs on transaction creation; limit cache size for transactions

cipherboy:
```
commit cebec7772febe00f286b9e3ad4f670d7784b2798 (HEAD -> fix-cache-spam-logs-on-txn, origin/fix-cache-spam-logs-on-txn)
Author: Alexander Scheel <ascheel@gitlab.com>
Date:   Sun Oct 13 19:02:38 2024 -0400

    Reduce size of cache for each transaction
    
    While transactions should be short-lived, we wish to prevent leaked
    transactions from consuming too much memory due to excessive cache
    usage. While the cache itself starts empty, a transaction reading and
    writing lots of entries could grow the cache: by decreasing by a factor
    of DefaultParallelTransactions, we can ensure we only grow the total
    memory used by the combined caches by 2x.
    
commit d3aeb49b73f756f5cb1f39b4ee21568a30ad65fc
Author: Alexander Scheel <ascheel@gitlab.com>
Date:   Sun Oct 13 19:01:54 2024 -0400

    Prevent spurious logging on transaction creation
    
    Every transaction created would result in details about the cache size
    being printed; this was always the same as the parent's cache and so is
    always extraneous information.
    
```

### DIFF
--- a/sdk/physical/cache.go
+++ b/sdk/physical/cache.go
@@ -113,6 +113,11 @@ func NewCache(b Backend, size int, logger log.Logger, metricSink metrics.MetricS
 	if logger.IsDebug() {
 		logger.Debug("creating LRU cache", "size", size)
 	}
+
+	return newCache(b, size, logger, metricSink)
+}
+
+func newCache(b Backend, size int, logger log.Logger, metricSink metrics.MetricSink) Cache {
 	if size <= 0 {
 		size = DefaultCacheSize
 	}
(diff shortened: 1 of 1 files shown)


sdk/physical/cache.go
```diff
@@ -274,7 +283,7 @@ func (c *cache) cloneWithStorage(b Backend) *cache {
 	// fresh, localized cache. This is globally sub-optimal (as it starts
 	// with an empty cache), but easiest to implement (as the transaction can
 	// modify its cache as it pleases).
-	cacheCopy := NewCache(b, c.size, c.logger, c.metricSink).(*cache)
+	cacheCopy := newCache(b, c.size/TransactionCacheFactor, c.logger, c.metricSink).(*cache)
```
DanGhita, line 286: I suppose there is no risk to have a division by zero here.

DanGhita (approved):

sdk/physical/cache.go
cipherboy: @DanGhita no, the numerator is non-zero by guard in NewCache and the denominator is a packages-level constant defined above (set to the value of another packages-level constant which is also non-zero).

cipherboy:
@DanGhita @JanMa Doing a warm cache here would be ideal, but `golang-lru` doesn't support a `Clone()` operation... So I think reducing the size of the new cache is probably ideal enough for now.

PR #625, opened 2024-10-13
Ensure Raft transaction complies with limits

cipherboy:
Ensure all storage operations can be equally performed inside and outside of a transaction; that is, that an individual Put within a transaction does not exceed the limits of a Put outside of a transaction.


DanGhita (approved):

cipherboy:
Thanks @DanGhita for your reviews!

PR #632, opened 2024-10-14
Move to faster golang-lru/v2

cipherboy:
This moves to the faster, maintained `golang-lru/v2`. We'll want to be on `v2` if we want to contribute a warm-cache implementation to fix benchmarks in transactions (see #624). 

So far I haven't seen many performance improvements with a warm cache; a later PR will introduce the performance improvements I wanted to see (thanks to Jan for identifying them!).

I will note we don't have many uses of generics in OpenBao so far, but many of these caches are heavily used so I'm inclined to take the faster implementation for them.


JanMa (approved):

cipherboy:
Thanks, @JanMa!

PR #636, opened 2024-10-16
import API v2

klaus-sap:
## Issue

Without v2 the old release v1.12.2 gets imported

## Target Release

2.0.3


cipherboy (approved):
Ahh, I hadn't replaced it in the website. Thanks, @klaus-sap!

PR #638, opened 2024-10-16
PKI - Allow revocation of expired certificates

### LINKED ISSUE #459: PKI - Allow revocation of expired certificates (Vault #27609)
As reported by @tchernobog on https://github.com/hashicorp/vault/issues/27609:

> This is a follow-up of [#19452](https://github.com/hashicorp/vault/issues/19452).
> 
> **Is your feature request related to a problem? Please describe.**
> 
> This is for embedded software development.
> 
> We use the Vault PKI to sign software releases. Releases themselves are signed by a certificate which has a very short expiration date (hours), since they should not be used for anything else than signing the current release and then the private key is thrown away.
> 
> If however we have a software release which is deemed insecure, we use the CRL to block software updates to that version. Since a CVE / security finding can happen at any point in the years (!) to come, we need to be able to revoke an already expired certificate.
> 
> **Describe the solution you'd like**
> 
> We would like an optional parameter to the `/pki/revoke` endpoint which skips the [expiration check](https://github.com/hashicorp/vault/blob/89276a56b2267ca49041be0bd9c7be7c96d3b94d/builtin/logical/pki/crl_util.go#L998) in `crl_util.go`.
> 
> Alternatively, it might be even better and even easier to augment `/pki/config/crl` to a boolean option `allow_expired_cert_revocation`.
> 
> **Describe alternatives you've considered**
> 
> Extending the lifetime of the code signing certificate to match the intermediate. However, since also intermediates are rotated on a yearly or 6-months basis, we still have the same issue as any leaf certificate expiration will be bounded by that, and a revocation can happen far later.
> 
> We can obviously use `/pki/issuer/:issuer_ref/sign-revocation-list` and maintain that ourselves.
> 
> Else we would need to sign the "CRL" with a detached signature (and handle that in code).
> 
> **Explain any additional use-cases**
> 
> See [#19452](https://github.com/hashicorp/vault/issues/19452).
> 
> **Additional context**
> 
> N/A

---

19452's use case was less compelling than this one.

Cosign uses short-lived certificates (10 minutes) but currently [lacks support for BYO PKI revocation](https://github.com/sigstore/cosign/issues/2568). However, a similar use case to OP could definitely be implemented in the future: point-valid certificates whose (much) later revocation signals that the original certificate's signature should no longer be trusted or an upgrade skipped. This could be supported today with Cosign's verify-by-public-key, assuming the relevant cert/CRL validation infrastructure was handled OOB.

While strictly the CRL isn't an ideal mechanism (the certificate itself was not compromised and the corresponding key likely isn't recoverable), it could still be a useful mechanism for this since it is rather ubiquitous -- and often has different failure modes than cert expiry.

As proposed, allowing non-RFC 5280 behavior conditionally via a flag could be viable. This requires careful attention to tidy (as it will remove any expired certificate within the `safety_buffer`). 

@tchernobog -- Is this an implementation of a standard by any chance?

fatima2003:
This PR introduces the ability to revoke expired certificates by adding an `allow_expired_cert_revocation` field to the CRL configuration. This field can be globally enabled by the operator in the `config/crl` endpoint, allowing expired certificates to be revoked across the system. The feature is a global setting to ensure deliberate use and avoid unnecessary overhead or arbitrary certificate revocations.

Resolves #459 Vault #27609

## Target Release

1.14.7

___
**TO DO LIST**

- [X] Add `allow_expired_cert_revocation` field and check there's no issues parsing.
- [X] Update revocation behavior.
- [X] Create tests for revoking expired-but-not-revoked certificates.

### DIFF
--- a/builtin/logical/pki/crl_test.go
+++ b/builtin/logical/pki/crl_test.go
@@ -1506,32 +1506,12 @@ func TestCRLIssuerRemoval(t *testing.T) {
 func TestRevokeExpiredCert(t *testing.T) {
 	t.Parallel()
 
-	coreConfig := &vault.CoreConfig{
-		LogicalBackends: map[string]logical.Factory{
-			"pki": Factory,
-		},
-		EnableRaw: true,
-	}
-	cluster := vault.NewTestCluster(t, coreConfig, &vault.TestClusterOptions{
-		HandlerFunc: vaulthttp.Handler,
-	})
-	cluster.Start()
-	defer cluster.Cleanup()
-	client := cluster.Cores[0].Client
-
-	err := client.Sys().Mount("pki", &api.MountInput{
-		Type: "pki",
-		Config: api.MountConfigInput{
-			DefaultLeaseTTL: "16h",
-			MaxLeaseTTL:     "60h",
-		},
-	})
-	require.NoError(t, err)
+	b, s := CreateBackendWithStorage(t)
 
 	// Generate a root certificate for the PKI.
-	resp, err := client.Logical().Write("pki/root/generate/internal", map[string]interface{}{
+	resp, err := CBWrite(b, s, "root/generate/internal", map[string]interface{}{
 		"ttl":         "40h",
-		"common_name": "Root X1",
+		"common_name": "myvault.com",
 		"key_type":    "ec",
 	})
 	require.NoError(t, err)
@@ -1540,30 +1520,15 @@ func TestRevokeExpiredCert(t *testing.T) {
 	require.NotEmpty(t, resp.Data["issuer_id"])
 
 	// Create a role in the PKI backend for issuing certificates.
-	_, err = client.Logical().Write("pki/roles/local-testing", map[string]interface{}{
+	_, err = CBWrite(b, s, "roles/local-testing", map[string]interface{}{
 		"allow_any_name":    true,
 		"enforce_hostnames": false,
 		"key_type":          "ec",
 	})
 	require.NoError(t, err)
 
-	// Ensure that the CRL config returns default values before being set.
-	resp, err = client.Logical().Read("pki/config/crl")
-	require.NoError(t, err)
-	require.NotNil(t, resp)
-	require.NotNil(t, resp.Data)
-	// Check that the default CRL configuration values are returned correctly.
-	require.Equal(t, resp.Data["expiry"], defaultCrlConfig.Expiry)
-	require.Equal(t, resp.Data["disable"], defaultCrlConfig.Disable)
-	require.Equal(t, resp.Data["ocsp_disable"], defaultCrlConfig.OcspDisable)
-	require.Equal(t, resp.Data["auto_rebuild"], defaultCrlConfig.AutoRebuild)
-	require.Equal(t, resp.Data["auto_rebuild_grace_period"], defaultCrlConfig.AutoRebuildGracePeriod)
-	require.Equal(t, resp.Data["enable_delta"], defaultCrlConfig.EnableDelta)
-	require.Equal(t, resp.Data["delta_rebuild_interval"], defaultCrlConfig.DeltaRebuildInterval)
-	require.Equal(t, resp.Data["allow_expired_cert_revocation"], defaultCrlConfig.AllowExpiredCertRevocation)
-
 	// Issue a short-lived certificate and revoke it after it expires whilst allow_expired_cert_revocation is false.
-	resp, err = client.Logical().Write("pki/issue/local-testing", map[string]interface{}{
+	resp, err = CBWrite(b, s, "issue/local-testing", map[string]interface{}{
 		"ttl":         "1s",
 		"common_name": "example.com",
 	})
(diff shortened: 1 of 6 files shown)


builtin/logical/pki/path_revoke.go
```diff
@@ -151,6 +156,11 @@ signed by an issuer in this mount.`,
 				Description: `Key to use to verify revocation permission; must
 be in PEM format.`,
 			},
+			"allow_expired_cert_revocation": {
```
cipherboy, line 159: I think this should go on `config/crl` (in `path_config_crl`), not `revoke`: we want this to be a conscious choice by the operator to enable, globally, and not a per-certificate action by someone potentially unrelated. This is because it requires extra bookkeeping and cost and so shouldn't be done arbitrarily. My 2c! :-)

fatima2003: Oh yea, that makes sense, I didn't look at it that way. I'll make the changes :)

fatima2003: @cipherboy I made the changes and created a test, I'll work on `revoked_safety_buffer` now :)

cipherboy: Nice! That one goes on the tidy operation, not the crl config, btw :-)

fatima2003: Alrighty!

changelog/638.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:feature
```
cipherboy, line 1: Features are generally large changes... Something like building a role system for intermediate CAs or adding ACME support or something similar. I think this is better suited as an `improvement`. :-)

builtin/logical/pki/crl_test.go
```diff
@@ -1502,3 +1502,119 @@ func TestCRLIssuerRemoval(t *testing.T) {
 	}
 	require.Equal(t, len(afterCRLList), len(crlList))
 }
+
+func TestRevokeExpiredCert(t *testing.T) {
+	t.Parallel()
+
+	coreConfig := &vault.CoreConfig{
```
cipherboy, line 1509: I'd use:

```go
	b, s := CreateBackendWithStorage(t)
```

instead of the cluster setup as it is much faster and nothing we are doing here relies on a full cluster. :-)

builtin/logical/pki/crl_test.go
```diff
@@ -1502,3 +1502,119 @@ func TestCRLIssuerRemoval(t *testing.T) {
...
+		"common_name": "example.com",
+	})
+	require.NoError(t, err)
+	require.NotNil(t, resp)
+	require.NotNil(t, resp.Data)
+	require.NotEmpty(t, resp.Data["serial_number"])
+	leafSerial := resp.Data["serial_number"].(string)
+
+	// Wait for 1 second to ensure the certificate expires before revocation.
+	time.Sleep(1 * time.Second)
```
cipherboy, line 1577: I'd probably set this one to 2s as well.

builtin/logical/pki/crl_test.go
```diff
@@ -1502,3 +1502,119 @@ func TestCRLIssuerRemoval(t *testing.T) {
...
+	require.NotNil(t, resp.Data)
+	// Check that the default CRL configuration values are returned correctly.
+	require.Equal(t, resp.Data["expiry"], defaultCrlConfig.Expiry)
+	require.Equal(t, resp.Data["disable"], defaultCrlConfig.Disable)
+	require.Equal(t, resp.Data["ocsp_disable"], defaultCrlConfig.OcspDisable)
+	require.Equal(t, resp.Data["auto_rebuild"], defaultCrlConfig.AutoRebuild)
+	require.Equal(t, resp.Data["auto_rebuild_grace_period"], defaultCrlConfig.AutoRebuildGracePeriod)
+	require.Equal(t, resp.Data["enable_delta"], defaultCrlConfig.EnableDelta)
+	require.Equal(t, resp.Data["delta_rebuild_interval"], defaultCrlConfig.DeltaRebuildInterval)
+	require.Equal(t, resp.Data["allow_expired_cert_revocation"], defaultCrlConfig.AllowExpiredCertRevocation)
```
cipherboy, line 1563: I think this is tested by `TestPKI_EmptyCRLConfigUpgraded` so we can probably remove these. :-) 

Otherwise, the test looks good!

builtin/logical/pki/crl_test.go
fatima2003: Alright, I was testing 1s and 2s and accidentally pushed without making them the same. 1 sec works fine but maybe on a faster system it would fail so I'll keep both at 2s as suggested

builtin/logical/pki/crl_test.go
fatima2003: Alright, I was using another test as a template for mine and picked a test that was setup with a cluster. I will keep this in mind when deciding next time :)

builtin/logical/pki/crl_test.go
fatima2003: Alright, I saw this in the auto rebuild test as well so I thought maybe it was needed to be safe... I'll remove em and add `require.Equal(t, resp.Data["allow_expired_cert_revocation"], defaultCrlConfig.AllowExpiredCertRevocation)` to `TestPKI_EmptyCRLConfigUpgraded ` so that it is also tested for.

builtin/logical/pki/crl_test.go
cipherboy: If you wanted to take the time to parse the certificate, you could use `time.Now().Until(cert.NotAfter)` to sleep the exact right amount :-)

builtin/logical/pki/crl_test.go
fatima2003: Oh, alright, this makes sense, thank you!

builtin/logical/pki/crl_test.go
fatima2003: I think I'll try that so that the test doesn't take a minimum of 4 sec to run

changelog/638.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:improvement
+**PKI Secret Engine**: Added support for revoking expired certificates by introducing the 
+`allow_expired_cert_revocation` flag. 
+**PKI Secret Engine**: Added a new field `revoked_safety_buffer` to the CRL configuration, 
+allowing differentiation between retention times for revoked certificates and 
+expired-but-not-revoked certificates.
```
cipherboy, line 6:
```suggestion
secret/pki: Support revoking expired certificates iwth the `allow_expired_cert_revocation` CRL configuration.
secret/pki: Add `revoked_safety_buffer` to control retention on revoked certificates separately from expired certificates. 
```

cipherboy (commented):
New test updates look good @fatima2003!

cipherboy (approved):
Cool, I'll open a follow-up PR adding this to the docs :-) Thanks @fatima2003!

Issue #599, opened 2024-10-07
Update of RPM package overwrites existing configuration files


rafaelorafaelo:
**Describe the bug**
Update of RPM package overwrites configuration files in /etc/openbao.  

**To Reproduce**
1. Install OpenBao 2.0.1 `dnf install ./bao_2.0.1_linux_amd64.rpm`  
2. Edit `/etc/openbao/openbao.env` and `/etc/openbao/openbao.hcl`
3. Update OpenBao with `dnf up ./bao_2.0.2_linux_amd64.rpm`
4. Check configuration files, if your changes are still there.

**Expected behavior**
Existing configuration files should not be overwritten. Instead, files like `/etc/openbao/openbao.env.rpmnew` and `/etc/openbao/openbao.hcl.rpmnew` should be created. https://access.redhat.com/solutions/60263

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.0.2
* OpenBao CLI Version (retrieve with `bao version`): v2.0.2
* Server Operating System/Architecture: Rocky Linux 9.4 x86_64

ajaygk95:
I can help with this. I can create a PR for this.

Question: How do we build rpms for openbao in order to test this?
` make bootstrap && make dev` ?

cipherboy:
@ajaygk95 No, we build RPMs via goreleaser: https://goreleaser.com/quick-start/

You'll have to do the transform in our build pipeline first: https://github.com/openbao/openbao/blob/0ce0c4f8f5c645bb694bf3b9301bc62f146f877f/.github/workflows/release.yml#L128

I think it'd be great to have a makefile target for this, if you're willing!

ajaygk95:
I'm new to using gorelease, I tried to execute the go-release command locally and I'm facing the below issues,

```
# sed "s/REPLACE_WITH_RELEASE_GOOS/linux/g" .goreleaser-template.yaml > .goreleaser.yaml
# sed -i "s/^#LINUXONLY#//g" .goreleaser.yaml
# goreleaser release --clean --timeout=60m --verbose --parallelism 2
  • verbose output enabled
  • only configurations files on  version: 2  are supported, yours is  version: 0 , please update your configuration
  • using configuration                              path=.goreleaser.yaml
  • parallelism: 2
  • cleaning distribution directory
  • loading environment variables
  ⨯ release failed after 0s                  error=missing GITHUB_TOKEN, GITLAB_TOKEN and GITEA_TOKEN
```

[...]

PR #639, opened 2024-10-16
Don't override conf files in OpenBao binaries

ajaygk95:
Fixing https://github.com/openbao/openbao/issues/599

Fix : 
1. Make the conf files as `type: config` in .goreleaser-template.yaml so the files are not overwritten during rpm upgrade.
2. Add a new target in makefile to build rpm locally

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -350,3 +350,13 @@ release-changelog: $(wildcard changelog/*.txt)
 	@:$(if $(LAST_RELEASE),,$(error please set the LAST_RELEASE environment variable for changelog generation))
 	@:$(if $(THIS_RELEASE),,$(error please set the THIS_RELEASE environment variable for changelog generation))
 	changelog-build -changelog-template changelog/changelog.tmpl -entries-dir changelog -git-dir . -note-template changelog/note.tmpl -last-release $(LAST_RELEASE) -this-release $(THIS_RELEASE)
+
+.PHONY: dev-gorelease
+dev-gorelease: export GORELEASER_PREVIOUS_TAG := $(shell git describe --tags --exclude "api/*" --exclude "sdk/*" --abbrev=0)
+dev-gorelease: export GORELEASER_CURRENT_TAG := $(shell git describe --tags --exclude "api/*" --exclude "sdk/*" )
+dev-gorelease: export GPG_KEY_FILE := /dev/null
+dev-gorelease:
+	@echo GORELEASER_CURRENT_TAG: $(GORELEASER_CURRENT_TAG)
+	@$(SED) 's/REPLACE_WITH_RELEASE_GOOS/linux/g' $(CURDIR)/.goreleaser-template.yaml > $(CURDIR)/.goreleaser.yaml
+	@$(SED) -i 's/^#LINUXONLY#//g' $(CURDIR)/.goreleaser.yaml
+	goreleaser release --clean --timeout=60m --verbose --parallelism 2 --snapshot --skip docker,sbom,sign
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,6 +125,7 @@ jobs:
         if: startsWith(github.ref, 'refs/tags/')
         run: |
           sed "s/REPLACE_WITH_RELEASE_GOOS/${{ matrix.release_os }}/g" .goreleaser-template.yaml > .goreleaser.yaml
+          sed -i "s/^#OTHERARCH#//g" .goreleaser.yaml
           [ "${{ matrix.release_os }}" == "linux" ] && sed -i "s/^#LINUXONLY#//g" .goreleaser.yaml || true
 
       - name: "GoReleaser: Release"
--- a/.goreleaser-template.yaml
+++ b/.goreleaser-template.yaml
@@ -33,11 +33,11 @@ builds:
       - REPLACE_WITH_RELEASE_GOOS
     goarch:
       - amd64
-      - arm
-      - arm64
-      - ppc64le
-      - riscv64
-      - s390x
+#OTHERARCH#      - arm
+#OTHERARCH#      - arm64
+#OTHERARCH#      - ppc64le
+#OTHERARCH#      - riscv64
+#OTHERARCH#      - s390x
     goarm:
       - "6"
     ignore:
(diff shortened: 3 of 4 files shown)


ajaygk95:
@cipherboy - The makefile target still builds the rpm, and fails at

```
    deb:
      signature:
        key_file: "{{ .Env.GPG_KEY_FILE }}"
    rpm:
      signature:
        key_file: "{{ .Env.GPG_KEY_FILE }}"
```

This stage, is there any way to skip this signature verification when building rpms locally

cipherboy:
> This stage, is there any way to skip this signature verification when building rpms locally

I was looking for that option and didn't see it... What do you think about creating an ephemeral key in the `Makefile` and using that?

ajaygk95:
> > This stage, is there any way to skip this signature verification when building rpms locally
> 
> I was looking for that option and didn't see it... What do you think about creating an ephemeral key in the `Makefile` and using that?

Pointing `GPG_KEY_FILE=/dev/null` works, there would be no need to create a new key

cipherboy:
\o hey @ajaygk95 two things -- we need [DCO sign off before merging](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco-sign-off-policy):

> ```
> git commit --amend --signoff
> ```

Also, could we get a [changelog entry](https://github.com/openbao/openbao/blob/main/changelog/README.md#changelog) since this is a user-facing change? Something like:

    ```release-note:improvement
    rpm: Fix packaging to properly annotate configs entries for noreplace
    ```

in a file named `changelog/639.txt`?

Thanks!

ajaygk95:
> \o hey @ajaygk95 two things -- we need [DCO sign off before merging](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco-sign-off-policy):
> 
> > ```
> > git commit --amend --signoff
> > ```
> 
> Also, could we get a [changelog entry](https://github.com/openbao/openbao/blob/main/changelog/README.md#changelog) since this is a user-facing change? Something like:
> 
> ```
> ```release-note:improvement
> rpm: Fix packaging to properly annotate configs entries for noreplace
> ```
> ```
> 
> in a file named `changelog/639.txt`?
> 
> Thanks!

ACK !

Makefile
```diff
@@ -350,3 +350,17 @@ release-changelog: $(wildcard changelog/*.txt)
...
+	@echo GORELEASER_PREVIOUS_TAG: $(GORELEASER_PREVIOUS_TAG)
+	@if [ -z $(GORELEASER_CURRENT_TAG) ]; then \
+		echo "ERROR: Set GORELEASER_CURRENT_TAG to a specific version"; \
+		exit 1; \
+	fi
+	@echo GORELEASER_CURRENT_TAG: $(GORELEASER_CURRENT_TAG)
+	@$(SED) 's/REPLACE_WITH_RELEASE_GOOS/linux/g' $(CURDIR)/.goreleaser-template.yaml > $(CURDIR)/.goreleaser.yaml
+	@$(SED) -i 's/^#LINUXONLY#//g' $(CURDIR)/.goreleaser.yaml
+	@export GPG_KEY_FILE=/dev/null
+	goreleaser release --clean --timeout=60m --verbose --parallelism 2 --snapshot --skip docker,sbom,sign
```
cipherboy, line 366: This doesn't work for me; each makefile invocation is [a separate command](https://unix.stackexchange.com/questions/656437/export-env-variable-does-not-work-from-makefile) so the export only is valid for a single line and so by the time the echo executes, it is already a new command and thus doesn't have the envvar set. 

I'd suggest:

```suggestion
	@$(SED) 's/REPLACE_WITH_RELEASE_GOOS/linux/g' $(CURDIR)/.goreleaser-template.yaml > $(CURDIR)/.goreleaser.yaml
	@$(SED) -i 's/^#LINUXONLY#//g' $(CURDIR)/.goreleaser.yaml
	@export GPG_KEY_FILE=/dev/null
	GORELEASER_PREVIOUS_TAG=$(shell git describe --tags $(shell git rev-list --tags --max-count=1)) \
		GORELEASER_CURRENT_TAG=v999.999.999 \
		GPG_KEY_FILE=/dev/null \
		goreleaser release --clean --timeout=60m --verbose --parallelism 2 --snapshot --skip docker,sbom,sign
```

(note the trailing `\` -- so that that is all one large command).

cipherboy: I chose v999.999.999 as a unique version number we'll probably never hit and thus will disambiguate the package version, but open to thoughts here.

JanMa: The Make-native way to do this would be to declare [target-specific variables](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html). An alternative would be to use the [`.ONESHELL`](https://www.gnu.org/software/make/manual/html_node/One-Shell.html) command for the target. One thing to be aware of though is that the default Make version shipped by MacOS is too old and doesn't support `.ONESHELL`

cipherboy: Ah yeah, that's probably cleaner. Do we expect RPMs to build on MacOS? 🤔

ajaygk95: Fixed

Makefile
```diff
@@ -350,3 +350,13 @@ release-changelog: $(wildcard changelog/*.txt)
 	@:$(if $(LAST_RELEASE),,$(error please set the LAST_RELEASE environment variable for changelog generation))
 	@:$(if $(THIS_RELEASE),,$(error please set the THIS_RELEASE environment variable for changelog generation))
 	changelog-build -changelog-template changelog/changelog.tmpl -entries-dir changelog -git-dir . -note-template changelog/note.tmpl -last-release $(LAST_RELEASE) -this-release $(THIS_RELEASE)
+
+.PHONY: dev-gorelease
+dev-gorelease: export GORELEASER_PREVIOUS_TAG := $(shell git describe --tags $(shell git rev-list --tags --max-count=1))
+dev-gorelease: export GORELEASER_CURRENT_TAG := v999.999.999
```
JanMa, line 356: Instead of setting this to a nonsense number, I'd prefer if we do something like this
```suggestion
dev-gorelease: export GORELEASER_CURRENT_TAG := $(shell git describe --tags --exclude "api/*" --exclude "sdk/*" )
```
It will result in a tag like `v2.0.0-115-g1c0c5647bc`

Makefile
```diff
@@ -350,3 +350,13 @@ release-changelog: $(wildcard changelog/*.txt)
 	@:$(if $(LAST_RELEASE),,$(error please set the LAST_RELEASE environment variable for changelog generation))
 	@:$(if $(THIS_RELEASE),,$(error please set the THIS_RELEASE environment variable for changelog generation))
 	changelog-build -changelog-template changelog/changelog.tmpl -entries-dir changelog -git-dir . -note-template changelog/note.tmpl -last-release $(LAST_RELEASE) -this-release $(THIS_RELEASE)
+
+.PHONY: dev-gorelease
+dev-gorelease: export GORELEASER_PREVIOUS_TAG := $(shell git describe --tags $(shell git rev-list --tags --max-count=1))
```
JanMa, line 355: not sure if this behaves like we want it to? On `main` this currently shows `v2.0.2` as the previous tag, even though this tag was created on the `release/v2.0.x` branch. I'd suggest to use this command instead to find the last tag in the parent of the current branch
```suggestion
dev-gorelease: export GORELEASER_PREVIOUS_TAG := $(shell git describe --tags --exclude "api/*" --exclude "sdk/*" --abbrev=0)
```

JanMa (changes requested):
Two small remarks regarding the tags of local release builds, otherwise LGTM

JanMa (approved):
LGTM :+1:  Thanks a lot @ajaygk95

ajaygk95:
> LGTM 👍 Thanks a lot @ajaygk95

Thanks

PR #641, opened 2024-10-19
Exclude changelog directory from release archive

cipherboy:


.goreleaser-template.yaml
```diff
@@ -788,6 +788,10 @@ archives:
     format_overrides:
       - goos: windows
         format: zip
+    files:
```
cipherboy, line 791: https://goreleaser.com/customization/archive/ defaults to including both `CHANGELOG*` and `changelog*`, the latter is a directory in our case.

JanMa (approved):

PR #653, opened 2024-10-23
Add revoked_safety_buffer Field to CRL Configuration for Differentiated Retention of Revoked Certificates

fatima2003:
This PR aims to add a new field `revoked_safety_buffer` to the CRL configuration to differentiate between the retention of revoked certificates and expired-but-not-revoked certificates.

### **Key Differences**
`safety_buffer`: Retains expired certificates.
`revoked_safety_buffer`: Retains revoked certificates, regardless of whether they are expired or not.

### **How it works**
**1. Neither field is explicitly set:** 
$\quad$ If neither `safety_buffer` nor `revoked_safety_buffer` is set, `revoked_safety_buffer` defaults to the value of `safety_buffer`, which is 72 hours.
**2. Both fields are explicitly set:**
$\quad$ When both fields are set, `safety_buffer` and `revoked_safety_buffer` take on the explicitly defined values.
**3. Only revoked_safety_buffer is explicitly set:**
$\quad$ In this case, `revoked_safety_buffer` uses the set value, while `safety_buffer` defaults to its default value of 72 hours.
**4. Only safety_buffer is explicitly set:**
$\quad$ When only `safety_buffer` is set, `revoked_safety_buffer` defaults to the value of `safety_buffer` to maintain backward compatibility.

## Follow up of 
#638 

## Target Release

1.14.7

---
**TO DO** 
- [X] Implement `revoked_safety_buffer` so that its value is set as described above in the 4 conditions. 
- [X] Create a test to check if `revoked_safety_buffer` and `safety_buffer` are correctly set in the above 4 conditions.
- [X] Implement `revoked_safety_buffer` so that it is the buffer for revoked certs and `safety_buffer` so that it is the buffer for expired certs.
- [X] Write a test to check if `revoked_safety_buffer` is the buffer for revoked certs and `safety_buffer` is the buffer for expired certs.


builtin/logical/pki/storage.go
```diff
@@ -1394,6 +1394,22 @@ func (sc *storageContext) getAutoTidyConfig() (*tidyConfig, error) {
 		result.IssuerSafetyBuffer = defaultTidyConfig.IssuerSafetyBuffer
 	}
 
+	// These cases pass as default
+	// Case 1: SafetyBuffer && RevokedSafetyBuffer are default, RevokedSafetyBuffer is defaultTidyConfig.SafetyBuffer
+	// Case 2: SafetyBuffer is default && RevokedSafetyBuffer has been explicitly set: RevokedSafetyBuffer should be set value
+	// Case 3: SafetyBuffer && RevokedSafetyBuffer have been explicitly set: RevokedSafetyBuffer should be its set value
+
+	// FAILS
+	// Case 4: SafetyBuffer explicitly set && RevokedSafetyBuffer is unset: RevokedSafetyBuffer should be updated SafetyBuffer
```
fatima2003, line 1403: @cipherboy, this is my dilemma with case 4: Only safety_buffer is explicitly set.

I used printed the memory addresses for debugging, these are the results for the lines 1335 - 1344 of the test **TestRevokedSafetyBufferConfig** : 

    fmt.Printf("Address defaultTidyConfig RevokedSafetyBuffer is pointing to: %p\n", defaultTidyConfig.RevokedSafetyBuffer)
    fmt.Printf("Address of Default SafetyBuffer: %p\n", &defaultTidyConfig.SafetyBuffer)
    fmt.Printf("Address RevokedSafetyBuffer is pointing to: %p\n", result.RevokedSafetyBuffer)
    fmt.Printf("Value of *RevokedSafetyBuffer: %v\n", *result.RevokedSafetyBuffer)
    fmt.Printf("Address of RevokedSafetyBuffer: %v\n", &result.RevokedSafetyBuffer)
    fmt.Printf("Value of Default SafetyBuffer: %v\n", defaultTidyConfig.SafetyBuffer)
    fmt.Printf("Address of result.SafetyBuffer: %p\n", &result.SafetyBuffer)
    fmt.Printf("Value of result.SafetyBuffer: %v\n", result.SafetyBuffer)

```
Address defaultTidyConfig RevokedSafetyBuffer is pointing to: 0x106ce75f8
Address of Default SafetyBuffer: 0x106ce75f8
Address RevokedSafetyBuffer is pointing to: 0x1400031f8a0
Value of *RevokedSafetyBuffer: 72h0m0s
Address of RevokedSafetyBuffer: 0x140009fc160
Value of Default SafetyBuffer: 72h0m0s
Address of result.SafetyBuffer: 0x140009fc158
Value of result.SafetyBuffer: 1h0m0s
```

As seen above,` result.RevokedSafetyBuffer` does not point to `defaultTidyConfig.SafetyBuffer` if it is unset because it is reassigned here: 
https://github.com/openbao/openbao/blob/981935e68736b3da290442543b832871c77a421a/builtin/logical/pki/storage.go#L1383-L1387

builtin/logical/pki/path_tidy_test.go
```diff
@@ -1313,3 +1320,39 @@ func waitForTidyToFinish(t *testing.T, client *api.Client, mount string) *api.Se
...
+	b, s := CreateBackendWithStorage(t)
+
+	// Verify that the default of revoked_safety_buffer is the default of safety_buffer when neither are set
+	resp, err := CBWrite(b, s, "config/auto-tidy", map[string]interface{}{})
+	resp, err = CBRead(b, s, "config/auto-tidy")
+	requireSuccessNonNilResponse(t, resp, err, "expected to read auto-tidy config")
+	require.Equal(t, resp.Data["safety_buffer"].(int), resp.Data["revoked_safety_buffer"].(int), "expected revoked_safety_buffer to be set to safetyBuffer")
+
+	// Verify that revoked_safety_buffer defaults to safety_buffer when safety_buffer is set
+	safetyBuffer := 3600
```
fatima2003, line 1336: This is the part causing the test failure for reasons specified in https://github.com/openbao/openbao/pull/653#discussion_r1813695576

builtin/logical/pki/storage.go
cipherboy: Hmmm, I think the clever use of pointers might be causing problems. Typically I think we've left it `nil` until a value has been set explicitly by the operator. This is a bit more complex as I think you'd need more fallback logic (in places, to ensure you read `safety_buffer` when `revoked_safety_buffer` is nil), but what do you think about that approach?

builtin/logical/pki/path_tidy.go
```diff
@@ -115,6 +117,13 @@ var defaultTidyConfig = tidyConfig{
 	PublishMetrics:          false,
 }
 
+// assign the default value of revoked_safety_buffer to point to the default of safety_buffer
+func init() {
```
cipherboy, line 121: If you leave it `nil`, this can be removed, which I think makes things cleaner :-)

builtin/logical/pki/path_tidy_test.go
```diff
@@ -548,6 +554,7 @@ func TestTidyIssuerConfig(t *testing.T) {
 	defaultConfigMap["interval_duration"] = int(time.Duration(defaultConfigMap["interval_duration"].(float64)) / time.Second)
 	defaultConfigMap["issuer_safety_buffer"] = int(time.Duration(defaultConfigMap["issuer_safety_buffer"].(float64)) / time.Second)
 	defaultConfigMap["safety_buffer"] = int(time.Duration(defaultConfigMap["safety_buffer"].(float64)) / time.Second)
+	defaultConfigMap["revoked_safety_buffer"] = int(time.Duration(defaultConfigMap["revoked_safety_buffer"].(float64)) / time.Second)
```
cipherboy, line 557: In some of these places, while the response will include `revoked_safety_buffer` (because we always want to reply with it, even if it is `nil`), we probably will refer to the value of `safety_buffer` in those cases.

builtin/logical/pki/ca_test.go
```diff
@@ -689,9 +689,10 @@ func runSteps(t *testing.T, rootB, intB *backend, client *api.Client, rootName,
 		// Run with a short safety buffer and both set to true, both should be cleared
 		{
 			resp, err := client.Logical().Write(rootName+"tidy", map[string]interface{}{
-				"safety_buffer":      "1s",
-				"tidy_cert_store":    true,
-				"tidy_revoked_certs": true,
+				"safety_buffer":         "1s",
+				"revoked_safety_buffer": "1s",
```
fatima2003, line 693: This is a bit weird... There is a nil deference error when `revoked_safety_buffer` is not defined although this should not be the case since this is already handled in the function `pathTidyWrite`. 

I'm a lil confused, I'm not sure how to read the values from this endpoint. When I read the parameters by `client.Logical().Read(rootName + "config/auto-tidy")`, the values are the default values set (i.e., 72h) which makes sense since I haven't written the buffer values to this endpoint.

fatima2003:
Hi @cipherboy, I can squash the last 2 commits for a cleaner history: 0fada7ac0382dd4de4fd27579953edf62a8b110e and 670f889fb47d8581babb6c38ac384959e2fe7fef. I don't think these fixes are needed in the history but I'm not too sure...

builtin/logical/pki/path_tidy.go
```diff
@@ -657,6 +679,13 @@ available on the tidy-status endpoint.`,
 
 func (b *backend) pathTidyWrite(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
 	safetyBuffer := d.Get("safety_buffer").(int)
+	revokedSafetyBuffer := 0
+	if revokedSafetyBufferRaw, ok := d.GetOk("revoked_safety_buffer"); ok {
+		revokedSafetyBuffer = revokedSafetyBufferRaw.(int)
+	} else {
+		// Default to safety buffer's default if missing
+		revokedSafetyBuffer = int(defaultTidyConfig.SafetyBuffer.Seconds())
```
cipherboy, line 687: I think we just want to default to the safety buffer's value, not to its default here. Otherwise it is a behavior change.

builtin/logical/pki/backend_test.go
```diff
@@ -4189,9 +4189,10 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 
 	// Issue a tidy on /pki
 	_, err = client.Logical().Write("pki/tidy", map[string]interface{}{
-		"tidy_cert_store":    true,
-		"tidy_revoked_certs": true,
-		"safety_buffer":      "1s",
+		"tidy_cert_store":       true,
+		"tidy_revoked_certs":    true,
+		"safety_buffer":         "1s",
+		"revoked_safety_buffer": "1s",
```
cipherboy, line 4195: I would probably leave this test alone as well :-) We want to show that `revoked_safety_buffer` is fine to leave off the request, if you don't want to set it separately from `safety_buffer`.

builtin/logical/pki/fields.go
```diff
@@ -511,11 +511,20 @@ deactivated ACME account is deleted.`,
...
 		Default: int(defaultTidyConfig.SafetyBuffer / time.Second), // TypeDurationSecond currently requires defaults to be int
 	}
 
+	fields["revoked_safety_buffer"] = &framework.FieldSchema{
+		Type: framework.TypeDurationSecond,
+		Description: `The amount of extra time that must have passed
+beyond certificate revocation before it is removed
+from the revocation list.
+Defaults to safety_buffer (which defaults to 72 hours).`,
+		Default: nil,
```
cipherboy, line 525: Ahh, good catch! I had left this as line 516 above (`defaultTidyConfig.SafetyBuffer`), but this is more correct. :-)

builtin/logical/pki/storage.go
```diff
@@ -1394,6 +1394,12 @@ func (sc *storageContext) getAutoTidyConfig() (*tidyConfig, error) {
 		result.IssuerSafetyBuffer = defaultTidyConfig.IssuerSafetyBuffer
 	}
 
+	// If revoked safety buffer is not specified OR it points to the default of safety buffer while safety buffer is specified
+	if result.RevokedSafetyBuffer == nil || (*result.RevokedSafetyBuffer == defaultTidyConfig.SafetyBuffer && result.SafetyBuffer != defaultTidyConfig.SafetyBuffer) {
```
cipherboy, line 1398: I don't think the latter part of this conditional will ever hold.

cipherboy: I'd also do this same fix on line 1385 to "fix" the returned default tidy config value (set `result.RevokedSafetyBuffer = &revokedBufferValue`) -- not modifying the default, just the returned one of course.

builtin/logical/pki/path_tidy.go
```diff
@@ -1459,6 +1496,21 @@ func (b *backend) pathConfigAutoTidyWrite(ctx context.Context, req *logical.Requ
...
+			config.RevokedSafetyBuffer = new(time.Duration) // Allocate memory for pointer
+		}
+		*config.RevokedSafetyBuffer = time.Duration(revokedSafetyBufferRaw.(int)) * time.Second
+		if *config.RevokedSafetyBuffer < 1*time.Second {
+			return logical.ErrorResponse(fmt.Sprintf("revoked_safety_buffer must be at least one second; got: %v", revokedSafetyBufferRaw)), nil
+		}
+	} else {
+		// If no explicit value for revoked_safety_buffer, default it to SafetyBuffer
+		if config.RevokedSafetyBuffer == nil {
+			config.RevokedSafetyBuffer = &config.SafetyBuffer
```
cipherboy, line 1510: I would probably clone the value here again. I know we're just writing it to storage but I'd probably be safe and prevent accidental modifications from causing issues.

builtin/logical/pki/path_tidy.go
```diff
@@ -1459,6 +1496,21 @@ func (b *backend) pathConfigAutoTidyWrite(ctx context.Context, req *logical.Requ
 		}
 	}
 
+	if revokedSafetyBufferRaw, ok := d.GetOk("revoked_safety_buffer"); ok {
+		if config.RevokedSafetyBuffer == nil {
+			config.RevokedSafetyBuffer = new(time.Duration) // Allocate memory for pointer
+		}
+		*config.RevokedSafetyBuffer = time.Duration(revokedSafetyBufferRaw.(int)) * time.Second
```
cipherboy, line 1503:
```suggestion
		revokedSafetyBuffer := time.Duration(revokedSafetyBufferRaw.(int)) * time.Second
		*config.RevokedSafetyBuffer = &revokedSafetyBuffer
```

Should be fewer lines and avoids a weird heap allocation. :-)

builtin/logical/pki/path_tidy_test.go
```diff
@@ -1313,3 +1253,192 @@ func waitForTidyToFinish(t *testing.T, client *api.Client, mount string) *api.Se
 	t.Logf("got tidy status: %v", statusResp.Data)
 	return statusResp
 }
+
+func waitForAutoTidyToFinish(t *testing.T, client *api.Client) {
```
cipherboy, line 1257: Nice refactor!

cipherboy (commented):
Looking fairly good @fatima2003! Some minor comments that I think are good simplifications to have.

builtin/logical/pki/path_tidy.go
fatima2003: Good idea! Updated to prevent potential issues from accidental modifications.

builtin/logical/pki/path_tidy.go
fatima2003: Looks much better 👀

builtin/logical/pki/path_tidy.go
fatima2003: Alright, I don't remember why I did that... this is better :D

builtin/logical/pki/storage.go
fatima2003: Thanks you, this approach was less confusing to work with :)

builtin/logical/pki/ca_test.go
fatima2003: Yes I think this test was failing because I was defaulting revoked safety buffer to safety buffers default value instead of safety buffer itself when safety buffer was not specified... the test passes now :D

builtin/logical/pki/storage.go
fatima2003: The condition `result.SafetyBuffer != defaultTidyConfig.SafetyBuffer` was used because I was defaulting revoked to the default cause of safety buffer when it was not specified, the logic has been updated and the redundant check has also been removed :)

builtin/logical/pki/path_tidy.go
```diff
@@ -1459,6 +1496,20 @@ func (b *backend) pathConfigAutoTidyWrite(ctx context.Context, req *logical.Requ
...
 
+	if revokedSafetyBufferRaw, ok := d.GetOk("revoked_safety_buffer"); ok {
+		revokedSafetyBuffer := time.Duration(revokedSafetyBufferRaw.(int)) * time.Second
+		config.RevokedSafetyBuffer = &revokedSafetyBuffer
+		if *config.RevokedSafetyBuffer < 1*time.Second {
+			return logical.ErrorResponse(fmt.Sprintf("revoked_safety_buffer must be at least one second; got: %v", revokedSafetyBufferRaw)), nil
+		}
+	} else {
+		// If no explicit value for revoked_safety_buffer, default it to SafetyBuffer
+		if config.RevokedSafetyBuffer == nil {
```
cipherboy, line 1507: Hmm, I think this conditional is wrong. Consider:

```
$ bao read pki/config/auto-tidy
Key                                         Value
---                                         -----
... snip ...
revoked_safety_buffer                       259200
safety_buffer                               259200
... snip ...
$ bao write pki/config/auto-tidy safety_buffer=5s
Key                                         Value
---                                         -----
revoked_safety_buffer                       5
safety_buffer                               5
$ bao write pki/config/auto-tidy safety_buffer=8s
Key                                         Value
---                                         -----
revoked_safety_buffer                       5
safety_buffer                               8
```

In particular, the value for `revoked_safety_buffer` gets stuck on an old value of `safety_buffer` because of this `nil` check. But TBH, I'm not sure how to fix that. :-D

I _think_ ideally we'd leave the value nil until explicitly set?

That would require us to update logic in a lot of places, e.g., to set `revoked_safety_buffer = safety_buffer` if it is `nil` on display... but never in the actual config itself. Which I think is technically the most correct but also a fair bit of work. 

What's your thoughts? I'm happy to do that in a follow-up PR myself if you don't want to deal with that :D

fatima2003: I wrote a quick test for this scenario (and pushed it), and it initially failed. After reading your suggestion to handle `revoked_safety_buffer` as display-only, I removed the assignment `revokedSafetyBuffer = safetyBuffer` from the getter in **storage.go** and handled it in **path_tidy.go** where needed. The tests are passing now :) 

I tried out a few other scenarios on my own (like only setting revoked and not safety) and they passed, I hope I haven't missed out on any other use cases... if this solution is still lacking, idm if you open up a follow up PR for it or leave suggestions for me :D Thank you!

builtin/logical/pki/path_tidy_test.go
```diff
@@ -1313,3 +1253,206 @@ func waitForTidyToFinish(t *testing.T, client *api.Client, mount string) *api.Se
...
+	})
+	requireSuccessNonNilResponse(t, resp, err, "expected to be able to set safety_buffer")
+
+	resp, err = CBRead(b, s, "config/auto-tidy")
+	requireSuccessNonNilResponse(t, resp, err, "expected to read auto-tidy config")
+	require.Equal(t, safetyBuffer, resp.Data["revoked_safety_buffer"].(int), "expected revoked_safety_buffer to be set to safetyBuffer")
+	require.Equal(t, safetyBuffer, resp.Data["safety_buffer"].(int), "expected safety_buffer to be set to safetyBuffer")
+
+	// Verify that revoked_safety_buffer defaults to safety_buffer when safety_buffer is explicitly set multiple times,
+	// and revoked_safety_buffer is not set explicitly.
```
cipherboy, line 1320: Nice!

cipherboy (approved):
Looks good, thanks @fatima2003!

PR #665, opened 2024-10-28
Delete invalid certs during tidy operation

### LINKED ISSUE #659: Tidy errors out on certificates with duplicate extension (Vault #27219)
This is an issue reported on vault [issues/27219](https://github.com/hashicorp/vault/issues/27219) by @Garagoth. 

**Description**
When a certificate with duplicate extensions exists and tidy is run, the process fails:
```
unable to parse stored certificate with serial \"06-e0-83-88-22-be-91-7f-8d-07-2e-21-35-01-f1-66-95-a0-de-35\": x509: certificate contains duplicate extensions
```

This is the line catching the failure: 
https://github.com/openbao/openbao/blob/c9201295ed833b431249f4592f32b1946b69f263/builtin/logical/pki/path_tidy.go#L875

PR [16700](https://github.com/hashicorp/vault/pull/16700) solved the problem of creating these invalid certificates but there's no way to get rid of them.

**To Reproduce**
1. Have a certificate with duplicate extensions. 
2. Run tidy process.

**Expected behavior**
Since certificates with duplicate extensions are deemed invalid in [RFC 5290, Section 4.2](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2), a way to delete these invalid certificates should exist.

If the tidy operation is run with certificates with extra extensions, they should be skipped OR the tidy operation stops and a warning which includes the command to delete them should be thrown. 

---
I'm working on the fix for this :)

fatima2003:
This PR addresses the issue where the tidy operation fails when certificates with duplicate extensions are present in storage as described by @Garagoth in [Vault Issue #27219](https://github.com/hashicorp/vault/issues/27219). Certificates that have duplicate X.509 extensions are deemed invalid according to [RFC 5280, Section 4.2](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2), and can no longer be parsed by the Go x509.ParseCertificate function since Go 1.19 (due to golang/go#50988).

Additionally, the [parseCertificate](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/crypto/x509/parser.go;l=829) function in the Go library returns other errors when a certificate is invalid, such as when the issuer is malformed, &c.. Instead of only handling errors related to certificates with duplicate extensions, I suggest deleting any certificate that encounters parsing errors, as these certificates will be unusable anyway. The potential downside of this approach is that it would deem only certificates parsable by the official Go function as valid. However, this may not be an issue since that function is currently used for parsing. 

#### Potential Solutions:
1. Delete these certs during tidy. However, this risks that some certificates could be unintentionally deleted or worst case scenario, a bug with parsing could wipe out the cert store.

2. (**Currently implemented**) Have a separate flag for this task similar to `tidy_revoked_certs` called `tidy_invalid_certs` so tidying these certs could look like: 
```
bao write pki_int/tidy tidy_invalid_certs=true
```
This was suggested to eliminate the risk of deleting certs unintentionally.

3. Similar to option 2,  but instead of tidying all the invalid certs, the user gives the serial number of the invalid cert so they can pick which ones to delete:
```
bao write pki_int/tidy tidy_invalid_certs=[XXXXXXX, YYYYYYY]
```
Same benefit as option 2 but allows for finer control for cert deletion, though I'm not sure if invalid certs have a use case so maybe this is an overkill. 

#### TO DO 

- [X] Add flag to tidy invalid certs.
- [X]  Create test for this.
- [X] Add docs.

---
Resolves #659 [Vault #27219](https://github.com/hashicorp/vault/issues/27219)

## Target Release
1.14.7


fatima2003:
I realized a flaw in the implementation yesterday... currently the `tidy_invalid_certs` flag cannot be called on its own since I purposely did not add it to `AnyTidyConfig` (to reduce the number of times all certs are iterated over) and the functionality only works when `tidy_cert_store` is also enabled. I think this design is confusing to the end user. I assume that people will not have to enable this flag often and for now I think only those with certificates with duplicate extensions will find this useful so I will make the changes so that it can be called on its own...

cipherboy:
@fatima2003 Here's an invalid cert to test with, courtesy of the Golang test suite:

```
-----BEGIN CERTIFICATE-----
MIIBrjCCARegAwIBAgIBATANBgkqhkiG9w0BAQsFADAPMQ0wCwYDVQQDEwR0ZXN0
MCIYDzAwMDEwMTAxMDAwMDAwWhgPMDAwMTAxMDEwMDAwMDBaMA8xDTALBgNVBAMT
BHRlc3QwgZ8wDQYJKoZIhvcNAQEBBQADgY0AMIGJAoGBAMiFchnHms9l9NninAIz
SkY9acwl9Bk2AtmJrNCenFpiA17AcOO5q8DJYwdXi6WPKlVgcyH+ysW8XMWkq+CP
yhtF/+LMzl9odaUF2iUy3vgTC5gxGLWH5URVssx21Und2Pm2f4xyou5IVxbS9dxy
jLvV9PEY9BIb0H+zFthjhihDAgMBAAGjFjAUMAgGAioDBAIFADAIBgIqAwQCBQAw
DQYJKoZIhvcNAQELBQADgYEAlhQ4TQQKIQ8GUyzGiN/75TCtQtjhMGemxc0cNgre
d9rmm4DjydH0t7/sMCB56lQrfhJNplguzsbjFW4l245KbNKHfLiqwEGUgZjBNKur
ot6qX/skahLtt0CNOaFIge75HVKe/69OrWQGdp18dkay/KS4Glu8YMKIjOhfrUi1
NZA=
-----END CERTIFICATE-----
```

The serial number is 1, but that shouldn't cause any issues.

changelog/665.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+secret/pki: Delete invalid certificates during tidy if they cannot be parsed due to Go's x509 handling.
```
cipherboy, line 2:
```suggestion
secret/pki: Delete invalid certificates during tidy if they cannot be parsed due to Go's x509 handling and `tidy_invalid_certs=true`.
```

builtin/logical/pki/path_tidy.go
```diff
@@ -778,6 +803,12 @@ func (b *backend) startTidyOperation(req *logical.Request, config *tidyConfig) {
 				return tidyCancelledError
 			}
 
+			if config.InvalidCerts {
+				if err := b.doTidyInvalidCerts(ctx, req, logger, config); err != nil {
```
cipherboy, line 807: I don't know that I like the idea of this being a separate pass. We already have one pass which does this (`config.CertStore` controlling `b.doTidyCertStore`), so I think it is fine to reuse that for this pass as well. In the case of 1M+ certs, we don't really want to iterate over all the certificates twice, especially if most certs aren't invalid. :-)

My 2c!

builtin/logical/pki/path_tidy.go
```diff
@@ -872,7 +903,11 @@ func (b *backend) doTidyCertStore(ctx context.Context, req *logical.Request, log
 
 		cert, err := x509.ParseCertificate(certEntry.Value)
 		if err != nil {
-			return fmt.Errorf("unable to parse stored certificate with serial %q: %w", serial, err)
+			// if tidying invalid certs is not enabled, throw warning
+			if !config.InvalidCerts {
+				logger.Warn("unable to parse stored certificate with serial %q: %w; tidy by enabling tidy_invalid_certs", serial, err)
+			}
```
cipherboy, line 909: In particular, I'd make this logic something like:

```suggestion
			if !config.InvalidCerts && !haveWarned {
				logger.Warn("unable to parse stored certificate with serial %q: %w; tidy by enabling tidy_invalid_certs", serial, err)
				haveWarned = true
			} else {
				if err := req.Storage.Delete(ctx, "certs/"+serial); err != nil {
					return fmt.Errorf("error deleting invalid certificate %s: %w", serial, err)
				}
				b.tidyStatusIncCertStoreCount()
			}
```

or something. :D

cipherboy (commented):
Some initial thoughts!

builtin/logical/pki/path_tidy.go
cipherboy: On line 784: `if config.CertStore || config.InvalidCerts {` ...

builtin/logical/pki/path_tidy.go
```diff
@@ -872,10 +898,31 @@ func (b *backend) doTidyCertStore(ctx context.Context, req *logical.Request, log
...
+
+			// only log warning once
+			if !haveWarned {
+				logger.Warn("unable to parse stored certificate with serial %q: %w; %s. Other invalid certificates may exist.", serial, err,
+					func() string {
+						if config.InvalidCerts {
+							return "tidying up since it is not usable"
+						}
+						return "tidy by enabling tidy_invalid_certs"
+					}())
```
cipherboy, line 910: Ahh, creative! Just a note though, this isn't a `printf`-style logger, this is a `key=value` style logger with explicit message.

So I think we'd need something like the following:

```suggestion
				msg := "Unable to parse stored certificate. Other invalid certificates may exist; "
				if config.InvalidCerts {
					msg += "tidying up since it is not usable."
				} else {
					msg += "tidy by enabling tidy_invalid_certs=true."
				}
				logger.Warn(msg, "serial", serial, "err", err)
```

as otherwise you'll get weird logger messages. I believe this wouldn't err, but would output something like:

```
[WARN] msg="unable to parse stored certificate with serial %q: %w; %s. Other invalid certificates may exist."  1d:de:f6:bc:77:2e:56:cb:b7:00:bc:5b:ea:db:37:63:0a:3e:2d:98="x509: certificate contains duplicate extensions"
```

which wouldn't make as much sense.

changelog/665.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+secret/pki: Delete invalid certificates during tidy if they cannot be parsed due to Go's x509 handling  and `tidy_invalid_certs=true`.
```
cipherboy, line 2: nit:

```suggestion
secret/pki: Delete invalid certificates during tidy via `tidy_invalid_certs=true` if they cannot be parsed due to Go's x509 handling.
```

builtin/logical/pki/path_tidy.go
fatima2003: Yes, I initially also wanted it to be a single pass but figured since the `tidy_invalid_certs` would be rarely used it wouldn't matter so much but this is much better, thank you :D

builtin/logical/pki/path_tidy.go
```diff
@@ -872,7 +903,11 @@ func (b *backend) doTidyCertStore(ctx context.Context, req *logical.Request, log
...
 		if err != nil {
-			return fmt.Errorf("unable to parse stored certificate with serial %q: %w", serial, err)
+			// if tidying invalid certs is not enabled, throw warning
+			if !config.InvalidCerts {
+				logger.Warn("unable to parse stored certificate with serial %q: %w; tidy by enabling tidy_invalid_certs", serial, err)
+			}
+			continue
 		}
 
 		if time.Since(cert.NotAfter) > config.SafetyBuffer {
```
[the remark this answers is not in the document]
fatima2003: I made the changes :D Thank you for the review + suggestions.

builtin/logical/pki/path_tidy.go
fatima2003: Ah interesting, I will keep this in mind for future logging :D Thank you

changelog/665.txt
fatima2003: I'll add all these suggestions in one commit with the test :D Thank you!

fatima2003:
> @fatima2003 Here's an invalid cert to test with, courtesy of the Golang test suite:
> 
> ```
> -----BEGIN CERTIFICATE-----
> MIIBrjCCARegAwIBAgIBATANBgkqhkiG9w0BAQsFADAPMQ0wCwYDVQQDEwR0ZXN0
> MCIYDzAwMDEwMTAxMDAwMDAwWhgPMDAwMTAxMDEwMDAwMDBaMA8xDTALBgNVBAMT
> BHRlc3QwgZ8wDQYJKoZIhvcNAQEBBQADgY0AMIGJAoGBAMiFchnHms9l9NninAIz
> SkY9acwl9Bk2AtmJrNCenFpiA17AcOO5q8DJYwdXi6WPKlVgcyH+ysW8XMWkq+CP
> yhtF/+LMzl9odaUF2iUy3vgTC5gxGLWH5URVssx21Und2Pm2f4xyou5IVxbS9dxy
> jLvV9PEY9BIb0H+zFthjhihDAgMBAAGjFjAUMAgGAioDBAIFADAIBgIqAwQCBQAw
> DQYJKoZIhvcNAQELBQADgYEAlhQ4TQQKIQ8GUyzGiN/75TCtQtjhMGemxc0cNgre
> d9rmm4DjydH0t7/sMCB56lQrfhJNplguzsbjFW4l245KbNKHfLiqwEGUgZjBNKur
> ot6qX/skahLtt0CNOaFIge75HVKe/69OrWQGdp18dkay/KS4Glu8YMKIjOhfrUi1
> NZA=
> -----END CERTIFICATE-----
> ```
> 
> The serial number is 1, but that shouldn't cause any issues.

cipherboy:
@fatima2003 Ah, do a direct `s.Put(...)` operation on an entry with a Key of certs/1 and a value which is the DER of the cert (you'll want to base64 decode it).

fatima2003:
@cipherboy Oh okay, I add it like that :)

sos I accidentally clicked close with comment instead of comment 🙇🏽‍♀️

fatima2003:
@cipherboy if all looks good, I can rebase :D

builtin/logical/pki/path_tidy.go
```diff
@@ -872,10 +898,31 @@ func (b *backend) doTidyCertStore(ctx context.Context, req *logical.Request, log
 
 		cert, err := x509.ParseCertificate(certEntry.Value)
 		if err != nil {
-			return fmt.Errorf("unable to parse stored certificate with serial %q: %w", serial, err)
+
+			// only log warning once
+			if !haveWarned {
```
cipherboy, line 903: Should the same logic be added to `doTidyRevocationStore` as well I wonder?

I _think_ so, so perhaps we should wait until #677 lands to rebase on top of that (so that we correctly handle tidying each individually)....

Though I don't think we should add the `config.InvalidCerts` as an or to `if config.RevokedCerts || config.IssuerAssocs {` as we're already doing one iteration over the certs in this pass. That means we might miss certs (if for some reason, we have a revocation entry but no actual cert), but that should be fine.

My 2c!

fatima2003: I think that would be a good idea, maybe in the future we can have the option for `- verbose` in the case that a user wants to debug with details for whatever reason.

fatima2003: @cipherboy I will wait for #677 to land before I rebase this PR and modify the logging style in `doTidyRevocationStore`, then after this PR lands, I will open up the draft PR for using pagination in tidy and rebase there :D

cipherboy:
@fatima2003 Do you mind rebasing this and I'll try to take a look and get it into Beta? Thanks!

builtin/logical/pki/path_tidy.go
```diff
@@ -1031,7 +1080,9 @@ func (b *backend) doTidyRevocationStore(ctx context.Context, req *logical.Reques
 		}
 
 		if revokedEntry.Value == nil || len(revokedEntry.Value) == 0 {
-			logger.Warn("revoked entry has nil value; tidying up since it is no longer useful for any server operations", "serial", serial)
+			if !haveWarned {
+				logger.Warn("Revoked entry has nil value. Other invalid entries may exist; tidying up since it is no longer useful for any server operations", "serial", serial)
+			}
 			if err := req.Storage.Delete(ctx, "revoked/"+serial); err != nil {
 				return false, fmt.Errorf("error deleting revoked entry with nil value with serial %s: %w", serial, err)
 			}
```
cipherboy, line 1088: On new line 1100 is where I think I expected the `config.InvalidCerts` check, with additional warning if the cert copy in the revocation entry was of a bad certificate.

builtin/logical/pki/path_tidy.go
```diff
@@ -794,7 +819,7 @@ func (b *backend) startTidyOperation(req *logical.Request, config *tidyConfig) {
 			var revokedDeleted uint
 			var rebuildCRL bool
 
-			if config.CertStore {
+			if config.CertStore || config.InvalidCerts {
 				deleted, err := b.doTidyCertStore(ctx, req, logger, config)
 				if err != nil {
 					return err
```
cipherboy, line 825: That also means I think on new line 836, you should also run the revocation store tidy if you are supposed to delete invalid certificates. They should be deleted from both places, I feel? My 2c!

cipherboy (commented):
Hey Fatima, sorry about the delay! Two more minor things, it seems like I had forgotten to finish describing what I had meant by revocation should also warn + tidy invalid certs :-) 

Let me know what you think!

website/content/api-docs/secret/pki.mdx
```diff
@@ -4468,12 +4468,16 @@ to ensure this gets run periodically.
 - `tidy_cert_store` `(bool: false)` - Specifies whether to tidy up the certificate
   store.
 
-- `tidy_revoked_certs` `(bool: false)` - Set to true to remove all invalid and
-  expired certificates from storage. A revoked storage entry is considered
+- `tidy_revoked_certs` `(bool: false)` - Set to true to remove all invalid revoked 
+  certs and expired certificates from storage. A revoked storage entry is considered
```
cipherboy, line 4472:
```suggestion
- `tidy_revoked_certs` `(bool: false)` - Set to true to remove all revoked 
  and expired certificates from storage. A revoked storage entry is considered
```

I think is what I actually meant to write here :D

website/content/api-docs/secret/pki.mdx
```diff
@@ -4468,12 +4468,16 @@ to ensure this gets run periodically.
...
   invalid if the entry is empty, or the value within the entry is empty. If a
   certificate is removed due to expiry, the entry will also be removed from the
   CRL, and the CRL will be rotated.
 
+- `tidy_invalid_certs` `(bool: false)` - Set to true to remove all invalid certificates 
+  from storage. A certificate is considered invalid if it cannot be parsed by X.509's
+  certificate parser.
```
cipherboy, line 4479: nit:

```suggestion
  from storage. A certificate is considered invalid if it cannot be parsed by Go's
  X.509 certificate parser.
```

website/content/api-docs/secret/pki.mdx
```diff
@@ -4637,6 +4641,7 @@ $ curl \
     "tidy_revocation_queue": false,
     "tidy_revoked_cert_issuer_associations": false,
     "tidy_revoked_certs": false
```
cipherboy, line 4643:
```suggestion
    "tidy_revoked_certs": false,
```

builtin/logical/pki/path_tidy_test.go
```diff
@@ -1293,6 +1293,115 @@ func waitForAutoTidyToFinish(t *testing.T, client *api.Client) {
...
+	resp, err := CBWrite(b, s, "issue/local-testing", map[string]interface{}{
+		"common_name": "short-lived",
+		"ttl":         "1s",
+	})
+	require.NoError(t, err)
+	shortLivedCert := parseCert(t, resp.Data["certificate"].(string))
+	shortLivedSerial := resp.Data["serial_number"].(string)
+
+	// Write invalid certificate to storage
+	err = s.Put(ctx, &logical.StorageEntry{
```
cipherboy, line 1343: Nice! :-)

builtin/logical/pki/path_tidy_test.go
```diff
@@ -1293,6 +1293,115 @@ func waitForAutoTidyToFinish(t *testing.T, client *api.Client) {
...
+		Key:   "certs/1",
+		Value: []byte(invalidCert),
```
cipherboy, line 1345: While technically this works, I think this is invalid for an incorrect reason (wrong format) -- I think ideally you'd remove the `----BEGIN ... ` and `---END ....` headers and just have the base64 data and call `base64.StdEncoding.DecodeString(...)` on the body, since storage entries are the raw byte value of the DER form of the certificate (not the PEM-encoded "nice-to-read" form).

builtin/logical/pki/path_tidy_test.go
```diff
@@ -1293,6 +1293,115 @@ func waitForAutoTidyToFinish(t *testing.T, client *api.Client) {
...
+
+	// Perform the tidy operation for cert store and invalid certs
+	resp, err = CBWrite(b, s, "tidy", map[string]interface{}{
+		"tidy_invalid_certs": true,
+		"tidy_cert_store":    true,
+		"safety_buffer":      1,
+	})
+	require.NoError(t, err, "tidy operation should complete without errors")
+
+	// Check if tidy operation has completed every 50ms
```
cipherboy, line 1374: I think you could reuse `waitForAutoTidyToFinish(...)` :-)

fatima2003: Hmm, but I don't have a client defined, should I refactor the test to use clusters instead?

cipherboy: Ahhh, I see. In that case, why not just invoke the relevant tidy method directly? Just a thought.

fatima2003: Oh wait, yea it is redundant

builtin/logical/pki/path_tidy_test.go
fatima2003: Good catch :) I missed this even though the error message states this 🙇🏽‍♀️

builtin/logical/pki/path_tidy_test.go
```diff
@@ -1293,6 +1293,91 @@ func waitForAutoTidyToFinish(t *testing.T, client *api.Client) {
...
+	resp, err = CBList(b, s, "certs")
+	require.NoError(t, err)
+	certKeys := resp.Data["keys"].([]string)
+	require.Len(t, certKeys, 4, "expected 4 certificates in the store")
+
+	// Wait for short lived cert to expire and the safety buffer to elapse.
+	time.Sleep(time.Until(shortLivedCert.NotAfter) + 1*time.Second)
+
+	// Perform the tidy operation for cert store and invalid certs
+	resp, err = CBWrite(b, s, "tidy", map[string]interface{}{
```
cipherboy, line 1361: @fatima2003 Sorry, I realized I was being confusing. I meant calling `doTidyCertStore` or `doTidyRevocationStore` with an appropriate `tidyConfig` object. It isn't fully realistic insofar as it isn't an end-to-end test, but for the purposes of what we're doing, I think it is fine!

This way, you don't even need the `time.Sleep` after this, as it is happening in the main thread and not in a background goroutine which may be interrupted.

fatima2003: Ahh okay, I didn't think that was an option. I pushed the update :)

cipherboy (approved):
Looks good, thanks @fatima2003!

cipherboy:
Race detector failures unrelated, merging...

Issue #74, opened 2024-01-27
Fork: Update Vault references within the UI


cipherboy:
The UI utilizes both the Vault name, but also logo and other branding features. We may want a new color scheme, a replacement logo, and name changes in several locations prior to release.

cipherboy:
Synicng some discussion from Matrix earlier:

To remove HashiCorp branding, we ~will need to fork https://github.com/hashicorp/design-system~ have forked https://github.com/openbao/design-system -- this is dep'd from https://github.com/openbao/openbao/blob/main/ui/package.json#L63 -> https://www.npmjs.com/package/@hashicorp/flight-icons -> https://github.com/hashicorp/design-system/tree/main/packages/ember-flight-icons

We may need to republish this to NPM and/or use a [local include](https://stackoverflow.com/questions/15806241/how-to-specify-local-modules-as-npm-package-dependencies) instead if we want to avoid that (probably preferable).

see e.g. packages/flight-icons/svg-original/hashicorp-24.svg or packages/flight-icons/svg/vault-24.svg

rhuddleston:
Random question but since the design-system repo is still MPL 2.0 do we really need to fork/maintain all of that? Can we instead change refernces in the UI code a few places with new icons where they are hashicorp specific or is it more complex than that?

cipherboy:
@rhuddleston IIRC, the issue is the design repo itself has HashiCorp logos &c that are used by OpenBao's UI. So without forking it, our only option would be to ask upstream to make those logos pluggable, which they're unlikely to accept as they have no need for it.

rhuddleston:
I guess was hoping we could change `hashicorp` and `vault-logo` references in ui/lib/core/icon-mappings.js to something else instead of having to fork that project? Though I'm not sure how that mappings file is even used.

Digging a but more it seems we can change ui/app/templates/vault.hbs `<Icon @name="hashicorp"` to change the footer I think

ui/app/templates/vault/cluster/auth.hbs and ui/app/templates/vault/error.hbs both reference `Icon @name="vault"` and could be changed to reference other icons but there are probably more places

cipherboy:
Go for it if you think it's doable! I will confess my frontend experience is quite minimal and limited to carefully supervised attempts. :D I have a little more familiarity with React, but even that isn't extensive enough to feel confident taking this on myself.

cipherboy:
This looks to be complete! Thanks @bozzo and @jenrik and whomever else I might've forgotten to thank!

PR #669, opened 2024-11-02
Draft: UI: replace documentation links

bozzo:
Small part of #74

Replaces the documentation links in the UI by links pointing to openbao.org.


bozzo:
Some links are related to Vault tutorials witch aren't available in https://openbao.org/docs/. Now they lead to a 404 not found error (for example: https://openbao.org/tutorials/auth-methods/multi-factor-authentication), what should I do with theses links? Should I remove them completely?

cipherboy:
@bozzo Yes, we've been removing links to tutorials. The issue is they were not licensed under the MPL to begin with. Like OpenTofu, we'd need to very carefully create our own tutorials from scratch, without relying on the existing ones. That's a lot of work that just hasn't been done yet. :-)

JanMa (approved):
LGTM :heavy_check_mark: Thanks a lot @bozzo

ui/app/templates/components/wizard/replication-setup.hbs
```diff
@@ -11,15 +11,9 @@
       you more interested in speed of access?
     </p>
 
-    <DocLink @path="/vault/tutorials/enterprise/performance-replication">
```
cipherboy, line 14: As an FYI, OpenBao likely will not have either mode as they're not on our roadmap. We do have HA mode though, so in the future, this doc page will likely have to be reworked.

ui/app/templates/components/clients/dashboard.hbs
```diff
@@ -2,7 +2,7 @@
   <p class="has-bottom-margin-xl">
     This dashboard will surface OpenBao's client usage over time. Clients represent a user or service that has authenticated
     to OpenBao. Documentation is available
-    <DocLink @path="/vault/docs/concepts/client-count">here.</DocLink>
+    <DocLink @path="/docs/concepts/client-count">here.</DocLink>
```
cipherboy, line 5: As a FYI @bozzo this page doesn't exist because we removed it in #194 -- this entire section on client counts can be removed from the UI. :-)

cipherboy (approved):
Looks good, thank you @bozzo!

ui/app/templates/components/clients/dashboard.hbs
bozzo: done in #734

PR #678, opened 2024-11-05
Add pagination to tidy 

### LINKED ISSUE #271: Broader usage of Paginated Lists

In #140 and #170, initial support of paginated lists were added, which included some API endpoints. However, some more endpoints could be updated to use these new APIs.

- [ ] ~PKI Tidy operation should use paginated lists everywhere.~ in #678 thanks to @fatima2003!
- [ ] Audit other tidy operations for usage of paginated lists
- [ ] Lease expiry should use paginated lists
- [ ] Token expiry should use paginated lists
- [ ] Secrets and auth engine listing should probably support pagination as well. ~This does not use a `storage.List` call but could still be made transactional.~ Transactional added in #622, but should still manually support pagination even though it won't be any more efficient.
- [ ] Policy listing should also use this, even though it uses a recursive storage view.
- [ ] Additional calls to `storage.List` should probably be audited as well.

Some places will continue to rely on `storage.List` as they have a low volume of responses, e.g., issuer counts should be in the 10s at most. 

Some additional use cases are available at the HashiCorp Vault issue: https://github.com/hashicorp/vault/issues/10887 

**Additional context**

This will help with memory utilization in more high-volume areas of the code. Would be a good GA milestone candidate.

fatima2003:
This PR modifies the tidy process to use pagination, allowing certificates to be processed in increments rather than loading the entire set into memory at once. This would be particularly helpful when the cert store contains millions of certs. The default page size is 1000 certificates but custom page sizes can be specified via a flag `page_size`. Minimum page size is 5 and to default to previous behavior, i.e., skip out on pagination, set `page_size=-1` or any negative int.

Partly resolves #271

## Target Release

1.14.7

---
### TO DO

- [X] Add pagination to tidy cert store.
- [X] Add pagination to all tidy functions.
- [X] Add flag to allow user to define page size with minimum size 5 and defaults to 1000.
- [X] Create pagination test.
- [X] Add docs.

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -4140,7 +4140,7 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 
 	// Sign the intermediate CSR using /pki
 	secret, err = client.Logical().Write("pki/root/sign-intermediate", map[string]interface{}{
-		"permitted_dns_domains": ".myvault.com",
+		"permitted_dns_domains": ".example.com",
 		"csr":                   intermediateCSR,
 		"ttl":                   "10s",
 	})
@@ -4188,17 +4188,11 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 	time.Sleep(3 * time.Second)
 
 	// Issue a tidy on /pki
-	_, err = client.Logical().Write("pki/tidy", map[string]interface{}{
+	waitForManualTidy(t, client, map[string]interface{}{
 		"tidy_cert_store":    true,
 		"tidy_revoked_certs": true,
 		"safety_buffer":      "1s",
 	})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	// Sleep a bit to make sure we're past the safety buffer
-	time.Sleep(2 * time.Second)
 
 	// Get CRL and ensure the tidied cert is still in the list after the tidy
 	// operation since it's not past the NotAfter (ttl) value yet.
@@ -4214,21 +4208,27 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 		t.Fatalf("expected: %v, got: %v", intermediateCertSerial, sn)
 	}
 
+	// Check the revoked-count metrics
+	mostRecentInterval2 := inmemSink.Data()[len(inmemSink.Data())-1]
+	expectedGauges := map[string]float32{
+		"secrets.pki.tidy.revoked_cert_total_entries":           1,
+		"secrets.pki.tidy.revoked_cert_total_entries_remaining": 1,
+	}
+	for gauge, value := range expectedGauges {
+		if metric, ok := mostRecentInterval2.Gauges[gauge]; !ok || metric.Value != value {
+			t.Fatalf("Expected gauge %s to have value %f, but got %f", gauge, value, metric.Value)
+		}
+	}
+
 	// Wait for cert to expire
 	time.Sleep(10 * time.Second)
 
 	// Issue a tidy on /pki
-	_, err = client.Logical().Write("pki/tidy", map[string]interface{}{
+	waitForManualTidy(t, client, map[string]interface{}{
 		"tidy_cert_store":    true,
 		"tidy_revoked_certs": true,
 		"safety_buffer":      "1s",
 	})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	// Sleep a bit to make sure we're past the safety buffer
-	time.Sleep(2 * time.Second)
 
 	// Issue a tidy-status on /pki
 	{
@@ -4247,7 +4247,7 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 			"tidy_expired_issuers":                  false,
 			"tidy_move_legacy_ca_bundle":            false,
 			"pause_duration":                        "0s",
-			"page_size":                             json.Number("50"),
+			"page_size":                             json.Number("1000"),
 			"state":                                 "Finished",
 			"error":                                 nil,
 			"time_started":                          nil,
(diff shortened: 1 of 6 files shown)


builtin/logical/pki/path_tidy.go
```diff
@@ -905,122 +938,136 @@ func (b *backend) startTidyOperation(req *logical.Request, config *tidyConfig) {
...
+			revokedSafetyBuffer = *config.RevokedSafetyBuffer
 		}
 
-		if certEntry == nil {
-			logger.Warn("certificate entry is nil; tidying up since it is no longer useful for any server operations", "serial", serial)
-			if err := req.Storage.Delete(ctx, "certs/"+serial); err != nil {
-				return 0, fmt.Errorf("error deleting nil entry with serial %s: %w", serial, err)
+		for i, serial := range serials {
+			// Set the `after` variable for the next ListPage call
+			after = serial
```
cipherboy, line 967: Out of curiosity, why not set `after = serials[len(serials)-1]` once at the top? 

I'm debating, I think this might be more forgiving if we `break` for some reason out of this inner loop, but I'm not honestly sure why we would. We'd be more inclined to `return` and exit the function entirely, IMO?

builtin/logical/pki/path_tidy.go
```diff
@@ -905,122 +938,136 @@ func (b *backend) startTidyOperation(req *logical.Request, config *tidyConfig) {
...
+
+			b.tidyStatusMessage(fmt.Sprintf("Tidying certificate store: checking entry %d of %d on current page; total certs checked: %d", i, config.PageSize, int(serialCount)+i))
```
cipherboy, line 969: Nice! I like this status update. Strictly to be correct, I think it needs to be:

```suggestion
			b.tidyStatusMessage(fmt.Sprintf("Tidying certificate store: checking entry %d of %d on current page; total certs checked: %d", i, len(serials), int(serialCount)-len(serials)+i))
```

Our last page will likely be a partial page and earlier on new line 958 we already adjusted `serialCount` to include `len(serials)`. We might be better off moving `serialCount`'s update to the end after this loop is done and then we can call this `serialCount+i` perhaps?

We might also want to keep `len(serials)` in a local variable as we reference it a lot.

cipherboy: Same I think also applies to the current entry below.

builtin/logical/pki/path_tidy.go
```diff
@@ -1038,152 +1085,163 @@ func (b *backend) doTidyRevocationStore(ctx context.Context, req *logical.Reques
...
+		if len(revokedSerials) == 0 {
+			break
 		}
 
-		revokedEntry, err := req.Storage.Get(ctx, "revoked/"+serial)
-		if err != nil {
-			return false, fmt.Errorf("unable to fetch revoked cert with serial %q: %w", serial, err)
-		}
+		revokedSerialsCount += uint(len(revokedSerials))
+		metrics.SetGauge([]string{"secrets", "pki", "tidy", "revoked_cert_total_entries"}, float32(revokedSerialsCount))
```
cipherboy, line 1106: I think you did this correctly for `doTidyCertStore` but I think this should only be done once, at the end of the function.

builtin/logical/pki/path_tidy.go
```diff
@@ -905,122 +938,136 @@ func (b *backend) startTidyOperation(req *logical.Request, config *tidyConfig) {
...
-		if certEntry == nil {
-			logger.Warn("certificate entry is nil; tidying up since it is no longer useful for any server operations", "serial", serial)
-			if err := req.Storage.Delete(ctx, "certs/"+serial); err != nil {
-				return 0, fmt.Errorf("error deleting nil entry with serial %s: %w", serial, err)
+		for i, serial := range serials {
+			// Set the `after` variable for the next ListPage call
+			after = serial
+
+			b.tidyStatusMessage(fmt.Sprintf("Tidying certificate store: checking entry %d of %d on current page; total certs checked: %d", i, config.PageSize, int(serialCount)+i))
+			metrics.SetGauge([]string{"secrets", "pki", "tidy", "cert_store_current_entry"}, float32(int(serialCount)+i-1))
```
cipherboy, line 970: I'm also confused why you have a `-1` on this...? I think it should just be the same value as in the status line above.

builtin/logical/pki/path_tidy.go
```diff
@@ -1396,35 +1454,49 @@ func (b *backend) doTidyAcme(ctx context.Context, req *logical.Request, logger h
...
+		thumbprints, err := sc.Storage.ListPage(ctx, acmeThumbprintPrefix, after, config.PageSize)
 		if err != nil {
-			logger.Warn("error tidying account %v: %v", thumbprint, err.Error())
+			return err
 		}
 
-		// Check for cancel before continuing.
-		if atomic.CompareAndSwapUint32(b.tidyCancelCAS, 1, 0) {
-			return tidyCancelledError
+		// If no thumbprints are returned, we've reached the end of the list
```
cipherboy, line 1466: While you're here, do you mind adding status messages for this and EAB as well?

builtin/logical/pki/path_tidy.go
```diff
@@ -1396,35 +1454,49 @@ func (b *backend) doTidyAcme(ctx context.Context, req *logical.Request, logger h
...
+			}
+		}
 	}
 
+	b.tidyStatusLock.Lock()
+	b.tidyStatus.acmeAccountsCount = uint(thumbprintsCount)
+	b.tidyStatusLock.Unlock()
+
 	// Clean up any unused EAB
 	eabIds, err := b.acmeState.ListEabIds(sc)
```
cipherboy, line 1501: You can do pagination with `b.acmeState.ListEabIdsPage` btw :-)

cipherboy (commented):
Some initial feedback! Really like what this will do to help operators. :-)

builtin/logical/pki/path_tidy.go
fatima2003: `TestBackend_RevokePlusTidy_Intermediate` was failing and after I looked at it and the value it was expecting for `cert_store_current_entry`, I thought it was erring because counting starts at 0 not 1 (which is true but this is not the correct solution if they are many entries), hence the `-1`. Moving the `serialCount` to below the for loop fixed this problem.

builtin/logical/pki/path_tidy.go
```diff
@@ -116,6 +122,7 @@ var defaultTidyConfig = tidyConfig{
 	IssuerSafetyBuffer:      365 * 24 * time.Hour,
 	AcmeAccountSafetyBuffer: 30 * 24 * time.Hour,
 	PauseDuration:           0 * time.Second,
+	PageSize:                50,
```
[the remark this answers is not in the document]
fatima2003: this sounds reasonable, I've updated it :)

builtin/logical/pki/path_tidy.go
fatima2003: Hmm, yes I agree, this would remove the unnecessary step of assigning a new value to `after` on each iteration of the loop

builtin/logical/pki/path_tidy.go
fatima2003: Good catch, I've updated it! :)

builtin/logical/pki/path_tidy.go
fatima2003: Alrighty, done! I'll push the changes in a minute after I double check all the changes :)

builtin/logical/pki/path_tidy.go
cipherboy: @fatima2003 Ahh right, because it isn't a count, it is the index. My bad!

builtin/logical/pki/backend_test.go
```diff
@@ -114,7 +114,7 @@ func TestPKI_RequireCN(t *testing.T) {
 	b, s := CreateBackendWithStorage(t)
 
 	resp, err := CBWrite(b, s, "root/generate/internal", map[string]interface{}{
-		"common_name": "myvault.com",
+		"common_name": "mybao.com",
```
cipherboy, line 117: Hmm, if we're going to go to a different domain, let's do it in a separate PR if that's alright.

I'd probably also use `example.com` as the RFC 2606 specifies :-)

builtin/logical/pki/backend_test.go
```diff
@@ -4290,7 +4291,7 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 			"secrets.pki.tidy.cert_store_current_entry":                         0,
 			"secrets.pki.tidy.cert_store_total_entries":                         1,
 			"secrets.pki.tidy.revoked_cert_current_entry":                       0,
-			"secrets.pki.tidy.revoked_cert_total_entries":                       1,
```
fatima2003, line 4293: @cipherboy since they are 2 tidy calls, the first tidy call cleans up the revoked cert and so the metrics will show `revoked_cert_total_entries = 0` in the end, if I add a check between the tidy calls on line 4191 and 4220 both of the following are true: 
```
"secrets.pki.tidy.revoked_cert_total_entries":           1,
"secrets.pki.tidy.revoked_cert_total_entries_remaining": 1,
```

fatima2003: I will add this to the test before I rebase

builtin/logical/pki/backend_test.go
fatima2003: Alright, lem fix that and I'll open up a PR shortly! :)

builtin/logical/pki/path_tidy.go
fatima2003: The way it was written was a tad bit confusing ><

fatima2003:
Hmm the test errs on the server but they all pass locally: 

```
Running tool: /usr/local/go/bin/go test -timeout 30s -run ^TestBackend_Tidy_MultipleExpiredCerts$ github.com/openbao/openbao/builtin/logical/pki

Running tool: /usr/local/go/bin/go test -timeout 30s -run ^TestBackend_RevokePlusTidy_MultipleCerts$ github.com/openbao/openbao/builtin/logical/pki

Running tool: /usr/local/go/bin/go test -timeout 30s -run ^TestBackend_RevokePlusTidy_Intermediate$ github.com/openbao/openbao/builtin/logical/pki

ok  	github.com/openbao/openbao/builtin/logical/pki	10.023s
ok  	github.com/openbao/openbao/builtin/logical/pki	14.937s
ok  	github.com/openbao/openbao/builtin/logical/pki	23.277s
```

`Error: -18T18:39:57.872Z [ERROR] TestBackend_RevokePlusTidy_MultipleCerts.core0.secrets.pki.pki_d944072a: failed loading existing ACME challenge validations:: err="failed loading list of validations from disk: context canceled"`

cipherboy (commented):
I ended up making the following set of changes locally, which I think helped with the flaky test in `TestBackend_RevokePlusTidy_Intermediate` and `TestBackend_RevokePlusTidy_MultipleCerts`... 

I think the changes to 600s is unnecessary, but switching to `time.Until` and `waitForManualTidy` are probably good to do. I wonder if it is worth refactoring the `doTidy` to have a named helper to do the tidy synchronously rather than asynchronously... 

<details>

<summary> patch </summary>

```diff 
diff --git a/builtin/logical/pki/backend_test.go b/builtin/logical/pki/backend_test.go
index b9af003fa9..25dbdc65ea 100644
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -4188,17 +4188,11 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 	time.Sleep(3 * time.Second)
 
 	// Issue a tidy on /pki
-	_, err = client.Logical().Write("pki/tidy", map[string]interface{}{
+	waitForManualTidy(t, client, map[string]interface{}{
 		"tidy_cert_store":    true,
 		"tidy_revoked_certs": true,
 		"safety_buffer":      "1s",
 	})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	// Sleep a bit to make sure we're past the safety buffer
-	time.Sleep(2 * time.Second)
 
 	// Get CRL and ensure the tidied cert is still in the list after the tidy
 	// operation since it's not past the NotAfter (ttl) value yet.
@@ -4230,17 +4224,11 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 	time.Sleep(10 * time.Second)
 
 	// Issue a tidy on /pki
-	_, err = client.Logical().Write("pki/tidy", map[string]interface{}{
+	waitForManualTidy(t, client, map[string]interface{}{
 		"tidy_cert_store":    true,
 		"tidy_revoked_certs": true,
 		"safety_buffer":      "1s",
 	})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	// Sleep a bit to make sure we're past the safety buffer
-	time.Sleep(2 * time.Second)
 
 	// Issue a tidy-status on /pki
 	{
@@ -4303,7 +4291,7 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 			"secrets.pki.tidy.cert_store_current_entry":                         0,
 			"secrets.pki.tidy.cert_store_total_entries":                         1,
 			"secrets.pki.tidy.revoked_cert_current_entry":                       0,
-			"secrets.pki.tidy.revoked_cert_total_entries":                       0,
+			"secrets.pki.tidy.revoked_cert_total_entries":                       1,
 			"secrets.pki.tidy.start_time_epoch":                                 0,
 			"secrets.pki." + backendUUID + ".total_certificates_stored":         0,
 			"secrets.pki." + backendUUID + ".total_revoked_certificates_stored": 0,
@@ -4438,21 +4426,25 @@ func TestBackend_RevokePlusTidy_MultipleCerts(t *testing.T) {
 	}
 
 	// Issue 3 short-lived certificates
+	lastTTL := time.Now()
 	for i := 1; i <= 3; i++ {
-		_, err := client.Logical().Write("pki/issue/test", map[string]interface{}{
+		resp, err := client.Logical().Write("pki/issue/test", map[string]interface{}{
 			"common_name": "short-lived.myvault.com",
 			"ttl":         "1s",
 		})
 		if err != nil {
 			t.Fatal(err)
 		}
+
+		leafCert := parseCert(t, resp.Data["certificate"].(string))
+		lastTTL = leafCert.NotAfter
 	}
 
 	// Issue 2 long-lived certificates
 	for i := 1; i <= 2; i++ {
 		_, err := client.Logical().Write("pki/issue/test", map[string]interface{}{
 			"common_name": "long-lived.myvault.com",
-			"ttl":         "60s",
+			"ttl":         "600s",
 		})
 		if err != nil {
 			t.Fatal(err)
@@ -4460,16 +4452,13 @@ func TestBackend_RevokePlusTidy_MultipleCerts(t *testing.T) {
 	}
 
 	// Wait for the short-lived certificates to expire
-	time.Sleep(4 * time.Second)
+	time.Sleep(time.Until(lastTTL) + 2*time.Second)
 
 	// Tidy the expired certificates
-	_, err = client.Logical().Write("pki/tidy", map[string]interface{}{
+	waitForManualTidy(t, client, map[string]interface{}{
 		"tidy_cert_store": true,
 		"safety_buffer":   "1s",
 	})
-	if err != nil {
-		t.Fatal(err)
-	}
 
 	// tidy metrics check
 	mostRecentInterval := inmemSink.Data()[len(inmemSink.Data())-1]
@@ -4484,18 +4473,19 @@ func TestBackend_RevokePlusTidy_MultipleCerts(t *testing.T) {
 	}
 
 	// Issue two certificates, a short-lived and a long-lived, then revoke them.
-	cert1, err := client.Logical().Write("pki/issue/test", map[string]interface{}{
+	cert1Resp, err := client.Logical().Write("pki/issue/test", map[string]interface{}{
 		"common_name": "short-lived.myvault.com",
 		"ttl":         "5s", // Short-lived certificate
 	})
 	if err != nil {
 		t.Fatal(err)
 	}
-	certSerial1 := cert1.Data["serial_number"].(string)
+	certSerial1 := cert1Resp.Data["serial_number"].(string)
+	cert1 := parseCert(t, cert1Resp.Data["certificate"].(string))
 
 	cert2, err := client.Logical().Write("pki/issue/test", map[string]interface{}{
 		"common_name": "long-lived.myvault.com",
-		"ttl":         "60s", // Long-lived certificate
+		"ttl":         "600s", // Long-lived certificate
 	})
 	if err != nil {
 		t.Fatal(err)
@@ -4516,19 +4506,13 @@ func TestBackend_RevokePlusTidy_MultipleCerts(t *testing.T) {
 	}
 
 	// Wait for the short-lived certificate to expire
-	time.Sleep(6 * time.Second)
+	time.Sleep(time.Until(cert1.NotAfter) + 2*time.Second)
 
 	// Tidy the revoked certificates
-	_, err = client.Logical().Write("pki/tidy", map[string]interface{}{
+	waitForManualTidy(t, client, map[string]interface{}{
 		"tidy_revoked_certs":    true,
 		"revoked_safety_buffer": "1s",
 	})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	// Wait for safety_buffer
-	time.Sleep(2 * time.Second)
 
 	// Verify that the revoked short-lived certificate has been tidied
 	crl := getParsedCrl(t, client, "pki")
diff --git a/builtin/logical/pki/path_tidy.go b/builtin/logical/pki/path_tidy.go
index d87b9881ef..8241e28e4a 100644
--- a/builtin/logical/pki/path_tidy.go
+++ b/builtin/logical/pki/path_tidy.go
@@ -8,7 +8,6 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
-	"math"
 	"net/http"
 	"sync/atomic"
 	"time"
@@ -778,12 +777,7 @@ func (b *backend) pathTidyWrite(ctx context.Context, req *logical.Request, d *fr
 		}
 	}
 
-	// If pageSize is negative, bypass pagination
-	if pageSize <= -1 {
-		pageSize = math.MaxInt // Use the maximum value as a practical way to represent "all"
-	}
-
-	if pageSize >= 0 && pageSize < 5 {
+	if pageSize > 0 && pageSize < 5 {
 		return logical.ErrorResponse("page_size must be greater than five"), nil
 	}
 
@@ -1261,6 +1255,9 @@ func (b *backend) doTidyRevocationStore(ctx context.Context, req *logical.Reques
 		totalRevokedSerialCount += lenSerials
 	}
 
+	totalRevokedSerialCount += int(revokedDeleted)
+	revokedDeletedCount += int(revokedDeleted)
+
 	b.tidyStatusLock.RLock()
 	metrics.SetGauge([]string{"secrets", "pki", "tidy", "revoked_cert_total_entries"}, float32(totalRevokedSerialCount))
 	metrics.SetGauge([]string{"secrets", "pki", "tidy", "revoked_cert_total_entries_remaining"}, float32(totalRevokedSerialCount-revokedDeletedCount))
diff --git a/builtin/logical/pki/path_tidy_test.go b/builtin/logical/pki/path_tidy_test.go
index 3a643fa871..049b663e5f 100644
--- a/builtin/logical/pki/path_tidy_test.go
+++ b/builtin/logical/pki/path_tidy_test.go
@@ -1294,6 +1294,46 @@ func waitForAutoTidyToFinish(t *testing.T, client *api.Client) {
 	}
 }
 
+func waitForManualTidy(t *testing.T, client *api.Client, tidyConfig map[string]interface{}) {
+	status, err := client.Logical().Read("pki/tidy-status")
+	require.NoError(t, err, "got error reading initial tidy status")
+
+	t.Logf("initial status resp: %v", status)
+
+	_, err = client.Logical().Write("pki/tidy", tidyConfig)
+	require.NoError(t, err, "got error starting tidy")
+
+	timeoutChan := time.After(120 * time.Second)
+
+	for {
+		select {
+		case <-timeoutChan:
+			t.Fatalf("expected manual tidy to run before timeout")
+		default:
+			time.Sleep(50 * time.Millisecond)
+
+			newStatus, err := client.Logical().Read("pki/tidy-status")
+			require.NoError(t, err, "got error reading subsequent tidy status")
+
+			if newStatus.Data["state"].(string) == "Finished" {
+				thisStart, err := time.Parse(time.RFC3339, newStatus.Data["time_started"].(string))
+				require.NoError(t, err, "failed to parse time")
+				lastStartRaw, ok := status.Data["time_started"]
+				if !ok || lastStartRaw == nil {
+					return
+				}
+
+				lastStart, err := time.Parse(time.RFC3339, lastStartRaw.(string))
+				require.NoError(t, err, "failed to parse time")
+
+				if thisStart.After(lastStart) {
+					return
+				}
+			}
+		}
+	}
+}
+
 func TestTidyWithInvalidCertInStore(t *testing.T) {
 	t.Parallel()
 	b, s := CreateBackendWithStorage(t)
```

</details>

builtin/logical/pki/path_tidy.go
```diff
@@ -749,6 +778,15 @@ func (b *backend) pathTidyWrite(ctx context.Context, req *logical.Request, d *fr
 		}
 	}
 
+	// If pageSize is negative, bypass pagination
+	if pageSize <= -1 {
+		pageSize = math.MaxInt // Use the maximum value as a practical way to represent "all"
```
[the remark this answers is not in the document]
fatima2003: Alrighty, thank you, I should've double checked with the docs! :)

builtin/logical/pki/path_tidy.go
```diff
@@ -1038,152 +1096,174 @@ func (b *backend) doTidyRevocationStore(ctx context.Context, req *logical.Reques
...
 			}
 		}
+
+		// Update the cumulative count after processing the page
+		totalRevokedSerialCount += lenSerials
 	}
 
 	b.tidyStatusLock.RLock()
-	metrics.SetGauge([]string{"secrets", "pki", "tidy", "revoked_cert_total_entries_remaining"}, float32(uint(revokedSerialsCount)-b.tidyStatus.revokedCertDeletedCount))
+	metrics.SetGauge([]string{"secrets", "pki", "tidy", "revoked_cert_total_entries"}, float32(totalRevokedSerialCount))
```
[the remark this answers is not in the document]
fatima2003: This makes so much sense, the reason why revoked certs where "magically" decrementing the count when `tidyCertStore` was being called first 😭 That should've been the red flag for me, this solves that as well, thank you!

cipherboy (approved):
Looks great, thanks for the patience on this one @fatima2003!

PR #680, opened 2024-11-08
Return certificate details in pki/certs/detailed endpoint

### LINKED ISSUE #675: Return some certificate details in pki/certs endpoint
As requested by @anastas001 and @celesteking in Vault issues [27249](https://github.com/hashicorp/vault/issues/27249) and [24997](https://github.com/hashicorp/vault/issues/24997) respectively, when listing certificates in pki, only the serial numbers are returned. This makes it hard to find specific certificates so displaying a few certificates details like CN, SAN, notAfter and encryption type would be helpful when the `- detailed` flag is provided.

---
I will open up a PR for this soon :)

fatima2003:
This PR introduces the pki/certs/detailed endpoint to provide additional details for each certificate under pki/certs. With this new endpoint, users can view information such as:
- common_name
- serial_number
- expiration
- issuer

This is not the final list of details and I will be adding more details like DNS names, etc. 

This feature allows for more comprehensive inspection of stored certificates, enabling users to retrieve specific certificate attributes without additional GET requests for each individual certificate.

---
### TO DO
- [x] Create new endpoint `pki/certs/detailed`
- [ ] Choose which details to return
- [ ] Add documentation

Resolves #675 [Vault 27249](https://github.com/hashicorp/vault/issues/27249) [Vault 24997](https://github.com/hashicorp/vault/issues/24997) 

## Target Release

1.14.7


fatima2003:
@cipherboy, I created a new endpoint `pki/certs/detailed` but since this uses format.go 


the details are not printed to CLI unless `-detailed` is provided.

fatima2003:
Hi @cipherboy, I created the new endpoint `pki/cert/detailed`, however since it uses `ListResponseWithInfo` which is formatted in [format.go](https://github.com/openbao/openbao/blob/main/command/format.go) based off whether the detailed flag is provided, the CLI does not print the details unless `-detailed` is provided in `pki/cert/detailed` endpoint so the command looks like this: 
```
bao list -detailed pki/certs/detailed
```

I tried to enable the flag in the code itself but was unsuccessful. Any pointers will be very helpful :)

builtin/logical/pki/path_fetch.go
```diff
@@ -261,6 +262,106 @@ func (b *backend) pathFetchCertList(ctx context.Context, req *logical.Request, d
...
+		dnsNames := certData.DNSNames
+		if len(dnsNames) > 5 {
+			dnsNames = dnsNames[:5]
+		}
+
+		responseInfo[string(entries[i])] = map[string]interface{}{
+			"common_name": certData.Subject.CommonName,
+			"issuer":      certData.Issuer.String(),
+			"key_type":    certData.PublicKeyAlgorithm.String(),
+			"expiration":  certData.NotAfter,
```
cipherboy, line 356: nit: I'd maybe call this `not_after` for consistency with `not_before`.

builtin/logical/pki/path_fetch.go
```diff
@@ -261,6 +262,106 @@ func (b *backend) pathFetchCertList(ctx context.Context, req *logical.Request, d
...
+		// limit DNS names to 5
+		dnsNames := certData.DNSNames
+		if len(dnsNames) > 5 {
+			dnsNames = dnsNames[:5]
+		}
+
+		responseInfo[string(entries[i])] = map[string]interface{}{
+			"common_name": certData.Subject.CommonName,
+			"issuer":      certData.Issuer.String(),
+			"key_type":    certData.PublicKeyAlgorithm.String(),
```
cipherboy, line 355: How about a `key_bits` parameter as well? I think this is a touch harder, because you'll have to go from `PublicKey` and infer the bits from there (excluding it for `ed25519` certs).

Thoughts?

builtin/logical/pki/path_fetch.go
```diff
@@ -261,6 +262,106 @@ func (b *backend) pathFetchCertList(ctx context.Context, req *logical.Request, d
...
+		}
+
+		// limit DNS names to 5
+		dnsNames := certData.DNSNames
+		if len(dnsNames) > 5 {
+			dnsNames = dnsNames[:5]
+		}
+
+		responseInfo[string(entries[i])] = map[string]interface{}{
+			"common_name": certData.Subject.CommonName,
```
cipherboy, line 353: Trying to decide if I like just `common_name` (which is, I guess, most common), or if the full `subject` would be preferable?

What's your take?

builtin/logical/pki/path_fetch.go
```diff
@@ -261,6 +262,106 @@ func (b *backend) pathFetchCertList(ctx context.Context, req *logical.Request, d
...
+				},
+			},
+		},
+
+		HelpSynopsis:    pathFetchHelpSyn,
+		HelpDescription: pathFetchHelpDesc,
+	}
+}
+
+func (b *backend) pathFetchCertListDetailed(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
```
cipherboy, line 313: I like this approach, very clean!

I could see an argument that we might reuse `pathFetchCertList`, but I think this is far more complicated than that one and I like keeping that simpler so I don't think I'd recommend it in this case.

website/content/api-docs/secret/pki.mdx
```diff
@@ -1908,6 +1908,82 @@ $ curl \
 }
 ```
 
+### List certificates with details
+
+This endpoint returns a list of the current certificates by serial number and provides details such as CN, issuer, key_type, expiration, not_before and upto 5 DNS names.
```
cipherboy, line 1913: nit:

```suggestion
This endpoint returns a list of the current certificates by serial number and provides details such as CN, issuer, key_type, expiration, not_before and up to 5 DNS names.
```

What do you think of collapsing this documentation into the `LIST /pki/certs` version instead? Some of the info is redundant and you could say detailed information is only present on this endpoint instead.

I think you can see it with things like import: https://openbao.org/api-docs/secret/pki/#import-ca-certificates-and-keys 

Otherwise, the page gets rather long... 

Thoughts?

builtin/logical/pki/path_fetch.go
```diff
@@ -261,6 +262,106 @@ func (b *backend) pathFetchCertList(ctx context.Context, req *logical.Request, d
...
+			dnsNames = dnsNames[:5]
+		}
+
+		responseInfo[string(entries[i])] = map[string]interface{}{
+			"common_name": certData.Subject.CommonName,
+			"issuer":      certData.Issuer.String(),
+			"key_type":    certData.PublicKeyAlgorithm.String(),
+			"expiration":  certData.NotAfter,
+			"not_before":  certData.NotBefore,
+			"DNSNames":    dnsNames,
```
cipherboy, line 358: nit for consistency: 

```suggestion
			"dns_names":    dnsNames,
```

builtin/logical/pki/path_fetch.go
```diff
@@ -261,6 +262,106 @@ func (b *backend) pathFetchCertList(ctx context.Context, req *logical.Request, d
...
+		return nil, err
+	}
+	for i := range entries {
+		// Fetch the full certificate entry by key
+		entry, err := req.Storage.Get(ctx, "certs/"+entries[i])
+		if err != nil {
+			return nil, err
+		}
+		if entry == nil {
+			return logical.ErrorResponse(fmt.Sprintf("failed to retrieve entry for %s", entries[i])), nil
```
cipherboy, line 334: You might consider using a [read-only transaction](https://github.com/openbao/openbao/blob/main/sdk/logical/storage_transactions.go#L16-L19) here so that you never have entries which show up in the list but were since deleted by tidy before you could read them.

I'd take a look at `builtin/logical/pki/path_root.go` to see how we optionally use (writable) transactions and follow a similar pattern here. :-)

cipherboy (commented):
Very clean! I like this @fatima2003! Mostly just a few nits here and there.

cipherboy:
> I also tried to implement detailed cert listing so that bao list -detailed pki/certs/ would not process all the details unless the detailed flag is provided but was also unsuccessful. If I can do this then we can potentially optimize a lot of other places that use detailed lists (like list issuers) so they don't fetch and parse all details unless explicitly requested.

Hmm, the issue with this approach is that if you add a new flag, `detailed=true`, to the LIST endpoints, an admin who wouldn't want users listing detailed information (because it is too resource intensive) would have to modify their ACL policies to explicitly add a `denied_parameters=detailed` line. This isn't... obvious for most administrators.

So I'm kinda preferential to having detailed info be on separate API endpoints so that users don't get access to them by default unless an operator has set up wildcards.

My 2c.!

builtin/logical/pki/path_fetch.go
fatima2003: Ooo this is a good one, I will add checks for the key types rsa and ec; ed25519 will be fixed to 256 bits. 

I didn't add checks for other key types since:
https://github.com/openbao/openbao/blob/acbeddafc5d6721479f3a705783511b995cae3db/builtin/logical/pki/path_roles.go#L235-L241

builtin/logical/pki/path_fetch.go
fatima2003: Hmm the full [subject](https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/crypto/x509/pkix/pkix.go;l=123-139) does have a lot of details which includes String lists of Country, Organization, OrganizationalUnit, Locality, Province, StreetAddress, PostalCode &c.  In the tests I wrote, the only subject details returned was the CN, I'm not quite sure how often these are filled out and how long the lists typically are.

Maybe restricting each list to <= 5 names if they exist, OR to return just the first 5 subject details in cases where more than 5 are present, OR both. 

Let me know your thoughts!

builtin/logical/pki/path_fetch.go
fatima2003: Yep, it would add quite a bit of overhead :)

builtin/logical/pki/path_fetch.go
fatima2003: Alright, I tried this :)

builtin/logical/pki/path_fetch.go
cipherboy: Yeah, that is true that Subject can grow long. A lot of organizations like setting it, though many components will be the same (except maybe O/OU)... Let's stick with `common_name` for now and someone can always file a RFE later and we can add additional fields then. :-)

builtin/logical/pki/path_fetch.go
```diff
@@ -261,6 +265,136 @@ func (b *backend) pathFetchCertList(ctx context.Context, req *logical.Request, d
...
+		case ed25519.PublicKey:
+			keyBits = 256 // Fixed size for Ed25519
+		default:
+			keyBits = 0 // Unknown key type
+		}
+
+		responseInfo[string(entries[i])] = map[string]interface{}{
+			"common_name": certData.Subject.CommonName,
+			"issuer":      certData.Issuer.String(),
+			"key_type":    certData.PublicKeyAlgorithm.String(),
```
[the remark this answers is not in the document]
fatima2003: Alrighty, I'm on it!  I'm almost done writing a test for the metrics for the tidy pagination PR, I clean this up right after :)

website/content/api-docs/secret/pki.mdx
```diff
@@ -1877,6 +1880,7 @@ of third-party verification of CA integrity.
 | Method | Path         |
 | :----- | :----------- |
 | `LIST` | `/pki/certs` |
+| `LIST` | `/pki/certs/detailed?detailed` |
```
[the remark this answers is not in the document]
fatima2003: I'll fix this as well and rebase :)

builtin/logical/pki/path_fetch.go
fatima2003: Alrighty, it is a simple add so if anyone does request it, we can always add it :)

fatima2003:
Thank you for the review @cipherboy!

cipherboy:
Perfect, thanks for the updates, @fatima2003!

PR #691, opened 2024-11-11
Bump github.com/golang-jwt/jwt/v4

tsaarni:
This PR resolves CVE-2024-51744 in github.com/golang-jwt/jwt/v4. Although this vulnerability does not directly affect OpenBao's security, it is flagged in vulnerability scans. Upgrading the dependency will resolve this issue in future scans.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,7 +46,7 @@ require (
 	github.com/go-sql-driver/mysql v1.6.0
 	github.com/go-test/deep v1.1.0
 	github.com/gocql/gocql v1.0.0
-	github.com/golang-jwt/jwt/v4 v4.5.0
+	github.com/golang-jwt/jwt/v4 v4.5.1
 	github.com/golang/protobuf v1.5.4
 	github.com/golangci/revgrep v0.0.0-20220804021717-745bb2f7c2e6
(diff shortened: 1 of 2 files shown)


tsaarni:
I think no changelog entry is necessary for this PR.

cipherboy (approved):
Thanks @tsaarni!

cipherboy:
I do have to say, maybe I'm misattributing it, but while race testing passed here, I swear this change has made test 11 fail more often with the message from https://github.com/openbao/openbao/pull/693#issuecomment-2471330331 

I wonder if there's something we can do to make that more resilient outside of using a fake clock? 

(cc @JanMa / @tsaarni)

tsaarni:
The strangest thing to me is that `github.com/golang-jwt/jwt/v4` does not seem to be even used in test case `builtin/credential/jwt TestLogin_Leeways` which is failing. I could find its use only in `vault/login_mfa.go`.  

Within `builtin/credential/jwt/*` I see it uses `github.com/hashicorp/cap/` which on depends on `github.com/coreos/go-oidc/v3` and `github.com/go-jose/go-jose/v3` which does the actual signature verification.

PR #692, opened 2024-11-11
Bump golang.org/x/net

tsaarni:
This PR resolves CVE-2023-45288 in golang.org/x/net versions used in `api` and `sdk` submodules. The main module had already been updated to a non-vulnerable version previously. This PR now brings all modules to the latest version level.

Although this vulnerability does not directly affect security, it is flagged in vulnerability scans. Upgrading the dependency will resolve this issue in future scans.


cipherboy:
@tsaarni I think you need to run `make tidy-all` on this.

tsaarni:
Thanks @cipherboy! I missed that tidy was needed for the sub-sub-modules as well.

cipherboy:
@tsaarni Hmm  looks like Kubernetes might be relying on the deprecated behavior. Do you want to take a look or should I try to adjust the tests? Do you have allow maintainer to push to branch turned on?

tsaarni:
@cipherboy I'll take a look myself!

By the way, I don't see option "Maintainers are allowed to edit this pull request".  I never paid attention when it is shown, but I'm wondering if it is because the repository I'm pushing from belongs to an organization instead of being personal repo 🤔

cipherboy:
@tsaarni Sorry about the delay. Its a little option on the bottom right of the side bar (from my PR):

image

On one of Fatima's PRs, I see:

image

But on yours, I don't see that:

image

At any rate, let me know if you have questions!

tsaarni:
I'm re-running the test by pushing the same code again, as I couldn't reproduce the failure locally. I suspect it might be a race so I attempted to execute the test case with a high `-count` etc but didn't find any problem...

The error previously was:

```
=== FAIL: builtin/credential/kubernetes TestLoginProjectedToken/projected-token-expired (0.00s)
2024-11-11T15:25:24.761Z [DEBUG] failed to read local service account token, will use client token: error="open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory"
2024-11-11T15:25:24.761Z [DEBUG] login unauthorized: err="error verifying token signature: no known key successfully validated the token signature"
    path_login_test.go:1229: error mismatch, expected (invalid expiration time (exp) claim: token is expired) got (permission denied)

=== FAIL: builtin/credential/kubernetes TestLoginProjectedToken (0.00s)
2024-11-11T15:25:24.757Z [TRACE] Root CA certificate pool has changed, updating the client's transport
```

The test expects to default to using the JWT provided from the client request towards Kubernetes token review, rather than the Kubernetes service account token, so the initial message "failed to read local service account token" is expected. The final message "Root CA certificate pool has changed" is also expected.

However, the message "verifying token signature" was unexpected and led to "permission denied" being returned [here](https://github.com/openbao/openbao/blob/9e874309f5e549fb4f7a1fd7582f4222ba341ea4/builtin/credential/kubernetes/path_login.go#L135-L138).  The failure happens during JWT verification in github.com/hashicorp/cap  [here](https://github.com/hashicorp/cap/blob/8c4d3db5ad15bc5a19dc1b35fed4d07e00b549a2/jwt/keyset.go#L170-L172) , which should have used the public key originally provided [here](https://github.com/openbao/openbao/blob/9e874309f5e549fb4f7a1fd7582f4222ba341ea4/builtin/credential/kubernetes/path_login_test.go#L250) and stored in inmem storage as part of config. Other [sub-tests](https://github.com/openbao/openbao/blob/9e874309f5e549fb4f7a1fd7582f4222ba341ea4/builtin/credential/kubernetes/path_login_test.go#L1159-L1187) seem to validate tokens with the same key with no error, so it does not make sense to me why this particular test failed.

cipherboy:
> @tsaarni Hmm looks like Kubernetes might be relying on the deprecated behavior. Do you want to take a look or should I try to adjust the tests? Do you have allow maintainer to push to branch turned on?

Sorry @tsaarni I was looking at this CI run and thinking about #691 and the JWT update. I concur, I think in memory I might've seen this race before there. Sorry for the noise!

tsaarni:
Ok, no problem! I’m curious what might be causing the flaky test, since the key pair and tokens are pre-generated at package `init()` and in `TestLoginProjectedToken()` they are configured in `setupBackend()` outside the sub-test loop and goroutines lauched by `t.Run()`. It would make sense for either all or none of the sub-tests to fail, and I wouldn’t have expected delays from the race detector to interfere with e.g. public keys’ availability for token validation, causing the error "no known key successfully validated the token signature".

cipherboy (approved):

cipherboy:
Thank you, @tsaarni!

PR #693, opened 2024-11-11
Bump github.com/go-jose/go-jose/v3 to v3.0.3

tsaarni:
This PR resolves CVE-2024-28180 in github.com/go-jose/go-jose/v3 used in `api` and `sdk` submodules.  Although this vulnerability does not directly affect OpenBao's security, it is flagged in vulnerability scans. Upgrading the dependency will resolve this issue in future scans.


tsaarni:
It looks like the error below in "Run Go tests with data race detection / test-go (11)" has been recurring flake? I was seeing it occurring in older workflow runs as well.

```
=== FAIL: builtin/credential/jwt TestLogin_Leeways (4.33s)
    path_login_test.go:1091: [test 19: not yet valid custom nbf leeway using exp with no clock_skew_leeway jws: {<nil> 0xc00040e840 map[]  [] <nil> map[]}] expected token not valid yet error, got : true
```

cipherboy (approved):

PR #696, opened 2024-11-12
Move deadlock tests to separate file

cipherboy:
This lets us disable the deadlock tests in the race detector, keeping CI clean. Unfortunately, the deadlock library in use does not allow atomic control over these values and doesn't recommend setting them post-startup. While running in tests are serial and nothing else is executing in the process state, it is not ideal but workable. This code is not used in production.


JanMa:
> @JanMa After 4 reruns (discussed in #691), this is finally fully CI clean :D

That is one flaky test :joy:

JanMa (approved):

PR #697, opened 2024-11-12
Clean up invalid certs tidy conditions

cipherboy:
This proposes a minor refactor from #665 and adds a new test for ensuring revocation entries are cleaned up as well. 

/cc @fatima2003

### DIFF
--- a/builtin/logical/pki/path_tidy.go
+++ b/builtin/logical/pki/path_tidy.go
@@ -971,7 +971,10 @@ func (b *backend) doTidyCertStore(ctx context.Context, req *logical.Request, log
 				haveWarned = true
 			}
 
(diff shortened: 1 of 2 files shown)


builtin/logical/pki/path_tidy.go
```diff
@@ -982,14 +985,21 @@ func (b *backend) doTidyCertStore(ctx context.Context, req *logical.Request, log
 			continue
 		}
 
+		// We could be exclusively looking for invalid certificates; skip
+		// fetching a known-good revocation entry here if so. This also lets
+		// us avoid guarding each deletion check below.
+		if !config.CertStore {
```
cipherboy, line 991: This is a behavior change and would be good to get in before Beta if we can.

JanMa (approved):
LGTM

PR #698, opened 2024-11-13
Adjust the JWT leeway tests to reduce time sensitivity.

tsaarni:
This PR attempts to address a flaky test in the "Run Go tests with data race detection / test-go (11)" group, which contains a time-sensitive test case that frequently fails with the following error:

```
=== FAIL: builtin/credential/jwt TestLogin_Leeways (4.33s)
    path_login_test.go:1091: [test 19: not yet valid custom nbf leeway using exp with no clock_skew_leeway jws: {<nil> 0xc00040e840 map[]  [] <nil> map[]}] expected token not valid yet error, got : true
```

The test seems to use unnecessarily strict timing constraints.

https://github.com/openbao/openbao/blob/737c77b39b1ba28dd474141d5a20bab48c41ca20/builtin/credential/jwt/path_login_test.go#L1065-L1066

It checks that a JWT with an expiration set 7 seconds in the future is invalid when the not-before leeway is 5 seconds. I think this leaves a 2 second margin, but the problem is made worse by using `time.Now()` at the time of initializing the table-driven test, rather than making each sub-test use the current time during each test iteration. Also, the strictest timing requirement is on the last test, which is executed the longest time after the table initialization time.

I believe this can be tagged with "pr/no-changelog" label.

xref #691

### DIFF
--- a/builtin/credential/jwt/path_login_test.go
+++ b/builtin/credential/jwt/path_login_test.go
@@ -1062,8 +1062,8 @@ func testLogin_NotBeforeClaims(t *testing.T, jwks bool) {
 		// exp, no clock_skew_leeway (0s), custom nbf leeway (5s)
 		{"custom nbf leeway using exp with no clock_skew_leeway", true, jwks, time.Time{}, time.Time{}, time.Now().Add(3 * time.Second), -1, 5},
 		{"custom nbf leeway using exp with no clock_skew_leeway", true, jwks, time.Time{}, time.Time{}, time.Now().Add(3 * time.Second), -100, 5},
-		{"not yet valid custom nbf leeway using exp with no clock_skew_leeway", false, jwks, time.Time{}, time.Time{}, time.Now().Add(7 * time.Second), -1, 5},
-		{"not yet valid custom nbf leeway using exp with no clock_skew_leeway", false, jwks, time.Time{}, time.Time{}, time.Now().Add(7 * time.Second), -100, 5},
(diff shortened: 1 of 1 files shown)


cipherboy (approved):
Makes sense, many thanks for this @tsaarni!

cipherboy:
@tsaarni If you're up for it, `not yet valid auto nbf leeway using exp with no clock_skew_leeway` also failed on #699 fwiw:

https://github.com/openbao/openbao/actions/runs/11824640035/attempts/1?pr=699

> ```
> 2024-11-13T19:57:07.2318911Z === [31mFAIL[0m: builtin/credential/jwt TestLogin_Leeways (3.40s)
> 2024-11-13T19:57:07.2320470Z     path_login_test.go:1091: [test 11: not yet valid auto nbf leeway using exp with no clock_skew_leeway jws: {<nil> 0xc000257340 map[]  [] <nil> map[]}] expected token not valid yet error, got : true
> ```

tsaarni:
No, I think it is the same problem after all:

When `nbf` is unset github.com/hashicorp/cap will [automatically set](https://github.com/hashicorp/cap/blob/8c4d3db5ad15bc5a19dc1b35fed4d07e00b549a2/jwt/jwt.go#L217-L220) it to `exp` - `defaultLeewaySeconds` (which is 150 seconds [here](https://github.com/hashicorp/cap/blob/8c4d3db5ad15bc5a19dc1b35fed4d07e00b549a2/jwt/jwt.go#L19), so checking if token is valid yet has again 2 seconds margin.  These tests are bit hard to understand as they are written and one needs to read github.com/hashicorp/cap in order to make sense - it is almost like they are written to test _that library_ :)

I will submit a PR with the suggested change.

Issue #702, opened 2024-11-13
inmem - Occasional logging after test finishes


cipherboy:
See the following trace:

```
2024-11-13T19:31:27.9529024Z goroutine 539 [running]:
2024-11-13T19:31:27.9529558Z testing.(*common).logDepth(0xc0009dd040, {0xc0000221f8, 0x14}, 0x3)
2024-11-13T19:31:27.9530323Z 	/opt/hostedtoolcache/go/1.23.3/x64/src/testing/testing.go:1029 +0x6d4
2024-11-13T19:31:27.9530932Z testing.(*common).log(...)
2024-11-13T19:31:27.9531495Z 	/opt/hostedtoolcache/go/1.23.3/x64/src/testing/testing.go:1011
2024-11-13T19:31:27.9532202Z testing.(*common).Logf(0xc0009dd040, {0x929e6a, 0x14}, {0x0, 0x0, 0x0})
2024-11-13T19:31:27.9532962Z 	/opt/hostedtoolcache/go/1.23.3/x64/src/testing/testing.go:1062 +0xa5
2024-11-13T19:31:27.9533964Z github.com/openbao/openbao/sdk/v2/physical.testTransactionalMixedWriters.func2(0x12)
2024-11-13T19:31:27.9534753Z 	/home/runner/work/openbao/openbao/sdk/physical/testing.go:927 +0x639
2024-11-13T19:31:27.9535708Z created by github.com/openbao/openbao/sdk/v2/physical.testTransactionalMixedWriters in goroutine 14
2024-11-13T19:31:27.9536780Z 	/home/runner/work/openbao/openbao/sdk/physical/testing.go:920 +0x37b
2024-11-13T19:31:27.9537544Z FAIL	github.com/openbao/openbao/sdk/v2/physical/inmem	252.885s
```

This is coming from the following function: https://github.com/openbao/openbao/blob/main/sdk/physical/testing.go#L837

PR #703, opened 2024-11-14
[fix][inmem - Occasional logging after test finishes #702] 

DrGabriel:
# **Release-note:bug**

## file - sdk/physical/testing.go: 
1. Added Dedicated sync.WaitGroup for Writers, Readers, and Listers:
Purpose: To ensure proper synchronization and orderly completion of all goroutines.
Benefit: This approach ensures that the main test function waits for all writer, reader, and lister goroutines to complete before proceeding, preventing race conditions and ensuring consistent test results.

2. Removed time.Sleep Calls:
Purpose: To eliminate arbitrary delays and rely on proper synchronization mechanisms.
Benefit: This makes the code more efficient and predictable. Using sync.WaitGroup ensures that the test waits for all necessary operations to complete without relying on fixed sleep durations, which can be unreliable and lead to flaky tests.
2.1: The sleep times weren't 100% stable and could cause scenarios where the readers or listers would write to the log after the test completion, leading to the trace mentioned in the original issue.

3. Keep the "done" flag to signal readers and listeners instead of switching to context with cancellation to avoid applying too much of complex changes to something that already works

# Main issue
Resolves [#702](https://github.com/openbao/openbao/issues/702)

## Target Release

1.14.7

### DIFF
--- a/sdk/physical/testing.go
+++ b/sdk/physical/testing.go
@@ -606,7 +606,9 @@ func testTransactionalExclusiveWriters(t testing.TB, b TransactionalBackend) {
 	// transactions plus a lister outside of transactions (which should still
 	// be atomic relative to the transactions that are occurring). When the
 	// writers are done, signal the readers/listers to finish up.
-	var wg sync.WaitGroup
+	var wgWriters sync.WaitGroup
+	var wgReaders sync.WaitGroup
+	var wgListers sync.WaitGroup
 	var done atomic.Bool
 	var numFiles int = 25
 	var numWriters int = 100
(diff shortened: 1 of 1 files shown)


cipherboy:
@DrGabriel Any chance you could [DCO sign-off](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco-sign-off-policy) the commit and force-push it? Thanks!

DrGabriel:
> @DrGabriel Any chance you could [DCO sign-off](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco-sign-off-policy) the commit and force-push it? Thanks!

Will do! Thanks :)

cipherboy (approved):
Sometimes you just need more WaitGroups. Thanks, @DrGabriel!

PR #704, opened 2024-11-14
Get current time in each JWT leeway subtest

tsaarni:
This PR continues from #698, which did not fully resolve the CI time sensitivity issue.

In this change, I've "unrolled" the table-driven test so that each sub-test now uses the current time at the beginning of each-sub-test. This replaces the previous approach where tests ran relative to the time during table initialization at the start of the whole test case, which caused frequent failures due to strict timing sensitivity.

The diff looks larger than the actual changes. Although I (mostly) kept the formatting of the table and only added the function calls, it's showing as if the entire section was modified.  I made one change that might be hard to spot in the diff: I removed [this](https://github.com/openbao/openbao/blob/758e24181d71c93a75dc36cd0d17ab55d485c475/builtin/credential/jwt/path_login_test.go#L989) and [this](https://github.com/openbao/openbao/blob/758e24181d71c93a75dc36cd0d17ab55d485c475/builtin/credential/jwt/path_login_test.go#L1066) test, which ran the same test as the line above, but with default leeway of -100 instead of -1. When troubleshooting, the duplicate test names were confused, and I did not find what is the point to test with several different negative values since they all fall under [this single if branch](https://github.com/hashicorp/cap/blob/8c4d3db5ad15bc5a19dc1b35fed4d07e00b549a2/jwt/jwt.go#L226-L227) in github.com/hashicorp/cap - therefore I think it should be ok to test only -1, like it was done in all other sub-tests.

As previously, this PR can be tagged with "pr/no-changelog" label.

### DIFF
--- a/builtin/credential/jwt/path_login_test.go
+++ b/builtin/credential/jwt/path_login_test.go
@@ -928,7 +928,7 @@ func TestLogin_Leeways(t *testing.T) {
 }
 
 func testLogin_ExpiryClaims(t *testing.T, jwks bool) {
-	tests := []struct {
+	type params struct {
 		Context       string
 		Valid         bool
 		JWKS          bool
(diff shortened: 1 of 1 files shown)


tsaarni:
Is this failure new https://github.com/openbao/openbao/actions/runs/11832781040/job/32970215695?pr=704#step:7:14381 ?

cipherboy:
@tsaarni Kinda, that is what #702 mentions, fixed by #703. It is simply a test failure, not due to the underlying inmem library.

cipherboy (approved):
Appreciate it, thanks @tsaarni!

PR #705, opened 2024-11-14
Bump sdk to api/v2@v2.1.0

cipherboy:
~This replace directive meant that anyone consuming our SDK would have to include a similar replace directive in their own go module. This is an unfortunate side-effect of how modules work and undesirable in general; remove this.~

This may be incorrect, see: https://github.com/openbao/openbao/pull/705#pullrequestreview-2435486904

### DIFF
--- a/sdk/go.mod
+++ b/sdk/go.mod
@@ -46,7 +46,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/openbao/go-kms-wrapping/entropy/v2 v2.1.0
 	github.com/openbao/go-kms-wrapping/v2 v2.2.0
-	github.com/openbao/openbao/api/v2 v2.0.1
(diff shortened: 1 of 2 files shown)


sdk/go.mod
```diff
@@ -9,8 +9,6 @@ go 1.22.1
 
 toolchain go1.22.5
 
-replace github.com/openbao/openbao/api/v2 => ../api
```
cipherboy, line 12: While this makes consumption of the sdk easier, it does have the unfortunate side-effect that changes in the API aren't immediately reflected in local testing of the SDK... But presumably the combination of API+SDK functionality is adequately tested through the use in the main module, which does have the correct replace directives to make this work.

Thoughts?

cipherboy: I've reverted this change for the time being; I see Brave was able to pull the dependency successfully without it. I've asked on https://github.com/brave-intl/bat-go/pull/2707.

PR #706, opened 2024-11-14
Bump go-kms-wrapping versions for release

cipherboy:
@JanMa I don't think our release instructions adequately describe this, but there is an oddly circular dependency graph here:

```
[cipherboy@x1c go-kms-wrapping]$ gif 'openbao/sdk'
wrappers/transit/go.mod:13:	github.com/openbao/openbao/sdk/v2 v2.0.1
```

In particular go-kms-wrapping's transit wrapper depends on SDK for the testcluster, but SDK also depends on updated go-kms-wrapping. Since the former is in tests, the latter should take precedence, which means we should probably update go-kms-wrapping before tagging SDK, and not visa-versa.

Something to note in our release checklist for next time. This should thus merge ahead of #705 and before tagging `sdk/v2.1.0`


JanMa (approved):

cipherboy:
Thanks, @JanMa!

PR #713, opened 2024-11-14
Fix goreleaser prerelease status

cipherboy:
This fixes a bug in the goreleaser status that failed to wire up the prerelease status correctly. In particular, it meant that our beta release was both tagged as latest in GitHub but also in container images. 

See: 

https://github.com/openbao/openbao/blob/d45510adab420b225d6bf2c166b197358cfc4a0b/.github/workflows/release.yml#L144

Note the `_RELEASE_` that wasn't present earlier.


.goreleaser-template.yaml
```diff
@@ -595,7 +595,7 @@ checksum:
 #LINUXONLY#      - ghcr.io/{{ .Env.GITHUB_REPOSITORY_OWNER }}/openbao:{{ .Version }}-riscv64
 #LINUXONLY#      - ghcr.io/{{ .Env.GITHUB_REPOSITORY_OWNER }}/openbao:{{ .Version }}-s390x
 #LINUXONLY#  - name_template: ghcr.io/{{ .Env.GITHUB_REPOSITORY_OWNER }}/openbao:{{ .Major }}
-#LINUXONLY#    skip_push: false
+#LINUXONLY#    skip_push: ${{ .Prerelease }}
```
cipherboy, line 598: Notably, this will make it hard to do point releases for security issues in older branches when we have a newer GA major major version, given we have no way of pushing tags manually.

However, the TSC approved a strictly linear support policy so I think we're good to do so.

JanMa (approved):

PR #521, opened 2024-09-12
Update mitchellh/cli to hashicorp/cli

cipherboy:
This updates an archived dependency `mitchellh/cli` to the maintained variant, `hashicorp/cli`.

PR #714, opened 2024-11-14
Replace github.com/mholt/archiver/v3 with stdlib (#611)

JanMa:
* Replace github.com/mholt/archiver/v3 with stdlib

As suggested by Wojciech Kocjan, replace insecure archiver (which lacks
a patched release except via third-party fork) with the standard
library's archive/tar and compress/gzip. This removes it as a direct
dependency.

Notably, our use of archiver was not affected by the vulnerability, as
we only created archives (in `bao debug -compress`) and did not read
them (except for self-created archives as part of the test suite).

* Add changelog entry

---------

### DIFF
--- a/command/debug.go
+++ b/command/debug.go
@@ -4,10 +4,13 @@
 package command
 
 import (
+	"archive/tar"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/fs"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


cipherboy (approved):

command/debug.go
```diff
@@ -20,8 +23,6 @@ import (
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-secure-stdlib/gatedwriter"
 	"github.com/hashicorp/go-secure-stdlib/strutil"
-	"github.com/mholt/archiver/v3"
-	"github.com/mitchellh/cli"
 	"github.com/oklog/run"
```
cipherboy, line 26:
```suggestion
	"github.com/mitchellh/cli"
	"github.com/oklog/run"
```

cipherboy: @JanMa Too greedy with the import removal I think.

command/debug_test.go
```diff
@@ -15,8 +19,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/mholt/archiver/v3"
-	"github.com/mitchellh/cli"
+	"github.com/hashicorp/cli"
```
cipherboy, line 22:
```suggestion
	"github.com/mitchellh/cli"
```

cipherboy: Ah, this was before #521

PR #716, opened 2024-11-14
Compute raft peer bootstrap challenge via HKDF (#690)

JanMa:
* Compute raft peer bootstrap challenge via HKDF

Rather than storing a map of serverID->answer (and generating fully
random challenge answers), use HKDF on the serverID and a secret key
to create challenge answers using only a single constant-space key.
HKDF should be at least as fast as the underlying RNG source and this
ensures we do not incur any additional memory per request to this
un-authenticated endpoint.

See also: https://discuss.hashicorp.com/t/hcsec-2024-26-vault-vulnerable-to-denial-of-service-through-memory-exhaustion-when-processing-raft-cluster-join-requests/71047

* Add changelog entry

* Refactor limit to constant; clarify panic

---------

### DIFF
--- a/vault/logical_system_raft.go
+++ b/vault/logical_system_raft.go
@@ -5,6 +5,7 @@ package vault
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/subtle"
 	"encoding/base64"
 	"errors"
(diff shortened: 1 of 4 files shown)


vault/logical_system_raft.go
```diff
@@ -292,6 +324,9 @@ func (b *SystemBackend) handleRaftBootstrapAnswerWrite() framework.OperationFunc
 		if len(serverID) == 0 {
 			return logical.ErrorResponse("no server_id provided"), logical.ErrInvalidRequest
 		}
+		if len(serverID) > 16384 {
```
cipherboy, line 327: @JanMa Ah, I had missed the constant here. I'll update that in a bit on `main`.

cipherboy (approved):

PR #723, opened 2024-11-15
PKI: clarify usage of set-signed command

klaus-sap:
Not only a single certificate, but a bundle of PEM-encoded certificates may be passed

Resolves #

## Target Release

2.1.1

### DIFF
--- a/website/content/docs/secrets/pki/quick-start-intermediate-ca.mdx
+++ b/website/content/docs/secrets/pki/quick-start-intermediate-ca.mdx
@@ -109,10 +109,11 @@ serial_number   10:dc:50:0f:b2:88:26:2d:73:13:f8:c4:89:8a:80:1b:55:42:e0:dc
 ```
 
 Now set the intermediate certificate authorities signing certificate to the
-root-signed certificate.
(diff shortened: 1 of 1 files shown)


website/content/docs/secrets/pki/quick-start-intermediate-ca.mdx
```diff
@@ -109,10 +109,11 @@ serial_number   10:dc:50:0f:b2:88:26:2d:73:13:f8:c4:89:8a:80:1b:55:42:e0:dc
 ```
 
 Now set the intermediate certificate authorities signing certificate to the
-root-signed certificate.
+root-signed certificate. You may pass a bundle of PEM-encoded certificates,
```
cipherboy, line 112: You may provide it, but need not:

https://github.com/openbao/openbao/blob/74f2aa48da97b4d4a0a3d1ccb92b449faffac4c7/builtin/logical/pki/chain_test.go#L186-L192

This test shows that we can set exactly a single PEM certificate via this endpoint. So I think your failures from Matrix were unrelated...

cipherboy (approved):

PR #732, opened 2024-11-16
Add version to download page, prerelease warning

cipherboy:
When switching releases, add a version to the query string (replacing the current state) so users can send the link to the exact version to others. Additionally, add a warning box to prerelease versions.

image


JanMa (approved):
Nice idea :heavy_check_mark:

PR #733, opened 2024-11-16
Update website metadata

cipherboy:
This updates various website metadata and adds a robots.txt to link the sitemap to hopefully make it more visible to search engines. 

Sadly I don't think the `<link rel="sitemap" ... />` is actually working though: https://docusaurus.io/docs/api/docusaurus-config#headTags


JanMa (approved):
LGTM

PR #735, opened 2024-11-16
Remove HCP Link status in UI

cipherboy:
This was already hidden from the UI (likely because the server removed support for HCP Link in #63), but there was still code lying around to support it. 

/cc @bozzo -- do you mind reviewing this? Thanks!

Also, if you have time, it'd be great to get a sense of what's left in the UI itself before we're happy calling it ready for release and use!


bozzo:
Hi @cipherboy, I think you can also remove `ui/app/components/link-status.js`.
In `ui/app/components/sidebar/frame.hbs` on line 36, I think you can also remove link to that status.
Apart from that, LGTM.

bozzo:
> Also, if you have time, it'd be great to get a sense of what's left in the UI itself before we're happy calling it ready for release and use!

From what I've seen:
* there are Sentinel related stuff (they don't seem to be visible)
* replication related stuff
* `ui/app/services/version.js` which is used in many places (I started working on it but broke everything so far)
* `ui/public/images/mfa-landing.png` & `ui/public/images/oidc-landing.png` still have Vault logo and name

I also saw some namespace related tests in the Enterprise section that probably need to be removed or updated to cover the usage of https://openbao.org/api-docs/system/internal-ui-namespaces/.

JanMa (approved):

PR #736, opened 2024-11-16
Allow listing prefixed policies

cipherboy:
The `sys/policies/acl/*` paths allow arbitrary structuring of policies into paths, e.g., `a/b` and `a/c` may exist. Using path prefixes are useful e.g., when structuring policies into pipelines or projects or tenants. However, policy listing was not capable of only listing policies under a given prefix. So results for `d/e`, `default`, and `root` would also be returned, even if only policies under `a/` were desired by the caller.

Support list operations on prefixes to allow callers to view policies only under some root prefix.

### DIFF
--- a/changelog/736.txt
+++ b/changelog/736.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+core/policies: Allow listing policies under a given prefix.
+```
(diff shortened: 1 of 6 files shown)


DanGhita (approved):
LGTM

cipherboy:
Thanks, @DanGhita!

Issue #578, opened 2024-10-02
incomplete non-voting Raft member implementation


zarganum:
**Is your feature request related to a problem? Please describe.**
The non-voting cluster member is used for native Raft replication (originally Enterprise-only).

**Describe the solution you'd like**
[Re-enable](https://github.com/openbao/openbao/blob/main/physical/raft/raft.go#L997) the `non_voter` option.

**Describe alternatives you've considered**
The `retry_join_as_non_voter` option is available but impacts only cluster join.

**Additional context**
The `retry_join_as_non_voter` is marked Enterprise-only in parent product, but still ported here without problems.
However the accompanying `non_voter` option is [explicitly disabled](https://github.com/openbao/openbao/blob/main/physical/raft/raft.go#L997). How so?

cipherboy:
\o hey @zarganum! If you're interested in contributing this, I'd definitely be interested in reviewing it.

This feature was only in Vault Enterprise so not part of the Vault Community edition pre-fork so we never had an implementation for it. :-) I'm not quite sure how large of a feature this is, but if you think its small enough, feel free to file a PR for it, else if you think its larger, a RFC would be appreciated!

zarganum:
... and if you'd ask me how I've got here: because I broke my cluster badly and start investigating different offline recovery mechanisms that will work when online API fails. Fortunately that was an experimental deployment and I've found [enough info](https://developer.hashicorp.com/vault/tutorials/monitoring/inspect-data-boltdb) about auto-recovery mechanism. What authors did not share is that `latest_config` simply contains a [protobuf version](https://github.com/openbao/openbao/blob/main/physical/raft/types.proto#L39) of `peers.json`.

cipherboy:
\o hello @zarganum:

> Before further investments, may I know if there are legal limitations on re-implementing (or simply porting) the original enterprise features to the OpenBao public fork?

Clean room implementations without looking at upstream's code are the standard here. This is mentioned in the [`CONTRIBUTING.md`](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#source-available-license-policy) file:

> This policy prohibits the utilization or copy from any source that is protected by a source-available license agreement including Business Source License 1.1. This specifically includes all source repository members under the Hashicorp GitHub organization. It is the intention of this policy that when developing for feature parity, bug fixes, CVEs, etc. in comparison to Hashicorp Vault that a clean room methodology be used and all deliverables made original.

[...]

PR #741, opened 2024-11-17
Allow Nodes to Join the Raft Cluster as Non-Voters

JanMa:
#### Summary
This PR reimplements a previous enterprise feature to allow nodes to join a Raft cluster as non-voting members. Nodes can now be configured as non-voters by either using the new `-non-voter` flag in the `operator raft join` command or by setting the existing configuration option `retry_join_as_non_voter`.

#### Key Changes
- Introduced a `-non-voter` flag to the `operator raft join` command.
- Enhanced join logic to support both the `-non-voter` flag and the pre-existing `retry_join_as_non_voter` configuration option.
- Updated relevant Raft implementation files and improved the handling of non-voter joins.
- Added documentation updates to reflect these changes.

Closes: #578

### DIFF
--- a/changelog/741.txt
+++ b/changelog/741.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+raft: Added support for nodes to join the Raft cluster as non-voters.
+```
--- a/physical/raft/raft_autopilot.go
+++ b/physical/raft/raft_autopilot.go
@@ -473,6 +473,7 @@ func autopilotToAPIState(state *autopilot.State) (*AutopilotState, error) {
 		FailureTolerance: state.FailureTolerance,
 		Leader:           string(state.Leader),
 		Voters:           stringIDs(state.Voters),
+		NonVoters:        []string{},
 		Servers:          make(map[string]*AutopilotServer),
 	}
 
@@ -482,6 +483,10 @@ func autopilotToAPIState(state *autopilot.State) (*AutopilotState, error) {
 			return nil, err
 		}
 		out.Servers[string(id)] = aps
+
+		if aps.NodeType == "non-voter" {
+			out.NonVoters = append(out.NonVoters, string(id))
+		}
 	}
 
 	return out, nil
--- a/physical/raft/raft_autopilot.go
+++ b/physical/raft/raft_autopilot.go
@@ -1,3 +1,4 @@
+// Copyright (c) 2024 OpenBao a Series of LF Projects, LLC
 // Copyright (c) HashiCorp, Inc.
 // SPDX-License-Identifier: MPL-2.0
 
@@ -15,7 +16,6 @@ import (
 
 	"github.com/armon/go-metrics"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
-	"github.com/hashicorp/go-secure-stdlib/strutil"
 	"github.com/hashicorp/raft"
 	autopilot "github.com/hashicorp/raft-autopilot"
 	"github.com/mitchellh/mapstructure"
@@ -29,6 +29,9 @@ const (
 	CleanupDeadServersUnset CleanupDeadServersValue = 0
 	CleanupDeadServersTrue  CleanupDeadServersValue = 1
 	CleanupDeadServersFalse CleanupDeadServersValue = 2
+
+	// NonVoterPath is the path to the non-voters in the storage.
+	NonVoterPath = "autopilot/non-voters"
 )
 
 func (c CleanupDeadServersValue) Value() bool {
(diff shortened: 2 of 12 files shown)


website/content/api-docs/system/storage/raft.mdx
```diff
@@ -52,6 +52,11 @@ leader node.
 
 - `auto_join_port` `(int: 8200)` - Port to be used for `auto_join`.
 
+- `non_voter` `(bool: false)` - If set, will make the server not
+  participate in the Raft quorum, and have it only receive the data replication
+  stream. This can be used to add read scalability to a cluster in cases where a
+  high volume of reads to servers are needed. The default is false.
```
cipherboy, line 58: One nit if you want to look it up: does adding non-voter nodes increase application latency? Is data  sent to non-voter nodes handled inline to the application or is is queued separately? If you run a cluster with, say, 5 voters and 206 non-voters, does this greatly increase the time to apply logs still?

Just something we might want to call out in docs if it does.

changelog/741.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
cipherboy, line 1: Should probably be a feature :-D

physical/raft/raft_autopilot.go
```diff
@@ -263,30 +268,37 @@ type Delegate struct {
...
+	}
 	d.l.RLock()
 	config := &autopilot.Config{
 		CleanupDeadServers:      d.autopilotConfig.CleanupDeadServers,
 		LastContactThreshold:    d.autopilotConfig.LastContactThreshold,
 		MaxTrailingLogs:         d.autopilotConfig.MaxTrailingLogs,
 		MinQuorum:               d.autopilotConfig.MinQuorum,
 		ServerStabilizationTime: d.autopilotConfig.ServerStabilizationTime,
-		Ext:                     nil,
+		Ext:                     d.permanentNonVoters,
```
cipherboy, line 301: I guess it doesn't truly matter since this value isn't persisted, but would we want to preemptively set this to a struct of our own and not directly place the non-voters map here in case we want to add more fields?

physical/raft/raft_autopilot.go
```diff
@@ -803,3 +825,131 @@ func (b *RaftBackend) SetupAutopilot(ctx context.Context, storageConfig *Autopil
 
 	go b.startFollowerHeartbeatTracker()
 }
+
+// Ensure that the CustomPromoter implements the autopilot.Promoter interface
```
cipherboy, line 829: Do you want to move this to a new file with our project copyright header?

physical/raft/raft_autopilot.go
```diff
@@ -803,3 +825,131 @@ func (b *RaftBackend) SetupAutopilot(ctx context.Context, storageConfig *Autopil
...
+var _ autopilot.Promoter = (*CustomPromoter)(nil)
+
+type CustomPromoter struct{}
+
+func (_ *CustomPromoter) GetServerExt(_ *autopilot.Config, srv *autopilot.ServerState) interface{} {
```
cipherboy, line 834: nit, consistency:

```suggestion
func (_ *CustomPromoter) GetServerExt(_ *autopilot.Config, _ *autopilot.ServerState) interface{} {
```

physical/raft/raft_autopilot.go
```diff
@@ -803,3 +825,131 @@ func (b *RaftBackend) SetupAutopilot(ctx context.Context, storageConfig *Autopil
...
+	return types
+}
+
+func (_ *CustomPromoter) FilterFailedServerRemovals(_ *autopilot.Config, _ *autopilot.State, failed *autopilot.FailedServers) *autopilot.FailedServers {
+	return failed
+}
+
+// CalculatePromotionsAndDemotions will return a list of all promotions and demotions to be done as well as the server id of
+// the desired leader. This particular interface implementation maintains a stable leader and will promote healthy servers
+// to voting status if they are not makred as permanent non-voters. It will never change the leader ID nor will it perform demotions.
```
cipherboy, line 862:
```suggestion
// to voting status if they are not marked as permanent non-voters. It will never change the leader ID nor will it perform demotions.
```

physical/raft/raft_autopilot.go
```diff
@@ -803,3 +825,131 @@ func (b *RaftBackend) SetupAutopilot(ctx context.Context, storageConfig *Autopil
...
+func (_ *CustomPromoter) CalculatePromotionsAndDemotions(c *autopilot.Config, s *autopilot.State) autopilot.RaftChanges {
+	var changes autopilot.RaftChanges
+
+	now := time.Now()
+	minStableDuration := s.ServerStabilizationTime(c)
+	nonVoters := c.Ext.(map[raft.ServerID]bool)
+	for id, server := range s.Servers {
+		// ignore non-voters
```
cipherboy, line 870: I suppose, we write non-voter status into the config file, but would we want to make it API controlled in the future and be able to demote voter->non-voter?

physical/raft/raft_autopilot.go
```diff
@@ -263,30 +268,37 @@ type Delegate struct {
...
 
 // AutopilotConfig is called by the autopilot library to know the desired
 // autopilot configuration.
 func (d *Delegate) AutopilotConfig() *autopilot.Config {
+	// Always fetch non-voters before returning the config
+	err := d.FetchNonVoters()
+	if err != nil {
+		d.logger.Error("failed to fetch non-voters", "error", err)
+	}
 	d.l.RLock()
```
cipherboy, line 294: What is the ordering/locking requirements w.r.t. `d.l` vs `d.dl`?

cipherboy (commented):
If I were to summarize the current design:

Leader's config dictates the non-voting status of peers, so this implies all peers have the same configuration for non-voting status for retry-join? This clobbers the local storage version on startup, which is subsequently used throughout the lifecycle of the node, until a leadership election changes to a different leader?

But for manually joined nodes, the value chosen at initial join is persisted throughout its lifecycle and there's no way to update it without removing and re-joining?

Just making sure I understand the implications of this...

website/content/api-docs/system/storage/raft.mdx
JanMa: Honestly I am not sure. I copy pasted the text from the previous enterprise docs

physical/raft/raft_autopilot.go
JanMa: This value is only used in the `CustomPromoter` struct. If we ever want to add more fields here in the future, I think this will be very straightforward to extend

physical/raft/raft_autopilot.go
JanMa: Good question. `d.l` is the lock which the delegate inherits from it's `RaftBackend` instance, `d.dl` is the lock for it's own fields. Which means, I am missing a read lock for the non-voter map here :+1:

vault/logical_system_raft.go
```diff
@@ -375,9 +375,9 @@ func (b *SystemBackend) handleRaftBootstrapAnswerWrite() framework.OperationFunc
 
 		switch nonVoter {
 		case true:
-			err = errors.New("adding non voting peer is not allowed")
+			err = raftBackend.AddPeer(ctx, serverID, clusterAddr, false)
 		default:
-			err = raftBackend.AddPeer(ctx, serverID, clusterAddr)
+			err = raftBackend.AddPeer(ctx, serverID, clusterAddr, true)
```
cipherboy, line 380: Ahh, right, I see now, `AddPeer` is called here by non-autopilot instances. I gotcha.

cipherboy (approved):
Looks like the UI even works!

image

cipherboy:
@JanMa said:

> I also thought it might be good to expose this via an API so it can be configured at runtime which nodes are permanent non-voters and which not.

This probably requires updating the CustomPromoter to do demotions as well then?

physical/raft/raft_autopilot_promoter.go
```diff
@@ -0,0 +1,55 @@
...
+
+	now := time.Now()
+	minStableDuration := s.ServerStabilizationTime(c)
+	nonVoters := c.Ext.(map[raft.ServerID]bool)
+	for id, server := range s.Servers {
+		// ignore non-voters
+		if _, ok := nonVoters[id]; ok {
+			continue
+		}
+		// ignore staging state as they are not ready yet
```
codergs, line 48: this comment can be bit more clear. If I am reading this correct, what you are trying to say is - if a server node isn't a voter yet (and also not part of the permanent non-voters configured by us), then we mark it for promotion to voter, is that correct? I would try to stay away from terminologies like staging state as those can be bit convoluted.

website/content/api-docs/system/storage/raft.mdx
codergs: yeah, perf will take a hit as leader will be replicating data to all those non-voters. I would throw in a caution on fine tuning this for intended use case.

physical/raft/raft_autopilot_delegate.go
```diff
@@ -0,0 +1,347 @@
...
+	config := &autopilot.Config{
+		CleanupDeadServers:      d.autopilotConfig.CleanupDeadServers,
+		LastContactThreshold:    d.autopilotConfig.LastContactThreshold,
+		MaxTrailingLogs:         d.autopilotConfig.MaxTrailingLogs,
+		MinQuorum:               d.autopilotConfig.MinQuorum,
+		ServerStabilizationTime: d.autopilotConfig.ServerStabilizationTime,
+		Ext:                     d.permanentNonVoters,
+	}
+	d.l.RUnlock()
+	d.dl.RUnlock()
```
codergs, line 67: nit: defer could be cleaner here

codergs (commented):
LGTM overall

website/content/api-docs/system/storage/raft.mdx
codergs: I am not familiar with raft library as such so can't estimate this easily. Conceptually, makes sense but don't know how hard it is. need a spike.

physical/raft/raft_autopilot_delegate.go
```diff
@@ -0,0 +1,347 @@
...
+	d.dl.RLock()
+	defer d.dl.RUnlock()
+	nonVoters := make([]raft.ServerID, 0, len(d.permanentNonVoters))
+	for id := range d.permanentNonVoters {
+		nonVoters = append(nonVoters, id)
+	}
+	return nonVoters
+}
+
+func (d *Delegate) UpdateNonVoters() error {
```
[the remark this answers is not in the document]
JanMa: Changed the name of the function to make the intention a bit clearer.

cipherboy (approved):
Looks good, thanks Jan!

PR #750, opened 2024-11-21
Clean up legacy keyring code

cipherboy:
This cleans up various legacy migrations that should have already occurred by Vault 1.14 (the last Vault release users should've come from).

1. Remove support for pre-keyring barrier encryption
2. Remove support for legacy (direct) Shamir seals.
3. Better enforce that there are only zero or one StoredShares (root keys). 
4. Simplify all references of master->root keys. This was partially renamed but never completed, which makes evaluating the key hierarchy in code much more difficult.

### DIFF
--- a/changelog/750.txt
+++ b/changelog/750.txt
@@ -1,0 +1,4 @@
+```release-note:change
+core/seal: remove support for legacy pre-keyring barrier entries
+core/seal: remove support for legacy (direct) shamir unseal keys
+```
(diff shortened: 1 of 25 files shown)


cipherboy:
Apologies, I guess this needs revisions :D

DanGhita (approved):
LGTM.
This code cleaning was really necessary, thank you @cipherboy !

cipherboy:
Just minor typing errors. :-)

cipherboy:
Thanks @DanGhita!

PR #766, opened 2024-11-27
Add detailed metadata list to KV

cipherboy:
Allow KV to list detailed information about metadata keys. With paginated listing and transactional storage, this allows a single, consistent view of metadata (including `current_version`, which in turn gives a consistent view over all data). This incurs additional work (reading each entry) and thus is done under a new path to allow separate ACLing.

---

This will be rebased once #763 lands, as it conflicts and also to take advantage of the new SCAN operation to support recursively listing all keys in K/V.

### DIFF
--- a/builtin/logical/kv/path_metadata.go
+++ b/builtin/logical/kv/path_metadata.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


builtin/logical/kv/path_metadata.go
```diff
@@ -81,6 +82,36 @@ version-agnostic information about a secret.
...
+				Description: `Optional number of entries to return; defaults to all entries. Only used for listing.`,
+			},
+		},
+		Callbacks: map[logical.Operation]framework.OperationFunc{
+			logical.ListOperation: b.upgradeCheck(b.pathMetadataList()),
+		},
+
+		ExistenceCheck: b.metadataExistenceCheck(),
+
+		HelpSynopsis:    confHelpSyn,
```
cipherboy, line 110: Interestingly, this is a shortcoming with the existing `pathMetadata` as well; probably needs to be fixed in a follow-up PR.

cipherboy:
@JanMa @DanGhita This should be ready for review after #763 merged. :-)

builtin/logical/kv/path_metadata.go
```diff
@@ -97,6 +98,37 @@ version-agnostic information about a secret.
 	}
 }
 
+// pathDetailedMetadata returns the path configuration for LIST operations on
```
DanGhita, line 101: comment: for LIST **and SCAN** operations ?

DanGhita (approved):
LGTM

builtin/logical/kv/path_metadata.go
cipherboy: Ahh, good catch. Thanks!

PR #789, opened 2024-12-11
Import issuer response includes friendly name

### LINKED ISSUE #785: Imported Issuer clarity
**Describe the bug**
As @dannyleesmith has reported on [Vault 26037](https://github.com/hashicorp/vault/issues/26037), when importing an issuer for CA rotation, the response is unclear and there is no way to consistently identify the newly imported issuer which is especially a problem for automation.

**Expected behavior**
The response should include an explicit identifier or metadata for issuers to simplify automation workflows and make it easier for operators to reference the correct issuer.

**Suggested Solution**
In RFC #679, @Gabrielopesantos has suggested importing an issuer with a friendly name which can later be referenced for update operations:

> Submit CA with Name (POST /ssh/issuers/import/:issuer_name): This is an extension of the ‘Submit CA Information’ endpoint but accepts an optional issuer_name parameter so the CA information has a name reference. If the name is left empty, a Unix timestamp will be used instead and an actual name can then be set using the 'Update CA' endpoint.

This idea can be similarly implemented in PKI, by introducing a new endpoint and enhancements:
- Add new issuer end point to accept the `issuer_name` and imports 1 issuer only. 
- Current endpoints cannot be modified to accept `issuer_name` since multiple can be imported at once so this would cause ambiguity.
- Update endpoint can take in `issuer_name` as reference.
---

This can be labelled **secret/pki**
Thank you @cipherboy for the suggestion! I am working on a solution :)

fatima2003:
As described in [Vault Issue #26037](https://github.com/hashicorp/vault/issues/26037), the current issuer import response does not provide a clear or consistent way to identify newly imported issuers, particularly for automation purposes. This makes it challenging to reference specific issuers in subsequent operations. 

This update enhances the `/pki/issuers/import/bundle` and `/pki/intermediate/set-signed` endpoints by allowing an optional `issuer_name` parameter, with responses now including a `issuer_name` field for easier reference. Issuers can also be updated by using their `issuer_name`, streamlining automation workflows.

Resolves #785 [Vault #26037](https://github.com/hashicorp/vault/issues/26037)

**TO DO** 
- [X] User can specify `issuer_name` and response includes this name.
- [ ] User can update issuer using `issuer_name`
- [ ] Create Comprehensive test
- [ ] Add docs

### DIFF
--- a/builtin/logical/pki/path_manage_issuers.go
+++ b/builtin/logical/pki/path_manage_issuers.go
@@ -220,6 +220,10 @@ func pathImportIssuer(b *backend) *framework.Path {
 				Description: `PEM-format, concatenated unencrypted
 secret-key (optional) and certificates.`,
 			},
+			"issuer_name": {
+				Type:        framework.TypeString,
+				Description: "Optional friendly name to assign to the imported issuer.",
+			},
 		},
 
 		Operations: map[logical.Operation]framework.OperationHandler{
@@ -296,13 +300,19 @@ func (b *backend) pathImportIssuers(ctx context.Context, req *logical.Request, d
 
 	var pemBundle string
 	var certificate string
+	var issuerName string
 	rawPemBundle, bundleOk := data.GetOk("pem_bundle")
 	rawCertificate, certOk := data.GetOk("certificate")
+	rawIssuerName, nameOk := data.GetOk("issuer_name")
+
 	if bundleOk {
 		pemBundle = rawPemBundle.(string)
 	}
 	if certOk {
 		certificate = rawCertificate.(string)
+	}
+	if nameOk {
+		issuerName = rawIssuerName.(string) // Type assertion to convert interface{} to string
 	}
 
 	if len(pemBundle) == 0 && len(certificate) == 0 {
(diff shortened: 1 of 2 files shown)


cipherboy:
Hey @fatima2003 -- do you recall where this one was at?

fatima2003:
> Hey @fatima2003 -- do you recall where this one was at?

Hi @cipherboy, lemme wrap her up

fatima2003:
Hello @cipherboy o/  

I was looking back into what the blocker for this PR was. Multiple issuers can be imported at once, so assigning a single `issuer_name` like done in [SSH: Submit issuer with optional name](https://openbao.org/api-docs/secret/ssh/#submit-issuer-with-optional-name) would not work in pki.

What do you think of a field, `issuer_names` that maps the name to the cert serial number similar to what was suggested in https://github.com/hashicorp/vault/issues/26037?

cipherboy:
@fatima2003 The issue is that multiple certificates might share a serial number... :thinking: I think our last discussion on this was a new import-single-issuer with name parameter like this one? Perhaps taking `issuer`, `private_key`, `issuer_name`, `key_name` as fields?

What's your thoughts on that?

fatima2003:
@cipherboy ooh yes serial numbers are not a good identifier 🫢

I like that idea and it's similar to the ssh endpoint as well. When I thought of this solution earlier, I wasn't sure if it solved the issue of reducing ambiguity when multiple issuers are imported since, currently, if you import one issuer in your bundle, you can grab the issuer reference from the returned `imported_issuers` safely.

That said, a dedicated single-issuer import endpoint sidesteps the whole ambiguity thing more explicitly so I think its a solid direction :))

cipherboy:
@fatima2003 Cool, let's go for it then :-D

cipherboy:
Hey @fatima2003 are you interested in finishing this? Hope all is well! :-)

PR #794, opened 2024-12-12
CEL for Certificate Issuance Policy

### LINKED ISSUE #753: RFC - CEL for Certificate Issuance Policy
## Summary

This feature proposes the use of [Common Expression Language (CEL)](https://github.com/google/cel-spec) in OpenBao's PKI plugin to create a flexible framework for administrators to craft custom certificate issuance policies. CEL will allow administrators to define fine-grained rules and dynamic templates for certificate issuance and signing, replacing the limitations of static role-based policies.

## Problem Statement

The [OpenBao role system](https://openbao.org/api-docs/secret/pki/#createupdate-role) allows operators to define constraints that users must meet to have certificates issued under a specific role. However, this role system is not expressive. Administrators cannot enforce complex constraints, such as requiring specific combinations of domains without modifying the codebase. This restricts flexibility and requires frequent code changes for new use cases.

The challenge is to provide a dynamic, user-configurable solution that can address these limitations while minimizing complexity and maintenance overhead.

## User-facing description

**_Administrators_** can define CEL policies under a new endpoint, `pki/cel/role/:name`, for example, a cel policy called `simple_policy` that requires 2 specific domains (example.com and example.org) to be present.

**_Administrators_** can use ACLs to restrict which CEL policies users/tokens can call.

**_Users_** can reference the new policy in a certificate issuance request using the new endpoints, `pki/cel/issue/:name` or `pki/cel/sign/:name`. When a user submits a request, the associated CEL policy validates the request. For example, a user can request a certificate by calling `pki/cel/issue/example-policy` with certificate parameters such as a `CSR` if this is a certificate signing request,`alt_names`, `key_type`, `key_bits` &c.

- If the validation is successful, it generates the certificate based on the approved request. For example, if a user requests a certificate with a `common_name` example.com, an `alt_name` example.org, the request will be successful.
-  If the validation fails, an error message is returned detailing why the request was denied. For example, if a user requests a certificate with a `common_name` of example.com, an `alt_name` of example.net, the request would fail because the domain example.net is not allowed in the policy and example.org has not been specified. 

### **API Endpoints**
The following new endpoints will be created for CEL-based certificate issuance and signing workflows:

| Use                                                 | Method   | Path                                    |
| ---------------------------- | ---------- | ------------------------ |
| List CEL policies                 | `LIST`       | `pki/cel/roles`              |
| Retrieve a specific CEL policy  | `GET`       | `pki/cel/roles/:name`  |
| Create a CEL policy                    | `POST`    | `pki/cel/role/:name`      |
| Update a CEL policy                   | `PATCH ` | `pki/cel/role/:name`     |
| Delete a CEL policy                    | `DELETE` |	`pki/cel/role/:name`    |
| Generate certificate and key  | `POST`   | `pki/cel/issue/:name`        |
| Sign certificate                           | `POST`   | `pki/cel/sign/:name`         |

**List CEL Policies** 
* This endpoint returns a list of available cel policies. Only the policy names are returned, not any values. It is useful to both operators and users.
* List with pagination will be implemented similarly to #678.

**Read CEL Policy**
* This endpoint fetches the full definition of the specified policy by its `name`, including its `validation_program`. It is useful for operators to manage policies and for users to understand policy configurations.

**Create a CEL Policy**
* Parameter `name` `(string: <required>)` specifies the name of the policy to be updated. This is part of the request URL.
* A policy consists of a `validation_program`, which evaluates incoming parameters to either approve or reject a request, and dynamically generates the certificate structure upon approval. 
* The response includes an error message, which is `nil` if the policy is successfully created or updated, or provides details if an error occurs.

**Update a CEL Policy**
* Similar to creating a policy, but in this case, the policy already exists. Only the parameters included in the request are updated, leaving any unspecified parameters unchanged.

**Generate Certificate and Key**
* This endpoint generates a new set of credentials (private key and certificate) based on the policy named in the endpoint. It takes in a set of parameter  like `common_name`, `alt_names`, &c similar to [pki/issue/:name](https://openbao.org/api-docs/secret/pki/#generate-certificate-and-key).The issuing CA certificate and full CA chain is returned as well, so that only the root CA need be in a client's trust store. Choice of issuing CA is determined by the policy.

**Sign Certificate**
* This endpoint signs a new certificate based upon the provided `CSR` and the supplied parameters (like `common_name`, `alt_names`, &c similar to [pki/sign/:name](https://openbao.org/api-docs/secret/pki/#sign-certificate)), subject to the restrictions contained in the policy named in the endpoint. The issuing CA certificate and the full CA chain is returned as well, so that only the root CA need be in a client's trust store.

_Access to these endpoints should be managed via ACLs. For instance, admin roles can create or update CEL policies, while operator roles can only list or retrieve them. Issuance and signing endpoints should also be restricted to roles with specific permissions._

### CEL Create Role Request Format
This defines the parameters the operator sends to create a CEL role.
- `name` `(string: required)`
A unique name for the CEL role name.
- `message` `(string: optional)`
Optional failure message. This is a static message, for a more customizable message use `message_expression`.
_E.g., "Required domain names not supplied."_
- `validation_program`
     - `variables` `(string: optional)`
A variable is a named expression that can be referred later in other variables and expressions. The order of variables matters since later variables can reference previous ones. This is useful if an expression is too complicated or has reusable parts.
     - `expressions` `(string: required)`
CEL expressions which are used to validate the request. If an expression evaluates to false, the validation check is enforced according to the `failure_policy` field. 

### CEL Request Format (Bao to CEL)
This defines the parameters that OpenBao sends to the CEL policy engine for validation and templating.

- `request_id` `(string: required)`
A unique identifier for tracking the request. The response must include this ID.
- `user_request_parameters` `(map[string]interface{}: required)`
Unvalidated parameters sent by the user. It is up to the CEL policy to validate these prior to using them. The following fields may be present, among others submitted by the user:
   * `csr` `(string)` A PEM-format Certificate Signing Request.
- `identity_request_values` `(map[string]interface{}: optional)`
Information about the user making the request:
   * `user_id` `(string)` Unique identifier for the user.
   * `groups` `(array[string])` Groups associated with the user.
- `bao_request_values` `(map[string]interface{}: optional)`
System-generated values validated by OpenBao before sending to CEL. These have higher trust than the unvalidated `user_request_parameters`.
    * `cel_role` `(string: required)` The CEL policy name that the request should validate against.   
   * `mount` `(string)` - The request's mount path as known by the PKI plugin.
   * `namespace` `(string)` Namespace of the request.
   * `requested_issuer`: Details about the requested issuer (name, ID, certificate).
       * `requested_issuer_name` `(string)` - Name of the user's requested issuer.
       * `requested_issuer_id` `(string)` - UUID of the user's requested issuer.
       * `requested_issuer_cert` `(string)` - PEM format certificate of the user's requested issuer.

### CEL Reply Format (CEL to Bao)
The response from the CEL engine indicates whether the request is approved or rejected, and provides the final certificate or error details.

- `request_id` `(string: required)` 
The unique identifier from the request. Used for tracking purposes.
- `success` `(bool: required)` 
Status of the request. True if the request was validated else false if it was rejected due to validation errors .
- `template` `(map: required)`
The signed certificate in PEM format. Only included if status is success.
- `issuer` `(string: optional)`
The issuer used to sign the certificate.
- `warnings` `(array[string]: optional)`
Warnings about the request or adjustments made by the CEL policy engine. 
_E.g., "common_name was empty so added example.com"_
- `error` `(string: optional)`
Detailed error message if status is failure.
- `generate_lease` `(string: optional)`
- `no_store` `(string: optional)`

### CEL Environment Functions and Extensibility
CEL supports injecting custom functions into the execution environment. This allows CEL policies to call application-defined functions to perform additional processing, validation, or transformations. For example, these functions could allow the CEL engine to parse inputs, decode values, or verify signatures.

#### Proposed Functions
* Base64 Operations
    * `base64_decode(string)` -> `bytes`
Decodes a base64-encoded string.
    * `base64_encode(bytes)` -> `string`
Encodes a byte array into a base64 string.
* JSON Parsing
    * `parse_json(string)` -> `map[string]interface{}`
Parses a JSON-encoded string into a CEL map for further evaluation.
* Hex Encoding/Decoding
    * `hex_decode(string)` -> `bytes`
Decodes a hex-encoded string into a byte array.
    * `hex_encode(bytes)` -> `string`
Encodes a byte array into a hex string.
* Key Operations
    * `verify_signature(bytes, signature, public_key)` -> `bool`
Verifies that the given signature matches the provided data using a specific public key.
* Validation Utilities
    * `is_domain(string)` -> `bool`
Checks if the string conforms to a valid domain format.

In some scenarios, CEL rules might need to make network requests for dynamic lookups or verifications (e.g., querying an external CRL, validating an identity). To support this safely:

* Fetch data from a given URL.
    * `http_get(url)` -> `string` 
**Security Consideration**: To avoid misuse or unnecessary network exposure, this functionality must be gated behind a global configuration parameter: `cel_http_requests_enabled`. If disabled, any CEL rule that attempts to use `http_get` would fail validation. **_Administrators_** must explicitly enable this feature to mitigate security risks.

### Edge Cases
* **Policy Execution Failures and Validation:** If the `validation_program` contains syntax errors or runtime issues, the system will validate these during policy creation or updates and deny the request with a detailed error message explaining the issue. 
* **Undefined Behavior:** Handling of unknown or unexpected parameters passed to CEL policies. CEL engine should ignore unrecognized parameters or handle them gracefully.

## Rationale and alternatives

Certain policies can be implemented using [roles](https://openbao.org/api-docs/secret/pki/#createupdate-role), for instance, `allow_localhost` but for more flexible scenarios, such as validating specific request metadata, or generating custom certificate templates based on runtime parameters, CEL policies would provide the necessary adaptability and control. For example, a CEL policy could enforce a rule that `alt_name` should contain `sale.org`.

 Alternatively, a web hook approach like [Vault's CIEPS](https://developer.hashicorp.com/vault/docs/secrets/pki/cieps) could be used to achieve similar flexibility. In this approach, requests are sent to an external service for validation and certificate generation logic, allowing for custom behaviors. However, relying on a web hook introduces additional network overhead, potential latency, and dependency on an external system's availability. CEL policies, in contrast, embed the logic directly into OpenBao, eliminating external dependencies on out-of-process mechanisms. 

[Kubernetes](https://kubernetes.io/docs/reference/using-api/cel/) also uses CEL to declare validation rules, policy rules, and other constraints or conditions. They have really nice documentation of it [here](https://kubernetes.io/docs/reference/using-api/cel/#language-overview). I would like to add something like this to our docs too after the feature is implemented :)

## Downsides

Downgrading to a previous version of OpenBao or upstream wouldn’t break anything but new paths will be unreachable.

## Security Implications

There is no known increase/decrease in OpenBao's security.

## User/Developer Experience

This feature will not affect current users/developers or their existing setups. They are not required to make changes unless they wish to adopt this new functionality.

## Unresolved Questions
N/a

**Previously Unresolved Question**
>  I am a little uncertain on whether the `template` for generating the certificate should be explicitly provided by the operator or if the `policy_program` with CEL expressions alone can provide enough flexibility and simplicity. In my opinion, the template shouldn’t need to be provided by the operator, as CEL rules should be sufficient to fully define and customize the rules and outputs for the certificate generation process. Ideally, the template would be “automatically” filled in by the CEL engine based on the defined rules.

**Resolution** Only 1 input parameter should be sufficient.

## Related Issues
n/a

## Proof of Concept

WIP PR #794

fatima2003:
Resolves #753

**TO DO**
- [X] Create paths and basic CRUD functions.
- [X] Test CRUD functions.
- [X] Import CEL and allow creation and validation of simple CEL rules.
- [X] Implement `Variables` for CEL expressions.
- [X] Create `pki/cel/issue/:name` 
- [x] Create `pki/cel/sign/:name`
- [x] Comprehensive tests.

**Follow up PR's**
* Add global config flag for HTTP request functions.
* Docs.
---

**TO DO AFTER REVISIONS**
- [X] Protobuf representation of X509 certs.
- [X] Translation helpers for proto representation to X509 cert.
- [X] Generate/ sign cert template directly, without creation bundle.


builtin/logical/pki/path_issue_sign_test.go
```diff
@@ -0,0 +1,605 @@
...
+				{
+					"name":       "b",
+					"expression": "has(request.common_name) && request.common_name == 'example.com'",
+				},
+				{
+					"name":       "a",
+					"expression": "b && has(request.ip_sans) && size(request.ip_sans) > 0",
+				},
+			},
+			"expressions": "a",
```
cipherboy, line 91:
```suggestion
				{
					"name":       "validate_cn",
					"expression": "has(request.common_name) && request.common_name == 'example.com'",
				},
				{
					"name":       "cn_value",
					"expression": "request.common_name",
				},
				{
					"name":       "dns_san_value",
					"expression": "!request.dns_sans.include("my.example.com") ? [request.common_name, request.dns_sans..., "my.example.com"] : [request.common_name, request.dns_sans...]",
				},
				{
				  "name": "dns_san_missing_value",
				  "expression": "request.dns_sans.include("my.example.com") ? "" : "adding my.example.com "",
				},
			},
			"expressions": "
			 {
			  "request_id": env.request_id,
			  "success": validate_cn,
			  "issuer": default,
			  "warnings": dns_san_missing_value,
			  "certificate": validate_cn ? { "subject": { "common_name" } } : null,
			 }
			",
```

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+}
+
+message ObjectIdentifier {
+  // 1.2.840.113549 → [1, 2, 840, 113549]
+  repeated uint32 arc = 1;
+}
+
+message IPMask {
+    bytes IPMask = 1;
+}
```
cipherboy, line 36: Should this live in `net`? https://pkg.go.dev/net#IPMask

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+    }
+    message IPNet {
+        net.IP IP = 1;          // network number
+	    IPMask Mask = 2;
+    }
+}
+
+message Userinfo {
+	string username      = 1;   // RFC‑2396 allows empty.
+    string password      = 2;   // Empty when not supplied.
```
cipherboy, line 50: Formatting seems a little weird in this file. Wonder if we have an auto-formatter hiding somewhere?

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+	DSAWithSHA256 = 8;          // Unsupported.
+	ECDSAWithSHA1 = 9;          // Only supported for signing, and verification of CRLs, CSRs, and OCSP responses.
+	ECDSAWithSHA256 = 10;
+	ECDSAWithSHA384 = 11;
+	ECDSAWithSHA512 = 12;
+	SHA256WithRSAPSS = 13;
+	SHA384WithRSAPSS = 14;
+	SHA512WithRSAPSS = 15;
+	PureEd25519 = 16;
+}
```
cipherboy, line 106: I think we can drop this field; it will be determined by PKI internally (+/- `use_pss` as a flag on the role) and won't really be a CEL-author decision. Half of the value is determined by the issuer key type anyways.

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+	OID SubjectDomainPolicy = 2;
+}
+
+enum PublicKeyAlgorithm {
+    UnknownPublicKeyAlgorithm = 0;
+    RSA = 1;
+    DSA = 2;
+    ECDSA = 3;
+    Ed25519 = 4;
+}
```
cipherboy, line 86: Same here, I think we can drop this. This is entirely determined by the specified CSR (or key generation parameters).

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+	ExtKeyUsageMicrosoftCommercialCodeSigning = 12;
+	ExtKeyUsageMicrosoftKernelCodeSigning = 13;
+}
+
+// Mirrors x509.Certificate
+message CertTemplate {
+    // Complete ASN.1 DER content (certificate, signature algorithm and signature).
+    bytes Raw = 1; 
+	bytes RawTBSCertificate = 2; 
+	bytes RawSubjectPublicKeyInfo  = 3;
```
cipherboy, line 143: In general, there's a few `Raw` values we'll keep but most we'll drop. 

Let's drop these three. The first two are determined by the rest of the structure and the last is determined by the provided CSR and/or generated key.

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// Mirrors x509.Certificate
+message CertTemplate {
+    // Complete ASN.1 DER content (certificate, signature algorithm and signature).
+    bytes Raw = 1; 
+	bytes RawTBSCertificate = 2; 
+	bytes RawSubjectPublicKeyInfo  = 3; 
+	bytes RawSubject = 4; 
+	bytes RawIssuer = 5;
```
cipherboy, line 145: This value can also be dropped, it is determined by the issuing CA's `RawSubject` value and modifying it from CEL would break signature verification.

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+message CertTemplate {
+    // Complete ASN.1 DER content (certificate, signature algorithm and signature).
+    bytes Raw = 1; 
+	bytes RawTBSCertificate = 2; 
+	bytes RawSubjectPublicKeyInfo  = 3; 
+	bytes RawSubject = 4; 
+	bytes RawIssuer = 5; 
+
+	bytes Signature = 6; 
+    SignatureAlgorithm SignatureAlgorithm = 7;
```
cipherboy, line 148: Signature and its algorithm will be determined by the rest of the engine, and so can drop.

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+
+	PublicKeyAlgorithm PublicKeyAlgorithm = 8;
+	google.protobuf.Any PublicKey = 9;
```
cipherboy, line 151: These too will be determined by the CSR/generation capabilities and so can be dropped.

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+
+	int64 Version = 10;
+	bytes SerialNumber = 11;
+    PKIX.Name Issuer = 12;
```
cipherboy, line 155: As above about Issuer.

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+	bytes RawIssuer = 5; 
+
+	bytes Signature = 6; 
+    SignatureAlgorithm SignatureAlgorithm = 7;
+
+	PublicKeyAlgorithm PublicKeyAlgorithm = 8;
+	google.protobuf.Any PublicKey = 9;
+
+	int64 Version = 10;
+	bytes SerialNumber = 11;
```
cipherboy, line 154: Serial Number could also be dropped; we'll be generating a fully random one in `certutil`.

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+    PKIX.Name Issuer = 12;
+    PKIX.Name Subject = 13;
+    google.protobuf.Timestamp NotBefore = 14;
+    google.protobuf.Timestamp NotAfter  = 15;
+    KeyUsage KeyUsage = 16;
+    
+    repeated PKIX.Extension Extensions = 17;
+	repeated PKIX.Extension ExtraExtensions = 18;
+    repeated ObjectIdentifier UnhandledCriticalExtensions = 19;
```
cipherboy, line 163: I think we can drop these two fields. If you read the comments on https://pkg.go.dev/crypto/x509#Certificate:

```
// Extensions contains raw X.509 extensions. When parsing certificates,
	// this can be used to extract non-critical extensions that are not
	// parsed by this package. When marshaling certificates, the Extensions
	// field is ignored, see ExtraExtensions.
	Extensions [][pkix](https://pkg.go.dev/crypto/x509/pkix).[Extension](https://pkg.go.dev/crypto/x509/pkix#Extension)
```

and

```
// UnhandledCriticalExtensions contains a list of extension IDs that
	// were not (fully) processed when parsing. Verify will fail if this
	// slice is non-empty, unless verification is delegated to an OS
	// library which understands all the critical extensions.
	//
	// Users can access these extensions using Extensions and can remove
	// elements from this slice if they believe that they have been
	// handled.
	UnhandledCriticalExtensions [][asn1](https://pkg.go.dev/encoding/asn1).[ObjectIdentifier](https://pkg.go.dev/encoding/asn1#ObjectIdentifier)
```

So really only `ExtraExtensions` is actually put on the final certificate, but I don't know why we'd keep it as `Extra` and not just `Extensions` :thinking: 

Thoughts?

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+	repeated ObjectIdentifier UnknownExtKeyUsage = 21;
+
+	bool BasicConstraintsValid =22;
+	bool IsCA = 23;
+
+    int64 MaxPathLen = 24;
+    bool MaxPathLenZero = 25;
+
+    bytes SubjectKeyId = 26;
+	bytes AuthorityKeyId = 27;
```
cipherboy, line 175: As above about issuer material.

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+	repeated ExtKeyUsage ExtKeyUsage = 20;
+	repeated ObjectIdentifier UnknownExtKeyUsage = 21;
+
+	bool BasicConstraintsValid =22;
+	bool IsCA = 23;
+
+    int64 MaxPathLen = 24;
+    bool MaxPathLenZero = 25;
+
+    bytes SubjectKeyId = 26;
```
cipherboy, line 174: SKID has a few calculation modes, so let's leave this one. :-D

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+	bytes AuthorityKeyId = 27;
+
+    // RFC 5280, 4.2.2.1 (Authority Information Access)
+	repeated string OCSPServer = 28;
+	repeated string IssuingCertificateURL = 29;
```
cipherboy, line 179: These are set by PKI from `config/urls` and so can be dropped.

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,211 @@
...
+	repeated string PermittedDNSDomains = 35;
+	repeated string ExcludedDNSDomains = 36;
+	repeated net.IPNet PermittedIPRanges = 37;
+	repeated net.IPNet ExcludedIPRanges = 38;
+	repeated string PermittedEmailAddresses = 39;
+	repeated string ExcludedEmailAddresses = 40;
+	repeated string PermittedURIDomains = 41;
+	repeated string ExcludedURIDomains = 42;
+    
+	repeated string CRLDistributionPoints = 43;
```
cipherboy, line 197: Same thing with this one.

cipherboy (commented):
Some comments to make your conversions easier :-)

builtin/logical/pki/path_cel.go
```diff
@@ -0,0 +1,473 @@
...
+		return nil, nil
+	}
+
+	resp := &logical.Response{
+		Data: role.ToResponseData(),
+	}
+	return resp, nil
+}
+
+func (b *backend) pathCelRolePatch(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
```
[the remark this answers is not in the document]
fatima2003: We can patch a single variable without clearing the others 👀

See [path_cel_test.go](https://github.com/openbao/openbao/pull/794/files#diff-e664aadbd5ed16c4477a9ca9315cde10d32187ab362bbc277d782e68b5cfe9b2R109-R126)

builtin/logical/pki/cert_template.proto
fatima2003: Thank you, lemme fix :) I tried checking the whitespaces but was just free balling with highlight so might've missed them..

builtin/logical/pki/path_issue_sign.go
```diff
@@ -583,6 +797,284 @@ func (b *backend) pathIssueSignCert(ctx context.Context, req *logical.Request, d
...
+		// Translate to *x509.Certificate struct.
+		cert, err = CertProtoToX509(certTemp)
+		if err != nil {
+			return nil, err
+		}
+
+	case string:
+		// The CEL program decided the request is invalid and returned a human-readable error string.
+		return nil, fmt.Errorf("%s", v)
+
```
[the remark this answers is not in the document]
cipherboy: Right, I saw that in the JWT engine too. What happens is we also later reject the request, but with a CEL programming error rather than an obtuse permission denied error.

builtin/logical/pki/path_cel.go
```diff
@@ -0,0 +1,473 @@
...
+	// Validate the main CEL expression
+	ast, issues := env.Parse(validationProgram.Expressions.MainProgram)
+	if issues != nil && issues.Err() != nil {
+		return false, fmt.Errorf("invalid CEL syntax for main expression: %v", issues.Err())
+	}
+
+	// Ensure the AST is non-nil
+	if ast == nil {
+		return false, fmt.Errorf("failed to compile CEL main expression: AST is nil")
+	}
```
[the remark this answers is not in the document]
fatima2003: good catch!

builtin/logical/pki/cert_template.proto
fatima2003: Interesting, how about we drop `Extensions` and `UnhandledCriticalExtensions` and keep `ExtraExtensions` as is... idk if it'd be less surprising to have `ExtraExtensions` as `Extensions` or that `Extensions` is usually provided by Der and ignored in issuing 🧐

Perhaps a comment: 
```
// ExtraExtensions are copied verbatim into the certificate when it is
// issued. They are not populated when parsing an existing cert.
```

builtin/logical/pki/cert_template.proto
```diff
@@ -0,0 +1,175 @@
...
+        net.IP IP = 1;          // network number
+        IPMask Mask = 2;
+    }
+    message IPMask {
+        bytes IPMask = 1;
+    }
+}
+
+message Userinfo {
+	string username      = 1;   // RFC‑2396 allows empty.
```
cipherboy: @fatima2003 Sorry, it looks like you got just about all of them but one! This one still uses a tab rather than four spaces! :(

builtin/logical/pki/path_cel.go
```diff
@@ -0,0 +1,381 @@
...
+	// Warnings about the request or adjustments made by the CEL policy engine.
+	// E.g., "common_name was empty so added example.com"
+	Warnings string
+
+	// Specifies if certificates issued/signed against this role will have OpenBao leases attached to them.
+	GenerateLease string
+	// If set, certificates issued/signed against this role will not be stored in the storage backend.
+	NoStore string
+	// The issuer used to sign the certificate.
+	Issuer string
```
cipherboy:
```suggestion
```

These are handled by the new `ValidationOutput` fields :-)

builtin/logical/pki/path_cel.go
```diff
@@ -0,0 +1,381 @@
...
+	// CSR ignored if certificate is being issued.
+	CSR string
```
cipherboy: I'm curious what this field does? I would expect the CSR to be passed from the `/sign` endpoint from the client's request, not via role parameter.

builtin/logical/pki/path_cel.go
```diff
@@ -0,0 +1,381 @@
...
+)
+
+type CELRoleEntry struct {
+	// Required, the name of the role
+	Name string `json:"name"`
+	// Required, defines validation logic
+	CelProgram celhelper.CelProgram `json:"validation_program"`
+	// Warnings about the request or adjustments made by the CEL policy engine.
+	// E.g., "common_name was empty so added example.com"
+	Warnings string
```
cipherboy: I like `Warnings`, but maybe we just want to make it `repeated string warnings = 7` or something on the above return type? Then we are down to just a pair of fields here and it is simple to understand. Just a thought :-)

builtin/logical/pki/path_cel.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	resp := &logical.Response{
+		Data: role.ToResponseData(),
+	}
+	return resp, nil
+}
+
+func (b *backend) pathCelRolePatch(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+	txRollback, err := logical.StartTxStorage(ctx, req)
```
cipherboy, line 219: :partying_face:

builtin/logical/pki/path_cel.go
```diff
@@ -0,0 +1,381 @@
...
+	pathCelRoleHelpSyn  = `Manage the cel roles that can be created with this backend.`
+	pathCelRoleHelpDesc = `This path lets you manage the cel roles that can be created with this backend.`
+)
+
+func (r *CELRoleEntry) ToResponseData() map[string]interface{} {
+	return map[string]interface{}{
+		"name":        r.Name,
+		"cel_program": r.CelProgram,
+	}
+}
```
cipherboy, line 327: @fatima2003 Ah I see, you had already removed the other fields here, so `warnings` should probably be directly added to the `ValidationOutput` message and then removed as parameter. :-)

builtin/logical/pki/path_cel_test.go
```diff
@@ -0,0 +1,324 @@
...
+				return
+			}
+
+			// Assert the result matches the expected result
+			if result.Value().(bool) != tt.expectedResult {
+				t.Fatalf("Expected result '%v', but got '%v'", tt.expectedResult, result)
+			}
+		})
+	}
+}
```
cipherboy, line 337: :partying_face: Nice tests!

sdk/helper/cel/program.go
```diff
@@ -21,7 +24,7 @@ type CelVariable struct {
 	// Name of the variable.
 	Name string `json:"name"`
 	// CEL expression for the variable
-	Expression string `json:"explression"`
+	Expression string `json:"expression"`
```
cipherboy, line 27: :blue_heart: Wow, I totally missed that. Thanks for fixing!

sdk/helper/cel/program.go
```diff
@@ -72,3 +75,74 @@ func ParseCompileAndEvaluateExpression(env *cel.Env, expression string, evaluati
...
+	}
+
+	// Validate the main CEL expression
+	ast, issues := env.Parse(celProgram.Expression)
+	if issues != nil && issues.Err() != nil {
+		return false, fmt.Errorf("invalid CEL syntax for main expression: %v", issues.Err())
+	}
+
+	// Ensure the AST is non-nil
+	if ast == nil {
```
cipherboy: @fatima2003 Sorry, I think you copied this from the other place and forgot to remove :D

cipherboy: Like the refactor though!

ui/app/styles/app.scss
```diff
@@ -12,7 +12,9 @@
 @mixin font-face($name) {
   @font-face {
     font-family: $name;
-    src: url('/ui/fonts/#{$name}.woff2') format('woff2'), url('/ui/fonts/#{$name}.woff') format('woff');
+    src:
```
cipherboy, line 15: Not sure why these came in?  Same with two other files below.

cipherboy:
@fatima2003 Looks like test 2 and 8 are failing in CI with legitimate errors. :-)

builtin/logical/pki/path_cel.go
fatima2003: oops, I removed em now 😅

builtin/logical/pki/cert_template.proto
fatima2003: ha!

builtin/logical/pki/path_cel.go
fatima2003: Good point... I was putting it on the CEL Author to explicitly do this:
```
CSR: request.csr
```
But it doesn't have to be so, fixed it :-)

ui/app/styles/app.scss
fatima2003: hmm, I think I may have run some type of lint on my lappy before caz the changes look like formatting changes. Lemme revert!

fatima2003: Weird, I reverted and pushed but the changes 'vanished' in my latest commit...

Found the culprit:
```
fatima@fattiesPatties-2 openbao % git commit --signoff                                                   
ui/app/styles/app.scss
ui/app/styles/components/console-ui-panel.scss
ui/app/styles/core/checkbox-and-radio.scss
==> Changes detected in ui/: Running linter...
⚠ Some of your tasks use `git add` command. Please remove it from the config since all modifications made by tasks will be automatically added to the git commit index.

✔ Backed up original state in git stash (5a1d611f6)
✔ Running tasks for staged files...
✔ Applying modifications from tasks...
✔ Cleaning up temporary files...
==> Checking that code complies with gofmt requirements...
Checking changed files...
[cel 154425cb4] Minor fixes
 6 files changed, 217 insertions(+), 279 deletions(-)
```

fatima2003:
@cipherboy the CI is clean now! :D 

Test 9 is flakey, check [previous run](https://github.com/openbao/openbao/actions/runs/15374505841/job/43257770962 ).

builtin/logical/pki/cert_template.proto
cipherboy: Yep, exactly what I've done in my follow-up which I'll open shortly.

PR #805, opened 2024-12-18
Enhance tests

klaus-sap:
just for practice


vault/util_test.go
```diff
@@ -3,19 +3,40 @@
...
+	"testing"
+)
+
+const n = 1024
 
 func TestMemZero(t *testing.T) {
-	b := []byte{1, 2, 3, 4}
+	r := rand.Intn(n)
+	b := randbytes(r)
 	memzero(b)
```
cipherboy, line 16: TBH, given the rather extensive disclaimer around this function:

https://github.com/openbao/openbao/blob/88945c0791e71caa61883ba0851e856e4955c076/vault/util.go#L11-L29

I wonder if it is better to simply remove it? memguard is probably preferable: https://github.com/awnumar/memguard -- but it'd be great if this was a native part of the Go toolchain.

klaus-sap: We may discuss your proposal (or using https://github.com/awnumar/memcall for storing key material?), but this PR is about tests only :-)

cipherboy: @DanGhita You might also have opinions about removing this. :-) for the future, though!

vault/util_test.go
```diff
@@ -3,19 +3,40 @@
...
-	if b[0] != 0 || b[1] != 0 || b[2] != 0 || b[3] != 0 {
-		t.Fatalf("bad: %v", b)
+	if len(b) != r {
+		t.Fatalf("buffer has wrong length: %d", len(b))
+	}
+	for i := range b {
```
cipherboy, line 20:
```suggestion
	for i, v := range b {
```

Also works to get the value without having to reindex below.

vault/util_test.go
```diff
@@ -3,19 +3,40 @@
...
+import (
+	"math/rand"
+	"testing"
+)
+
+const n = 1024
 
 func TestMemZero(t *testing.T) {
-	b := []byte{1, 2, 3, 4}
+	r := rand.Intn(n)
```
cipherboy, line 14: I don't think it matters but doesn't this have a 1/1024 chance of being zero-length?

cipherboy (approved):
This is fine, though perhaps less necessary.

PR #811, opened 2024-12-20
Add support for ECDH key agreement in Transit Secret Engine

### LINKED ISSUE #655: Add support for ECDH key agreement in Transit Secret Engine
### Summary

Transit Secret Engine currently supports ECC keys for signature purpose, but it does not support ECDH key agreement allowing Alice and Bob to derive a shared secret.

### Problem Statement

This feature would allow  different OpenBao instances to exchange data protected with keys derived from a shared secret. The security of the exchanged data would not rely solely on the TLS security. 

### User-facing description

When deploying a hierarchical structure of OpenBao instances, this feature allows to setup a secure channel between two OpenBao instances (Alice = instance 1, Bob = instance 2, for example).

### Technical Description

The Transit API should support new ECC key types for ECDH: ecdh-p256, ecdh-p384, ecdh-p521.

When creating a ECDH key, the API should support an attribute for specifying whether the new key is ephemeral or not (i.e. persistent, if not ephemeral).

The Transit API should support a new path ("derive") to be used with the new key type ECDH for deriving a shared secret.
The derivation endpoint should support the creation of a symmetric key.
A specific attribute of the API should indicate whether this derived key is ephemeral or not.

This is a new feature that should not break the compatibility with the existing Transit features.

### Downsides

_No response_

### Security Implications

_No response_

### User/Developer Experience

_No response_

### Unresolved Questions

_No response_

### Related Issues

The initial discussion around supporting ECDH is here: #556

### Proof of Concept

_No response_

DanGhita:
Resolves #655 

## Target Release

1.14.7


builtin/logical/transit/path_derive_key.go
```diff
@@ -0,0 +1,160 @@
...
+				Type:    framework.TypeString,
+				Default: "aes256-gcm96",
+				Description: `
+The type of the output derived key. Currently, "aes128-gcm96" , "aes256-gcm96", "chacha20-poly1305", "xchacha20-poly1305" are supported.
+Defaults to "aes256-gcm96".
+`,
+			},
+		},
+
+		Callbacks: map[logical.Operation]framework.OperationFunc{
```
cipherboy, line 70: Do you mind using `Operations:` ? `Callbacks:` is deprecated: https://github.com/openbao/openbao/blob/main/sdk/framework/path.go#L109

cipherboy: This is addressed in #812 so there's no more examples of `Callbacks:`. :-)

builtin/logical/transit/path_derive_key.go
```diff
@@ -0,0 +1,165 @@
+// Copyright (c) HashiCorp, Inc.
```
cipherboy, line 1:
```suggestion
// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
```

This is net-new code that you've introduced.

sdk/helper/keysutil/policy.go
```diff
@@ -161,6 +162,14 @@ func (kt KeyType) DerivationSupported() bool {
 	return false
 }
 
+func (kt KeyType) KeyAgreementSupported() bool {
```
cipherboy, line 165: Little surprised you don't support Curve25519 EdDH derivation :-)

sdk/helper/keysutil/policy.go
```diff
@@ -882,6 +891,87 @@ func (p *Policy) DeriveKey(context, salt []byte, ver int, numBytes int) ([]byte,
...
+	ownPrivateKey := &ecdsa.PrivateKey{
+		PublicKey: ecdsa.PublicKey{
+			Curve: curve,
+			X:     baseKeyEntry.EC_X,
+			Y:     baseKeyEntry.EC_Y,
+		},
+		D: baseKeyEntry.EC_D,
+	}
+
+	peerPublicKey, err := certutil.ParsePublicKeyPEM(peerPublicKeyPem)
```
cipherboy, line 940: Nice! Glad you are using this helper.

builtin/logical/transit/path_derive_key_test.go
```diff
@@ -0,0 +1,240 @@
+// Copyright (c) HashiCorp, Inc.
```
cipherboy, line 1: Here as well.

builtin/logical/transit/path_derive_key.go
```diff
@@ -0,0 +1,165 @@
...
+
+		Fields: map[string]*framework.FieldSchema{
+			"key_derivation_algorithm": {
+				Type:    framework.TypeString,
+				Default: defaultKeyDerivationAlgorithm,
+				Description: `Key derivation algorithm to use. Valid values are:
+* ecdh
+Defaults to "ecdh".`,
+			},
+
```
cipherboy, line 36: Maybe there needs to be a `kdf` parameter, which accepts `""` or `"hkdf"`, and if using HKDF, can add HKDF-specific parameters (like salt or context)? (to answer your TODO below). 

Then HKDF can also be a key type potentially, and we can do multiple HKDF invocations (without doing multiple ECDH invocations), if they don't want to use the shared secret directly.

builtin/logical/transit/path_derive_key_test.go
```diff
@@ -0,0 +1,240 @@
...
+		t.Fatalf("err:%v resp:%#v", err, resp)
+	}
+
+	if resp.Data["plaintext"] != plaintext {
+		t.Fatalf("bad: plaintext. Expected: %q, Actual: %q", plaintext, resp.Data["plaintext"])
+	}
+
+}
+
+func TestTransit_EncryptWithRSAPublicKeyDGH(t *testing.T) {
```
cipherboy, line 197: I don't quite know why you added this test? I think it is covered by others and not related to this PR.

sdk/helper/keysutil/policy.go
```diff
@@ -882,6 +891,87 @@ func (p *Policy) DeriveKey(context, salt []byte, ver int, numBytes int) ([]byte,
...
+	if err != nil {
+		return nil, errutil.UserError{Err: fmt.Sprintf("failed to parse peer public key PEM: %s ", err.Error())}
+	}
+
+	if ownPrivateKey.Curve != peerPublicKey.(*ecdsa.PublicKey).Curve {
```
cipherboy, line 945: With thanks to @fatima2003, this will crash if `peerPublicKey` cannot be safely converted. E.g., if you pass in a RSA public key in the PEM. :-)

sdk/helper/keysutil/policy.go
cipherboy, line 940:
```suggestion
	peerPublicKeyRaw, err := certutil.ParsePublicKeyPEM(peerPublicKeyPem)
```

sdk/helper/keysutil/policy.go
```diff
@@ -882,6 +891,87 @@ func (p *Policy) DeriveKey(context, salt []byte, ver int, numBytes int) ([]byte,
...
+			Y:     baseKeyEntry.EC_Y,
+		},
+		D: baseKeyEntry.EC_D,
+	}
+
+	peerPublicKey, err := certutil.ParsePublicKeyPEM(peerPublicKeyPem)
+	if err != nil {
+		return nil, errutil.UserError{Err: fmt.Sprintf("failed to parse peer public key PEM: %s ", err.Error())}
+	}
+
```
cipherboy, line 944:
```suggestion

	peerPublicKey, ok := peerPublicKeyRaw.(*ecdsa.PublicKey)
	if !ok {
		return nil, errutil.UserError{Err: "supplied peer public key is not an ECDSA key"}
	}

```

cipherboy (commented):
Some comments!

builtin/logical/transit/path_derive_key_test.go
cipherboy: @DanGhita I agree that's the description of `TestTransit_ECDH_NominalCase1`. But `TestTransit_EncryptWithRSAPublicKeyDGH` just imports a RSA key (line 217) and encrypts some data with it (line 230). Am I mistaken?

DanGhita: Oh, sorry, you are right, TestTransit_EncryptWithRSAPublicKeyDGH is to be removed, it's a silly copy -paste

sdk/helper/keysutil/policy.go
DanGhita: Strange, my previous comment seems to be lost  :(
I was saying: OK, I will try to add KeyType_ED25519, since they are pretty used in the field.
However, if I'm not wrong, they are considered as less secure than the other curves; for example, ANSSI (French Security Agency) considers ED25519 just as an "acceptable" alternative.
Furthermore, digging into the subject, I found this in the go crypto library (so, probably I should change the current ECDH implementation):
[https://pkg.go.dev/crypto/elliptic#P521:~:text=Direct%20use%20of%20this%20package%20is%20deprecated%2C%20beyond%20the%20P224%2C%20P256%2C%20P384%2C%20and%20P521%20values%20necessary%20to%20use%20crypto/ecdsa.%20Most%20other%20uses%20should%20migrate%20to%20the%20more%20efficient%20and%20safer%20crypto/ecdh](url)

cipherboy: NIST P-Curves are vulnerable to certain problems that Ed25519 isn't: https://safecurves.cr.yp.to/

NIST recently certified EdDSA at the same level as ECDSA in FIPS 186-5.

fatima2003: I think switching to `crypto/ecdh` would be a good move since it supports both NIST P-Curves and Ed25519.

DanGhita: Yes, that was my idea too, I will play a little bit with the crypto/ecdh API and if everything works fine, I will update my implementation accordingly. Thank you @fatima2003 !

DanGhita: Hi @cipherboy and  @fatima2003,

After digging into the subject, I have my doubts about using Ed25519 for ECDH.
Actually, the "standard" method would be to use [Curve25519 for ECDH](https://crypto.stackexchange.com/questions/27866/why-curve25519-for-encryption-but-ed25519-for-signatures).

While using Ed25519  for ECDH seems to be possible (and secure), it would require a "manual" implementation. Indeed, the [ecdh package](https://pkg.go.dev/crypto/ecdh#section-documentation) implements ECDH for Montgomery Curve25519.
On the other hand, in OpenBao the KeyType_ED25519 keys are (slightly) different from  Curve25519 keys (mainly because of the ED25519 [keys clamping](https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about/)).

Any thoughts ?

cipherboy: Ahh sorry. Yes, let's add a new keytype Curve25519 and do that in a subsequent commit. Thanks @DanGhita, my memory was bad.

cipherboy:
@DanGhita Do you mind running `make fmt` and [adding a changelog entry](https://github.com/openbao/openbao/blob/main/changelog/README.md)? Thanks!

sdk/helper/keysutil/policy.go
```diff
@@ -882,6 +891,100 @@ func (p *Policy) DeriveKey(context, salt []byte, ver int, numBytes int) ([]byte,
...
+		ownPrivateKey := &ecdsa.PrivateKey{
+			PublicKey: ecdsa.PublicKey{
+				Curve: curve,
+				X:     baseKeyEntry.EC_X,
+				Y:     baseKeyEntry.EC_Y,
+			},
+			D: baseKeyEntry.EC_D,
+		}
+
+		if ownPrivateKey.Curve != peerPublicKey.Curve {
```
fatima2003, line 951: Nice! I tried a few invalid cases and they err perfectly.

builtin/logical/transit/path_derive_key.go
cipherboy: This can be done in a later PR as we make HKDF a first-class operation.

builtin/logical/transit/path_derive_key.go
```diff
@@ -0,0 +1,165 @@
...
+				Callback: b.pathPolicyDeriveKeyWrite,
+			},
+		},
+
+		HelpSynopsis:    pathDeriveKeyHelpSyn,
+		HelpDescription: pathDeriveKeyDesc,
+	}
+}
+
+func (b *backend) pathPolicyDeriveKeyWrite(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
```
cipherboy, line 80: @DanGhita Since #956 merged, can you add the transaction helpers to this API? Thanks!

builtin/logical/transit/path_derive_key_test.go
```diff
@@ -0,0 +1,218 @@
...
+	transit_ECDH_NominalCase(t, "ecdsa-p384", "aes256-gcm96")
+	transit_ECDH_NominalCase(t, "ecdsa-p521", "aes256-gcm96")
+
+	transit_ECDH_NominalCase(t, "ecdsa-p256", "chacha20-poly1305")
+	transit_ECDH_NominalCase(t, "ecdsa-p384", "chacha20-poly1305")
+	transit_ECDH_NominalCase(t, "ecdsa-p521", "chacha20-poly1305")
+
+	transit_ECDH_NominalCase(t, "ecdsa-p256", "xchacha20-poly1305")
+	transit_ECDH_NominalCase(t, "ecdsa-p384", "xchacha20-poly1305")
+	transit_ECDH_NominalCase(t, "ecdsa-p521", "chacha20-poly1305")
```
cipherboy, line 32:
```suggestion
	transit_ECDH_NominalCase(t, "ecdsa-p521", "xchacha20-poly1305")
```

cipherboy (commented):
@DanGhita Small things worth addressing, IMHO, and then I'll get it in beta :-)

builtin/logical/transit/path_derive_key.go
DanGhita: Yes, the next version of the API should handle this. I was also wondering if the key version (of the derived, resulting key) could be a good enough salt for the hkdf function (and thus Alice and Bob should agree on the same version of the derived key). But I'm not sure this is a good idea...

cipherboy: I don't know that I'd choose an arbitrary value. I'd let the user control these values, so that they can interoperate with other, non-Transit use cases.

cipherboy (approved):

PR #812, opened 2024-12-21
Use semgrep to enforce callbacks deprecation

cipherboy:
`framework.Path` has a deprecated field, `Callbacks` in favor of `Operations`, which better supports OpenAPI schemas. Remove all deprecated usage of `Callbacks` and have semgrep enforce its removal. 

Eventually we can annotate these new Operations with response typing information so the generated OpenAPI spec is more correct.


JanMa (approved):
LGTM :heavy_check_mark:

cipherboy:
Thanks, @JanMa!

PR #814, opened 2024-12-21
remove target version from PR template

JanMa:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -17,16 +17,3 @@ All Pull Requests must have an issue attached to them
 -->
 
 Resolves #
-
-## Target Release
-
-<!--
-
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -29,4 +29,4 @@ all targeted releases.
 
 -->
 
-1.14.7
+2.2.0
```
cipherboy, line 32: Do we just want to remove this? I think this matters more when we have PRs coming in from lots of people that depend on landing in exact versions, but so far we just set milestone+`blocker` label for those that do.

JanMa: Fine by me :+1:

cipherboy (approved):

PR #816, opened 2024-12-22
Ignore overlapping -config file/dir options passed to server

cipherboy:
When using the official container image with no arguments, OpenBao automatically invokes the server as:

```
$ bao server -config=/openbao/config "$@"
```

A user might reasonably expect to mount their configuration into /openbao/config, perhaps using a filename ending in .hcl or .json, and pass -config=/openbao/config/my-config.hcl, not expecting it to be auto-loaded. When this occurs, a strange error such as:

> bind: cannot assign requested address

occurs, because there are two identical listener configurations with the same address: one from loading the configuration file as a directory and one from adding it as a config file.

We opt to not entirely deduplicate arguments, but do warn if a server uses both a config directory and a config path.

This is a potentially breaking change: while ideal for listeners, if a user has a layout such as:

 - `config/00-ui-enable.hcl`
 - `config/01-ui-disable.hcl`

and invokes the CLI as:

```
$ bao server -config=config/ -config=00-ui-enable.hcl
```

before this change, the UI would be enabled (due to the last version winning), but after this change, the UI would be disabled. It is suggested admins update the configuration for clarity.

---

> ```
> $ podman run --replace --name vault  --mount=type=bind,src=./config.hcl,dst=/openbao/config/config.hcl,chown=true,relabel=shared --mount=type=bind,src=/home/cipherboy/GitHub/cipherboy/openbao/bin/bao,dst=/bin/bao,chown=true,relabel=shared   openbao/openbao:latest   server -config=/openbao/config/config.hcl 
> WARNING: ignoring duplicate configuration found in directory: /openbao/config/config.hcl
> ...
> ```

### DIFF
--- a/changelog/816.txt
+++ b/changelog/816.txt
@@ -1,0 +1,3 @@
+```release-note:change
+command/server: Prevent and warn about loading of duplicate config file from config directory.
+```
--- a/command/server.go
+++ b/command/server.go
@@ -372,17 +372,23 @@ func (c *ServerCommand) parseConfig() (*server.Config, []configutil.ConfigError,
 	// Load the configuration
 	var config *server.Config
 	for _, path := range c.flagConfigs {
-		current, err := server.LoadConfig(path)
+		current, err := server.LoadConfig(path, c.flagConfigs)
 		if err != nil {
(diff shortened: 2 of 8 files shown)


JanMa (approved):
LGTM

PR #817, opened 2024-12-22
ui: ensure logo is responsive

JanMa:
So it is resized on smaller screens and does not leak outside the sidebar


cipherboy (approved):
Looks nice!

Issue #694, opened 2024-11-11
Add Dependabot for Go dependencies


tsaarni:
Currently OpenBao has Dependabot enabled for website and ui ([link](https://github.com/openbao/openbao/blob/46e4a0ce50b7fde948b32a94a2b585b4862a5c33/.github/dependabot.yml)), but not for Go dependencies. I'd like to propose enabling it for Go. 

While it will initially generate bit of a flood of PRs and require some extra effort from maintainers to merge them, it should prove beneficial over time. As a trade-off, I suggest temporarily disabling UI updates while there is no capacity to maintain it. This would significantly reduce notification noise and prevent Dependabot from getting stalled, since it pauses new PRs when a certain number of its PRs are pending.

If you agree on this, I can submit a PR to set it up.

PR #825, opened 2024-12-23
Add dependabot for go.mod, container images

cipherboy:
Resolves: #694 

---

I hadn't realized this would also handle container images, which would be nice for avoiding the issues with our October release.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,38 +1,42 @@
 version: 2
 
 updates:
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


DrDaveD (approved):
Looks OK to me.  You might want to change the intervals to weekly instead of daily though, because updates aren't usually urgent and it makes reviewing the PRs a little more efficient.

cipherboy:
@DrDaveD said:

> Looks OK to me. You might want to change the intervals to weekly instead of daily though, because updates aren't usually urgent and it makes reviewing the PRs a little more efficient.

Yeah, I think we can enable dependabot alerts to handle the case where we're about to do a release and there's a security issue reported between when dependabot last ran and when we wanted to cut the release.

JanMa (approved):

cipherboy:
Thanks all!

PR #830, opened 2024-12-23
Sync dependencies between api, sdk, and root go.mod

cipherboy:
This pushes the dependencies we build with (from the root `go.mod`) into every other go.mod file. This means anyone pulling `api/` or `sdk/` will get the same versions of dependencies we use in the core, hopefully improving stability and compatibility.


JanMa (approved):

cipherboy:
Thanks, @JanMa!

Issue #17, opened 2023-12-11
Ability to seal/unseal own TLS certificate


bnevis-i:
Although there is a built-in PKI secrets engine, OpenBao can't bootstrap its own TLS certificate. A handy feature would be able to use the PKI secrets engine to create OpenBao's own TLS serving certificate, and include it as part of the seal data (and deal with certificate rotation as well). While this would only be doable for a self-unsealing mechanism that doesn't require the API to be up to perform the unseal, it would resolve what today is a difficult chicken-and-egg problem.  (Though this would be less of a problem if unsealing cloud be done over a named pipe or other secure non-network mechanism.)

Cloned from https://github.com/openbao/openbao/issues/16

cipherboy:
\o hey @bnevis-i --- some thoughts here.

I think this might be hard to achieve as written. Plugins cannot load until the underlying storage is decrypted and this can require manual involvement (Shamir's unseal, which is done remotely via API -- needing certificates in the HTTPS secure case), and generally requires the core + listeners to already be started.  This bootstrapping leads to a chicken-and-egg problem, and breaking that cycle might be impossible without local unseal (or strictly requiring auto-unseal). However, if you have an existing certificate, this could be reused and refreshed periodically more easily... though this also requires Bao to be able to issue potentially privileged tokens (bypassing authentication plugins entirely) to itself (or, tightly scoped tokens for single API endpoints, using authentication information it can present to itself). This is all very hard, IMO, as written.

However, more broadly applicable (and doing 90% of what you want -- the steady-state renewal) I think, is perhaps using ACME similar to how Caddy works. Let the listener point at an ACME server to fetch certificates from, allow local caching, and perhaps have a temporary self-signed certificate mode when necessary.  Luckily the version Bao is based on supports ACME in the PKI plugin, so this should be fairly achievable much more easily. Some detection (for pointing at ourselves) would be necessary -- so we opportunistically fetch via ACME but fallback to self-signing if no locally cached cert and no ACME responder is (yet) available -- but otherwise this should be much more achievable. This would also benefit people not using Bao's PKI capabilities.

Let me know your thoughts and if you'd like to formalize that into a [proper RFC](https://github.com/openbao/openbao/blob/development/.github/ISSUE_TEMPLATE/rfc.yml).

bnevis-i:
[...]

PR #857, opened 2024-12-25
Auto-TLS listeners via ACME

cipherboy:
This adds the ability to provision a TLS listener without specifying certificates; instead, [CertMagic](https://github.com/caddyserver/certmagic) (from the authors of [Caddy](https://caddyserver.com/)) will be used to provision a just-in-time certificate, received via ACME. We support HTTP-01 and TLS-ALPN-01 challenge types.

This required updating the Dockerfile so that port 80 can be optionally exposed. Notably, ACME requires solving ALPN challenges on port 443 and HTTP challenges on port 80. If directly exposing the container, it can now bind to privileged ports by using the `BAO_SKIP_DROP_ROOT`. However, if a load balancer or firewall port redirection is used, the challenges will be exposed on all ports and root permissions can be dropped.

Resolves: #17 

---

TODO:

- [x] Finish ALPN listener configuration
- [x] Add ALPN tests
- [x] Add DNS tests for allow-listing


cipherboy:
@DanGhita @JanMa This should be ready for review now!

internalshared/listenerutil/tls_acme.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+
+		acg.Magic.KeySource = certmagic.StandardKeyGenerator{
+			KeyType: certmagic.KeyType(l.TLSACMEKeyType),
+		}
+	}
+
+	template := certmagic.ACMEIssuer{
+		CA:                      l.TLSACMECADirectory,
+		TestCA:                  l.TLSACMECADirectory,
```
JanMa, line 133: I think we shouldn't set this to the same value as `CA` by default. Either leave it empty or allow the user to set this using a config variable

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -213,6 +213,62 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
...
+
+The following parameters replace direct certificate acquisition when
+`tls_cert_file` is empty. ACME defaults to being enabled. Set the
+`tls_acme_ca_directory` to the empty string to disable. These options belong
+in the root of the TCP listener configuration.
+
+- `tls_acme_ca_directory` `(string: "https://acme-v02.api.letsencrypt.org/directory")` - Path
+  to the ACME directory; defaults to Let's Encrypt's production CA.
+
+- `tls_acme_cache_path` `(string: "~/.local/share/certmagic")` - Specifies the
```
JanMa, line 226: This default implies that the (linux) user running OpenBao has a home directory. In our Docker images and DEB/RPM packages, openbao is running as a system user without a home directory. So accessing the default path might return an error. 

Not sure what a sensible default might be though. Maybe `/var/lib/openbao/certmagic` :thinking:

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -213,6 +213,62 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
...
+- `tls_acme_eab_key_id` `(string: "")` - Specify the optional External Account
+  Binding account reference.
+
+- `tls_acme_eab_mac_key` `(string: "")` - Specify the optional External Account
+  Binding account protection key.
+
+- `tls_acme_email` `(string: "")` - Specify an optional account the CA may
+  contact you at for expiring certificates or other service-related emails.
+
+- `tls_acme_domains` `(string: "")` - An optional but strongly recommended
```
JanMa, line 243: This is a list of strings

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -213,6 +213,62 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
...
+
+:::
+
+- `tls_acme_disable_http_challenge` `(bool: false)` - An optional but strongly
+  recommended option to allow disabling HTTP challenges. Use of this method
+  without an HTTP listener causes a temporary bind on port 80; this allows
+  a remote peer to connect (if misconfigured to use HTTP) and potentially
+  send sensitive authentication data in the clear over the wire for the
+  duration the challenge solving listener is active. While this data is
+  ignored on the OpenBao side, it is.
```
JanMa, line 265: I think you are missing the rest of the sentence here?

JanMa (changes requested):
Not completely done yet, but here are already some remarks

internalshared/listenerutil/tls_acme.go
```diff
@@ -0,0 +1,280 @@
...
+func (c *ACMECertGetter) Reload() error { return nil }
+
+func (c *ACMECertGetter) GetCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	// ACMECertGetter follows the strategy by Caddy of on-demand TLS
+	// configuration: when a connection comes in for a domain, attempt to
+	// get a valid certificate from the ACME responder, just-in-time.
+	return c.Magic.GetCertificate(hello)
+}
+
+type zapHclCore struct {
```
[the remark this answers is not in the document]
cipherboy: @JanMa The issue is there's two different formats and stdout vs file writing, so it gets hard to implement it in a 1:1 compatible way. The other thing you risk is, but not calling into hclog, that you interrupt another log line (because you don't have locking), such as:

```
{START OF HCLOG LINE}{SOME ZLOG LINE}
{END OF HCLOG LINE}
```

and so you'll get corrupted data in, say, JSON format. I think you're really better off with this hack, though I'm happy extracting it out somewhere else if you prefer (e.g. `tls_acme_zap_hclog_compat.go`).

internalshared/listenerutil/tls_acme.go
cipherboy: It is not clear from the certmagic docs when this is used. I think it is used when `attempts > 0` perhaps? But that makes no sense.

website/content/docs/configuration/listener/tcp.mdx
cipherboy: > This default implies that the (linux) user running OpenBao has a home directory. In our Docker images and DEB/RPM packages, openbao is running as a system user without a home directory. So accessing the default path might return an error.

Hmmm, I'm not sure how we'd be able to conditionally check if we can use any other path, tbh. This is also not our default, but certmagic's default: https://pkg.go.dev/github.com/caddyserver/certmagic#readme-storage

`/var/lib/openbao` may not exist, and so the `openbao` user may not have permissions to write to it. I think that means we'd have to fall back to `$TEMPDIR`?

website/content/docs/configuration/listener/tcp.mdx
cipherboy: Ah, yes, good catch.

internalshared/listenerutil/tls_acme.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/mholt/acmez/v2/acme"
+	"github.com/mitchellh/cli"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
+	"github.com/openbao/openbao/internalshared/configutil"
+	"github.com/openbao/openbao/version"
+)
+
+type Reloadable interface {
```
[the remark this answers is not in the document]
cipherboy: Thanks, fixed.

internalshared/listenerutil/tls_acme.go
cipherboy: I set this to the empty string. I think that means we won't ever use the TestCA, which is probably ideal behavior, given you can't explicitly tell CertMagic to use the TestCA>

JanMa: By default certmagic sets this to the Let's Encrypt Staging CA
https://github.com/caddyserver/certmagic/blob/v0.21.7/acmeissuer.go#L638-L643

cipherboy: @JanMa But only if `CA == DefaultACME.CA`: 

https://github.com/caddyserver/certmagic/blob/3f666293f0e588b804d5bc4e792a841d68a3f27c/acmeissuer.go#L183-L189

Which it doesn't, so it'll be blank?

JanMa: Right, so let's set this blank by default and enable the user to override it if they want?

website/content/docs/configuration/listener/tcp.mdx
JanMa: Interesting, the certmagic default is subtly different from what you wrote

> By default, CertMagic stores assets on the local file system in $HOME/.local/share/certmagic (and honors $XDG_DATA_HOME if set). CertMagic will create the directory if it does not exist. If writes are denied, things will not be happy, so make sure CertMagic can write to it!

I guess this is fine then

internalshared/listenerutil/tls_acme.go
cipherboy: @JanMa Hmm... Like I said, I don't think we ever use it. There's no switch in certmagic to actually force it to use the TestCA, so I'm not sure what value it serves. I think the only case it gets used is if the main issuance fails for some reason, it'll try falling back to the test server to see if that one works, before retrying on the main server.

Mostly I think that's in the interest of avoiding rate limiting in the event of a configuration failure, though CertMagic should already respect the rate limiting (and has its own).

cipherboy: Filed as https://github.com/caddyserver/certmagic/issues/330

JanMa (approved):

JanMa (approved):

cipherboy:
Thanks @JanMa!

PR #868, opened 2024-12-31
CEL auth program evaluation during JWT login

### LINKED ISSUE #493: Expand JWT claim matching capabilities

[GitLab provisions](https://docs.gitlab.com/ee/ci/jobs/job_rules.html#common-if-clauses-with-predefined-variables) complex JWT claims on ID tokens used with its [Vault integration](https://docs.gitlab.com/ee/ci/examples/authenticating-with-hashicorp-vault/). However, the JWT role isn't expressive enough to fully authenticate these fields.

In particular, the `sub` claim contains the following structure, which is unique to pipeline jobs:

> `project_path:{group}/{project}:ref_type:{type}:ref:{branch_name}`

Matching on a prefix of this (`bound_subject=project_path:my-group/my-project:ref_type:*:ref:*`) isn't possible, to allow a pipeline running in any branch to have access to protected secrets. Similarly, for `bound_claims`, it isn't possible to match a glob or an either-or choice (say, `bound_claims[environment]=prod-*` or `bound_claims[ref]=v*` (with `ref_type=tag`)).

I'm not sure what tools we have currently, or what make the best sense to solve it:

 - Globs could work, but would be rather greedy.
 - Regexes could also work but be more error-prone.
 - Templating is a usual design approach but because we're authenticating, we won't have it yet.
 - Lists of possible matches (`allowed_domains` in PKI secrets enging).

Manually specifying every possible value could potentially require a large increase in the number of roles, or tight coupling to their current values which could be inflexible (e.g., binding strictly to version number from a tag, while not binding at all could result in say, requests from another branch).

n/a

**Additional context**

n/a

suprjinx:
This is a POC PR to rough out an approach for adding CEL auth_programs to the JWT auth method. CEL entries are posted to path `cel/roles` for the auth method, and chiefly contain a string attribute `auth_program`. All `cel/roles` for the auth method are executed at the end of login and can modify the `logical.Auth` before it is returned.

A CEL auth pogram that returns `false` or `{"authorized": false}` will reject the login.

A CEL auth program that returns `{"add_policies": ["some_policy","some_other"]}` will add the policies to the `logical.Auth` returned from `path_login`.

A CEL auth program that returns `{"remove_policies": ["some_policy"]}` will remove the policy from the `logical.Auth`.

The `cel/roles` CRUD pieces are borrowed from @fatima2003's PR #753 

Resolves #493
Co-authored-by @fatima2003


suprjinx:
closing in favor of PR #869

PR #869, opened 2024-12-31
CEL auth program evaluation during JWT login

### LINKED ISSUE #493: Expand JWT claim matching capabilities

suprjinx:
This is a POC PR to rough out an approach for adding CEL auth_programs (CEL roles) to the JWT auth method. CEL role entries are posted to path `cel/role` for the auth method, and chiefly contain a string attribute `auth_program`. A login seeking to use the role will login to `{jwt auth mount}/cel/login -role {celRole.Name}`

A CEL auth pogram that returns `false` will reject the authorization attempt.

A CEL auth program that returns `true` or null will authorize a token according to the auth program.

The auth program uses CEL functions to set attributes of a (transient) role, such as: `SetPolicies`, `SetBoundCIDRs`, etc, and may check any attributes of the jwt `claims`

The `cel/role` CRUD pieces are borrowed from @fatima2003's PR #753 

Resolves #493
Co-authored-by @fatima2003


builtin/credential/jwt/path_login.go
```diff
@@ -194,11 +205,99 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
 		return nil, err
 	}
 
+	if err := b.applyCelRoles(ctx, req, allClaims, auth); err != nil {
```
cipherboy, line 208: Ah interesting, I think this is different than I'd expect based on discussion in https://github.com/openbao/openbao/issues/493#issuecomment-2549033976 -- though, reading the RFC more carefully, I think there's a mismatch between RFC sections.

In "User-facing Description":

> **Administrators** can define CEL extensions under a new endpoint, `auth/{oidc|jwt}/cel/roles/:name`, to be executed during authentication. CEL roles evaluate incoming claims and generate `logical.Auth{}` instances having appropriate token policies.

Namely, the "executed during authentication" makes me think it supplements the existing role-based system, as is the case in this code. However, I thought we had decided on _replacing_ the role system entirely, which seems to match what is in the "Technical Description" section:

> CEL roles integrate with Vault's Auth engine (OIDC and JWT) to dynamically evaluate claims. Instead of using static role-to-policy assignment with `bound_claim`, CEL programs validate claims and return a `logical.Auth` instance containing dynamically assigned token policies.

(from the "Instead of using static role-to-policy..." portion of that quote).

My 2c., but I think I'd prefer the latter to the former as it is simpler to reason about and has fewer moving parts.  

---

In particular, though not articulated in the RFC, I think I'd have expected a CEL/regular role _name_ to be globally unique (across both types). Login (with a role name) would then lookup the single unique role (whether regular or CEL) and evaluate just that one role on the request. 

Or, perhaps easier, rather than globally unique, have a `cel/login/:role` path, which logs in with the specified CEL role.

builtin/credential/jwt/path_login.go
```diff
@@ -194,11 +205,99 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
+func (b *jwtAuthBackend) applyCelRoles(ctx context.Context, req *logical.Request, allClaims map[string]any, auth *logical.Auth) error {
+	celRoleKeys, err := req.Storage.List(ctx, "cel/roles")
+	if err != nil {
+		return err
+	}
+	if len(celRoleKeys) == 0 {
+		return nil
+	}
+	for _, key := range celRoleKeys {
+		celRole, err := b.getCelRole(ctx, req.Storage, key)
```
cipherboy, line 226: In particular, I don't quite like this behavior of evaluating all CEL policies against all tokens. It seems like it could lead to high complexity/latency if lots of CEL roles exist, which each validate a small portion of a token, rather than evaluating a single larger role.

cipherboy (commented):
Nice! Some comments inline, I think we might've crossed wires on the RFC, reading it more closely...

builtin/credential/jwt/path_login.go
suprjinx: can do -- I was thinking a layering/additive approach might make it easier to compose a bunch of uncomplicated CEL programs -- but your idea makes more sense with the way regular roles work.

suprjinx: I do think there's a lot of complexity here if we want to completely replace role -- everything that `roles` do, the `cel/roles` will need to do. The approach in my PR was more supplemental (to just post-process the token_policies, essentially) based on some role (eg, default) already being applied. I think it's worth considering this strategy before discarding since it's much simpler -- the "cel_auth_program" could even become an reference attribute on the role?

suprjinx: Another thought -- what if `cel_auth_program` is just an optional field belonging to `jwtRole`. We apply if it exists for the active role, otherwise it's just a normal role. This avoids issue of name uniquesness across different role categories; and allows a `CelRole` to behave like existing for other fields such as `OIDCScopes`, etc

fatima2003: Hmm I've also been contemplating this. In particular, with pure CEL roles, authors would have to write rules for cases that could be handled by simply enabling or disabling a flag in a regular role. , for instance:

Flag: `allow_localhost = true`
CEL Equivalent: `request.common_name == 'localhost' || request.common_name == '127.0.0.1'`

This level of detail might not always be necessary. For instance, what if the author’s intent was simply to ensure that two specific domains (e.g., example.com and example.org) were included in the request in addition to a few other regular role constraints.

A possible solution (at least for #794) might be to allow role authors to optionally combine CEL roles with regular roles. I haven’t worked out a concrete design yet, but perhaps CEL roles could take an additional parameter referencing the path to a regular role. When validation requests are made, both the CEL rules and the traditional role's validation rules could be applied. 

That said, I share the same concerns as @cipherboy’s first point. Semantically invalid combination of roles could be easily created. For example, a CEL role might specify a rule like `request.common_name == "example.com"`, which is valid on its own. But if combined with a regular role that lacks `cn_validations: ["hostname"]`, the rule becomes obsolete and could fail in unexpected ways.

This idea seems like the inverse of @suprjinx's proposal:
> Another thought -- what if cel_auth_program is just an optional field belonging to jwtRole. We apply if it exists for the active role, otherwise it's just a normal role. 

Alternatively, CEL roles optionally inherit the same fields as regular roles, enforcing those rules in addition to CEL rules. However, I think this is too complex and has similar problems as the previous suggestion.

fatima2003: Another idea that just came to mind is introducing the CEL rule equivalents of regular role parameters as globally accessible supplemental variables for CEL role authors. That way, they can simply specify the variable in their cel expressions. For example, rather than requiring authors to rewrite something like:

`request.common_name == 'localhost' || request.common_name == '127.0.0.1'`
They could simply reference a predefined variable like `allow_localhost` in their CEL expressions.

This would reduce the burden on CEL role authors by allowing them to leverage the existing regular role logic without manually replicating it.

However, I'm uncertain about the complexity and potential cost this could add to the CEL program itself, particularly around managing and validating these supplemental variables. For instance:

* Would these variables need to be defined globally, even when the corresponding feature isn’t enabled in the system? For example, if the system defines a global variable `allow_localhost` as part of the CEL environment, should it always exist (perhaps defaulting to false) even if no rule explicitly enables `allow_localhost`?
* How should conflicting logic between predefined variables and custom CEL rules be handled? For example, what happens if a predefined variable suggests one outcome, but the custom CEL rule implies another?

suprjinx: Thanks for the feedback @fatima2003! You raise good questions here. My reaction is that the CEL rule should be supplemental to the role, so the CEL author doesn't have to specify all of role attributes. As a "post-processing program" it would be pretty clear that the CEL rule has final say on an attribute (in this PR, only `token_policies`)

suprjinx: @cipherboy looks like I found the wrong meeting on the wiki for the community call today, sorry I missed.

cipherboy: @suprjinx No worries! You might've been on the TSC call. I've published slides and recordings here: https://lf-edge.atlassian.net/wiki/spaces/OP/pages/15211863/OpenBao+Meetings  

We discussed this on the call, but rather than using variables, I think we can inject helper functions into the environment that will behave similarly to the roles. We could even go all the way to allowing full role evaluation (using the same logic as the Go side) using either a named role or using a stubbed role configuration. That way, CEL authors would have a very easy way of either using pieces of roles policy or using a full policy along side their additional CEL policy, if they'd like to. 

I'm also soliciting feedback from the Dev WG (Jan, Dan mostly unless Nathan wishes to comment) -- but I think that with Fatima's approach, we can have a global policy/suggested way of using CEL as being an independent evaluation language (free of _requiring_ a role evaluation), but with the option to call back into the role system if desired. This would also work later for Sentinel-like policies or a replacement to the ACL policy system, where the CEL policy could be evaluated and may optionally invoke additional ACL policy evaluations on the request, if desired.

builtin/credential/jwt/path_cel_login.go
```diff
@@ -0,0 +1,329 @@
...
+	}, nil
+}
+
+func (b *jwtAuthBackend) runCelProgram(ctx context.Context, celRoleEntry *celRoleEntry, allClaims map[string]any) (*jwtRole, error) {
+	role := jwtRole{}
+	// these functions are closures around a temporary `role` and will affect it
+	env, err := cel.NewEnv(
+		cel.Variable("claims", cel.MapType(cel.StringType, cel.DynType)),
+		cel.Function("SetPolicies",
+			cel.Overload("SetPolicies",
```
suprjinx, line 219: @cipherboy I am changing this PR to evaluate CEL role login at a separate path and to use CEL functions for adjusting the resulting Auth token -- let me know if this feels like the right direction!

suprjinx:
@cipherboy let me know how I should proceed with this one. Would it make sense to refactor to an auth plugin?

builtin/credential/jwt/path_cel_role.go
```diff
@@ -0,0 +1,581 @@
...
+		return false, fmt.Errorf("failed to compile CEL rule")
+	}
+
+	return true, nil
+}
+
+func (b *jwtAuthBackend) celEnv(role *jwtRole) (*cel.Env, error) {
+	return cel.NewEnv(
+		// these functions are closures around `role` and can alter it
+		cel.Variable("claims", cel.MapType(cel.StringType, cel.DynType)),
```
cipherboy, line 370: Hmm, I see what you've done here.

So I ended up using the [`identity` engine](https://openbao.org/api-docs/secret/identity/tokens/#generate-a-signed-id-token) to generate ID tokens.

```
# Create userpass
bao auth enable userpass
bao write auth/userpass/users/testing password=testing token_policies=example
bao write sys/policies/acl/example policy='path "identity/oidc/token/example" { capabilities = ["read"] }'

# Set up identity
bao write identity/oidc/key/default algorithm=EdDSA
bao write identity/oidc/role/example key=default

# Setup JWT
bao auth enable jwt
bao write auth/jwt/config oidc_discovery_url=http://127.0.0.1:8200/v1/identity/oidc

# Get new JWT
BAO_TOKEN=$(bao login -field=token -method=userpass username=testing password=testing) bao read identity/oidc/token/example

# Prime JWT login role
bao write auth/jwt/role/testing user_claim=aud role_type=jwt bound_audiences=iebGrGKN1D0a8sCJB58NgSdNSX # output from above

# Try login
bao write auth/jwt/login role=testing jwt=...
```

So then it becomes a question of how to auth with cel... For one, I think there's a lot of missing functions below. I started with:

```
bao write auth/jwt/cel/role/example auth_program='SetPolicies(["example"]) && SetUserClaim("aud")'
```

but this results in:

```
$ bao write auth/jwt/cel/login role=example jwt=...
Error writing data to auth/jwt/cel/login: Error making API request.

URL: PUT http://localhost:8200/v1/auth/jwt/cel/login
Code: 400. Errors:

* audience claim found in JWT but no audiences bound to the role
```

because there is no way to set `bound_audiences` on this role.

This means the restrictions of the role system still apply and you can't quite free cel from having to extend the role each time... I do like the idea of reusing common elements though, if we can... Hm.

builtin/credential/jwt/path_cel_login.go
```diff
@@ -0,0 +1,286 @@
...
+	signingAlgorithms := toAlg(config.JWTSupportedAlgs)
+	if len(signingAlgorithms) == 0 {
+		signingAlgorithms = []jwt.Alg{
+			jwt.RS256, jwt.RS384, jwt.RS512, jwt.ES256, jwt.ES384,
+			jwt.ES512, jwt.PS256, jwt.PS384, jwt.PS512, jwt.EdDSA,
+		}
+	}
+
+	// Set expected claims values to assert on the JWT
+	expected := jwt.Expected{
```
cipherboy, line 150: Maybe the comment about `celEnv` is by design; it seems from here like audiences, subjects, claims, &c would have to be fully vetted by Cel. Perhaps with something like:

```cel
'aud' in claims ? 
    (
        claims["aud"] != "" &&
        SetPolicies(["example"]) &&
        SetUserClaim("aud")
    ) : false
```

In particular, I don't see a way to modify the environment, so we can't remove `claim["aud"]` dynamically, unless we add a helper function for that (`DeleteTokenClaim(...)` perhaps).

I'm not sure, tbh, what the best approach would be. The verification isn't exactly modular and returning a map type seems to require we know keys in advance (and can't necessarily incrementally build the map...).

cipherboy: (cc @fatima2003)

In PKI, I think because the role system was already very modular, we opted to expose each of the role's helper functions in the Cel environment, letting role-like validation occur from cel via chaining -- e.g., `ValidateNames(csr.dns) && ValidateIP(...)` &c -- some more refactoring would make that a touch more explicit.

But I'll have to think more what the best operator experience would be here. It'd be nice not to have to enforce the entire role-based system here... but also have the option to use pieces of it, if desired.

builtin/credential/jwt/path_cel_login.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	if err := jwtRole.PopulateTokenAuth(auth, req); err != nil {
+		return nil, fmt.Errorf("failed to populate auth information: %w", err)
+	}
+
+	if err := jwtRole.maybeTemplatePolicies(auth, allClaims); err != nil {
+		return nil, err
+	}
+
```
cipherboy, line 211: If I'm not mistaken, we don't have a way of modifying aliases or groups or anything from cel afterwards? Perhaps an `validation_program` versus `auth_program` makes sense? Not sure.

cipherboy (commented):
@suprjinx To answer your question, now that #880 is done, I should have more time to think about Cel. I would like to see it in this plugin and not fork another plugin off. I still think there's a fair amount of reuse that makes a fork less desirable. 

That said, I left some initial comments here. I wasn't quite able to get a successful cel validation performed, so, hm. Will have to think more!

builtin/credential/jwt/path_cel_role.go
suprjinx: Thanks for testing! Should the `celRole` should have the attributes of a regular jwt role, eg `bound_audiences`, so that we don't need to include this in the auth program?
```
// Set expected claims values to assert on the JWT
	expected := jwt.Expected{
		Issuer:            config.BoundIssuer,
		Subject:           role.BoundSubject,
		Audiences:         role.BoundAudiences,
		SigningAlgorithms: signingAlgorithms,
		NotBeforeLeeway:   role.NotBeforeLeeway,
		ExpirationLeeway:  role.ExpirationLeeway,
		ClockSkewLeeway:   role.ClockSkewLeeway,
	}
```

suprjinx: i've got this working now with an audiences attribute now:
```
gwilson@homeslice-linux ~/P/openbao (cel-roles-auth2) [2]> BAO_TOKEN="dev-only-token" bao write auth/jwt/cel/role/example auth_program='SetPolicies(["example"]) && SetUserClaim("aud")' bound_audiences='aud,aud1,pnaOezgsRgqKAW0vxmV1lN9Okr'
Key                  Value
---                  -----
auth_program         SetPolicies(["example"]) && SetUserClaim("aud")
bound_audiences      [aud aud1 pnaOezgsRgqKAW0vxmV1lN9Okr]
clock_skew_leeway    60
expiration_leeway    150
failure_policy       Deny
message              n/a
name                 example
not_before_leeway    150
gwilson@homeslice-linux ~/P/openbao (cel-roles-auth2)> bao write auth/jwt/cel/login role=example jwt=eyJhbGciOiJFZERTQSIsImtpZCI6ImFkNjAxNTRhLTcyMjMtYzJiOS1hZjlhLTAyYTFiMWFlMjhkOSJ9.eyJhdWQiOiJwbmFPZXpnc1JncUtBVzB2eG1WMWxOOU9rciIsImV4cCI6MTc0MzAxMjY5OCwiaWF0IjoxNzQyOTI2Mjk4LCJpc3MiOiJodHRwOi8vMTI3LjAuMC4xOjgyMDAvdjEvaWRlbnRpdHkvb2lkYyIsIm5hbWVzcGFjZSI6InJvb3QiLCJzdWIiOiJkYzlhZjMwOS1lYTljLTRlMjAtNGFkMi03MGZlZTA5N2YyZGYifQ.pxRw7xA4QkIaqlckeyjC02bHgZvm0O9ctwgIeQFHE98GmLot5G5TD7yYhZT9gdI3XIxAdQVnpSYwL9FA2cr7Bw
Key                  Value
---                  -----
token                s.Y2uJO5FC8WOMoRVMgZ5pnw2H
token_accessor       bu8HpQbueHv8Ln911hRBOjlL
token_duration       768h
token_renewable      true
token_policies       ["default" "example"]
identity_policies    []
policies             ["default" "example"]
token_meta_role      example
```

builtin/credential/jwt/path_cel_login.go
suprjinx: I was assuming that template policies could proceed here as they do with normal jwt flow -- if the policies that result from the cel program are templated.

cipherboy: @suprjinx So the short coming with that is we're limited by the power of our templating languages (as discussed a bit in #1207 but also as noted in #618 -- templates in JWT roles (to set policies) can only create one policy at a time and can't do a 1->n mapping very easily, without repeating the template for each of a maximum number of indices.

Phrased alternatively, you have to create a template for each possible index in any array-based claims (e.g., using GitLab as an example, the list of groups a user is part of), which is a bit hard to maintain and results in large role sizes.

This is mostly a shortcoming of _where_ policies are created, along with a shortcoming of the only return type of `text/template` (which is, well, text). So having CEL directly set the full `logical.Auth` would be most ideal to avoid this. Then we could use [the `map` macro](https://github.com/google/cel-spec/blob/master/doc/langdef.md#macros) to expand a list of claims into a list of policies, with only a single map invocation. 

---

(On the policy evaluation side, #1207 stands and that `identitytpl` language is very limited as well so CEL as a policy language would be great to have, too.)

cipherboy (commented):
Some random in-progress comments I've completely forgotten about. Not sure if they still make sense, but the patch ones do so I figured I'd post them all.

builtin/credential/jwt/path_cel_role.go
```diff
@@ -0,0 +1,664 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if oldEntry == nil {
+		return logical.ErrorResponse("Unable to fetch cel role entry to patch"), nil
+	}
+
+	entry := &celRoleEntry{
+		Name:          roleName,
+		AuthProgram:   data.Get("auth_program").(string),
```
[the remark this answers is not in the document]
suprjinx: moved `getWithExplicitDefault` to `FieldData`, and used here

builtin/credential/jwt/path_cel_role.go
```diff
@@ -0,0 +1,485 @@
...
+	"github.com/openbao/openbao/sdk/v2/framework"
+	celhelper "github.com/openbao/openbao/sdk/v2/helper/cel"
+	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/openbao/openbao/sdk/v2/plugin/pb"
+)
+
+type celRoleEntry struct {
+	Name          string               `json:"name"`                     // Required
+	CelProgram    celhelper.CelProgram `json:"cel_program"`              // Required
+	FailurePolicy string               `json:"failure_policy,omitempty"` // Default: Deny
```
cipherboy, line 25: If I'm not mistaken, isn't `FailurePolicy` unused? I don't know that it has a role under the single `CelProgram` design, tbh.

builtin/credential/jwt/path_cel_role.go
```diff
@@ -0,0 +1,485 @@
...
+		return nil, fmt.Errorf("%w", err)
+	}
+
+	resp.Data = entry.ToResponseData()
+	return resp, nil
+}
+
+func (b *jwtAuthBackend) validateCelProgram(program celhelper.CelProgram) (bool, error) {
+	// adding a minimal jwtClaims collection here, for validating usages in CEL expression
+	_, err := b.celEvalProgram(program, map[string]any{"sub": "email@example.com", "aud": "audience", "iss": "issuer"})
```
cipherboy, line 418: :thinking: Hmm, I do wonder how scalable this will be, but I suppose enforcing optionality for some of these might be hard.

builtin/credential/jwt/path_cel_role.go
```diff
@@ -0,0 +1,485 @@
...
+		// Add the evaluated result for subsequent CEL evaluations.
+		// This ensures variables can reference each other and build a cumulative evaluation context.
+		evaluationData[variable.Name] = result.Value()
+	}
+
+	// Evaluate the CEL Role success expression
+	return celhelper.ParseCompileAndEvaluateExpression(env, program.Expression, evaluationData)
+}
+
+func (b *jwtAuthBackend) celEnv(program celhelper.CelProgram) (*cel.Env, error) {
```
cipherboy, line 454: In the future, not for this PR as I'd like to see it merged, but we can add `EvaluateRole(...)` as a helper, so the caller can have easy enforcement.

cipherboy (commented):
I'll need a minute to review the tests, but they look a lot more comprehensive now, many thanks @suprjinx! This is looking great, really appreciate your patience here.

builtin/credential/jwt/path_cel_role.go
suprjinx: I'll remove -- I think that was left over from adapting `Role`. Now, we could return Auth with failure policy from the CEL expression.

builtin/credential/jwt/path_cel_login.go
```diff
@@ -0,0 +1,266 @@
...
+	pathCelLoginHelpSyn = `
+	Authenticates to OpenBao using a JWT (or OIDC) token against a CEL role.
+	`
+	pathCelLoginHelpDesc = `
+Authenticates JWTs against a CEL role.
+`
+)
+
+// authFromProto will convert a protobuf auth message to logical.Auth
+func authFromProto(pbAuth pb.Auth) logical.Auth {
```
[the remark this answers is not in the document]
cipherboy: Ah sorry, there's the inverse method there too. ProtoAuthToLogicial.

builtin/credential/jwt/path_cel_role.go
cipherboy: Thank you!

builtin/credential/jwt/path_cel_role.go
```diff
@@ -0,0 +1,551 @@
...
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/checker/decls"
+	"github.com/openbao/openbao/sdk/v2/framework"
+	"github.com/openbao/openbao/sdk/v2/logical"
+)
+
+type celRoleEntry struct {
+	Name          string     `json:"name"`                     // Required
+	AuthProgram   CelProgram `json:"auth_program"`             // Required
+	FailurePolicy string     `json:"failure_policy,omitempty"` // Default: Deny
```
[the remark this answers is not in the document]
cipherboy: Ah, this was a duplicate comment, my bad.

builtin/credential/jwt/path_cel_login.go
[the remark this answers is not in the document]
cipherboy: @suprjinx I ended up with:

```diff
diff --git a/builtin/credential/jwt/path_cel_login.go b/builtin/credential/jwt/path_cel_login.go
index 1182a5c30d..3dd34c43bd 100644
--- a/builtin/credential/jwt/path_cel_login.go
+++ b/builtin/credential/jwt/path_cel_login.go
@@ -8,7 +8,6 @@ import (
        "errors"
        "fmt"
        "reflect"
-       "time"
 
        "github.com/google/cel-go/common/types/ref"
        "github.com/hashicorp/cap/jwt"
@@ -165,14 +164,17 @@ func (b *jwtAuthBackend) pathCelLogin(ctx context.Context, req *logical.Request,
                return logical.ErrorResponse("error executing cel program: %s", err.Error()), nil
        }
 
-       auth := authFromProto(*pbAuth)
+       auth, err := pb.ProtoAuthToLogicalAuth(pbAuth)
+       if err != nil {
+               return nil, fmt.Errorf("failed unmarshaling response from CEL: %w", err)
+       }
 
        if err := logical.EndTxStorage(ctx, req); err != nil {
                return nil, err
        }
 
        return &logical.Response{
-               Auth: &auth,
+               Auth: auth,
        }, nil
 }
 
@@ -238,23 +240,6 @@ Authenticates JWTs against a CEL role.
 `
 )
 
-// authFromProto will convert a protobuf auth message to logical.Auth
-func authFromProto(pbAuth pb.Auth) logical.Auth {
-       return logical.Auth{
-               DisplayName: pbAuth.DisplayName,
-               // InternalData:     pbAuth.InternalData, // TODO requires unmarshaling
-               Policies:         pbAuth.Policies,
-               TokenPolicies:    pbAuth.TokenPolicies,
-               IdentityPolicies: pbAuth.IdentityPolicies,
-               NoDefaultPolicy:  pbAuth.NoDefaultPolicy,
-               Metadata:         pbAuth.Metadata,
-               Period:           time.Duration(pbAuth.Period),
-               ExplicitMaxTTL:   time.Duration(pbAuth.ExplicitMaxTTL),
-               NumUses:          int(pbAuth.NumUses),
-               BoundCIDRs:       marshalCIDRs(pbAuth.BoundCIDRs),
-       }
-}
-
 func marshalCIDRs(cidrs []string) []*sockaddr.SockAddrMarshaler {
        sockaddrs := []*sockaddr.SockAddrMarshaler{}
        for _, cidr := range cidrs {
```

that seems to work.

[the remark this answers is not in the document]
suprjinx: Removed this func and now uses the pb/translation func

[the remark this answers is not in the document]
cipherboy: @suprjinx Ahh, I see, this doesn't work:

```
    --- FAIL: TestCelRoleAuth/Subjects_match,_expect_success (0.01s)
        path_cel_login_test.go:319: unexpected error: err:failed unmarshaling response from CEL: unexpected end of JSON input resp:(*logical.Response)(nil)
    --- FAIL: TestCelRoleAuth/Audience_match,_expect_success (0.00s)
        path_cel_login_test.go:319: unexpected error: err:failed unmarshaling response from CEL: unexpected end of JSON input resp:(*logical.Response)(nil)
```

We need to make the decoding optional:


```diff
diff --git a/sdk/plugin/pb/translation.go b/sdk/plugin/pb/translation.go
index d42cbaafb7..16b86f7fc1 100644
--- a/sdk/plugin/pb/translation.go
+++ b/sdk/plugin/pb/translation.go
@@ -551,9 +551,12 @@ func ProtoAuthToLogicalAuth(a *Auth) (*logical.Auth, error) {
        }
 
        data := map[string]interface{}{}
-       err := json.Unmarshal([]byte(a.InternalData), &data)
-       if err != nil {
-               return nil, err
+
+       if len(a.InternalData) > 0 {
+               err := json.Unmarshal([]byte(a.InternalData), &data)
+               if err != nil {
+                       return nil, err
+               }
        }
 
        lo, err := ProtoLeaseOptionsToLogicalLeaseOptions(a.LeaseOptions)
```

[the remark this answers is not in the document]
cipherboy: Ah, you're faster than me. Thank you!!

cipherboy (approved):
Assuming CI is happy, I am happy. Thank you very much @suprjinx!! Glad to see this merged for v2.3.0-beta.

cipherboy:
Test 8 is PeriodicFunc failures discussed in #1177.

DrDaveD:
@cipherboy We need another PR for changelog and documentation, right?

cipherboy:
@DrDaveD #869 had a changelog entry; feel free to suggest improvements: https://github.com/openbao/openbao/blob/main/changelog/869.txt

I do think documentation is outstanding, but I think we can get that in ahead of GA as it only affects the website.

DrDaveD:
That's fine, I just missed it.

PR #878, opened 2025-01-08
Use transactions in credential write operations

### LINKED ISSUE #607: Broader usage of Transactions

suprjinx:
Added helpers to assist TX storage operations; used this in `put` operations related to the builtin auth methods where the existing config is read and then overwritten.

Related: #607

### DIFF
--- a/builtin/credential/jwt/path_config.go
+++ b/builtin/credential/jwt/path_config.go
@@ -316,6 +316,12 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 		ProviderConfig:       d.Get("provider_config").(map[string]interface{}),
 	}
 
+	txRollback, err := logical.StartTxStorage(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	defer txRollback()
+
 	// Check if the config already exists, to determine if this is a create or
 	// an update, since req.Operation is always 'update' in this handler, and
 	// there's no existence check defined.
@@ -431,13 +437,15 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 	if err != nil {
 		return nil, err
 	}
-	if err := logical.WithTransaction(ctx, req.Storage, func(storage logical.Storage) error {
-		return storage.Put(ctx, entry)
-	}); err != nil {
+	if err := req.Storage.Put(ctx, entry); err != nil {
 		return nil, err
 	}
 
 	b.reset()
+
+	if err := logical.EndTxStorage(ctx, req); err != nil {
+		return nil, err
+	}
 
 	return nil, nil
 }
--- a/builtin/credential/jwt/path_config.go
+++ b/builtin/credential/jwt/path_config.go
@@ -431,7 +431,9 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 	if err != nil {
 		return nil, err
 	}
-	if err := req.Storage.Put(ctx, entry); err != nil {
+	if err := logical.WithTransaction(ctx, req.Storage, func(storage logical.Storage) error {
+		return storage.Put(ctx, entry)
+	}); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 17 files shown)


builtin/credential/jwt/path_config.go
```diff
@@ -431,7 +431,9 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 	if err != nil {
 		return nil, err
 	}
-	if err := req.Storage.Put(ctx, entry); err != nil {
+	if err := logical.WithTransaction(ctx, req.Storage, func(storage logical.Storage) error {
```
cipherboy, line 434: Don't we want the transaction created earlier on the read call as well, for consistency between when the read occurred and when the write executed?

suprjinx: yes that's a good point -- i wasn't sure if the cost of a longer transaction would be too high

builtin/credential/kubernetes/path_role.go
```diff
@@ -246,10 +246,11 @@ func (b *kubeAuthBackend) pathRoleDelete(ctx context.Context, req *logical.Reque
 	defer b.l.Unlock()
 
 	// Delete the role itself
-	if err := req.Storage.Delete(ctx, "role/"+strings.ToLower(roleName)); err != nil {
+	if err := logical.WithTransaction(ctx, req.Storage, func(storage logical.Storage) error {
```
cipherboy, line 249: I believe we operate on the assumption that individual storage operations are atomic. I don't know if PostgreSQL behaves differently, but I'd assume if we're modifying a single entry that single modification should be fine...

There's a lot of places where single entries are modified that I don't think make sense to move to transactions, given we have a lower pool limit (256 -> 128 I believe).

---

I think this is true of every usage of `WithTransaction(...)` in credentials, tbh. There are some places where we might want a straight-shot transaction, such as in PKI Tidy though, e.g., inside the item callback: 

https://github.com/openbao/openbao/blob/29e4b7df84312f2e8e68adccbe762f50902af8da/builtin/logical/pki/path_tidy.go#L967-L1050

So I'd keep the helper, but perhaps not use it here.

cipherboy (approved):
Open to discussion on this point.

changelog/878.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+credentials: Use transactions for write and delete config for various auth methods.
```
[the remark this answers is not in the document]
suprjinx: 👍 making change in next commit

builtin/credential/kubernetes/path_role.go
suprjinx: removed the WithTransaction helper on the simple operations

builtin/credential/ldap/path_config.go
```diff
@@ -208,6 +215,10 @@ func (b *backend) pathConfigWrite(ctx context.Context, req *logical.Request, d *
 		}, nil
 	}
 
+	if err := logical.EndTxStorage(ctx, req); err != nil {
```
cipherboy, line 218: @suprjinx One bug: this needs to go above the `checkConfigUserFilter` call; those are simply warnings, not fatal errors and the transaction should still be committed rather than rolled back by the defer above.

cipherboy:
Thank you @suprjinx!

Issue #679, opened 2024-11-07
RFC - SSH CA Multi-issuer


Gabrielopesantos:
### Summary

OpenBao’s SSH engine is limited to one Certificate Authority (CA) issuer per mount. This complicates the key rotation process, which is usually a desired feature for security and operational continuity. As previously done in the [PKI engine](https://github.com/hashicorp/vault/pull/15277), this document proposes an implementation to support multiple issuers per SSH mount. This proposal tries to stay consistent with the PKI implementation, having a ‘default’ issuer and a set of new endpoints to read/update it, and not break any backward compatibility with older versions of OpenBao or Upstream.

### Problem Statement

OpenBao’s current SSH engine design imposes limitations by allowing only one Certificate Authority (CA) issuer per mount. This restriction complicates the key rotation process, an important feature for security and operational continuity. To perform key rotation, new SSH mounts must be created, which necessitates duplicating all existing configurations (roles, ACLs, etc.) and updating all references to the new mounts or swapping the old and the new mount points via [two move operations](https://openbao.org/api-docs/system/remount/). This is not only time-consuming but prone to misconfigurations and errors.

While an alternative approach–deleting and recreating the CA issuer within an existing mount–can be used, it introduces significant risks. During the gap between deleting the old key and creating a new one, any requests for issuing access certificates fail. If the import or generation of a new CA issuer fails, this leads to service disruption. Even if the new key is created quickly, it lacks trust, as it is a fresh random key that cannot be cross-signed like in PKI. The key distribution process to all hosts is not instantaneous, leaving a window where authentication requests will fail.

This limitation highlights the need for a safer key rotation mechanism that ensures the system can securely fetch and trust new keys before the actual rotation occurs. Moreover, the ability to revert to a previous key in case of a failure or unexpected events would provide fallback capabilities.

### User-facing description

With this feature, OpenBao’s ‘privileged’ users can submit CA information for more than one issuer, having the ability to support multiple issuers in the same SSH mount. Besides the unique identifier, each issuer will optionally have a unique name that can be used to identify it. The concept of a ‘default’ issuer will be introduced to the existing endpoints, that use the issuer’s key material, stay backwards compatible. The ‘default’ issuer may be read/updated through the newly introduced endpoint, config/issuers. Roles allow the selection of an explicit issuer, allowing operators to choose issuers for users without them necessarily noticing the impact.
The behavior of existing endpoints will be adapted to conform with this feature. New endpoints will also be added so operations on CA’s can be performed.

[...]

PR #880, opened 2025-01-09
Support Multiple Issuers for SSH Secret Engine Mounts

Gabrielopesantos:
# Description

This pull requests adds support for configuration and use of multiple issuers under the same SSH secrets engine mount point, as described in #679. A new major API path is exposed, /issuer{s}, for managing issuers.

Existing APIs are preserved and will continue to function, performing slightly different operations but maintaining expected behavior when an issuer is configured. A "default" issuer has been introduced to automatically determine the issuer when existing APIs require this information. This can be configured through multiple endpoints, such as config/ca, config/issuers, and issuers/import.

Roles now support explicit issuer selection, allowing operators to assign issuers to users without them necessarily noticing any changes.

Unless otherwise stated, many of these new APIs should be considered privileged and should have appropriate access controls (ACLs) in place.

Additionally, when migrating to the Bao version that introduces this feature, existing CA information will be converted to the new format, with the issuer set as the default.

# Issue

#679


builtin/logical/ssh/fields.go
```diff
@@ -0,0 +1,41 @@
+package ssh
+
+import "github.com/openbao/openbao/sdk/v2/framework"
+
+func addSubmitIssuerCommonFields(fields map[string]*framework.FieldSchema) map[string]*framework.FieldSchema {
```
cipherboy, line 5: I do rather like this pattern, since it keeps `path_*.go` files shorter. Thanks, @Gabrielopesantos!

builtin/logical/ssh/path_roles.go
```diff
@@ -554,6 +562,13 @@ func (b *backend) createCARole(allowedUsers, defaultUser, signer string, data *f
 	role.DefaultExtensions = defaultExtensions
 	role.AllowedUserKeyTypesLengths = allowedUserKeyLengths
 
+	// Just like in PKI, we ensure that the issuers ref is set to an non-empty value,
```
cipherboy, line 565: Nice!

builtin/logical/ssh/path_fetch.go
```diff
@@ -23,24 +23,27 @@ func pathFetchPublicKey(b *backend) *framework.Path {
...
+	issuer, err := sc.fetchDefaultIssuer()
 	if err != nil {
-		return nil, err
+		return handleStorageContextErr(err)
 	}
-	if publicKeyEntry == nil || publicKeyEntry.Key == "" {
-		return nil, nil
+
+	if issuer == nil {
+		return logical.ErrorResponse("No keyuaslkdjnmaskldjaskld"), nil
```
cipherboy, line 40: @Gabrielopesantos This looks unintentional?

```suggestion
		return logical.ErrorResponse("No key corresponding to issuer"), nil
```

builtin/logical/ssh/path_issue_sign.go
```diff
@@ -101,15 +98,19 @@ func (b *backend) pathSignIssueCertificateHelper(ctx context.Context, req *logic
 		return logical.ErrorResponse(err.Error()), nil
 	}
 
-	privateKeyEntry, err := caKey(ctx, req.Storage, caPrivateKey)
+	// TODO (gabrielopesantos): Review
```
cipherboy, line 101: I think this makes sense. Did you have thoughts about what you wanted to review?

builtin/logical/ssh/path_config_issuers.go
```diff
@@ -0,0 +1,139 @@
...
+	configIssuerSchema := map[int][]framework.Response{
+		http.StatusOK: {{
+			Description: "OK",
+			Fields: map[string]*framework.FieldSchema{
+				"default": {
+					Type:        framework.TypeString,
+					Description: `Reference (name or identifier) to the default issuer.`,
+					Required:    true,
+				},
+			},
```
cipherboy, line 21: I forget, did we talk about `default_follows_latest_issuer` and what we felt about that?

builtin/logical/ssh/backend.go
```diff
@@ -20,6 +21,8 @@ type backend struct {
 	view      logical.Storage
 	salt      *salt.Salt
 	saltMutex sync.RWMutex
+	// Write lock around issuers
+	issuersLock sync.RWMutex
```
cipherboy, line 25:
```suggestion
	issuersLock sync.Mutex
```

I don't think we use the read lock anywhere?

builtin/logical/ssh/storage.go
```diff
@@ -0,0 +1,301 @@
...
+
+	// NOTE (gabrielopesantos): These might be to high for SSH issuers
+	maxRolesToScanOnIssuerChange = 100
+	maxRolesToFindOnIssuerChange = 10
+
+	// Used as a quick sanity check for a reference id lookups...
+	uuidLength = 36
+)
+
+type issuerID string
```
cipherboy, line 35: This is nice in the PKI land, so to not confuse `keyID` vs `issuerID` but may not be necessary here. It added a bit of problems in PKI in particular, though I think one of those problems will be fixed by #887 tbh (typing).

cipherboy (commented):
I've not yet had time to look at everything, but some initial thoughts! 

Looking fairly good from what I see :-)

builtin/logical/ssh/backend.go
```diff
@@ -53,6 +57,8 @@ func Backend(conf *logical.BackendConfig) (*backend, error) {
 				caPrivateKey,
 				caPrivateKeyStoragePath,
 				keysStoragePrefix,
+				fmt.Sprintf("%s+", issuerPrefix),
```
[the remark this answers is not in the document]
Gabrielopesantos: Yes, you are right. Accepting suggestion.

builtin/logical/ssh/backend_test.go
```diff
@@ -2738,18 +2738,24 @@ func TestProperAuthing(t *testing.T) {
...
+		"roles/test-otp":            shouldBeAuthed,
+		"roles":                     shouldBeAuthed,
+		"sign/test-ca":              shouldBeAuthed,
+		"tidy/dynamic-keys":         shouldBeAuthed,
+		"verify":                    shouldBeUnauthedWriteOnly,
+		"issuer/default/public_key": shouldBeUnauthedReadList,
+		"issuer/default":            shouldBeAuthed,
+		"issuers/import/test":       shouldBeAuthed,
+		"issuers/import":            shouldBeAuthed,
+		"issuers":                   shouldBeAuthed,
```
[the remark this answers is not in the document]
Gabrielopesantos: Sure, I have committed a "sorted" order but let me know if its aligned with what you were lf.

builtin/logical/ssh/path_issue_sign.go
Gabrielopesantos: I must have addressed whatever I wanted to review and forgot to review the comment.

builtin/logical/ssh/path_config_issuers.go
Gabrielopesantos: Yes, we did. I don't quite remember the details, but I believe we agreed that it was unnecessary as there's an endpoint to submit issuers with optional names. Let me know what you think, it's also possible that I have extrapolated this out of some discussion we had.

builtin/logical/ssh/storage.go
Gabrielopesantos: That sounds good, I defined this type to stay consistent with PKI but don't seem much use in having it, going to revert.

builtin/logical/ssh/path_config_issuers.go
cipherboy: Nope, you're right. There's no hierarchy and the import-with-name handles most of the remaining use case here.

builtin/logical/ssh/util.go
```diff
@@ -135,3 +148,134 @@ func substQuery(tpl string, data map[string]string) string {
 
 	return tpl
 }
+
+// keys parses the input parameters and returns the public and private keys
+// by either generating them or using the provided ones.
+func (b *backend) keys(data *framework.FieldData) (string, string, error) {
```
cipherboy, line 154: Could we name this a little more descriptively? I was expecting this to be storage-level but fetching but was confused by the comment initially that it was about the parameter parsing.

This was moved from `path_config_ca` if I'm not mistaken?

builtin/logical/ssh/path_issue_sign.go
```diff
@@ -101,15 +98,18 @@ func (b *backend) pathSignIssueCertificateHelper(ctx context.Context, req *logic
...
 	}
-	if privateKeyEntry == nil || privateKeyEntry.Key == "" {
-		return nil, errors.New("failed to read CA private key")
+	issuer, err := sc.fetchIssuerById(issuerId)
+	if err != nil {
+		return handleStorageContextErr(err)
 	}
 
-	signer, err := ssh.ParsePrivateKey([]byte(privateKeyEntry.Key))
+	privateKey := issuer.PrivateKey
```
cipherboy, line 110: This will be fun to deal with when it comes time to do PKCS#11 HSM support from the working group, but that's not a you problem now :-)

builtin/logical/ssh/path_issuers.go
```diff
@@ -0,0 +1,517 @@
...
+		},
+
+		HelpSynopsis:    pathListIssuersSyn,
+		HelpDescription: pathListIssuersDesc,
+	}
+}
+
+func pathGetIssuerPublicKeyUnauthenticated(b *backend) *framework.Path {
+	return &framework.Path{
+		Pattern: "issuer/" + framework.GenericNameRegex("issuer_ref") + "/public_key",
```
cipherboy, line 173: Is this strictly a raw endpoint or do we have JSON endpoints like PKI does?

builtin/logical/ssh/path_issuers.go
```diff
@@ -0,0 +1,517 @@
...
+		DisplayAttrs: &framework.DisplayAttributes{
+			OperationPrefix: operationPrefixSSH,
+			OperationVerb:   "get",
+			OperationSuffix: "issuer",
+		},
+
+		Fields: map[string]*framework.FieldSchema{
+			"issuer_ref": {
+				Type:        framework.TypeString,
+				Description: `Issuer reference. It can be the issuer's unique identifier, or the optionally given name.`,
```
cipherboy, line 184: This might apply in other places, but can't it also take the value `default`?

builtin/logical/ssh/path_issuers.go
```diff
@@ -0,0 +1,517 @@
...
+	b.issuersLock.Lock()
+	defer b.issuersLock.Unlock()
+
+	publicKey, privateKey, err := b.keys(d)
+	if err != nil {
+		return handleStorageContextErr(err)
+	}
+
+	// Use the transaction storage if there's one.
+	if txnStorage, ok := req.Storage.(logical.TransactionalStorage); ok {
```
cipherboy, line 357: As above about the helpers.

builtin/logical/ssh/path_issuers.go
```diff
@@ -0,0 +1,517 @@
...
+	if setDefault {
+		// Update issuers config to set new issuer as the 'default'
+		err = sc.setIssuersConfig(&issuerConfigEntry{DefaultIssuerID: id})
+		if err != nil {
+			// It is not possible to have this error in the transaction, so check
+			// storage type and skip if is a transaction
+			if _, ok := req.Storage.(logical.Transaction); !ok {
+				// Even if the new issuer fails to be set as default, we want to return
+				// the newly submitted issuer with a warning
+				response.AddWarning(fmt.Sprintf("Unable to update the default issuers configuration: %s", err.Error()))
```
cipherboy, line 403: I'd probably do this regardless of whether or not we're in a transaction. I'd hate to swallow an error from the transaction layer just because we didn't think it was possible.

builtin/logical/ssh/path_config_ca.go
```diff
@@ -186,127 +193,73 @@ func caKey(ctx context.Context, storage logical.Storage, keyType string) (*keySt
...
-
-		_, err = parsePublicSSHKey(publicKey)
-		if err != nil {
-			return logical.ErrorResponse(fmt.Sprintf("Unable to parse public_key as an SSH public key: %v", err)), nil
-		}
-
-	// not set and no public/private key provided so generate
-	case publicKey == "" && privateKey == "":
-		generateSigningKey = true
+	// Since we're planning on updating issuers here, grab the lock so we've
```
cipherboy, line 196: Is there some way to unify the code with the other version of this handler a bit more? This goes to the same `b.pathImportIssuers` in PKI (in `pki/path_config_ca.go`). 

`req` gives you `Path` as a field if you need to condition some small behavior difference on a given path. 

Same applies for delete or read.

builtin/logical/ssh/path_config_issuers.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	// Update the config
+	config, err := sc.getIssuersConfig()
+	if err != nil {
+		return handleStorageContextErr(err, "unable to fetch existing issuers configuration")
+	}
+
+	oldDefault := config.DefaultIssuerID
+	config.DefaultIssuerID = parsedIssuer
```
cipherboy, line 115: In PKI we left this in ref form and didn't resolve it down to an ID:

```go
    config.DefaultIssuerId = parsedIssuer
    if followOk {
        config.DefaultFollowsLatestIssuer = followIssuer
    }
```

I'd do the same here, so that you can rename issuers and have the renamed one be the new default.

builtin/logical/ssh/path_config_issuers.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	response := &logical.Response{
+		Data: map[string]interface{}{
+			defaultRef: config.DefaultIssuerID,
+		},
+	}
+	if len(oldDefault) != 0 {
+		warningMessage := fmt.Sprintf("Previous default (%s) has been updated with the issuer '%s'", oldDefault, newDefault)
```
cipherboy, line 134: I'm curious why this warning was added?

builtin/logical/ssh/storage.go
```diff
@@ -0,0 +1,294 @@
...
+			}
+		}
+		checkedRoles = checkedRoles + 1
+		if checkedRoles >= maxRolesToScanOnIssuerChange {
+			return true, inUseBy, nil
+		}
+	}
+
+	return false, inUseBy, nil
+}
```
cipherboy, line 294: Wow this really is so much simpler than PKI :-D

builtin/logical/ssh/storage.go
```diff
@@ -0,0 +1,294 @@
...
+func (sc *storageContext) deleteIssuer(id string) (bool, error) {
+	config, err := sc.getIssuersConfig()
+	if err != nil {
+		return false, err
+	}
+
+	wasDefault := false
+	if config.DefaultIssuerID == id {
+		wasDefault = true
+		config.DefaultIssuerID = ""
```
cipherboy, line 89: Just pointing out that there'll be a few places that need to be updated to handle the default being a ref rather than a concrete ID.

builtin/logical/ssh/storage_migrations.go
```diff
@@ -0,0 +1,114 @@
...
+
+// This allows us to record the version of the migration code within the log entry
+const (
+	latestMigrationVersion = 1
+)
+
+type migrationInfo struct {
+	Created          time.Time `json:"created"`
+	MigrationVersion int       `json:"migrationVersion"`
+}
```
cipherboy, line 20: Don't we still need to store the last-migrated legacy bundle and compare it to the current version in storage? If it is present and differs, it means we were downgraded and modified and thus need to rerun migration on upgrade.

builtin/logical/ssh/storage.go
```diff
@@ -0,0 +1,294 @@
...
+func (b *backend) makeStorageContext(ctx context.Context, s logical.Storage) *storageContext {
+	return &storageContext{
+		Context: ctx,
+		Storage: s,
+		Backend: b,
+	}
+}
+
+// writeIssuer writes an issuerEntry to storage
+func (sc *storageContext) writeIssuer(issuer *issuerEntry) error {
```
cipherboy, line 67: @Gabrielopesantos On the call today, we noticed that we need an `importIssuer(...)` so that if you call the `config/ca` endpoint multiple times with the same key material, it won't endlessly create new issuers with the same key material, but instead point out that it already exists (single entry).

builtin/logical/ssh/util.go
Gabrielopesantos: Yes, most of it is the code that was in "Submit CA Information" endpoint handler.

builtin/logical/ssh/path_issue_sign.go
Gabrielopesantos: Interesting, how so?

builtin/logical/ssh/path_issuers.go
Gabrielopesantos: It should as, If I'm not mistaken and going to confirm, we always use `resolveIssuerReference` to resolve issuer references provided. And this function does handle `default` being passed.

```
	if ref == defaultRef {
		// If reference is 'default', fetch the default issuer ID from the configuration
		issuerConfig, err := sc.getIssuersConfig()
		if err != nil {
			return "", err
		}
		if len(issuerConfig.DefaultIssuerID) == 0 {
			return IssuerRefNotFound, errutil.UserError{Err: "no default issuer currently configured"}
		}

		return issuerConfig.DefaultIssuerID, nil
	}
...
```

builtin/logical/ssh/path_issuers.go
Gabrielopesantos: Didn't see what PKI did here, I have based myself more on what we had before. This is raw, unauthenticated, returning only the public key in plain text format. 

Then there's the authenticated endpoint which returns a JSON response with the fields `issuer_id`, `issuer_name` and `public_key`.

Of course, both of these accept an issuer reference or `default`. I'm going to see how does this align with PKI.

builtin/logical/ssh/path_issuers.go
cipherboy: Perfect, want to add that to this description? Thanks!

builtin/logical/ssh/backend.go
```diff
@@ -128,3 +141,52 @@ After mounting this backend, before generating credentials, configure the
...
+
+func (b *backend) initializeIssuersStorage(ctx context.Context) error {
+	// Grab the lock prior to the updating of the storage lock preventing us flipping
+	// the storage flag midway through the request stream of other requests.
+	b.issuersLock.Lock()
+	defer b.issuersLock.Unlock()
+
+	// Use the transaction storage if there's one.
+	storage := b.view
+	if txnStorage, ok := b.view.(logical.TransactionalStorage); ok {
```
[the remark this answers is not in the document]
Gabrielopesantos: Going to leave this for the end.

builtin/logical/ssh/path_issuers.go
Gabrielopesantos: Yes, added. Just haven't pushed the commit yet.

builtin/logical/ssh/path_config_ca.go
Gabrielopesantos: I'm not sure if I like the approach that I have followed but let me know what you think. I introduced a new variable that checks the incoming request path and based on that decides what code to execute.
```go
isConfigCARequest := req.Path == "config/ca"
```

builtin/logical/ssh/path_config_issuers.go
Gabrielopesantos: As discussed, the issuer's identifier being resolved at creation is also the of PKI and for that reason we are going to stay with that implementation here.

builtin/logical/ssh/storage.go
Gabrielopesantos: Given what we discussed [here](https://github.com/openbao/openbao/pull/880#discussion_r1950871123), not anymore I assume?

cipherboy: Yep :-)

builtin/logical/ssh/path_issue_sign.go
```diff
@@ -53,9 +52,7 @@ type creationBundle struct {
 	Extensions      map[string]string
 }
 
-func (b *backend) pathSignIssueCertificateHelper(ctx context.Context, req *logical.Request, data *framework.FieldData, role *sshRole, publicKey ssh.PublicKey) (*logical.Response, error) {
-	// Note that these various functions always return "user errors" so we pass
-	// them as 4xx values
+func (b *backend) pathSignIssueCertificateHelper(sc *storageContext, req *logical.Request, data *framework.FieldData, role *sshRole, publicKey ssh.PublicKey) (*logical.Response, error) {
```
cipherboy, line 55: In the response, it'd be great to inject the `issuer_ref` as well. I don't know if PKI does this, but given SSH certs are a little opaque, it'd be good to have a correlation ID.

builtin/logical/ssh/path_issuers.go
```diff
@@ -0,0 +1,542 @@
...
+
+	setDefault := isConfigCARequest || d.Get("set_default").(bool)
+	if setDefault {
+		err = sc.updateDefaultIssuerId(issuer.ID)
+		if err != nil {
+			// Even if the new issuer fails to be set as default, we want to return
+			// the newly submitted issuer with a warning
+			response.AddWarning(fmt.Sprintf("Unable to update the default issuers configuration: %s", err.Error()))
+		}
+	}
```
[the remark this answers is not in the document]
Gabrielopesantos: Right, I did not consider this behavior. This endpoint does have a flag `set_default` that, if set, will make the issuer as default otherwise not.

builtin/logical/ssh/storage.go
Gabrielopesantos: What if the key material is from an issuer that is not set as default and when one being imported is being set as default. Do we want to set the existing issuer as the default?

cipherboy: If there's an existing default, I'd leave it alone in `importIssuer(...)` and only set one if there's no issues at all. Then the `config/ca` can do the write always and `issuers/import/<name>` can do it conditionally on the `set_default` value.

builtin/logical/ssh/backend.go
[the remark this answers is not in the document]
Gabrielopesantos: Yes, nicer to user. Updated.

cipherboy:
@Gabrielopesantos Looks like you're missing DCO in a few commits :/ Do you want to `git rebase --signoff main` perhaps?

builtin/logical/ssh/storage.go
```diff
@@ -0,0 +1,385 @@
...
+
+	caPublicKey                       = "ca_public_key"
+	caPrivateKey                      = "ca_private_key"
+	caPublicKeyStoragePath            = "config/ca_public_key"
+	caPublicKeyStoragePathDeprecated  = "public_key"
+	caPrivateKeyStoragePath           = "config/ca_private_key"
+	caPrivateKeyStoragePathDeprecated = "config/ca_bundle"
+
+	maxRolesToScanOnIssuerChange = 100
+	maxRolesToFindOnIssuerChange = 10
```
Gabrielopesantos, line 31: Thoughts on these values for SSH? I copied them from PKI but might be too high here?

cipherboy: I think it makes sense, shouldn't be too much work to do.

builtin/logical/ssh/backend_test.go
```diff
@@ -2820,3 +2729,774 @@ func TestProperAuthing(t *testing.T) {
...
+	return logicaltest.TestStep{
+		Operation: logical.UpdateOperation,
+		Path:      "config/issuers",
+		Data:      parameters,
+	}
+}
+
+// TestSSHBackend_IssuerLifecycle tests the complete lifecycle of issuers including
+// creation, configuration, updates, and deletion
+func TestSSHBackend_IssuerLifecycle(t *testing.T) {
```
cipherboy, line 2755: I'd mark this `t.Parallel()` and maybe consider doing the same for the subtests.

builtin/logical/ssh/backend_test.go
```diff
@@ -2820,3 +2729,774 @@ func TestProperAuthing(t *testing.T) {
...
+		require.EqualValues(t, issuerPublicKey1, issuerPublicKey2, "expected public keys to match")
+	})
+}
+
+// TestSSHBackend_BasicIssuerOperations tests the basic creation and usage of issuers
+func TestSSHBackend_BasicIssuerOperations(t *testing.T) {
+	b, s := CreateBackendWithStorage(t)
+
+	// Test creating first issuer and automatic default setting
+	t.Run("first_issuer_becomes_default", func(t *testing.T) {
```
cipherboy, line 3078: As above, though you could remove the subtest here.

builtin/logical/ssh/backend_test.go
```diff
@@ -2820,3 +2729,774 @@ func TestProperAuthing(t *testing.T) {
...
+			t.Fatalf("did not expect error, got err: %v", err)
+		}
+	})
+}
+
+// TestSSHBackend_DefaultIssuerBehavior tests default issuer mechanics
+func TestSSHBackend_DefaultIssuerBehavior(t *testing.T) {
+	b, s := CreateBackendWithStorage(t)
+
+	t.Run("default_issuer_override", func(t *testing.T) {
```
cipherboy, line 3173: And here.

builtin/logical/ssh/backend_test.go
```diff
@@ -2820,3 +2729,774 @@ func TestProperAuthing(t *testing.T) {
...
+
+		issuer2Req := &logical.Request{
+			Operation: logical.UpdateOperation,
+			Path:      "issuers/import/issuer4",
+			Storage:   s,
+		}
+		resp, err = b.HandleRequest(context.Background(), issuer2Req)
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+		issuer4ID := resp.Data["issuer_id"].(string)
```
cipherboy, line 3416: The numbers differ from the request numbers :-)

builtin/logical/ssh/path_config_issuers.go
cipherboy: Hmm, ok, we'll keep it for now. I guess it is more useful in SSH than PKI since the default behavior is a touch more complex over here.

builtin/logical/ssh/path_issue_sign.go
cipherboy: We'll need a `crypto.Signer` interface, rather than the actual byte material of the private key; it tends not to be extractable.

builtin/logical/ssh/path_issuers.go
cipherboy: Good enough for now. We can revisit and add a JSON endpoint in the future. Thanks!

builtin/logical/ssh/storage_migrations.go
cipherboy: @Gabrielopesantos I guess, perhaps because issuer id is 1:1 to issuer key  material, IDs is sufficient as long as you never remove the key material as well.

builtin/logical/ssh/storage_migrations_test.go
```diff
@@ -0,0 +1,142 @@
...
+	require.NoError(t, err)
+	require.Equal(t, log, newLog)
+
+	// Update key material in CA and run storage migration again,
+	// as the key material is different a new migration should run
+	startTime = time.Now()
+
+	// Set CA public key
+	json, err = logical.StorageEntryJSON(caPublicKeyStoragePath, &keyStorageEntry{
+		Key: testCAPublicKeyEd25519,
```
cipherboy, line 111: A scenario I'm thinking about is, start with an empty mount, run a migration of nothing, downgrade, write key material, and try and upgrade again.

cipherboy (commented):
Hopefully the last round!

builtin/logical/ssh/storage_migrations.go
cipherboy: Ah, sorry, I think I was looking at an old version of this code. Resolving! Thank you!!

website/content/api-docs/secret/ssh.mdx
```diff
@@ -349,23 +459,23 @@ $ curl \
     http://127.0.0.1:8200/v1/ssh/roles/my-role
 ```
 
-## List Zero-Address roles
+### List Zero-Address Roles
```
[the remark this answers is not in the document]
Gabrielopesantos: Yes, sure.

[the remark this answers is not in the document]
Gabrielopesantos: Sure, it seems I wasn't really following this rule anyway as only a few titles were updated.

builtin/logical/ssh/storage_migrations.go
```diff
@@ -0,0 +1,158 @@
...
+	publicKeyCaEntry, err := caKey(ctx, s, caPublicKey)
+	if err != nil {
+		return
+	}
+
+	if publicKeyCaEntry != nil {
+		publicKey = publicKeyCaEntry.Key
+	}
+
+	privateKeyCaEntry, err := caKey(ctx, s, caPublicKey)
```
cipherboy, line 140: @Gabrielopesantos Is this a bug? Shouldn't this be `caPrivateKey`?

Gabrielopesantos: Yes, it is. Nice catch, this probably should have been caught by a test. :thinking:

Gabrielopesantos: Ok, fix pushed.

cipherboy (approved):
I think this is great, thank you @Gabrielopesantos! If there's any feedback from others, we can pick it up and revise it in later PRs but I don't see anything substantial to prevent merge now.

builtin/logical/ssh/util.go
```diff
@@ -135,3 +148,124 @@ func substQuery(tpl string, data map[string]string) string {
...
+		}
+
+		_, err = ssh.ParsePrivateKey([]byte(privateKey))
+		if err != nil {
+			err = errutil.UserError{Err: fmt.Sprintf("could not parse private_key provided value: %v", err)}
+			return
+		}
+	// generation of signing key not set and no key material provided so generate
+	case publicKey == "" && privateKey == "":
+		generateSigningKey = true
```
cipherboy, line 182: @Gabrielopesantos Did we talk about this? In the current path, it throws an error if `config/ca` is called without an explicit `generate_signing_key=true`, but after multi-issuers, it'll now silently create a key. Thoughts? 

cc: @fatima2003

Gabrielopesantos: Not sure I understood what's happening, going to have a look later today. I remember testing this execution path and it was working correctly. 🤔

Gabrielopesantos: > In the current path, it throws an error if config/ca is called without an explicit generate_signing_key=true
I'm not seeing this behavior, new issuers are being created even if I don't include the `generate_signing_key` param.

cipherboy: @Gabrielopesantos Sorry, wrong description of the bug.

On main:

```
$ bao write -f ssh/config/ca generate_signing_key=false
Key            Value
---            -----
issuer_id      f65db5f5-8644-0c42-4542-fe962f1be2f4
issuer_name    n/a
public_key     ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDwjqVyzjVDu+xHQP9ZkcjisByU78E5UgfOtU4Tg8ksvKoJWLvvlgbYg0Aiy5bz2/5SheGHbQK8Ywb7a/k3rORTxDbCx3xY7hkpwmtYcrLF+W5HaXcz0WbWIkY5DU9nb943k6Uov+GnnUYkNosGmtsSexbBLP1JYrlI/fGI3hEUXB14VofpB2l0PpdtGQ1NAcOjFLv5gXf0h5/pfNlY2fO0cQeKsMNPCUk295n+YnooYa0xZVI45FnnUwWyILoc2O9QgphiMMsHVFPwPOOw7TTsivBHTQdQOfMsywpWQmwe3fQGOuZmxKK7DJU7wJ/QxA23nA/hInUmuX7WxT2A79bZpoii0H+uLcflmb30vOFH3KsJyz9sdCZD+aff6DFKbpn6xpeJ4z4+pIl46GtMZgXimxm+40zAueoRhoxhBbKJX0mD5Z+geexgbUVlODsIOKfd/Vw4ekDMFJe8sZ7OSa2q6L6BbFnMlvko/8znUjp98mDdcJ9B+lOq6+MnoeYKRNgdX0xSXe9X2/XRzBC2KC6uAmc4qexOCklDYkjQLYUPWB069FsF/DTTducX0roS0pDjX8tyxOGnr/utj4wogRCh1FXznA6WAshIiJ7lolX4PG0xfjAYSAg6+LDmKO2R8yAkcA8j5a1mgeeUAYXpubWKHfJokWQMdAH1AxEJdjkNiQ==
```

On v2.2.0:

```
$ bao write -f ssh/config/ca generate_signing_key=false
Error writing data to ssh/config/ca: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/ssh/config/ca
Code: 400. Errors:

* missing public_key
```

cipherboy: I think this needs to be:

```
case publicKey == "" && privateKey == "" && !ok
```

and we need another case:

```
case publicKey == "" && privateKey == "" && ok && !generateSigningKeyRaw.(bool):
    err = errutil.UserError{Err: "missing public_key"}
    return
```

Gabrielopesantos: https://github.com/openbao/openbao/pull/1235

Worth having a changelog for this fix?

cipherboy: @Gabrielopesantos Nah, multi-issuers haven't shipped yet

PR #886, opened 2025-01-13
Fix getting started guide

katmayo7:
Updated step 4 Bash script to include echo statement to return "Secret written successfully" when the command terminates successfully.

Updated step 5 so the documentation only mentions a return of "Access granted" when looking at the Go code.

### DIFF
--- a/website/content/docs/get-started/developer-qs.mdx
+++ b/website/content/docs/get-started/developer-qs.mdx
@@ -126,7 +126,8 @@ $ curl \
     --header "Content-Type: application/json" \
     --request POST \
     --data '{"data": {"password": "OpenBao123"}}' \
-    http://127.0.0.1:8200/v1/secret/data/my-secret-password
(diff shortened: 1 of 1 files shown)


website/content/docs/get-started/developer-qs.mdx
```diff
@@ -184,17 +185,21 @@ if value != "OpenBao123" {
...
 ```
 
+If the secret was fetched successfully, you should see the `Access granted!` message after you run the code.
+
 </TabItem>
 <TabItem value="Bash" group="bash" heading="Bash" group="bash">
 
 ```shell-session
-$ cat secrets.json | jq '.data.data'
+$ cat secrets.json | jq '.data.data'
```
cipherboy, line 194:
```suggestion
$ cat secrets.json | jq '.data.data'
```

cipherboy (approved):

cipherboy:
Thanks @katmayo7!

Issue #478, opened 2024-08-20
Audit logging panic investigation


cipherboy:
**Describe the bug**

A well-known problem with the audit logging subsystem today is that, because it rewrites the response in-place using reflection, we end up being unable to audit certain types, causing panics like #97. 

In particular, the root cause in this case is that the struct is referenced from a map by value, rather than by reference:

```go
// Role that defines the capabilities of the credentials issued against it.
// Maps are used because the names of vhosts and exchanges will vary widely.
// VHosts is a map with a vhost name as key and the permissions as value.
// VHostTopics is a nested map with vhost name and exchange name as keys and
// the topic permissions as value.
type roleEntry struct {
    Tags        string                                      `json:"tags" structs:"tags" mapstructure:"tags"`
    VHosts      map[string]vhostPermission                 `json:"vhosts" structs:"vhosts" mapstructure:"vhosts"`
    VHostTopics map[string]map[string]vhostTopicPermission `json:"vhost_topics" structs:"vhost_topics" mapstructure:"vhost_topics"`
}
```

when using `structs.New(role).Map()`, we do get a `map[string]interface{}` out, but this is not recursive: the interior type of the `vhost` key is still `map[string]vhostPermission`. When we go to [walk the response for auditing](https://github.com/openbao/openbao/blob/main/audit/hashstructure.go#L369-L371), we panic because the struct field is not addressable because it comes from a value and not a pointer reference.

In this case, we have two options:

1. Build a parallel, addressable copy of the struct from the parent map (when the current struct is not accessible), and write the modified struct to the map when finished.
2. Don't [copy the struct](https://github.com/openbao/openbao/blob/7fca5c0baebd3f55254da06d26bc160f465a7e1a/audit/hashstructure.go#L131-L134), but instead, serialize it to JSON (because `resp.Data` must be serializable), and audit log the de-serialized form.

While 1 is likely strictly faster (because it only incurs cost in the few places where structs are responded with due to this bug), 2 is likely more robust and allows us to avoid the dependency on an unmaintained package (as `mitchellh` has archived many of his repositories, some without replacement). Thus, I'd personally lean towards two (deserializing into `map[string]interface{}`).

We should also probably remove the [last usages of `structs.New(..)`](https://github.com/search?q=repo%3Aopenbao%2Fopenbao%20structs.New(&type=code) and replace it with custom serializers in all places so it isn't seen as a pattern for others. This is also a deprecated package.

**To Reproduce**

[...]

PR #887, opened 2025-01-13
Modify audit logging to handle nested structs without panicing

GeorgeJahad:
## The Problem
Audit logging uses reflection to write the hashed value of logged fields to a copy of the log data.  That write is inherently problematic.  If the log data contains structs, the struct fields may be "unaddressable" with reflection and the write attempt causes a panic, (as described here): https://github.com/openbao/openbao/issues/478

## The fix in this PR
The fix is to remove all structs by marshalling the data to json and back.  This creates a deep copy where all structs/fields have been converted to maps/keys.

Prior to this fix, the hashing code made a copy of the data structure to be logged, and uses reflection to walk the copy and hash the data.

Because the fix uses the json unmarshalled copy, many of the data types get changed, (like structs to maps).  So the fixed code must walk both the original and the copy.  It walks the original to get the type info and the data to hash, and the copy to get the location to be written to.

## Implications
Because of the type changes caused by the unmarshalling, I've had to modify some of the tests, for example to expect "interface{}"'s instead of strings, and floats instead of ints.

I've removed the "logical.optMarshaler" tests because I was told that feature doesn't need to be maintained, (and is broken by unmarshalling.)

I've had to remove one of the audit.TestHashWalker_TimeStructs() tests because it tested that keys of type time.Time should not be modified.  With the unmarshalling of the data, there is no way to avoid modifying non-string keys.  In json all keys must be strings, so the unmarshalling converts all keys to strings, (including int keys and struct keys.)

## Issues
I'm a little wary of this fix, because it is hard to predict all implications of the type changes implicit in unmarshalling.  (And I'm an openbao newbie and have very little context about those implications may affect openbao.)  So please give that some consideration when evaluating this PR.

Also, https://github.com/openbao/openbao/issues/478 requests that all uses of structs.New() be replaced with custom serializers.  I haven't done that in this PR.  It is complicated enough already and recommend a second PR be created for the custom serializers.

### DIFF
--- a/audit/format.go
+++ b/audit/format.go
@@ -583,11 +583,11 @@ func doElideListResponseDataWithCopy(inputData map[string]interface{}, outputDat
 	for k, v := range inputData {
 		if k == "keys" {
 			if vSlice, ok := v.([]string); ok {
-				outputData[k] = float64(len(vSlice))
+				outputData[k] = len(vSlice)
 			}
 		} else if k == "key_info" {
 			if vMap, ok := v.(map[string]interface{}); ok {
-				outputData[k] = float64(len(vMap))
+				outputData[k] = len(vMap)
 			}
 		}
 	}
--- a/audit/format.go
+++ b/audit/format.go
@@ -576,14 +576,18 @@ func NewTemporaryFormatter(format, prefix string) *AuditFormatter {
 //
 // See the documentation of the controlling option in FormatterConfig for more information on the purpose.
 func doElideListResponseData(data map[string]interface{}) {
-	for k, v := range data {
+	doElideListResponseDataWithCopy(data, data)
+}
+
+func doElideListResponseDataWithCopy(inputData map[string]interface{}, outputData map[string]interface{}) {
+	for k, v := range inputData {
 		if k == "keys" {
 			if vSlice, ok := v.([]string); ok {
-				data[k] = len(vSlice)
+				outputData[k] = float64(len(vSlice))
 			}
 		} else if k == "key_info" {
 			if vMap, ok := v.(map[string]interface{}); ok {
-				data[k] = len(vMap)
+				outputData[k] = float64(len(vMap))
 			}
 		}
 	}
--- a/audit/hashstructure.go
+++ b/audit/hashstructure.go
@@ -426,23 +426,24 @@ func (w *hashWalker) Primitive(v reflect.Value) error {
 
 }
 
+// get current value from copy to be written to.
 func (w *hashWalker) getValueFromCopy() reflect.Value {
 	size := len(w.cs)
-	newStruct := w.UnmarshalledCopy
+	currentValue := w.UnmarshalledCopy
 	for i := 0; i < size-1; i++ {
 		switch w.loc[2+2*i] {
 		case reflectwalk.MapValue:
-			newStruct = newStruct.MapIndex(w.csKey[i]).Elem()
+			currentValue = currentValue.MapIndex(w.csKey[i]).Elem()
 		case reflectwalk.SliceElem:
 			index := w.csKey[i].Int()
-			newStruct = newStruct.Index(int(index)).Elem()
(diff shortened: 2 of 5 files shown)


audit/format.go
```diff
@@ -576,14 +576,18 @@ func NewTemporaryFormatter(format, prefix string) *AuditFormatter {
...
-	for k, v := range data {
+	doElideListResponseDataWithCopy(data, data)
+}
+
+func doElideListResponseDataWithCopy(inputData map[string]interface{}, outputData map[string]interface{}) {
+	for k, v := range inputData {
 		if k == "keys" {
 			if vSlice, ok := v.([]string); ok {
-				data[k] = len(vSlice)
+				outputData[k] = float64(len(vSlice))
```
cipherboy, line 586: Just for my own curiosity... I'm curious why you didn't leave this as `int` here?

cipherboy: Ah, I guess JSON marshaling will marshal floats which contain integers to ints, so perhaps this isn't problematic except for direct testing.

audit/hashstructure.go
```diff
@@ -187,11 +193,19 @@ func HashWrapInfo(salter *salt.Salt, in *wrapping.ResponseWrapInfo, HMACAccessor
...
+// The original is walked with the reflectwalk.Walk() method below.
+// That method leaves a trail in w.loc[]/w.csKey[] which is then used
+// by getValueFromCopy() to walk the copy.
 //
 // For the HashCallback, see the built-in HashCallbacks below.
-func HashStructure(s interface{}, cb HashCallback, ignoredKeys []string) error {
-	walker := &hashWalker{Callback: cb, IgnoredKeys: ignoredKeys}
-	return reflectwalk.Walk(s, walker)
+func HashStructure(original interface{}, copy interface{}, cb HashCallback, ignoredKeys []string, elideListResponseData bool) error {
+	walker := &hashWalker{UnmarshalledCopy: reflect.ValueOf(copy), Callback: cb, IgnoredKeys: ignoredKeys, ElideListResponseData: elideListResponseData}
```
cipherboy, line 207: My only comment here is that I wonder if we can walk `copy` directly? Since it is the output of JSON unmarshaling into `map[string]interface{}`, we know the expected structure and can directly index it with type assertions, always to `map[string]interface{}` by virtue of how the JSON deserialization works.

cipherboy: I guess we'd need to reflect at the very last step of `getValueFromCopy` anyways, so probably not worth it.

cipherboy (approved):
Wonderful! And this passes on 8d42b5f684f35dccac9ad5822f77e87277a3f366 as hoped. Thank you @GeorgeJahad!

audit/format.go
GeorgeJahad: I'll take a look.  My initial reason for this change may no longer be valid.

audit/hashstructure.go
GeorgeJahad: walking directly might be a good idea.  (it didn't occur to me to be honest.)

Let me play with it a bit and see.

audit/format.go
cipherboy: Hmm, I think you might be correct. I strangely get type issues if I remove this and/or the value in the tests. Interestingly though, the typing for `Unconventional other values in a list response are not touched` is left as an int, so I do wonder if there's an inconsistency lurking somewhere...

GeorgeJahad: the problem is that the tfw.hashExpectedValueForComparison()  test function doesn't handle the elided data types.  I've added a fix.  Hopefully that will be acceptable.

audit/hashstructure.go
GeorgeJahad: @cipherboy I changed it to do direct writes.  I'm not sure it is any better, but you can take a look.

cipherboy (commented):
Appreciate your updates @GeorgeJahad! Really great to see this type of improvement done to core. Merging...

Issue #300, opened 2024-04-17
Requesting support for HSM Unseal method


ats0stv:
**Is your feature request related to a problem? Please describe.**
Yes, this feature request is related to a problem. Currently, OpenBao does not support integration with Hardware Security Modules (HSM), a critical feature for enhancing security in sensitive environments. Many organizations, especially in sectors like finance and government, require HSM integration to meet compliance standards for cryptographic operations. This feature will enable OpenBao to automatically unseal using secure hardware, reduce operational complexity, and ensure that cryptographic keys are protected according to industry standards such as FIPS 140-2.

**Describe the solution you'd like**
I would like OpenBao to implement HSM integration to support automatic unsealing and enhanced security for encrypted data. This solution should include seamless setup and configuration for HSM devices that adhere to industry standards like FIPS 140-2. The integration should feature detailed documentation and command-line tools to facilitate easy and secure key management operations. Additionally, implementing the Seal Wrap feature will provide an additional layer of security, ensuring sensitive data is wrapped with an extra cryptographic barrier. This integration will make OpenBao a more robust tool for enterprises requiring high security standards.

**Describe alternatives you've considered**
As an alternative to integrating HSM support directly into OpenBao, we considered using software-based encryption keys with enhanced security measures. However in a private environment (non-cloud), the most securely viable is an HSM integration. Additionally, using HashiCorp Vault Enterprise was explored; however, it does not meet our cost constraints, making it less feasible for our project's budget. These alternatives, while potentially viable, do not fully meet the stringent compliance and security requirements that direct HSM integration would achieve.

**Explain any additional use-cases**
Adding HSM integration to OpenBao would also cater to use cases involving high-security applications, such as in government, healthcare, and banking sectors, where regulatory compliance demands hardware-level protection of encryption keys. I feel though an enterprise feature, this is a top priority item to implement as it will enable the adoption of OpenBao project in all environments.

**Additional context**
Integrating HSM with OpenBao aligns with industry best practices for security, particularly in sectors dealing with sensitive information. This feature would not only enhance the platform's attractiveness to prospective enterprise users but also leverage existing community expertise in HSM technologies to foster collaboration and innovation. By providing detailed documentation and examples, the adoption of this feature can be streamlined, encouraging widespread use and contributing to the robustness of OpenBao's security capabilities.

cipherboy:
I think #16 is a duplicate of this.

[...]

PR #889, opened 2025-01-14
Add PKCS#11 auto-unseal to OpenBao

cipherboy:
Major thanks to @glatigny for support in https://github.com/openbao/go-kms-wrapping/pull/14!

Resolves: #300

### DIFF
--- a/website/content/docs/configuration/seal/pkcs11.mdx
+++ b/website/content/docs/configuration/seal/pkcs11.mdx
@@ -1,0 +1,289 @@
+---
+layout: docs
+page_title: PKCS11 - Seals - Configuration
+description: |-
+  The PKCS11 seal configures Vault to use an HSM with PKCS11 as the seal
+  wrapping mechanism.
+---
+<!-- TODO: Remove these Enterprise only docs -->
+# `pkcs11` seal
+
+
+<Note title="Auto-unseal and seal wrapping requires Vault Enterprise">
+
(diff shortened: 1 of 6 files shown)


cipherboy:
@JanMa This will require some careful work on the release process so that we start building cgo-enabled binaries.

website/content/docs/configuration/seal/pkcs11.mdx
```diff
@@ -0,0 +1,156 @@
...
+```
+
+## `pkcs11` parameters
+
+These parameters apply to the `seal` stanza in the OpenBao configuration file:
+
+- `lib` `(string: <required>)`: The path to the PKCS#11 library shared object
+  file. May also be specified by the `BAO_HSM_LIB` environment variable.
+
+  ~> **Note:** Depending on your HSM, the value of the `lib` parameter may be
```
cipherboy, line 63: Needs to be formatted properly

cipherboy: @JanMa @DanGhita This is fixed and should be ready for review. I don't know how easy it is to build a containerized test of this so I might suggest we do manual testing for now.

JanMa (approved):
LGTM, really happy to see this land :rocket: 
Will you adjust the build config to conditionally enable CGO in a follow-up PR?

cipherboy:
@JanMa Yes, I'm working on UI release workflow changes as well; I'll likely spin up another repository for nightly release builds if possible so we can test the workflow before v2.2.0 beta.

PR #892, opened 2025-01-15
Fix expired cert auth test-fixture

cipherboy:


fatima2003 (approved):
LGTM!

cipherboy:
Thanks @fatima2003!

Given this is breaking main, I'm going to go ahead and merge so we get cleaner CI. We may want to consider moving to dynamically generated CAs so that these types of expiration errors do not happen, though we may need to wait for CEL to land so that we can generate certain types of CA and leaf certificates.

Issue #711, opened 2024-11-14
CI: All release platforms should have weekly builds & tests


cipherboy:
As seen in #700 and #710, we need CI to verify all release platforms at least weekly and ideally be triggerable on demand prior to release. Tests using Docker need not pass (though, ideally, an external Docker host should be provided on platforms which don't have native Docker support), but the binaries should be buildable on these platforms.

- [ ] Linux (native GitHub Actions)
    - [x] amd64
    - [ ] arm64
    - [ ] riscv
    - [ ] s390x
    - [ ] ppc64
- [ ] Windows
- [ ] MacOS
- [ ] OpenBSD
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Illumos

As seen with Solaris, if a platform lacks support for OpenBao and its dependencies, it may be dropped from certain releases.

cipherboy:
@DanGhita @JanMa https://github.com/nodejs/build/issues/2876 will be interesting for v2.2.0 assuming #739 is finished. I think what will need to happen is we build the UI once in an earlier stage and then depend on it from cache for all other arches.

cipherboy:
@JanMa @DanGhita Hmm, perhaps problematic, but if we continue to use the matrix form, we'll be limited by https://github.com/gotestyourself/gotestsum/releases , which doesn't support Windows for one.

cipherboy:
So far this would've prevented:

- v2.1.0 beta failure, on solaris/illumos
- v2.3.0 beta failure, on darwin/amd64
- v2.3.0 GA failure on illumos

dc-tec:
I would like to help out with this one

gaige:
I've been building for SmartOS/Illumos regularly and now have a onsistent successful build environment. I'd be interested in contributing the small changes back. Is there a plan for how to support these less-popular platforms in terms of build/test environments? (for example: using cross compilation, or donated runners)?

pree:
I think for tests themselves we can also use something like binfmt_misc through [QEMU User space emulator](https://www.qemu.org/docs/master/user/main.html)

PR #406, opened 2024-07-15
Bump gopsutil/v3 from 3.22.6 to 3.24.5

znley:
This can be help for build and run openbao on loongarch64(a risc architecture). https://github.com/shirou/gopsutil/pull/1228

Resolves #

## Target Release

1.14.7

PR #894, opened 2025-01-15
Update to gopsutil v4

cipherboy:
Updates to the supported series of gopsutil. 

---

Theoretically this should supersede #406. However, unless as shown in #711, unless we have CI validation, we will be unable to release official images for this architecture.


DWwanghao:
LGTM, The loongarch64 architecture mentioned in [#406](https://github.com/openbao/openbao/pull/406) may only be validated through QEMU simulation and cannot be verified via CI. However, overall, as long as the upstream continues to maintain compatibility, v4 is indeed a more reasonable choice.

cipherboy:
Thanks @DWwanghao! 

> validated through QEMU simulation and cannot be verified via CI.

I wonder if the MacOS runner can be used to execute QEMU for this?

JanMa (approved):

cipherboy:
Test failures were because this was opened before #892; merging. Thanks @JanMa!

PR #895, opened 2025-01-15
Go 1.24 fixes

cipherboy:
This fixes various format string errors (using a formatter where none is required mostly), which prevent testing with the upcoming Go 1.24 toolchain. This was tested with 1.24rc1. 

This will be required to address #496 in the future.

### DIFF
--- a/api/auth.go
+++ b/api/auth.go
@@ -5,6 +5,7 @@ package api
 
 import (
 	"context"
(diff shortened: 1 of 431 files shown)


DrDaveD (approved):

cipherboy:
Thanks @DrDaveD, merging :-)

PR #911, opened 2025-01-18
Attempt to add reusable components to dependabot

cipherboy:
This might address why `actions/cache` is out of date in `.github/actions/set-up-go/action.yml` despite dependabot saying there is nothing to do: https://github.com/openbao/openbao/pull/163#issuecomment-2599722950

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,9 @@ updates:
     schedule:
       interval: "weekly"
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/"
+      - "/.github/actions"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


fatima2003 (commented):
I think this would work! :D
Nit: I think we can combine the 2 "github-actions" into 1 entry using [directories](https://docs.github.com/en/enterprise-cloud@latest/code-security/dependabot/working-with-dependabot/dependabot-options-reference?learn=dependency_version_updates&learnProduct=code-security#directories-or-directory--)

fatima2003 (approved):

cipherboy:
Thanks @fatima2003, updated!

cipherboy:
Merging with approval from @fatima2003 as it has no code changes and only affects dependabot.

PR #921, opened 2025-01-20
Allow tuning HMAC status on protected mounts

cipherboy:
When protected mounts were disallowed from tuning in b9a5a137c002211328cc90d9d7e92170d0d86908, only TTL status was present in the tune interface at the time. Tuning HMAC status of request/response in the audit log was not present but makes sense for all mounts, even protected mounts.

Thanks to @jficz for pair coding this!

Resolves: #474

---

Live coding video: https://youtu.be/lr7_YmEIOro


JanMa (approved):
LGTM

PR #924, opened 2025-01-21
Mitigate security risks found using Zizmor

fatima2003:
Found a few vulnerabilities in the workflow using [zizmor](https://github.com/woodruffw/zizmor). 

**Fixed 33 warnings by:**
- Using actions/checkout with persisted credentials to false where possible.
- Using Github Variables.

**Zizmor still returns some warnings:** 
- Persisted credentials are required for further authentication of Git operations after the checkout step in some workflows.
- Secrets unconditionally inherited by some workflows. I didn't modify those since I'm not sure which secrets are required in the workflow.

**References:** 
- [Github default variables]( https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables)
- [Preventing script injection by using env var](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable)
- [Exploring Github Vulnerabilities](https://snyk.io/blog/exploring-vulnerabilities-github-actions/)


.github/workflows/verify_changes.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Get changed directories
         id: get-changeddir
         env:
-          TYPE: ${{ github.event_name }}
-          REF_NAME: ${{ github.ref_name }}
-          BASE: ${{ github.base_ref }}
+          TYPE: ${GITHUB_EVENT_NAME}
```
cipherboy, line 28: I'm slightly confused by this versus `.github/workflows/changelog-checker.yml` -- why are the substitutions OK there but not here?

.github/workflows/verify_changes.yml
```diff
@@ -25,7 +25,7 @@ jobs:
...
+          REF_NAME: ${GITHUB_REF_NAME}
+          BASE: ${GITHUB_BASE_REF}
         run: ./.github/scripts/verify_changes.sh ${{ env.TYPE }} ${{ env.REF_NAME }} ${{ env.BASE }}
```
cipherboy, line 31: Regardless, the actual error seems here: I think we need to quote the environment variables.

cipherboy (commented):
Some thoughts!

.github/workflows/ci.yml
```diff
@@ -33,9 +33,9 @@ jobs:
       # control checking out head instead of default ref by a GH label
       # if checkout-head label is added to a PR, checkout HEAD otherwise checkout ref
       - if: ${{ !contains(github.event.pull_request.labels.*.name, 'checkout-head') }}
-        run: echo "CHECKOUT_REF=${{ github.ref }}" >> "$GITHUB_ENV"
+        run: echo "CHECKOUT_REF=${GITHUB_REF}" >> "$GITHUB_ENV"
       - if: ${{ contains(github.event.pull_request.labels.*.name, 'checkout-head') }}
-        run: echo "CHECKOUT_REF=${{ github.event.pull_request.head.sha }}" >> "$GITHUB_ENV"
+        run: echo "CHECKOUT_REF=${BASE_SHA}" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
fatima2003: Ah I might've removed the initialization by accident when reviewing. Yes, it's supposed to be defined in the env similar to `REPOSITORY_NAME` in `.github/workflows/changelog-checker.yml`

.github/workflows/verify_changes.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Get changed directories
         id: get-changeddir
         env:
-          TYPE: ${{ github.event_name }}
-          REF_NAME: ${{ github.ref_name }}
-          BASE: ${{ github.base_ref }}
+          TYPE: "${GITHUB_EVENT_NAME}"
+          REF_NAME: "${GITHUB_REF_NAME}"
+          BASE: "${GITHUB_BASE_REF}"
         run: ./.github/scripts/verify_changes.sh ${{ env.TYPE }} ${{ env.REF_NAME }} ${{ env.BASE }}
```
cipherboy, line 31: @fatima2003 Ah, I had meant something like this:

```suggestion
        run: ./.github/scripts/verify_changes.sh "${{ env.TYPE }}" "${{ env.REF_NAME }}" "${{ env.BASE }}"
```

fatima2003: Aah makes sense, it felt wrong wrapping those in quotes :D I also think instead of initializing env variables with the default GitHub env variables, we can just use the default variables... maybe that's also why there's an error

fatima2003:
Hi @andreaso, thanks a lot! 😃 I'll prioritize this PR to get it merged sooner. Looking forward to collaborating on workflow security improvements!

fatima2003:
Hi @andreaso, thank you for your input! I'm still getting familiar with workflows, so I was referencing other articles and PR's: like [packaging.python.org #1748](https://github.com/pypa/packaging.python.org/issues/1748), [Exploring Github Vulnerabilities](https://snyk.io/blog/exploring-vulnerabilities-github-actions/) and [Preventing script injection by using env var](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable). This is prolly why my approach has been heavily focused on using environment variables 😅 I appreciate your advice and will try setting output parameters and $GITHUB_PATH where it feels appropriate.

The warnings I was focusing on addressing were `persisting credentials` (by setting `persist-credentials: false`)  and `code injection via template expansion` (by replacing template substitution with environment variables to ensure proper shell quoting). 

Regarding the last point on evaluating whether each Zizmor warning is applicable, I don’t have a specific test for that. My approach was primarily focused on replacing template substitutions in Bash scripts, as that’s where I believe the risk of template injection is most significant, as advised in [GitHub's security hardening guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable).

Really appreciate the advice and look forward to more feedback as I work on this :D

andreaso:
Yeah, setting `persist-credentials: false` (where doable) and getting rid of `${{ expressions }}` directly within `run:` commands are definitely no-brainers.

I should probably clarify my dislike for setting environment variables is strictly a dislike for setting environment variables that persist through the entire _job_, like when writing to `$GITHUB_ENV`. Something I'm on the other hand is very much in favor of is defining environment variables on a _step_ level, since they only persist throughout that particular step.

To me setting environment variables that persist through the entire job vs. defining environment variables on a step level is similar to the distinction between working with global variables vs with local variables. Yes, having access to global variables can be very convenient, but once your code base passes a certain size it can easily become really hairy to keep track of them all.

Of course, working with environment variables (that persist through the entire job) has the potential to become even hairier, since there's a whole bunch of them that will affect ones runtime behavior even if they aren't explicitly referenced.

If we now look at this from a security perspective the term _blast radius_ ought to apply.

Yes, it's very common that documentation examples as well as third party Actions will set environment variables that persist through out a job, and I understand why. Persistent environment variables are familiar, easy to use and they do work fine as long as your workflows are sufficiency simple.

`</rant>`

.github/workflows/test-go.yml
```diff
@@ -68,9 +68,26 @@ jobs:
...
+          INPUTS_GO_ARCH: ${{ inputs.go-arch }}
+          INPUTS_EXTRA_FLAGS: ${{ inputs.extra-flags }}
+          INPUTS_GO_TAGS: ${{ inputs.go-tags }}
+        run: |
+          echo "INPUTS_NAME=$INPUTS_NAME" >> "$GITHUB_OUTPUT"
+          echo "INPUTS_TOTAL_RUNNERS=$INPUTS_TOTAL_RUNNERS" >> "$GITHUB_OUTPUT"
+          echo "INPUTS_BINARY_TESTS=$INPUTS_BINARY_TESTS" >> "$GITHUB_OUTPUT"
+          echo "INPUTS_GO_ARCH=$INPUTS_GO_ARCH" >> "$GITHUB_OUTPUT"
+          echo "INPUTS_EXTRA_FLAGS=$INPUTS_EXTRA_FLAGS" >> "$GITHUB_OUTPUT"
+          echo "INPUTS_GO_TAGS=$INPUTS_GO_TAGS" >> "$GITHUB_OUTPUT"
```
andreaso, line 90: As much as I'm a fan of using GitHub Outputs, I am curious on what specific benefit we are after here, shuffling values from input objects into corresponding output objects?

fatima2003: Using `${{ ... }}` directly in `run` can be exploited by a malicious actor by inputting shell commands. These should be replaced with either env variables since these are parsed safely as strings or as you've suggested as outputs so that they can be parsed safely AND calls to it can be scoped.

References: [Zizmor docs](https://woodruffw.github.io/zizmor/audits/#template-injection), [Ken Muse: GitHub Actions Injection Attacks](https://www.kenmuse.com/blog/github-actions-injection-attacks/).

I've also devised 2 workflows: [a safe and unsafe workflow](https://github.com/fatima2003/GitHub-Workflows-Playground) to test out the vulnerability.

I sent it this as inputs payload: 
`"\"$(echo MALICIOUS PAYLOAD EXECUTED! && echo SECOND && echo THIRD)\""`

Results when workflow uses `$GITHUB_OUTPUTS `or `$GITHUB_ENVIRONMENT`: 
```
> Run echo "doing a thing: $INPUTS_NAME"

doing a thing: "$(echo MALICIOUS PAYLOAD EXECUTED! && echo SECOND && echo THIRD)"
```

Results when workflow uses `${{ inputs. }}` code directly in `run` statements:
```
> Run echo "doing a thing: "$(echo MALICIOUS PAYLOAD EXECUTED! && echo SECOND && echo THIRD)""

doing a thing: MALICIOUS PAYLOAD EXECUTED! SECOND THIRD
```

As seen, in the case where ENV variables are not used, arbitrary code can be executed.

I've got to confirm if this is possible in the exact workflow setup we've got going... but I think to be safe we should use ENV variables

andreaso: I'm obviously not advocating for the use of `${{ expression }}` directly within run statements.

Let me attempt to clarify what I was really meant to ask by highlight that piece of code.

Why do you find the following....

```yaml
- name: Prepare failure summary
  env:
    INPUTS_NAME: ${{ steps.set-variables.outputs.INPUTS_NAME }}
  run: |
    ...
    ...
```

Preferable to the following?

```yaml
- name: Prepare failure summary
  env:
    INPUTS_NAME: ${{ inputs.name }}
  run: |
    ...
    ...
```

In the latter case there's no need to shuffle values from inputs objects to outputs objects.

fatima2003: Aaah yes, I was using the same technique for setting global ENV variables (so I don't have to set them everywhere) which doesn't make sense here 🫠 Thank you, I'll update it

.github/workflows/test-go.yml
```diff
@@ -276,9 +293,9 @@ jobs:
           # | github.com/openbao/openbao/package | TestName | fips | 0 | 2 | [view results](github.com/link-to-logs) |
           jq -r -n 'inputs
           | select(.Action == "fail")
-          | "| ${{inputs.name}} | \(.Package) | \(.Test // "-") | \(.Elapsed) | ${{ matrix.id }} | [view test results :scroll:](${{ env.GH_JOB_URL }}) |"' \
-          failure-summary-${{ matrix.id }}${{ inputs.name != '' && '-' || '' }}${{inputs.name}}.json \
-          >> failure-summary-${{ matrix.id }}${{ inputs.name != '' && '-' || '' }}${{inputs.name}}.md
+          | "| $INPUTS_NAME | \(.Package) | \(.Test // "-") | \(.Elapsed) | ${{ matrix.id }} | [view test results :scroll:](${{ env.GH_JOB_URL }}) |"' \
+          failure-summary-${{ matrix.id }}${{ inputs.name != '' && '-' || '' }}$INPUTS_NAMEjson \
```
cipherboy, line 297:
```suggestion
          failure-summary-${{ matrix.id }}${{ inputs.name != '' && '-' || '' }}$INPUTS_NAME.json \
```

cipherboy (commented):
Few minor nits about shell quoting and then I'll merge!

fatima2003:
test-go (7) failure seems unrelated and exists in https://github.com/openbao/openbao/commit/64ce299a2ecaa96cfd2f04652130c7ef65263247. I'm looking into it :)

cipherboy (approved):
Thank you @fatima2003 and appreciate the assist @andreaso!

fatima2003:
Thank you for the review @cipherboy and @andreaso!

PR #935, opened 2025-01-22
Ensure pki health-check errs on non-pki mounts

cipherboy:
When `bao pki health-check` runs on a non-PKI mount, we see results like:

    ca_validity_period
    ------------------
    status             endpoint                        message
    ------             --------                        -------
    invalid_version    /auth/userpass/users/issuers    This health check requires Vault 1.11+ but an earlier version of Vault Server was contacted, preventing this health check from running.

which isn't correct: we are running on a non-PKI mount, not on an invalid version of Vault or OpenBao.

As reported by @pgnd on Matrix.

### DIFF
--- a/api/sys_mounts.go
+++ b/api/sys_mounts.go
@@ -242,6 +242,39 @@ func (c *Sys) MountConfigWithContext(ctx context.Context, path string) (*MountCo
 	return &result, err
 }
 
+func (c *Sys) MountInfo(path string) (*MountOutput, error) {
+	return c.MountInfoWithContext(context.Background(), path)
+}
+
+func (c *Sys) MountInfoWithContext(ctx context.Context, path string) (*MountOutput, error) {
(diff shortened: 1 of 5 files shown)


api/sys_mounts.go
```diff
@@ -242,6 +242,39 @@ func (c *Sys) MountConfigWithContext(ctx context.Context, path string) (*MountCo
 	return &result, err
 }
 
+func (c *Sys) MountInfo(path string) (*MountOutput, error) {
```
cipherboy, line 245: @JanMa Interestingly, I didn't see an existing function for this. `MountConfig(...)` returns the tune information, not the mount info.

JanMa (changes requested):
LGTM, but the `TestPKIHC_NoMount` test is failing. Please have a look :-)

cipherboy:
@JanMa Ah yes, thanks! Fixed :-)

JanMa (approved):

Issue #739, opened 2024-11-17
Tracking: Reintroduction of UI to Release Builds


cipherboy:
This issue tracks the work necessary to reintroduce the UI to release builds:

- [x] #74, to remove references to HashiCorp assets
    - [x] #643 to remove Vault references on sign-in page
    - [x] #668 to remove Vault references on sidebar, &c
    - [x] #669 to remove references to HashiCorp documentation
    - [x] #683 to fix more documentation links & tests
    - [x] #788 to remove the Vault logo in MFA & OIDC graphics
- [x] #129, to remove references to Vault Enterprise features
    - [x] #684, to remove the license banner
    - [x] #734, to remove the client counts mirroring #194
    - [x] #735 to remove HCP Link status
    - [x] #737, to remove Sentinel Policies from
    - [x] #738, to remove PR/DR replication references
- [x] #154, to fix the main test suite
    - [x] #784, to start fixing several test failures
- [ ] #731, to build with modern versions of NodeJS
- [x] #752, to address outdated dependencies in the UI.
   - [x] #807 addresses this in the current release but will need to be redone before release.
- [ ] (unscoped) to add a shared pipeline stage to build the UI and add it to release artifacts

Optionally:

- [ ] #417, to fix UI builds on Windows

arminfelder:
@cipherboy I fixed a few issues, but there are still around 20 failing 

https://github.com/openbao/openbao/pull/784

cipherboy:
@arminfelder Thank you! I'll try to take a look at the remaining ones shortly :-)

PR #940, opened 2025-01-25
Add UI to release workflow

cipherboy:
Resolves: #739 

---

I think I will setup a new repository, `openbao-nightly`, which mirrors `openbao` and launches nightly release builds... I'll provision fresh keys, fresh access tokens, and fresh repositories (with `-nightly` suffix) which contain updated nightly releases.

Thoughts?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,6 +54,16 @@ jobs:
       - name: go-check
         run: go version
 
+      - name: Node setup
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version-file: "./ui/package.json"
+
+      - name: Yarn setup
+        id: install-yarn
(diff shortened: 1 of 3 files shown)


JanMa (approved):
:tada: 
Regarding nightlies: I would prefer if we could somehow have them running in the main repo and maybe just create a prerelease?

cipherboy:
>  Regarding nightlies: I would prefer if we could somehow have them running in the main repo and maybe just create a prerelease?

Hmm, if we create true pre-releases, we'll have lots of nightlies, even if we only do it when something changed. That seems... like a lot of pre-releases to clutter the main repo with? That's why I was thinking a separate repo would be ideal.

I think even pre-releases show on our download page, so if we push a lot of releases here, we'd have to either prune nightlies out (and may still need to implement paging) and/or hit rate limiting on requests to view all releases... It becomes more data to transfer and slows things down.

My 2c.

PR #943, opened 2025-01-26
Add nightly release workflow

cipherboy:
This adds nightly release workflows for testing.

Repositories are with the `-nightly` suffix:

 - https://github.com/openbao/openbao-nightly
 - https://hub.docker.com/repository/docker/openbao/openbao-ubi-nightly/general
 - https://hub.docker.com/repository/docker/openbao/openbao-nightly/general
 - https://quay.io/repository/openbao/openbao-nightly
 - https://quay.io/repository/openbao/openbao-ubi-nightly

By pushing them to a separate repository, we do not clutter the downloads page (https://openbao.org/downloads/) with non-necessary releases (as this fetches releases via the public API, potentially incurring rate limiting). 

All secrets used are unique: a fresh GPG key, fresh cosign keys, and fresh Dockerhub and Quay keys. The Quay token is scoped just to nightly repositories, though the Dockerhub token doesn't seem to have fine-grained permission control. 

<details>
<summary> Nightly GPG key </summary>

```
-----BEGIN PGP PUBLIC KEY BLOCK-----

mQGNBGeVu+gBDADPVCBBpIbySpoL5mhUR82e8xqbV/GXumxK4XTmGLCOALAzj9fO
g2jGzvEa9TfAyRYaYL3amKX3PV64GPQKDc6cQg5xv175vds109KiYLYVDphBQL7r
H0wW/io0L+Cdvstffr//bVJMR3fLU4ykgGPnIQAHvSv95LQAToxKm1Iic/hKa1JH
7GgVdCtZA/rjMw3/DAOOy19nZto3jIrkksJ7I2JYiUKguypy+NalRdVtKr6GxSRE
xKJSC1Pd3L2Pu4z9E4m91jtghR/e18HplXBY4KLJvLlQH4BmP55Cr36r7KMKeGDz
diPUVoVTVKTdpNZD9+VxzsblOgCL5RoRutEFOLsta0PeO/93vSQTdxPmhgEQI2bh
W2OsAlq7xDrijAY0J+a2MkVy3tis9iNXrYt6tTASdZPSJWWkOpfbru8wVRHTHJ0k
oo9Oi/latLYP06il6ZISLlbB4k7nSjvrl/0jRZZZHFjBOa1CB9brzXTV83Bcw491
YABxJt1cMYnFcgsAEQEAAbQsT3BlbkJhbyAoTmlnaHRseSkgPG9wZW5iYW9AbGlz
dHMubGZlZGdlLm9yZz6JAdcEEwEIAEEWIQSDYYskQx88u0CDu7o0bC+8CEnyZwUC
Z5W76AIbAwUJBaOagAULCQgHAgIiAgYVCgkICwIEFgIDAQIeBwIXgAAKCRA0bC+8
CEnyZ5VSC/0SxIoK3paDzji7DBeW17u4CQ0IdjKX1NSQpr+q0YxI5HXnHDkcRfml
1A6j9OgaeSVf4pnbr8L4eOVZiq4O6Wpv6ba4cYp3ZbKny6lY8rgKcSAvLeK/m9oT
bKAlK7zd6dzAS/zx4RGJWrOTHQgna+PyjgT3NpG87KGUiJEnCgowUcrXYoo4oTvl
svT3wY97tzdluFoMSoO2qhgHoYulMMSyU/OMDEiVtlK03g6TZtrkQ7hHURtdhmBj
cPgGkJljOwOvYrSmaNyFigLgA+HPql974ESUDcctfxs5NCNgJojQ7/xpCk7++lof
vAxRIuH+IeHMa7xSZuxaK462suOz9Rsf5h+7OJs68+GQniDV9tiMaQ8zW5qqzZOb
PpC2Vq4pgqd7zWxSjLqW19KqoFGS8Gp4a0BifVVpG/yTLTM1EyFD3NFVbbkBYq9T
7hcFLAeFei3oVS1h2IbLuPfkp7e3Jf2oYIqv5XvLScyBR2npGYfeCsjq3xCAiNEk
GTamGWbkpfU=
=NrPR
-----END PGP PUBLIC KEY BLOCK-----
```

</details>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,11 @@ on:
         required: false
         default: true
         type: boolean
+      nightly:
+        description: "Mark as nightly build"
+        required: true
+        default: true
+        type: boolean
 
 permissions:
   contents: write
@@ -43,10 +48,45 @@ jobs:
           - openbsd
           - windows
     steps:
+      - name: Disable release on nightly repository
+        if: ${{ !inputs.nightly }}
+        run: |
+          if [ "$GITHUB_REPOSITORY" == "openbao/openbao-nightly" ]; then
+            echo "Refusing to run non-nightly release on nightly repo" 1>&2
+            exit 1
+          fi
+
       - name: Checkout
         uses: actions/checkout@v4.1.7
         with:
           fetch-depth: 0 # Required by GoRelease
+
+      - name: Disable nightly on primary repository
+        if: inputs.nightly
+        env:
+          PRERELEASE: ${{ inputs.prerelease }}
+        run: |
+          if [ "$GITHUB_REPOSITORY" != "openbao/openbao-nightly" ]; then
+            echo "Refusing to run nightly release on non-nightly repo" 1>&2
+            exit 1
+          fi
+
+          # Locally remove previous nightly tags so they do not interfere
+          # with nightly tag computation. Also add the upstream.
+          git remote add upstream https://github.com/openbao/openbao
+          git fetch --all
+
+          for tag in $(git tag); do
+            if [[ "$tag" == *nightly* ]]; then
+              echo "Deleting $tag"
+              git tag --delete "$tag"
+            fi
+          done
+
+          # Then create a nightly tag for this commit. Because this uses the
+          # time of commit, it is stable across different runners.
+          nightly_tag="$(bash ./scripts/nightly_tag.sh)"
+          git tag "$nightly_tag"
 
       - name: Golang Setup
         uses: ./.github/actions/set-up-go
@@ -98,7 +138,7 @@ jobs:
           key: ${{ github.ref }}
 
       - name: "Docker Login: ghcr.io"
-        if: startsWith(github.ref, 'refs/tags/')
+        if: startsWith(github.ref, 'refs/tags/') || inputs.nightly
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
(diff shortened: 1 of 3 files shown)


cipherboy:
@JanMa Build is available here: https://github.com/openbao/openbao-nightly/releases/tag/v9999.9999.9999-n20250126 

Let me know your thoughts, and if you want to merge it into the main repo's releases or not. So far this is manually executed, though we could change that.

JanMa (changes requested):
As discussed offline, I am not a fan of the `v9999` tag and constantly creating new releases in the `openbao-nightly` repo. I'd prefer if we could actually create a valid tag representing the current commit using git describe like this:

```
git describe --tags --exclude "api/*" --exclude "sdk/*" --abbrev=0
```

This way, if someone installs a nightly manually and we release a new version, they can just install the release package and OpenBao will be properly updated.

I also think that creating a new release every night for the nightly build does not provide much value. I don't see why people would want to download a nightly from two or more days ago. Instead I'd like it if we could just overwrite the existing nightly release and keep that history clean

cipherboy:
@JanMa Hmm... Do you have a more concrete tagging structure? The command as posted returns:

```
$ git describe --tags --exclude "api/*" --exclude "sdk/*" --abbrev=0
v2.1.0-beta20241114.3
```

on a fresh checkout because `v2.1.1` and `v2.1.0` are on the `release/2.1.x` branch, not on the `main` branch.

This means our nightly won't be picked up over the latest release, which violates another property we'd prefer to have (nightly beats latest release, but if the nightly isn't updated since installation and a new full release is cut, we should prefer that one).

We could use like so:

```
$ git describe --tags --exclude "api/*" --exclude "sdk/*"
v2.1.0-beta20241114.3-147-g0b1117c166
```

But this doesn't help with the issue of versioning...

Maybe we'd need something a lot more complex...

cipherboy:
@JanMa Take a look at this: https://github.com/openbao/openbao-nightly/releases --- I think that should be about what we expect?

JanMa (approved):

cipherboy:
@JanMa said:

> In hindsight I think this was a mistake.

Hmm, you're going to have the problem anyways when you go to do point releases anyhow. In our case, we bump the minor version for this nightly, but we'd have the same problem for patch release nightlies if we want to do that in the future.

What this means is that, to tag on `main`, you have to wait to delay every single change after beta until GA is cut. That means any breaking changes you want to make after the branch is cut, needs to wait. That's probably 3-4 weeks, during which they might conflict (e.g., `go.mod` changes most commonly). 

I think that's a fine trade-off.

Tbh, I'm not sure we need this, I think we can perhaps just use the main branch and bump the minor version and be fine (since beta is still the same minor version as the later GA).

cipherboy:
Let's revise tagging latter :-)

PR #945, opened 2025-01-26
Add PKCS#11 auto-unseal builds to release workflow

cipherboy:
This creates release tooling for HSM binaries. 

---

To be merged after #943.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,6 +41,7 @@ jobs:
       matrix:
         release_os:
           - linux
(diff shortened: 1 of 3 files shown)


cipherboy:
We have a changelog entry from #889.

JanMa (approved):
While this seems to work, your goreleaser template is becoming more and more unreadable :thinking: Would you mind opening an issue to figure out how we can improve this and enable us to use the same commands for local and release builds?

PR #956, opened 2025-01-28
Add transactions to transit functions

suprjinx:
Adds transactional storage to `transit` functions with read+write operations

Relates #607

### DIFF
--- a/builtin/logical/transit/path_byok.go
+++ b/builtin/logical/transit/path_byok.go
@@ -60,6 +60,13 @@ func (b *backend) pathBYOKExportKeys() *framework.Path {
 }
 
 func (b *backend) pathPolicyBYOKExportRead(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
+
+	txRollback, err := logical.StartTxStorage(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	defer txRollback()
+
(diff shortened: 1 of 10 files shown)


cipherboy (changes requested):
One set of changes off the top of my head: it is terrible UX but https://openbao.org/api-docs/secret/transit/#encrypt-data mentions that if a CREATE operation is used and the key does not exist, it will be created with default values (AES256-GCM96 I believe).

This means we'll want write transactions for encrypt calls as well.

suprjinx:
> One set of changes off the top of my head: it is terrible UX but https://openbao.org/api-docs/secret/transit/#encrypt-data mentions that if a CREATE operation is used and the key does not exist, it will be created with default values (AES256-GCM96 I believe).
> 
> This means we'll want write transactions for encrypt calls as well

Correct me if I'm wrong, but from a little spelunking it appears that transit is not really using `Storage` but rather the cache -- which has a lot of mutex locking around keys already -- eg: https://github.com/openbao/openbao/blob/main/builtin/logical/transit/backend.go#L143-L174

suprjinx:
@cipherboy thanks for the help! I think https://github.com/openbao/openbao/pull/956/commits/a8148eeca5c92e55f23c0528e583ef549dbb3a18 addresses this

cipherboy (approved):
Thank you @suprjinx!

PR #969, opened 2025-02-05
disable openbao-k8s and CSI driver docs

JanMa:
The projects are still unfinished and having the docs visible mostly
leads to confusion-


website/content/docs/agent-and-proxy/agent/caching/persistent-caches/kubernetes.mdx
```diff
@@ -15,7 +15,7 @@ and leases between initialization and sidecar OpenBao Agent containers. This cac
 using a memory volume between the OpenBao Agent containers.
 
 If the OpenBao Agent Injector for Kubernetes is being used, the persistent cache is automatically configured
-and used if the annotation [`vault.hashicorp.com/agent-cache-enable: true`](/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-cache-enable) is set.
```
cipherboy, line 18: If we're going to use the HashiCorp version, should we link to their docs?

cipherboy (approved):
One non-blocking comment.

PR #970, opened 2025-02-06
Docs: Add a simple OIDC keycloak setup example

OdyX:
As I had to retro-discover this setup, let's document it for everyone's sake. It's clearly not complete, but a good starting point I think.


DrDaveD (changes requested):
Thanks for the PR.  Since there's only one item in the setup, we don't need a list.  So please remove the `1. `.

DrDaveD (approved):

PR #992, opened 2025-02-11
Add transactions to AppRole funcs

suprjinx:
Adds transactional storage to the AppRole endpoint funcs.

Relates #607


suprjinx:
@cipherboy I think this is all we need for AppRole -- but there is some hairy storage business with `path_tidy_user_id.go` which I left alone. What do you think?

cipherboy:
@suprjinx Let's do that in a follow-up. I think we can untangle the tidy a bit and use paginated lists. I wonder if we can grab a read transaction over the list itself and a write transaction for tidying each of the various related entries. Thoughts?

cipherboy (approved):

PR #994, opened 2025-02-12
Set BAO_CONFIG_DIR to a default but let it be overridable.

ekarlso:
Allow the config directory env var to be overridable so that it will work inside of bankvaults deployments.


cipherboy (approved):

JanMa (approved):

cipherboy:
Thank you @ekarlso!

PR #996, opened 2025-02-12
Add API and CLI commands to promote/demote nodes in the Raft cluster

JanMa:
This is a follow up of #741 to allow managing the voter state of a node in the
cluster at runtime.  We provide new CLI commands `bao operator raft promote` and
`demote` to promote a non-voter to a voter and vice versa.  Matching API
endpoints at `sys/storage/raft/promte` and `demote` are also added for
convenience.

I am no sure yet if I should also expose this directly in the UI's Raft Storage
overview.

### DIFF
--- a/command/operator_raft_demote.go
+++ b/command/operator_raft_demote.go
@@ -1,0 +1,108 @@
+// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package command
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/cli"
+	"github.com/posener/complete"
+)
+
+var (
+	_ cli.Command             = (*OperatorRaftDemoteCommand)(nil)
+	_ cli.CommandAutocomplete = (*OperatorRaftDemoteCommand)(nil)
+)
+
+type OperatorRaftDemoteCommand struct {
+	*BaseCommand
+	flagDRToken string
+}
+
+func (c *OperatorRaftDemoteCommand) Synopsis() string {
+	return "Demotes a voter to a permanent non-voter"
+}
+
+func (c *OperatorRaftDemoteCommand) Help() string {
+	helpText := `
+Usage: bao operator raft demote <server_id>
+
+  Demotes voter to a permanent non-voter.
+
+	  $ bao operator raft demote node1
+
+` + c.Flags().Help()
+
+	return strings.TrimSpace(helpText)
+}
+
+func (c *OperatorRaftDemoteCommand) Flags() *FlagSets {
+	set := c.flagSet(FlagSetHTTP | FlagSetOutputFormat)
+
+	f := set.NewFlagSet("Command Options")
+
+	f.StringVar(&StringVar{
+		Name:       "dr-token",
+		Target:     &c.flagDRToken,
+		Default:    "",
+		EnvVar:     "",
+		Completion: complete.PredictAnything,
+		Usage:      "DR operation token used to authorize this request (if a DR secondary node).",
+	})
+
+	return set
+}
+
+func (c *OperatorRaftDemoteCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictAnything
+}
+
+func (c *OperatorRaftDemoteCommand) AutocompleteFlags() complete.Flags {
+	return c.Flags().Completions()
+}
+
+func (c *OperatorRaftDemoteCommand) Run(args []string) int {
+	f := c.Flags()
+
+	if err := f.Parse(args); err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	serverID := ""
+
+	args = f.Args()
+	switch len(args) {
+	case 1:
+		serverID = strings.TrimSpace(args[0])
+	default:
+		c.UI.Error(fmt.Sprintf("Incorrect arguments (expected 1, got %d)", len(args)))
+		return 1
+	}
+
+	if len(serverID) == 0 {
+		c.UI.Error("Server id is required")
+		return 1
+	}
+
+	client, err := c.Client()
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 2
+	}
+
+	_, err = client.Logical().Write("sys/storage/raft/demote", map[string]interface{}{
+		"server_id":          serverID,
+		"dr_operation_token": c.flagDRToken,
+	})
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error promoting server: %s", err))
+		return 2
+	}
+
+	c.UI.Output("Server demoted successfully!")
+
+	return 0
+}
--- a/command/operator_raft_demote.go
+++ b/command/operator_raft_demote.go
@@ -18,7 +18,6 @@ var (
 
 type OperatorRaftDemoteCommand struct {
 	*BaseCommand
-	flagDRToken string
 }
 
 func (c *OperatorRaftDemoteCommand) Synopsis() string {
@@ -39,20 +38,7 @@ Usage: bao operator raft demote <server_id>
 }
 
 func (c *OperatorRaftDemoteCommand) Flags() *FlagSets {
-	set := c.flagSet(FlagSetHTTP | FlagSetOutputFormat)
-
-	f := set.NewFlagSet("Command Options")
-
-	f.StringVar(&StringVar{
-		Name:       "dr-token",
-		Target:     &c.flagDRToken,
-		Default:    "",
-		EnvVar:     "",
-		Completion: complete.PredictAnything,
-		Usage:      "DR operation token used to authorize this request (if a DR secondary node).",
-	})
-
-	return set
+	return c.flagSet(FlagSetHTTP | FlagSetOutputFormat)
 }
 
 func (c *OperatorRaftDemoteCommand) AutocompleteArgs() complete.Predictor {
@@ -94,8 +80,7 @@ func (c *OperatorRaftDemoteCommand) Run(args []string) int {
 	}
 
 	_, err = client.Logical().Write("sys/storage/raft/demote", map[string]interface{}{
-		"server_id":          serverID,
-		"dr_operation_token": c.flagDRToken,
+		"server_id": serverID,
 	})
 	if err != nil {
 		c.UI.Error(fmt.Sprintf("Error promoting server: %s", err))
--- a/physical/raft/raft.go
+++ b/physical/raft/raft.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -1245,7 +1246,29 @@ func (b *RaftBackend) PromotePeer(ctx context.Context, peerID string) error {
 		if b.raft == nil {
 			return errors.New("raft storage is not initialized")
 		}
-		return errors.New("raft storage autopilot must be enabled to promote a peer")
+		peers, err := b.Peers(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get Raft peers: %s", err)
+		}
+
+		found := false
+		addr := ""
+		for _, peer := range peers {
+			if peer.ID == peerID {
+				addr = peer.Address
+				found = true
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("server %s not found in raft configuration", peerID)
+		}
+
+		future := b.raft.AddVoter(raft.ServerID(peerID), raft.ServerAddress(addr), 0, 0)
+		if err := future.Error(); err != nil {
+			return fmt.Errorf("failed to promote non-voter to voter: %s", err)
+		}
+		return nil
 	}
 
 	if b.autopilot == nil {
@@ -1273,7 +1296,37 @@ func (b *RaftBackend) DemotePeer(ctx context.Context, peerID string) error {
 		if b.raft == nil {
 			return errors.New("raft storage is not initialized")
 		}
-		return errors.New("raft storage autopilot must be enabled to promote a peer")
+
+		// refuse to demote current leader to not trigger a leader election
+		// when the leader is demoted. This is not necessary if autopilot is enabled,
+		// as it will handle this case for us and only demote the leader after a
+		// leader election
+		if strings.EqualFold(peerID, b.localID) {
+			return errors.New("refusing to demote current leader")
+		}
+
+		peers, err := b.Peers(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get Raft peers: %s", err)
+		}
+
+		found := false
+		for _, peer := range peers {
+			if peer.ID == peerID {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			return fmt.Errorf("server %s not found in raft configuration", peerID)
+		}
+
+		future := b.raft.DemoteVoter(raft.ServerID(peerID), 0, 0)
+		if err := future.Error(); err != nil {
+			return fmt.Errorf("failed to demote voter to non-voter: %s", err)
+		}
+		return nil
 	}
 
 	if b.autopilot == nil {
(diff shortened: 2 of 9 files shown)


command/operator_raft_demote.go
```diff
@@ -0,0 +1,108 @@
...
+
+	f.StringVar(&StringVar{
+		Name:       "dr-token",
+		Target:     &c.flagDRToken,
+		Default:    "",
+		EnvVar:     "",
+		Completion: complete.PredictAnything,
+		Usage:      "DR operation token used to authorize this request (if a DR secondary node).",
+	})
+
```
cipherboy, line 54:
```suggestion
```

We don't have DR support and likely never will. This won't be compatible with Vault Enterprise either.

command/operator_raft_demote.go
```diff
@@ -0,0 +1,108 @@
...
+	client, err := c.Client()
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 2
+	}
+
+	_, err = client.Logical().Write("sys/storage/raft/demote", map[string]interface{}{
+		"server_id":          serverID,
+		"dr_operation_token": c.flagDRToken,
```
cipherboy, line 98:
```suggestion
```

command/operator_raft_demote.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+var (
+	_ cli.Command             = (*OperatorRaftDemoteCommand)(nil)
+	_ cli.CommandAutocomplete = (*OperatorRaftDemoteCommand)(nil)
+)
+
+type OperatorRaftDemoteCommand struct {
+	*BaseCommand
+	flagDRToken string
```
cipherboy, line 21:
```suggestion
```

command/operator_raft_demote.go
cipherboy: And a few other places in the CLI. :-)

physical/raft/raft.go
```diff
@@ -1232,6 +1232,63 @@ func (b *RaftBackend) RemovePeer(ctx context.Context, peerID string) error {
...
+
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
+	if b.disableAutopilot {
+		if b.raft == nil {
+			return errors.New("raft storage is not initialized")
+		}
+		return errors.New("raft storage autopilot must be enabled to promote a peer")
```
cipherboy, line 1248: I thought we also supported manual joining of servers as non-voter nodes, outside of auto-pilot?

In the CLI for `bao operator raft join`:

```go
    f.BoolVar(&BoolVar{
        Name:    "non-voter",
        Target:  &c.flagNonVoter,
        Default: false,
        Usage:   "Join the Raft cluster as non-voter.",
    })
```

Can we not promote/demote these as well?

physical/raft/raft.go
```diff
@@ -1232,6 +1232,63 @@ func (b *RaftBackend) RemovePeer(ctx context.Context, peerID string) error {
...
+	}
+
+	if b.autopilot == nil {
+		return errors.New("raft storage autopilot is not initialized")
+	}
+
+	b.logger.Trace("demoting voter to non-voter", "id", peerID)
+
+	if b.delegate.IsNonVoter(raft.ServerID(peerID)) {
+		return errors.New("server is already a non-voter")
```
cipherboy, line 1286: Just making sure--is this the right check? If we have a temporary non-voter, because it is catching up, would we be unable to permanently demote it until it caught back up and then we could demote it?

What sets `RaftBackend.nonVoter` out of curiosity? Or is that why we only support demoting via auto-pilot?

cipherboy (commented):
> I am no sure yet if I should also expose this directly in the UI's Raft Storage overview.

I think it'd be good to do long-term, though doesn't need to happen today.

command/operator_raft_demote.go
JanMa: TBH I just copy-pasted that boilerplate :-)

physical/raft/raft.go
JanMa: Hmm I think we might be able to do that, but without autopilot we can't really guarantee the promoted/demoted node stays in it's given role. I wonder if there's a valid use-case of running OpenBao without Raft autopilot at all :thinking:

physical/raft/raft.go
JanMa: This checks if the server ID is persisted in our permanent non-voter store, the current role in the Raft cluster. So calling to demote a server which is currently stabilizing will work

physical/raft/raft.go
cipherboy: How is auto-pilot state persisted? I thought we had storage level state that stored voter/non-voter status that we needed to update?

JanMa: We do, but this is managed by the Raft Autopilot Delegate here:
https://github.com/openbao/openbao/blob/main/physical/raft/raft_autopilot_delegate.go#L35
And persisted to the Raft store here:
https://github.com/openbao/openbao/blob/main/physical/raft/raft_autopilot_delegate.go#L309-L324

cipherboy (approved):
Thanks, Jan! I'll slate for backport to v2.2.0 GA .

pree:
Just tested this with @eyenx , In case all voter nodes are down we cannot promote nonvoter nodes, the request always end up in I/O timeout.

cipherboy:
@pree Is this true even with inline authentication? Are there log messages?

eyenx:
Well, we tested it on a single k8s with a 6 node cluster (3 voters, 3 non-voters). They were retry-joining over svc which was always pointing to the "Leader node" and joining worked out:

```
/ $ bao status
Seal Type                static
Recovery Seal Type       shamir
Initialized              true
Total Recovery Shares    1
Threshold                1
Version                  2.4.1
Build Date               2025-09-11T20:28:09Z
Storage Type             raft
Cluster Name             vault-cluster-524bf3c2
Cluster ID               c98e4fb5-1146-5d9b-e70d-4abee68baa1a
HA Enabled               true
HA Cluster               https://a-openbao-2.a-openbao-internal:8201
HA Mode                  standby
Active Node Address      http://10.244.0.27:8200
Raft Committed Index     52
Raft Applied Index       52
/ $ bao operator raft list-peers
Node                                    Address                                State       Voter
----                                    -------                                -----       -----
5119cbd5-fde0-c92a-bd83-f7ad528e96f9    a-openbao-0.a-openbao-internal:8201    follower    true
1b396e3a-eef3-2e0b-5ec8-5dceb5382675    a-openbao-2.a-openbao-internal:8201    leader      true
e778a7e7-e264-30a4-e468-f30ab07764a0    a-openbao-1.a-openbao-internal:8201    follower    true
f9c24c89-1eec-20b7-385f-c71811bbef15    b-openbao-1.b-openbao-internal:8201    follower    false
a02b6493-7bce-5a17-08c8-76f3302f953e    b-openbao-2.b-openbao-internal:8201    follower    false
e9d2b03b-c332-840f-07c9-ea77b787447e    b-openbao-0.b-openbao-internal:8201    follower    false
/ $

# k get po
NAME          READY   STATUS    RESTARTS   AGE
a-openbao-0   1/1     Running   0          3m14s
a-openbao-1   1/1     Running   0          3m14s
a-openbao-2   1/1     Running   0          3m14s
b-openbao-0   1/1     Running   0          2m59s
b-openbao-1   1/1     Running   0          2m59s
b-openbao-2   1/1     Running   0          2m58s
```

We scaled down cluster A (with the voter nodes)

```
# k scale --replicas=0 sts/a-openbao
statefulset.apps/a-openbao scaled
#  kgp
NAME          READY   STATUS        RESTARTS   AGE
a-openbao-0   1/1     Terminating   0          3m58s
a-openbao-1   1/1     Terminating   0          3m58s
a-openbao-2   1/1     Terminating   0          3m58s
b-openbao-0   1/1     Running       0          3m43s
b-openbao-1   1/1     Running       0          3m43s
b-openbao-2   1/1     Running       0          3m42s
# kgp
NAME          READY   STATUS    RESTARTS   AGE
b-openbao-0   1/1     Running   0          3m53s
b-openbao-1   1/1     Running   0          3m53s
b-openbao-2   1/1     Running   0          3m52s
```

Now connecting to the service the `bao status` stil replies with the last leader address, and any `operator` command will be redirected to that leader, failing as it is down. Even a `bao login` won't work

```
/ $ bao status
Seal Type                static
Recovery Seal Type       shamir
Initialized              true
Total Recovery Shares    1
Threshold                1
Version                  2.4.1
Build Date               2025-09-11T20:28:09Z
Storage Type             raft
Cluster Name             vault-cluster-524bf3c2
Cluster ID               c98e4fb5-1146-5d9b-e70d-4abee68baa1a
HA Enabled               true
HA Cluster               https://a-openbao-2.a-openbao-internal:8201
HA Mode                  standby
Active Node Address      http://10.244.0.27:8200
Raft Committed Index     53
Raft Applied Index       52
/ $ bao operator raft list-peers
Error reading the raft cluster configuration: Get "http://10.244.0.27:8200/v1/sys/storage/raft/configuration": dial tcp 10.244.0.27:8200: i/o timeout
```

Logs are showing failing connection to the leader 
```
2025-10-08T17:03:34.148Z [WARN]  storage.raft: not part of stable configuration, aborting election
2025-10-08T17:04:51.428Z [ERROR] core: error during forwarded RPC request: error="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup a-openbao-2.a-openbao-internal on 10.96.0.10:53: no such host\""
2025-10-08T17:04:51.429Z [ERROR] core: forward request error: error="error during forwarding RPC request"
2025-10-08T17:05:04.637Z [ERROR] core: error during forwarded RPC request: error="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup a-openbao-2.a-openbao-internal on 10.96.0.10:53: no such host\""
2025-10-08T17:05:04.637Z [ERROR] core: forward request error: error="error during forwarding RPC request"
```

No such host is normal, as the pod is not existing anymore, but would be similar when a VM would be down, as then the connection to the leader won't work.

I'm reading through your inline authentication PR and will try it next.

eyenx:
I see the link in the [Documentation](https://openbao.org/docs/rfcs/inline-auth/) is wrong (goes to https://github.com/cipherboy/openbao/pull/new/inline-authentication) I think your PR is https://github.com/openbao/openbao/pull/1433

Correct me if I'm wrong

eyenx:
I was able to make the non-voter cluster or at least one of the nodes work by doing the following:

Under `/openbao/data/raft/` I created a `peers.json` with the following content:
```json
[
  {
    "id": "a02b6493-7bce-5a17-08c8-76f3302f953e",
    "address": "b-openbao-2.b-openbao-internal:8201",
    "non_voter": false
  }
]


```
This was then read on restart of the Pod:

```
2025-10-08T19:55:03.749Z [INFO]  storage.raft: raft recovery initiated: recovery_file=peers.json
2025-10-08T19:55:03.750Z [INFO]  storage.raft: raft recovery found new config: config="{[{Voter a02b6493-7bce-5a17-08c8-76f3302f953e b-openbao-2.b-openbao-internal:8201}]}"
2025-10-08T19:55:03.751Z [INFO]  storage.raft: snapshot restore progress: id=bolt-snapshot last-index=52 last-term=5 size-in-bytes=0 read-bytes=0 percent-complete="NaN%"
2025-10-08T19:55:03.762Z [INFO]  storage.raft: raft recovery deleted peers.json
2025-10-08T19:55:03.762Z [INFO]  storage.raft: creating Raft: config="&raft.Config{ProtocolVersion:3, HeartbeatTimeout:15000000000, ElectionTimeout:15000000000, CommitTimeout:50000000, MaxAppendEntries:64, BatchApplyCh:true, ShutdownOnRemove:true, TrailingLogs:0x2800, SnapshotInterval:120000000000, SnapshotThreshold:0x2000, LeaderLeaseTimeout:2500000000, LocalID:\"a02b6493-7bce-5a17-08c8-76f3302f953e\", NotifyCh:(chan<- bool)(0xc000b48000), LogOutput:io.Writer(nil), LogLevel:\"DEBUG\", Logger:(*hclog.interceptLogger)(0xc000a260f0), NoSnapshotRestoreOnStart:true, PreVoteDisabled:false, NoLegacyTelemetry:false, skipStartup:false}"
2025-10-08T19:55:03.764Z [INFO]  storage.raft: initial configuration: index=1 servers="[{Suffrage:Voter ID:a02b6493-7bce-5a17-08c8-76f3302f953e Address:b-openbao-2.b-openbao-internal:8201}]"
2025-10-08T19:55:03.764Z [INFO]  storage.raft: entering follower state: follower="Node at b-openbao-2.b-openbao-internal:8201 [Follower]" leader-address= leader-id=
2025-10-08T19:55:03.764Z [INFO]  core: entering standby mode
2025-10-08T19:55:03.765Z [WARN]  storage.raft: heartbeat timeout reached, starting election: last-leader-addr= last-leader-id=
2025-10-08T19:55:03.765Z [INFO]  storage.raft: entering candidate state: node="Node at b-openbao-2.b-openbao-internal:8201 [Candidate]" term=6
2025-10-08T19:55:03.765Z [INFO]  storage.raft: pre-vote successful, starting election: term=6 tally=1 refused=0 votesNeeded=1
2025-10-08T19:55:03.764Z [INFO]  core: vault is unsealed
2025-10-08T19:55:03.774Z [INFO]  storage.raft: election won: term=6 tally=1
2025-10-08T19:55:03.774Z [INFO]  storage.raft: entering leader state: leader="Node at b-openbao-2.b-openbao-internal:8201 [Leader]"
```

After that I was able to login into the node (with the old root token) and ask for the peers;

```
# bao login
Token (will be hidden):
Success! You are now authenticated. The token information displayed below is
already stored in the token helper. You do NOT need to run "bao login" again.
Future OpenBao requests will automatically use this token.

Key                  Value
---                  -----
token                s.NOTAROOTTOKEN
token_accessor       xyz
token_duration       ∞
token_renewable      false
token_policies       ["root"]
identity_policies    []
policies             ["root"]

# bao operator raft list-peers
Node                                    Address                                State     Voter
----                                    -------                                -----     -----
a02b6493-7bce-5a17-08c8-76f3302f953e   b-openbao-2.b-openbao-internal:8201    leader    true
```

I could join the other nodes by deleting their data and restart:

```
Node                                    Address                                State       Voter
----                                    -------                                -----       -----
a02b6493-7bce-5a17-08c8-76f3302f953e    b-openbao-2.b-openbao-internal:8201    leader      true
45206e90-6cd8-4e9d-c73c-cd58b8147a24    b-openbao-0.b-openbao-internal:8201    follower    false
417c0ee0-95e7-9d61-c6d6-1b388e2a5e3f    b-openbao-1.b-openbao-internal:8201    follower    false
```

eyenx:
I think we should put that in our documentation I could create a quick PR for this

cipherboy:
@eyenx @pree The correct (live) documentation link is https://openbao.org/docs/auth/#inline-authentication 

There's no way to do this with the CLI currently, but you'd do something like:

```
$ curl \
  --header 'X-Vault-Inline-Auth-Path: /auth/userpass/login/admin' \
  --header "X-Vault-Inline-Auth-Parameter-Password: $(echo '{"key":"password","value":"supersecretadmin123456"}' | basenc --base64url)" \
  -X POST \
  --data '{"server_id":"..."}' \
  "$BAO_ADDR/v1/sys/storage/raft/promote"
```

The thing about `bao login` _in general_ (if using auth primitives and not an already issued token) is that it incurs several write operations. This makes it impossible to actually work on leaderless clusters.

---

However, if you're doing this with a token already set, `bao login` in the environment is completely useless and just reads information about the token from storage.

Looking for a better root cause, I think the issue is that these paths weren't added to `http/handler.go` as `handleLogicalNoForward(core)`, so secondaries will _always_ attempt to forward them.

JanMa:
> Just tested this with @eyenx , In case all voter nodes are down we cannot promote nonvoter nodes, the request always end up in I/O timeout.

Yes this is to be expected. Before you shutdown the current leader you need to promote the follower nodes in cluster B.
Otherwise they will not participate in the upcoming leader election. In general the promotion/demotion of nodes will only work if you have an active leader. If you manage to break the state of the Raft cluster, you will need to resort to a `peers.json` file as @eyenx did

pree:
> Before you shutdown the current leader

We were testing this to determine the best setup for disaster recovery scenarios, where you loose for example a whole data center or region :)

PR #997, opened 2025-02-12
Add transaction to rabbitmq module

suprjinx:
Adds transaction wrapper in one rabbitmq method that appears to need

Relates #607


cipherboy (approved):

cipherboy:
This really is a small plugin :-D

PR #1000, opened 2025-02-13
Update to Go 1.24.0

cipherboy:
This has one change not identified in earlier testing, that `sync.Map` differs under `reflect.DeepEqual`. As far as I can tell, no other usage of `reflect.DeepEqual(...)` in runtime is affected.

I've filed it as https://github.com/golang/go/issues/71702. 

There were also minor adjustments to PKI due to the newer toolchain semantics and additional logging fixes.


JanMa (approved):
LGTM

cipherboy:
Thanks @JanMa!

PR #896, opened 2025-01-15
Add NamespaceStore to the namespace branch

tabilet:
This is the first step toward the full namespace feature in Openbao, described in  <a href="https://github.com/openbao/openbao/issues/787">RFC - Add Namespace Support to OpenBao #787</a>

Changes:

<ul>
<li><em>vault/namespace_oss.go</em>, which describes the <em>NamespaceStore</em> struct, is renamed to <em>vault/namespace_openbao.go</em>
</li>
<li> Changes in <em>http/util.go</em> to pass namespace string in request context.</li>
<li>API-based testing cases using backend <em>file</em> are placed in <em>command/ns_openbao/file</em>
</li>
<li> More changes are needed to support plugins, mounts and auths in namespace.</li>
</ul>

<p>To run the tests:</p>
<p> $ ./bin/bao server -config=./command/ns_openbao/file.hcl </p>
<p> go to another terminal and run </p>
<p> $ cd command/ns_openbao/file/ </p>
<p> $ go test ./... </p>

Issue #787, opened 2024-12-10
OpenBao Namespaces


genelet:
### Summary

Introduce namespaces within OpenBao to allow strong multi-tenancy and logical separation of disjoint ACL policies, quotas, plugins etc. Many fragments of namespace support already exist and we wish to keep compatibility with the existing underlying storage framework.

### Problem Statement

OpenBao currently does not support a multi-tenant environment, where each team within an organization can work in its own **namespace** with separate ACL policies, auths, engines, identities, plugins etc. In HashiCorp’s Vault Enterprise, namespaces are implemented as an extra URL path or accessed via HTTP request header. According to [the stated limits](https://developer.hashicorp.com/vault/docs/internals/limits#namespace-limits) for Vault Enterprise, and like upstream’s mount table implementation, all namespaces must fit within a single storage entry. Because each namespace incurs an additional two mount entries, namespaces have a lower conceptual limit than mounts in upstream’s implementation (as mounts must also fit within a single entry). This leads to a scalability problem we wish to avoid in our implementation.

By introducing namespaces, we seek to be able to effectively segment and permit different tenants, preventing them from granting themselves permissions in a parent namespace even if they have full permissions within a child namespace. This segmentation could eventually extend all the way to the storage and potentially compute resource level.

### User-facing description

Namespaces are a mechanism by which operators can achieve multi-tenancy and separation of users within a single OpenBao instance. They allow operators to delegate specific permissions to a group of users, without allowing those users to interfere with the permissions of other users or gain access to the broader instance.

In OpenBao, namespaces are the combination of the following pieces of data:

1. An accessor ID to refer to the namespace in API calls;
2. A UUID to canonically refer to namespaces internally;
3. The path the namespace takes when not using the `X-Vault-Namespace` header;
4. An implicit parent namespace; and
5. Optional custom metadata.

A privileged system token can create a namespace with specific authentication methods, secret engines, and ACL or password policies. APIs to interact with namespaces [already exist](https://openbao.org/api-docs/system/namespaces/) in OpenBao but are not currently implemented, returning errors. An end user can authenticate and manage secrets within the namespace using the same API as the current, non-namespace OpenBao either via using an additional, prefixed path (the namespace path), or via the accessor in conjunction with the `X-Vault-Namespace` header. However, using that same authentication token, they’d be unable to interact with APIs or secrets available in parent namespaces without first authenticating to that namespace separately. The default namespace is the root namespace, a parent namespace of all namespaces.

[...]

Issue #1022, opened 2025-02-20
Namespace Creation Behavior Issues


driif:
Description:

I've identified two issues related to namespace creation in OpenBao:

1. Missing Command Response
When creating a namespace, there is no feedback provided to the user:
```bash
$ ./bao namespace create -ns=a/b d
# No response is shown
```

The command should return useful information such as:
- Namespace path
- Creation confirmation
- Any relevant metadata

2. Parent Namespace Validation
Currently, it's possible to create namespaces under non-existing parent namespaces:
```bash
$ ./bao namespace create -ns=a/b d
$ ./bao namespace list
Keys
----
a/b/d
root
```

Expected Behavior:
1. The namespace creation command should return a response with relevant information about the created namespace.
2. The system should validate parent namespace existence before creating child namespaces.

Current Behavior:
1. No response is provided after namespace creation.
2. Namespaces can be created under non-existing parent namespaces.

Impact:
- Lack of feedback makes it difficult to confirm successful namespace creation
- Creating namespaces under non-existing parents could lead to unexpected behavior and management issues
- Both issues may cause confusion for users and potential problems in production environments

satoqz:
> 1. Missing Command Response
>    When creating a namespace, there is no feedback provided to the user:
>
> ```sh
> $ ./bao namespace create -ns=a/b d
> # No response is shown
> ```

The same goes for namespace patching. I've omitted this in #1027 so far, once this issue is tackled the same change that is required for create/set should be included for patch.

voigt:
Just tested against the current namespace branch. It seems like both scenarios are obsolete:

**Scenario 1: Missing Command Response**

Proper error message when we create a NS in a non-existing NS.

```bash
$ bao namespace create -ns=a/b d
Error creating namespace: Error making API request.

Namespace: a/b/
URL: PUT http://127.0.0.1:8200/v1/sys/namespaces/d
Code: 500. Errors:

* requested namespace was not found
```

The Happy-Path works as expected:

```bash
$ bao namespace create a
Key                Value
---                -----
custom_metadata    map[]
id                 G3CKZl
path               a/
uuid               adf0a6d2-218b-cf95-bc6b-6e3c5a61f550

$ bao namespace create -ns a b
Key                Value
---                -----
custom_metadata    map[]
id                 RhiSfO
path               a/b/
uuid               cb852d76-b432-3902-a1e7-fffd74ffb0ca

$ bao namespace create -ns=a/b d
Key                Value
---                -----
[...]

PR #1026, opened 2025-02-21
Polish NamespaceStore implementation

cipherboy:
This refactors the `NamesapceStore` implementation to conform to the RFC in #787 and adds some hints about adding child namespace support eventually. It also addresses some leftover comments from #896.

---

cc: @genelet @driif @klaus-sap @satoqz @phyrog -- I'll take an ACK or two from anyone and then merge to the feature branch.


vault/logical_system_namespaces.go
```diff
@@ -220,17 +228,14 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
 // handleNamespacesPatch handles the "/sys/namespace/<path>" endpoints to update a namespace
 func (b *SystemBackend) handleNamespacesPatch() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-		path := data.Get("path").(string)
+		// TODO(ascheel): Implement this in terms of get+set calls.
```
cipherboy, line 231: @klaus-sap You'll need to rebase #1016 on top of this. Patch isn't really a NamespaceStore operation but something that applies based on the API. I think @satoqz and @phyrog are on the right path in their comments there.

vault/logical_system_namespaces.go
```diff
@@ -240,9 +245,21 @@ func (b *SystemBackend) handleNamespacesDelete() framework.OperationFunc {
...
-		if err := b.Core.namespaceStore.DeleteNamespace(ctx, path); err != nil {
+		ns, err := b.Core.namespaceStore.GetNamespaceByPath(ctx, path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load namespace: %w", err)
+		}
+
+		if ns == nil {
+			resp := &logical.Response{}
+			resp.AddWarning("requested namespace does not exist")
+			return resp, nil
```
cipherboy, line 256: @driif I ended up taking care of the warning while I was here doing cleanup if that's alright. :-)

vault/namespaces_oss.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/openbao/openbao/helper/namespace"
+)
+
+func (c *Core) NamespaceByID(ctx context.Context, nsID string) (*namespace.Namespace, error) {
+	return namespaceByID(ctx, nsID, c)
+}
+
+func (c *Core) ListNamespaces(includePath bool) []*namespace.Namespace {
+	return []*namespace.Namespace{namespace.RootNamespace}
```
klaus-sap, line 17: Could we fix this by calling ListNamespaces from namespaces_store.go?

vault/logical_system_namespaces.go
satoqz: I'll likely take over this task and create a new PR (As discussed with Klaus)

vault/namespaces_oss.go
satoqz: I'm curious if we can't simplify the `vault/namespaces.go` and `vault/namespaces_oss.go` part - couldn't it just be a single file with just `ListNamespaces()` and `NamespaceById`? I'm likely missing context here.

vault/logical_system_namespaces.go
driif: @cipherboy Thank you, i will remove this part from my code than.

vault/namespaces_store.go
```diff
@@ -5,70 +5,167 @@ package vault
...
 
-// NamespaceEntry is used to store a namespace by name
+// NamespaceEntry is used to store a namespace. We wrap namespace.Namespace
+// in case there is additional data we wish to store that isn't relevant to
+// a namespace instance.
 type NamespaceEntry struct {
-	namespace.Namespace
+	UUID      string               `json:"uuid"`
+	Namespace *namespace.Namespace `json:"namespace"`
+}
```
voigt, line 63: According to the RFC we also want to have a field `parent`:

> `parent`, the uuid of the parent namespace of this namespace and not returned from the response API.

I guess we would add this one here?

```suggestion
	UUID      string               `json:"uuid"`
	Namespace *namespace.Namespace `json:"namespace"`
	Parent      string
}
```

cipherboy: Yeah, I think @driif had filed #1022 which has: 

> 2. Parent Namespace Validation
>     Currently, it's possible to create namespaces under non-existing parent namespaces:

and in general the first iteration didn't really support hierarchical namespaces properly. I think we should probably split that point out as a separate ticket and add `parent` as an explicit separate PR.

vault/namespaces_oss.go
cipherboy: @satoqz The context that I think you're missing is @genelet had moved this into a new file with the OpenBao copyright header, so I had to revert the changes back to the original HashiCorp header in the original file.

---

That said, I agree with you in that I don't see value in keeping both. Personally I'd probably remove `vault/namespaces.go` and keep the `namespaces_oss.go` form -- the latter are methods on the `vault.Core` whereas the former are called with `Core` as an explicit parameter. Since the latter are exported, I'm not sure I see the value in keeping the former as well. 

That said, there are ~50 uses of the `namespaces.go` version and relatively few of the `namespaces_oss.go` version which is rather weird.

cipherboy: @klaus-sap Yeah, I think we need a new helper that returns fat namespaces and not just thin ids.

vault/namespaces_store.go
```diff
@@ -91,141 +187,298 @@ func (c *Core) teardownNamespaceStore() error {
...
+		entry.UUID, err = uuid.GenerateUUID()
+		if err != nil {
+			return err
+		}
+	} else {
+		// Ensure we have no conflicts for paths.
+		for idx, existing := range ns.namespaces {
+			if existing.UUID == entry.UUID {
+				index = idx
+				continue
```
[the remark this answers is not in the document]
cipherboy: Ahh, good eyes. Fixed locally.

vault/logical_system_namespaces.go
```diff
@@ -208,8 +204,20 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
...
+		}
+
+		ns.Namespace.Path = path
+		ns.Namespace.CustomMetadata = metadata
+
 		// Update the namespace
-		if err := b.Core.namespaceStore.SetNamespace(ctx, path, metadata); err != nil {
+		if err := b.Core.namespaceStore.SetNamespace(ctx, ns); err != nil {
 			return handleError(err)
 		}
```
satoqz, line 222: Should we have a locking mechanism here? As far as I can tell another write could be lost in between those calls, even if very unlikely.

cipherboy: @satoqz What do you think of a `ModifyNamespace(...)` pattern; added in the latest push.

satoqz: Great solution, thanks!

vault/namespaces_store.go
cipherboy: Ahh, I think I remember (keep me honest @genelet / @klaus-sap) -- we had originally thought of a flat namespace store, strictly under `/core/namespaces/<uuid>`. Child namespaces wouldn't be under `/namespace/<parent-uuid>/core/namespaces/<child-uuid>` but instead all part of the global root store.

Curious what we feel about either design approach? Parent would be necessary in the former case, but the latter would keep all of a parent namespace's data in one place.

cipherboy:
Test 2 and all of the race test failures [were already present](https://github.com/openbao/openbao/pull/896/checks) on the `namespaces` branch; it is due to request handling. @driif and I were going to take a look at that in a bit, but I think it is net-new work and so should be a separate PR.

At any rate, I think this is ready for re-review @genelet @driif @klaus-sap @satoqz @phyrog

satoqz (approved):

vault/namespaces_store.go
phyrog: @cipherboy `/namespace/<parent-uuid>/core/namespaces/<child-uuid>` feels weird to me. If we'd go hierarchical, I think the way it would make the most sense is just using the (already hierarchical) path instead of uuids (`/core/namespaces/<parent-name>/<child-name>`, so the equivalent to `/core/namespaces/<full-namespace-path>`).

Otherwise I'd go with the flat hierarchy with uuids.

klaus-sap: I would prefer a flat hierarchy with uuids.

cipherboy: @phyrog @klaus-sap Note that in the additional work section, we have:
 
https://github.com/openbao/openbao/issues/787

> Per-Namespace Storage Segments

If we place all of this information in (root) Core, we mix between a self-contained namespace and not. Where do you draw the line and why stop at the namespace table? Wouldn't it (by similar logic) make sense to keep the former mount table implementation, which is a single global entry, and not move mount information into each namespace? And leases... And tokens... And ACLs... &c.

With the interactions with lazy-loading and seal/unseal of namespaces (which are closely related), you end up with a root namespaces which has references to namespaces which don't have storage mounted/accessible/may be broken/... and so have to continually skip them, despite them contributing to LIST operations for the above operations (namespaces, token store tidying, lease tidying, &c). 

For scalability, my opinion is that it is better to push, consistently, all of this information into the relevant parent namespace and not clutter root.

Though, I'd probably suggest a single-level hierarchy and say all namespaces, regardless of type, should be rooted at `/namespace/<uuid>` so the path doesn't grow too long (which impacts performance on Psql or File backend). So `/namespaces/<parent-uuid>` and `/namespaces/<child-uuid>` are siblings in the hierarchy. 

That said, the use of UUID aligns with the mount table and `logical` barrier view. I prefer this as we hope to reduce information leakage and paths aren't necessarily guaranteed to be encrypted. So I'd avoid using `path` there, even if we're sure it wouldn't cause injection issues. 

My 2c.

PR #1027, opened 2025-02-21
Add PATCH support to namespaces API

satoqz:
Implements JSON Merge Patch for the Namespaces API.

Replaces #1016.

---

@cipherboy I'll keep this in draft state for now as I can't sign-off DCO yet, for internal reasons. ETA is sometime next week... I don't think much else depends on this PR so I'd argue the delay is acceptable. Nevertheless this is ready for review :)

cc: @genelet @driif @klaus-sap @phyrog @voigt

### DIFF
--- a/vault/logical_system_namespaces.go
+++ b/vault/logical_system_namespaces.go
@@ -5,6 +5,7 @@ package vault
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


vault/logical_system_namespaces.go
```diff
@@ -215,17 +216,47 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
...
+	return metadata, nil
+}
+
 // handleNamespacesPatch handles the "/sys/namespace/<path>" endpoints to update a namespace
 func (b *SystemBackend) handleNamespacesPatch() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-		// TODO(ascheel): Implement this in terms of get+set calls.
+		path := data.Get("path").(string)
+		if err := b.Core.namespaceStore.ModifyNamespaceByPath(ctx, path, func(ctx context.Context, ns *NamespaceEntry) (*NamespaceEntry, error) {
+			current := make(map[string]interface{})
```
cipherboy, line 240: Ahh yes. Nasty that you can't directly go `map[string]string` to `map[string]interface{}` without the for loop.

cipherboy (approved):
Do you want to start adding tests for namespace APIs?

vault/logical_system_namespaces.go
satoqz: Could be solved via modifications in `framework`, but not sure if worth it, and if so what the API there should look like.

satoqz:
> Do you want to start adding tests for namespace APIs?

Would you like to see those in this PR (if so, just for patch, or all?) or another one? (In that case, let's add it to the backlog, but I'd agree that it makes sense for me to pick that up as a followup)

cipherboy:
@satoqz Given you've still gotta get approvals for DCO, I'm happy to have you add them here if you want.

satoqz:
@cipherboy will do the tests in #1041 - since that also includes returning the right data in the responses for create/patch.

satoqz:
@cipherboy we're finally there. We'll need to merge + rebase our PRs in order, starting with this one, #1041 second and #1051 third once ready.

phyrog (approved):
LGTM

voigt (approved):

cipherboy:
@satoqz This looks to be the oldest PR. I just force-pushed the `namespaces` branch; do you mind rebasing? Thanks!

cipherboy (approved):

PR #1028, opened 2025-02-21
Correctly handle request routing

cipherboy:
During [live-coding with Andrii](https://youtu.be/Gm15qRTtiUY), we worked through correctly handling request routing with namespaces from paths or headers. Right now there is no support for child namespaces, but we've solved the issue with using the root namespace and the data race from the concurrent modification issue in http/util.go.

This doesn't yet allow secret engine mounting within a namespace; we still need to enable the `sys/`, `cubbyhole/`, and `identity/` mounts within the namespace's mount table and load them at runtime.

---

cc: @genelet @driif @klaus-sap @phyrog @voigt @satoqz

### DIFF
--- a/http/util.go
+++ b/http/util.go
@@ -74,28 +74,26 @@ var (
 	adjustRequest = func(c *vault.Core, r *http.Request) (*http.Request, int) {
 		u := r.URL
 		p := u.Path[3:]
+
+		nsHeaderPath := namespace.Canonicalize(r.Header.Get(consts.NamespaceHeaderName))
 		for _, api := range restrictedAPIs {
-			if strings.HasSuffix(p, api) {
-				return r, 0
-			}
-		}
-		for _, api := range containsAPIs {
-			if strings.Contains(p, api) {
-				return r, 0
+			if strings.HasSuffix(p, api) && (p != "/"+api || nsHeaderPath != "") {
+				return r, http.StatusBadRequest
 			}
 		}
 
-		id := namespace.RootNamespaceID
-		namespace.RootNamespace.ID = id // this is reset to root namespace id
-		ns := namespace.Canonicalize(r.Header.Get(consts.NamespaceHeaderName))
-		if ns != "" && ns != "/" {
-			id += "/" + ns
+		for _, api := range containsAPIs {
+			if strings.Contains(p, api) && (!strings.HasPrefix(p, "/"+api) && nsHeaderPath != "") {
+				return r, http.StatusBadRequest
+			}
 		}
-		r = r.WithContext(namespace.ContextWithNamespace(r.Context(), &namespace.Namespace{
-			ID:             id,
-			Path:           ns,
-			CustomMetadata: map[string]string{},
-		}))
+
+		// TODO(ascheel): Consider if we can combine header with actual request path here without triggering redirect logic.
+		if nsHeaderPath != "" {
+			r = r.WithContext(namespace.ContextWithNamespace(r.Context(), &namespace.Namespace{
+				Path: nsHeaderPath,
+			}))
+		}
 
 		return r, 0
 	}
--- a/vault/namespaces_store.go
+++ b/vault/namespaces_store.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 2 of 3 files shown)


http/util.go
```diff
@@ -74,28 +74,28 @@ var (
 	adjustRequest = func(c *vault.Core, r *http.Request) (*http.Request, int) {
 		u := r.URL
 		p := u.Path[3:]
+
+		nsHeaderPath := namespace.Canonicalize(r.Header.Get(consts.NamespaceHeaderName))
 		for _, api := range restrictedAPIs {
-			if strings.HasSuffix(p, api) {
-				return r, 0
+			if strings.HasSuffix(p, api) && (p != "/"+api || nsHeaderPath != "") {
+				c.Logger().Error("got request to restricted API with namespace", "p", p, "api", api, "nsPath", nsHeaderPath)
```
satoqz, line 81:
```suggestion
				c.Logger().Error("got request to restricted API with namespace", "path", p, "api", api, "nsPath", nsHeaderPath)
```
Just a nit, just "p" is not that clear reading the logs not knowing the underlying code.

vault/namespaces_store.go
```diff
@@ -554,3 +555,58 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, uuid string) erro
...
+	if entry == nil {
+		return baseCtx, nil, fmt.Errorf("requested namespace was not found")
+	}
+
+	newCtx := namespace.ContextWithNamespace(baseCtx, entry.Namespace)
+	return newCtx, entry.Namespace, nil
+}
+
+func (ns *NamespaceStore) ResolveNamespaceFromRequest(baseCtx context.Context, httpCtx context.Context, reqPath string) (context.Context, *namespace.Namespace, string, error) {
+	// TODO(ascheel): Reconsider which takes precedence. Currently Header->Path is the final order.
```
satoqz, line 581: The way I believe this works upstream is when both header and path are provided, they stack, treating the header as a prefix to the path. So `X-Vault-Token: abc` and a path of `def` would become `abc/def`.

I guess this only makes sense once we handle child namespaces anyway.

satoqz (commented):
Just one concern that might have been discussed before. Given we have a namespace `abc`:

- Path `v1/abc` will resolve the `root` namespace
- Path `v1/abc/` will resolve the `abc` namespace

However:

- `X-Vault-Namespace: abc` will resolve the `abc` namespace
- `X-Vault-Namespace: abc/` will not resolve at all, request handling exits beforehand

phyrog:
> - `X-Vault-Namespace: abc/` will not resolve at all, request handling exits beforehand

This is what the cli sets when using -namespace, so we should definitely handle this.

http/util.go
cipherboy: Ah, that was a debug line, I will remove!

cipherboy:
> Just one concern that might have been discussed before. Given we have a namespace `abc`:
> 
>     * Path `v1/abc` will resolve the `root` namespace
> 
>     * Path `v1/abc/` will resolve the `abc` namespace
> 
> 
> However:
> 
>     * `X-Vault-Namespace: abc` will resolve the `abc` namespace
> 
>     * `X-Vault-Namespace: abc/` will not resolve at all, request handling exits beforehand

Sorry @satoqz -- is this what we should be doing but aren't -- or are currently doing but shouldn't be?

I think the second part (`X-Vault-Namespace: abc` and `X-Vault-Namespace: abc/`) should behave the same on this branch due to the Canonicalize call: https://github.com/openbao/openbao/pull/1028/files#diff-278ba6b34b839340763b2dc9d4278acf4d4b5826d5d41b13dc81d584e5d446bcR78

satoqz:
@cipherboy huh, excuse the confusion - I tested again and you are absolutely right about the second part.

Question remains only for the first case, what I described above is what we are currently doing in this PR. Should the trailing slash be required?

satoqz (approved):

cipherboy:
@satoqz Yeah I think I agree. In particular things like:

https://github.com/openbao/openbao/blob/82008e99a956736abd3f833bc9ec62de895615d7/vault/request_handling.go#L538-L549

Suggest terminal paths should be treated differently from `/` suffixed paths.

cipherboy:
OK, comment should be clarified and log statement removed.

Issue #1004, opened 2025-02-14
goreleaser - fix pre-release container image taging & latest


cipherboy:
See: https://hub.docker.com/r/openbao/openbao-ubi/tags

|  |  | 
|---|---|
| ![Image](https://github.com/user-attachments/assets/75278608-aeee-4b94-9a5e-4cbcd2ef0986) | ![Image](https://github.com/user-attachments/assets/102a8fd7-c690-47fe-a559-640e3e4b73e0) |

Pre-release was set like so:

2025-02-13T21:40:32.4578367Z   GITHUB_RELEASE_PRERELEASE: true
2025-02-13T21:40:32.4580114Z   GITHUB_RELEASE_MAKE_LATEST: false

But the release was marked latest on GitHub and minor versions tagged to registries anyways.

This likely means we need to split out the release workflows even further to make the pre-release process disjoint because the filters aren't working.

PR #1030, opened 2025-02-21
Fix prerelease image tagging

JanMa:
Turns out `.Prerelease` is not a boolean, so `skip_push` is never set to
false on a pre release. To fix this, we evaluate the environment
variable `GITHUB_RELEASE_PRERELEASE` instead.

Closes: #1004


cipherboy (approved):
Thanks @JanMa!

Issue #991, opened 2025-02-11
Simplify goreleaser build process


cipherboy:
per @JanMa on #945:

> While this seems to work, your goreleaser template is becoming more and more unreadable 🤔 Would you mind opening an issue to figure out how we can improve this and enable us to use the same commands for local and release builds?

I wholeheartedly agree. Perhaps we need a build system for goreleaser? Long-term, paid support might also give us some options for reusable components and project variants.

PR #1039, opened 2025-02-24
Simplify Goreleaser templates

JanMa:
This splits goreleaser template in multiple files to get rid of the `sed` magic
in our release workflow and improve readability of the individual files. Also
this allows you to run goreleaser locally without any additional steps.

Resolves: #991


JanMa:
@cipherboy tested the release workflow in the nightly repo and it works as expected now.
Releases are also automatically created as a prelease if their semver tag indicates it. See:
https://github.com/openbao/openbao-nightly/releases/tag/v2.3.0-nightly1740521276

.github/workflows/release.yml
```diff
@@ -95,15 +150,12 @@ jobs:
...
         with:
-          node-version-file: "./ui/package.json"
-
-      - name: Yarn setup
-        id: install-yarn
-        run: |
-          npm install -g yarn
+          path: |
+            http/web_ui/
+          key: ${{ github.ref }}-ui
```
cipherboy, line 158: This will mean for actual releases (e.g., `v2.2.0`), if the UI fails to build and we need to patch it, we'll have to bump the minor version I believe?

This should be fine, but probably warrants a follow-up note in our checklist.

goreleaser.hsm.yaml
```diff
@@ -0,0 +1,282 @@
+version: 2
+
+project_name: bao
```
cipherboy, line 3: This is fine, but we'll have to keep in mind that any update to a section like `docker` or something probably needs to be redone in several places.

cipherboy (approved):
Thanks @JanMa! I do wish there was a way of splitting of common sections a bit better, but eh.

.github/workflows/release.yml
JanMa: If the UI fails to build, the rest of the pipeline shouldn't run

goreleaser.hsm.yaml
JanMa: Yes, this is unfortunately the case.

Issue #1037, opened 2025-02-24
Namespaces: Incorrect list response schema


satoqz:
**Describe the bug**
Currently, on the namespaces branch, the response returned for `LIST`/`GET` namespace calls has the wrong format.

**To Reproduce**

1. Create a namespace: `bao namespace create test`
2. Get the raw API response from the list endpoint
```sh
curl -s -H "X-Vault-Request: true" -H "X-Vault-Token: $(bao print token)" 'http://127.0.0.1:8200/v1/sys/namespaces?list=true' | jq
```

...results in something like:

```json
{
  "keys": [
    "test/"
  ],
  "data": {
    "keys": [
      "test/"
    ]
  },
}
```

(with all generic response fields such as `request_id` removed)

The sample response at https://openbao.org/api-docs/system/namespaces looks different though:

```json
{
  "data": {
    "key_info": {
      "bar/": {
        "custom_metadata": {},
        "id": "HWmNL",
        "path": "bar/"
      },
      "foo/": {
        "custom_metadata": {},
        "id": "5q39x",
        "path": "foo/"
      }
    },
    "keys": [
      "bar/",
      "foo/"
    ]
  }
}
```

**Expected behavior**
The JSON returned by the `LIST`/`GET` endpoint should adhere to the expected schema, as outlined in the sample response.

Related to #1022

@cipherboy can you make this an RFC subissue?

voigt:
Issue seems to be resolved. Tested on current namespace branch:

**Without namespace parameter:**

```
$ bao namespace list
Keys
----
reply/
```

```json
$ curl -s -H "X-Vault-Request: true"  -H "X-Vault-Token: $(bao print token)" 'http://127.0.0.1:8200/v1/sys/namespaces?list=true' | jq
{
  "request_id": "dcc6970a-32a1-f92e-9139-61749fd98490",
  "lease_id": "",
  "renewable": false,
  "lease_duration": 0,
  "data": {
    "key_info": {
      "reply/": {
        "custom_metadata": {},
        "id": "iQtQLh",
        "path": "reply/",
        "uuid": "56e1a1c5-bbd8-fe4a-46c0-a884a182fe0a"
      }
    },
    "keys": [
      "reply/"
    ]
  },
  "wrap_info": null,
  "warnings": null,
  "auth": null
}
```

**With namespace parameter:**

```
$ bao namespace list -ns=reply
Keys
----
comsysto/
ki/
liquid/
```

```json
[...]

PR #1041, opened 2025-02-24
Polish namespaces API

satoqz:
This resolves the following issues related to the namespaces API:

- [x] Part 1 of #1022
- [x] #1037

Additional goals:
- [x] Polishing of the backend definition (schemas, help text, that kind of stuff)
- [x] Initial test coverage of the namespaces API

### DIFF
--- a/vault/logical_system.go
+++ b/vault/logical_system.go
@@ -5788,4 +5788,35 @@ This path responds to the following HTTP methods.
         Returns a list historical version changes sorted by installation time in ascending order.
 		`,
 	},
+
+	"list-namespaces": {
+		"List namespaces.",
+		`
+This path responds to the following HTTP methods.
+
+	GET /
+		Returns a list of all namespaces.
+
+	LIST /
+		Returns a list of all namespaces.
+		`,
+	},
+	"namespaces": {
+		"Create, read, update and delete namespaces.",
+		`
+This path responds to the following HTTP methods.
+
+	GET /<path>
+		Retrieve a namespace.
+
+	PUT /<path>
+		Create or update a namespace.
+
+	PATCH /<path>
+		Update a namespace's custom metadata.
+
+	DELETE /<path>
+		Delete a namespace.
+		`,
+	},
 }
--- a/vault/logical_system.go
+++ b/vault/logical_system.go
@@ -5794,9 +5794,6 @@ This path responds to the following HTTP methods.
 		`
 This path responds to the following HTTP methods.
 
-	GET /
-		Returns a list of all namespaces.
-
 	LIST /
 		Returns a list of all namespaces.
 		`,
--- a/vault/logical_system.go
+++ b/vault/logical_system.go
@@ -5795,7 +5795,10 @@ This path responds to the following HTTP methods.
 This path responds to the following HTTP methods.
 
 	LIST /
-		Returns a list of all namespaces.
+		List namespaces.
+
+	SCAN /
+		Scan (recursively list) namespaces.
 		`,
 	},
 	"namespaces": {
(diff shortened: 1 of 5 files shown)


cipherboy:
@satoqz Do you mind rebasing this on top of the new `upstream/namespaces` branch and signing off? I think something like `git rebase -i --signoff upstream/namespaces` might work well.

satoqz:
@cipherboy that's totally planned, no worries! For now this is rebased on top of #1027, I'd like to get that merged first and then this can be cleanly rebased on just `namespaces`.

vault/logical_system.go
```diff
@@ -5788,4 +5788,35 @@ This path responds to the following HTTP methods.
         Returns a list historical version changes sorted by installation time in ascending order.
 		`,
 	},
+
+	"list-namespaces": {
+		"List namespaces.",
+		`
+This path responds to the following HTTP methods.
+
+	GET /
```
cipherboy, line 5797: I don't think we actually need to respond to this actually; perhaps we should just support `LIST` and `SCAN`.

satoqz: Not familiar with `SCAN`, I see it is specific to the project.. Intuition tells me that would be akin to pagination?

cipherboy: Recursive listing, yes.

cipherboy: Thinking about it, I think @driif might be working on the recursive list as part of delete cascading but I might be mistaken.

driif: @cipherboy
After our previous talks, we've decided not to delete namespaces if they still have child namespaces under them. For this approach, we just need to check if a namespace contains any child namespaces. If it has children, we won't allow it to be deleted.

satoqz:
@cipherboy I'm technically done here, assuming we'll cover `SCAN` at a later date. Commits should be well-separated, so this is probably easiest to review commit-by-commit.

Now on to hoping I can sign-off those commits in near time...

cc @driif @phyrog @voigt would appreciate if someone could have a short look over this!

vault/logical_system_namespaces.go
```diff
@@ -14,154 +15,154 @@ import (
...
 	}
 }
 
-// handleNamespacesList handles /sys/namespaces/ endpoints to provide the enabled namespaces
+// handleNamespacesList handles "/sys/namespaces" endpoint to list the enabled namespaces.
 func (b *SystemBackend) handleNamespacesList() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-		namespaces, err := b.Core.namespaceStore.ListNamespacePaths(ctx, false /* includeRoot */)
+		// TODO(satoqz): Use ListNamespaceEntries once this can be rebased so we can also return the UUID fields.
+		namespaces, err := b.Core.namespaceStore.ListNamespaces(ctx, false /* includeRoot */)
```
phyrog, line 141: `ListNamespaceEntries` as currently implemented by #1047 is not really great here. It currently returns ALL namespaces, not just the ones directly under the parent namespace we want to list. The same problem applies to `ListNamespaces`.

We really should distinguish between functions that are used internally (like `ListNamespaces` and `ListNamespaceEntries` - listing every namespace there is) and the ones used by the namespaces API (filter by direct parent namespace, do not list grandchildren, etc).

satoqz: Right. I think we want to refactor/rethink the namespace store APIs once all the current PRs are merged.

wslabosz-reply (approved):

phyrog:
@cipherboy I rebased this one for @satoqz. Should also be good to merge.

driif (approved):

vault/logical_system_namespaces_test.go
```diff
@@ -0,0 +1,230 @@
...
+			require.NoError(t, mapstructure.Decode(info, &ns), "key_info entry is not a namespace")
+			require.Equal(t, ns.Path, path, "path in key does not match path in namespace struct")
+			require.NotEmpty(t, ns.ID, "namespace ID should not be empty")
+			require.NotEqual(t, ns.ID, namespace.RootNamespaceID, "list should not include root namespace")
+		}
+	})
+}
+
+func TestNamespaceBackend_Scan(t *testing.T) {
+	// TODO(satoqz): Implement scan once a clear API is available from NamespaceStore.
```
cipherboy, line 229: Wouldn't scan just list all? Regardless, I think this can be a follow-up ticket.

cipherboy (approved):
I'll merge this, based on other conversations here.

PR #1042, opened 2025-02-24
fix(ui): repairs missing checkmarks, in checkboxes, due to invalid us…

arminfelder:
…age of sass-svg-uri

Resolves #


JanMa (approved):
LGTM, thanks a lot @arminfelder.
Not sure why the UI test suite didn't run, but all tests pass when I run it locally

Issue #1029, opened 2025-02-21
Resolve Namespace Routing and Initialization Issues for Compatibility


driif:
Description:

To address the issue where the route `v1/ns1/ns2/sys/namespaces/ns3` is not found in OpenBao and to ensure compatibility with expected namespace logic, the following steps are recommended:

1. **Namespace-Aware Routing**
   - **Problem:** The router does not correctly handle namespace prefixes, leading to route mismatches.
   - **Solution:** Adjust the routing logic to resolve namespaces before matching routes.
   - **Implementation:**
     - Parse the request path to extract namespace segments (e.g., `ns1/ns2`).
     - Set the namespace context for the request.
     - Remove namespace prefixes from the path, leaving only the endpoint path (e.g., `sys/namespaces/ns3`).
     - Route the adjusted path to the correct handler.

2. **Mount System Backends in Every Namespace**
   - **Problem:** Critical system backends like `sys/` are not automatically mounted in child namespaces.
   - **Solution:** Ensure system backends are mounted when a new namespace is created.
   - **Implementation:**
     - During namespace creation, initialize a mount table for the namespace.
     - Add default system mounts to the namespace's mount table:

3. **Initialize Identity Stores Per Namespace**
   - **Problem:** Identity stores are not isolated per namespace.
   - **Solution:** Each namespace should have its own identity store.
   - **Implementation:**
     - Include an identity store in the namespace struct.
     - Initialize the identity store during namespace creation.

4. **Key Code Areas to Review**
   - **Router Logic:** Ensure namespaces are parsed correctly in `router.go` (e.g., `parseNamespacePath`).
   - **Namespace Initialization:** Confirm `CreateNamespace` triggers mount and identity setup.
   - **System Backend Mounting:** Verify `sys` is mounted in all namespaces (check `mountSystemBackend` calls).

cipherboy:
Identity is in #1110 so I think we can close this.

PR #1047, opened 2025-02-25
Create required mounts when setting up namespace

cipherboy:
When setting up namespaces, we're required to create additional mounts
for sys/, identity/, cubbyhole/, and token/ within the namespace. Of
these, cubbyhole is never persisted but is instead a dynamic mount,
whereas the remainder exist in the underlying storage.

We don't yet handle the separation of these mounts' data by namespace;
this will be handled in later updates to this branch. Many singleton
stores like TokenStore and PolicyStore will need broader updates to
align with the RFC's data storage model.

This was created during live-coding with @satoqz.

`Signed-off-by: Alexander Scheel <ascheel@gitlab.com>`

----

Resolves: #1029

cc: @genelet @driif @klaus-sap @phyrog @voigt @satoqz

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -59,6 +59,11 @@ var (
 
 // enableCredential is used to enable a new credential backend
 func (c *Core) enableCredential(ctx context.Context, entry *MountEntry) error {
+	// Ensure the token backend is a singleton
+	if entry.Type == mountTypeToken || entry.Type == mountTypeNSToken {
+		return errors.New("token credential backend cannot be instantiated")
+	}
+
 	// Enable credential internally
 	if err := c.enableCredentialInternal(ctx, entry, MountTableUpdateStorage); err != nil {
 		return err
@@ -98,9 +103,6 @@ func (c *Core) enableCredentialInternal(ctx context.Context, entry *MountEntry, 
 	entry.NamespaceID = ns.ID
 	entry.namespace = ns
 
-	// Populate cache
-	c.NamespaceByID(ctx, ns.ID)
-
 	// Basic check for matching names
 	for _, ent := range c.auth.Entries {
 		if ns.ID == ent.NamespaceID {
@@ -113,11 +115,6 @@ func (c *Core) enableCredentialInternal(ctx context.Context, entry *MountEntry, 
 				return logical.CodedError(409, fmt.Sprintf("path is already in use at %s", ent.Path))
 			}
 		}
-	}
-
-	// Ensure the token backend is a singleton
-	if entry.Type == mountTypeToken {
-		return errors.New("token credential backend cannot be instantiated")
 	}
 
 	// Check for conflicts according to the router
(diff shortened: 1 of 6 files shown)


cipherboy:
Note that this failure is expected:

```

2025-02-25T22:18:00.4965510Z ##[group][31mFAIL[0m vault.TestNamespaceStore (0.03s)
2025-02-25T22:18:00.4966136Z     namespaces_store_test.go:133: 
2025-02-25T22:18:00.4967243Z         	Error Trace:	/home/runner/work/openbao/openbao/vault/namespaces_store_test.go:133
2025-02-25T22:18:00.4968138Z         	Error:      	Received unexpected error:
2025-02-25T22:18:00.4968853Z         	            	permission denied
2025-02-25T22:18:00.4969419Z         	Test:       	TestNamespaceStore
```

In particular, what happens is our singleton tokenStore is overrided by the namespace-specific one post-unseal, so we lose our root token. I think @satoqz might be interested in working on this.

vault/auth.go
```diff
@@ -849,6 +891,13 @@ func (c *Core) persistAuth(ctx context.Context, barrier logical.Storage, table *
 					continue
 				}
 
+				ne, err := c.namespaceStore.GetNamespaceByAccessor(ctx, mtEntry.NamespaceID)
```
phyrog, line 894:
```suggestion
				ne, err := c.NamespaceByID(ctx, mtEntry.NamespaceID)
```

vault/auth.go
```diff
@@ -565,47 +562,75 @@ func (c *Core) loadCredentials(ctx context.Context) error {
...
 
-	var needPersist bool
 	if len(globalEntries) == 0 {
+		// TODO(ascheel) Assertion: globalEntries is empty iff there is only
+		// one namespace (the root namespace).
 		c.logger.Info("no auth mounts in transactional auth mount table; adding default auth mount table")
-		c.auth = c.defaultAuthTable()
+		c.auth, err = c.defaultAuthTable(ctx)
+		if err != nil {
+			panic(err.Error())
```
satoqz, line 597:
```suggestion
			return err
```

vault/auth.go
```diff
@@ -652,7 +691,10 @@ func (c *Core) loadLegacyCredentials(ctx context.Context, barrier logical.Storag
 			return false, nil
 		}
 		c.logger.Info("no mounts in legacy auth table; adding default mount table")
-		c.auth = c.defaultAuthTable()
+		c.auth, err = c.defaultAuthTable(ctx)
+		if err != nil {
+			panic(err.Error())
```
satoqz, line 696:
```suggestion
			return false, err
```

vault/mount.go
```diff
@@ -1235,47 +1234,72 @@ func (c *Core) loadMounts(ctx context.Context) error {
...
 
-	var needPersist bool
 	if len(globalEntries) == 0 {
+		// TODO(ascheel) Assertion: globalEntries is empty iff there is only
+		// one namespace (the root namespace).
 		c.logger.Info("no mounts in transactional mount table; adding default mount table")
-		c.mounts = c.defaultMountTable()
+		c.mounts, err = c.defaultMountTable(ctx)
+		if err != nil {
+			panic(err.Error())
```
satoqz, line 1268:
```suggestion
			return err
```

vault/mount.go
```diff
@@ -1323,7 +1361,10 @@ func (c *Core) loadLegacyMounts(ctx context.Context, barrier logical.Storage) (b
 			return false, nil
 		}
 		c.logger.Info("no mounts in legacy mount table; adding default mount table")
-		c.mounts = c.defaultMountTable()
+		c.mounts, err = c.defaultMountTable(ctx)
+		if err != nil {
+			panic(err.Error())
```
satoqz, line 1366:
```suggestion
			return false, err
```

vault/mount.go
```diff
@@ -1373,7 +1414,11 @@ func (c *Core) runMountUpdates(ctx context.Context, barrier logical.Storage, nee
 		needPersist = true
 	}
 
-	for _, requiredMount := range c.requiredMountTable().Entries {
+	requiredMounts, err := c.requiredMountTable(ctx)
+	if err != nil {
+		panic(err.Error())
```
satoqz, line 1419:
```suggestion
		return err
```

vault/mount.go
```diff
@@ -1882,11 +1949,16 @@ func (c *Core) newLogicalBackend(ctx context.Context, entry *MountEntry, sysView
...
-func (c *Core) defaultMountTable() *MountTable {
+func (c *Core) defaultMountTable(ctx context.Context) (*MountTable, error) {
 	table := &MountTable{
 		Type: mountTableType,
 	}
-	table.Entries = append(table.Entries, c.requiredMountTable().Entries...)
+
+	requiredMounts, err := c.requiredMountTable(ctx)
+	if err != nil {
+		panic(err.Error())
```
satoqz, line 1959:
```suggestion
		return nil, err
```

vault/mount.go
```diff
@@ -2042,10 +2135,10 @@ func (c *Core) singletonMountTables() (mounts, auth *MountTable) {
...
 	switch entry.Type {
-	case mountTypeSystem:
+	case mountTypeSystem, mountTypeNSSystem:
 		c.systemBackend = backend.(*SystemBackend)
 		c.systemBarrierView = view
-	case mountTypeCubbyhole:
+	case mountTypeCubbyhole, mountTypeNSCubbyhole:
 		ch := backend.(*CubbyholeBackend)
 		ch.saltUUID = entry.UUID
 		ch.storageView = view
```
satoqz, line 2144: I don't think we intend to modify `c.(systemBackend|cubbyholeBackend|identityStore)` here if it's a non-root-namespace mount.

vault/mount.go
```diff
@@ -1735,7 +1805,7 @@ func (c *Core) setupMounts(ctx context.Context) error {
 			backendType := backend.Type()
 
 			if backendType != logical.TypeLogical {
-				if entry.Type != mountTypeKV && entry.Type != mountTypeSystem && entry.Type != mountTypeCubbyhole {
+				if entry.Type != mountTypeKV && entry.Type != mountTypeSystem && entry.Type != mountTypeCubbyhole && entry.Type != mountTypeNSSystem && entry.Type != mountTypeNSCubbyhole {
```
satoqz, line 1808: Should probably have something like `var knownMounts`.

vault/namespaces_store.go
```diff
@@ -212,14 +220,20 @@ func (ns *NamespaceStore) SetNamespace(ctx context.Context, namespace *Namespace
...
-// NamespaceStore.
+// NamespaceStore. This function unlocks the lock once finished.
 func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, namespace *NamespaceEntry) error {
+	var unlocked bool
+	defer func() {
+		if !unlocked {
+			ns.lock.Unlock()
+			unlocked = true
+		}
+	}()
```
satoqz, line 235: For understanding, why do we need this bit of logic again?

vault/mount.go
satoqz: here too:
https://github.com/openbao/openbao/blob/02410e2ea07cb1616a4d0183be0333db076b56d4/vault/mount.go#L714

satoqz:
Just found another couple of interesting points in the codebase where I assume the upstream namespace implementation would hook into:

https://github.com/openbao/openbao/blob/02410e2ea07cb1616a4d0183be0333db076b56d4/vault/core_util.go#L30-L36

this is akin to our `(*NamespaceEntry).View()`, but we have direct access to core so we can be sure we're always only extending `c.systemBarrierView`.

Furthermore:

https://github.com/openbao/openbao/blob/02410e2ea07cb1616a4d0183be0333db076b56d4/vault/token_store_util.go#L10-L12

`TokenStore` (and IIUC the request auth codepaths using it) already supports the idea of being namespace-aware.

based on this together with all the calls to `c.tokenStore` in the request authentication codepath I'd argue that a single cross-namespace yet namespace-storage-aware `TokenStore` instance is preferable. Right now we're also creating an additional `TokenStore` instance when we create the token mount, which causes issues as mentioned.

The `TokenStore` is deeply dependent on a `CubbyholeBackend`, which we're currently also reinstantiating for each namespace. `CubbyholeBackend` holds a `storageView` field:

https://github.com/openbao/openbao/blob/02410e2ea07cb1616a4d0183be0333db076b56d4/vault/logical_cubbyhole.go#L42-L46

but that is only ever used once, in a test. Otherwise, `CubbyholeBackend` just uses `req.Storage`, which we need to centrally override to point to namespace storage anyways.

TLDR; @cipherboy is my assumption correct that a single cross-namespace `TokenStore` + `CubbyholeBackend` instance is the way to go?

vault/auth.go
cipherboy: @satoqz @phyrog @voigt I am curious about these. I think it is a behavior change if we `err`, no? Panicing is better; the root namespace (that these are called with) will always exist and the fatal behavior should continue?

vault/mount.go
cipherboy: @satoqz Yes, I think this is the bug we were seeing earlier. But I'm not sure which structure for namespaces we'll want. We could either create a cubbyhole as a `map[ns]backend` or just use a singleton instance and have it pull namespace information from `ctx`. Thoughts?

cipherboy: Ah, I see, you talked about this below:

> based on this together with all the calls to `c.tokenStore` in the request authentication codepath I'd argue that a single cross-namespace yet namespace-storage-aware `TokenStore` instance is preferable. Right now we're also creating an additional `TokenStore` instance when we create the token mount, which causes issues as mentioned.

I will update this, then.

vault/auth.go
cipherboy: All of the returns were already annotated but I opted to keep `defaultAuthTable` panicing and instead have the `c.requiredMountTable` helper return the error.

vault/namespaces_store.go
cipherboy: Added a comment!

satoqz (approved):
Looks great, thank you!

Once this is merged I'll create issues for all the follow up tasks, like storage path awareness for several backends and plugin mounting.

vault/mount_test.go
```diff
@@ -734,7 +735,8 @@ func TestCore_Remount_Protected(t *testing.T) {
 
 func TestDefaultMountTable(t *testing.T) {
 	c, _, _ := TestCoreUnsealed(t)
-	table := c.defaultMountTable()
+	table, err := c.defaultMountTable(context.Background())
+	require.NoError(t, err)
```
phyrog, line 739: `defaultMountTable` does not return an `error`

vault/mount.go
```diff
@@ -1558,6 +1607,13 @@ func (c *Core) persistMounts(ctx context.Context, barrier logical.Storage, table
 					continue
 				}
 
+				ne, err := c.namespaceStore.GetNamespaceByAccessor(ctx, mtEntry.NamespaceID)
+				if err != nil {
+					return -1, err
+				}
+
+				view := ne.View(barrier)
```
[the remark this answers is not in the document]
cipherboy: Ah, previously `TestNamespaceStore` was failing because of the double creation of TokenStore, let me take a look...

cipherboy:
@phyrog @satoqz This passes tests, so am I good to merge?

satoqz:
@cipherboy looks good to me.

Note: I just realized that we always re-persist _all_ mounts in the entire system whenever any path is (un-)mounted, presumably since that's how the legacy mounts work. Could this become an issue at scale? Just a thought.

PR #1051, opened 2025-02-28
Add support for child namespaces

### LINKED ISSUE #1040: CRUD child namespaces
**Describe the bug**
Creating child namespaces is currently not possible. The CRUD operations for namespaces should respect the passed parent namespace.

**To Reproduce**
1. `bao namespace create parent`
2. `bao namespace create -namespace=parent child`
3. `bao namespace list`
4. `bao namespace list -namespace=parent`

**Expected behavior**
Command 3 should return only the `parent` namespace, command 4 should return only the `child` namespace.

**Environment:**
* `namespace` branch

phyrog:
Resolves #1040 and the server part of #1063.

/cc @cipherboy @satoqz @wslabosz-reply @voigt @driif

### DIFF
--- a/helper/namespace/namespace.go
+++ b/helper/namespace/namespace.go
@@ -48,12 +48,6 @@ func (n *Namespace) Validate() error {
 		return fmt.Errorf("%v is a reserved path and cannot be used as a namespace", n.Path)
 	}
 
-	// Canonicalize ensures we have a trailing slash; remove it for this
-	// comparison to ensure we have no other slashes.
-	if strings.Contains(n.Path[:len(n.Path)-1], "/") {
-		return errors.New("path separator ('/') cannot be used in namespace path")
-	}
-
 	return nil
 }
 
--- a/helper/namespace/namespace.go
+++ b/helper/namespace/namespace.go
@@ -71,6 +71,8 @@ var (
 	}
 )
 
+// HasParent returns true if possibleParent is a parent namespace of n.
+// Otherwise it returns false.
 func (n *Namespace) HasParent(possibleParent *Namespace) bool {
 	switch {
 	case possibleParent.Path == "":
@@ -82,14 +84,38 @@ func (n *Namespace) HasParent(possibleParent *Namespace) bool {
 	}
 }
 
+// HasDirectParent returns true if possibleParent is the direct parent of n. Otherwise
+// it returns false.
+func (n *Namespace) HasDirectParent(possibleParent *Namespace) bool {
+	parentPath, ok := n.ParentPath()
+	return ok && parentPath == possibleParent.Path
+}
+
+// ParentPath returns the path of the parent namespace. n.Path must be a
+// canonicalized path.
+func (n *Namespace) ParentPath() (string, bool) {
+	if n.Path == "" {
+		return "", false
+	}
+	segments := strings.SplitAfter(n.Path, "/")
+	if len(segments) <= 2 {
+		return "", true
+	}
+	return strings.Join(segments[:len(segments)-2], ""), true
+}
+
+// TrimmedPath trims n.Path from the given path
 func (n *Namespace) TrimmedPath(path string) string {
 	return strings.TrimPrefix(path, n.Path)
 }
 
+// ContextWithNamespace adds the given namespace to the given context
 func ContextWithNamespace(ctx context.Context, ns *Namespace) context.Context {
 	return context.WithValue(ctx, contextNamespace, ns)
 }
 
+// RootContext adds the root namespace to the given context or returns a new
+// context if the given context is nil
 func RootContext(ctx context.Context) context.Context {
 	if ctx == nil {
 		return ContextWithNamespace(context.Background(), RootNamespace)
@@ -118,7 +144,7 @@ func FromContext(ctx context.Context) (*Namespace, error) {
 }
 
 // Canonicalize trims any prefix '/' and adds a trailing '/' to the
-// provided string
+// provided string. The canonical root namespace path is the empty string.
 func Canonicalize(nsPath string) string {
 	if nsPath == "" || nsPath == "/" {
 		return ""
--- a/vault/logical_system_namespaces.go
+++ b/vault/logical_system_namespaces.go
@@ -168,8 +168,29 @@ func (b *SystemBackend) handleNamespacesList() framework.OperationFunc {
 // handleNamespacesScan handles "/sys/namespaces" endpoint to scan the enabled namespaces.
 func (b *SystemBackend) handleNamespacesScan() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-		// TODO(satoqz): Implement scan once a clear API is available from NamespaceStore.
-		return nil, fmt.Errorf("scan is unimplemented")
+		parent, err := namespace.FromContext(ctx)
+		if err != nil {
+			return nil, err
+		}
+		entries, err := b.Core.namespaceStore.ListNamespaceEntries(ctx, false, true)
+		if err != nil {
+			return nil, err
+		}
+
+		var keys []string
+		keyInfo := make(map[string]interface{})
+		for _, entry := range entries {
+			p := parent.TrimmedPath(entry.Namespace.Path)
+			keys = append(keys, p)
+			keyInfo[p] = map[string]any{
+				"uuid":            entry.UUID,
+				"id":              entry.Namespace.ID,
+				"path":            entry.Namespace.Path,
+				"custom_metadata": entry.Namespace.CustomMetadata,
+			}
+		}
+
+		return logical.ListResponseWithInfo(keys, keyInfo), nil
 	}
 }
 
--- a/vault/logical_system_namespaces.go
+++ b/vault/logical_system_namespaces.go
@@ -6,10 +6,12 @@ package vault
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
 
+	"github.com/openbao/openbao/helper/namespace"
 	"github.com/openbao/openbao/sdk/v2/framework"
 	"github.com/openbao/openbao/sdk/v2/logical"
 )
@@ -137,17 +139,26 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
 // handleNamespacesList handles "/sys/namespaces" endpoint to list the enabled namespaces.
 func (b *SystemBackend) handleNamespacesList() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-		// TODO(satoqz): Use ListNamespaceEntries once this can be rebased so we can also return the UUID fields.
-		namespaces, err := b.Core.namespaceStore.ListNamespaces(ctx, false /* includeRoot */)
+		parent, err := namespace.FromContext(ctx)
+		if err != nil {
+			return nil, err
+		}
+		entries, err := b.Core.namespaceStore.ListNamespaceEntries(ctx, false, false)
 		if err != nil {
 			return nil, err
 		}
 
 		var keys []string
 		keyInfo := make(map[string]interface{})
-		for _, ns := range namespaces {
-			keys = append(keys, ns.Path)
-			keyInfo[ns.Path] = ns
+		for _, entry := range entries {
+			p := parent.TrimmedPath(entry.Namespace.Path)
+			keys = append(keys, p)
+			keyInfo[p] = map[string]any{
+				"uuid":            entry.UUID,
+				"id":              entry.Namespace.ID,
+				"path":            entry.Namespace.Path,
+				"custom_metadata": entry.Namespace.CustomMetadata,
+			}
 		}
 
 		return logical.ListResponseWithInfo(keys, keyInfo), nil
@@ -165,7 +176,11 @@ func (b *SystemBackend) handleNamespacesScan() framework.OperationFunc {
 // handleNamespacesRead handles the "/sys/namespaces/<path>" endpoints to read a namespace.
 func (b *SystemBackend) handleNamespacesRead() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-		path := data.Get("path").(string)
+		path := namespace.Canonicalize(data.Get("path").(string))
+
+		if len(path) > 0 && strings.Contains(path[:len(path)-1], "/") {
+			return nil, errors.New("path must not contain /")
+		}
 
 		ns, err := b.Core.namespaceStore.GetNamespaceByPath(ctx, path)
 		if err != nil {
@@ -192,7 +207,12 @@ func (b *SystemBackend) handleNamespacesRead() framework.OperationFunc {
 // handleNamespaceSet handles the "/sys/namespaces/<path>" endpoint to set a namespace.
 func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-		path := data.Get("path").(string)
+		path := namespace.Canonicalize(data.Get("path").(string))
+
+		if len(path) > 0 && strings.Contains(path[:len(path)-1], "/") {
+			return nil, errors.New("path must not contain /")
+		}
+
 		imetadata, ok := data.GetOk("custom_metadata")
 		var metadata map[string]string
 		if ok {
@@ -204,8 +224,7 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
 			}
 		}
 
-		ns, err := b.Core.namespaceStore.ModifyNamespaceByPath(ctx, path, func(ctx context.Context, ns *NamespaceEntry) (*NamespaceEntry, error) {
-			ns.Namespace.Path = path
+		entry, err := b.Core.namespaceStore.ModifyNamespaceByPath(ctx, path, func(ctx context.Context, ns *NamespaceEntry) (*NamespaceEntry, error) {
 			ns.Namespace.CustomMetadata = metadata
 			return ns, nil
 		})
@@ -214,10 +233,10 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
 		}
 
 		resp := &logical.Response{Data: map[string]interface{}{
-			"uuid":            ns.UUID,
-			"path":            ns.Namespace.Path,
-			"id":              ns.Namespace.ID,
-			"custom_metadata": ns.Namespace.CustomMetadata,
+			"uuid":            entry.UUID,
+			"path":            entry.Namespace.Path,
+			"id":              entry.Namespace.ID,
+			"custom_metadata": entry.Namespace.CustomMetadata,
 		}}
 		return resp, nil
 	}
@@ -242,7 +261,12 @@ func customMetadataPatchPreprocessor(input map[string]interface{}) (map[string]i
 // handleNamespacesPatch handles the "/sys/namespace/<path>" endpoints to update a namespace's custom metadata.
 func (b *SystemBackend) handleNamespacesPatch() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-		path := data.Get("path").(string)
+		path := namespace.Canonicalize(data.Get("path").(string))
+
+		if len(path) > 0 && strings.Contains(path[:len(path)-1], "/") {
+			return nil, errors.New("path must not contain /")
+		}
+
 		ns, err := b.Core.namespaceStore.ModifyNamespaceByPath(ctx, path, func(ctx context.Context, ns *NamespaceEntry) (*NamespaceEntry, error) {
 			if ns.UUID == "" {
 				return nil, fmt.Errorf("requested namespace does not exist")
@@ -283,7 +307,11 @@ func (b *SystemBackend) handleNamespacesPatch() framework.OperationFunc {
 // handleNamespacesDelete handles the "/sys/namespace/<path>" endpoints to delete a namespace.
 func (b *SystemBackend) handleNamespacesDelete() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
-		path := data.Get("path").(string)
+		path := namespace.Canonicalize(data.Get("path").(string))
+
+		if len(path) > 0 && strings.Contains(path[:len(path)-1], "/") {
+			return nil, errors.New("path must not contain /")
+		}
 
 		ns, err := b.Core.namespaceStore.GetNamespaceByPath(ctx, path)
 		if err != nil {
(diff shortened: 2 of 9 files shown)


vault/logical_system_namespaces_test.go
```diff
@@ -0,0 +1,413 @@
...
+			require.True(t, ok, fmt.Sprintf("key_info does not have path %q which is present in keys", path))
+			var ns namespace.Namespace
+			require.NoError(t, mapstructure.Decode(info, &ns), "key_info entry is not a namespace")
+			require.Equal(t, ns.Path, path, "path in key does not match path in namespace struct")
+			require.NotEmpty(t, ns.ID, "namespace ID should not be empty")
+			require.NotEqual(t, ns.ID, namespace.RootNamespaceID, "list should not include root namespace")
+		}
+	})
+
+	t.Run("list nested namespaces", func(t *testing.T) {
```
phyrog, line 376: Test is currently failing due to ListNamespaces implementation, see https://github.com/openbao/openbao/pull/1041#discussion_r1975059934

vault/namespaces_store.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/go-secure-stdlib/base62"
 	uuid "github.com/hashicorp/go-uuid"
 	"github.com/openbao/openbao/helper/namespace"
+	hnamespace "github.com/openbao/openbao/helper/namespace"
```
satoqz, line 21: Nit: I'd prefer not aliasing the `namespace` helper package and instead rename variables of type `NamespaceEntry` named `namespace` to e.g. `entry`. The package name has quite ubiquitous use in all kinds of places already, a variable with the same name is confusing.

satoqz (approved):

vault/namespaces_store.go
phyrog: Fixed. I believe this was there from before a rebase on your branch, since the line above it already imports the same package as `namespace`.

phyrog:
FYI I will refactor this PR again. I do not think it is a good idea anymore to handle the context stuff within the namespace store. This should instead be handled in the calling functions.

The reason is that some of the functions are used in contexts other than requests (ListNamespaces and GetNamespaceByAccessor).

helper/namespace/namespace.go
```diff
@@ -71,25 +71,51 @@ var (
...
-	case possibleParent.Path == "":
+	case possibleAncestor.Path == "":
 		return true
 	case n.Path == "":
 		return false
 	default:
-		return strings.HasPrefix(n.Path, possibleParent.Path)
+		return strings.HasPrefix(n.Path, possibleAncestor.Path)
+	}
+}
```
wslabosz-reply, line 79: so this would look like
```
nsParent = namespace{}
nsChild = namespace{parent: ns1}

nsChild.HasAncestor(nsParent) -> true
nsParent.HasAncestor(nsChild) -> false
```
having a hard time wrapping my head around possibly due to use of language here, so assuming we want to go only up in hierarchy (or down if we consider tree structure), then what about calling it `IsAncestor`, but then the call would look like `nsChild.IsAncestor(nsParent)` which is also misleading if we consider that the method is invoked on the `nsChild` and the "ancestor" we actually match against is an argument...

overall this doesn't really matter, but I'd others opinion how to phrase this relationship between namespaces a bit more clear

wslabosz-reply (approved):

vault/logical_system_namespaces.go
```diff
@@ -193,6 +230,12 @@ func (b *SystemBackend) handleNamespacesRead() framework.OperationFunc {
 func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
 		path := data.Get("path").(string)
```
cipherboy, line 232: It is slightly confusing that we don't do the validation here, but handle it in the namespace store; should we add a comment?

vault/logical_system_namespaces_test.go
```diff
@@ -16,25 +16,35 @@ import (
...
+		ID:             res.Data["id"].(string),
+		Path:           res.Data["path"].(string),
+		CustomMetadata: res.Data["custom_metadata"].(map[string]string),
+	}
 }
 
 func TestNamespaceBackend_Set(t *testing.T) {
 	t.Parallel()
 	b := testSystemBackend(t)
+	rootCtx := namespace.RootContext(context.Background())
```
cipherboy, line 41: Do we not infer the root context? It seems like this might be a breaking change for testing; we don't export `namespace`, which makes it a bit hard for third-party packages to do testing I think?

I wonder if we can update it to return root namespace if there was no namespace on the request?

helper/namespace/namespace.go
```diff
@@ -71,25 +65,51 @@ var (
...
+
+// HasParent returns true if possibleParent is the direct parent of n. Otherwise
+// it returns false.
+func (n *Namespace) HasParent(possibleParent *Namespace) bool {
+	parentPath, ok := n.ParentPath()
+	return ok && parentPath == possibleParent.Path
```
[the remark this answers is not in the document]
cipherboy: It seems like a lot of changes have occurred as a result of this, kinda makes me wonder if this should be `HasDirectParent` or something similar, tbh.

vault/namespaces_store.go
```diff
@@ -604,7 +666,13 @@ func (ns *NamespaceStore) ListNamespaceAccessors(ctx context.Context, includeRoo
 
 	entries := make([]string, 0, len(ns.namespaces))
 	for _, item := range ns.namespaces {
-		if !includeRoot && item.Namespace.ID == namespace.RootNamespaceID {
+		if !includeParent && item.Namespace.ID == parent.ID {
```
cipherboy, line 669: Wonder if we can simplify the `List` behavior to take a filter to return the correct field rather than reimplementing it all again every time. 

Problem for later.

vault/namespaces_store_test.go
```diff
@@ -14,18 +14,18 @@ func TestNamespaceStore(t *testing.T) {
 	c, keys, root := TestCoreUnsealed(t)
 	s := c.namespaceStore
 
-	ctx := context.TODO()
+	ctx := namespace.RootContext(context.TODO())
```
cipherboy, line 17: Here as well, as above.

cipherboy: But maybe it is fine, since changing `FromContext(...)` will cause other problems... Hmm.

cipherboy (commented):
Curious to get others' thoughts on this before merging, but looks fairly good from some local testing! 

I think changes for policies/auth into a parent namespace granting access to a child namespace would have to come later? I'll have to play with that later to see how it works now.

vault/namespaces_store.go
phyrog: Probably, but this is completely refactored in #1113 anyway.

cipherboy: Ah, gotcha. Not yet looked at that one yet.

helper/namespace/namespace.go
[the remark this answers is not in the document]
driif: isChildOf(parent)? 🤔

cipherboy:
Hmm, the panic seems real:

```

2025-03-14T18:30:56.3476472Z 2025-03-14T18:30:56.300Z [INFO]  TestPolicyStore_ListPoliciesByNamespace.core: pre-seal teardown complete
2025-03-14T18:30:56.3476821Z 2025-03-14T18:30:56.300Z [DEBUG] TestPolicyStore_ListPoliciesByNamespace.core: clustering disabled, not stopping listeners
2025-03-14T18:30:56.3477065Z 2025-03-14T18:30:56.300Z [DEBUG] TestPolicyStore_ListPoliciesByNamespace.core: sealing barrier
2025-03-14T18:30:56.3477306Z 2025-03-14T18:30:56.301Z [INFO]  TestPolicyStore_ListPoliciesByNamespace.core: vault is sealed
2025-03-14T18:30:56.3477551Z panic: runtime error: invalid memory address or nil pointer dereference [recovered]
2025-03-14T18:30:56.3477836Z 	panic: runtime error: invalid memory address or nil pointer dereference
2025-03-14T18:30:56.3478034Z [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x2f17fc4]
2025-03-14T18:30:56.3478039Z 
2025-03-14T18:30:56.3478132Z goroutine 1405852 [running]:
2025-03-14T18:30:56.3478265Z testing.tRunner.func1.2({0x355dc20, 0x64345b0})
2025-03-14T18:30:56.3478567Z 	/opt/hostedtoolcache/go/1.24.0/x64/src/testing/testing.go:1734 +0x21c
2025-03-14T18:30:56.3478670Z testing.tRunner.func1()
2025-03-14T18:30:56.3478882Z 	/opt/hostedtoolcache/go/1.24.0/x64/src/testing/testing.go:1737 +0x35e
2025-03-14T18:30:56.3478972Z panic({0x355dc20?, 0x64345b0?})
2025-03-14T18:30:56.3479182Z 	/opt/hostedtoolcache/go/1.24.0/x64/src/runtime/panic.go:787 +0x132
2025-03-14T18:30:56.3479457Z github.com/openbao/openbao/vault.TestPolicyStore_ListPoliciesByNamespace(0xc00262b340)
2025-03-14T18:30:56.3479693Z 	/home/runner/work/openbao/openbao/vault/policy_store_test.go:661 +0x4a4
2025-03-14T18:30:56.3479800Z testing.tRunner(0xc00262b340, 0x3f4af30)
2025-03-14T18:30:56.3480010Z 	/opt/hostedtoolcache/go/1.24.0/x64/src/testing/testing.go:1792 +0xf4
2025-03-14T18:30:56.3480119Z created by testing.(*T).Run in goroutine 1
2025-03-14T18:30:56.3480502Z 	/opt/hostedtoolcache/go/1.24.0/x64/src/testing/testing.go:1851 +0x413
2025-03-14T18:30:56.3480512Z 
2025-03-14T18:30:56.3480729Z ##[endgroup]
2025-03-14T18:30:56.3480930Z   [31mFAIL[0m Package vault (3m37.716s)
```

vault/logical_system_namespaces.go
```diff
@@ -157,31 +168,57 @@ func (b *SystemBackend) handleNamespacesList() framework.OperationFunc {
...
+		ns, err := namespace.FromContext(ctx)
+		if err != nil {
+			return nil, err
+		}
+		path = namespace.Canonicalize(p.Join(ns.Path, path))
 
-		ns, err := b.Core.namespaceStore.GetNamespaceByPath(ctx, path)
+		nsEntry, err := b.Core.namespaceStore.GetNamespaceByPath(ctx, path)
```
driif, line 207: The error reported by Alex is due to the logic in `GetNamespaceByPath`. This function is now expecting a full path instead of a single namespace, as the handling of the full path was previously implemented one level below in `namespace_store`.

helper/namespace/namespace.go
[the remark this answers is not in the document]
phyrog: Since #1113 will remove all uses of this version of `HasParent` again, I'll probably just name it `HasDirectParent` like @cipherboy proposed and undo the changes to the old `HasParent` function. This is currently only used because there is no better way to find the direct children of a namespace.

vault/logical_system_namespaces.go
phyrog: Not sure I follow. Before this PR there were only top-level (well, under root) namespaces, so any single namespace path you had was the full path.

driif: Sorry for the confusion. What I meant was that joining paths to create a full namespace path was happening in the namespace store.

phyrog: I think I get what you mean, but what you describe did not really happen before this PR.

When this is run on the `namespaces` branch at the moment: https://github.com/openbao/openbao/blob/93be1b0e3f26be759457156fc4279e61fe451865/vault/policy_store_test.go#L649-L653

It does not actually create a child namespace of parent. It instead creates a sibling namespace to parent. So you only have `parent/` and `child/`, but they are not nested. There was no joining at all going on, which is what this PR is all about 🙂

But you're right, the logic should be in the namespace store and not in the logical system handlers.

vault/logical_system_namespaces_test.go
phyrog: There was no inferring going on. The implementation just assumed that namespaces are not nested and all namespaces are on the same level below root. So more an implicit assumption (that is not true anymore).

Returning the root namespace when it's missing in context sounds like we are just one little bug away from root access 🙂

vault/logical_system_namespaces.go
cipherboy: Sorry, guess this comment was illegible :D 

I meant the validation that all parents exist, and we don't create foo/bar/baz without foo/bar existing, if our request sets bar/baz as path.

helper/namespace/namespace.go
```diff
@@ -48,12 +48,6 @@ func (n *Namespace) Validate() error {
...
 	}
 
-	// Canonicalize ensures we have a trailing slash; remove it for this
-	// comparison to ensure we have no other slashes.
-	if strings.Contains(n.Path[:len(n.Path)-1], "/") {
-		return errors.New("path separator ('/') cannot be used in namespace path")
-	}
-
 	return nil
 }
```
driif, line 52: I believe the validation that Alex mentioned above [link](https://github.com/openbao/openbao/pull/1051#discussion_r1995949085), is important for validation of path. Previously, you were not able to write:

```bash
./bao namespace create foo/bar/baz
```

Instead, you had to use:

```bash
./bao namespace create -ns=foo/bar baz
```

At first glance, this seems acceptable, but it could introduce edge cases when using HTTP requests instead of the CLI tool. For example, we cannot write:

```
PUT http://127.0.0.1:8202/v1/sys/namespaces/foo/bar
```

This could lead to various issues with paths. Therefore, I suggest adding this path validation in `logical_system_namespaces.go` to ensure consistency and prevent potential errors.

phyrog: From the API docs I think it is ambiguous if this should work or not: `http://127.0.0.1:8202/v1/sys/namespaces/foo/bar`. The segment is called path, so if it means a full path (or just partial), this should be valid. But it could also mean the last path segment.

driif: Allowing slashes ("/") in paths could introduce potential side effects. This means we need to ensure our implementation can handle more complex cases, such as the one illustrated in the `curl` command below:

```bash
curl --location --request PUT 'http://127.0.0.1:8200/v1/bar/baz/sys/namespaces/qux/quux/corge' \
--header 'X-Vault-Token: s.EQyJhxtoxTZfWqfZLDJjet5q' \
--header 'X-Vault-Namespace: foo'
```

I executed this request, but encountered issues with path resolution. Specifically, the application seems to ignore or headers or namespaces after `v1/`, depending on the number of namespaces present between `v1/` and `/sys`.

I plan to create a new issue to address this. 

It seems that allowing slashes in the path introduces more side effects and edge cases without providing significant advantages for the client using this application.

phyrog: I am relatively sure this behavior is fixed in #1113. In the current storage implementation this is difficult to do, which is why I left the todo in place in ResolveNamespaceFromRequest.

With the refactoring this became pretty easy, so the routes you mentioned should work there.

phyrog: That is not to say that I'm not open to discussing limits that we could/should impose on combining paths in different sources. As I see it at the moment, we have 4 different places that define path fragments:

- Token (via accessor id, not path)
- Header
- URL path prefix
- URL path segment (for /sys/namespaces routes)

Whatever we decide, we should add it to the documentation somewhere. I believe the order I wrote is the order in which the namespace path fragments are combined.

vault/logical_system_namespaces_test.go
cipherboy: Ah, I see, we're in an OpenBao test file. I was thinking this was an upstream test file that were modifying to add a root context where there was none before, indicating to me that we had a behavior difference.

But I think we're fine. I concur though, we could be one bug away from root access, though I suppose that's true anyways since we'd always start at the root namespace anyways. :-)

helper/namespace/namespace.go
cipherboy: I don't think token itself actually defines any path segments. The header and URL path prefixes define them, but if you use a token from an arbitrary child namespace and leave off all header/URL paths, I don't think you can access the child namespace given your token comes from there.

More precisely, I don't think the following works:

```
BAO_TOKEN="$(bao write -field=token -namespace=foo/bar/baz userpass)" bao write kv-v2/data/secret a=b
```

(where, we "infer" that the full request path is `foo/bar/baz/kv-v2/data/secret`).

phyrog: I might be wrong here as I haven't looked into it too deeply yet, but I believe in (*Token store).lookupInternal, it does find out the namespace a token belongs to and returns it in the TokenEntry. That in turn is put into context somewhere else (currently on mobile, so searching the code is hard).

phyrog: On further reading, the namespace in the TokenEntry might be only used for policy related stuff and not for setting the base namespace.

cipherboy: Comment here seems relevant:

https://github.com/openbao/openbao/blob/2012bfda69111e79d2945a6dbd1b3291ac95fec0/vault/token_store.go#L1582-L1584

So if the request comes in to the root namespace with a token from a child, we should fail the request IMO.

phyrog: I'll test that on Monday! @cipherboy any opinion on whether /sys/namespaces/:path should allow more than one namespace level (e.g. `PATCH /sys/namespaces/foo/bar`?

IMO the parameter is called path, so would support full (or partial in combination with namespace prefix and header) paths here. Would also be more consistent with what's possible in the header and prefix (those are also not limited to one ns level).

vault/logical_system_namespaces.go
phyrog: All good, I think I understand what you mean. In fact I had it like this for a while but moved it to the logical handlers later to simplify the store (but I did one or two more iterations on the code since then, so this might not be necessary or useful anymore).

I believe the reasoning was that the header and prefix paths are concatenated already before hitting the handler, so doing the last piece of concatenation outside of the store as well we could just forget about the different sources of path fragments in the store itself and just work with absolute paths.

helper/namespace/namespace.go
cipherboy: Hmmm, it is weird. To be consistent with mounts, it should accept an arbitrary path and not create intermediate child namespaces. I.e., if you call:

```
PATCH sys/namespaces/foo/bar
```

You end up with one namespace with `path=foo/bar` and cannot create new namespaces/mounts at `foo`.

When you call:

```
PATCH foo/sys/namespaces/bar
```

you already have to have `foo` existing and create just `./bar` (full path: `foo/bar`).


---

The bigger problem is reserved namespace names. How would you handle:

```
PATCH sys/namespaces/bar/sys/baz
```

for instance? Here, there's an implicit `sys` sibling namespace that gets created presently.

IMHO, it is best to only allow a single path segment and not a full path. So you have to come in to the parent namespace and create the child namespace.

Otherwise, you have a policy problem. A namespace can either be created from `sys/namespace/<prefix>/child` or `<prefix>/sys/namespace/child` and thus needs to be ACL'd both ways. Best to only allow the latter.

vault/logical_system_namespaces.go
driif: Hi [Tom](@phyrog),

I appreciate the changes you've made and understand your reasoning for moving the concatenation outside the store. It makes sense to use "/" in paths to create a full path as early as possible. However, if we can't use "/" in the path, it seems that this approach adds some friction. In context, like the list endpoint, we typically have the full path or the parent path - in create or delete endpoints. This separation simplifies things, as we always have the parent path in context along with the name of the child namespace we're trying to create, delete, or modify.

phyrog: Makes sense, you and @cipherboy convinced me :)

vault/logical_system_namespaces.go
phyrog: Based on the feedback from you and @driif, I added checks to all the handlers that accept a `:path` URL param that forbids `/` in the paths. This means, that we can now rely on the parent namespace existing for sure in context inside the handlers (`(*NamespaceStore).ResolveNamespaceFromRequest` would otherwise fail before), so I don't think any more checking is necessary here.

phyrog:
> Hmm, the panic seems real:
> 
> ...

Fixed now.

helper/namespace/namespace.go
phyrog: This is fixed now. All handlers accepting `:path` URL parameters have a check that makes sure path does not contain any `/`. That means `PATCH sys/namespaces/foo/bar` is no longer possible, but `PATCH foo/sys/namespaces/bar` is.

vault/logical_system_namespaces.go
```diff
@@ -137,17 +139,26 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
...
-			keyInfo[ns.Path] = ns
+		for _, entry := range entries {
+			p := parent.TrimmedPath(entry.Namespace.Path)
+			keys = append(keys, p)
+			keyInfo[p] = map[string]any{
+				"uuid":            entry.UUID,
+				"id":              entry.Namespace.ID,
+				"path":            entry.Namespace.Path,
+				"custom_metadata": entry.Namespace.CustomMetadata,
+			}
```
phyrog, line 161: I am not sure if this is the right thing to do here with the namespace paths. The keys contain the abbreviated version of the path (relative to the parent namespace), while the info field contains the full namespace path.

Same goes for `handleNamespaceScan`.

Any opinions, @driif @cipherboy?

vault/logical_system_namespaces.go
phyrog: Fixed by moving the path handling into the namespace store

vault/logical_system_namespaces.go
driif: @phyrog You are right, it's a right thing to do here.

cipherboy: I think that makes sense. `keys` and any value that appears in the info (`path`) are semantically different, IMO.

cipherboy:
@phyrog hmm, latest test failure also seems valid:

```
2025-03-18T18:09:25.7495013Z ##[group][31mFAIL[0m helper/namespace.TestHasDirectParent (0.00s)
2025-03-18T18:09:25.7502142Z     namespace_test.go:305: bad parent calculation; name: "is root the parent of root", actual: true, expected: false
2025-03-18T18:09:25.7502834Z 
2025-03-18T18:09:25.7503218Z ##[endgroup]
```

phyrog:
> @phyrog hmm, latest test failure also seems valid:
> 
> ```
> 2025-03-18T18:09:25.7495013Z ##[group]�[31mFAIL�[0m helper/namespace.TestHasDirectParent (0.00s)
> 2025-03-18T18:09:25.7502142Z     namespace_test.go:305: bad parent calculation; name: "is root the parent of root", actual: true, expected: false
> 2025-03-18T18:09:25.7502834Z 
> 2025-03-18T18:09:25.7503218Z ##[endgroup]
> ```

@cipherboy Yup, missed a `HasParent` -> `HasDirectParent` rename in the test. Just pushed a fix.

cipherboy (approved):
Thanks @phyrog! Looking good :-)

driif:
Great Job, Tom @phyrog ;)

PR #1055, opened 2025-03-02
fix: fix the wrong error return value

cangqiaoyuzhuo:
In fact, err is incorrect and should be  `intErr ` .

Resolves #


JanMa (approved):
Nice catch, thank you @cangqiaoyuzhuo

PR #1064, opened 2025-03-03
Bump actions/cache to v4, use pinning

cipherboy:


cipherboy:
Due to v3.x deprecation starting [March 1](https://github.com/actions/cache/discussions/1510), we need to backport this as well.

JanMa (approved):

JanMa:
LGTM, but we need to open a follow-up issue to either fix or remove the `test-collect-reports`. In it's current form it just checks if _any_ of the test runs successfully uploaded their test results.

PR #1081, opened 2025-03-04
Add hex to templating for ACL policies

cipherboy:
Per PolicyStore.sanitizeName, ACL policies must have lower case names, not reflected in documentation. This makes the original choice of base64 (in #618) not work well: base64 is case sensitive and thus does not survive lowercase conversion.

Using hex allows us to retain the original purpose but without making the policy store case-sensitive (as that would likely be a breaking change).


JanMa (approved):
LGTM

cipherboy:
Thanks @JanMa!

PR #1087, opened 2025-03-05
Add HSM to download page

cipherboy:
@JanMa This might be one too many tabs, but I thought all of the architecture info was too busy. Thoughts? I can back that part out if you want. 

Screenshot From 2025-03-05 08-51-07

image

### DIFF
--- a/website/src/components/Releases/index.tsx
+++ b/website/src/components/Releases/index.tsx
@@ -1,3 +1,6 @@
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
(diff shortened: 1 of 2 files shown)


JanMa:
I agree that the current page is too busy, but I'm not entirely sure if adding another layer of tabs makes it any better. Maybe a number of dropdowns would be easier to navigate. :thinking: 
Here's an example that works quite well:
https://adoptium.net/temurin/releases/?package=any&os=any&arch=any

cipherboy:
Ah, that might nice. Let me try that!

cipherboy:
This conflicts; I suspect someone else can refactor the downloads page much better than I can right now. Closing!

PR #1102, opened 2025-03-08
Support clearing views via pagination

cipherboy:
I think we need a performance comparison before adopting this. However, this might help address the comments in https://github.com/orgs/openbao/discussions/966: rather than loading the full paths to all 150k secrets, we can load only a page (7.5k secrets) at a time, which should help use less memory.

### DIFF
--- a/sdk/logical/storage.go
+++ b/sdk/logical/storage.go
@@ -28,6 +28,11 @@ var ErrSetupReadOnly = errors.New("cannot write to storage during setup")
 // for more information.
 const PBPWFClusterSentinel = "{{clusterId}}"
 
+// Default number of elements returned from a single call to ListPage. This
+// should roughly fit in 2MB of memory assuming a reasonable path length
+// (275 characters).
+const DefaultScanViewPageLimit = 7500
+
 // Storage is the way that logical backends are able read/write data.
 type Storage interface {
 	List(context.Context, string) ([]string, error)
@@ -67,8 +72,25 @@ type ClearableView interface {
 	Delete(context.Context, string) error
 }
 
+var _ ClearableView = Storage(nil)
+
+type PaginatedClearableView interface {
+	ClearableView
+	ListPage(context.Context, string, string, int) ([]string, error)
+}
+
+var _ PaginatedClearableView = Storage(nil)
+
 // ScanView is used to scan all the keys in a view iteratively
 func ScanView(ctx context.Context, view ClearableView, cb func(path string)) error {
+	// Prefer pagination if we have it.
+	if paginated, ok := view.(PaginatedClearableView); ok {
+		return ScanViewPaginated(ctx, paginated, DefaultScanViewPageLimit, func(page int, index int, path string) (cont bool, err error) {
+			cb(path)
+			return true, nil
+		})
+	}
+
 	frontier := []string{""}
 	for len(frontier) > 0 {
 		n := len(frontier)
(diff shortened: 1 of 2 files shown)


Gabrielopesantos:
Interesting, so the performance is about the same with pagination. Taken into account other places where pagination is used is this expected?

cipherboy:
@Gabrielopesantos The performance we're hoping to improve here isn't time, it is memory. There'll be a little more churn since we're making more function calls (which incur compute & memory), but we don't keep as much in memory at one time. If we have 5 million entries, a single List will store all entries in memory at once, whereas here we might make 1k ListPage calls (each 5k entries returned), but have smaller overhead in memory at any time.

Gabrielopesantos (approved):
Ok, I understand. As the issue linked mentions the deletion timing out I somewhat assumed the idea was improve the overall time.

cipherboy:
> As the issue linked mentions the deletion timing out I somewhat assumed the idea was improve the overall time.

In the event that the sever has very small amounts of memory but large amounts of swap, or lots of parallel deletion operations, this would lead to very large deletion time caused by constant swapping. Using pagination will improve the overall time in that event. :-)

cipherboy:
I have updated this to also acquire a transaction during pagination, if available. While this doesn't guarantee nothing has changed (as each pagination call adds to a write transaction, which will bloat the size of it), it does at least guarantee consistency of the scan like previously. So semantics are overall completely unchanged.

sdk/logical/storage.go
```diff
@@ -64,34 +69,92 @@ func StorageEntryJSON(k string, v interface{}) (*StorageEntry, error) {
...
-			if strings.HasSuffix(c, "/") {
-				frontier = append(frontier, fullPath)
-			} else {
-				cb(fullPath)
+
+			if after == "" && len(contents) == 1 && pageSize > 1 {
+				// In this case, we likely had a mostly-empty "directory" with
+				// only a single entry, with the same name as this "directory".
+				// Exit out of it to avoid an infinite loop.
+				break
```
wslabosz-reply, line 157: I understand the need for this short-circuit, but wouldn't this break some (probably very unlikely) storage paths? like when you create an user with `curl -X POST -H "X-Vault-Token: $(bao print token)" -d '{"password": "123"}' "http://127.0.0.1:8200/v1/auth/userpass/users/user"` you end up with `sys/raw/auth/<mount_uuid>/user/user`

wslabosz-reply (approved):
thank you for adding recursion test!

sdk/logical/storage.go
```diff
@@ -64,34 +69,99 @@ func StorageEntryJSON(k string, v interface{}) (*StorageEntry, error) {
...
+
+			if after == "" && len(contents) == 1 && pageSize > 1 {
+				// In this case, contents[0] == after == "". We hit this when
+				// a key is written to storage with a trailing slash (baz/);
+				// this is still a valid entry, so by semantics of list,
+				// list(baz/) = "", if nothing else resides under baz/. This
+				// is hit in some incorrect path joining operations and so
+				// must still be able to function correctly. Setting after=""
+				// is the default value and must include "" in the listing, so
+				// if we have an adequately large page size and the empty empty
```
wslabosz-reply, line 161:
```suggestion
				// if we have an adequately large page size and the empty
```

cipherboy: Thanks!

cipherboy:
@wslabosz-reply Do you see a failure in the tests? I've rerun three times and each time it says job 7 exited with code 143, but I do not see why. Is there a memory problem here perhaps?

wslabosz-reply:
```
2025-05-26T13:25:43.2390480Z ##[endgroup]
2025-05-26T13:25:43.2391328Z   [32mPASS[0m Package vault/external_tests/router (32.442s)
2025-05-26T13:25:43.2391737Z 
2025-05-26T13:25:58.7943907Z   [33mEMPTY[0m Package . (30ms)
2025-05-26T13:25:58.7947721Z 
2025-05-26T13:28:03.2755154Z ##[error]Process completed with exit code 143.
2025-05-26T13:28:03.2763373Z ##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
```

cipherboy:
Per review from @wslabosz-reply, future maintainer of `/vault`, merging.

Issue #1069, opened 2025-03-04
Namespace-aware Policy Store


satoqz:
**Describe the bug**

Following #1047, the system backend is automatically mounted into namespaces, which enables interaction with the Policy Store via `sys/policy`, `sys/policies/*` paths within namespaces.

However, the Policy Store is not yet aware of namespace storage organization, i.e. it does not write to a view under `namespaces/<uuid>`, but into the root namespace's storage hierarchy.

**To reproduce**

1. Create a namespace:

```sh
bao namespace create foo
```

2. Create a policy in the namespace

```sh
bao policy read default > policy.hcl
bao policy write -namespace=foo bar policy.hcl
```

3. Note that the policy is available under the raw path `sys/raw/sys/policy/bar`:

```sh
bao read sys/raw/sys/policy/bar
```

4. It also shows up if you query policies for the root namespace:

```sh
bao policy list
```

**Expected behavior**

Policies should be stored under their namespace's storage prefix:

```sh
# The `bar` policy should have been listed here, not at `sys/raw/sys/policy`
bao list sys/raw/namespace/$uuid/sys/policy

# Should not have returned `bar` policy.
bao policy list

# Should still return `bar` policy.
bao policy list -namespace=foo
```

**Additional context**

Resolution of this issue likely involves filling in the following "stubs" that presently ignore namespaces:

https://github.com/openbao/openbao/blob/12d005462ba336643d6bb4495f26e91f08c790d0/vault/policy_store_util.go#L12-L22

driif:
Working on It.

PR #1106, opened 2025-03-10
Namespace aware policy store #1069

driif:
Implementation of Policy Store in Namespaces - #1069 
Following #1047 - implementation to conform #787 

---
cc: [@genelet](https://github.com/genelet) @cipherboy @klaus-sap [@satoqz](https://github.com/satoqz) [@phyrog](https://github.com/phyrog)


vault/policy_store.go
```diff
@@ -509,34 +524,41 @@ func (ps *PolicyStore) ListPoliciesWithPrefix(ctx context.Context, policyType Po
...
-	// Scan the view, since the policy names are the same as the
-	// key names.
-	var keys []string
-	switch policyType {
-	case PolicyTypeACL:
-		keys, err = logical.CollectKeysWithPrefix(ctx, view, prefix)
-	default:
-		return nil, fmt.Errorf("unknown policy type %q", policyType)
+	// Use direct List instead of CollectKeysWithPrefix to avoid collecting keys from all namespaces
+	keys, err := view.List(ctx, prefix)
```
cipherboy, line 528: This might break the LIST endpoint, tbh.

If we put the policies for different namespaces in separate space (`/namespace/<uuid>/sys/...`) we shouldn't have this problem FWIW.

driif: okay, understood.

vault/namespaces_store.go
```diff
@@ -392,6 +397,28 @@ func (ns *NamespaceStore) createMounts(ctx context.Context, entry *NamespaceEntr
 	return nil
 }
 
+// createPolicyStore creates a policy store for the specified namespace entry
+// and loads default policies for this namespace.
+func (ns *NamespaceStore) createPolicyStore(ctx context.Context, entry *NamespaceEntry) error {
+	// Create a context with the namespace we're working in
+	nsCtx := namespace.ContextWithNamespace(ctx, entry.Namespace)
```
wslabosz-reply, line 404: I think that `ctx` passed to this method already contain the metadata of the namespace.

driif: @wslabosz-reply Yes, you're absolutely right.

vault/namespaces_store.go
```diff
@@ -92,7 +92,7 @@ func (ne *NamespaceEntry) View(barrier logical.Storage) BarrierView {
 		return NewBarrierView(barrier, "")
 	}
 
-	return NewBarrierView(barrier, path.Join("namespace", ne.UUID)+"/")
+	return NewBarrierView(barrier, path.Join("namespaces", ne.UUID)+"/")
```
satoqz, line 95: Are you sure? IIUC the RFC specifies `/sys/namespaces` for the namespace store, but `/namespace/<uuid>` for per-namespace storage.

driif: @satoqz,
I'm not sure, but this approach might improve our API compatibility.

@cipherboy Alex – what are your thoughts on this?

vault/policy_store.go
```diff
@@ -363,6 +384,32 @@ func (ps *PolicyStore) GetNonEGPPolicyType(nsID string, name string) (*PolicyTyp
...
+		return ps.aclView
+	}
+
+	// First, get the namespace's entry from the namespace store
+	nsEntry, err := ps.core.namespaceStore.GetNamespaceByAccessor(context.Background(), ns.ID)
+	if err != nil || nsEntry == nil {
+		// If we can't find the namespace, fall back to the root view
+		ps.logger.Error("failed to find namespace entry, using root view", "namespace_id", ns.ID)
+		return ps.aclView
+	}
```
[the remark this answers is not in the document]
driif: @satoqz You're right, Jonas. I'll fix it.

vault/policy_store.go
```diff
@@ -363,6 +384,32 @@ func (ps *PolicyStore) GetNonEGPPolicyType(nsID string, name string) (*PolicyTyp
 	return &policyType, nil
 }
 
+// getACLView returns the ACL view for the root namespace or a subview for a non-root namespace.
+func (ps *PolicyStore) getACLView(ns *namespace.Namespace) BarrierView {
+	if ns == nil || ns.ID == namespace.RootNamespaceID {
+		// For root namespace, use the original ACL view
+		return ps.aclView
+	}
```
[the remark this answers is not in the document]
driif: @satoqz That's a great point! I agree it would be better to use (*NamespaceEntry).View() as the single source of truth for path behavior in the root namespace as well. I'll work on figuring out a solution that might allow us to eliminate the aclView field entirely. A simpler, more consistent approach would definitely be beneficial.

vault/policy_store.go
```diff
@@ -711,3 +744,23 @@ func (ps *PolicyStore) sanitizeName(name string) string {
 func (ps *PolicyStore) cacheKey(ns *namespace.Namespace, name string) string {
 	return path.Join(ns.ID, name)
 }
+
+// loadDefaultPolicies loades default policies for the namespace in the provided context
```
voigt, line 748:
```suggestion
// loadDefaultPolicies loads default policies for the namespace in the provided context
```

voigt (changes requested):

vault/policy_store.go
driif: @voigt Thnx, fixed.

vault/namespaces_store_test.go
```diff
@@ -161,6 +161,10 @@ func TestNamespaceStore(t *testing.T) {
 		},
 	}
 
+	// Override the store.
+	s = c.namespaceStore
+	// After sealing and unsealing, the namespace stored in the core is replaced with a new one.
+	// owever, the s.SetNamespace function is still using the previous namespace.
```
satoqz, line 167: Should probably do this following both of the re-unseals in the test, even if it results in no unexpected behavior yet for first re-unseal.

satoqz (approved):
Looks good to me so far!

vault/namespaces_store.go
```diff
@@ -367,6 +367,11 @@ func (ns *NamespaceStore) createMounts(ctx context.Context, entry *NamespaceEntr
 	// so create a new context with the newly created child namespace.
 	nsCtx := namespace.ContextWithNamespace(ctx, entry.Clone().Namespace)
 
+	// Create default policies for this namespace
+	if err := ns.core.policyStore.loadDefaultPolicies(nsCtx); err != nil {
+		return fmt.Errorf("error creating default policies: %w", err)
+	}
+
```
wslabosz-reply, line 374: very ignorable nitpick: comment on the `(*NamespaceStore).createMounts` doesn't mention the creation of default policies, could we add something like that?

driif: @wslabosz-reply I agree, it could be confusing from a naming perspective. I'll refactor the code to separate these concerns.

wslabosz-reply (approved):
great work on answering all questions, also finding that bug in the namespace_store test, LGTM

vault/policy_store.go
```diff
@@ -35,6 +35,9 @@ const (
 	// responseWrappingPolicyName is the name of the fixed policy
 	responseWrappingPolicyName = "response-wrapping"
 
+	// controlGroupPolicyName is the name of the fixed policy
+	controlGroupPolicyName = "control-group"
```
cipherboy, line 39: We do not have control groups, so I'd remove this. We'd need a separate RFC for adding them and we can introduce this then.

vault/policy_store_test.go
```diff
@@ -193,7 +194,7 @@ func testPolicyStorePredefined(t *testing.T, ps *PolicyStore, ns *namespace.Name
 	if err != nil {
 		t.Fatalf("err: %v", err)
 	}
-	// This shouldn't contain response-wrapping since it's non-assignable
+	// This shouldn't contain response-wrapping and control-group since it's non-assignable
```
cipherboy, line 197:
```suggestion
	// This shouldn't contain response-wrapping since it's non-assignable
```

as above

cipherboy (changes requested):
One set of changes and then I'll merge @driif!

vault/policy_store.go
driif: @cipherboy Okay, I'll remove them. Since it’s a small change, we can easily add them later.

vault/policy_store_test.go
driif: @cipherboy  fixed it!

vault/policy_store.go
```diff
@@ -363,6 +361,29 @@ func (ps *PolicyStore) GetNonEGPPolicyType(nsID string, name string) (*PolicyTyp
...
 }
 
+// getACLView returns the ACL view for the given namespace
+func (ps *PolicyStore) getACLView(ns *namespace.Namespace) (BarrierView, error) {
+	if ns == nil {
+		ns = namespace.RootNamespace
+	}
+
+	// First, get the namespace's entry from the namespace store
+	nsEntry, err := ps.core.namespaceStore.GetNamespaceByAccessor(context.Background(), ns.ID)
```
phyrog, line 371: Let's use (*Core).NamespaceById here to keep some boundaries.

satoqz: I don't think that works, since a full NamespaceEntry is required to get the view (and in turn that is because it has the UUID). Generally something I'm unhappy with.

satoqz: To go into more detail, we already resolve the namespace for each request and place it into context, yet in many backends we then end up having to perform the same namespace store lookup _again_ to get the view just because the UUID isn't included in context.

phyrog: Yeah you're right, I missed that it uses the entry here.

cipherboy: I think that can be simplified if we just extend the definition of namespace. I had commented elsewhere but my original mental model was that most of `helper/namespace.Namespace` was made accessible under the external, plugin-facing sdk/ side, but it wasn't so we don't need to worry about compatibility and can extend our definition of `Namespace` as we see fit.

satoqz: That's great, highly in favor of ultimately unifying namespace.Namespace and NamespaceE try then.

Issue #1110, opened 2025-03-12
Namespace-aware Identity Store


satoqz:
**Describe the bug**

Following #1047, we create an entirely new instance of `IdentityStore` per namespace. We should probably keep it to a single one as:
1) `Core.identityStore` is used in many places, and is assumed to account for namespaces
2) A singleton `IdentityStore` already knows to distinguish namespaces internally, and provides any cross-namespace identity functionality.

At the same time, `IdentityStore` currently _heavily_ assumes a single storage view, saved in the `IdentityStore.view` field and initially set via `BackendConfig` (this is equal to what `req.Storage` provides, so it could be determined dynamically). This storage view is propagated to the store's `entityPacker`, `localAliasPacker` and `groupPacker`, which are all fields of the store struct and the main interfaces for persistence in request handling.

This is where it may get a bit complicated: storage packers are essentially stateful views (as they manage a view + list of locks), however with the namespace RFC's design we have a storage view for each namespace. Some rethinking is required here, do we store a list of storage packers, do we make a single storage packer handle views dynamically?

**Additional notes**

As far as I can tell from a quick look, it seems that for [cross-namespace access via identity outside of the namespace hierarchy](https://developer.hashicorp.com/vault/docs/enterprise/namespaces/configure-cross-namespace-access) we might only be missing the API endpoint at `sys/config/group_policy_application`.

cipherboy:
> As far as I can tell from a quick look, it seems that for [cross-namespace access via identity outside of the namespace hierarchy](https://developer.hashicorp.com/vault/docs/enterprise/namespaces/configure-cross-namespace-access) we might only be missing the API endpoint at `sys/config/group_policy_application`.

If memory serves, this was added well after the initial namesapces application and so we could delay this portion to future work.

driif:
Start working on it.

cipherboy:
I believe this was fixed in #1159.

wslabosz-reply:
@cipherboy @driif 
I've noticed while working on https://github.com/openbao/openbao/pull/1336, that we never addressed the
> At the same time, IdentityStore currently heavily assumes a single storage view, saved in the IdentityStore.view field and initially set via BackendConfig (this is equal to what req.Storage provides, so it could be determined dynamically). This storage view is propagated to the store's entityPacker, localAliasPacker and groupPacker, which are all fields of the store struct and the main interfaces for persistence in request handling.

[...]

Issue #1068, opened 2025-03-04
Namespace-aware Token Store


satoqz:
**Describe the bug**

Following #1047, the Token Store (+backend) is automatically mounted into namespaces, e.g. for a namespace `foo/` the `foo/token/*` endpoints are available.

However, the Token Store is not yet aware of namespace storage organization, i.e. it does not write to a view under `namespaces/$uuid`, but into the root namespace's storage hierarchy.

Furthermore, creating a namespace currently instantiates an additional Token Store per namespace. This is unwanted, see discussion in https://github.com/openbao/openbao/pull/1047#discussion_r1975566865 and https://github.com/openbao/openbao/pull/1047#issuecomment-2690396137. Rather, the singleton Token Store should be re-used across namespaces, yet remain aware of per-namespace storage. This is also in line with our singleton Cubbyhole as described in #1067, as the Token Store is directly linked to the Cubbyhole.

**To Reproduce**

1. In an empty OpenBao dev instance, note that there is only one token (the root token) under `/sys/raw/sys/token`:

```sh
bao list sys/raw/sys/token/accessor
# ... one entry
```

2. Create a namespace, then create a token in it:

```
bao namespace create test
bao token create -namespace=test
```

3. Note that there are now two tokens under `sys/raw/sys/token`:

```sh
bao list sys/raw/sys/token/accessor
# ... two entries
```

**Expected behavior**

Per-namespace token storage should be under the namespace's storage prefix:

```sh
bao list sys/raw/namespace/$uuid/sys/token/accessor
# ... token should be here
```

**Additional notes**

From reading the existing request handling and token store logic, it seems that the associated authentication logic (i.e., which namespaces does which token allow access to) should mostly be in place already. This should be double-checked, and any missing bits filled in if needed.

wslabosz-reply:
opened PR for #1067 and I'd like to start working on this one, @cipherboy could you assign me to it? Thanks!

PR #1112, opened 2025-03-12
Implement namespace aware Cubbyhole #1067

### LINKED ISSUE #1067: Namespace-aware Cubbyhole backend
**Describe the bug**

Following #1047, the Cubbyhole backend is automatically mounted into namespaces, e.g. for a namespace `foo/` the `foo/cubbyhole/*` endpoints are available.

However, Cubbyhole is not yet aware of namespace storage organization, i.e. it does not write to a view under `namespaces/$uuid`, but into the root namespace's storage hierarchy.

Furthermore, creating a namespace currently instantiates an additional Cubbyhole backend per namespace. This is unwanted, see discussion in https://github.com/openbao/openbao/pull/1047#discussion_r1975566865. Rather, the singleton Cubbyhole backend should be re-used across namespaces, yet remain aware of per-namespace storage.

**To Reproduce**

1. Create a namespace and write to its Cubbyhole:

```
bao namespace create test
bao write test/cubbyhole/foo bar=baz
```

2. Check where it is writing to:

```sh
bao list sys/raw/logical
# ... should see several entries, one of them being the namespace's cubbyhole data

uuid=$(bao namespace lookup -format json test | jq -r .data.uuid)
bao list sys/raw/namespace/$uuid/logical
# ... no matching entry found
```

**Expected behavior**

Per-namespace Cubbyhole data should be stored under the namespace's storage prefix:

```sh
bao list sys/raw/namespace/$uuid/logical
# ... cubbyhole entry is displayed
```

wslabosz-reply:
Resolves #1067

 - Implements Cubbyhole (storage-wise) awareness of namespace by refactoring 
 `(*mountEntry).View()` into `(*Core).mountEntryView()`
 - Adjusts the logical backend creation of `CubbyholeNS` to reuse the existing root owned Cubbyhole backend

___

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -147,8 +147,10 @@ func (c *Core) enableCredentialInternal(ctx context.Context, entry *MountEntry, 
 	// Sync values to the cache
 	entry.SyncCache()
 
-	viewPath := entry.ViewPath()
-	view := NewBarrierView(c.barrier, viewPath)
+	view, err := c.mountEntryView(ctx, entry)
+	if err != nil {
+		return err
+	}
 
 	origViewReadOnlyErr := view.GetReadOnlyErr()
 
@@ -1034,10 +1036,11 @@ func (c *Core) setupCredentials(ctx context.Context) error {
 	for _, entry := range c.auth.sortEntriesByPathDepth().Entries {
 		var backend logical.Backend
 
-		// Create a barrier view using the UUID
-		viewPath := entry.ViewPath()
-
-		view := NewBarrierView(c.barrier, viewPath)
+		var view BarrierView
+		view, err = c.mountEntryView(ctx, entry)
+		if err != nil {
+			return err
+		}
 
 		origViewReadOnlyErr := view.GetReadOnlyErr()
 
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -1032,12 +1032,8 @@ func (c *Core) setupCredentials(ctx context.Context) error {
 	c.authLock.Lock()
 	defer c.authLock.Unlock()
 
-	var err error
 	for _, entry := range c.auth.sortEntriesByPathDepth().Entries {
-		var backend logical.Backend
-
-		var view BarrierView
-		view, err = c.mountEntryView(ctx, entry)
+		view, err := c.mountEntryView(ctx, entry)
 		if err != nil {
 			return err
 		}
@@ -1055,6 +1051,7 @@ func (c *Core) setupCredentials(ctx context.Context) error {
 		// Initialize the backend
 		sysView := c.mountEntrySysView(entry)
 
+		var backend logical.Backend
 		backend, entry.RunningSha256, err = c.newCredentialBackend(ctx, entry, sysView, view)
 		if err != nil {
 			c.logger.Error("failed to create credential entry", "path", entry.Path, "error", err)
--- a/vault/core.go
+++ b/vault/core.go
@@ -1241,7 +1241,7 @@ func (c *Core) configureLogicalBackends(backends map[string]logical.Factory, log
 
 	// Cubbyhole
 	logicalBackends[mountTypeCubbyhole] = CubbyholeBackendFactory
-	logicalBackends[mountTypeNSCubbyhole] = func(ctx context.Context, conf *logical.BackendConfig) (logical.Backend, error) {
+	logicalBackends[mountTypeNSCubbyhole] = func(ctx context.Context, config *logical.BackendConfig) (logical.Backend, error) {
 		if c.cubbyholeBackend != nil {
 			return c.cubbyholeBackend, nil
 		}
--- a/vault/core.go
+++ b/vault/core.go
@@ -1241,7 +1241,12 @@ func (c *Core) configureLogicalBackends(backends map[string]logical.Factory, log
 
 	// Cubbyhole
 	logicalBackends[mountTypeCubbyhole] = CubbyholeBackendFactory
-	logicalBackends[mountTypeNSCubbyhole] = logicalBackends[mountTypeCubbyhole]
+	logicalBackends[mountTypeNSCubbyhole] = func(ctx context.Context, conf *logical.BackendConfig) (logical.Backend, error) {
+		if c.cubbyholeBackend != nil {
+			return c.cubbyholeBackend, nil
+		}
+		return nil, errors.New("cubbyhole backend does not exist")
+	}
 
 	// System
 	logicalBackends[mountTypeSystem] = func(ctx context.Context, config *logical.BackendConfig) (logical.Backend, error) {
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -2194,6 +2194,20 @@ func (c *Core) readMigrationStatus(migrationID string) *MountMigrationInfo {
 	return &migrationInfo
 }
 
+func (c *Core) namespaceMountEntryView(ctx context.Context, namespaceID, prefix string) (BarrierView, error) {
+	nsEntry, err := c.namespaceStore.GetNamespaceByAccessor(ctx, namespaceID)
+	// corrupted namespace store
+	if err != nil {
+		return nil, err
+	}
+
+	if nsEntry == nil {
+		return nil, errors.New("namespace not found")
+	}
+
+	return nsEntry.View(c.barrier).SubView(prefix), nil
+}
+
 // mountEntryView returns the barrier view object with prefix depending on the mount entry type, table and namespace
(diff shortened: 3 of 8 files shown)


vault/mount.go
```diff
@@ -2190,3 +2193,38 @@ func (c *Core) readMigrationStatus(migrationID string) *MountMigrationInfo {
...
+		return NewBarrierView(c.barrier, systemBarrierPrefix), nil
+	case mountTypeToken:
+		return NewBarrierView(c.barrier, systemBarrierPrefix+tokenSubPath), nil
+	}
+
+	switch me.Table {
+	case mountTableType:
+		// NSCubbyhole should be stored under the namespace prefix (UUID)
+		if me.Type == mountTypeNSCubbyhole {
+			nsEntry, err := c.namespaceStore.GetNamespaceByAccessor(ctx, me.NamespaceID)
```
phyrog, line 2210: Let's use (*Core).NamespaceById here instead

phyrog (approved):

vault/core.go
```diff
@@ -1241,7 +1241,12 @@ func (c *Core) configureLogicalBackends(backends map[string]logical.Factory, log
 
 	// Cubbyhole
 	logicalBackends[mountTypeCubbyhole] = CubbyholeBackendFactory
-	logicalBackends[mountTypeNSCubbyhole] = logicalBackends[mountTypeCubbyhole]
+	logicalBackends[mountTypeNSCubbyhole] = func(ctx context.Context, conf *logical.BackendConfig) (logical.Backend, error) {
+		if c.cubbyholeBackend != nil {
+			return c.cubbyholeBackend, nil
+		}
+		return nil, errors.New("cubbyhole backend does not exist")
+	}
```
wslabosz-reply, line 1249: This doesn't sit right with me, but I don't have a clear alternative, maybe @cipherboy you've got some thought about it?

Same pattern (or very similar) would have to be implemented for other singleton backends (e.g. Token), so maybe introduction of more readable and reusable pattern would be beneficial here...

vault/mount.go
```diff
@@ -2190,3 +2193,38 @@ func (c *Core) readMigrationStatus(migrationID string) *MountMigrationInfo {
...
+			return nsEntry.View(c.barrier).SubView(backendBarrierPrefix + me.UUID + "/"), nil
+		}
+		return NewBarrierView(c.barrier, backendBarrierPrefix+me.UUID+"/"), nil
+	case credentialTableType:
+		return NewBarrierView(c.barrier, credentialBarrierPrefix+me.UUID+"/"), nil
+	case auditTableType:
+		return NewBarrierView(c.barrier, auditBarrierPrefix+me.UUID+"/"), nil
+	}
+
+	panic("invalid mount entry")
```
driif, line 2229: IMHO, the current implementation using `panic("invalid mount entry")` as a fallback for unrecognized mount entry types could be problematic. While this might be intended as an assertion that should never occur in practice, I'm concerned about using panic for input validation, especially since the function already returns errors elsewhere.

wslabosz-reply: Changing the panic for returning error makes sense to me, will adjust.

vault/core.go
wslabosz-reply: Looking at the code I see one main difference that `CredentialBackend` is not a field on the `Core`, so we wouldn't be able to reuse the same pattern as this one

vault/auth.go
```diff
@@ -1034,10 +1036,11 @@ func (c *Core) setupCredentials(ctx context.Context) error {
...
 
-		// Create a barrier view using the UUID
-		viewPath := entry.ViewPath()
-
-		view := NewBarrierView(c.barrier, viewPath)
+		var view BarrierView
+		view, err = c.mountEntryView(ctx, entry)
+		if err != nil {
+			return err
+		}
```
driif, line 1043:
```suggestion
        view, err := c.mountEntryView(ctx, entry)
        if err != nil {
            return err
        }
```

driif: **CONSIDER**: Use the short variable declaration (`:=`) directly instead of pre-declaring `view`, as this is the pattern you use consistently elsewhere in your code. Error shadowing isn't a concern here since you handle the error immediately.

vault/core.go
cipherboy: I think it makes sense; I don't see many alternatives unless we rewrite the meaning of `logicalBackends[...]` or its usage.

driif: @wslabosz-reply @cipherboy I think there's an opportunity for architectural improvement here.
This pattern creates a factory that doesn't actually construct anything new - it just returns the existing singleton.

For a more declarative architecture, we could consider:

1. Initialize the cubbyhole backend once during core setup
2. Directly assign this instance to the mountTypeNSCubbyhole entry without a factory
3. Focus only on ensuring the proper storage view routing via mountEntryView

This would make the singleton pattern more explicit in the code and separate the concerns of backend instantiation from storage routing.

While the current PR correctly fixes the immediate issue with storage paths, this might be worth considering in a future refactoring to make the codebase more maintainable.

LGTM for the current changes addressing the storage path issue:)

driif: I mean in general - looks like using factory here for CubbyholeBackendFactory isn't the obvious choice since we only need to create one instance of the backend.

cipherboy (commented):
Some comments and then I'll merge, curious what you think. Looking good though!

vault/core.go
cipherboy: @driif Right, but this otherwise gates several mount decisions so we need something in this list unless we remove the list entirely in favor of something else, which seems like a much larger architectural change than worth it in this PR.

I suspect we could maybe change the direct calls of the factory to become a singleton-check/set operation though, with other usage being just a check.

But for another PR.

wslabosz-reply: > 1. Initialize the cubbyhole backend once during core setup

This definitely makes sense to me, but as Alex said, it would require some adjustments to the `logicalBackends[...]` pattern usage

> 2. Directly assign this instance to the mountTypeNSCubbyhole entry without a factory

This would also require refactor of the `logicalBackends`

> 3. Focus only on ensuring the proper storage view routing via mountEntryView

Not sure if I follow, do you mean it as not having one Cubbyhole backend instance?

> This would make the singleton pattern more explicit in the code and separate the concerns of backend instantiation from storage routing.

Agree, although "Factory" pattern doesn't necessarily mean it cannot return a singleton instance. Definitely would like to change that name as I think that it introduces confusion.

@driif @cipherboy
I would propose to revisit the "make the singleton backend mounts more architecturally clean" after token store ns awareness [implementation](https://github.com/openbao/openbao/issues/1068) which from the initial glance will introduce more changes into a `TokenStore` object implementation (as a Singleton)

vault/mount.go
```diff
@@ -2190,3 +2191,41 @@ func (c *Core) readMigrationStatus(migrationID string) *MountMigrationInfo {
...
+
+// mountEntryView returns the barrier view object with prefix depending on the mount entry type, table and namespace
+func (c *Core) mountEntryView(ctx context.Context, me *MountEntry) (BarrierView, error) {
+	switch me.Type {
+	case mountTypeSystem:
+		return NewBarrierView(c.barrier, systemBarrierPrefix), nil
+	case mountTypeToken:
+		return NewBarrierView(c.barrier, systemBarrierPrefix+tokenSubPath), nil
+	// Namespace mounts should be stored under the namespace prefix (UUID)
+	case mountTypeNSCubbyhole:
```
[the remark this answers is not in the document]
wslabosz-reply: I wasn't sure about `mountTypeNSSystem` but I assumed `Identity` would be tackled with #1110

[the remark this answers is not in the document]
wslabosz-reply: I also noticed while working on #1068 that some of these views (like for `mountTypeToken`) are totally disregarded later on with no real usage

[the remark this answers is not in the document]
wslabosz-reply: > Ah interesting, it wasn't handled in the original function. But we'll need it for the namespace-aware variant at least, no?

haven't tested the system backend functionality in context of a namespace, how would you approach verifying the correct storage path?

Also if I'm gonna tackle the system mount in this PR, I'd also ensure that the system backend is a singleton.

vault/core.go
cipherboy: Yeah, I think later is fine for this.

vault/mount.go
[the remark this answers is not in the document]
cipherboy: Hm, fair, let's split into separate tickets to investigate the sys as well.

The main storage that I have in mind right now that sys/ would write to is for child namespaces, but that is in #1051 so perhaps this is fine for now.

cipherboy (approved):

driif (approved):

PR #1122, opened 2025-03-17
Fix ACME TLS documentation

### LINKED ISSUE #1121: Duplicated documentation for ACME TLS settings
**Describe the bug**

The documentation for `tls_acme_ca_directory` is duplicated.

**To Reproduce**

View the documentation here: https://github.com/openbao/openbao/blob/7a1cee6468219fba85a0c0d61281d1061f8a311a/website/content/docs/configuration/listener/tcp.mdx?plain=1#L223-L227

**Expected behavior**

I think this should be `tls_acme_ca_directory` and `tls_acme_test_ca_directory`.

**Environment:**

N/A, documentation issue.

**Additional context**

Will open a PR (contributing guide says to open an issue for all PRs except "trivial" fixes, not sure if this counts as trivial).

kwohlfahrt:
The documentation had a confusing duplication of `tls_acme_ca_directory`. From looking at the defaults & the code of `certmagic`, it looks like one of them should be `tls_acme_test_ca_directory` instead.

Resolves #1121

Also add a note about the `tls_acme_cache_path` - if I undestand correctly, this is the `Storage` passed to the `certmagic` instance, and needs to be shared across cluster members if issuing certificates for a cluster behind a load-balancer or otherwise using a shared hostname. I was confused about how this might work when reading about the feature. Happy to remove this change though, since it's not directly related to the issue.

### DIFF
--- a/website/content/docs/configuration/listener/tcp.mdx
+++ b/website/content/docs/configuration/listener/tcp.mdx
@@ -233,6 +233,9 @@ in the root of the TCP listener configuration.
   certificates. If this not writable, `/var/lib/openbao/certmagic` and then
   a temporary directory in `$TEMPDIR` will be used as fallback.
 
+  If issuing certificates for a shared cluster hostname, this must be shared
+  across all nodes.
+
 - `tls_acme_ca_root` `(string: "")` - Path to an optional Root CA to use to
   validate connections to the ACME directory. Defaults to all system
   certificates when empty.
--- a/website/content/docs/configuration/listener/tcp.mdx
+++ b/website/content/docs/configuration/listener/tcp.mdx
@@ -223,8 +223,10 @@ in the root of the TCP listener configuration.
 - `tls_acme_ca_directory` `(string: "https://acme-v02.api.letsencrypt.org/directory")` - Path
   to the ACME directory; defaults to Let's Encrypt's production CA.
 
-- `tls_acme_ca_directory` `(string: "")` - Path
-  to the ACME directory; defaults to Let's Encrypt's staging CA.
+- `tls_acme_test_ca_directory` `(string: "")` - Path
+  to the test ACME directory; defaults to Let's Encrypt's staging CA if
+  `tls_acme_ca_directory` is the Let's Encrypt production CA (otherwise defaults
+  to empty/disabled).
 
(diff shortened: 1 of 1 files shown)


cipherboy:
> Also add a note about the `tls_acme_cache_path` - if I undestand correctly, this is the Storage passed to the certmagic instance, and needs to be shared across cluster members if issuing certificates for a cluster behind a load-balancer or otherwise using a shared hostname.

I think that's one way of using it, but I don't think it necessarily needs to be true. If not using a shared cache, my understanding is each node would fetch certificates on its own, and you'll end up with N certificates. You might, if using a public CA, hit rate limits, but otherwise if you're running against an internal CA, this should be just fine.

kwohlfahrt:
> If not using a shared cache, my understanding is each node would fetch certificates on its own, and you'll end up with N certificates.

I'm not sure how this would work, but I might be missing something.

Let's say I have OpenBao at `bao.example.com`. This DNS name resolves to a load-balancer, which distributes traffic to 3 instances of OpenBao. The instances all start up, talk to Let's Encrypt, which tells them to put a particular response at `.well-known/acme-challenge/$TOKEN`, where each instance gets a unique `$TOKEN` and response combination.

When Let's Encrypt sends a request to check each of the `bao.example.com/.well-known/acme-challenge/$TOKEN` URLs, the issue is that the load-balancer will pick an instance at random to serve that request, which 2/3 of the time will be an isntance that doesn't know about that particular token, so the verification will fail.

That also seems supported by [this section of the certmagic docs](https://github.com/caddyserver/certmagic?tab=readme-ov-file#behind-a-load-balancer-or-in-a-cluster):

> CertMagic runs effectively behind load balancers and/or in cluster/fleet environments. In other words, you can have 10 or 1,000 servers all serving the same domain names, all sharing certificates and OCSP staples.
>
> To do so, simply ensure that each instance is using the same Storage. That is the sole criteria for determining whether an instance is part of a cluster.

cipherboy:
@kwohlfahrt Ahh, I'm with you now. A DNS round-robin load balancer or similar. Yes, then this would require a shared cache.

I was thinking separate DNS names for each node, behind a TLS-terminating LB. Your internal names would all be separate and distinct for the node, so you wouldn't need a shared cache. And your LB would acquire certificates for the shared name separately.

Thank you!

cipherboy (approved):

Issue #1086, opened 2025-03-05
Namespace-aware Plugin Mounting/Storage


satoqz:
**Describe the bug**

Mounting and unmounting secrets engines and auth methods within namespaces appears to work out of the box following #1047. Mounts are also correctly persisted under `namespace/$uuid/core/{mounts,local-mounts}`.

However, once mounted, the plugins themselves don't write to per-namespace storage yet.

**To reproduce**

Create a namespace, enable the kv engine inside, write an entry to kv:

```sh
bao namespace create foo
bao secrets enable -namespace=foo kv
bao kv put -namespace=foo -mount=kv foo bar=baz
```

See how the entry isn't stored in namespace storage:

```sh
bao list sys/raw/logical/$(bao read -namespace=foo -format=json sys/mounts/kv | jq -r .data.uuid)
```

**Expected behavior**

"Plugin data" should be written to `namespace/$namespace_uuid/logical/$mount_uuid`.

**Additional context**

Most likely the fix would consist of tweaking `vault/mount_util.go` such that `req.Storage` points to the right path. This is also required for related issues, see #1067 and #1084.

phyrog:
@cipherboy could you assign this to @voigt and me?

cipherboy:
@voigt you'll have to comment to be assignable but yes.

voigt:
This is my comment to be assignable. ;)

PR #1123, opened 2025-03-17
Namespace-aware token store #1068

### LINKED ISSUE #1068: Namespace-aware Token Store

wslabosz-reply:
Resolves #1068 

 - [x] Implements CredentialBackend (TokenStore) storage awareness of namespace, removing `View` fields from `TokenStore` object and implementing `(*TokenStore) [x]View (*Namespace)` methods 
 - [x] Adjusts the logical backend creation of `CredentialBackend[mountTypeNSToken]` to reuse the existing root owned `TokenStore`
- [x] Removes the `storageView logical.Storage` field from `Cubbyhole Backend` as it was only used in one specific test which was refactored
- [x] Adds unit tests for proper token storage validation

Open questions (**answered**):
- (c *Core) mountEntryView for `mountTypeNSToken` doesn't seem to have impact on the storage logic for the token store. This is because the `NewTokenStore()` constructor doesn't use this "view" but rather was reusing the system one (this has changed with this PR to instead dynamically create a "view" based on the namespace). This introduces a question whether `(*Router) Mount()` requires this view to be created or not. -> **Answered by Alex, it might not be used by the token store operations, but is definitely used by plugins.**
___


cipherboy:
> * This introduces a question whether `(*Router) Mount()` requires this view to be created or not.

If I understand your question correctly, I believe the answer is yes. We'll keep using the system view for token store (which we'll want to make sure points to the relevant namespace system view), but the mount view being updated will help fix #1086, though this can be under a different PR.

vault/token_store.go
```diff
@@ -844,6 +830,41 @@ func NewTokenStore(ctx context.Context, logger log.Logger, core *Core, config *l
...
+	return ts.core.namespaceMountEntryView(ctx, ns.ID, systemBarrierPrefix+tokenSubPath+parentPrefix)
+}
+
+func (ts *TokenStore) rolesView(ctx context.Context, ns *namespace.Namespace) (BarrierView, error) {
+	if ns.ID == namespace.RootNamespaceID {
+		return ts.core.systemBarrierView.SubView(tokenSubPath + rolesPrefix), nil
+	}
+	return ts.core.namespaceMountEntryView(ctx, ns.ID, systemBarrierPrefix+tokenSubPath+rolesPrefix)
+}
+
```
driif, line 867:
```suggestion
// viewFactory is a helper function to create token store barrier views based on the prefix and namespace
func (ts *TokenStore) viewFactory(ctx context.Context, ns *namespace.Namespace, subPath string) (BarrierView, error) {
    if ns.ID == namespace.RootNamespaceID {
        return ts.core.systemBarrierView.SubView(tokenSubPath + subPath), nil
    }
    return ts.core.namespaceMountEntryView(ctx, ns.ID, systemBarrierPrefix+tokenSubPath+subPath)
}

// baseView returns the base token store view for the given namespace
func (ts *TokenStore) baseView(ctx context.Context, ns *namespace.Namespace) (BarrierView, error) {
    return ts.viewFactory(ctx, ns, "")
}

// idView returns the token ID storage view for the given namespace
func (ts *TokenStore) idView(ctx context.Context, ns *namespace.Namespace) (BarrierView, error) {
    return ts.viewFactory(ctx, ns, idPrefix)
}

// accessorView returns the accessor storage view for the given namespace
func (ts *TokenStore) accessorView(ctx context.Context, ns *namespace.Namespace) (BarrierView, error) {
    return ts.viewFactory(ctx, ns, accessorPrefix)
}

// parentView returns the parent token reference storage view for the given namespace
func (ts *TokenStore) parentView(ctx context.Context, ns *namespace.Namespace) (BarrierView, error) {
    return ts.viewFactory(ctx, ns, parentPrefix)
}

// rolesView returns the token role storage view for the given namespace
func (ts *TokenStore) rolesView(ctx context.Context, ns *namespace.Namespace) (BarrierView, error) {
    return ts.viewFactory(ctx, ns, rolesPrefix)
}
```

driif: looks like a candidate for a new factory:)

phyrog: You could probably also do something like this and get rid of the factory method:

```go
func (ts *TokenStore) baseView(ctx context.Context, ns *namespace.Namespace) (BarrierView, error) {
	if ns.ID == namespace.RootNamespaceID {
		return ts.core.systemBarrierView.SubView(tokenSubPath), nil
	}
	return ts.core.namespaceMountEntryView(ctx, ns.ID, systemBarrierPrefix+tokenSubPath)
}

func (ts *TokenStore) idView(ctx context.Context, ns *namespace.Namespace) (BarrierView, error) {
    return ts.baseView(ctx, ns).SubView(idPrefix)
}

```

phyrog (approved):
The view code could be simplified, but otherwise LGTM

driif:
You've already added unit tests for basic namespace isolation functionality. I suggest adding more tests for crossnamespace isolation could be beneficial:
- Token lookup across namespaces
- Token revocation effects across namespaces
- Parent-child relationships spanning namespaces

vault/token_store.go
wslabosz-reply: agree, Alex already gave us a greenlight on that, so I'd be happy if someone started working on this (me included)

vault/testing.go
```diff
@@ -358,6 +358,16 @@ func TestCoreSeal(core *Core) error {
...
+func TestCoreCreateNamespaces(core *Core, namespaces ...*namespace.Namespace) error {
+	for _, ns := range namespaces {
+		err := core.namespaceStore.SetNamespace(context.Background(), &NamespaceEntry{Namespace: ns})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
cipherboy: Completely unrelated but I hate how tests don't give you the ability to create a context which expires when the test should finish and/or is canceled. Would be nice to have in many places.

vault/token_store.go
cipherboy: I don't mind removing the errors, but wonder if we'll ever want them back. Things like namespace lazy loading, per-namespace seal mechanisms, hybrid cluster namespaces, &c could inject errors into a lot of places upstream didn't have.

vault/testing.go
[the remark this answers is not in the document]
wslabosz-reply: @driif
revisiting your request for tests after some time spent on reading through code
> Token lookup across namespaces

token lookup behavior is fully controlled by the token's policies, meaning that all token's are separated by default (no matter the namespace) and they have visibility into any given token with appropriate ACL policy. This means that I cannot write any sensible tests for this on the level of the token store methods.

> Token revocation effects across namespaces

I've concocted something

> For parent-child relationships spanning namespaces, we should test whether a token stored in a child or parent namespace only affects the namespace where the token store is used. This means testing both sibling namespaces and parent-child namespaces to ensure token stores are isolated.

This is also solely policy based, testing this would require testing the ACL's actually, so I don't think this should be required in the context of this PR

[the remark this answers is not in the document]
driif: @wslabosz-reply Thank you for your feedback. I completely agree with your points. I've reviewed the tests you've added, and your code looks good to me. Thank you for your contribution to the project! :)

driif (approved):

wslabosz-reply:
@cipherboy I think most of the issues touched upon in comments were resolved, could you revisit this PR?

cipherboy (approved):
The comments before notwithstanding, I do see how this would be much simpler if we hadn't gotten the view to error.

But I think it is still a step in a positive direction!

Issue #1031, opened 2025-02-22
OIDC Test Failure


cipherboy:
```
2025-02-21T22:31:47.7116669Z ##[group][31mFAIL[0m vault.TestOIDC_PeriodicFunc/test-key (8.43s)
2025-02-21T22:31:47.7117579Z 2025-02-21T22:31:39.283Z [INFO]  TestOIDC_PeriodicFunc/test-key: creating error injector
2025-02-21T22:31:47.7119176Z     identity_store_oidc_test.go:1165: For key: test-key at cycle: 2 expected namedKey's KeyRing to be at least of length 3 but was: 2
2025-02-21T22:31:47.7120873Z     identity_store_oidc_test.go:1176: For key: test-key at cycle: 2 expected public keys to be at least of length 3 but was: 2
2025-02-21T22:31:47.7121960Z 2025-02-21T22:31:47.710Z [DEBUG] TestOIDC_PeriodicFunc/test-key.core: shutdown called
2025-02-21T22:31:47.7123059Z 2025-02-21T22:31:47.710Z [INFO]  TestOIDC_PeriodicFunc/test-key.core: marked as sealed
2025-02-21T22:31:47.7124011Z 2025-02-21T22:31:47.710Z [DEBUG] TestOIDC_PeriodicFunc/test-key.core: clearing forwarding clients
2025-02-21T22:31:47.7125073Z 2025-02-21T22:31:47.710Z [DEBUG] TestOIDC_PeriodicFunc/test-key.core: done clearing forwarding clients
2025-02-21T22:31:47.7126139Z 2025-02-21T22:31:47.710Z [INFO]  TestOIDC_PeriodicFunc/test-key.core: pre-seal teardown starting
2025-02-21T22:31:47.7127114Z 2025-02-21T22:31:47.710Z [DEBUG] TestOIDC_PeriodicFunc/test-key.expiration: stop triggered
2025-02-21T22:31:47.7128226Z 2025-02-21T22:31:47.710Z [TRACE] TestOIDC_PeriodicFunc/test-key.expiration.job-manager: terminating job manager...
2025-02-21T22:31:47.7129460Z 2025-02-21T22:31:47.710Z [TRACE] TestOIDC_PeriodicFunc/test-key.expiration.job-manager: terminating dispatcher
2025-02-21T22:31:47.7130696Z 2025-02-21T22:31:47.710Z [DEBUG] TestOIDC_PeriodicFunc/test-key.expiration: finished stopping
[...]

PR #1129, opened 2025-03-18
Allow periodicFunc to execute "now"

suprjinx:
Small additional fix for flakey test.

Relates #1031

### DIFF
--- a/changelog/1129.txt
+++ b/changelog/1129.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+vault: addresses a timing issue revealed in OIDC_PeriodicFunc test
+```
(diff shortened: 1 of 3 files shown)


suprjinx:
@cipherboy I think this change will better resolve the PeriodicFunc test flakiness

cipherboy (approved):
If this still causes us issues, I realized I should probably remind people about these helpers:

https://github.com/openbao/openbao/blob/93be1b0e3f26be759457156fc4279e61fe451865/vault/testing.go#L1016-L1032

The "periodic" nature can't be tested (which I think is under test here, to be fair), but the contents of the helper can be.

PR #1133, opened 2025-03-19
feat: add `bao namespace scan`

### LINKED ISSUE #1063: Support `SCAN` in Namespaces API
Following #1041, the CRUD functionality of the namespaces API is mostly complete, with the omission of the `SCAN/logical.ScanOperation` endpoint. While `LIST` returns only the direct child namespaces of the request's namespace, `SCAN` would return a recursive list of all child namespaces.

#1041 adds related stubs:
https://github.com/openbao/openbao/blob/72be6cdf11f22f31b2cc1af33f6b8d88faaaf621/vault/logical_system_namespaces.go#L157-L163

An exemplary `LIST` response currently looks like:

```json
{
  "data": {
    "keys": [
      "bar/",
      "foo/"
    ],
    "key_info": {
      "bar/": {
        "custom_metadata": {},
        "id": "HWmNL",
        "path": "bar/"
      },
      "foo/": {
        "custom_metadata": {},
        "id": "5q39x",
        "path": "foo/"
      }
    }
  }
}
```

accordingly, a `SCAN` response would look like:

```json
{
  "data": {
    "keys": [
      "bar/",
      "foo/",
      "foo/baz/"
    ],
    "key_info": {
      "bar/": {
        "custom_metadata": {},
        "id": "HWmNL",
        "path": "bar/"
      },
      "foo/": {
        "custom_metadata": {},
        "id": "5q39x",
        "path": "foo/"
      },
      "foo/baz/": {
        "custom_metadata": {},
        "id": "6r40y",
        "path": "foo/baz/"
      }
    }
  }
}
```

...assuming there is a single further child namespace `baz` under `foo`.

> *@cipherboy I'm making the assumption here that `SCAN` would also return the entire `key_info` + `custom_metadata` like `LIST` would - correct?*

#1051 is a hard requirement for this to be implemented, and waiting for #1052 to be addressed beforehand is likely also a good idea.

voigt:
This PR adds `bao namespace scan` subcommand.

This command recursively lists all (child) namespaces and is namespace aware.

This resolves #1063

```
$ bao namespace scan
Keys
----
reply/
reply/comsysto/
reply/ki/
reply/ki/christoph/
reply/liquid/

$ bao namespace scan -ns=reply
Keys
----
comsysto/
ki/
ki/christoph/
liquid/
```

### DIFF
--- a/command/namespace_scan.go
+++ b/command/namespace_scan.go
@@ -26,7 +26,7 @@ func (c *NamespaceScanCommand) Synopsis() string {
 
 func (c *NamespaceScanCommand) Help() string {
 	helpText := `
-Usage: bao namespace san [options]
+Usage: bao namespace scan [options]
 
   Lists the enabled child namespaces recursively.
 
--- a/command/namespace_scan.go
+++ b/command/namespace_scan.go
@@ -34,6 +34,10 @@ Usage: bao namespace san [options]
 
       $ bao namespace scan
 
+  List enabled child namespaces relative to parent:
+
+      $ bao namespace scan -namespace=my-parent
+
 ` + c.Flags().Help()
 
 	return strings.TrimSpace(helpText)
--- a/command/namespace_scan.go
+++ b/command/namespace_scan.go
@@ -1,4 +1,4 @@
-// Copyright (c) HashiCorp, Inc.
+// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
 // SPDX-License-Identifier: MPL-2.0
 
 package command
--- a/command/namespace_scan.go
+++ b/command/namespace_scan.go
@@ -1,0 +1,123 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package command
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/cli"
+	"github.com/posener/complete"
+)
+
+var (
+	_ cli.Command             = (*NamespaceScanCommand)(nil)
+	_ cli.CommandAutocomplete = (*NamespaceScanCommand)(nil)
+)
+
+type NamespaceScanCommand struct {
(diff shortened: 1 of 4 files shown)


command/namespace_scan.go
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) HashiCorp, Inc.
```
cipherboy, line 1: I'd add our copyright header as well. Scan is net-new to OpenBao, though I understand the structure of commands is very similar across all of them.

command/namespace_scan.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (c *NamespaceScanCommand) Help() string {
+	helpText := `
+Usage: bao namespace san [options]
+
+  Lists the enabled child namespaces recursively.
+
+  List all enabled child namespaces recursively:
+
+      $ bao namespace scan
```
cipherboy, line 35: Maybe also list relative to a parent?

```
$ bao namespace scan -namespace=my-parent
```

command/namespace_scan.go
```diff
@@ -0,0 +1,123 @@
...
+	*BaseCommand
+}
+
+func (c *NamespaceScanCommand) Synopsis() string {
+	return "List all (child) namespaces recursively"
+}
+
+func (c *NamespaceScanCommand) Help() string {
+	helpText := `
+Usage: bao namespace san [options]
```
wslabosz-reply, line 29:
```suggestion
Usage: bao namespace scan [options]
```

voigt: There are already options like `-ns` global options like `-format`...

wslabosz-reply: you are right, I was thinking about "local" options, sorry!

wslabosz-reply (approved):
LGTM

cipherboy (approved):
Thanks @voigt!

PR #1135, opened 2025-03-20
fix(docs/pkcs11): remove documentation regarding key generation

pree:
OpenBao requires the key to be generated externally inside the HSM. This PR removes phrases suggesting OpenBao can generate the key by itself.


JanMa (approved):
LGTM, thanks @pree

Issue #1137, opened 2025-03-20
Docs need clarity re: PostgreSQL


udf2457:
The docs needs to make its mind up about PostgreSQL 😆 

At the top we have a big warning:

> The PostgreSQL Storage Backend is in early preview. Use at your own risk.

Meanwhile, a tiny bit lower down the page, we have:

> Production Ready – the PostgreSQL backend is considered production ready

So, which is it ? 🤷

cipherboy:
Ah, the first warning should be removed now. Do you want to open a PR removing `The PostgreSQL Storage Backend is in early preview. Use at your own risk.`?

PR #1138, opened 2025-03-20
Remove Postgres beta warning

udf2457:
@cipherboy says the PostgreSQL cookie is baked 😉 

#1137


cipherboy:
Yes, I think there's more work to do as part of the namespaces WG, but that shouldn't hold things up yet.

cipherboy (approved):

cipherboy:
Thank you @udf2457!

PR #1146, opened 2025-03-21
docs: Add SSH CA Multi-issuer RFC to website docs

### LINKED ISSUE #679: RFC - SSH CA Multi-issuer

Gabrielopesantos:
Following the work done on #880, this PR adds the RFC, #679, to the website's RFCs.

Resolves #679

### DIFF
--- a/website/content/docs/rfcs/index.mdx
+++ b/website/content/docs/rfcs/index.mdx
@@ -37,3 +37,5 @@ Steering Committee.
    removes mount table limits, allowing potentially hundreds of thousands of
    mounts on a single scaled-up server. This landed in
    [622](https://github.com/openbao/openbao/pull/622).
+ - [SSH CA Multi-issuer](/docs/rfcs/ssh-ca-multi-issuer), adds support for multiple issuers on SSH secret engine mounts.
+    Landed in [880](https://github.com/openbao/openbao/pull/880).
--- a/website/content/docs/rfcs/ssh-ca-multi-issuer.mdx
+++ b/website/content/docs/rfcs/ssh-ca-multi-issuer.mdx
@@ -1,0 +1,97 @@
+---
+sidebar_label: SSH CA Multi-issuer
+description: |-
+    Adds support for multiple issuers on SSH secret engine mounts
+---
+
+# SSH CA Multi-issuer
+
+### Summary
+
+OpenBao’s SSH engine is limited to one Certificate Authority (CA) issuer per mount. This complicates the key rotation process, which is usually a desired feature for security and operational continuity. As previously done in the [PKI engine](https://github.com/hashicorp/vault/pull/15277), this document proposes an implementation to support multiple issuers per SSH mount. This proposal tries to stay consistent with the PKI implementation, having a ‘default’ issuer and a set of new endpoints to read/update it, and not break any backward compatibility with older versions of OpenBao or Upstream.
+
+### Problem Statement
+
+OpenBao’s current SSH engine design imposes limitations by allowing only one Certificate Authority (CA) issuer per mount. This restriction complicates the key rotation process, an important feature for security and operational continuity. To perform key rotation, new SSH mounts must be created, which necessitates duplicating all existing configurations (roles, ACLs, etc.) and updating all references to the new mounts or swapping the old and the new mount points via [two move operations](https://openbao.org/api-docs/system/remount/). This is not only time-consuming but prone to misconfigurations and errors.
+
+While an alternative approach–deleting and recreating the CA issuer within an existing mount–can be used, it introduces significant risks. During the gap between deleting the old key and creating a new one, any requests for issuing access certificates fail. If the import or generation of a new CA issuer fails, this leads to service disruption. Even if the new key is created quickly, it lacks trust, as it is a fresh random key that cannot be cross-signed like in PKI. The key distribution process to all hosts is not instantaneous, leaving a window where authentication requests will fail.
+
+This limitation highlights the need for a safer key rotation mechanism that ensures the system can securely fetch and trust new keys before the actual rotation occurs. Moreover, the ability to revert to a previous key in case of a failure or unexpected events would provide fallback capabilities.
+
+### User-facing description
+
+With this feature, OpenBao’s ‘privileged’ users can submit CA information for more than one issuer, having the ability to support multiple issuers in the same SSH mount. Besides the unique identifier, each issuer will optionally have a unique name that can be used to identify it. The concept of a ‘default’ issuer will be introduced to the existing endpoints, that use the issuer’s key material, stay backwards compatible. The ‘default’ issuer may be read/updated through the newly introduced endpoint, config/issuers. Roles allow the selection of an explicit issuer, allowing operators to choose issuers for users without them necessarily noticing the impact.
+The behavior of existing endpoints will be adapted to conform with this feature. New endpoints will also be added so operations on CA’s can be performed.
+
+### Technical Description
+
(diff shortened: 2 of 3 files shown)


cipherboy (approved):
Thanks @Gabrielopesantos!

PR #1157, opened 2025-03-25
implement namespace aware plugin/auth storage

### LINKED ISSUE #1086: Namespace-aware Plugin Mounting/Storage

phyrog:
Resolves #1086 

/cc @cipherboy @driif @voigt @wslabosz-reply @satoqz

### DIFF
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -2213,15 +2213,18 @@ func (c *Core) mountEntryView(ctx context.Context, me *MountEntry) (BarrierView,
 		return NewBarrierView(c.barrier, systemBarrierPrefix), nil
 	case mountTypeToken:
 		return NewBarrierView(c.barrier, systemBarrierPrefix+tokenSubPath), nil
-	// Namespace mounts should be stored under the namespace prefix (UUID)
-	case mountTypeNSCubbyhole:
-		return c.namespaceMountEntryView(ctx, me.NamespaceID, backendBarrierPrefix+me.UUID+"/")
 	}
 
 	switch me.Table {
 	case mountTableType:
+		if me.Namespace() != nil && me.NamespaceID != namespace.RootNamespaceID {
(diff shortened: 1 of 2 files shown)


vault/mount.go
```diff
@@ -2213,13 +2213,13 @@ func (c *Core) mountEntryView(ctx context.Context, me *MountEntry) (BarrierView,
 		return NewBarrierView(c.barrier, systemBarrierPrefix), nil
 	case mountTypeToken:
 		return NewBarrierView(c.barrier, systemBarrierPrefix+tokenSubPath), nil
-	// Namespace mounts should be stored under the namespace prefix (UUID)
-	case mountTypeNSCubbyhole:
```
phyrog, line 2217: this is now also caught by the mount table case below

wslabosz-reply (approved):
LGTM, love that this makes it more clear for other mountTypes/backends as well

cipherboy (approved):

cipherboy:
Thanks @phyrog!

PR #1158, opened 2025-03-25
Namespace-aware Expiration Manager #1125

### LINKED ISSUE #1125: Namespace-aware Expiration Manager
**Describe the bug**

ExpirationManager stores the tokens expiration entries on root's `/sys/raw/sys/expire`, this will will introduce unnecessary complexity for the namespace sealing (or deletion, revoking) operations. We should adjust that so the expiration manager stores these entries under the namespace owned path `/sys/raw/namespaces/$uuid/sys/expire` 

~Also since https://github.com/openbao/openbao/pull/1047, we create a new instance of `ExpirationManager` per namespace. We should probably keep it to a single one.~
EDIT: `ExpirationManager` kept being a singleton, so no need to adjust this.

**To Reproduce**

1. In an empty OpenBao dev instance, note that there is only one token (the root token) under /sys/raw/sys/token with one corresponding expire entry under /sys/raw/sys/expire/id/auth/token/create
```
bao list sys/raw/sys/token/accessor
# ... one entry

bao list sys/raw/sys/expire/id/auth/token/create
# ... one entry
```

2. Create a namespace, then create a token in it:
```
bao namespace create test
bao token create -namespace=test
```

3. List the tokens in the namespace (assuming #1068 is implemented and merged)
```
bao list sys/raw/namespaces/$uuid/sys/token/accessor
# ... token entry

bao list sys/raw/namespaces/$uuid/sys/expire.... <-- doesn't exist
```

**Expected behavior**
Per-namespace expiration manager should be under the namespace's storage prefix:
```
bao list sys/raw/namespace/$uuid/sys/expire/id/auth/token/create
# ... expiration entry should be here
```

**Additional context**
Expiration of the leases should be correctly handled by the manager, although should be double checked during implementation.

wslabosz-reply:
Resolves #1125  

Note: this branch was created from #1123, will rebase as soon as the #1123 gets merged

 - [x] Implements Expiration Manager storage awareness of namespace, removing `View` fields from `ExpirationManager` object and refactoring `(*ExpirationManager) leaseView()`, `(*ExpirationManager) tokenIndexView()`, `(*ExpirationManager) collectLeases()` methods
 - [x] Adjusts `ExpirationManager` unit tests with namespace awareness

___


vault/expiration.go
```diff
@@ -453,6 +446,44 @@ func (c *Core) stopExpiration() error {
...
+	if err != nil {
+		return nil, 0, fmt.Errorf("failed to list all namespaces: %w", err)
+	}
+
+	for _, namespace := range namespaces {
+		view, err := m.leaseView(m.quitContext, namespace)
+		if err != nil {
+			return nil, 0, fmt.Errorf("failed to determine lease view for namespace: %w", err)
+		}
+		keys, err := logical.CollectKeys(m.quitContext, view)
```
cipherboy, line 476: In #1102 I'm adding support for clearing views with pagination. I might hold on that change until Namespaces land given I'm worried it'll conflict, but it'd be great to update this to behave similarly. There's no point collecting all leases from all namespaces in a single batch, better to use pagination and iteratively handle lease revocation. My 2c.

cipherboy (approved):
I think this looks good from my side, but will let @driif and others take a look first in case I missed anything.

driif (approved):

Issue #1084, opened 2025-03-05
Namespace-aware Quota subsystem


satoqz:
Currently, the quota subsystem (which at this time is only used for HTTP rate limits) does not respect namespaces yet.

There are two parts to this that each need work:
1. Quota configuration/evaluation
2. Quota storage

Regarding 1), Configuring rate limit quota doesn't pick up the namespace that it is created in, and always uses the root namespace. This is because the request's namespace is ignored here:

https://github.com/openbao/openbao/blob/12d005462ba336643d6bb4495f26e91f08c790d0/vault/logical_system_quotas.go#L304-L307

Create a namespace + quota:

```sh
bao namespace create foo
bao write -namespace=foo sys/quotas/rate-limit/bar rate=1 interval=5s
```

Now issue two requests to some endpoint in the root namespace that is not exempt from rate limiting, within the span of 5 seconds:

```sh
bao list sys/raw
bao list sys/raw
```

You should not get rate limited on the second request because you're talking to the root namespace, not the foo namespace. Except you _do_ get rate limited...

We can also do some digging in `sys/raw` to find the resulting saved quota configuration. You'll need to look in `sys/raw/logical/<uuid of namespace's sys mount>/quotas/rate-limit/bar`:

```sh
bao read sys/raw/logical/2bb121ff-daaf-0dca-9511-676b6cd14b58/quotas/rate-limit/bar
# {"id":"c88f2c32-d04f-00bc-b56a-94a025a623f7","type":"rate-limit","name":"bar","namespace_path":"","mount_path":"","role":"","path_suffix":"","rate":100,"interval":1000000000,"block_interval":0}
```

Note how `namespace_path` is empty.

Regarding 2), the path above is of course entirely wrong and unexpected, and the result of how `req.Storage` is currently determined:

https://github.com/openbao/openbao/blob/12d005462ba336643d6bb4495f26e91f08c790d0/vault/mount_util.go#L10-L29

Adapting the `mount_util.go` file is probably also required for (at least) #1067.

`namespace_path` should reflect the namespace that the quota was configured in. The related quota evaluation logic ([here](https://github.com/openbao/openbao/blob/12d005462ba336643d6bb4495f26e91f08c790d0/vault/quotas/quotas.go#L567-L568)) seems to already handle that field correctly, it is just never set.

Outstanding question: **Under what path exactly do we want per-namespace quota to be stored?**

An additional relevant excerpt of the code related to the quota storage path is here:

[...]

PR #1161, opened 2025-03-25
Add namespace-aware rate-limit quota

voigt:
This PR addresses #1084 and replicates namespace-aware quota behavior.

#### 1. Enables quota configuration per namespace
  ```bash
$ bao namespace create ns1
Key                Value
---                -----
custom_metadata    map[]
id                 g7B5Hn
path               ns1/
uuid               4f4afac3-adbf-baad-6576-6444a9e19a64
$ bao write sys/quotas/rate-limit/foo-limit path="ns1" rate=100 interval=2s
Success! Data written to: sys/quotas/rate-limit/foo-limit
$ bao read sys/quotas/rate-limit/foo-limit
Key               Value
---               -----
block_interval    0
inheritable       false
interval          2
name              foo-limit
path              ns1/
rate              100
role              n/a
type              rate-limit
  ```

The quota api endpoints under `/sys/quotas` remain restricted and can't be called in namespaces. Consequently, writing a quota with namespace header will fail.

```bash
$ bao write -ns ns1 sys/quotas/rate-limit/foo-limit rate=100 interval=2s
Error writing data to sys/quotas/rate-limit/foo-limit: Error making API request.

Namespace: ns1/
URL: PUT http://127.0.0.1:8200/v1/sys/quotas/rate-limit/foo-limit
Code: 400. Errors:

* unavailable operation (PUT /v1/sys/quotas/rate-limit/foo-limit)
```

#### 2. Writes quota configuration to `/sys/quotas`

Global, as well as namespace specific quotas will be stored in the same, global, location.

```bash
$ bao write sys/quotas/rate-limit/global-limit rate=3000 interval=1m
Success! Data written to: sys/quotas/rate-limit/global-limit
$ bao list sys/quotas/rate-limit/
Keys
----
foo-limit
global-limit
```

#### 4. Adds `inheritable` flag to quotas
  ```bash
# Update `foo-limit` to inherit quota configuration to child namespaces
$ bao write sys/quotas/rate-limit/foo-limit path="ns1" rate=2 interval=2s inheritable=true
Success! Data written to: sys/quotas/rate-limit/foo-limit
$ bao namespace create -ns ns1 ns1.1
Key                Value
---                -----
custom_metadata    map[]
id                 3i7Ufc
path               ns1/ns1.1/
uuid               382c8d0d-f713-413c-c724-36e1f0712045
# Quota of `ns1` now also applies to `ns1/ns1.1`

$ while true; sleep 0.1; bao namespace list -ns=ns1/ns1.1; end
No namespaces found
No namespaces found
No namespaces found
No namespaces found
Error listing namespaces: Error making API request.

Namespace: ns1/ns1.1/
URL: GET http://127.0.0.1:8200/v1/sys/namespaces?list=true
Code: 429. Errors:

* request path "sys/namespaces/": rate limit quota exceeded
  ```

### DIFF
--- a/http/util.go
+++ b/http/util.go
@@ -143,12 +143,6 @@ func wrapMaxRequestSizeHandler(handler http.Handler, props *vault.HandlerPropert
 
 func rateLimitQuotaWrapping(handler http.Handler, core *vault.Core) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ns, err := namespace.FromContext(r.Context())
-		if err != nil {
-			respondError(w, http.StatusInternalServerError, err)
-			return
-		}
-
 		// We don't want to do buildLogicalRequestNoAuth here because, if the
 		// request gets allowed by the quota, the same function will get called
 		// again, which is not desired.
@@ -157,6 +151,7 @@ func rateLimitQuotaWrapping(handler http.Handler, core *vault.Core) http.Handler
 			respondError(w, status, err)
 			return
 		}
+		ns, _ := core.NamespaceByPath(r.Context(), path)
 		mountPath := strings.TrimPrefix(core.MatchingMount(r.Context(), path), ns.Path)
 
 		quotaReq := &quotas.Request{
--- a/http/util.go
+++ b/http/util.go
@@ -98,6 +98,11 @@ var (
 			}
 		}
 
+		namespacedQuotaRequest := (strings.Contains(fullURL, "sys/quotas") && ns.Path != "")
+		if namespacedQuotaRequest {
+			return http.StatusBadRequest
+		}
+
 		for _, api := range containsAPIs {
 			if strings.Contains(fullURL, api) && !strings.HasPrefix(fullURL, api) {
 				return http.StatusBadRequest
--- a/vault/logical_system_quotas.go
+++ b/vault/logical_system_quotas.go
@@ -301,13 +301,7 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 		}
 
 		mountPath := sanitizePath(d.Get("path").(string))
-		ns, err := namespace.FromContext(ctx)
-		if err != nil {
-			return logical.ErrorResponse("failed to find namespace in context: %v", err), nil
-		}
-		if ns.ID != namespace.RootNamespaceID {
-			mountPath = strings.TrimPrefix(mountPath, ns.Path)
-		}
+		ns, mountPath := b.Core.NamespaceByPath(ctx, mountPath)
 
 		var pathSuffix string
 		if mountPath != "" {
--- a/vault/logical_system_quotas.go
+++ b/vault/logical_system_quotas.go
@@ -303,7 +303,7 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 		mountPath := sanitizePath(d.Get("path").(string))
 		ns, err := namespace.FromContext(ctx)
 		if err != nil {
-			return logical.ErrorResponse("invalid namespace"), nil
+			return logical.ErrorResponse("failed to find namespace in context: %v", err), nil
 		}
 		if ns.ID != namespace.RootNamespaceID {
 			mountPath = strings.TrimPrefix(mountPath, ns.Path)
--- a/vault/logical_system_quotas.go
+++ b/vault/logical_system_quotas.go
@@ -149,6 +149,11 @@ The 'rate' must be positive.`,
 					Description: `If set, when a client reaches a rate limit threshold, the client will be prohibited
 from any further requests until after the 'block_interval' has elapsed.`,
 				},
+				"inheritable": {
+					Type: framework.TypeBool,
+					Description: `If set to true, child namespaces will use this quota, unless another more specific
+quota exists. Can only be set on namespace quotas. A quota on the root namespace will by default be inheritable.`,
+				},
 			},
 			Operations: map[logical.Operation]framework.OperationHandler{
 				logical.UpdateOperation: &framework.PathOperation{
@@ -197,6 +202,10 @@ from any further requests until after the 'block_interval' has elapsed.`,
 								},
 								"block_interval": {
 									Type:     framework.TypeInt,
+									Required: true,
+								},
+								"inheritable": {
+									Type:     framework.TypeBool,
 									Required: true,
 								},
 							},
@@ -351,9 +360,17 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 			return nil, err
 		}
 
+		inheritable := false
+		inheritableI, ok := d.GetOk("inheritable")
+		if ok {
+			inheritable = inheritableI.(bool)
+		} else if ns.Path == "" && mountPath == "" {
+			inheritable = true
+		}
+
 		switch {
 		case quota == nil:
-			quota = quotas.NewRateLimitQuota(name, ns.Path, mountPath, pathSuffix, role, rate, interval, blockInterval)
+			quota = quotas.NewRateLimitQuota(name, ns.Path, mountPath, pathSuffix, role, rate, interval, blockInterval, inheritable)
 		default:
 			// Re-inserting the already indexed object in memdb might cause problems.
 			// So, clone the object. See https://github.com/hashicorp/go-memdb/issues/76.
@@ -365,6 +382,7 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 			rlq.Rate = rate
 			rlq.Interval = interval
 			rlq.BlockInterval = blockInterval
+			rlq.Inheritable = inheritable
 			quota = rlq
 		}
 
@@ -413,6 +431,7 @@ func (b *SystemBackend) handleRateLimitQuotasRead() framework.OperationFunc {
 			"rate":           rlq.Rate,
 			"interval":       int(rlq.Interval.Seconds()),
 			"block_interval": int(rlq.BlockInterval.Seconds()),
+			"inheritable":    rlq.Inheritable,
 		}
 
 		return &logical.Response{
--- a/vault/logical_system_quotas.go
+++ b/vault/logical_system_quotas.go
@@ -301,7 +301,10 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 		}
 
 		mountPath := sanitizePath(d.Get("path").(string))
-		ns := namespace.RootNamespace
+		ns, err := namespace.FromContext(ctx)
+		if err != nil {
+			return logical.ErrorResponse("invalid namespace"), nil
+		}
 		if ns.ID != namespace.RootNamespaceID {
 			mountPath = strings.TrimPrefix(mountPath, ns.Path)
 		}
--- a/vault/mount_test.go
+++ b/vault/mount_test.go
@@ -1153,6 +1153,28 @@ func TestCore_MountEntryView(t *testing.T) {
 			wantError: true,
 		},
 		{
+			name: "entry of 'ns_system' mount type",
+			mountEntry: &MountEntry{
+				UUID:        testMountEntryUUID,
+				Type:        mountTypeNSSystem,
+				NamespaceID: testNamespace1.Namespace.ID,
+				namespace:   testNamespace1.Namespace,
+			},
+
+			wantViewPrefix: namespaceBarrierPrefix + testNamespace1.UUID + "/" + systemBarrierPrefix,
+		},
+		{
+			name: "entry of 'ns_system' mount type for nested namespace",
+			mountEntry: &MountEntry{
+				UUID:        testMountEntryUUID,
+				Type:        mountTypeNSSystem,
+				NamespaceID: testNamespace2.Namespace.ID,
+				namespace:   testNamespace2.Namespace,
+			},
+
+			wantViewPrefix: namespaceBarrierPrefix + testNamespace2.UUID + "/" + systemBarrierPrefix,
+		},
+		{
 			name: "entry of 'system' mount type",
 			mountEntry: &MountEntry{
 				UUID: testMountEntryUUID,
--- a/vault/mount_test.go
+++ b/vault/mount_test.go
@@ -1153,28 +1153,6 @@ func TestCore_MountEntryView(t *testing.T) {
 			wantError: true,
 		},
 		{
-			name: "entry of 'ns_system' mount type",
-			mountEntry: &MountEntry{
-				UUID:        testMountEntryUUID,
-				Type:        mountTypeNSSystem,
-				NamespaceID: testNamespace1.Namespace.ID,
-				namespace:   testNamespace1.Namespace,
-			},
-
-			wantViewPrefix: namespaceBarrierPrefix + testNamespace1.UUID + "/" + systemBarrierPrefix,
-		},
-		{
-			name: "entry of 'ns_system' mount type for nested namespace",
-			mountEntry: &MountEntry{
-				UUID:        testMountEntryUUID,
-				Type:        mountTypeNSSystem,
-				NamespaceID: testNamespace2.Namespace.ID,
-				namespace:   testNamespace2.Namespace,
-			},
-
-			wantViewPrefix: namespaceBarrierPrefix + testNamespace2.UUID + "/" + systemBarrierPrefix,
-		},
-		{
 			name: "entry of 'system' mount type",
 			mountEntry: &MountEntry{
 				UUID: testMountEntryUUID,
--- a/vault/namespaces_oss.go
+++ b/vault/namespaces_oss.go
@@ -31,3 +31,7 @@ func NamespaceView(barrier logical.Storage, ns *namespace.Namespace) BarrierView
 
 	return NewBarrierView(barrier, path.Join(namespaceBarrierPrefix, ns.UUID)+"/")
 }
+
+func (c *Core) NamespaceByPath(ctx context.Context, path string) (*namespace.Namespace, string) {
+	return c.namespaceStore.GetNamespaceByLongestPrefix(ctx, path)
+}
--- a/vault/namespaces_oss.go
+++ b/vault/namespaces_oss.go
@@ -32,6 +32,8 @@ func NamespaceView(barrier logical.Storage, ns *namespace.Namespace) BarrierView
 	return NewBarrierView(barrier, path.Join(namespaceBarrierPrefix, ns.UUID)+"/")
 }
 
+// NamespaceByPath returns the namespace and the path prefix for the given path.
+// Note, that it is on the caller to ensure that the namespace is resolved, as NamespaceByPath otherwise resolves to root.
 func (c *Core) NamespaceByPath(ctx context.Context, path string) (*namespace.Namespace, string) {
 	return c.namespaceStore.GetNamespaceByLongestPrefix(ctx, path)
 }
--- a/vault/namespaces_store.go
+++ b/vault/namespaces_store.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 	"path"
+	"strings"
 	"sync"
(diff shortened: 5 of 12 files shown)


klaus-sap:
When defining a quota on a namespace, should we check that this quota does **not** exceed the quota of the parent namespace?

vault/logical_system_quotas.go
```diff
@@ -301,7 +301,10 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 		}
 
 		mountPath := sanitizePath(d.Get("path").(string))
-		ns := namespace.RootNamespace
+		ns, err := namespace.FromContext(ctx)
+		if err != nil {
+			return logical.ErrorResponse("invalid namespace"), nil
```
wslabosz-reply, line 306: By doing a exact string matching against the repo, there are 6 examples of following error format when we cannot retrieve the namespace from the context, so I'd suggest using something like that
```suggestion
			return logical.ErrorResponse(fmt.Errorf("failed to find namespace in context: %w", err).Error()), nil
```

klaus-sap:
@cipherboy @voigt I wonder if we shouldn't combine horizontal scaling with multiple storage trees. IMHO, one tree should not be distributed over two or more clusters. In that case, it would still be consistent to check the quota of a namespace with its parent namespace. And as the quota is non-secret configuration data, I would not see an issue with the veil of the parent namespace being opaque to a child.

cipherboy:
> IMHO, one tree should not be distributed over two or more clusters.

Note that only one active node (in one cluster) would have (write) ownership of it at a time and would be predicated on having the same storage backend shared by both. Mostly it decouples the plugin logic from the storage/clustering layer a bit. It doesn't change the strict write safety guarantees we have currently.

This is a rather powerful concept, IMHO. Right now we need symmetric scaling: the entire cluster needs to ramp up to support the needs of all users. However, if there are latency-sensitive workloads, there's no way to scale memory usage or compute for particular areas. Creating fully separate clusters complicates the DR recovery plan, so having specific namespaces with specific resource requirements (e.g., a collection of PKI and Transit engines could be very compute heavy with low memory requirements) would allow primary leadership of that namespace be owned by high-CPU nodes (with high quota) and secondary ownership for DR purposes on the main cluster, with very low quota in the event of failure.

Similarly, if one wanted to run a K/V store with a larger storage LRU cache (localized to the workload), this'd require all nodes to have high memory to hit latency requirements when combined with other workloads. Instead, pushing this to a memory-heavy cluster within dedicated NSes and replicating back to the primary cluster for DR purposes would be useful. Performance would take a hit in the event of failure. 

This could be solved with multiple clusters, but would require separating workloads by type, not by tenant. By placing these resource-namespaces within the tenant's base NS, you still get logical tenant grouping in storage/path while allowing the namespace itself to float over the relevant namespace cluster. If the thin parent namespace is mounted in both clusters as well, you get the added benefit of single authentication rather than having separate tokens for separate clusters.

---

But, I don't think this discussion is quite suited for a discussion on quotas outside of some examples where having disparate quotas at some parent level would be useful.

My 2c.

klaus-sap:
@cipherboy We should take care to keep OpenBao simple, because in general complicated concepts come with high operation costs.

In our operation model for multi-tenancy, we want to assign first level namespaces to customers, which then in turn can create child namespaces within their parent namespace. In this model, it would be illogical, if customers could assign themselves higher quotas than they got assigned on first level.

In case of disparate quota requirements, I would vote for using separate namespaces. IMHO, multiple authentication to separate namespaces does not really hurt, does it?

vault/logical_system_quotas.go
```diff
@@ -301,7 +302,10 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 		}
 
 		mountPath := sanitizePath(d.Get("path").(string))
-		ns := namespace.RootNamespace
+		ns, err := namespace.FromContext(ctx)
+		if err != nil {
+			return logical.ErrorResponse(fmt.Errorf("failed to find namespace in context: %w", err).Error()), nil
+		}
```
driif, line 308:
```suggestion
		if err != nil {
			return logical.ErrorResponse("failed to find namespace in context: %v", err), nil
		}
```

driif: We could simplify it here. Also Using %v instead of %w since we're immediately converting to string via Error().

http/util.go
```diff
@@ -44,9 +44,6 @@ var (
 		"sys/loggers",
 		"sys/metrics",
 		"sys/monitor",
-		"sys/quotas/config",
-		"sys/quotas/lease-count",
-		"sys/quotas/rate-limit",
```
driif, line 49: I think this might explain why the Vault API uses paths instead of namespaces.

https://developer.hashicorp.com/vault/tutorials/operations/resource-quotas#vault-enterprise-namespaces

Historically, sys/quotas/... might have been in restrictedAPIs. This meant you couldn't even route a request like POST /v1/ns1/sys/quotas/rate-limit/some-quota to the handler. The request would be rejected upfront by the HTTP layer based purely on the path structure. Removing it from the list allows the request to reach the SystemBackend handler, even when the request has a namespace context.

Seems like Removing sys/quotas/config, sys/quotas/lease-count, and sys/quotas/rate-limit could cause some issues. 
In this case the Safeguard should be ACLs, but if they are not properly configured, than a user within ns1 could have a possibility to change global quotas. 

@cipherboy @voigt @klaus-sap @wslabosz-reply

vault/mount_test.go
```diff
@@ -1152,6 +1152,28 @@ func TestCore_MountEntryView(t *testing.T) {
 
 			wantError: true,
 		},
+		{
+			name: "entry of 'ns_system' mount type",
+			mountEntry: &MountEntry{
```
driif, line 1157: I suggest adding tests to check quota rate limits in namespaces. These could be placed in `vault/logical_system_quotas_test.go` or `vault/external_tests_quotas/quotas_test.go`.

Currently, we have tests for mounting, but we aren't testing rate limits in namespaces yet. It would also be valuable to add tests for checking Global Quota Modification from namespaces.

phyrog:
There are a bunch of ideas here that require a bigger refactoring of the quota manager. As discussed in the WG call today, we'll go with feature parity with Vault for now and create a separate issue or RFC for the enhancements of quotas. That means, for the first namespaces release, quotas will be implemented in the following way:

- the quota api endpoints unter `/sys/quotas` remain restricted and can't be called in namespaces
- quotas are located in storage under `/sys/quotas`
- quotas are applied as per https://github.com/openbao/openbao/blob/279271da7371719c7939fe493aed172f28562f72/vault/quotas/quotas.go#L481-L490
- there is no enforcement or requirement that quotas can only get smaller the more specific the path, which also means child namespaces can get a bigger quota than parent namespaces

Anything that goes beyond this should be part of a new issue or RFC to properly discuss implications of changes.

cipherboy:
Ahh, it looks like our startup function has some unfortunate ordering issues too, related to resolving quotas in `http` layer:

```
2025-04-29T12:24:56.3643354Z WARNING: DATA RACE
2025-04-29T12:24:56.3643738Z Write at 0x00c001b7d068 by goroutine 5891:
2025-04-29T12:24:56.3644628Z   github.com/openbao/openbao/vault.(*Core).setupNamespaceStore()
2025-04-29T12:24:56.3645690Z       /home/runner/work/openbao/openbao/vault/namespaces_store.go:173 +0xbb
2025-04-29T12:24:56.3646597Z   github.com/openbao/openbao/vault.standardUnsealStrategy.unseal()
2025-04-29T12:24:56.3647576Z       /home/runner/work/openbao/openbao/vault/core.go:2269 +0x42b
2025-04-29T12:24:56.3648414Z   github.com/openbao/openbao/vault.(*standardUnsealStrategy).unseal()
2025-04-29T12:24:56.3649135Z       <autogenerated>:1 +0x6e
2025-04-29T12:24:56.3649711Z   github.com/openbao/openbao/vault.(*Core).postUnseal()
2025-04-29T12:24:56.3650587Z       /home/runner/work/openbao/openbao/vault/core.go:2389 +0x4b9
2025-04-29T12:24:56.3651312Z   github.com/openbao/openbao/vault.(*Core).postUnseal()
2025-04-29T12:24:56.3652164Z       /home/runner/work/openbao/openbao/vault/core.go:2385 +0x464
2025-04-29T12:24:56.3652936Z   github.com/openbao/openbao/vault.(*Core).waitForLeadership()
2025-04-29T12:24:56.3654004Z       /home/runner/work/openbao/openbao/vault/ha.go:634 +0x1285
2025-04-29T12:24:56.3654828Z   github.com/openbao/openbao/vault.(*Core).performKeyUpgrades()
2025-04-29T12:24:56.3655775Z       /home/runner/work/openbao/openbao/vault/ha.go:995 +0x21b
2025-04-29T12:24:56.3656576Z   github.com/openbao/openbao/vault.(*Core).performKeyUpgrades()
2025-04-29T12:24:56.3657497Z       /home/runner/work/openbao/openbao/vault/ha.go:991 +0x129
2025-04-29T12:24:56.3658266Z   github.com/openbao/openbao/vault.(*Core).performKeyUpgrades()
2025-04-29T12:24:56.3659157Z       /home/runner/work/openbao/openbao/vault/ha.go:987 +0x5b
2025-04-29T12:24:56.3659919Z   github.com/openbao/openbao/vault.(*Core).waitForLeadership()
2025-04-29T12:24:56.3660801Z       /home/runner/work/openbao/openbao/vault/ha.go:578 +0xc75
2025-04-29T12:24:56.3661556Z   github.com/openbao/openbao/vault.(*Core).runStandby.func9()
2025-04-29T12:24:56.3662439Z       /home/runner/work/openbao/openbao/vault/ha.go:448 +0x54
2025-04-29T12:24:56.3663221Z   github.com/oklog/run.(*Group).Run.func1()
2025-04-29T12:24:56.3664309Z       /home/runner/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:38 +0x39
2025-04-29T12:24:56.3665045Z   github.com/oklog/run.(*Group).Run.gowrap1()
2025-04-29T12:24:56.3665943Z       /home/runner/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:39 +0x4f
2025-04-29T12:24:56.3666547Z 
2025-04-29T12:24:56.3666794Z Previous read at 0x00c001b7d068 by goroutine 5899:
2025-04-29T12:24:56.3667515Z   github.com/openbao/openbao/vault.(*Core).NamespaceByPath()
2025-04-29T12:24:56.3668493Z       /home/runner/work/openbao/openbao/vault/namespaces_oss.go:36 +0x46
2025-04-29T12:24:56.3669440Z   github.com/openbao/openbao/http.handler.rateLimitQuotaWrapping.func49()
2025-04-29T12:24:56.3670606Z       /home/runner/work/openbao/openbao/http/util.go:159 +0xd8
2025-04-29T12:24:56.3671216Z   net/http.HandlerFunc.ServeHTTP()
2025-04-29T12:24:56.3672185Z       /opt/hostedtoolcache/go/1.24.2/x64/src/net/http/server.go:2294 +0x47
2025-04-29T12:24:56.3673023Z   github.com/openbao/openbao/http.wrapGenericHandler.func1()
2025-04-29T12:24:56.3674156Z       /home/runner/work/openbao/openbao/http/handler.go:437 +0x1999
2025-04-29T12:24:56.3674800Z   net/http.HandlerFunc.ServeHTTP()
2025-04-29T12:24:56.3675644Z       /opt/hostedtoolcache/go/1.24.2/x64/src/net/http/server.go:2294 +0x47
2025-04-29T12:24:56.3676607Z   github.com/openbao/openbao/http.handler.wrapMaxRequestSizeHandler.func50()
2025-04-29T12:24:56.3677634Z       /home/runner/work/openbao/openbao/http/util.go:145 +0x1fa
2025-04-29T12:24:56.3678266Z   net/http.HandlerFunc.ServeHTTP()
2025-04-29T12:24:56.3679136Z       /opt/hostedtoolcache/go/1.24.2/x64/src/net/http/server.go:2294 +0x47
2025-04-29T12:24:56.3680096Z   github.com/openbao/openbao/http.handler.PrintablePathCheckHandler.func51()
2025-04-29T12:24:56.3681355Z       /home/runner/go/pkg/mod/github.com/hashicorp/go-cleanhttp@v0.5.2/handlers.go:42 +0xb4
2025-04-29T12:24:56.3682156Z   net/http.HandlerFunc.ServeHTTP()
2025-04-29T12:24:56.3683010Z       /opt/hostedtoolcache/go/1.24.2/x64/src/net/http/server.go:2294 +0x47
2025-04-29T12:24:56.3683695Z   net/http.serverHandler.ServeHTTP()
2025-04-29T12:24:56.3684723Z       /opt/hostedtoolcache/go/1.24.2/x64/src/net/http/server.go:3301 +0x2a1
2025-04-29T12:24:56.3685420Z   net/http.initALPNRequest.ServeHTTP()
2025-04-29T12:24:56.3686298Z       /opt/hostedtoolcache/go/1.24.2/x64/src/net/http/server.go:3974 +0x358
2025-04-29T12:24:56.3686988Z   net/http.(*initALPNRequest).ServeHTTP()
2025-04-29T12:24:56.3687522Z       <autogenerated>:1 +0x7b
2025-04-29T12:24:56.3687980Z   net/http.Handler.ServeHTTP-fm()
2025-04-29T12:24:56.3688486Z       <autogenerated>:1 +0x67
2025-04-29T12:24:56.3688959Z   net/http.(*http2serverConn).runHandler()
2025-04-29T12:24:56.3689865Z       /opt/hostedtoolcache/go/1.24.2/x64/src/net/http/h2_bundle.go:6529 +0x253
2025-04-29T12:24:56.3690654Z   net/http.(*http2serverConn).scheduleHandler.gowrap1()
2025-04-29T12:24:56.3691616Z       /opt/hostedtoolcache/go/1.24.2/x64/src/net/http/h2_bundle.go:6463 +0x5d
```

@driif @satoqz it looks like this would be another vote in favor of moving quota enforcement and namespace resolution into core, once unseal has finished.

satoqz:
@cipherboy ah, that race shouldn't occur in either of #1262 or #1279. The call from `http.rateLimitQuotaWrapping()` into `core.NamespaceByPath()` does not read-lock core's state, so accessing `c.namespaceStore` is bad.

I wonder what the best Git-strategy is here? Optimally we would merge #1279 first and then rebase this on top, if we want to merge this before #1279 however then the below should suffice:

```go
func (c *Core) NamespaceByPath(ctx context.Context, path string) (*namespace.Namespace, string) {
        c.stateLock.Rlock()
        defer c.stateLock.RUnlock()
	return c.namespaceStore.GetNamespaceByLongestPrefix(ctx, path)
}
```

cipherboy:
@satoqz If we're happy with this PR, I'm fine merging it as-is and then fixing the race test in #1279 or a later PR. My 2c.

But would like another review on this before I look at it. :-)

satoqz (approved):
LGTM, let's get this merged!

Edit: Just discussed with @voigt that we'll include the `req.Storage`/`mountEntryView` changes from #1282 in here first.

vault/namespaces_oss.go
```diff
@@ -31,3 +31,7 @@ func NamespaceView(barrier logical.Storage, ns *namespace.Namespace) BarrierView
 
 	return NewBarrierView(barrier, path.Join(namespaceBarrierPrefix, ns.UUID)+"/")
 }
+
+func (c *Core) NamespaceByPath(ctx context.Context, path string) (*namespace.Namespace, string) {
```
wslabosz-reply, line 35: I'd add a comment that this won't necessarily returns the desired namespace, but the one that got matched as the longest prefix of the path, and it's on the caller to ensure that the namespace is resolved before calling this, otherwise this could resolve to the `root`

vault/logical_system_quotas.go
```diff
@@ -428,8 +445,7 @@ func (b *SystemBackend) handleRateLimitQuotasDelete() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
 		name := d.Get("name").(string)
 		qType := quotas.TypeRateLimit.String()
-
-		if err := req.Storage.Delete(ctx, quotas.QuotaStoragePath(qType, name)); err != nil {
+		if err := b.Core.systemBarrierView.Delete(ctx, quotas.QuotaStoragePath(qType, name)); err != nil {
```
wslabosz-reply, line 448: just an open question, as #1206 was merged, are quotas considered when deleting a namespace?

vault/quotas/quotas.go
```diff
@@ -576,18 +579,27 @@ func (m *Manager) queryQuota(txn *memdb.Txn, req *Request) (Quota, error) {
...
+			path = "root"
+		}
+		quota, err = quotaFetchFunc(indexNamespace, path, false, false, false)
+		if err != nil {
+			return nil, err
+		}
+		if quota != nil && quota.IsInheritable() {
+			return quota, nil
+		} else if quota != nil {
+		}
```
wslabosz-reply, line 598: is this empty `else if` block required?

vault/namespaces_oss.go
voigt: I'll add a comment.

vault/quotas/quotas.go
voigt: Its totally not. Removing it!

vault/logical_system_quotas.go
satoqz: In short, yes

wslabosz-reply: @voigt 
investigated with @satoqz and turns out the `(Core).unmountInternal` handles the quota deletion for the ns (by `ns.path`) deleting it both from storage and in-mem so we are fine on this front!

wslabosz-reply (approved):
Looks good to me, the one path that let's you create a quota for a namespace (mainly when you write the quota with ns path specified and not the header) will be accounted for in #1279, so LGTM

vault/logical_system_quotas.go
voigt: Just double checked: quotas are not deleted.
Also I'm not quite sure about the expected behavior here tbh. Should quotas be deleted as well?

vault/logical_system_quotas.go
```diff
@@ -354,9 +360,17 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 			return nil, err
 		}
 
+		inheritable := false
+		inheritableI, ok := d.GetOk("inheritable")
+		if ok {
+			inheritable = inheritableI.(bool)
```
cipherboy, line 366: (nit) For consistency:

```suggestion
		inheritableRaw, ok := d.GetOk("inheritable")
		if ok {
			inheritable = inheritableRaw.(bool)
```

vault/namespaces_store.go
```diff
@@ -444,6 +445,24 @@ func (ns *NamespaceStore) GetNamespaceByAccessor(ctx context.Context, id string)
 	return item.Clone(), nil
 }
 
+func (ns *NamespaceStore) GetNamespaceByLongestPrefix(ctx context.Context, path string) (*namespace.Namespace, string) {
+	ctxNs, err := namespace.FromContext(ctx)
+	if err != nil {
+		ctxNs = namespace.RootNamespace
+	}
+
+	// e.g. needed for /sys/unseal, where the namespace store is not yet initialized
```
cipherboy, line 454: While necessary, this is less than ideal. I wish we could ensure we either have primitive paths (which never exist within a namespace; those routed by `http` directly without calling `core.RouteRequest(...)`) or a fully-initialized store so that the response is always correct (+/- unloaded namespaces). 

Then again, if this request were to a missing namespace, the subsequent route should fail so that's fine :thinking:... 

My 2c.

cipherboy (approved):
I think this looks good as well. I'm surprised but mildly happy that these changes were so small.

cipherboy:
Given my nit is small, I'll go ahead and merge and we can improve in a subsequent follow-up when we review #1165.

cipherboy:
Thank you @voigt and @phyrog!

vault/namespaces_store.go
satoqz: I have plans in that direction :D

Issue #934, opened 2025-01-22
feature request: Add Check-and-Set (CAS) support to policies


fcbajao:
**Is your feature request related to a problem? Please describe.**

Policies are susceptible to lost updates during concurrent modifications. When multiple clients read and modify the same policy simultaneously, later modifications can inadvertently overwrite changes made by earlier ones.

For example:
1. Client A reads policy (contains capabilities X, Y)
2. Client B reads same policy (contains X, Y)
3. Client B adds capability Z, saves policy (now X, Y, Z)
4. Client A removes capability Y, saves policy (now just X)
-> Client B's addition of Z was lost

This is a common issue in systems that manage shared state without concurrency control mechanisms.

**Describe the solution you'd like**

Add Check-and-Set (CAS) support to policy operations, similar to how OpenBao's KV secrets already support CAS. Each policy would track a version number that increments on updates. The `set_policy` operation would accept an optional `cas` parameter.

The update would fail if the current version doesn't match the provided CAS value.

**Describe alternatives you've considered**

1. External locking (e.g., using Redis/PostgreSQL locks) - adds operational complexity and external dependencies
2. Updating policy first in atomic operations - can lead to temporary access issues if subsequent operations fail
3. Background reconciliation jobs - adds complexity and doesn't prevent the underlying issue

**Explain any additional use-cases**

- CI/CD systems managing secrets and their access policies
- Multi-tenant systems where different components modify shared policies
- Automation tools that need to safely add/remove capabilities
- Access control systems that need to ensure policy modifications aren't lost
- Migration tools that need to safely transform policies without losing concurrent updates

**Additional context**

This is particularly important for systems that manage policies programmatically rather than through manual updates. Without CAS, implementing safe concurrent policy modifications requires external coordination mechanisms that add complexity and potential points of failure.

Similar systems like etcd and Consul provide CAS functionality for their key-value stores to handle these scenarios safely.

PR #1162, opened 2025-03-26
Draft: Add check and set for policies

cipherboy:
Based on #1142; will be rebased when that lands.

---

When updating ACL policies in an automated system, it is difficult to
know if a policy has been changed between reading the policy and
subsequently writing it. By using check-and-set semantics similar to
KVv2, we can ensure that callers can update policies and detect if
conflicts have occurred.

We add two parameters, `cas` and `cas_required`, to write operations,
allowing us to restrict sets of policies to CAS-only updates (sans
deletion). `cas` must be set to the previous value of `version` returned by
the read API or `-1` if the entry does not yet exist. CAS is required if
either the current (stored) version or future version requires CAS.

Resolves: #934


wslabosz-reply (approved):
LGTM

wslabosz-reply (approved):
yes, this looks more clean to me, thanks for adjusting!

PR #1163, opened 2025-03-26
Namespaces Regression: Restricted APIs in `sys/raw`

### LINKED ISSUE #1070: Namespaces Regression: Restricted APIs in `sys/raw`
**Describe the bug**

We've made an oversight in the handling of APIs that are restricted to the root namespace that breaks functionality in `sys/raw`:

https://github.com/openbao/openbao/blob/83ecd4f960340aafb32899d2fa74d3adaf0994c8/http/util.go#L78-L83

For example, the `sys/raw/sys/quotas/rate-limit` path breaks as `sys/quotas/rate-limit` is part of `restrictedAPIs` and above logic makes the assumption that `sys/raw` refers to a namespace, thus rejecting the request.

Some additional notes:
- Should `sys/quotas/rate-limit` be restricted in the first place?
- We have `sys/quotas/lease-count` in the restricted APIs - I don't think those exist in OpenBao?
- Has this broken any other APIs besides `sys/raw`?

**To Reproduce**

1. Attempt to read a restricted API path appended to `sys/raw`, from the root namespace:
```sh
bao list sys/raw/sys/quotas/rate-limit
```

...and receive a 400 response

```
Error listing sys/raw/sys/quotas/rate-limit: Error making API request.

URL: GET http://127.0.0.1:8200/v1/sys/raw/sys/quotas/rate-limit?list=true
Code: 400. Errors:
```

**Expected behavior**

Given the above example, the result should either be a 404 response if no rate limits are configured, or a 200 response if rate limits exist. Either way, the request should not be rejected.

**Environment**
- `namespaces` branch

wslabosz-reply:
Resolves #1070  

 - [x] Fixes the restriction on operations (e.g. list, read) on `sys/raw/sys` resources while using root namespace, retaining the restriction on the same operations requested with namespace specification
 - [x] Simplifies logic for preventing serving requests to restricted APIs for namespaces

_Questions from the issue:_
> Should sys/quotas/rate-limit be restricted in the first place?

Depends on the implementation of quotas we want to do within OpenBao (not this PR)
> We have sys/quotas/lease-count in the restricted APIs - I don't think those exist in OpenBao?

Definitely a leftover from [upstream](https://developer.hashicorp.com/vault/api-docs/system/lease-count-quotas) implementation 
> Has this broken any other APIs besides sys/raw?

To be honest it hasn't broken the listing of the `sys/raw/sys` resources, as `bao list sys/quotas/rate-limit` is the equivalent of `bao list sys/raw/sys/quotas/rate-limit` and that worked as always. But to answer the question, yes this broke the usage of `bao list sys/raw/sys...` for restricted APIs

**Before**:
```
bao list sys/raw/sys/quotas/rate-limit
Error listing sys/raw/sys/quotas/rate-limit: Error making API request.

URL: GET http://127.0.0.1:8200/v1/sys/raw/sys/quotas/rate-limit?list=true
Code: 400. Errors:
```

```
bao list -ns=test sys/raw/sys/quotas/rate-limit
Error listing sys/raw/sys/quotas/rate-limit: Error making API request.

URL: GET http://127.0.0.1:8200/v1/sys/raw/sys/quotas/rate-limit?list=true
Code: 400. Errors:
```

And for the policies names (thanks @cipherboy)
```
bao policy write -ns=test sys/raw ./test.hcl 
Error uploading policy: Error making API request.

Namespace: test/
URL: PUT http://127.0.0.1:8200/v1/sys/policies/acl/sys/raw
Code: 400. Errors:
```

**After**:

```
bao list sys/raw/sys/quotas/rate-limit
No value found at sys/raw/sys/quotas/rate-limit

bao write sys/quotas/rate-limit/test rate=5.2
Success! Data written to: sys/quotas/rate-limit/test
bao list sys/raw/sys/quotas/rate-limit
Keys
----
test
```

```
bao list test/sys/raw/sys/quotas/rate-limit
Error listing test/sys/raw/sys/quotas/rate-limit: Error making API request.

URL: GET http://127.0.0.1:8200/v1/test/sys/raw/sys/quotas/rate-limit?list=true
Code: 400. Errors:

* unavailable operation (GET /v1/test/sys/raw/sys/quotas/rate-limit)
...

bao list sys/raw/test/sys/quotas/rate-limit
Error listing sys/raw/test/sys/quotas/rate-limit: Error making API request.

URL: GET http://127.0.0.1:8200/v1/sys/raw/test/sys/quotas/rate-limit?list=true
Code: 400. Errors:

* unavailable operation (GET /v1/sys/raw/test/sys/quotas/rate-limit)
...

bao list /sys/quotas/rate-limit       
No value found at sys/quotas/rate-limit
...

bao list sys/raw/sys/quotas/rate-limit 
No value found at sys/raw/sys/quotas/rate-limit
```

```
bao list -ns=test sys/raw/sys/quotas/rate-limit
Error listing sys/raw/sys/quotas/rate-limit: Error making API request.

Namespace: test/
URL: GET http://127.0.0.1:8200/v1/sys/raw/sys/quotas/rate-limit?list=true
Code: 400. Errors:

* unavailable operation (GET /v1/sys/raw/sys/quotas/rate-limit)
```

```
bao policy write -ns=test sys/raw test.hcl
Success! Uploaded policy: sys/raw
```
--- 

With this PR I also try to address following TODO comment:
```		
// TODO(ascheel): Consider if we can combine header with actual request path here without triggering redirect logic.
```
___

### DIFF
--- a/http/util.go
+++ b/http/util.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
@@ -70,29 +71,40 @@ var (
 		"sys/rekey/",
 	}
 
-	adjustRequest = func(r *http.Request) (*http.Request, int) {
-		u := r.URL
-		p := u.Path[3:]
-
+	validateNamespace = func(r *http.Request) int {
 		ns, err := namespace.FromContext(r.Context())
-		if err != nil || ns == namespace.RootNamespace {
-			return r, 0
+		if err != nil {
+			return http.StatusBadRequest
+		}
+
+		fullURL, err := url.JoinPath(ns.Path, r.URL.Path[4:])
+		if err != nil {
+			return http.StatusBadRequest
 		}
 
 		for _, api := range restrictedAPIs {
-			// exclude any occurences of possible restriced APIs paths when prefix indicates it's a policy name
-			if strings.HasSuffix(p, api) && !strings.HasPrefix(p, "/sys/policies") {
-				return r, http.StatusBadRequest
+			apiSysRawRouted := "sys/raw/" + api
+			// if path not equals the restricted api path or not equals restricted api path with "sys/raw/" prefix
+			namespaceUseOfRestrictedEndpoint := (fullURL != api && fullURL != apiSysRawRouted) && strings.HasSuffix(fullURL, api)
+			namespacePath, _, ok := strings.Cut(fullURL, api)
+
+			// exclude any occurences of possible restricted APIs paths when path dictates that the suffix is a policy name
+			if ok && strings.HasSuffix(namespacePath, "sys/policies/acl/") {
+				return 0
+			}
+
+			if namespaceUseOfRestrictedEndpoint {
+				return http.StatusBadRequest
 			}
 		}
 
 		for _, api := range containsAPIs {
-			if strings.Contains(p, api) {
-				return r, http.StatusBadRequest
+			if strings.Contains(fullURL, api) {
+				return http.StatusBadRequest
 			}
 		}
 
-		return r, 0
+		return 0
 	}
 
 	genericWrapping = func(core *vault.Core, in http.Handler, props *vault.HandlerProperties) http.Handler {
--- a/http/util.go
+++ b/http/util.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
-	"github.com/openbao/openbao/sdk/v2/helper/consts"
 	"github.com/openbao/openbao/sdk/v2/logical"
 
 	"github.com/openbao/openbao/helper/namespace"
@@ -71,28 +70,25 @@ var (
 		"sys/rekey/",
 	}
 
-	adjustRequest = func(c *vault.Core, r *http.Request) (*http.Request, int) {
+	adjustRequest = func(r *http.Request) (*http.Request, int) {
 		u := r.URL
 		p := u.Path[3:]
 
-		nsHeaderPath := namespace.Canonicalize(r.Header.Get(consts.NamespaceHeaderName))
+		ns, err := namespace.FromContext(r.Context())
+		if err != nil || ns == namespace.RootNamespace {
+			return r, 0
+		}
+
 		for _, api := range restrictedAPIs {
-			if strings.HasSuffix(p, api) && (p != "/"+api || nsHeaderPath != "") {
+			if strings.HasSuffix(p, api) {
 				return r, http.StatusBadRequest
 			}
 		}
 
 		for _, api := range containsAPIs {
-			if strings.Contains(p, api) && (!strings.HasPrefix(p, "/"+api) && nsHeaderPath != "") {
+			if strings.Contains(p, api) {
 				return r, http.StatusBadRequest
 			}
-		}
-
-		// TODO(ascheel): Consider if we can combine header with actual request path here without triggering redirect logic.
-		if nsHeaderPath != "" {
-			r = r.WithContext(namespace.ContextWithNamespace(r.Context(), &namespace.Namespace{
-				Path: nsHeaderPath,
-			}))
 		}
 
 		return r, 0
--- a/http/util.go
+++ b/http/util.go
@@ -89,7 +89,7 @@ var (
 			namespacePath, _, ok := strings.Cut(fullURL, api)
 
 			// exclude any occurences of possible restricted APIs paths when path dictates that the suffix is a policy name
-			if ok && strings.HasSuffix(namespacePath, "sys/policies/acl/") {
+			if ok && (strings.HasSuffix(namespacePath, "sys/policies/acl/") || strings.HasSuffix(namespacePath, "sys/policy/")) {
 				return 0
 			}
 
--- a/http/util.go
+++ b/http/util.go
@@ -80,7 +80,8 @@ var (
 		}
 
 		for _, api := range restrictedAPIs {
-			if strings.HasSuffix(p, api) {
+			// exclude any occurences of possible restriced APIs paths when prefix indicates it's a policy name
+			if strings.HasSuffix(p, api) && !strings.HasPrefix(p, "/sys/policies") {
 				return r, http.StatusBadRequest
 			}
 		}
--- a/http/util.go
+++ b/http/util.go
@@ -99,7 +99,7 @@ var (
 		}
 
 		for _, api := range containsAPIs {
-			if strings.Contains(fullURL, api) {
+			if strings.Contains(fullURL, api) && !strings.HasPrefix(fullURL, api) {
 				return http.StatusBadRequest
 			}
 		}
--- a/http/handler.go
+++ b/http/handler.go
@@ -370,13 +370,12 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
 
 		switch {
 		case strings.HasPrefix(r.URL.Path, "/v1/"):
-			newR, status := adjustRequest(r)
+			status := validateNamespace(r)
 			if status != 0 {
-				respondError(nw, status, fmt.Errorf("unavailable operation (%s %s) from namespace [%s]", r.Method, r.URL.Path, ns.Path))
+				respondError(nw, status, fmt.Errorf("unavailable operation (%s %s)", r.Method, r.URL.Path))
 				cancelFunc()
 				return
 			}
-			r = newR
 
 		case strings.HasPrefix(r.URL.Path, "/ui"), r.URL.Path == "/robots.txt", r.URL.Path == "/":
 		case strings.HasPrefix(r.URL.Path, "/.well-known/acme-challenge/"):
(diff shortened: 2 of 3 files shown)


cipherboy:
@wslabosz-reply re:

> as I'm not sure what the "redirect" means in this context

When I attempted to update the path in that function (i.e., handling combining the namespace header with the path and setting that as the new request path -- like is done elsewhere in routing under `vault/` code now), I was getting a literal 302 redirect because the path was detected as changed somewhere in `http/`, which broke clients as many weren't expecting to follow redirects (it is optional).

wslabosz-reply:
okay, that gives me another perspective, I was under impression that listing bao's hierarchical resources by the `list sys/raw` command was a "frequent" operation, but I get the point now. This still technically means that the intended operation would be to use `bao list sys/quotas/rate-limit` which the introduction of the namespaces did not break.

wslabosz-reply:
@cipherboy
what do we gain from injecting the namespace path to url of the request?

cipherboy (commented):
This looks fine, though I'm not sure if you want to include the discussion on Matrix of https://github.com/openbao/openbao/blob/namespaces/vault/request_handling.go#L512-L515.

wslabosz-reply:
@cipherboy 
yes, regarding that, I've been looking at possible adjustments for that TODO (discussion on matrix), and I'm gonna summarize my findings:
1. currently http handler only takes the namespace header and resolves it as the request namespace -> this is incorrect as the path can also contain the namespace path, as we know

```
nsHeader = "team_1"
path: "v1/subteam_a/auth/token/create"
actual namespace: "team_1/subteam_a/"
```
there's a `rateLimitQuotaWrapping` that considers the namespace from the context (in this case only the header part), when calculating the ... well rate limits, so when you make the request for the "team_1/subteam_a" it is actually consider a request to "team_1"

2. because we do not validate whether namespace exists on http level, and we only validate if the namespace stored in context exists on the core handle request level (`(*namespaceStore).ResolveNamespaceFromRequest`) specifying invalid namespace in path like: `v1/non-existing-ns/auth/token/create` will result in `{"errors":["no handler for route \"non-existing-ns/auth/token/create\". route entry not found."]}` coming from internal bao core router. This might be intended, or not...

3. RE: whether we can combine the header with the request path to have the actual request with full namespace specification
 a) so initially I thought we could remove whole "storing ns in ctx" on http level and just pass along the url with the ns header value inserted into that. This makes sense to me (http handler doesn't need to know about the internal logic of the namespaces), but it's actually hard to implement on the core level, as we don't know where does the namespace path end in the url path. This is currently solved using the `LongestPrefix` func that causes the issue number 2. Technically there's a way to resolve the namespace from any arbitrary url, but will require some string manipulation

b) after talking with @voigt I've tried different approach, not passing the url, but actually resolving the full namespace on the http handler, which then would enable core to just copy the namespace from the context. This also simplifies things (fixes point 1 & 2), but I had my reservations because in the end it's the http level so we shouldn't really uncover the internal implementation of the core, but we already kinda do, as the `*Core` is accessible by the handler wrappers. The issue is, we cannot resolve the namespace (& validate it exists) as `*Core` doesn't have a method like that, we'd need `*namespaceStore` which is private (and should be), I'd reckon we could add a "NamespaceByPath" method to (*Core), as there are already "NamespaceByID" and "ListNamespaces" (`namespace_oss.go`) methods declared.

curious what you think @cipherboy

phyrog:
> but it's actually hard to implement on the core level, as we don't know where does the namespace path end in the url path.

I may be misunderstanding, but the namespace part from the header is **prepended** to any namespace in the url path, so the part from the header would just go after /v1/ (if that is still there at this point during handling) and then the rest from the url path.

> I'd reckon we could add a "NamespaceByPath" method to (*Core)

Also sounds reasonable, but it does mean we move some (more) of the domain logic into the http layer before routing; though it sounds like some of the sub-systems could benefit from that (e.g. the quota system).

wslabosz-reply:
@phyrog 
> I may be misunderstanding, but the namespace part from the header is prepended to any namespace in the url path, so the part from the header would just go after /v1/ (if that is still there at this point during handling) and then the rest from the url path.

yes, we do know where does the namespace path specification start (after `/v1/`), but we do not know where does it end. 
- we can do a match on the `LongestPrefix` but it will resolve on the longest path of the known namespace, meaning you could have specified something that doesn't exist, and this will resolve to `root` or other parent of the specified namespace
- to do exact namespace path specification we would need to determine the full namespace path from the url, so we'd have to maintain all possible endpoints and rip out the segments after `/v1/` and before the endpoint path.

cipherboy:
> Also sounds reasonable, but it does mean we move some (more) of the domain logic into the http layer before routing; though it sounds like some of the sub-systems could benefit from that (e.g. the quota system).

I agree that most of the namespace logic should live in Core and not http. 

---

I guess, I've lost sight of the reason for this change. I think just fixing `sys/raw` and `sys/quotas` and stuff, I assumed, was a much smaller change just limited to `adjustRequest(...)`?

wslabosz-reply:
> I guess, I've lost sight of the reason for this change. I think just fixing sys/raw and sys/quotas and stuff (including policies), I assumed, was a much smaller change just limited to adjustRequest(...)?

yes, I'm doing just that in this PR, but I wanted to touch upon the TODO that I've removed and other things I've found while working on it.

> FWIW, the top commit fails most tests

looking into it, will fix.

wslabosz-reply:
@cipherboy 
okay, I don't think that the rest of the failing tests are caused by the changes in this PR

driif:
Once the major tickets in namespaces are completed, the next step should be refactoring the existing code in namespaces. This could be a proposal for that refactoring.

I agree that moving the resolution of the namespace path for earlier execution is a valid point, but not resolving the full path for `delete/create` endpoints. Splitting the *FULL PARENT*->*CTX* and the *CHILD* path can offer advantages by handling them separately, reducing friction with strings in the code.

--- 
LGTM

driif (approved):

cipherboy:
@wslabosz-reply @driif I think fundamentally, moving namespaces into `http` only works _iff_ you can guarantee no storage reads. Storage is one of the portions that is very clearly on the side of Core, IMHO. The HTTP layer itself should never incur storage operations (I hope).

In particular, (while admittedly horizontal scalability WG isn't this far yet), consider what happens on a (future, read-enabled) standby node when accepting a request with a namespace not yet known.

We'd have gotten an invalidation hook trigger (which is simply a notification of a path in storage which has changed and _cannot_ perform any storage operations in the invalidation handler itself) for this new namespace. Assume we're called early enough that no other request to this standby node has occurred and no rollback/periodic func has executed.

There are two options:

1. Fail the request, because we don't yet know about this namespace (in-memory). Somewhere in the future will have to deal with the invalidation, but for now this isn't our problem.
2. Treat the invalidation as storage refresh point, talk to underlying storage, and figure out that this is a legitimate request.

IMHO, if we do 2, this should strictly be on the side of Core.

From my Pov, I think this is a (relatively) common scenario: when using a RR LB (DNS or otherwise), namespace creation happens on the active node as will initial subsequent namespace use (likely mount creation or policy creation). However, these requests may come into a read-enabled standby node; if a non-read-only error is generated early (e.g., namespace does not exist), the request will not be forwarded to the active node.

---

All this to say, we probably need another ticket about adding invalidation hooks to NamespaceStore, at minimum.

http/util.go
```diff
@@ -71,31 +71,40 @@ var (
...
 		for _, api := range restrictedAPIs {
-			if strings.HasSuffix(p, api) && (p != "/"+api || nsHeaderPath != "") {
-				return r, http.StatusBadRequest
+			apiSysRawRouted := "sys/raw/" + api
+			// if path not equals the restricted api path or not equals restricted api path with "sys/raw/" prefix
+			namespaceUseOfRestrictedEndpoint := (fullURL != api && fullURL != apiSysRawRouted) && strings.HasSuffix(fullURL, api)
+			namespacePath, _, ok := strings.Cut(fullURL, api)
+
+			// exclude any occurences of possible restricted APIs paths when path dictates that the suffix is a policy name
+			if ok && strings.HasSuffix(namespacePath, "sys/policies/acl/") {
```
cipherboy, line 92:
```suggestion
			if ok && (strings.HasSuffix(namespacePath, "sys/policies/acl/") || strings.HasSuffix(namespacePath, "sys/policy/")) {
```

As there are two policy paths (latter is legacy, prior to introduction of EGP in Vault Enterprise).

cipherboy: Really makes me think, if someone wants to take it on, we do something like the SSH or PKI backend tests and interrogate OpenAPI and make sure that all paths that work in the root namespace work in a namespace or have an explicit exclusion as to why it shouldn't.

cipherboy (approved):
@wslabosz-reply Sorry I missed that earlier! One change and then I'll go ahead and merge.

wslabosz-reply:
@cipherboy 
commited your suggested change. If we merge #1201 first, then we'd have reinforced test coverage for this PR overall. Let's maybe do #1201 first, then this one?

driif:
@wslabosz-reply I'm sorry, but it might take me some time to figure out ther real reason for the test failure. I appreciate your effort to help identify the cause of failure in #1201, but it seems to be due to another issue.

http/util.go
wslabosz-reply: > As there are two policy paths (latter is legacy, prior to introduction of EGP in Vault Enterprise).
adjusted

> Really makes me think, if someone wants to take it on, we do something like the SSH or PKI backend tests and interrogate OpenAPI and make sure that all paths that work in the root namespace work in a namespace or have an explicit exclusion as to why it shouldn't.
Agree, still for future.

cipherboy:
test-go (5) is fixed by #1221, merging. Thanks @wslabosz-reply!

PR #1165, opened 2025-03-26
Merge Namespaces to Main

cipherboy:
WIP; do not merge.

Resolves: #787

---

Currently I'm checking mergability but also as a reminder that we'll want to avoid merging conflicting changes (especially #1142)  as long as we can.


vault/core.go
```diff
@@ -2472,6 +2497,9 @@ func (c *Core) preSeal() error {
 	if err := c.unloadMounts(context.Background()); err != nil {
 		result = multierror.Append(result, fmt.Errorf("error unloading mounts: %w", err))
 	}
+	if err := c.teardownNamespaceStore(); err != nil {
```
cipherboy, line 2500: Does LoginMFA shutdown need to occur before NamespaceStore teardown? Can it reference namespaces?

@schultz-is thinks it might be able to.

vault/expiration.go
```diff
@@ -453,6 +446,41 @@ func (c *Core) stopExpiration() error {
...
+}
+
+func (m *ExpirationManager) tokenIndexView(ns *namespace.Namespace) BarrierView {
+	if ns.ID == namespace.RootNamespaceID {
+		return m.core.systemBarrierView.SubView(expirationSubPath + tokenViewPrefix)
+	}
+	return m.core.namespaceMountEntryView(ns, systemBarrierPrefix+expirationSubPath+tokenViewPrefix)
+}
+
+func (m *ExpirationManager) collectLeases() (map[*namespace.Namespace][]string, int, error) {
```
cipherboy, line 463: @voigt Do you want to file a follow-up issue to make this work with a callback? Then after this merges and #1102 merges, we can update this to not load all leases into memory :-) 

That will need to be a separate RFC to improve scalability of this particular subsystem.

vault/namespaces_store.go
```diff
@@ -0,0 +1,757 @@
...
+// a singleton store across the Core and contains all child namespaces.
+type NamespaceStore struct {
+	core    *Core
+	storage logical.Storage
+
+	// This lock ensures we don't concurrently modify the store while using
+	// a namespace entry. We also store an atomic to check if we need to
+	// reload all namespaces.
+	lock        sync.RWMutex
+	invalidated atomic.Bool
```
cipherboy, line 53: @voigt Can you file a ticket about this as well? When horizontal scalability is added, we'll need to hook path-level invalidations here in namespace store and reload. Thanks!

vault/namespaces_store.go
```diff
@@ -0,0 +1,757 @@
...
+	// invalidation this will pre-allocate enough space to reload everything
+	// as we'll likely be essentially in sync already. However, at startup, this
+	// will mostly just give us space for the root namespace.
+	namespacesByPath := newNamespaceTree(namespace.RootNamespace)
+	namespacesByUUID := make(map[string]*namespace.Namespace, len(ns.namespacesByUUID)+1)
+	namespacesByAccessor := make(map[string]*namespace.Namespace, len(ns.namespacesByAccessor)+1)
+	namespacesByUUID[namespace.RootNamespaceUUID] = namespace.RootNamespace
+	namespacesByAccessor[namespace.RootNamespaceID] = namespace.RootNamespace
+
+	if err := logical.WithTransaction(ctx, ns.storage, func(s logical.Storage) error {
```
cipherboy, line 122: I think we should actually split this so child namespaces' storage references are in the parent, not root. We need another recursion here, but that way if we have an external import of a namespace ([see vision](https://openbao.org/blog/vision-for-namespaces/)), you get children for free without having to explicitly register it. My 2c.

cipherboy: I think this is necessary for out-of-hierarchy namespaces (parent-less namespaces), based on conversation with @schultz-is.

sdk/helper/pathmanager/pathmanager.go
```diff
@@ -137,3 +137,28 @@ func (p *PathManager) HasExactPath(path string) bool {
...
+			exception = exceptionRaw.(bool)
+		}
+
+		strVal := string(val)
+		if path == strVal || // exact match, down to the trailing '/'
+			// prefix match is continued by a new path segment, where `strVal` has no trailing '/'
+			path[len(strVal)] == '/' ||
+			// prefix match is continued by a new path segment, where `strVal` has an trailing '/'
+			// -- in this case we enforce the trailing '/' in `path` as well.
+			(path[len(strVal)-1] == '/' && strVal[len(strVal)-1] == '/') {
```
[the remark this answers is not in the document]
cipherboy: By definition of `val := LongestPrefix([]byte(path))`, this is invalid unless the value returned by `LongestPrefix` results in a non-prefix value.

JanMa (approved):
Not 100% done with the review yet but things are looking good so far.
I'm fine with merging this now to keep our intended release timeline

http/handler.go
```diff
@@ -277,7 +299,8 @@ func handleAuditNonLogical(core *vault.Core, h http.Handler) http.Handler {
 		input := &logical.LogInput{
 			Request: req,
 		}
-		err = core.AuditLogger().AuditRequest(r.Context(), input)
+		ctx := namespace.RootContext(r.Context())
```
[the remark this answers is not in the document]
cipherboy: I realized I was looking at the wrong function. This should be fine as all specially handled `sys/` paths only occur in the root namespace.

vault/core.go
cipherboy: LoginMFA is handled by #1336.

PR #1167, opened 2025-03-27
move namespace uuid into namespace struct

### LINKED ISSUE #1166: Move UUID field from NamespaceEntry to Namespace
**Describe the bug**
We have many cases of requiring the namespace UUID to construct storage paths. Since the UUID is not part of the `Namespace` struct stored in context, this requires additional lookups of the `NamespaceEntry` from the namespace store. To simplify this and decouple otherwise unrelated code from the namespace store, we should move the UUID into the `Namespace` struct.

**Examples**
- https://github.com/openbao/openbao/blob/a3edd3119ed66b7ad309fed983ab664e7627789b/vault/mount.go#L1612
- https://github.com/openbao/openbao/blob/a3edd3119ed66b7ad309fed983ab664e7627789b/vault/mount.go#L2195
  - and places where this is transitively used
- https://github.com/openbao/openbao/blob/a3edd3119ed66b7ad309fed983ab664e7627789b/vault/policy_store.go#L371
- other examples e.g. in #1123 

**Expected behavior**
UUID is contained in the `Namespace` struct, so we can avoid unnecessary redundant lookups in the namespace store.

phyrog:
Resolves #1166.

- Moves UUID field from `NamespaceEntry` into `Namespace`
- Removes unnecessary fetches from NamespaceStore
- Use `(*Core).ListNamespaces` and `(*Core).NamespaceByID` where now possible
- Removes `NamespaceEntry` completely

cc: @cipherboy, @driif, @voigt, @wslabosz-reply

### DIFF
--- a/helper/namespace/namespace.go
+++ b/helper/namespace/namespace.go
@@ -29,6 +29,7 @@ type contextValues struct{}
 
 type Namespace struct {
 	ID             string            `json:"id" mapstructure:"id"`
+	UUID           string            `json:"uuid" mapstructure:"uuid"`
 	Path           string            `json:"path" mapstructure:"path"`
 	CustomMetadata map[string]string `json:"custom_metadata" mapstructure:"custom_metadata"`
 }
@@ -64,7 +65,8 @@ func (n *Namespace) Validate() error {
 }
 
 const (
-	RootNamespaceID = "root"
+	RootNamespaceID   = "root"
+	RootNamespaceUUID = "00000000-0000-0000-0000-000000000000"
 )
 
 var (
@@ -72,6 +74,7 @@ var (
 	ErrNoNamespace   error         = errors.New("no namespace")
 	RootNamespace    *Namespace    = &Namespace{
 		ID:             RootNamespaceID,
+		UUID:           RootNamespaceUUID,
 		Path:           "",
 		CustomMetadata: make(map[string]string),
 	}
--- a/helper/namespace/namespace.go
+++ b/helper/namespace/namespace.go
@@ -111,6 +111,20 @@ func (n *Namespace) TrimmedPath(path string) string {
 	return strings.TrimPrefix(path, n.Path)
 }
 
+func (n *Namespace) Clone() *Namespace {
+	meta := make(map[string]string, len(n.CustomMetadata))
+	for k, v := range n.CustomMetadata {
+		meta[k] = v
+	}
+
+	return &Namespace{
+		ID:             n.ID,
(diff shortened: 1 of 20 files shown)


cipherboy:
> Removes `NamespaceEntry` completely (do we want to keep this?)

I'm happy removing it.

helper/namespace/namespace.go
```diff
@@ -64,14 +65,16 @@ func (n *Namespace) Validate() error {
 }
 
 const (
-	RootNamespaceID = "root"
+	RootNamespaceID   = "root"
+	RootNamespaceUUID = "00000000-0000-0000-0000-000000000000"
```
cipherboy, line 69: FWIW, `RootNamespaceID` doesn't conform to ID requirements (Vault's were 5 characters base 62, we're at 6 ourselves anyways).

cipherboy: Ah, sorry, the latter.

vault/mount.go
```diff
@@ -1609,17 +1609,7 @@ func (c *Core) persistMounts(ctx context.Context, barrier logical.Storage, table
 					continue
 				}
 
-				ne, err := c.namespaceStore.GetNamespaceByAccessor(ctx, mtEntry.NamespaceID)
```
cipherboy, line 1612: This will be nice :-)

vault/auth.go
```diff
@@ -564,7 +564,7 @@ func (c *Core) loadCredentials(ctx context.Context) error {
 
 // This function reads the transactional split auth (credential) table.
 func (c *Core) loadTransactionalCredentials(ctx context.Context, barrier logical.Storage) error {
-	allNamespaces, err := c.namespaceStore.ListAllNamespaceEntries(ctx, true)
+	allNamespaces, err := c.ListNamespaces(ctx)
```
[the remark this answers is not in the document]
phyrog: Are these safeguards needed to get this PR merged, or are we fine with doing this later?

wslabosz-reply (approved):

vault/auth.go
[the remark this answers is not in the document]
phyrog: @cipherboy Alright, then we'd be done here from my perspective and ready for review.

cipherboy (approved):
This looks slick, many thanks @phyrog!

helper/namespace/namespace.go
cipherboy: I left this as-is for now @phyrog but we can revisit this later if we want.

PR #1178, opened 2025-03-31
Another attempt to fix TestOIDC_PeriodicFunc

### LINKED ISSUE #1177: OIDC PeriodicFunc test is flakey
**Describe the bug**
This the same as issue #1031, which I thought was fixed because it is no longer reproducible on my workstation. It is still appearing in CI data race condition tests, eg: https://github.com/openbao/openbao/actions/runs/14116297347/job/39547273769?pr=989. It seems like when the test calls the 'oidcPeriodicFunc', the TTL of the key has sometimes not yet expired so it's not cycled.

```
2025-03-27T20:52:47.5556480Z === [31mFAIL[0m: vault TestOIDC_PeriodicFunc/test-key-nil-next-signing-key (6.80s)
2025-03-27T20:52:47.5557097Z 2025-03-27T20:50:04.811Z [INFO]  TestOIDC_PeriodicFunc/test-key-nil-next-signing-key: creating error injector
2025-03-27T20:52:47.5558180Z     identity_store_oidc_test.go:1158: For key: test-key-nil-next-signing-key at cycle: 2 expected namedKey's KeyRing to be at least of length 3 but was: 2
2025-03-27T20:52:47.5559537Z     identity_store_oidc_test.go:1169: For key: test-key-nil-next-signing-key at cycle: 2 expected public keys to be at least of length 3 but was: 2
```

**To Reproduce**
1. Submit a PR and it may fail, may not.
3. See error in log

**Expected behavior**
The test should pass in CI as it passes on workstation.

**Environment:**
CI

Data race test setup

**Additional context**
None

suprjinx:
Added additional sleep buffer between call to `oidcPeriodicFunc` and verification logic.

Resolves #1177


suprjinx:
@cipherboy I think this tweak is looking pretty good on CI (2 successes in a row)

JanMa (approved):
Looks like this finally passes :tada: Thanks @suprjinx

cipherboy (approved):
Many thanks :-) Test only changes still don't need changelog entries, but that's fine.

vault/identity_store_oidc_test.go
```diff
@@ -1139,6 +1139,8 @@ func TestOIDC_PeriodicFunc(t *testing.T) {
 				// sleep for a rotation period
 				time.Sleep(testSet.namedKey.RotationPeriod)
 				c.identityStore.oidcPeriodicFunc(ctx)
+				// sleep a moment before collecting key info
+				time.Sleep(50 * time.Millisecond)
```
cipherboy, line 1143: Hmm, this actually failed again:

```
2025-04-02T14:26:20.9867977Z     identity_store_oidc_test.go:1160: For key: test-key-nil-next-signing-key at cycle: 2 expected namedKey's KeyRing to be at least of length 3 but was: 2
2025-04-02T14:26:20.9869840Z     identity_store_oidc_test.go:1171: For key: test-key-nil-next-signing-key at cycle: 2 expected public keys to be at least of length 3 but was: 2
```

suprjinx: D'oh!

cipherboy:
> I'm not allowed to edit the labels!

Ah, I'm happy to add them later, the PR bot is a little over eager in that regard. :-)

vault/identity_store_oidc_test.go
suprjinx: @cipherboy I simplified the test so there's just one loop, and added a bit more buffer time (100ms). it seems to be mostly green (4/5 runs) -- but now there's a vulncheck failure.

driif: @suprjinx thank you for fixing. 

https://pkg.go.dev/vuln/GO-2025-3563

Vulncheck failure message suggests moving to the fixed version go1.24.2
https://groups.google.com/g/golang-announce/c/Y2uBTVKjBQk/m/cs_6qIK5BAAJ?pli=1

--- 
@cipherboy @JanMa

cipherboy: Vulncheck can be ignored for PRs (that don't introduce new dependencies), it is more informational.

I was hoping dependabot would've updated these since they've been out for a bit, but I guess we'll need manual PR for the JWT change anyways.

And yes, go toolchain bump will need to happen manually.

cipherboy (approved):
I think this feels a lot more substantial to address the timing issue. Thank you for your patience, @suprjinx!

PR #1179, opened 2025-03-31
Do not set LimitNOFILE

arianvp:
This is not needed anymore since https://github.com/golang/go/issues/46279 was resolved.   golang automatically raises the soft limit on startup.

Quoting https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Process%20Properties

> **Don't use.** Be careful when raising the soft limit above 1024, since select(2) cannot function with file descriptors above 1023 on Linux. Nowadays, the hard limit defaults to 524288, a very high value compared to historical defaults. Typically applications should increase their soft limit to the hard limit on their own, if they are OK with working with file descriptors above 1023, i.e. do not use select(2). Note that file descriptors are nowadays accounted like any other form of memory, thus there should not be any need to lower the hard limit. Use MemoryMax= to control overall service memory use, including file descriptor memory.

Resolves #

### DIFF
--- a/.release/linux/package/usr/lib/systemd/system/openbao.service
+++ b/.release/linux/package/usr/lib/systemd/system/openbao.service
@@ -26,7 +26,6 @@ KillSignal=SIGINT
 Restart=on-failure
 RestartSec=5
 TimeoutStopSec=30
-LimitNOFILE=65536
 MemorySwapMax=0
 
 [Install]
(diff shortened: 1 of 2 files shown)


cipherboy (approved):
Thanks @arianvp! Do you mind adding a changelog entry for this one?

https://github.com/openbao/openbao/blob/main/changelog/README.md

cipherboy:
:( I broke CI, but it was previously OK and I only added a changelog entry, so going ahead and merging.

PR #1186, opened 2025-04-03
Add action to mirror repo to Codeberg

JanMa:
We decided it might be a good idea to provide an official repository mirror other than GitHub. This adds an action to mirror the main branch to Codeberg.

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -1,0 +1,22 @@
+name: Mirror Repo
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["main"]
+
+jobs:
+  mirror:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: yesolutions/mirror-action@v0.7.0
+        with:
+          REMOTE: "git@codeberg.org:openbao/openbao.git"
+          GIT_SSH_PRIVATE_KEY: ${{ secrets.GIT_SSH_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror.yml
```diff
@@ -0,0 +1,22 @@
...
+  mirror:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: yesolutions/mirror-action@v0.7.0
+        with:
+          REMOTE: "git@codeberg.org:openbao/openbao.git"
+          GIT_SSH_PRIVATE_KEY: ${{ secrets.GIT_SSH_PRIVATE_KEY }}
```
cipherboy, line 18:
```suggestion
          GIT_SSH_PRIVATE_KEY: ${{ secrets.CODEBERG_GIT_SSH_PRIVATE_KEY }}
```

cipherboy: Let's use separate keys for separate places.

cipherboy (approved):

.github/workflows/mirror.yml
```diff
@@ -0,0 +1,22 @@
...
+    branches: ["main"]
+
+jobs:
+  mirror:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: yesolutions/mirror-action@v0.7.0
```
cipherboy, line 15: In the future, we could probably simplify this down to a shell script, though I think later @karras will be suggesting policies around third-party GHAction usage.

PR #1188, opened 2025-04-04
Replace Vault with OpenBao in init error msg

Nerkho:
I just noticed while doing some tests that the error msg when initializing fails still contains `Vault`

### DIFF
--- a/vault/barrier.go
+++ b/vault/barrier.go
@@ -15,15 +15,15 @@ import (
 var (
 	// ErrBarrierSealed is returned if an operation is performed on
 	// a sealed barrier. No operation is expected to succeed before unsealing
-	ErrBarrierSealed = errors.New("Vault is sealed")
+	ErrBarrierSealed = errors.New("OpenBao is sealed")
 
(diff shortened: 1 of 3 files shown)


cipherboy:
Hey @Nerkho -- I think we decided these specific error messages were constants. Some third-party code relies on them verbatim, e.g.: https://github.com/OneOaaS/vault/blob/0645606f8489f2cb208dadd9f0a84b1e015b93a3/command/status.go#L12 

This is rather unfortunate and means we likely won't be able to update these ever.

Nerkho:
> Hey @Nerkho -- I think we decided these specific error messages were constants. Some third-party code relies on them verbatim, e.g.: https://github.com/OneOaaS/vault/blob/0645606f8489f2cb208dadd9f0a84b1e015b93a3/command/status.go#L12
> 
> This is rather unfortunate and means we likely won't be able to update these ever.

Aaaah I see okay. Thanks for the explanation @cipherboy 

Too bad :(

PR #1190, opened 2025-04-05
Add guide for listing detailed endpoints

fatima2003:
Lists benefits of implementing `/detailed` endpoint vs using `- detailed` flag.


cipherboy (approved):
Thanks!

PR #1191, opened 2025-04-05
fix mirror workflow

JanMa:
The previously used action turned out to be broken, so provide a simple
alternative workflow.


cipherboy (approved):
This looks better, thanks @JanMa!

PR #1199, opened 2025-04-07
Add cert details to cert list in UI

fatima2003:
The cert list in the UI currently only shows the serial number. It now also shows the common name if it exists.

image

Pairs with #680
Resolves [Vault 27249](https://github.com/hashicorp/vault/issues/27249)

### DIFF
--- a/ui/app/adapters/pki/certificate/base.js
+++ b/ui/app/adapters/pki/certificate/base.js
@@ -11,7 +11,7 @@ export default class PkiCertificateBaseAdapter extends ApplicationAdapter {
 
   getURL(backend, id) {
     const uri = `${this.buildURL()}/${encodePath(backend)}`;
-    return id ? `${uri}/cert/${id}` : `${uri}/certs`;
+    return id ? `${uri}/cert/${id}` : `${uri}/certs/detailed`;
   }
 
   fetchByQuery(query) {
--- a/ui/app/serializers/pki/certificate/base.js
+++ b/ui/app/serializers/pki/certificate/base.js
@@ -27,4 +27,22 @@ export default class PkiCertificateBaseSerializer extends ApplicationSerializer 
     }
     return super.normalizeResponse(...arguments);
   }
+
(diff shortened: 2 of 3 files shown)


ui/app/adapters/pki/certificate/base.js
```diff
@@ -11,7 +11,7 @@ export default class PkiCertificateBaseAdapter extends ApplicationAdapter {
 
   getURL(backend, id) {
     const uri = `${this.buildURL()}/${encodePath(backend)}`;
-    return id ? `${uri}/cert/${id}` : `${uri}/certs`;
+    return id ? `${uri}/cert/${id}` : `${uri}/certs/detailed`;
```
fatima2003, line 14: hmm I'm not sure about this... would this mean the users need access to the `certs/detailed` path as well?

fatima2003: I also think the fallback to `certs/` is needed incase the certs were created before the `detailed` endpoint was introduced... or does that not matter

cipherboy: Ah, no, the detailed list response is generated on the fly so it shouldn't matter when a cert was created.

fatima2003:
@cipherboy it works and I managed to display the TTL as well :D 

1. When a user has permissions to the `certs/detailed/` endpoint:
image

2. When a user does not have permissions to the `certs/detailed/` endpoint:
image
I created a policy like follows to demonstrate behavior without `certs\detailed\` permissions: 
```
# Allow reading certificate list
path "pki/certs" {
  capabilities = ["read", "list"]
}

# Deny access to detailed certificate list
path "pki/certs/detailed" {
  capabilities = []
}
```

ui/app/adapters/pki/certificate/base.js
fatima2003: Oh, perfect!

cipherboy:
@fatima2003 Do you mind rebasing this? :-)

Issue #1198, opened 2025-04-07
Strange behavior with namespace shadowing existing mount


cipherboy:
**Describe the bug**

Namespace (#787) shadowing mount causes strange behavior.

**To Reproduce**

1. Run `bao secrets enable kv-v2`
2. Run `bao kv put -mount=kv-v2 data a=c`
3. Run `bao read kv-v2/data/data` (exists)
4. Create namespace `bao namespace create kv-v2` (this fails, correctly).
    ```
   Error creating namespace: Error making API request.
    
    URL: PUT http://127.0.0.1:8200/v1/sys/namespaces/kv-v2
    Code: 500. Errors:
    
    * 1 error occurred:
    	* failed to modify namespace: existing mount at kv-v2/
    ```
5. Run `bao namespace list` (`kv-v2` namespace exists!)
    ```
    kv-v2/
    ```
6. Run `bao read kv-v2/data/data` (this works)
7. Run `bao namespace create kv-v2` again (this shows metadata for the mount).
    ```
    Key                Value
    ---                -----
    custom_metadata    map[]
    id                 sdx7e3
    path               kv-v2/
    uuid               80f8db9e-08d0-4f93-ab6d-3a3a130b5253
    ```
8. Run `bao read kv-v2/data/data` (this works)
9. Run `bao secrets enable -namespace=kv-v2 pki` (this 404s, implying namespace doesn't work).

**Expected behavior**

Namespace should not be able to take the name of mount and visa-versa.

As of `4413a4cf451c86ff6d8c270630fa6573ffd8b88c`.

cc: @phyrog @voigt @driif @wslabosz-reply

wslabosz-reply:
@cipherboy 
does this issue cause failures in test like: https://github.com/openbao/openbao/actions/runs/14315484244/job/40120588263?pr=1163
because of some cross-tests dependencies?

cipherboy:
Right, I think it's a general bug that needs to be fixed. I'll try and get to it soon.

PR #1202, opened 2025-04-08
Fix namespace, mount name exclusion logic

cipherboy:
When a namespace is created that conflicts with a mount (or visa-versa), ensure that the creation fails and does not leave residuals lying around.

Resolves: #1198

--

cc: @driif @klaus-sap @satoqz @phyrog @voigt @wslabosz-reply @genelet

### DIFF
--- a/vault/mount_test.go
+++ b/vault/mount_test.go
@@ -1307,3 +1307,121 @@ func TestCore_MountEntryView(t *testing.T) {
 		})
 	}
 }
+
+func TestNamespaceMount_Exclusion(t *testing.T) {
+	c, _, _ := TestCoreUnsealed(t)
+	c.logicalBackends["noop"] = func(ctx context.Context, config *logical.BackendConfig) (logical.Backend, error) {
+		return &NoopBackend{}, nil
+	}
+
+	// Creating a mount and then a namespace with the same name should fail.
+	me := &MountEntry{
+		Table:       mountTableType,
+		Path:        "foo/",
+		Type:        "noop",
+		NamespaceID: namespace.RootNamespaceID,
+	}
+	err := c.mount(namespace.RootContext(nil), me)
+	require.NoError(t, err)
+
+	ns, err := c.namespaceStore.ModifyNamespaceByPath(namespace.RootContext(nil), "foo/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
+
+	// In openbao#1198, the first creation erred but left a ghost namespace
+	// object lying around. This meant that list and subsequent create
+	// namespace operations returned this ghost structure and did not error
+	// properly. Retrying the create ensures no ghost object exists.
+	ns, err = c.namespaceStore.ModifyNamespaceByPath(namespace.RootContext(nil), "foo/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
+
+	// Creating a deeply nested mount should also cause failures.
+	me = &MountEntry{
+		Table:       mountTableType,
+		Path:        "fud/bar/baz/foo/",
+		Type:        "noop",
+		NamespaceID: namespace.RootNamespaceID,
+	}
+	err = c.mount(namespace.RootContext(nil), me)
+	require.NoError(t, err)
+
+	ns, err = c.namespaceStore.ModifyNamespaceByPath(namespace.RootContext(nil), "fud/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
+
+	ns, err = c.namespaceStore.ModifyNamespaceByPath(namespace.RootContext(nil), "fud/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
+
+	// Creating a new namespace should succeed.
+	nsBar, err := c.namespaceStore.ModifyNamespaceByPath(namespace.RootContext(nil), "bar/", nil)
+	require.NoError(t, err)
+	require.NotNil(t, nsBar)
+
+	barCtx := namespace.ContextWithNamespace(context.Background(), nsBar.Namespace)
+
+	// Doing the above inside bar should behave the same.
+	me = &MountEntry{
+		Table:       mountTableType,
+		Path:        "foo/",
+		Type:        "noop",
+		NamespaceID: nsBar.Namespace.ID,
+	}
+	err = c.mount(barCtx, me)
+	require.NoError(t, err)
+
+	ns, err = c.namespaceStore.ModifyNamespaceByPath(barCtx, "foo/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
+
+	ns, err = c.namespaceStore.ModifyNamespaceByPath(barCtx, "foo/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
+
+	me = &MountEntry{
+		Table:       mountTableType,
+		Path:        "fud/bar/baz/foo/",
+		Type:        "noop",
+		NamespaceID: nsBar.Namespace.ID,
+	}
+	err = c.mount(barCtx, me)
+	require.NoError(t, err)
+
+	ns, err = c.namespaceStore.ModifyNamespaceByPath(barCtx, "fud/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
+
+	ns, err = c.namespaceStore.ModifyNamespaceByPath(barCtx, "fud/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
+
+	// Creating a mount at the root level that conflicts with bar/ should fail.
+	me = &MountEntry{
+		Table:       mountTableType,
+		Path:        "bar/",
+		Type:        "noop",
+		NamespaceID: namespace.RootNamespaceID,
+	}
+	err = c.mount(namespace.RootContext(nil), me)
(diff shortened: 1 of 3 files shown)


vault/namespaces_store.go
```diff
@@ -270,6 +271,46 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *Names
...
+	// Before attempting to modify it, ensure we don't have a mount table
+	// entry for it. We assume we only need to look at our parent's
+	// namespace for the last path component; while mount paths can have
+	// any number of components, our namespace only has one and is relative
+	// to some parent path. So, we have to manually iterate over mounts in
+	// a given namespace and ensure no mount has a prefix which is the last
+	// component of our child namespace.
+	if err := func() error {
+		ns.core.mountsLock.RLock()
+		defer ns.core.mountsLock.RUnlock()
```
cipherboy, line 283: Should I just call `ns.core.router.MountConflict(...)` on creation I wonder?

vault/namespaces_store.go
```diff
@@ -270,6 +271,46 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *Names
...
+
+			path = namespace.Canonicalize(strings.TrimPrefix(entry.Namespace.Path, parent.Path))
+		}
+
+		for _, mEntry := range ns.core.mounts.Entries {
+			if mEntry.NamespaceID != parent.ID {
+				continue
+			}
+
+			if strings.HasPrefix(mEntry.Path, path) {
```
wslabosz-reply, line 304: so I don't think that the `*MountEntry` `Path` (in case of a namespace) contains namespace path

wslabosz-reply: at least that's what I see locally, all namespace mounts (whether mounted on namespace creation, or after) have path of "<mountName>/" (like `sys/`, `identity/` or `kv-v2/` if we consider `bao secrets enable -ns=test kv-v2`)

cipherboy: @wslabosz-reply Right. `mEntry.Path` is the mount path with trailing slash, but without namespace prefix. `path` is namespace's final path segment with trailing slash. See line 296 above, where we canonicalize the removal of the _parent_ path (`TrimPrefix(...)`).

vault/namespaces_store.go
driif:
```suggestion
	// Before attempting to modify it, ensure we don't have a mount table
	// entry for it.
	parent, err := namespace.FromContext(ctx)
	if err == nil {
		path := entry.Namespace.Path
		if parent.ID != namespace.RootNamespaceID {
			if !strings.HasPrefix(entry.Namespace.Path, parent.Path) {
				return errors.New("namespace path lacks parent as a prefix")
			}

			path = namespace.Canonicalize(strings.TrimPrefix(entry.Namespace.Path, parent.Path))
		}

		// Use MountConflict to check for conflicts
		if conflict := ns.core.router.MountConflict(ctx, path); conflict != "" {
			return errors.New("namespace path conflicts with existing mount; remount or delete the mount or choose an alternative namespace path")
		}
	}
```

driif: It looks like we could use `mountConflict` here, and then in `assignIdentifier`, we can delete this check since we're already verifying if the namespace exists.
https://github.com/openbao/openbao/blob/6c37482619c0dcbb849ffabef4e0a08973fd61c9/vault/namespaces_store.go#L390-L392

vault/namespaces_store.go
wslabosz-reply: forgot we actually care about mounts with the same name as the namespace...
okay, just explained it to myself, all good!

vault/namespaces_store.go
driif: > @driif shouldn't line 286 be `err != nil`? also I'd refine the comment
> 
> ```
> // Before attempting to modify it, ensure we don't have a mount table
> // entry for it.
> - >
> // Before attempting to modify namespace, ensure we don't have a mount table
> // with same name as the namespace.
> ```

I don't think checking if `err != nil` would have any impact in this case because it should fail later anyway. However, we could definitely check it and return nil or an error instead of continuing further. I believe it's unlikely to get an error here.

wslabosz-reply: > I don't think checking if err != nil would have any impact in this case because it should fail later anyway. However, we could definitely check it and return nil or an error instead of continuing further. I believe it's unlikely to get an error here.

yeah, I've misread your suggested change, but overall I'd advocate for as flat code structure as possible

vault/namespaces_store.go
```diff
@@ -298,6 +305,28 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *Names
...
+		// path component of this new namespace; while mount paths can have
+		// any number of components, our namespace only has one and is relative
+		// to some parent path.
+		path := entry.Namespace.Path
+		if parent.ID != namespace.RootNamespaceID {
+			if !strings.HasPrefix(entry.Namespace.Path, parent.Path) {
+				return errors.New("namespace path lacks parent as a prefix")
+			}
+
+			path = namespace.Canonicalize(strings.TrimPrefix(entry.Namespace.Path, parent.Path))
```
phyrog, line 321:
```suggestion
			path = namespace.Canonicalize(parent.TrimmedPath(entry.Namespace))
```

vault/namespaces_store.go
```diff
@@ -298,6 +305,28 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *Names
...
+	if !exists {
+		// Before attempting to create it, ensure we don't have a mount table
+		// entry that conflicts with this new namespace. We assume we only
+		// need to look at our parent's namespace's mount table for the last
+		// path component of this new namespace; while mount paths can have
+		// any number of components, our namespace only has one and is relative
+		// to some parent path.
+		path := entry.Namespace.Path
+		if parent.ID != namespace.RootNamespaceID {
+			if !strings.HasPrefix(entry.Namespace.Path, parent.Path) {
```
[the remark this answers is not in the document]
cipherboy: Ah, thanks @phyrog !

vault/namespaces_store.go
cipherboy: @wslabosz-reply @driif This has been updated btw. I called a different submethod because we don't need the full `MountConflict(...)`.

driif (approved):

vault/mount_test.go
```diff
@@ -1307,3 +1307,117 @@ func TestCore_MountEntryView(t *testing.T) {
...
+	err := c.mount(namespace.RootContext(nil), me)
+	require.NoError(t, err)
+
+	ns, err := c.namespaceStore.ModifyNamespaceByPath(namespace.RootContext(nil), "foo/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
+
+	ns, err = c.namespaceStore.ModifyNamespaceByPath(namespace.RootContext(nil), "foo/", nil)
+	require.Error(t, err)
+	require.Nil(t, ns)
```
wslabosz-reply, line 1333: why do we try the same operation twice?

vault/namespaces_store.go
```diff
@@ -298,6 +304,28 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *Names
...
+			if !entry.Namespace.HasParent(parent) {
+				return errors.New("namespace path lacks parent as a prefix")
+			}
+
+			path = namespace.Canonicalize(parent.TrimmedPath(entry.Namespace.Path))
+		}
+
+		conflict := ns.core.router.matchingPrefixInternal(ctx, path)
+		if conflict != "" {
+			return fmt.Errorf("new namespace conflicts with existing mount: %v", conflict)
```
wslabosz-reply, line 325: nitpick
```suggestion
			return fmt.Errorf("new namespace conflicts with existing mount: %q", conflict)
```

wslabosz-reply (approved):
left two non-blocking comments

vault/namespaces_store.go
cipherboy: Hmm, there's a lot of places where `%v` is used:

```
[cipherboy@x1c openbao]$ gif 'fmt\.[a-zA-Z]*f(' | grep -i '%q' | wc -l
591
[cipherboy@x1c openbao]$ gif 'fmt\.[a-zA-Z]*f(' | grep -i '%v' | wc -l
892
```

I think identifying and converting all of these to strictly typed variants would be a lot of work and not sure it matters much? Thoughts?

wslabosz-reply: it was a nit of "it looks better with quotes", but overall I'm with you on the repo-wide consistency.

vault/mount_test.go
wslabosz-reply: I thought so, but then could you maybe add a comment before the test, describing the behavior you want to test against? For anyone not familiar with how it behaved before the change, this might be looking like a copying mistake with repeating the same operation.

vault/namespaces_store.go
cipherboy: Ah, gotcha. Very good :-)

wslabosz-reply (approved):
Looks great!

PR #1203, opened 2025-04-09
Forbid illegal namespace characters

### LINKED ISSUE #1183: Namespace name fuzzing results
**Describe the bug**
Based on discussions in Matrix and during the Namespace WG call, I am running some fuzz tests on namespace names to discover which characters might be causing issues. These tests are end-to-end, doing actual HTTP requests, so the whole stack should be covered.

These are the results I have so far (will update once I discover more):

- [ ] `+*` (and any other broken wildcard):
  - ```
    1 error occurred:\n\t* failed to modify namespace: error creating default policies: failed to load default policy: error parsing default policy: failed to parse policy: path \"+*/auth/token/lookup-self\": invalid use of wildcards ('+*' is forbidden)\n\n
    ```
- [ ] ` ` (space)
  - The error message is correct, but should return `400` instead of `500`
  - ```
    "1 error occurred:\n\t* failed to modify namespace: failed validating namespace: \" Q\" contains space characters and cannot be used as a namespace name\n\n"
  ```

**Expected behavior**
For any `500` error that we encounter, we should decide if we want to fix the code to allow namespaces with the failing name, or decide to forbid this name/character and return the proper `400` status.

**Additional context**
`namespaces` branch

phyrog:
This adds a fuzz test for the namespaces API to check for characters that cause problems. It also adds checks to the `(*Namespace).Validate` function for the characters found so far. All illegal characters now return a `400` error code, except for:

- `.`, and `..` return `405` and `404` respectively. They are interpreted by the HTTP stack, so these never reach the actual namespace validation. Not much that we can do here, I'm afraid.

Resolves #1183 

cc: @cipherboy @driif @klaus-sap @satoqz @voigt @wslabosz-reply @genelet

### DIFF
--- a/helper/namespace/namespace.go
+++ b/helper/namespace/namespace.go
@@ -10,6 +10,8 @@ import (
 	"path"
 	"slices"
 	"strings"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/openbao/openbao/sdk/v2/helper/consts"
 )
@@ -17,6 +19,9 @@ import (
 // reservedNames is the list of string names that
 // shouldn't be used as namespace name, hence are forbidden to use
 var reservedNames = []string{
+	"",
+	".",
+	"..",
 	"root",
 	"sys",
 	"audit",
@@ -52,11 +57,21 @@ func (n *Namespace) Validate() error {
 		if segment == "" {
 			return fmt.Errorf("namespace name cannot be empty")
 		}
-		if strings.Contains(segment, " ") {
-			return fmt.Errorf("%q contains space characters and cannot be used as a namespace name", segment)
-		}
 		if slices.Contains(reservedNames, segment) {
 			return fmt.Errorf("%q is a reserved path and cannot be used as a namespace name", segment)
+		}
+
+		for _, r := range segment {
+			switch {
+			case !utf8.ValidRune(r):
+				return fmt.Errorf("%q contains invalid utf-8 characters that cannot be used in a namespace name: %q", segment, r)
+			case !unicode.IsGraphic(r):
+				return fmt.Errorf("%q contains unicode characters that cannot be used in a namespace name: %q", segment, r)
+			case unicode.IsSpace(r):
+				return fmt.Errorf("%q contains space characters that cannot be used in a namespace name", segment)
+			case r == '+' || r == '*':
+				return fmt.Errorf("%q contains wildcard characters that cannot be used in a namespace name: %q", segment, r)
+			}
 		}
 	}
 
--- a/vault/logical_system_namespaces.go
+++ b/vault/logical_system_namespaces.go
@@ -231,7 +231,7 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
 		path := namespace.Canonicalize(data.Get("path").(string))
 
 		if len(path) > 0 && strings.Contains(path[:len(path)-1], "/") {
-			return nil, errors.New("path must not contain /")
+			return logical.ErrorResponse("path must not contain /"), logical.ErrInvalidRequest
 		}
 
 		imetadata, ok := data.GetOk("custom_metadata")
@@ -240,7 +240,7 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
 			metadata = make(map[string]string)
 			for k, v := range imetadata.(map[string]interface{}) {
 				if metadata[k], ok = v.(string); !ok {
-					return nil, fmt.Errorf("custom_metadata values must be strings")
+					return logical.ErrorResponse("custom_metadata values must be strings"), logical.ErrInvalidRequest
 				}
 			}
 		}
@@ -250,7 +250,7 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
 			return ns, nil
 		})
 		if err != nil {
-			return nil, fmt.Errorf("failed to modify namespace: %w", err)
+			return handleError(err)
 		}
 
 		resp := &logical.Response{Data: map[string]interface{}{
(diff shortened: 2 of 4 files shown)


helper/namespace/namespace.go
```diff
@@ -52,11 +57,21 @@ func (n *Namespace) Validate() error {
...
 			return fmt.Errorf("namespace name cannot be empty")
 		}
-		if strings.Contains(segment, " ") {
-			return fmt.Errorf("%q contains space characters and cannot be used as a namespace name", segment)
-		}
-		if slices.Contains(reservedNames, segment) {
-			return fmt.Errorf("%q is a reserved path and cannot be used as a namespace name", segment)
+
+		for _, r := range segment {
+			if slices.Contains(reservedNames, segment) {
```
wslabosz-reply, line 62: this if clause should be outside (before the loop)

helper/namespace/namespace.go
```diff
@@ -52,11 +57,21 @@ func (n *Namespace) Validate() error {
...
+				return fmt.Errorf("%q is a reserved path and cannot be used as a namespace name", segment)
+			}
+			switch {
+			case !utf8.ValidRune(r):
+				return fmt.Errorf("%q contains invalid utf-8 characters that cannot be used in a namespace name: %q", segment, r)
+			case !unicode.IsGraphic(r):
+				return fmt.Errorf("%q contains unicode characters that cannot be used in a namespace name: %q", segment, r)
+			case unicode.IsSpace(r):
+				return fmt.Errorf("%q contains space characters and cannot be used as a namespace name", segment)
```
wslabosz-reply, line 71:
```suggestion
				return fmt.Errorf("%q contains space characters and cannot be used in a namespace name", segment)
```

helper/namespace/namespace.go
phyrog: We're checking each namespace segment for the reserved names, so it should be correct like this?

phyrog: You're right :)

helper/namespace/namespace.go
```diff
@@ -52,12 +57,22 @@ func (n *Namespace) Validate() error {
...
+		for _, r := range segment {
+			switch {
+			case !utf8.ValidRune(r):
+				return fmt.Errorf("%q contains invalid utf-8 characters that cannot be used in a namespace name: %q", segment, r)
+			case !unicode.IsGraphic(r):
+				return fmt.Errorf("%q contains unicode characters that cannot be used in a namespace name: %q", segment, r)
+			case unicode.IsSpace(r):
+				return fmt.Errorf("%q contains space characters and cannot be used in a namespace name", segment)
+			case r == '+' || r == '*':
+				return fmt.Errorf("%q contains wildcard characters and cannot be used in a namespace name: %q", segment, r)
```
wslabosz-reply, line 73: sorry v2 😅 
```suggestion
				return fmt.Errorf("%q contains wildcard characters that cannot be used in a namespace name: %q", segment, r)
```

phyrog:
> I'm not sure about the impact of changes in sdk/logical (probably non-existent), but everything looks right to me, great job!

I did not find any direct usages of `codedError` outside of that file, and since it still fulfills the `HTTPCodedError` interface it should be fine (as far as I can tell).

sdk/logical/error.go
```diff
@@ -63,20 +63,27 @@ type HTTPCodedError interface {
...
 	}
 }
 
 var _ HTTPCodedError = (*codedError)(nil)
 
 type codedError struct {
-	Status  int
-	Message string
+	Status int
+	Err    error
```
driif, line 82: @phyrog What are your thoughts on using "Error" instead of "Message" in coded errors? In my opinion, this change brings no advantages because you don't need to wrap error but could return CodedError(status, err.Error()) without needing to wrap the error. Could you please explain?

phyrog: The thought was to allow proper go error wrapping with this, so things like `errors.Unwrap`, `errors.Is`, `errors.As` can be used on this (although I realize I did not add the `Unwrap` implementation`). `error` itself might already be wrapped, so using the existing implementation and call `err.Error()`, you'll loose that error wrapping hierarchy.

driif: @phyrog Thank you for explanation. Seems like logical errors are mainly used to return response with the status code and simple error message. I don't think you'll need to check error at this stage, but I might be mistaken. IMO, we're adding a feature that might not be utilized.

I suggest not including this changes in this PR. If we want to improve error handling and implement a more standerdized and better approach for returning response errors, we should create an issue for it.

phyrog: @driif Fair, I'll remove it 👍

vault/logical_system_namespaces.go
```diff
@@ -231,7 +231,7 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
 		path := namespace.Canonicalize(data.Get("path").(string))
 
 		if len(path) > 0 && strings.Contains(path[:len(path)-1], "/") {
-			return nil, errors.New("path must not contain /")
+			return handleError(logical.CodedError(http.StatusBadRequest, "path must not contain /"))
```
cipherboy, line 234:
```suggestion
			return return logical.ErrorResponse("path must not contain /"), logical.ErrInvalidRequest
```

I think should also work fwiw.

phyrog: @cipherboy fixed

driif (approved):

cipherboy (approved):
Thanks @phyrog! Always appreciate fuzz testing.

cipherboy:
Ah, @phyrog this conflicts now that I merged @driif's tests. Sorry!

phyrog:
@cipherboy rebased

wslabosz-reply (approved):

cipherboy:
Thanks @phyrog!

PR #1210, opened 2025-04-10
Fix test-ui failing test

fatima2003:
`test-ui` is failing and this is the log: 

```
not ok 1 Chrome 135.0 - [1250 ms] - Acceptance | settings/configure/secrets/ssh: it configures ssh ca
    ---
        actual: >
            SSH Certificate Authority Configuration saved!
        expected: >
            missing public_key
        stack: >
                at Object.<anonymous> (http://localhost:7357/ui/assets/tests.js:6411:14)
        message: >
            renders warning flash message for failed save
        ...
```

The SSH secrets engine no longer requires a mount‑level public_key after #880: 
```
fatima@fattiesPatties-2 openbao % bao secrets enable -path=ssh-test ssh 
Success! Enabled the ssh secrets engine at: ssh-test/
fatima@fattiesPatties-2 openbao % curl -s -o /dev/null -w "%{http_code}\n" \
     -H "X-Vault-Token: $BAO_TOKEN" \
     -X
```

This updates the acceptance test to expect the success flash “SSH Certificate Authority Configuration saved!” instead of the old “missing public_key” error.

### DIFF
--- a/ui/tests/acceptance/settings/configure-secret-backends/configure-ssh-secret-test.js
+++ b/ui/tests/acceptance/settings/configure-secret-backends/configure-ssh-secret-test.js
@@ -38,8 +38,8 @@ module('Acceptance | settings/configure/secrets/ssh', function (hooks) {
     await click(SELECTORS.saveConfig);
     assert.strictEqual(
       flashMessage.latestMessage,
-      'missing public_key',
-      'renders warning flash message for failed save'
+      'SSH Certificate Authority Configuration saved!',
+      'shows success flash when saving mount‑level config without a key'
     );
     await click(SELECTORS.generateSigningKey);
     await click(SELECTORS.saveConfig);


cipherboy (approved):

Gabrielopesantos:
> @Gabrielopesantos I think this is the expected behavior, don't you?

Yes, from an implementation point of view no longer requiring a mount‑level public_key can be the expected behavior as the endpoints that use key material, through the issuer configured on the role, error out when nothing is configured. I'm just not sure if I understood how did that affect this test.
Must say that I did not take UI changes into consideration.

cipherboy:
> I'm just not sure if I understood how did that affect this test.

I think the test was just over-fit for the purpose.

>  Must say that I did not take UI changes into consideration.

Yeah, UI is really hard to update, so I wouldn't expect the feature design to. As people are interested, we'll definitely welcome contributions to improve the UI, but I don't think it prevents releasing any features.

PR #1206, opened 2025-04-09
Implement Cascading Delete for Namespaces and Associated Resources

wslabosz-reply:
Resolves #1025

implements the `delete namespace` operation with following assumptions:
- requests to core have a deadline of 90 seconds, and the deletion operation might take an unknown period of time, depending on the amount of resources to free up, we opt for asynchronous operation
- delete namespace now taints the namespace, blocking all future requests to the tainted namespace
- namespace taint is persisted, all other non-leader nodes will also reject the request
- delete namespace returns a response of `"status": "in-progress"`
- TODO:  for further enhancements, add the status endpoint for checking the deletion progress 
- TODO: in case of the interruption of the deletion operation, the operator should retry the delete operation resulting in a one more attempt at the deletion

```
bao namespace create ns1

bao namespace delete ns1
Key       Value
---       -----
status    in-progress

----

// subsequent calls also lead to
Key       Value
---       -----
status    in-progress
// until the namespace deletion succeeds

----

// during that time all requests using path of a namespace being in process of deletion result in
bao token create -ns=ns1
Error creating token: Error making API request.

Namespace: ns1/
URL: POST http://127.0.0.1:8200/v1/auth/token/create
Code: 404. Errors:

* no handler for route "auth/token/create" on namespace "ns1/". route entry or namespace is tainted.

----

bao write ns1/cubbyhole/test foo=bar
Error writing data to ns1/cubbyhole/test: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/ns1/cubbyhole/test
Code: 404. Errors:

* no handler for route "cubbyhole/test" on namespace "ns1/". route entry or namespace is tainted.
```

```
bao namespace create ns1

bao namespace create -ns=ns1 ns2

bao namespace delete ns1
Error deleting namespace: Error making API request.

URL: DELETE http://127.0.0.1:8200/v1/sys/namespaces/ns1
Code: 400. Errors:

* cannot delete namespace ("ns1/") containing child namespaces
----
curl -X DELETE -H "X-Vault-Token: $(bao print token)" -H "X-Vault-Request: true" http://127.0.0.1:8200/v1/sys/namespaces/ns1
{"errors":["cannot delete namespace (\"ns1/\") containing child namespaces"]}
```
___
cc: @cipherboy @klaus-sap @satoqz @phyrog @voigt @driif

PR #1201, opened 2025-04-08
Tests for namespace  route resolution - issue #1020

driif:
Resolves [#1020](https://github.com/openbao/openbao/issues/1020)

Added Tests for PR [#1028](https://github.com/openbao/openbao/pull/1028)

---
cc: @cipherboy @klaus-sap [@satoqz](https://github.com/satoqz) [@phyrog](https://github.com/phyrog) @voigt @wslabosz-reply

PR #1219, opened 2025-04-15
fix namespace tests

phyrog:
Fixes the broken tests on the namespace branch. Affects all PRs to that branch currently, including #1161, #1206, #1203, #1201

cc: @cipherboy, @driif, @wslabosz-reply, @voigt


wslabosz-reply (approved):
makes most sense, unblocks most

wslabosz-reply:
@phyrog 
I think that it's fixed now, so this can be closed

PR #1220, opened 2025-04-15
Clarify that region is required

cipherboy:
We recently stumbled across this internally when attempting to start OpenBao with GCP CloudKMS without explicitly setting region.

Thanks to Gonzalo Servat for identifying this.


wslabosz-reply (approved):

cipherboy:
Since these are docs-only changes, I'm going to go ahead and merge based on @wslabosz-reply's review. Thanks!

PR #1221, opened 2025-04-15
Fix vault/diagnose test failure

cipherboy:
The Baltimore CyberTrust CA is expiring soon; it is a global root CA installed in most system trust stores but otherwise appears on our diagnose tests. Exclude this particular certificate's serial number from our test suite for now.

See also: https://tls-observatory.services.mozilla.com/static/certsplainer.html?id=16

---

As reported by @driif, thank you!

### DIFF
--- a/vault/diagnose/tls_verification_test.go
+++ b/vault/diagnose/tls_verification_test.go
@@ -429,10 +429,25 @@ func TestTLSLeafCertInClientCAFile(t *testing.T) {
 	}
 	warnings, errs := ListenerChecks(context.Background(), listeners)
 	if errs == nil || len(errs) != 1 {
-		t.Fatal("TLS Config check on bad ClientCAFile certificate should fail once")
+		t.Fatalf("TLS Config check on bad ClientCAFile certificate should fail once: %v", errs)
 	}
 	if warnings == nil || len(warnings) != 1 {
-		t.Fatal("TLS Config check on bad ClientCAFile certificate should warn once")
+		// CN=Baltimore CyberTrust Root,OU=CyberTrust,O=Baltimore,C=IE global
+		// root CA is expiring soon (May '25); it has serial number 33554617
+		// and is affecting tests.
+		var skipBaltimore bool
+		if len(warnings) > 1 {
+			for _, warning := range warnings {
+				if strings.Contains(warning, "33554617") {
+					skipBaltimore = true
+					break
+				}
+			}
+		}
+
+		if !skipBaltimore {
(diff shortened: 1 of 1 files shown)


vault/diagnose/tls_verification_test.go
```diff
@@ -429,10 +429,25 @@ func TestTLSLeafCertInClientCAFile(t *testing.T) {
...
-		t.Fatal("TLS Config check on bad ClientCAFile certificate should fail once")
+		t.Fatalf("TLS Config check on bad ClientCAFile certificate should fail once: %v", errs)
 	}
 	if warnings == nil || len(warnings) != 1 {
-		t.Fatal("TLS Config check on bad ClientCAFile certificate should warn once")
+		// CN=Baltimore CyberTrust Root,OU=CyberTrust,O=Baltimore,C=IE global
+		// root CA is expiring soon (May '25); it has serial number 33554617
+		// and is affecting tests.
+		var skipBaltimore bool
+		if len(warnings) > 1 {
```
driif, line 439:
```suggestion
		if len(warnings) == 2 {
```

driif: Who knows, just to make sure we don’t miss anything:)

wslabosz-reply (approved):
Completely fine to exclude it for now, but what's the long term solution? renewal? if so can it be stated somewhere?

(worrying that it will be left like this for unknown, perhaps long time)

cipherboy:
@wslabosz-reply said:

> but what's the long term solution? renewal? if so can it be stated somewhere?

This is not our certificate, this is a root CA installed by the OS. That means the expected behavior is that the OS root store will eventually update and remove or replace this certificate. 

Not clear is the impact or timeline for this. 

It seems others have run into this: https://www.cisco.com/c/en/us/support/docs/field-notices/742/fn74259.html

Mozilla policy looks to be removal 15 years after creation, which would be here in May on expiry.

DanGhita (commented):
Sorry, I suppose it's obvious, but I don't get how do you convert the SN "020000b9" to "33554617".

cipherboy:
@DanGhita Mozilla has the hash in hexadecimal; `%d` was used in formatting so we output in decimal:

```
>>> 0x020000b9
33554617
```

DanGhita:
Sure, but I had used Copilot to convert and I'm getting "33554489". I suspect a new hallucination (mine or Copilot's) :)

DanGhita (approved):

PR #1225, opened 2025-04-16
ci: increase total-runners and adjust go-test-parallelism

driif:
Increase the total-runners from 16 to 20 in ci.yml and adjust the go-test-parallelism default from 20 to 16 in test-go.yml to optimize test execution and resource allocation.

Resolves occasional CI test failures under high load.

### DIFF
--- a/changelog/1225.txt
+++ b/changelog/1225.txt
@@ -1,0 +1,2 @@
+```release-note:change
+Adjusted CI configuration, improved test stability.
--- a/changelog/1225.txt
+++ b/changelog/1225.txt
@@ -1,2 +1,0 @@
-```release-note:change
-Adjusted CI configuration, improved test stability.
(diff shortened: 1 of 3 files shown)


changelog/1225.txt
cipherboy, line 1: Changelog entry is only required if it affects code, this doesn't so I'd remove it.

driif: Thanks for advice.

driif:
@cipherboy After providing more CI Tests in the main and namespaces branches with tuned CI configs,  I still got an error when testing Expiration Manager:
https://github.com/openbao/openbao/actions/runs/14511955186/job/40712573308?pr=1226

While reviewing recent PRs on the main branch, I didn't notice any errors related to Expiration Manager or nil Pointer dereferences in core.

Changing test parallelism from 20 to 12 and total-runners from 16 to 20 - saw no significant performance impact. 

Since the error still occurs occasionally in the namespaces branch but not in main, I think your suggestion is likely correct: we need to investigate the code for the cause. 

Because the error persists despite CI tuning, I don't see a reason to make further changes in CI for now. 

I'd close both PRs.

cipherboy:
Thanks @driif! Sorry that this didn't seem to help.

Issue #1111, opened 2025-03-12
Loss of barrier sanity when using vault.db from 2.1.0 #1006


LuyeT:
**Describe the bug**
The compiled binary from main branch loses barrier sanity when vault.db was initialised in an older version, release 2.1.0 in this case, though I'm sure this applies to release 2.2.0 as well or vault 1.14.8)

After unsealing, the process enters an halting state which it does not recover from, it does not respond to anything (not even `bao status`) I've narrowed this down the block down to the block starting at L572 in vault/ha.go by adding some log entries at various points. Additionally the process has to be sent SIGKILL to fully exit.

With reverting #1006 fixed this

```go
		{
			c.logger.Info("Entering barrier check")
			c.seal.SetBarrierConfig(activeCtx, nil)
			if c.seal.RecoveryKeySupported() {
				c.seal.SetRecoveryConfig(activeCtx, nil)
			}

			if err := c.performKeyUpgrades(activeCtx); err != nil {
				c.logger.Error("error performing key upgrades", "error", err)

				// If we fail due to anything other than a context canceled
				// error we should shutdown as we may have the incorrect Keys.
				if !strings.Contains(err.Error(), context.Canceled.Error()) {
					// We call this in a goroutine so that we can give up the
					// statelock and have this shut us down; sealInternal has a
					// workflow where it watches for the stopCh to close so we want
					// to return from here
					c.logger.Info("Barrier check shutdown")
					go c.Shutdown()
				}

				c.heldHALock = nil
				lock.Unlock()
				c.stateLock.Unlock()
				metrics.MeasureSince([]string{"core", "leadership_setup_failed"}, activeTime)

				// If we are shutting down we should return from this function,
				// otherwise continue
				if !strings.Contains(err.Error(), context.Canceled.Error()) {
					c.logger.Info("Exiting barrier check - continue")
					continue
				} else {
					c.logger.Info("Exiting barrier check - return")
					return
				}
			}
```

```
2025-03-12T14:16:00.674+0100 [INFO]  core: vault is unsealed
2025-03-12T14:16:00.674+0100 [INFO]  storage.raft: entering follower state: follower="Node at spc-lewis-vault1.p1.lab.basefarm.net:8201 [Follower]" leader-address= leader-id=
2025-03-12T14:16:00.674+0100 [WARN]  storage.raft: heartbeat timeout reached, starting election: last-leader-addr= last-leader-id=
[...]

PR #1234, opened 2025-04-18
Fix deadlock on legacy root key path migration

cipherboy:
As reported by @LuyeT, migration of the root key entry causes a
deadlock. This is because, while holding the write lock, we attempt to
call `barrier.Get(...)`, which acquires its own read lock. Use the
lockless variant instead to avoid this.
    
Resolves: #1111

---

Many thanks to @DanGhita for identifying the cause of the bug.

This can be tested with the attached test case, or by upgrading from v2.2.0 to v2.3.0 (with e.g., Raft) prior to this change.

### DIFF
--- a/vault/barrier_aes_gcm.go
+++ b/vault/barrier_aes_gcm.go
@@ -451,7 +451,7 @@ func (b *AESGCMBarrier) ReloadRootKey(ctx context.Context) error {
 	}
 
 	if out == nil {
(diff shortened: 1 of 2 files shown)


cipherboy:
@DanGhita @JanMa Could I get a review here? Thanks!

DanGhita (approved):

cipherboy:
Thank you @DanGhita!

PR #1235, opened 2025-04-18
fix(ssh): `generate_signing_key=false` not honored on issuer submission endpoints

Gabrielopesantos:
As discussed [here](https://github.com/openbao/openbao/pull/880#discussion_r2051177603), there is a bug in the issuer submission/generation endpoints where the parameter `generate_signing_key=false` was not being honored. This diff fixes it, returning a 400 response with the error message `missing public_key` and introduces a test for the parameters.


cipherboy (approved):
Thanks @Gabrielopesantos!

PR #1257, opened 2025-04-22
Fix make goreleaser-check after split

cipherboy:
After splitting the goreleaser template into separate files, we needed to update the Makefile so that release verification could proceed.


JanMa (approved):

cipherboy:
Thank you @JanMa!

PR #1265, opened 2025-04-24
Add script to verify release artifacts

cipherboy:
In the slow path towards automating more and more of the release process, this script verifies all cosign and GPG signatures as well as checksum files.

### DIFF
--- a/scripts/verify-release.sh
+++ b/scripts/verify-release.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+
+RELEASE_TAG="$1"
+
+GITHUB_TOKEN="${GITHUB_TOKEN:-}"
+OWNER="${OWNER:-openbao}"
+REPO="${REPO:-openbao}"
+TMPDIR="${TMPDIR:-/tmp}"
+GPG_KEY_NAME="${GPG_KEY_NAME:-openbao-gpg-pub-20240618.asc}"
+
+DOWNLOAD_FOLDER="${TMPDIR}/${OWNER}-${REPO}-${RELEASE_TAG}"
+
+if [ ! -e "$DOWNLOAD_FOLDER" ]; then
+  mkdir -p "$DOWNLOAD_FOLDER"
+fi
+
+release_cache="${TMPDIR}/${OWNER}-${REPO}-release-${RELEASE_TAG}.json"
+release_api="https://api.github.com/repos/${OWNER}/${REPO}/releases/tags/${RELEASE_TAG}"
+bearer_header=""
+if [ -n "$GITHUB_TOKEN" ]; then
+  bearer_header="Authorization: Bearer $GITHUB_TOKEN"
+fi
+
+release_info=""
+if [ -e "$release_cache" ]; then
+  release_info="$(cat "$release_cache")"
+else
+  release_info="$(curl -sSL \
+    -H "Accept: application/vnd.github+json" \
+    -H "$bearer_header" \
+    -H "X-GitHub-Api-Version: 2022-11-28" \
+    "$release_api")"
+  echo "$release_info" > "$release_cache"
+fi
+
+assets_count="$(jq '.assets | length' <<< "$release_info")"
+
+for index in $(seq 0 $(( assets_count - 1 ))); do
+  download_url="$(jq -r ".assets[$index].browser_download_url" <<< "$release_info")"
+  download_file="$(basename "$download_url")"
+  download_path="${DOWNLOAD_FOLDER}/${download_file}"
+
+  if [ ! -e "$download_path" ]; then
+    echo "Downloading $download_file..."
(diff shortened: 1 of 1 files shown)


chicks-net:
Please consider trying https://www.shellcheck.net/ on this code.  The code is good, but it could be much better with a few cleanups.

Is the subshell so that you scope the strict mode `set` stuff?

cipherboy:
@chicks-net Hmm, maybe my package is broken?

```
[cipherboy@x1c openbao]$ shellcheck scripts/verify-release.sh 
[cipherboy@x1c openbao]$ echo $?
0
[cipherboy@x1c openbao]$ rpm -ql ShellCheck
/usr/bin/shellcheck
/usr/share/doc/ShellCheck
/usr/share/doc/ShellCheck/CHANGELOG.md
/usr/share/doc/ShellCheck/README.md
/usr/share/licenses/ShellCheck
/usr/share/licenses/ShellCheck/LICENSE
/usr/share/man/man1/shellcheck.1.gz
[cipherboy@x1c openbao]$ sudo dnf info ShellCheck
Installed packages
Name            : ShellCheck
Epoch           : 0
Version         : 0.10.0
Release         : 4.fc42
Architecture    : x86_64
Installed size  : 20.6 MiB
Source          : ShellCheck-0.10.0-4.fc42.src.rpm
From repository : fedora
Summary         : Shell script analysis tool
URL             : https://www.shellcheck.net/
License         : GPL-3.0-or-later
Description     : The goals of ShellCheck are:
                : 
                : * To point out and clarify typical beginner's syntax issues,
                :   that causes a shell to give cryptic error messages.
                : 
                : * To point out and clarify typical intermediate level semantic
                :  problems,
                :   that causes a shell to behave strangely and counter-intuitiv
                : ely.
                : 
                : * To point out subtle caveats, corner cases and pitfalls, that
                :  may cause an
                :   advanced user's otherwise working script to fail under futur
                : e circumstances.
Vendor          : Fedora Project
```

scripts/verify-release.sh
```diff
@@ -0,0 +1,96 @@
...
+
+if [ ! -e "${DOWNLOAD_FOLDER}/${GPG_KEY_NAME}" ]; then
+  echo "Missing GPG file ($GPG_KEY_NAME); upload to release and rerun" 1>&2
+  exit 1
+fi
+
+cd "$DOWNLOAD_FOLDER"
+
+(
+  set -euxo pipefail
```
cipherboy, line 58: @chicks-net Ahh, I see. If I remove this line, I get a warning about the line above:

```
[cipherboy@x1c openbao]$ shellcheck scripts/verify-release.sh 

In scripts/verify-release.sh line 55:
cd "$DOWNLOAD_FOLDER"
^-------------------^ SC2164 (warning): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

Did you mean: 
cd "$DOWNLOAD_FOLDER" || exit

For more information:
  https://www.shellcheck.net/wiki/SC2164 -- Use 'cd ... || exit' or 'cd ... |...
```

so I think shellcheck assumes it has global impact, rather than understanding it only affects execution within the subshell and only after it has been executed.

JanMa (approved):
LGTM

scripts/verify-release.sh
```diff
@@ -0,0 +1,96 @@
...
+  download_url="$(jq -r ".assets[$index].browser_download_url" <<< "$release_info")"
+  download_file="$(basename "$download_url")"
+  download_path="${DOWNLOAD_FOLDER}/${download_file}"
+
+  if [ ! -e "$download_path" ]; then
+    echo "Downloading $download_file..."
+    curl -sSL "$download_url" --output "$download_path"
+    sleep 2
+  fi
+done
```
JanMa, line 48: Instead of looping over each individual asset here, it would be much quicker to make use of the `gh` CLI to download all assets in parallel.  For example:

```
gh release download v2.2.1 -D /tmp/openbao-openbao-v2.2.1/ --skip-existing
```

In general, making use of the GitHub CLI would also simplify other parts of the script, because we don't need to manually fetch all release assets first.

cipherboy: Ah, I hadn't realized it has a downloader. I don't have it locally though... It looks like they don't maintain a container image either, so we'd have to build a custom Dockerfile for this. :laughing: 

https://github.com/cli/cli/issues/2027#issuecomment-977241407

JanMa: It is preinstalled on all GitHub Actions runners. I guess we're planning to run this script in CI anyways, so the missing image will not be an issue

cipherboy:
I'm going ahead and closing this; I'll let the next adventurous person pick up a rewrite and hopefully sometime we'll have a proper release validation workflow.

PR #1267, opened 2025-04-24
Add CEL best practices RFC

cipherboy:
This RFC seeks to unify behavior and expectations around #753 and #493, plus any upcoming future RFCs around KV or policies.

fyi: @suprjinx @fatima2003 @achetronic

### DIFF
--- a/website/content/docs/rfcs/cel-best-practices.mdx
+++ b/website/content/docs/rfcs/cel-best-practices.mdx
@@ -1,0 +1,164 @@
+---
+sidebar_label: Best Practices for Integrating CEL in Auth, Secrets, and Elsewhere
+description: |-
+  An OpenBao RFC to discuss how Google's Common Expression Language should be integrated into OpenBao.
+---
+
+# Best Practices for Integrating CEL in Auth, Secrets, and Elsewhere
+
+## Summary
+
+Two RFCs ([#493](https://github.com/openbao/openbao/issues/493#issuecomment-2549033976) and [#753](https://github.com/openbao/openbao/issues/753)) and two or more issues ([#514](https://github.com/openbao/openbao/issues/514), [#1207](https://github.com/openbao/openbao/issues/1207), and [#1227](https://github.com/openbao/openbao/issues/1227)) have sought to include CEL in OpenBao.
+
+This RFC attempts to set project-wide guidelines for its adoption to ensure a common operator experience regardless of which area of OpenBao they're using CEL in.
+
+## Problem Statement
+
+Many areas of OpenBao have a programability problem. For authentication against arbitrary objects (e.g., JWT tokens which can include custom claims of arbitrary JSON types or X.509 certificate validation or construction which may include arbitrary ASN.1-encoded extensions), OpenBao cannot know how to parse, validate, or generate everything.
+
+This leaves two alternatives for end-operators:
+
+1. Change OpenBao either through upstream contributions or a custom plugin version, or
+2. In our desired method, using [Google's Common Expression Language](https://cel.dev/) to implement their own authentication or generation.
+
+However, every place CEL is used presents an opportunity for confusion and different semantics.
+
+Currently, we have two open PRs for CEL support in OpenBao: [#794 for PKI Secrets](https://github.com/openbao/openbao/pull/794) by @fatima2003 and [#869 for JWT Auth](https://github.com/openbao/openbao/pull/869) by @suprjinx.
+
+These have two slightly different semantics:
+
+1. [PKI seeks to give CEL full issuance control](https://github.com/openbao/openbao/blob/9acf8fc25e64b670cfa20f3feb6b5c14023510e6/builtin/logical/pki/path_issue_sign.go#L868-L908) over the final certificate.
+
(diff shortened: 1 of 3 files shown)


website/content/docs/rfcs/cel-best-practices.mdx
```diff
@@ -0,0 +1,164 @@
...
+## Unresolved Questions
+
+Not clear is the behavior of [Object construction](https://github.com/google/cel-spec):
+
+```cel
+// Object construction
+common.GeoPoint{ latitude: 10.0, longitude: -5.5 }
+```
+
+This would be ideal (for auth and for certificate creation), but may be more complex than desired.
```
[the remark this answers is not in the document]
cipherboy: @suprjinx Thanks!

That's fair, I think that's a reasonable change. The benefit of that is it allows both validation and output to happen in one step (which may be nice, but may also be confusing from a safety/correctness PoV if you're actively modifying the output while also incrementally validating). You'd hate to risk having a partial object lying around that accidentally gets used and has broader scope as a result, IMO.

I'd really like a JSON-like output mode, tbh. Then the underlying output object can change and we wouldn't necessarily have to update the CEL bindings to add a new helper. In PKI for instance, we can loosely bind to x509.Certificate and json.Unmarshal to/from that to let CEL authors pick up new x509.Certificate fields without involving us.

Thoughts @fatima2003?

website/content/docs/rfcs/cel-best-practices.mdx
```diff
@@ -0,0 +1,164 @@
...
+### Libraries
+
+In pursuit of 3, we should introduce plugin-wide (and eventually, global) libraries for common components. These are collections of variables with the format:
+
+```json
+{
+  "variables": [
+    {
+      "name": "<var-name>",
+      "expression": "<cel-expression>"
```
fatima2003, line 79: To take this a step further, I think having both `internal` and `external` variables inside CEL libraries would be a good idea: 
• **Internal variables** would hold small, reusable sub-expressions that aren’t useful alone but are referenced by other variables.
• **External variables** the set of variables which are exposed to the public and can be imported.
The evaluation order would be: compile all internal variables first, then compile external variables (since they may reference the internals).

website/content/docs/rfcs/cel-best-practices.mdx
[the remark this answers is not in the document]
fatima2003: Sorry for the late follow up! Took some time trying to understand and compare both @suprjinx's approach and what I did 😅

I checked out the JWT CEL implementation in #868 and it takes on a different approach on evaluating and modifying the users parameters through the CEL program. 

In #794 (the PKI CEL implementation), each certificate attribute is a CEL expression and evaluated one by one in the go program (vs calling custom CEL functions with the attributes). This also helps with tracing errors since the exact expression causing the error is thrown. All the expressions share a CEL environment where the variables and user request are declared. Instead of using custom CEL functions (like SetPolicies, SetTTL, etc.) that directly mutate the role struct, the `certificate_template` is put together in the go program. 

#868 approach works well for JWT since only the user doesn't have to specify what attributes they want and just specify the role they would like request against. Versus for certificate issuance policies, the CEL program would become overly complex and one large evaluation... debugging it would be hard for the CEL Author and the person requesting a certificate.

website/content/docs/rfcs/cel-best-practices.mdx
cipherboy: I like this, I think it is a good approach. I don't think we need all of this up front for this release (especially the library functionality), but I think we'll get there over time.

cipherboy:
@suprjinx @fatima2003 This has been updated, let me know what you think and I shall merge :-)

website/content/docs/rfcs/cel-best-practices.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+   The net result of this evaluation should be a one of three values:
+
+   1. A boolean literal `false`, indicating the default error message should be returned to the caller. For e.g., login requests, this would be `permission denied` or similar.
+   2. A string error message, giving more detailed failure rationale.
+   3. The final success response object, a Protobuf `message` mimicking e.g., `logical.Auth` or some other plugin-specific structure.
+
+   Notably, while object construction is hard to handle in general (to/from `ref.Val`), using Protobuf is simpler to cross the Go<->CEL boundary. Certain areas, like [`logical.Auth`](https://github.com/openbao/openbao/blob/9bdcc26ef9ed7114f74075d196aeb46c96357a7b/sdk/logical/auth.go#L13-L117) already have [Protobuf variants](https://github.com/openbao/openbao/blob/9bdcc26ef9ed7114f74075d196aeb46c96357a7b/sdk/plugin/pb/backend.proto#L150-L227), along with [conversion helpers](https://github.com/openbao/openbao/blob/9bdcc26ef9ed7114f74075d196aeb46c96357a7b/sdk/plugin/pb/translation.go#L506-L594).
```
fatima2003, line 66: I think this line is throwing an error because of the '-' in '<->'.
Perhaps changing it to '⇌'

fatima2003 (commented):
Otherwise this looks great, thank you @cipherboy!

fatima2003 (approved):

suprjinx (approved):
Looks great, but there is a markdown error in CI that @fatima2003 mentions

cipherboy:
Fixed. Thanks all!

PR #1269, opened 2025-04-24
Fix preSeal ordering w.r.t NamespaceStore

cipherboy:
Within preSeal, we faced a race condition:

 - NamespaceStore was first destroyed
 - Then RollbackManager was stopped.

In particular, this ordering means that any rollback functions which
relied on namespaces, such as the token or identity subsystems, would
occasionally panic.

This was only reported in CI, but the code path is reachable in
production systems and thus may have manifested in heavy load scenarios.

Tear down the namespace store as the very last thing, so that rollback
manager and any dangling mount options can access namespaces if
necessary.

---

cc: @wslabosz-reply @klaus-sap @satoqz @phyrog @voigt @driif

### DIFF
--- a/vault/namespaces_store.go
+++ b/vault/namespaces_store.go
@@ -529,6 +529,9 @@ func (ns *NamespaceStore) ModifyNamespaceByPath(ctx context.Context, path string
 func (ns *NamespaceStore) ListAllNamespaces(ctx context.Context, includeRoot bool) ([]*namespace.Namespace, error) {
 	defer metrics.MeasureSince([]string{"namespace", "list_all_namespaces"}, time.Now())
 
+	ns.lock.RLock()
+	defer ns.lock.RUnlock()
+
 	namespaces := make([]*namespace.Namespace, 0, len(ns.namespacesByUUID))
 	for _, entry := range ns.namespacesByUUID {
 		if !includeRoot && entry.ID == namespace.RootNamespaceID {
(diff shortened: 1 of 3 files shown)


vault/namespaces_store.go
```diff
@@ -599,6 +602,13 @@ func (ns *NamespaceStore) taintNamespace(ctx context.Context, namespaceToTaint *
 func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, uuid string) (string, error) {
 	defer metrics.MeasureSince([]string{"namespace", "delete_namespace"}, time.Now())
 
+	// We grab a read lock here; the goroutine will acquire the write lock to
+	// actually remove the namespace, but we'll have exited prior to that (or,
+	// will be able to as we're not actively blocking that goroutine on lock
+	// write acquisition while holding this read lock as well).
+	ns.lock.RLock()
+	defer ns.lock.RUnlock()
```
satoqz, line 610: Sadly not the case, since we're also grabbing the write lock in the same Goroutine down below in `ns.taintNamespace(...)` so this is a deadlock. This is why `test-go(7)` currently times out.

Suggestion for lines 605-619:

```go
namespaceToDelete, err := ns.GetNamespace(ctx, uuid)
if err != nil {
	return "", err
}
if namespaceToDelete == nil {
	return "", nil
}
```

`GetNamespace` already checks invalidation, and releases the lock enough.

satoqz (approved):
Besides the locking party, makes sense to me!

driif:
Looks also good to me.

vault/namespaces_store.go
```diff
@@ -599,12 +602,20 @@ func (ns *NamespaceStore) taintNamespace(ctx context.Context, namespaceToTaint *
...
 		return "", err
 	}
 
-	namespaceToDelete, ok := ns.namespacesByUUID[uuid]
-	if !ok {
+	namespaceToDelete, err := ns.GetNamespace(ctx, uuid)
+	if err != nil {
+		return "", err
+	}
+	if namespaceToDelete == nil {
```
satoqz, line 618: nit: `GetNamespace` already calls `checkInvalidation`

cipherboy: Ah, nice catch! Fixed.

PR #1270, opened 2025-04-25
fix data races in namespace store

satoqz:
`test-go(7)` with data race detection currently fails `TestNamespaceStore`.

Both `ListAllNamespaces` and `DeleteNamespace` miss a read lock. It seems the data race test failure was caused by `ListAllNamespaces` running without a read lock while `clearNamespaceResources` held the write lock.


satoqz:
Ah, didn't notice this was already hidden in #1269. `test-go(7)` still fails over there though? Investigating 😄.

PR #1272, opened 2025-04-25
chore: remove artifacts in `dependencies/`

satoqz:
Can't help but notice `dependencies/2-25-21` stick out like a sore thumb any time I visit the repo's homepage.

Is `scripts/deps_upgrade.py` still in use, or can it be removed as well?


cipherboy:
I've not seen it used in a while, I'm happy removing it.

cipherboy (approved):
Thank you!

PR #1280, opened 2025-04-28
Support retrying PostgreSQL connection

cipherboy:
When starting OpenBao, we formerly assumed that the database instance was already up and running. In certain scenarios, such as executing in the context of Kubernetes, with a PGBouncer instance, or in Podman Compose, we might have a looser coupling of service availability. Support graceful retry logic in the initial connection handler to accommodate startup ordering issues.

### DIFF
--- a/physical/postgresql/postgresql.go
+++ b/physical/postgresql/postgresql.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/armon/go-metrics"
+	"github.com/cenkalti/backoff/v4"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
 	"github.com/hashicorp/go-uuid"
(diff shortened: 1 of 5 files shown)


website/content/docs/configuration/storage/postgresql.mdx
```diff
@@ -79,6 +79,12 @@ to disable SSL.
   PostgreSQL 9.5 or later. Set to `true` if the database user does not have
   the required permissions; otherwise, OpenBao will fail to start.
 
+- `max_connect_retries` `(int: 1)` - Maximum number of retries to perform
+  when waiting for the database to be active. This uses exponential backoff,
+  starting with 15ms and increasing to 5s between retries. To ensure the
+  connection is active, issues a `SELECT RANDOM() as random` query. Can be
```
cipherboy, line 85: This is incorrect, we call `db.Ping(...)`, whose behavior isn't specified: https://pkg.go.dev/database/sql#DB.PingContext

> Ping verifies a connection to the database is still alive, establishing a connection if necessary.

physical/postgresql/postgresql.go
```diff
@@ -301,6 +317,38 @@ func connectionURL(conf map[string]string) string {
 	return connURL
 }
 
+func doRetryConnect(logger log.Logger, connURL string, retries uint64) (*sql.DB, error) {
+	db, err := sql.Open("pgx", connURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to postgres: %w", err)
```
satoqz:
```suggestion
		return nil, err
```

else we get `failed to connect to postgres: failed to connect to postgres: ...`

satoqz (approved):

physical/postgresql/postgresql.go
cipherboy: Ahh yes, thank you!

cipherboy:
Based on @satoqz's review and the size of this change, I'm happy merging this. Thank you!

PR #1282, opened 2025-04-29
Namespace-aware password policy storage & lookup

### LINKED ISSUE #1263: Namespace-aware Password Policies
**Describe the Bug**

Password policies are a [system endpoint](https://openbao.org/api-docs/system/policies-password/) fully distinct from Bao's "normal" policies.

Based on [this guide](https://developer.hashicorp.com/vault/tutorials/db-credentials/password-policies?productSlug=vault&tutorialSlug=policies&tutorialSlug=password-policies), password policies can be used in two ways:

1) Manually generate a password via `/sys/policies/password/:name/generate`
2) Attach a password policy to an auth engine that supports it

Based on 2), we can infer that password policies should be per-namespace, as (auth) mounts are also per-namespace. So any namespace administrator creating in auth mount in their namespace should probably also manage the password policies that can be associated to it.

I don't think something like password policy inheritance from ancestor namespaces should work as password policies are simply referred to by name, so we wouldn't be able to disambiguate between a password policy in the auth mount's namespace and one in an ancestor namespace if they share the same name.

**Suggested solution**

That said, let's look at the current implementation:

Request handlers for password policy CRUD + password generation use `req.Storage`, which is namespace-aware storage already. https://github.com/openbao/openbao/blob/dc1ae8d828794dab659fd28cb936b8f4c9025a79/vault/logical_system.go#L2903

When an auth engine plugin uses a password policy to generate a password however, the call goes through

https://github.com/openbao/openbao/blob/dc1ae8d828794dab659fd28cb936b8f4c9025a79/vault/dynamic_system_view.go#L389

which then calls

https://github.com/openbao/openbao/blob/dc1ae8d828794dab659fd28cb936b8f4c9025a79/vault/password_policy_util.go#L17

to read the policy from storage. The latter file is obviously a stub and always reads from `systemBarrierView`, which is not in line with what we do in request handling already. The context passed to `retrievePasswordPolicy()` should have the correct namespace for the calling auth engine mount, so it should be possible to use the correct storage view via `NamespaceView()`.

satoqz:
Resolves #1263

In addition to checking the written storage paths, we did some testing with the database secrets engine + Postgres, ensuring that it can only find the password policies for the namespace the engine is enabled in.

### DIFF
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -2210,7 +2210,10 @@ func (c *Core) mountEntryView(me *MountEntry) (BarrierView, error) {
 	}
 
 	switch me.Type {
-	case mountTypeSystem:
+	case mountTypeSystem, mountTypeNSSystem:
+		if me.Namespace() != nil && me.NamespaceID != namespace.RootNamespaceID {
+			return c.namespaceMountEntryView(me.Namespace(), systemBarrierPrefix), nil
+		}
 		return NewBarrierView(c.barrier, systemBarrierPrefix), nil
 	case mountTypeToken:
 		return NewBarrierView(c.barrier, systemBarrierPrefix+tokenSubPath), nil
--- a/vault/dynamic_system_view_test.go
+++ b/vault/dynamic_system_view_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/openbao/openbao/helper/namespace"
 	"github.com/openbao/openbao/sdk/v2/framework"
 	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/stretchr/testify/require"
 )
 
 var (
@@ -33,6 +34,12 @@ rule "charset" {
 }
 rule "charset" {
 	charset = "0123456789"
+	min_chars = 1
+}`
+	rawTestPasswordPolicy2 = `
+length = 10
+rule "charset" {
+	charset = "abcdefghijklmnopqrstuvwxyz"
 	min_chars = 1
 }`
 )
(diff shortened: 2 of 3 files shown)


vault/mount.go
```diff
@@ -2210,7 +2210,10 @@ func (c *Core) mountEntryView(me *MountEntry) (BarrierView, error) {
 	}
 
 	switch me.Type {
-	case mountTypeSystem:
+	case mountTypeSystem, mountTypeNSSystem:
+		if me.Namespace() != nil && me.NamespaceID != namespace.RootNamespaceID {
+			return c.namespaceMountEntryView(me.Namespace(), systemBarrierPrefix), nil
+		}
```
satoqz, line 2216: FWIW, it seems that the only components of the system backend that use `req.Storage` are quotas and password policies. So since password policies also go under `sys/`, @voigt you could actually retain the usage of `req.Storage` in #1161.

Generally, I don't think it ever makes sense to have a storage path like:
```
/namespaces/<namespace uuid>/logical/<sys mount uuid>/...
```

cipherboy: @satoqz We were intending to have that setup for mounts, yes. We wanted all Namespace storage to be under the namespace path, so that it could later be moved to an external storage backend and still be transactional. My 2c but I think the longer path is fine.

cipherboy: @satoqz My bad, I thought you meant for a non-sys plugin. Yes, `sys/` is preferred to UUID.

voigt (approved):
Looks good!

cipherboy (commented):
It might be nice to have a test for usage of policies in a namespace, but other than that, I'm happy with this. 

I'm surprised it doesn't conflict, though.

satoqz:
> It might be nice to have a test for usage of policies in a namespace, but other than that, I'm happy with this.

Added!

> I'm surprised it doesn't conflict, though.

Same.

voigt (approved):

cipherboy (approved):
Thank you!

PR #1279, opened 2025-04-28
Move namespace resolution from http to core

satoqz:
Alternative to #1262
Resolves #1229, #1268

PR #1283, opened 2025-04-29
#1161 rebased on top of #1279, for testing :)

satoqz:
See https://github.com/openbao/openbao/pull/1279#issuecomment-2838831912


satoqz:
Having a hard time locally reproducing the failure in `vault.TestExpiration_Tidy`, but I've seen that one before while creating #1279 (and it then disappeared while fixing a bunch of other tests?)

Other than that, green I suppose.

cipherboy:
@satoqz Neat! That's good.

PR #1284, opened 2025-04-29
Add support for non-voting PostgreSQL nodes

cipherboy:
When running OpenBao against a replicated PostgreSQL cluster, certain
PostgreSQL nodes may not be able to perform writes as they are read-only
followers. OpenBao nodes on these clusters should not attempt write
operations and thus should not attempt to participate in leadership
elections.

Changing this value at runtime is currently impossible; standby nodes do
not have token information and thus if the active node is actively
marked as a non-voter (and no other voters remain), the cluster will be
without a leader. Promoting a leader would thus require changing the
configuration of a node and restarting. When standby nodes can handle
read-only requests (from an underlying storage PoV), live modification
could be added.

---

Blocked on #1280; will be rebased once that merges.


physical/postgresql/postgresql.go
```diff
@@ -238,6 +257,14 @@ func NewPostgreSQLBackend(conf map[string]string, logger log.Logger) (physical.B
 		upsert_function: quoted_upsert_function,
 	}
 
+	// Determine if we should participate in leadership elections. This can
+	// be changed at runtime but is not persisted.
+	nonVoter, err := permanentStandby(conf)
```
cipherboy, line 262: I'm not quite sure how to test this feature in CI. I think we'll want a replicated PGSQL environment, which I can set up with docker-compose, and then we'll also need the ability to modify a node's config and restart it. The restart part is difficult; the dockerized testcluster would let us do full binary testing (I'm not sure if it supports config file snippets but we could theoretically add it and PostgreSQL support), but we can't reliably restart these nodes with different configs after they were initially started up...

cipherboy:
I think this code is just broken, so we should close it and redo it.

PR #1288, opened 2025-05-01
Minor cleanup of the rekey webpage

OdyX:
The https://openbao.org/docs/commands/operator/rekey/ page has two minor issues:
* still references hashicorp as keybase key names
* one example has broken end-of-line handling in shell excerpt


cipherboy (approved):
Thank you @OdyX!

Out of curiosity, does Keybase support still work? For some reason I thought their service was being sunset.

OdyX:
> Out of curiosity, does Keybase support still work? For some reason I thought their service was being sunset.

It did for me today. Not a fan of their service, but it was an easy test to do as I already had an account with my GnuPG key there. Thanks for the quick merge!

PR #1296, opened 2025-05-05
Delete namespace-level quota on namespace deletion

satoqz:
Followup to #1161 and #1206 because it turns out "it depends" would have been the correct answer to https://github.com/openbao/openbao/pull/1161#discussion_r2068161550 😄

Quotas attached to a mount are automatically deleted on unmount, which ususally clears all namespace-specific quotas on namespace deletion.

There's one edge case where quota is not attached to the mount and thus not removed on namespace deletion yet:

`path=foo/bar` where `foo/bar` is a namespace, no mount path, no API path

---

cc @cipherboy @voigt @wslabosz-reply

### DIFF
--- a/vault/quotas/quotas.go
+++ b/vault/quotas/quotas.go
@@ -1204,6 +1204,49 @@ func (m *Manager) HandleRemount(ctx context.Context, from, to namespace.MountPat
 	return nil
 }
 
+// pruneQuota is a helper to delete quota entries from memdb and storage as part of
+// mount disabling or namespace deletion.
+// The caller must hold quotaLock in write mode and dbAndCacheLock in read/write mode.
+func (m *Manager) pruneQuota(ctx context.Context, txn *memdb.Txn, idx string, args ...any) error {
+	for _, quotaType := range quotaTypes() {
+		iter, err := txn.Get(quotaType, idx, args...)
+		if err != nil {
+			return err
+		}
+		for raw := iter.Next(); raw != nil; raw = iter.Next() {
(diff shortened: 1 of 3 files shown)


vault/quotas/quotas.go
```diff
@@ -1222,39 +1265,20 @@ func (m *Manager) HandleBackendDisabling(ctx context.Context, nsPath, mountPath
...
-					return fmt.Errorf("failed to delete quota from db after mount disabling; namespace %q, err %v", nsPath, err)
-				}
-				quota := raw.(Quota)
-				if err := m.storage.Delete(ctx, QuotaStoragePath(quotaType, quota.QuotaName())); err != nil {
-					return fmt.Errorf("failed to delete quota from storage after mount disabling; namespace %q, err %v", nsPath, err)
-				}
-			}
-		}
-		return nil
-	}
```
satoqz, line 1242: I decided to factor this out and have both the existing `HandleBackendDisabling` and the new `HandleNamespaceDeletion`. Technically this could have remained as a single function checking if `mountPath == ""`, but I thought I'd make intent clearer.

wslabosz-reply (commented):
Looks correct to me, I'd reckon this could use a single unit test that will showcase the deletion of the mount-less quota (with namespace deletion)

satoqz:
Tests added!

voigt (approved):
Looks Clean! Appreciate the comments.
LGTM

cipherboy (approved):
Definitely! Thanks @satoqz!

PR #1292, opened 2025-05-02
chore(tools): rm codechecker bin

bazaah:
## Change

This was erroneously checked in many moons ago in:

- hashicorp/vault@8c18f24b9da475c13f7908e609c5d4be24c773e6 hashicorp/vault#21611

and is now removed, in service of openbao/dev-wg#37

---

Resolves: openbao/dev-wg#37

PR #1297, opened 2025-05-05
Allow empty PostgreSQL connection URLs

cipherboy:
When using the PostgreSQL storage implementation, libpgx supports component-wise standard environment variables for connection information. This means that one complete connection string may not be present but the connection would still succeed. Support empty connection information as a result.

### DIFF
--- a/physical/postgresql/postgresql_test.go
+++ b/physical/postgresql/postgresql_test.go
@@ -6,6 +6,7 @@ package postgresql
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"testing"
 	"time"
@@ -15,6 +16,7 @@ import (
 	"github.com/openbao/openbao/helper/testhelpers/postgresql"
 	"github.com/openbao/openbao/sdk/v2/helper/logging"
 	"github.com/openbao/openbao/sdk/v2/physical"
+	"github.com/stretchr/testify/require"
 )
 
 func TestPostgreSQLBackend(t *testing.T) {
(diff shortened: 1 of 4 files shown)


physical/postgresql/postgresql_test.go
```diff
@@ -444,3 +446,43 @@ func TestPostgreSQLBackend_NoCreateTables(t *testing.T) {
 	logger.Info("Running basic backend tests")
 	physical.ExerciseBackend(t, b)
 }
+
+func TestPostgreSQLBackend_PGEnv(t *testing.T) {
```
wslabosz-reply, line 450: interesting how it's requiring comment here, I could swear that it was not mandatory (by linter) for the `test` files, maybe new change?

cipherboy: @wslabosz-reply Yeah, it was enforced in the upstream repository. I think what happened was that @bazaah removed the existing precompiled binary in #1292, which was compiled for macOS. 

If you go to an earlier commit, #1251, and look at the code checks, it seems like it didn't actually execute: https://github.com/openbao/openbao/actions/runs/14597860000/job/40948279036

```
Run make ci-vet-codechecker
  make ci-vet-codechecker
  shell: /usr/bin/bash -e {0}
==> Checking that build is using go version >= 1.24.2...
==> Using go version 1.24.2...
go: not generating in packages in dependency modules
go: not generating in packages in dependency modules
```

It seems to silently fail to execute on Linux, if you drop in a MacOS binary. Joy!

JanMa (approved):
LGTM :+1:

physical/postgresql/postgresql_test.go
cipherboy: @wslabosz-reply I fixed this before merging. Thanks!

PR #1299, opened 2025-05-08
PostgreSQL - Remove redundant PermitPool

cipherboy:
Calling `db.SetMaxOpenConns(...)` has the same effect as using a permit
pool; Go will internally do locking and ensure that no more connections
are created.

Adds a test case to verify this behavior, which should allow us to avoid
double locking and the latency that brings with.

---

This is based on #1280 and will be rebased when it lands.

### DIFF
--- a/physical/postgresql/postgresql_test.go
+++ b/physical/postgresql/postgresql_test.go
@@ -542,7 +542,7 @@ func TestPostgreSQLBackend_Parallel(t *testing.T) {
 
 	b := bRaw.(physical.TransactionalBackend)
 
-	// Put should succeed with an error even with massively parallel
+	// Put should succeed without an error even with massively parallel
 	// requests.
 	errors := make([]error, 100)
 	var wg sync.WaitGroup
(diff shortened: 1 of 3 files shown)


physical/postgresql/postgresql_test.go
```diff
@@ -488,3 +490,142 @@ func TestPostgreSQLBackend_PGEnv(t *testing.T) {
...
+		}(j)
+
+		wg.Wait()
+
+		for j := range errors {
+			if errors[j] != nil {
+				t.Fatalf("process %v: %v", j, errors[j])
+			}
+		}
+	}
```
satoqz, line 630: Shouldn't this be outside of the loop?

```suggestion
	}
	
	wg.Wait()

	for j := range errors {
		if errors[j] != nil {
			t.Fatalf("process %v: %v", j, errors[j])
		}
	}
```

satoqz (approved):
Makes sense.

cipherboy:
Thanks @satoqz! I've rebased this one.

cipherboy:
Race test should be happy this time; I forgot to redefine `err` within the goroutine in the test.

cipherboy:
Based on @satoqz's approval and the size and scope of this change, I'm happy merging. Thank you!

PR #1298, opened 2025-05-08
Namespaces locking/unlocking operation

wslabosz-reply:
EDIT: closed in favor of #1300 

Resolves #1181
Based on #1279

Implements `bao namespace lock <path>` and `bao namespace unlock <path>` operations.

- Locking the namespace results in all requests to that namespace being rejected with `503` error code and a error message
```
bao namespace create test   
bao namespace lock test   
Key           Value
---           -----
unlock_key   <unlock_key>
(...)
bao write -ns=test cubbyhole/foo bar=baz
Error writing data to cubbyhole/foo: Error making API request.

Namespace: test/
URL: PUT http://127.0.0.1:8200/v1/cubbyhole/foo
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "test/" must be unlocked to gain access.
(...)
bao token create -ns=test
Error creating token: Error making API request.

Namespace: test/
URL: POST http://127.0.0.1:8200/v1/auth/token/create
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "test/" must be unlocked to gain access.
```

- Namespace locking respects the hierarchy of the namespaces, meaning that you cannot lock a namespace that is already locked or is a descendant of already locked namespace
```
// namespaces: "test/", "test/test2", "test/test2/test3"
bao namespace lock test/test2
// success
bao namespace lock test/test2/test3
Error locking namespace: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/sys/namespaces/api-lock/lock/test/test2/test3
Code: 400. Errors:

* cannot lock namespace "test/test2/test3/" as "test/test2/" is already locked

// but namespace "test" as an ancestor can be locked.
bao namespace lock test
// success
```

- Namespaces inherit the locked status, meaning that if any namespace in a upwards hierarchy is locked, then all the descendants cannot serve any requests
```
bao write -ns=test/test2/test3 cubbyhole/foo bar=baz
Error writing data to cubbyhole/foo: Error making API request.

Namespace: test/test2/test3/
URL: PUT http://127.0.0.1:8200/v1/cubbyhole/foo
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "test/" must be unlocked to gain access.
// notice how the error mentions the locked namespace, and not the namespace of the request.
```

- Namespace locks are persisted through sealing of the bao instance
```
bao namespace lookup test
Key                Value
---                -----
custom_metadata    map[]
id                 IiZhx7
locked             true
path               test/
tainted            false
[...]

PR #1300, opened 2025-05-09
Namespaces locking unlocking implementation

### LINKED ISSUE #1181: Locking & Unlocking of Namespaces
To meet API expectations for projects migrating to OpenBao, we need to implement an API for locking/unlocking namespaces. The feature is an emergency mechanism to mitigate unexpected behavior and, therefore, crucial for secure multitenant deployments.

In a locked state, Openbao blocks all (authenticated) API endpoints from responding to clients operating in a locked namespace (or child-namespace of a locked namespace).

### Lock Namespace

Locks a given namespace and child namespaces, preventing responses to requests to the respective namespace.

Method | Path
-- | --
`POST` | `/sys/namespaces/api-lock/lock/:subpath`

Note: `:subpath` parameter is optional when `X-Vault-Namespace` header is set:

```
$ curl \
    --header "X-Vault-Namespace: some/path
    --request POST \
    http://127.0.0.1:8200/v1/sys/namespaces/api-lock/lock
```

#### Response

```json
{
    "unlock_key": "<unlock key for current/ns/path>"
}
```

### Unlock Namespace

This endpoint unlocks the api for the current namespace path or optional subpath.

Method | Path
-- | --
POST | /sys/namespaces/api-lock/unlock/:subpath

Note: `:subpath` parameter is optional when `X-Vault-Namespace` header is set:

```
curl \
    --request POST \
    --data @payload.json \
    http://127.0.0.1:8200/v1/sys/namespaces/api-lock/unlock
```

#### Sample Payload

```json
{
  "unlock_key": "<unlock key for current/ns/path>"
}
```

### API Endpoints that are excluded from lock:

* `sys/namespaces/api-lock/lock`
* `sys/namespaces/api-lock/unlock`
* `sys/health`
* (possibly others)

wslabosz-reply:
(recreated as a copy of #1298)

Resolves #1181

Implements `bao namespace lock <path>` and `bao namespace unlock <path>` operations.

- Locking the namespace results in all requests to that namespace being rejected with `503` error code and a error message
```
bao namespace create test   
bao namespace lock test   
Key           Value
---           -----
unlock_key   <unlock_key>
(...)
bao write -ns=test cubbyhole/foo bar=baz
Error writing data to cubbyhole/foo: Error making API request.

Namespace: test/
URL: PUT http://127.0.0.1:8200/v1/cubbyhole/foo
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "test/" must be unlocked to gain access.
(...)
bao token create -ns=test
Error creating token: Error making API request.

Namespace: test/
URL: POST http://127.0.0.1:8200/v1/auth/token/create
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "test/" must be unlocked to gain access.
```

- Namespace locking respects the hierarchy of the namespaces, meaning that you cannot lock a namespace that is already locked or is a descendant of already locked namespace
```
// namespaces: "test/", "test/test2", "test/test2/test3"
bao namespace lock test/test2
// success
bao namespace lock test/test2/test3
Error locking namespace: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/sys/namespaces/api-lock/lock/test/test2/test3
Code: 400. Errors:

* cannot lock namespace "test/test2/test3/" as "test/test2/" is already locked

// but namespace "test" as an ancestor can be locked.
bao namespace lock test
// success
```

- Namespaces inherit the locked status, meaning that if any namespace in a upwards hierarchy is locked, then all the descendants cannot serve any requests
```
bao write -ns=test/test2/test3 cubbyhole/foo bar=baz
Error writing data to cubbyhole/foo: Error making API request.

Namespace: test/test2/test3/
URL: PUT http://127.0.0.1:8200/v1/cubbyhole/foo
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "test/" must be unlocked to gain access.
// notice how the error mentions the locked namespace, and not the namespace of the request.
```

- Namespace locks are persisted through sealing of the bao instance
```
bao namespace lookup test
Key                Value
---                -----
custom_metadata    map[]
id                 IiZhx7
locked             true
path               test/
tainted            false
uuid               8f4959b7-7733-4f0c-5a01-22b010e9407b

bao operator seal
bao operator unseal

bao namespace lookup test
Key                Value
---                -----
custom_metadata    map[]
id                 IiZhx7
locked             true
path               test/
tainted            false
uuid               8f4959b7-7733-4f0c-5a01-22b010e9407b
```

- Unlocking can be done using unlock key from the locking operation response, or using the token with `sudo` privileges
```
bao namespace lock test
Key           Value
---           -----
unlock_key   <unlock_key>
(...)
bao namespace unlock -unlock-key=<key> test
bao namespace lookup test
Key                Value
---                -----
custom_metadata    map[]
id                 IiZhx7
locked             false
path               test/
tainted            false
uuid               8f4959b7-7733-4f0c-5a01-22b010e9407b
(...)
bao namespace lock test
bao namespace unlock test // assuming your token has sudo capabilities
// success
```

- Namespace hierarchy has to be respected during unlocking, meaning that you have to unlock the top-most namespace in a hierarchy of locked namespaces
```
// notice how we have to lock namespaces starting from the most deeply nested one
bao namespace lock test/test2/test3
bao namespace lock test/test2
bao namespace lock test
(...)
bao namespace unlock test/test2/test3
Error unlocking namespace: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/sys/namespaces/api-lock/unlock/test/test2/test3
Code: 500. Errors:

* 1 error occurred:
        * cannot unlock "test/test2/test3/" with namespace "test/" being locked


bao namespace unlock test/test2      
Error unlocking namespace: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/sys/namespaces/api-lock/unlock/test/test2
Code: 500. Errors:

* 1 error occurred:
        * cannot unlock "test/test2/" with namespace "test/" being locked


bao namespace unlock test
// success
bao namespace unlock test/test2
// success
bao namespace unlock test/test2/test3
// success
```

-----
cc: @cipherboy @satoqz @phyrog @voigt @driif

### DIFF
--- a/command/namespace_api_lock.go
+++ b/command/namespace_api_lock.go
@@ -86,10 +86,11 @@ func (c *NamespaceAPILockCommand) Run(args []string) int {
 		return 2
 	}
 
-	c.UI.Warn(fmt.Sprintf(`Unlock Key: %s
-	
+	c.UI.Output(fmt.Sprintf("Unlock Key: %s\n", resp.Data["unlock_key"]))
+	c.UI.Info(`
 Namespace is now locked. Please securely store the unlock key printed above.
 You must supply the unlock key to re-enable access to the namespace.
-If the unlock key is lost, sudo capability is required to unlock the namespace without it.`, resp.Data["unlock_key"]))
+If the unlock key is lost, sudo capability is required to unlock the namespace without it.`)
+
 	return 0
 }
--- a/command/namespace_api_lock.go
+++ b/command/namespace_api_lock.go
@@ -86,7 +86,7 @@ func (c *NamespaceAPILockCommand) Run(args []string) int {
 		return 2
 	}
 
-	c.UI.Output(fmt.Sprintf("Unlock Key: %s\n", resp.Data["unlock_key"]))
+	c.UI.Output(fmt.Sprintf("Unlock Key: %s", resp.Data["unlock_key"]))
 	c.UI.Info(`
 Namespace is now locked. Please securely store the unlock key printed above.
 You must supply the unlock key to re-enable access to the namespace.
--- a/command/namespace_api_lock.go
+++ b/command/namespace_api_lock.go
@@ -86,6 +86,10 @@ func (c *NamespaceAPILockCommand) Run(args []string) int {
 		return 2
 	}
 
-	c.UI.Warn("Copy and store your unlock key, as losing it may cause data loss.\nIn case key gets missing, contact the administrator.\n")
-	return OutputSecret(c.UI, resp)
+	c.UI.Warn(fmt.Sprintf(`Unlock Key: %s
+	
+Namespace is now locked. Please securely store the unlock key printed above.
+You must supply the unlock key to re-enable access to the namespace.
+If the unlock key is lost, sudo capability is required to unlock the namespace without it.`, resp.Data["unlock_key"]))
+	return 0
 }
--- a/command/namespace_api_lock.go
+++ b/command/namespace_api_lock.go
@@ -86,11 +86,16 @@ func (c *NamespaceAPILockCommand) Run(args []string) int {
 		return 2
 	}
 
-	c.UI.Output(fmt.Sprintf("Unlock Key: %s", resp.Data["unlock_key"]))
-	c.UI.Info(`
+	switch Format(c.UI) {
+	case "table":
+		c.UI.Output(fmt.Sprintf("Unlock Key: %s", resp.Data["unlock_key"]))
+		c.UI.Info(`
 Namespace is now locked. Please securely store the unlock key printed above.
 You must supply the unlock key to re-enable access to the namespace.
 If the unlock key is lost, sudo capability is required to unlock the namespace without it.`)
+	default:
+		return OutputData(c.UI, resp.Data)
+	}
 
 	return 0
 }
--- a/command/namespace_api_lock.go
+++ b/command/namespace_api_lock.go
@@ -86,5 +86,6 @@ func (c *NamespaceAPILockCommand) Run(args []string) int {
 		return 2
 	}
 
+	c.UI.Warn("Copy and store your unlock key, as losing it may cause data loss.\nIn case key gets missing, contact the administrator.\n")
 	return OutputSecret(c.UI, resp)
 }
--- a/command/namespace_api_unlock.go
+++ b/command/namespace_api_unlock.go
@@ -92,5 +92,6 @@ func (c *NamespaceAPIUnlockCommand) Run(args []string) int {
 		return 2
 	}
 
+	c.UI.Info("Namespace unlocked successfully!")
 	return 0
 }
--- a/command/namespace_api_unlock.go
+++ b/command/namespace_api_unlock.go
@@ -39,7 +39,7 @@ Usage: bao namespace unlock [options] PATH
 
 	Unlock a child namespace, and all of its descendants (e.g. ns1/ns2/):
 
-		$ bao namespace lock -unlock-key=<key> ns1/ns2
+		$ bao namespace unlock -unlock-key=<key> ns1/ns2
 
 ` + c.Flags().Help()
 
@@ -84,7 +84,7 @@ func (c *NamespaceAPIUnlockCommand) Run(args []string) int {
 		optionalChildNSPath = fmt.Sprintf("/%s", namespace.Canonicalize(args[0]))
 	}
 
-	_, err = client.Logical().Write(fmt.Sprintf("sys/namespaces/api-lock/unlock%s", optionalChildNSPath), map[string]interface{}{
+	secret, err := client.Logical().Write(fmt.Sprintf("sys/namespaces/api-lock/unlock%s", optionalChildNSPath), map[string]interface{}{
 		"unlock_key": c.flagUnlockKey,
 	})
 	if err != nil {
@@ -92,6 +92,11 @@ func (c *NamespaceAPIUnlockCommand) Run(args []string) int {
 		return 2
 	}
 
-	c.UI.Info("Namespace unlocked successfully!")
+	if secret != nil && len(secret.Warnings) > 0 {
+		c.UI.Warn(secret.Warnings[0])
+	} else {
+		c.UI.Info("Namespace unlocked successfully!")
+	}
+
 	return 0
 }
--- a/vault/logical_system_namespaces.go
+++ b/vault/logical_system_namespaces.go
@@ -121,8 +121,8 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
 				},
 			},
 
-			HelpSynopsis:    strings.TrimSpace(sysHelp["namespaces"][0]),
-			HelpDescription: strings.TrimSpace(sysHelp["namespaces"][1]),
+			HelpSynopsis:    strings.TrimSpace(sysHelp["namespaces-lock"][0]),
+			HelpDescription: strings.TrimSpace(sysHelp["namespaces-lock"][1]),
 		},
 
 		{
@@ -155,8 +155,8 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
 				},
 			},
 
-			HelpSynopsis:    strings.TrimSpace(sysHelp["namespaces"][0]),
-			HelpDescription: strings.TrimSpace(sysHelp["namespaces"][1]),
+			HelpSynopsis:    strings.TrimSpace(sysHelp["namespaces-unlock"][0]),
+			HelpDescription: strings.TrimSpace(sysHelp["namespaces-unlock"][1]),
 		},
 
 		{
--- a/vault/logical_system_namespaces.go
+++ b/vault/logical_system_namespaces.go
@@ -44,6 +44,16 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
 			Required:    true,
 			Description: "Path of the namespace.",
 		},
+		"tainted": {
+			Type:        framework.TypeBool,
+			Required:    true,
+			Description: "Flag representing the taint status of the namespace.",
+		},
+		"locked": {
+			Type:        framework.TypeBool,
+			Required:    true,
+			Description: "Flag representing the lock status of the namespace.",
+		},
 		"custom_metadata": {
 			Type:        framework.TypeMap,
 			Required:    true,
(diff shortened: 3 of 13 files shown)


wslabosz-reply:
@cipherboy 
Sorry for the mess, I've created second PR for this 😅, could you remove the tag from the #1298 and it here?

vault/logical_system_namespaces.go
```diff
@@ -80,6 +90,75 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
 			HelpDescription: strings.TrimSpace(sysHelp["list-namespaces"][1]),
 		},
 
+		{
+			Pattern: "namespaces/api-lock/lock(?P<path>.*)",
```
satoqz, line 94: This still matches e.g. `namespaces/api-lock/lockfoo`. I came up and tested with

```regex
namespaces/api-lock/lock($|/(?P<path>.+))
```

but maybe there's a simpler one.

vault/namespaces_store.go
```diff
@@ -755,3 +773,176 @@ func (ns *NamespaceStore) ResolveNamespaceFromRequest(nsHeader, reqPath string)
...
+		return err
+	}
+
+	if namespaceToUnlock == nil {
+		return errors.New("requested namespace does not exist")
+	}
+
+	if !namespaceToUnlock.Locked {
+		return fmt.Errorf("namespace %q is not locked", namespaceToUnlock.Path)
+	}
```
satoqz, line 815:
```suggestion

	if namespaceToUnlock.ID == namespace.RootNamespaceID {
		return fmt.Errorf("cannot unlock root namespace")
	}

	if !namespaceToUnlock.Locked {
		return fmt.Errorf("namespace %q is not locked", namespaceToUnlock.Path)
	}
```

else we get `* namespace "" is not locked` for a simple `bao namespace unlock`.

vault/logical_system_namespaces.go
wslabosz-reply: good catch!
I've tested with `(?:$|/(?P<path>.+))` now

vault/namespaces_store.go
wslabosz-reply: good catch

wslabosz-reply: @satoqz 
I wonder if we should respond with error like that here, it's not possible to lock `root` namespace, so trying to unlock it is completely invalid operation, and the response of `cannot unlock root namespace` could imply that it is locked and cannot be unlocked for some reason...

command/namespace_api_lock.go
```diff
@@ -86,5 +86,6 @@ func (c *NamespaceAPILockCommand) Run(args []string) int {
 		return 2
 	}
 
+	c.UI.Warn("Copy and store your unlock key, as losing it may cause data loss.\nIn case key gets missing, contact the administrator.\n")
```
satoqz, line 89: Inspired by the style of the `bao operator init` output:

```
Unlock Key: <unlock key>

Namespace <namespace> is now locked. Please securely store the unlock key printed above.
You must supply the unlock key to re-enable access to the namespace.
If the unlock key is lost, a root token is required to unlock the namespace without it.
```

...and then display the key as part of the message, not via the automatic table.

satoqz (approved):

vault/logical_system_namespaces.go
```diff
@@ -317,6 +397,50 @@ func (b *SystemBackend) handleNamespacesPatch() framework.OperationFunc {
...
+		te, err := b.Core.LookupToken(ctx, req.ClientToken)
+		if err != nil {
+			return nil, err
+		}
+
+		isRootRequest := te.IsRoot()
+		if unlockKey == "" && !isRootRequest {
+			return nil, errors.New("provided empty key")
+		}
+
```
voigt, line 434: Uh, this one just got me. I was wondering why I was able to unlock with a wrong unlock key.
While this is compliant with VE's unlock behavior, should we at least give a hint that unlock happened with root permissions/without unlock key?

voigt (approved):
Looks good to me. Works as intended.

One UX nit:
I kind of expected the lookup on the locked child namespace to work. Or other way around: why should a lookup on the locked namespace succeed, but not the one on of the consequently locked child namespace?
From my PoV it would be better UX if lookups always reveal namespace status info. In the case of ns1/ns1.1/ns1.1.1, maybe even with an info (third column?) about the cause of the lock (the cause ns1/ns1.1 being locked).

What I tried:

```bash
bao namespace create ns1
bao namespace create -ns ns1 ns1.1
bao namespace create -ns ns1/ns1.1 ns1.1.1

bao namespace lock ns1/ns1.1
Key           Value
---           -----
unlock_key    D1D3Vij6rzXvdYT4vlDdb1tj

bao namespace lookup -ns ns1 ns1.1
Key                Value
---                -----
custom_metadata    map[]
id                 0p25Jw
locked             true
path               ns1/ns1.1/
tainted            false
uuid               5db9bf5c-8ae9-c65c-c855-da5164b82251

bao namespace lookup -ns ns1/ns1.1 ns1.1.1
Error looking up namespace: Error making API request.

Namespace: ns1/ns1.1/
URL: GET http://127.0.0.1:8200/v1/sys/namespaces/ns1.1.1
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "ns1/ns1.1/" must be unlocked to gain access.
```

command/namespace_api_unlock.go
voigt, line 1: `s/lock/unlock/`

https://github.com/openbao/openbao/blob/7187ddf273ed66e2833a37281634a842a47a3847/command/namespace_api_unlock.go#L42

vault/logical_system_namespaces.go
satoqz: Two thoughts:

- I think this is a good use case for response warnings (need to make sure to also print them in custom CLI output, like the table formatter does)
- I would argue that an incorrect unlock key should fail the request even if the caller is root, and to only attempt unlocking via root capabilities if the unlock key is empty. That way you avoid accidentally unlocking a namespace you didn't mean to unlock - if you provide an unlock key and it's incorrect, there's a high chance you're attempting to unlock the wrong namespace. Callers using a root token might be oblivious to the fact that they can just unlock whatever without a key.

command/namespace_api_unlock.go
wslabosz-reply: good catch!

vault/logical_system_namespaces.go
wslabosz-reply: @voigt wouldn't notice your comment as this thread was resolved before it. I'll look into your proposal.

vault/namespaces_store.go
```diff
@@ -755,3 +773,180 @@ func (ns *NamespaceStore) ResolveNamespaceFromRequest(nsHeader, reqPath string)
...
+		return false
+	})
+	if lockedNS != nil {
+		return lockedNS.Clone()
+	}
+
+	return nil
+}
+
+// UnlockNamespace attempts to unlock the namespace with provided provided uuid.
```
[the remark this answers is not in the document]
wslabosz-reply: good catch!

vault/logical_system_namespaces.go
cipherboy: I like those ideas fwiw.

wslabosz-reply: @satoqz @voigt @cipherboy adjusted

satoqz (approved):

vault/logical_system_namespaces.go
```diff
@@ -317,6 +407,54 @@ func (b *SystemBackend) handleNamespacesPatch() framework.OperationFunc {
...
+	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+		path := namespace.Canonicalize(data.Get("path").(string))
+		unlockKey := data.Get("unlock_key").(string)
+
+		te, err := b.Core.LookupToken(ctx, req.ClientToken)
+		if err != nil {
+			return nil, err
+		}
+
+		if unlockKey == "" && !te.IsRoot() {
```
cipherboy, line 440: I am happy loosening this to a sudo check.

api/logical.go
```diff
@@ -325,8 +325,12 @@ func (c *Logical) Write(path string, data map[string]interface{}) (*Secret, erro
 
 func (c *Logical) WriteWithContext(ctx context.Context, path string, data map[string]interface{}) (*Secret, error) {
 	r := c.c.NewRequest(http.MethodPut, "/v1/"+path)
-	if err := r.SetJSONBody(data); err != nil {
-		return nil, err
+
+	// don't set stringified nil body on the request
```
cipherboy, line 329: I'm curious why this change was made? I don't think it'll have an impact, but it seems unrelated?

wslabosz-reply: @cipherboy this was caused because of a bug within `--output-curl-string` flag usage, as the PUT request of the `lock` operation does not contain a request body ([here](https://github.com/openbao/openbao/blob/48e7bd6f8358191056bdd0d2ac8b11e9e253254d/command/namespace_api_lock.go#L83)) the output curl string would take shape of 
```
$ curl -X PUT -H "X-Vault-Request: true" -H "X-Vault-Token: $(bao print token)" -d 'null' "http://127.0.0.1:8200/v1/sys/namespaces/api-lock/lock/test"
```
which is simply an invalid request

the other possible fix of that would be to change ([this line](https://github.com/openbao/openbao/blob/48e7bd6f8358191056bdd0d2ac8b11e9e253254d/command/namespace_api_lock.go#L83)) to `map[string]interface{}{}`, but I've chosen to add this if clause as it seems invalid to explicitly stringify a nil passed as a http request body.

cipherboy: Ahhhh, I see. No, that's fine. Do you mind opening it as a separate PR against `main`?

wslabosz-reply: No problem, just will do that tomorrow.

wslabosz-reply: @cipherboy 
https://github.com/openbao/openbao/pull/1315

vault/logical_system_namespaces.go
wslabosz-reply: adjusted

wslabosz-reply:
> Looks good to me. Works as intended.
> 
> One UX nit: I kind of expected the lookup on the locked child namespace to work. Or other way around: why should a lookup on the locked namespace succeed, but not the one on of the consequently locked child namespace? From my PoV it would be better UX if lookups always reveal namespace status info. In the case of ns1/ns1.1/ns1.1.1, maybe even with an info (third column?) about the cause of the lock (the cause ns1/ns1.1 being locked).
> 
> What I tried:
> 
> ```shell
> bao namespace create ns1
> bao namespace create -ns ns1 ns1.1
> bao namespace create -ns ns1/ns1.1 ns1.1.1
> 
> bao namespace lock ns1/ns1.1
> Key           Value
> ---           -----
> unlock_key    D1D3Vij6rzXvdYT4vlDdb1tj
> 
> bao namespace lookup -ns ns1 ns1.1
> Key                Value
> ---                -----
> custom_metadata    map[]
> id                 0p25Jw
> locked             true
> path               ns1/ns1.1/
> tainted            false
> uuid               5db9bf5c-8ae9-c65c-c855-da5164b82251
> 
> bao namespace lookup -ns ns1/ns1.1 ns1.1.1
> Error looking up namespace: Error making API request.
> 
> Namespace: ns1/ns1.1/
> URL: GET http://127.0.0.1:8200/v1/sys/namespaces/ns1.1.1
> Code: 503. Errors:
> 
> * API access to this namespace has been locked by an administrator - "ns1/ns1.1/" must be unlocked to gain access.
> ```

It makes sense to not lookup namespaces through locked namespace, this also matches the behavior of the upstream.

driif (approved):

voigt (approved):

wslabosz-reply:
@cipherboy 
should I open this against main now?

cipherboy:
@wslabosz-reply Sorry, I did not intentionally close this, it auto-closed when I deleted the `namespaces` branch. It looks like that's the only option!

Issue #1301, opened 2025-05-09
RFC - Support inline authentication for non-leased operations


cipherboy:
### Summary

Inline authentication allows direct requests to OpenBao to perform actions
which are authenticated by including login information with the action
request. This authentication occurs prior to the main request body being
processed, with the caveat that the generated token is not stored. This allows
for better scalability to future standby nodes which support read-request
processing, by avoiding write operations that occur in the course of regular
login requests.

### Problem Statement

Authentication is one of the few non-scalable requests in OpenBao; while
[theoretical work to allow per-namespace storage and active leader
designation](https://openbao.org/blog/vision-for-namespaces/) would allow
scaling, this still requires all login requests be processed by the active
leader. Certain access patterns, such as listing or reading from K/V or
issuing certificates without storing, are otherwise broadly horizontally
scalable. Furthermore, sometimes this authentication is discarded; such as
when fetching secrets from a CI/CD pipeline, each pipeline run will start
with a fresh token as usually authentication is tied to some parent JWT and
no authentication information or secrets are persisted across runs, usually
due to lack of persistent, secure storage and reproducibility.

Scaling login requests thus becomes important depending on the workload.

Furthermore, using the standard authentication flow incurs three storage
write operations:

Two to the token store:

- sys/token/id/{id}
- sys/token/accessor/{accessor}

and one to the expiration manager:

- sys/expire/id/auth/{mount_uuid}/login/{id}

As leadership changes in a cluster, the standby nodes read these entries
to check for expirations of tokens and thus incur storage pressure on
leadership changes. Switching to inline, writeless tokens will reduce this
pressure significantly, depending on workload.

This also helps integration for developers: rather than having to manage a
specific OpenBao identity or token, the authentication source can instead
be directly used on many requests and workflows. Confusion, such as expressed
by a few users on Matrix about authentication, can be avoided; an auth method
can be configured and "tokens" can be ignored in favor of inline
authentication.

We suggest calling the existing, persistent auth flow **standard**
authentication and the new flow **inline** authentication.

### User-facing Description

Inline authentication requires the user be able to modify headers on the
request; no other modification to the request body is necessary. The
`X-Vault-Inline-Auth-Path` header indicates that inline authentication is
to be performed; one or more headers beginning with the prefix
`X-Vault-Inline-Auth-Parameter-` specify parameters. These are URL-safe,
without padding, Base64 encoded JSON encoded maps of the form:

```
{
    "key": "<name of key>",
    "value": <value of key
}
[...]

PR #1302, opened 2025-05-09
Add RFC for inline authentication

cipherboy:
See also: https://github.com/openbao/openbao/issues/1301


website/content/docs/rfcs/inline-auth.mdx
```diff
@@ -0,0 +1,179 @@
...
+- Should `core.LocalUpdateUserFailedLoginInfo(...)` be skipped on inline
+  auth requests?
+
+## Related Issues
+
+- n/a
+
+## Proof of Concept
+
+https://github.com/cipherboy/openbao/pull/new/inline-authentication
```
wslabosz-reply, line 175: @cipherboy 
I don't think this is the right link (or is it a call to action?)

cipherboy: Ah, I kinda like these links :-) They show both the branch and the diff over current `main` of what this PoC entails. I would actually finish this and create a PR, of course, if there's sufficient approvals for the RFC.

Otherwise if you just do https://github.com/cipherboy/openbao/tree/inline-authentication, you can't easily see what the diff is, if I add multiple commits.

wslabosz-reply: oh yeah, because you have the diff below the PR description form... yeah. All's good then!

website/content/docs/rfcs/inline-auth.mdx
```diff
@@ -0,0 +1,179 @@
+---
+sidebar_label: Signed commits
```
cipherboy, line 2: Gotta fix this label.

website/content/docs/rfcs/inline-auth.mdx
```diff
@@ -0,0 +1,179 @@
...
+request's headers or authentication values are sufficiently static, this
+should work with third-party clients, especially those that expect to talk
+to Agent for transparent authentication.
+
+## Unresolved Questions
+
+- Is the header format sufficiently defensive?
+- Will header-size limiting become an issue for certain request types?
+- Not clear is if this will affect quotas or other subsystems.
+- Not clear if this should have a separate `X-Vault-Inline-Auth-Namespace`
```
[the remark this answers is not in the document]
satoqz: Regarding quota... if you want to then rate limit _both_ the auth route and the final API route, I guess that's one more reason to move quota evaluation mostly to core.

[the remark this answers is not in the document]
cipherboy: @satoqz Yeah, I think it is fine if we rate limit only one, though I do wonder if it is worth counting as two... :thinking: It is one inbound request that does twice the work, but I suppose we want counts to align with what an external LB would see.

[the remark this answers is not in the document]
cipherboy: @satoqz I pushed support for a namespace header to the PoC branch: https://github.com/cipherboy/openbao/tree/inline-authentication 

This behaves as follows:

- If no `X-Vault-Namespace` and no `X-Vault-Inline-Auth-Namespace`, both req and authReq are relative to root namespace.
- If `X-Vault-Namespace` and no `X-Vault-Inline-Auth-Namespace`, both req and authReq are relative to `X-Vault-Namespace`.
- If `X-Vault-Inline-Auth-Namespace`, authReq is relative to ``X-Vault-Inline-Auth-Namespace` and req behaves regularly.

LMKWYT and I'll update the RFC.

website/content/docs/rfcs/inline-auth.mdx
```diff
@@ -0,0 +1,179 @@
...
+- Should `core.LocalUpdateUserFailedLoginInfo(...)` be skipped on inline
+  auth requests?
+
+## Related Issues
+
+- https://gitlab.com/gitlab-org/gitlab/-/issues/540889
+
+## Proof of Concept
+
+https://github.com/cipherboy/openbao/pull/new/inline-authentication
```
cipherboy, line 175: The proof of concept had a bug, wherein token policies were not respected. I refactored this to behave better, using the cached TokenEntry value.

cipherboy: It also now invalidates leases and errs out accordingly, with a simple test of this functionality (policy behavior & lease revocation).

website/content/docs/rfcs/inline-auth.mdx
```diff
@@ -0,0 +1,179 @@
...
+HashiCorp's Vault API client. As long as the operator can modify each
+request's headers or authentication values are sufficiently static, this
+should work with third-party clients, especially those that expect to talk
+to Agent for transparent authentication.
+
+## Unresolved Questions
+
+- Is the header format sufficiently defensive?
+- Will header-size limiting become an issue for certain request types?
+- Not clear is if this will affect quotas or other subsystems.
```
cipherboy, line 163: I think quotas should behave like it was a single request; we're trading off load and we're not incuring more (inbound) network I/O. 

This is also consistent with how [profiles](https://openbao.org/blog/profiles/) would eventually behave: it is a single request to a complex endpoint.

Thoughts?

wslabosz-reply: makes sense for quotas to interpret it as a single request, otherwise we impose on ourselves to add additional documentation details which might be counter-intuitive to some users.

website/content/docs/rfcs/inline-auth.mdx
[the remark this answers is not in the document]
satoqz: Looks good to me! I think those are sane defaults and should be easy to explain to users if approach `X-Vault-Inline-Auth-Namespace` as an auth-specific override to `X-Vault-Namespace`.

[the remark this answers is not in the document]
cipherboy: @wslabosz-reply @satoqz I wonder if we really want this. Part of me thinks we can just use `X-Vault-Namespace` and use path prefixes on both the actual `req.Path` plus the inline auth path... Then it would only be relative to `X-Vault-Namespace` and we wouldn't have as much complexity around a second header.

Do we trim the namespace header out of the request flow? Or is context always resolved so far that it is past the point of the header by the time we call into `HandleRequest` and attempt inline authentication?

Thoughts?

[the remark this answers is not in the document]
satoqz: @cipherboy sorry for taking so long to get back on this!

I'm not sure that would really reduce complexity, both for us and the user. It irks me a bit that you would have to understand that you can put only the common namespace prefix shared between auth engine + dest path into `X-Vault-Namespace`, and that you have to potentially relevel your original request path to an upper namespace when switching to inline authentication. With the existing design you could just add the new headers, change nothing about the existing request, done.

> Do we trim the namespace header out of the request flow? Or is context always resolved so far that it is past the point of the header by the time we call into HandleRequest and attempt inline authentication?

Trimmed URL + Namespace + Context are resolved exactly once, in `switchedLockHandleRequest`.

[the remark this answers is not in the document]
cipherboy: @satoqz No, that's fair. I'll work on polishing this RFC and get the PR opened then!

cipherboy:
@wslabosz-reply @satoqz Thanks for the comments! RFC is now updated. :-)

website/content/docs/rfcs/inline-auth.mdx
```diff
@@ -0,0 +1,189 @@
...
+Using the standard authentication flow incurs three storage write operations:
+
+Two to the token store:
+
+- sys/token/id/{id}
+- sys/token/accessor/{accessor}
+
+and one to the expiration manager:
+
+- sys/expire/id/auth/{mount_uuid}/login/{id}
```
satoqz:
```suggestion
- sys/token/id/\{id\}
- sys/token/accessor/\{accessor\}

and one to the expiration manager:

- sys/expire/id/auth/\{mount_uuid\}/login/\{id\}
```

To fix the docs build. Yay, MDX.

cipherboy: Ah, we should probably put them in backticks, I'll do that so they're inline code blocks.

satoqz: Right, that works even better.

cipherboy: @satoqz Ok, this should be fixed now :-)

satoqz (approved):

cipherboy:
Thanks all!

Issue #1303, opened 2025-05-09
RFC - Automatic unseal using environment-provided keys


cipherboy:
### Summary

Certain environments lack KMS or HSM systems but otherwise have a backing
secure secret store. This store would usually be used for KMS or HSM access
credentials and so relying on it for storing a static, auto-unseal key is of
equivalent security but much improved operational experience.

### Problem Statement

In certain provisioning processes, secret material must be carefully accounted
for. Deployment mechanisms may not have direct access to create new secrets;
instead, they are explicitly created ahead of time and referenced by the
provisioner. Some environments, such as Kubernetes, lack widely-deployed KMS or
HSM solutions (though, when available, should be preferred) but have secure
password or secret storage. Allowing auto-unseal with rotation with statically
injected secrets is of equivalent security and thus should be natively supported
as a fully-automated alternative to Shamir's or any roundabout approach with
PKCS#11 and SoftHSM. Thus, to some extent, this approach already exists in two
different forms, they are just _complex_ to operate.

In particular, usage of a KMS or HSM in this environment would usually result
in storing some connection information (such as certificate, token, or
password) in this platform secret store and thus would be equivalent to a
static but rotatable key as discussed more clearly in the [security
implications](#security-implications) section below.

### User-facing Description

This option is meant to be used when other alternatives are not available,
but a trusted base platform exists and a fully auto-unseal-able solution
is required.

OpenBao would support a new auto-unseal mechanism, `static`, as follows:

```hcl
seal "static" {
    previous_key = "file:///volumes/secrets/last-unseal-key"
    previous_key_id = "33c9886e-7c93-4252-bf7c-8b15f4fd937c"
    current_key = "file:///volumes/secrets/current-unseal-key"
    current_key_id = "0c94ff77-42b6-4de2-b27a-8f7c166fb162"
}
```

This will [use](https://github.com/openbao/go-kms-wrapping/blob/1a1c5b0c79b112d9d08dadc53b4d98c7e5166397/options.go#L159-L181)
`go-kms-wrapping`'s existing [`parseutil.ParsePath`](https://pkg.go.dev/github.com/hashicorp/go-secure-stdlib/parseutil#ParsePath)
helpers, allowing the use of keys from either the environment (`env://`) or
the filesystem (`file://`), in addition to literal explicit keys in raw,
hex, or base64 encodings.

The key identifiers (`previous_key_id` and `current_key_id`) must
remain the same through the lifecycle of the corresponding key.

[...]

Issue #1021, opened 2025-02-20
RFC - Parallel Unseal of OpenBao


cipherboy:
### Summary

OpenBao needs the ability to unseal with one of many pieces of key material to truly solve multi-cloud deployment issues and to provide [outage recovery procedures](https://en.wikipedia.org/wiki/Business_continuity_planning). Setting up a recoverable, distributed system shouldn't rely on a single node having access to all key management systems and operators should have the option to manually recover using Shamir's keys if cloud services are not available.

To differentiate from upstream's solution which has different properties, we prefer to call our solution "Parallel Unseal" rather than "Multi-Unseal".

### Problem Statement

OpenBao's upstream has a history of [introducing](https://github.com/hashicorp/vault/pull/18683), [removing](https://github.com/hashicorp/vault/pull/18942), and subsequently [reintroducing](https://developer.hashicorp.com/vault/docs/configuration/seal/seal-ha) their multi-unseal feature in Vault Enterprise. The following [statement from Melissa Gurney Greene](https://github.com/hashicorp/vault/pull/18683#issuecomment-1425065124) on behalf of HashiCorp is quoted verbatim regrading this feature's history:

> During the development of this feature and corresponding enterprise enhancements, we determined that adoption of this version of the solution would create a potential conflict when migrating to enterprise or vice versa. For example, the original implementation was incompatible with seal wrap, so we withdrew this for a more holistic solution that does not have this inconsistency, which could lead to data loss in some scenarios if the seal wrapped values could no longer be decrypted. Currently we do not have any plans to bring this feature into OSS. We welcome any feedback or suggestions related to OSS seal recovery options due to non-availability of KMS.

From this we learn that the core problem was seal wrapping. While [Seal Wrap](https://developer.hashicorp.com/vault/docs/enterprise/sealwrap) is indeed an issue for Vault Enterprise, it is not present in OpenBao and thus gives us flexibility to choose an alternative solution. Furthermore, Vault Enterprise also has advanced clustering semantics we're unlikely to adopt, [Performance Secondary and Disaster Recovery](https://developer.hashicorp.com/vault/tutorials/enterprise/performance-replication) nodes. This poses a problem for Vault's Multi-Unseal as it introduces node complex topologies; however, they require the remote HSM(s) to be co-online at a single point in time to encrypt data for all nodes.

In designing an alternative for OpenBao, we'd like to consider the following requirements:

[...]

PR #1304, opened 2025-05-09
Add Static Auto-Unseal RFC

cipherboy:
See also: #1303

### DIFF
--- a/website/content/docs/rfcs/static-auto-unseal.mdx
+++ b/website/content/docs/rfcs/static-auto-unseal.mdx
@@ -1,0 +1,157 @@
+---
+sidebar_label: Static Auto-Unseal
+description: |-
+  An RFC for OpenBao detailing a new auto-unseal mechanism supporting static
+  keys loaded from the environment.
+---
+
+# Automatic unseal using environment-provided keys
+
+## Summary
+
+Certain environments lack KMS or HSM systems but otherwise have a backing
+secure secret store. This store would usually be used for KMS or HSM access
+credentials and so relying on it for storing a static, auto-unseal key is of
+equivalent security but much improved operational experience.
+
+## Problem Statement
+
+In certain provisioning processes, secret material must be carefully accounted
+for. Deployment mechanisms may not have direct access to create new secrets;
+instead, they are explicitly created ahead of time and referenced by the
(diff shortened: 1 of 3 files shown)


website/content/docs/rfcs/static-auto-unseal.mdx
```diff
@@ -0,0 +1,163 @@
...
+helpers, allowing the use of keys from either the environment (`env://`) or
+the filesystem (`file://`), in addition to literal explicit keys in raw,
+hex, or base64 encodings.
+
+The key identifiers (`previous_key_id` and `current_key_id`) must
+remain the same through the lifecycle of the corresponding key.
+
+Because key identifiers cannot be changed (as they are part of the
+configuration), keys are loaded in memory at startup and are not reloaded
+on SIGHUP currenlty. When [Parallel Unseal](https://github.com/openbao/openbao/issues/1021)
```
wslabosz-reply:
```suggestion
on SIGHUP currently. When [Parallel Unseal](https://github.com/openbao/openbao/issues/1021)
```

wslabosz-reply (commented):
Sound design doc, the only thing that I'd like for here to be mentioned, would be how does it interact (if at all) with following two RFCs:
- https://github.com/openbao/openbao/issues/1340
- https://github.com/openbao/openbao/issues/1021

cipherboy:
@wslabosz-reply I think this is independent of both of those. It should "just" be another auto-unseal mechanism for the purpose of those RFCs.

The one it is related to actually is #1320; we wouldn't use this as a true HSM/KMS thing. 

The other minor caveat for parallel unseal is that currently we can only do symmetric crypto, which means any future recommendation for asymmetric crypto for cross-node configuration wouldn't work, and static unseal would need the same keys on all systems. However, as discussed in #1425, we can have `algorithm` later and support asymmetric algorithms if we wish. So it would be a future improvement I can add. 

---

However, I think neither of those have landed in the RFC slot, so I guess conceptually, if I were to land this in the next week or two, I'd expect it not to reference unlanded RFCs and have those future RFCs reference this one? :thinking:

wslabosz-reply (approved):
@cipherboy 
Great, makes sense, thanks for clarification!

cipherboy:
Thanks all, merging!

PR #1291, opened 2025-05-02
fairshare job manager/worker: stop synchronously

satoqz:
Naive attempt at #1290. Turns out there was already a wait group used to synchronize stopping in tests.

I don't see any component that depends on asynchronous stopping, otherwise it would also be no issue to have both `stop`/`stopSync` methods.

I'm opening this against `namespaces` since that's where it causes panics (as of #1158), this should also be an issue on `main` but cause graceful errors at most, no `nil` pointer dereferences.

Resolves #1290

PR #1311, opened 2025-05-12
Namespaces: Test flake, deadlock, race condition cleanup

satoqz:
[namespace deletion: avoid outdated reads](https://github.com/openbao/openbao/pull/1311/commits/3b8663123f3ab88aa5b930cdfbefcc3eec453299) fixes an edge case where tests reached
https://github.com/openbao/openbao/blob/7e053be408876c51c35d54d1763f02828b50e8c1/vault/logical_system_namespaces.go#L353
on my machine :-)

[expiration manager: fix deadlock on seal](https://github.com/openbao/openbao/pull/1311/commits/f370e97c1973010c774fc2d809a250594ed73127) fixes https://github.com/openbao/openbao/actions/runs/14926238096/job/41931694374

[token store: wait for tidy before teardown](https://github.com/openbao/openbao/pull/1311/commits/047f2786ac59098390d0e3b1109d84a66475830b) fixes https://github.com/openbao/openbao/actions/runs/14969013713/job/42045556523

### DIFF
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -226,11 +226,13 @@ func (r *revocationJob) Execute() error {
 	default:
 	}
 
-	r.m.coreStateLock.RLock()
-	err := r.m.Revoke(revokeCtx, r.leaseID)
-	r.m.coreStateLock.RUnlock()
-
-	return err
+	// We call Revoke(...) without read-locking core's state lock. This prevents a deadlock if
+	// this job is still running while sealing begins and a write lock is acquired. As sealing
+	// synchronously ensures that this very job has finished before advancing any further, we
+	// inherently already hold the lock here. Assuming seal/unseal transitions are the only
+	// conditions where not holding a read lock on core is racy, we are safe to never
+	// lock here at all.
+	return r.m.Revoke(revokeCtx, r.leaseID)
 }
 
 func (r *revocationJob) OnFailure(err error) {
--- a/vault/logical_system_namespaces.go
+++ b/vault/logical_system_namespaces.go
@@ -326,30 +326,21 @@ func (b *SystemBackend) handleNamespacesDelete() framework.OperationFunc {
 			return nil, errors.New("path must not contain /")
 		}
 
-		ns, err := b.Core.namespaceStore.GetNamespaceByPath(ctx, path)
+		status, err := b.Core.namespaceStore.DeleteNamespace(ctx, path)
 		if err != nil {
-			return nil, fmt.Errorf("failed to load namespace: %w", err)
+			return handleError(err)
 		}
 
-		if ns == nil {
+		if status == "" {
 			resp := &logical.Response{}
 			resp.AddWarning("requested namespace does not exist")
 			return resp, nil
 		}
 
-		status, err := b.Core.namespaceStore.DeleteNamespace(ctx, ns.UUID)
-		if err != nil {
-			return handleError(err)
-		}
-
-		if status != "" {
-			return &logical.Response{
-				Data: map[string]interface{}{
-					"status": status,
-				},
-			}, nil
-		}
-
-		return nil, nil
+		return &logical.Response{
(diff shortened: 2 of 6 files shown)


vault/logical_system_namespaces.go
```diff
@@ -326,30 +326,21 @@ func (b *SystemBackend) handleNamespacesDelete() framework.OperationFunc {
...
-				},
-			}, nil
-		}
-
-		return nil, nil
+		return &logical.Response{
+			Data: map[string]interface{}{
+				"status": status,
+			},
+		}, nil
```
wslabosz-reply, line 344: great stuff actually.

vault/expiration.go
```diff
@@ -226,11 +226,7 @@ func (r *revocationJob) Execute() error {
 	default:
 	}
 
-	r.m.coreStateLock.RLock()
-	err := r.m.Revoke(revokeCtx, r.leaseID)
-	r.m.coreStateLock.RUnlock()
-
-	return err
+	return r.m.Revoke(revokeCtx, r.leaseID)
```
cipherboy, line 229: This is the one change I'm confused about. Don't we risk panics, if we call revoke without holding the state lock?

satoqz: Glad you asked! This is indeed the toughest one to justify, bear with me for a minute.

In #1291 we changed the expiration manager to block and finish all jobs during `(*Core).stopExpiration()`. Only then can the rest of sealing continue. This solved a nil pointer dereference we had before (Note: Not a data race, a plain nil pointer dereference!).

Said nil pointer dereference happened because:
- The sealing process holds core's state lock in write mode
- Revocation jobs continue to run asynchronously during and after sealing
- Revocation jobs try to acquire the core state lock in read mode! So, if a job is still running _while sealing is ongoing_ it will block here until sealing is entirely finished, and _only then_ resume the job! At this point fields like `core.systemBarrierView` and friends are `nil` of course, and the job reads into the void. This behavior turned into a panic only with the implementation of namespaces because of changes to storage view calculation.

Coming back, with #1291 we ensure that jobs all finish while sealing, so no more nil pointer dereference possible. However, we started getting a deadlock instead:
- The sealing process holds a write lock on core's state lock
- The sealing process is waiting for the expiration manager to shut down (via WaitGroup)
- Revocation jobs (and thereby the expiration manager) are trying to acquire core's state lock in read mode, essentially trying to wait until the sealing process is done D:

Ok, now why do you believe simply removing the read lock acquisition from the revocation job is a safe solution?

- Why do we acquire a read lock? Well, to avoid reading anything while someone else may have the write lock. If we were sure that either nobody has a write lock _or_ that we already inherently hold the write lock, we don't have to ever read lock.
- Who holds core's state lock in write mode, ever? Sealing & unsealing.

Unsealing: The expiration manager is not started until the resources it depends on (namespace store, barrier, `systemBarrierView`) are already "unsealed" and available. Reading from or even writing to those at this point is fine, even if unseal is still ongoing and holding the write lock.

Sealing: As the expiration manager now shuts down synchronously during sealing, it indirectly inherits the core state write lock from the unsealing process. No actual further writes are possible until the expiration manager has shut down, so revocation jobs may read as they wish.

phyrog: @satoqz Maybe add a comment about this in the code for the future 🙂

wslabosz-reply (approved):
looks good to me

vault/expiration.go
satoqz: > @satoqz Maybe add a comment about this in the code for the future 🙂

Added a comment

phyrog (approved):

vault/expiration.go
cipherboy: Really appreciate the explanation, thanks @satoqz!

cipherboy (approved):
Appreciate all the investigations here, thanks @satoqz!

PR #1312, opened 2025-05-12
Add JWT CEL Role RFC

suprjinx:
This RFC documents CEL usage for JWT/OIDC.

Relates #493

### DIFF
--- a/website/content/docs/rfcs/cel-jwt.mdx
+++ b/website/content/docs/rfcs/cel-jwt.mdx
@@ -2,21 +2,21 @@
 
 ## Summary
 
-This feature offers support for [Common Expression Language (CEL)](https://github.com/google/cel-spec) in Vault's jwt-based auth engines (oidc, jwt) to create a flexible framework for validating claims and dynamically assigning token policies. CEL will allow administrators to define fine-grained rules for claim validation and dynamic policy assignment, extending the more static assignment of role- and identity-based policies.
+This feature offers support for [Common Expression Language (CEL)](https://github.com/google/cel-spec) in OpenBao's jwt-based auth engines (oidc, jwt) to create a flexible framework for validating claims and dynamically assigning token policies. CEL will allow administrators to define fine-grained rules for claim validation and dynamic policy assignment, extending the more static assignment of role- and identity-based policies.
 
-JWT CEL support is implemented as a new type of CEL role and follows the [general plan for CEL support](https://github.com/openbao/openbao/blob/main/website/content/docs/rfcs/cel-best-practices.mdx) in Openbao 
+JWT CEL support is implemented as a new type of CEL role and follows the [general plan for CEL support](https://github.com/openbao/openbao/blob/main/website/content/docs/rfcs/cel-best-practices.mdx) in OpenBao.
 
 ---
 
 ## Problem Statement
 
-Vault's jwt-based auth engines (OIDC and JWT) currently assemble token policies from `roles` and `identities`. However, administrators cannot enforce complex constraints, such as validating combinations of claims or dynamically assigning token policies based on claim values.
+OpenBao's jwt-based auth engines (OIDC and JWT) currently assemble token policies from `roles` and `identities`. However, administrators cannot enforce complex constraints, such as validating combinations of claims or dynamically assigning token policies based on claim values.
 
 ---
 
 ## User-facing description
 
-**_Administrators_** can define CEL roles under a new endpoint, `auth/{oidc|jwt}/cel/roles/:name`, to be executed during authentication. CEL roles evaluate incoming claims and generate `logical.Auth{}` instances having appropriate token policies.
+**_Administrators_** can define CEL roles under a new endpoint, `auth/{oidc|jwt}/cel/roles/:name`, to be executed during authentication. CEL roles should evaluate all incoming claims and generate `logical.Auth{}` instances having appropriate token policies.
 
 **_Users_** authenticating with JWT or OIDC tokens will have their claims validated through the CEL inidicated role. If successful, they receive a token with policies and other attributes. 
 
@@ -25,13 +25,17 @@ For example:
 - A CEL role validates that the token contains specific claims like `"admin" in claims.roles && matches(claims.tenant_id, [0-9]+)"`.
 - If validation passes, the CEL program assigns the token policies `admin` and `team-12345` (using `claims.tenant_id` to form the policy name).
 
+This could be codified in the following CEL program:
+
+    "admin" in claims.roles && matches(claims.tenant_id, [0-9]+)" ? pb.Auth{ policies: ["admin", "team-12345"] } : false
+
 ---
 
 ## Technical Description
 
 ### CEL in Vault Auth Engines (OIDC, JWT)
 
-CEL roles integrate with Vault's Auth engine (OIDC and JWT) to dynamically evaluate claims. Instead of using static role-to-policy assignment with `bound_claim`, CEL programs validate claims and return a `logical.Auth` instance containing dynamically assigned token policies.
+CEL roles integrate with OpenBao's Auth engine (OIDC and JWT) to dynamically evaluate claims. Instead of using static role-to-policy assignment with `bound_claim`, CEL programs validate claims and return a `logical.Auth` instance containing dynamically assigned token policies.
 
 **Workflow:**
 1. A user authenticates with a JWT or OIDC token.
--- a/website/content/docs/rfcs/cel-jwt.mdx
+++ b/website/content/docs/rfcs/cel-jwt.mdx
@@ -26,8 +26,9 @@ For example:
 - If validation passes, the CEL program assigns the token policies `admin` and `team-12345` (using `claims.tenant_id` to form the policy name).
 
 This could be codified in the following CEL program:
-
-    "admin" in claims.roles && matches(claims.tenant_id, [0-9]+)" ? pb.Auth{ policies: ["admin", "team-12345"] } : false
+```json
+"admin" in claims.roles && matches(claims.tenant_id, [0-9]+)" ? pb.Auth{ policies: ["admin", "team-12345"] } : false
+```
 
 ---
 
--- a/website/content/docs/rfcs/cel-jwt.mdx
+++ b/website/content/docs/rfcs/cel-jwt.mdx
@@ -2,7 +2,9 @@
 
 ## Summary
 
-This feature proposes the use of [Common Expression Language (CEL)](https://github.com/google/cel-spec) in Vault's jwt-based auth engines (oidc, jwt) to create a flexible framework for validating claims and dynamically assigning token policies. CEL will allow administrators to define fine-grained rules for claim validation and dynamic policy assignment, extending the more static assignment of role- and identity-based policies.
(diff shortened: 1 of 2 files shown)


website/content/docs/rfcs/cel-jwt.mdx
```diff
@@ -0,0 +1,177 @@
+# RFC: CEL Expression Support for JWT Authentication
+
+## Summary
+
+This feature offers support for [Common Expression Language (CEL)](https://github.com/google/cel-spec) in Vault's jwt-based auth engines (oidc, jwt) to create a flexible framework for validating claims and dynamically assigning token policies. CEL will allow administrators to define fine-grained rules for claim validation and dynamic policy assignment, extending the more static assignment of role- and identity-based policies.
```
cipherboy, line 5:
```suggestion
This feature offers support for [Common Expression Language (CEL)](https://github.com/google/cel-spec) in OpenBao's jwt-based auth engines (oidc, jwt) to create a flexible framework for validating claims and dynamically assigning token policies. CEL will allow administrators to define fine-grained rules for claim validation and dynamic policy assignment, extending the more static assignment of role- and identity-based policies.
```

website/content/docs/rfcs/cel-jwt.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+JWT CEL support is implemented as a new type of CEL role and follows the [general plan for CEL support](https://github.com/openbao/openbao/blob/main/website/content/docs/rfcs/cel-best-practices.mdx) in Openbao
```
cipherboy, line 7:
```suggestion
JWT CEL support is implemented as a new type of CEL role and follows the [general plan for CEL support](https://github.com/openbao/openbao/blob/main/website/content/docs/rfcs/cel-best-practices.mdx) in OpenBao.
```

website/content/docs/rfcs/cel-jwt.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+This feature offers support for [Common Expression Language (CEL)](https://github.com/google/cel-spec) in Vault's jwt-based auth engines (oidc, jwt) to create a flexible framework for validating claims and dynamically assigning token policies. CEL will allow administrators to define fine-grained rules for claim validation and dynamic policy assignment, extending the more static assignment of role- and identity-based policies.
+
+JWT CEL support is implemented as a new type of CEL role and follows the [general plan for CEL support](https://github.com/openbao/openbao/blob/main/website/content/docs/rfcs/cel-best-practices.mdx) in Openbao 
+
+---
+
+## Problem Statement
+
+Vault's jwt-based auth engines (OIDC and JWT) currently assemble token policies from `roles` and `identities`. However, administrators cannot enforce complex constraints, such as validating combinations of claims or dynamically assigning token policies based on claim values.
```
cipherboy, line 13:
```suggestion
OpenBao's jwt-based auth engines (OIDC and JWT) currently assemble token policies from `roles` and `identities`. However, administrators cannot enforce complex constraints, such as validating combinations of claims or dynamically assigning token policies based on claim values.
```

website/content/docs/rfcs/cel-jwt.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+---
+
+## User-facing description
+
+**_Administrators_** can define CEL roles under a new endpoint, `auth/{oidc|jwt}/cel/roles/:name`, to be executed during authentication. CEL roles evaluate incoming claims and generate `logical.Auth{}` instances having appropriate token policies.
```
cipherboy, line 19:
```suggestion
**_Administrators_** can define CEL roles under a new endpoint, `auth/{oidc|jwt}/cel/roles/:name`, to be executed during authentication. CEL roles should evaluate all incoming claims and generate `logical.Auth{}` instances having appropriate token policies.
```

website/content/docs/rfcs/cel-jwt.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+**_Users_** authenticating with JWT or OIDC tokens will have their claims validated through the CEL inidicated role. If successful, they receive a token with policies and other attributes. 
+
+For example:
+
+- A CEL role validates that the token contains specific claims like `"admin" in claims.roles && matches(claims.tenant_id, [0-9]+)"`.
+- If validation passes, the CEL program assigns the token policies `admin` and `team-12345` (using `claims.tenant_id` to form the policy name).
+
```
cipherboy, line 27: Perhaps also give an example of such a program?

```suggestion

This could be codified in the following CEL program:

    "admin" in claims.roles && matches(claims.tenant_id, [0-9]+)" ? pb.Auth{ policies: ["admin", "team-12345"] } : false

```

website/content/docs/rfcs/cel-jwt.mdx
```diff
@@ -0,0 +1,177 @@
...
+---
+
+## Technical Description
+
+### CEL in Vault Auth Engines (OIDC, JWT)
+
+CEL roles integrate with Vault's Auth engine (OIDC and JWT) to dynamically evaluate claims. Instead of using static role-to-policy assignment with `bound_claim`, CEL programs validate claims and return a `logical.Auth` instance containing dynamically assigned token policies.
```
cipherboy, line 34:
```suggestion
CEL roles integrate with OpenBao's Auth engine (OIDC and JWT) to dynamically evaluate claims. Instead of using static role-to-policy assignment with `bound_claim`, CEL programs validate claims and return a `logical.Auth` instance containing dynamically assigned token policies.
```

cipherboy (commented):
Some minor nits and then I'll merge!

cipherboy (approved):
Thank you!

luke-clifton:
I have a couple of quick questions/comments:

1. With the standard way of specifying JWT roles, I can write a policy that restricts how a role is instantiated by using the `required` and `allowed` parameter feature of policies. Example, I want to allow our k8s team to create JWT mounts for each cluster they spin up, but I want to restrict exactly what they do (the most important one being, I don't want to allow them to assign arbitrary policies to a role). As I understand this RFC, we can't restrict what the CEL roles can do at a policy level?
2. Does the CEL role allow for the creation of entity-alias and the implicit entity associated with it, and if so, do we have control over the alias name?

cipherboy:
@luke-clifton

1. There's no easy way to restrict CEL role behavior at the moment. Putting CEL in the policy system would let you have more fined grain control over allowed properties but wouldn't solve this problem. This type of problem in general seems like it's suited to an alternative solution as `allowed_parameters` isn't a very flexible tool for policies in particular. You can (I believe) only specify exact values for `token_policies` and can't easily do subset or deny-list behaviors. Perhaps this warrants some auth mount tunable parameter that would let you restrict auth mounts to particular sets of policies. 
2. CEL for JWT gives you full control over the `logical.Auth` object returned by the auth method and thus yes, lets you set alias name. https://github.com/openbao/openbao/blob/dcba72cfdf0c738df78782e4899d68d0f2365175/sdk/logical/auth.go#L85

1 is maybe solvable by moving back to the earlier design for CEL, where each field in the CEL program output object was a separate parameter in the program, but I don't think this really is the case given it still is wrapped inside another parameter and the ACL system doesn't handle nested parameter values, just top-level parameter values. So the entire CEL system would need to be flattened, but custom variables can't easily be flattened given you need an explicit evaluation order between them. The reason we moved away from this is to allow us to bind to some lower level representation for objects (e.g., the protobuf representation of `logical.Auth`), allowing us to pick up changes in underlying types without having to add a new parameter.

luke-clifton:
Yep, I totally understand why it was done like this. But the "picking up changes" without having to add a new parameter is double edged. If some new change were to break some assumptions that the policy author had made, it could lead to compromise.

The "allowed parameters" model would allow the policy author to allow-list parameters that are considered safe, and any new parameters that are added later can be reviewed, and allowed, at some arbitrary point after their release. People are going to have to review changelogs (possibly even commits) very carefully without that safeguard in place.

I agree that the existing mechanisms are pretty terrible at dealing with nested data.

Perhaps there could be a mount, or even role, level validation script (itself a CEL script), which would take as input the output of the "untrusted" script, and validate it. (Your libraries proposal in the CEL best practices RFC would come in handy here...).

This way I could grant the k8s team a policy like (excuse my freehand policy writing here)

```
path "sys/auth/jwt" {
  capabilities = ["create", "update"]
  allowed_parameters = {
    "description" = [],
    "type" = "jwt",
    "auth_validation_script" = "<.. reference to a script stored somewhere, or an inline script ..>",
   ... other restrictions ...
  }
}

path "auth/jwt/role/+" {
  capabilities = ["create", "update"]
}
```

And they can go ahead and create the mount and roles, but the validation script will catch any disallowed behaviour.

The main downside I see with this is that the validation happens later, so there may be some wasted work if there are a lot of requests that get denied by the validation script.

As a bonus, the `auth_validation_script` could apply to ALL authentication this mount does, including regular old-style roles. This could allow centralisation of policy, as well as more interesting policies being able to be applied to existing roles.

cipherboy:
@luke-clifton Thank you again for all the details and the well-thought out responses. :-) 

Philosophically, I don't know that we can ever really defend partial access to any of policies, auth mounts, or the identity system (arguably, the plugin catalog as well -- since auth plugins can have arbitrary control over entities or policies as we see here). There's too many ways, and not precise enough control over resources, that one could escalate to a higher set of permissions... I'd argue that's true even if we add more advanced CEL capabilities here to the policy engine (whether embedded in ACLs as discussed here or otherwise).

I agree, that any validation would have to be on the auth response from the plugin -- because simulating a full CEL role would be nearly impossible, to understand what possible outputs it may have, even though CEL is not Turing complete -- but would any extra validation be allowed to mutate that response, or simply reject it? Arguably we'd need to give some feedback to Auth admins (whether writing traditional roles or CEL roles) that their configuration does not comply with organization policy (as set by this validation program author).

---

The ways I've seen to address this in organizations in the wild are two-fold:

1. Either by using namespaces, and agreeing that they fall into a broad admin category within the namespace. Each k8s cluster would thus be in a new namespace. Secrets would have to be structured accordingly.
2. By using a trusted, external provisioning system. Users would not directly have mount + role creation policies, but these would be managed in e.g., Terraform or a third-party application.

The second pattern in particular solves this rather well in a number of organizations, but does admittedly require you have a separate provisioning service.

---

Back to the topic at hand, I think I realized splitting this out has no net impact; we'll _always_ have to have some default here for missing fields, so that programs can continue to work on upgrade. Largely this default will have to preserve backwards compatibility.
 
Unless the community feels strongly here (especially cc @suprjinx and @fatima2003 for your thoughts), I think leaving this as-is is thus _roughly_ equivalent to the per-field separation (minus this validation component discussed above) and thus perhaps not worth the effort of splitting it out, versus the benefits of a common CEL UX.

Thoughts?

suprjinx:
> 
> Unless the community feels strongly here (especially cc @suprjinx and @fatima2003 for your thoughts), I think leaving this as-is is thus _roughly_ equivalent to the per-field separation (minus this validation component discussed above) and thus perhaps not worth the effort of splitting it out, versus the benefits of a common CEL UX.
> 
> Thoughts?

There's definitely a risk here that a CEL role author could grant any policy in the CEL expression. Maybe we could add a "permitted-policies" attribute to the auth mount, or the like, but I do think namespaces are the best mitigation -- then each k8s cluster could operate with a sandboxed namespace.

PR #10, opened 2023-12-08
Replace Vault with OpenBao

jmls:
replaced a number of occurrences ( where it makes sense ) of Vault with OpenBao

PR #1313, opened 2025-05-12
Add OSPS to policies

xskrasek:
I just added the page to the sidebar. I think, that it is ok to stay in the policies tree for now. I would only create dedicated tree in case the documentation for OSPS grows by a lot. 
Resolves [#10](https://github.com/openbao/dev-wg/issues/10)


cipherboy (approved):

cipherboy:
Thank you @xskrasek !

PR #1314, opened 2025-05-13
Fix various govet warnings

phyrog:
Fixes:
- unreachable code
- ctxCancel function is not used on all paths (possible context leak)
- malformed struct field tag
- struct literal uses unkeyes fields
- discarded ctx cancel functions

There are still a lot of warnings left after this that might need some closer look:
- calling (*testing.T).Fatal from non-test goroutine
- copying lock values (at least most of these should be ok, the lock is recreated after the copy)

### DIFF
--- a/command/secrets_move.go
+++ b/command/secrets_move.go
@@ -126,6 +126,4 @@ func (c *SecretsMoveCommand) Run(args []string) int {
 		c.UI.Output(fmt.Sprintf("Waiting for terminal status in migration of secrets engine %s to %s, with migration ID %s", source, destination, remountResp.MigrationID))
 		time.Sleep(10 * time.Second)
 	}
-
-	return 0
 }
--- a/sdk/logical/response_util.go
+++ b/sdk/logical/response_util.go
@@ -203,7 +203,7 @@ func RespondErrorAndData(w http.ResponseWriter, status int, data interface{}, er
 
 	type ErrorAndDataResponse struct {
 		Errors []string    `json:"errors"`
-		Data   interface{} `json:"data""`
+		Data   interface{} `json:"data"`
 	}
 	resp := &ErrorAndDataResponse{Errors: make([]string, 0, 1)}
 	if err != nil {
--- a/vault/logical_system_raft.go
+++ b/vault/logical_system_raft.go
@@ -695,6 +695,11 @@ func (b *SystemBackend) handleStorageRaftSnapshotWrite(force bool) framework.Ope
 			}()
 
 			ctx, ctxCancel := context.WithCancel(namespace.RootContext(nil))
+			defer func() {
+				if retErr != nil {
+					ctxCancel()
+				}
+			}()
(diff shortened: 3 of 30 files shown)


sdk/logical/response_util.go
```diff
@@ -203,7 +203,7 @@ func RespondErrorAndData(w http.ResponseWriter, status int, data interface{}, er
 
 	type ErrorAndDataResponse struct {
 		Errors []string    `json:"errors"`
-		Data   interface{} `json:"data""`
```
cipherboy, line 206: Was this a problem? Or did Go silently ignore it?

vault/logical_system_raft.go
```diff
@@ -707,18 +707,21 @@ func (b *SystemBackend) handleStorageRaftSnapshotWrite(force bool) framework.Ope
 			b.Core.logger.Info("shutting down prior to restoring snapshot")
 			if err := b.Core.preSeal(); err != nil {
 				b.Core.logger.Error("raft snapshot restore failed preSeal", "error", err)
+				ctxCancel()
```
cipherboy, line 710: Should we just defer the `ctxCancel`? :thinking:

cipherboy (approved):
One comment of substance. Otherwise, I'm happy to merge after namespaces lands, as I'm cautious about conflicts at this point in the release cycle.

vault/logical_system_raft.go
phyrog: Yes, but: the context should only be cancelled when there is an error. Otherwise it and the cancel function will be saved in `*Core`. I'll update the code to include that.

sdk/logical/response_util.go
phyrog: It was silently ignored: https://go.dev/play/p/aCrsHnHtj18

vault/logical_system_raft.go
phyrog: Fixed

command/secrets_move.go
```diff
@@ -126,6 +126,4 @@ func (c *SecretsMoveCommand) Run(args []string) int {
 		c.UI.Output(fmt.Sprintf("Waiting for terminal status in migration of secrets engine %s to %s, with migration ID %s", source, destination, remountResp.MigrationID))
 		time.Sleep(10 * time.Second)
 	}
-
-	return 0
```
wslabosz-reply, line 130: I might not understand this removal, I think that this path might execute, so the reasonable thing is to return something from a function with `int` return signature, and If I understand it correctly `0` is not the implicit returned value (it might be `1`)

wslabosz-reply: nvm, it's in loop, so it never executes it, resolving myself.

wslabosz-reply (approved):
Looking as all tests pass, I'm happy how it looks.

cipherboy:
Thank you @phyrog!

PR #1315, opened 2025-05-13
Fix marshaling and setting nil request body

wslabosz-reply:
Caused by discussion on https://github.com/openbao/openbao/pull/1300#discussion_r2084843533

## Issue:
whenever there's explicit `nil` data provided to either `WriteWithContext` or `JSONMergePatch`, `(*request).setJSONBody` would marshal and set the marshaled data as a request body. This introduces a bug in a `output-curl-string` flag usage, as it creates invalid requests with a body of `-d 'null'`

## Fix:
Check for `nil` data in `SetJSONBody` method and return immediately from function if it's encountered.

In the PR also included removal of unused function parameter `path` in `(*Logical).write` method.

### DIFF
--- a/api/request.go
+++ b/api/request.go
@@ -42,6 +42,10 @@ type Request struct {
 
 // SetJSONBody is used to set a request body that is a JSON-encoded value.
 func (r *Request) SetJSONBody(val interface{}) error {
+	if val == nil {
+		return nil
+	}
+
 	buf, err := json.Marshal(val)
 	if err != nil {
 		return err
--- a/api/logical.go
+++ b/api/logical.go
@@ -329,7 +329,7 @@ func (c *Logical) WriteWithContext(ctx context.Context, path string, data map[st
 		return nil, err
 	}
 
-	return c.write(ctx, path, r)
+	return c.write(ctx, r)
 }
 
 func (c *Logical) JSONMergePatch(ctx context.Context, path string, data map[string]interface{}) (*Secret, error) {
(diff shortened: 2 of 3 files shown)


wslabosz-reply:
@cipherboy 
what's the approach to the go-vet error output of
```api/logical.go:356:1: Function write can return an error, and has a statement that returns only nils```
here?

should I change the
```
switch parseErr {
case nil:
case io.EOF:
	return nil, nil
default:
	return nil, parseErr
}
```
to 
```
if parseErr != nil {
	return nil, parseErr
}
```

**Comment by satoqz** [comment]
> what's the approach to the go-vet error output of `api/logical.go:356:1: Function write can return an error, and has a statement that returns only nils` here?

Existing code seems valid to me tbh, neither would there be additional value ensuring `write()` never returns `(nil, nil)` as `ParseSecret()` also has a couple paths that may return `(nil, nil)` anyways.

**Review by satoqz — APPROVED** [review]

**Comment by cipherboy** [comment]
I think go-vet is just confused; write isn't an implementation of `io.Writer` but instead a `logical.Operation` verb (http-verb-esque).

File: api/request.go [hunk 1]
```diff
@@ -42,6 +42,10 @@ type Request struct {
 
 // SetJSONBody is used to set a request body that is a JSON-encoded value.
 func (r *Request) SetJSONBody(val interface{}) error {
+	if val == nil {
```
cipherboy, line 45: Since this changes the behavior, do you mind [adding a changelog entry](https://github.com/openbao/openbao/tree/main/changelog/README.md)? Thanks!

cipherboy (approved):

api/request.go
wslabosz-reply: added!

PR #1317, opened 2025-05-13
Hierarchical namespace storage

### LINKED ISSUE #1316: Hierarchical namespace entry storage
With the introduction of namespaces we've opted for centralized logical storage structure located in root (`sys/raw/core/namespaces`). Currently it is visible that this decision is not in line with pattern we've introduced within logical structure of system backends of the namespaces. This also complicates things when it comes to non-hierarchical namespaces.

Alex's comments about the motivation for this change:
> I think we should actually split this so child namespaces' storage references are in the parent, not root. We need another recursion here, but that way if we have an external import of a namespace ([see vision](https://openbao.org/blog/vision-for-namespaces/)), you get children for free without having to explicitly register it. My 2c.

also

> I think this is necessary for out-of-hierarchy namespaces (parent-less namespaces)

Proposed solution would be to change the logical structure of the namespaces storage to be nested inside a parent namespace. 

Having three namespaces of: "ns1/ns2/ns3" order:
```
core/
    ...
    namespaces/
        <ns1_uuid>
namespaces/
    <ns1_uuid>/
        core/
        ...
        namespaces/
           <ns2_uuid>
        ...
    <ns2_uuid>/
        core/
        ...
        namespaces/
           <ns3_uuid>
        ...
    <ns3_uuid>/
        core/
        ...
       // ns3 doesn't have any child namespaces
```

An alternative we've pursed till now is the flat structure of all namespaces in the root, this does not support fully use-cases we want to implement in the future, and change in the storage structure would introduce unnecessary complexities and require migration.

Non-hierarchical namespaces, so the namespaces that do not have a "parent" namespace. This would require us to revisit the naming convention, as all non-hierarchical namespaces would be a root in itself, which would require some kind of "root of all roots" entity.

**Additional context**
none

satoqz:
Resolves #1316

### DIFF
--- a/vault/namespaces_store.go
+++ b/vault/namespaces_store.go
@@ -119,11 +119,18 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
 	namespacesByAccessor[namespace.RootNamespaceID] = namespace.RootNamespace
 
 	if err := logical.WithTransaction(ctx, ns.storage, func(s logical.Storage) error {
-		return ns.loadNamespacesRecursive(ctx, s, NewBarrierView(s, namespaceStoreSubPath), func(namespace *namespace.Namespace) {
-			namespacesByPath.unsafeInsert(namespace)
-			namespacesByUUID[namespace.UUID] = namespace
-			namespacesByAccessor[namespace.ID] = namespace
-		})
+		return ns.loadNamespacesRecursive(ctx, s, NewBarrierView(s, namespaceStoreSubPath),
+			func(namespace *namespace.Namespace) error {
+				if _, ok := namespacesByUUID[namespace.UUID]; ok {
+					return fmt.Errorf("namespace with UUID %q is not unique in storage", namespace.UUID)
+				}
+				if err := namespacesByPath.Insert(namespace); err != nil {
+					return err
+				}
+				namespacesByUUID[namespace.UUID] = namespace
+				namespacesByAccessor[namespace.ID] = namespace
+				return nil
+			})
 	}); err != nil {
 		return err
 	}
@@ -143,7 +150,7 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
 // loadNamespacesRecursive reads all child namespaces from a given namespace store storage view
 // from disk, recursing into the namespace storage of any discovered namespaces to load an entire
 // namespace tree.
-func (ns *NamespaceStore) loadNamespacesRecursive(ctx context.Context, barrier, view logical.Storage, callback func(*namespace.Namespace)) error {
+func (ns *NamespaceStore) loadNamespacesRecursive(ctx context.Context, barrier, view logical.Storage, callback func(*namespace.Namespace) error) error {
 	return logical.HandleListPage(view, "", 100, func(page int, index int, entry string) (bool, error) {
 		item, err := view.Get(ctx, entry)
 		if err != nil {
@@ -159,7 +166,11 @@ func (ns *NamespaceStore) loadNamespacesRecursive(ctx context.Context, barrier, 
 			return false, fmt.Errorf("failed to decode namespace %v (page %v / index %v): %w", entry, page, index, err)
 		}
 
-		callback(&namespace)
+		err = callback(&namespace)
+		if err != nil {
+			return false, err
+		}
+
 		err = ns.loadNamespacesRecursive(ctx, barrier, NamespaceView(barrier, &namespace).SubView(namespaceStoreSubPath), callback)
 		if err != nil {
 			return false, err
--- a/vault/namespaces_store.go
+++ b/vault/namespaces_store.go
@@ -149,9 +149,9 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


cipherboy (approved):
Thank you @satoqz!

vault/namespaces_store.go
```diff
@@ -163,6 +140,35 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
...
+			return false, fmt.Errorf("%v has an empty namespace definition (page %v / index %v)", entry, page, index)
+		}
+
+		var namespace namespace.Namespace
+		if err := item.DecodeJSON(&namespace); err != nil {
+			return false, fmt.Errorf("failed to decode namespace %v (page %v / index %v): %w", entry, page, index, err)
+		}
+
+		callback(&namespace)
+		err = ns.loadNamespacesRecursive(ctx, barrier, NamespaceView(barrier, &namespace).SubView(namespaceStoreSubPath), callback)
```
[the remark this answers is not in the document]
phyrog: After some offline discussions, I think we can leave it like this and revisit later if necessary.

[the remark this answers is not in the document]
satoqz: Discussed the tradeoffs with @phyrog and we've chosen to keep this recursive for now, with an iterative approach it would be a lot more complex to achieve the same efficiency of reading pages as needed instead of buffering them.

vault/namespaces_store.go
```diff
@@ -163,6 +149,41 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
 	return nil
 }
 
+// loadNamespacesRecursive reads all child namespaces from a given namespace store storage view
```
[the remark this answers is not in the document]
satoqz: I think it's important to differentiate between "namespace storage view" and "namespace store view" as those are different, but yes, mostly agree, updating.

phyrog (approved):

wslabosz-reply (approved):
approving, although I'm co-contributor

driif (approved):

vault/namespaces_store.go
```diff
@@ -163,6 +149,41 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
...
+		}
+
+		if err := callback(&namespace); err != nil {
+			return false, err
+		}
+
+		childView := NamespaceView(barrier, &namespace).SubView(namespaceStoreSubPath)
+		if err := ns.loadNamespacesRecursive(ctx, barrier, childView, callback); err != nil {
+			return false, err
+		}
```
cipherboy, line 181: I think it is fine for now, but in the context of https://github.com/openbao/openbao/pull/1102, this might consume endless transactions. Using an iterative approach in the future would be better, but it shouldn't matter unless you have >64 _nested_ namespaces.

vault/namespaces_store.go
```diff
@@ -119,34 +118,21 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
...
-			namespacesByUUID[namespace.UUID] = &namespace
-			namespacesByAccessor[namespace.ID] = &namespace
-
-			return true, nil
-		}, nil); err != nil {
+	loadingCallback := func(namespace *namespace.Namespace) error {
+		if _, ok := namespacesByUUID[namespace.UUID]; ok {
+			return fmt.Errorf("namespace with UUID %q is not unique in storage", namespace.UUID)
+		}
+		if err := namespacesByPath.Insert(namespace); err != nil {
```
cipherboy, line 125: Not needing unsafe insert is nice.

cipherboy:
Thank you @satoqz!

vault/namespaces_store.go
satoqz: Can you elaborate the constraints here a bit further? Very curious.

cipherboy: Ah, my bad, `HandleListPage` is not transactional; we probably should update it to be optionally transactional.

At any rate, transactions are subject to limits: 

https://github.com/openbao/openbao/blob/cc0df5ca5853e829bcb22bfbb68212472bc0d7a1/sdk/physical/physical.go#L15

So if you are inside a transaction, (one held by `HandleListPage` in the future), and recurse to call `HandleListPage` again from the _base_ object (thereby creating another transaction -- which will be necessary if/when namespaces live on separate storage engines), if you keep hitting the same storage engine, you'll be consuming transactions here, up to the depth of the recursion.

This will eventually deadlock as the permit pool acquisition will stall, causing issues.

satoqz: Makes sense. FWIW all the `HandleListPage` calls here share the same transaction started from `loadNamespacesLocked`.

> HandleListPage is not transactional

Do you mean `HandleListPage` doesn't start a transaction by itself and needs to be wrapped in one, or will ignore the transaction it is part of somehow? Do we anticipate the need to split namespace loading into several transactions? I guess that would need to happen with per-namespace physical backends, but for another reason?

cipherboy: > Do you mean HandleListPage doesn't start a transaction by itself and needs to be wrapped in one

The first part, yes.

> I guess that would need to happen with per-namespace physical backends, but for another reason?

No other reasons than the before, which will be a big change in and of itself, so we'll get there when we get there. Not important for today, just things on my mind.

PR #1318, opened 2025-05-13
Bump dependencies

cipherboy:
These seem mostly to have been missed by dependabot, though I know it doesn't do major version upgrades.


cipherboy:
Other than another certificate expiring... This fixes the other failure:

```diff
diff --git a/builtin/logical/database/backend_test.go b/builtin/logical/database/backend_test.go
index 3353ed22c8..0edf585350 100644
--- a/builtin/logical/database/backend_test.go
+++ b/builtin/logical/database/backend_test.go
@@ -350,7 +350,7 @@ func TestBackend_BadConnectionString(t *testing.T) {
                }
                err = resp.Error()
                if strings.Contains(err.Error(), "localhost") {
-                       t.Fatal("error should not contain connection info")
+                       t.Fatalf("error should not contain connection info: %#v", err)
                }
        }
 
diff --git a/builtin/logical/database/path_config_connection.go b/builtin/logical/database/path_config_connection.go
index e7505db6e4..fd8041820f 100644
--- a/builtin/logical/database/path_config_connection.go
+++ b/builtin/logical/database/path_config_connection.go
@@ -465,7 +465,8 @@ func (b *databaseBackend) connectionWriteHandler() framework.OperationFunc {
                initResp, err := dbw.Initialize(ctx, initReq)
                if err != nil {
                        dbw.Close()
-                       return logical.ErrorResponse("error creating database object: %s", err), nil
+                       b.Logger().Error("error creating database object", "err", err)
+                       return logical.ErrorResponse("error creating database object; check server logs for information"), nil
                }
                config.ConnectionDetails = initResp.Config
 
```



While maybe we shouldn't be _logging_ connection parse failures, we do contradict a rather explicit test failure. Not sure what our best course of action is here. Thoughts? 

cc: @openbao/openbao-org-maintainers

cipherboy:
@JanMa Not quite; PGXv4 also threw an parse error here. There's different root cause. I'll post an updated patch:

```
func (mw *DatabaseErrorSanitizerMiddleware) sanitize(err error) error {
    if err == nil {
        return nil
    }
    if errwrap.ContainsType(err, new(url.Error)) {
        return errors.New("unable to parse connection url")
    }
```

no longer is true in PGXv5. I'll see if I can fix it.

cipherboy:
@JanMa Hmm, or not. That also seems not to be the full root cause.

The intent of the test _is_ to cause a parse failure, though, and ensure that the message is redacted.

cipherboy:
Ah, here we go:


```
  err=
  | unable to initialize: rpc error: code = Internal desc = failed to initialize: 
  | 
  | ALEXWASHERE Walking error: error verifying connection: cannot parse `postgresql://:xxxxxx@[localhost?timezone=UTC&application_name=vault`: failed to parse as URL (missing ']' in host)
  | \tgot err (*errors.errorString): error verifying connection: cannot parse `postgresql://:xxxxxx@[localhost?timezone=UTC&application_name=vault`: failed to parse as URL (missing ']' in host)
  | \tgot err (*pgconn.ParseConfigError): cannot parse `postgresql://:xxxxxx@[localhost?timezone=UTC&application_name=vault`: failed to parse as URL (missing ']' in host)
  | \tgot err (*errors.errorString): missing ']' in host
   took="777.254µs"
```

So it changed from a wrapped `url.Error` to `pgconn.ParseConfigError`, which is unfortunate.

cipherboy:
@JanMa Ok, this should be good now.

JanMa (approved):
LGTM

cipherboy:
Thanks @JanMa!

Issue #1319, opened 2025-05-14
RFC - External Key Configuration for KMS and HSM Access


cipherboy:
### Summary

For better security of cryptographic material, keys within OpenBao should
optionally be backed by a KMS or HSM solution. We design a per-namespace
repository of seal configurations, with mount-specific key information to
prevent cross-mount usage of keys except via explicit reuse.

We call this solution External Keys to differentiate it from HashiCorp
Vault Enterprise's Managed Keys.

### Problem Statement

Upstream HashiCorp Vault Enterprise implemented Managed Keys, which have
the following properties:

1. KMS library configurations live in storage and apply to any namespace;
   this takes a `type` and a `name` in a `kms_library` stanza. This sets
   global configuration for the provider; per
   [documentation](https://developer.hashicorp.com/vault/docs/configuration/kms-library)
   this is only a PKCS#11 library configuration as it is the only type that
   requires elements in storage.

2. Within each namespace, a repository of keys accessible only to that
   namespace exist.

  - Each `type` of key can be listed (via  `LIST /sys/managed-keys/<:type>`),
    to show which keys with the particular type exist).
  - Creating a key sets its name (`POST /sys/managed-keys/<:type>/<:key-name>`),
    the name of the library (if PKCS#11), and all credential information.

  In particular, this means that there is no central repository (within a
  namespace) of credentials, making rotation harder.

Further, while less restrictive, any namespace can use any `kms_library` or
seal type; a namespace admin has full control over policies and thus a global
operator cannot restrict types.

### User-facing Description

We suggest the following design for external keys:

1. The `external_keys` stanzas take an optional `namespaces = []` parameter,
   taking items of the form `uuid:<ns-uuid>`, `id:<ns-accessor>`, or
   `path:<ns-path>`, to identify namespaces that this library is allowed to
   be used in. This prevents cross-tenant library usage problems by preventing
   them from using the underlying library.

   - In the future, when support for pluginized KMS integrations are added, we
     can use an `external_keys "binary" { ... }` stanza to limit access to
     specific external libraries or hosts that they run on.

   - The `name` of this stanza gets registered as the library type; it must
     not conflict with any existing type and `pkcs11` will not be a valid
     type, instead requiring a named configuration. E.g., `gcpckms`, `thales`,
     `entrust`, and `securosys` might all be valid `types` for the endpoints
     below.

2. Within `sys/namespaces/<:path>`, we add a new option, `types`, which limits
[...]

PR #1320, opened 2025-05-14
Add RFC on External Keys

cipherboy:
See also: https://github.com/openbao/openbao/issues/1319

### DIFF
--- a/website/content/docs/rfcs/external-keys.mdx
+++ b/website/content/docs/rfcs/external-keys.mdx
@@ -64,7 +64,7 @@ We suggest the following design for external keys:
      not conflict with any existing type and `pkcs11` will not be a valid
      type, instead requiring a named configuration. E.g., `gcpckms`, `thales`,
      `entrust`, and `securosys` might all be valid `types` for the endpoints
-     below.
+     below, specified in the `name` field in the stanza.
 
    For example, to register the SoftHSM PKCS#11 library as library type
    `"softhsm"` and allow its usage in two namespaces:
@@ -77,13 +77,13 @@ We suggest the following design for external keys:
    }
    ```
 
-2. Within `sys/namespaces/<:path>/*`, we add a new option, `external_key_types`,
-   which limits the allowed HSM and KMS types within this namespace and all
-   children. This allows both the root operator to restrict immediate children,
-   but also for future tenant operators to restrict their own child namespaces.
-   This reflects the hierarchical nature of access (that a parent namespace
-   operator can create a policy that effectively grants them full access to a
-   child namespace).
+2. We add a new path `sys/namespaces/<:path>/external_keys`, with an option,
+   `external_key_types`, which limits the allowed HSM and KMS types within
+   this namespace and all children. This allows both the root operator to
+   restrict immediate children, but also for future tenant operators to
+   restrict their own child namespaces. This reflects the hierarchical nature
+   of access (that a parent namespace operator can create a policy that
+   effectively grants them full access to a child namespace).
 
 3. Within the `sys/external-keys` space, we implement the following APIs:
 
@@ -93,7 +93,8 @@ We suggest the following design for external keys:
 
      A configuration may be inherited from a direct parent namespace via the
      special `inherits = "<:config-name>"` key. This inheritance mechanism
-     is chosen over a top-level `inheritable = true` flag to avoid potential
+     is chosen over a `inherited = true` flag in the child namespace or over
+     a `inheritable=true` flag in the parent namespace to avoid potential
      naming conflicts across the namespace hierarchy, or alternatively to avoid
      the need for a mechanism/format to address a specific ancestor namespace.
      Inherited configurations are similar to symbolic links; they do not copy
@@ -102,9 +103,13 @@ We suggest the following design for external keys:
      several levels of child namespaces, you would build a chain of inherited
      configurations.
 
+     `/` and `:` will be forbidden identifiers for `config-name`.
+
    - `configs/<:config-name>/keys/<:key-name>`, to maintain mappings of keys
      and their access information. A `key-name` is unique per `config-name`, not
-     per namespace.
+     per namespace. The key will have a UUID associated with it to allow it to
+     be referenced independently of `config-name`. Otherwise, use in plugins
+     will require the `<config-name>/<key-name>` format.
 
    - `configs/<:config-name>/keys/<:key-name>/grants/<:mount-path>` will be
      used to add or remove mounts from accessing the key.
@@ -136,16 +141,21 @@ will be done with the `config-name` plus `key-name` association.
 This requires an improvement to the configuration, along with allowing these new
 stanzas to be reloaded via `SIGHUP`.
 
+Unlike the seal mechanism, where key rotation is automatically detected by the
+wrapper and the root key is transparently re-encrypted, binding in the
+external keys layer is assumed to be to an exact underlying key at a very
+specific version and not to a set of key (or a keyring).
+
 ### SystemView Changes
 
-However, elided from the above user-facing description is the core technical
+Elided from the above user-facing description is the core technical
 improvement: plugins interact with core via the `logical.SystemView`
 [interface](https://github.com/openbao/openbao/blob/48e7bd6f8358191056bdd0d2ac8b11e9e253254d/sdk/logical/system_view.go#L18-L99),
 which will need to be extended to support the following new APIs:
 
 1. `ListKeys() (map[string]*ExternalKeyInfo, error)` - to return information
    about external keys accessible to this mount.
-2. `GetKey() (go-kms-wrapper.ExternalKey, error)` - to return a helper to
+2. `GetKey() (logical.ExternalKey, error)` - to return a helper to
    use external keys. This will be implemented in `go-kms-wrapper` as a new
    key type which _always_ performs direct operations via underlying keys.
    We will need to make this compatible with the `crypto` standard library
@@ -154,14 +164,31 @@ which will need to be extended to support the following new APIs:
    does). `GetKey()` would likely take both a `config-name` and `key-name`
    parameter to uniquely identify a key.
 
-`ExternalKey` will implement two standard Go types, depending on the
+`ExternalKey` will implement one or two standard Go types, depending on the
 capabilities of the underlying keys: `crypto.Decrypter` and `crypto.Signer`.
 Once an `ExternalKey` has been retrieved via `GetKey()`, a secret engine can
 check for interface support by type casting. Access to private keys will not be
 returned.
 
-In the future, support for HMAC-based keys may be supported. This means that
-initially, Transit will not support modes which require HMAC.
+In the future, HMAC-based keys may be supported. This means that initially,
+Transit will not support modes which require HMAC.
+
+### Plugin Key Usage
+
+Upstream HashiCorp Vault Enterprise uses the `managed_key_name` and
+`managed_key_uuid` fields. We will support these as well as the
+`external_key_name` and `external_key_uuid` fields.
+
+The intention here is that plugin authors will enable external keys via an
+explicit key type (`key_type=managed_key` or `key_type=external_key`). The
+actual key type will be inferred from the underlying external key instance.
+When providing these parameters, the above `external_key_name` or
+`external_key_uuid` fields will need to be provided, which will be used for
+the lookup via the SystemView API.
+
+In Transit in particular, each key version could reference different values
+for `external_key_name` and `external_key_uuid`, giving different versions.
+Any automatic rotation will be disabled.
 
 ## Rationale and Alternatives
 
--- a/website/content/docs/rfcs/external-keys.mdx
+++ b/website/content/docs/rfcs/external-keys.mdx
@@ -1,0 +1,178 @@
+---
+sidebar_label: External Keys from KMSs and HSMs
+description: |-
+  External Key Configuration for KMS and HSM Access
+---
+
+# External Key Configuration for KMS and HSM Access
+
+## Summary
+
+For better security of cryptographic material, keys within OpenBao should
+optionally be backed by a KMS or HSM solution. We design a per-namespace
+repository of seal configurations, with mount-specific key information to
+prevent cross-mount usage of keys except via explicit reuse.
+
+We call this solution External Keys to differentiate it from HashiCorp
+Vault Enterprise's Managed Keys.
+
+## Problem Statement
+
+Upstream HashiCorp Vault Enterprise implemented Managed Keys, which have
+the following properties:
+
+1. KMS library configurations live in storage and apply to any namespace;
+   this takes a `type` and a `name` in a `kms_library` stanza. This sets
+   global configuration for the provider; per
+   [documentation](https://developer.hashicorp.com/vault/docs/configuration/kms-library)
+   this is only a PKCS#11 library configuration as it is the only type that
+   requires elements in storage.
+
+2. Within each namespace, a repository of keys accessible only to that
+   namespace exist.
+
+  - Each `type` of key can be listed (via  `LIST /sys/managed-keys/<:type>`),
+    to show which keys with the particular type exist).
+  - Creating a key sets its name (`POST /sys/managed-keys/<:type>/<:key-name>`),
+    the name of the library (if PKCS#11), and all credential information.
+
+  In particular, this means that there is no central repository (within a
+  namespace) of credentials, making rotation harder.
+
+Further, while less restrictive, any namespace can use any `kms_library` or
+seal type; a namespace admin has full control over policies and thus a global
+operator cannot restrict types.
+
+## User-facing Description
+
+We suggest the following design for external keys:
+
+1. The `external_keys` stanzas take an optional `namespaces = []` parameter,
+   taking items of the form `uuid:<ns-uuid>`, `id:<ns-accessor>`, or
+   `path:<ns-path>`, to identify namespaces that this library is allowed to
+   be used in. This prevents cross-tenant library usage problems by preventing
+   them from using the underlying library.
+
+   - In the future, when support for pluginized KMS integrations are added, we
+     can use an `external_keys "binary" { ... }` stanza to limit access to
+     specific external libraries or hosts that they run on.
+
+   - The `name` of this stanza gets registered as the library type; it must
+     not conflict with any existing type and `pkcs11` will not be a valid
+     type, instead requiring a named configuration. E.g., `gcpckms`, `thales`,
+     `entrust`, and `securosys` might all be valid `types` for the endpoints
+     below.
+
+2. Within `sys/namespaces/<:path>`, we add a new option, `types`, which limits
+   the allowed HSM and KMS types within this namespace and all children. This
+   allows both the root operator to restrict immediate children, but also for
+   future tenant operators to restrict their own child namespaces. This
+   reflects the hierarchical nature of access (that a parent namespace
+   operator can create a policy that effectively grants them full access to a
+   child namespace).
+
+3. Within the `sys/external-keys` space, we implement the following APIs:
+
+   - `configs/<:config-name>`, to configure KMS or HSM _access_ information,
+     such as type (above) or credentials. This gives a single point of
+     rotation for any given key using this provider.
+
+   - `configs/<:config-name>/keys/<:key-name>`, to maintain mappings of keys
+     and their access information.
+
+   - `configs/<:config-name>/keys/<-key-name>/grants/:mount_path` will be
+     used to add or remove mounts from accessing the key.
+
+Notably, unlike Vault Enterprise, no key material is implicitly created at
+the key association step; it must already exist within the KMS or HSM and this
+is just a linking.
+
+Furthermore, rather than relying on mount tuning, the use of explicit grants,
+with mounts in the path, allows for fine-grained delegation of permission via
+the standard ACL models, assuming the operator does not have policy
+modification permissions.
+
+In the future, key creation may be supported as a follow-up RFC.
+
+Individual mount types (PKI, SSH, Identity, ...) will need to be updated to
+support "creating" keys via using external keys instead of existing keys. This
+will be done with the `key-name` association.
+
+## Technical Description
+
+This requires an improvement to the configuration, along with allowing these new
+stanzas to be reloaded via `SIGHUP`.
+
+### SystemView Changes
+
+However, elided from the above user-facing description is the core technical
+improvement: plugins interact with core via the `logical.SystemView`
+[interface](https://github.com/openbao/openbao/blob/48e7bd6f8358191056bdd0d2ac8b11e9e253254d/sdk/logical/system_view.go#L18-L99),
+which will need to be extended to support the following new APIs:
+
+1. `ListKeys() (map[string]*ExternalKeyInfo, error)` - to return information
+   about external keys accessible to this mount.
+2. `GetKey() (go-kms-wrapper.ExternalKey, error)` - to return a helper to
+   use external keys. This will be implemented in `go-kms-wrapper` as a new
+   key type which _always_ performs direct operations via underlying keys.
+   We will keep to make this compatible with the `crypto` standard library
+   (to support e.g., `crypto.Signer` and other such interfaces like the
+   limited [`crypto11` library](https://github.com/ThalesGroup/crypto11)
+   does).
+
+`ExternalKey` will implement two standard Go types, depending on the
+capabilities of the underlying keys: `crypto.Decrypter` and `crypto.Signer`;
+access to private keys will not be returned.
+
+In the future, support for HMAC-based keys may be supported. This means that
+initially, Transit will not support modes which require HMAC.
+
+## Rationale and Alternatives
+
+Reimplementing Managed Keys as it exists in Vault Enterprise does not match
+the future semantics we hope to include for namespaces, such as stronger
+multi-tenant separation (e.g., [namespace-level sealing](https://github.com/openbao/openbao/issues/1170)).
+
+Another alternative would be to implement a HSM or KMS library as a top-level
+secrets engine in OpenBao. This would directly expose actions on keys as
+top-level APIs. With a cross-plugin communication mechanism, this could open
+up a fully-pluggable system for such keys. However, this opens a significant
+weakness: OpenBao is assumed (in e.g., the PKI, SSH, and OIDC Identity
+Provider) to _control_ signing capabilities. Users are not allowed to sign
+arbitrary blobs but are instead restricted to values explicitly allowed by
+these engines' roles. This means that opening up such cross-plugin
+communication directly via users' existing token on a request would mean they
+have to have API access to bypass these controls. Thus the only viable approach
+is a per-mount identity, policy, and tokens, and would be significantly more
+work.
+
+## Downsides
+
+One downside of this change is that users of managed keys will not be able
+to directly transition to external keys. However, because managed keys are
+transparent from an API caller's perspective and are only visible to
+operators, this should only be a one-time impact on migration from Vault
+Enterprise to OpenBao.
+
+## Security Implications
+
+This improves the overall security posture of the upstream feature and of
+OpenBao: backing key material by a HSM or KMS is one of the strongest storage
+mechanisms we could support.
+
+## User/Developer Experience
+
+This may have some impact on request time depending on the latency and
+throughput of the external HSM or KMS.
+
+## Unresolved Questions
+
+n/a
+
+## Related Issues
+
+n/a
+
+## Proof of Concept
+
+n/a
--- a/website/content/docs/rfcs/external-keys.mdx
+++ b/website/content/docs/rfcs/external-keys.mdx
@@ -77,13 +77,13 @@ We suggest the following design for external keys:
    }
    ```
 
-2. We add a new path `sys/namespaces/<:path>/external_keys`, with an option,
-   `external_key_types`, which limits the allowed HSM and KMS types within
-   this namespace and all children. This allows both the root operator to
-   restrict immediate children, but also for future tenant operators to
-   restrict their own child namespaces. This reflects the hierarchical nature
-   of access (that a parent namespace operator can create a policy that
-   effectively grants them full access to a child namespace).
+2. We add a new path `sys/namespaces/<:path>/external-keys`, with an option,
+   `types`, which limits the allowed HSM and KMS types within this namespace
+   and all children. This allows both the root operator to restrict immediate
+   children, but also for future tenant operators to restrict their own child
+   namespaces. This reflects the hierarchical nature of access (that a parent
+   namespace operator can create a policy that effectively grants them full
+   access to a child namespace).
 
 3. Within the `sys/external-keys` space, we implement the following APIs:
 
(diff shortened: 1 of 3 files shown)


website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,178 @@
...
+     can use an `external_keys "binary" { ... }` stanza to limit access to
+     specific external libraries or hosts that they run on.
+
+   - The `name` of this stanza gets registered as the library type; it must
+     not conflict with any existing type and `pkcs11` will not be a valid
+     type, instead requiring a named configuration. E.g., `gcpckms`, `thales`,
+     `entrust`, and `securosys` might all be valid `types` for the endpoints
+     below.
+
+2. Within `sys/namespaces/<:path>`, we add a new option, `types`, which limits
```
cipherboy, line 66:
```suggestion
2. Within `sys/namespaces/<:path>`, we add a new option, `external_key_types = [ ... ]`, which limits
```

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,178 @@
...
+   the allowed HSM and KMS types within this namespace and all children. This
+   allows both the root operator to restrict immediate children, but also for
+   future tenant operators to restrict their own child namespaces. This
+   reflects the hierarchical nature of access (that a parent namespace
+   operator can create a policy that effectively grants them full access to a
+   child namespace).
+
+3. Within the `sys/external-keys` space, we implement the following APIs:
+
+   - `configs/<:config-name>`, to configure KMS or HSM _access_ information,
```
cipherboy, line 76: This should probably have a flag, `inheritable`, to show whether or not a given config is inheritable by child namespaces, defaulting to `false`. If so, we should try to make a best-effort at identifying sensitive fields (per-type) and restricting them from viewing on `read` connections in inherited child namespaces. 

Not clear: if child namespace is sealed and `inherited=true`, what should be the result? This could cause a conflict between parent/child version. Maybe `:` should be denied as a special character in config names and used to prefix the NS Accessor in the child view? `ns:<accessor>:<config-name>`?

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+Further, while less restrictive, any namespace can use any `kms_library` or
+seal type; a namespace admin has full control over policies and thus a global
+operator cannot restrict types.
+
+## User-facing Description
+
+We suggest the following design for external keys:
+
+1. The `external_keys` stanzas take an optional `namespaces = []` parameter,
```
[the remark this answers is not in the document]
cipherboy: @satoqz I was thinking about deliberately changing the name to make it clear that we're not aiming for 1:1 compatibility with Vault Enterprise's version and have different operator API semantics. If we apply that consistently, starting at the config level, it'll be more obvious to users.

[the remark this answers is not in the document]
cipherboy: @satoqz That's fair, we can do that. I wasn't sure if we wanted controls over other library types as well potentially? Not sure what those would be off the top of my head tbh.

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+1. `ListKeys() (map[string]*ExternalKeyInfo, error)` - to return information
+   about external keys accessible to this mount.
+2. `GetKey() (go-kms-wrapper.ExternalKey, error)` - to return a helper to
+   use external keys. This will be implemented in `go-kms-wrapper` as a new
+   key type which _always_ performs direct operations via underlying keys.
+   We will keep to make this compatible with the `crypto` standard library
+   (to support e.g., `crypto.Signer` and other such interfaces like the
+   limited [`crypto11` library](https://github.com/ThalesGroup/crypto11)
+   does).
```
satoqz, line 121: @voigt and I were looking at `go-kms-wrapping` today to figure out where to get started, assuming implementing support for `ExternalKey` over there is a good starting point.

Couple of questions we had:

- `ExternalKeyInfo` would be a type in `openbao`, not in `go-kms-wrapping`, correct? What fields would it potentially have? Presumably `config-name` and `key-name`? Anything more specific that `go-kms-wrapping`/`ExternalKey` would have to account for?

- So holding an `ExternalKey` would not guarantee that you hold a `crypto.Signer` and/or a `crypto.Decrypter`, i.e. you have to cast like `signer, ok := key.(crypto.Signer)` to check for support?

- Have you already had any thoughts on what the `ExternalKey` type/interface would look like, how it would be accessed/constructed? Does a `go-kms-wrapping.Wrapper` yield an `ExternalKey`, does an `ExternalKey` hold a `go-kms-wrapping.Wrapper`?

cipherboy: @satoqz Sorry, never found time to reply:

1. Right, I think `ExternalKeyInfo` would be in OpenBao and be a "pointer type"; it wouldn't have the actual key _details_ but instead just pretty names for users. Internally, OpenBao would hold the actual key information and not expose it to plugins at all. However, it'll be the soft reference used for looking up/resolving the actual key, e.g., over the GRPC bridge. If the plugin is builtin, it might be able to downcast all the way to the wrapper version, but not sure. Let me review this to see if my design doc has issues with typing :-D 
2. Right, you would have to check for support. This is probably unavoidable unless you add two separate List endpoints (`GetSigners(...)`, `GetDecrypters(...)`), and then you'd have to still return a custom type that includes `crypto.Signer` plus your high-level naming helper. 
3. It needs to be fairly high level, and you may not actually have a direct reference to something in g-k-w; it may be a GRPC bridged version of this interface. I don't yet know the relationship between wrapper and external key; to me, most wrappers live in two pieces (client, and actual wrapper), so perhaps both `Wrapper` and `ExternalKey` use the same underlying `client`, and `Wrapper` is stuck exclusive to seal wrapping. 

Thoughts as they are now, open to suggestions :-)

cipherboy:
> (`key_type` is a new option I'm suggesting, lets you specify just "RSA" instead of the full mechanism to resolve a key)

I definitely understand this mirrors the `CKK_*` pattern, but this means that we can't restrict keys to particular algorithms. Not as much a problem for RSA as PKCS#1v1.5 and OAEP/PSS seem to be used interchangeably, but for things like AES this is important as you really shouldn't mix e.g., CTR and GCM with the same key. 

Thoughts?

satoqz:
@cipherboy oh certainly, for keys where we can't choose the mechanism automatically based on the key type (no matter if specified by the user or inferred from the PKCS#11 object) we would require `mechanism` to be set (or have the more commonly used/originally supported one, in case of AES that would be AES-GCM, be the default).

The main motivation behind `key_type` is indeed signing via asymmetric keys, with the `ExternalKey` interface you want to be able to register an opaque "RSA key" and then later pass further options into `Sign(...)`.

satoqz:
I can also see the point in being restrictive by default and not allowing key usage at all unless config specifies what mechanism to use it with. Might be simpler ultimately. FWIW the current g-k-w PKCS#11 wrapper does not require setting a mechanism and will automatically resolve the mechanism from the key type. In that case we would have to break that capability, which is fair though.

cipherboy:
@satoqz Maybe mechs should be plural?

satoqz:
> @satoqz Maybe mechs should be plural?

Yup, thought about this too, else RSA users will be unhappy :D
~~Currently I think we can only pass strings to `WithConfigMap`, I suppose we can comma-separate.~~

Edit: Actually, why not just add the key multiple times then, with a different mechanism each. go-kms-wrapping will always know how to use the key right away, and the user has ultimate control with a tiny bit of duplication.

cipherboy:
@satoqz said:

> I can also see the point in being restrictive by default and not allowing key usage at all unless config specifies what mechanism to use it with. Might be simpler ultimately. FWIW the current g-k-w PKCS#11 wrapper does not require setting a mechanism and will automatically resolve the mechanism from the key type. In that case we would have to break that capability, which is fair though.

Yeah, I think the rationale there is, for a given key type, there is exactly one optimal algorithm for some definition, with an ordered list of alternatives. OAEP with SHA256 tends to be the best answer. Similarly, AES-GCM is most preferable, if supported.

That choice will never change.

But it's not unreadable to think that for different applications, there's different optimal algorithms (transit: sign vs ecdh) which in turn mean a reused key may break the algorithm's security model. Explicit is better, here.

Which reminds me, we probably need to expand the definition of ExternalKey to support ecdh and Kem.

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+`ExternalKey` will implement two standard Go types, depending on the
+capabilities of the underlying keys: `crypto.Decrypter` and `crypto.Signer`;
+access to private keys will not be returned.
+
+In the future, support for HMAC-based keys may be supported. This means that
+initially, Transit will not support modes which require HMAC.
+
```
cipherboy, line 198: Missing how to connect keys in PKi, Transit, &c.

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,178 @@
...
+   - `configs/<:config-name>`, to configure KMS or HSM _access_ information,
+     such as type (above) or credentials. This gives a single point of
+     rotation for any given key using this provider.
+
+   - `configs/<:config-name>/keys/<:key-name>`, to maintain mappings of keys
+     and their access information.
+
+   - `configs/<:config-name>/keys/<-key-name>/grants/:mount_path` will be
+     used to add or remove mounts from accessing the key.
```
satoqz: Would key names be unique across configs, or per config? This would decide how you later link to a key in an engine.

website/content/docs/rfcs/external-keys.mdx
satoqz: > If so, we should try to make a best-effort at identifying sensitive fields (per-type) and restricting them from viewing on read connections in inherited child namespaces.

Isn't this already handled as a child namespace token wouldn't be able to interact with the parent namespace in the first place?

satoqz: > If so, we should try to make a best-effort at identifying sensitive fields (per-type) and restricting them from viewing on read connections in inherited child namespaces.

Isn't this already handled as a child namespace token wouldn't be able to interact with the parent namespace in the first place?

satoqz: How does `inheritable` play with `grants/:mount_path`? Are mount path grants always absolute starting from the root namespace?

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,178 @@
...
+stanzas to be reloaded via `SIGHUP`.
+
+### SystemView Changes
+
+However, elided from the above user-facing description is the core technical
+improvement: plugins interact with core via the `logical.SystemView`
+[interface](https://github.com/openbao/openbao/blob/48e7bd6f8358191056bdd0d2ac8b11e9e253254d/sdk/logical/system_view.go#L18-L99),
+which will need to be extended to support the following new APIs:
+
+1. `ListKeys() (map[string]*ExternalKeyInfo, error)` - to return information
```
satoqz, line 156: How would this ultimately be exposed to the user? An engine-specific info endpoint?

satoqz:
I'd like to start implementing support in core, but in hindsight I'm unhappy with the complexity of access control mechanisms.

To summarize, we have:

1. Restricting PKCS#11 library access per-namespace in the configuration file
2. Restricting allowed KMS types for a namespace in the namespace entry itself
3. Restricting child namespace inheritance via `inheritable` at the `/:config-name` level
4. Restricting secret engine access via mount path grants at the `/:key-name` level

satoqz:
> 1. Restricting PKCS#11 library access per-namespace in the configuration file

I would keep this as-is.

> 2. Restricting allowed KMS types for a namespace in the namespace entry itself

I don't see the value here. 1) makes total sense as PKCS#11 access is a bit closer to the underlying system, so the system administrator gets to grant access. Restricting generic KMS types per namespace ("you can't use Azure!!") is like restricting the database engine based on database type per namespace.

> 3. Restricting child namespace inheritance via `inheritable` at the `/:config-name` level
> 4. Restricting secret engine access via mount path grants at the `/:key-name` level

I wonder if these two can be merged into one. For starters, I don't like `inheritable` for access control since you can't set a limit to the depth of inheritance - you either enable access for all child namespaces or none. Since mount path grants would have to scale across child namespaces anyways, why not forget about `inheritable` and use grants for everything?

Suppose we have following external key in the `foo` namespace:

```
/foo/sys/external-keys/configs/my-config/keys/my-key
```

to grant an engine in the same namespace access, we add the following grant:

```
/foo/sys/external-keys/configs/my-config/keys/my-key/grants/pki
```

To provide access to another engine in a child namespace:

```
/foo/sys/external-keys/configs/my-config/keys/my-key/grants/bar/baz/transit
```

which would let the `transit` mount in namespace `foo/bar/baz` access the key, without ever providing access to `foo/bar` or a namespace that's nested in `foo/bar/baz`.

Edit: I think https://github.com/openbao/openbao/pull/1320#discussion_r2150667846 is the way I'd ultimately like to do this.

website/content/docs/rfcs/external-keys.mdx
satoqz: Ah, how about this:

- Rather than an `inheritable` flag on the parent namespace's config, we add the child namespace name to `grants`
- In the child namespace, we add a config with `inherits = <config name in parent namespace>`
- Child namespace can then set its own `grants` for the inherited config
- To propagate access downwards infinitely, you just inherit from an inherited config generation by generation

This lets us avoid:
- `inheritable=true` providing too much access
- The `ns:<accessor>:<config-name>` construct and checking for cross-namespace collisions generally, child namespaces just reference the "local" inherited config
- The parent namespace needing to know about the child namespace's mounts

cipherboy: I think `inheritable=true` is unnecessary, the grants could be grants at the top level could be all the way down to the child namespace's mount.

satoqz:
Updated based on past WG discussions and resolved conflicts :)

website/content/docs/rfcs/external-keys.mdx
satoqz: Left this one up to you :-D

website/content/docs/rfcs/index.mdx
```diff
@@ -63,3 +63,6 @@ Steering Committee.
  - [Best Practices for Integrating CEL in Auth, Secrets, and Elsewhere](/docs/rfcs/cel-best-practices),
    to discuss how Google's Common Expression Language should be integrated
    into OpenBao.
+ - [External Key Configuration for KMS and HSM Access](/docs/rfcs/external-keys),
+   to add support for secure key material storage and interaction from
```
wslabosz-reply: not sure
```suggestion
   to add support for secure key material storage and interaction with
```

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,211 @@
...
+     below.
+
+   For example, to register the SoftHSM PKCS#11 library as library type
+   `"softhsm"` and allow its usage in two namespaces:
+
+   ```hcl
+   external_keys "pkcs11" {
+     name       = "softhsm"
+     library    = "/usr/lib/softhsm/libsofthsm2.so"
+     namespaces = ["path:foo/bar", "uuid:c0de1570-688e-41ad-ad97-1395fe30cbbd"]
```
wslabosz-reply, line 76: just a thought (on a implementation side), what would happen when an operator specifies all three possible methods of namespace annotation, do we register it three times? probably should safeguard against multiple links between same external_key and a namespace.

website/content/docs/rfcs/index.mdx
satoqz: I think "from" makes sense since we're interacting with external keys _from_ secrets engines, but core doesn't do anything novel _with_ secrets engines.

website/content/docs/rfcs/external-keys.mdx
satoqz: I think it's fine as we would never fetch a namespace based on the values here, rather we would already have a resolved `namespace.Namespace` (e.g. the one that a config is being loaded for at startup, or the one that is currently creating a config) and just compare it to these strings by value.

wslabosz-reply (approved):
I've got nothing to be said against this RFC

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,214 @@
...
+
+   - In the future, when support for pluginized KMS integrations are added, we
+     can use an `external_keys "binary" { ... }` stanza to limit access to
+     specific external libraries or hosts that they run on.
+
+   - The `name` of this stanza gets registered as the library type; it must
+     not conflict with any existing type and `pkcs11` will not be a valid
+     type, instead requiring a named configuration. E.g., `gcpckms`, `thales`,
+     `entrust`, and `securosys` might all be valid `types` for the endpoints
+     below.
```
cipherboy:
```suggestion
     below, specified in the `name` field in this stanza.
```

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,214 @@
...
+
+   ```hcl
+   external_keys "pkcs11" {
+     name       = "softhsm"
+     library    = "/usr/lib/softhsm/libsofthsm2.so"
+     namespaces = ["path:foo/bar", "uuid:c0de1570-688e-41ad-ad97-1395fe30cbbd"]
+   }
+   ```
+
+2. Within `sys/namespaces/<:path>/*`, we add a new option, `external_key_types`,
```
cipherboy:
```suggestion
2. We add a new path `sys/namespaces/<:path>/external_keys`, with an option, `external_key_types`,
```

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,214 @@
...
+
+3. Within the `sys/external-keys` space, we implement the following APIs:
+
+   - `configs/<:config-name>`, to configure KMS or HSM _access_ information,
+     such as type (above) or credentials. This gives a single point of rotation
+     for any given key using this provider.
+
+     A configuration may be inherited from a direct parent namespace via the
+     special `inherits = "<:config-name>"` key. This inheritance mechanism
+     is chosen over a top-level `inheritable = true` flag to avoid potential
```
cipherboy:
```suggestion
     is chosen over a top-level `inherited = true` flag to avoid potential
```
`inheritable = true` would be in the parent namespace.

website/content/docs/rfcs/external-keys.mdx
cipherboy: @satoqz I left some comments up there, but I'll address them.

I think we'll use the existing `managed_key` format from Vault Enterprise: https://developer.hashicorp.com/vault/api-docs/secret/pki#managed-keys  --- but support `external_*` prefixes as well.

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,214 @@
...
+will be done with the `config-name` plus `key-name` association.
+
+## Technical Description
+
+This requires an improvement to the configuration, along with allowing these new
+stanzas to be reloaded via `SIGHUP`.
+
+### SystemView Changes
+
+However, elided from the above user-facing description is the core technical
```
cipherboy:
```suggestion
Elided from the above user-facing description is the core technical
```

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,214 @@
...
+improvement: plugins interact with core via the `logical.SystemView`
+[interface](https://github.com/openbao/openbao/blob/48e7bd6f8358191056bdd0d2ac8b11e9e253254d/sdk/logical/system_view.go#L18-L99),
+which will need to be extended to support the following new APIs:
+
+1. `ListKeys() (map[string]*ExternalKeyInfo, error)` - to return information
+   about external keys accessible to this mount.
+2. `GetKey() (go-kms-wrapper.ExternalKey, error)` - to return a helper to
```
cipherboy:
```suggestion
2. `GetKey() (logical.ExternalKey, error)` - to return a helper to
```

I think wrapping the g-k-w type with an internal type would be good. Would let us implement it more easily in the GRPC layer.

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,214 @@
...
+   use external keys. This will be implemented in `go-kms-wrapper` as a new
+   key type which _always_ performs direct operations via underlying keys.
+   We will need to make this compatible with the `crypto` standard library
+   (to support e.g., `crypto.Signer` and other such interfaces like the
+   limited [`crypto11` library](https://github.com/ThalesGroup/crypto11)
+   does). `GetKey()` would likely take both a `config-name` and `key-name`
+   parameter to uniquely identify a key.
+
+`ExternalKey` will implement two standard Go types, depending on the
```
cipherboy:
```suggestion
`ExternalKey` will implement one or two standard Go types, depending on the
```

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,214 @@
...
+capabilities of the underlying keys: `crypto.Decrypter` and `crypto.Signer`.
+Once an `ExternalKey` has been retrieved via `GetKey()`, a secret engine can
+check for interface support by type casting. Access to private keys will not be
+returned.
+
+In the future, support for HMAC-based keys may be supported. This means that
```
cipherboy:
```suggestion
In the future, HMAC-based keys may be supported. This means that
```

cipherboy (commented):
Some minor cleanup I'll pick up :-)

cipherboy:
@satoqz @DanGhita I think this RFC is strictly focused on the high level OpenBao server changes, so would be fine to merge ahead of having a design for #1467. #1467 will influence the g-k-w changes, but won't impact this level at all.

I think in the interim of such a design @satoqz, you can start on the high-level changes.

website/content/docs/rfcs/external-keys.mdx
satoqz: The parent namespace is what I mean here, but I suppose it's unclearly worded.

website/content/docs/rfcs/external-keys.mdx
satoqz: I think it doesn't matter as much concerning the GRPC layer as long as you don't plan to change something about the interface by wrapping it, and as long as it remains an interface (not a concrete type) on the g-k-w side?

The only advantage I can see for having the interface in api or sdk is that plugins don't need to concern themselves with importing g-k-w to reference the ExternalKey type.

satoqz:
@cipherboy yeah, happy to get started. We are of course still missing a way for core to actually load a config on creation or startup and use it to construct an `ExternalKey`, but I suppose the basic API/storage layer can already be done. Would you prefer to merge a partial feature like that to main or to a feature branch?

website/content/docs/rfcs/external-keys.mdx
satoqz: Should we really support both? I agree with keeping the general format with `external_*`, but wouldn't support anything `managed_*`. This RFC so far has been explicitly distinct from and incompatible with VE's managed keys, this would suggest the opposite.

cipherboy:
@satoqz Let's merge to a feature branch. We have three outstanding candidates for this next release (this, horizontal scalability, and namespace seals) plus a smattering of other relatively interesting smaller features, so its really whichever feature(s) get done first to cause us to cut v2.4.0. We'll not want to hold up on all three if the timelines end up being too different.

website/content/docs/rfcs/external-keys.mdx
cipherboy: @satoqz I think the value will be slightly different than with `managed_key`, with the `<config_name>:<key_name>` format.

From an operator configuration perspective they're different but from a usage perspective, I think (minus not having globally unique key names), they're mostly the same and we could benefit from supporting `managed_key_name`? My 2c. 

E.g., we can interop with things like https://github.com/pulumi/pulumi-vault/blob/0c6b2327ebf2bb901753cf1faa974328cb6cbb76/sdk/python/pulumi_vault/pkisecret/secret_backend_key.py#L149. Not the key creation side, of course.

satoqz: Fair, we can do that. @cipherboy should external keys also have UUIDs? (at that point it becomes debatable whether we need to go through the whole inheritance ordeal)

cipherboy: > at that point it becomes debatable whether we need to go through the whole inheritance ordeal

I think having inheritance is necessary so that we can solve the permission problem, no? Even if a key was referenced by UUID, I'd still want a two-way explicit permission sharing reference.

website/content/docs/rfcs/external-keys.mdx
cipherboy: > The only advantage I can see for having the interface in api or sdk is that plugins don't need to concern themselves with importing g-k-w to reference the ExternalKey type.

Right, it is a rather fat library so I'd like to avoid it if we can. Also it is somewhat unusual, most of the other pieces of logical live there so I'd prefer to keep it that way.

cipherboy:
@satoqz @wslabosz-reply Pushed a new section on usage and fixed my comments above, lmkwyt!

website/content/docs/rfcs/external-keys.mdx
satoqz: The present inheritance design doesn't cater well to UUID key references then. Looking up a key based on `<config-name>/<key-name` is easy:

1. Check for `<config-name>` in the local namespace
2. If the config has an `inherits` field, repeat step 1 with the parent namespace (setting `config-name` to  the value of `inherits`)
3. Check for `<key-name>` in the final non-inherited configuration

How would you do this with UUIDs if you still want to ensure that the config was explicitly inherited? Assuming we can easily index a key + its config by UUID, it'll be expensive to then resolve whether the config is inherited into that namespace.

cipherboy: Ah, we could always use the `<config-name>/<key-uuid>` format instead I suppose.

satoqz: That's the simple fix, if you're fine with that, I'm fine with it too. A bit unsure because `external_key_id` really sounds like we'd want some opaque ID value, but _at least_ it's not named `external_key_uuid`, which would set an even higher expectation.

satoqz: I'm also not sure why we need UUIDs (or why managed keys has UUIDs) as there's no config/key rename operation (neither in our design, nor in managed keys).

Specifically, the comment below:

https://github.com/openbao/openbao/blob/f012eef04fb5e1bcad84b22f5a588b331fbfb013/sdk/logical/managed_key.go#L30-L32

does not agree with the operations available at https://developer.hashicorp.com/vault/api-docs/system/managed-keys.

Unless of course we use the UUID in the storage path to obscure key names, at which point you'd likely want the same for config names. And even then, why should plugins care about those UUIDs?

cipherboy: Hmm, I think that's a fair criticism. I assume the current system just didn't support rename; with transactions, I think it is more possible, but you still have to update any _named_ references to the key on rename otherwise they'll break. Many places don't really let you do that (e.g., PKI). 

So probably best to do a delete+recreate tbh.

satoqz: Yeah, I've thought about this for a good while now and have convinced myself that we shouldn't have UUIDs and thus no `external_key_id`/`managed_key_id`, just `external_key_name`,`managed_key_name`.

Upsides:
- Our inheritance design remains sound and clean. `<config-name>/<key-uuid>` could be an approach, but really you would expect to only need the UUID here. If configs are renamable, you'd also need `<config-uuid>/<key-uuid>`, which is horrifying.
- We avoid special rename endpoints and logic; rather you would delete+recreate which is generally more idiomatic within the application
- We can map names 1:1 to storage entries, lookups are simple. No need to linearly search storage that uses UUID keys (see PKI engine when looking up issuers by name). Alternatively, no need to keep all keys in memory and index in e.g. a memdb.
- Less ambiguity and choices to make for the operator, there's only one way to link to an external key.

Downsides:
- Specifically PKI wouldn't play well with changing the name of an external key via recreate+delete. You'd want to keep the issuer with its CA cert but swap out the key (or modify the key itself) which isn't currently possible. I believe we should fix this in PKI itself, and allow mutating a key itself or the `key_id` of an issuer if it is of `type = external_key`. You can't guarantee the immutability of the underlying key material with external keys in the first place, so why pretend. Transit doesn't have this issue, changing the external key can be done as part of the key rotation endpoint (that's what upstream does).

Thoughts?

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,241 @@
...
+   use external keys. This will be implemented in `go-kms-wrapper` as a new
+   key type which _always_ performs direct operations via underlying keys.
+   We will need to make this compatible with the `crypto` standard library
+   (to support e.g., `crypto.Signer` and other such interfaces like the
+   limited [`crypto11` library](https://github.com/ThalesGroup/crypto11)
+   does). `GetKey()` would likely take both a `config-name` and `key-name`
+   parameter to uniquely identify a key.
+
+`ExternalKey` will implement one or two standard Go types, depending on the
+capabilities of the underlying keys: `crypto.Decrypter` and `crypto.Signer`.
```
satoqz: `crypto.Decrypter` is a bit of a niche interface and only really used by `crypto/tls` with RSA (at least what concerns std). E.g. for transit, it would be much more important to have our own interface with generic `Encrypt`/`Decrypt` methods that cover both symmetric and asymmetric keys.

website/content/docs/rfcs/external-keys.mdx
cipherboy: > Specifically PKI wouldn't play well with changing the name of an external key via recreate+delete. You'd want to keep the issuer with its CA cert but swap out the key (or modify the key itself) which isn't currently possible. I believe we should fix this in PKI itself, and allow mutating a key itself or the key_id of an issuer if it is of type = external_key. You can't guarantee the immutability of the underlying key material with external keys in the first place, so why pretend. 

Only if the underlying key material is exactly the same is this operation is safe. Otherwise if you have different key material your public key in the CA certificate changes, which either involves getting it re-signed by a parent authority (easy-ish, some places this requires 3mo+ of tickets + coordination + review) or if it is self-signed, placing it in every trust store (hard).

So in PKI, you'd just create another issuer and re-import the same certificate if the key name really has changed by the key material is the same. 

Otherwise yes, I agree.

website/content/docs/rfcs/external-keys.mdx
cipherboy: I think implement the relevant symmetric interfaces (https://pkg.go.dev/crypto/cipher#AEAD most importantly) is probably the best approach here. I wouldn't implement non-AEAD in our first pass.

website/content/docs/rfcs/external-keys.mdx
satoqz: > So in PKI, you'd just create another issuer and re-import the same certificate if the key name really has changed by the key material is the same.

You're completely right, I was even thinking about that path but somehow concluded there were limitations and forgot 😄

satoqz:
@cipherboy updated to include recent discussion, and snuck in some cleanup to the proposed `/sys/namespaces` extensions.

I'm happy with this now, let's hope it stays that way :^)

website/content/docs/rfcs/external-keys.mdx
```diff
@@ -0,0 +1,247 @@
...
+   }
+   ```
+
+2. We add a new path `sys/namespaces/<:path>/external-keys`, with an option,
+   `types`, which limits the allowed HSM and KMS types within this namespace
+   and all children. This allows both the root operator to restrict immediate
+   children, but also for future tenant operators to restrict their own child
+   namespaces. This reflects the hierarchical nature of access (that a parent
+   namespace operator can create a policy that effectively grants them full
+   access to a child namespace).
```
satoqz, line 86: Actually, to clarify: If namespace `foo` allows its child namespace `foo/bar` to use type `utimaco`, this does not count unless `foo` itself has access to `utimaco`, correct?

cipherboy: @satoqz correct.

cipherboy:
@satoqz Are we good to merge this btw? I lost track of where we're at :-D

satoqz:
@cipherboy I was going to ask you the same in today's WG meeting :^).

Let's merge this.

PR #1323, opened 2025-05-15
Prevent information disclosure on invalid request

cipherboy:
When sending an API request with incorrect typing of request data, OpenBao (and HashiCorp Vault) would leak information in service and audit logs. The original disclosure from HashiCorp states this is limited to KVv2, but this is applicable to any secrets engine.

See also: https://discuss.hashicorp.com/t/hcsec-2025-09-vault-may-expose-sensitive-information-in-error-logs-when-processing-malformed-data-with-the-kv-v2-plugin/74717

### DIFF
--- a/changelog/1323.txt
+++ b/changelog/1323.txt
@@ -1,0 +1,3 @@
+```release-note:security
+sdk/framework: prevent information disclosure on invalid request. HCSEC-2025-09 / CVE-2025-4166.
+```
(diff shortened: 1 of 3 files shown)


wslabosz-reply (approved):

cipherboy:
Based on impact and @wslabosz-reply I feel comfortable merging this. Thank you!

Issue #1340, opened 2025-05-21
RFC - Declarative Self-Initialization


cipherboy:
### Summary

OpenBao needs to make changes to improve provisioning and initialization in
order to facilitate day-zero automation and audited-by-default changes. We
suggest using an initialization structure based on request/response patterns,
with an optional output section for preserving any necessary responses.

### Problem Statement

OpenBao currently has a stateful, one-time initialization process. This makes
operating it in a fully declarative environment, such as NixOS or OpenTofu,
rather hard. In particular, initialization returns two items: a set of unseal
or recovery key shards and a highly privileged root token; some initial setup
using this root token should also be taken (such as initial auth method
creation, policies, and audit logging), before this root token is ideally
revoked. For auto-unseal, the initialization process does not yield necessary
unseal shards and all information required to unseal exists immediately on
first binary startup. Instead, recovery shards, used to generate new root
tokens, are returned from auto-unseal initializations. Without an initial root
token, no access to OpenBao is possible as no authentication methods are
configured.

Thus, simply initializing OpenBao and removing the need for the root token's
one-off use is insufficient unless recovery shares are placed elsewhere (in
a secure location) and used on subsequent configuration steps. Changes to
recovery key handling will be discussed in a subsequent RFC.

Ideally, initialization should set up an audit method, an initial policy, and
an auth method which would be used by the administrator via any declarative
policy. Then subsequent configuration could be entirely declarative and not
feature any one-time steps or one-time authentication methods. Only when an
operator is ready should one-time steps be taken, such as generating and
preserving long-term access to recovery keys.

However, each auth method and organization's policy requirements are
different. While we cannot seek to fully replace a declarative framework like
OpenTofu, we should have sufficient flexibility to define all initialization
up-front. Adding to this, output or request chaining is necessary: if OpenBao
creates a certificate authority, we may wish to make that available on disk
for other automation to use and deploy or reuse it in subsequent requests to
the instance during initialization (to e.g., configure a certificate auth
mount).

To bound the security risk, initialization configuration must only be applied
on first startup. It should not be treated as the long-term owner of these
resources and an import to a proper IaC tool would be preferable long-term.

We need this to interact nicely with existing configuration management
techniques; HCL and JSON-equivalent representations should be possible for
any initialization and we should not attempt to build a new DSL for this.

Lastly, whatever format is chosen should ultimately form the basis of the
[profile system](https://openbao.org/blog/profiles/), though the exact
driving mechanism may be different.

### User-facing Description

[...]

PR #1341, opened 2025-05-21
Add RFC on declarative self-initialization

cipherboy:
See also: https://github.com/openbao/openbao/issues/1340

### DIFF
--- a/website/content/docs/rfcs/self-init.mdx
+++ b/website/content/docs/rfcs/self-init.mdx
@@ -1,0 +1,292 @@
+---
+sidebar_label: Declarative self-initialization
+description: |-
+  An OpenBao RFC to discuss self-initializing OpenBao from a configuration
+  file entry rather than manual API calls.
+---
+
+# Declarative self-initialization
+
+## Summary
+
+OpenBao needs to make changes to improve provisioning and initialization in
+order to facilitate day-zero automation and audited-by-default changes. We
+suggest using an initialization structure based on request/response patterns,
+with an optional output section for preserving any necessary responses. This
+will improve operator experience by replacing manual post-startup steps with
+greater end-to-end automation especially when orchestrated by an IaC tool.
+
+## Problem Statement
+
+OpenBao currently has a stateful, one-time initialization process. This makes
+operating it in a fully declarative environment, such as NixOS or OpenTofu,
+rather hard. In particular, initialization returns two items: a set of unseal
+or recovery key shards and a highly privileged root token; some initial setup
+using this root token should also be taken (such as initial auth method
+creation, policies, and audit logging), before this root token is ideally
+revoked. For auto-unseal, the initialization process does not yield necessary
+unseal shards and all information required to unseal exists immediately on
+first binary startup. Instead, recovery shards, used to generate new root
+tokens, are returned from auto-unseal initializations. Without an initial root
+token, no access to OpenBao is possible as no authentication methods are
+configured.
+
+Thus, simply initializing OpenBao and removing the need for the root token's
+one-off use is insufficient unless recovery shares are placed elsewhere (in
+a secure location) and used on subsequent configuration steps. Changes to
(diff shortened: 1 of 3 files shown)


website/content/docs/rfcs/self-init.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+ - https://github.com/openbao/openbao/issues/944
+ - https://github.com/orgs/openbao/discussions/1115
+ - https://bank-vaults.dev/docs/concepts/external-configuration/
+ - https://github.com/orgs/openbao/discussions/1339
+ - https://github.com/openbao/openbao-helm/issues/54
+
+## Proof of Concept
+
+n/a
```
cipherboy, line 192: https://github.com/cipherboy/openbao/pull/new/self-initialization

Config:

```hcl
ui            = true
cluster_addr  = "https://127.0.0.1:8201"
api_addr      = "https://127.0.0.1:8200"

storage "raft" {
  path = "raft"
}

listener "tcp" {
  tls_disable   = true
}

initialize "audit" {
  request "enable-audit" {
    operation = "update"
    path = "sys/audit/stdout"
    data = {
      type = "file"
      options = {
        file_path = "/dev/stdout"
        log_raw = true
      }
    }
  }
}

initialize "identity" {
  request "mount-userpass" {
    operation = "update"
    path = "sys/auth/userpass"
    data = {
      type = "userpass"
      path = "userpass/"
      description = "admin"
    }
  }

  request "userpass-add-admin" {
    operation = "update"
    path = "auth/userpass/users/admin"
    data = {
      "password" = {
        eval_type = "string"
        eval_source = "env"
        env_var = "INITIAL_ADMIN_PASSWORD"
      }
      "token_policies" = ["superuser"]
    }
  }
}

initialize "policy" {
  request "add-superuser-policy" {
    operation = "update"
    path = "sys/policies/acl/superuser"
    data = {
      policy = <<EOP
path "*" {
  capabilities = ["create", "update", "read", "delete", "list", "scan", "sudo"]
}
EOP
    }
  }
}
```

cipherboy: I have an equivalent JSON I'll put in #1506 as well.

website/content/docs/rfcs/self-init.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+Items which ultimately need literals (such as operations, paths, &c) can
+alternatively take an object, defining an output type (`string`, `int`, &c)
+and a source (`env`, `file`, `request.data`, `response.data`, &c) to serve
+as the source value for the field.
+
+Initialization will occur in sorted order by name by component (first by
+each named initialization block in alphabetical order and then by each named
+request within the block in alphabetical order). Names must be unique within
+the specified scope.
```
satoqz: Hmm, would the user expect this, is there a better approach? I suppose aware users will end up with blocks such as `00-audit`, `99-policy` and others might be caught by surprise when they rename components? Especially within a single `initialize` block I would intuitively expect `request` blocks to run in order of declaration.

cipherboy: @satoqz Yeah, I'm updating this in my local copy. I started looking through the RFC and realized it was not aligned with #1506 lol.

This should be in server configuration order (so ordered according to the file, and then alphabetically across different files) --- this manual sorting was going to be a way of making the latter obvious (alphabetically across files also applies to across and within blocks).

cipherboy:
@satoqz Updated, let me know what you think!

website/content/docs/rfcs/self-init.mdx
```diff
@@ -0,0 +1,292 @@
...
+}
+EOP
+    }
+  }
+}
+```
+
+Equivalently in JSON this looks like:
+
+```json
```
cipherboy, line 114: @KrzysztofKornalewski-Reply This is an equivalent JSON configuration if you want to test it. :-)

satoqz (approved):

PR #1346, opened 2025-05-23
Display seal configuration info keys in logs

satoqz:
Seal configuration collects "info keys" from e.g. KMS configuration but fails to ultimately display these in log output because Go slices are complicated.

With this fix, the configuration is printed as intended:

```
==> OpenBao server configuration:

   PKCS#11 KMS Key Label: bao-root-key-aes
     PKCS#11 KMS Library: /opt/homebrew/lib/softhsm/libsofthsm2.so
 PKCS#11 KMS Token Label: OpenBao
Administrative Namespace:
             Api Address: http://127.0.0.1:8200
                     Cgo: enabled
         Cluster Address: https://127.0.0.1:8201
```

I'm a bit unhappy with how all the "PKCS#11 KMS" prefixes align here (same issue for different KMS types) but not sure what to do about it 😄

### DIFF
--- a/internalshared/configutil/pkcs11.go
+++ b/internalshared/configutil/pkcs11.go
@@ -19,8 +19,12 @@ func GetPKCS11KMSFunc(kms *KMS, opts ...wrapping.Option) (wrapping.Wrapper, map[
 	info := make(map[string]string)
 	if wrapperInfo != nil {
 		info["PKCS#11 KMS Library"] = wrapperInfo.Metadata["lib"]
-		info["PKCS#11 KMS Key Label"] = wrapperInfo.Metadata["key_label"]
-		info["PKCS#11 KMS Key ID"] = wrapperInfo.Metadata["key_id"]
+		if val, present := wrapperInfo.Metadata["key_id"]; present {
+			info["PKCS#11 KMS Key ID"] = val
+		}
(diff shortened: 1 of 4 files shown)


internalshared/configutil/pkcs11.go
```diff
@@ -19,8 +19,12 @@ func GetPKCS11KMSFunc(kms *KMS, opts ...wrapping.Option) (wrapping.Wrapper, map[
...
 	if wrapperInfo != nil {
 		info["PKCS#11 KMS Library"] = wrapperInfo.Metadata["lib"]
-		info["PKCS#11 KMS Key Label"] = wrapperInfo.Metadata["key_label"]
-		info["PKCS#11 KMS Key ID"] = wrapperInfo.Metadata["key_id"]
+		if val, present := wrapperInfo.Metadata["key_id"]; present {
+			info["PKCS#11 KMS Key ID"] = val
+		}
+		if val, present := wrapperInfo.Metadata["key_label"]; present {
+			info["PKCS#11 KMS Key Label"] = val
+		}
```
satoqz, line 27: Either of key id or key label may be empty, the user only needs to specify one, which would also be more typical than specifying both.

cipherboy (approved):
Thank you!

wslabosz-reply (approved):

PR #1347, opened 2025-05-23
Namespaces locking/unlocking implementation

### LINKED ISSUE #1181: Locking & Unlocking of Namespaces

wslabosz-reply:
Resolves #1181

Implements `bao namespace lock <path>` and `bao namespace unlock <path>` operations.

- Locking the namespace results in all requests to that namespace being rejected with `503` error code and a error message
```
bao namespace create test   
bao namespace lock test   
Key           Value
---           -----
unlock_key   <unlock_key>
(...)
bao write -ns=test cubbyhole/foo bar=baz
Error writing data to cubbyhole/foo: Error making API request.

Namespace: test/
URL: PUT http://127.0.0.1:8200/v1/cubbyhole/foo
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "test/" must be unlocked to gain access.
(...)
bao token create -ns=test
Error creating token: Error making API request.

Namespace: test/
URL: POST http://127.0.0.1:8200/v1/auth/token/create
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "test/" must be unlocked to gain access.
```

- Namespace locking respects the hierarchy of the namespaces, meaning that you cannot lock a namespace that is already locked or is a descendant of already locked namespace
```
// namespaces: "test/", "test/test2", "test/test2/test3"
bao namespace lock test/test2
// success
bao namespace lock test/test2/test3
Error locking namespace: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/sys/namespaces/api-lock/lock/test/test2/test3
Code: 400. Errors:

* cannot lock namespace "test/test2/test3/" as "test/test2/" is already locked

// but namespace "test" as an ancestor can be locked.
bao namespace lock test
// success
```

- Namespaces inherit the locked status, meaning that if any namespace in a upwards hierarchy is locked, then all the descendants cannot serve any requests
```
bao write -ns=test/test2/test3 cubbyhole/foo bar=baz
Error writing data to cubbyhole/foo: Error making API request.

Namespace: test/test2/test3/
URL: PUT http://127.0.0.1:8200/v1/cubbyhole/foo
Code: 503. Errors:

* API access to this namespace has been locked by an administrator - "test/" must be unlocked to gain access.
// notice how the error mentions the locked namespace, and not the namespace of the request.
```

- Namespace locks are persisted through sealing of the bao instance
```
bao namespace lookup test
Key                Value
---                -----
custom_metadata    map[]
id                 IiZhx7
locked             true
path               test/
tainted            false
uuid               8f4959b7-7733-4f0c-5a01-22b010e9407b

bao operator seal
bao operator unseal

bao namespace lookup test
Key                Value
---                -----
custom_metadata    map[]
id                 IiZhx7
locked             true
path               test/
tainted            false
uuid               8f4959b7-7733-4f0c-5a01-22b010e9407b
```

- Unlocking can be done using unlock key from the locking operation response, or using the token with `sudo` privileges
```
bao namespace lock test
Key           Value
---           -----
unlock_key   <unlock_key>
(...)
bao namespace unlock -unlock-key=<key> test
bao namespace lookup test
Key                Value
---                -----
custom_metadata    map[]
id                 IiZhx7
locked             false
path               test/
tainted            false
uuid               8f4959b7-7733-4f0c-5a01-22b010e9407b
(...)
bao namespace lock test
bao namespace unlock test // assuming your token has sudo capabilities
// success
```

- Namespace hierarchy has to be respected during unlocking, meaning that you have to unlock the top-most namespace in a hierarchy of locked namespaces
```
// notice how we have to lock namespaces starting from the most deeply nested one
bao namespace lock test/test2/test3
bao namespace lock test/test2
bao namespace lock test
(...)
bao namespace unlock test/test2/test3
Error unlocking namespace: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/sys/namespaces/api-lock/unlock/test/test2/test3
Code: 500. Errors:

* 1 error occurred:
        * cannot unlock "test/test2/test3/" with namespace "test/" being locked


bao namespace unlock test/test2      
Error unlocking namespace: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/sys/namespaces/api-lock/unlock/test/test2
Code: 500. Errors:

* 1 error occurred:
        * cannot unlock "test/test2/" with namespace "test/" being locked


bao namespace unlock test
// success
bao namespace unlock test/test2
// success
bao namespace unlock test/test2/test3
// success
```

-----
cc: @cipherboy @satoqz @phyrog @voigt @driif


cipherboy:
See previous discussions: https://github.com/openbao/openbao/pull/1300

vault/namespaces_store.go
```diff
@@ -34,6 +35,9 @@ const (
 	// Namespace storage location.
 	namespaceStoreSubPath = "core/namespaces/"
 
+	// Namespaces lock location.
+	namespaceLockPrefix = "core/namespace-lock"
```
cipherboy, line 39: Out of curiosity, why wouldn't this be added to the existing namespace entry? It seems easier than maintaining a separate entry just for locking.

vault/namespaces_store.go
```diff
@@ -63,6 +67,12 @@ type NamespaceStore struct {
 	logger hclog.Logger
 }
 
+// namespaceLock marks a locked namespace in storage.
+// It holds the unlock key for the locked namespace.
+type namespaceLock struct {
+	Key string `json:"key" mapstructure:"key"`
```
cipherboy, line 73: Especially since it is only a single entry, tbh.

vault/namespaces_store.go
```diff
@@ -785,3 +805,181 @@ func (ns *NamespaceStore) ResolveNamespaceFromRequest(nsHeader, reqPath string)
...
+	err = logical.WithTransaction(ctx, ns.storage, func(s logical.Storage) error {
+		return lockView.Delete(ctx, "")
+	})
+	if err != nil {
+		return fmt.Errorf("error deleting namespace lock: %w", err)
+	}
+
+	ns.namespacesByAccessor[namespace.ID].Locked = false
+	ns.namespacesByUUID[namespace.UUID].Locked = false
+	namespace.Locked = false
```
cipherboy, line 903: Ah, I see. 

I suppose we'd need to sanitize the actual lock key out from the return value to avoid external consumers from having this value. But that'd be a small change to `GetNamespace(...)` or whatever.

cipherboy (commented):
Just one commentary about storage layouts, happy to change my mind I suppose, but I feel like it is just extra work, tbh?

vault/namespaces_store.go
satoqz: Good point actually. @wslabosz-reply and I chose this storage path before we made the namespace store storage hierarchical, so this makes less sense now.

vault/namespaces_store.go
wslabosz-reply: @cipherboy
so what you observed is probably the whole story. Yes we don't need to store a lock as a separate entry to the namespace, but this way we don't need to sanitize the namespace entry, to not uncover the lock key by accident.

cipherboy: @wslabosz-reply If this were seal configuration, I might agree (that you'd want to use the parent's key), but not sure I follow here.

If we lock a namespace, why can't we put the unlock key on the namespace itself? Its namespace description _already_ lives in the parent's storage space. So there's functionally no difference, other than not having to reimplement hierarchical storage of locks.

But if the redaction concerns are too much, I can drop the matter. :-)

wslabosz-reply: > mean that we can lock /a/b/c from /d/e/f, if we have permissions? It seems weird, given the HashiCorp docs:

no, sorry, just a "ancestor" namespace, so we cannot do what you describe, as we won't be able to write a path that leads from `/d/e/f` to `/a/b/c`.

> use subpath, which seems to suggest it is only for a child namespace of our given namespace. We know ctx has a distant parent. So that means that a call to ModifyNamespaceByPath(...) should be sufficient for actual namespace entry modification purposes? Unless I'm missing something? GetLockingNamespace(...) will still function as expected, though.

`writeNamespace()` invoked by the `setNamespaceLocked()` invoked by the `ModifyNamespacesByPath()` assume that we modify the namespace, as it's *parent*, that assumption is required, otherwise we would save the namespace entry in a wrong location.

wslabosz-reply: > We should be strictly validating that entry is the direct child of parent and not a distant child.

but that's not how this documentation https://developer.hashicorp.com/vault/api-docs/system/namespaces#lock-namespace describes it

cipherboy: @wslabosz-reply Sorry, that was in the context of the `setNamespaceLocked()` call. The documentation is misleading, see the implementation:

```
// HasParent returns true if possibleParent is a parent namespace of n.
// Otherwise it returns false.
func (n *Namespace) HasParent(possibleParent *Namespace) bool {
    switch {    
    case possibleParent.Path == "":
        return true
    case n.Path == "":
        return false
    default:
        return strings.HasPrefix(n.Path, possibleParent.Path)
    }
}           
```

`/a` is a prefix of `/a/b/c` and so `(/a/b/c).HasParent(/a) == true`, which would mean if we ever call `ModifyNamespaceByEntry(...)` or `setNamespaceLocked(...)` with `/a` as context and `/a/b/c` as the target namespace to modify, we'd cause problems.

Luckily, we limited the modification APIs to one level, but we don't enforce this internally to the store.

cipherboy: @wslabosz-reply I tried a change locally, too much work :-D

wslabosz-reply: @cipherboy 
I can do that in a follow-up PR, but the question we've started with, still stands. I'm not sure whether we have the same understanding of the complications in the storage.

cipherboy: @wslabosz-reply The one test I'd probably add is to test that we can modify namespaces while locked, just in case that happens for some reason, without affecting status. But we should be able to do that based on the changes I added.

cipherboy: @wslabosz-reply Never mind, tests seem to hang, I have a deadlock somewhere?

wslabosz-reply: @cipherboy 
probably the removed
```
-	defer ns.lock.Unlock()
```
in `LockNamespace()`

cipherboy (approved):
I shall push a follow-up PR with my adjustments to use a single entry so we can review that separately.

Thanks @wslabosz-reply!

PR #1351, opened 2025-05-23
Renew `goodcertbadroot.pem`

satoqz:
Fixes `TestTLSInvalidRoot` and `TestTLSLeafCertInClientCAFile`, assuming my understanding of _the kind of cert_ it wants is correct 😄

New root cert was generated via:

```sh
bao write pki/root/generate/internal ttl=876000h
```

Valid for 100 years.

---

cc: @cipherboy


cipherboy (approved):
TBH, not chaining to Baltimore is probably a more realistic test anyways. Thank you @satoqz!

PR #1353, opened 2025-05-24
add hsm docker distribution to downloads page

voigt:
So far https://openbao.org/downloads/ did not reference OpenBaos HSM docker distribution. This PR adds docker pull instructions for HSM.

### DIFF
--- a/website/src/pages/downloads.tsx
+++ b/website/src/pages/downloads.tsx
@@ -204,30 +204,45 @@ const Docker = ({ version, name }) => {
     return (
         <Tabs>
             <TabItem value="quay" label="quay.io">
+                <p>Alpine Image Distribution</p>
                 <CodeBlock language="shell">
                     {`docker pull quay.io/openbao/openbao:${version.slice(1)}`}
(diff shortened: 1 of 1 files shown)


website/src/pages/downloads.tsx
```diff
@@ -204,31 +204,46 @@ const Docker = ({ version, name }) => {
     return (
         <Tabs>
             <TabItem value="quay" label="quay.io">
+                <p>Main Image Distribution</p>
```
cipherboy, line 207:
```suggestion
                <p>Alpine Image Distribution</p>
```

Maybe its my ex-Red Hat bias, but I'm not sure I'd call the Alpine one `main`. That said, there is an argument for consistency (distro name + distro name) or libc implementation (`musl image distribution` vs `glibc image distribution`) ;-) 

Your call :-)

cipherboy: The rest of the work is in #1087 if you want to take that on too :-)

cipherboy (approved):

Issue #1170, opened 2025-03-28
RFC: Per-Namespace Sealing [WIP]


voigt:
### Summary

> ⚠️ NOTE: WIP - filling out the document

Introduce a per-namespace sealing mechanism to extend OpenBaos multi-tenancy capabilities. Using Barrier + Keyring functionality, we want to create a mechanism to enable per-tenant encryption, which lets a tenant control their own encryption keys. By implementing per-namespace encryption, OpenBao can provide a more robust and flexible multi-tenancy solution, addressing the current limitations and offering tenants more significant control over their data security and availability.

### Problem Statement

The introduction of namespaces in OpenBao has paved the way for multi-tenancy capabilities, allowing multiple tenants to operate on the same instance while maintaining logical isolation of data. However, this isolation is currently limited to logical hierarchy, and several areas need further improvement to provide tenants with greater control, sovereignty, and security over their data. These limitations include:

* **Global Sealing:** The current implementation of (un)sealing affects all namespaces simultaneously. This means tenants lack independent control over the availability of their specific namespace.
* **Limited Namespace Control:** Tenants do not have full control over the lifecycle of their namespace, including the ability to lock, unlock, seal, or unseal independently of other tenants.
* **Shared Encryption Keys:** Tenants must rely on the OpenBao operator's cryptographic implementation, as they have no authority over the keys used to encrypt and decrypt their secrets.

To address these limitations, this RFC proposes implementing "Per-Namespace Sealing". This feature would provide the following benefits:

* Independent Sealing: Tenants would be able to seal and unseal their namespaces independently without affecting other tenants' data availability.
* Tenant-Controlled Encryption: Each namespace would have its own encryption keys, giving tenants full authority over the cryptographic security of their data.
* Enhanced Namespace Lifecycle Management: Tenants would gain more control over the lifecycle of their namespace, including the ability to lock, unlock, seal, and unseal as needed.

## Creating a Sealed namespace

A sealed namespace is created by providing a special option to the namespace create API call.
The default sealing method is [Shamir’s Secret Sharing Algorithm](https://en.wikipedia.org/wiki/Shamir%27s_Secret_Sharing). However, other [sealing configurations](https://openbao.org/docs/configuration/seal/) that are already supported for global sealing should also be supported for per-namespace sealing.

## Sealed Namespace

[...]

PR #1357, opened 2025-05-26
RFC - Per-namespace sealing

voigt:
RFC PR for #1170

### DIFF
--- a/website/content/docs/rfcs/index.mdx
+++ b/website/content/docs/rfcs/index.mdx
@@ -19,6 +19,8 @@ Steering Committee.
    to define initial mounts, auth, policies, and audit logs from a static
    configuration applied once on early startup, preventing the need to call
    manual `sys/init` except with manual unseal mechanisms.
+ - [Per-Namespace Sealing](/docs/rfcs/namespace-sealing), to add per-tenant 
+   encryption key separation.
  - [External Key Configuration for KMS and HSM Access](/docs/rfcs/external-keys),
    to add support for secure key material storage and interaction from
    secrets engines like PKI, Transit, and SSH.
@@ -76,6 +78,3 @@ Steering Committee.
  - [Best Practices for Integrating CEL in Auth, Secrets, and Elsewhere](/docs/rfcs/cel-best-practices),
    to discuss how Google's Common Expression Language should be integrated
    into OpenBao.
- - [Using CEL Roles For JWT/OIDC Auth](/docs/rfcs/cel-jwt), adds support for CEL roles in JWT Auth
-   Landed in [769](https://github.com/openbao/openbao/pull/769)
- - [Per-Namespace Sealing](/docs/rfcs/namespace-sealing), tbd
--- a/website/content/docs/rfcs/index.mdx
+++ b/website/content/docs/rfcs/index.mdx
@@ -76,3 +76,6 @@ Steering Committee.
  - [Best Practices for Integrating CEL in Auth, Secrets, and Elsewhere](/docs/rfcs/cel-best-practices),
    to discuss how Google's Common Expression Language should be integrated
    into OpenBao.
+ - [Using CEL Roles For JWT/OIDC Auth](/docs/rfcs/cel-jwt), adds support for CEL roles in JWT Auth
+   Landed in [769](https://github.com/openbao/openbao/pull/769)
+ - [Per-Namespace Sealing](/docs/rfcs/namespace-sealing), tbd
--- a/website/content/docs/rfcs/namespace-sealing.mdx
+++ b/website/content/docs/rfcs/namespace-sealing.mdx
@@ -1,0 +1,421 @@
+---
+sidebar_label: Per-Namespace Sealing
+description: |-
+  An OpenBao RFC to enable sealing of individual namespaces.
+---
+
+# Per-Namespace Sealing
+
+## Summary
+
+Introduce a per-namespace sealing mechanism to extend OpenBaos multi-tenancy capabilities. Using Barrier + Keyring functionality, we want to create a mechanism to enable per-tenant encryption, which lets a tenant control their own encryption keys. By implementing per-namespace encryption, OpenBao can provide a more robust and flexible multi-tenancy solution, addressing the current limitations and offering tenants more significant control over their data security and availability.
+
+## Problem Statement
+
+The introduction of namespaces in OpenBao has paved the way for multi-tenancy capabilities, allowing multiple tenants to operate on the same instance while maintaining logical isolation of data. However, this isolation is currently limited to logical hierarchy, and several areas need further improvement to provide tenants with greater control, sovereignty, and security over their data. These limitations include:
+
+* **Global Sealing:** The current implementation of (un)sealing affects all namespaces simultaneously. This means tenants lack independent control over the availability of their specific namespace.
+* **Limited Namespace Control:** Tenants do not have full control over the lifecycle of their namespace, including the ability to lock, unlock, seal, or unseal independently of other tenants.
+* **Shared Encryption Keys:** Tenants must rely on the OpenBao operator's cryptographic implementation, as they have no authority over the keys used to encrypt and decrypt their secrets.
+
+To address these limitations, this RFC proposes implementing "Per-Namespace Sealing". This feature would provide the following benefits:
+
+* Independent Sealing: Tenants would be able to seal and unseal their namespaces independently without affecting other tenants' data availability.
+* Tenant-Controlled Encryption: Each namespace would have its own encryption keys, giving tenants full authority over the cryptographic security of their data.
+* Enhanced Namespace Lifecycle Management: Tenants would gain more control over the lifecycle of their namespace, including the ability to lock, unlock, seal, and unseal as needed.
+
+
+## User-facing description
+
+### Creating a Sealed namespace
+
+A sealed namespace is created by providing a special option to the namespace create API call.
+The default sealing method is [Shamir’s Secret Sharing Algorithm](https://en.wikipedia.org/wiki/Shamir%27s_Secret_Sharing). However, other [sealing configurations](https://openbao.org/docs/configuration/seal/) that are already supported for global sealing should also be supported for per-namespace sealing.
+
+### Sealed Namespace
+
+Sealing an OpenBao Namespace pauses the Server's operations aimed at that namespace until a tenant operator decides to unseal it. Equal to global sealing, this means that a sealed namespace discards its in-memory namespace-root key (which is required to decrypt the dedicated namespace encryption key), which physically prevents responding to operations when sealed.
+
+If an unsealing is in progress, sealing the namespace will reset the unsealing process. Users will have to re-enter their shares of the namespace key again.
+
+Sealing an already sealed namespace has no effect.
+
+Namespace sealing is optional. By default, namespaces are created without per-namespace sealing, and they are part of the global sealing/unsealing process. Namespace sealing only affects custom namespaces. The root namespace and namespaces without a sealing config are not affected.
+
+While it has some behavior in common, sealing is a distinct feature from locking/unlocking a namespace.
+
+### Unsealing a Namespace
+
+OpenBao starts in a sealed state. Only after the global OpenBao unseal process is concluded, unsealing a namespace is possible.
+
+Unsealing a namespace allows the tenant to decrypt its data and allow operations on that namespace. No operations can be performed on a sealed namespace.
+
+Once an OpenBao namespace is unsealed, it remains unsealed until one of these things happens:
+
+* It is resealed via the API (see below).
+* The server is restarted.
+* OpenBao's storage layer encounters an unrecoverable error.
+
+### Auto-Unsealing
+
+Auto-Unsealing (non-interactive unsealing) is an option that can be used to automatically unseal a namespace with the namespace-root key and appropriate configuration. 
+
+Auto-Unsealing is considered an important feature, as it will be extremely handy for nested (sealed) namespaces, as structures like `sealed-n1/sealed-ns2/ns3/sealed-ns4` are possible.
+
+As the default sealing mechanism is Shamir, auto-unsealing is not the default. Shamir requires a tenant to interactively enter key shares, which makes the auto-unsealing impossible.
+
+### Namespace Lifecycle Options
+
+Per-namespace encryption moves parts of the responsibility away from the operator, closer to the tenant. Therefore, some life cycle features known from the global sealing/unsealing level also need to exist on namespace level:
+
+* Rekeying: In case a namespace-root-key needs to be changed (e.g. because of a leak), revoking the old and creating a new key needs to be possible.
+  * Rekeying can only be started with valid namespace key
+* Rekeying of Unseal Key
+  * Rekeying of Unseal Keys can only be started with a valid namespace key
+* Key Recovery
+* Rekeying of recovery key
+
+The global root key is unable to decrypt data of sealed namespaces.
+Only users with appropriate permissions can trigger any kind of rekey.
+
+
+## Technical Description
+
+### Unseal Process
+
+#### Key Hierarchy
+
+##### Existing Key Encryption-Design
+
+Currently, OpenBao has a three-key hierarchy:
+
+```mermaid
+graph LR
+A1[Share 1] -->|Recovers| B[Shamir's Key Encryption Key]
+A2[Share 2] -->|Recovers| B[Shamir's Key Encryption Key]
+A3[Share 3] -->|Recovers| B[Shamir's Key Encryption Key]
+    B -->|Encrypts| RK[Root Key]
+    RK -->|Encrypts| RBK[Root Barrier Keyring]
+    RBK -->|Encrypts| ESK[Encrypted Shamir's Key Encryption Key]
+    RBK -->|Encrypts| ERK[Encrypted Root Key]
+```
+
+This is the same for auto-unseal keys, swapping Shamir's recovery of the shard's combined key encryption key (KEK) for a key stored in an HSM or KMS solution.
+
+**Proposal for a Namespace scoped Barrier-Keyring:**
+
+```mermaid
+graph LR
+RBE2[Root Barrier Keyring] -->|Encrypts| B[Shamir's Key Encryption Key]
+A1[Share 1] -->|Recovers| B[Namespace Shamir's Key Encryption Key]
+A2[Share 2] -->|Recovers| B[Namespace Shamir's Key Encryption Key]
+A3[Share 3] -->|Recovers| B[Namespace Shamir's Key Encryption Key]
+    B -->|Encrypts| C[Namespace-Root-Key]
+    C -->|Encrypts| RBE[Namespace-Barrier-Keyring]
+    RBE -->|Encrypts| ESK[Encrypted Namespace Shamir's Key Encryption Key]
+    RBE -->|Encrypts| ENSRK[Encrypted Namespace Root Key]
+```
+
+With Per-Namespace Sealing, the responsibility of Root Barrier Keyring will be extended. Root Barrier Keyring will be used to encrypt the Shamirs Key Encryption Key (KEK) for a Namespace. This has the implication that a Namespace Shamirs KEK first needs to be decrypted (Global needs to be unsealed) before a namespace can be unsealed.
+
+### Unsealing Process
+
+1. OpenBao will attempt an unblocking unseal of each sealed namespace (all namespaces with sealing config), skipping Shamir seals
+   - on success: namespace will work as expected and requests can be answered
+   - on fail: namespace remains sealed; returns applicable errors; status can be retrieved from namespace-status endpoint; periodic retry (as HSM can be temporarily unavailable etc.)
+
+1. OpenBao global unseal is successful
+2. OpenBao will attempt an unblocking unseal of each sealed namespace (all namespaces with sealing config), skipping Shamir seals
+   - on success: namespace will work as expected and requests can be answered
+   - on fail: namespace remains sealed; returns applicable errors; status can be retrieved from namespace-status endpoint; periodic retry (as HSM can be temporarily unavailable etc.)3. For namespaces that could not be auto-unsealed (and do not have a sealing config), OpenBao allows manual shamir unseal via the respective namespace specific api endpoint.  
+   Note: OpenBao will not globally block if single Namespaces remain sealed
+4. When various namespace-barrier or namespace-root-key rotations have been performed, unseal will locally perform the rotation finalisation steps outlined in the rotation section below.
+
+### Encryption
+
+While namespaces logically can be used in a hierarchy, on storage level they are laid out in a flat list. This means we can encrypt each namespace individually and on a global level.
+
+<img width="1433" alt="Image" src="https://github.com/user-attachments/assets/72c71f7c-6e5d-4173-96a9-f4943dd3497a" />
+
+At the level of the global sealing we use the root-key to encrypt the list of namespaces including their metadata and configuration. With access to this data, each individual namespace can be decrypted by auto-unseal or a tenant with the namespace-root-key.
+
+For data this means, we encrypt it twice: first with the namespace encryption key, second with the root encryption key.
+
+**Pros:**
+leaked namespace encryption keys are less of an issue as data at rest is still protected by root encryption key
+
+**Cons:**
+in backup/restore scenarios the root encryption key must be known
+
+### Storage Considerations
+
+We will introduce a new namespace scoped root key `ns-root-key`, which has similar responsibility as the root key on global level.
+
+We do not need to change the existing root key entries, as they are unrelated. However, we need namespace specific entries for keys and config. The idea is, to generally follow the storage logic of the [Parallel Unseal RFC](https://github.com/openbao/openbao/issues/1021), so we will be semantically in harmony with future changes to  global sealing mechanisms. Please note, this RFC does not support parallel unsealing of namespaces.
+
+| **Location** | **Description** |
+|------------ | -------------|
+|`namespaces/<ns-uuid>/core/seals/shamir-config` OR <br/>`namespaces/<ns-uuid>/core/seals/recovery-config` | It stores the configuration of the seal itself.<br/><br/>Can be shamir-config or config for auto-unseal mechanism.<br/><br/>This is a *raw* (unencrypted, from Barrier's PoV) storage entry, as it needs to be accessible in case of a sealed namespace. |
+|`namespaces/<ns-uuid>/core/seals/encrypted-ns-root` | Used for storing the encrypted namespace-root key.<br/><br/>This is a raw (unencrypted, from Barrier's PoV) storage entry; however, this raw entry is encrypted with the seal mechanism's key information. |
+|`namespaces/<ns-uuid>/core/seals/kek`<br/>(when applicable) | Storing a copy of the Shamir’s key for Shamir rekey operations and to restore raft snapshots.<br/><br/>This is a **barrier-encrypted** storage entry. |
+|`namespaces/<ns-uuid>/core/seals/autounseal-confi` | This allows for retrieval of the seal mechanisms and determining if configuration has changed since the last startup. <br /><br />This is a **barrier-encrypted** storage entry. |
+|`namespaces/<ns-uuid>/core/seals/barrier-keyrin` | We'll store a copy of the barrier keyring encrypted with that seal's current namespace-root key.<br/><br/>(Enables decryption of the current latest root key.) |
+|`namespaces/<ns-uuid>/core/seals/current-roo` | We'll store an encrypted copy of the seal's current namespace-root key<br/><br/>This is a **barrier-encrypted** storage entry. (Allows other nodes to rotate the barrier keyring keys.) |
+|`namespaces/<ns-uuid>/core/ns-root-ke` | It stores the cluster’s latest namespace specific root key.<br/><br/>This is a **barrier-encrypted** storage entry. |
+
+### Namespace specific Keyring and Barrier
+
+To let a tenant control their own encryption keys, we will need to introduce a per-namespace keyring and barrier. While I assume we can reuse code in [vault/keyring.go](https://github.com/openbao/openbao/blob/main/vault/keyring.go) and [vault/barrier.go](https://github.com/openbao/openbao/blob/main/vault/barrier.go#L73), we might consider some naming/semantic changes. A namespace specific Keyring does not have a rootKey, but a namespaceRootKey. Similarly, some method names of the SecurityBarrierCore interface might be misleading in the context of a namespace (e.g. SetRootKey, ReloadRootKey ).
+
+### Migration
+
+To keep this feature simple, the migration path from an existing normal namespace to a sealed namespace is not discussed here. The proposed migration path would be via export/import.
+
+### API & CLI Changes
+
+We need to extend the create namespace API to enable creating sealed namespaces.
+We need to create new endpoints per namespace to trigger sealing and unsealing.
+The functionality needs to be exposed by the HTTP API and the bao namespace subcommand of the CLI.
+
+If a namespace is sealed, any requests to this namespace's resources should be handled similar to a locked namespace.
+
+#### Creating a Sealed Namespace
+
+| **Method** | **Path**                |
+| ---------- | ----------------------- |
+| `POST`     | `/sys/namespaces/:path` |
+
+##### Additional Parameters
+
+The following Parameter will be added to the existing list of parameters:
+
+- seals (`[]map<string|string>: nil`) - A list containing maps of user-provided string to string key value pairs, containing sealing configuration.
+
+<details>
+
+<summary>Collapse Example</summary>
+
+##### Payload
+
+```json
+{
+  "custom_metadata": {
+    "foo": "bar"
+  },
+  "seals": [{
+    "type": "shamir",
+    "name": "default",
+    "key_shares": 5,
+    "key_threshold": 3,
+  }]
+}
+```
+
+##### Example
+
+```bash
+$ curl \
+    --header "X-Vault-Token: ..." \
+    --request POST \
+    --data @payload.json \
+    http://127.0.0.1:8200/v1/sys/namespaces/sealednamespace1
+{
+  "request_id": "1871d69e-38fe-1804-68be-a4a049e349cd",
+  "lease_id": "",
+  "renewable": false,
+  "lease_duration": 0,
+  "data": {
+    "custom_metadata": {
+      "foo": "bar"
+    },
+    "id": "1XFxHV",
+    "path": "foo/",
+    "uuid": "35c784cf-90b8-42a6-dd29-c8e33edfb4db"
+    unseal_keys: [
+        {
+            key: "sP/4C/fwIDjJmHEC2bi/1Pa43uKhsUQMmiB31GRzFc0R",
+            number: "1"
+        },
+        {
+            key: "kHkw2xTBelbDFIMEgEC8NVX7NDSAZ",
+            number: "2"
+        },
+        {
+            key: "+1+1ZnkQDfJFHDZPRq0wjFxEuEEHxDDOQxa8JJ/AYWcb",
+            number: "3"
+        },
+        {
+            key: "cewseNJTLovmFrgpyY+9Hi5OgJlJgGGCg7PZyiVdPwN0",
+            number: "4"
+        },
+        {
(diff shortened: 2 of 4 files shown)


website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+ "token": "token"
+}
+```
+
+##### Example
+
+```bash
+$ curl \
+  --header "X-Vault-Token: ..." \
+  --request PATCH \
```
phyrog, line 319:
```suggestion
  --request POST \
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+- Creating a sealed namespace requires a root key. After creation, (un)sealing of the namespace requires the namespace key. A root key can not (un)seal.
+
+#### Retrieve Sealing status of a Namespace
+
+Note: contrary to existing /sys/seal-status, the namespace specific endpoint is authenticated.
+
+| **Method** | **Path**           |
+| ---------- | ------------------ |
+| `GET`      | `/sys/namespaces/:path/seal-status` |
```
phyrog, line 334:
```suggestion
| `GET`      | `/sys/namespaces/:name/seal-status` |
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+  --header "X-Vault-Token: ..." \
+  --request GET \
+  http://127.0.0.1:8200/v1/sys/namespaces/sealednamespace1/seal-status
+```
+
+</details>
+
+##### Additional Endpoints
+
+Furthermore, the following endpoints will need to be adjusted, so that they are (optionally) namespace aware, by providing X-Vault-Namespace and X-Vault-Token headers.
```
phyrog, line 353: I assume that means we need to provide them too under `/sys/namespaces/:name`?

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+</details>
+
+Sealing stanzas for KMS and PKCS#11 follow [the existing ones](https://openbao.org/docs/configuration/seal/).
+
+For PKCS#11 there are some more restrictions: first, OpenBao requires a physical connection to an existing HSM. Second, it can only support the PKCS#11 library that is already present on the machine OpenBao is running on. Therefore, the lib field should not be configurable.
+
+Note: [Parallel Unsealing](https://github.com/openbao/openbao/issues/1021) on namespace level is considered a future improvement
+
+#### Sealing of a Namespace
+
```
phyrog, line 272: I think this is not mentioned anywhere else yet.

```suggestion

Sealing a namespace also seals all of it's children recursively.

```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+| **Location** | **Description** |
+|------------ | -------------|
+|`namespaces/<ns-uuid>/core/seals/shamir-config` OR <br/>`namespaces/<ns-uuid>/core/seals/recovery-config` | It stores the configuration of the seal itself.<br/><br/>Can be shamir-config or config for auto-unseal mechanism.<br/><br/>This is a *raw* (unencrypted, from Barrier's PoV) storage entry, as it needs to be accessible in case of a sealed namespace. |
+|`namespaces/<ns-uuid>/core/seals/encrypted-ns-root` | Used for storing the encrypted namespace-root key.<br/><br/>This is a raw (unencrypted, from Barrier's PoV) storage entry; however, this raw entry is encrypted with the seal mechanism's key information. |
+|`namespaces/<ns-uuid>/core/seals/kek`<br/>(when applicable) | Storing a copy of the Shamir’s key for Shamir rekey operations and to restore raft snapshots.<br/><br/>This is a **barrier-encrypted** storage entry. |
+|`namespaces/<ns-uuid>/core/seals/autounseal-confi` | This allows for retrieval of the seal mechanisms and determining if configuration has changed since the last startup. <br /><br />This is a **barrier-encrypted** storage entry. |
```
wslabosz-reply, line 161:
```suggestion
|`namespaces/<ns-uuid>/core/seals/autounseal-config` | This allows for retrieval of the seal mechanisms and determining if configuration has changed since the last startup. <br /><br />This is a **barrier-encrypted** storage entry. |
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+|`namespaces/<ns-uuid>/core/seals/barrier-keyrin` | We'll store a copy of the barrier keyring encrypted with that seal's current namespace-root key.<br/><br/>(Enables decryption of the current latest root key.) |
+|`namespaces/<ns-uuid>/core/seals/current-roo` | We'll store an encrypted copy of the seal's current namespace-root key<br/><br/>This is a **barrier-encrypted** storage entry. (Allows other nodes to rotate the barrier keyring keys.) |
```
wslabosz-reply, line 163:
```suggestion
|`namespaces/<ns-uuid>/core/seals/current-root` | We'll store an encrypted copy of the seal's current namespace-root key<br/><br/>This is a **barrier-encrypted** storage entry. (Allows other nodes to rotate the barrier keyring keys.) |
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+| **Method** | **Path**                |
+| ---------- | ----------------------- |
+| `POST`     | `/sys/namespaces/:path` |
+
+##### Additional Parameters
+
+The following Parameter will be added to the existing list of parameters:
+
+- seals (`[]map<string|string>: nil`) - A list containing maps of user-provided string to string key value pairs, containing sealing configuration.
```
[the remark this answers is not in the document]
wslabosz-reply: to be exactly precise it would be something like `map[string]map[string]interface{}` as the properties can be both strings and ints

website/content/docs/rfcs/namespace-sealing.mdx
wslabosz-reply: I think so

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+#### Unsealing of a Namespace
+
+| **Method** | **Path**                       |
+| ---------- | ------------------------------ |
+| `PATCH`    | `/sys/namespaces/:path/unseal` |
+
+##### Additional Parameters
+
+- token (`string: nil`) - A Shamirs token.
```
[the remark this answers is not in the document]
wslabosz-reply: my nit: can we call it `key` and change the description to `A Shamir's key share`?

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+##### Additional Endpoints
+
+Furthermore, the following endpoints will need to be adjusted, so that they are (optionally) namespace aware, by providing X-Vault-Namespace and X-Vault-Token headers.
+
+- `/sys/generate-recovery-token`
+- `/sys/generate-root`
+- `/sys/key-status`
+- `/sys/rekey`
+- `/sys/rekey-recovery-key`
```
phyrog: Does it actually make sense to add the recovery endpoints? I don't think they make much sense in the namespace context right now.

cipherboy: @voigt @phyrog I think we have the opportunity to make either approach viable. When it comes time to do non-hierarchical namespaces (or, when we do parallel unseal), we'll definitely need the recovery key endpoints.

However, for the existing namespaces, it goes either way. It is mostly about whether a tenant should have the ability to create a root key in their own environment or not. If they don't, what is the path for bootstrapping a sealed namespace? They have to rely on the parent namespace (root) to provision initial auth/policies/... and remediating any issues, versus having a privileged fallback token to rely on.

Thoughts?

wslabosz-reply: As a general rule of thumb I'd like us to go with hardest route for now, if it will mean that we do not introduce any unnecessary breaking changes (storage migration are awful).

> They have to rely on the parent namespace (root) to provision initial auth/policies/... and remediating any issues, versus having a privileged fallback token to rely on.

I'd hesitate to impose on the operator to bootstrap the tenant namespace, although it will come down to how much setup we are talking about, but still prefer the bare minimum. I'd consider going this route, if a second degree namespace (so with a parent that is not the root) could be bootstrapped by the parent, but right now I don't think this is possible(?). Considering this I'd say that a recovery token fallback is the most reasonable approach.

Lastly, I'd say that if with https://github.com/openbao/openbao/issues/1340 we are going to be able to setup the namespaces (at least one level deep in hierarchy) then I'd be happy to exclude recovery token from this initial work (before parallel-unseal).

My 0.018€

cipherboy: @wslabosz-reply You certainly can use #1340 to initialize and set up a sealed namespace, but it only will initialize the whole system, so cannot be used to onboard a future namespace post-initialization. (Maybe it will be useful for out-of-hierarchy namespaces but not for arbitrary child namespaces.)

wslabosz-reply: Will let @voigt @phyrog @rencooo speak, but I think we should add support for recovery token sooner than later

phyrog: IMO `/sys/generate-root` is reasonable to support/implement now, but I still don't really see a use case for recovery keys **yet**. Would they only be used if OpenBao is in recovery mode to decrypt namespaces? Is the endpoint only available in recovery mode? I think there is a lot of details unclear on how all of that works.

cipherboy: @phyrog Yes, only in recovery mode:

https://github.com/openbao/openbao/blob/1e3ee07dfe96935e2c562dd53b806fbba95a83b2/http/handler.go#L160-L168

I think you cannot use root tokens in recovery mode and recovery tokens in normal mode; if we allow one, IMHO, it makes sense to allow both. A root namespace recovery mode token will certainly work against a child namespace, so maybe that is sufficient outside of non-hierarchical namespaces.

I suppose, the main difference with recovery tokens is their use requires affecting all tenants anyways (by virtue of being in recovery mode) and thus wouldn't be useful for any namespace but the root (which, presumably, the binary operator controls).

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+* The server is restarted.
+* OpenBao's storage layer encounters an unrecoverable error.
+
+### Auto-Unsealing
+
+Auto-Unsealing (non-interactive unsealing) is an option that can be used to automatically unseal a namespace with the namespace-root key and appropriate configuration. 
+
+Auto-Unsealing is considered an important feature, as it will be extremely handy for nested (sealed) namespaces, as structures like `sealed-n1/sealed-ns2/ns3/sealed-ns4` are possible.
+
+As the default sealing mechanism is Shamir, auto-unsealing is not the default. Shamir requires a tenant to interactively enter key shares, which makes the auto-unsealing impossible.
```
phyrog: We were discussing about this today and found kind of a conceptual problem with auto-unsealing and the purpose of sealed namespaces. The purpose of sealing a namespace, as stated in the Problem Statement, is:

> giving tenants full authority over the cryptographic security of their data.

Now, when we introduce auto-unsealing, we need to store the auto-unseal config somewhere where it can be read before the namespace is unsealed (because it is needed for the unsealing), e.g. encrypted by the parent namespace barrier. Now we need to trust the parent namespace, given that the auto-unseal config contains sensitive data (cloud credentials, HSM pin, ...), which kind of defeats the purpose of sealed namespaces.

So IMO there's two options:
- allow auto-unsealing, store the config encrypted by the parent namespace and degrade the security and data protection of the sealed namespace
- only allow shamir seals for namespaces, which makes nested sealed namespaces super annoying to work with

Or am I missing a third option?

@cipherboy Could use your thoughts on this.

cc: @wslabosz-reply @voigt @rencooo

cipherboy: @phyrog While this is strictly true:

> given that the auto-unseal config contains sensitive data (cloud credentials, HSM pin, ...), which kind of defeats the purpose of sealed namespaces.

_Useful_ KMS providers will still let you:

1. [Rotate](https://cloud.google.com/iam/docs/key-rotation) these credentials whenever you want to, thereby revoking the provider's access.
2. [Audit](https://cloud.google.com/kms/docs/audit-logging) access requests to/from the service.

&c. -- this minimizes the risk.

No matter what you do, you have to trust the provider or you wouldn't be using a cloud service. Presumably, in a real-world situation, you'd have contractual agreements in place with the provider to ensure they do not misuse this KMS or retain the decrypted keyring keys at all. Perhaps this would extend to binary/code audit capabilities to show that this OpenBao instance has not been modified to retain these anyways.

This is true even for Shamir's fwiw: you have to trust the cloud provider not to retain copies of the barrier keyring, the recovered Shamir keys, the decrypted root key. And it is true of GCP Secrets Manager/... 

So at some point, you have to put trust somewhere. If you don't trust the cloud provider, move your workloads back on-prem.

In a much later hypothetical future with secure remote computation, you could avoid this trust, but IMHO, it is impossible or severely impractical at the current moment. 

---

(And, encrypted in barrier storage is nominally more secure than unencrypted in the configuration file like is handled today.)

website/content/docs/rfcs/index.mdx
```diff
@@ -44,3 +44,4 @@ Steering Committee.
    into OpenBao.
  - [Using CEL Roles For JWT/OIDC Auth](/docs/rfcs/cel-jwt), adds support for CEL roles in JWT Auth
    Landed in [769](https://github.com/openbao/openbao/pull/769)
+ - [Per-Namespace Sealing](/docs/rfcs/namespace-sealing), tbd
```
cipherboy: Do you mind placing this in the new "Accepted" section at the top? ty!

```suggestion
 - [Per-Namespace Sealing](/docs/rfcs/namespace-sealing), to add per-tenant
   encryption key separation. 
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+sidebar_label: Per-Namespace Sealing
+description: |-
+  An OpenBao RFC to enable sealing of individual namespaces.
+---
+
+# Per-Namespace Sealing
+
+## Summary
+
+Introduce a per-namespace sealing mechanism to extend OpenBaos multi-tenancy capabilities. Using Barrier + Keyring functionality, we want to create a mechanism to enable per-tenant encryption, which lets a tenant control their own encryption keys. By implementing per-namespace encryption, OpenBao can provide a more robust and flexible multi-tenancy solution, addressing the current limitations and offering tenants more significant control over their data security and availability.
```
cipherboy:
```suggestion
Introduce a per-namespace sealing mechanism to extend OpenBao's tenant separation capabilities. Expanding the Barrier + Keyring functionality, we want to create a mechanism to enable per-tenant encryption, which lets a tenant control their own seal keys in their namespace. OpenBao can thus provide a more robust and flexible multi-tenancy solution, addressing the shared encryption limitations and offering tenants more significant control over their data security and availability.
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+* **Limited Namespace Control:** Tenants do not have full control over the lifecycle of their namespace, including the ability to lock, unlock, seal, or unseal independently of other tenants.
+* **Shared Encryption Keys:** Tenants must rely on the OpenBao operator's cryptographic implementation, as they have no authority over the keys used to encrypt and decrypt their secrets.
+
+To address these limitations, this RFC proposes implementing "Per-Namespace Sealing". This feature would provide the following benefits:
+
+* Independent Sealing: Tenants would be able to seal and unseal their namespaces independently without affecting other tenants' data availability.
+* Tenant-Controlled Encryption: Each namespace would have its own encryption keys, giving tenants full authority over the cryptographic security of their data.
+* Enhanced Namespace Lifecycle Management: Tenants would gain more control over the lifecycle of their namespace, including the ability to lock, unlock, seal, and unseal as needed.
```
cipherboy:
```suggestion
* Enhanced Namespace Lifecycle Management: Tenants would gain more control over the lifecycle of their namespace, including the ability to seal and unseal as needed.
```

Locking and unlocking already landed and is a system-level feature; it doesn't prevent the operator from bypassing this lock at the storage level.

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+
+## User-facing description
+
+### Creating a Sealed namespace
```
cipherboy:
```suggestion
### Creating a Sealed Namespace
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+A sealed namespace is created by providing a special option to the namespace create API call.
+The default sealing method is [Shamir’s Secret Sharing Algorithm](https://en.wikipedia.org/wiki/Shamir%27s_Secret_Sharing). However, other [sealing configurations](https://openbao.org/docs/configuration/seal/) that are already supported for global sealing should also be supported for per-namespace sealing.
+
+### Sealed Namespace
+
+Sealing an OpenBao Namespace pauses the Server's operations aimed at that namespace until a tenant operator decides to unseal it. Equal to global sealing, this means that a sealed namespace discards its in-memory namespace-root key (which is required to decrypt the dedicated namespace encryption key), which physically prevents responding to operations when sealed.
```
cipherboy:
```suggestion
Sealing an OpenBao Namespace pauses the Server's operations aimed at that namespace until a tenant operator decides to unseal it. Equal to global sealing, this means that a sealed namespace discards its in-memory namespace barrier keyring and namespace root key (which is required to decrypt the dedicated namespace barrier keyring), which physically prevents responding to operations when sealed.
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+If an unsealing is in progress, sealing the namespace will reset the unsealing process. Users will have to re-enter their shares of the namespace key again.
```
cipherboy: Is this true of the `sys/seal` endpoint? :thinking:  https://openbao.org/api-docs/system/seal/ is not clear.

cipherboy: It seems like it'd contradict the line below:

> Sealing an already sealed namespace has no effect.

(because an in-progress unsealing namespace is still sealed)

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+Sealing an already sealed namespace has no effect.
+
+Namespace sealing is optional. By default, namespaces are created without per-namespace sealing, and they are part of the global sealing/unsealing process. Namespace sealing only affects custom namespaces. The root namespace and namespaces without a sealing config are not affected.
```
cipherboy:
```suggestion
Namespace sealing is optional. By default, namespaces are created without per-namespace sealing and they are part of their parent namespace's sealing/unsealing process. Namespace sealing only affects new namespaces created with this option. The root namespace and existing namespaces without a sealing config are not affected.
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+OpenBao starts in a sealed state. Only after the global OpenBao unseal process is concluded, unsealing a namespace is possible.
+
+Unsealing a namespace allows the tenant to decrypt its data and allow operations on that namespace. No operations can be performed on a sealed namespace.
+
+Once an OpenBao namespace is unsealed, it remains unsealed until one of these things happens:
+
+* It is resealed via the API (see below).
+* The server is restarted.
+* OpenBao's storage layer encounters an unrecoverable error.
```
cipherboy, line 90:
```suggestion
* OpenBao's storage layer for the namespace encounters an unrecoverable error.
```

cipherboy: Or is this global and triggers a seal of all namespaces?

website/content/docs/rfcs/namespace-sealing.mdx
cipherboy: @voigt @phyrog Going further even, I'd like to see a "light" version of per-namespace seals using the `static` auto-unseal mechanism. Perhaps keys for this would be stored on the namespace object (like the lock sentinel). The result is namespaces can be completely transparently encrypted with a per-tenant encryption key without (much) performance overhead and behave like a regular namespace, just that it has tenant-level separation by default.

The operator wouldn't need recovery key shares either for that matter (and https://openbao.org/docs/rfcs/authenticated-rekey/ can apply to them).

But this can be a future improvement, not required for first implementation pass.

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+### Namespace Lifecycle Options
+
+Per-namespace encryption moves parts of the responsibility away from the operator, closer to the tenant. Therefore, some life cycle features known from the global sealing/unsealing level also need to exist on namespace level:
+
+* Rekeying: In case a namespace-root-key needs to be changed (e.g. because of a leak), revoking the old and creating a new key needs to be possible.
+  * Rekeying can only be started with valid namespace key
+* Rekeying of Unseal Key
+  * Rekeying of Unseal Keys can only be started with a valid namespace key
+* Key Recovery
```
cipherboy: Is this the recovery key PGP-encrypted backup procedure or something else?

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+Auto-Unsealing is considered an important feature, as it will be extremely handy for nested (sealed) namespaces, as structures like `sealed-n1/sealed-ns2/ns3/sealed-ns4` are possible.
+
+As the default sealing mechanism is Shamir, auto-unsealing is not the default. Shamir requires a tenant to interactively enter key shares, which makes the auto-unsealing impossible.
+
+### Namespace Lifecycle Options
+
+Per-namespace encryption moves parts of the responsibility away from the operator, closer to the tenant. Therefore, some life cycle features known from the global sealing/unsealing level also need to exist on namespace level:
+
+* Rekeying: In case a namespace-root-key needs to be changed (e.g. because of a leak), revoking the old and creating a new key needs to be possible.
+  * Rekeying can only be started with valid namespace key
```
cipherboy: Sorry, not sure I understand this? Rekeying can be progressed with a valid namespace (recovery) key _share_, but initialization should be authenticated under https://openbao.org/docs/rfcs/authenticated-rekey/.

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+* Rekeying of Unseal Key
+  * Rekeying of Unseal Keys can only be started with a valid namespace key
```
cipherboy: I think this happens automatically for auto-unseal when the seal information changes, or, from the above rekey for Shamir's.

I'm curious how you'll store config here?

Also, please call these rotations, to align with the new guidance in https://openbao.org/docs/rfcs/authenticated-rekey/.

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+A1[Share 1] -->|Recovers| B[Namespace Shamir's Key Encryption Key]
+A2[Share 2] -->|Recovers| B[Namespace Shamir's Key Encryption Key]
+A3[Share 3] -->|Recovers| B[Namespace Shamir's Key Encryption Key]
+    B -->|Encrypts| C[Namespace-Root-Key]
+    C -->|Encrypts| RBE[Namespace-Barrier-Keyring]
+    RBE -->|Encrypts| ESK[Encrypted Namespace Shamir's Key Encryption Key]
+    RBE -->|Encrypts| ENSRK[Encrypted Namespace Root Key]
+```
+
+With Per-Namespace Sealing, the responsibility of Root Barrier Keyring will be extended. Root Barrier Keyring will be used to encrypt the Shamirs Key Encryption Key (KEK) for a Namespace. This has the implication that a Namespace Shamirs KEK first needs to be decrypted (Global needs to be unsealed) before a namespace can be unsealed.
```
cipherboy:
```suggestion
With Per-Namespace Sealing, the responsibility of Root Barrier Keyring will be extended. The parent namespace's Barrier Keyring will be used to encrypt the sealed namespace's stored Shamirs Key Encryption Key (KEK). This has the implication that a Namespace Shamirs KEK first needs to be decrypted (implying that the parent or root namespace needs to be unsealed) before a namespace can be unsealed.
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+   - on success: namespace will work as expected and requests can be answered
+   - on fail: namespace remains sealed; returns applicable errors; status can be retrieved from namespace-status endpoint; periodic retry (as HSM can be temporarily unavailable etc.)3. For namespaces that could not be auto-unsealed (and do not have a sealing config), OpenBao allows manual shamir unseal via the respective namespace specific api endpoint.  
+   Note: OpenBao will not globally block if single Namespaces remain sealed
+4. When various namespace-barrier or namespace-root-key rotations have been performed, unseal will locally perform the rotation finalisation steps outlined in the rotation section below.
+
+### Encryption
+
+While namespaces logically can be used in a hierarchy, on storage level they are laid out in a flat list. This means we can encrypt each namespace individually and on a global level.
+
+<img width="1433" alt="Image" src="https://github.com/user-attachments/assets/72c71f7c-6e5d-4173-96a9-f4943dd3497a" />
```
cipherboy: Would love this as a mermaid diagram, but if not, that's fine.

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+At the level of the global sealing we use the root-key to encrypt the list of namespaces including their metadata and configuration. With access to this data, each individual namespace can be decrypted by auto-unseal or a tenant with the namespace-root-key.
+
+For data this means, we encrypt it twice: first with the namespace encryption key, second with the root encryption key.
```
cipherboy: We don't, though, right? We never encrypt data entries with the root encryption key.

At least, that's the impression I was under from #1170.

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+**Pros:**
+leaked namespace encryption keys are less of an issue as data at rest is still protected by root encryption key
+
+**Cons:**
+in backup/restore scenarios the root encryption key must be known
```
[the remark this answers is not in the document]
cipherboy: As above, I think we may just want to remove the pros/cons and put that in the security section.

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+|`namespaces/<ns-uuid>/core/seals/current-roo` | We'll store an encrypted copy of the seal's current namespace-root key<br/><br/>This is a **barrier-encrypted** storage entry. (Allows other nodes to rotate the barrier keyring keys.) |
+|`namespaces/<ns-uuid>/core/ns-root-ke` | It stores the cluster’s latest namespace specific root key.<br/><br/>This is a **barrier-encrypted** storage entry. |
+
+### Namespace specific Keyring and Barrier
+
+To let a tenant control their own encryption keys, we will need to introduce a per-namespace keyring and barrier. While I assume we can reuse code in [vault/keyring.go](https://github.com/openbao/openbao/blob/main/vault/keyring.go) and [vault/barrier.go](https://github.com/openbao/openbao/blob/main/vault/barrier.go#L73), we might consider some naming/semantic changes. A namespace specific Keyring does not have a rootKey, but a namespaceRootKey. Similarly, some method names of the SecurityBarrierCore interface might be misleading in the context of a namespace (e.g. SetRootKey, ReloadRootKey ).
```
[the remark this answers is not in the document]
cipherboy: I think the other observation I'm having now is that ideally the root namespace is "just" another namespace, it just happens to have no parent and thus lacks a parent seal, writing directly to physical.

Also, this sets up for out-of-hierarchy namespaces: sealing works as expected, with no parent.

(The message here is just really that terminology is/should be consistent between all use cases)

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+### Lazy Loading of Mounts, Namespaces
+
+Coupled with per-namespace seal mechanisms, one issue with this is that it requires all devices to be online at the time of cluster restart. Instead, one could imagine swapping loading of mounts to be lazily done: this might allow the cluster to start if certain tenants do not have online KMS or HSMs at the time of cluster start, but puts greater availability requirements on these systems in general.
+
+However, this work is strictly disjoint from implementing per-namespaces sealing as a feature and so will be done later.
+
+
+### Downsides
+
+This will add another complex feature to OpenBao; however, this substantially increases competitiveness in the broader Vault Enterprise ecosystem.
```
cipherboy:
```suggestion
This will add another complex feature to OpenBao; however, this substantially increases competitiveness in the broader Vault Enterprise ecosystem and increase the security posture for tenant separation.
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,421 @@
...
+
+## Security Implications
+
+Per-Namespace encryption would increase tenant security, as there is no single key encrypting all of OpenBaos data.
```
cipherboy:
```suggestion
Per-Namespace encryption would increase tenant security, as there is no single key encrypting all of OpenBao's data.
```

cipherboy (commented):
Some random thoughts @voigt! 

Now that we have a proper "Accepted" section on the website, do we want to clean this up and merge it?

website/content/docs/rfcs/namespace-sealing.mdx
wslabosz-reply: Open question from [POC](https://github.com/openbao/openbao/pull/1474)
> Or maybe namespaces should only ever be created with a single share and more added later?

cipherboy: Just to clarify this, I originally meant single _seal_, though I realized I messed up the wording there.

Like, even in a post-parallel-unseal world, I think you'll almost always start with one seal and then migrate to multiple rather than attempt them all in parallel on initialization? :thinking: 

And I think the same comment holds true here. 

We couly simplify the sealed namespace creation response by just including flat keys, which would only apply to a single seal.

website/content/docs/rfcs/namespace-sealing.mdx
phyrog: > A root namespace recovery mode token will certainly work against a child namespace, so maybe that is sufficient outside of non-hierarchical namespaces.

Just stumbled about this. Does this mean, a root namespace recovery token would be able to decrypt sealed namespace data? I assumed you would need to generate a recovery token for each sealed (parent) namespace.

cipherboy: @phyrog Sorry, I'm trying to draw a somewhat artificial distinction between:

- In a [parallel unseal world](https://github.com/openbao/openbao/issues/1021) where recovery keys _are_ unseal keys; the "unsealing key material"
- In a recovery situation, the API authentication token used.

My understanding is when booting into recovery mode, you still go through unseal as normal (from the PoV of the internals -- you don't use the recovery key and only use the KMS itself for auto-unseal -- for Shamir's you obviously still only have one set of key shares). The `generate-recovery-token` does one of two things:

1. With Shamir's unseal, advances the unseal process by one step.
2. When a sufficient number of recovery key shares have been provided (under auto-unseal or regular shares with Shamir's), an API authentication token is returned. And when using auto-unseal, this triggers the actual unseal operation which is delayed until then.

In particular, you're correct that the root namespace's recovery keys would (be expected) not function to perform the first unseal step of the namespace, but _if_ a namespace is unsealed in recovery mode (say, they too wished to recover), I think the root namespace's generated recovery API token would suffice for authentication to a namespace's storage segment during this time.

Thoughts?

website/content/docs/rfcs/namespace-sealing.mdx
wslabosz-reply: I'd opt for `Sealing an already sealed namespace has no effect.`
We should only reset the unsealing process using the dedicated endpoint.

website/content/docs/rfcs/namespace-sealing.mdx
wslabosz-reply: probably?

website/content/docs/rfcs/namespace-sealing.mdx
cipherboy: Yeah, I concur.

wslabosz-reply:
@cipherboy 
It's ready for re-review.

website/docusaurus.config.ts
```diff
@@ -39,7 +39,7 @@ const config: Config = {
     mermaid: true,
   },
   themes: ['@docusaurus/theme-mermaid'],
-
+
```
cipherboy:
```suggestion

```

website/content/docs/rfcs/index.mdx
```diff
@@ -76,3 +78,5 @@ Steering Committee.
  - [Best Practices for Integrating CEL in Auth, Secrets, and Elsewhere](/docs/rfcs/cel-best-practices),
    to discuss how Google's Common Expression Language should be integrated
    into OpenBao.
+ - [Using CEL Roles For JWT/OIDC Auth](/docs/rfcs/cel-jwt), adds support for CEL roles in JWT Auth
+   Landed in [769](https://github.com/openbao/openbao/pull/769)
```
cipherboy: Mistake in merge conflict fixing?

wslabosz-reply: yup, fixed

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,532 @@
...
+connection to an existing HSM. Second, it can only support the PKCS#11 library
+that is already present on the machine the OpenBao instance is running on. 
+Therefore, the lib field should not be configurable.
+
+Note: [Parallel Unsealing](https://github.com/openbao/openbao/issues/1021) on
+namespace level is considered a future improvement
+
+#### Sealing of a Namespace
+
+Sealing a namespace also seals all of it's children recursively.
```
satoqz:
```suggestion
Sealing a namespace also seals all of its children recursively.
```

I would move this info to the `### Sealed Namespace` section up above, it's very out of place here.

wslabosz-reply: adjusted

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,532 @@
...
+* The server process is restarted.
+* OpenBao's storage layer encounters an unrecoverable error.
+
+### Auto-Unsealing
+
+Auto-Unsealing (non-interactive unsealing) is an option that can be used to
+automatically unseal a namespace with the namespace root key and appropriate
+configuration. 
+
+Auto-Unsealing is considered an important feature, as it is convienient in case
```
satoqz:
```suggestion
Auto-Unsealing is considered an important feature, as it is convenient in case
```

website/content/docs/rfcs/namespace-sealing.mdx
```diff
@@ -0,0 +1,532 @@
...
+of multiple nested (sealed) namespaces, with an example structure of 
+`sealed-n1/sealed-ns2/ns3/sealed-ns4`. The default sealing mechanism is Shamir, 
+requiring a tenant to manually enter unseal key shares, which is incompatibile
```
satoqz:
```suggestion
requiring a tenant to manually enter unseal key shares, which is incompatible
```

satoqz (commented):
I should run https://github.com/crate-ci/typos... sooner...

PR #1360, opened 2025-05-27
Use per-namespace storage layouts for identity

cipherboy:
Two relevant commits here:

```
    Move to per-namespace identity storage
    
    This sets up all of identity storage to use packers available under the
    entity's namespace. This means less pressure on each packer, conforms
    with the layout of OpenBao's namespace design, and allows for
    per-namespace seals to also protect identity information.
    
    In the future, we can move to per-namespace MemDB objects as well.
    
```

and:

```
    Move identity memdb to per-namespace
    
    While a large change, this allows us to quickly drop a single namespace
    from the identity subsystem. We no longer need to purge all entries
    matching a given namespace from the MemDB and can instead simply remove
    all references to the MemDB, letting the garbage collection handle the
    rest.
    
    In particular, memdb access typically did not require a context; this
    meant that very few methods took a context (containing a namespace),
    like we have now provided.
    
    This cleanup was already handled by a new call to
    IdentityStore.RemoveNamespaceview(...).
    
```

The refactor in the second named commit was nasty, but worthwhile. To my view, we never iterated over the MemDB and removed since-deleted entities that were only part of a given (deleted) namespace. This would result in a memory leak until we restarted the instance.


wslabosz-reply (approved):
great job, love the refactor

PR #1363, opened 2025-05-28
Add NeoNephos to supporters section of homepage

klaus-sap:
https://neonephos.org/about/


cipherboy (approved):
Thank you @klaus-sap! Appreciate the support of the NeoNephos foundation.

Issue #1364, opened 2025-05-28
RFC - Web UI Modernization


driif:
### Summary

**OpenBao’s** current Web UI — based on **EmberJS** — is outdated and limits community contributions and modern feature development. This RFC proposes a complete rewrite using **React with TypeScript** to build a modern, secure, and extensible interface. The new UI will be developed as a separate application, introducing a secure **plugin architecture**, robust **observability dashboards**, and an **integration hub**. This initiative aims to revitalize community involvement, improve the user and operator experience, and establish a strong foundation for future innovation.

### Problem Statement

- **Obsolete Frontend:** EmberJS lacks community momentum; is hard to attract contributors and maintain.

- **No Plugin Support:** The UI can’t be extended dynamically for operator or user needs.

- **Clunky Permissions UX:** Users see, but can’t use, features; operators can’t easily hide UI elements based on roles/policies.

- **Confusing URL Pathing:** Users often copy/paste UI URLs directly into the CLI, which fails due to mismatched endpoint paths, causing frustration and inconsistent outcomes.

- **Multiple, Disjointed UIs:** Differing interfaces (API, web UI, CLI, web console) create confusion and inconsistent experience.

- **Hard to Maintain:** Coupled release cycles with OpenBao backend slow down development (including security patching).

- **Limited Operator Tools:** No native dashboards for observability; lacking integration templates for external systems.

- **Branding/UX Stagnation:** UI theme deviates from OpenBao identity; lacks clarity and customization.

- **No Clear Developer Entry Point:** Onboarding for contributors is painful; existing architecture deters engagement.
  
- **Insufficient User Guidance:** The UI lacks contextual help, links to documentation, and visual support — raising user frustration and support load.

- **No Built-in Global Messaging:** Operators cannot announce system incidents, upgrades, or downtime through banners or broadcast notifications in the UI.

- **Lack of i18n Support**.

- **Missing Branding Options:** Limited support for theming/customization.

### User-facing description

**For End Users**
- Faster, intuitive UI, focused on core secret/key management workflows.
- Guided navigation, contextual help, and clear error reporting.
- Only see actions/features they’re authorized for, improving focus/security.

**For Operators**
- Dashboards: built-in metrics, health, usage, audit visibility.
- Configurable message banners for announcements.
- Fine-grained branding and theme controls.

**For Developers**
- React/TypeScript codebase with modern tooling — easy to contribute to.
- Secure plugin SDK/interface for rapid UI or workflow extension.
[...]

PR #1365, opened 2025-05-28
Add RFC for Web UI Modernization

driif:
See also: #1364

### DIFF
--- a/website/content/docs/rfcs/web-ui-modernization.mdx
+++ b/website/content/docs/rfcs/web-ui-modernization.mdx
@@ -123,7 +123,7 @@ We invite the community to provide feedback on these key decisions:
 
 * What are the most critical features or secrets engines to prioritize for the initial release?
 [**Current community input:**](https://github.com/openbao/openbao/pull/1365#discussion_r2141249117)
-  - **Auth:** OIDC/JWT, Userpass, AppRole, Kubernetes
+  - **Auth:** OIDC/JWT, Userpass, AppRole
   - **Secrets:** KVv1, KVv2, Database, Transit
   *(Other engines such as PKI and SSH are acknowledged as more complex, may warrant separate design docs.)*
 * What specific metrics would you want to see in the built-in observability dashboards?


satoqz:
@driif What JavaScript package manager are you considering? A big pain with the current UI is not just Ember, but the version of Yarn it uses. I'd love to see something modern like PNPM or Bun, at the same time let's try not to make the tooling too bleeding-edge such that packagers (e.g. Linux distributions) don't have a hard time.

driif:
@wslabosz-reply I’d like to see Material Design or, as an alternative, shadcn/ui elements used in the UI—reusing well-supported components instead of creating custom ones. Adfinis is currently looking for someone to help create storyboards, which we’ll share with the community.
We believe - minimalist design focused on good UX is more valuable than a flashy UI.

website/content/docs/rfcs/web-ui-modernization.mdx
```diff
@@ -0,0 +1,137 @@
...
+- **React (TypeScript)** for massive dev pool, ecosystem, and maintainability.
+- **Material Design** for modern visual language and accessibility.
+- **Build:** Vite or RSPack for rapid iterative development and optimized bundles.
+- **Linter/Formatter:** BiomeJS (or ESLint).
+- **API:** Consumes OpenBao’s OpenAPI definition; codegen for type safety.
+- **Security:** Enforce CSP, SRI, CSRF, always HTTPS. Plugins sandboxed (Web Workers/iframed), NO backend credential exposure.
+- **Repo:** Separate repo, included as submodule or published artifact; decouples releases from backend.
+- **Caching** likely none, except for static content like images.
+
+**Plugin System**
```
phyrog, line 75: Are these purely frontend plugins? Have you given thought on how these would be persisted? Would we have a separate backend server that manages plugins, or is this something we need to integrate into the OpenBao API?

driif: @phyrog Hi Tom, thanks for your questions. The concept for how plugins should work is still a work in progress.

Right now, I see backend plugins and frontend plugins as two separate things. For example, if you have an Azure plugin, integrating it with Bao wouldn’t affect the UI at all at this point. I imagine it might be possible to install an extension—similar to how VS Code does it—that would let you extend the UI with the necessary interfaces to interact with that plugin.

That said, I’m definitely open to other ideas. For example, we could explore plugins that provide information to the frontend about which components to render. That could be interesting too.

Overall, I think this is a topic where community input will be really valuable. It will help us figure out the most practical and user-friendly approach for plugins.

Let me know what you think, or if you have any suggestions!

cipherboy: Conceptually @phyrog this kinda already exists. The main UI _could_ check for a `<plugin>/.well-known/OpenBao/ui-iframe` path and load it as a sign of plugins supporting UI. We don't currently support this because tokens don't pass through, but you could build up something more complex in a similar manner (load a react element from a JS endpoint and eval it), or similar.

So other than defining some standard and providing usable tools to plugin authors, I don't think it requires _much (if anything)_ in the way of internal OpenBao Go server changes.

cipherboy: This would let us move the UI and plugins to a consistently versioned interface. The main UI becomes shipped components and every plugin (builtin or not) exposes its own UI in the plugin, so you can do consistent updates for them.

website/content/docs/rfcs/index.mdx
```diff
@@ -44,3 +44,6 @@ Steering Committee.
    into OpenBao.
  - [Using CEL Roles For JWT/OIDC Auth](/docs/rfcs/cel-jwt), adds support for CEL roles in JWT Auth
    Landed in [769](https://github.com/openbao/openbao/pull/769)
+ - [Web UI Modernization](/docs/rfcs/web-ui-modernization), for rewriting OpenBao’s web UI from EmberJS to React
```
cipherboy: In #1389, which hopefully might be merged soon-ish, I proposed an `Accepted`, `Landed`, and `Strategy` category for RFCs. I think this would probably fall in the latter. 

But since we've begun opening RFCs as PRs, I think it'd be great to land them as Accepted to communicate the state more clearly, even ahead of the code landing.

website/content/docs/rfcs/web-ui-modernization.mdx
```diff
@@ -0,0 +1,137 @@
...
+## User/Developer Experience
+
+- Easy onboarding: clear contributing guide, modern stack, fast dev loop.
+- Community plugin gallery starts simple; grows with use.
+- Built-in operator tools for observability; extensions for those needing deep integration.
+
+## Unresolved Questions
+We invite the community to provide feedback on these key decisions:
+
+* What are the most critical features or secrets engines to prioritize for the initial release?
```
cipherboy, line 124: From my side, I think the candidates are:

**Auth**

- OIDC/JWT
- Userpass
- AppRole
- Kubernetes?

**Secrets**

- KVv1 + KVv2
- Database
- Transit

Many of the other types are more complex, especially PKI and SSH, and might warrant design docs of their own.

website/content/docs/rfcs/web-ui-modernization.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+## Problem Statement
+
+- **Obsolete Frontend:** EmberJS lacks community momentum; is hard to attract contributors and maintain.
+
+- **No Plugin Support:** The UI can’t be extended dynamically for operator or user needs.
+
+- **Clunky Permissions UX:** Users see, but can’t use, features; operators can’t easily hide UI elements based on roles/policies.
+
+- **Confusing URL Pathing:** Users often copy/paste UI URLs directly into the CLI, which fails due to mismatched endpoint paths, causing frustration and inconsistent outcomes.
```
eyenx: Not only this paths are being used by CLI, but in my experience it's also quiet difficult to built policies around UI & API usage alike. Some UI stuff needs other paths access than what path we use through the CLI. I would very much welcome a streamlining of the URL Paths.

suprjinx:
I'd like to make a plug for [HTMX](https://htmx.org/docs/) as an alternative to React. It has some very nice advantages:
1. does not need pnpm/node to compile -- no compile step during development at all
2. uses server-side rendering so does not need protobuf api for every possible UI action
3. much simpler markup approach (instead of markup/code component mixtures of React)

I agree that React has the market cornered from UI developer point of view, but I do think HTMX would be simpler for the typical backend/Go developer (ahem, like me :-))

cipherboy:
@suprjinx one concern I have on:

> uses server-side rendering so does not need protobuf api for every possible UI action

The current UI uses the standard OpenBao API which is JSON, not protobuf.

That said, I'm a little against server-side rendering. I agree, heavy PWAs aren't nice but in this case it serves a useful security property: we don't need server-side session management and we prevent certain classes of bugs (authorization bypass, session confusion, ...). The UI _should_ just be another client (like any other customer application), though a very complete one. It just happens to have its HTML/JS/CSS payloads served from the same server as the API.

driif:
Hi @suprjinx. I’m also a big fan of htmx and really like its approach. But as someone who’s used htmx in my personal projects, I’ve run into a few challenges:

1. It can be tricky to organize a clean frontend and backend structure.
2. You need to rewrite your endpoints to return HTML templates instead of JSON, which can be a hassle (it requires to rewrite not only frontend).
3. It's not always easy to use with custom request methods like LIST.

I agree that htmx is a great option, especially for backend developers, but I’m still not completely convinced it’s the best fit in this particular case. 
https://htmx.org/essays/htmx-sucks/

suprjinx:
> I agree that htmx is a great option, especially for backend developers, but I’m still not completely convinced it’s the best fit in this particular case. https://htmx.org/essays/htmx-sucks/

You do make good points @driif @cipherboy. I'm not sure I agree with much in that essay, except for "Backend Engineers Make Garbage UIs" :-).

driif:
@suprjinx Oh, I wouldn’t share this essay either if it wasn’t written by the creator of htmx:) I think his main point is that if your project needs more complex business logic, htmx alone probably won’t be enough. Eventually, you’ll end up building extra layers or your own framework on top of htmx, maybe adding TypeScript for type safety and more structure.

website/content/docs/rfcs/web-ui-modernization.mdx
```diff
@@ -0,0 +1,137 @@
+---
+sidebar_label: Web UI modernization
+description: |-
+  Proposal to rewrite OpenBao's Web UI from EmberJS to React with TypeScript,
+introducing a modern, extensible, secure interface with plugin architecture,
```
cipherboy: I think these need to be indented @driif; it is failing the build right now:

```
  [cause]: Error: Can't process doc metadata for doc at path path=/home/runner/work/openbao/openbao/website/content/docs/rfcs/web-ui-modernization.mdx in version name=current 
```

website/content/docs/rfcs/web-ui-modernization.mdx
driif: Added to Unresolved Questions.

eyenx: I would leave k8s out. We usually configure JWT when integrating into k8s.

driif:
@cipherboy this PR might be ready for merge.

cipherboy:
@driif So merged. Thank you!

PR #1366, opened 2025-05-28
Namespace sealing POC

phyrog:
POC for #1357 / #1170 

Currently implemented features:
- per namespace barriers, root keys, keyrings (data is encrypted only once)
- create sealable namespace (incl. support in cli command) using a seal config
    - currently only supports shamir
    - namespace is unsealed after creation for now, we could reseal after namespace initialization
- store seal config permanently, encrypted with closest parent barrier
- `/sys/raw` works for keys encrypted with non-root barriers

Missing in this POC (not exhaustive):
- loading/unloading of sealed namespaces from storage
- sealing/unsealing lifecycle functionality, api, cli command
- remembering seal shares when calling unseal api
- most other api actions (rekeying, recovery, generate-root, key rotation, seal status, ...)
  - most should be straight forward to implement since we are reusing the existing AESGCMBarrier implementation
- handling requests towards sealed namespaces

Example for creating a sealable namespace:

```sh
bao namespace create -seals=seals.json ns1
```

with seals.json (hcl not supported yet):

```json
[
  {
    "type":"shamir",
    "secret_shares":3,
    "secret_threshold":2
  }
]
```

/cc @cipherboy, @driif, @voigt, @wslabosz-reply, @rencooo, @satoqz

### DIFF
--- a/vault/barrier_aes_gcm.go
+++ b/vault/barrier_aes_gcm.go
@@ -69,6 +69,8 @@ type AESGCMBarrier struct {
 	l      sync.RWMutex
 	sealed bool
 
+	metaPrefix string
+
 	// keyring is used to maintain all of the encryption keys, including
 	// the active key used for encryption, but also prior keys to allow
 	// decryption of keys encrypted under previous terms.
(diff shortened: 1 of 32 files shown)


vault/barrier_aes_gcm.go
```diff
@@ -69,6 +69,8 @@ type AESGCMBarrier struct {
 	l      sync.RWMutex
 	sealed bool
 
+	metaPrefix string
```
cipherboy, line 72: I think `physical.Backend` has a view, just like `logical.Storage` does fwiw.

phyrog: FYI `metaPrefix` is not used to restrict the view on the backend here, but to set the path where the barrier metadata is saved. The barrier itself could access all of the storage (but won't be able to decrypt anything it did not write itself).

wslabosz-reply:
my bad, deleted the source branch of this PR by mistake.

satoqz:
> my bad, deleted the branch this orginated from by mistake.

My fault for telling y'all to clean up 😁

wslabosz-reply:
merged this one without any approval (by mistake), opened up same pull request against new branch `per-namespace-sealing` https://github.com/openbao/openbao/pull/1471

PR #1368, opened 2025-05-28
Fix flaky PostgreSQL backend connection test

satoqz:
Fixes flaky test introduced in https://github.com/openbao/openbao/pull/1299.
In the wild: https://github.com/openbao/openbao/actions/runs/15303683740/job/43050865821

Committing the transaction may already decrement std's internal connection count, we need to decrement our own beforehand.


cipherboy (approved):
D'oh. Thanks @satoqz!

wslabosz-reply (approved):

PR #1369, opened 2025-05-28
Add a guard rail if CA cert is expired while using the truncate behavior

### LINKED ISSUE #1322: PKI - Truncate should error on expired certificates
As reported by @andy-d2019 on https://github.com/hashicorp/vault/issues/30621:

> **Describe the bug** Vault issues expired certificates when the **leaf_not_after_behavior** is set to **truncate** and the **default issuer** in a **pki** has expired.
> 
> **To Reproduce** Steps to reproduce the behavior:
> 
>     1. Setup a vault pki with a given default issuer
> 
>     2. Let the default Issuer expire
> 
>     3. Request a certificate from the default issuer
> 
> 
> This bug can be easily replicated with the following setup:
> 
> `id=$(kubectl exec vault-0 -i -- vault list -format json pki/issuers | jq -r '.[0]') kubectl exec vault-0 -i -- vault write pki/issuer/$id leaf_not_after_behavior=truncate`
> 
> With the err behavior, Vault errors because the cert's NotAfter goes beyond the issuer's. This prevents Vault from issuing when the CA has expired.
> 
> With truncate, Vault copies the NotAfter e.g.:
> 
> `min(cert's NotAfter, CA's NotAfter)`
> 
> Which leads to Vault issuing expired certificates.
> 
> **Expected behavior** Vault should not issue expired certificates with the truncate leaf_not_after_behavior.
> 
> **Environment:**
> 
>     * Vault Server Version: 1.17.2
> 
>     * Vault CLI Version: 1.17.2
> 
>     * Server Operating System/Architecture: Kubernetes
> 
> 
> **Additional context** This issue lead to a problem with cert-manager for us: [cert-manager/cert-manager#7741](https://github.com/cert-manager/cert-manager/issues/7741) But the original root cause seems to be in vault.

---

This makes sense to me fwiw. The change should be fairly minimal, to add a guard clause here:

https://github.com/openbao/openbao/blob/cc0df5ca5853e829bcb22bfbb68212472bc0d7a1/builtin/logical/pki/cert_util.go#L1741-L1742

Preventing the use of a truncated timestamp if it has already expired. 

---

One concern would be on systems with clockskew. If a user provides `not_after` to a date in the past (but still before issuer's date), we'll happily issue an expired certificate. This may be a "feature" if OpenBao's time is in the present but a disconnected system is running with offset clocks that the certs would be valid for.

Regardless, that is likely an edge case to be addressed later and more globally.

sudhanvaghebbale:
Add a guard rail if CA cert is expired while using the truncate behavior

Followed this doc to replicate created an expired CA/int CA - https://developer.hashicorp.com/vault/tutorials/pki/pki-engine

Resolves #1322

### DIFF
--- a/builtin/logical/pki/cert_util.go
+++ b/builtin/logical/pki/cert_util.go
@@ -1739,6 +1739,11 @@ func getCertificateNotAfter(b *backend, data *inputBundle, caSign *certutil.CAIn
 		case certutil.PermitNotAfterBehavior:
 			// Explicitly do nothing.
 		case certutil.TruncateNotAfterBehavior:
+			// Error out if notAfter is in the past
+			if notAfter.Before(time.Now()) {
+				return time.Time{}, warnings, errutil.UserError{Err: fmt.Sprintf(
+					"cannot satisfy request, as NotAfter date %s is in the past", notAfter)}
+			}
 			notAfter = caSign.Certificate.NotAfter
 		case certutil.ErrNotAfterBehavior:
 			fallthrough
--- a/builtin/logical/pki/cert_util_test.go
+++ b/builtin/logical/pki/cert_util_test.go
@@ -5,13 +5,17 @@ package pki
 
 import (
 	"context"
+	"crypto/x509"
 	"fmt"
 	"reflect"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/openbao/openbao/sdk/v2/framework"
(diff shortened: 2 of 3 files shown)


sudhanvaghebbale:
> One nit @sudhanvaghebbale, then I'll merge! Thank you!

@cipherboy Done, let me know if that looks good?

cipherboy (approved):
Thank you @sudhanvaghebbale! Appreciate it :-)

PR #1372, opened 2025-05-28
Add CAS support to kv metadata updates

### LINKED ISSUE #1253: feature request: Add CAS support to KV secrets engine create/update metadata endpoint

The KV secrets engine currently lacks Compare-And-Swap (CAS) support for metadata updates, while it exists for secret value updates. This creates a race condition vulnerability when multiple clients attempt to update a secret's metadata concurrently. Without metadata version checking, clients with outdated metadata can unintentionally overwrite more recent metadata changes made by other clients, potentially leading to data loss or security issues.

Add a `cas` parameter to the OpenBao KVv2 create/update metadata endpoint (https://openbao.org/api-docs/secret/kv/kv-v2/#createupdate-metadata) that works independently from the existing secret value versioning system. This would:

1. Create a synthetic version counter for metadata that's separate from secret value versions
2. Include the metadata version in responses when reading a secret
3. Allow clients to specify the expected metadata version when updating metadata
4. Return appropriate errors when metadata version mismatch occurs
5. Increment the metadata version with each successful metadata update

The behavior should mirror the existing CAS functionality for secret values but apply specifically to metadata operations.

1. Implementing application-level locking in client applications - This adds complexity to client applications and doesn't solve the fundamental issue.
2. Using the existing secret version for metadata - This would conflate two different types of changes and make independent updates impossible.
3. Having a single atomic update for both secret value and metadata - While convenient, this reduces flexibility for applications that need to update these separately.

Multi-user collaboration on secrets management - Multiple users working with the same secrets need assurance that their metadata changes won't overwrite others'

**Additional context**

This feature would complete the KV secrets engine's Compare-And-Swap implementation by extending it to metadata operations, creating a comprehensive solution for handling concurrent updates. 

The implementation would ideally keep metadata versioning separate from the secret version history to maintain the clean separation between value data (which may have multiple historical versions) and metadata (which typically only needs the current version).

fcbajao:
# Add CAS Support for KV v2 Metadata Updates

Resolves #1253

## Description

This PR adds Compare-And-Swap (CAS) support for metadata updates in the KV v2 secrets engine. Previously, CAS was only available for secret data updates, which could lead to race conditions when multiple clients attempted to update metadata concurrently.

### What does this PR do?

This implementation adds:
- A new `metadata_cas` parameter to metadata update endpoints
- A `metadata_cas_required` configuration option (both global and per-key)
- A `current_metadata_version` field that tracks metadata changes independently from data versions
- Validation that prevents metadata updates without the correct CAS value when required

### Background

The KV v2 secrets engine provides versioning for secret data with CAS protection, but metadata updates (such as `max_versions`, `cas_required`, custom metadata) lacked this protection. This created potential race conditions where concurrent metadata updates could overwrite each other.

This feature was requested by GitLab (see GitLab issue [gitlab-org/gitlab#538174](https://gitlab.com/gitlab-org/gitlab/-/issues/538174)) to ensure data integrity in their secrets management system.

## Design of Change

The implementation follows the existing CAS pattern used for secret data:

1. **Metadata Versioning**: Added `current_metadata_version` field to `KeyMetadata` that increments on every metadata update
2. **CAS Validation**: Before any metadata update, check if CAS is required and validate the provided version matches
3. **Configuration**: Added `metadata_cas_required` to both global config and per-key metadata
4. **Backward Compatibility**: Metadata CAS is optional by default, maintaining compatibility with existing workflows

Key files changed:
- `builtin/logical/kv/types.proto` - Added new fields to protobuf definitions
- `builtin/logical/kv/path_metadata.go` - Implemented CAS logic for metadata operations
- `builtin/logical/kv/path_config.go` - Added global configuration support

## Related Issues & PRs

- Fixes #1253
- Related to GitLab issue [[#538174](https://gitlab.com/gitlab-org/gitlab/-/issues/538174)](https://gitlab.com/gitlab-org/gitlab/-/issues/538174)

### Unit Tests
All existing tests pass, plus new tests added for:
- Metadata CAS validation on create/update operations
- Metadata version increment behavior  
- PATCH operations with metadata CAS
- Configuration warnings when backend mandates CAS
- Proper handling of `metadata_cas_required` settings

```bash
make test TEST=./builtin/logical/kv TESTARGS="-run TestVersionedKV_Metadata"
```

### Example Usage

```bash
# Enable metadata CAS globally
bao write secret/config metadata_cas_required=true

# Create metadata with CAS
bao write secret/metadata/my-secret max_versions=5 metadata_cas=0

# Update metadata with correct CAS value  
bao write secret/metadata/my-secret max_versions=10 metadata_cas=1

# Attempting update without CAS fails
bao write secret/metadata/my-secret max_versions=15
# Error: metadata check-and-set parameter required for this call

# PATCH operations also respect CAS
bao patch secret/metadata/my-secret max_versions=20 metadata_cas=2
```

## Backwards Compatibility

This change is fully backwards compatible:
- Metadata CAS is disabled by default
- Existing metadata update operations continue to work unchanged
- The `current_metadata_version` field is initialized to 0 for existing metadata
- Only when explicitly enabled does CAS validation occur

## Notes for Reviewers

- The implementation mirrors the existing CAS functionality for secret data to maintain consistency
- Metadata versioning is independent from data versioning as they serve different purposes
- The protobuf field numbers (400, 401) were chosen to avoid conflicts with potential future fields

Please let me know if you need any additional information or changes to this implementation.


fcbajao:
@cipherboy here's my initial attempt on implementing this. Please take a look and let me know how we can further improve it :)

builtin/logical/kv/path_metadata.go
```diff
@@ -461,9 +475,20 @@ func (b *versionedKVBackend) pathMetadataWrite() framework.OperationFunc {
 		}
 
 		var resp *logical.Response
+
+		addWarning := func(msg string) {
```
cipherboy, line 479: Neat, I like the pattern :-)

builtin/logical/kv/path_metadata.go
```diff
@@ -486,14 +511,33 @@ func (b *versionedKVBackend) pathMetadataWrite() framework.OperationFunc {
...
 			meta = &KeyMetadata{
-				Key:         key,
-				Versions:    map[uint64]*VersionMetadata{},
-				CreatedTime: now,
-				UpdatedTime: now,
+				Key:                    key,
+				Versions:               map[uint64]*VersionMetadata{},
+				CreatedTime:            now,
+				UpdatedTime:            now,
+				CurrentMetadataVersion: 0, // Initialize to 0 for new metadata
```
cipherboy:
```suggestion
				CurrentMetadataVersion: 1, // Initialize to 1 for new metadata
```

Do we want to go to 1? It allows us to differentiate new versus upgraded entries in case there is some bug. From 1+, everything should behave the same.

builtin/logical/kv/path_metadata_test.go
```diff
@@ -58,10 +59,17 @@ func TestVersionedKV_Metadata_Put(t *testing.T) {
...
 	}
+	if resp.Data["metadata_cas_required"] != true {
+		t.Fatalf("Bad response: %#v", resp)
+	}
 	if resp.Data["delete_version_after"] != d.String() {
 		t.Fatalf("Bad response: %#v", resp)
 	}
 
+	if resp.Data["current_metadata_version"] != uint64(0) {
+		t.Fatalf("Expected initial current_metadata_version to be 0, got: %v", resp.Data["current_metadata_version"])
```
cipherboy:
```suggestion
	if resp.Data["current_metadata_version"] != uint64(1) {
		t.Fatalf("Expected initial current_metadata_version to be 1, got: %v", resp.Data["current_metadata_version"])
```

This will have a lot of other implications elsewhere fwiw.

builtin/logical/kv/path_metadata_test.go
```diff
@@ -180,6 +188,10 @@ func TestVersionedKV_Metadata_Put(t *testing.T) {
 		t.Fatalf("Bad response: %#v", resp)
 	}
 
+	if resp.Data["current_metadata_version"] != uint64(0) {
+		t.Fatalf("Expected current_metadata_version to still be 0, got: %v", resp.Data["current_metadata_version"])
```
cipherboy:
```suggestion
	if resp.Data["current_metadata_version"] != uint64(1) {
		t.Fatalf("Expected current_metadata_version to still be 1, got: %v", resp.Data["current_metadata_version"])
```

builtin/logical/kv/path_metadata_test.go
```diff
@@ -188,11 +200,35 @@ func TestVersionedKV_Metadata_Put(t *testing.T) {
...
+		t.Fatalf("expected error, %#v", resp)
+	}
+
+	// Next, attempt to update the metadata while also passing the correct metadata_cas value.
 	// Update the metadata settings, remove the cas requirement and lower the
 	// max versions.
 	data = map[string]interface{}{
 		"max_versions": 1,
 		"cas_required": false,
+		"metadata_cas": 0,
```
cipherboy:
```suggestion
		"metadata_cas": 1,
```

builtin/logical/kv/path_metadata_test.go
```diff
@@ -249,6 +285,11 @@ func TestVersionedKV_Metadata_Put(t *testing.T) {
 		t.Fatalf("Bad response: %#v", resp)
 	}
 
+	// Expect that the metadata version was incremented
+	if resp.Data["current_metadata_version"] != uint64(1) {
```
cipherboy:
```suggestion
	if resp.Data["current_metadata_version"] != uint64(2) {
```

builtin/logical/kv/path_metadata_test.go
```diff
@@ -279,6 +320,84 @@ func TestVersionedKV_Metadata_Put(t *testing.T) {
 	if diff := deep.Equal(actual, expected); len(diff) > 0 {
 		t.Fatalf("expected detailed-metadata/ listing to have same contents as read on foo/\ndiff: %#v", diff)
 	}
+
+	// Next, attempt to update the metadata while passing an incorrect metadata_cas value which we expect to fail
+	data = map[string]interface{}{
+		"metadata_cas_required": false,
+		// This will not match the current metadata version which is 1
```
cipherboy:
```suggestion
		// This will not match the current metadata version which is 2
```

builtin/logical/kv/path_metadata_test.go
```diff
@@ -279,6 +320,84 @@ func TestVersionedKV_Metadata_Put(t *testing.T) {
...
+	}
+	if resp.Error().Error() != "metadata check-and-set parameter does not match the current version" {
+		t.Fatalf("expected error, %#v", resp)
+	}
+
+	// Correctly update the metadata and disable metadata cas check
+	data = map[string]interface{}{
+		"metadata_cas_required": false,
+		// This will match the current metadata version
+		"metadata_cas": 1,
```
cipherboy:
```suggestion
		"metadata_cas": 2,
```

builtin/logical/kv/path_metadata_test.go
```diff
@@ -1043,6 +1162,63 @@ func TestVersionedKV_Metadata_Patch_CasRequiredWarning(t *testing.T) {
...
+	}
+
+	// Create metadata
+	req = &logical.Request{
+		Operation: logical.CreateOperation,
+		Path:      "metadata/foo",
+		Storage:   storage,
+		Data: map[string]interface{}{
+			"max_versions": 5,
+			"metadata_cas": 0, // Required because global setting
```
cipherboy:
```suggestion
			"metadata_cas": 1, // Required because global setting
```

builtin/logical/kv/path_metadata_test.go
```diff
@@ -1043,6 +1162,63 @@ func TestVersionedKV_Metadata_Patch_CasRequiredWarning(t *testing.T) {
...
+	}
+
+	// Try to disable metadata_cas_required via PATCH
+	req = &logical.Request{
+		Operation: logical.PatchOperation,
+		Path:      "metadata/foo",
+		Storage:   storage,
+		Data: map[string]interface{}{
+			"metadata_cas_required": false,
+			"metadata_cas":          0,
```
cipherboy:
```suggestion
			"metadata_cas":          1,
```

Shouldn't this be 2 though? :thinking: We should've set 1 earlier, so I would've expected this to fail since we updated it above?

builtin/logical/kv/types.proto
```diff
@@ -12,6 +12,7 @@ message Configuration {
 	uint32 max_versions = 1;
 	bool cas_required = 2;
 	google.protobuf.Duration delete_version_after = 3;
+  bool metadata_cas_required = 400;
```
cipherboy: This needs to be a tab ideally, rather than two spaces. Same as below.

builtin/logical/kv/path_metadata.go
```diff
@@ -486,14 +511,33 @@ func (b *versionedKVBackend) pathMetadataWrite() framework.OperationFunc {
 		if err != nil {
 			return nil, err
 		}
+
+		now := ptypes.TimestampNow()
 		if meta == nil {
-			now := ptypes.TimestampNow()
 			meta = &KeyMetadata{
```
cipherboy: It seems like it'd be best to have CAS checked here as well and prevent the request if the caller sets `Metadata >= 1`  since that implies they think it already exists. Otherwise, you could have an inbound request with `metadata == 53` and a delete call which just beats it, causing the write to create a new entry.

builtin/logical/kv/path_metadata.go
cipherboy: In particular, based on the comment on the lines above (below in the GitHub viewer), you could set this to zero, move the check below about CAS value to global, and _then_ increment, so zero means entry does not exist and/or is upgraded from earlier, and then always increment (getting 1 on net-new entries). Caller would then provider `0` as the CAS value for creation, and get an entry with 1, which is expected.

cipherboy (commented):
@fcbajao One functionality change, that results in a few test changes, but otherwise looks good :-) Thank you!

builtin/logical/kv/path_metadata_test.go
cipherboy: This one is wrong. We create with value zero still. Ignore.

builtin/logical/kv/path_metadata_test.go
cipherboy: This would be 1, not 2. Ignore my text comment, but the suggestion is correct.

fcbajao:
> @fcbajao One [functionality change](https://github.com/openbao/openbao/pull/1372#discussion_r2136120377), that results in a few test changes, but otherwise looks good :-) Thank you!

@cipherboy Thank you for the valuable feedback. I've updated the PR to set the metadata CAS default to 1 and also add validation on metadata_cas upon creation.

Please take another look.

cipherboy:
@fcbajao Hmm, we're now failing DCO. Do you mind also adding [a changelog entry](https://github.com/openbao/openbao/blob/main/changelog/README.md)? Thanks!

Test failure also looks real:

```
2025-06-10T17:43:09.0699601Z ##[group][31mFAIL[0m command.TestKvMetadataPatchCommand_Flags/cas_required_success (4.08s)
2025-06-10T17:43:09.0700339Z     kv_metadata_patch_test.go:252: patched "current_metadata_version" mismatch, diff: []string{"1 != 2"}
```

though the metadata patch might just need an exclusion to avoid flakiness:

```
2025-06-10T17:43:25.1772337Z     kv_metadata_patch_test.go:252: patched "updated_time" mismatch, diff: []string{"2025-06-10T17:43:22.17370456Z != 2025-06-10T17:43:23.176245799Z"}
```

wslabosz-reply (approved):
LGTM

PR #1385, opened 2025-05-30
Deprecation notice for undocumented duplicate PKCS#11 seal options

satoqz:
Also see https://github.com/openbao/go-kms-wrapping/pull/33#discussion_r2112177962.

Does this need a changelog entry?


cipherboy:
@satoqz I think that'd be a good idea given it's a deprecation. Can you check if it's natively supported? Else will call it `change` to highlight it.

PR #1388, opened 2025-05-31
Fix detailed metadata on list results

cipherboy:
Unlike scan, list results include directories in their results. This results in KVv2 panicing when attempting to fetch metadata about the entry, as it doesn't exist and thus returns nil. We should handle this gracefully.


cipherboy:
Thank you @wslabosz-reply! Merging...

Issue #769, opened 2024-11-29
Restrict LIST and SCAN to only accessible paths


cipherboy:
### Summary

Add a new keyword, `list_scan_response_keys_filter_path`, to ACL policies to support filtering well-formed `list` and [`scan`](https://github.com/openbao/openbao/pull/763) responses to only subpaths accessible by the token.

### Problem Statement

Administrators grant broad permissions via the `list` endpoint. Notably, if a user only has `read` or `list` access to select subkeys, but is granted `list` or `scan` on a parent, all results are visible to the user even if they are not otherwise accessible. Complicating this is that certain paths are unauthenticated (such as PKI's `certs/` listing, which is itself authenticated but any given entry under `cert/:serial` is not), or similarly, that a list handler may not necessarily be a prefix of the single entry read handler (again, in the case of PKI's certificate or issuers lists).

This makes automatic, backend-agnostic filtering hard to achieve without some guidance from the plugin or policy author.

We opt to do the latter and push this mapping requirement onto the policy author.

This also gives us an indication at the system level that filtering should be applied, as it is a potentially expensive operation and different callers may have differing requirements (e.g., in KVv2, some users may wish to filter based on `metadata/{{ key }}` access whereas others may wish to filter on `data/{{ key }}` access).

If necessary, this parameter could be made to accept a list of paths. However, semantics would then also have to be defined (`OR` versus `AND`) to decide how many need to match to show the entry in the result.

### User-facing Description

Policy authors may make a breaking (from a consuming application's PoV) change by introducing a `list_scan_response_keys_filter_path` argument inside a given `path` grant in an ACL policy. This modifies the specific `list` response indicated by removing any paths whose corresponding item path (generated by evaluating `{{ key }}` in `list_scan_response_keys_filter_path` with each item in the response's `"keys"` data field) which the token does not have access to. This is relatively expensive, requiring a policy check for every response key, so should likely be combined with `required_parameters=limit` on list operations to prevent having to evaluate thousands or potentially millions (in the case of PKI certificates or recursive K/V listing over the entire mount) of policy checks.

Consuming applications will see a given key in the `keys` response field if one of two conditions holds:

 1. The key ends in `/` and the application has the `list` capability on the resulting path, or
 2. The key does not end in `/` and the application has the `read` on the resulting path.

### Technical Description

[...]

PR #1389, opened 2025-05-31
Add response_keys_filter_path to filter lists

cipherboy:
This adds a new policy parameter, list_scan_response_keys_filter_path, to create a path (containing `{{ .key }}` as a required substring for replacement) for evaluating each list response item against the token's ACL. Keys and associated data whose templated path are allowed read access (or if it ends with a `/`, list access) to the given path are preserved in the response; otherwise, they are filtered and removed.

Note that desired path for filtering is potentially application and plugin-dependent and thus is best left to the policy author to do rather than have plugin authors indicate best paths.

In the future, there are two possible avenues for expansion: using text/template as a templating engine, giving more control to policy authors over destination paths, and using a language like CEL for fully controlling response filtering regardless of request type and allowing it to dispatch additional ACL policy checks.

Note that this implementation is limited: only `read` and `list` operations are considered.

Resolves: #769

---

Blocked on #1388; will be rebased once that merges.

### DIFF
--- a/vault/request_handling.go
+++ b/vault/request_handling.go
@@ -381,7 +381,7 @@ func (c *Core) fetchACLTokenEntryAndEntity(ctx context.Context, req *logical.Req
 	return acl, te, entity, identityPolicies, nil
 }
 
-func (c *Core) CheckToken(ctx context.Context, req *logical.Request, unauth bool) (*logical.Auth, *logical.TokenEntry, error) {
+func (c *Core) CheckToken(ctx context.Context, req *logical.Request, unauth bool) (*logical.Auth, *ACL, *logical.TokenEntry, *identity.Entity, error) {
 	defer metrics.MeasureSince([]string{"core", "check_token"}, time.Now())
 
 	var acl *ACL
@@ -399,24 +399,24 @@ func (c *Core) CheckToken(ctx context.Context, req *logical.Request, unauth bool
 		// unauth, we just have no information to attach to the request, so
 		// ignore errors...this was best-effort anyways
 		if err != nil && !unauth {
-			return nil, te, err
+			return nil, acl, te, entity, err
 		}
 	}
 
 	if entity != nil && entity.Disabled {
 		c.logger.Warn("permission denied as the entity on the token is disabled")
-		return nil, te, logical.ErrPermissionDenied
+		return nil, acl, te, entity, logical.ErrPermissionDenied
 	}
 	if te != nil && te.EntityID != "" && entity == nil {
 		c.logger.Warn("permission denied as the entity on the token is invalid")
-		return nil, te, logical.ErrPermissionDenied
+		return nil, acl, te, entity, logical.ErrPermissionDenied
 	}
 
 	// Check if this is a root protected path
 	rootPath := c.router.RootPath(ctx, req.Path)
 
 	if rootPath && unauth {
-		return nil, nil, errors.New("cannot access root path in unauthenticated request")
+		return nil, nil, nil, nil, errors.New("cannot access root path in unauthenticated request")
 	}
 
 	// At this point we won't be forwarding a raw request; we should delete
@@ -449,18 +449,18 @@ func (c *Core) CheckToken(ctx context.Context, req *logical.Request, unauth bool
 			// fail later via bad path to avoid confusing items in the log
 			checkExists = false
 		case logical.ErrRelativePath:
-			return nil, te, errutil.UserError{Err: err.Error()}
+			return nil, acl, te, entity, errutil.UserError{Err: err.Error()}
 		case nil:
 			if existsResp != nil && existsResp.IsError() {
-				return nil, te, existsResp.Error()
+				return nil, acl, te, entity, existsResp.Error()
 			}
 			// Otherwise, continue on
 		default:
 			c.logger.Error("failed to run existence check", "error", err)
 			if _, ok := err.(errutil.UserError); ok {
-				return nil, te, err
+				return nil, acl, te, entity, err
 			} else {
-				return nil, te, ErrInternalError
+				return nil, acl, te, entity, ErrInternalError
 			}
 		}
 
@@ -516,13 +516,17 @@ func (c *Core) CheckToken(ctx context.Context, req *logical.Request, unauth bool
 		Allowed: authResults.Allowed,
 	}
 
+	if authResults.ACLResults != nil {
+		auth.ResponseKeysFilterPath = authResults.ACLResults.ResponseKeysFilterPath
+	}
+
 	if !authResults.Allowed {
 		retErr := authResults.Error
 
 		if authResults.Error.ErrorOrNil() == nil || authResults.DeniedError {
 			retErr = multierror.Append(retErr, logical.ErrPermissionDenied)
 		}
-		return auth, te, retErr
+		return auth, acl, te, entity, retErr
 	}
 
 	if authResults.ACLResults != nil && len(authResults.ACLResults.GrantingPolicies) > 0 {
@@ -532,7 +536,7 @@ func (c *Core) CheckToken(ctx context.Context, req *logical.Request, unauth bool
 		auth.PolicyResults.GrantingPolicies = append(auth.PolicyResults.GrantingPolicies, authResults.SentinelResults.GrantingPolicies...)
 	}
 
-	return auth, te, nil
+	return auth, acl, te, entity, nil
 }
 
 // HandleRequest is used to handle a new incoming request
@@ -923,7 +927,7 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
 	}
 
 	// Validate the token
-	auth, te, ctErr := c.CheckToken(ctx, req, false)
+	auth, acl, te, entity, ctErr := c.CheckToken(ctx, req, false)
 	if ctErr == logical.ErrRelativePath {
 		return logical.ErrorResponse(ctErr.Error()), nil, ctErr
 	}
@@ -1072,6 +1076,11 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
 				CreationPath: creationPath,
 				SealWrap:     sealWrap,
 			}
+		}
+
+		// Ensure compliance with List operation filtering.
+		if err := c.filterListResponse(ctx, req, false, auth, acl, te, entity, resp); err != nil {
+			return nil, nil, err
 		}
 	}
 
@@ -1283,8 +1292,11 @@ func (c *Core) handleLoginRequest(ctx context.Context, req *logical.Request) (re
 
 	// Do an unauth check. This will cause EGP policies to be checked
 	var auth *logical.Auth
+	var acl *ACL
+	var te *logical.TokenEntry
+	var entity *identity.Entity
 	var ctErr error
-	auth, _, ctErr = c.CheckToken(ctx, req, true)
+	auth, acl, te, entity, ctErr = c.CheckToken(ctx, req, true)
 	if ctErr == logical.ErrPerfStandbyPleaseForward {
 		return nil, nil, ctErr
 	}
@@ -1417,6 +1429,11 @@ func (c *Core) handleLoginRequest(ctx context.Context, req *logical.Request) (re
 				CreationPath: creationPath,
 				SealWrap:     sealWrap,
 			}
+		}
+
+		// Ensure compliance with List operation filtering.
+		if err := c.filterListResponse(ctx, req, true, auth, acl, te, entity, resp); err != nil {
+			return nil, nil, err
 		}
 	}
 
--- a/vault/request_handling_list_filtering.go
+++ b/vault/request_handling_list_filtering.go
@@ -1,0 +1,129 @@
+// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package vault
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/openbao/openbao/helper/identity"
+	"github.com/openbao/openbao/sdk/v2/logical"
+)
+
+func (c *Core) filterListResponse(ctx context.Context, req *logical.Request, unauth bool, auth *logical.Auth, acl *ACL, te *logical.TokenEntry, entity *identity.Entity, resp *logical.Response) error {
+	// Ignore lists with no results.
+	if resp == nil {
+		return nil
+	}
+
+	// Ignore non-list operations.
+	switch req.Operation {
+	case logical.ListOperation:
+	case logical.ScanOperation:
+	default:
+		return nil
+	}
+
+	// No filtering, so nothing to do.
+	if auth.ResponseKeysFilterPath == "" {
+		return nil
+	}
+
+	// Secret and Auth should not be set on List operation responses which
+	// expect to be filtered.
+	if resp.Secret != nil {
+		c.logger.Error("non-empty secret on filtered list response", "path", req.Path)
+		return ErrInternalError
+	}
+	if resp.Auth != nil {
+		c.logger.Error("non-empty secret on filtered list response", "path", req.Path)
+		return ErrInternalError
+	}
+
+	// Exit early when there is no data.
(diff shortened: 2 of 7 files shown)


cipherboy:
Test failure legitimate:

```
2025-06-04T09:03:33.0965222Z ##[group][31mFAIL[0m vault.TestPolicy_Parse/HCL (0.00s)
2025-06-04T09:03:33.0965535Z     policy_test.go:240: err: failed to parse policy: 1 error occurred:
2025-06-04T09:03:33.0965905Z         	* path "test/metadata/*": invalid key "response_keys_filter_path" on line 113
2025-06-04T09:03:33.0965994Z         
2025-06-04T09:03:33.0966000Z 
2025-06-04T09:03:33.0966205Z ##[endgroup]
2025-06-04T09:03:33.0966392Z   [31mFAIL[0m vault.TestPolicy_Parse (0.00s)
```

vault/request_handling_list_filtering.go
```diff
@@ -0,0 +1,153 @@
...
+	if keyInfoPresent {
+		keyInfo, ok = keyInfoRaw.(map[string]interface{})
+		if !ok {
+			c.logger.Error("invalid type for parameter key_info on filtered list response", "path", req.Path, "type", fmt.Sprintf("%T", keysRaw))
+			return ErrInternalError
+		}
+	}
+
+	filteredKeys := make([]string, 0, len(keys))
+	tmpl, err := compileTemplatePathForFiltering(auth.ResponseKeysFilterPath)
```
cipherboy, line 76: I ended up adjust things to take templates already, as I realized we need them to support path capability grants containing wildcards.

cipherboy (commented):
This should be ready for review @openbao/openbao-core-committers @openbao/openbao-org-maintainers.

website/content/docs/rfcs/filtering-list.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+This can be done later if users utilizing these features wish to have support.
+
+### Cross-Mount/Invalid Path Confusion
+
+One additional downside is that, because policy authors set `list_scan_response_keys_filter_path`, they could set this to a mount which doesn't exist or isn't the same as the list. That is, there is no common prefix requirement between the list and the templated path for filtering.
+
+### Wildcard Matching Difficulties
+
+Lastly, one downside is that there are difficulties in supporting a wildcard match: if _any_ path within a given prefix is allowed by any policy, show the key in the list result. For instance, if the list path was `sys/mounts`, one could envision a `list_scan_response_keys_filter_path` value of `{{ .key }}/*` to indicate a matching mount should only be returned if the token's policy grants at least one operation on said mount. This is hard as the underlying trie structures do not allow reverse wildcard lookup.
```
wslabosz-reply:
```suggestion
Lastly, one downside is that there are difficulties in supporting a wildcard match: if _any_ path within a given prefix is allowed by any policy, show the key in the list result. For instance, if the list path was `sys/mounts`, one could envision a `list_scan_response_keys_filter_path` value of `{{ .key }}/*` to indicate a matching mount should only be returned if the token's policy grants at least one operation on said mount. This is hard as the underlying tree structures do not allow reverse wildcard lookup.
```

website/content/docs/rfcs/filtering-list.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+For policy authors, this is another variable to consider but may make quality of life better on certain endpoints.
+
+However, the greatest improvement comes from web UI users: list results will now only contain relevant paths that a policy author sets and thus.
+
+## Unresolved Questions
+
+### Interactions with Pagination
+
+When implementing pagination, the design was given as two request keys (`after` and `limit`), returning the same response structure. The semantics were implicitly that an empty result meant no further data was present. As a user may not have access to a given range of keys (say, there are four-digit numbers encoded in a K/V entry and the user only has access to the one `2024`, `list` called with `after=1000` and `limit=100` would return 100 results from the plugin call which would be filtered to the empty list as the allowed `2024` is not in the returned `[1000, 1100)`). This would look to the client like they should finish iteration (as there are no results) instead of having ot increment their `after` value.
```
wslabosz-reply:
```suggestion
When implementing pagination, the design was given as two request keys (`after` and `limit`), returning the same response structure. The semantics were implicitly that an empty result meant no further data was present. As a user may not have access to a given range of keys (say, there are four-digit numbers encoded in a K/V entry and the user only has access to the one `2024`, `list` called with `after=1000` and `limit=100` would return 100 results from the plugin call which would be filtered to the empty list as the allowed `2024` is not in the returned `[1000, 1100)`). This would look to the client like they should finish iteration (as there are no results) instead of having to increment their `after` value.
```

website/content/docs/rfcs/filtering-list.mdx
cipherboy: Tries are a specialized form of trees fwiw: https://en.wikipedia.org/wiki/Trie

wslabosz-reply: 🤦 I went with reading the RFC notes before the actual change and it seems that I barely understand the RFC...

cipherboy: @wslabosz-reply Maybe I should move this section to a future work section? We opt not to support this sub-point.

At any rate, if you have suggestions for the RFC let me know :D

vault/request_handling.go
```diff
@@ -381,7 +381,7 @@ func (c *Core) fetchACLTokenEntryAndEntity(ctx context.Context, req *logical.Req
 	return acl, te, entity, identityPolicies, nil
 }
 
-func (c *Core) CheckToken(ctx context.Context, req *logical.Request, unauth bool) (*logical.Auth, *logical.TokenEntry, error) {
+func (c *Core) CheckToken(ctx context.Context, req *logical.Request, unauth bool) (*logical.Auth, *ACL, *logical.TokenEntry, *identity.Entity, error) {
```
wslabosz-reply, line 384: readability nit: as we've went from 3 to 5 returns, would it be beneficial to make them named returns?

vault/request_handling_list_filtering.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/openbao/openbao/helper/identity"
+	"github.com/openbao/openbao/sdk/v2/helper/template"
+	"github.com/openbao/openbao/sdk/v2/logical"
+)
+
+func (c *Core) filterListResponse(ctx context.Context, req *logical.Request, unauth bool, auth *logical.Auth, acl *ACL, te *logical.TokenEntry, entity *identity.Entity, resp *logical.Response) error {
+	// Ignore lists with no results.
+	if resp == nil {
+		return nil
+	}
```
wslabosz-reply: I'd collapse these cases into one
```suggestion
	// Ignore lists with no results.
	if resp == nil || resp.Data == nil {
		return nil
	}
```

vault/request_handling_list_filtering.go
```diff
@@ -0,0 +1,153 @@
...
+	if resp.Auth != nil {
+		c.logger.Error("non-empty secret on filtered list response", "path", req.Path)
+		return ErrInternalError
+	}
+
+	// Exit early when there is no data.
+	if resp.Data == nil {
+		return nil
+	}
```
wslabosz-reply, line 44: remove due to collapse in the first lines of the function
```suggestion
```

vault/request_handling_list_filtering.go
```diff
@@ -0,0 +1,153 @@
...
+	// Secret and Auth should not be set on List operation responses which
+	// expect to be filtered.
+	if resp.Secret != nil {
+		c.logger.Error("non-empty secret on filtered list response", "path", req.Path)
+		return ErrInternalError
+	}
+	if resp.Auth != nil {
+		c.logger.Error("non-empty secret on filtered list response", "path", req.Path)
+		return ErrInternalError
+	}
```
wslabosz-reply: I'd happily keep them separated if these differed at least in log message (or return value), but in this case:
```suggestion
	if resp.Secret != nil || resp.Auth != nil {
		c.logger.Error("non-empty secret on filtered list response", "path", req.Path)
		return ErrInternalError
	}
```

vault/request_handling_list_filtering.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil
+	}
+
+	// Ignore non-list operations.
+	switch req.Operation {
+	case logical.ListOperation:
+	case logical.ScanOperation:
+	default:
+		return nil
+	}
```
wslabosz-reply, line 23: nit: if case of `if req.Operation != logical.ListOperation || req.Operation != logical.ScanOperation` would be more readable for me, but it's not a hill I'd die on, just if we never consider more "list type operations" then the switch is a tad bit less readable to me

vault/request_handling_list_filtering.go
```diff
@@ -0,0 +1,153 @@
...
+	"strings"
+
+	"github.com/openbao/openbao/helper/identity"
+	"github.com/openbao/openbao/sdk/v2/helper/template"
+	"github.com/openbao/openbao/sdk/v2/logical"
+)
+
+func (c *Core) filterListResponse(ctx context.Context, req *logical.Request, unauth bool, auth *logical.Auth, acl *ACL, te *logical.TokenEntry, entity *identity.Entity, resp *logical.Response) error {
+	// Ignore lists with no results.
+	if resp == nil {
```
wslabosz-reply: in both two references this function is used `vault/response_handling.go#1435` and `vault/response_handling.go#1082` the immediate preceding if clause is like `resp != nil` could we consider invoking this function in a `else` clause and remove the check?

website/content/docs/rfcs/filtering-list.mdx
wslabosz-reply: If we plan to add that in future update, then yes, please move that there

vault/request_handling_test.go
```diff
@@ -473,3 +476,259 @@ func TestRequestHandling_SecretLeaseMetric(t *testing.T) {
...
+
+	req.Path = "sys/mounts/secret"
+	req.Operation = logical.UpdateOperation
+	req.Data = map[string]interface{}{
+		"type": "kv-v2",
+	}
+
+	resp, err = core.HandleRequest(namespace.RootContext(nil), req)
+	require.NoError(t, err)
+	require.False(t, resp.IsError())
```
wslabosz-reply: just checking: this whole flow is about removing default mount and adding manually the kv-v2 mount? (is this really required?)

vault/request_handling_test.go
```diff
@@ -473,3 +476,259 @@ func TestRequestHandling_SecretLeaseMetric(t *testing.T) {
...
+		"type": "userpass",
+	}
+
+	resp, err = core.HandleRequest(namespace.RootContext(nil), req)
+	require.NoError(t, err)
+	require.False(t, resp.IsError())
+
+	// Add policy
+	req.Path = "sys/policies/acl/list-filtered"
+	req.Data = map[string]interface{}{
```
wslabosz-reply, line 511: (not a request for a change)

@cipherboy wdyt?
in future I'd be very happy if we started using https://pkg.go.dev/embed for defining any long text variables, might be my personal preference, but I prefer when the input data like that is separated from the code logic.

vault/request_handling_test.go
```diff
@@ -473,3 +476,259 @@ func TestRequestHandling_SecretLeaseMetric(t *testing.T) {
...
+				}
+
+				for _, entry := range resp.Data["keys"].([]string) {
+					if strings.HasSuffix(entry, "/") {
+						continue
+					}
+
+					onList, present := entries[entry]
+					if !present {
+						t.Fatalf("list included %v but shouldn't have; path: %v\n\texpected: %#v\n\tactual: %#v", entry, req.Path, entries, resp.Data["keys"].([]string))
```
wslabosz-reply: nit: if you are already using the `require` lib, is there a reason for spontaneous usage of the `t.Fatalf()`?

vault/request_handling_list_filtering.go
cipherboy: Ah, copy/paste typo in the second one, it should be non-empty _auth_ on filtered list response.

vault/request_handling_test.go
cipherboy: Right, the default mount is KVv1, which doesn't have the distinction between data/metadata. See https://openbao.org/api-docs/secret/kv/kv-v1/ vs https://openbao.org/api-docs/secret/kv/kv-v2/.  

We want to test that we can use either path (data or metadata) to control access to keys visible in the list response (where in KVv2, list is a metadata-associated item).

wslabosz-reply: that's what I thought, I'm not feeling great how we have this 10+ lines of code not related to what we want to test because we need setup the secret engine, but I guess there's no other way.

cipherboy: Yeah, since the APIs are different enough, we can't just upgrade the global helper's version to v2 sadly :(

website/content/docs/rfcs/filtering-list.mdx
cipherboy: I added a note to the end of the section that this could be done in the future in a follow-up RFC, since the lack of support for wildcard based filtering (i.e., allowing an entry `x` if any path under `<prefix>/x/*` is allowed) is still a downside that some people could've wanted.

vault/request_handling_test.go
cipherboy: @wslabosz-reply Hmm, I'd ordinarily use embed or similar (we have helpers in `pkiext` for example) if this were just a config file/prerequisite and not what we're actually testing.

However, in our case, we're very much testing the behavior of this rather complex policy. We could split it up into separate entities, but I wanted to see how they'd interact with multiple rules &c in a single (large) policy. Splitting that into another place just serves to confuse the reader as to where that policy lives.

vault/request_handling.go
cipherboy: @wslabosz-reply I think named returns make a lot of sense when there's duplicate or generic types, but in this case, all of the variable names would very closely match the unique returned types:

 - `auth`, for `*logical.Auth`,
 - `acl` for `*ACL`,
 - `te` for `*logical.TokenEntry` 
 - `entity` for `*identity.Entity`,
 - and `err` obviously :-D

Which I think doesn't buy us _that_ much? If it were `(string, string, int, int, err)` or similar, I think we could definitely benefit from naming returns. 

That said, the Google Go style guide doesn't actually mention anything about named returns https://google.github.io/styleguide/go/best-practices.html

Happy to change if you have stronger preferences otherwise. :-)

vault/request_handling_list_filtering.go
cipherboy: I've opted to remove this condition as you pointed out we already enforce it in the caller.

vault/request_handling_list_filtering.go
cipherboy: @wslabosz-reply I think I hinted at this in #1400, but @Gabrielopesantos has started a design doc for a new `SEARCH` operation. The output will likely look like a (detailed?) list for interoperability. This is part of the reason why I want to land list filtering. :-)

vault/request_handling_test.go
cipherboy: Ah, mistake :-D Fixed!

cipherboy:
@wslabosz-reply This should be ready for re-review :-)

vault/request_handling_test.go
wslabosz-reply: in this case you are right, as it's that one policy that we want to test against, and not a defined set of input files.

vault/request_handling.go
wslabosz-reply: yup, you are right, when the types are distinct as in this case, then this is explicit enought.

cipherboy: @wslabosz-reply do you want to see if there's a good style guide for Go, preferably one with some automated enforcement tooling, and make a recommendation to DevWG on where to go? Seems like you've had some good suggestions in this front.

cipherboy:
Thanks @wslabosz-reply! I'm going ahead and admin merging; as far as I can tell, we're blocked on CODEOWNERS for `changelog/` so I'll open up a follow-up PR to fix that and the `go.*` files.

PR #1398, opened 2025-06-02
Add new core, kubernetes teams to CODEOWNERS

cipherboy:
After the creation of the new teams, we should add them to the CODEOWNERS file.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,9 +3,33 @@
 # Everything
 * @openbao/openbao-org-maintainers
 
+# Core team ownership
+api/ @openbao/openbao-org-maintainers @openbao/openbao-core-committers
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -3,9 +3,33 @@
 # Everything
 * @openbao/openbao-org-maintainers
 
+# Core team ownership
```
cipherboy, line 6: @DanGhita @JanMa @naphelps I think the `vault` team should probably have fairly broad ownership of things, though I've excluded most plugins (other than KV -- but maybe should also include either AppRole or UserPass, which are fairly core anyways).

We can certainly create more granular additional teams if we pick up interested parties (e.g., for juts the CLI or just the API/SDK), but core team will probably want to be tagged at least for informational purposes.

Thoughts?

cipherboy:
Thanks @JanMa, merging.

PR #1406, opened 2025-06-03
Namespaces UI support

### LINKED ISSUE #1405: Namespaces UI support
Add namespace support to the existing UI.

driif:
Resolves #1405

### DIFF
--- a/ui/app/templates/vault/cluster/access/namespaces/index.hbs
+++ b/ui/app/templates/vault/cluster/access/namespaces/index.hbs
@@ -21,7 +21,7 @@
         Create Namespace
       </LinkTo>
 
-      <DocLink @path="/vault/tutorials/enterprise/namespaces">
+      <DocLink @path="/blog/namespaces-announcement/#how-to-use-namespaces">
         Learn more
       </DocLink>
     </list.empty>
--- a/ui/app/templates/vault/cluster/access/namespaces/index.hbs
+++ b/ui/app/templates/vault/cluster/access/namespaces/index.hbs
@@ -1,0 +1,63 @@
+<PageHeader as |p|>
+  <p.levelLeft>
+    <h1 class="title is-3">
+      Namespaces
+    </h1>
+  </p.levelLeft>
+</PageHeader>
+
+<Toolbar>
+  <ToolbarActions>
+    <ToolbarLink @route="vault.cluster.access.namespaces.create" @type="add">
+      Create namespace
+    </ToolbarLink>
+  </ToolbarActions>
+</Toolbar>
+
+<ListView @items={{this.model}} @itemNoun="namespace" @paginationRouteName="vault.cluster.access.namespaces" as |list|>
+  {{#if list.empty}}
+    <list.empty>
+      <LinkTo @route="vault.cluster.access.namespaces.create">
+        Create Namespace
+      </LinkTo>
+
+      <DocLink @path="/vault/tutorials/enterprise/namespaces">
+        Learn more
+      </DocLink>
+    </list.empty>
+  {{else}}
+    <ListItem as |Item|>
+      <Item.content>
+        {{list.item.id}}
+      </Item.content>
+      <Item.menu as |m|>
+        {{#let (concat this.currentNamespace (if this.currentNamespace "/") list.item.id) as |targetNamespace|}}
+          {{#if (includes targetNamespace this.accessibleNamespaces)}}
+            <li class="action">
+              <NamespaceLink @targetNamespace={{targetNamespace}} @unparsed={{true}} @class="is-block">
+                Switch to Namespace
+              </NamespaceLink>
+            </li>
+          {{/if}}
+        {{/let}}
+        <li class="action">
+          <m.Message
+            @id={{list.item.id}}
+            @confirmButtonText="Remove"
+            @message="Any engines or mounts in this namespace will also be removed."
+            @onConfirm={{action
+              (perform
+                Item.callMethod
+                "destroyRecord"
+                list.item
+                (concat "Successfully deleted namespace: " list.item.id)
+                "There was an error deleting this namespace: "
+                (action "refreshNamespaceList")
+              )
+            }}
+          />
+        </li>
+      </Item.menu>
+    </ListItem>
+  {{/if}}
+</ListView>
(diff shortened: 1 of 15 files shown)


a-chartier:
@driif  

Just looking at this - I am noticing some strange behaviour when logging into a child namespace.

so in this example I am logging into the namespace `something-new` 
image

image


couple of things I want to mention here . `Testing` is not within the child namespace of `something-new`  and the `something-new` namespace actually has child namespaces  which for some reason seem to be nested incorrectly 

image

This is the structure of the test namespaces

```shell
.
└── root/
    ├── testing
    └── something-new/
        ├── a
        ├── b
        └── c
```

a-chartier:
@driif  - I apologize I was short on time yesterday, I could have done some more digging into this.  

This is the response I am getting back from `v1/sys/internal/ui/namespaces`

```json
{
    "request_id": "79ee03cb-9070-d55f-1c6e-3cf7020cfe30",
    "lease_id": "",
    "renewable": false,
    "lease_duration": 0,
    "data": {
        "keys": [
            "something-new/a/",
            "something-new/b/",
            "something-new/c/",
            "",
            "testing/",
            "something-new/"
        ]
    },
    "wrap_info": null,
    "warnings": null,
    "auth": null
}
```
It's interesting since this call is passing the  correct header 

```
x-vault-namespace something-new
```

which I would would assume should make all calls relative to that namespace 


when I check the `/v1/sys/namespaces?list=true` (manage namespaces list)  I do se the correct child namespaces listed 

```json
{
    "request_id": "47cab0a8-dacb-2f42-ace4-55865175bdff",
    "lease_id": "",
    "renewable": false,
    "lease_duration": 0,
    "data": {
        "key_info": {
            "a/": {
                "custom_metadata": {},
                "id": "0QSphn",
                "locked": false,
                "path": "something-new/a/",
                "tainted": false,
                "uuid": "fb1a696b-daa2-837b-ae3c-f75b7c60755e"
            },
            "b/": {
                "custom_metadata": {},
                "id": "T9ImXi",
                "locked": false,
                "path": "something-new/b/",
                "tainted": false,
                "uuid": "ac14e76f-307b-79fe-bea8-1aecd31d3940"
            },
            "c/": {
                "custom_metadata": {},
                "id": "t59aiz",
                "locked": false,
                "path": "something-new/c/",
                "tainted": false,
                "uuid": "629dbe4f-5c61-0069-c506-449756a9d538"
            }
        },
        "keys": [
            "a/",
            "b/",
            "c/"
        ]
    },
    "wrap_info": null,
    "warnings": null,
    "auth": null
}
```

ui/app/templates/vault/cluster/access/namespaces/index.hbs
```diff
@@ -0,0 +1,63 @@
...
+<ListView @items={{this.model}} @itemNoun="namespace" @paginationRouteName="vault.cluster.access.namespaces" as |list|>
+  {{#if list.empty}}
+    <list.empty>
+      <LinkTo @route="vault.cluster.access.namespaces.create">
+        Create Namespace
+      </LinkTo>
+
+      <DocLink @path="/vault/tutorials/enterprise/namespaces">
+        Learn more
+      </DocLink>
```
cipherboy: We probably want to change this link?

cipherboy (commented):
Hmm, looks like it works better for top-level namespace but the child is still broken:

image

image

image

ui/app/templates/vault/cluster/access/namespaces/index.hbs
driif: @cipherboy sure! Changed it to navigate to https://openbao.org/blog/namespaces-announcement/#how-to-use-namespace

driif:
@a-chartier @cipherboy Fixed the issue of always using the root namespace instead of the current one in namespace-picker. Thank you for the detailed feedback and help with debugging.

ui/app/templates/vault/cluster/access/namespaces/index.hbs
```diff
@@ -0,0 +1,63 @@
...
+            @message="Any engines or mounts in this namespace will also be removed."
+            @onConfirm={{action
+              (perform
+                Item.callMethod
+                "destroyRecord"
+                list.item
+                (concat "Successfully deleted namespace: " list.item.id)
+                "There was an error deleting this namespace: "
+                (action "refreshNamespaceList")
+              )
```
satoqz, line 56: Namespace deletion is async, so I'm not sure it is really "successfully deleted" here. Optimally we would poll until deletion is finished, else "scheduled for deletion" is probably more accurate.

vault/logical_system_helpers.go
```diff
@@ -35,7 +36,23 @@ var pathInternalUINamespacesRead = func(b *SystemBackend) framework.OperationFun
 			return nil, logical.ErrPermissionDenied
 		}
 
-		return logical.ListResponse([]string{""}), nil
+		parent, err := namespace.FromContext(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		list, err := b.Core.namespaceStore.ListNamespaces(ctx, false, false)
```
satoqz, line 44: Should these recursively list so you can navigate downwards more easily? When you want to go from `a/` to `a/b/c/` right now, you'd have to go to `a/b/` first. Maybe we can even have a toggle in the UI to switch between direct descendants and all descendants only?

ui/app/templates/vault/cluster/access/namespaces/index.hbs
driif: @satoqz Hi Jonas, I probably won’t change it in this PR, but thanks for the reminder. We should add support for Server-Side Events in the new UI to receive backend notifications. It would be great to send a normal OK response when the process starts, and later send an event to show updates.

vault/logical_system_helpers.go
driif: Like above, I wouldn’t invest time on this in the current PR, but it’s a great suggestion. It’s something we could improve in the future using the backend functions we already have.

JanMa (changes requested):
Can you please address the failing UI test? 
>   Global error: Uncaught TypeError: Cannot read properties of undefined (reading 'userRootNamespace') at http://localhost:7357/ui/assets/vendor.js, line 144490
While executing test: Integration | Component | sidebar-frame: it should hide and show sidebar
: Chrome 136.0}

Other that that, looks grat. Thanks a lot @driif !

cipherboy:
@driif Since this came after beta, do you mind adding a changelog entry? Thanks :-)

PR #1408, opened 2025-06-04
Fix some code deprecations

phyrog:
Fixes a usage of a couple of deprecated functions/types/packages.

More to come 🙂

### DIFF
--- a/builtin/logical/pki/crl_test.go
+++ b/builtin/logical/pki/crl_test.go
@@ -5,7 +5,6 @@ package pki
 
 import (
 	"context"
-	"encoding/asn1"
 	"encoding/json"
 	"fmt"
 	"strings"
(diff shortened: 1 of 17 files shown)


builtin/logical/pki/crl_test.go
```diff
@@ -198,7 +198,7 @@ func TestBackend_CRL_AllKeyTypeSigAlgos(t *testing.T) {
 		if strings.HasSuffix(tc.SigAlgo, "PSS") {
 			algo := crl.SignatureAlgorithm
 			pssOid := asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 1, 10}
-			if !algo.Algorithm.Equal(pssOid) {
+			if algo.String() != pssOid.String() {
```
cipherboy: Does this actually work? Aren't we using a comparison between an OID (on the right) and a pretty name on the left (or an [id if missing](https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/crypto/x509/x509.go;l=254))?

I wish they had exposed `isRSAPSS` lol.

But I think you need to do a `strings.Contains(algo.String(), "PSS")` if I understand the refactor correctly.

cipherboy: Ah, test failures look legit; I missed the other ones (about malformed CRLs), but:

```
    crl_test.go:202: tc 3 failed: expected sig-alg to be 1.2.840.113549.1.1.10 / got SHA256-RSAPSS
```

looks about what I expected. :-)

phyrog: Yeah I need to look into this. Some of the tests are also failing.

phyrog: Fixed

satoqz (approved):

Issue #1409, opened 2025-06-05
Add FIPS-140-3 Compliant Build


voigt:
**Is your feature request related to a problem? Please describe.**  
Organizations in regulated industries such as government, finance, and healthcare often require software that complies with cryptographic standards like FIPS 140-3. OpenBao currently does not offer an official build that meets FIPS 140-3 requirements, which limits its use in environments with strict compliance mandates.

**Describe the solution you'd like**  
We would like OpenBao to provide an official FIPS 140-3 compliant build.

Starting with Go 1.24, the Go toolchain supports FIPS 140-3 through a dedicated cryptographic module (`GOFIPS140`) and runtime toggles (`GODEBUG=fips140=on|only`). 

The OpenBao build/release process should support these options to provide a FIPS-140-3 compliant distribution.

**Additional Considerations:**

As OpenBao depends on wrappers/plugins in [go-kms-wrappers](github.com/openbao/go-kms-wrapping/), we possibly need to consider it as well.

**Describe alternatives you've considered**  
Yet another build increases OpenBaos (already complex) distribution matrix by yet another dimension. Building OpenBao by default FIPS-140-3 compliant could be an alternative. Admittedly, at this point in time I have no understanding of all consequences and dependencies.

**Explain any additional use-cases**  
- Government agencies deploying OpenBao for secrets management.
- Enterprises requiring FIPS-validated cryptographic operations for audit and compliance.
- Vendors integrating OpenBao into their products needing FIPS assurance.

**Additional context**  
FIPS 140-3 is the latest U.S. government standard for cryptographic modules and aligns with international standards (ISO/IEC 19790:2012). By enabling a FIPS-compliant build, OpenBao can gain adoption in security-sensitive environments and demonstrate strong cryptographic assurance.

Relevant reference:  
- https://go.dev/doc/security/fips140

cipherboy:
Happy to take a FIPS build upstream.


Some general thoughts:

- FIPS compatibility is a breaking change, so we'd need separate binaries for supported targets for it. Likely Linux/AMD64 in HSM and non-HSM variants.
- It'd be good to know what [level](https://www.entrust.com/resources/learn/what-fips-140-3) we're looking at; software is easy but moving to a higher hardware-backed level will require more work that upstream Go toolchains have not done.
- Supported/desired architectures and algorithms might also require we choose an alternative to the upstream Go toolchain and move to e.g., Red Hat's or similar.


[...]

PR #1410, opened 2025-06-05
Add FIPS-140-3 Build of OpenBao

voigt:
This PR adresses #1409, by providing a FIPS-140-3 build of OpenBao.
It furthermore extends InfoKeys by a field indicating whether FIPS-140 is enabled, similar to the existing `cgo` InfoKey.

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -5,6 +5,7 @@ package command
 
 import (
 	"context"
+	"crypto/fips140"
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/hex"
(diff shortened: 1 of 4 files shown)


klaus-sap:
https://go.dev/doc/security/fips140: should we use GOFIPS140=v1.0.0 or GOFIPS140=latest?

cipherboy:
@klaus-sap I suspect in a bit we'll formally kick off a FIPS WG where we'll examine the answers to that question. I concur with `needs-rfc` label though.

cipherboy:
I'll close this for now; @-voigt has left Reply and as mentioned above we want to kick off the FIPS WG first, before we get too far down this route.

PR #1411, opened 2025-06-05
Clarify confusing error if auto-unsealing & uninitialized

satoqz:
I've seen several people, including myself, be confused by

```log
[WARN]  failed to unseal core: error="stored unseal keys are supported, but none were found"
```

when starting a server that uses auto-unsealing for the first time, eventually figuring out that this is telling you to `bao operator init` 😄

I suggest we add a little more context to the message to make it easier to understand what the issue is likely to be.

### DIFF
--- a/changelog/1411.txt
+++ b/changelog/1411.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-auto-unsealing: The warning logged when the server is still uninitialized and auto-unsealing is configured is now less confusing
+auto-unsealing: Improved the clarity of the warning message logged when the server is uninitialized and auto-unsealing is configured.
 ```


vault/init.go
```diff
@@ -448,7 +448,7 @@ func (c *Core) UnsealWithStoredKeys(ctx context.Context) error {
 	// This usually happens when auto-unseal is configured, but the servers have
 	// not been initialized yet.
 	if len(keys) == 0 {
-		return NewNonFatalError(errors.New("stored unseal keys are supported, but none were found"))
+		return NewNonFatalError(errors.New("stored unseal keys are supported, but none were found: is the server initialized?"))
```
wslabosz-reply, line 451: can we make it more direct? Like "if your instance is setup with auto-unsealing, check your initialization status"?

satoqz: IIUC if we hit this condition we already know that we are supposed to auto-unseal, so no reason for "if your instance is setup with auto-unsealing".

changelog/1441.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+auto-unsealing: The warning logged when the server is still uninitialized and auto-unsealing is configured is now less confusing
```
voigt: A less colloquial proposal:

```suggestion
auto-unsealing: Improved the clarity of the warning message logged when the server is uninitialized and auto-unsealing is configured.
```

cipherboy:
I'm fine with it in this instance since it was adjusting an error message, but let's try to get a review from another organization in the future for larger changes. Thanks! :-)

satoqz:
@cipherboy noted, that's a good policy.

Issue #1413, opened 2025-06-06
Replace LF Edge references with OpenSSF


karras:
A few days ago OpenBao was migrated from the LF Edge to the OpenSSF. With that we also need to adjust various references across our repositories, code base, website, and more. This issue helps us track those changes and PRs.

We would want to avoid changing all entries in bulk since various references (e.g. Wiki) are still valid and must only be changed if and once the content is migrated.

* [x] Mailing lists
* [x] Wiki (will those pages be migrated?)
* [x] Contact information
  - GPG key
  - goreleaser
  - ...
* [x] openbao.org website
* [x] Other websites
* [x] Logo

cipherboy:
>  Wiki (will those pages be migrated?)

I have not yet heard back on wiki fwiw.

cipherboy:
@karras The mailing list has been updated if you want to finish that up. :-) We'll use google docs instead of the wiki going forward.

karras:
> [@karras](https://github.com/karras) The mailing list has been updated if you want to finish that up. :-) We'll use google docs instead of the wiki going forward.

Thanks! I'll do another org-level search to check if I've missed anything.

karras:
We'll skip modifying the GPG key for the time being since the domain is being forwarded, and it would be quite some effort for the small benefits.

karras:
All references should have been properly cleaned up in the meantime.

PR #1415, opened 2025-06-06
Point goreleaser and container image contacts to new OpenSSF domain

karras:
Replaces the contact information in our go and container image builds with the new OpenSSF address. Please let me know if a changelog entry is needed or not.

Related to #1413.


satoqz (approved):
Not sure about the changelog, just double-checked that you've caught everything 😄

voigt (approved):

cipherboy:
Thanks @karras! 

No, changelog are only needed for binary changes. Arguably the mailing list changes the binary I suppose but in a trivial way.

PR #1424, opened 2025-06-09
Remove deprecated recovery config location fallback

wslabosz-reply:
During beta version of upstream (v0.11.4) there was a deprecation of a `recovery-seal-config` entry path of recovery config. This PR removes the fallback checking old location for a possible migration to a new path.

This removal doesn't contradict the [OpenBao compatibility proposal](https://openbao.org/docs/policies/migration/#proposal). OpenBao supports migration from Vault v1.14, not guaranteeing the proper migration path from older versions (especially beta versions of Vault). This change removes complexity from the already complex system of sealing and seal migrations.

---
cc: @cipherboy

### DIFF
--- a/changelog/1424.txt
+++ b/changelog/1424.txt
@@ -1,0 +1,3 @@
+```release-note:change
+vault/seal: removal of deprecated migration path of an old (encrypted) recovery config location
+```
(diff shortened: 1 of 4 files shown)


changelog/1424.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:change
+vault/seal: removal of deprecated migration path of an old (encrypted) recovery config location
```
cipherboy:
```suggestion
vault/seal: removal of deprecated migration path of an old pre-Vault v1.0 (encrypted) recovery config location
```

cipherboy (approved):
One nit, then I think things are good.

I validated that the plaintext configuration is safe to have in, well, plaintext.

satoqz (approved):

changelog/1424.txt
wslabosz-reply: added

PR #1425, opened 2025-06-09
Add static key auto-unseal

cipherboy:
This adds support for the static key auto-unseal mechanism, useful for
chaining from another secrets manager in an automated environment like
Kubernetes. Currently the encryption algorithm is fixed at
AES-256-GCM-96, but this could be made configurable in the future.

See also: https://github.com/openbao/go-kms-wrapping/pull/31
See also: https://github.com/openbao/openbao/pull/1304


website/content/docs/configuration/seal/static.mdx
```diff
@@ -0,0 +1,86 @@
...
+description: |-
+  The static key seal configures OpenBao to use static keys provided alongside
+  the configuration file as the autounseal mechanism.
+---
+# `static` seal
+
+The static key seal configures OpenBao to use static keys provided alongside 
+the configuration file as the autounseal mechanism. These can either be provided
+directly, in base64 or hex form, as an environment variable (using the `env://`
+syntax), or as a file (using the `file://` syntax). This key must be
```
karras: Unfinished sentence

website/content/docs/configuration/seal/static.mdx
```diff
@@ -0,0 +1,86 @@
...
+directly, in base64 or hex form, as an environment variable (using the `env://`
+syntax), or as a file (using the `file://` syntax). This key must be 
+
+OpenBao's static key unseal support is activated by the following:
+
+- The presence of a `seal "static"` block in OpenBao's configuration file.
+
+:::warning
+
+Carefully evaluate use of static key auto-unseal to see if its use meets the
```
karras: What's our preferred spelling of auto-unseal? Auto-unseal, autounseal or auto unseal? I'd recommend `auto unseal`.

website/content/docs/configuration/seal/static.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+- `current_key_id` `(string: <required>)`: permanent identifier of the key
+  in the `current_key` field; whenever the final value of `current_key`
+  changes (either directly or via modifying the referenced environment
+  variable or file).
+
+- `current_key` `(string: <required>)`: key to use for new seal operations;
+  may have an `env://` or `file://` prefix to reference an environment
+  variable or file respectively. Must be a 32-byte key to be used as an
+  AES-256-GCM-96 key.
```
karras, line 64: Maybe emphasize the algorithm part to ensure operators read this important remark twice, e.g. add another sentence here or elsewhere such as "Other key algorithms are currently not supported.".

website/content/docs/configuration/seal/static.mdx
```diff
@@ -0,0 +1,86 @@
...
+These parameters apply to the `static` stanza in the OpenBao configuration
+file:
+
+- `current_key_id` `(string: <required>)`: permanent identifier of the key
+  in the `current_key` field; whenever the final value of `current_key`
+  changes (either directly or via modifying the referenced environment
+  variable or file).
+
+- `current_key` `(string: <required>)`: key to use for new seal operations;
+  may have an `env://` or `file://` prefix to reference an environment
```
JanMa, line 62: Instead of allowing an environment variable reference via `env://` I would prefer if we come up with predefined variable names, similar to the other unseal mechanisms. For example `BAO_SEAL_TYPE=static`, `BAO_STATIC_KEY` and `BAO_STATIC_KEY_ID`.

cipherboy: @JanMa This was due to the use of `ParsePath`, which seems to be the new pattern in `go-kms-wrapping` after upstream commit https://github.com/hashicorp/go-kms-wrapping/commit/9aac87b758e1aa0b9734b9a499484fad96568985 + https://github.com/hashicorp/go-kms-wrapping/commit/a1337fd541fe9d97ee2d9c8ef67c7d4c1646d3fc (MPL) that we pulled. Credentials _should_ (per pattern)  --- and thus presumably keys in this case --- also be pulled from values which `parseutil.ParsePath(...)` understands, which is `env://` and `file://`, with no option in https://pkg.go.dev/github.com/hashicorp/go-secure-stdlib/parseutil#ParsePath to disable environment parsing...

So I don't think we have much choice, but if you strongly disagree, we can back that feature out of g-k-w (though, it is a bit late as it landed in v2.2.0 though we could back it out just for static keys). 


When it comes time to land external keys (cc @voigt @satoqz), we should disable the environment variables as it is unsafe from a strict multi-tenancy perspective. This might require reviving our fork of go-secure-stdlib (or trying to contribute again upstream) to add an option to disable it.

satoqz: @cipherboy For external keys, I believe this can be achieved in just the wrapper-specific `options.go` files.. you would want to differentiate between wrapper options and external keys options anyways (even if many are shared), and then call `GetEnv` and `ParsePath` for wrapper options only.

website/content/docs/configuration/seal/static.mdx
satoqz: True, the docs are very inconsistent here already. My vote is:
- The noun: Auto Unseal (Explicitly capitalized)
- The verb: to auto-unseal, auto-unsealing

internalshared/configutil/kms.go
```diff
@@ -381,6 +385,25 @@ func GetKmipKMSFunc(kms *KMS, opts ...wrapping.Option) (wrapping.Wrapper, map[st
...
 
+func GetStaticKMSFunc(kms *KMS, opts ...wrapping.Option) (wrapping.Wrapper, map[string]string, error) {
+	wrapper := statickms.NewWrapper()
+	wrapperInfo, err := wrapper.SetConfig(context.Background(), append(opts, wrapping.WithConfigMap(kms.Config))...)
+	if err != nil {
+		// If the error is any other than logical.KeyNotFoundError, return the error
+		if !errwrap.ContainsType(err, new(logical.KeyNotFoundError)) {
+			return nil, nil, err
+		}
+	}
```
satoqz, line 396: I see most other KMS wrappers check this too, but looking at g-k-w I don't think we would ever get that error?

cipherboy: Yeah, I kinda wondered how the wrapper would behave in the future with asymmetric keys. There are a few PQC algorithms where the public key isn't necessarily derivable from the private key, so maybe you'd get a KeyNotFoundError if a field is missing? Maybe the wrapper should follow this construct anyways, since there's no way to create the key later? Thoughts?

satoqz: I'm missing why we would want to discard a `KeyNotFoundError` in the first place, doesn't that yield a dysfunctional sealing mechanism?

cipherboy: @satoqz I think this goes back to Friday's call -- the seal mechanism in certain scenarios (e.g., Vault Enterprise) _may_ implicitly generate key material. This isn't something we want to be implicit and should make it explicit. It is typically generated on encrypt/... calls and not on startup.

satoqz: Ah, got it, thanks. Fair enough.

satoqz (approved):

satoqz:
Changelog + Go mod checks outstanding, but looks good!

website/content/docs/configuration/seal/static.mdx
cipherboy: @JanMa https://github.com/openbao/go-kms-wrapping/pull/36 should have the basis for the proper environment variables.

cipherboy:
Thank you all! Merging...

website/content/docs/configuration/seal/static.mdx
```diff
@@ -0,0 +1,88 @@
+---
+sidebar_label: Static Key
+description: |-
+  The static key seal configures OpenBao to use static keys provided alongside
+  the configuration file as the Auto Anseal mechanism.
```
DrDaveD, line 5: Typo, change `Anseal` -> `Unseal`

website/content/docs/configuration/seal/static.mdx
```diff
@@ -0,0 +1,88 @@
...
+---
+# `static` seal
+
+The static key seal configures OpenBao to use static keys provided alongside
+the configuration file as the Auto Unseal mechanism. These can either be provided
+directly, in base64 or hex form, as an environment variable (using the `env://`
+syntax), or as a file (using the `file://` syntax). This key must be a 32-bytes
```
DrDaveD, line 12: change "a 32-bytes" to "32 bytes"

Issue #1407, opened 2025-06-03
Build HSM enabled packages for ARM64


sebastianw:
**Is your feature request related to a problem? Please describe.**
I have a redundant setup where two of three nodes are running on ARM64 Debian. I would like to use HSM Auto-Unseal which doesn't work on ARM64 as there are no packages available.

**Describe the solution you'd like**
ARM64 packages with HSM support.

**Describe alternatives you've considered**
None as I'm fixed to this system architecture.

**Additional context**
I'm using p11-kit remote forwarding from the ARM boxes to the HSM, but that is probably not relevant for this issue. Normal pkcs11 commands work fine.

satoqz:
Hi @sebastianw, thanks for the issue! @voigt and I have also been interested in this and had a preliminary look some days ago, but weren't sure if there was demand from the community -- seems like there is 😄

FWIW, auto-unsealing via HSM works fine on ARM64 if you build it yourself. Maybe that can be a temporary solution for you while we figure out how to integrate this into the release pipeline properly.

sebastianw:
@satoqz Thanks, I'll look into compiling it myself in the meantime.

cipherboy:
@satoqz I forget, we decided arm64 support was just a single line in goreleaser, right? But it was musl support that was harder? If that's true, if someone wants to open a PR, we can add it to v2.3.0 GA.

satoqz:
@cipherboy all of it is hard :D

For any arch other than x86_64 we need to either cross-compile C or emulate or have a runner with that arch... all of those take a bit more effort to integrate into the existing pipeline.

cipherboy:
@satoqz Luckily I think based on [this](https://goreleaser.com/cookbooks/cgo-and-crosscompiling/), it'd be a feature of GoReleaser Pro, which the OpenSSF is working on purchasing: https://github.com/ossf/tac/issues/475

So this might be easier to do in a little bit.

sebastianw:
@satoqz Could you maybe give me a hint how to compile with HSM support? I looked into the `Makefile` and `goreleaser.hsm.yaml` files but can't figure it out right now.

cipherboy:
@sebastianw Depending on if you want the UI or not, it really just is a fancy `go build ...`: https://openbao.org/docs/contributing/packaging/ :-) 

(`hsm` is a tag and you should set a few values, though those are kinda optional but strongly recomended)

satoqz:
To expand:

```sh
[...]

PR #1427, opened 2025-06-10
HSM: Add arm64 builds & Alpine containers

satoqz:
See https://github.com/openbao/openbao/issues/1407#issuecomment-2952858958, we install cross-compiling GCC from the Ubuntu repositories and build a single `aarch64-unknown-linux-gnu` binary, then reuse it in Alpine images via `gcompat`.

We could easily add support for other architectures with the same pattern if there is interest.

Resolves https://github.com/openbao/openbao/issues/1407

### DIFF
--- a/goreleaser.hsm.yaml
+++ b/goreleaser.hsm.yaml
@@ -20,10 +20,14 @@ builds:
       - -X github.com/openbao/openbao/version.fullVersion={{.Version}} -X github.com/openbao/openbao/version.GitCommit={{.Commit}} -X github.com/openbao/openbao/version.BuildDate={{ .Date }} -X github.com/openbao/openbao/version.VersionMetadata=hsm
     env:
       - CGO_ENABLED=1
+      - >-
+        {{- if eq .Arch "amd64"}}CC=x86_64-linux-gnu-gcc{{- end}}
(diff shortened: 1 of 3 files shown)


goreleaser.hsm.yaml
```diff
@@ -20,10 +20,14 @@ builds:
       - -X github.com/openbao/openbao/version.fullVersion={{.Version}} -X github.com/openbao/openbao/version.GitCommit={{.Commit}} -X github.com/openbao/openbao/version.BuildDate={{ .Date }} -X github.com/openbao/openbao/version.VersionMetadata=hsm
     env:
       - CGO_ENABLED=1
+      - >-
+        {{- if eq .Arch "amd64"}}CC=x86_64-linux-gnu-gcc{{- end}}
```
satoqz, line 24: This is just `/usr/bin/gcc` on `x86_64`, but more consistent and makes builds work the other way around, too (on an `aarch64` host that has the `x86_64` cross-compiler installed)

JanMa (approved):
LGTM, thanks @satoqz

cipherboy:
Thank you @satoqz! Do you want this backported to v2.3.x?

satoqz:
> Do you want this backported to v2.3.x?

Yeah, I don't see why not

PR #1432, opened 2025-06-11
Fix identity store resolution

cipherboy:
This PR does three things:

1. Fix a bug wherein identities weren't being loaded for non-root namespaces.
2. Attach namespace accessors to automatically generated identifiers in the identity subsystem, similar to tokens. This will provide the basis for any safe group inclusion logic in the future.
3. Add an unsafe, default off, config flag for the identity system to behave like Vault Enterprise and allow weak multi-tenancy in the identity system. All that is needed to do to enable this is to use only a single MemDB instance, rather than attempting to look up entities across MemDBs. 

Resolves: #1110

### DIFF
--- a/helper/namespace/namespace.go
+++ b/helper/namespace/namespace.go
@@ -14,6 +14,7 @@ import (
 	"unicode"
 	"unicode/utf8"
 
+	"github.com/hashicorp/go-uuid"
 	"github.com/openbao/openbao/sdk/v2/helper/consts"
 )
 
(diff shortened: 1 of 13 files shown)


helper/namespace/namespace.go
```diff
@@ -167,6 +168,33 @@ func (n *Namespace) Clone(withUnlock bool) *Namespace {
 	return data
 }
 
+func (n *Namespace) GenerateUUID() (string, error) {
```
wslabosz-reply, line 179: nit: could we have a comment, what uuid is it generating, what shape?

cipherboy: @wslabosz-reply Good idea, added.

cipherboy (commented):
@openbao/openbao-core-committers This should now be ready for review. I fixed a bug with missed identity loading for non-root namespaces and also added a test for this cross-namespace behavior. 

@luke-clifton do you mind taking a look as well, and seeing if this resolves the issues you saw? I believe it should, but want to verify :-)

wslabosz-reply (approved):
with my limited knowledge of the identity subsystem, haven't found anything invalid

luke-clifton:
I just substituted Vault for OpenBao built from this branch in one of our test suites which notionally covers this use case, and it worked!

I'll do some further investigation to make sure all the corner cases are covered on Monday, but it's looking like it works.

Edit: so I remember to check:

- [ ] Are entity and alias mappings locked to the originating namespace. In enterprise vault, it's an error to associate an alias with an entity if they are not from the same namespace.

driif:
@cipherboy @luke-clifton 
I also tested this with the config `unsafe_cross_namespace_identity = true`, and we were able to add a namespace entity to a root group and vice versa.

- [x]  Are entity and alias mappings locked to the originating namespace. In enterprise vault, it's an error to associate an alias with an entity if they are not from the same namespace.

I tested this as well, and alias mappings are locked to the originating namespace.

cipherboy:
@luke-clifton @driif Thank you both for the testing! I'm going to go ahead and merge on Andrii's approval but feel free to validate the behavior Luke and if there's more bugs we should have a week or more to get them sorted for GA.

luke-clifton:
I tested a few other scenarios this morning and it all looked good.

PR #1433, opened 2025-06-11
Add inline authentication

cipherboy:
This implements inline authentication, a form of writeless authentication that does not result in a token returned to the caller. We introduce four new headers:

 - `X-Vault-Inline-Auth-Path`, the path to authenticate against; required.
 - `X-Vault-Inline-Auth-Operation`, the operation to use; optional, defaults to UPDATE.
 - `X-Vault-Inline-Auth-Namespace`, the namespace to authenticate against; optional, defaults to the X-Vault-Namespace value if specified.
 - `X-Vault-Inline-Auth-Parameter-*`, user-specified parameters taking a URL-Safe Base64 encoded JSON object having a `key` and `value` parameter to inject into the request data.

In the flow of the request, this results in an initial call to the specified authentication endpoint; when successful, it returns a token which is used for the main request these headers are attached to, removing these headers from the subsequent operation. This token is not persisted and any leases created during the subsequent operation are revoked immediately and an error is returned.

This allows for greater scalability: by moving ephemeral interactions which cannot persist OpenBao auth tokens, such as CI/CD pipelines, to inline authentication, they have less resource usage on the instance. Additionally, as future read-only replicas cannot handle authentication (due to it incurring several writes), we get more value from these replicas if workloads use inline authentication unless they incur leases.

---

Resolves: #1301

### DIFF
--- a/website/content/docs/auth/index.mdx
+++ b/website/content/docs/auth/index.mdx
@@ -55,3 +55,57 @@ However, any existing token remain valid for the original grant period unless
 they are explicitly revoked within OpenBao. Operators should set appropriate
 [token TTLs](/docs/concepts/tokens#the-general-case) when using external
 authN methods.
+
+## Authentication Flows
+
+There are two primary flows client applications use to talk to OpenBao.
+
+### Standard Authentication
+
+In this flow, client applications first make calls to specific login
+endpoints, specific to the desired auth engine. These operations return an
+auth token which are subsequently passed in the `X-Vault-Token` header on
+following authenticated requests. This token has an expiration and may be
+renewable. Any leases created by the subsequent authenticated requests are
+associated with this token and are usually revoked when the token expires or
+is revoked.
+
+### Inline Authentication
+
+In contrast to [standard authentication](#standard-authentication), inline
+authentication is sent with the main, authenticated request. No ahead-of-time
+token acquisition occurs; an ephemeral token is created but not persisted to
+storage so authenticated operations which incur leases will not work and will
+result in an error (with any lease created being immediately revoked). This
+token will not be returned to the client application.
+
+To use, specify the following headers:
+
+ - `X-Vault-Inline-Auth-Path`, the path to authenticate against; required.
+ - `X-Vault-Inline-Auth-Operation`, the operation to use; optional,
+   defaults to `update`.
+ - `X-Vault-Inline-Auth-Namespace`, the namespace to authenticate against;
+   optional, defaults to the `X-Vault-Namespace` value if specified.
+ - `X-Vault-Inline-Auth-Parameter-*`, user-specified parameters taking a
+   URL-Safe Base64 encoded JSON object having a `key` and `value`
+   parameter to inject into the request data.
+
+
+:::warning
+
+Login calls made with this authentication mechanism must be one-shot: MFA
+(without the use of `X-Vault-MFA`) and other multi-step auth schemes such as
+certain OIDC flows will not work.
+
+:::
+
+If an auth method imposes a number of uses on source identity information,
+such as AppRole's `secret_id_num_uses`, each request with inline
+authentication will count against those limits separately.
+
+This improves scalability: standard authentication incurs several storage
+writes as a result of successful authentication and thus must happen on the
+active node. Additionally, these persisted entries are refreshed on leadership
+changes, which may cause churn and excessive disk and CPU usage. Depending on
+the workload patterns, a client application using inline authentication may
+result in a more performant OpenBao instance than standard authentication.
--- a/website/content/docs/concepts/auth.mdx
+++ b/website/content/docs/concepts/auth.mdx
@@ -64,6 +64,9 @@ revoking tokens, and renewing tokens. This is all covered on the
 
 ## Authenticating
 
+For more information on authenticating and the various forms of
+authentication, see the [authentication landing page](/docs/auth).
+
 ### Via the CLI
 
 To authenticate with the CLI, `bao login` is used. This supports many


vault/login_mfa.go
```diff
@@ -796,7 +796,7 @@ func (b *LoginMFABackend) handleMFALoginValidate(ctx context.Context, req *logic
 	}
 
 	// MFA validation has passed. Let's generate the token
-	resp, err := b.Core.LoginMFACreateToken(ctx, cachedResponseAuth.RequestPath, cachedResponseAuth.CachedAuth, req.Data)
+	resp, err := b.Core.LoginMFACreateToken(ctx, cachedResponseAuth.RequestPath, cachedResponseAuth.CachedAuth, req.Data, req.IsInlineAuth)
```
cipherboy: Besides being a typo, (condition should be negated), I think this also means that login MFA would not work with inline authentication, because Login MFA returns a token for subsequent use to provide the second factor. 

This is likely covered by the "one-shot auth" nature of inline auth (in #1302), but is worth calling out in documentation as an example of a form of multi-step auth.

cipherboy: Ah, that's incorrect. We can use the `X-Vault-MFA` parameter to occasionally do one-shot MFA. I've updated the code to err appropriately.

website/content/docs/auth/index.mdx
```diff
@@ -55,3 +55,57 @@ However, any existing token remain valid for the original grant period unless
...
+
+In this flow, client applications first make calls to specific login
+endpoints, specific to the desired auth engine. These operations return an
+auth token which are subsequently passed in the `X-Vault-Token` header on
+following authenticated requests. This token has an expiration and may be
+renewable. Any leases created by the subsequent authenticated requests are
+associated with this token and are usually revoked when the token expires or
+is revoked.
+
+### Inline Authentication
```
cipherboy, line 73: @fcbajao Let me know what you think of the documentation or if there's more places you'd like to see this documented.

https://patch-diff.githubusercontent.com/raw/openbao/openbao/pull/1433.patch should be a patch form if you want to experiment with it before it lands in a release. :-)

api/client.go
```diff
@@ -60,6 +60,30 @@ const (
...
+	// performed must be single-request.
+	InlineAuthPathHeaderName = "X-Vault-Inline-Auth-Path"
+
+	// Request operation to perform inline authentication with. Defaults to
+	// update.
+	InlineAuthOperationHeaderName = "X-Vault-Inline-Auth-Operation"
+
+	// Namespace to perform inline authentication with. Defaults to
+	// the value of X-Vault-Namespace; can be combined with a namespace
+	// in the path itself.
```
satoqz: Clarity:

```suggestion
	// Namespace to perform inline authentication with. Defaults to
	// the value of X-Vault-Namespace; can be combined with a namespace
	// in X-Vault-Inline-Auth-Path itself.
```

sdk/helper/consts/consts.go
```diff
@@ -43,4 +43,28 @@ const (
...
+	// { "key": <name>, "value": <value> }
+	//
+	// so that typing of the value and case sensitivity of the key can be
+	// preserved. The remainder of the header value (after the trailing
+	// dash) is ignored. Any repeated header keys result in request failure.
+	InlineAuthParameterHeaderPrefix = "X-Vault-Inline-Auth-Parameter-"
+
+	// Namespace to perform inline authentication with. Defaults to
+	// the value of X-Vault-Namespace; can be combined with a namespace
+	// in the path itself.
```
satoqz: Clarity:

```suggestion
	// Namespace to perform inline authentication with. Defaults to
	// the value of X-Vault-Namespace; can be combined with a namespace
	// in X-Vault-Inline-Auth-Path itself.
```

vault/request_handling.go
```diff
@@ -608,12 +618,148 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
 }
 
+func (c *Core) handleInlineAuth(ctx context.Context, req *logical.Request, nsHeader string) error {
+	// Find the path of the request.
+	authPath, present := req.Headers[consts.InlineAuthPathHeaderName]
+	if !present {
+		return nil
+	}
+	if len(authPath) == 0 || len(authPath) > 1 {
+		return fmt.Errorf("expected only one value for %v", consts.InlineAuthPathHeaderName)
```
```suggestion
	if len(authPath) != 1 {
		return fmt.Errorf("expected exactly one value for %v", consts.InlineAuthPathHeaderName)
```

vault/request_handling.go
```diff
@@ -608,12 +618,148 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
+		IsInlineAuth: true,
+	}
+
+	// Find the optional operation; this defaults to Update if missing.
+	authOperation, present := req.Headers[consts.InlineAuthOperationHeaderName]
+	if !present {
+		authOperation = []string{logical.UpdateOperation}
+	}
+	if len(authOperation) == 0 || len(authOperation) > 1 {
+		return fmt.Errorf("expected only one value for %v", consts.InlineAuthOperationHeaderName)
```
```suggestion
	if len(authOperation) != 1 {
		return fmt.Errorf("expected exactly one value for %v", consts.InlineAuthOperationHeaderName)
```

vault/request_handling.go
```diff
@@ -608,12 +618,148 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
+			return fmt.Errorf("inline auth namespace was not found")
+		}
+	} else {
+		// No header specified on request so assume relative to root namespace.
+		authNs, authReq.Path = c.namespaceStore.ResolveNamespaceFromRequest("", authReq.Path)
+		if authNs == nil {
+			return fmt.Errorf("inline auth namespace was not found")
+		}
+	}
+	authCtx = namespace.ContextWithNamespace(authCtx, authNs)
```
satoqz: Tried to clean this up a bit:
```suggestion
	switch {
	case !present:
		// Default to X-Vault-Namespace:
		authNamespace = []string{nsHeader}
	case len(authNamespace) == 0:
		// No header specified on request so assume relative to root namespace:
		authNamespace = []string{""}
	default:
		return fmt.Errorf("expected only one value for %v", consts.InlineAuthNamespaceHeaderName)
	}

	var authNs *namespace.Namespace
	authNs, authReq.Path = c.namespaceStore.ResolveNamespaceFromRequest(authNamespace[0], authReq.Path)
	if authNs == nil {
		return fmt.Errorf("inline auth namespace was not found")
	}
	authCtx := namespace.ContextWithNamespace(ctx, authNs)
```

vault/request_handling.go
```diff
@@ -608,12 +618,148 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
+	// parameters: a token of some sort and a role. This becomes our request
+	// data.
+	loginParams := make(map[string]interface{}, 2)
+	for header, values := range req.Headers {
+		if !strings.HasPrefix(header, consts.InlineAuthParameterHeaderPrefix) {
+			continue
+		}
+
+		if len(values) == 0 || len(values) > 1 {
+			return fmt.Errorf("expected only one value for each auth header parameter")
```
```suggestion
		if len(values) != 1 {
			return fmt.Errorf("expected exactly one value for each auth header parameter")
```

vault/request_handling.go
```diff
@@ -1121,6 +1267,21 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
...
+			if req.HasInlineAuth {
+				// When we've performed inline authentication and see a lease created,
+				// we must revoke this lease even though it isn't yet persisted. Throw
+				// an error.
+				revokeLease, err := c.router.Route(ctx, logical.RevokeRequest(req.Path, resp.Secret, resp.Data))
+				if err != nil {
+					return nil, nil, fmt.Errorf("failed to revoke ephemeral lease generated during inline authentication: %w", err)
+				}
+				if revokeLease != nil && revokeLease.IsError() {
+					return nil, nil, fmt.Errorf("failed to revoke ephemeral lease generated during inline authentication: %v", revokeLease.Error())
```
```suggestion
					return nil, nil, fmt.Errorf("failed to revoke ephemeral lease generated during inline authentication: %w", revokeLease.Error())
```

vault/request_handling.go
```diff
@@ -1536,6 +1697,10 @@ func (c *Core) handleLoginRequest(ctx context.Context, req *logical.Request) (re
 					}
 				}
 			} else if len(matchedMfaEnforcementList) > 0 && len(req.MFACreds) == 0 {
+				if req.IsInlineAuth {
+					return nil, nil, fmt.Errorf("unable to perform inline authentication with login MFA; use the X-Vault-MFA header to specify MFA information on the inline auth request")
```
satoqz, line 1712: `"X-Vault-MFA"` doesn't really have a convenient constant anywhere, should we add one?

vault/request_handling.go
```diff
@@ -1725,6 +1890,10 @@ func (c *Core) LoginCreateToken(ctx context.Context, ns *namespace.Namespace, re
 		},
 	)
 
+	if !persistToken {
+		resp.InlineAuthTokenEntry = te
+	}
```
satoqz: Mhh, while for now inline auth is the only effect driving `persistToken = false`, I don't like the implication here that it's because of inline auth.

satoqz (commented):
Finally got to this!

vault/request_handling.go
cipherboy: I had to rework the logic a bit, but hopefully this will be better :-)

vault/request_handling.go
cipherboy: Ah, hadn't realized this wasn't a string. Good catch :-)

vault/request_handling.go
cipherboy: @satoqz Hmm, yeah, I've renamed the variable to `isInlineAuth` and updated the logic accordingly.

vault/request_handling.go
cipherboy: There's a lot of constants missing... Just looking at API:

```go
    if len(r.WrapTTL) != 0 {
        req.Header.Set("X-Vault-Wrap-TTL", r.WrapTTL)
    }

    if len(r.MFAHeaderVals) != 0 {
        for _, mfaHeaderVal := range r.MFAHeaderVals {
            req.Header.Add("X-Vault-MFA", mfaHeaderVal)
        }
    } 
        
    if r.PolicyOverride {
        req.Header.Set("X-Vault-Policy-Override", "true")
    }
```

Let's file an issue for this and address separately? Seems like it'd be a good first issue for someone looking to improve the API/SDK libraries :-)

cipherboy:
Thanks @satoqz! Updated! :-)

cipherboy:
@satoqz And it actually compiles now lol :D

cipherboy:
Thank you @satoqz!

PR #1434, opened 2025-06-12
Bump to Go 1.24.4

wslabosz-reply:
Upgrades go version from `1.23.3` to `1.23.4`

Addresses following vulnerabilities:
- https://pkg.go.dev/vuln/GO-2025-3751
- https://pkg.go.dev/vuln/GO-2025-3750
- https://pkg.go.dev/vuln/GO-2025-3749

---
cc: @satoqz @cipherboy @JanMa


satoqz (approved):

cipherboy:
@wslabosz-reply Do you want to run the makefile target to go update the `golang.org/x/*` packages? I forget what it is called off the top of my head.

cipherboy:
```
path_login_test.go:105: Could not start local MSSQL docker container: API error (500): toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```

😢

wslabosz-reply:
> ```
> path_login_test.go:105: Could not start local MSSQL docker container: API error (500): toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
> ```
> 
> 😢

that's not result of this change... right?

cipherboy:
> that's not result of this change... right?

Right, just another thing we have to check on and see if we can run through a proxy or otherwise handle.

satoqz:
> > that's not result of this change... right?
> 
> Right, just another thing we have to check on and see if we can run through a proxy or otherwise handle.

Could look into caching the images we use in tests in GitHub actions cache, they don't need to update much after all. Does depend on how big the images are though, as max size of all cache items per repository combined seems to be capped at 10GB.

wslabosz-reply:
> > > that's not result of this change... right?
> > 
> > 
> > Right, just another thing we have to check on and see if we can run through a proxy or otherwise handle.
> 
> Could look into caching the images we use in tests in GitHub actions cache, they don't need to update much after all. Does depend on how big the images are though, as max size of all cache items per repository combined seems to be capped at 10GB.

doesn't matter as the gh will automatically delete the oldest entries?
we could also add a cleanup of artifacts created out of branches on the merge

satoqz:
> doesn't matter as the gh will automatically delete the oldest entries?

I was thinking what if the total size of images needed for one CI run exceeds 10GB, we would never have a full cache hit - reduced amount of pulls yes.

PR #1437, opened 2025-06-12
Adjust go.*, changelog codeowners

cipherboy:
Anyone should be able to approve changelog/ entries, and core should also own our go.{mod,sum} and .go-version files.


wslabosz-reply (approved):

cipherboy:
Thanks @JanMa and @wslabosz-reply!

PR #1438, opened 2025-06-12
chore(tls): refactor dev certificates to openbao

pree:
Refactors the dev certificates. Additionally I saw that `-dev-tls-san` is not implemented yet.
This is required for the current implementation of the `openbao-k8s` integration tests.

For now I will remove the `-dev-tls-san` from the integration tests.

### DIFF
--- a/command/server/config.go
+++ b/command/server/config.go
@@ -28,9 +28,9 @@ import (
 )
 
 const (
-	VaultDevCAFilename   = "vault-ca.pem"
-	VaultDevCertFilename = "vault-cert.pem"
-	VaultDevKeyFilename  = "vault-key.pem"
+	BaoDevCAFilename   = "openbao-ca.pem"
+	BaoDevCertFilename = "openbao-cert.pem"
+	BaoDevKeyFilename  = "openbao-key.pem"
 )
 
 var entConfigValidate = func(_ *Config, _ string) []configutil.ConfigError {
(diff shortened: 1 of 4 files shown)


command/server/config.go
```diff
@@ -28,9 +28,9 @@ import (
 )
 
 const (
-	VaultDevCAFilename   = "vault-ca.pem"
```
cipherboy, line 31: I realized we called this out as a change, but could we adjust the changelog message to be more precise? I guess strictly we only have an _API_ compatibility promise with Vault 1.14, but it seems like it'd break compatibility with previous OpenBao releases.

Thoughts? @pree

pree: This is only related to the dev server using `-dev-tls`. 
I don't think it breaks compatibility other than with unit test or the odd script for a dev env.

I just noticed that I also missed https://github.com/openbao/openbao/blob/c038a659e784162db590c37522a0a2fb3d3a8e60/command/server.go#L895

To clarify this is what changed.
openbao 2.2.2:
```
$ bao server -dev -dev-tls
    $ export BAO_CACERT='/tmp/vault-tls1140775264/vault-ca.pem'

$ ls /tmp/vault-tls1140775264 
vault-ca.pem  vault-cert.pem  vault-key.pem
```

openbao main:
```
$ bao server -dev -dev-tls
    $ export BAO_CACERT='/tmp/vault-tls199555210/openbao-ca.pem'

$ ls /tmp/vault-tls199555210/        
openbao-ca.pem  openbao-cert.pem  openbao-key.pem
```

pree: If you prefer this to be left unchanged we can revert this commit and change it in the `openbao-k8s` integration tests. I just thought we would like to have this changed to be prefixed with `openbao`.

cipherboy: @pree Understood, if memory serves, I helped to review the feature when it landed in Vault. ;-)

> other than with unit test or the odd script for a dev env.

It does change the behavior of a `bao server` command, which should ideally be fairly stable. I'd like it if, for the most part, we had full compatibility if someone were to `alias bao=vault`, we'd give the same output for especially this command. 

That's why we put care into the environment variable compatibility (#138, thanks @Mab879!).

[A quick search](https://github.com/search?q=vault-ca.pem+dev-tls&type=code) seems to suggest there's a few legitimate usages that would be broken:

 - https://github.com/escalopa/vault-playground/blob/e8afdd013304bd135201dd9dcbbe77c3f0aeff96/Makefile#L24 (admittedly a dev script as mentioned)
 - https://github.com/KokoiRuby/service-identity-provider/blob/9b687d60040afd4550dd08ff86190e67c7e27ead/vault/tutorials/foundations.md?plain=1#L436-L438 (tutorials will also be broken)
- https://github.com/ecadlabs/signatory/blob/659e44f099ff9003eb1bf764755dfd52b640cad1/integration_test/docker-compose.yml#L131 (integration tests)

So I do think reverting this might be better, and we just live with a few random `vault` references.

pree: Understood, haven't thought about aliasing to vault. I agree that we should stay compatible this way. Then lets revert this and I'll change the reference in `openbao-k8s` to use `vault-ca` :)

cipherboy: Thanks, @pree!

PR #1442, opened 2025-06-13
Transactions expiration mgr

suprjinx:
Relates #607

### DIFF
--- a/changelog/1442.txt
+++ b/changelog/1442.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+**ExpirationManager transaction support**: Added datastore transaction support to the expiration manager.
+```


vault/expiration.go
```diff
@@ -924,30 +933,110 @@ func (m *ExpirationManager) LazyRevoke(ctx context.Context, leaseID string) erro
...
 
+// WithTransaction is used to run a function with a transaction for the lease and token index views.
+func (m *ExpirationManager) WithTransaction(
+	ctx context.Context,
+	callbackFn func(context.Context) error) error {
+
+	var lvTx *barrierViewTransaction
+	var tivTx *barrierViewTransaction
+
+	if m.transactional {
```
cipherboy: Doesn't this mean that we can only have a single thread operating at a time? 🤔

vault/expiration.go
```diff
@@ -924,30 +933,110 @@ func (m *ExpirationManager) LazyRevoke(ctx context.Context, leaseID string) erro
...
+			return bv
+		}
+
+		m.tokenIndexView = func(ns *namespace.Namespace) BarrierView {
+			if tivTx != nil {
+				return tivTx
+			}
+			bv := originalStorageView(ns)
+			if bvtx, ok := bv.(*transactionalBarrierView); ok {
+				tivTx, err := bvtx.BeginTx(ctx)
```
cipherboy: Additionally, by creating a write transaction here, we place size limits on the amount of operations. 

I guess, if I were to suggest transactions, I'd use a read transaction to do paginated listing on startup and then use write transactions to handle removal of all storage with individual entries, rather than across multiple entries. My 2c.

vault/expiration.go
```diff
@@ -924,30 +933,110 @@ func (m *ExpirationManager) LazyRevoke(ctx context.Context, leaseID string) erro
...
-	le, err := m.loadEntry(ctx, leaseID)
-	if err != nil {
-		return err
-	}
-
-	// If there is no entry, nothing to revoke
-	if le == nil {
-		return nil
-	}
+	err := m.WithTransaction(ctx, func(ctx context.Context) error {
```
cipherboy, line 936: Ah I see, it is scoped to individual operations. This will definitely race then I think, which is not ideal. But I like the thought... I wonder if we can instead update all of these to take explicit storage references? 🤔

suprjinx: the storage is nested pretty deep in various interfaces, so it's not straightforward. What do you think of WithTransaction placing a storage transaction in `ctx` which can then be accessed and preferred where present much further down the call stack? Context should be unique per thread, if the manager itself is a shared instance.

suprjinx: Adjusted this to use `context` for conveying transaction -- let me know what you think

vault/expiration.go
```diff
@@ -447,14 +447,49 @@ func (c *Core) stopExpiration() error {
...
+	if err != nil {
+		m.logger.Error("failed to get namespace from lease ID", "error", err, "lease_id", leaseID)
+		return nil
+	}
+
+	return m.leaseView(ctx, ns)
+}
+
+func (m *ExpirationManager) leaseView(ctx context.Context, ns *namespace.Namespace) BarrierView {
+	if tx, err := logical.TransactionFromContext(ctx); err == nil && tx != nil {
```
suprjinx, line 453: this is where Tx is fetched from context if present

PR #1447, opened 2025-06-16
Add CEL concept and pki docs

fatima2003:
Related #794


fatima2003:
I tried to add @suprjinx as a reviewer but the username wouldn't show up, any thoughts/suggestions are welcome @suprjinx :))

suprjinx:
> I tried to add @suprjinx as a reviewer but the username wouldn't show up, any thoughts/suggestions are welcome @suprjinx :))

Thank you, I think i have to be approved as committer? But I can make comments :-)

website/content/docs/concepts/cel.mdx
```diff
@@ -0,0 +1,133 @@
...
+- The main expression in the `CelProgram` should return the engine's
+  specific output object on success, for instance:
+    * PKI → `ValidationOutput`
+    * JWT/OIDC → `pb.Auth`
+
+- A custom `string` error message or `bool` should be returned on
+  failure. The error message can be constructed using a CEL expression.
+  For instance: 
+  ```go
+  !has(request.common_name) ? 'request should have a common_name' : 'request rejected'
```
suprjinx: does this example fail in both cases? might be better to only fail when missing

website/content/docs/concepts/cel.mdx
```diff
@@ -0,0 +1,133 @@
...
+---
+
+## Main expression
+- This is the main expression that determines whether a CEL program has accepted or rejected
+a request.
+
+- The main expression in the `CelProgram` should return the engine's
+  specific output object on success, for instance:
+    * PKI → `ValidationOutput`
+    * JWT/OIDC → `pb.Auth`
```
suprjinx, line 34: Should we move `ValidationOutput` into the `pb` package so these can be more similar?

website/content/docs/concepts/cel.mdx
fatima2003: Ahh it looks ambiguous, but I wanted to show that the error message can be different depending on the reason for failure.. For instance in the case that we require `require_cn` and `validate_ttl` for success, the error message can be different depending on what failed first. So: 

`require_cn ? 'request should have a common_name' : (validate_ttl ? 'request has invalid TTL' : "Request Rejected")`

suprjinx:
Looks great, just a couple of comments

website/content/docs/concepts/cel.mdx
suprjinx: perhaps this should also have name like `PKIValidation`?

fatima2003: Sorry for getting back a bit late, thank you for the suggestions!

> Should we move `ValidationOutput` into the `pb` package so these can be more similar?

I think moving to the `pb` package is a good idea and other CEL protos can live here in the future as well. Perhaps creating folders according to the engine would be a good idea? 

> perhaps this should also have name like `PKIValidation`?

I agree, I think we would have to change up the name if we refactor to avoid confusion :))

cipherboy: @suprjinx @fatima2003 What about moving the X509 Certificate to `sdk/plugins/pb` as this might be of use to a GRPC-connected Cert Auth plugin as well in the future (in addition to CEL for Cert Auth)... The `ValidationOutput` I think is fine staying in the PKI package, as it'll never be used by any other plugin. The K/V plugin is an example of a plugin with local protobuf files.

cipherboy (approved):
Thanks @fatima2003!

website/content/docs/concepts/cel.mdx
cipherboy: Also, fwiw, since this shipped in v2.3.0, I wouldn't necessarily affect typing now if we can avoid it.

PR #1436, opened 2025-06-12
Move to minio openldap container

cipherboy:
See also: https://github.com/openbao/openbao/pull/1434#issuecomment-2966650712

PR #1449, opened 2025-06-16
Fix LDAP login test timing out on connection

satoqz:
An attempt to fix https://github.com/openbao/openbao/pull/1436#issuecomment-2968254702.

### DIFF
--- a/builtin/credential/kerberos/path_login_test.go
+++ b/builtin/credential/kerberos/path_login_test.go
@@ -6,10 +6,8 @@ package kerberos
 import (
 	"context"
 	"fmt"
-	"net"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/go-ldap/ldap/v3"
 	"github.com/openbao/openbao/sdk/v2/logical"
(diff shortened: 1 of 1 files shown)


satoqz:
Hmm, doesn't seem to fix it. Unfortunate, I'd really like to get this resolved as it flakes in CI more often than not ☹️

satoqz:
Ehrmmmm... `TestLogin` never even initiates a connection to the LDAP server. The login test never tests login. Ok.

The `quay.io/minio/openldap` container wants us to use `min.io` over `example.org`, otherwise it shuts down a few seconds after startup (But still starts a server for those few seconds, making it reachable within the test on fast machines).

😵‍💫

...so well, the container starts up fine now, but someone still needs to add a test case that uses it :D

wslabosz-reply (approved):
great find

cipherboy:
Wat

cipherboy (approved):

cipherboy:
Thanks @satoqz!

PR #1451, opened 2025-06-16
Add "namespace specifier" helpers

satoqz:
From #1320:

> The `external_keys` stanzas take an optional `namespaces = []` parameter, taking items of the form `uuid:<ns-uuid>`, `id:<ns-accessor>`, or `path:<ns-path>`, to identify namespaces that this library is allowed to be used in.

This adds utilities around these "namespace specifier" strings to easily parse them and compare them to an instance of `namespace.Namespace`.

We'll probably want to use them in various places in the future; the logic is quite encapsulated, so I'm extracting this out of the bigger feature.

### DIFF
--- a/helper/namespace/namespace.go
+++ b/helper/namespace/namespace.go
@@ -206,6 +206,37 @@ func (n *Namespace) ValidateUUID(candidate string) error {
 	return nil
 }
 
+// ParseSpecifier parses a "namespace specifier" kind and value from a
+// colon-separated string. Valid namespace specifier kinds are: path, id, uuid.
+func ParseSpecifier(value string) (string, string, error) {
+	kind, value, ok := strings.Cut(value, ":")
+	if !ok {
+		return "", "", fmt.Errorf("invalid namespace specifier")
+	}
+	switch kind {
+	case "path", "id", "uuid":
+	default:
+		return "", "", fmt.Errorf("unknown namespace specifier kind: %q", kind)
+	}
+	return kind, value, nil
+}
+
+// CompareSpecifier returns true if the namespace matches the passed specifier
+// value, comparing by Path, ID and UUID respectively depending on the passed
+// kind. Also see [ParseSpecifier].
+func (n *Namespace) CompareSpecifier(kind, value string) bool {
+	switch kind {
+	case "path":
+		return n.Path == Canonicalize(value)
+	case "id":
+		return n.ID == value
+	case "uuid":
+		return n.UUID == value
+	default:
+		return false
+	}
+}
+
 // ContextWithNamespace adds the given namespace to the given context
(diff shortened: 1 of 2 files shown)


wslabosz-reply (approved):
LGTM

satoqz:
Hmm, namespace paths may include `:` characters, which breaks the "trailing value" part. Do we want to change that?

helper/namespace/namespace.go
```diff
@@ -206,6 +206,48 @@ func (n *Namespace) ValidateUUID(candidate string) error {
 	return nil
 }
 
+// ParseSpecifier parses a "namespace specifier" into its kind, spec value
+// and an optional trailing rest.
+//
+// A namespace specifier has the following format:
+// <kind>:<spec>[:<rest>]
```
cipherboy: @satoqz I guess, what does `rest` do here? One alternative might be a URI-like scheme (`namespace://by-id/<ID>[?key=value]`).

satoqz: I was initially thinking this could be useful for what you proposed here: https://github.com/openbao/openbao/pull/1320#discussion_r2090924969, so we could prepend a namespace specifier to the key name in a common format. But based on further thoughts (noted down in my rambling in #1320) I'd prefer not to go that route altogether.

So I'd simplify this PR to omit the "rest" handling part.

cipherboy: @satoqz Ah, I see. You could still split left and right a single `:` and as long as the `rest` doesn't have `:` we'd be fine... Though I do kinda like the URI schemes for #1320 fwiw. Would it be `namespace://by-id/satoqz?key=my-named-key` ?

satoqz: I'd prefer to avoid absolute references to namespaces, it seems cumbersome for a child namespace to refer to its parent by id/uuid/path (more in https://github.com/openbao/openbao/pull/1320#discussion_r2150667846). The main use case I see for specifiers right now is in the server configuration for PKCS#11 library namespace whitelisting (where just `<kind>:<spec>` is enough).

A lot of this spills over into #1320 RFC discussion, maybe I was a bit too hasty here (but it also helped develop some thoughts :D). I'd love if you could have another pass at my recent comments there, so we can clarify the ultimate requirements.

satoqz:
Simplified this down to handle just `<kind>:<value>`, precisely what we need for https://github.com/cipherboy/openbao/blob/47468f9511184ac252ab62c141f04f9f7975efec/website/content/docs/rfcs/external-keys.mdx?plain=1#L50-L52.

wslabosz-reply (approved):

satoqz:
Thanks @wslabosz-reply @cipherboy, merging as first commit to `external-keys` :D

PR #1454, opened 2025-06-17
Add job which checks protobuf files

cipherboy:
This will help to ensure that when we update protoc, protobuf definitions, or other dependencies, we remember to regenerate the artifacts as well.


.github/workflows/code-checker.yml
```diff
@@ -100,3 +100,26 @@ jobs:
...
+  protobuf:
+    name: Protobuf checks
+    runs-on: ubuntu-latest
+    if: github.base_ref == 'main'
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: ./.github/actions/set-up-go
+      - run: |
+          pushd ~/.protoc
+          cd ~/.protoc
```
chicks-net, line 108: Isn't the `cd` redoing what the `pushd` already did for you?

chicks-net: It also isn't clear why you need to do the `pushd` and `popd` thing.  Maybe sticking with the `cd` and dropping the other two lines would be cleaner.

cipherboy: @chicks-net Ah, d'oh, good eyes. I changed the `mkdir` call to `pushd` instead lol. :facepalm:  I agree `cd` should be sufficient, I was wondering if it'd impact the subsequent command but I believe it shouldn't from quick glance at https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions

Fixed!

.github/workflows/code-checker.yml
```diff
@@ -100,3 +100,25 @@ jobs:
...
+          sha256sum -c <<< '96553041f1a91ea0efee963cb16f462f5985b4d65365f3907414c360044d8065  protoc-31.1-linux-x86_64.zip' || exit 1
+          unzip protoc-31.1-linux-x86_64.zip
+          sudo ln -s $PWD/bin/protoc /usr/local/bin
+          protoc --version
+        name: Install protobuf
+      - run: |
+          git diff --quiet || exit 1
+          make proto
+          git diff --quiet || (echo -e "\n\nModified files:" && git status --short && echo -e "\n\nRun 'make proto' locally and commit the changes.\n" && exit 1)
+        name: Run protobuf and check for differences
```
satoqz: I'd prefer

```yaml
- name:
  run:
```

over

```yaml
- run:
  name:
```

especially when `run` goes over several lines. But I'm just trying to find something to nit.

satoqz (approved):

.github/workflows/code-checker.yml
```diff
@@ -100,3 +100,25 @@ jobs:
...
+    name: Protobuf checks
+    runs-on: ubuntu-latest
+    if: github.base_ref == 'main'
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: ./.github/actions/set-up-go
+      - run: |
+          mkdir -p ~/.protoc
+          cd ~/.protoc
+          wget https://github.com/google/protobuf/releases/download/v31.1/protoc-31.1-linux-x86_64.zip
```
JanMa, line 109: Would be great if we could somehow keep this dependency up to date automatically with dependabot :thinking: 
There is a package in the ubuntu repos, but it seems way too old to me.

JanMa (approved):

.github/workflows/code-checker.yml
cipherboy: @satoqz Fixed consistently.

.github/workflows/code-checker.yml
cipherboy: @JanMa Yeah, I wish so too. And yes, it is sad the Ubuntu repo is out of date. I'd prefer not to rely on a third-party PPA.

cipherboy:
@openbao/openbao-core-committers Since protobuf files changed, could I get a re-review? Thanks!

satoqz (approved):

cipherboy:
Thanks! Merging...

PR #1457, opened 2025-06-17
Broaden bump-critical target to include more dependencies

cipherboy:
This adds more dependencies into `bump-critical` target, ensuring we get additional dependencies that we rely on in every release. This includes all of the missing cryptography-adjacent libraries and core storage libraries, plus CEL.


driif (approved):

cipherboy:
As this mostly adjusts dependencies, going ahead and merging based on reviews above. Thanks all!

PR #1461, opened 2025-06-17
change version to 2.3.0 in upcoming release

driif:
Change version to 2.3.0 in upcoming release.

### DIFF
--- a/version/version_base.go
+++ b/version/version_base.go
@@ -19,7 +19,7 @@ var (
 	CgoEnabled bool
 
 	// Filled by Goreleaser
-	fullVersion                   = "2.0.0-HEAD"
+	fullVersion                   = "2.3.0"
 	Version, VersionPrerelease, _ = strings.Cut(strings.TrimSpace(fullVersion), "-")
 	VersionMetadata               = ""
 )


version/version_base.go
```diff
@@ -19,7 +19,7 @@ var (
 	CgoEnabled bool
 
 	// Filled by Goreleaser
-	fullVersion                   = "2.0.0-HEAD"
+	fullVersion                   = "2.3.0"
```
cipherboy, line 22: This version number is unused: https://openbao.org/docs/contributing/packaging/

cipherboy: Additionally, `main` targets `v2.4.0` right now.

driif: @cipherboy UI uses it in the footer.

cipherboy: @driif Sorry, let me clarify. The version checked into source is unused in release versions. During build time, the value this variable takes on is replaced via an `-X github.com/openbao/openbao/version.fullVersion=...` option. 

See `podman run -p 8200:8200 quay.io/openbao/openbao:2.2.2 server -dev -dev-root-token-id=devroot`:

image

So this version is only used during builds from the filesystem and helps to serve as a reminder that production builds should follow the guide I linked above.

driif: @cipherboy Oh yes, great, thanks for clarifying, Alex. I'll close this PR.

PR #1463, opened 2025-06-17
Add nonce parameter to be used as salt by the underlying HKDF

DanGhita:
Add nonce parameter in the ECDH transit API  to be used as salt by the underlying HMAC Key Derivation Function.

Resolves #811

### DIFF
--- a/builtin/logical/transit/path_derive_key.go
+++ b/builtin/logical/transit/path_derive_key.go
@@ -5,6 +5,7 @@ package transit
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


builtin/logical/transit/path_derive_key.go
```diff
@@ -39,9 +40,16 @@ func (b *backend) pathDeriveKey() *framework.Path {
 Defaults to "ecdh".`,
 			},
 
+			"nonce": {
```
cipherboy, line 43: Hmm, it seems like we'd want a new parameter, `key_expansion_algorithm` (`=hkdf` by default). If we add other algorithms in the future (or support `none`), we'd have algorithm-specific parameters here, which I'm not sure how we'd want to handle. Thoughts?

DanGhita: Since this feature is probably not yet used in production,  an alternative would be to rename the current 'key_derivation_algorithm' from "ecdh" to "ecdh_hkdf".

That keeps the API simple. And the documentation  would explain the "nonce" is only for "*_hkdf" schemes. 

This "trick" is already used in Bao APIs (the "X" parameter is expected only for the "Y" option) . That is a less "academic" solution, but, once again, it keeps the API simple.

The most academic solution would be to have both  'key_agreement_algorithm' (ecdh) and 'key_derivation_algorithm' (hkdf), but it's probably overkill (not even sure the cloud KMS APIs are making this distinction !)

Furthermore, this  key-derivation API/endpoint is intended to manage  (at least theoretically) all kind of key derivations (for ex, concatenation of two symmetric keys for getting a HMAC key). In this (theoretical) example the parameter "peer_public_key" is meaningless and other specific parameters would be expected.
BTW, it might be already a confusion with the pre-existing API: key derivation operation by encrypting some data with a key could be implemented by this endpoint , but something similar already exists in Transit (see derived (bool: false)  in Create key).

Once again, my first option would be to keep the API simple and to do the change "ecdh" -> "ecdh_hkdf".

What do you think ?

PR #1464, opened 2025-06-17
Add changelog, Makefile to core committers

cipherboy:


cipherboy:
Thanks @JanMa! Merging since there's only CODEOWNERS file changes and GitHub has confirmed them to be valid and CI doesn't test this...

PR #1468, opened 2025-06-19
fix(test-go): Verify test results before passing

### LINKED ISSUE #1065: fix or remove the test-collect-reports CI step
In it's current configuration, the `test-collect-reports` CI step just checks if any of our test runs actually uploaded a valid test result. This has the effect that for example if our UI test uploads a result, but all the Go tests don't, the step succeeds nevertheless.

dc-tec:
## Fix test-collect-reports validation to verify all Go test matrix jobs

Implemented fix in the `test-collect-reports` step. It retrieves the expected matrix IDs from the `test-matrix` job output and validates each matrix job individually by checking for the presence of `results-{id}.*` files. If any expected Go test matrix job results are missing, the job fails.

Resolves #1065

### DIFF
--- a/.github/workflows/test-go.yml
+++ b/.github/workflows/test-go.yml
@@ -310,7 +310,34 @@ jobs:
           path: test-results/go-test
           pattern: test-results-*
           merge-multiple: true
-      - run: |
-          ls -lhR test-results/go-test
-          find test-results/go-test -type f -mindepth 1 -mtime +3 -delete
-          ls -lhR test-results/go-test
+      - env:
+          EXPECTED_IDS: ${{ needs.test-matrix.outputs.matrix_ids }}
+        run: |
+          echo "Expected matrix IDs: $EXPECTED_IDS"
+
+          # Parse expected IDs and check for corresponding result files
+          missing_results=0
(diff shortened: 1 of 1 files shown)


wslabosz-reply (approved):

wslabosz-reply:
test failure not related, flaky `TestOIDC_PeriodicFunc`.

.github/workflows/test-go.yml
```diff
@@ -310,7 +310,34 @@ jobs:
           path: test-results/go-test
           pattern: test-results-*
           merge-multiple: true
-      - run: |
-          ls -lhR test-results/go-test
-          find test-results/go-test -type f -mindepth 1 -mtime +3 -delete
-          ls -lhR test-results/go-test
+      - env:
+          EXPECTED_IDS: ${{ needs.test-matrix.outputs.matrix_ids }}
```
JanMa, line 314: This evaluates to an empty string because `test-matrix` is missing from the `needs` for this job. Interesting behavior, I would have expected this to fail :thinking:

dc-tec: Interesting indeed, it should have failed. Ill get it fixed.

JanMa (approved):
LGTM, thanks @dc-tec

PR #1474, opened 2025-06-23
Namespace sealing POC

wslabosz-reply:
POC for #1357 / #1170 

Currently implemented features:
- per namespace barrier, root key, keyring (data is encrypted only once)
- create sealable namespace (incl. support in cli command) using a seal config
    - currently only supports shamir
    - namespace is unsealed after creation for now, we could reseal after namespace initialization
- store seal config permanently, encrypted with closest parent barrier
- `/sys/raw` works for keys encrypted with non-root barriers

Missing in this POC (not exhaustive):
- loading/unloading of sealed namespaces from storage
- sealing/unsealing lifecycle functionality, api, cli command
- remembering seal shares when calling unseal api
- most other api actions (rekeying, recovery, generate-root, key rotation, seal status, ...)
- most should be straight forward to implement since we are reusing the existing AESGCMBarrier implementation
- handling requests towards sealed namespaces

Example for creating a sealable namespace:

```sh
bao namespace create -seals=seals.json ns1
```

with seals.json (hcl not supported yet):

```json
[
  {
    "type": "shamir",
    "secret_shares": 3,
    "secret_threshold": 2
  }
]
```

/cc @cipherboy, @driif, @voigt, @wslabosz-reply, @rencooo, @satoqz


vault/seal_manager.go
```diff
@@ -0,0 +1,77 @@
...
+// SealManager is used to provide storage for the seals.
+// It's a singleton that associates seals (configs) to the namespaces.
+type SealManager struct {
+	core    *Core
+	storage logical.Storage
+
+	// lock        sync.RWMutex
+	// invalidated atomic.Bool
+
+	sealsByNamespace map[string][]*Seal
```
satoqz: `Seal` is an interface that's internally already a pointer.

vault/barrier_aes_gcm.go
```diff
@@ -69,6 +69,8 @@ type AESGCMBarrier struct {
 	l      sync.RWMutex
 	sealed bool
 
+	metaPrefix string
```
cipherboy: I forget, why'd we go this route rather than a proper View?

cipherboy: I think that'd help some of the churn in `vault/barrier_aes_gcm_test.go` tbh, since a `physical.BarrierView` will be implemented by `physical.Barrier`. My 2c.

command/namespace_create.go
```diff
@@ -59,6 +64,14 @@ func (c *NamespaceCreateCommand) Flags() *FlagSets {
 			"This can be specified multiple times to add multiple pieces of metadata.",
 	})
 
+	f.StringVar(&StringVar{
+		Name:       "seals",
+		Target:     &c.flagSealsConfigPath,
+		Completion: complete.PredictFiles("*.json"),
+		Usage: "Path to a JSON file with at least one or several SealConfigs." +
+			"MUST be an JSON array.",
```
satoqz:
```suggestion
		Usage: "Path to a JSON file with a list of namespace seal configurations. " +
			"Must be a JSON array with at least one configuration.",
```

Above `custom-metadata` flag is also missing a space after the first sentence.

command/namespace_create.go
```diff
@@ -96,10 +109,20 @@ func (c *NamespaceCreateCommand) Run(args []string) int {
 		return 2
 	}
 
+	seals, err := c.parseSeals()
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error while parsing seal configs: %s", err))
```
satoqz:
```suggestion
		c.UI.Error(fmt.Sprintf("Error parsing seal configs: %s", err))
```

command/namespace_create.go
```diff
@@ -111,5 +134,59 @@ func (c *NamespaceCreateCommand) Run(args []string) int {
...
+	path := c.flagSealsConfigPath
+	if path == "" {
+		return nil, nil
+	}
+	var rawSeals []map[string]interface{}
+
+	absolutePath, err := filepath.Abs(path)
+	if err != nil {
+		return nil, err
+	}
```
satoqz: Is that needed?

command/namespace_create.go
```diff
@@ -111,5 +134,59 @@ func (c *NamespaceCreateCommand) Run(args []string) int {
...
+		// TODO Handle Output for multiple seals
+		keySharesMap := secret.Data["key_shares"].(map[string]interface{})
+		if keySharesMap != nil {
+			defaultKeyShares := keySharesMap["default"].([]interface{})
+			for i, key := range defaultKeyShares {
+				c.UI.Output(fmt.Sprintf("Unseal Key %d: %s", i+1, key))
+			}
+		}
+		secretShares := int(seals[0]["secret_shares"].(float64))
+		secretThreshold := int(seals[0]["secret_threshold"].(float64))
```
satoqz: You're trusting the server to validate that all of these fields do indeed exist. For example, `int(seals[0]["secret_shares"].(float64))` will panic if `"secret_shares"` is not in the map. You can import `vault.SealConfig` here and share parsing/validation logic and only send the seals to the server if they're ok locally. Probably also spares you the `float64` roundtrip.

vault/logical_system.go
```diff
@@ -5893,6 +5893,12 @@ This path responds to the following HTTP methods.
 
 	DELETE /<path>
 		Delete a namespace.
+
+	POST /<name>/seal
+		Seal a namespace.
+
+	POST /<name/unseal
+		Unseal a namespace.
```
satoqz:
```suggestion
	POST /<name>/seal
		Seal a namespace.

	POST /<name>/unseal
		Unseal a namespace.
```

We've been using `<path>` in the namespace backend (also for patterns). I see the appeal in using `<name>` where you can't specify a full namespace path but just an immediate child, but it would be nice to backport this to existing endpoints as well.

vault/logical_system_namespaces.go
```diff
@@ -238,14 +309,19 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
...
+	}
 	return map[string]any{
 		"uuid":            ns.UUID,
 		"path":            ns.Path,
 		"id":              ns.ID,
 		"tainted":         ns.Tainted,
 		"locked":          ns.Locked,
 		"custom_metadata": ns.CustomMetadata,
+		"key_shares":      keySharesPerSeal,
 	}
```
```suggestion
	ret := map[string]any{
		"uuid":            ns.UUID,
		"path":            ns.Path,
		"id":              ns.ID,
		"tainted":         ns.Tainted,
		"locked":          ns.Locked,
		"custom_metadata": ns.CustomMetadata,
	}
	if len(keySharesMap) != 0 {
		ret["key_shares"] = keySharesMap[0]
	}
	return ret
```

vault/logical_system_namespaces.go
```diff
@@ -339,15 +415,50 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
...
+		if new {
+			// TODO(wslabosz): write all the provided configs
+			if len(sealConfigs) > 0 {
+				if err := b.Core.sealManager.SetSeal(ctx, sealConfigs[0], entry); err != nil {
+					return handleError(err)
+				}
+
+				nsSealKeyShares, err := b.Core.sealManager.InitializeBarrier(ctx, entry)
+				if err != nil {
+					return logical.ErrorResponse(fmt.Sprintf("%s", err.Error())), err
```
```suggestion
					return handleError(err)
```

vault/logical_system_namespaces.go
```diff
@@ -339,15 +415,50 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
...
+
+		entry, new, err := b.Core.namespaceStore.ModifyNamespaceByPath(ctx, path, func(ctx context.Context, ns *namespace.Namespace) (*namespace.Namespace, error) {
 			ns.CustomMetadata = metadata
 			return ns, nil
 		})
 		if err != nil {
 			return handleError(err)
 		}
+		keySharesMap := make(map[string][]string)
+		if new {
```
satoqz: Nit: Could rearrange to early return

vault/namespace_store.go
```diff
@@ -226,66 +226,60 @@ func (ns *NamespaceStore) SetNamespace(ctx context.Context, namespace *namespace
...
-	return ns.setNamespaceLocked(ctx, namespace)
+	// setNamespaceLocked handles unlocking
+	new, err := ns.setNamespaceLocked(ctx, namespace)
+	if err != nil {
+		return err
+	}
+
+	// TODO: might want to support calling (*SealManager).SetSeal here
+	if new {
+		ns.initializeNamespace(ctx, namespace)
```
satoqz: Unhandled error

vault/namespace_store.go
```diff
@@ -557,20 +545,21 @@ func (ns *NamespaceStore) ModifyNamespaceByPath(ctx context.Context, path string
...
 		}
 
 		// ModifyNamespaceByPath can never modify lock status.
 		entry.UnlockKey = unlockKey
 	}
 
-	// setNamespaceLocked will unlock ns.lock
-	if err := ns.setNamespaceLocked(ctx, entry); err != nil {
-		return nil, err
+	// setNamespaceLocked handles unlocking
```
satoqz: It really doesn't need to anymore though. Could rearrange callers to do the unlocking now.

vault/seal_manager.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (sm *SealManager) StorageAccessForPath(path string) StorageAccess {
+	_, v, _ := sm.barrierByStoragePath.LongestPrefix(path)
+	if v == nil {
+		return &directStorageAccess{physical: sm.core.physical}
+	}
+	barrier := v.(SecurityBarrier)
+	return &secureStorageAccess{barrier: barrier}
+}
```
satoqz, line 116: Would be nice to have a comment somewhere (doesn't have to be on this function) that explains the entire `StorageAccess` and write-straight-to-physical edge case mechanism.

vault/seal_test.go
```diff
@@ -7,28 +7,31 @@ import (
...
-// of an API.  In other words if your change break this test, it might be more
+// of an API. In other words if your change break this test, it might be more
 // the test's fault than your changes.
 func TestDefaultSeal_Config(t *testing.T) {
 	bc := &SealConfig{
 		SecretShares:    4,
 		SecretThreshold: 2,
 	}
 	core, _, _ := TestCoreUnsealed(t)
+	ctx := namespace.ContextWithNamespace(context.Background(), namespace.RootNamespace)
```
satoqz:
```suggestion
	ctx := namespace.RootContext(nil)
```

vault/seal_test.go
```diff
@@ -39,11 +42,45 @@ func TestDefaultSeal_Config(t *testing.T) {
...
+		SecretShares:    10,
+		SecretThreshold: 5,
+	}
+
+	nsTest := &namespace.Namespace{Path: "test/"}
+	TestCoreCreateNamespaces(t, core, nsTest)
+
+	defSeal = NewDefaultSeal(nil)
+	defSeal.SetCore(core)
+	err = defSeal.SetBarrierConfig(ctx, nsSealConfig, nsTest)
```
satoqz: Only realized now that the type is `SealConfig` but all the functions and fields are "barrier config". If anything this PR is the opportunity to change that (to always be seal config)

vault/seal.go
```diff
@@ -54,14 +55,15 @@ const (
...
 	StoredKeysSupported() seal.StoredKeysSupport // SealAccess
 	SealWrapable() bool
 	SetStoredKeys(context.Context, [][]byte) error
 	GetStoredKeys(context.Context) ([][]byte, error)
-	BarrierType() wrapping.WrapperType                  // SealAccess
-	BarrierConfig(context.Context) (*SealConfig, error) // SealAccess
-	SetBarrierConfig(context.Context, *SealConfig) error
+	BarrierType() wrapping.WrapperType                                        // SealAccess
+	BarrierConfig(context.Context, *namespace.Namespace) (*SealConfig, error) // SealAccess
+	SetBarrierConfig(context.Context, *SealConfig, *namespace.Namespace) error
```
satoqz: Can we not get the namespace from context? This causes sooo many changed lines just to pass the root namespace in almost all cases.

satoqz (commented):
Nothing serious, you could fix all of these later. Looks good generally.

command/namespace_create.go
cipherboy: There has to be a better way of parsing this. Maybe this should be added to the API client somehow?

vault/init.go
```diff
@@ -250,7 +250,8 @@ func (c *Core) Initialize(ctx context.Context, initParams *InitParams) (*InitRes
 		return nil, fmt.Errorf("error initializing seal: %w", err)
 	}
 
-	barrierKey, barrierKeyShares, err := c.generateShares(barrierConfig)
+	// TODO: we are here
```
cipherboy: Maybe the TODO should clarify that these shares are only used in tests?

I also wonder if we can drop these tests, given that I think they were pre-Vault 1.0. We don't necessarily support earlier Vault versions and assume you come through v1.14 for migrations, but maybe they were only ever read compatible and not migrated.

vault/logical_system_namespaces.go
cipherboy: Yeah, I wonder what the most optimal data type would be here. I think the `map name->[]shares` might be best?

cipherboy: Or maybe namespaces should only ever be created with a single share and more added later? :thinking:

vault/seal.go
cipherboy: I like this idea FWIW. Worst case, you'll have to find the actual sealed parent namespace but that's about it.

vault/seal.go
```diff
@@ -76,9 +78,10 @@ type Seal interface {
 }
 
 type defaultSeal struct {
-	access seal.Access
-	config atomic.Value
-	core   *Core
+	access     seal.Access
+	config     atomic.Value
+	core       *Core
+	metaPrefix string
```
cipherboy, line 84: Not sure if I suggested this elsewhere, but not sure I like the whole `metaPrefix` thing again here, but that's fine.

cipherboy (commented):
Some thoughts on the proof of concept, but looking fairly good!

command/namespace_create.go
wslabosz-reply: I mean if we want to enable operators to provide a path to config files then yes

satoqz: Can't you just read the relative path if it is relative? I don't see why you need resolve it to an absolute one.

vault/seal.go
wslabosz-reply: that is completely valid, I even adjusted that in my poc multiple seals storage adjustment

vault/seal_test.go
wslabosz-reply: `SealConfig` might be a "barrier config" or a "recovery config", so no, this PR is unfortunately not a opportunity to remove that confusion

vault/init.go
wslabosz-reply: > Maybe the TODO should clarify that these shares are only used in tests?

this TODO is a leftover from a debug session

> I also wonder if we can drop these tests, given that I think they were pre-Vault 1.0. We don't necessarily support earlier Vault versions and assume you come through v1.14 for migrations, but maybe they were only ever read compatible and not migrated.

valid point, not the scope of this PR per say, but definitely worth a look

vault/seal_test.go
satoqz: The point I'm making is that it never configures the barrier, but the sealing mechanism. A "barrier config" does not exist.

wslabosz-reply: okay I get it, so what about
`SetSealConfig` and `SetRecoverySealConfig`
@cipherboy wdyt?

command/namespace_create.go
wslabosz-reply: so you are suggesting assuming it is relative, and attempt at reading so, then fallback to absolute?

vault/namespace_store.go
wslabosz-reply: will wait on adjustment after https://github.com/openbao/openbao/pull/1538

vault/seal_manager.go
wslabosz-reply: will add something

command/namespace_create.go
satoqz: No, just read whatever path is. `os.Open` should understand both types of paths.

vault/seal_test.go
cipherboy: Barrier to me refers to keyring specifically, and not to the root keys or seal mechanism that this is.

So I do think I agree the latter set of names are most clear (`SetSealConfig` and `SetRecoverySealConfig`).

vault/logical_system_namespaces.go
wslabosz-reply: > Or maybe namespaces should only ever be created with a single share and more added later?

would need to think more about it, could we add that to the RFC comments?

vault/seal_test.go
wslabosz-reply: I've simplified them even a bit more (removing the `Seal` as it is in the name of the interface and ended up with:
`Config`, `SetConfig`, `SetCachedConfig` (old barrierConfig) and
`RecoveryConfig`, `SetRecoveryConfig`, `SetCachedRecoveryConfig`

vault/seal.go
wslabosz-reply: this caused a ripple effect in tests and some other places, requiring a manual addition of root namespace to the context

vault/barrier_aes_gcm.go
wslabosz-reply: @cipherboy
I think it's doable to make it a `physical.View` instead, but I don't think there's a way to leave it with just that, as we need to be able to converse it into `Transactional` barrier, and that requires `physical.Backend`, so we still will be left with:

```
		backend:                  storage,
		view:                     physical.NewView(storage, metaPrefix),
```

> since a physical.BarrierView will be implemented by physical.Barrier

there are no `physical.BarrierView` nor `physical.Barrier` (did you mean `View` and `Backend`?)

cipherboy: Types are soft things that kinda sorta exist in my head but it isn't a type-checking compiler. :laughing:

---

Ahh, right. And we have no way to do a `TransactionalView`. Though, I suppose we could add it...

vault/logical_system_namespaces.go
cipherboy: @wslabosz-reply Yeah, feel free :-)

vault/namespace_store.go
wslabosz-reply: adjusted

vault/barrier_aes_gcm.go
wslabosz-reply: @cipherboy 
should I add it in the scope of this PoC?

cipherboy: Eh, you can add it as a separate PR if you want or we can drop the idea, either way. I suspect there are other places where this'd be useful, for example in @suprjinx's expiration manager transactional view handling, which I've not had time to give feedback on. :(

wslabosz-reply: added it in this PR, not sure if it's correct implementation (I've mirrored the `StorageView` approach)

phyrog: As mentioned in a previous version of this PR, this `metaPrefix` does not construct a view, it just points the barrier to where its config is stored. The barrier will still be able to write to all of the storage (child namespace unseal configs are encrypted with the (next higher sealable) parent namespace barrier).

phyrog: Not sure if I understood correctly, but you can't replace `backend` with a scoped view. The barriers need to be able to write outside of their namespaces path. A namespace's unseal config needs to be encrypted by the next higher sealable namespace barrier, so that barrier needs to be able to write into child namespace data.

What you possibly could do is replace the `metaPrefix` field with a view of just the storage paths that contain the barrier config, while still using the complete `backend` for writing.

vault/logical_system_namespaces.go
```diff
@@ -238,15 +291,21 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
...
 		"uuid":            ns.UUID,
 		"path":            ns.Path,
 		"id":              ns.ID,
 		"tainted":         ns.Tainted,
 		"locked":          ns.Locked,
 		"custom_metadata": ns.CustomMetadata,
 	}
+	if len(keySharesMap) > 0 {
+		ret["key_shares"] = make(map[string][]string, len(keySharesMap))
+		maps.Copy(ret["key_shares"].(map[string][]string), keySharesMap)
```
satoqz: I don't think you need to copy here? `maps.Copy` also won't deepcopy the arrays.

sdk/physical/physical_view.go
```diff
@@ -11,42 +11,80 @@ import (
...
+}
+
+type ViewTransaction interface {
+	Transaction
+	ViewCore
+}
+
+var (
+	_ Backend = &view{}
+	_ View    = &view{}
```
satoqz: `View` implies `Backend`

sdk/physical/physical_view.go
```diff
@@ -11,42 +11,80 @@ import (
...
+	}
+
+	if _, ok := backend.(Transaction); ok {
+		return &viewTransaction{
+			*v,
+		}
+	}
+
+	return v
 }
```
satoqz:
```suggestion
func NewView(backend Backend, prefix string) View {
	v := view{
		backend: backend,
		prefix:  prefix,
	}

	if _, ok := backend.(TransactionalBackend); ok {
		return &transactionalView{v}
	}

	if _, ok := backend.(Transaction); ok {
		return &viewTransaction{v}
	}

	return &v
}
```

vault/logical_system_namespaces.go
wslabosz-reply: right

cipherboy (approved):
Thank you @wslabosz-reply for your patience and the refactor :-) I think this looks fairly good for the first steps! Really like the use of view and the suggestions from @satoqz, I think it is cleaner.

satoqz (approved):

vault/barrier_aes_gcm.go
wslabosz-reply: @phyrog 
you were right from the beginning, I couldn't make it nicely with the scoped view of the storage, so I've reverted the change.

PR #1477, opened 2025-06-24
deps(go-jose): Update go-jose from v3 to v4

### LINKED ISSUE #1455: Move to github.com/go-jose/go-jose/v4
We are currently on v3 of go-jose; we should evaluate getting to /v4 as it is a relatively critical dependency and we don't want to miss any vulnerability fixes as a result of being on the last major version.

dc-tec:
Updated go-jose v3 to v4 throughout the codebase.

**This dependency was used in the following files:**

`Makefile`
`api/plugin_helpers.go`
`api/secret.go`
`audit/format.go`
`builtin/credential/jwt/path_cel_login_test.go`
`builtin/credential/jwt/path_cel_role.go`
`builtin/credential/jwt/path_login_test.go`
`builtin/credential/jwt/path_oidc_test.go`
`builtin/credential/jwt/path_role.go`
`builtin/credential/kubernetes/path_login.go`
`builtin/credential/kubernetes/path_login_test.go`
`builtin/logical/kubernetes/integrationtest/helpers.go`
`builtin/logical/kubernetes/path_creds.go`
`builtin/logical/pki/acme_jws.go`
`command/agent/testing.go`
`vault/identity_store_oidc.go`
`vault/identity_store_oidc_provider.go`
`vault/identity_store_oidc_provider_util.go`
`vault/identity_store_oidc_test.go`
`vault/wrapping.go`

**The following changes were needed:**

- The `ParseSigned()` function now expects a slice of supported algorithms to validate.
- Type `Builder` now calls `Serialize()` instead of `CompactSerialize()`

**Additional information:**
Commit [da824f0](https://github.com/openbao/openbao/commit/da824f002cde82fec0eacba6b3d9797197996e25) includes a fix that prevented some tests from running after running go mod tidy. We might want to split that out into a separate PR if needed

Resolves #1455


api/secret.go
```diff
@@ -372,7 +372,7 @@ func ParseSecret(r io.Reader) (*Secret, error) {
 			if err := json.Unmarshal(errBytes, &errStrArray); err != nil {
 				return nil, err
 			}
-			return nil, fmt.Errorf(strings.Join(errStrArray, " "))
+			return nil, errors.New(strings.Join(errStrArray, " "))
```
DrDaveD, line 375: Was this an accident?  It doesn't appear to have anything to do with the subject of the PR.

audit/format.go
```diff
@@ -536,7 +537,17 @@ func parseVaultTokenFromJWT(token string) *string {
...
-	parsedJWT, err := jwt.ParseSigned(token)
+	supportedAlgs := []jose.SignatureAlgorithm{
+		jose.ES256,
+		jose.ES384,
+		jose.ES512,
+		jose.RS256,
+		jose.RS384,
+		jose.RS512,
+		jose.EdDSA,
+	}
```
DrDaveD: It's rather a shame to have to hardcode these, in several places.  Is there not a better way?

At minimum I think we should have a single place for these to come from in openbao.

vault/wrapping.go
```diff
@@ -404,7 +404,7 @@ func (c *Core) validateWrappingToken(ctx context.Context, req *logical.Request)
 	// and then a dot.
 	if IsJWT(token) {
 		// Implement the jose library way
-		parsedJWT, err := jwt.ParseSigned(token)
+		parsedJWT, err := jwt.ParseSigned(token, []jose.SignatureAlgorithm{jose.ES512})
```
DrDaveD, line 407: Why is it ok to have only one supported algorithm here?

audit/format.go
dc-tec: I agree with that, I couldn't really find a correct place where we could define it. If you have suggestions please let me know!

api/secret.go
dc-tec: No, was not by accident. See the additional information in the PR description about this commit.

audit/format.go
JanMa: I suggest you add this to the SDK helpers package at https://github.com/openbao/openbao/tree/main/sdk/helper/consts

vault/wrapping.go
dc-tec: My assumption to only specify this algorithm was based on L217 in `wrapping.go` https://github.com/openbao/openbao/blob/12a1bb859c1db62bcb777921b2c9abae5b039217/vault/wrapping.go#L217

api/secret.go
dc-tec: I see that some other tests are now failing in CI with the same reason it failed when I ran the tests locally, which caused me to make this commit.

dc-tec:
Created a new constants package. We might want to tweak it, but should be much cleaner now.

sdk/helper/consts/jwt.go
```diff
@@ -0,0 +1,61 @@
...
+		jose.RS512,
+		jose.ES256,
+		jose.ES384,
+		jose.ES512,
+		jose.EdDSA,
+	}
+
+	// AllowedJWTSignatureAlgorithmsBao defines the JWT signature algorithms
+	// that are supported by OpenBao.
+	AllowedJWTSignatureAlgorithmsBao = []jose.SignatureAlgorithm{
```
DrDaveD: Can any of these constants be defined in terms of the others?  For example this one has the same list as the previous one.

vault/identity_store_oidc.go
```diff
@@ -1539,7 +1539,12 @@ func (i *IdentityStore) pathOIDCIntrospect(ctx context.Context, req *logical.Req
 	clientID := d.Get("client_id").(string)
 
 	// validate basic JWT structure
-	parsedJWT, err := jwt.ParseSigned(rawIDToken)
+	// Convert all supported algorithms to jose.SignatureAlgorithm types
+	supportedSigAlgs := make([]jose.SignatureAlgorithm, len(supportedAlgs))
```
DrDaveD: Can `supportedArgs` be updated to be based on the shared constants?

DrDaveD (dismissed):
Now I see (by searching for "RS256") that similar lists appear in other places in the codebase as well.  There should probably be another issue made to consolidate those using the new constants.

sdk/helper/consts/jwt.go
dc-tec: Currently they (Bao and OIDC) are the same, i've split them up as they are used in different parts of the codebase. Where `AllowedJWTSignatureAlgorithmsBao` can be used more generally and the others are more specified to the specific features (i.e PKI, Kubernetes, OIDC). 

I don't mind consolidating them for now tho, it was done so it's easier in the future to change the algorithms for a specific feature.

vault/identity_store_oidc.go
dc-tec: Done!

sdk/helper/consts/jwt.go
DrDaveD: I wasn't suggesting using the same name, but having one be a copy of the other in this file so the whole list doesn't need to be duplicated.

vault/identity_store_oidc.go
DrDaveD: Oh but I meant all the uses of supportedAlgs in that file.  Oh well it could wait for the other issue that looks over the code base.

dc-tec: Ah yeah I didn't change that as it was not directly related. On the other hand it's in-line with the change at hand.

sdk/helper/consts/jwt.go
dc-tec: Removed the duplication.

DrDaveD:
Please rebase on the current main so the CI checks can be run.

JanMa (changes requested):
@dc-tec please have another look at the build failues. As far as I can see they are not present on the `main` branch right now so they might have been introduced by your changes

dc-tec:
> @dc-tec please have another look at the build failures. As far as I can see they are not present on the `main` branch right now so they might have been introduced by your changes

@JanMa Got the test failures fixed. However I'm still not sure if these were (all) due to my changes, those about `non-constant format string` errors for example. However they are sorted now. 

If you would, could you verify `sdk/helper/consts/jws.go` and double check if the algorithms.

JanMa:
> > @dc-tec please have another look at the build failures. As far as I can see they are not present on the `main` branch right now so they might have been introduced by your changes
> 
> @JanMa Got the test failures fixed. However I'm still not sure if these were (all) due to my changes, those about `non-constant format string` errors for example. However they are sorted now.
> 
> If you would, could you verify `sdk/helper/consts/jws.go` and double check if the algorithms.

Thanks a lot for fixing the CI issues @dc-tec. I'll have another look at the algorithms :+1:

api/plugin_helpers.go
```diff
@@ -128,7 +129,7 @@ func VaultPluginTLSProviderContext(ctx context.Context, apiTLSConfig *TLSConfig)
 	return func() (*tls.Config, error) {
 		unwrapToken := ReadBaoVariable(PluginUnwrapTokenEnv)
 
-		parsedJWT, err := jwt.ParseSigned(unwrapToken)
+		parsedJWT, err := jwt.ParseSigned(unwrapToken, []jose.SignatureAlgorithm{jose.ES512})
```
JanMa, line 135: Is this token always using `ES512` or does it make sense to use `AllowedJWTSignatureAlgorithmsBao` here?

builtin/credential/kubernetes/path_login_test.go
```diff
@@ -1429,7 +1430,7 @@ func Test_kubeAuthBackend_getAliasName(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			tok, err := josejwt.ParseSigned(s)
+			tok, err := josejwt.ParseSigned(s, []jose.SignatureAlgorithm{jose.ES256, jose.ES512, jose.EdDSA})
```
JanMa: This should use `AllowedJWTSignatureAlgorithmsK8s`

sdk/helper/consts/jwt.go
```diff
@@ -0,0 +1,56 @@
...
+		jose.PS256,
+		jose.PS384,
+		jose.PS512,
+		jose.ES256,
+		jose.ES384,
+		jose.ES512,
+		jose.EdDSA,
+		jose.HS256,
+		jose.HS384,
+		jose.HS512,
```
JanMa, line 39: I don't think it makes sense to allow `HS*` algorithms in the PKI. @cipherboy any insights if this is okay or not?

sdk/helper/consts/jwt.go
```diff
@@ -0,0 +1,56 @@
...
+	// AllowedJWTSignatureAlgorithmsOIDC defines the JWT signature algorithms
+	// allowed for OIDC operations.
+	AllowedJWTSignatureAlgorithmsOIDC = []jose.SignatureAlgorithm{
+		jose.RS256,
+		jose.RS384,
+		jose.RS512,
+		jose.ES256,
+		jose.ES384,
+		jose.ES512,
+		jose.EdDSA,
```
JanMa: I think it makes sense to also add the `PS*` algorithms here so the list is equivalent to `AllowedJWTSignatureAlgorithmsK8s`.

sdk/helper/consts/jwt.go
dc-tec: I've added those because it caused test failures

dc-tec:
> Some questions regarding the chosen algorithms, otherwise LGTM

Thanks for the feedback I'll go over them later today.

api/plugin_helpers.go
dc-tec: I might have made a wrong assumption, but I used only this algo based on: 
https://github.com/openbao/openbao/blob/b5cfc43f006f8c6fff4f6e7e1192fe4145a58c86/vault/wrapping.go#L217

sdk/helper/consts/jwt.go
JanMa: Then I guess they are needed :smile:

sdk/helper/consts/jwt.go
```diff
@@ -38,19 +38,11 @@ var (
...
 	}
 
-	// AllowedJWTSignatureAlgorithmsOIDC defines the JWT signature algorithms
-	// allowed for OIDC operations.
-	AllowedJWTSignatureAlgorithmsOIDC = []jose.SignatureAlgorithm{
-		jose.RS256,
-		jose.RS384,
-		jose.RS512,
-		jose.ES256,
-		jose.ES384,
```
JanMa: @dc-tec `jose.ES384` is missing from the consolidated list. This is causing the current test failures as far as I can see.

JanMa (changes requested):

sdk/helper/consts/jwt.go
dc-tec: Thanks! I had some things I had to attend to and didn't had the time to further investigate why tests failed.

JanMa (approved):
Lgtm, thanks again @dc-tec

JanMa:
@dc-tec would you be so kind to rebase the branch onto main and resolve the conflicts. Seems I forgot to merge this PR in time :sweat_smile:

dc-tec:
> @dc-tec would you be so kind to rebase the branch onto main and resolve the conflicts. Seems I forgot to merge this PR in time 😅

Ofcourse!

dc-tec:
We still have transitive deps that use go-jose v3:

Root module graph shows these packages still pull in v3: 
- github.com/hashicorp/cap@v0.9.1 
- github.com/openbao/go-kms-wrapping/wrappers/{pkcs11,transit}/v2
- github.com/okta/okta-sdk-golang/v2
- github.com/openbao/openbao-template.

In api/auth/jwt, v3 comes via: 
- github.com/openbao/openbao/api/v2 v2.3.1 (that published tag depends on go-jose/v3/jwt), even though the local api/ in this repo is on v4

DrDaveD (approved):

JanMa:
Thanks for rebasing @dc-tec

PR #1483, opened 2025-06-24
Fix detailed metadata on list results (#1388)

cipherboy:
* Fix detailed metadata on list results

Unlike scan, list results include directories in their results. This results in KVv2 panicing when attempting to fetch metadata about the entry, as it doesn't exist and thus returns nil. We should handle this gracefully.

* Add changelog entry

---------

---

Backport of https://github.com/openbao/openbao/pull/1388.


wslabosz-reply (approved):
LGTM, although you will have to overwrite the changelog req

cipherboy:
@wslabosz-reply Backports use the original changelog entry upstream as well. We don't really need to point users at the backport separately from the main PR.

Merging on review from the core team.

PR #1490, opened 2025-06-25
MFA Login namespace adjustments (#1348)

cipherboy:
* refactor: login mfa namespace awarness adjusted

* refactor: removed legacy helper namespace helper methods

* refactor: login mfa teardown before namespace store teardown

* cr: applied comments with additonal fixes

* cr: remove commented code and fix one listing view

* cr: fix invalid locked users storing path

* refactor: prune locked users entries on namespace delete

* changelog added

* test: added mfa config storage test

* cr: fixed lock bug and added test

---------

Backport of https://github.com/openbao/openbao/pull/1348.

### DIFF
--- a/changelog/1348.txt
+++ b/changelog/1348.txt
@@ -1,0 +1,5 @@
+```release-note:bug
+vault/login_mfa: config storage storage change
+vault/core: locked users storage structure change
+vault/namespaces: deleting namespace removes namespace locked users entries
+```


satoqz:
Huh, why was the changelog removed in c6fbfdedd1f4d8932413b78dcefd9271d841c1c8 in the first place?

cipherboy:
@satoqz Ahhh, I misread this change date. I thought it had landed in `main` after v2.3.0-beta was cut, not before. So the changelog was removed in that commit because we didn't need the changelog for a change also in the release (MFA bugs in namespcaes didn't matter since that beta was the very first release with namespace support). 

Closing!

PR #1488, opened 2025-06-24
Backport go-viper/mapstructure/v2 move

cipherboy:
---

Backport of https://github.com/openbao/openbao/pull/1459 with https://github.com/openbao/openbao/pull/1480; this needs a rebase after #1482 is merged.

PR #1491, opened 2025-06-25
Fix identity store resolution (#1432)

cipherboy:
* Attach namespace to identity information

When allowing cross-namespace identity store operations, we need to attach namespace information to avoid expensive searches across all child namespaces to find the correct namespace for a given identifier.

This change is not backwards compatible with previous versions; we will fail sanitization for any existing entities in namespaces due to the lack of the namespace identifier suffix.

* Make StoragePacker transaction-friendly

The current StoragePacker design retains a copy of logical.Storage, avoiding the need for callers to use their own. However, it doesn't actually use the View interface, meaning we can apply transactions over the top for sensitive operations.

Expose storage as a parameter so that we can have cross-operation transactions.

* Support upgrading entity ids to new format

This resolves the incompatibilities in the previous entity alias change, restoring compatibility with beta, though potentially affecting callers who retained the entity ID long-term. We also make this change on stored entity IDs in the token store.

* Support unsafe_cross_namespace_identity flag

This flag will allow server operators to opt into Vault Enterprise compatible namespace identity store, which uses a single memdb instance across all tenants to allow for cross-namespace group association.

* Ensure loading identity entries across namespaces

Before this commit, on restarts of OpenBao, identities within namespaces were never loaded into memdb, resulting in a failure to use newly created identities. This has been fixed, ensuring all identities are loaded on startup into their respective MemDB instances.

* Add test for cross-namespace identity behavior

This ensures the new unsafe_cross_namespace_identity flag functions as expected, allowing cross-namespace group inclusion in the identity subsystem.

* Add changelog entry

---------

Backport of https://github.com/openbao/openbao/pull/1432. Will conflict with https://github.com/openbao/openbao/pull/1488.

### DIFF
--- a/changelog/1432.txt
+++ b/changelog/1432.txt
@@ -1,0 +1,6 @@
+```release-note:bug
+core/identity: load namespace entities, groups into MemDB preventing them from disappearing on restart.
+```
+```release-note:improvement
+core/identity: add unsafe_cross_namespace_identity to give compatibility with Vault Enterprise's cross-namespace group membership.
+```
--- a/command/server.go
+++ b/command/server.go
@@ -2590,6 +2590,7 @@ func createCoreConfig(c *ServerCommand, config *server.Config, backend physical.
 		EnableResponseHeaderHostname:   config.EnableResponseHeaderHostname,
 		EnableResponseHeaderRaftNodeID: config.EnableResponseHeaderRaftNodeID,
 		AdministrativeNamespacePath:    config.AdministrativeNamespacePath,
+		UnsafeCrossNamespaceIdentity:   config.UnsafeCrossNamespaceIdentity,
 	}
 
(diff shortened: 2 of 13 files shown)


satoqz:
What happened to CI here 😮

cipherboy:
@satoqz I changed the base branch so it got confused. When I rebase it after the mapstructure change finishes it should fix itself.

cipherboy:
@satoqz This is rebased, lmkwyt!

wslabosz-reply (approved):
very neat update

EDIT: I know it's a backport, but I've also reviewed the update when it was made, and it rocks nonetheless

satoqz (approved):

cipherboy:
Merging based on approvals from core team. Thanks!

PR #1496, opened 2025-06-25
Allow disabling unauthenticated rekey

cipherboy:
OpenBao allows interacting with the rekey root and rekey recovery key endpoints without authentication. While the rekey itself requires existing key material to be given on subsequent calls, an unauthenticated attacker could interrupt any rekey operation by calling the cancel endpoint (DELETE /sys/rekey/init) before it could be completed. These interactions were not audited but are going forward.

We add a new option to listeners (disable_unauthed_rekey_endpoints), allowing disabling rekey operations on a per-listener basis. This would allow an operator to safely use this operation from localhost or a Unix socket, but disable it from other global listeners.

In the future when an authenticated variant of the API is available, this option will be set to true by default, disabling all requests to this endpoint. Because rekeying is a rarely-used but highly sensitive operation, the impact should be minimal and the security benefits outweigh the operational risks.

Please update your configurations accordingly.

Resolves: https://github.com/openbao/openbao/security/advisories/GHSA-prpj-rchp-9j5h


website/content/docs/deprecation/unauthed-rekey.mdx
```diff
@@ -0,0 +1,33 @@
+---
+sidebar_label: Unauthenticated Rekey Endpoints
+description: |-
+  Deprecating the `sys/rekey/*` and `sys/rekey-recovery-key/*` endpoints.
```
wslabosz-reply, line 4:
```suggestion
  Deprecating unauthorized use of `sys/rekey/*` and `sys/rekey-recovery-key/*` endpoints.
```

cipherboy: @wslabosz-reply Actually, I'm going to revert this. I don't think we'll ever be able to make these endpoints authenticated, so we'll likely need to make entirely new endpoints for this purpose.

cipherboy:
Thanks @wslabosz-reply! This was additionally approved by @DanGhita on an internal branch; merging.

PR #1390, opened 2025-06-01
Minor improvements to CEL for PKI 

cipherboy:
I was testing this with the following CEL program:

<details>

<summary>Program</summary>

```cel
{
	"cel_program": {
		"variables": [
			{
				"name": "have_cn",
				"expression": "'common_name' in request && size(request['common_name']) > 0"
			},
			{
				"name": "valid_cn",
				"expression": "have_cn && request['common_name'].endsWith('.local.cipherboy.com')"
			},
			{
				"name": "valid",
				"expression": "valid_cn"
			},
			{
				"name": "cn_value",
				"expression": "have_cn ? request['common_name'] : ''"
			},
			{
				"name": "subject",
				"expression": "have_cn ? PKIX.Name{CommonName: cn_value} : null"
			},
			{
				"name": "dns_sans",
				"expression": "have_cn ? [cn_value] : null"
			},
			{
				"name": "extensions",
				"expression": "[PKIX.Extension{Id: [1, 2, 3, 4, 5], Critical: true, Value: b'issued by OpenBao'}]"
			},
			{
				"name": "cert",
				"expression": "CertTemplate{Subject: subject, DNSNames: dns_sans, ExtraExtensions: extensions}"
			},
			{
				"name": "key_type",
				"expression": "'key_type' in request ? request['key_type'] : 'ec'"
			},
			{
				"name": "key_bits",
				"expression": "'key_bits' in request ? uint(request['key_bits']) : 521u"
			},
			{
				"name": "signature_bits",
				"expression": "512u"
			},
			{
				"name": "use_pss",
				"expression": "true"
			},
			{
				"name": "warnings",
				"expression": "request.map(parameter, (parameter in ['role', 'common_name', 'key_type', 'key_bits']) ? '' : ('unrecognized parameter: ' + parameter)).filter(x, x != '')"
			},
			{
				"name": "output",
				"expression": "ValidationOutput{template: cert, generate_lease: false, no_store: true, issuer_ref: 'default', key_type: key_type, key_bits: key_bits, signature_bits: signature_bits, use_pss: use_pss, warnings: warnings}"
			},
			{
				"name": "reason",
				"expression": "!valid_cn ? 'invalid common name' : false"
			}
		],
		"expression": "valid ? output : reason"
	}
}
```

</details>

And found the following minor issues with #794:

```
     - Proper handling of key_bits, key_type, signature_bits, &c,
       to set use the global default handlers from the role system.
     - Move the fields into the privileged evaluationData space, rather
[...]

PR #1499, opened 2025-06-25
Minor improvements to CEL for PKI  (#1390)

cipherboy:
* Regenerate PKI CEL protobuf

* Restructure certutil issuance, cleanup bugs

This fixes a few minor bugs with CEL for PKI issuance:

 - Proper handling of key_bits, key_type, signature_bits, &c, to set use the global default handlers from the role system.
 - Move the fields into the privileged evaluationData space, rather than using the untrusted "request" values: these are not necessarily validated by CEL and so we must set our preferred values output by the program.
 - Clean up handling of extensions on the template.
 - Hide unrecognized parameter warnings via use of TakesArbitraryInput.
 - Support repeated warnings

* Refactor signature bits handling for CSRs

* Add identity into CEL environment

---------

This had conflicts in the protobuf file; since this is self-contained to the PKI engine, I opted to take the version merged to main completely rather than rebuilding. The only other modifying commit, 6225830d096eac837412decb55094c4d3ed0abef, will not be backported to v2.3.0 but shouldn't impact functionality at all.

Backport of https://github.com/openbao/openbao/pull/1390


voigt (approved):

voigt (approved):

cipherboy:
Merging based on reviews from the core team. Thanks!

PR #1506, opened 2025-06-26
Add one-time, declarative self-initialization

cipherboy:
See also: #1341
Resolves: #1340

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -8,6 +8,7 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
@@ -1363,6 +1364,15 @@ func (c *ServerCommand) Run(args []string) int {
 		return 1
 	}
 
+	// Else if we're in production mode, initialize the core as well.
+	c.UI.Error("Trying self-init...")
+	err = c.Initialize(core, config)
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+	c.UI.Error("Self-init done...")
+
 	// Initialize the HTTP servers
 	err = startHttpServers(c, core, config, lns)
 	if err != nil {
@@ -1672,6 +1682,136 @@ func (c *ServerCommand) notifySystemd(status string) {
 			c.logger.Debug("would have sent systemd notification (systemd not present)", "notification", status)
 		}
 	}
+}
+
+func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) error {
+	if len(config.Initialization) == 0 {
+		c.UI.Error("skipping initialization")
+		return nil
+	}
+
+	c.UI.Error("beginning initialization")
+
+	var history server.InitializationContext
+	ctx := namespace.ContextWithNamespace(context.Background(), namespace.RootNamespace)
+
+	var recoveryConfig *vault.SealConfig
+	barrierConfig := &vault.SealConfig{
+		SecretShares:    1,
+		SecretThreshold: 1,
+	}
+
+	if core.SealAccess().RecoveryKeySupported() {
+		recoveryConfig = &vault.SealConfig{
+			SecretShares:    1,
+			SecretThreshold: 1,
+		}
+	}
+
+	barrierConfig.StoredShares = 1
+
+	// Initialize it with a basic single key
+	init, err := core.Initialize(ctx, &vault.InitParams{
+		BarrierConfig:  barrierConfig,
+		RecoveryConfig: recoveryConfig,
+	})
+	if err != nil {
+		return fmt.Errorf("self-initialization failed: %w", err)
+	}
+
+	c.UI.Error("beginning unseal")
+	if ok, err := core.Unseal(init.SecretShares[0]); err != nil || !ok {
+		return fmt.Errorf("unesal after self-initialization failed: err=%w ok=%v", err, ok)
+	}
+
+	c.UI.Error("waiting for leadership")
+
+	isLeader, _, _, err := core.Leader()
+	if err != nil && err != vault.ErrHANotEnabled {
+		return fmt.Errorf("failed to check active status: %w", err)
+	}
+	if err == nil {
+		// Raft is slower than dev mode.
+		leaderCount := 35
+		for !isLeader {
+			if leaderCount == 0 {
+				buf := make([]byte, 1<<16)
+				runtime.Stack(buf, true)
+				return fmt.Errorf("failed to get active status after five seconds; call stack is\n%s", buf)
+			}
+			time.Sleep(1 * time.Second)
+			isLeader, _, _, err = core.Leader()
+			if err != nil {
+				return fmt.Errorf("failed to check active status: %w", err)
+			}
+			leaderCount--
+		}
+	}
+
+	c.UI.Error("beginning post-unseal configuration")
+	for initIndex, initBlock := range config.Initialization {
+		initBlockName := initBlock.Type
+
+		c.UI.Error("initializing: " + initBlockName)
+
+		for reqIndex, reqBlock := range initBlock.Requests {
+			reqBlockName := reqBlock.Type
+
+			c.UI.Error("\trequest: " + reqBlockName + fmt.Sprintf(" %#v", reqBlock.DataRaw))
+
+			req := &logical.Request{
+				ID:        fmt.Sprintf("initialize[%d].%s.request[%d].%s", initIndex, initBlockName, reqIndex, reqBlockName),
+				Operation: logical.Operation(reqBlock.Operation),
+				Path:      reqBlock.Path,
+			}
+
+			var err error
+
+			req.ClientToken, err = reqBlock.GetToken(init.RootToken, history)
+			if err != nil {
+				return fmt.Errorf("initialize.%d[%s].request.%d[%s]: error getting token: %w", initIndex, initBlockName, reqIndex, reqBlockName, err)
+			}
+
+			req.Data, err = reqBlock.GetData(history)
+			if err != nil {
+				return fmt.Errorf("initialize.%d[%s].request.%d[%s]: error getting data: %w", initIndex, initBlockName, reqIndex, reqBlockName, err)
+			}
+
+			reqMarshaled, err := json.Marshal(req)
+			if err != nil {
+				return fmt.Errorf("initialize.%d[%s].request.%d[%s]: failed to marshal request: %w", initIndex, initBlockName, reqIndex, reqBlockName, err)
+			}
+
+			c.UI.Error("\trequest data: " + string(reqMarshaled))
+
+			resp, err := core.HandleRequest(ctx, req)
+			if err != nil {
+				return fmt.Errorf("initialize.%d[%s].request.%d[%s]: failed to make request: %w", initIndex, initBlockName, reqIndex, reqBlockName, err)
+			}
+
+			respMarshaled, err := json.Marshal(resp)
+			if err != nil {
+				return fmt.Errorf("initialize.%d[%s].request.%d[%s]: failed to marshal response: %w", initIndex, initBlockName, reqIndex, reqBlockName, err)
+			}
+
+			c.UI.Error("\tresponse: " + string(respMarshaled))
+
+			var respData map[string]interface{}
+			if err := json.Unmarshal(respMarshaled, &respData); err != nil {
+				return fmt.Errorf("initialize.%d[%s].request.%d[%s]: failed to unmarshal response: %w", initIndex, initBlockName, reqIndex, reqBlockName, err)
+			}
+
+			if err := history.AddRequest(initBlockName, reqBlockName, req.Data); err != nil {
+				return fmt.Errorf("initialize.%d[%s].request.%d[%s]: error saving request %w", initIndex, initBlockName, reqIndex, reqBlockName, err)
+			}
+
+			if err := history.AddResponse(initBlockName, reqBlockName, respData); err != nil {
+				return fmt.Errorf("initialize.%d[%s].request.%d[%s]: error saving request %w", initIndex, initBlockName, reqIndex, reqBlockName, err)
+			}
+		}
+	}
+
+	return nil
 }
 
 func (c *ServerCommand) enableDev(core *vault.Core, coreConfig *vault.CoreConfig) (*vault.InitResult, error) {
--- a/command/server/initialization.go
+++ b/command/server/initialization.go
@@ -1,0 +1,412 @@
+package server
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/hashicorp/hcl"
+	"github.com/hashicorp/hcl/hcl/ast"
+	"github.com/openbao/openbao/internalshared/configutil"
+)
+
+type InitializationContext struct {
+	// Most maps are multi-dimensional:
+	//  - Name of initialization block
+	//  - Name of request block
+	//  - Actual data (usually map string->interface)
+	Requests  map[string]map[string]map[string]interface{}
+	Responses map[string]map[string]map[string]interface{}
+}
+
+func (ic *InitializationContext) AddRequest(initBlock string, requestBlock string, request map[string]interface{}) error {
+	if ic.Requests == nil {
+		ic.Requests = make(map[string]map[string]map[string]interface{})
+	}
+	if ic.Requests[initBlock] == nil {
+		ic.Requests[initBlock] = make(map[string]map[string]interface{})
+	}
+	if ic.Requests[initBlock][requestBlock] != nil {
+		return fmt.Errorf("existing request with same initialize (%v) and request (%v) block names", initBlock, requestBlock)
+	}
+
+	ic.Requests[initBlock][requestBlock] = request
+	return nil
+}
+
+func (ic *InitializationContext) AddResponse(initBlock string, responseBlock string, response map[string]interface{}) error {
+	if ic.Responses == nil {
+		ic.Responses = make(map[string]map[string]map[string]interface{})
+	}
+	if ic.Responses[initBlock] == nil {
+		ic.Responses[initBlock] = make(map[string]map[string]interface{})
+	}
+	if ic.Responses[initBlock][responseBlock] != nil {
+		return fmt.Errorf("existing response with same initialize (%v) and response (%v) block names", initBlock, responseBlock)
+	}
+
+	ic.Responses[initBlock][responseBlock] = response
+	return nil
+}
+
+func (ic *InitializationContext) GetRequest(initBlock string, requestBlock string, fieldSelector interface{}) (interface{}, error) {
+	init, ok := ic.Requests[initBlock]
+	if !ok {
+		return nil, fmt.Errorf("missing initialize block '%v'", initBlock)
+	}
+
+	req, ok := init[requestBlock]
+	if !ok {
+		return nil, fmt.Errorf("missing request block '%v' inside initialize block '%v'", requestBlock, initBlock)
+	}
+
+	val, err := ic.getField(req, fieldSelector)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving field; %w", err)
+	}
+
+	return val, nil
+}
+
+func (ic *InitializationContext) GetResponse(initBlock string, responseBlock string, fieldSelector interface{}) (interface{}, error) {
+	init, ok := ic.Responses[initBlock]
+	if !ok {
+		return nil, fmt.Errorf("missing initialize block '%v'", initBlock)
+	}
+
+	req, ok := init[responseBlock]
+	if !ok {
+		return nil, fmt.Errorf("missing response block '%v' inside initialize block '%v'", responseBlock, initBlock)
+	}
+
+	val, err := ic.getField(req, fieldSelector)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving field; %w", err)
+	}
+
+	return val, nil
+}
+
+func (ic *InitializationContext) getField(obj map[string]interface{}, fieldSelector interface{}) (interface{}, error) {
+	switch typed := fieldSelector.(type) {
+	case string:
+		val, present := obj[typed]
+		if !present {
+			return nil, fmt.Errorf("field '%v' is missing", typed)
+		}
+
+		return val, nil
+	case []string:
+		for i, selector := range typed {
+			val, present := obj[selector]
+			if !present {
+				return nil, fmt.Errorf("field '%v' at depth %v is missing", selector, i)
+			}
+
+			if i == len(typed)-1 {
+				return val, nil
+			}
+
+			obj, present = val.(map[string]interface{})
+			if !present {
+				return nil, errors.New("object did not have sufficient depth for selector")
+			}
+		}
+
+		return nil, errors.New("selector had zero length")
+	default:
+		return nil, fmt.Errorf("unknown type for selector: %T", fieldSelector)
+	}
+}
+
+func (ic *InitializationContext) Evaluate(obj map[string]interface{}) (interface{}, error) {
+	var err error
+	source, sourcePresent := obj["eval_source"]
+	objType, objPresent := obj["eval_type"]
+
+	if !sourcePresent && !objPresent {
+		// Walk object and see if any of its keys are typed objects.
+		for key, value := range obj {
+			if subObj, ok := value.(map[string]interface{}); ok {
+				ret, err := ic.Evaluate(subObj)
+				if err != nil {
+					return nil, fmt.Errorf("[%v]: %w", key, err)
+				}
+				obj[key] = ret
+			} else if listObj, ok := value.([]map[string]interface{}); ok {
+				// If we have a single item, which satisfies the above, then
+				// swap.
+				if len(listObj) == 1 {
+					subObj := listObj[0]
+					_, sourcePresent := subObj["eval_source"]
+					_, objPresent := subObj["eval_type"]
+					if sourcePresent || objPresent {
+						ret, err := ic.Evaluate(subObj)
+						if err != nil {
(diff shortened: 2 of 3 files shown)


command/server/initialization.go
```diff
@@ -0,0 +1,423 @@
+package server
```
cipherboy: Given we might implement profiles, I might move the general pattern of parsing this type of config into `helper/` in the future. 

`env` and `file` sources will need to be disabled, though we might support `cel`.

command/server.go
```diff
@@ -1680,6 +1688,147 @@ func (c *ServerCommand) notifySystemd(status string) {
...
+		return nil
+	}
+
+	if !core.SealAccess().RecoveryKeySupported() {
+		return errors.New("self-initialization requires auto-unseal as there is no way to persist the Shamir's keys")
+	}
+
+	// Initialize it with a basic single recovery key.
+	//
+	// XXX (ascheel): identify if we can drop these to zero and generate
```
cipherboy: @KrzysztofKornalewski-Reply This will likely be the subject of future RFCs now that #1496 has landed. I'd like to set this to zero and allow operators to create these later, from a root token. 

(this is in regards to `5. There are hardcoded “1 share / 1 threshold”. Is this tmp?`)

Maybe I'll draft that in the interim as a separate RFC.

KrzysztofKornalewski-Reply:
I'm looking over the code and see some possible improvements.

**1. There are no retry or fallback mechanisms for failed API requests.**

When making requests to the core (core.HandleRequest method), any error is immediately returned and interrupts the entire initialization process. There are no retries.

**2. There are hard timeouts when waiting for a cluster leader.**

**3. Limited validation and no recovery of data sources**

Initialization blocks can retrieve data from various sources (env, file, request, response), but PoC does not check:
- whether the file exists and has the correct format,
- whether the environment variable is not empty or invalid,
- whether the response from the earlier request contains the expected fields for stringing.

In case of incompatibility, the data goes straight to error, and initialization fails without the option to define default values or an alternative path.

**4.  Perhaps we should also verify whether an initial attempt has already occurred?**

**5.  There are hardcoded “1 share / 1 threshold”. Is this tmp?**

**6. Maybe there should be also post‐unseal sanity checks to confirm that configuration requests actually changed server state.**

That's my general thoughts. I would like to help you with something. Would you like to split work somehow?

command/server.go
```diff
@@ -1680,6 +1688,147 @@ func (c *ServerCommand) notifySystemd(status string) {
...
+		// other authentication information but on subsequent startups
+		// presumably the admin has created an alternative mechanism we should
+		// defer to.
+		return nil
+	}
+
+	if !core.SealAccess().RecoveryKeySupported() {
+		return errors.New("self-initialization requires auto-unseal as there is no way to persist the Shamir's keys")
+	}
+
```
cipherboy: @KrzysztofKornalewski-Reply We need a check for whether or not the node is a non-voter node to help resolve (`2. There are hard timeouts when waiting for a cluster leader.`).

However, I'm not sure how we get around timeboxed retries. It is an asynchronous process for leadership election. I'm not sure we can always solve this...

Maybe:

1. Check if initialized, exit early if so.
2. Grab leadership lock (if possible without initializing?)
3. Check if initialized, if so, exit again.
4. Initialize seal + do self-init

cipherboy (commented):
Hey @KrzysztofKornalewski-Reply! Thanks for the comments.

> **1. There are no retry or fallback mechanisms for failed API requests.**
> 
> When making requests to the core (core.HandleRequest method), any error is immediately returned and interrupts the entire initialization process. There are no retries.

The request itself is entirely local; there is no network involved here (not even loopback) so the request should never fail in a manner which a retry would be necessary or which would be expected to succeed.

The request may return an error (e.g., due to bad parameter value), in which case initialization would fail. But there's no retry mechanism that's possible, in that scenario. They have to modify config, which would require a restart. In that case, they should also delete the database to re-initialize from scratch.

IMHO, the goal shouldn't be to add a full orchestration system into OpenBao, but merely to provide enough of a scaffold to set something up enough to let a proper orchestration tool like OpenTofu/... take over.

This is a design point as well:

https://github.com/openbao/openbao/pull/1341/files#diff-0e28e4297fe3dad4f326602f412c73d516ebf481f497aaa8849e320cb93e898aR53-R55

explicitly states this is one-time. I don't really want to weaken the security model; it would be hard to detect if self-initialization was finished except by some other sentinel besides barrier/seal. This leads to a problem where, if an existing cluster is upgraded with a config that does self-init, it _could_ modify things again, which Would Be Bad (TM). 

My 2c., but I'm curious how much you think this is worth solving?

----



> 3. Limited validation and no recovery of data sources
> 
> Initialization blocks can retrieve data from various sources (env, file, request, response), but PoC does not check:
>
>     whether the file exists and has the correct format,
>     whether the environment variable is not empty or invalid,
>     whether the response from the earlier request contains the expected fields for stringing.
> 
> In case of incompatibility, the data goes straight to error, and initialization fails without the option to define default values or an alternative path.

I'm not quite sure how we'd deal with incompatibility; we're reading the file verbatim for use as the request parameter value (and in the future, potentially the full request). I don't think we can easily do any validation there on the contents itself if the type of the parameter is `string`? (If it is more strongly typed, the conversion will detect format bugs). I'd also probably not err on empty environment variable values; empty string is different than undefined environment and some values could be set at an empty-but-present-variable and still work. 

That said, yes, startup should probably fail sooner if files/environment variables are missing. If you want to add logic for that, we could move this over to a separate branch and add a config validation call.

The response field thing I think is fixed in my last push fwiw.

---

> 4. Perhaps we should also verify whether an initial attempt has already occurred?

My new code should handle this; sorry, I had other things going on before I could push the WIP branch. 

---

> 6. Maybe there should be also post‐unseal sanity checks to confirm that configuration requests actually changed server state.

I'm not sure how we'd validate that, thoughts? 

It might be nice to make sure we have a call to an `sys/auth` endpoint, and warn otherwise, to help ensure that users can auth, but there's many ways to misconfigure (most) auth endpoints so that auth is impossible anyways. :-D 

Hence, I think these might mostly be trial-and-error configurations fwiw.

KrzysztofKornalewski-Reply:
@cipherboy OK, so for the beginning I can take care about adding validation for file existence before startup.

command/server.go
cipherboy: @KrzysztofKornalewski-Reply Hmm, this approach does not work. Leadership lock requires unsealed (which makes sense, we want to ensure potential leaders can read the barrier successfully), which means our order has to be:

1. Check if initialized, exit early if so.
2. Do seal initialization
3. Grab leadership lock

and have to live with the race of 2-3.

command/server/initialization.go
cipherboy: Done.

command/server.go
cipherboy: This is opened in #1512.

cipherboy:
@KrzysztofKornalewski-Reply I've pushed a rather large refactor. Let me know what you think of the improvements and if you have ways around the leadership locking problem. 

I think this is reasonably complete from an initial feature PoV. Do you want to work on the TODOs I've left (in the code and in the description of the PR), in a new branch/PR, and also CEL support as a source (see [guidelines](https://openbao.org/docs/rfcs/cel-best-practices/))?

---

I think this largely fixes problem 3 you identified as well. If we add CEL as a support, you could do something like:

```hcl
initialize "block" {
  request "optional" {
    allow_failure = {
      eval_type = "bool"
      eval_source = "cel"
      expr = "response.status_code == 404"
    }
  }
}
```

to allow failure only if the status is 404 not found. We could additionally define a `skip` value at both the request and outer block level to allow skipping requests or sections of requests, conditionally, based on some previous value. That would also be a good item for you to work on (`skip` support, once CEL support lands as a source).

One trick with CEL support is we'll want to be able to control what helpers are injected into the CEL library context, but I can help you with that. 

Thoughts?

KrzysztofKornalewski-Reply:
> @KrzysztofKornalewski-Reply I've pushed a rather large refactor. Let me know what you think of the improvements and if you have ways around the leadership locking problem.
> 
> I think this is reasonably complete from an initial feature PoV. Do you want to work on the TODOs I've left (in the code and in the description of the PR), in a new branch/PR, and also CEL support as a source (see [guidelines](https://openbao.org/docs/rfcs/cel-best-practices/))?
> 
> I think this largely fixes problem 3 you identified as well. If we add CEL as a support, you could do something like:
> 
> ```hcl
> initialize "block" {
>   request "optional" {
>     allow_failure = {
>       eval_type = "bool"
>       eval_source = "cel"
>       expr = "response.status_code == 404"
>     }
>   }
> }
> ```
> 
> to allow failure only if the status is 404 not found. We could additionally define a `skip` value at both the request and outer block level to allow skipping requests or sections of requests, conditionally, based on some previous value. That would also be a good item for you to work on (`skip` support, once CEL support lands as a source).
> 
> One trick with CEL support is we'll want to be able to control what helpers are injected into the CEL library context, but I can help you with that.
> 
> Thoughts?


I take these tasks from ToDo lists
UT branch:
https://github.com/openbao/openbao/commit/c433c6cd6c49c9cc1c9b1d2bd0828a981d3e607c

cipherboy:
@openbao/openbao-core-committers this should be ready for review :-) @KrzysztofKornalewski-Reply has graciously provided test cases and cleanups for todos in #1520.

command/server.go
```diff
@@ -1680,6 +1688,150 @@ func (c *ServerCommand) notifySystemd(status string) {
...
+func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) error {
+	if len(config.Initialization) == 0 {
+		return nil
+	}
+
+	if !core.SealAccess().RecoveryKeySupported() {
+		return errors.New("self-initialization requires auto-unseal as there is no way to persist the Shamir's keys")
+	}
+
+	ctx := namespace.ContextWithNamespace(context.Background(), namespace.RootNamespace)
```
wslabosz-reply:
```suggestion
	ctx := namespace.RootContext(context.Background())
```

command/server.go
```diff
@@ -1680,6 +1688,150 @@ func (c *ServerCommand) notifySystemd(status string) {
...
+		// way of sidestepping authentication. At first startup there is no
+		// other authentication information but on subsequent startups
+		// presumably the admin has created an alternative mechanism we should
+		// defer to.
+		return nil
+	}
+
+	// Initialize it with a basic single recovery key.
+	//
+	// XXX (ascheel): identify if we can drop these to zero and generate
```
wslabosz-reply: pointing out that it's doable with the current state of `main` branch.

command/server.go
```diff
@@ -1680,6 +1688,150 @@ func (c *ServerCommand) notifySystemd(status string) {
...
+	init, err := core.Initialize(ctx, &vault.InitParams{
+		BarrierConfig:  barrierConfig,
+		RecoveryConfig: recoveryConfig,
+	})
+	if err != nil {
+		return fmt.Errorf("self-initialization failed: %w", err)
+	}
+
+	if ok, err := core.Unseal(init.RecoveryShares[0]); err != nil || !ok {
+		return fmt.Errorf("unesal after self-initialization failed: err=%w ok=%v", err, ok)
```
wslabosz-reply:
```suggestion
		return fmt.Errorf("unseal after self-initialization failed: err=%w ok=%v", err, ok)
```

helper/profiles/config.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/hcl"
+	"github.com/hashicorp/hcl/hcl/ast"
+	"github.com/openbao/openbao/internalshared/configutil"
+)
+
+// Outer is a named configuration object that contains one or more request
```
wslabosz-reply:
```suggestion
// OuterConfig is a named configuration object that contains one or more request
```

helper/profiles/env_source.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+)
+
+// EnvSourceBuilder allows reading environment variables from the system.
+func EnvSourceBuilder(ctx context.Context, engine *ProfileEngine, field map[string]interface{}) (Source, error) {
```
wslabosz-reply: do we need the err return signature?

helper/profiles/profiles.go
```diff
@@ -0,0 +1,416 @@
...
+// only a single outer block is allowed, which may be empty.
+func WithLogger(logger hclog.Logger) func(*ProfileEngine) {
+	return func(p *ProfileEngine) {
+		p.logger = logger
+	}
+}
+
+// SourceBuilder creates a new concrete source mapped to a particular
+// instance of a field.
+type SourceBuilder func(ctx context.Context, engine *ProfileEngine, field map[string]interface{}) (Source, error)
```
wslabosz-reply: For now not a single struct implementing this interface returns back an error, is this needed?

helper/profiles/history.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (eh *EvaluationHistory) GetRequestField(outerBlock string, requestBlock string, fieldSelector interface{}) (interface{}, error) {
+	values, err := eh.getValue(eh.Requests, outerBlock, requestBlock)
+	if err != nil {
+		return nil, err
+	}
+
+	val, err := eh.getField(values, fieldSelector)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving field; %w", err)
```
wslabosz-reply:
```suggestion
		return nil, fmt.Errorf("error resolving field: %w", err)
```

helper/profiles/request_source.go
```diff
@@ -0,0 +1,94 @@
...
+			return nil, nil, fmt.Errorf("field '%v' is of wrong type: expected 'string' got '%T'", outerFieldName, rawOuterName)
+		}
+
+		requestName = outerName + "."
+		s.outerName = outerName
+	}
+
+	rawReqName, present := s.field["req_name"]
+	if !present {
+		return nil, nil, fmt.Errorf("request source is missing required field '%v'", "req_name")
```
wslabosz-reply:
```suggestion
		return nil, nil, fmt.Errorf("request source is missing required field %q", "req_name")
```

helper/profiles/response_source.go
```diff
@@ -0,0 +1,94 @@
...
+var _ Source = &ResponseSource{}
+
+func (s *ResponseSource) Validate(_ context.Context) ([]string, []string, error) {
+	var responseName string
+
+	if s.outer != "" {
+		outerFieldName := fmt.Sprintf("%v_name", s.outer)
+		rawOuterName, present := s.field[outerFieldName]
+		if !present {
+			return nil, nil, fmt.Errorf("response source is missing required field '%v'", outerFieldName)
```
wslabosz-reply:
```suggestion
			return nil, nil, fmt.Errorf("response source is missing required field %q", outerFieldName)
```

helper/profiles/response_source.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		outerName, ok := rawOuterName.(string)
+		if !ok {
+			return nil, nil, fmt.Errorf("field '%v' is of wrong type: expected 'string' got '%T'", outerFieldName, rawOuterName)
```
wslabosz-reply: these are all nitpicks to use the `%q` formatting
```suggestion
			return nil, nil, fmt.Errorf("field %q is of wrong type: expected 'string' got '%T'", outerFieldName, rawOuterName)
```

helper/profiles/profiles.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+// Sets the default token for the profile evaluation engine.
+func WithDefaultToken(token string) func(*ProfileEngine) {
+	return func(p *ProfileEngine) {
+		p.defaultToken = token
+	}
+}
+
+// Sets the profile
```
wslabosz-reply: consistency nit
```suggestion
// Sets the profile.
```

helper/profiles/env_source.go
cipherboy: @wslabosz-reply Yes, to implement `profiles.SourceBuilder` type:

```go
// SourceBuilder creates a new concrete source mapped to a particular
// instance of a field.
type SourceBuilder func(ctx context.Context, engine *ProfileEngine, field map[string]interface{}) (Source, error)
```

Not every type (like env) will implement this, but `SourceBuilder` is used in `profile.WithSourceBuilder(...)` as an opt to `NewEngine(...)`.

helper/profiles/profiles.go
cipherboy: @wslabosz-reply Ah, I saw this one after the above. I believe CEL might incur an error on construction potentially. I suppose conceptually we could move all processing into the `.Validate(...)` method?

command/server.go
```diff
@@ -1680,6 +1688,150 @@ func (c *ServerCommand) notifySystemd(status string) {
...
+	isLeader, _, _, err := core.Leader()
+	if err != nil && err != vault.ErrHANotEnabled {
+		return false, fmt.Errorf("failed to check active status: %w", err)
+	}
+	if err == nil {
+		// Raft is slower than dev mode.
+		leaderCount := 35
+		for !isLeader {
+			if leaderCount == 0 {
+				buf := make([]byte, 1<<16)
```
satoqz: Would love to see an explanation on the magic numbers here.

command/server.go
```diff
@@ -1680,6 +1688,150 @@ func (c *ServerCommand) notifySystemd(status string) {
...
+			time.Sleep(1 * time.Second)
+			isLeader, _, _, err = core.Leader()
+			if err != nil {
+				return false, fmt.Errorf("failed to check active status: %w", err)
+			}
+			leaderCount--
+		}
+	}
+
+	return isLeader, nil
```
satoqz: Not quite sure on the background here, but what happens if we hit `ErrHANotEnabled` and get `isLeader = false`? Shouldn't that be okay and just do the self init?

command/server.go
cipherboy: This was copied from the existing version under `enableDev(...)`:

https://github.com/openbao/openbao/blob/390a273353dd0b1fa75a47906350f8f48aeb259d/command/server.go#L1731-L1750

So I am afraid they're lost to time. Leader count was extended to provide a longer sleep period for leadership to be acquired (35 seconds versus 5 in dev mode), which I explained above.

But I think the buffer size is perhaps is just an upper bound on how much stack trace to print at this point in time. It is somewhat weird for our use case that we have to pre-allocate and there is no complete version of it, but it makes sense in the broader usage of the `runtime.Stack(...)` function. Presumably if you need more than 65kb of stack trace to debug a failed leadership election, you might have problems.

---

That said, I do find it weird we aren't guaranteed to be the leader. I think it would be _rare_ for us not to be, but  I didn't see a way of first acquiring leadership and then initializing OpenBao. If you point 8 nodes at empty storage, they will race to initialize the seal mechanism and then one will become the leader (which may be different than the one that initialized the seal mechansim -- thus causing us to bail from the initialization code of the one which ultimately became leader).

I don't know how likely this scenario is to occur; I think an invalidation of some sort would have to happen first for the other nodes to realize storage is initialized and thus attempt to become leader, but I might be wrong.

command/server.go
cipherboy: @satoqz Like I mentioned in the [earlier comment](https://github.com/openbao/openbao/pull/1506#discussion_r2219595152), this is the same behavior as devmode and is equally undocumented. 

If we hit `ErrHANotEnabled`, it means that we don't have HA mode enabled at all, so we know we'll be the only active node ever, so we'll skip the subsequent loop. You can't run a HA/Standby/... setup on a non-HA-enabled instance.

command/server.go
satoqz: I see, hmm, welp. Thanks for the explanation, I suppose this is fine then.

command/server.go
satoqz: @cipherboy in that case, if we return `false, nil`, won't we hit line 1787 unintendedly as there was no election in the first place?

```go
if !isLeader {
		return fmt.Errorf("initializing node did not win leadership election: ensure only one active, voting node during initialization")
}
```

helper/profiles/profiles.go
cipherboy: @wslabosz-reply Fixed in this next push :-)

cipherboy:
@wslabosz-reply @satoqz Wow, that was entertaining. I think this should have the behavior we want. While we can't strictly ensure that we acquire leadership (that'd require we modify how the UUID for HA leadership gets generated), it so far hasn't happened in my testing that we fail to acquire leadership.

I've attached some additional configurations:

<details>

<summary> Local testing </summary>

A single node raft:

```hcl
ui            = true
cluster_addr  = "https://127.0.0.1:8201"
api_addr      = "https://127.0.0.1:8200"

storage "raft" {
  path = "raft"
}

listener "tcp" {
  tls_disable   = true
}

seal "static" {
  current_key_id = "20250630"
  current_key = "542a91c2aff86797e8dfa16adf86dcbea5706410791b319798936a1b08e4ef8b"
}

initialize "audit" {
  request "enable-audit" {
    operation = "update"
    path = "sys/audit/stdout"
    data = {
      type = "file"
      options = {
        file_path = "/dev/stdout"
        log_raw = true
      }
    }
  }
}

initialize "identity" {
  request "mount-userpass" {
    operation = "update"
    path = "sys/auth/userpass"
    data = {
      type = "userpass"
      path = "userpass/"
      description = "admin"
    }
  }

  request "userpass-add-admin" {
    operation = "update"
    path = "auth/userpass/users/admin"
    data = {
      "password" = {
        eval_type = "string"
        eval_source = "env"
        env_var = "INITIAL_ADMIN_PASSWORD"
      }
      "token_policies" = ["superuser"]
    }
  }
}

initialize "policy" {
  request "add-superuser-policy" {
    operation = "update"
    path = "sys/policies/acl/superuser"
    data = {
      policy = <<EOP
path "*" {
  capabilities = ["create", "update", "read", "delete", "list", "scan", "sudo"]
}
EOP
    }
  }
}
```

----

Or, a PostgreSQL parallel testing script:

```bash
#!/bin/bash

killall bao
podman kill postgres
rm ./server-*.log
rm config-postgresql-*.json

sleep 2

podman run --replace --name postgres -e POSTGRES_USER=username -e POSTGRES_PASSWORD=password -p 5432:5432 --detach postgres:latest

if (( $(date +%S) > 50 )); then
  echo "Waiting for next minute"
  sleep 10
fi

then="$(date +%H:%M:)$(( $(date +%S) + 2 )).9999"
target_epoch=$(date -d "$then" +%s.%N)


for i in `seq 821 830`; do
  (
    sed "s/820/$i/g" config-postgresql.json > config-postgresql-$i.json

    current_epoch=$(date +%s.%N)
    sleep_seconds=$(echo "$target_epoch - $current_epoch"|bc)
    echo "Sleeping $sleep_seconds"
    sleep -- $sleep_seconds

    INITIAL_ADMIN_PASSWORD=Secret123 bao server -config=config-postgresql-$i.json > server-$i.log 2>&1
  ) &
done

echo "Waiting for target epoch..."
current_epoch=$(date +%s.%N)
sleep_seconds=$(echo "$target_epoch - $current_epoch"|bc)
sleep $sleep_seconds

echo "Waiting for startup..."
sleep 10

for i in `seq 821 830`; do
  BAO_ADDR="http://localhost:${i}0" BAO_TOKEN="" bao write auth/userpass/login/admin password=Secret123 || echo "Failed on $i"
done

killall bao
podman kill postgres
wait
```

and base config file which also works for a single node:

```json
{
  "ui": true,
  "cluster_addr": "https://127.0.0.1:8201",
  "api_addr": "https://127.0.0.1:8200",
  "storage": {
    "postgresql": {
      "connection_url": "postgres://username:password@127.0.0.1:5432/postgres",
      "max_connect_retries": 0,
      "ha_enabled": true
    }
  },
  "listener": [
    {
      "tcp": {
        "address": "127.0.0.1:8200",
        "tls_disable": true
      }
    }
  ],
  "seal": {
    "static": {
      "current_key_id": "20250630",
      "current_key": "542a91c2aff86797e8dfa16adf86dcbea5706410791b319798936a1b08e4ef8b"
    }
  },
  "initialize": [
    {
      "audit": {
        "request": [
          {
            "enable-audit": {
              "operation": "update",
              "path": "sys/audit/stdout",
              "data": {
                "type": "file",
                "options": {
                  "file_path": "/dev/stdout",
                  "log_raw": true
                }
              }
            }
          }
        ]
      }
    },
    {
      "identity": {
        "request": [
          {
            "mount-userpass": {
              "operation": "update",
              "path": "sys/auth/userpass",
              "data": {
                "type": "userpass",
                "path": "userpass/",
                "description": "admin"
              }
            }
          },
          {
            "userpass-add-admin": {
              "operation": "update",
              "path": "auth/userpass/users/admin",
              "data": {
                "password": {
                  "eval_type": "string",
                  "eval_source": "env",
                  "env_var": "INITIAL_ADMIN_PASSWORD"
                },
                "token_policies": ["superuser"]
              }
            }
          }
        ]
      }
    },
    {
      "policy": {
        "request": [
          {
            "add-superuser-policy": {
              "operation": "update",
              "path": "sys/policies/acl/superuser",
              "data": {
                "policy": "path \"*\" {\n  capabilities = [\"create\", \"update\", \"read\", \"delete\", \"list\", \"scan\", \"sudo\"]\n}"
              }
            }
          }
        ]
      }
    }
  ]
}
```

</details>

@KrzysztofKornalewski-Reply Do you mind turning the second example into a proper unit test?

KrzysztofKornalewski-Reply:
> @wslabosz-reply @satoqz Wow, that was entertaining. I think this should have the behavior we want. While we can't strictly ensure that we acquire leadership (that'd require we modify how the UUID for HA leadership gets generated), it so far hasn't happened in my testing that we fail to acquire leadership.
> 
> I've attached some additional configurations:
> 
> Local testing
> @KrzysztofKornalewski-Reply Do you mind turning the second example into a proper unit test?

Sure @cipherboy , I will do it :)

PR #1513, opened 2025-06-29
Add constants for commonly used vault headers #1443

### LINKED ISSUE #1443: `"X-Vault-MFA"` doesn't really have a convenient constant anywhere, should we add one?
`"X-Vault-MFA"` doesn't really have a convenient constant anywhere, should we add one?

_Originally posted by @satoqz in https://github.com/openbao/openbao/pull/1433#discussion_r2145359851_
            

---

Looking more closely, at least `X-Vault-Wrap-TTL`, `X-Vault-MFA`, `X-Vault-Policy-Override`, and likely many more constants are missing. It'd be great to add these to `api/client.go` and `sdk/helper/consts/consts.go`.

sudhanvaghebbale:
Resolves #1443


api/client.go
```diff
@@ -83,6 +83,31 @@ const (
 	// dash) is ignored. Any repeated header keys result in request failure.
 	InlineAuthParameterHeaderPrefix = "X-Vault-Inline-Auth-Parameter-"
 
+	// HostNameHeader is the header containing the hostname information for the Vault server
+	// It helps identify which specific Vault instance is handling the request.
+	HostNameHeader = "X-Vault-Hostname"
```
satoqz, line 88: All existing header constants end in `HeaderName` instead of just `Header`, I'd prefer to keep that consistent.

This constant specifically would be `HostnameHeaderName` (no capital `N` in `Hostname` as it's a single word)

sdk/helper/consts/consts.go
```diff
@@ -15,10 +15,53 @@ const (
...
+	// Ensures requests are processed locally and not forwarded to other nodes
+	NoRequestForwardingHeader = "X-Vault-No-Request-Forwarding"
+
+	// WrapFormatHeader is the header for specifying the format of the wrapped response
+	// Allows clients to request a specific format (e.g., JSON) for response wrapping
+	WrapFormatHeader = "X-Vault-Wrap-Format"
+
 	// PerformanceReplicationALPN is the negotiated protocol used for
 	// performance replication.
 	PerformanceReplicationALPN = "replication_v1"
```
satoqz, line 67: Hmm, ordering in this file is a bit different compared to `api/client.go` because inline auth header constants are below other non-header constants. I think it's worth keeping the ordering between these two files consistent so you can easily diff them.

api/client.go
```diff
@@ -83,6 +83,31 @@ const (
...
+	// RawErrorHeader is the header for raw error information that bypasses error formatting
+	// When this header is present, Vault returns unprocessed error details for debugging purposes.
+	RawErrorHeader = "X-Vault-Raw-Error"
+
+	// WrapTTLHeader is the header for response wrapping TTL (Time To Live)
+	// Specifies how long a wrapped response should be valid before automatic unwrapping.
+	WrapTTLHeader = "X-Vault-Wrap-TTL"
+
+	// MFAHeader is the header for multi-factor authentication credentials
+	// Used to pass MFA challenge responses during authentication.
```
satoqz, line 104: Very nitpicky: Existing header constant doc comments consistently end sentences with a dot and wrap to 80 characters.

satoqz (commented):
Great work @sudhanvaghebbale! Thanks!

There are a couple leftover literal header strings in `vault/external_tests` that can be replaced.

It seems `http/handler.go` also defines some header constants that are now also available in `api`/`sdk`, so those can be deduplicated and removed from `http`.

satoqz:
Hi @sudhanvaghebbale, are you still interested in bringing this over the finish line? I can pick this up otherwise.

satoqz:
Superseded by https://github.com/openbao/openbao/pull/2323.

Issue #1511, opened 2025-06-27
RFC - Delay recovery key generation for auto-unseal mechanisms and make rotation authenticated


cipherboy:
### Summary

We propose adding authenticated root and recovery key rotation endpoints, allow
delayed recovery key generation (setting initial shares to 0) and add a CLI helper
to manually perform offline recovery key generation on an auto-unseal cluster if
keys were lost. This will allow better side-effect-less self-initialization and
and solves the issue with the unauthenticated recovery key rotation APIs.

### Problem Statement

OpenBao's initialization with an auto-unseal mechanism is side-effecting in
a way that requires long-term storage of recovery keys, used to create root
tokens for use during initialization (to create any initial audit, auth, and
secret engines) and [recovery mode](https://openbao.org/docs/concepts/recovery-mode/).

When running in certain environments, like Kubernetes or provisioning via
OpenTofu, having such a side-effecting change makes writing automation
difficult. For instance, [in the context of GitLab deployments](https://docs.gitlab.com/charts/installation/secrets/),
the GitLab helm chart does not assume RBAC access to create secrets and
many optionally be installed with pre-created secrets provisioned manually
by an operator. We wish to support this type of ahead-of-time static
definition of required secrets; creating a random, long-term, persistent
secret (such as recovery key shards) during the installation process is at
odds with this.

While a [separate RFC](https://github.com/openbao/openbao/issues/1340)
discusses approaches for declarative self-initialization and another focuses
on [static unseal](https://github.com/openbao/openbao/pull/1304), this focuses
on changes to the unseal process to improve operator experience.

This effort is made more important by the recent vulnerability in rekeying
[recovery and root keys](https://github.com/openbao/openbao/pull/1496),
where an unauthenticated attacker could effectively disable rekey without
notification to a server operator. Creating authenticated varieties of these
endpoints thus serve a dual purpose: to adequately defend against this
category of attack and to allow us to default to zero recovery keys returned
on initialization, instead using a privileged token to create them.

Lastly, this document proposes explicit use of `rotate` + `<descriptive noun>`
as the authoritative descriptive of key rotation processes. `rekey` as a term
should not be used to describe something unique as it is a synonym to `rotate`.
This should help to address issues with clarity and ambiguity.

### User-facing Description

This change lets the `sys/init` API endpoint take the `0` value for number of
shares when using an auto-unseal mechanism. When specified, the recovery keys
will not be generated at initialization time and thus will not be returned.
[...]

PR #1518, opened 2025-07-01
Add authenticated Rekey endpoints

wslabosz-reply:
Implements https://github.com/openbao/openbao/issues/1511

List of changes:
- `api/sys_rekey.go` functions marked as deprecated
- added functions for calling new endpoints to the `api/sys_rotate.go` 
- added to the `injectDataIntoTopRoutes` in the `http/handler.go`:
```
"/v1/sys/rotate/root/backup",
"/v1/sys/rotate/recovery/backup",
"/v1/sys/rotate/keyring",
```
as aliases to:
```
"/v1/sys/rekey/backup",
"/v1/sys/rekey/recovery-key-backup",
"/v1/sys/rotate",
```
- added `sysHelp` docs for:
```
"rotate_root",
"rotate_init",
"rotate_update",
"rotate_verify",
```
- moved "rotate" operation related code from `vault/logical_system.go` and `vault/logical_system_paths.go` to newly created `vault/logical_system_rotate.go`
- added `(*SealConfig).ValidateRecovery()` allowing `0` as a valid value for the `secret_shares` property
- relaxed the validation of the recovery config in the `init.go` function (based on the new `(*SealConfig).ValidateRecovery()` function
- changed the validation of the `(*SealConfig)` in `(*Core).PhysicalSealConfigs()` and `(*AutoSeal).RecoveryConfig()` for recovery seal configs from `Validate()` to `ValidateRecovery()`
- in case of server initialization with `0` `secret_shares` for recovery config, the call to the `/sys/rotate/recovery/init` now generates the recovery key and returns back the shares
- added entries to `PathsSpecial.Root` of the `SystemBackend`
```
"rotate/keyring",
"rotate/root",
```
- added entries to `restrictedSysAPIs` list in `vault/request_handling.go`:
```
"rotate/root",
"rotate/keyring",
"rotate/keyring/config",
```

-----
Follow up work:
- `sys/rotate/root/config` endpoint implementation
- documentation updates (new endpoints, deprecation marking)
- CLI command support

-----
cc: @cipherboy


http/handler.go
```diff
@@ -99,6 +99,7 @@ var (
 		"/v1/sys/mounts/",
 		"/v1/sys/policy",
 		"/v1/sys/policy/",
+		// TODO: verify
```
cipherboy: Given this note https://openbao.org/api-docs/system/rekey/#start-rekey that the `backup` parameter controls whether the below endpoint works, yes, I'd make that authenticated as well. We can always include it as part of the new `bao operator recovery ...` CLIs if desired.

vault/logical_system.go
```diff
@@ -86,6 +86,7 @@ func NewSystemBackend(core *Core, logger log.Logger) *SystemBackend {
 				"raw",
 				"raw/*",
 				"rotate",
+				"rotate/root",
```
cipherboy, line 90: `rotate/recovery` is missing I think?

vault/seal.go
```diff
@@ -329,14 +329,9 @@ type SealConfig struct {
...
-	if s.SecretShares < 1 {
-		return errors.New("shares must be at least one")
-	}
-	if s.SecretThreshold < 1 {
-		return errors.New("threshold must be at least one")
-	}
+// baseValidate is used as a shared base between `Validate` and `ValidateRecovery`
+// functions
+func (s *SealConfig) baseValidate() error {
 	if s.SecretShares > 1 && s.SecretThreshold == 1 {
```
cipherboy:
```suggestion
	if s.SecretShares > 1 && s.SecretThreshold < 2 {
```

(In particular, with the removal of `s.SecretThreshold < 1` check above, a recovery key could've come in and set shares > 1 but threshold == 0, which would be bad)

http/handler.go
wslabosz-reply: okay, then adding:
- `/v1/sys/rotate/root/backup`
- `/v1/sys/rotate/recovery/backup`
- `/v1/sys/rotate/keyring` (as an equivalent of `/v1/sys/rotate`)

vault/logical_system.go
wslabosz-reply: I think these three are missing:
```
"rotate/keyring",
"rotate/root/*",
"rotate/recovery/*"
```

cipherboy: Actually never mind, I misread this, I thought this was the no-namespace helper in core routing. Only the first should be added, the latter two we don't particularly want to be root due to their added authorization mechanism (providing shares).

command/operator_init_test.go
```diff
@@ -86,16 +86,17 @@ func TestOperatorInitCommand_Run(t *testing.T) {
...
 			},
-			"incorrect number",
+			"count mismatch between number of provided PGP keys and number of shares",
 			2,
 		},
 		{
 			"key_shares_pgp_more",
 			[]string{
 				"-key-shares", "1",
+				"-key-threshold", "1",
```
wslabosz-reply, line 96: had to add that due to changed order of validations (before the pgp_keys were double checked, with the first pass being before threshold)

vault/logical_system_rotate.go
```diff
@@ -0,0 +1,864 @@
...
+		},
+	}
+}
+
+// handleRotate handles the GET `/sys/rotate` and `/sys/rotate/keyring`
+// endpoints used to trigger an encryption key rotation.
+func (b *SystemBackend) handleRotate() framework.OperationFunc {
+	return func(ctx context.Context, _ *logical.Request, _ *framework.FieldData) (*logical.Response, error) {
+		if err := b.rotateBarrierKey(ctx); err != nil {
+			b.Backend.Logger().Error("error handling key rotation", "error", err)
```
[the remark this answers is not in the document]
wslabosz-reply: haven't really changed that, just moved the function to different file for navigable separation

@cipherboy wdyt, is this a valid alert?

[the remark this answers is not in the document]
cipherboy: @wslabosz-reply No, this has the same broken jump from api's output curl string to logical_system: 

image

I'm really sad that the recent CVEs weren't flagged by this last I looked and instead we get these terrible false positives.

wslabosz-reply:
@cipherboy 
due to `rekey.go` methods (which were the base for the old rekey and now new rotation ones) acquiring a lock over the core state, I've had to rewrite them, as the `switchedLockHandleRequest()` (`request_handling.go`) acquires the lock in the first place.

api/sys_rotate.go
```diff
@@ -9,23 +9,331 @@ import (
 	"errors"
 	"net/http"
 	"time"
+
+	"github.com/go-viper/mapstructure/v2"
 )
 
+func (c *Sys) RotateStatus(recovery bool) (*RekeyStatusResponse, error) {
```
cipherboy: Should these just be separate methods? `RotateRecovery` vs `RotateRoot`? &c Since we're renaming to `RotateKeyring` that might be nice from a consistency PoV.

vault/logical_system.go
```diff
@@ -5586,7 +5514,27 @@ Enable a new audit backend or disable an existing backend.
...
+		"",
+	},
+	"rotate_update": {
+		"Progress the rotation process by providing a single key share.",
+		`This endpoint is used to enter a single key share to progress the
+		rotation of the recovery or root key. If the threshold number of key
+		shares is reached, rotation will be completed. Otherwise, this API
+		must be called multiple times until that threshold is met.
+		The rotation nonce operation must be provided with each call.
+		`,
```
cipherboy, line 5529:
```suggestion
		On the final call, any new key shares will be returned immediately.
		`,
```

vault/logical_system_rotate.go
```diff
@@ -0,0 +1,884 @@
...
+		verificationReq, ok, err := data.GetOkErr("require_verification")
+		if err != nil {
+			return handleError(err)
+		}
+		if ok {
+			rotateConf.VerificationRequired = verificationReq.(bool)
+		}
+
+		recovery := strings.Contains(req.Path, "recovery")
+		if err := b.Core.InitRotation(rotateConf, recovery); err != nil {
```
cipherboy: Just curious why this is different from `RekeyInit`? I know you mentioned that the statelock was grabbed later in the variant called from `http/sys_rekey.go`, but I think some of the validation logic between `BarrierRekeyInit` and `initBarrierRotation` could be synced, perhaps with a helper with state lock grabbing? My 2c.

cipherboy (commented):
Some initial comments.

vault/logical_system_rotate.go
wslabosz-reply: > Just curious why this is different from RekeyInit?

exactly due to `RekeyInit()` invoking either `BarrierRekeyInit` or `RecoveryRekeyInit` which both do acquire a lock over `c.stateLock`

also now it is returning back a recovery key shares in the specific case the recovery config was initialized with `0` as a `secret_shares` value

api/sys_rotate.go
wslabosz-reply: splitting introduces unnecessary duplication imho, also don't see the consistency due to `RotateKeyring` introduction, old `Rotate` also exists, although I've marked it as deprecated, which I probably shouldn't as it is just an alias to the `RotateKeyring`, will adjust.

vault/logical_system_rotate.go
cipherboy: Ah, the latter is a bigger issue. I was thinking we could adjust the locking to be more global like the other handler does implicitly and have better reuse here. The broader locking shouldn't be much of a problem here, IMHO.

wslabosz-reply: @cipherboy 
to be honest, I don't think I follow:
> I was thinking we could adjust the locking to be more global like the other handler does implicitly and have better reuse here.

but we don't have to acquire a statelock (of core) as it is handled by `request_handler.go`
If it is about the `rekeyLock` then I don't think we have that much of a "reuse" possibilities.

api/sys_rotate.go
cipherboy: Hmm, given we're naming the endpoints explicitly (yes, `/rotate` exists but `/rotate/keyring` should be used), I figured the API forms of all of them (`/rotate/keyring` -> `RotateKeyring(...)`) should match the actual endpoints. I don't know that _many_ clients will want to rotate either the root key or the recovery keys on a boolean basis. In most (auto-unsealed) environments, they're two separate sets of keys with two separate sets of semantics and so you wouldn't _typically_ write the same rotation logic.

(Importantly, root key recovery doesn't return shares and can optionally be done without requiring shares be inputted whereas recovery key rotation does.)

My 2c.

cipherboy: (Like, my critiques are from an API design PoV; I'm happy if you keep the internal implementation the same.)

api/sys_rotate.go
```diff
@@ -9,23 +9,331 @@ import (
...
+	r := c.c.NewRequest(http.MethodDelete, requestPath)
+
+	resp, err := c.c.rawRequestWithContext(ctx, r)
+	if err != nil {
+		return err
+	}
+	return resp.Body.Close()
+}
+
+// Alias of RotateKeyring function.
```
cipherboy: I'd still keep these deprecated for reasons above, my 2c.

vault/logical_system_rotate.go
```diff
@@ -0,0 +1,908 @@
...
+// handleRotateUpdate handles the POST `/sys/rotate/root/update` and `/sys/rotate/recovery/update`
+// endpoints used for providing a single root key share progressing the rotation of the key.
+func (b *SystemBackend) handleRotateUpdate() framework.OperationFunc {
+	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+		ikey, ok, err := data.GetOkErr("key")
+		if err != nil {
+			return handleError(err)
+		}
+		if !ok || ikey.(string) == "" {
+			return handleError(errors.New("'key' must be specified in request body as JSON"))
```
cipherboy: nit:

```suggestion
			return handleError(errors.New("missing required field 'key'"))
```

&c.

I suppose they are JSON request bodies, but this is more inline with other parameters. My 2c.

cipherboy: I also think `GetOk(...)` is _probably_ fine, but this is OK too.

cipherboy (approved):
I'll likely attempt a follow-up PR when I go to add the CLI helper, but I think this is fine.

cipherboy (approved):
I think this looks fairly good, thanks @wslabosz-reply! 

Does anyone else from @openbao/openbao-core-committers want to take a look before we merge?

rencooo (approved):

cipherboy:
Merging based on reviews here. Thank you @wslabosz-reply!

cipherboy:
@wslabosz-reply Would you mind adding documentation of these endpoints to the https://openbao.org/api-docs/system/rotate/ page?

I think we could unify many of the other pages (https://openbao.org/api-docs/system/rekey-recovery-key/, https://openbao.org/api-docs/system/rekey/, &c) into a single one. 

Thanks!

wslabosz-reply:
@cipherboy sure, will get to it this week.

cipherboy:
@wslabosz-reply If not, no worries, just let me know :-)

PR #1519, opened 2025-07-01
Add missing sidebar entry for Static Key seal doc page

karras:
Adds the missing sidebar entry for the recent Static Key seal functionality.

Related to #1425


cipherboy (approved):
Ah, good catch. Thanks @karras!

satoqz (approved):

PR #1521, opened 2025-07-02
Cleanup of redundant code

phyrog:
Removes and simplifies some redundant or unnecessarily complicated code. I recommend reviewing this commit by commit.

I'm adding a couple of line comments for places I think could be simplified further, but I want some feedback on first.

### DIFF
--- a/builtin/credential/cert/path_crls.go
+++ b/builtin/credential/cert/path_crls.go
@@ -137,7 +137,7 @@ func (b *backend) populateCRLs(ctx context.Context, storage logical.Storage) err
 	if err != nil {
 		return fmt.Errorf("error listing CRLs: %w", err)
 	}
-	if keys == nil || len(keys) == 0 {
+	if len(keys) == 0 {
 		return nil
 	}
 
--- a/physical/raft/raft.go
+++ b/physical/raft/raft.go
@@ -916,9 +916,9 @@ func (b *RaftBackend) SetupCluster(ctx context.Context, opts SetupOpts) error {
 			return true
 		default:
 			// Concrete type checks
-			switch cl.(type) {
+			switch listener := cl.(type) {
 			case *cluster.Listener:
-				return cl.(*cluster.Listener) == nil
+				return listener == nil
 			}
 		}
 		return false
--- a/sdk/database/dbplugin/client.go
+++ b/sdk/database/dbplugin/client.go
@@ -76,9 +76,9 @@ func NewPluginClient(ctx context.Context, sys pluginutil.RunnerUtil, pluginRunne
 	// We should have a database type now. This feels like a normal interface
 	// implementation but is in fact over an RPC connection.
 	var db Database
-	switch raw.(type) {
+	switch d := raw.(type) {
 	case *gRPCClient:
-		db = raw.(*gRPCClient)
+		db = d
 	default:
 		return nil, errors.New("unsupported client type")
 	}
--- a/sdk/logical/storage.go
+++ b/sdk/logical/storage.go
@@ -220,11 +220,7 @@ func ClearViewWithLogging(ctx context.Context, view ClearableView, logger hclog.
 		logger = hclog.NewNullLogger()
 	}
 
-	if paginated, ok := view.(ClearableView); ok {
-		return ClearViewWithPagination(ctx, paginated, logger)
-	}
-
-	return ClearViewWithoutPagination(ctx, view, logger)
+	return ClearViewWithPagination(ctx, view, logger)
 }
 
 func ClearViewWithPagination(ctx context.Context, view ClearableView, logger hclog.Logger) error {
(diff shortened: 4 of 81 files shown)


sdk/logical/storage.go
```diff
@@ -220,11 +220,7 @@ func ClearViewWithLogging(ctx context.Context, view ClearableView, logger hclog.
 		logger = hclog.NewNullLogger()
 	}
 
-	if paginated, ok := view.(ClearableView); ok {
-		return ClearViewWithPagination(ctx, paginated, logger)
-	}
-
-	return ClearViewWithoutPagination(ctx, view, logger)
```
phyrog, line 227: @cipherboy `ClearViewWithoutPagination` was never used with this code, so I removed it completely. I do not know if that is the intended outcome, or if the check here was just wrong.

wslabosz-reply (approved):
LGTM

sdk/database/dbplugin/client.go
```diff
@@ -76,9 +76,9 @@ func NewPluginClient(ctx context.Context, sys pluginutil.RunnerUtil, pluginRunne
 	// We should have a database type now. This feels like a normal interface
 	// implementation but is in fact over an RPC connection.
 	var db Database
-	switch raw.(type) {
+	switch d := raw.(type) {
 	case *gRPCClient:
-		db = raw.(*gRPCClient)
+		db = d
 	default:
 		return nil, errors.New("unsupported client type")
```
phyrog, line 83: We might as well just do

```go
	db, ok := raw.(Database)
	if !ok {
		return nil, errors.New("unsupported client type")
	}
```

physical/raft/raft.go
```diff
@@ -916,9 +916,9 @@ func (b *RaftBackend) SetupCluster(ctx context.Context, opts SetupOpts) error {
 			return true
 		default:
 			// Concrete type checks
-			switch cl.(type) {
+			switch listener := cl.(type) {
```
phyrog, line 919: All of this could also become:

```go
	listenerIsNil := func(cl cluster.ClusterHook) bool {
		if opts.ClusterListener == nil {
			return true
		}
		if listener, ok := cl.(*cluster.Listener); ok {
			return listener == nil
		}
		return false
	}
```

builtin/credential/cert/path_crls.go
```diff
@@ -360,7 +360,7 @@ Manage Certificate Revocation Lists checked during authentication.
 
 const pathCRLsHelpDesc = `
 This endpoint allows you to list, create, read, update, and delete the Certificate
-Revocation Lists checked during authentication, and/or CRL Distribution Point 
+Revocation Lists checked during authentication, and/or CRL Distribution Point
```
cipherboy: This space is structural as otherwise the next line will be formatted weird I believe. Feel free to adjust globally in another PR if you want to fix this problem, you can see these with `bao path-help some/path/`.

sdk/logical/storage.go
cipherboy: @phyrog I think theoretically this is reachable if you build an external plugin with the OpenBao SDK and run on HashiCorp Vault.

phyrog: `view` is already of type `ClearableView` (as per function signature), so the cast here will always succeed and `ClearViewWithoutPagination` will never be called.

I did not find any other references to `ClearViewWithoutPagination` outside of tests. Am I missing something?

phyrog: Ohh I think I got what you mean. A third party uses the function directly in their code via the sdk import. Gotcha. I'll revert the deletion in that case.

builtin/credential/cert/path_crls.go
phyrog: This was automatically removed by gofmt. I'll revert these here and leave that for another PR then.

cipherboy: Does `gofumpt` behave differently? Otherwise our pipeline job would fail. We don't use `gofmt`

sdk/logical/storage.go
cipherboy: Right, exactly.

builtin/credential/cert/path_crls.go
phyrog: No you're right, it probably was my editor. Running gofmt manually did not change the spaces.

sdk/database/dbplugin/client.go
phyrog: I'll keep this as is for now, to not accidentally introduce any functional changes.

physical/raft/raft.go
phyrog: I'll keep this as is for now, to not accidentally introduce any functional changes.

phyrog:
Changes to strings are reverted and `ClearViewWithoutPagination` is restored (but unused by our code now).

If there are no objections, I'll merge this later.

driif (approved):
Thank you, @phyrog . You’re doing a great job keeping the project maintainable!

PR #1523, opened 2025-07-03
Fix potential problems found by staticcheck/gopls

phyrog:
This fixes a range of issues raised by gopls/staticcheck. While most were basically a no-op (always true/false conditions, ...), some I think were legit bugs or could lead to bugs in the future (I'll add line comments to those, to make sure I understood the intent of the code correctly).

I recommend to review commit by commit.

### DIFF
--- a/builtin/logical/database/mockv4.go
+++ b/builtin/logical/database/mockv4.go
@@ -92,7 +92,7 @@ func (m MockDatabaseV4) RotateRootCredentials(ctx context.Context, statements []
 	if err != nil {
 		return config, fmt.Errorf("failed to generate credentials: %w", err)
 	}
-	config["password"] = newPassword
+	m.config["password"] = newPassword
 
 	return m.config, nil
 }
--- a/builtin/logical/pki/acme_state_test.go
+++ b/builtin/logical/pki/acme_state_test.go
@@ -29,6 +29,7 @@ func TestAcmeNonces(t *testing.T) {
 		nonce, _, err = a.GetNonce()
 		require.NoError(t, err)
 		require.NotEmpty(t, nonce)
+		nonces = append(nonces, nonce)
 	}
 
 	for i := len(nonces) - 1; i >= 0; i-- {
--- a/http/handler.go
+++ b/http/handler.go
@@ -305,9 +305,7 @@ func handleAuditNonLogical(core *vault.Core, h http.Handler) http.Handler {
 			respondError(w, status, err)
 			return
 		}
-		if origBody != nil {
-			r.Body = io.NopCloser(origBody)
-		}
+		r.Body = io.NopCloser(origBody)
 		input := &logical.LogInput{
 			Request: req,
 		}
--- a/http/handler.go
+++ b/http/handler.go
@@ -731,13 +731,9 @@ func parseJSONRequest(r *http.Request, w http.ResponseWriter, out interface{}) (
 	// Limit the maximum number of bytes to MaxRequestSize to protect
 	// against an indefinite amount of data being read.
 	reader := r.Body
-	var origBody io.ReadWriter
 	err := jsonutil.DecodeJSONFromReader(reader, out)
 	if err != nil && err != io.EOF {
 		return nil, fmt.Errorf("failed to parse JSON input: %w", err)
-	}
-	if origBody != nil {
-		return io.NopCloser(origBody), err
 	}
 	return nil, err
 }
--- a/physical/postgresql/postgresql_test.go
+++ b/physical/postgresql/postgresql_test.go
@@ -586,7 +586,7 @@ func TestPostgreSQLBackend_Parallel(t *testing.T) {
 				errors[i] = fmt.Errorf("value for job %v exceeded max_parallel: %v", i, value)
 			}
 
-			time.Sleep(1)
+			time.Sleep(1 * time.Second)
 
 			value = count.Load()
 			if value > 2 {
@@ -604,7 +604,7 @@ func TestPostgreSQLBackend_Parallel(t *testing.T) {
 				errors[i] = fmt.Errorf("value for job %v exceeded max_parallel: %v", i, value)
 			}
 
-			time.Sleep(1)
+			time.Sleep(1 * time.Second)
 
 			value = count.Load()
 			if value > 2 {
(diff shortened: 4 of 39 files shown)


builtin/logical/database/mockv4.go
```diff
@@ -92,7 +92,7 @@ func (m MockDatabaseV4) RotateRootCredentials(ctx context.Context, statements []
 	if err != nil {
 		return config, fmt.Errorf("failed to generate credentials: %w", err)
 	}
-	config["password"] = newPassword
+	m.config["password"] = newPassword
```
phyrog, line 95: `config` is always nil at this point, so I assume the intent was to assign the password to `m.config` instead.

builtin/logical/pki/acme_state_test.go
```diff
@@ -29,6 +29,7 @@ func TestAcmeNonces(t *testing.T) {
 		nonce, _, err = a.GetNonce()
 		require.NoError(t, err)
 		require.NotEmpty(t, nonce)
+		nonces = append(nonces, nonce)
```
phyrog, line 32: I am not 100% this is what was intended here, but it's my best guess. Otherwise `nonces` is always empty and the code below is a no-op.

@cipherboy in case you remember what you wanted to do here 2 years ago 😄

physical/postgresql/postgresql_test.go
```diff
@@ -586,7 +586,7 @@ func TestPostgreSQLBackend_Parallel(t *testing.T) {
 				errors[i] = fmt.Errorf("value for job %v exceeded max_parallel: %v", i, value)
 			}
 
-			time.Sleep(1)
+			time.Sleep(1 * time.Second)
```
phyrog, line 589: I assume this was supposed to be 1s and not 1ns.

sdk/helper/ocsp/client.go
```diff
@@ -549,7 +550,7 @@ func (c *Client) GetRevocationStatus(ctx context.Context, subject, issuer *x509.
 			case ocspStatusRevoked:
 				ret = ocspStatuses[i]
 				ocspRes = ocspResponses[i]
-				break
+				break LOOP
```
phyrog, line 553: I think the intent here was to break out of the loop and not the switch.

wslabosz-reply (approved):
looks good

vault/identity_store_util.go
```diff
@@ -2442,7 +2442,7 @@ func (i *IdentityStore) countEntities(ctx context.Context) (int, error) {
 
 		iter, err := txn.Get(entitiesTable, "id")
 		if err != nil {
-			err = fmt.Errorf("failed to get entities table for namespace %v: %w", uuid, err)
+			outErr = fmt.Errorf("failed to get entities table for namespace %v: %w", uuid, err)
```
phyrog, line 2445: `err` was local to the inner function and was thus unused, I think it was meant to be passed to the outer function.

phyrog:
CodeQL findings are the usual false positives.

builtin/logical/pki/acme_state_test.go
cipherboy: Ah, yes, good catch.

sdk/helper/ocsp/client.go
cipherboy: Yes, it must be because switch statements are single-shot anyways.

http/handler.go
```diff
@@ -731,14 +731,10 @@ func parseJSONRequest(r *http.Request, w http.ResponseWriter, out interface{}) (
 	// Limit the maximum number of bytes to MaxRequestSize to protect
 	// against an indefinite amount of data being read.
 	reader := r.Body
-	var origBody io.ReadWriter
```
cipherboy, line 736: Is this a problem? The comment above seems to be ignored here.

vault/request_handling.go
```diff
@@ -667,7 +667,7 @@ func (c *Core) handleInlineAuth(ctx context.Context, req *logical.Request, nsHea
 	// Find the optional operation; this defaults to Update if missing.
 	authOperation, present := req.Headers[consts.InlineAuthOperationHeaderName]
 	if !present {
-		authOperation = []string{logical.UpdateOperation}
+		authOperation = []string{string(logical.UpdateOperation)}
```
cipherboy, line 670: Wondered how this typing worked :laughing:

sdk/logical/request.go
```diff
@@ -364,21 +364,21 @@ type Operation string
...
-	PatchOperation                    = "patch"
-	DeleteOperation                   = "delete"
-	ListOperation                     = "list"
-	ScanOperation                     = "scan"
-	HelpOperation                     = "help"
-	AliasLookaheadOperation           = "alias-lookahead"
-	ResolveRoleOperation              = "resolve-role"
-	HeaderOperation                   = "header"
+	ReadOperation           Operation = "read"
+	UpdateOperation         Operation = "update"
```
cipherboy, line 368: Strictly I think this is a breaking change but that's fine.

http/handler.go
cipherboy: @phyrog I think this is the only question I have -- I'm happy splitting to a separate issue after investigation but should probably understand it before we do the removal and merge this one.

sdk/logical/request.go
phyrog: Oof, you're right, I hadn't thought about that 😬

http/handler.go
phyrog: I guess we can just wrap body in an io.LimitedReader to comply with the comment. But I think the removal here is unrelated to the comment as it is now (maybe there used to be more code that handled that there).

sdk/logical/request.go
phyrog: Should I write a change log entry for that?

Edit: wrote one

http/handler.go
phyrog: @cipherboy Request size limit should be handled by the `wrapMaxRequestSizeHandler` middleware, though I have not checked yet if this is called before `parseJSONRequest` in all cases, or if there are edge cases where it is not (recovery mode maybe?).

Edit: recovery mode should be fine as well

cipherboy: @phyrog I see, it was refactored in 446f213c47cabf47d52d065647ef666ce4bf8692 but the comment was left alone. We can drop this comment I think.

cipherboy (approved):
Thanks @phyrog!

PR #1525, opened 2025-07-03
Fix race condition with sealing in namespace deletion operation

### LINKED ISSUE #1524: Namespace deletion cleanup process race
**Describe the bug**
Namespace deletion "cleanup" goroutine can encounter a panic when the `namespaceStore` gets tear-downed, this happens in case of the sealing of bao instance. For now the issue was observed in tests, which fail with the following trace:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x30a9974]

goroutine 2502947 [running]:
github.com/openbao/openbao/vault.(*NamespaceStore).checkInvalidation(0xc0200a7d28?, {0x49f0c10?, 0xc02e71ccf0?})
	/home/runner/work/openbao/openbao/vault/namespace_store.go:89 +0x34
github.com/openbao/openbao/vault.(*NamespaceStore).GetNamespace(0x0, {0x49f0c10, 0xc02e71ccf0}, {0xc0343c3710, 0x24})
	/home/runner/work/openbao/openbao/vault/namespace_store.go:443 +0xb7
github.com/openbao/openbao/vault.(*Core).runLockedUserEntryUpdatesForNamespace(0xc02007c708, {0x49f0c10, 0xc02e71ccf0}, 0xc0341cb800)
	/home/runner/work/openbao/openbao/vault/core.go:3392 +0xf2
github.com/openbao/openbao/vault.(*NamespaceStore).clearNamespaceResources(0xc03486e240, {0x49f0c10, 0xc02e71ccf0}, 0xc0341cb800)
	/home/runner/work/openbao/openbao/vault/namespace_store.go:789 +0x127c
created by github.com/openbao/openbao/vault.(*NamespaceStore).DeleteNamespace in goroutine 2447342
	/home/runner/work/openbao/openbao/vault/namespace_store.go:690 +0x414
```

(found by @satoqz [here](https://github.com/openbao/openbao/pull/1522#issuecomment-3029109839))

**To Reproduce**
1. Create a namespace
2. Delete a namespace
3. Seal the bao instance - this step happens automatically with the unit test finishing

For this panic to happen, the sealing operation would have to teardown the `namespaceStore` before the cleanup goroutine finishes. It's unknown how frequently can this panic occur.

**Expected behavior**
The goroutine gracefully errors, or the deletion process completes before the sealing.

**Additional context**
One of the possible solutions proposed by @satoqz is for the `namespaceStore` to hold a read-only lock over `Core.stateLock`, but that would mean the namespace deletion operation would take precedence over the bao instance sealing, which is considered not ideal.

wslabosz-reply:
Resolves #1524

### DIFF
--- a/vault/core.go
+++ b/vault/core.go
@@ -2496,8 +2496,8 @@ func (c *Core) preSeal() error {
 	var result error
 
 	c.stopForwarding()
-
 	c.stopRaftActiveNode()
+	c.cancelNamespaceDeletion()
 
 	if err := c.teardownAudits(); err != nil {
 		result = multierror.Append(result, fmt.Errorf("error tearing down audits: %w", err))
@@ -3365,7 +3365,7 @@ func (c *Core) runLockedUserEntryUpdates(ctx context.Context) error {
 
 	totalLockedUsersCount := 0
 	for _, ns := range nsList {
-		nsLockedUsers, err := c.runLockedUserEntryUpdatesForNamespace(ctx, ns)
+		nsLockedUsers, err := c.runLockedUserEntryUpdatesForNamespace(ctx, ns, false)
 		if err != nil {
 			return err
 		}
@@ -3378,16 +3378,11 @@ func (c *Core) runLockedUserEntryUpdates(ctx context.Context) error {
 }
 
 // runLockedUserEntryUpdatesForNamespace runs updates for locked users storage entries
-// for a single namespace if the specified namespace was deleted, all login entries are also deleted.
-func (c *Core) runLockedUserEntryUpdatesForNamespace(ctx context.Context, namespace *namespace.Namespace) (int, error) {
+// for a single namespace. If a forceDelete flag is passed all login entries are deleted.
+func (c *Core) runLockedUserEntryUpdatesForNamespace(ctx context.Context, namespace *namespace.Namespace, forceDelete bool) (int, error) {
 	// get the list of mount accessors of locked users of a namespace
 	view := NamespaceView(c.barrier, namespace).SubView(coreLockedUsersPath)
 	mountAccessors, err := view.List(ctx, "")
-	if err != nil {
-		return 0, err
-	}
-
-	ns, err := c.namespaceStore.GetNamespace(ctx, namespace.UUID)
 	if err != nil {
 		return 0, err
 	}
@@ -3400,7 +3395,7 @@ func (c *Core) runLockedUserEntryUpdatesForNamespace(ctx context.Context, namesp
 	// the userFailedLoginInfo map has correct failed login information
 	// if incorrect, update the entry in userFailedLoginInfo map
 	for _, mountAccessorPath := range mountAccessors {
-		lockedAliasesCount, err := c.runLockedUserEntryUpdatesForMountAccessor(ctx, view.SubView(mountAccessorPath), mountAccessorPath, ns == nil)
+		lockedAliasesCount, err := c.runLockedUserEntryUpdatesForMountAccessor(ctx, view.SubView(mountAccessorPath), mountAccessorPath, forceDelete)
 		if err != nil {
 			return namespaceLockedUsers, err
 		}
@@ -3439,59 +3434,63 @@ func (c *Core) runLockedUserEntryUpdatesForMountAccessor(ctx context.Context, vi
 			aliasName:     alias,
 			mountAccessor: mountAccessor,
 		}
-
-		existingEntry, err := view.Get(ctx, alias)
-		if err != nil {
-			return 0, err
-		}
-
-		if existingEntry == nil {
-			continue
-		}
-
-		var lastLoginTime int
-		err = jsonutil.DecodeJSON(existingEntry.Value, &lastLoginTime)
-		if err != nil {
-			return 0, err
-		}
-
-		lastFailedLoginTimeFromStorageEntry := time.Unix(int64(lastLoginTime), 0)
-		lockoutDurationFromConfiguration := userLockoutConfiguration.LockoutDuration
-
 		// get the entry for the locked user from userFailedLoginInfo map
 		failedLoginInfoFromMap := c.LocalGetUserFailedLoginInfo(ctx, loginUserInfoKey)
 
-		// check if the storage entry for locked user is stale
+		var staleEntry bool
+		// if not forcing the delete, check if the entry is stale
+		if !forceDelete {
+			existingEntry, err := view.Get(ctx, alias)
+			if err != nil {
+				return 0, err
+			}
+
+			if existingEntry == nil {
+				continue
+			}
+
+			var lastLoginTime int
+			err = jsonutil.DecodeJSON(existingEntry.Value, &lastLoginTime)
+			if err != nil {
+				return 0, err
+			}
+
+			lastFailedLoginTimeFromStorageEntry := time.Unix(int64(lastLoginTime), 0)
+			lockoutDurationFromConfiguration := userLockoutConfiguration.LockoutDuration
+			staleEntry = time.Now().After(lastFailedLoginTimeFromStorageEntry.Add(lockoutDurationFromConfiguration))
+			if !staleEntry {
+				// not a stale entry
+				// update the map with actual failed login information
+				actualFailedLoginInfo := FailedLoginInfo{
+					lastFailedLoginTime: lastLoginTime,
+					count:               uint(userLockoutConfiguration.LockoutThreshold),
+				}
+
+				if failedLoginInfoFromMap != &actualFailedLoginInfo {
+					// outdated entry, updating the entry in userFailedLoginMap with correct information
+					if err = c.LocalUpdateUserFailedLoginInfo(ctx, loginUserInfoKey, &actualFailedLoginInfo, false); err != nil {
+						return 0, err
+					}
+				}
+			}
+		}
+
+		// delete if the storage entry for locked user is stale
 		// or force delete if the namespace is being deleted
-		if time.Now().After(lastFailedLoginTimeFromStorageEntry.Add(lockoutDurationFromConfiguration)) || forceDelete {
+		if staleEntry || forceDelete {
 			// stale entry, remove from storage
 			// leaving this as it is as this happens on the active node
-			// also handles case where namespace is deleted
+			// also handles case where entry is force deleted
 			if err := view.Delete(ctx, alias); err != nil {
 				return 0, err
 			}
-			// remove entry for this user from userFailedLoginInfo map if present as the user is not locked
-			if failedLoginInfoFromMap != nil {
-				if err = c.LocalUpdateUserFailedLoginInfo(ctx, loginUserInfoKey, nil, true); err != nil {
-					return 0, err
-				}
-			}
-			lockedAliasesCount -= 1
-			continue
-		}
-
-		// this is not a stale entry
-		// update the map with actual failed login information
-		actualFailedLoginInfo := FailedLoginInfo{
-			lastFailedLoginTime: lastLoginTime,
-			count:               uint(userLockoutConfiguration.LockoutThreshold),
-		}
-
-		if failedLoginInfoFromMap != &actualFailedLoginInfo {
-			// entry is invalid, updating the entry in userFailedLoginMap with correct information
-			if err = c.LocalUpdateUserFailedLoginInfo(ctx, loginUserInfoKey, &actualFailedLoginInfo, false); err != nil {
+			// remove entry for this user from userFailedLoginInfo map
+			// as the user is not locked (if not present, this is no-op)
+			if err = c.LocalUpdateUserFailedLoginInfo(ctx, loginUserInfoKey, failedLoginInfoFromMap, true); err != nil {
 				return 0, err
 			}
+
+			lockedAliasesCount -= 1
 		}
 	}
 	return lockedAliasesCount, nil
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -60,6 +60,10 @@ type NamespaceStore struct {
 	namespacesByUUID     map[string]*namespace.Namespace
 	namespacesByAccessor map[string]*namespace.Namespace
 
+	deletionJobGroup      sync.WaitGroup
+	deletionJobContext    context.Context
+	deletionJobCancelFunc context.CancelFunc
+
 	// logger is the server logger copied over from core
 	logger hclog.Logger
 }
@@ -76,6 +80,8 @@ func NewNamespaceStore(ctx context.Context, core *Core, logger hclog.Logger) (*N
 		namespacesByAccessor: make(map[string]*namespace.Namespace),
 	}
 
+	ns.deletionJobContext, ns.deletionJobCancelFunc = context.WithCancel(core.activeContext)
+
 	// Add namespaces from storage to our table. We can do this without
 	// holding a lock as we've not returned ns to anyone yet.
 	if err := ns.loadNamespacesLocked(ctx); err != nil {
@@ -83,6 +89,11 @@ func NewNamespaceStore(ctx context.Context, core *Core, logger hclog.Logger) (*N
 	}
 
 	return ns, nil
+}
+
+func (c *Core) cancelNamespaceDeletion() {
+	c.namespaceStore.deletionJobCancelFunc()
+	c.namespaceStore.deletionJobGroup.Wait()
 }
 
 func (ns *NamespaceStore) checkInvalidation(ctx context.Context) error {
@@ -686,13 +697,16 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
 		return "", fmt.Errorf("error loading parent namespace from context: %w", err)
 	}
 
-	quitCtx := namespace.ContextWithNamespace(ns.core.activeContext, parent)
+	// quitCtx := namespace.ContextWithNamespace(ns.core.activeContext, parent)
+	quitCtx := namespace.ContextWithNamespace(ns.deletionJobContext, parent)
+	ns.deletionJobGroup.Add(1)
 	go ns.clearNamespaceResources(quitCtx, namespaceToDelete)
 
 	return "in-progress", nil
 }
 
 func (ns *NamespaceStore) clearNamespaceResources(ctx context.Context, namespaceToDelete *namespace.Namespace) {
+	defer ns.deletionJobGroup.Done()
 	nsCtx := namespace.ContextWithNamespace(ctx, namespaceToDelete)
 	// clear ACL policies
 	policiesToClear, err := ns.core.policyStore.ListPolicies(nsCtx, PolicyTypeACL, false)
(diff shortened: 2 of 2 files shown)


vault/namespace_store.go
```diff
@@ -686,13 +697,16 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
 		return "", fmt.Errorf("error loading parent namespace from context: %w", err)
 	}
 
-	quitCtx := namespace.ContextWithNamespace(ns.core.activeContext, parent)
+	// quitCtx := namespace.ContextWithNamespace(ns.core.activeContext, parent)
```
satoqz: Stale comment
```suggestion
```

vault/core.go
```diff
@@ -2496,8 +2496,8 @@ func (c *Core) preSeal() error {
 	var result error
 
 	c.stopForwarding()
-
 	c.stopRaftActiveNode()
+	c.cancelNamespaceDeletion()
```
cipherboy, line 2500: Ah, so the way it works in mounts is that:

```go
func (c *Core) unloadMounts(ctx context.Context) error {
    c.mountsLock.Lock()
    defer c.mountsLock.Unlock()
```

plus:

```
// disableAudit is used to disable an existing audit backend
func (c *Core) disableAudit(ctx context.Context, path string, updateStorage bool) (bool, error) {
    // Ensure we end the path in a slash
    if !strings.HasSuffix(path, "/") {
        path += "/"
    }

    // Ensure there is a name
    if path == "/" {
        return false, errors.New("backend path must be specified")
    }

    // Remove the entry from the mount table
    c.auditLock.Lock()
    defer c.auditLock.Unlock()
```

We can't use this as we want to release our namespace lock during deletion of a namespace, so I think this is probably fine.

cipherboy:
@wslabosz-reply Actually, there seems to be a bug:


```

goroutine 59480 [running]:
github.com/openbao/openbao/vault.(*Core).cancelNamespaceDeletion(0xc002dac308)
	/home/runner/work/openbao/openbao/vault/namespace_store.go:95 +0x1a
github.com/openbao/openbao/vault.(*Core).preSeal(0xc002dac308)
	/home/runner/work/openbao/openbao/vault/core.go:2500 +0x1a5
github.com/openbao/openbao/vault.(*Core).waitForLeadership(0xc002dac308, 0x0?, 0xc002a46a80, 0xc00250d810)
	/home/runner/work/openbao/openbao/vault/ha.go:692 +0x1265
github.com/openbao/openbao/vault.(*Core).runStandby.func9()
	/home/runner/work/openbao/openbao/vault/ha.go:448 +0x25
github.com/oklog/run.(*Group).Run.func1({0xc002d487e0?, 0xc00418e900?})
	/home/runner/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:38 +0x29
created by github.com/oklog/run.(*Group).Run in goroutine 59440
	/home/runner/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:37 +0x5a
```

satoqz:
> @wslabosz-reply Actually, there seems to be a bug:
> 
> ```
> 
> goroutine 59480 [running]:
> github.com/openbao/openbao/vault.(*Core).cancelNamespaceDeletion(0xc002dac308)
> 	/home/runner/work/openbao/openbao/vault/namespace_store.go:95 +0x1a
> github.com/openbao/openbao/vault.(*Core).preSeal(0xc002dac308)
> 	/home/runner/work/openbao/openbao/vault/core.go:2500 +0x1a5
> github.com/openbao/openbao/vault.(*Core).waitForLeadership(0xc002dac308, 0x0?, 0xc002a46a80, 0xc00250d810)
> 	/home/runner/work/openbao/openbao/vault/ha.go:692 +0x1265
> github.com/openbao/openbao/vault.(*Core).runStandby.func9()
> 	/home/runner/work/openbao/openbao/vault/ha.go:448 +0x25
> github.com/oklog/run.(*Group).Run.func1({0xc002d487e0?, 0xc00418e900?})
> 	/home/runner/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:38 +0x29
> created by github.com/oklog/run.(*Group).Run in goroutine 59440
> 	/home/runner/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:37 +0x5a
> ```

Right, `namespaceStore` may already be `nil` at this point and `preSeal` is expected to be idempotent-ish, so just return if it's not there.

changelog/1525.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+core/namespaces: fixed race condition in namespace deletion operation
```
cipherboy:
```suggestion
core/namespaces: fixed race condition in namespace deletion operation during instance sealing
```

vault/namespace_store.go
```diff
@@ -686,13 +704,22 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
...
 	return "in-progress", nil
 }
 
 func (ns *NamespaceStore) clearNamespaceResources(ctx context.Context, namespaceToDelete *namespace.Namespace) {
+	succeded := false
+	defer func() {
+		if !succeded {
+			ns.deletionMap[namespaceToDelete.UUID] = false
+		}
+	}()
```
satoqz: - No need to mark as `false` just `delete`
- I think you'll want to do this regardless of whether we fail or not

```suggestion
	defer delete(ns.deletionMap, namespaceToDelete.UUID)
```

satoqz: We also can't be sure we hold `ns.lock` here, I just realized 😄

satoqz: Suggestion, bit of a refactor to decouple the parts of `clearNamespaceResources` where we need locking vs the ones where we don't:

<details>

<summary>Patch:</summary>

```diff
diff --git a/vault/namespace_store.go b/vault/namespace_store.go
index 81e2da57e8..815e302cce 100644
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -642,7 +642,6 @@ func (ns *NamespaceStore) taintNamespace(ctx context.Context, namespaceToTaint *
 	ns.namespacesByUUID[namespaceToTaint.UUID].Tainted = true
 	ns.namespacesByAccessor[namespaceToTaint.ID].Tainted = true
 	namespaceToTaint.Tainted = true
-	ns.deletionMap[namespaceToTaint.UUID] = true
 	err := ns.namespacesByPath.Insert(namespaceToTaint)
 	if err != nil {
 		return fmt.Errorf("failed to modify namespace tree: %w", err)
@@ -704,114 +703,112 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
 		return "", fmt.Errorf("error loading parent namespace from context: %w", err)
 	}
 
-	quitCtx := namespace.ContextWithNamespace(ns.deletionJobContext, parent)
 	ns.deletionJobGroup.Add(1)
-	go ns.clearNamespaceResources(quitCtx, namespaceToDelete)
+	ns.deletionMap[namespaceToDelete.UUID] = true
+	go func() {
+		ctx := namespace.ContextWithNamespace(ns.deletionJobContext, namespaceToDelete)
 
-	return "in-progress", nil
-}
+		defer ns.deletionJobGroup.Done()
+		ok := ns.clearNamespaceResources(ctx, namespaceToDelete)
 
-func (ns *NamespaceStore) clearNamespaceResources(ctx context.Context, namespaceToDelete *namespace.Namespace) {
-	succeded := false
-	defer func() {
-		if !succeded {
-			ns.deletionMap[namespaceToDelete.UUID] = false
+		// Now grab write lock so that we can write to storage.
+		ns.lock.Lock()
+
+		defer ns.lock.Unlock()
+		// Make sure this happens _before_ the unlock.
+		defer delete(ns.deletionMap, namespaceToDelete.UUID)
+
+		// If we failed to clear any resources, stop here and don't delete the namespace.
+		if !ok {
+			return
+		}
+
+		err = ns.namespacesByPath.Delete(namespaceToDelete.Path)
+		if err != nil {
+			ns.logger.Error("failed to delete namespace entry in namespace tree", "namespace", namespaceToDelete.Path, "error", err.Error())
+		}
+		delete(ns.namespacesByUUID, namespaceToDelete.UUID)
+		delete(ns.namespacesByAccessor, namespaceToDelete.ID)
+
+		view := NamespaceView(ns.storage, parent).SubView(namespaceStoreSubPath)
+		err = logical.WithTransaction(ctx, view, func(s logical.Storage) error {
+			return s.Delete(ctx, namespaceToDelete.UUID)
+		})
+		if err != nil {
+			ns.logger.Error("failed to delete namespace storage", "namespace", namespaceToDelete.Path, "error", err.Error())
 		}
 	}()
-	defer ns.deletionJobGroup.Done()
 
-	nsCtx := namespace.ContextWithNamespace(ctx, namespaceToDelete)
+	return "in-progress", nil
+}
+
+func (ns *NamespaceStore) clearNamespaceResources(ctx context.Context, namespaceToDelete *namespace.Namespace) bool {
 	// clear ACL policies
-	policiesToClear, err := ns.core.policyStore.ListPolicies(nsCtx, PolicyTypeACL, false)
+	policiesToClear, err := ns.core.policyStore.ListPolicies(ctx, PolicyTypeACL, false)
 	if err != nil {
 		ns.logger.Error("failed to retrieve namespace policies", "namespace", namespaceToDelete.Path, "error", err.Error())
-		return
+		return false
 	}
 
 	for _, policy := range policiesToClear {
-		err := ns.core.policyStore.deletePolicyForce(nsCtx, policy, PolicyTypeACL)
+		err := ns.core.policyStore.deletePolicyForce(ctx, policy, PolicyTypeACL)
 		if err != nil {
 			ns.logger.Error(fmt.Sprintf("failed to delete policy %q", policy), "namespace", namespaceToDelete.Path, "error", err.Error())
-			return
+			return false
 		}
 	}
 
 	// clear auth mounts
-	authMountEntries, err := ns.core.auth.findAllNamespaceMounts(nsCtx)
+	authMountEntries, err := ns.core.auth.findAllNamespaceMounts(ctx)
 	if err != nil {
 		ns.logger.Error("failed to retrieve namespace credentials", "namespace", namespaceToDelete.Path, "error", err.Error())
-		return
+		return false
 	}
 
 	for _, me := range authMountEntries {
-		err := ns.core.disableCredentialInternal(nsCtx, me.Path, true)
+		err := ns.core.disableCredentialInternal(ctx, me.Path, true)
 		if err != nil {
 			ns.logger.Error(fmt.Sprintf("failed to unmount %q", me.Path), "namespace", namespaceToDelete.Path, "error", err.Error())
-			return
+			return false
 		}
 	}
 
 	// clear mounts
-	mountEntries, err := ns.core.mounts.findAllNamespaceMounts(nsCtx)
+	mountEntries, err := ns.core.mounts.findAllNamespaceMounts(ctx)
 	if err != nil {
 		ns.logger.Error("failed to retrieve namespace mounts", "namespace", namespaceToDelete.Path, "error", err.Error())
-		return
+		return false
 	}
 
 	for _, me := range mountEntries {
-		err := ns.core.unmountInternal(nsCtx, me.Path, true)
+		err := ns.core.unmountInternal(ctx, me.Path, true)
 		if err != nil {
 			ns.logger.Error(fmt.Sprintf("failed to unmount %q", me.Path), "namespace", namespaceToDelete.Path, "error", err.Error())
-			return
+			return false
 		}
 	}
 
 	// clear identity store
 	if err := ns.core.identityStore.RemoveNamespaceView(namespaceToDelete); err != nil {
 		ns.logger.Error("failed to clean identity store", "namespace", namespaceToDelete.Path, "error", err.Error())
-		return
+		return false
 	}
 
 	// clear quotas
-	err = ns.core.quotaManager.HandleNamespaceDeletion(nsCtx, namespaceToDelete.Path)
+	err = ns.core.quotaManager.HandleNamespaceDeletion(ctx, namespaceToDelete.Path)
 	if err != nil {
 		ns.logger.Error("failed to update quotas after deleting namespace", "namespace", namespaceToDelete.Path, "error", err.Error())
-		return
+		return false
 	}
 
 	// clear locked users entries
 	_, err = ns.core.runLockedUserEntryUpdatesForNamespace(ctx, namespaceToDelete, true)
 	if err != nil {
 		ns.logger.Error("failed to clean up locked user entries", "namespace", namespaceToDelete.Path, "error", err.Error())
-		return
-	}
-
-	// Now grab write lock so that we can write to storage.
-	ns.lock.Lock()
-	defer ns.lock.Unlock()
-
-	err = ns.namespacesByPath.Delete(namespaceToDelete.Path)
-	if err != nil {
-		ns.logger.Error("failed to delete namespace entry in namespace tree", "namespace", namespaceToDelete.Path, "error", err.Error())
-		return
-	}
-	delete(ns.namespacesByUUID, namespaceToDelete.UUID)
-	delete(ns.namespacesByAccessor, namespaceToDelete.ID)
-
-	parent, err := namespace.FromContext(ctx)
-	if err != nil {
-		ns.logger.Error("error loading parent namespace from context", "namespace", namespaceToDelete.Path, "error", err.Error())
-		return
+		return false
 	}
 
-	view := NamespaceView(ns.storage, parent).SubView(namespaceStoreSubPath)
-	err = logical.WithTransaction(ctx, view, func(s logical.Storage) error {
-		return s.Delete(ctx, namespaceToDelete.UUID)
-	})
-	if err != nil {
-		ns.logger.Error("failed to delete namespace storage", "namespace", namespaceToDelete.Path, "error", err.Error())
-	}
-	succeded = true
+	return true
 }
 
 // ResolveNamespaceFromRequest resolves a namespace from the 'X-Vault-Namespace'

```

</details>

wslabosz-reply: applied your suggestion in its entirety

wslabosz-reply:
@cipherboy 
fixed the tests, refactored the code (once again, thanks @satoqz), could you re-review the PR?

helper/namespace/namespace.go
```diff
@@ -38,15 +38,12 @@ type (
...
 type Namespace struct {
-	ID        string `json:"id" mapstructure:"id"`
-	UUID      string `json:"uuid" mapstructure:"uuid"`
-	Path      string `json:"path" mapstructure:"path"`
-	Tainted   bool   `json:"tainted" mapstructure:"tainted"`
-	Locked    bool   `json:"-"`
-	UnlockKey string `json:"unlock_key" mapstructure:"unlock_key"`
-	// IsDeleting tracks whether there's an ongoing deletion process of the specified namespace
-	// If tainted is true, but IsDeleting not, then namespace deletion operation has to be retried.
-	IsDeleting     bool              `json:"-"`
```
cipherboy, line 49: Hmm, @wslabosz-reply @satoqz Don't we want to go the other way and mark this boolean as persistent and set `json:"is_deleting"`? Otherwise, we could kill the binary in this new setup and not realize that our namespace is in-deletion and we should allow the operator to retry/restart deletion if it failed. We shouldn't allow any other operations (like actually loading the namespace), though it should probably appear in lists with this status. My 2c.

satoqz: > we could kill the binary in this new setup and not realize that our namespace is in-deletion

We do realize this because `Tainted` is persisted. `Tainted` being true means that a namespace has, at some point, been in-deletion. It may not be in-deletion anymore if the server was restarted while the deletion was ongoing, but the operator can just retry it from there.

> We shouldn't allow any other operations (like actually loading the namespace)

I suspect we might have to load the namespace such that all the components of core know how to properly delete their resources associated to the namespace. This is what the initial `clearNamespaceResources` run expects, if another one is fired off following a restart we wouldn't want the namespace to be in a "shallow" mode that requires a different deletion procedure.

wslabosz-reply: @cipherboy 
so the initial approach was meant to work as @satoqz described (although had that flaw that if the `isDeleting` value did not get reset to false, you'd have to "kill the binary" (reload the server) and reloading the namespace would not include that status (as it is not persisted), then operator could just run namespace deletion operation once again. With this fix it is not required to reload the server, as the `deletionMap` should track whether deletion completed successfully. Operator will know that deletion operation did not pass if the namespace is `tainted` (that property is on the list/read endpoint response).

> We shouldn't allow any other operations (like actually loading the namespace), though it should probably appear in lists with this status.

I mean yes, not loading mounts makes sense, but if deletion did fail on storage removals, then without loading the namespace, we wouldn't be able to clean it up using delete operation (still could be done with recovery mode probably).

cipherboy: @wslabosz-reply @satoqz Hm, I guess we'll have to try this. I worry with certain places (thinking expiration manager which maintains multiple storage entries per item) we might panic if we load a partially deleted namespace.

wslabosz-reply: @cipherboy 
fair point, in the end if something goes that awry, that's what the recovery mode should be for, right?

cipherboy: @wslabosz-reply Yes though I'd prefer to avoid it and fix the problem automatically ;-)

wslabosz-reply: @cipherboy 
I mean, yes, but we are in dead end of:
- if we load partially deleted namespace, we are (possibly) exposing ourselves to the nil deferences
- if we don't load partially deleted namespace, the operator won't be able to retry the delete operation

wslabosz-reply: @cipherboy 
the single thing that comes to my mind now is omitting the load of namespace related resources of a namespace that is `tainted` in every subsystem apart from `namespaceStore`

satoqz: @cipherboy @wslabosz-reply another option that comes to mind is permitting an interrupt-via-shutdown only _inbetween_ deletion steps to ensure that a certain subsystem or mount doesn't end up half-deleted (unless we hit an unexpected error ofc).

Something like this, repeated for each component:

```go
for _, me := range mountEntries {
	select {
	case <-quitCtx.Done():
		//   ^ canceled on shutdown
		return false
	default:
		err := ns.core.unmountInternal(nsCtx, me.Path, true)
		//                             ^ never canceled (or a longer independent timeout)
		if err != nil {
			return false
		}
	}
}
```

cipherboy: @wslabosz-reply Sorry, from this PoV:

> if we load partially deleted namespace, we are (possibly) exposing ourselves to the nil deferences

I was thinking if we load it into the namespace store so it shows in lists/... but not load any values (mount table, skip it in invalidations, ...) so we avoid the _bulk_ of the nil pointer derefs? Then the namespace otherwise behaves like it is not there (it is pending deletion anyways) and I believe `taint` already catches most everything else.

Thoughts?

@satoqz That approach is OK, but I don't think it stops the "power was pulled on server" forced outage. :-)

wslabosz-reply: > I was thinking if we load it into the namespace store so it shows in lists/... but not load any values (mount table, skip it in invalidations, ...) so we avoid the bulk of the nil pointer derefs? Then the namespace otherwise behaves like it is not there (it is pending deletion anyways) and I believe taint already catches most everything else.

So basically:
```
if entry.namespace.Tainted {
	continue
}
```
on setup auth/mounts?

cipherboy: @wslabosz-reply I think so, yes, plus a few more skips in expiration manager if it doesn't already exist.

wslabosz-reply: @cipherboy 

I don't see a place where a skip in expiration manager would be required, could you describe a situation where we could encounter a nil dereference in the context of the partially deleted namespace in exp manager?

wslabosz-reply:
@cipherboy 
From the internal wg discussion, let's merge it in this state, and see if we can replicate a situation when this breaks.

driif:
https://github.com/openbao/openbao/issues/1543

Not directly related to this PR, but could you please also take a look to an issue - a Race Condition performing parallel deletion of namespace and mounts

@wslabosz-reply @satoqz

driif:
LGTM

driif (approved):

satoqz:
Had a call with @driif, we went over https://github.com/openbao/openbao/issues/1543 and possible solutions to that - we agreed though that it shouldn't hold up this PR which at least partially resolves the issue.

vault/namespace_store.go
```diff
@@ -686,110 +703,110 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
...
+
+		err = ns.namespacesByPath.Delete(namespaceToDelete.Path)
+		if err != nil {
+			ns.logger.Error("failed to delete namespace entry in namespace tree", "namespace", namespaceToDelete.Path, "error", err.Error())
+		}
+		delete(ns.namespacesByUUID, namespaceToDelete.UUID)
+		delete(ns.namespacesByAccessor, namespaceToDelete.ID)
+
+		view := NamespaceView(ns.storage, parent).SubView(namespaceStoreSubPath)
+		err = logical.WithTransaction(ctx, view, func(s logical.Storage) error {
```
cipherboy, line 733: Single deletes do not need to be in a transaction.

PR #1526, opened 2025-07-03
api/auth/jwt: initial implementation of JWT Auth Method

dnaeon:
This PR adds an implementation of `auth.AuthMethod` for JWT Auth Method in the SDK.


JanMa:
Hello @dnaeon,

thank you for your contribution. Can you please add some background as to why you opened the PR, what problem or open issue it solves? The usual approach we follow is to first discuss problems or new feature requests in an issue before a PR is opened

DrDaveD:
Is there something missing in the [builtin jwt auth method](https://openbao.org/docs/auth/jwt/)?

satoqz:
> Is there something missing in the [builtin jwt auth method](https://openbao.org/docs/auth/jwt/)?

If I understand correctly this would be a helper for clients that consume the builtin JWT auth method, not an alternative. The same pattern exists for `approle/kubernetes/ldap/userpass`.

dnaeon:
> thank you for your contribution. Can you please add some background as to why you opened the PR, what problem or open issue it solves? The usual approach we follow is to first discuss problems or new feature requests in an issue before a PR is opened

@JanMa , sorry about that, I should have done that initially. I'll try to explain the rationale here as well, let me know if you still think I should submit an issue for it.

> > Is there something missing in the [builtin jwt auth method](https://openbao.org/docs/auth/jwt/)?
> 
> If I understand correctly this would be a helper for clients that consume the builtin JWT auth method, not an alternative. The same pattern exists for `approle/kubernetes/ldap/userpass`.

Correct. This is just an implementation of the [api.AuthMethod](https://github.com/openbao/openbao/blob/f012eef04fb5e1bcad84b22f5a588b331fbfb013/api/auth.go#L17-L19) interface, for API clients to use and authenticate against a JWT Auth Method login endpoint as @satoqz mentioned.

While the [auth/kubernetes.KubernetesAuth](https://github.com/openbao/openbao/blob/f012eef04fb5e1bcad84b22f5a588b331fbfb013/api/auth/kubernetes/kubernetes.go#L15-L19) implementation can be used to accomplish the same thing, it's name (and internal details such as Kubernetes service accounts, default token paths, etc.) suggest it's meant to be used for Kubernetes workloads (even though, that is not a requirement).

This implementation on the other hand I think is a bit more generic and can be used with any JWT/OIDC IdP, without suggesting in any way where the JWT token comes from - Kubernetes, Github Actions, or any other IdP.

Technically, both implementations (`JWTAuth` from this PR and the existing `KubernetesAuth`) can do the same thing, but this `JWTAuth` can be used in the broader sense by not making any suggestions that it's Kubernetes (or other IdP) related.

Hopefully that makes sense, thanks!

JanMa:
Thanks a lot for the additional details @dnaeon

DrDaveD:
That sounds reasonable to me but I'd like to see what @cipherboy thinks.

cipherboy:
I think this is a good idea, though I've not looked at the code. Thanks @dnaeon!

dnaeon:
Seems like a test unrelated to this change has failed. Is this test failure a known issue/expected?

@DrDaveD , @cipherboy , @JanMa ?

satoqz:
@dnaeon `TestOIDC_PeriodicFunc` is a known test flake, nothing related to your changes :)

api/auth/jwt/jwt.go
```diff
@@ -0,0 +1,166 @@
...
+	switch {
+	case a.token != "":
+		token = a.token
+	case a.tokenPath != "":
+		data, err := os.ReadFile(filepath.Clean(a.tokenPath))
+		if err != nil {
+			return nil, err
+		}
+		token = string(data)
+	}
```
satoqz: Do you want to error if no token was set? `New(...)` mostly ensures this is not the case, but the file at `tokenPath` may be empty. Alternatively maybe already read `tokenPath` in `New(...)`.

api/auth/jwt/jwt.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return opt
+}
+
+// WithMountPath is an [Option], which configures [JWTAuth] to use the given
+// mount path for the OpenBao Authentication Method.
+func WithMountPath(mountPath string) Option {
+	opt := func(a *JWTAuth) error {
+		// Remove any trailing slashes from the given mount path
+		a.mountPath = strings.TrimRight(mountPath, "/")
```
satoqz: Nit: I don't believe this is necessary, unless you hit a case where it was?

api/auth/jwt/jwt.go
dnaeon: I initially had the code that way -- the token was read when using the `WithTokenFromPath` option, but then I realized that if the JWT token gets refreshed in the meantime we won't notice this, e.g. this is the case with [Projected Service Account Tokens](https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken), where the token will be periodically refreshed. In that case during `Login()` I want to make sure that I have the latest JWT token from the path, thus reading it only during `Login()`.

> Do you want to error if no token was set? New(...) mostly ensures this is not the case

This one is covered here as well: 

- https://github.com/openbao/openbao/pull/1526/files#diff-56836077c3418aae93f4982d0fbb28fc100a0649aa78a8a55f0338c6b6b5ca9cR83-R85

If no `token` and no `tokenPath` are specified `New()` will error out.

api/auth/jwt/jwt.go
dnaeon: Interestingly, I had an issue when using extra slashes for the JWT Auth Method mount point... Tried that again now and that's not the case, so not sure what went wrong during my initial tests.

Will submit a change which removes this, thanks!

EDIT: Just pushed a fix in https://github.com/openbao/openbao/pull/1526/commits/a8c0ab0590335e94a8ae63888f9dc2539e28b085

api/auth/jwt/jwt.go
satoqz: > the token was read when using the WithTokenFromPath option, but then I realized that if the JWT token gets refreshed in the meantime we won't notice this, e.g. this is the case with [Projected Service Account Tokens](https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken), where the token will be periodically refreshed. In that case during Login() I want to make sure that I have the latest JWT token from the path, thus reading it only during Login().

Got it, that makes sense to me.

The only edge case I'm concerned about then is if the file _contents_ are empty, not its name.

You could technically bypass `New(...)` by manually constructing `JWTAuth{}`, but I don't think we should worry about that 😄

dnaeon: > The only edge case I'm concerned about then is if the file contents are empty, not its name.

Such errors would bubble up when the client tries to `Login()` - invalid JWT token, etc. I would say such errors are expected and visible to the client. For example when external process refreshes the JWT token and our `JWTAuth` tries to `Login()` and catches the token refresh in the middle, which results in no token data read from the file.

In such cases the client may decide to retry or use a custom a [Config.CheckRetry](https://pkg.go.dev/github.com/openbao/openbao/api#Config) custom policy.

> You could technically bypass New(...) by manually constructing JWTAuth{}, but I don't think we should worry about that 😄

Agree :)

satoqz: > Such errors would bubble up when the client tries to Login() - invalid JWT token, etc. I would say such errors are expected and visible to the client. For example when external process refreshes the JWT token and our JWTAuth tries to Login() and catches the token refresh in the middle, which results in no token data read from the file.

Totally, I'm just saying you could also catch this on the client-side inside of `Login()` once you've read the file contents. No need to get the error from the server.

So just:

```go
data, err := os.ReadFile(filepath.Clean(a.tokenPath))
if err != nil {
	return nil, err
}
if len(data) == 0 {
    return nil, ErrNoToken
}
token = string(data)
```

dnaeon: > Totally, I'm just saying you could also catch this on the client-side inside of Login() once you've read the file contents. No need to get the error from the server.

Got it, makes sense.

Just pushed a fix for this one: https://github.com/openbao/openbao/pull/1526/commits/86453817e8c57e4e3fe4902ec42ed9d4d5f7e68a

Let me know how it looks, thanks.

satoqz (approved):

api/auth/jwt/jwt.go
satoqz: Thanks! Sorry for being pedantic 😄

api/auth/jwt/jwt.go
```diff
@@ -0,0 +1,168 @@
...
+		if err != nil {
+			return nil, err
+		}
+		token = string(data)
+		if token == "" {
+			return nil, fmt.Errorf("%w: got empty token from %s", ErrNoToken, a.tokenPath)
+		}
+	}
+
+	path := fmt.Sprintf("auth/%s/login", a.mountPath)
```
cipherboy, line 112: This is consistent with other auth helpers but also weird. I guess namespace _has_ to come from the client and not from a custom opt in this auth helper.

PR #1527, opened 2025-07-04
Fix typo in dev mode

Swinir:
When running in development mode using the Windows version of the binary, there is a typo in the command used to set the "BAO_ADDR" environment variable. This PR aims to correct this typo so that the command works correctly.

As described in the documentation, I did not create an issue for this PR, as it is just a minor typo.

### DIFF
--- a/changelog/1527.txt
+++ b/changelog/1527.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+command: fixes typo in Windows command for setting BAO_ADDR in development mode
+```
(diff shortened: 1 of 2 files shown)


satoqz:
Thanks @Swinir! Could you add a simple changelog entry? I think it's warranted given this isn't just a documentation typo, but fixes an actual bug.

Swinir:
Sure ! I will add a small entry.

satoqz (approved):

satoqz:
Merging as this is a simple fix.

Swinir:
Was this commit missed in the new version ? Just a heads-up if it got lost. @satoqz

cipherboy:
@Swinir Ah yes, it wasn't marked for backport. We'll pick it up in v2.4.0 which should be soon-ish.

breca:
Still present in v2.4.3

satoqz:
Hi @breca, are you 100% sure? This patch has been part of the 2.4.x release series from the start.

PR #1529, opened 2025-07-06
add support for inline kerberos auth

### LINKED ISSUE #1516: Using kerberos with inline authentication doesn't work
**Describe the bug**
I'm trying to use [inline auth](https://openbao.org/docs/rfcs/inline-auth/) with [kerberos](https://openbao.org/docs/auth/kerberos/). However, the inline auth handler prevents the use of the `Authorization` header in the request and doesn't forward the response headers and status code from the auth plugin, which prevents kerberos auth from working with inline auth.

**To Reproduce**
1. Setup [kerberos auth](https://openbao.org/docs/auth/kerberos/) endpoint and a simple kv mount
2. Verify kerberos auth works:
```bash
$ curl "https://$HOSTNAME:8200/v1/auth/kerberos/login" -X POST --negotiate -u :
{..., "auth":{"client_token":"s.xxxxxxxxxxxxxxxxxxxxxxxx", ...}, ...}
```
3. Try using inline-auth with kv put:
```bash
$ curl "https://$HOSTNAME:8200/v1/kv/data/test" -X POST --negotiate -u : -H "X-Vault-Inline-Auth-Path: auth/kerberos/login" -d '{"data": {"a": "b"}}'
{"errors":["failed to perform inline authentication: authentication required"]}
```

**Expected behavior**
Step 3 should succeed:
```bash
$ curl "https://$HOSTNAME:8200/v1/kv/data/test" -X POST --negotiate -u : -H "X-Vault-Inline-Auth-Path: auth/kerberos/login" -d '{"data": {"a": "b"}}'
{...,"data":{"created_time":"2025-07-01T02:08:43Z","custom_metadata":null,...}}
```

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.0.0-HEAD (commit 29ebb4a)
* OpenBao CLI Version (retrieve with `bao version`): OpenBao v2.0.0-HEAD (commit 29ebb4a)
* Server Operating System/Architecture: Linux x86_64

OpenBao server was started with `-dev`

**Additional context**
When handling kerberos authentication, this seems to be the process
1. Client sends request to server without authentication
2. Server responds with `401 Unauthorized` along with a `WWW-Authenticate: Negotiate` header
3. Client retries the original request with an `Authorization: Negotiate .....` header
4. Server checks the header and responds with authenticated content

I've thought of one way to fix the problem:

1. Forward the response code and headers generated by the auth plugin from inside [`handleInlineAuth`](https://github.com/openbao/openbao/blob/29ebb4a0104acbabe26c43df3e468cfcc0cd6886/vault/request_handling.go#L744) all the way out to [`switchedLockHandleRequest`](https://github.com/openbao/openbao/blob/29ebb4a0104acbabe26c43df3e468cfcc0cd6886/vault/request_handling.go#L626)
2. Don't [reject requests](https://github.com/openbao/openbao/blob/29ebb4a0104acbabe26c43df3e468cfcc0cd6886/vault/request_handling.go#L648) containing an `Authorization` header if it starts with `Negotiate `

jackyliao123:
Allows kerberos auth to work with inline authentication.

To do this, this PR does these 2 things:

1. Instead of manually checking the headers for `X-Vault-Index` or `Authorization: Bearer ...`, we instead check for the whether there's a `ClientToken` associated with the request. This `ClientToken` should be populated by the `http/handler.go`, and would let the `handleInlineAuth` function not fail when there's a `Authorization: Negotiate ...` header. 
2. When the auth endpoint fails, instead of returning a generic 500, we instead include both the response status code and the response headers from the auth endpoint to the client. This should allow the `WWW-Authenticate: Negotiate` header to reach the client to start the kerberos auth.

Resolves #1516


vault/request_handling.go
```diff
@@ -622,37 +622,33 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
 	if ok {
 		ctx = logical.CreateContextOriginalBody(ctx, body)
 	}
-	if err = c.handleInlineAuth(ctx, req, nsHeader); err != nil {
-		return nil, fmt.Errorf("failed to perform inline authentication: %w", err)
+	if resp, err := c.handleInlineAuth(ctx, req, nsHeader); err != nil {
+		return resp, multierror.Append(errors.New("failed to perform inline authentication"), err)
```
satoqz: We had discussed (on the weekly community call) that there could be a custom response header that indicates an error originates from inline auth, but I also kind of like this. We should likely have both such that it's easy to tell both by reading the message and via programmatic checks.

vault/request_handling.go
```diff
@@ -622,37 +622,33 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
 	}
 	resp, err = c.handleCancelableRequest(ctx, req)
 	req.SetTokenEntry(nil)
 	cancel()
 	return resp, err
 }
 
-func (c *Core) handleInlineAuth(ctx context.Context, req *logical.Request, nsHeader string) error {
+func (c *Core) handleInlineAuth(ctx context.Context, req *logical.Request, nsHeader string) (resp *logical.Response, err error) {
```
```suggestion
func (c *Core) handleInlineAuth(ctx context.Context, req *logical.Request, nsHeader string) (*logical.Response, error) {
```

Nit: No real need to name these. Down below just keep using:

```go
resp, err := c.handleCancelableRequest(authCtx, authReq)
```

vault/request_handling.go
cipherboy: I'm curious what the multi-error does for us that wrapping doesn't? I suppose it makes it a little easier to parse the response as they're in two separate errors? 

My personal view is that while `multierror` has a place, perhaps it is better left for places where the are actually two or more separate errors rather than a single (wrapped) error.

changelog/1529.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+core/auth: allow inline auth to work with kerberos auth
```
cipherboy: Inline auth isn't released yet, so you don't need a changelog entry btw :-)

A test would be nice, we have some in `vault/external_tests/api/secret_test.go` :-)

vault/request_handling.go
jackyliao123: Ah, I only used `multierror` because it preserved the status code. 

At [sdk/logical/response_util.go#L156](
https://github.com/openbao/openbao/blob/374fc02d2d060c4915b198f138a610b46d05ccd1/sdk/logical/response_util.go#L156), it seems like only `multierror` and plain `HTTPCodedError` are supported. I tried using `errwrap` but the response code was ignored (not sent to the client).

Are you thinking of something else or another way of achieving this?

vault/request_handling.go
jackyliao123: When I make this change locally, the `ci-vet-codechecker` target seems to be unhappy:

```
vault/request_handling.go:634:1: Function handleInlineAuth can return an error, and has a statement that returns only nils
```

My assumption is that it's tripped up the `return nil, nil` I added. Any advice on this?

(I'm also not sure why the automated checks are complaining about this, even though this pull request has the named returns)

satoqz: Pretty sure it complains about this either way: https://github.com/openbao/openbao/actions/runs/16094653461/job/45461650562?pr=1529

You can ignore it for all that matters, the check doesn't keep erroring once changes are on `main`.

vault/request_handling.go
cipherboy: @jackyliao123 Ah, interesting. TIL! I was expecting that this:

https://github.com/openbao/openbao/blob/374fc02d2d060c4915b198f138a610b46d05ccd1/sdk/logical/response_util.go#L176-L179

would instead be:

```go
if errwrap.ContainsType(err, HTTPCodedError) {
    t := errwrap.GetType(err, HTTPCodedError)
    *status = t.Code()
}
```

or similar.

But I guess this is not what is done. It seems weird that an `HTTPCodedError` isn't allowed to be wrapped at all but e.g., `StatusBadRequest` errors are (see `RespondErrorCommon` above)? 

Both types of errors seem just about equally likely, so I guess it doesn't matter too much, it is not like either are very widely used so I think this should be a safe change.

vault/request_handling.go
cipherboy: @jackyliao123 @satoqz Yeah, the context here is that in a _request_ handler in OpenBao `return nil, nil` has a very specific meaning (404), which applies even if the endpoint exists and otherwise has no data. Returning 204 no content requires a different format of response. So `return nil, nil` here is fine in this context.

cipherboy:
@satoqz Do you mind doing a review pass on this? I've pushed some changes. Thanks!

vault/request_handling.go
```diff
@@ -704,46 +709,63 @@ func (c *Core) handleInlineAuth(ctx context.Context, req *logical.Request, nsHea
...
+		// If we have a non-error response object, ensure it has a header
+		// indicating it is from the auth step.
+		if resp != nil {
+			if resp.Headers == nil {
+				resp.Headers = make(map[string][]string)
+			}
+			resp.Headers[consts.InlineAuthErrorResponseHeader] = []string{"true"}
+		}
+
+		// We could a case where err == resp == nil; set error to 404 not
```
```suggestion
		// We could hit a case where err == resp == nil; set error to 404 not
```
?

cipherboy:
@satoqz Do you think we should go back to `fmt.Errorf(...)`? :thinking: `semgrep` gives an error, though I think that is more convention than actual issue (as https://github.com/hashicorp/go-multierror?tab=readme-ov-file#usage suggests it should be fine). 

Thoughts?

satoqz:
@cipherboy not opposed

vault/request_handling.go
```diff
@@ -622,37 +622,42 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
-	if err = c.handleInlineAuth(ctx, req, nsHeader); err != nil {
-		return nil, fmt.Errorf("failed to perform inline authentication: %w", err)
+
+	// Perform inline authentication. This returns nil, nil if the request
+	// succeeds and the passed request is mutated to now have the token. In
+	// the event it fails, it may have either a request, an error, or both,
+	// depending on what the auth method does.
+	if resp, err := c.handleInlineAuth(ctx, req, nsHeader); err != nil || resp != nil {
+		if err != nil {
+			err = multierror.Append(errors.New("failed to perform inline authentication"), err)
```
cipherboy:
```suggestion
			err = fmt.Errorf("failed to perform inline authentication: %w", err)
```

sdk/logical/response_util.go
```diff
@@ -174,8 +174,14 @@ func AdjustErrorStatusCode(status *int, err error) {
...
-	if t, ok := err.(HTTPCodedError); ok {
-		*status = t.Code()
+	var hce HTTPCodedError = &codedError{}
+	if errwrap.ContainsType(err, hce) {
+		t := errwrap.GetType(err, hce)
+		if t != nil {
+			if coded, ok := t.(HTTPCodedError); ok {
+				*status = coded.Code()
+			}
+		}
```
satoqz, line 184: I think this should use `errors.As` now?

cipherboy: @satoqz Hm, does it matter? I was thinking this one was independent of type (multierror, wrapping, ...) whereas .As was less robust? But I might be wrong.

We don't really care about the exact error or it's message, just the status code.

satoqz: Ah, I guess didn't expect `errwrap` to be aware of conventionally wrapped errors, my bad. This is fine then.

cipherboy: @satoqz `errwrap.GetType(...)` calls `Walk` internally, which seems to support many types of wrapped errors: https://github.com/hashicorp/errwrap/blob/d914f8498083a6cd0136ffe977cd9701acd46c5b/errwrap.go#L137-L162

cipherboy:
Thank you @jackyliao123!

PR #1530, opened 2025-07-06
sdk/bptree: Introduces B+ Tree implementation (with custom storage wrapper) to be used for indexing purposes

Gabrielopesantos:
Draft opened to be linked in the RFC. Not ready to review yet.
EDIT: Need to update this.

<!--

Resolves #

<!--

### DIFF
--- a/sdk/bptree/bptree.go
+++ b/sdk/bptree/bptree.go
@@ -160,7 +160,11 @@ func (t *Tree) setRootID(ctx context.Context, storage Storage, newRootID string)
 
 // contextWithTreeID returns a context with the tree's ID added, enabling multi-tree storage
 func (t *Tree) contextWithTreeID(ctx context.Context) context.Context {
-	return t.config.contextWithTreeID(ctx)
+	if t.config == nil {
+		return ctx
+	}
+
+	return withTreeID(ctx, t.config.TreeID)
 }
 
 // maxChildrenNodes returns the maximum number of children an internal node can have
--- a/sdk/bptree/bptree.go
+++ b/sdk/bptree/bptree.go
@@ -28,7 +28,7 @@ func InitializeBPlusTree(
 ) (*BPlusTree, error) {
 	if config == nil {
 		config = NewDefaultBPlusTreeConfig()
-	} else if err := config.Validate(); err != nil {
+	} else if err := ValidateTreeConfig(config); err != nil {
 		return nil, fmt.Errorf("invalid configuration: %w", err)
 	}
 
@@ -66,7 +66,7 @@ func NewBPlusTree(
 	if config == nil {
 		return nil, fmt.Errorf("config is required for tree creation")
 	}
-	if err := config.Validate(); err != nil {
+	if err := ValidateTreeConfig(config); err != nil {
 		return nil, fmt.Errorf("invalid configuration: %w", err)
 	}
 
--- a/sdk/bptree/bptree.go
+++ b/sdk/bptree/bptree.go
@@ -42,6 +42,20 @@ func InitializeBPlusTree(
 	return NewBPlusTree(ctx, storage, config)
 }
 
+// InitializeBPlusTreeWithOptions initializes a tree using functional options.
+// Creates the tree if it doesn't exist, or loads it if it does.
+func InitializeBPlusTreeWithOptions(
+	ctx context.Context,
+	storage Storage,
+	opts ...TreeOption,
+) (*BPlusTree, error) {
+	config, err := NewBPlusTreeConfig(opts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create tree config: %w", err)
+	}
+	return InitializeBPlusTree(ctx, storage, config)
+}
+
 // NewBPlusTree creates a new B+ tree with the given configuration.
 // Fails if a tree with the same ID already exists.
 func NewBPlusTree(
@@ -89,6 +103,20 @@ func NewBPlusTree(
 	}
 
 	return tree, nil
+}
+
+// NewBPlusTreeWithOptions creates a new B+ tree using functional options.
+// Fails if a tree with the same ID already exists.
+func NewBPlusTreeWithOptions(
+	ctx context.Context,
+	storage Storage,
+	opts ...TreeOption,
+) (*BPlusTree, error) {
+	config, err := NewBPlusTreeConfig(opts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create tree config: %w", err)
+	}
+	return NewBPlusTree(ctx, storage, config)
 }
 
 // LoadExistingBPlusTree loads an existing B+ tree from storage using the stored configuration
--- a/sdk/bptree/bptree.go
+++ b/sdk/bptree/bptree.go
@@ -61,7 +61,7 @@ func NewBPlusTree(
 
 	// Check if tree already exists
 	existingConfig, err := storage.GetConfig(ctx)
-	if err != nil {
+	if err != nil && !errors.Is(err, ErrConfigNotFound) {
 		return nil, fmt.Errorf("failed to check for existing tree: %w", err)
 	}
 	if existingConfig != nil {
@@ -106,11 +106,11 @@ func LoadExistingBPlusTree(
 
 	// Get stored configuration - this is the source of truth
 	storedConfig, err := storage.GetConfig(ctx)
-	if err != nil {
+	if err != nil && !errors.Is(err, ErrConfigNotFound) {
 		return nil, fmt.Errorf("failed to load tree configuration: %w", err)
 	}
 	if storedConfig == nil {
-		return nil, fmt.Errorf("tree '%s' does not exist", treeID)
+		return nil, fmt.Errorf("tree (%s) does not exist", treeID)
 	}
 
 	// Create tree with stored configuration
--- a/sdk/bptree/bptree.go
+++ b/sdk/bptree/bptree.go
@@ -64,7 +64,7 @@ func newTree(
 	tree := &Tree{config: config}
 
 	// Create new leaf root
-	root := NewLeafNode(generateUUID())
+	root := NewLeafNode()
 	if err := storage.PutNode(ctx, root); err != nil {
 		return nil, fmt.Errorf("failed to save root node: %w", err)
 	}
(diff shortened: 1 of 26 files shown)


sdk/lru/lru.go
cipherboy, line 1: @Gabrielopesantos Not sure I understand what this file does?

sdk/lru/lru.go
```diff
@@ -0,0 +1,46 @@
+package lru
+
+import lru "github.com/hashicorp/golang-lru/v2"
+
+type LRU[K comparable, V any] struct {
+	size int
```
cipherboy, line 6: iiuc it just allows exposing `size` since `TwoQueueCache` assumes you already know it and `Len` returns current length not total? 🤔

cipherboy: If so I think you can do something like:

```go
type LRU[K comparable, V any] struct {
    lru.TwoQueueCache[K, V]
    size int
}

func (c *LRU[K, V]) Size() int {
    return c.size
}
```

and have a simpler implementation (still need `NewLRU` though).

sdk/bptree/node.go
```diff
@@ -0,0 +1,762 @@
...
+)
+
+// Node represents a node in the B+ tree
+type Node struct {
+	ID     string `json:"id"`
+	IsLeaf bool   `json:"isLeaf"`
+	// NOTE (gabrilopesantos): Consider making keys a generic type instead of string
+	Keys []string `json:"keys"`
+	// NOTE (gabrielopesantos): Have some limit in the number of values per key?
+	Values      [][]string `json:"values"`      // Only for leaf nodes
```
cipherboy, line 53: Nodes would need to be split if this grows too large, yeah. 🤔  Raft has a relatively small (512kb) default operation limit, which interacts with transactions non-trivially (4x?). So writing lots of large nodes could potentially be a problem.

sdk/bptree/node.go
```diff
@@ -0,0 +1,762 @@
...
+	if !n.IsLeaf {
+		return ErrNotALeafNode
+	}
+
+	idx, keyExists := n.FindKeyIndex(key)
+
+	if keyExists {
+		// Key exists, append value if not already present
+		if !slices.Contains(n.Values[idx], value) {
+			n.Values[idx] = append(n.Values[idx], value)
```
cipherboy, line 263: Should values be sorted? Then we can do pagination over values as well. 🤔 Or is that too complicated?

sdk/bptree/storage_keys.go
```diff
@@ -0,0 +1,34 @@
...
+// Storage path constants
+const (
+	nodesPath  = "nodes"
+	rootPath   = "root"
+	configPath = "config"
+)
+
+// configKey constructs the storage key for tree metadata, using tree ID from context if available
+func configKey(ctx context.Context) string {
+	treeID := getTreeIDOrDefault(ctx, defaultTreeID)
```
cipherboy, line 14: After #2910, we should be careful to use `storage.View` to ensure we don't traverse across boundary of the btree, IMO.

cipherboy: I don't know that we have a great way of working with them from a transaction PoV, though. Deriving transactions through/bringing transaction to a view. 

(Happy to take new, scoped PRs on that topic.)

sdk/bptree/storage_transactional.go
```diff
@@ -0,0 +1,310 @@
...
+	// Either Commit or Rollback must be called to release resources.
+	Rollback(context.Context) error
+}
+
+// TransactionalStorage is implemented if a storage backend implements
+// Transactional as well.
+type TransactionalStorage interface {
+	Storage
+	Transactional
+}
```
cipherboy, line 34: Aren't these interfaces 1:1 for `sdk/logical/storage_transactions.go`? I think you can use them interchangeably and therefore drop these.

sdk/bptree/storage_caching.go
```diff
@@ -0,0 +1,53 @@
...
+	return s.cachingEnabled
+}
+
+// PurgeCache clears all entries from the cache
+func (s *NodeStorage) PurgeCache() {
+	if s.cache != nil {
+		s.cache.Purge()
+	}
+}
+
```
cipherboy, line 26: Probably needs an `invalidate(...path...)` method to be called from the backend's invalidate func.

sdk/bptree/storage_config.go
```diff
@@ -0,0 +1,157 @@
...
+	"errors"
+	"fmt"
+)
+
+const (
+	// defaultCacheSize is the default cache size for node caching
+	defaultCacheSize = 1000
+)
+
+// NOTE (gabrielopesantos): This is probably not needed (standard JSON serialization is fine for now)
```
cipherboy, line 14: I do wonder if we want to just start with protobuf tbh.

sdk/bptree/storage_config.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+type StorageConfig struct {
+	// NodeSerializer defines how nodes are serialized/deserialized
+	NodeSerializer NodeSerializer
+
+	// Enable or disable caching at the storage layer
+	CachingEnabled bool
+	// Cache size for node caching (if caching is enabled)
+	CacheSize int
```
cipherboy, line 45: Ah I see, this is where the LRU fork comes from, 🤔  Can we just keep it here and not allow modifications?

sdk/bptree/storage_config.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// Enable or disable built-in write buffering at the storage layer
+	BufferingEnabled bool
```
cipherboy, line 48: Sounds like a risky behavior, tbh. I'd probably simplify and remove it for now unless someone asks for it.

sdk/bptree/util.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2024 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package bptree
+
+import "github.com/hashicorp/go-uuid"
+
+// generateUUID generates a UUID
```
cipherboy, line 8: Go 1.27 has a new UUID package finally 🥳

sdk/bptree/bptree.go
```diff
@@ -0,0 +1,1130 @@
...
+		current, err = storage.GetNode(ctx, current.NextID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load next leaf node: %w", err)
+		}
+	}
+
+	return results, nil
+}
+
+// TODO: Instead of all these methods we can probably have a single method with an opts struct
```
cipherboy, line 299: +1

cipherboy (commented):
@Gabrielopesantos Random larger thoughts.

sdk/bptree/storage_keys.go
suprjinx: I do have an ancient PR that was placing current transactions in the context, which might be a way to work with them here -- https://github.com/openbao/openbao/pull/1442

suprjinx:
This is a very clean and impressive PR! I am wondering how it impacts performance, since a write-heavy index can actually slow you down vs. non-indexed. Would a native index in Postgres storage provide the same benefits for free? I believe that raft (bbolt) also has indexing built-in -- could we better leverage these for faster retrieval?

Maybe we need to see how it's wired into KV to better grok how it would be used.

sdk/bptree/node.go
Gabrielopesantos: Isn't the default op limit 1024kB?
```
defaultMaxEntrySize    = uint64(2 * raftchunking.ChunkSize)
```
And max transaction size 8 MiB?
```
defaultMaxTxnSize      = 8 * defaultMaxEntrySize
```

cipherboy: @Gabrielopesantos Yeah, the other two things we could do are protobuf encoding and compressed entries. 🤷 

Interesting, guess our limits page is wrong

yordis:
@Gabrielopesantos are you active on this one still or should I fork it and try to get it to the finish line?

Gabrielopesantos:
@yordis Yes, I'm working on it. Got back to it recently and intend to make some good progress in the near future.

sdk/bptree/storage_config.go
Gabrielopesantos: Sure, it also solves node schema evolution, which otherwise is hand-rolled version field.

sdk/bptree/storage_transactional.go
Gabrielopesantos: Not quite. They are the same  shape but they embed they embed `bptree.Storage` (node-level) rather than `logical.Storage`. They are being unexported regardless, which removes the confusion.
I can re-evaluate whether it makes sense to keep them.

sdk/lru/lru.go
Gabrielopesantos: Makes sense. Even the tests that I hjave added can be removed since they are kind of redundant.

sdk/bptree/storage_config.go
Gabrielopesantos: Or, this might be simpler even.

sdk/bptree/node.go
Gabrielopesantos: So currently each key holds a list of values ordered by insert-order, and deduped on insertion. We could sort the slice on insert on every insertion. However, with my review I also realized that by not having some sort of limit on key-values on a node we might also hit some sort of maximum entry size and this list can't be divided so it will keep growing.
What if we instead killed the `Values` and switched to a composite key: `{key} + 0x00 + {value}`. The tree becomes a set of these. This seems to be a commonly employed technique.
Every entry for a key shares the prefix `{key}+0x00`, and the tree already orders keys byte-wise, so values emerge sorted from the leaf scan with no sort step. Each `(key, value)` pair is its own independent entry and pagination over values works the same after-cursor way pagination over keys already would. 
The disadvantages is that the key prefix is duplicated per value and we will need byte encoders for some types, such as number and timestamps.

cipherboy: @Gabrielopesantos I was thinking about this more, and what if we split out value storage from key storage? I think this is a design trick in another database format, using B+ trees for keys (assuming they're largely constant -- which might be more true in our case) and then using a different structure for efficient value storage? 

I think we could store something like (to continue abusing uuids):

```
b+-tree: key -> uuid

storage/value-sets/<uuid>/...
```

Depending on how you wanted to do this, you could store them directly as keys (base64-encoded?) or as hashes or similar. With something like HMAC+SHA384, you'd probably be guaranteed collision resistance and masking from a storage PoV (without #3055). 

This means:

 - For an already existing key, add is always easy (just put to storage).
 - For an existing key with 2+ entries, remove is easy (just a delete to storage).

But:

 - Deleting all values for a key is easy-ish (remove b+ tree and clearview on uuid).
 - Insert of a new key also becomes a little more costly by an additional write.

Thoughts?

sdk/bptree/storage_transactional.go
cipherboy: Ah, I had missed that `bptree.Storage` was so different in this pass. Makes sense.

PR #1512, opened 2025-06-27
Add RFC on requiring authenticated rekeys

cipherboy:
See also: #1511

PR #1533, opened 2025-07-07
Support TLS Verification Against Explicit Alt Name

SamJoan:
Hi all,

This MR addresses the following issue, highlighted by @cipherboy, in the [following GitLab issue](https://github.com/openbao/openbao/pull/1512).

> We should support explicit server name, which may differ from connection URL. E.g., PostgreSQL might be available at `postgres.service` but report a certificate with SAN `postgres.local` that we'd otherwise trust. Same for JWT auth against a secondary Geo site; it may connect to `secondary.gitlab.instance` but TLS in the cluster report a hostname of `gitlab.local` because it is in-cluster traffic.
> [...]
> there's a lot of OpenBao that we don't use so don't particularly care to update. But the gist of it is that, due to Kubernetes routing, some places we'll set a URL which we may wish to connect to (e.g., `https://gitlab.com`) but which'll present an internal certificate for in-cluster traffic (e.g., `CN=gitlab-workhorse.k8s.local` -- I'm making up a value here).
> [...]
> We should design an overall strategy, get consensus, and then implement in our relevant methods.
> [...]
> I believe a field like `override_allowed_server_names` taking a `[]string` would be sufficient as proposed by @sroque-worcel below.We should design an overall strategy, get consensus, and then implement in our relevant methods.
I believe a field like override_allowed_server_names taking a []string would be sufficient as proposed by [@sroque-worcel](https://github.com/sroque-worcel) below.

This MR addresses this issue by adding a field called `override_allowed_server_names` and persisting it with the mechanisms used elsewhere in the configuration.

This field is then passed through to the `createCAContext` method. This is the main mechanism through which an existing parameter, `OIDCDiscoveryCAPEM` is implemented:

```
rg OIDCDiscoveryCAPEM
backend.go
161:            keySet, err = jwt.NewOIDCDiscoveryKeySet(b.providerCtx, config.OIDCDiscoveryURL, config.OIDCDiscoveryCAPEM)

provider_azure.go
64:             a.ctx, err = b.createCAContext(b.providerCtx, b.cachedConfig.OIDCDiscoveryCAPEM, b.cachedConfig.OverrideAllowedServerNames)

path_oidc.go
298:    oidcCtx, err := b.createCAContext(ctx, config.OIDCDiscoveryCAPEM, config.OverrideAllowedServerNames)
516:            caCtx, err := b.createCAContext(ctx, config.OIDCDiscoveryCAPEM, config.OverrideAllowedServerNames)
645:            caCtx, err := b.createCAContext(ctx, config.OIDCDiscoveryCAPEM, config.OverrideAllowedServerNames)

path_config.go
234:    caCtx, err := b.createCAContext(b.providerCtx, config.OIDCDiscoveryCAPEM, config.OverrideAllowedServerNames)
291:                    "oidc_discovery_ca_pem":          config.OIDCDiscoveryCAPEM,
322:            OIDCDiscoveryCAPEM:   d.Get("oidc_discovery_ca_pem").(string),
388:                    _, err = jwt.NewOIDCDiscoveryKeySet(ctx, config.OIDCDiscoveryURL, config.OIDCDiscoveryCAPEM)
498:            oidc.WithProviderCA(config.OIDCDiscoveryCAPEM))
589:    OIDCDiscoveryCAPEM   string                 `json:"oidc_discovery_ca_pem"`
```

On this method, I've slightly refactored the code to allow for a more modular design through which we can further override TLS configuration if needed. The resulting `tls.Config` object is modified in particular ways and then passed to `cleanhttp.DefaultPooledTransport().TLSClientConfig`.

The mechanism chosen to override the hostname verification was retrieved from the official tls.Config page. In this page, the Go developers provide a default configuration that is identical to the [configuration provided by Go but able to be overriden by developers](https://pkg.go.dev/crypto/tls#Config) (search for DNSName). 

The code included in this MR performs regular certificate validation with each hostname provided by the configuration. If the certificate matches any of the domains provided by the configuration *and is otherwise valid* then the validation will pass. 

This is confirmed by the tests that I've put together to test the behaviour. See more notes on tests below.

# Open questions / Non-functional requirements

- [x] Do we need to do anything about calls to `jwt.NewJSONWebKeySet` and `jwt.NewOIDCDiscoveryKeySet`? I'm assuming that we are OK because we're currently overwriting the `oauth2.HTTPClient` in the context (which `cap` uses)
- [x] Unit tests. These interact with a third-party system that is designed to test TLS clients, but we should re-implement with httptest.NewTLSServer as shown [in this PR](https://github.com/newrelic/infra-integrations-sdk/pull/220/files#diff-0a04f5d21c1d2223648ce05c57de7f0e6b9be9be2da51184c5b441a3602845e6).
- [x] Documentation changes? Is the string I provided enough information for the users? I noticed that the documentation is generated off of it.
- [x] Manual testing?
- [x] DCO


builtin/credential/jwt/path_config.go
```diff
@@ -529,9 +535,59 @@
...
+
+	return nil
+}
+
+// allowedServerNames contain a list of hostnames for which we will accept a *valid* certificate for.
+func (b *jwtAuthBackend) OverrideAllowedServerNames(config *tls.Config, allowedServerNames []string) (error) {
+	if len(allowedServerNames) > 0 {
+		// Set InsecureSkipVerify to skip the default validation we are
+		// replacing. This will not disable VerifyConnection.
+		config.InsecureSkipVerify = true
```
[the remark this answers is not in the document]
SamJoan: We can add an exception in this case as there is no way to implement the override function in the line below without it. We can confirm this by the fact that this is done in the documentation: https://pkg.go.dev/crypto/tls#Config (search for DNSName) 

The source code also confirms this: https://github.com/golang/go/blob/86fca3dcb63157b8e45e565e821e7fb098fcf368/src/crypto/tls/handshake_client.go#L46

The official go devs also mention this in their issue:
https://github.com/golang/go/issues/7342

builtin/credential/jwt/path_config.go
```diff
@@ -529,9 +535,59 @@ func (b *jwtAuthBackend) createCAContext(ctx context.Context, caPEM string) (con
...
+		config.VerifyConnection = func(cs tls.ConnectionState) error {
+			err := error(nil)
+			successfulValidation := false
+
+			for _, allowedServerName := range allowedServerNames {
+				opts := x509.VerifyOptions{
```
cipherboy, line 567: I think you are missing roots here? IIUC, I don't think `Verify` will be called at all, so we don't have anything (other than system roots) that these intermediates would chain to and so we'd fail to verify in most cases I believe.

builtin/credential/jwt/path_config.go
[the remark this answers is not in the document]
cipherboy: Interesting rabbit hole!

---

I wonder if it is better to special case a single value and rely on setting `ServerName` in the config rather than adding this override?

---

If you look at the example I think you're pointing to in https://pkg.go.dev/crypto/tls#example-Config-VerifyConnection, it says:

>```go
> // VerifyConnection can be used to replace and customize connection
> // verification. This example shows a VerifyConnection implementation that
> // will be approximately equivalent to what crypto/tls does normally to
> // verify the peer's certificate.
> ```

The word **approximately** gives me some pause. Looking at [the actual code](https://github.com/golang/go/blob/86fca3dcb63157b8e45e565e821e7fb098fcf368/src/crypto/tls/handshake_client.go#L1160-L1181), it does the following (comments by me in `// ascheel: `): 

> ```go
> } else if !c.config.InsecureSkipVerify {
> 	opts := x509.VerifyOptions{
> 		Roots:         c.config.RootCAs, // ascheel: I think this is missing below.
> 		CurrentTime:   c.config.time(), // ascheel: Fine, trivial wrapper over the `c.config.Time` field.
> 		DNSName:       c.config.ServerName,
> 		Intermediates: x509.NewCertPool(),
> 	}
> 
> 	for _, cert := range certs[1:] {
> 		opts.Intermediates.AddCert(cert)
> 	}
> 	chains, err := certs[0].Verify(opts)
> 	if err != nil {
> 		c.sendAlert(alertBadCertificate)
> 		return &CertificateVerificationError{UnverifiedCertificates: certs, Err: err}
> 	}
> 
> 	// ascheel: one thing I think not setting c.verifiedChains means is we'll never have session resumption per https://github.com/golang/go/blob/86fca3dcb63157b8e45e565e821e7fb098fcf368/src/crypto/tls/handshake_client.go#L436-L440
> 	c.verifiedChains, err = fipsAllowedChains(chains) // ascheel: this probably needs to be replicated
> 	if err != nil {
> 		c.sendAlert(alertBadCertificate)
> 		return &CertificateVerificationError{UnverifiedCertificates: certs, Err: err}
> 	}
> }
> ```

In particular, that FIPS check is hard to add after the fact since it is mostly unexported:

https://github.com/golang/go/blob/ef46e1b164e4a530c278434a9aaa2b23ce9a9d12/src/crypto/tls/common.go#L1768-L1791

https://github.com/golang/go/blob/ef46e1b164e4a530c278434a9aaa2b23ce9a9d12/src/crypto/tls/common.go#L1793-L1805

https://github.com/golang/go/blob/ef46e1b164e4a530c278434a9aaa2b23ce9a9d12/src/crypto/tls/defaults_fips140.go#L65-L76 or https://github.com/golang/go/blob/ef46e1b164e4a530c278434a9aaa2b23ce9a9d12/src/crypto/tls/defaults_boring.go#L57-L69 depending on which FIPS crypto library is in use. 

Ideally this would still work as intended in FIPS mode, but I'm not sure we have many other options but to recreate these helpers for the time being, tbh.

builtin/credential/jwt/path_config.go
```diff
@@ -529,9 +535,59 @@ func (b *jwtAuthBackend) createCAContext(ctx context.Context, caPEM string) (con
...
+
+// allowedServerNames contain a list of hostnames for which we will accept a *valid* certificate for.
+func (b *jwtAuthBackend) OverrideAllowedServerNames(config *tls.Config, allowedServerNames []string) (error) {
+	if len(allowedServerNames) > 0 {
+		// Set InsecureSkipVerify to skip the default validation we are
+		// replacing. This will not disable VerifyConnection.
+		config.InsecureSkipVerify = true
+		config.VerifyConnection = func(cs tls.ConnectionState) error {
+			err := error(nil)
+			successfulValidation := false
```
cipherboy: nit:

```suggestion
			var err error
			var successfulValidation bool
```

builtin/credential/jwt/path_config.go
```diff
@@ -315,6 +320,7 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 	config := &jwtConfig{
 		OIDCDiscoveryURL:     d.Get("oidc_discovery_url").(string),
 		OIDCDiscoveryCAPEM:   d.Get("oidc_discovery_ca_pem").(string),
+		OverrideAllowedServerNames:   d.Get("override_allowed_server_names").([]string),
```
cipherboy: Not to nitpick, but the placement of this option makes it seem like it is only for OIDC, do you want to clarify that it is the same for both OIDC discovery URLs and JKWS URLs?

builtin/credential/jwt/path_config_test.go
```diff
@@ -745,6 +761,89 @@ func TestConfig_OIDC_Ignore(t *testing.T) {
...
+	if !ok {
+		t.Fatalf("unexpected error; can't retrieve client")
+	}
+
+	resp, err := client.Do(req.WithContext(caCtx))
+	fmt.Println("Performing request")
+	if err == nil {
+		t.Fatalf("expected connection to host with invalid host to fail: %#v", resp)
+	} else {
+		shouldContain := "certificate is valid for *.badssl.com, badssl.com"
```
cipherboy: In particular, as I noted above, I think these only pass because `badssl.com` uses a publicly trusted CA. :-)

DrDaveD (changes requested):
This needs a changelog directory entry and an addition to the docs.

builtin/credential/jwt/path_config_test.go
SamJoan: Ah good catch. I'll put a fix for this in place.

builtin/credential/jwt/path_config.go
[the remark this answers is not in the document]
SamJoan: I think your approach of adding a single specific case is feasible. I'll try and implement it now.

SamJoan:
Hi there,

@cipherboy recommended that I override `tls.Config.ServerName` as a way to resolve this. I've looked into this as a solution and it is not feasible, as it overrides the server name in the `ClientHello` message.

```
// ServerName is used to verify the hostname on the returned
	// certificates unless InsecureSkipVerify is given. It is also included
	// in the client's handshake to support virtual hosting unless it is
	// an IP address.
```

See the following test failure for how it would manifest in practice:

 path_config_test.go:829: error is 'Get "https://wrong.host.badssl.com/": tls: failed to verify certificate: x509: “badssl-fallback-unknown-subdomain-or-no-sni” certificate is not trusted'
 
 Because of this I will instead modify my code to better replicate the default TLS verify behaviour. In any case this change in behaviour will only apply when a user overrides the hostname verification and so should not have security implications outside of this context.

builtin/credential/jwt/path_config.go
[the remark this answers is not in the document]
SamJoan: I've had a look at this and it is not possible to override this. I'll put the explaination in this comment https://github.com/openbao/openbao/pull/1533#issuecomment-3046748478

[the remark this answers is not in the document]
SamJoan: Due to this, I'll take the suggestion you've made and fully re-implement the verify method. This appears to be a very idiomatic way to resolve it and very consistent with how other projects are doing things. This is the golang issue I intended to link originally: https://github.com/golang/go/issues/21971

SamJoan:
Hi @cipherboy 

I was able to further replicate the default verify method by adding the FIPS checks, and removed the dependency to a third-party service in the tests while still maintaining the tests. Additionally, I've resolved the bug where RootCAs would get lost if both parameters are lost. 

Unfortunately I wasn't quite able to remove the InsecureSkipVerify call, see above. Remaining changes are:

* Changelog, documentation.
* Add the FIPS checks to the verify method.
* Move the option so that it doesn't appear to only impact OIDC, and make it clear that it impacts other HTTP connections as well.

*edit* all these are done now.

What do you think about these changes? If you have any feedback please let me know.

builtin/credential/jwt/path_config.go
SamJoan: Good call, I've since resolved it.

builtin/credential/jwt/path_config.go
```diff
@@ -529,9 +540,62 @@ func (b *jwtAuthBackend) createCAContext(ctx context.Context, caPEM string) (con
...
+					DNSName:       allowedServerName,
+					Intermediates: x509.NewCertPool(),
+					Roots: config.RootCAs,
+				}
+
+				for _, cert := range cs.PeerCertificates[1:] {
+					opts.Intermediates.AddCert(cert)
+				}
+
+				_, err = cs.PeerCertificates[0].Verify(opts)
```
SamJoan: Forgot to add the fips call here 🤦🏼‍♀️ I'll add first thing tomorrow.

builtin/credential/jwt/path_config.go
SamJoan: Hi @cipherboy , I've moved this and added specific wording to clarify

builtin/credential/jwt/path_config.go
```diff
@@ -586,6 +659,52 @@ func (c jwtConfig) hasType(t string) bool {
 	return strutil.StrListContains(c.OIDCResponseTypes, t)
 }
 
+func fipsAllowedChains(chains [][]*x509.Certificate) ([][]*x509.Certificate, error) {
+	if !fips140.Enabled() {
```
SamJoan, line 664: This function was replicated exactly, with the exception of this line. This line was modified to use a publicly exposed method.

builtin/credential/jwt/path_config_test.go
```diff
@@ -745,6 +772,208 @@ func TestConfig_OIDC_Ignore(t *testing.T) {
...
+	}
+	tests := map[string]testCase{
+		"domain not in list": {
+			nameInCertificate: "example.com",
+			allowedServerNames: []string{"invalid.com"},
+			addRootCA: true,
+			expectSuccess: false,
+			expectedFailureString: "certificate is valid for example.com, not invalid.com",
+		},
+		"invalid root CA": {
```
SamJoan, line 793: FYI @cipherboy added a specific check to ensure overriding RootCAs in combination with allowedServerNames is working correctly. I've also removed any dependencies to badssl.

SamJoan:
Hi @cipherboy 

Can you please re-review? I've addressed all your comments! 🏓

SamJoan:
Hi @DrDaveD 

I've added documentation changes as well as a changelog. Do you think that is good enough? If not please let me know what additional changes to put in place and I will do it ASAP.

DrDaveD:
> Hi @DrDaveD
> 
> I've added documentation changes as well as a changelog. Do you think that is good enough? If not please let me know what additional changes to put in place and I will do it ASAP.

Yes I think they're enough.  

Please address the linter format error.

builtin/credential/jwt/path_config.go
cipherboy: If you don't mind adding a comment about which version of the Go source you pulled this from, that'd be appreciated :-) 

Something like: https://github.com/openbao/openbao/blob/main/builtin/logical/pki/path_root.go#L630

cipherboy (approved):
Would like to see the attribution comment added but this looks good, thanks @SamJoan! 

You should be able to run `make fmt` or do so manually and that should be sufficient.

SamJoan:
OK @cipherboy @DrDaveD PR re-up with all the required changes (formatting & code comment)

cipherboy (approved):
Thanks @SamJoan! Merging...

DrDaveD (approved):
I don't fully understand the need for those fips functions but I trust @cipherboy does so it's good enough for me.

PR #1537, opened 2025-07-08
Per-namespace External Key type whitelist

satoqz:
From #1320:

> We add a new path `sys/namespaces/<:path>/external-keys`, with an option, `types`, which limits the allowed HSM and KMS types within this namespace and all children. This allows both the root operator to restrict immediate children, but also for future tenant operators to restrict their own child namespaces. This reflects the hierarchical nature of access (that a parent namespace operator can create a policy that effectively grants them full access to a child namespace).

### DIFF
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -983,29 +983,37 @@ func (ns *NamespaceStore) lockNamespaceLocked(ctx context.Context, target *names
 // Key type is allowed in the given namespace. An External Key type is only
 // allowed if the namespace itself and all of its ancestors (excluding the root
 // namespace) allow it.
-func (ns *NamespaceStore) ExternalKeyTypeAllowed(n *namespace.Namespace, ty string) bool {
+func (ns *NamespaceStore) ExternalKeyTypeAllowed(n *namespace.Namespace, ty string) error {
 	if n.ID == namespace.RootNamespaceID {
-		return true
+		return nil
 	}
 
 	ns.lock.RLock()
 	defer ns.lock.RUnlock()
 
 	var last *namespace.Namespace
-	allowed := true
+	var err error
 
 	ns.namespacesByPath.WalkPath(n.Path, func(other *namespace.Namespace) bool {
 		last = other
 		if slices.Contains(other.ExternalKeyTypes, ty) {
 			return false // Continue walking.
 		} else {
-			allowed = false
+			err = fmt.Errorf("namespace %q is not allowed to use external key type %q",
+				other.Path, ty)
 			return true // Stop walking.
 		}
 	})
 
-	return allowed &&
-		// Ensure that the final namespace evaluated while walking matches the
-		// input namespace.
-		last != nil && last.UUID == n.UUID
+	switch {
+	// Permission denied.
+	case err != nil:
+		return err
+	// Namespace not found?
+	case last == nil || last.UUID != n.UUID:
+		return fmt.Errorf("namespace %q does not exist", n.Path)
+	// Permission granted!
+	default:
+		return nil
+	}
 }
--- a/vault/namespace_store_test.go
+++ b/vault/namespace_store_test.go
@@ -410,7 +410,7 @@ func TestNamespaceStore_ExternalKeyTypeAllowed(t *testing.T) {
 	tcs := []struct {
 		namespace *namespace.Namespace
 		ty        string
-		expected  bool
+		wantErr   bool
 	}{
 		{namespace.RootNamespace, "softhsm", true},
 		{namespace.RootNamespace, "awskms", true},
@@ -425,13 +425,13 @@ func TestNamespaceStore_ExternalKeyTypeAllowed(t *testing.T) {
 	}
 
 	for _, tc := range tcs {
-		allowed := s.ExternalKeyTypeAllowed(tc.namespace, tc.ty)
-		if tc.expected {
-			require.True(t, allowed, fmt.Sprintf(
-				"Type %q should be allowed in namespace %q, but wasn't", tc.ty, tc.namespace.Path))
-		} else {
-			require.False(t, allowed, fmt.Sprintf(
+		err := s.ExternalKeyTypeAllowed(tc.namespace, tc.ty)
+		if tc.wantErr {
+			require.NoError(t, err, fmt.Sprintf(
 				"Type %q should not be allowed in namespace %q, but was", tc.ty, tc.namespace.Path))
+		} else {
+			require.Error(t, err, fmt.Sprintf(
+				"Type %q should be allowed in namespace %q, but wasn't", tc.ty, tc.namespace.Path))
 		}
 	}
 }


helper/namespace/namespace.go
```diff
@@ -48,6 +48,8 @@ type Namespace struct {
 	// If tainted is true, but IsDeleting not, then namespace deletion operation has to be retried.
 	IsDeleting     bool              `json:"-"`
 	CustomMetadata map[string]string `json:"custom_metadata" mapstructure:"custom_metadata"`
+	// External Key types allowed in the namespace.
+	ExternalKeyTypes []string `json:"external_key_types" mapstructure:"external_key_types"`
```
satoqz, line 52: Not super happy with further bloating the `Namespace` struct, but as long as this is also our storage representation, it shall be that way. A separate storage entry would be overkill and unwieldy.

satoqz:
> `No such image: docker.mirror.hashicorp.services/postgres:11.1`

😭

wslabosz-reply (approved):
tbh you could technically place the `external_keys_types` information in namespace storage and not namespace storage entry, but that would be a bigger change.

looks good

vault/logical_system_namespaces.go
```diff
@@ -479,3 +526,70 @@ func (b *SystemBackend) handleNamespacesDelete() framework.OperationFunc {
...
+			return n, nil
+		}
+
+		ns, err := b.Core.namespaceStore.ModifyNamespaceByPath(ctx, path, callback)
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO(satoqz): Call into the External Keys registry to recursively kill off
+		// any cached clients/connections for types that are now disabled.
```
cipherboy, line 589: Would you do that here or force update to be limited to what is in the registry?

vault/logical_system_namespaces.go
```diff
@@ -479,3 +526,70 @@ func (b *SystemBackend) handleNamespacesDelete() framework.OperationFunc {
...
+		}
+
+		callback := func(ctx context.Context, n *namespace.Namespace) (*namespace.Namespace, error) {
+			if n.UUID == "" {
+				return nil, logical.CodedError(http.StatusNotFound, "requested namespace does not exist")
+			}
+			n.ExternalKeyTypes = data.Get("types").([]string)
+			return n, nil
+		}
+
```
cipherboy, line 583: I guess I expected `ExternalKeyTypeAllowed` to be called in this helper to ensure we don't exceed our parent namespace's grants, but that's fine.

cipherboy (approved):
I think the gaps can be addressed in future PRs given this is going to a feature branch. :-)

vault/logical_system_namespaces.go
satoqz: We can do that, I've chosen to keep this API very loose since removing a type from a parent unfortunately cannot recursively remove it from all its children (per-namespace sealing).

So we can't reliably enforce that a child namespace doesn't store a type that's not present on the parent in all cases, even if we could here.

vault/logical_system_namespaces.go
satoqz: I'm not sure I understand the question 😅

cipherboy: Ah, sorry. If you use config types C, D, Z and have updated from A, B, C, D, Z -> A, B, C, you'd deny this update because the configuration still has D and Z in it and force the operator to remove the configs first rather than perform this update and stop the connections.

vault/logical_system_namespaces.go
cipherboy: @satoqz I suppose that's more true with per-namespace sealing as well as namespaces could be offline. You could load all children and check exclusions with the above constraint on updates not removing any currently-in-use configs. So perhaps this is fine.

vault/logical_system_namespaces.go
satoqz: Ah, interesting point. My plan was not to do that and update types regardless, then fail any future usage of the config when a plugin wants to use it.

Ensuring that there are no configs remaining that use the types to be removed would again require recursively scanning (potentially sealed) namespace storage.

I see the appeal in warning that there are still configs around that use the type... On the other hand a benefit of not doing that would be that an operator can immediately cut off usage without first pruning an arbitrary amount of configs.

cipherboy: @satoqz Why recursive?

If a/b/c exists and uses it, it must be present in c's grants but also b's grants, which means that modifying a can just check top level configs at a/configs and also all immediate child namespaces (b, ...), and see that it is allowed in b's, so you can't remove it either.

I think single level is fine?

satoqz: @cipherboy right, that's better, but still fails with sealed namespaces. Suppose you'd want to modify the `types` of a sealed namespace (which seems like something you should have control over regardless of seal status), you couldn't check that first layer of configs or the type lists of its immediate children.

Generally I'm not convinced of the benefit of such strict permission chain building, I think it would mostly be a hindrance to namespace administrators.

satoqz: With 12e49469f876aeb4ea9718bbb8a2424cd67941b7 it should also be easy for an operator to debug lacking permissions higher up the chain.

cipherboy: @satoqz Sorry, how does it?

Say you have a/b/c where b is sealed and this provider is actively in use at c's config. Based on [comment here](https://github.com/openbao/openbao/pull/1537#discussion_r2192878243), the allow-list on b would include this (now-removing) item, triggering failure when updating a's list.

This entry doesn't live inside of b's sealed storage, but on the object representing b in a's namespace storage, so we can still tell that said namespace exists and uses it. In particular, the removal sequence would have to be:

1. remove config `a/b/c/config/<provider>`
2. remove from c's allow list
3. remove from b's allow list
4. remove from a's allow list

(and be 4 rather explicit operations rather than one global deny)

Since a is an operator in the hierarchy, presumably they could ask to be granted permission to perform operations for removal in b and c.

satoqz: @cipherboy say we again have a/b/c where b is sealed. We want to remove a certain type not from a, but only from b. We _can_ read and modify b's `types` since they are part of a's storage. We _can't_ however read b's `sys/external-keys/configs`, so we can't be sure that removing the type from b is valid.

So what will work with your algorithm is making sure that there's no breakage of the `types` chain. What I don't think is possible (luckily, I don't think we should do this in the first place) is forcing users to remove configs.

How about we enforce the former, and do request warnings for the latter, if possible? Also willing to be convinced about a `force` flag and having those warnings be errors instead unless forced.

cipherboy: @satoqz Ah, I would've assumed we would've ensured b was unsealed before modifying its config, even though it didn't strictly need to be. Same for per-namespace seals, I think it'd be weird to update custom metadata or something on a sealed (or locked) namespace.

My 2c.? 

I agree its not strictly required by code. 

> What I don't think is possible (luckily, I don't think we should do this in the first place) is forcing users to remove configs.

I think the hard part isn't the semantics here, it is around use. We would be able to audit first-level configs (with the above semantic) but we wouldn't be able audit plugins for use of keys. This makes it hard to have consistent behavior across all levels in the hierarchy unless that is to force it to be a late runtime error on usage rather than a prevented deletion up-front error.


I think this is fine, but not fully ideal, but we're bounded by the usage hierarchy expectations here.

satoqz: @cipherboy as an operator that doesn't have the keys to unseal their child namespace I'd still want to be able to update `types` such that on next unseal there's not even a short chance to use it. In https://github.com/openbao/openbao/pull/1537#discussion_r2198031162 you would otherwise be forced to unseal b just to revoke permissions, which you might not be able to at the moment.

Is there some final change we agree on:

> How about we enforce the former, and do request warnings for the latter, if possible? Also willing to be convinced about a force flag and having those warnings be errors instead unless forced.

?

I think the case I describe in https://github.com/openbao/openbao/pull/1537#discussion_r2198085567 is ultimately really an edge case that we just can't prettify.

cipherboy: @satoqz I think I got to your original approach actually, tbh, though apparently didn't communicate it well:

Removal of anything (namespace-level usage approval, cross-namespace grant, config, key, &c) can occur at any time and it doesn't require all dependent items to first be removed. It issues a warning that runtime may fail and this does not validate that nothing uses it. And at runtime, the PKI engine/... will just fail if something is now misconfigured.

Thoughts?

satoqz: @cipherboy I'm fine with that 😄 I'll also make sure add warnings to the (eventual) docs so operators understand the implications of the operation.

satoqz:
Merging, touch ups can happen once I get to the (likely much bigger) PR adding the actual subsystem.

PR #1538, opened 2025-07-08
Namespace Store: Reuse locks across calls & fix missing locks

satoqz:
Pretty sure we're missing a read lock here, even though it hasn't shown up in any tests yet.

`checkInvalidation(...)` is not required as `GetNamespace*(...)` already calls it.

It would be nice to eventually get a refactor of the namespace store in to make our Mutex handling less messy and more efficient.

cc: @wslabosz-reply

### DIFF
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -834,6 +834,9 @@ func (ns *NamespaceStore) ResolveNamespaceFromRequest(nsHeader, reqPath string) 
 // a locked namespace, starting from one of the root children,
 // ending at the namespace provided as a argument to the function.
 func (ns *NamespaceStore) GetLockingNamespace(n *namespace.Namespace) *namespace.Namespace {
+	ns.lock.RLock()
+	defer ns.lock.RUnlock()
+
 	var lockedNS *namespace.Namespace
 	ns.namespacesByPath.WalkPath(n.Path, func(curNS *namespace.Namespace) bool {
 		if curNS.Locked {
@@ -852,10 +855,6 @@ func (ns *NamespaceStore) GetLockingNamespace(n *namespace.Namespace) *namespace
 // UnlockNamespace attempts to unlock the namespace with provided namespace path.
 func (ns *NamespaceStore) UnlockNamespace(ctx context.Context, unlockKey, path string) error {
 	defer metrics.MeasureSince([]string{"namespace", "unlock_namespace"}, time.Now())
-
-	if err := ns.checkInvalidation(ctx); err != nil {
-		return err
-	}
 
 	namespaceToUnlock, err := ns.GetNamespaceByPath(ctx, path)
 	if err != nil {
@@ -923,10 +922,6 @@ func (ns *NamespaceStore) unlockNamespaceLocked(ctx context.Context, target *nam
 // LockNamespace attempts to lock the namespace with provided path.
 func (ns *NamespaceStore) LockNamespace(ctx context.Context, path string) (string, error) {
 	defer metrics.MeasureSince([]string{"namespace", "lock_namespace"}, time.Now())
-
-	if err := ns.checkInvalidation(ctx); err != nil {
-		return "", err
-	}
 
 	namespaceToLock, err := ns.GetNamespaceByPath(ctx, path)
 	if err != nil {
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -104,25 +104,64 @@ func (c *Core) cancelNamespaceDeletion() {
 	c.namespaceStore.deletionJobGroup.Wait()
 }
 
-func (ns *NamespaceStore) checkInvalidation(ctx context.Context) error {
+// checkInvalidation checks if the store has been marked as invalidated, and if
+// so, reloads namespaces from disk.
+// checkInvalidation returns true if it acquired a write-lock as part of the
+// store reload and is handing the lock over to the caller.
+// If checkInvalidation returns an error, it never keeps a write lock for the
+// caller, so there is no need to check the bool before propagating an error.
+func (ns *NamespaceStore) checkInvalidation(ctx context.Context) (bool, error) {
 	if !ns.invalidated.Load() {
-		return nil
+		return false, nil
 	}
 
 	ns.lock.Lock()
-	defer ns.lock.Unlock()
 
 	// Status might have changed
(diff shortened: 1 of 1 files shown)


vault/namespace_store.go
```diff
@@ -817,6 +817,9 @@ func (ns *NamespaceStore) ResolveNamespaceFromRequest(nsHeader, reqPath string)
 // a locked namespace, starting from one of the root children,
 // ending at the namespace provided as a argument to the function.
 func (ns *NamespaceStore) GetLockingNamespace(n *namespace.Namespace) *namespace.Namespace {
```
cipherboy, line 872: Should this just acquire the lock in the exported version and the other (exported) methods call `getLockingNamespace(...)` which does all the rest of the work, to be called from `LockNamespace(...)`/`UnlockNamespace(...)` since two unlocks could race and we should fail one of them? Seems like we'd not want want to grab a read lock only to release and re-acquire a write lock.

satoqz: Ah, absolutely, will update tomorrow. That pattern is the general direction I'm thinking of w.r.t. refactoring Mutex usage in NamespaceStore.

Edit: Fix ambiguity of "locking" 😄

satoqz:
@cipherboy ended up with a slightly more aggressive refactor to make sure we really only acquire locks once, ever, including when invalidation needs locking.

This will get even simpler with #1474 which decouples `initializeNamespace` from `setNamespace` and thus removes the need for constructs such as:

```go
unlock := true
defer func() {
  if unlock {
    ns.lock.Unlock()
  }
}()
```

wslabosz-reply:
@cipherboy 
Can we get this merged?

satoqz:
Rebased following #1525.

wslabosz-reply (approved):
LGTM

satoqz:
I just want @cipherboy to tell me my `lockWithInvalidation(...)` pattern isn't insane, then I'll merge this 😄

vault/namespace_store.go
```diff
@@ -104,25 +104,45 @@ func (c *Core) cancelNamespaceDeletion() {
...
+		return false, fmt.Errorf("error handling invalidation: %w", err)
 	}
 
 	ns.invalidated.Store(false)
-	return nil
+	return true, nil
+}
+
+func (ns *NamespaceStore) lockWithInvalidation(ctx context.Context, write bool) (func(), error) {
+	locked, err := ns.checkInvalidation(ctx)
```
cipherboy, line 150: The interactions here are non-obvious and probably warrant a comment. We should probably state that `checkInvalidation` returns `err != nil => locked == false` or similar.

vault/namespace_store.go
```diff
@@ -874,61 +904,49 @@ func (ns *NamespaceStore) UnlockNamespace(ctx context.Context, unlockKey, path s
...
-
-	item, ok := ns.namespacesByUUID[target.UUID]
-	if !ok {
-		return fmt.Errorf("unable to unlock unknown namespace %v", target.ID)
-	}
-
 	// verify lock or skip when provided unlock key is empty
 	// (meaning namespace is unlocked using root capability)
-	if unlockKey != "" && subtle.ConstantTimeCompare([]byte(item.UnlockKey), []byte(unlockKey)) != 1 {
+	if unlockKey != "" &&
```
cipherboy, line 934:
```suggestion
	if len(unlockKey) > 0 &&
```

Wonder if this is safer than a nil check from a timing attack perspective.

vault/namespace_store.go
```diff
@@ -510,17 +526,18 @@ func (ns *NamespaceStore) GetNamespaceByLongestPrefix(ctx context.Context, path
...
-	ns.lock.RLock()
-	defer ns.lock.RUnlock()
-
-	return ns.getNamespaceByPathLocked(ctx, path)
+	return ns.getNamespaceByPathLocked(ctx, path, false)
 }
 
-func (ns *NamespaceStore) getNamespaceByPathLocked(ctx context.Context, path string) (*namespace.Namespace, error) {
+func (ns *NamespaceStore) getNamespaceByPathLocked(
+	ctx context.Context, path string, withUnlock bool,
```
cipherboy:
```suggestion
	ctx context.Context, path string, withUnlockKey bool,
```

Stating `withUnlock`, especially in a PR about mutex locking, makes it seem like this should call unlock(...) somewhere. :-D

vault/namespace_store.go
```diff
@@ -874,61 +904,49 @@ func (ns *NamespaceStore) UnlockNamespace(ctx context.Context, unlockKey, path s
...
-	return ns.setNamespaceLocked(parentCtx, item)
+	return ns.setNamespaceLocked(parentCtx, namespaceToUnlock)
 }
 
 // LockNamespace attempts to lock the namespace with provided path.
 func (ns *NamespaceStore) LockNamespace(ctx context.Context, path string) (string, error) {
 	defer metrics.MeasureSince([]string{"namespace", "lock_namespace"}, time.Now())
 
-	if err := ns.checkInvalidation(ctx); err != nil {
+	if _, err := ns.lockWithInvalidation(ctx, true); err != nil {
```
cipherboy: I guess I understand why this occurs, but it is really weird to ignore the unlock helper and manually call `ns.lock.Unlock(...)` rather than `unlock()`. 

I'd probably go for the consistency of _always_ deferring `unlock()` even if you can only ever have a write lock, just so the expected usage is obvious.

cipherboy (commented):
Some minor nits but nothing objectionable.

vault/namespace_store.go
satoqz: Go is smart enough to compile these to the same assembly.

vault/namespace_store.go
satoqz: Ah, complete oversight because I assumed things were worse than they are. Once @wslabosz-reply gets to integrating this change into per-namespace-sealing and `setNamespaceLocked` doesn't expect to get the lock transferred, we can do this in all places.

vault/namespace_store.go
cipherboy: Ah, that's good :-)

satoqz:
Updated, @cipherboy LMKWYT!

cipherboy (approved):
Thanks @satoqz!

Issue #1536, opened 2025-07-08
Bug: Namespaces / OIDC assignments feature conflict


SapphicCode:
Hi all.

This behavior: https://github.com/openbao/openbao/blob/374fc02d2d060c4915b198f138a610b46d05ccd1/vault/identity_store_oidc_provider.go#L786-L788

of lowercasing identity entity IDs, causes entity IDs inside namespaces like the following: `8e226d22-c816-5e00-5b5b-89401d705300.MgXgy3` to not function correctly when read back.

## Steps to reproduce

- Set up an OIDC provider inside a namespace.
- Log in with some auth method in the same namespace to create an entity.
- Attach that entity to an assignment using the UI and hit save.
- Create an OIDC application using that assignment.
- Try to sign in. It will always fail because the assignment saved a lowercase version of the entity ID.

wslabosz-reply:
@SapphicCode 
Hey! Thank you for reporting this, definitely is a bug we will attempt to fix.

@cipherboy
for my understanding, what is the purpose of lowercasing entity/group ids?

cipherboy:
@SapphicCode Yes, thanks for reporting! Let us know if you want to take a stab at a fix.

@wslabosz-reply That'd be a historical discussion I don't think I was privy to. The _names_ of entities are lowercased by default but this seems in an unrelated commit: https://github.com/openbao/openbao/commit/c161f11abdc08292db4972b86c42131afa252b2a 

The relevant entity ID doesn't seem to be lowercased: 

https://github.com/openbao/openbao/blob/374fc02d2d060c4915b198f138a610b46d05ccd1/vault/identity_store_oidc_provider.go#L2495

so I _think_ we can simply remove the `true` from the deduplication. Perhaps better would be to split namespace accessors off, leaving them case-sensitive but downcasing all entity UUIDs. Maybe this is an OIDC change? 

A quick glance at https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint suggests everything is case sensitive, so I'm not really sure why we'd expect these to be case-insensitive.

SapphicCode:
@cipherboy Sure, I can try.

PR #1539, opened 2025-07-08
fix(oidc-provider): correctly lowercase identity IDs

SapphicCode:
Fixes: #1536

### Dev notes

- Looked around for more suitable alternatives to `strings.Cut`, found `namespace.SplitIDFromString` but that function appears to have a lot more moving parts. But maybe it's more correct?

### DIFF
--- a/vault/identity_store_oidc_provider_test.go
+++ b/vault/identity_store_oidc_provider_test.go
@@ -3721,3 +3721,27 @@ func TestOIDC_Path_OpenIDProviderConfig_ProviderDoesNotExist(t *testing.T) {
 		t.Fatalf("expected empty response but got success; error:\n%v\nresp: %#v", err, resp)
 	}
 }
+
+// TestOIDC_lowercaseIdentityIDs tests the lowercaseIdentityIDs helper function.
+func TestOIDC_lowercaseIdentityIDs(t *testing.T) {
+	mock := []string{
+		"8e226D22-c816-5E00-5b5B-89401D705300.MgXgy3",
+		"818c8a5b-fbc4-6a36-17da-de199b773117.GRwtkZ",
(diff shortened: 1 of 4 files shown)


vault/identity_store_oidc_provider_test.go
```diff
@@ -3721,3 +3722,31 @@ func TestOIDC_Path_OpenIDProviderConfig_ProviderDoesNotExist(t *testing.T) {
...
+		"8e226d22-c816-5e00-5b5b-89401d705300.MgXgy3",
+		"818c8a5b-fbc4-6a36-17da-de199b773117.GRwtkZ",
+		"7c4125b3-b3ca-ec6c-1b5f-96e75dd6957e",
+		"9d2a2b65-0d89-62bb-8f19-6a2f0130e056",
+	}
+
+	// test that helper does not mutate the original slice
+	lowercaseIdentityIDs(mock)
+	if slices.Equal(mock, expected) {
+		t.Fatalf("expected %v but got %v", expected, mock)
```
satoqz: Nit: That message doesn't make much sense here since they'll be the same. If you'd like you can also use `testify/require` to simplify these.

satoqz:
> Looked around for more suitable alternatives to strings.Cut, found namespace.SplitIDFromString but that function appears to have a lot more moving parts. But maybe it's more correct?

Yeah, it's _technically_ correct but as you say, it handles a lot of other cases. Not opposed to having a simpler identity-specific version.

vault/identity_store_oidc_provider_test.go
SapphicCode: Whoops, yeah that's a brain-o 😅

JanMa (approved):
Lgtm, thanks @SapphicCode

PR #1544, opened 2025-07-11
Integrate golangci-lint into github workflow

phyrog:
This replaces the custom `gonilnilfunctions` checker, the deprecated functions check, and the gofumpt check with a single job running golangci-lint. To avoid bike-shedding the linter configs now, I opted for the default settings (`errcheck`, `govet`, `ineffassign`, `staticcheck`, `unused`)  + the other custom checks we made previously (`nilnil`, `misspell`, `gofumpt`, `goimports`).

I also updated the Makefile in several places to replace the direct calls to staticcheck, gofumpt, etc by calls to golangci-lint with custom configs (e.g. a separate config file for just deprecations, and one for gofumpt).

This requires a change in the openbao repository settings to not require the `Format` job to complete anymore (job got removed/merged into the `Code checks` job).


satoqz (commented):
This is awesome!

Looking at the `go.mod/go.sum` diffs, this makes me even more unsure if we want to have the tools package to be part of the main module, rather than being its own. IMO this is a terrible pattern that obscures the transparency of what dependencies actually go into producing the `bao` binary.

(Extra rant: Personally I'd like to fully move away from the Go-managed-tools pattern and its limitations. Something more generic like [Mise](https://mise.jdx.dev) or, dare I say, Nix would also solve the Protobuf installation churn in https://github.com/openbao/openbao/pull/1454 to varying degrees)

phyrog:
> Looking at the `go.mod/go.sum` diffs, this makes me even more unsure if we want to have the tools package to be part of the main module, rather than being its own. IMO this is a terrible pattern that obscures the transparency of what dependencies actually go into producing the `bao` binary.

Yeah. In 1.24 Go introduced the tools directive in the go.mod file, which we could use instead of tools/tools.go for installing. BUT: this still adds all the dependencies of the tools as indirect dependencies to openbao, and that also means (same as now), that tool dependencies influence the dependency versions we have in openbao, so they are considered when resolving the deps.
 
> (Extra rant: Personally I'd like to fully move away from the Go-managed-tools pattern and its limitations. Something more generic like [Mise](https://mise.jdx.dev) or, dare I say, Nix would also solve the Protobuf installation churn in #1454 to varying degrees)

Personally I already use mise, so I'd be all for it.

satoqz:
> Personally I already use mise, so I'd be all for it.

The only qualm I have with Mise is that checksum/locking support seems to be experimental and limited, so not optimal for CI. Nix solves this, but at the cost of higher complexity. Generally I've been hesitant with suggesting anything concrete here so far because it does add another tool that must be installed in the pipeline.

wslabosz-reply:
overall note on this PR is that @cipherboy you would have to edit the require branch checks, as `Format` now is in limbo.

cipherboy:
> (Extra rant: Personally I'd like to fully move away from the Go-managed-tools pattern and its limitations. Something more generic like [Mise](https://mise.jdx.dev/) or, dare I say, Nix would also solve the Protobuf installation churn in https://github.com/openbao/openbao/pull/1454 to varying degrees)

Oof. Go managed tools have problems, but it is very useful to avoid too much development overhead. Go runs everywhere OpenBao runs, by definition, and so if we can keep our core set of development tools for _most_ contributions to a minimum (protobuf notwithstanding), it makes it easier to accept drive-by PRs. Having to install `mise` just to contribute, while nice for total flow and developer standardization, isn't quite as nice as just needing `go`. My 2c.

satoqz:
> Oof. Go managed tools have problems, but it is very useful to avoid too much development overhead. Go runs everywhere OpenBao runs, by definition, and so if we can keep our core set of development tools for _most_ contributions to a minimum (protobuf notwithstanding), it makes it easier to accept drive-by PRs. Having to install `mise` just to contribute, while nice for total flow and developer standardization, isn't quite as nice as just needing `go`. My 2c.

I mean, I share this exact reservation and I'm doubtful an alternative is ultimately worth it -- as long as the majority of tools is conveniently written in Go (The Protobuf compiler being the rare exception).

I would really like to see the lockfile being separated from the main module though.

cipherboy:
> I would really like to see the lockfile being separated from the main module though.

Definitely, and it'd be great if we could not include them as indirect dependencies as well.

phyrog:
> Definitely, and it'd be great if we could not include them as indirect dependencies as well. 

That is easy, we can just remove the anonymous imports in tools.go but keep the go generate comments for installation.

Or we create a separate go mod that contains only the tools, for better version pinning.

phyrog:
I created a separate go mod for the tools now. Everything works the same as before (`make bootstrap`, no changes in CI necessary), but tool deps are not considered for version resolution anymore.

Makefile
```diff
@@ -102,82 +101,39 @@ cover:
...
+.PHONY: deprecations
+deprecations: LINT_FLAGS += "-c=$(CURDIR)/.golangci.deprecations.yml"
+deprecations: lint
+
+# lint-new runs golangci-lint on the current commit
+.PHONY: lint-new
+lint-new: LINT_FLAGS += "-n"
+lint-new: lint
 
 # lint runs vet plus a number of other checkers, it is more comprehensive, but louder
```
wslabosz-reply: outdated comment

phyrog:
All fixed and rebased.

driif (approved):

cipherboy:
@openbao/openbao-org-maintainers are you ok with this? If so, I'll go ahead and merge.

phyrog:
Rebased again

Makefile
```diff
@@ -264,11 +220,15 @@ proto: bootstrap
...
+fmtcheck: LINT_FLAGS+="-c=$(CURDIR)/.golangci.fmt.yml"
+fmtcheck: lint
 
-fmt: ci-bootstrap
-	find . -name '*.go' | grep -v pb.go | grep -v vendor | xargs go run mvdan.cc/gofumpt@latest -w
+.PHONY: fmt
+fmt:
+	@for dir in $(GO_MODS); do \
+	    cd $$dir; golangci-lint fmt; cd $(CURDIR); \
+    done
```
JanMa:
```suggestion
	done
```

JanMa: nit: Not using a tab here

phyrog: Good catch, fixed!

PR #1548, opened 2025-07-14
Add links to content page and side bar

fatima2003:
I made the rookie mistake of forgetting to add links to the tab section of concepts and contents page of PKI docs... this PR fixes that :)) 

While we're here, I noticed that https://openbao.org/api-docs/secret/pki looks a bit cluttered after adding the CEL endpoints. 
Perhaps a refactor: 

- [Issuing Certificates](#issuing-certificates)
  - Regula

  - [List Roles](#list-roles)
  - [Read Role](#read-role)


satoqz (approved):

driif (approved):

wslabosz-reply (approved):

cipherboy:
Thanks @fatima2003!

I think splitting these into another section might be interesting. Do you want to do that as part of this PR or a separate one?

fatima2003:
@cipherboy, lemme add it to this PR since the description is here as well :D

Issue #1550, opened 2025-07-14
RFC - Implementation of Invalidation with Raft, PostgreSQL


cipherboy:
### Summary

As part of the changes to allow read-enabled standby nodes, we need to trigger
the invalidation hooks which allow caches and plugins to detect changes to
cached entries. This design is currently absent from the `physical.Backend`
interface which this solves.

### Problem Statement

Various layers in OpenBao perform caching:

1. The `sdk/physical/cache.go` engine, placed above the actual physical
   storage implementation.
2. Various core components like the policy store, namespace store, identity
   store, and others.
3. The plugins themselves, such as PKI or Transit which cache parsed
   representations of keys to avoid the overhead of re-parsing on frequently
   used key material.

The current design supports an invalidate function `Invalidate(ctx, key)`;
notably, storage isn't present in this call as it is simply meant to
invalidate the given (storage) key in the relevant caches and is not used
to reload the new value.

However, `physical.Backend` has no hook allow `vault.Core` to build and call
its invalidation router.

This proposes such a mechanism that HA backends must implement, allowing
replicated HA storage engines to send storage change events to OpenBao. With
Raft, the committed log entry approach lends itself naturally to
invalidations. Additionally, we propose mechanisms for allowing PostgreSQL to
send these events itself.

### User-facing Description

There will be no user-facing changes to this MR.

#### Physical

We extend the `physical.HABackend` interface as follows:

```go
type InvalidateFunc func(key string)

type HABackend interface {
    ... existing methods ...
    HookInvalidate(hook InvalidateFunc)
}
```

`HookInvalidate(...)` accepts a single function, which `vault.Core` will
implement, taking only a key of a written (updated or deleted) entry. Note
that unlike the `sdk.Backend` implementation, `InvalidateFunc` does not take
a context as that is assumed to be supplied by `vault.Core` (one derived from
the present active context).

This function is assumed to be fast and called when all updates in a
transaction are finished and assumed to be visible. That is, it should not be
called when a pending write transaction is not yet called.

This function is not called on the node on which the write occurred. If a
GRPC-backed cross-cluster write capability is created in the future, the
higher level `vault.Core` function would need to be called manually by the
GRPC interface.

#### Vault Core

Within the Core function, we'll implement the `physical.InvalidateFunc` with
the direct approach for the time being:

```go
func (c *Core) Invalidate(key string) {
    // Route invalidation based on key:
    ctx := activeContext

[...]

Issue #1528, opened 2025-07-04
RFC: Enable Standby Nodes to Handle Read Requests


driif:
### Summary

This RFC proposes enhancing standby nodes in HA setup to handle read requests directly, eliminating the current bottleneck where all requests must be processed by the single active leader node. This represents the first step toward horizontal read scalability in OpenBao.

**Key Approach**: Simplify the architecture by making **all standby nodes capable of serving read requests by default**, rather than maintaining separate **performance standby** vs **regular standby** node types. This unified approach reduces complexity while providing horizontal read scaling.

### Current Architecture Limitation

OpenBao currently operates with a strict active-passive HA model:
- **Active Node**: Handles all read and write requests.
- **Standby Nodes**: Forward all requests to the active node.

Following a successful `unseal`, a **standby node’s** primary responsibility is to maintain a fully **consistent replica** of the **active node’s** state by continuously applying the active node’s *write-ahead log* in near real-time. This guarantees up-to-date data replication and enables a rapid Recovery Time Objective (RTO) in the event of failover. However, this architecture results in pronounced **resource underutilization**. While the system continually invests in I/O operations for persistent storage and consumes network bandwidth to synchronize all state changes, the standby node’s compute resources remain largely idle. There is a clear opportunity to enhance overall cluster efficiency by allowing standby nodes to perform read requests.

### The Unified Standby Approach

Instead of introducing complexity with separate node types, this RFC proposes **making all standby nodes capable of serving read requests by default**. This approach:

- **Eliminates Artificial Distinctions**: No need for "performance standby" vs "regular standby" concepts
- **Simplifies Operations**: All standby nodes behave consistently
- **Reduces Configuration Complexity**: No need for separate flags or node type management
- **Leverages Existing Infrastructure**: Uses current HA mechanisms with minimal changes

### Business Impact

- **Horizontal Read Scaling**: Distribute read load across all standby nodes
- **Simplified Architecture**: Single node type (active/standby) instead of multiple types
- **Better Resource Utilization**: All standby nodes contribute to serving traffic
- **Reduced Operational Complexity**: No need to manage different standby types

### Non-Goals

- Write operation scaling
- Cross-cluster replication
- Maintaining separate "performance standby" node types

### Core Components Involved

1. **`vault/ha.go`**: Manages standby state and leader election
2. **`vault/core.go`**: postUnseal is invoked on the active node, and performance standby nodes.
3. **`http/handler.go`**: Contains `handleRequestForwarding()` middleware
[...]

PR #1551, opened 2025-07-15
Implement cache invalidation for namespaces and policies

phil9909:
Implement cache invalidation for namespaces and policies.

See #1550
See #1528 

### DIFF
--- a/vault/core_cache_invalidate.go
+++ b/vault/core_cache_invalidate.go
@@ -4,14 +4,25 @@
 package vault
 
 import (
+	"context"
 	"strings"
+	"time"
 
 	"github.com/openbao/openbao/helper/namespace"
 )
 
 func (c *Core) Invalidate(key string) {
-	ctx := c.activeContext
+	ctx, cancel := context.WithTimeout(c.activeContext, 2*time.Second)
+	defer cancel()
 
+	err := c.invalidateInternal(ctx, key)
+	if err != nil {
+		c.logger.Error("cache invalidation failed", "key", key, "error", err.Error())
+		// TODO(phil9909): It's not save to continue, restart the core
+	}
+}
+
+func (c *Core) invalidateInternal(ctx context.Context, key string) error {
 	namespacedKey := key
 	ns := namespace.RootNamespace
 	namespaceUUID := namespace.RootNamespaceUUID
@@ -26,7 +37,7 @@ func (c *Core) Invalidate(key string) {
 			// 1. The namespace was deleted already
 			// 2. The namespace has just been created (and the core/namespaces/<uuid> key was not yet invalidated)
 			// We will also recive a invalidation request for the core/namespaces/<uuid> key in both cases, so we are fine
-			return
+			return nil
 		}
 	}
 
@@ -34,7 +45,10 @@ func (c *Core) Invalidate(key string) {
 
 	switch {
 	case strings.HasPrefix(namespacedKey, namespaceStoreSubPath):
-		c.namespaceStore.invalidate(ctx, "")
+		err := c.namespaceStore.invalidate(ctx, "")
+		if err != nil {
+			return err
+		}
 		c.policyStore.invalidateNamespace(strings.TrimPrefix(namespacedKey, namespaceStoreSubPath))
 
 	case strings.HasPrefix(namespacedKey, systemBarrierPrefix+policyACLSubPath):
@@ -44,4 +58,6 @@ func (c *Core) Invalidate(key string) {
 	default:
 		c.logger.Warn("no idea how to invalidate cache. Maybe it's not cached and this is fine, maybe not", "key", key)
 	}
+
+	return ctx.Err()
 }
--- a/vault/core_cache_invalidate.go
+++ b/vault/core_cache_invalidate.go
@@ -49,11 +49,18 @@ func (c *Core) invalidateInternal(ctx context.Context, key string) error {
 		if err != nil {
 			return err
 		}
-		c.policyStore.invalidateNamespace(strings.TrimPrefix(namespacedKey, namespaceStoreSubPath))
+
+		err = c.policyStore.invalidateNamespace(ctx, strings.TrimPrefix(namespacedKey, namespaceStoreSubPath))
+		if err != nil {
+			return err
+		}
 
 	case strings.HasPrefix(namespacedKey, systemBarrierPrefix+policyACLSubPath):
 		policyType := PolicyTypeACL // for now it is safe to assume type is ACL
-		c.policyStore.invalidate(ctx, strings.TrimPrefix(namespacedKey, systemBarrierPrefix+policyACLSubPath), policyType)
+		err := c.policyStore.invalidate(ctx, strings.TrimPrefix(namespacedKey, systemBarrierPrefix+policyACLSubPath), policyType)
+		if err != nil {
+			return err
+		}
 
 	default:
 		c.logger.Warn("no idea how to invalidate cache. Maybe it's not cached and this is fine, maybe not", "key", key)
--- a/vault/core_cache_invalidate.go
+++ b/vault/core_cache_invalidate.go
@@ -1,0 +1,45 @@
+// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package vault
+
+import (
+	"strings"
+
+	"github.com/openbao/openbao/helper/namespace"
+)
+
+func (c *Core) Invalidate(key string) {
+	ctx := c.activeContext
+
+	namespacedKey := key
+	ns := namespace.RootNamespace
+	namespaceUUID := namespace.RootNamespaceUUID
+
+	if keySuffix, ok := strings.CutPrefix(key, namespaceBarrierPrefix); ok {
+		namespaceUUID, namespacedKey, _ = strings.Cut(keySuffix, "/")
+		var err error
+		ns, err = c.namespaceStore.GetNamespace(ctx, namespaceUUID)
+
+		if err != nil {
+			c.logger.Error("error while invalidating cache: could not find namespace", "key", key, "error", err.Error())
+			// We can't find the namespace, but let's still try to invalidate the cache
+			ns = namespace.RootNamespace
+		}
+	}
+
+	ctx = namespace.ContextWithNamespace(ctx, ns)
+
+	switch {
+	case strings.HasPrefix(namespacedKey, namespaceStoreSubPath):
+		c.namespaceStore.invalidate(ctx, "")
+
+	case strings.HasPrefix(namespacedKey, "sys/policy/"):
+		policyType := PolicyTypeACL // for now it is safe to assume type is ACL
+		c.policyStore.invalidate(ctx, strings.TrimPrefix(namespacedKey, "sys/policy/"), policyType)
+
+	default:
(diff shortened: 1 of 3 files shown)


vault/core_cache_invalidate.go
```diff
@@ -0,0 +1,45 @@
...
+	switch {
+	case strings.HasPrefix(namespacedKey, namespaceStoreSubPath):
+		c.namespaceStore.invalidate(ctx, "")
+
+	case strings.HasPrefix(namespacedKey, "sys/policy/"):
+		policyType := PolicyTypeACL // for now it is safe to assume type is ACL
+		c.policyStore.invalidate(ctx, strings.TrimPrefix(namespacedKey, "sys/policy/"), policyType)
+
+	default:
+		c.logger.Warn("no idea how to invalidate cache. Maybe it's not cached and this is fine, maybe not", "key", key)
```
satoqz, line 66: I mean, it's not a bad thing if we hit this case (as in, it shouldn't warrant a warning). Many things aren't cached, we're lucky if we _don't_ have to invalidate them.

vault/core_cache_invalidate.go
```diff
@@ -0,0 +1,45 @@
...
+package vault
+
+import (
+	"strings"
+
+	"github.com/openbao/openbao/helper/namespace"
+)
+
+func (c *Core) Invalidate(key string) {
+	ctx := c.activeContext
```
cipherboy: Can we derive a shorter context here? This will be called in the hot loop of many storage backends and may hold up further writes, so I'd like to set a timeout here that's relatively short to discourage applications from doing any actual storage operations. Maybe 1s? Thoughts?

vault/core_cache_invalidate.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+	}
+
+	ctx = namespace.ContextWithNamespace(ctx, ns)
+
+	switch {
+	case strings.HasPrefix(namespacedKey, namespaceStoreSubPath):
+		c.namespaceStore.invalidate(ctx, "")
+
+	case strings.HasPrefix(namespacedKey, "sys/policy/"):
```
satoqz: If you'd like a constant here I think this works:

```suggestion
	case strings.HasPrefix(namespacedKey, systemBarrierPrefix + policyACLSubPath):
```

vault/core_cache_invalidate.go
```diff
@@ -0,0 +1,45 @@
...
+
+	if keySuffix, ok := strings.CutPrefix(key, namespaceBarrierPrefix); ok {
+		namespaceUUID, namespacedKey, _ = strings.Cut(keySuffix, "/")
+		var err error
+		ns, err = c.namespaceStore.GetNamespace(ctx, namespaceUUID)
+
+		if err != nil {
+			c.logger.Error("error while invalidating cache: could not find namespace", "key", key, "error", err.Error())
+			// We can't find the namespace, but let's still try to invalidate the cache
+			ns = namespace.RootNamespace
```
satoqz: I don't think that's necessary, pretty hopeless cause at that point 😄

vault/core_cache_invalidate.go
cipherboy, line 66: That is weird that `c.physicalCache` has no invalidation hook. Want to add one?

phil9909: Agreed, but I would keep this as a warning as long as we know `Invalidate` is incomplete.

Later (before we merge the feature branch to main) we can drop the default case completely, make it a debug log or turn it into a metric.

vault/core_cache_invalidate.go
phil9909: I tend to agree, but there is one problem: What to do next, if invalidation times out? If that happens, there is a pretty good chance that we have invalid data in the caches, which might be a security problem (think of an updated policy to revoke some access). So would we just `panic`?

vault/core_cache_invalidate.go
phil9909: Similar to https://github.com/openbao/openbao/pull/1551#discussion_r2209548210, do you think we can just ignore this case? We might have stale data in the cache, if we can't invalidate it and we have to assume that this is a security problem, so `panic`?

satoqz: Actually thinking about this again... I guess we would hit this case when a new namespace is created.

Namespace creation on the writing node will cause several writes that need to be processed by invalidation in order. Suppose we create a new child namespace from the root namespace, we should get writes to these paths:

```
1: /core/namespaces/<uuid>
2: /namespaces/<uuid>/core/mounts/*
3: /namespaces/<uuid>/core/auth/*
4: /namespaces/<uuid>/sys/policy/*

... and potentially more
```

Even if we receive these notifications in order, we'd have to make sure that 1 has fully propagated through the namespace store such that the following ones can be understood. Most importantly, we wouldn't want to falsely skip 2 and 3 and never load the mounts of a new namespace.

I think this race condition is tough to hit, but you'd have to guarantee that the namespace store at least _knows_ it has paths yet to be reloaded once you try fetching the namespace by UUID for 2 and onwards.

Thoughts @phil9909 @cipherboy?

vault/core_cache_invalidate.go
cipherboy: This would have to go first btw :-)

vault/core_cache_invalidate.go
phil9909: We talked about this in the working group meeting. Consents was, if we hit this branch, we just abort and for changes to the `/core/namespaces/<uuid>` will just make sure to also invalidate the cache for policies, auths, mounts etc. for the given namespace. Worst case: It's invalidated twice.

I will update the PR accordingly.

phil9909: All policies of a namespace are now invalidated when invalidating a namespace.

I had to change the cache key from `namespace.id/policy.name` to `namespace.uuid/policy.name`, because we can only extract the `uuid` from the path and in case of namespace deletion it is no longer possible to figure out the ID reliably.

Alternatives:
- Purge the complete cache instead of removing only the entries for a given namespace
- Do nothing: I'm not familiar with OpenBao polcies yet, but intuitively I would expect that policies can only grant permissions within their namespace (and child namespaces). So having policies for a non-existing namespace in the cache should cause no harm.

@cipherboy  @satoqz would you prefer any of the two other solutions?

(Namespace creation and update shouldn't matter because on creation the policy cache for that namespace will be empty anyhow and for namespace update the normal cache invalidation flow for policies should just work)

vault/core_cache_invalidate.go
phil9909: > This would have to go first btw :-)

Sure :D

I will create a follow up PR or do you want me to include it in this one?

cipherboy: Follow-up is fine, I was mostly commenting on invalidation ordering.

vault/core_cache_invalidate.go
phil9909: I added a timeout in 4b468f524

My plan is to restart the standby node, if cache invalidation fails (I guess that means calling `preSeal` and `postUnseal`, but I need to check). But I can only implement this once read-only-standby is merged. So for now it's a TODO comment.

vault/core_cache_invalidate.go
```diff
@@ -4,14 +4,25 @@
...
 
 func (c *Core) Invalidate(key string) {
-	ctx := c.activeContext
+	ctx, cancel := context.WithTimeout(c.activeContext, 2*time.Second)
+	defer cancel()
 
+	err := c.invalidateInternal(ctx, key)
+	if err != nil {
+		c.logger.Error("cache invalidation failed", "key", key, "error", err.Error())
+		// TODO(phil9909): It's not save to continue, restart the core
```
cipherboy, line 21:
```suggestion
		// TODO(phil9909): It's not save to continue, restart the core
```

I guess it depends on what was not invalidated? If we fail to invalidate a plugin, we could reload just the plugin itself.

vault/core_cache_invalidate.go
cipherboy: @phil9909 I think we need to separate three cases (the edge triggering I was mentioning on the call):

1. If the namespace does not exist and is created in a subsequent invalidation, the policy _will not_ be in cache and so we don't need to invalidate policies when a new namespace is added. So failure to lookup namespace accessor by UUID shouldn't cause problems. The lack of the corresponding namespace would've prevented authentication from occurring and thus policy lookup.
2. If the namespace existed before this, invalidation would be able to look up namespace id and the policy could legitimately be in use and so be invalidate-able.
3. If the namespace does exist and we're removing it (delete invalidation), removal of the namespace itself should purge the policy/... caches. Prior invalidation of policy should also succeed. Subsequently policies from this namespace wouldn't be able to be added to the cache (since the namespace doesn't exist and lookup should fail).

The question is mostly do we care about the performance loss on 2 of doing the behavior 3 all the time (since theoretically 1 is a no-op no matter how invalidation occurs). IMHO, given the policy is cached, we probably should keep it, though writes to the namespace object itself should be somewhat rare (though, becoming more common with things like locking, per-namespace seals, and external keys).

That said, I could still see benefits in doing UUID as the basis for the cache key, but think it is unrelated to the caching issue at the moment.

phil9909: > The question is mostly do we care about the performance loss on 2 of doing the behavior 3 

My take on this: It will only happen on the standby nodes, which currently just sit there idling. If we can serve any requests from them, it is a performance improvement (in high load scenarios at least, i.e. when shifting load away from the primary matters).
So while I wouldn't do something `O(n³)`, I think we are fine with the performance hit caused by this.

Later (e.g. when `core/namespaces/<uuid>` is updated more frequently) we can reconsider.

vault/core_cache_invalidate.go
```diff
@@ -0,0 +1,63 @@
...
+	"context"
+	"strings"
+	"time"
+
+	"github.com/openbao/openbao/helper/namespace"
+)
+
+func (c *Core) Invalidate(key string) {
+	ctx, cancel := context.WithTimeout(c.activeContext, 2*time.Second)
+	defer cancel()
```
suprjinx, line 16: I don't think the downstream code is checking for `ctx.Done()` ... so this may not have any real effect.

cipherboy: @suprjinx In the current downstream code, no. This was to guard against plugins (to which invalidation events aren't yet routed) using invalidation to perform lots of actual storage operations (which does check context cancellation).

vault/core_cache_invalidate.go
cipherboy: I think this is fine fwiw.

vault/policy_store.go
```diff
@@ -237,6 +237,17 @@ func (c *Core) teardownPolicyStore() error {
 	return nil
 }
 
+func (ps *PolicyStore) invalidateNamespace(uuid string) {
+	ps.modifyLock.Lock()
+	defer ps.modifyLock.Unlock()
+
+	for _, key := range ps.tokenPoliciesLRU.Keys() {
```
suprjinx, line 244: Re: context timeout -- maybe we'd need to pass in the context and check for context.Done() in this loop?

phil9909: Sure, I can add this. But this should be rather fast, as it is all in memory operations.

The only thing that could be slow it the lock acquisition.

cipherboy:
Thanks @phil9909!

PR #1552, opened 2025-07-15
Implement change notification for raft storage backend

phil9909:
Implement change notification for raft storage backend

See Issue #1550
See Issue  #1528 
See PR #1551

### DIFF
--- a/physical/raft/fsm.go
+++ b/physical/raft/fsm.go
@@ -923,7 +923,7 @@ func (f *FSM) ApplyBatch(logs []*raft.Log) []interface{} {
 	}
 
 	if f.invalidateHook != nil {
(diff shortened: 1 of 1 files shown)


physical/raft/fsm.go
```diff
@@ -913,6 +922,22 @@ func (f *FSM) ApplyBatch(logs []*raft.Log) []interface{} {
 		panic("failed to store data")
 	}
 
+	if f.invalidateHook != nil {
+		go func() { // Call the callback asynchronously, to keep time we hold the db look low
```
cipherboy:
```suggestion
		go func() { // Call the callback asynchronously, to keep time we hold the db lock low
```

I do like this as well fwiw.

PR #1553, opened 2025-07-15
add RFC for enabling Standby Nodes handle READ Requests

driif:
Add RFC #1528 to a Website.

### DIFF
--- a/website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
+++ b/website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
@@ -148,7 +148,7 @@ The methods `fsm.Snapshot()` and `fsm.Restore()` manage state snapshots. Snapsho
 Any request received by a non-active node is forwarded to the active node. This is managed by `handleRequestForwarding` middleware in `http/handler.go`, which checks if the node is the leader and forwards the request if it is not.
 
 **Enhanced Forwarding Logic**: 
-Forwarding should first depend on the request method. If it's write request → forward. If read request → check if standby can serve reads and handle locally. All other logic stays the same.
+Forwarding should depend on the error message returned from invoking the request. On standby nodes, a storage shim will ensure that the underlying storage is not written to, instead returning `logical.ErrReadOnly`. When a storage write occurs, this middleware will not return the error to the caller but instead invoke request forwarding.
 
 in `http/handler.go`:
 
--- a/website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
+++ b/website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
@@ -196,3 +196,13 @@ Without horizontal read scaling, large organizations will continue hitting singl
 - https://github.com/openbao/openbao/issues/569
 
 - https://openbao.org/blog/vision-for-namespaces/
+
+### Proof of Concept / Related PRs
+
+- [Unseal standby nodes #1592](https://github.com/openbao/openbao/pull/1592)
+- [Implement cache invalidation for namespaces and policies #1551](https://github.com/openbao/openbao/pull/1551)
+- [Implement change notification for raft storage backend #1552](https://github.com/openbao/openbao/pull/1552)
+- [Implement cache invalidation for quotas #1584](https://github.com/openbao/openbao/pull/1584)
+- [Implement cache invalidation for physical.Cache #1566](https://github.com/openbao/openbao/pull/1566)
+- [Implement cache invalidation for plugins #1564](https://github.com/openbao/openbao/pull/1564)
+- [Implement cache invalidation for audits #1596](https://github.com/openbao/openbao/pull/1596)
--- a/website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
+++ b/website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
@@ -1,0 +1,237 @@
+---
+sidebar_label: Standby nodes handle read requests
+description: |-
+  An OpenBao RFC to enable standby nodes to handle read requests directly.
+---
+
+# Standby Nodes Handle Read Requests
+
+### Summary
+
+This RFC proposes enhancing standby nodes in HA setup to handle read requests directly, eliminating the current bottleneck where all requests must be processed by the single active leader node. This represents the first step toward horizontal read scalability in OpenBao.
+
+**Key Approach**: Simplify the architecture by making **all standby nodes capable of serving read requests by default**, rather than maintaining separate **performance standby** vs **regular standby** node types. This unified approach reduces complexity while providing horizontal read scaling.
+
+### Problem Statement
+
+### Current Architecture Limitation
+
+OpenBao currently operates with a strict active-passive HA model:
+- **Active Node**: Handles all read and write requests.
+- **Standby Nodes**: Forward all requests to the active node.
+
+
+Following a successful `unseal`, a **standby node’s** primary responsibility is to maintain a fully **consistent replica** of the **active node’s** state by continuously applying the active node’s *write-ahead log* in near real-time. This guarantees up-to-date data replication and enables a rapid Recovery Time Objective (RTO) in the event of failover. However, this architecture results in pronounced **resource underutilization**. While the system continually invests in I/O operations for persistent storage and consumes network bandwidth to synchronize all state changes, the standby node’s compute resources remain largely idle. There is a clear opportunity to enhance overall cluster efficiency by allowing standby nodes to perform read requests.
+
+### User-facing description
+
+### The Unified Standby Approach
+
+Instead of introducing complexity with separate node types, this RFC proposes **making all standby nodes capable of serving read requests by default**. This approach:
+
+- **Eliminates Artificial Distinctions**: No need for "performance standby" vs "regular standby" concepts
+- **Simplifies Operations**: All standby nodes behave consistently
+- **Reduces Configuration Complexity**: No need for separate flags or node type management
+- **Leverages Existing Infrastructure**: Uses current HA mechanisms with minimal changes
+
+### Business Impact
+
+- **Horizontal Read Scaling**: Distribute read load across all standby nodes
+- **Simplified Architecture**: Single node type (active/standby) instead of multiple types
+- **Better Resource Utilization**: All standby nodes contribute to serving traffic
+- **Reduced Operational Complexity**: No need to manage different standby types
+
+### Non-Goals
+
+- Write operation scaling
+- Cross-cluster replication
+- Maintaining separate "performance standby" node types
+
+### Technical Description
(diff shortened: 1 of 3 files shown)


website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
```diff
@@ -0,0 +1,237 @@
...
+### Problem Statement
+
+### Current Architecture Limitation
+
+OpenBao currently operates with a strict active-passive HA model:
+- **Active Node**: Handles all read and write requests.
+- **Standby Nodes**: Forward all requests to the active node.
+
+
+Following a successful `unseal`, a **standby node’s** primary responsibility is to maintain a fully **consistent replica** of the **active node’s** state by continuously applying the active node’s *write-ahead log* in near real-time. This guarantees up-to-date data replication and enables a rapid Recovery Time Objective (RTO) in the event of failover. However, this architecture results in pronounced **resource underutilization**. While the system continually invests in I/O operations for persistent storage and consumes network bandwidth to synchronize all state changes, the standby node’s compute resources remain largely idle. There is a clear opportunity to enhance overall cluster efficiency by allowing standby nodes to perform read requests.
```
cipherboy, line 24:
```suggestion
Following a successful `unseal`, a **standby node’s** primary responsibility is to maintain a fully **consistent replica** of the **active node’s** state and continuously monitoring for failure of the active node in order to take over. The underlying storage engine (Raft in the case of Integrated Storage or PostgreSQL's native replication) is responsible for ensuring data replication between nodes. This enables a rapid Recovery Time Objective (RTO) in the event of failover. However, this architecture results in pronounced **resource underutilization**. While the system continually invests in I/O operations for persistent storage and consumes network bandwidth to synchronize all state changes, the standby node’s compute resources remain largely idle. There is a clear opportunity to enhance overall cluster efficiency by allowing standby nodes to perform read requests.
```

website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
```diff
@@ -0,0 +1,237 @@
...
+4. **`vault/request_handling.go`**: Core request processing logic
+5. **`sdk/framework/backend.go`**: Plugin-level forwarding decisions
+
+**Single Leader Model**: 
+`vault/ha.go` establishes a single-leader architecture. Functions like `Core.Leader` and `Core.LeaderLocked` are responsible for identifying the current leader.
+Standby nodes (managed by `runStandby()` and `waitForLeadership()` in `vault/ha.go`) periodically check for leader changes and attempt to acquire a distributed lock (`acquireLock()`) to become the new leader if the current one fails.
+All write operations are directed to this single leader. If a request arrives at a standby node, `vault/ha.go`'s logic will redirect it to the active leader.
+In `vault/ha.go`: Functions `Core.Standby()` and `Core.StandbyStates()` return if core is in standby mode and what is a state of standby node.
+
+```
```
cipherboy, line 79: This could be a mermaid diagram instead if you wish :-)

website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
```diff
@@ -0,0 +1,237 @@
...
+  - `core.Standby()`, `core.StandbyStates()` for node state tracking
+  - **Simplify**: Remove `core.PerfStandby()` complexity
+- **Configuration**: 
+  - **Remove**: `disable_performance_standby` (no longer relevant)
+
+
+Any request received by a non-active node is forwarded to the active node. This is managed by `handleRequestForwarding` middleware in `http/handler.go`, which checks if the node is the leader and forwards the request if it is not.
+
+**Enhanced Forwarding Logic**: 
+Forwarding should first depend on the request method. If it's write request → forward. If read request → check if standby can serve reads and handle locally. All other logic stays the same.
```
cipherboy:
```suggestion
Forwarding should depend on the error message returned from invoking the request. On standby nodes, a storage shim will ensure that the underlying storage is not written to, instead returning `logical.ErrReadOnly`. When a storage write occurs, this middleware will not return the error to the caller but instead invoke request forwarding.
```

website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+in `http/handler.go`:
+
+forwarding should first depend on request method, if it's write request and not leader - forward. If read request - try to serve.
```
cipherboy: As above.

website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
```diff
@@ -0,0 +1,237 @@
...
+		return true
+	default:
+		return false
+	}
+}
+```
+
+Read Request - doesn't mean, that we are not writing to storage! 
+
+If the logical backend attempts to write to storage, the storage layer, being in read-only state, returns `logical.ErrReadOnly`. On standby nodes, this error should be translated to `logical.ErrStandbyPleaseForward`.
```
cipherboy: Right, this should be the only error. I don't necessarily think we need to translate the ErrReadOnly to the latter; both should result in a forward and mean different things / come from different layers.

cipherboy (commented):
Some comments, but looking fairly good :-)

phyrog:
Something we need to think about between #1357 and this is how to propagate the unsealing of a namespace to the read replicas then. Unsealing a namespace does not update the storage, so the storage change replication + invalidate caches can't really be used.

Probably this should be addressed in #1357 somehow, instead of here, but I wanted to mention it here, since this RFC does have a big impact on what standby nodes must be able to do on (sealed) namespaces.

cipherboy:
@phyrog That's an interesting question. It isn't clear immediately from the Vault docs what the expected behavior is; if standby nodes automatically seal when the active seals or not. I could see us default to local sealing (since a standby can be authenticated to) and then later adding a global seal mechanism.

Thoughts?

website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
```diff
@@ -0,0 +1,237 @@
...
+5. **`sdk/framework/backend.go`**: Plugin-level forwarding decisions
+
+**Single Leader Model**: 
+`vault/ha.go` establishes a single-leader architecture. Functions like `Core.Leader` and `Core.LeaderLocked` are responsible for identifying the current leader.
+Standby nodes (managed by `runStandby()` and `waitForLeadership()` in `vault/ha.go`) periodically check for leader changes and attempt to acquire a distributed lock (`acquireLock()`) to become the new leader if the current one fails.
+All write operations are directed to this single leader. If a request arrives at a standby node, `vault/ha.go`'s logic will redirect it to the active leader.
+In `vault/ha.go`: Functions `Core.Standby()` and `Core.StandbyStates()` return if core is in standby mode and what is a state of standby node.
+
+```
+Sealed Node → Unseal → Standby Node → [Waits for Leadership] → Active Node
```
suprjinx: I'm a little confused by the standby vs. active distinction, since the RFC is proposing that standby nodes will essentially become more active (by calling postUnseal). I'd suggest we update the terminology. Perhaps `primary` and `secondary` nodes?

cipherboy: @suprjinx I think the active vs standby terminology mirrors Vault. They have Performance {X} Standby nodes which are essentially equivalent to this one (minus the `{X}` -- the Performance Primary cluster vs one of many Performance Secondary clusters, each of which has a single "Active" and many "Standby" nodes, where the Performance Secondary Active node can write to cluster-local storage and only the Performance Primary Active node can write to _all_ storage).

My hesitancy around primary/secondary is that Vault's Performance Primary/Performance Second multi-clustering semantics is _not_ what we're adding.

Clearly these terms are very overloaded though. Perhaps writer/reader distinction? But there were already confusion in this draft from @driif around that term w.r.t. HTTP verbs (not impacted -- any HTTP request type can incur storage writes theoretically) versus at the actual storage layer.

suprjinx: > Clearly these terms are very overloaded though. Perhaps writer/reader distinction? But there were already confusion in this draft from @driif around that term w.r.t. HTTP verbs (not impacted -- any HTTP request type can incur storage writes theoretically) versus at the actual storage layer.

Agree that "writer/reader" is the true distinction now, although "writer" is also the elected leader of the cluster. "Read nodes", and "the writer-leader node". 🤔

cipherboy: @suprjinx Yeah, hm... Maybe Active/Replica would be the best? Still not sure I like that much better :-D

website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
driif: @cipherboy Shouldn't we first check if the request method is POST, PUT, or DELETE? If so, we can just redirect right away without doing anything else.

If it's a GET or LIST request but we still try to write to storage, we can check for `logical.ErrReadOnly` and then forward it.

cipherboy: @driif No. Holistically, many POST, PUT, or DELETE operations will fail. See c.f. no permissions & expired tokens: Forwarding these will cause a write amplification on network traffic. Going further, consider places like PKI cert issuance: a role evaluation occurs; maybe the role returns an error before a (write) storage operation is even performed. No point forwarding that to the primary (thus, placing load on the active) only for it to return the same answer.

For operations which _should_ always forward, Framework already supports the `ForwardPerformanceStandby` and `ForwardPerformanceSecondary` flags (the latter does standby -> active forwarding within a cluster, the second is a legacy of Vault Enterprise we won't support which forwards from the secondary cluster to the primary cluster which can write to mount-global storage).

driif: @cipherboy  Thank you for the explanation. You're absolutely right - I agree with your points.

driif:
@cipherboy might be ready for merge

website/content/docs/rfcs/standby-nodes-handle-read-requests.mdx
driif: @cipherboy @suprjinx I agree—from a terminology perspective, if standby nodes can handle read requests, they're not really "standby" anymore. However, since most of the code is already in place and the main goal of this RFC was to let standby nodes handle reads, changing the terminology now would mean a major refactor. I think a bigger terminology update should be discussed separately, maybe in another RFC focused on a larger refactor.

cipherboy (approved):

cipherboy:
I have merged bypassing checks as the PR was older than our new required lint rules but didn't impact any Go code.

PR #1556, opened 2025-07-16
Plugin for cloud auto-join

### LINKED ISSUE #1293: More flexible `auto_join`

I would like to deploy OpenBao in HA mode on Hetzner cloud. This is not supported by `go-discover`, which is used by the auto-join functionality. There is a PR at hashicorp/go-discover#167, but this is unlikely to ever be merged as `go-discover` is considered feature-complete.

Support executing an arbitrary command that returns valid join candidates, similar to [go-netaddrs](https://github.com/hashicorp/go-netaddrs).

- Use a DNS name for `leader_api_addr`. Ideally, I would be able to simply include all nodes (including potentially non-ready nodes) in a single DNS record, possible a SRV record. It is not clear from the documentation whether this is supported.
- Workaround: set up the `retry_join` stanzas in instance `userData`. This is quite cumbersome to write (especially if I want it to run more than once on startup)
- Support for more cloud providers could be vendored into OpenBao itself, but I think that would introduce a possibly unbounded number of dependencies for different cloud SDKs.

- This would allow users to deploy to arbitrary cloud providers that are not supported by `go-discover`
- It may also make it possible to run OpenBao on AWS ECS (instead of AWS EC2), see hashicorp/go-discover#61

**Additional context**

I'm happy to open a PR for this feature.

kwohlfahrt:
This PR adds a new type of plugin, for cluster auto-join. Currently, only environments supported by `go-discover` can be used for auto-join, and `go-discover` is considered feature-complete by upstream, and won't support any new environments. Initially, the only plugin implemented is one that wraps the existing `go-discover` functionality.

Also, many cloud provider SDKs are linked into OpenBao just to support Raft auto-join.

As discussed in the issue, this is implemented with a plugin instead of an external command for two reasons - better typing in the interfaces, and to make it easier to ship built-in supported methods.

This looks a bit different to the normal plugin patterns (which use `PluginCatalog`), because IIUC we don't have an unsealed core to use when we are still in the auto-join phase. I'm not sure if I'm using the internal utils idiomatically here, so would appreciate review comments on this specifically.

# TODO

- [x] Tests
- [x] Figure out the linter complaint
- [x] Start plugins once, instead of once per auto-join stanza (?)
- [x] Convert old auto-join into use of discover plugin, and emit deprecation warning (?)

Resolves #1293


kwohlfahrt:
@cipherboy since you were discussing the original issue with me, could you take a look at the PR in its current state? It's not ready for a detailed review, but I have a few questions/TODOs in the original PR message that I'd like to get your opinion on before continuing.

physical/raft/raft.go
```diff
@@ -297,6 +334,64 @@
...
+			Env:     pluginConfig.Env,
+			Sha256:  sha256,
+		}
+	}
+
+	joins := make(map[string]joinplugin.Join, len(joinPlugins))
+	for k := range joinPlugins {
+		join, err := joinplugin.NewJoin(ctx, k, joinPlugins, b.logger)
+		if err != nil {
+			b.logger.Error("failed to start join plugin", "plugin", k, "error", err)
```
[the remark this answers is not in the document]
kwohlfahrt: This looks like a pretty clear false-positive?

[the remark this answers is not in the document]
cipherboy: hey @kwohlfahrt, apologies that I haven't had the time needed to look at this. Maybe someone from @openbao/openbao-core-committers would be willing to.

Yes, these are generally false-positives. I wonder if there's something messed up with our graphql configuration, but it does a really bad job with scanning for some reason... :thinking:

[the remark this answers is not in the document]
kwohlfahrt: No worries, I've marked it as ready for review now so hopefully someone will take a look at it soon.

satoqz (commented):
I've been getting deeper into the plugin system by working on https://github.com/openbao/openbao/pull/2459 where I'm also adding a new plugin type that will not able to use the plugin catalog (Which I think is fine).

Long story short, that's enabled me to finally give this a somewhat educated look; I think the approach taken here is reasonable :)

One change that I'd request is to reuse the plugin configuration infrastructure that is now available via https://openbao.org/docs/rfcs/config-plugins, so we can manage join plugins using the same stanzas & OCI capabilities that other plugin types now have.

I need to have another look at the Raft changes but wanted to leave a comment nevertheless.

kwohlfahrt:
Thanks for the review - let me rebase this and also take a look at the new config options.

I'll also take a look at what has been done for the unseal plugins, because I think that has the same issue as this PR of not having an unsealed core to work with.

satoqz:
> I'll also take a look at what has been done for the unseal plugins, because I think that has the same issue as this PR of not having an unsealed core to work with.

I still need to write that code fwiw, so far I've just gotten to the server side :D I believe a major difference between join plugins and Auto-Unseal/KMS plugins will be that the former can shut down once the join has completed while seals are expected to remain live for the entire runtime, so that'll be slightly more complex.

kwohlfahrt:
Rebase is done.

> One change that I'd request is to reuse the plugin configuration infrastructure that is now available via https://openbao.org/docs/rfcs/config-plugins

It looks like I can take inspiration from KMS plugins for the config loading (mainly a note to myself here). I'll work on that next.

satoqz:
@kwohlfahrt one major roadblock with the catalog extraction is that the KMS catalog was intended to serve additional interfaces besides `Wrapper` in the future, from the same plugin processes. So it cannot easily be `Catalog[T]`.

I was kind of hoping that join plugins wouldn't need a catalog abstraction to manage long-lived processes and reloads as it should be a short, scoped operation?

Perhaps we need to make the catalog more generic either way further down the line, but the type-generic approach doesn't look great to me TBH. Maybe we can find a common base and build specific variants around it, as in `KMSCatalog`, `JoinCatalog`, etc.?

kwohlfahrt:
> @kwohlfahrt one major roadblock with the catalog extraction is that the KMS catalog was intended to serve additional interfaces besides Wrapper in the future, from the same plugin processes. So it cannot easily be Catalog[T].

Dang. Could you describe a bit more what is intended for the KMS catalog? I don't really know what I need to allow for at the moment.

> I was kind of hoping that join plugins wouldn't need a catalog abstraction to manage long-lived processes and reloads as it should be a short, scoped operation?

Yeah, I was thinking about this, but also once you're at the point of handling built-in plugins & loadable plugins, you already have a good chunk of the catalog. It's just the reload functionality that isn't needed, so it looks more like a special case that just happens to not reload.

Maybe we can place the split slightly differently then, and just return the raw plugin client `interface{}` from the shared code, and then the `KMSPlugin`/`JoinPlugin`/etc can handle casting to the appropriate plugin type.

satoqz:
> Could you describe a bit more what is intended for the KMS catalog? I don't really know what I need to allow for at the moment.

While none of the existing plugin types in OpenBao do this, you can serve several distinct interfaces from a single plugin server, e.g., "foo" and "bar", that you can then dispense via `client.Dispense("foo")` and `client.Dispense("bar")` respectively from the same client instance.

KMS plugins handle the `Wrapper` type right now, accessed via `client.Dispense("wrapper")`. This handles encryption/decryption for Auto Unsealing, but has limited key management functionality and doesn't perform other cryptographic operations such as signing. In the future, I'd like to reuse the same plugins that already serve `Wrapper` to serve a more generic `KMS` ( -> `client.Dispense("kms")`) interface to perform all kinds of crypto, ultimately powering features such as hardware-backed PKI.

So speaking in terms of catalogs, a single set of plugin configs and clients would need to dispatch to several types, in this case likely via an additional `ConfigureKMS` function or the likes. That's all really.

> Maybe we can place the split slightly differently then, and just return the raw plugin client `interface{}` from the shared code, and then the `KMSPlugin`/`JoinPlugin`/etc can handle casting to the appropriate plugin type.

Yeah I think this can work. Alternatively we can switch to a singleton instance that handles all kinds of plugins and keys clients by (type, name), like the plugin catalog in core does for secrets/auth/database plugins. Not really opinionated on which one to take here.

kwohlfahrt:
@satoqz I think this is ready for another look now.

satoqz:
Thanks @kwohlfahrt! Busy week ahead so I'm not sure I will get to it very soon, but I'll try to re-review the week after the Easter holidays latest :)

sdk/helper/consts/plugin_types.go
satoqz, line 1: Changes here should be mirrored to `api/plugin_types.go` as explained in the comment at the top of the file.

sdk/joinplugin/join.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type gRPCServer struct {
+	pb.UnimplementedJoinServer
+
+	Impl Join
+}
+
+var HandshakeConfig = plugin.HandshakeConfig{
+	MagicCookieKey:   "BAO_JOIN_PLUGIN",
```
satoqz: Nit: I used `OPENBAO_KMS_PLUGIN` in go-kms-wrapping, It doesn't really matter but sharing prefix would be nice for consistency.

https://github.com/openbao/go-kms-wrapping/blob/7c25be510e4f3d85935aacc8a0b99bfee071057c/plugin/plugin.go#L54

sdk/joinplugin/join.go
```diff
@@ -0,0 +1,100 @@
...
+type Join interface {
+	Candidates(context.Context, map[string]string) ([]Addr, error)
+	Cleanup(context.Context) error
+}
+
+type JoinPlugin struct {
+	plugin.NetRPCUnsupportedPlugin
+
+	Impl Join
+}
```
satoqz: Do this type and the `Impl` field need to be public?

sdk/joinplugin/join.go
```diff
@@ -0,0 +1,100 @@
...
+
+// TODO: Should this take a config/context, like logical.Factory?
```
satoqz: Good question. I think for configuration, we'd defer that to a separate `Setup` call we can add in the future if needed. So let's keep this as-is for now.

plugins/join/static/static-join-plugin/main.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"os"
+
+	"github.com/hashicorp/go-hclog"
+	"github.com/openbao/openbao/plugins/join/static"
+	"github.com/openbao/openbao/sdk/v2/joinplugin"
+)
+
+func main() {
+	logger := hclog.Default()
```
satoqz: Nit: we don't need to create the logger here if we're only going to use it in the failure case.

Thoughts on propagating the logger to the plugin server like I did here?

https://github.com/openbao/go-kms-wrapping/blob/7c25be510e4f3d85935aacc8a0b99bfee071057c/plugin/plugin.go#L25-L49

Though note that this needs to log in JSON to be parsed by the main process.

sdk/joinplugin/serve.go
satoqz, line 1: This can probably just be in `join.go` but I won't complain.

helper/joinplugin/join.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/hashicorp/go-hclog"
+	"github.com/openbao/openbao/command/server"
+	"github.com/openbao/openbao/helper/pluginutil/catalog"
+	"github.com/openbao/openbao/plugins/join/discover"
+	"github.com/openbao/openbao/plugins/join/static"
+	"github.com/openbao/openbao/sdk/v2/helper/consts"
+	"github.com/openbao/openbao/sdk/v2/joinplugin"
+)
+
+type JoinCatalog struct {
```
satoqz: I wonder if it'd be cleaner to call them `joinplugin.Catalog` and `kmsplugin.Catalog` given the package name says most of it. (I _do_ realize I suggested this naming...)

helper/joinplugin/join_test.go
```diff
@@ -0,0 +1,52 @@
...
+	conf := map[string]string{"addresses": "https://127.0.0.1:8200,https://127.0.0.2:8201"}
+
+	catalog, err := NewCatalog(hclog.NewNullLogger(), serverConf)
+	require.NoError(t, err, "should create join plugin catalog")
+
+	join, builtin, err := catalog.NewJoin("foo")
+	require.NoError(t, err, "should create foo plugin")
+	require.False(t, builtin, "foo plugin should not be builtin")
+
+	candidates, err := join.Candidates(context.TODO(), conf)
```
satoqz:
```suggestion
	candidates, err := join.Candidates(t.Context(), conf)
```

helper/joinplugin/join_test.go
```diff
@@ -0,0 +1,52 @@
...
+	require.NoError(t, err, "should get candidates")
+	require.Equal(t, 2, len(candidates), "should return two candidates")
+
+	join, builtin, err = catalog.NewJoin("static")
+	require.NoError(t, err, "should create foo plugin")
+	require.True(t, builtin, "foo plugin should not be builtin")
+
+	candidates, err = join.Candidates(context.TODO(), conf)
```
satoqz:
```suggestion
	candidates, err = join.Candidates(t.Context(), conf)
```

helper/kmsplugin/wrapper.go
```diff
@@ -177,18 +200,18 @@ func (w *wrapper) reload(ctx context.Context, canary *client) error {
...
 
-	client, err := w.client.catalog.reloadClient(w.client)
+	client, err := w.client.Reload()
 	if err != nil {
 		return err
 	}
 
 	raw, err := client.Dispense("wrapper")
 	if err != nil {
+		client.Close()
```
satoqz, line 225: good catch

helper/pluginutil/catalog/catalog.go
satoqz:
```suggestion
		panic("pluginutil/catalog: tried to close client more than once")
```

helper/pluginutil/catalog/catalog.go
```diff
@@ -61,7 +69,7 @@ func NewCatalog(logger hclog.Logger, config *server.Config) (*Catalog, error) {
 	plugins := make(map[string]*server.PluginConfig)
 	for _, plugin := range config.Plugins {
 		// Ignore plugins that aren't type KMS.
```
satoqz:
```suggestion
		// Ignore plugins that don't concern this catalog.
```

helper/pluginutil/catalog/catalog.go
```diff
@@ -72,23 +80,35 @@ func NewCatalog(logger hclog.Logger, config *server.Config) (*Catalog, error) {
...
+func (c *Catalog) GetClient(name string) (*Client, bool, error) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
 	return c.getClientLocked(name)
 }
 
-func (c *Catalog) getClientLocked(name string) (*client, bool, error) {
+func (c *Catalog) CloseAll() {
+}
```
satoqz: Do we need this?

helper/pluginutil/catalog/catalog.go
```diff
@@ -72,23 +80,35 @@ func NewCatalog(logger hclog.Logger, config *server.Config) (*Catalog, error) {
...
+
+func (c *Catalog) removeClientLocked(client *Client) {
+	if stored, ok := c.clients[client.name]; ok && stored == client {
+		delete(c.clients, client.name)
+	}
+}
```
satoqz: Is this used somewhere other than `Close()`?

sdk/helper/consts/plugin_types.go
```diff
@@ -15,6 +15,7 @@ var PluginTypes = []PluginType{
 	PluginTypeDatabase,
 	PluginTypeSecrets,
 	PluginTypeKMS,
+	PluginTypeJoin,
```
satoqz, line 18: Can you add a branch for this at https://github.com/openbao/openbao/blob/4976a7b1c09002b1d15e37b26a8f2e7218ab0f42/vault/plugin_catalog.go#L1212-L1213?

sdk/helper/pluginutil/run_config.go
satoqz, line 1: I think we can drop these changes given this is now driven by the catalog.

satoqz (commented):
Here's another pass @kwohlfahrt; nothing major, just small things to clean up.

plugins/join/static/static-join-plugin/main.go
kwohlfahrt: > Thoughts on propagating the logger to the plugin server like I did here?

Sounds good, the more consistent the better IMO.

helper/joinplugin/join.go
kwohlfahrt: Yes, I prefer the shorter, `fooplugin.FooCatalog` is redundant.

helper/pluginutil/catalog/catalog.go
kwohlfahrt: Whoops, that was left over from some early prototyping.

helper/pluginutil/catalog/catalog.go
kwohlfahrt: No, this was left over from the `CloseAll()` prototyping above. Have inlined it.

sdk/joinplugin/join.go
kwohlfahrt: I was thinking about whether we should support plugin multiplexing as well? IIUC currently, if the same plugin type is instantiated twice, only one client process will be launched, so all of these plugin instantiations will share the same state.

sdk/joinplugin/pb/join.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+package join;
+
+option go_package = "github.com/openbao/openbao/sdk/v2/plugin/join/pb";
+
+message Empty {}
+
+message CandidateArgs {
+    map<string, string> config = 1;
```
satoqz: I wonder if we should use https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto here, so we can pass the JSON-serializable subset of `map[string]any` and then have plugins parse that with `mapstructure.WeakDecode`. This gets you much nicer array parsing, for example.

Available via:

```proto
import "google/protobuf/struct.proto";

message CandidateArgs {
    google.protobuf.Struct config = 1;
}
```

sdk/joinplugin/join.go
satoqz: I'm not sure yet what the value in serving multiple `Join` instances at once would be for join plugins. Based on the current design, a single instance of `Join` should be enough to handle several configurations passed via `Candidates()` while the `Join` itself remains largely stateless, no?

sdk/joinplugin/pb/join.proto
satoqz: For an almost out-of-context reference, I had similar plans with https://github.com/Ki-Reply-GmbH/go-kms-wrapping/blob/cf7ab09a9c0d714f3442a404ad394f9896fb686e/kms/kms.go#L120-L127 which is part of the proposed future addition for KMS plugins. Existing plugin interfaces (such as `Wrapper`) take `map[string]string` for configuration, but I found that very cumbersome to work with; it probably contributes to 1000+ lines of parsing boilerplate across all of go-kms-wrapping 😄

sdk/joinplugin/join.go
kwohlfahrt: I also don't have an immediate use-case, if it can be added later in a backwards compatible way (like the `Setup` method), let's do it when there is demand.

sdk/joinplugin/join.go
kwohlfahrt: Looks like no (and also `gRPCServer.Impl`).

plugins/join/static/static-join-plugin/main.go
kwohlfahrt: Wait, what is the intended way to pass the logger through to the actual plugin code? In the linked section above, the logger is just passed to `plugin.ServeConfig`, how does it get to the actual plugin code?

sdk/joinplugin/pb/join.proto
kwohlfahrt: I like this idea, but if you know how to implement it off-hand, I'd appreciate if you could push/PR an example for the static plugin config (especially how the decoding should work).

Otherwise I'll figure it out, but it might take another couple of days, I only have a few hours in the evening to work on this.

plugins/join/static/static-join-plugin/main.go
satoqz: My bad, didn't set this up for KMS plugins yet because there was no API to pass loggers to `Wrapper` anyway.

Backend plugins are the better reference, the logger is propagated like so:

https://github.com/openbao/openbao/blob/ef9fb04a52d744e53853ec4232e882294225ac01/sdk/plugin/serve.go#L33-L68

And later passed to the plugin's factory here:

https://github.com/openbao/openbao/blob/ef9fb04a52d744e53853ec4232e882294225ac01/sdk/plugin/grpc_backend_server.go#L78-L117

...so coming back to join plugins, I suppose we'd just pass it to the one-off factory right away.

Might be smart to pass a

```go
type JoinConfig struct {
  Logger log.Logger
}
```

to the factory for now to ensure future space for anything we may want later on.

sdk/joinplugin/pb/join.proto
satoqz: This is the gist of how you'd parse it in the plugin code:

https://github.com/Ki-Reply-GmbH/go-kms-wrapping/blob/f7fdf585923421cd0cb324baaa899b561f025610/kms/transit/transit.go#L51-L70

And here's how you'd convert between `map[string]any` and the `google.protobuf.Struct` type in Go:

Client: https://github.com/Ki-Reply-GmbH/go-kms-wrapping/blob/5c010f42dec73eec66d3bedd75eb1fe1b403bb8c/plugin/kms_client.go#L49-L60
Server: https://github.com/Ki-Reply-GmbH/go-kms-wrapping/blob/5c010f42dec73eec66d3bedd75eb1fe1b403bb8c/plugin/kms_server.go#L80

If time runs low, just let me know and I'll try to get this over the finish line (though I think we're very close). I've optimistically added this to the v2.6-beta milestone for now, but there's no hard requirement at all to ship until then.

plugins/join/static/static-join-plugin/main.go
kwohlfahrt: OK, I think I've followed the pattern from the backend plugin. For the AWS backend at least, the ability to pass in a logger isn't actually used though:

https://github.com/openbao/openbao-plugins/blob/c88c2889e3e70b9c609887c115421dd7744cd743/secrets/aws/cmd/main.go#L23-L28

satoqz: @kwohlfahrt hmm I don't think it matters much with the backend plugins because there's almost nothing inside of `ServeMultiplex` can actually fail. I would assume though that for the logging call referenced above to forward to the client correctly, it'd need to be a `JSONFormat=true` logger, not a standard logger.

This does make me think that it's probably better to call the join plugin factory not inside `Serve()`, but inside `GRPCServer()`. `Serve()` can then have no return value at all, and any error produced by the factory will be forwarded after already establishing a gRPC connection, with the correct logging settings by go-plugin.

Edit: Logs are forwarded over stdout, but it probably makes sense to actually handshake/connect before we even call the factory and see if that fails.

satoqz: @kwohlfahrt here's a quick patch based on that idea:

```diff
diff --git a/helper/joinplugin/join.go b/helper/joinplugin/join.go
index 6e5074b81f..d92dc685f4 100644
--- a/helper/joinplugin/join.go
+++ b/helper/joinplugin/join.go
@@ -44,7 +44,9 @@ func (c *Catalog) NewJoin(name string) (joinplugin.Join, bool, error) {
 	}
 	if !ok {
 		if factory, ok := builtins[name]; ok {
-			plugin, err := factory(joinplugin.JoinConfig{c.Logger})
+			plugin, err := factory(&joinplugin.JoinConfig{
+				Logger: c.Logger,
+			})
 			return plugin, true, err
 		}
 		return nil, false, fmt.Errorf("unknown join: %s", name)
diff --git a/plugins/join/discover/discover-join-plugin/main.go b/plugins/join/discover/discover-join-plugin/main.go
index 047a44c947..a1cbd9ac7d 100644
--- a/plugins/join/discover/discover-join-plugin/main.go
+++ b/plugins/join/discover/discover-join-plugin/main.go
@@ -1,24 +1,12 @@
 package main
 
 import (
-	"os"
-
-	"github.com/hashicorp/go-hclog"
 	"github.com/openbao/openbao/plugins/join/discover"
 	"github.com/openbao/openbao/sdk/v2/joinplugin"
 )
 
 func main() {
-	logger := hclog.New(&hclog.LoggerOptions{
-		Level:      hclog.Info,
-		JSONFormat: true,
-	})
-
-	if err := joinplugin.Serve(joinplugin.ServeOpts{
+	joinplugin.Serve(joinplugin.ServeOpts{
 		Factory: discover.Factory,
-		Logger:  logger,
-	}); err != nil {
-		logger.Error(err.Error())
-		os.Exit(1)
-	}
+	})
 }
diff --git a/plugins/join/discover/discover.go b/plugins/join/discover/discover.go
index 3fb3bcb23c..44be0650a2 100644
--- a/plugins/join/discover/discover.go
+++ b/plugins/join/discover/discover.go
@@ -12,7 +12,7 @@ import (
 	"github.com/openbao/openbao/sdk/v2/joinplugin"
 )
 
-func Factory(cfg joinplugin.JoinConfig) (joinplugin.Join, error) {
+func Factory(cfg *joinplugin.JoinConfig) (joinplugin.Join, error) {
 	providers := make(map[string]discover.Provider)
 	for k, v := range discover.Providers {
 		providers[k] = v
diff --git a/plugins/join/static/static-join-plugin/main.go b/plugins/join/static/static-join-plugin/main.go
index fe9c84ae2d..a5fac27ab7 100644
--- a/plugins/join/static/static-join-plugin/main.go
+++ b/plugins/join/static/static-join-plugin/main.go
@@ -1,24 +1,12 @@
 package main
 
 import (
-	"os"
-
-	"github.com/hashicorp/go-hclog"
 	"github.com/openbao/openbao/plugins/join/static"
 	"github.com/openbao/openbao/sdk/v2/joinplugin"
 )
 
 func main() {
-	logger := hclog.New(&hclog.LoggerOptions{
-		Level:      hclog.Info,
-		JSONFormat: true,
-	})
-
-	if err := joinplugin.Serve(joinplugin.ServeOpts{
+	joinplugin.Serve(joinplugin.ServeOpts{
 		Factory: static.Factory,
-		Logger:  logger,
-	}); err != nil {
-		logger.Error(err.Error())
-		os.Exit(1)
-	}
+	})
 }
diff --git a/plugins/join/static/static.go b/plugins/join/static/static.go
index d2b47b5ec4..f06a81e7fe 100644
--- a/plugins/join/static/static.go
+++ b/plugins/join/static/static.go
@@ -11,7 +11,7 @@ import (
 	"github.com/openbao/openbao/sdk/v2/joinplugin"
 )
 
-func Factory(cfg joinplugin.JoinConfig) (joinplugin.Join, error) {
+func Factory(cfg *joinplugin.JoinConfig) (joinplugin.Join, error) {
 	return &Static{logger: cfg.Logger}, nil
 }
 
diff --git a/sdk/joinplugin/join.go b/sdk/joinplugin/join.go
index c31e7d7997..8444c9792c 100644
--- a/sdk/joinplugin/join.go
+++ b/sdk/joinplugin/join.go
@@ -23,14 +23,14 @@ type Join interface {
 type joinPlugin struct {
 	plugin.NetRPCUnsupportedPlugin
 
-	impl Join
+	factory func() (Join, error)
 }
 
 type JoinConfig struct {
 	Logger hclog.Logger
 }
 
-type Factory func(cfg JoinConfig) (Join, error)
+type Factory func(*JoinConfig) (Join, error)
 
 type gRPCClient struct {
 	pb.JoinClient
@@ -94,7 +94,11 @@ func (j *joinPlugin) GRPCClient(ctx context.Context, b *plugin.GRPCBroker, c *gr
 }
 
 func (j *joinPlugin) GRPCServer(b *plugin.GRPCBroker, s *grpc.Server) error {
-	pb.RegisterJoinServer(s, &gRPCServer{impl: j.impl})
+	impl, err := j.factory()
+	if err != nil {
+		return err
+	}
+	pb.RegisterJoinServer(s, &gRPCServer{impl: impl})
 	return nil
 }
 
diff --git a/sdk/joinplugin/serve.go b/sdk/joinplugin/serve.go
index 0265c96597..f508587c91 100644
--- a/sdk/joinplugin/serve.go
+++ b/sdk/joinplugin/serve.go
@@ -10,7 +10,7 @@ type ServeOpts struct {
 	Logger  hclog.Logger
 }
 
-func Serve(opts ServeOpts) error {
+func Serve(opts ServeOpts) {
 	logger := opts.Logger
 	if logger == nil {
 		logger = hclog.New(&hclog.LoggerOptions{
@@ -19,19 +19,18 @@ func Serve(opts ServeOpts) error {
 		})
 	}
 
-	impl, err := opts.Factory(JoinConfig{logger})
-	if err != nil {
-		// Don't log returned error, caller will handle it,
-		return err
+	factory := func() (Join, error) {
+		return opts.Factory(&JoinConfig{
+			Logger: logger,
+		})
 	}
 
 	plugin.Serve(&plugin.ServeConfig{
 		HandshakeConfig: HandshakeConfig,
 		VersionedPlugins: map[int]plugin.PluginSet{
-			1: {"join": &joinPlugin{impl: impl}},
+			1: {"join": &joinPlugin{factory: factory}},
 		},
 		GRPCServer: plugin.DefaultGRPCServer,
 		Logger:     logger,
 	})
-	return nil
 }
```

sdk/joinplugin/serve.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-plugin"
+)
+
+type ServeOpts struct {
+	Factory Factory
+	Logger  hclog.Logger
+}
+
+func Serve(opts ServeOpts) error {
```
satoqz:
```suggestion
func Serve(opts *ServeOpts) error {
```

Let's make this take a reference just out of consistency with other plugins.

kwohlfahrt:
@satoqz OK, I think that's everything from the review - could you take another look please?

plugins/join/static/static.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		host := url.Hostname()
+		portStr := url.Port()
+		if portStr == "" {
+			portStr = "8200"
+		}
+		port, err := strconv.ParseUint(portStr, 10, 16)
+		if err != nil {
+			d.logger.Warn("failed to parse port", portStr)
+		}
```
satoqz, line 45: Should the error branches return the error rather than logging?

plugins/join/static/static.go
```diff
@@ -0,0 +1,54 @@
...
+func Factory(cfg *joinplugin.JoinConfig) (joinplugin.Join, error) {
+	return &Static{logger: cfg.Logger}, nil
+}
+
+type Static struct {
+	logger hclog.Logger
+}
+
+type staticConfig struct {
+	Addresses []string
```
satoqz:
```suggestion
	Addresses []string `mapstructure:"addresses"`
```

I think it's good to be explicit about the expected name (in the discover plugin, too).

physical/raft/raft_test.go
```diff
@@ -283,6 +283,45 @@ func TestRaft_ParseNonVoter(t *testing.T) {
...
+		logger: hclog.NewNullLogger(),
+		conf: map[string]string{
+			"retry_join": `[
+				{"auto_join": "provider=aws foo=baz"},
+				{"auto_join_plugin": {
+					"plugin": "discover",
+					"config": {
+					  "provider": "aws",
+					  "args": {"foo": "bar"}
+		            }
```
satoqz:
```suggestion
					"config": {
					  "provider": "aws",
					  "args": {"foo": "bar"}
		      }
```

vault/raft.go
```diff
@@ -1034,16 +1028,44 @@ func (c *Core) JoinRaftCluster(ctx context.Context, leaderInfos []*raft.LeaderJo
...
+		joinPlugins[name] = plugin
+	}
+
+	if retryFailures {
 		go func() {
+			defer func() {
+				for name, plugin := range joinPlugins {
+					err := plugin.Cleanup(ctx)
+					if err != nil {
+						c.logger.Error("error cleaning up join plugin %s: %w", name, err)
```
satoqz:
```suggestion
						c.logger.Error("error cleaning up join plugin", "name", name, "error", err.Error())
```

vault/raft.go
```diff
@@ -1054,8 +1076,16 @@ func (c *Core) JoinRaftCluster(ctx context.Context, leaderInfos []*raft.LeaderJo
...
 		// Backgrounded so return false
 		return false, nil
-	default:
-		if err := join(); err != nil {
+	} else {
+		defer func() {
+			for name, plugin := range joinPlugins {
+				err := plugin.Cleanup(ctx)
+				if err != nil {
+					c.logger.Error("error cleaning up join plugin %s: %w", name, err)
```
satoqz:
```suggestion
					c.logger.Error("error cleaning up join plugin", "name", name, "error", err.Error())
```

changelog/1556.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+storage/raft: auto-join is now a pluggable feature, allowing support for new environments with custom plugins.
```
satoqz:
```suggestion
Add **Auto Join plugins** for Raft to support discovery in arbitrary environments via custom external plugins. 
```

`feature` type release notes take a slightly different format.

helper/pluginutil/catalog/catalog.go
```diff
@@ -29,11 +28,14 @@ import (
 // created. Additionally, we get to simplify many things as KMS plugins are
 // declarative only.
```
satoqz:
```suggestion
// Catalog manages long-lived plugin server processes and clients of a
// particular plugin type. It is disjoint from the plugin catalog in core
// and services plugins that must be available before core starts up and are
// configurable via the server configuration only.
```

helper/pluginutil/catalog/catalog.go
```diff
@@ -42,7 +44,13 @@ type Catalog struct {
 }
 
 // NewCatalog returns a new KMS plugin catalog.
```
satoqz:
```suggestion
// NewCatalog returns a new Catalog.
```

physical/raft/raft.go
```diff
@@ -231,6 +237,12 @@ type LeaderJoinInfo struct {
 	// auto-join.
 	AutoJoinPort uint `json:"auto_join_port"`
 
+	// AutoJoinScript defines an executable and arguments that outputs peer API
+	// addresses (e.g. `https://vault.example.com:8200`) to standard output. If
+	// supplied, Vault will attempt to automatically discover peers in addition
+	// to what can be provided via 'leader_api_addr'.
```
satoqz: This seems outdated? FWIW we also don't ever refer to OpenBao as Vault.

satoqz (commented):
Final suggestions. @openbao/openbao-org-maintainers do you want to have a look? I think this is in fairly good shape now.

physical/raft/raft.go
kwohlfahrt: Yes it is out of date. Sorry about the stray reference, I'll go fix the comment I copied this from too then 😉  https://github.com/openbao/openbao/blob/7d2db749fdc3b4b7df2ef55d385a26a9202c8aab/physical/raft/raft.go#L221-L223

plugins/join/static/static.go
kwohlfahrt: Hm, they should at least `continue`. Depends whether we want to consider the remaining addresses if one of them is invalid. Just returning the error is simpler, and I don't think anyone will complain about an invalid config not working.

satoqz: Yeah I'd just return the error on the spot.

ragnar-howler (commented):

This PR has been scanned for:
- ✅ Common security issues
- ✅ Code style consistency  
- ✅ Test coverage indicators

**Stats:** +1316 -123 in 26 files

helper/kmsplugin/wrapper_test.go
```diff
@@ -11,11 +11,53 @@ import (
...
+	"github.com/openbao/openbao/helper/osutil"
+	"github.com/openbao/openbao/sdk/v2/helper/consts"
 	"github.com/stretchr/testify/require"
 )
 
+const TestPluginServerEnv = "BAO_TEST_PLUGIN_SERVER"
+
+// TestPluginServer is not an actual test but a hack to reuse the test binary
+// as a plugin binary that is called into by the same test binary from the main
+// test runner process.
```
cipherboy, line 27: 🤯

physical/raft/raft.go
```diff
@@ -270,6 +281,19 @@ type LeaderJoinInfo struct {
 	TLSConfig *tls.Config `json:"-"`
 }
 
+func (info *LeaderJoinInfo) ValidateJoinMethods() error {
+	// TODO: At config loading, convert all join methods (legacy discover and
```
cipherboy, line 285: Since we're making a lot of pluginization related changes in v2.7.0, let's add a deprecation warning about `discover` and move it to `openbao-plugins` by then @satoqz @kwohlfahrt.

cipherboy (dismissed):
Code is looking good, though I don't run Raft enough to test live.

Will let @satoqz do the final review, but thank you @kwohlfahrt and appreciate your patience! This is looking rather excellent.

physical/raft/raft.go
satoqz: That's what I had in mind, yeah.

satoqz: I've marked this as follow-up work here, together with general documentation: https://github.com/openbao/openbao/issues/2956.

vault/raft.go
```diff
@@ -1034,16 +1028,44 @@ func (c *Core) JoinRaftCluster(ctx context.Context, leaderInfos []*raft.LeaderJo
...
+
+	if retryFailures {
 		go func() {
+			defer func() {
+				for name, plugin := range joinPlugins {
+					err := plugin.Cleanup(ctx)
+					if err != nil {
+						c.logger.Error("error cleaning up join plugin", "name", name, "error", err.Error())
+					}
+				}
```
satoqz: This also needs to run if we fail at `NewJoin` above, and already spawned plugins just in case several are configured. Probably best to pull it out as a `cleanupPlugins` closure and call in all 3 places.

satoqz:
I wanted to give this a final test run but I'm having trouble defining the config as HCL.

I'm using the following:

```hcl
  retry_join {
    auto_join_plugin = {
      plugin = "static"
      config = {
        addresses = ["http://127.0.0.1:8200"]
      }
    }
  }
```

where the contents of the `retry_join` stanza are parsed in `command/server/config.go` and converted to the following JSON, which is then passed to Raft and deserialized back from JSON:

```json
[
  {
    "auto_join_plugin": [
      {
        "config": [
          {
            "addresses": [
              "http://127.0.0.1:8200"
            ]
          }
        ],
        "plugin": "static"
      }
    ]
  }
]
```

Note how we expect the outer array (this is `[]*LeaderJoinInfo`), but don't expect `auto_join_plugin` to be an array, neither do we expect `config` to be an array.

We can adapt code to expect these, but that's really ugly, especially since this expands to any potentially plugin-defined nested key.

Not immediately sure how to work around this, thoughts?

vault/raft.go
satoqz: Similarly, nothing waits for this Goroutine to eventually complete so we're not sure to actually run the defer.

satoqz: Yeah, this Goroutine always leaks if `join()` never succeeds. `ctx` always being `context.Background()` doesn't help either...

satoqz: This also led me to discover that `/sys/storage/raft/join` allows passing `auto_join` and `retry`, should we support passing `auto_join_plugin` there? See `http/sys_raft.go`.

satoqz:
To illustrate, attached is an ugly patch that resolves most of the described HCL problem, but not very nicely. This would result in:
- Allowing several `auto_join_plugin` stanzas per `retry_join` stanza, which seems reasonable.
- Allowing only a single `config` block.
- When join plugins define nested config objects (likely rare, but supported), they must perform the same de-duplication down to a single object if those are the wanted semantics.

<details>

<summary>
Expand patch
</summary>

```diff
diff --git a/physical/raft/raft.go b/physical/raft/raft.go
index 71879061ed..60fcd26f61 100644
--- a/physical/raft/raft.go
+++ b/physical/raft/raft.go
@@ -217,8 +217,9 @@ func (r *RaftBackend) HookInvalidate(hook physical.InvalidateFunc) {
 }
 
 type AutoJoinPlugin struct {
-	Plugin string         `json:"plugin"`
-	Config map[string]any `json:"config"`
+	Plugin    string           `json:"plugin"`
+	ConfigRaw []map[string]any `json:"config"`
+	Config    map[string]any   `json:"-"`
 }
 
 // LeaderJoinInfo contains information required by a node to join itself as a
@@ -237,10 +238,10 @@ type LeaderJoinInfo struct {
 	// auto-join.
 	AutoJoinPort uint `json:"auto_join_port"`
 
-	// AutoJoinPlugin specifies the name and configuration of an existing join
-	// plugin. If supplied, OpenBao will attempt to automatically discover peers
-	// in addition to what can be provided via 'leader_api_addr'.
-	AutoJoinPlugin *AutoJoinPlugin `json:"auto_join_plugin"`
+	// AutoJoinPlugins specify names and configurations of existing join
+	// plugins. If supplied, OpenBao will attempt to automatically discover
+	// peers in addition to what can be provided via 'leader_api_addr'.
+	AutoJoinPlugins []*AutoJoinPlugin `json:"auto_join_plugin"`
 
 	// LeaderAPIAddr is the address of the leader node to connect to
 	LeaderAPIAddr string `json:"leader_api_addr"`
@@ -285,7 +286,7 @@ func (info *LeaderJoinInfo) ValidateJoinMethods() error {
 	// TODO: At config loading, convert all join methods (legacy discover and
 	// static addresses) to plugins. Then, this can be removed.
 	haveJoinMethod := false
-	for _, joinMethodEnabled := range []bool{len(info.AutoJoin) != 0, len(info.LeaderAPIAddr) != 0, info.AutoJoinPlugin != nil} {
+	for _, joinMethodEnabled := range []bool{len(info.AutoJoin) != 0, len(info.LeaderAPIAddr) != 0, len(info.AutoJoinPlugins) != 0} {
 		if haveJoinMethod && joinMethodEnabled {
 			return errors.New("only one of leader_api_addr, auto_join, and auto_join_script may be enabled")
 		}
@@ -325,10 +326,10 @@ func (b *RaftBackend) JoinConfig() ([]*LeaderJoinInfo, error) {
 			}
 			provider := config["provider"]
 			delete(config, "provider")
-			info.AutoJoinPlugin = &AutoJoinPlugin{
+			info.AutoJoinPlugins = []*AutoJoinPlugin{{
 				Plugin: "discover",
 				Config: map[string]any{"provider": provider, "args": map[string]string(config)},
-			}
+			}}
 			info.AutoJoin = ""
 		}
 
@@ -336,6 +337,20 @@ func (b *RaftBackend) JoinConfig() ([]*LeaderJoinInfo, error) {
 			return nil, fmt.Errorf("invalid scheme %q; must either be http or https", info.AutoJoinScheme)
 		}
 
+		for _, plugin := range info.AutoJoinPlugins {
+			// Lower the raw config array into a single config. We always
+			// receive an array of configs from the server config because it is
+			// interpreted as an HCL block, even in JSON mode.
+			switch len(plugin.ConfigRaw) {
+			case 0:
+			case 1:
+				plugin.Config = plugin.ConfigRaw[0]
+			default:
+				return nil, fmt.Errorf("invalid auto_join_plugin config: expected a single 'config' block, got %d", len(plugin.ConfigRaw))
+			}
+			plugin.ConfigRaw = nil
+		}
+
 		info.Retry = true
 		info.TLSConfig, err = b.parseTLSInfo(info)
 		if err != nil {
diff --git a/physical/raft/raft_test.go b/physical/raft/raft_test.go
index 2a01c37b57..7a2f3dbd74 100644
--- a/physical/raft/raft_test.go
+++ b/physical/raft/raft_test.go
@@ -289,13 +289,13 @@ func TestRaft_JoinConfig(t *testing.T) {
 		conf: map[string]string{
 			"retry_join": `[
 				{"auto_join": "provider=aws foo=baz"},
-				{"auto_join_plugin": {
+				{"auto_join_plugin": [{
 					"plugin": "discover",
-					"config": {
+					"config": [{
 						"provider": "aws",
 						"args": {"foo": "bar"}
-					}
-				}}
+					}]
+				}]}
 			]`,
 		},
 	}
@@ -305,17 +305,17 @@ func TestRaft_JoinConfig(t *testing.T) {
 	}
 	expected := []*LeaderJoinInfo{
 		{
-			AutoJoinPlugin: &AutoJoinPlugin{
+			AutoJoinPlugins: []*AutoJoinPlugin{{
 				Plugin: "discover",
 				Config: map[string]any{"provider": "aws", "args": map[string]string{"foo": "baz"}},
-			},
+			}},
 			Retry: true,
 		},
 		{
-			AutoJoinPlugin: &AutoJoinPlugin{
+			AutoJoinPlugins: []*AutoJoinPlugin{{
 				Plugin: "discover",
 				Config: map[string]any{"provider": "aws", "args": map[string]any{"foo": "bar"}},
-			},
+			}},
 			Retry: true,
 		},
 	}
diff --git a/vault/external_tests/raft/raft_test.go b/vault/external_tests/raft/raft_test.go
index d46a05e268..d9e26240f2 100644
--- a/vault/external_tests/raft/raft_test.go
+++ b/vault/external_tests/raft/raft_test.go
@@ -220,12 +220,12 @@ func TestRaft_RetryAutoJoin_Plugin(t *testing.T) {
 
 	leaderInfos := []*raft.LeaderJoinInfo{
 		{
-			AutoJoinPlugin: &raft.AutoJoinPlugin{
+			AutoJoinPlugins: []*raft.AutoJoinPlugin{{
 				Plugin: "static",
 				Config: map[string]any{
 					"addresses": []string{leaderCore.Client.Address()},
 				},
-			},
+			}},
 			TLSConfig: leaderCore.TLSConfig(),
 			Retry:     true,
 		},
diff --git a/vault/raft.go b/vault/raft.go
index f69b752e9d..12a03002e4 100644
--- a/vault/raft.go
+++ b/vault/raft.go
@@ -1035,18 +1035,16 @@ func (c *Core) JoinRaftCluster(ctx context.Context, leaderInfos []*raft.LeaderJo
 
 	joinPlugins := make(map[string]joinsdk.Join, 0)
 	for _, info := range leaderInfos {
-		if info.AutoJoinPlugin == nil {
-			continue
-		}
-		name := info.AutoJoinPlugin.Plugin
-		if _, ok := joinPlugins[name]; ok {
-			continue
-		}
-		plugin, _, err := catalog.NewJoin(name)
-		if err != nil {
-			return false, fmt.Errorf("failed to create join plugin: %w", err)
+		for _, plugin := range info.AutoJoinPlugins {
+			if _, ok := joinPlugins[plugin.Plugin]; ok {
+				continue
+			}
+			join, _, err := catalog.NewJoin(plugin.Plugin)
+			if err != nil {
+				return false, fmt.Errorf("failed to create join plugin: %w", err)
+			}
+			joinPlugins[plugin.Plugin] = join
 		}
-		joinPlugins[name] = plugin
 	}
 
 	if retryFailures {
@@ -1104,17 +1102,19 @@ func (c *Core) raftLeaderInfo(ctx context.Context, joinPlugins map[string]joinsd
 	switch {
 	case leaderInfo.LeaderAPIAddr != "":
 		ret = append(ret, leaderInfo)
-	case leaderInfo.AutoJoinPlugin != nil:
-		join := joinPlugins[leaderInfo.AutoJoinPlugin.Plugin]
-		addrs, err := join.Candidates(ctx, leaderInfo.AutoJoinPlugin.Config)
-		if err != nil {
-			return nil, fmt.Errorf("failed to run join plugin: %w", err)
-		}
+	case len(leaderInfo.AutoJoinPlugins) != 0:
+		for _, plugin := range leaderInfo.AutoJoinPlugins {
+			join := joinPlugins[plugin.Plugin]
+			addrs, err := join.Candidates(ctx, plugin.Config)
+			if err != nil {
+				return nil, fmt.Errorf("failed to run join plugin: %w", err)
+			}
 
-		for _, addr := range addrs {
-			info := *leaderInfo
-			info.LeaderAPIAddr = fmt.Sprintf("%s://%s:%d", addr.Scheme, addr.Host, addr.Port)
-			ret = append(ret, &info)
+			for _, addr := range addrs {
+				info := *leaderInfo
+				info.LeaderAPIAddr = fmt.Sprintf("%s://%s:%d", addr.Scheme, addr.Host, addr.Port)
+				ret = append(ret, &info)
+			}
 		}
 	default:
 		return nil, errors.New("must provide leader address or auto-join metadata")
```

</details>

---

@cipherboy notably, we can't apply the profile system hack we talked about so easily since the storage plugin itself may want to choose between both behaviors. For example, Raft expects multiple `retry_join` stanzas, so we can't eliminate `[]map[string]any` _before_ passing the config to the storage backend, neither can we eliminate it _after_ it is passed to the storage backend because it will see `[]any` instead after roundtripping from JSON.

vault/raft.go
kwohlfahrt: > This also needs to run if we fail at NewJoin above, and already spawned plugins just in case several are configured. Probably best to pull it out as a cleanupPlugins closure and call in all 3 places.

Done.

> Yeah, this Goroutine always leaks if join() never succeeds. ctx always being context.Background() doesn't help either...

Is this not expected? If the join hasn't succeeded, the aim is to retry forever, no? And as long as we're retrying, we need to keep the plugins alive.

> This also led me to discover that /sys/storage/raft/join allows passing auto_join and retry, should we support passing auto_join_plugin there? See http/sys_raft.go.

Probably? Is that endpoint intended to be the API version of the `retry_join` config? So you `POST` it on a new member instead of specifying `retry_join` in the config file?

There aren't any tests for this endpoint as far as I can see, so I'd have to put those together from scratch.

satoqz: > Is this not expected? If the join hasn't succeeded, the aim is to retry forever, no? And as long as we're retrying, we need to keep the plugins alive.

By leak I meant to say it does not synchronously shut down before `main()` exits, and as a result we won't run `defer`s, which means we leak the plugin process (which go-plugin doesn't attach to the parent fsr). To give an example: https://go.dev/play/p/uEZ6UlynXl0.

> Probably? Is that endpoint intended to be the API version of the retry_join config? So you POST it on a new member instead of specifying retry_join in the config file?

I assume so too (guesswork for me as well). But let's defer this to a follow up since it's a niche detail we can add separately.

kwohlfahrt: Ah I see. Interestingly the playground does run them all about 90% of the time for me, but sometimes it does leak.

Is it a good idea to put a `ctx context.Context` on the core, and then do a `defer ctx.Cancel()` in the main `Run()` method? It should probably be used in more places, lots of `context.Background()` everywhere.

cipherboy: @kwohlfahrt I have a patch locally that makes active context become atomic, but I think that's more narrow in scope than this one. Context on core could be an interesting idea, though I think I'd need to understand how it interacts with the command line and that weird `-exit-on-core-shutdown` flag. Either way, should probably be bound to core lifetime. 

I think that'd be good.

kwohlfahrt: ~Actually there is already `core.GetContext()`, which looks like something that can be used here. Let me give that a spin.~

```go
	// This can be used to trigger operations to stop running when Vault is
	// going to be shut down, stepped down, or sealed
	activeContext           context.Context
```

EDIT: Actually nevermind, probably not. We need to join while the core is sealed, so it's not a great fit. I'll just use a new context for the main `Run()` function.

cipherboy: @kwohlfahrt Sadly that's tied to the lifetime of the active/read-enabled standby state, not the entire server process lifetime. It isn't created early enough I don't think either.

kwohlfahrt:
Hi folks, sorry for the radio silence, I haven't quite forgotten about this, just been too busy to devote the necessary time to it. Just wanted to outline my plans for this PR - I think the remaining issue is the config parsing, and the cleanest way to solve it would be to switch to HCL v2 parsing, which doesn't wrap everything in lists.

We could do this for just the join stanzas, or _maybe_ switch over the entire config parser (agent already uses v2 parsing) - I'll test both, and if the latter turns out to be feasible open a separate PR for that, since it probably shouldn't be bundled with this change.

PR #1561, opened 2025-07-18
Add QR option to the OIDC device flow login

### LINKED ISSUE #1558: Add QR option to the OIDC device flow login
The OIDC device flow is a great authentication method to use when the normal browser flow isn't usable, e.g., inside an SSH session, but copying the link to an external browser can be a bit annoying. It would be nice to add a QR code option to the client, which could improve the user experience in such cases.

It would be nice to add a "generateqr=yes" option to the callbackmode=device to generate a QR that can be scanned by the user's mobile device.

It could be done externally with scripts like this:
```
#!/bin/bash
./bao login -method=oidc callbackmode=device skip_browser=true 2>&1 | while read -r line; do
    # Output the current line (this will allow the process to still display the login prompts)
    echo "$line"

    # Check if the line contains the URL (regex for URLs)
    if [[ "$line" =~ https?://.+ ]]; then
        # Extract the URL
        auth_url="${BASH_REMATCH[0]}"

        # Output the URL
        echo "Authorization URL: $auth_url"

        # Generate the QR code for the URL
        echo "Generating QR code for the URL..."
        qrencode -t UTF8 "$auth_url"  # Displays QR code directly in terminal

        # After generating the QR code, stop further URL extraction (we only need the first URL)
    fi
done
```
But it would be better to implement in the bao binary

phil9909:
How it looks:

openbao-feat-login-qr

This adds a new dependency to https://github.com/yeqown/go-qrcode/, which itself has one transitive dependency (from the same author, the error correction algorithm for QR codes in a separate repo)

How it works: as terminal fonts tend to be about twice as high as wide it uses UTF-8 "[Block Elements](https://en.wikipedia.org/wiki/Block_Elements)":
- Upper half block `▀`
- Lower half block `▄`
- Full block `█`

Therefore we display two pixels per character (and two row per line)

Also [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit) are used to make sure the chars are black and background is white. Terminals which don't support this should just ignore it.

Alternatives:
- We could display the QR code with terminal image protocols (like SIXEL), but this is supported by fewer terminals (compared to UTF-8).
- We could make the browser command customizable and users would be free to set their "browser" to `qrencode -t UTF8` or even a script, which automates the flow, etc.

Resolves #1558


builtin/credential/jwt/cli.go
```diff
@@ -415,13 +426,16 @@ Configuration:
   %s=<bool>
     Toggle the automatic launching of the default browser to the login URL. (default: false).
 
+  %s=<bool>
+    Display a QR code to the login URL. Require UTF-8 support (default: false).
```
satoqz:
```suggestion
    Display a QR code to the login URL. Requires UTF-8 support from your terminal emulator. (default: false).
```

builtin/credential/jwt/cli_qr.go
```diff
@@ -0,0 +1,75 @@
...
+			return false
+		}
+		return b[row][col]
+	}
+
+	builder := strings.Builder{}
+
+	for row := -2; row < mat.Height()+2; row += 2 {
+		builder.WriteString("\033[38;2;255;255;255m") // set forground color to white
+		builder.WriteString("\033[48;2;0;0;0m")       // set background color to black
```
satoqz: I'd suggest using 3 bit colors over 24 bit colors for better terminal emulator compatibility.

```suggestion
		builder.WriteString("\033[37m") // set forground color to white
		builder.WriteString("\033[40m") // set background color to black
```

You could also conditionally do this by checking for support via https://github.com//termstandard/colors?tab=readme-ov-file#truecolor-detection, but overkill.

satoqz (commented):
Pretty cool, just some drive by nits.

driif:
Looks cool :+1:

changelog/1561.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+auth/oidc: Add new `show_qr=true` option to display a QR code of the login URL.
```
DrDaveD: Change `option` to `cli option`

builtin/credential/jwt/cli_qr.go
phil9909: The problem with this predefined colors is, that they are usually remapped to some different color (e.g. to match the operating system theme).

You can try by running the following:

```bash
echo -e "\033[40m\033[37m3-bit colors: This should be white on black, but is it?\n\033[38;2;255;255;255m\033[48;2;0;0;0m24-bit colors: This is actually white on black\033[0m"
```

On my machine, this looks like this:

image

phil9909:
> Please add the option to the [OIDC login (CLI) doc section](https://github.com/openbao/openbao/blob/main/website/content/docs/auth/jwt/index.mdx#oidc-login-cli).

Done

> I'm not sure what we should put in the header of new files. Maybe @cipherboy could comment on that.

Sorry, I forgot the headers. Fixed.

builtin/credential/jwt/cli_qr.go
satoqz: I would assume that the contrast is still high enough in the average case for the scanner to figure this out. But maybe just do the `$COLORTERM` check then.

DrDaveD (approved):

changelog/1561.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
cipherboy: This should just be an improvement not a full feature.

builtin/credential/jwt/cli_qr.go
DrDaveD: Yes.  The 24-bit color looks right in iTerm2 on a Mac
Screenshot 2025-07-21 at 10 26 21 AM
but not in the standard Terminal app
Screenshot 2025-07-21 at 10 27 33 AM
which doesn't support 24 bits.

DrDaveD (changes requested):
Need to address the issue of terminal emulators that don't support 24-bit colors.

builtin/credential/jwt/cli_qr.go
phil9909: The problem I see is that this is a solution for situations where automatically opening the browser is not available. For example (as mentioned in #1558) over ssh. Now `COLORTERM` is usually not forwarded:

> Having an extra environment variable (separate from `TERM`) is not ideal: by default it is not forwarded via sudo, ssh, etc, and so it may still be unreliable even where support is available in programs. 
> https://github.com//termstandard/colors?tab=readme-ov-file#truecolor-detection

I'm not strictly against it (after all. there are solutions to forward this env), but I'd prefer a solution that works out of the box for most people (and most people will be using a modern terminal emulator I guess).
How do you @satoqz @DrDaveD feel about this: make `show_qr` a string option and if it is set to `true`, `truecolor` or `24bit` we use 24-bit colors, if is set to `3bit` we use 3-bit colors, if it is empty or `false` we do not generate a QR code.

satoqz: > and most people will be using a modern terminal emulator I guess

You'd hope so, but no, unfortunately, not the case at all. For example, the MacOS default terminal and PuTTY are likey used by millions of engineers that don't know how to choose better, so this would be broken out of the box. Users of those terminals likely also wouldn't understand that they'd have to pass `3bit` explicitly.

> Now COLORTERM is usually not forwarded

Yes, it's imperfect, but works very well in the average case. Falling back to 3bit even if 24bit may be supported will also work in the average case, as most themes should have sufficient contrast (QR codes do not need to be white on black, try some lower constrast combinations on https://www.qrcode-monkey.com/, you'll be surprised how well your phone is able to detect them). The worse thing to do would be to fallback to 24bit, which is essentially undefined behavior in a terminal emulator that doesn't support it.

phil9909: Ok, the PR is now changed. Truecolor ANSI will only be used if `COLORTERM` is set.

changelog/1561.txt
phil9909: Changed

builtin/credential/jwt/cli_qr.go
satoqz: Thanks! I think you would check explicitly for `COLORTERM=truecolor`, not sure. Maybe setting no colors at all in non-truecolor mode is even better? That way you're guaranteed the contrast of the general terminal fore-/background, the explicit black/white colors may indeed have a bit less contrast typically.

But, not going to drag this out as it's really not that important, do as you see fit.

satoqz (approved):
Thanks @phil9909!

builtin/credential/jwt/cli.go
```diff
@@ -377,7 +388,7 @@ Usage: bao login -method=oidc [CONFIG K=V...]
 
           https://accounts.google.com/o/oauth2/v2/...
 
-  The default browser will be opened for the user to complete the login. 
+  The default browser will be opened for the user to complete the login.
```
cipherboy: @phil9909 This space is structural and needs to be retained.

builtin/credential/jwt/cli_qr.go
```diff
@@ -0,0 +1,87 @@
...
+		return b[row][col]
+	}
+
+	builder := strings.Builder{}
+
+	clearColors := "\033[0m"
+	setColor := "\033[37m" + // set forground color to white
+		"\033[40m" // set background color to black
+
+	if _, ok := os.LookupEnv("COLORTERM"); ok {
```
cipherboy:
```suggestion
	if value, ok := os.LookupEnv("COLORTERM"); ok && (value == "truecolor" || value == "24bit") {
```

It looks like it can take on other values depending on which piece of internet advice is accepted at any given moment:

https://github.com/Textualize/rich/blob/9c9b011187bba772dca57653c9114005b489f005/rich/console.py#L811-L812

phil9909: My idea was: If someone comes along with let's say HDR support for the terminal and indicates this with `COLORTERM="hdr"` or whatever, then I would expect this to also support the truecolor standard.

I don't really expect this env to be used for anything else.

Anyway, if we copy the python implementation, I would fully copy the logic (maybe they have good reasons trim and lowercase): 

```suggestion
	colorterm := strings.TrimSpace(strings.ToLower(os.Getenv("COLORTERM")))
	if colorterm == "truecolor" || colorterm == "24bit" {
```

BTW In this case, we don't need the `ok` check because non existing env will evaluate to empty.

DrDaveD: I agree with checking for truecolor & 24bit after lower casing like that.

cipherboy: @phil9909 Copying fully works with me. If we get HDR support, I'll only want proper 12-bit Dolby Vision next. ;-)

builtin/credential/jwt/cli_qr.go
cipherboy: @satoqz this ended up being forked into https://github.com/openbao/openbao/pull/1561#discussion_r2222437906 fwiw, with a source for this in another proper terminal library :-)

cipherboy (approved):

PR #1564, opened 2025-07-22
Implement cache invalidation for plugins

phil9909:
Implement cache invalidation for plugins (assuming plugins already properly implement the `InvalidateKey` method of the  `sdk/logical.Backend` interface).

See #1550
See #1528 


vault/core_cache_invalidate.go
```diff
@@ -68,5 +68,7 @@ func (c *Core) invalidateInternal(ctx context.Context, key string) error {
 		c.logger.Warn("no idea how to invalidate cache. Maybe it's not cached and this is fine, maybe not", "key", key)
 	}
 
+	c.router.Invalidate(ctx, key)
```
cipherboy: Won't we get the warning above here? Perhaps we should return a boolean as to whether we found a matching entry and add it to the switch above?

phil9909: good idea, will do that

PR #1566, opened 2025-07-22
Implement cache invalidation for physical.Cache

phil9909:
Implement cache invalidation for `physical.Cache` and call it from `core.Invalidate`

See #1550
See #1528 

### DIFF
--- a/sdk/physical/cache.go
+++ b/sdk/physical/cache.go
@@ -412,3 +412,13 @@ func (c *cacheTransaction) Rollback(ctx context.Context) error {
 
 	return nil
 }
+
+// Invalidate removes the value for key from the cache.
+// This will not affect transactions that have already been started.
+func (c *cache) Invalidate(ctx context.Context, key string) {
+	lock := locksutil.LockForKey(c.locks, key)
+	lock.Lock()
+	defer lock.Unlock()
+
+	c.lru.Remove(key)
+}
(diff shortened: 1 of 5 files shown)


sdk/physical/cache.go
```diff
@@ -412,3 +412,11 @@ func (c *cacheTransaction) Rollback(ctx context.Context) error {
 
 	return nil
 }
+
+func (c *cache) Invalidate(ctx context.Context, key string) {
```
cipherboy, line 418: Do you mind just adding a note saying that it is expected not to invalidate keys in the transactions here? The cache for a transaction is empty at the time the transaction started and any in-flight transaction would not see any new storage updates so would not need to be invalidated either way.

cipherboy (approved):
One nit, but makes sense otherwise.

sdk/physical/cache.go
phil9909: Done.

Would it make sense to explicitly test for this behavior? i.e. start a transaction, read, modify/delete value, read again => validate it's unchanged from first read.

cipherboy: I think it wouldn't hurt.

phil9909: done

cipherboy:
@phil9909 I have also rebased the `read-replication` branch against `main`.

Issue #1555, opened 2025-07-16
Add transit secret engine wrappers to openbao Go client


olegdayo:
**Is your feature request related to a problem? Please describe.**

Wanted to create a key in transit secret engine via Go client

cURL request:
```sh
curl --header "X-Vault-Token: $VAULT_TOKEN" \
   --request POST \
   $VAULT_ADDR/v1/transit/keys/autounseal
```

But for now, only KV is supported

I tried using `client.Logical().WriteWithContext()`, but it only supports PUT verb

**Describe the solution you'd like**

Similar to client.KVv2, but for transit: `client.Transit().CreateKey()`, `client.Transit().ImportKey()`, etc...

**Describe alternatives you've considered**

Right now, I am using the following solution with a deprecated method:

```go
r := client.NewRequest(http.MethodPost, "/v1/transit/keys/autounseal")
resp, err = client.RawRequestWithContext(ctx, r)
```

cipherboy:
@olegdayo If you (or anyone else!) wants to implement this, happy to take a PR!

I will note that `PUT` verb should work just fine with Transit.

Nivedita-coder:
Hi @cipherboy,
I would like to contribute on this issue. if @olegdayo is not working feel free to assign me!:)

olegdayo:
@Nivedita-coder 
I started working a little on the issue, but totally ok if you want to take it
Let me share a list of methods that would be cool to see:
- CreateKey (`bao write -output-curl-string -f transit/keys/my-key`)
- RotateKey (`bao write -output-curl-string -f transit/keys/my-key/rotate`)
- RewrapKey (`bao write -output-curl-string transit/rewrap/my-key ciphertext=...`)
- ExportKey (`bao read -output-curl-string transit/export/public-key/my-key`)
- Encrypt (`bao write -output-curl-string transit/encrypt/my-key plaintext=...`)
- Decrypt (`bao write -output-curl-string transit/decrypt/my-key ciphertext=...`)
- GetWrappingKey (`bao read -output-curl-string transit/wrapping_key`)

cipherboy:
@Nivedita-coder @olegdayo Let me know if you want a feature branch and I can create one. Then you can both provide PRs and we can merge it when it is complete :-)

olegdayo:
> Let me know if you want a feature branch and I can create one

@cipherboy yeah, it would come in handy, thank you

cipherboy:
[...]

PR #1567, opened 2025-07-22
issue-1555 methods init

olegdayo:
Resolves #

### DIFF
--- a/api/transit.go
+++ b/api/transit.go
@@ -1,0 +1,147 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package api
+
+import (
+	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"net/http"
+)
+
+const (
+	keyPath       = "%s/keys/%s"
+	rotateKeyPath = "%s/keys/%s/rotate"
+
+	wrappingKeyPath = "%s/wrapping_key"
+	rewrapKeyPath   = "%s/rewrap/%s"
+
+	encryptKeyPath = "%s/encrypt/%s"
+	decryptKeyPath = "%s/decrypt/%s"
+)
+
+const (
+	plaintextKey  = "plaintext"
+	ciphertextKey = "ciphertext"
+)
+
+type Transit struct {
+	c         *Client
+	mountPath string
+}
+
+func (c *Client) Transit(mountPath string) *Transit {
+	return &Transit{
+		c:         c,
+		mountPath: mountPath,
+	}
+}
+
+func (t *Transit) CreateKeyWithContext(ctx context.Context, keyName string) error {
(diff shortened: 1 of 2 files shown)


api/transit.go
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) HashiCorp, Inc.
```
cipherboy: Could we use the OpenBao header format?

> // Copyright The OpenBao Contributors


Thanks!

cipherboy:
@olegdayo @Nivedita-coder Do you want me to merge this now so you have a common place to begin implementing or should I wait for this to be more complete before we start merging? Happy to provide more in depth feedback if we want.

olegdayo:
@cipherboy 
I guess, you can merge, thank you
Please, share all the feedback you have, I'll be glad to make improvements;)

api/transit.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+func (t *Transit) CreateKeyWithContext(ctx context.Context, keyName string) error {
+	r := t.c.NewRequest(http.MethodPut, fmt.Sprintf(keyPath, t.mountPath, keyName))
+	_, err := t.c.rawRequestWithContext(ctx, r)
+	if err != nil {
+		return err
+	}
+	return nil
```
cipherboy, line 48: I think it is OK to do what KVv2 does in `api/kv_v2.go` and use `c.Logical()`'s helpers here.

api/transit.go
```diff
@@ -0,0 +1,147 @@
...
+	return t.ExportKeyWithContext(context.Background(), keyType, keyName)
+}
+
+func (t *Transit) Encrypt(keyName string, plaintext string) (string, error) {
+	return t.EncryptWithContext(context.Background(), keyName, plaintext)
+}
+
+func (t *Transit) Decrypt(keyName string, ciphertext string) (string, error) {
+	return t.DecryptWithContext(context.Background(), keyName, ciphertext)
+}
```
cipherboy, line 147: I think from #1555 we'll get a table of operations, but just commenting that I don't think anyone expects context-less calls. 

In other words, `Decrypt(...)` should just take `ctx` and we'll drop the `DecryptWithContext` version.

api/transit.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (c *Client) Transit(mountPath string) *Transit {
+	return &Transit{
+		c:         c,
+		mountPath: mountPath,
+	}
+}
+
+func (t *Transit) CreateKeyWithContext(ctx context.Context, keyName string) error {
```
cipherboy, line 42: I'd probably make a few fields required (mirroring the API itself) and then use Go's options pattern for the rest:

https://golang.design/research/generic-option/

Here, we'd probably use the actual request body as the configured struct. My 2c.

cipherboy (commented):
Some random thoughts!

cipherboy:
Hey @olegdayo! Let me know if you're still interested in picking this up; if so, feel free to reopen :-)

PR #1570, opened 2025-07-22
Remove upsert function usage in PostgreSQL backend

cipherboy:
On older PostgreSQL versions (before 9.5), native upsert capabilities were not supported and required a custom function. Switch to enforcing the use of PostgreSQL 9.5 as the oldest supported upstream PostgreSQL version is 12 per https://endoflife.date/postgresql.

Based on https://endoflife.date/rhel, the oldest maintenance supported RHEL release is 8.10, which ships natively with a newer PostgreSQL version per
https://rpmfind.net/linux/rpm2html/search.php?query=postgresql-server.

Deprecating this will be useful for later PostgreSQL read scalability.

----

@openbao/openbao-core-committers @openbao/openbao-org-maintainers let me know if anyone objects.

### DIFF
--- a/changelog/1570.txt
+++ b/changelog/1570.txt
@@ -1,0 +1,3 @@
+```release-note:deprecation
+storage/postgresql: remove support for legacy PostgreSQL versions before 9.5 which require a special upsert function.
+```


satoqz (approved):
Huh, I was looking at this code earlier today by coincidence. Makes sense to me.

wslabosz-reply (approved):

JanMa (approved):
LGTM

cipherboy:
Thanks all! I'll announce this as the next community call.

PR #1571, opened 2025-07-23
Implement fencing in PostgreSQL backend

cipherboy:
~This is based on #1570; will be rebased once that lands.~

----

    OpenBao already had mostly unused support for the fencing of writes to
    ensure the lock is actively held at the time of the write. In
    conjunction with our upcoming read-scalability mode, we'd like to guard
    at the storage layer against unintended writes that occur after we've
    lost the lock (whether it was intentionally grabbed by another node or
    otherwise).
    
    This fencing is already set up in vault/ha.go; we just need to ensure it
    is enforced at all writes.
    
    This makes one change to the write transaction: if a write transaction
    is acquired but no writes occur, we issue a rollback instead of a
    commit for consistency with Raft.

---

@KrzysztofKornalewski-Reply I realized I never finished the work for this; you can see the basis of the test structure here, you can base your changes off of this :-) 

(see `TestPostgreSQL_FencedWrites`)

I think you'll just need to add a declarative initialization hook to `DockerTestCluster` and maybe `TestCluster` as a whole.

### DIFF
--- a/physical/postgresql/postgresql.go
+++ b/physical/postgresql/postgresql.go
@@ -70,6 +70,7 @@ type PostgreSQLBackend struct {
 	haGetLockValueQuery      string
 	haUpsertLockIdentityExec string
 	haDeleteLockExec         string
+	haCheckLockHeldQuery     string
 
 	haEnabled     bool
 	logger        log.Logger
@@ -230,6 +231,10 @@ func NewPostgreSQLBackend(conf map[string]string, logger log.Logger) (physical.B
 		haDeleteLockExec:
 		// $1=ha_identity $2=ha_key
 		" DELETE FROM " + quotedHaTable + " WHERE ha_identity=$1 AND ha_key=$2 ",
+		haCheckLockHeldQuery:
+		// $1=ha_identity $2=ha_key $3=ha_value
+		" SELECT COUNT(*) FROM " + quotedHaTable + " WHERE " +
+			" ha_identity=$1 AND ha_key=$2 AND ha_value=$3 AND valid_until > NOW()  ",
 		logger:        logger,
 		txnPermitPool: physical.NewPermitPool(txnMaxParInt),
 		haEnabled:     conf["ha_enabled"] == "true",
@@ -553,12 +558,12 @@ func (p *PostgreSQLBackend) validateFence(ctx context.Context) error {
 		return nil
 	}
 
-	held, value, err := p.fence.Value()
+	held, err := p.fence.IsActivelyHeld(ctx)
 	if err != nil {
 		return fmt.Errorf("%v: err from database: %w", physical.ErrFencedWriteFailed, err)
 	}
-	if !held || value != p.fence.value {
-		return fmt.Errorf("%v: lock values differed", physical.ErrFencedWriteFailed)
+	if !held {
+		return fmt.Errorf("%v: lock changed ownership", physical.ErrFencedWriteFailed)
 	}
 
 	return nil
@@ -623,6 +628,37 @@ func (l *PostgreSQLLock) Value() (bool, string, error) {
 		return false, "", nil
 	default:
 		return false, "", err
+	}
+}
+
+// Whether or not this lock is currently held by this instance of the lock.
+//
+// Returns true if and only if this lock is active. Returns false if the lock
+// is held by another caller or an error occurred. While errors may occur
+// which prevent checking lock status, this likely also affects whether or not
+// the lock can be renewed and so should likely be treated as an error.
+//
+// While leaderLossCh returned from Lock() is the ultimate notification of
+// lock loss, this check is a online check and goes to the database to verify
+// the lock is held.
+func (l *PostgreSQLLock) IsActivelyHeld(ctx context.Context) (bool, error) {
+	pg := l.backend
+
+	// For simplicity and compatibility with all versions of PostgreSQL, we
+	// return the number of rows matching our lookup. This is zero if the
+	// lock isn't held by us (identity, key, and value all match and the lock
+	// is valid) and one if it is held by us. Uniqueness constraints prevent
+	// us from having more than one lock.
+	var result int
+	err := pg.client.QueryRowContext(ctx, pg.haCheckLockHeldQuery, l.identity, l.key, l.value).Scan(&result)
+
+	switch err {
+	case nil:
+		return result == 1, nil
+	case sql.ErrNoRows:
+		return false, nil
+	default:
+		return false, err
 	}
 }
 
--- a/sdk/physical/physical.go
+++ b/sdk/physical/physical.go
@@ -155,7 +155,7 @@ type RedirectDetect interface {
 type Lock interface {
 	// Lock is used to acquire the given lock
 	// The stopCh is optional and if closed should interrupt the lock
-	// acquisition attempt. The return struct should be closed when
+	// acquisition attempt. The returned channel will be closed when
 	// leadership is lost.
 	Lock(stopCh <-chan struct{}) (<-chan struct{}, error)
 
@@ -164,6 +164,14 @@ type Lock interface {
 
 	// Returns the value of the lock and if it is held by _any_ node
 	Value() (bool, string, error)
+}
+
+// When Lock is implemented by a remote server, typically RemoteLock
+// should also be implemented.
+type RemoteLock interface {
+	// IsActivelyHeld allows callers to online validate that this lock
+	// instance is valid, independently of a closed leader loss channel.
+	IsActivelyHeld(ctx context.Context) (bool, error)
 }
 
 // Factory is the factory function to create a physical backend.


physical/postgresql/postgresql.go
```diff
@@ -511,6 +552,48 @@ func (p *PostgreSQLBackend) HAEnabled() bool {
...
+		if result != p.fence.value {
+			return fmt.Errorf("%v: lock values differed", physical.ErrFencedWriteFailed)
+		}
+
+		return nil
+	case sql.ErrNoRows:
+		return fmt.Errorf("%v: lock was missing", physical.ErrFencedWriteFailed)
+	default:
+		return fmt.Errorf("%v: err from database: %w", physical.ErrFencedWriteFailed, err)
+	}
```
satoqz: This duplicates `(*PostgresSQLLock).Value()`, not sure if intentional

cipherboy: @satoqz It does, I couldn't see any easy way to reuse it if I'm not mistaken unless I acquired a transaction and threw it away I suppose, and made `valueTxn(...)` take the txn. I was hoping for a common interface between them but I guess https://github.com/golang/go/issues/14468 remains open. 

(Conceptually, I'm having a hard timing understanding if a read-only rolled back transaction behaves the same as no transaction would here. I think committing the read transaction would nominally behave differently because you'd validate that all your reads have successfully concluded and haven't changed due to a modifying write transaction, but I'm not sure.)

satoqz: @cipherboy ah, my bad, I missed that it needs to operate on the transaction.

Maybe you can come up with a test that you can (manually) run to understand the behavior? I like to just open two `psql` terminals, `begin transaction;` in each and see how things interact 😄.

cipherboy: @satoqz In this case, the repeatable read with commit did not actually fail in the way I expected/hoped:

(started second, committed first)
```sql
username=# BEGIN TRANSACTION ISOLATION LEVEL REPEATABLE READ;
BEGIN
username=*# UPDATE alex SET id=id+1 WHERE value='sam';
UPDATE 2
username=*# COMMIT
username-*# ;
```

(started first, committed second)
```sql
username=# BEGIN TRANSACTION ISOLATION LEVEL REPEATABLE READ;
BEGIN
username=*# SELECT * FROM alex WHERE value='sam';
 id | value 
----+-------
  1 | sam
  3 | sam
(2 rows)

#### (ascheel) above terminal fully executed here here ####

username=*# COMMIT;
COMMIT
username=# SELECT * FROM alex WHERE value='sam';
 id | value 
----+-------
  2 | sam
  4 | sam
(2 rows)
```

Putting this in our transaction commit logic is probably the correct answer, but still wouldn't be as durable as adjusting the queries themselves to check it.

My SQL has grown weak over the years, but something like:

```sql
WITH lock as (
  SELECT ha_value FROM " + quoted_ha_table + " WHERE NOW() <= valid_until AND ha_key = $3
)
DELETE FROM openbao_kv_store
  WHERE
    openbao_kv_store.path=$1 AND
    openbao_kv_store.key=$2 AND
    lock.value == $4
```

An alternative would be something even more ugly like using an `ACCESS SHARE` lock over the locks table to read the value (held through the entire transaction), with `ACCESS EXCLUSIVE` during lock updates. 

What's your thoughts on the approaches? Is this safe _enough_? Should we try for better?

satoqz: Adjusting the queries would be the safest indeed, not sure what side effects that would have though.
On the other hand, if we're fine with a good-enough sloppy read, wouldn't it be better not to make that read part of the transaction so you'd get potentially more up to date information?

I don't think just reading (without it enforcing a constraint visible to postgres) will ever cause serialization failures unless you write to the same row in some form, as you've discovered.

cipherboy: @satoqz Ah, that's fair. Yes, I'll move this outside of the transaction then.

cipherboy:
Looks like test failures since:

> Quay Container Registry  is currently in read-only mode. Pulls and other read-only operations will succeed but all other operations are currently suspended.

cipherboy:
@satoqz This should be updated :-)

satoqz (approved):

cipherboy:
@satoqz This could use a re-review, I pushed two new commits to guard it only on active, HA-enabled nodes.

physical/postgresql/invalidation.go
satoqz, line 1: New file really needed?

physical/postgresql/postgresql.go
```diff
@@ -541,10 +546,26 @@ func (p *PostgreSQLBackend) RegisterActiveNodeLock(l physical.Lock) error {
...
-	if p.fence == nil {
+	// Previously, standby nodes could theoretically attempt to write to
+	// storage. Explicitly guard against that with the active standby
+	// notification state.
+	active := p.active.Load()
+	if !active {
+		return logical.ErrReadOnly
+	}
+
+	if p.fence == nil && p.active.Load() {
```
satoqz: For understanding, I assume this one is hit during self-init?

satoqz (approved):
I apologize for taking a while, this looks good to me!

physical/postgresql/postgresql.go
cipherboy: @satoqz No, it is hit when using a non-HA enabled PostgreSQL backend (currently the default). In this case there are no locks acquire (assumed to be exclusive) and so there's no fence.

satoqz: Got it, yeah this makes sense 😄

satoqz: @cipherboy though I am slightly confused as to how this ties in with

https://github.com/openbao/openbao/blob/c7da3387e9776147b3f6e216aefca5ba10d2f938/physical/postgresql/postgresql.go#L549-L554

cipherboy: @satoqz Ah I see. Yes, earlier in the branch there were places were we'd be called in HA mode without the lock (namely, unseal). But I think that shouldn't be the case now. :thinking:

physical/postgresql/invalidation.go
cipherboy: @satoqz This was based on #1284, which actually fully populates this new file. So for this one thing, probably not. But going forward, yes. :laughing:

cipherboy:
@satoqz I've rebased this and checked that it doesn't conflict with #1986. I also dropped the change around active node, based on discussion here: https://github.com/openbao/openbao/pull/1986#discussion_r2524451674

cipherboy:
@satoqz @genelet Just to reason through the implications of this:

1. It means we're free to write from standby nodes still, if we bypass the barrier.
2. Active node lock being stolen will mean that the formerly active node doesn't attempt to perform any writes.
3. Lock holding allows writing all data. If we later do per-namespace active nodes, we almost should create separate physical PostgreSQL instances (like we would do with Raft) to ensure we correctly scope writes to specific database segments. That way we could strictly chroot the PostgreSQL instance to a specific table (and assign a corresponding lock to that) and not risk confusion.

Of these, 1 (and reintroducing https://github.com/openbao/openbao/commit/347388e5d7d860cbfcde78563828d8fc82fb98c0) seems like the only controversial statement.

Thoughts?

sdk/physical/physical.go
```diff
@@ -166,6 +166,14 @@ type Lock interface {
 	Value() (bool, string, error)
 }
 
+// When Lock is implemented by a remote server, typically RemoteLock
+// should also be implemented.
+type RemoteLock interface {
+	// IsActivelyHeld allows callers to online validate that this lock
+	// instance is valid, independently of a closed leader loss channel.
+	IsActivelyHeld(ctx context.Context) (bool, error)
+}
```
satoqz, line 175: I understand how this is useful internally for fencing, but is there a reason to expose it as a generic API?

cipherboy: So lock de-acquisition/theft only gets detected periodically, right? 

https://github.com/openbao/openbao/blob/7cf1ef02a93430fc5170fc44654acb8d639575fa/physical/postgresql/postgresql.go#L605-L614

Every `PostgreSQLLockRenewInterval = 5 * time.Second` we check for lock renewal and (after #2100) err if it isn't currently actively held by us.

 If this hangs for some longer period of time, we do not get notified about lock loss. We directly call `Exec(...)` without a context in this case (#2100 fixes that, too) and thus could actually hang _indefinitely_. 

Either way, we allow up to 2/3rds the total lock time to get an answer (2/3 * 15 seconds -> 10 seconds). 

That means there's a period of 10 seconds where we can't detect we're not the active leader. As long as we're fenced, we should be in good state there. But if we ever bypass this fence check, we'll actually succeed in the write.

My thought is we should likely ensure `IsActivelyHeld(...)` on certain sensitive operations (thinking mostly root/recovery/seal config/barrier keyring rotations) just to give us a good error message in this case, rather than a fence failure.

Thoughts?

cipherboy:
> ...unless we hit the fence? Are you describing the state before this PR or after? Or is this just a remark that the underlying postgres client is fundamentally never guaranteed read-only?

@satoqz No; standby nodes do not hold the lock and thus can't fence around it.

I pulled the commit from this series that sent active node status down to the PostgreSQL backend (or think I did? intended to :laughing:), as it is redundant w.r.t. the barrier changes merged in #1986.

satoqz:
> No; standby nodes do not hold the lock and thus can't fence around it.

but they still _have_ an _instance_ of the lock, no? So wouldn't `validateFence/IsActivelyHeld` fail?

sdk/physical/physical.go
satoqz: > My thought is we should likely ensure IsActivelyHeld(...) on certain sensitive operations (thinking mostly root/recovery/seal config/barrier keyring rotations) just to give us a good error message in this case, rather than a fence failure.

That makes sense yeah. It would be nice to document all the subtle differences and the concrete purpose of mechanisms a bit better in `physical.go`. I see `FencingBackend` has a long doc comment, but the one on `HABackend` is arguably vague and `Lock` doesn't have any additional information at all. Each time coming back to this PR I've had to slowly rebuild my world view around HA backends 😄

cipherboy:
> but they still _have_ an _instance_ of the lock, no? So wouldn't `validateFence/IsActivelyHeld` fail?

No, they don't. You don't want them to, either. You could call a two step:

```
rLock := newLock(name, "read")
rwLock := newLock(name, rLock.Value())
```

And technically (from a fencing PoV) take ownership of the lock, kinda. (It depends on your implement. This is true of Inmem HA but not true for PostgreSQL as it also validates the internal instance ID which is randomly assigned by the backend, independent of the value).

Lock is only actually set (w.r.t. fencing) on the active node and is cleared when it steps down. See `vault/ha.go` unless my memory is faulty.

satoqz:
@cipherboy got it, I had scanned the code too quickly. Thanks for clarifying!

sdk/physical/physical.go
cipherboy: Yeah, I think once PostgreSQL stabilizes with horizontal scalability, we'll have a better understanding of how all these interact and can create a proper overview of storage at that point in the file.

cipherboy:
Merging based on out-of-band discussion with @satoqz. Thank you!

PR #1581, opened 2025-07-23
Update run.mdx: HELM now exposes the web ui

LasseRosenow:
When reading the documentation while setting up openbao I was confused that it suddenly told me the web ui was not exposed even though I was already able to access it.

Then I found out that this documentation is out of date :) So here is a small fix :)


cipherboy (approved):
Thank you @LasseRosenow!

satoqz (approved):

Issue #1579, opened 2025-07-23
nilnil (returning both a nil value and nil error) - 9 findings


phyrog:
Findings:

```
vault/barrier_aes_gcm.go:847:3: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
		return nil, nil
vault/barrier_aes_gcm.go:943:3: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
		return nil, nil
vault/barrier_aes_gcm.go:957:3: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
		return nil, nil
sdk/database/dbplugin/v5/grpc_client.go:218:3: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
		return nil, nil
sdk/physical/cache.go:232:5: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
				return nil, nil
sdk/physical/physical_view.go:58:3: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
		return nil, nil
api/kv_v2.go:546:3: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
		return nil, nil
api/kv_v2.go:587:3: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
		return nil, nil
api/logical.go:198:4: return both a `nil` error and an invalid value: use a sentinel error instead (nilnil)
			return nil, nil
```

cipherboy:
@phyrog Wow this is far fewer than I expected. The issue is we've encoded `nil, nil` in our way of saying not found rather than `nil, ErrNorFound`, at both the storage and the API endpoint response layer.

It isn't very ideomatically Go, but not sure that can change in these layers.

I'm happy to fix some things but not if they're in those layers (as we're often taking `nil, nil` and stubbing in a default value since create/update are _frequently_ the same operation and we'd have to special case the error check which seems weirder than `nil, nil` after this point).

phyrog:
Yeah I'm totally fine to mark these with `//nolint`, as long as we have validated that this is the intended behavior. Would also document this for posterity.

satoqz:
> `ErrNorFound`

[...]

PR #1583, opened 2025-07-23
Add External Keys registry & API endpoints

satoqz:
Adds the `/sys/external-keys` endpoints described in #1320. 

Validation of configs / keys and client creation / cache management is of course missing, since we don't have the KMS interface for that yet.

I'd be happy if someone could take extra care to review the handling of transactions/locks to achieve consistency between configs and keys.. I wasn't so sure in all cases.

---

cc @phyrog @Huy-Dinh-Reply since you offered to review -- I hope you'll find that I was right saying this was 90% boilerplate 😄

### DIFF
--- a/vault/external_key_registry.go
+++ b/vault/external_key_registry.go
@@ -5,7 +5,6 @@ package vault
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"path"
@@ -36,9 +35,9 @@ type ExternalKeyRegistry struct {
 	core   *Core
 	logger hclog.Logger
 
-	// storageLock is used in some edge cases where transactions are not enough
-	// to ensure config <-> key consistency, essentially because we have no
-	// concept of foreign key constraints.
+	// storageLock ensures consistency both within and between configs and keys.
+	// In the unlikely event this ever becomes a bottleneck, we can shard this
+	// lock per config via locksutil.
 	storageLock sync.RWMutex
 }
 
@@ -233,35 +232,37 @@ func (r *ExternalKeyRegistry) PutConfig(
 	name := data.Get("config").(string)
 	storage := r.storageViewForRequest(req)
 
-	if err := logical.WithTransaction(ctx, storage, func(storage logical.Storage) error {
-		// Check for an existing config, we may have a conflict e.g. when trying
-		// to change a a typed config to an inherited config.
-		config, err := r.getConfigCommon(ctx, storage, name)
-		if err != nil {
-			return err
-		}
+	r.storageLock.Lock()
+	defer r.storageLock.Unlock()
 
-		// Create a new config if it doesn't exist yet.
-		if config == nil {
-			config = &ExternalKeyConfig{}
-		}
+	// Check for an existing config, we may have a conflict e.g. when trying
+	// to change a a typed config to an inherited config.
+	config, err := r.getConfigCommon(ctx, storage, name)
+	if err != nil {
+		return handleError(err)
+	}
 
-		if err := r.validateConfigTransition(
-			config.Type != "", ty != "", config.Inherits != "", inherits != "",
-		); err != nil {
-			return err
-		}
+	// Create a new config if it doesn't exist yet.
+	if config == nil {
+		config = &ExternalKeyConfig{}
+	}
 
-		config.Type = ty
-		config.Inherits = inherits
-		config.Values = values
+	if err := r.validateConfigTransition(
+		config.Type != "", ty != "", config.Inherits != "", inherits != "",
+	); err != nil {
+		return handleError(err)
+	}
 
-		if config.Inherits != "" && len(config.Values) != 0 {
-			return fmt.Errorf("setting field %q requires that no other fields are set", "inherits")
-		}
+	config.Type = ty
+	config.Inherits = inherits
+	config.Values = values
 
-		return r.putConfigCommon(ctx, storage, name, config)
-	}); err != nil {
+	if config.Inherits != "" && len(config.Values) != 0 {
+		return handleError(fmt.Errorf("setting field %q requires that no other fields are set",
+			"inherits"))
+	}
+
+	if err := r.putConfigCommon(ctx, storage, name, config); err != nil {
 		return handleError(err)
 	}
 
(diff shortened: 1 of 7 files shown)


vault/external_key_registry.go
```diff
@@ -0,0 +1,659 @@
...
+func (r *ExternalKeyRegistry) GetConfig(
+	ctx context.Context, req *logical.Request, data *framework.FieldData,
+) (*logical.Response, error) {
+	name := data.Get("config").(string)
+	storage := r.storageViewForRequest(req)
+
+	config, err := r.getConfigCommon(ctx, storage, name)
+	switch {
+	case err != nil:
+		handleError(err)
```
rencooo:
```suggestion
		return handleError(err)
```

vault/external_key_registry.go
```diff
@@ -0,0 +1,659 @@
...
+
+		return r.putConfigCommon(ctx, storage, name, config)
+	}); err != nil {
+		return handleError(err)
+	}
+
+	return nil, nil
+}
+
+// validateConfigTransition enforces the invariants in defined below.
```
rencooo:
```suggestion
// validateConfigTransition enforces the invariants defined below.
```

vault/external_key_registry.go
```diff
@@ -0,0 +1,659 @@
...
+
+	if err := logical.WithTransaction(ctx, storage, func(storage logical.Storage) error {
+		key, err := r.getKeyCommon(ctx, storage, fullName)
+		if err != nil {
+			return err
+		}
+
+		// Create a new key if it doesn't exist yet.
+		if key == nil {
+			// If the corresponding config gets deleted before we create a new
```
Huy-Dinh-Reply: I'm wondering if this race can happen in other places where `putKeyCommon` is called (`PatchKey`, `PutGrant`, `DeleteGrant`) and thus we would need to lock `storageLock` as well

satoqz: I don't think so as all of those require the key to already exist. In that case another write to the existing key would conflict with a delete in `DeleteConfig(...)`. But willing to be convinced otherwise :D

Huy-Dinh-Reply: So if I understood [this documentation](https://www.postgresql.org/docs/current/transaction-iso.html) correctly (and this is one big if 😄), in the default transaction isolation level, Read Committed, phantom reads can occur in transactions. So I'm assuming something like this can happen:
* Thread 1 starts its transaction in `PatchKey`, check if the key exists, it does at the time
* Context switch to thread 2 in `DeleteConfig`, it deletes the keys and the config
* Thread 1 continues and eventually calls `putKeyCommon`, then `storage.Put`, which (I assume) is an upsert (because the same function is called for key creation) and thus will be accepted even after the deletion. This causes the key to be re-created.

satoqz: Right, I did some testing and this indeed does not work out. Will update to lock more aggressively, I'll sleep better at night like that anyways.

OpenBao even uses `repeatable read` by default, but that still doesn't cause a put to fail because of a delete that committed earlier 🙃

https://github.com/openbao/openbao/blob/main/physical/postgresql/transaction.go?plain=1#L30

satoqz: @Huy-Dinh-Reply updated, LMKWYT. We don't use the read mode of the `RWMutex` for now, but it'll come in handy once we're able to construct a `logical.ExternalKey` for calls from plugins and read multiple configs + a key at once.

cipherboy: @Huy-Dinh-Reply @satoqz I guess my test case is flawed then:

> OpenBao even uses repeatable read by default, but that still doesn't cause a put to fail because of a delete that committed earlier 🙃

Oh hm, that's too bad. Should we go to serializable? There was some reason why I didn't... I think because (from testing) it seemed to lock the row, which meant we didn't get a conflict failure and also couldn't execute parallel transactions over the same data? At any rate, I think our test case validates this (our expected transaction semantics) with a conflicting `Put(...)`, it'd be interesting to update this for delete too.

But yes, I think you're going to still need locking, I don't think Transactions magically let us avoid locking sadly.

satoqz: `serializable` does fix this, yeah.

> it seemed to lock the row, which meant we didn't get a conflict failure and also couldn't execute parallel transactions over the same data?

As long as a transaction has an outstanding write to a row and it hasn't committed/rollbacked yet, it'll block other writes to the row. You get the conflict together with an unblock once the original transaction commits.

> I think our test case validates this (our expected transaction semantics) with a conflicting Put(...), it'd be interesting to update this for delete too.

Where can I find those tests?

cipherboy: @satoqz My bad; my memory is playing tricks on me. See https://github.com/openbao/openbao/commit/b2185b13c57eb2609971bb8567d7ed886fab5bff#diff-3934ca25dfaae234be5f9b0cf6ca59849c05973e8fae80fe69380f24dfba3a9f and #608, and comment here: https://github.com/openbao/openbao/pull/608#discussion_r1794426508 

I had also added this note:

> Raft and PostgreSQL will have different semantics: PostgreSQL prevents a single thread from interacting with two separate transactions, so developers will need to take care to avoid this.

vault/external_key_registry.go
```diff
@@ -0,0 +1,628 @@
...
+}
+
+// keyPath returns the storage path for a key.
+func (r *ExternalKeyRegistry) keyPath(name string) string {
+	return path.Join(externalKeyRegistryKeyPrefix, name)
+}
+
+// keyListPath returns the storage path to list keys at.
+func (r *ExternalKeyRegistry) keyListPath(name string) string {
+	return path.Join(externalKeyRegistryKeyPrefix, name) + "/"
```
cipherboy, line 117: Sorry, what's under this path? I think I expected it to be `externalKeyRegistryKeyPrefix/`.

vault/external_key_registry.go
```diff
@@ -0,0 +1,628 @@
...
+	return path.Join(externalKeyRegistryConfigPrefix, name)
+}
+
+// configListPath returns the storage path to list configs at.
+func (r *ExternalKeyRegistry) configListPath() string {
+	return externalKeyRegistryConfigPrefix
+}
+
+// keyPath returns the storage path for a key.
+func (r *ExternalKeyRegistry) keyPath(name string) string {
```
cipherboy, line 111: Given keys are under configs in the API definition (I thought?), I'm surprised `config` isn't a parameter here?

vault/external_key_registry.go
```diff
@@ -0,0 +1,628 @@
...
+
+	config, err := r.getConfigCommon(ctx, storage, name)
+	switch {
+	case err != nil:
+		return handleError(err)
+	case config == nil:
+		return nil, logical.CodedError(http.StatusNotFound, "config not found")
+	}
+
+	resp := &logical.Response{Data: make(map[string]any)}
```
cipherboy, line 195: I'm mildly entertained that you're going to `logical.Request` here versus in `logical_system_external_keys.go`, but I suppose it is fine.

vault/external_key_registry.go
```diff
@@ -0,0 +1,628 @@
...
+		if !ok {
+			return handleError(fmt.Errorf("expected field %q to be a string", k))
+		}
+		values[k] = str
+	}
+
+	// We remove these from the values map and place them in their own dedicated
+	// fields. This is not strictly needed, but should make it easy to pass the
+	// the remaining values right to the KMS interface.
+	ty, inherits := values["type"], values["inherits"]
```
cipherboy, line 228: :bow: tyvm

vault/external_key_registry.go
```diff
@@ -0,0 +1,628 @@
...
+		return handleError(err)
+	}
+
+	return nil, nil
+}
+
+// validateConfigTransition enforces invariants for modifications to External
+// Key config entries. There are a few here, but the general idea is that a
+// config must be either inherited or "typed", and we can move from an inherited
+// config to a typed config, but not the other way.
```
cipherboy, line 335: Hadn't even considered this, tbh. :laughing: 

I think it is fine though.

vault/external_key_registry.go
```diff
@@ -0,0 +1,628 @@
...
+
+	resp := &logical.Response{Data: make(map[string]any)}
+	for k, v := range key.Values {
+		resp.Data[k] = v
+	}
+
+	return resp, nil
+}
+
+// PUT /sys/external-keys/configs/:config-name/keys/:key-name
```
cipherboy, line 421: (ah, so I was correct above; `keys` are nested under configs. Given we want locally-unique key names, shouldn't we address that? :thinking:)

vault/external_key_registry.go
```diff
@@ -0,0 +1,628 @@
...
+
+	if !slices.Contains(key.Grants, mount) {
+		return nil, nil
+	}
+
+	key.Grants = slices.DeleteFunc(key.Grants, func(grant string) bool {
+		return grant == mount
+	})
+
+	if err := r.putKeyCommon(ctx, storage, name, key); err != nil {
```
cipherboy, line 621: Ah, grants are flat on the key. I think this should be fine, keys should mostly be single use. Otherwise we'll need to go much further to mount table with multiple separate entries.

cipherboy (commented):
Some comments @satoqz but looking fairly good. 

I'll trust @Huy-Dinh-Reply and @rencooo looked at the tests this time. :-)

vault/external_key_registry.go
satoqz: I see I've caused some confusion with names and paths, let me clear it up 😄

On the API level we indeed have `sys/external-keys/configs/:config-name/keys/:key-name`, and key names are unique per config, not per namespace.

I decided to go for a different storage layout however:

```
sys/external-keys/configs/:config-name
                          ^ this is an entry
sys/external-keys/keys/:config-name/:key-name
                                    ^ this is an entry
```

Mainly because this lets you maintain a directory-like structure, so you don't end up with:

```
sys/external-keys/configs/:config-name
                          ^ this is an entry
sys/external-keys/configs/:config-name/:key-name
                                       ^ this is also an entry, ew
```

In this particular instance (`keyListPath(name)`), `name` would be the config name and yield `sys/external-keys/keys/:config-name/`, listing the keys for that config only.

vault/external_key_registry.go
satoqz: `name` would already be format `:config-name/:key-name` here, i.e. "the full key name", similar to what it will be for engines. LMK if this is too confusing/if you'd prefer separate parameters/if a comment would suffice. I was initially passing `keyName, configName string` around the entire world which got very tedious.

vault/external_key_registry.go
satoqz: Initially I had a dedicated handler function in `logical_system_external_keys.go` and abstracted the entire request/response layer away from the registry itself, but it eventually felt like premature optimization and very boilerplatey. I don't expect these APIs to receive calls other than from request handling, and the unified version was much simpler to reason about.

vault/external_key_registry.go
satoqz: Curious to hear more about the alternative approach you're hinting at 😄

vault/external_key_registry.go
cipherboy: I think that's fine.

vault/external_key_registry.go
cipherboy: Why not move most of this into that file even without the separation then?

vault/external_key_registry.go
satoqz: `ty` is a Rust-ism leaking through, I'm terribly sorry.

vault/external_key_registry.go
satoqz: We can do that, it would sort of split brain the registry implementation though. I figured `logical_system_external_keys.go` could be fully declarative, and all actual functionality remains in `external_key_registry.go`. Your call 😄

vault/external_key_registry.go
cipherboy: @satoqz Ah just separate entries and using a transaction like we do for the split mount table.

vault/external_key_registry.go
```diff
@@ -0,0 +1,628 @@
...
+	if config.Inherits != "" && len(config.Values) != 0 {
+		return handleError(fmt.Errorf("setting field %q requires that no other fields are set",
+			"inherits"))
+	}
+
+	if err := r.putConfigCommon(ctx, storage, name, config); err != nil {
+		return handleError(err)
+	}
+
+	return nil, nil
```
wslabosz-reply, line 329: linter won't be happy

wslabosz-reply (approved):
I'd look into nilnils and disable linting preemptively if they make sense, or adjust otherwise

vault/external_key_registry.go
satoqz: It will be, actually. I don't know why, but as https://github.com/openbao/openbao/issues/1579 shows (and I've revalidated on this branch), the linter doesn't complain about `nil, nil` in request handling.

KrzysztofKornalewski-Reply:
> I'd look into nilnils and disable linting preemptively if they make sense, or adjust otherwise

I'm currently working on this issue. In the repository, there are nine places that need to be cleaned up.
https://github.com/openbao/openbao/issues/1579
The scope is quite large, as there are approximately 960 occurrences in the repository (grep -R --include="*.go" -n "return nil, nil" . | wc -l). Refactoring some of these places is difficult because there are no dedicated UTs, so the compatibility changes is not straightforward in the tests.

vault/external_key_registry.go
cipherboy: Ah I see. That's fine I suppose.

PR #1585, opened 2025-07-24
docs(guides): Securosys PrimusHSM using PKCS#11

pree:
This adds a new section to the website's docs section: Guides.
Also adds a Guide on how to configure OpenBao to Auto-Unseal using Securosys PrimusHSM.
I'm also in the process of create a guide on how to unseal using KMS / the Transaction Broker Service, but this will be handled in a seperate PR.

Resolves #1053

PR #1586, opened 2025-07-24
docs(pkcs11): add AES key generation example

pree:
Changes to example to also include AES key generation.


satoqz (commented):
Sorry, I believe you got me slightly wrong in https://github.com/openbao/openbao/pull/1585#discussion_r2228658801 😄

This command is correct, but if you were to use AES, you'd need to change `--keypairgen` to `--keygen`. So the better option is probably to add another example with server config + command tailored to AES.

pree:
Yeah I just realized that too ...

pree:
I think I've got it now correctly, thanks again @satoqz :)

PR #1587, opened 2025-07-24
Fix ergonomics around self-init, add documentation

cipherboy:
This fixes a few ergonomics around self-init (making names a little more legible, expanding field names, &c) and provides documentation for the feature.

---

@KrzysztofKornalewski-Reply Do you mind splitting this document (as you see fit) into a page under `/docs/concepts/profiles` for reference and reuse? I think it makes sense to avoid duplicating this when we go to add the rest of the profile system. 

Probably it is worth keeping the bit about the `initialize` block(s), that `request` and `response` sources take a `initialize_name` parameter, and declaring the (four) types of sources we support, but moving a lot of the rest over to the concepts page. 

Thanks!

### DIFF
--- a/website/content/docs/configuration/self-init.mdx
+++ b/website/content/docs/configuration/self-init.mdx
@@ -1,0 +1,265 @@
+---
+sidebar_label: initialize
+description: |-
+  The initialize stanza allows request-based one-time initialization of OpenBao
+  from the configuration file.
+---
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# `initialize` stanza
+
+The `initialize` stanza specifies various configurations for OpenBao to
+initialize itself once, on initial startup. To repeat the operation, remove
+all storage and re-initialize from scratch. Debugging can be performed by
+starting OpenBao with the `TRACE` log level, which shows all request/response
+pairs and thus **contains sensitive information**.
+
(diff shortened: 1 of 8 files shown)


website/content/docs/configuration/self-init.mdx
```diff
@@ -0,0 +1,265 @@
...
+    token = {
+      eval_type = "string"
+      eval_source = "response"
+      initialize_name = "authenticate"
+      request_name = "admin-userpass"
+      field_selector = ["auth", "client_token"]
+    }
+    data = {}
+  }
+}
```
wslabosz-reply: @cipherboy 
there's no `response` source here, and I wonder if the `request_name` just before is supposed to be `response_name`?

cipherboy: Ah, yep, should be `response_name`. My bad!

KrzysztofKornalewski-Reply:
@cipherboy sry for the delay. I haven't been online since Thursday afternoon. Thanks for the advice, I can split it. No problem!

website/content/docs/configuration/self-init.mdx
cipherboy: @wslabosz-reply Fixed, lmkwyt!

wslabosz-reply (approved):

cipherboy:
Thanks @wslabosz-reply! Merging...

PR #1590, opened 2025-07-25
Move inline auth to Client from Logical

cipherboy:
Keeping this on Logical() limits the consumption; uses like Sys() and KVv2() and other top-level clients cannot also use this without reimplementing it. By moving it to Client, it sits alongside SetToken(...) and thus should be more discoverable.

As this has not landed in a release, it shouldn't require a changelog entry.

---

cc: @fcbajao

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -6,6 +6,8 @@ package api
 import (
 	"context"
 	"crypto/tls"
+	"encoding/base64"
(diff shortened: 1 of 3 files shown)


wslabosz-reply (approved):

cipherboy:
Thanks all!

Issue #1543, opened 2025-07-11
Namespace stuck in "tainted" state after parallel deletion of namespace and mounts


driif:
## Description

When deleting a namespace that contains active mounts (e.g., PKI engine) - performing a `namespace delete` and `secrets disable` for the PKI engine concurrently can leave the namespace in a "tainted" state. This behavior suggests a race condition during deletion cleanup.

## Steps to Reproduce
Use this script:
https://github.com/driif/bao-assets/blob/main/delete-ns-race-condition/stress.sh

## Expected behavior

```
==> Deleting namespace stressns
Key       Value
---       -----
status    in-progress

==> Disabling PKI engine pki in stressns
Error disabling secrets engine at pki/: Error making API request.

Namespace: stressns/
URL: DELETE http://127.0.0.1:8200/v1/sys/mounts/pki
Code: 400. Errors:

* no handler for route "sys/mounts/pki" on namespace "stressns/". route entry or namespace is tainted.

==> ./bao namespace list:
No namespaces found
```

## Unexpected behavior

```
==> Ready to race deletion...
==> Deleting namespace stressns

==> Disabling PKI engine pki in stressns
Key       Value
---       -----
status    in-progress
Success! Disabled the secrets engine (if it existed) at: pki/

==> Deletion race trigger complete. Checking state:
Keys
----
stressns/
```

Logs:
```
2025-07-11T11:02:27.738+0200 [INFO]  core: disabled credential backend: namespace=stressns/ path=auth/token/
2025-07-11T11:02:27.745+0200 [INFO]  core: disabled credential backend: namespace=stressns/ path=auth/userpass1/
2025-07-11T11:02:27.750+0200 [INFO]  core: disabled credential backend: namespace=stressns/ path=auth/userpass2/
2025-07-11T11:02:27.755+0200 [INFO]  core: disabled credential backend: namespace=stressns/ path=auth/userpass3/
2025-07-11T11:02:27.759+0200 [INFO]  core: disabled credential backend: namespace=stressns/ path=auth/userpass4/
2025-07-11T11:02:27.763+0200 [INFO]  core: disabled credential backend: namespace=stressns/ path=auth/userpass5/
2025-07-11T11:02:27.767+0200 [INFO]  core: successfully unmounted: namespace=stressns/ path=cubbyhole/
2025-07-11T11:02:27.768+0200 [INFO]  core: successfully unmounted: namespace=stressns/ path=pki/
[...]

PR #1594, opened 2025-07-28
Ignore missing mount during namespace deletion

phil9909:
This can happen, if a mount deletion is requested in parallel.

Resolves https://github.com/openbao/openbao/issues/1543

### DIFF
--- a/changelog/1594.txt
+++ b/changelog/1594.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-Ignore missing mounts when deleting a namespace. This can happen when a mount is unmounted in prallel.
+Ignore missing mounts when deleting a namespace. This can happen when a mount is unmounted in parallel.
 ```
--- a/changelog/1594.txt
+++ b/changelog/1594.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Ignore missing mounts when deleting a namespace. This can happen when a mount is unmounted in prallel.
+```
--- a/vault/core.go
+++ b/vault/core.go
@@ -142,6 +142,9 @@ var (
 	// enabled in the configuration file
 	ErrIntrospectionNotEnabled = errors.New("The Vault configuration must set \"introspection_endpoint\" to true to enable this endpoint")
 
+	// ErrNoMatchingMount is returned if the mount is not found
+	errNoMatchingMount = errors.New("no matching mount")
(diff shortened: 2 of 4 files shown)


vault/core.go
```diff
@@ -142,6 +142,9 @@ var (
 	// enabled in the configuration file
 	ErrIntrospectionNotEnabled = errors.New("The Vault configuration must set \"introspection_endpoint\" to true to enable this endpoint")
 
+	// ErrNoMatchingMount is returned if the mount is not found
+	errNoMatchingMount = errors.New("no matching mount")
```
```suggestion
	// errNoMatchingMount is returned if the mount is not found
	errNoMatchingMount = errors.New("no matching mount")
```

changelog/1594.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+Ignore missing mounts when deleting a namespace. This can happen when a mount is unmounted in prallel.
```
satoqz:
```suggestion
Ignore missing mounts when deleting a namespace. This can happen when a mount is unmounted in parallel.
```

driif (approved):

vault/namespace_store.go
```diff
@@ -818,6 +818,9 @@ func (ns *NamespaceStore) clearNamespaceResources(nsCtx context.Context, namespa
 	for _, me := range mountEntries {
 		err := ns.core.unmountInternal(nsCtx, me.Path, true)
 		if err != nil {
+			if err == errNoMatchingMount {
```
wslabosz-reply: go docs suggest this is more preferable to the equality check
```suggestion
			if errors.Is(err, errNoMatchingMount) {
```

vault/core.go
phil9909: Why make the error public?

wslabosz-reply: the diff shows comment being lowercased

phil9909: You are absolutely right.
Sorry, I mixed up red and green :D

satoqz (approved):

phil9909:
@satoqz Can you merge this please?

satoqz:
@phil9909 sure. Not sure if we still wanted to explore alternative approaches first, but I suppose this is a small change we can easily iterate on later.

PR #1584, opened 2025-07-24
Implement cache invalidation for quotas

phil9909:
Implement cache invalidation for quotas.
Required as a prerequisite for enabling standby nodes to handle read requests.

See #1550
See #1528 

Notice: This PR targets the [read-replication](https://github.com/openbao/openbao/tree/read-replication) feature branch

PR #1592, opened 2025-07-27
Unseal standby nodes

fatima2003:
Enabling standby nodes to handle read request has 2 main steps:
1. unsealing standby nodes.
2. handling read request and forwarding writes.

I was having a bunch of bugs tryna shoot both down in one PR, so implementing them in separate steps will help with debugging and review :))

This PR targets step 1. 

---

Resolves #

PR #1596, opened 2025-07-28
Implement cache invalidation for audits

phil9909:
Implement cache invalidation for audits

See #1550
See #1528 

This change is a bit more involved than the previous PRs for cache invalidation (e.g. #1551, #1584), because the storage key does not contain an identifier for the audit, instead all audits are stored as an array in a single key (actually it's two keys, `core/audit` and `core/local-audit`) so if we are notified about a change we only know something changed, but not what. So I added a `reconcileAudits` method (naming inspired by [Kubernetes](https://book.kubebuilder.io/getting-started#reconciliation-process)) which will figure out the difference between the "desired" mount table and the "actual" mount table and add or remove audit backends as needed.

Also, I modified `setupAudits` to use `reconcileAudits` to limit duplication.

### DIFF
--- a/vault/audit.go
+++ b/vault/audit.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	uuid "github.com/hashicorp/go-uuid"
 	"github.com/openbao/openbao/audit"
 	"github.com/openbao/openbao/helper/namespace"
@@ -220,7 +221,7 @@ func (c *Core) disableAudit(ctx context.Context, path string, updateStorage bool
 }
 
 // loadAudits is invoked as part of postUnseal to load the audit table
-func (c *Core) loadAudits(ctx context.Context) error {
+func (c *Core) loadAudits(ctx context.Context, readonly bool) error {
 	auditTable := &MountTable{}
 	localAuditTable := &MountTable{}
 
@@ -235,9 +236,6 @@ func (c *Core) loadAudits(ctx context.Context) error {
 		c.logger.Error("failed to read local audit table", "error", err)
 		return errLoadAuditFailed
 	}
-
-	c.auditLock.Lock()
-	defer c.auditLock.Unlock()
 
 	if raw != nil {
 		if err := jsonutil.DecodeJSON(raw.Value, auditTable); err != nil {
@@ -300,6 +298,11 @@ func (c *Core) loadAudits(ctx context.Context) error {
 	}
 
 	if !needPersist {
+		return nil
+	}
+
+	if readonly {
+		c.logger.Warn("audit table needs update")
 		return nil
 	}
 
@@ -383,15 +386,64 @@ func (c *Core) persistAudit(ctx context.Context, table *MountTable, localOnly bo
 func (c *Core) setupAudits(ctx context.Context) error {
 	brokerLogger := c.baseLogger.Named("audit")
 	c.AddLogger(brokerLogger)
-	broker := NewAuditBroker(brokerLogger)
+	c.auditBroker = NewAuditBroker(brokerLogger)
 
+	err := c.reconcileAudits(reconcileAuditsRequests{
+		ctx:       ctx,
+		readonly:  false,
+		isInitial: true,
+	})
+	if err != nil {
+		if multiErr, ok := err.(*multierror.Error); ok {
+			for _, err := range multiErr.Errors {
+				c.logger.Error(err.Error())
+			}
+		} else {
+			return err
+		}
+	}
+
+	if len(c.audit.Entries) > 0 && c.auditBroker.Count() == 0 {
+		return errLoadAuditFailed
+	}
+
+	return nil
+}
+
+type reconcileAuditsRequests struct {
+	ctx       context.Context
+	readonly  bool
+	isInitial bool
+}
+
+func (c *Core) reconcileAudits(req reconcileAuditsRequests) error {
 	c.auditLock.Lock()
 	defer c.auditLock.Unlock()
 
-	var successCount int
+	var oldTable *MountTable
+	if c.audit != nil {
+		oldTable = c.audit.shallowClone()
+	}
 
-	for _, entry := range c.audit.Entries {
-		// Create a barrier view using the UUID
+	if err := c.loadAudits(req.ctx, req.readonly); err != nil {
+		c.audit = oldTable
+		return err
+	}
+
+	additions, deletions := oldTable.delta(c.audit)
+
+	var multiErr *multierror.Error
+
+	for _, entry := range deletions {
+		c.removeAuditReloadFunc(entry)
+
+		c.auditBroker.Deregister(entry.Path)
+		if c.logger.IsInfo() {
+			c.logger.Info("disabled audit backend", "path", entry.Path)
+		}
+	}
+
+	for _, entry := range additions {
 		view, err := c.mountEntryView(entry)
 		if err != nil {
 			return err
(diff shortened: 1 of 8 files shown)


vault/core_cache_invalidate.go
```diff
@@ -69,7 +69,17 @@ func (c *Core) invalidateInternal(ctx context.Context, key string) error {
 		c.quotaManager.Invalidate(strings.TrimPrefix(key, systemBarrierPrefix+quotas.StoragePrefix))
 
 	case c.router.Invalidate(ctx, key):
-		// if router.Invalidate returns true, a matching plugin was found and the invalidation is therefore dispatched
+	// if router.Invalidate returns true, a matching plugin was found and the invalidation is therefore dispatched
+
+	case key == coreAuditConfigPath || key == coreLocalAuditConfigPath:
+		err := c.reconcileAudits(reconcileAuditsRequests{
```
cipherboy: I don't think this is the right approach because it reconciles the audit inline with the storage. I would mark the table as invalid and reload it asynchronously rather than blocking the invalidation on audit reconciliation.

vault/audit.go
```diff
@@ -383,15 +386,64 @@ func (c *Core) persistAudit(ctx context.Context, table *MountTable, localOnly bo
...
+	err := c.reconcileAudits(reconcileAuditsRequests{
+		ctx:       ctx,
+		readonly:  false,
+		isInitial: true,
+	})
+	if err != nil {
+		if multiErr, ok := err.(*multierror.Error); ok {
+			for _, err := range multiErr.Errors {
+				c.logger.Error(err.Error())
+			}
```
cipherboy, line 400: Missing return?

vault/audit.go
```diff
@@ -220,7 +221,7 @@ func (c *Core) disableAudit(ctx context.Context, path string, updateStorage bool
 }
 
 // loadAudits is invoked as part of postUnseal to load the audit table
```
cipherboy: Comment should be adjusted at a minimum, to also clarify why the lock no longer needs to be held. 

(It isn't called directly from `postUnseal` but as a part of `setupMounts` which holds the lock.)

vault/audit.go
```diff
@@ -303,6 +301,11 @@ func (c *Core) loadAudits(ctx context.Context) error {
 		return nil
 	}
 
+	if readonly {
```
cipherboy, line 304: I wonder if this should be returned rather than just warned? :thinking:

vault/audit.go
```diff
@@ -383,15 +386,64 @@ func (c *Core) persistAudit(ctx context.Context, table *MountTable, localOnly bo
...
+	}
+
+	if len(c.audit.Entries) > 0 && c.auditBroker.Count() == 0 {
+		return errLoadAuditFailed
+	}
+
+	return nil
+}
+
+type reconcileAuditsRequests struct {
```
cipherboy, line 413: I don't think we need this struct.

vault/audit.go
```diff
@@ -383,15 +386,64 @@ func (c *Core) persistAudit(ctx context.Context, table *MountTable, localOnly bo
 func (c *Core) setupAudits(ctx context.Context) error {
 	brokerLogger := c.baseLogger.Named("audit")
 	c.AddLogger(brokerLogger)
-	broker := NewAuditBroker(brokerLogger)
+	c.auditBroker = NewAuditBroker(brokerLogger)
```
cipherboy, line 389: Isn't this destructive? It seems like we'd want to keep the existing pattern of keeping `broker` as a local until we're entirely sure we're in a good shape to swap.

vault/audit.go
```diff
@@ -383,15 +386,64 @@ func (c *Core) persistAudit(ctx context.Context, table *MountTable, localOnly bo
...
 
-	var successCount int
+	var oldTable *MountTable
+	if c.audit != nil {
+		oldTable = c.audit.shallowClone()
+	}
 
-	for _, entry := range c.audit.Entries {
-		// Create a barrier view using the UUID
+	if err := c.loadAudits(req.ctx, req.readonly); err != nil {
```
cipherboy, line 428: Same with loading, tbh. If you want to move to this pattern, we should use it to make it safer and to keep around the existing table until we know the new one is good. Not sure how we'd handle failures here; I think our only real option is to seal given that this is in the audit subsystem.

vault/audit.go
phil9909: no, as before we want to only return an error, if non of the audit backends could be loaded (see if below).

vault/audit.go
phil9909: This should usually only happen during an update, right? And to keep the system available, you probably don't want to update all nodes at the same time. When the first node is updated, it will come up as a follower and the leader will still be on the older version. Once leadership is transferred to a updated node the table will be updated.

If we instead return an error here, the cluster will be degraded until the leader has been updated.

vault/audit.go
phil9909: Sure, we don't need it. I added it, because the function takes two boolean arguments and I wanted to assign proper names to them to prevent issues caused by mixing up the order. I can remove it, if you prefer.

vault/audit.go
phil9909: Yes and no. While this would destroy a working audit broker, `setupAudits` is only called from within `unseal` and `preSeal` calls `teardownAudits` which does `c.auditBroker = nil`. So as long as you don't call `unseal` twice without re-sealing in between (which I guess causes all sorts of problems)  `auditBroker` should be `nil` anyway.

vault/audit.go
phil9909: In the line below (in case of an error) the table is restored. This all happens while holding a lock (this is why I moved the lock out of `loadAudits` to make the whole process atomic).

> I think our only real option is to seal given that this is in the audit subsystem.

The error will bubble up to either `unseal` which would fail or to `Invalidate` which will in turn re-seal and try again (not yet implemented).

**Update:** Well this is no longer true, if we make reloading the audit table asynchronous.

vault/audit.go
cipherboy: @phil9909 I think this might be where audits and other mounts differ. This code path is also used there for adding required system mounts, which there aren't any for audit backends.

I wasn't thinking of returning it as an error though, but instead point out that we need to flush the updated table later (after doing the reconciliation when we can update the full table). Note that standby nodes can't update the table too.

The other reason is (and I could be wrong, but from memory) there are some places where mounts are partially added and this logic helps finish the write. But this may only be for auth/secrets and not audit.

Strictly, I don't think we have that much control over updates. I think our only option is to update followers first to non-voting and then upgrade the leader as active before coming back to mark all followers as voting again. Otherwise you risk an active->updated transfer before you've completed all standby upgrades and risk having an issue mean you go upgraded->downgraded (which isn't allowed).

I think we should probably file an unrelated issue about improving upgrade paths, tbh.

vault/audit.go
cipherboy: @phil9909 I'm not sure sealing on invalidation failure is the right approach, but yeah. Keep in mind, invalidation is just a router, it isn't performing any actual invalidations.

phil9909: > I'm not sure sealing on invalidation failure is the right approach

I depends. For "important" things working on a stale cache can be dangerous (i.e. allowing access for leaked token that was revoked). If this fails using the sledgehammer approach of sealing and unsealing again seems justified. For plugins just unloading the plugin and loading it again should be fine.

> Keep in mind, invalidation is just a router, it isn't performing any actual invalidations.

You're saying that the `core.Invalidate` method should not be the one that handles invalidation errors, but rather the individual subsystems should decide what to do?

vault/audit.go
phil9909: > I wasn't thinking of returning it as an error though, but instead point out that we need to flush the updated table later (after doing the reconciliation when we can update the full table). Note that standby nodes can't update the table too.

I'm not sure if I understood that correctly. `readonly` should be `true` only in the invalidate code path, which should only be called on follower nodes. So I don't see the problem, the leader will take care of upgrading the table.
When thinking about this, we probably also need to set `readonly = true` in case of unsealing follower nodes. During normal operations any pending writes should have been handled by the leader already, but in cases when the follower runs a more recent version, it will run into problems (cc @fatima2003).

> I think we should probably file an unrelated issue about improving upgrade paths, tbh.

Yes, documentation could be improved there to include the "make as non voter", "update followers first" stuff.

Also, if we reconsider the "OpenBao upgrades are designed such that large jumps (ie 1.3.10 -> 1.7.x) are supported" part and only support `2.n.x` to `2.(n+1).y` updates, we can defer updating the table by one minor version.
E.g.:
Up to 2.5.x only the old format is supported
In version 2.6.0 support for the new format is introduced, but it will still use the old format for storage.
In version 2.7.0 the storage format will switch to the new version.

Same for any protocols, there has to be one version in between, that as a server already understands new request fields, but as a client will never use them.
If we do this, nodes can be upgraded in any order.

vault/audit.go
cipherboy: > You're saying that the `core.Invalidate` method should not be the one that handles invalidation errors, but rather the individual subsystems should decide what to do?

Exactly. In _most_ instances, I'd expect invalidate to be an atomic boolean/locked list and on next subsystem access/request/... we'd trigger a reload of the affected component(s). If that component is internal to the core and highly critical, we'd make all such loads trigger re-sealing.

(This is mostly because it is such a performance critical subsystem; we can't afford to stall Raft/... on invalidations taking long amounts of time.)

My 2c.

phil9909: Agreed, that's why I handled invalidation asynchronously right from within the FSM: https://github.com/openbao/openbao/blob/8b441003d6c8fcbc8391fdfecb05b388a943e002/physical/raft/fsm.go#L925-L939

vault/audit.go
cipherboy: I see, I think this is fine then.

vault/audit.go
cipherboy: Ahh, I see. Yes, gotcha now.

vault/core_cache_invalidate.go
phil9909: I changed it to be processed in a new go routine, but keep in mind that cache invalidation is already async (see https://github.com/openbao/openbao/pull/1596#discussion_r2247406552)

vault/audit.go
phil9909: For now all errors are handled in the Invalidate function, by simply logging them. Once https://github.com/openbao/openbao/pull/1592 is merged, I will refactor this accordingly (I don't want to implement the seal-unseal logic before we even have consensus on how standby nodes are unsealed).

vault/audit.go
phil9909: Do you want me to remove it?

vault/audit.go
cipherboy: > (I don't want to implement the seal-unseal logic before we even have consensus on how standby nodes are unsealed).

Sorry, what portion of consensus are we missing?

phil9909: Well, it's not merged :laughing: 
To me consensus is reached, once a PR is merged.

cipherboy: @phil9909 Ah my bad, maybe there was a miscommunication with @fatima2003. I was under the impression she was going to look into the panic before we merge? But otherwise I think we were both happy with the PR? 

If we want to leave that to a follow-up too, I'm happy to.

phil9909: I also think that there will be no more major changes, but I was burned by this before (in other Open Source projects).

> If we want to leave that to a follow-up too, I'm happy to.

If there is nothing else that you want to have changed, I would appreciate, if we could get this one merged.
I will immediately start working on refactoring the error handling for the whole `core.Invalidate`

PR #1597, opened 2025-07-28
goreleaser/hsm: Fix Docker manifests

satoqz:
Add some tags that I overlooked in #1427. For example, https://github.com/openbao/openbao/pkgs/container/openbao-hsm is entirely missing a `latest` tag. We have a total of 24 `docker_manifests` in `goreleaser.hsm.yaml` now, which is on par with the amount in `goreleaser.linux.yaml`.

### DIFF
--- a/goreleaser.hsm.yaml
+++ b/goreleaser.hsm.yaml
@@ -261,54 +261,126 @@ dockers:
       - ./CHANGELOG.md
 
 docker_manifests:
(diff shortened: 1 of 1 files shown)


wslabosz-reply (approved):

cipherboy (approved):

satoqz:
Thanks everyone, merging...

PR #1599, opened 2025-07-29
Sync per-ns sealing branch with post-poc changes

wslabosz-reply:
Implements #1170 

List of changes:
- Added request rejection for the sealed namespaces
- Implemented namespace unsealing
- Implemented loading of sealed namespaces after unsealing
- Implemented unmounting mounts on seal of the namespace
- Implemented `namespaces/<name>/seal-status` endpoint
- Implemented `namespaces/<name>/key-status` endpoint
- Implemented `namespaces/<name>/rotate` endpoint
- Implemented `namespaces/<name>/rotate/config` endpoint
- Implemented generation of root tokens for namespaces with CLI commands
- Added mount loading skip of sealed namespaces
- Added prevention of namespace creation in case of a bad seal config passed
- Added mounting of mounts after unsealing the namespace


cipherboy:
@wslabosz-reply Does this need to be rebased? The first commit is "add seal manager" which is already in https://github.com/openbao/openbao/commits/namespaces-seal/, just under a different name.

wslabosz-reply:
@cipherboy 
Will work something out, splitting it into smaller pieces.
EDIT: this does not make sense as smaller pieces are not passing the CI without heavy edit, sorry about that...

helper/namespace/namespace.go
```diff
@@ -30,6 +30,9 @@ var reservedNames = []string{
 	"auth",
 	"cubbyhole",
 	"identity",
+	"seal",
+	"unseal",
+	"seal-status",
```
satoqz: Is this really needed? I believe you should be able to call `/sys/namespaces/seal/seal` just fine, if you dare.

satoqz:
FYI I've started reviewing this, but might take quite a while 😄 Maybe I'll submit partial reviews when it makes sense, we'll see.

command/namespace_generate_root.go
```diff
@@ -0,0 +1,521 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
```
satoqz: Should include OpenBao contributors in the copyright header

vault/seal_manager.go
satoqz, line 1: Missing copyright header

vault/logical_system_namespaces_generate_root.go
satoqz, line 1: Missing copyright header

command/namespace_generate_root.go
```diff
@@ -0,0 +1,521 @@
...
+	flagGenerateOTP bool
+}
+
+func (c *NamespaceGenerateRootCommand) Synopsis() string {
+	return "Generate a new root token for a sealed namespace"
+}
+
+func (c *NamespaceGenerateRootCommand) Help() string {
+	helpText := `
+Usage: bao namespace generate-root [options] <NamespacePath>
```
satoqz:
```suggestion
Usage: bao namespace generate-root [options] PATH
```

to be consistent with other namespace commands.

command/namespace_generate_root.go
```diff
@@ -0,0 +1,521 @@
...
+      $ bao namespace generate-root -generate-otp
+
+  Start a root token generation:
+
+      $ bao namespace generate-root -init -otp="..." <NamespacePath>
+      $ bao namespace generate-root -init -pgp-key="..." <NamespacePath>
+
+  Enter an unseal key to progress root token generation:
+
+      $ bao namespace generate-root -otp="..." <NamespacePath>
```
satoqz: Same as above.

command/namespace_generate_root.go
```diff
@@ -0,0 +1,521 @@
...
+	"github.com/openbao/openbao/sdk/v2/helper/roottoken"
+	"github.com/posener/complete"
+)
+
+var (
+	_ cli.Command             = (*NamespaceGenerateRootCommand)(nil)
+	_ cli.CommandAutocomplete = (*NamespaceGenerateRootCommand)(nil)
+)
+
+type NamespaceGenerateRootCommand struct {
```
satoqz, line 26: Should add this command to the help text in `command/namespace.go`.

vault/barrier.go
```diff
@@ -15,15 +15,15 @@ import (
 var (
 	// ErrBarrierSealed is returned if an operation is performed on
 	// a sealed barrier. No operation is expected to succeed before unsealing
-	ErrBarrierSealed = errors.New("Vault is sealed")
+	ErrBarrierSealed = errors.New("Barrier is sealed")
```
cipherboy: I know of a few places that rely on these exact error strings so I think you're going to have to revert the values.

At that point, you can either fork it (for namespaces) or reuse the slightly incorrect error message here.

I think I had that discussion on an earlier PR but not sure I remember which one...

command/namespace_generate_root.go
```diff
@@ -0,0 +1,521 @@
...
+	flagCancel      bool
+	flagNonce       string
+	flagStatus      bool
+	flagDecode      string
+	flagOTP         string
+	flagGenerateOTP bool
+}
+
+func (c *NamespaceGenerateRootCommand) Synopsis() string {
+	return "Generate a new root token for a sealed namespace"
```
satoqz: Is it "sealed" or more like "sealable"?

command/namespace_generate_root.go
```diff
@@ -0,0 +1,521 @@
...
+// init is used to start the generation process
+func (c *NamespaceGenerateRootCommand) init(client *api.Client, otp, pgpKey string, namespacePath string) int {
+	// Validate incoming fields. Either OTP OR PGP keys must be supplied.
+	if otp != "" && pgpKey != "" {
+		c.UI.Error("Error initializing: cannot specify both -otp and -pgp-key")
+		return 1
+	}
+
+	// Start the root generation
+	secret, err := client.Logical().Write("sys/namespaces/"+namespacePath+"/generate-root/attempt", map[string]interface{}{})
```
satoqz:
```suggestion
	secret, err := client.Logical().Write("sys/namespaces/"+namespacePath+"/generate-root/attempt", nil)
```

command/namespace_generate_root.go
```diff
@@ -0,0 +1,521 @@
...
+	Started          bool   `json:"started"`
+	Progress         int    `json:"progress"`
+	Required         int    `json:"required"`
+	Complete         bool   `json:"complete"`
+	EncodedToken     string `json:"encoded_token"`
+	EncodedRootToken string `json:"encoded_root_token"`
+	PGPFingerprint   string `json:"pgp_fingerprint"`
+	OTP              string `json:"otp"`
+	OTPLength        int    `json:"otp_length"`
+}
```
satoqz: Isn't this the same as `logical.GenerateRootStatusResponse`?

sdk/helper/consts/error.go
```diff
@@ -10,6 +10,10 @@ var (
 	// No operation is expected to succeed before unsealing
 	ErrSealed = errors.New("Vault is sealed")
 
+	// ErrNamespaceSealed is returned if an operation is performed on a sealed namesapce barrier.
```
satoqz:
```suggestion
	// ErrNamespaceSealed is returned if an operation is performed on a sealed namespace barrier.
```

vault/acl.go
```diff
@@ -35,7 +35,7 @@ type ACL struct {
 	segmentWildcardPaths map[string]interface{}
 
 	// root is enabled if the "root" named policy is present.
```
satoqz:
```suggestion
	// root is non-nil if the "root" named policy is present.
```

vault/acl.go
```diff
@@ -345,15 +341,22 @@ func (a *ACL) Capabilities(ctx context.Context, path string) (pathCapabilities [
...
 	ret = new(ACLResults)
 
+	ns, err := namespace.FromContext(ctx)
+	if err != nil {
+		return
+	}
 	// Fast-path root
-	if a.root {
+	if a.root != nil {
+		if !ns.HasParent(a.root) {
```
cipherboy, line 354: This warrants a comment; we earlier enforce `root` is an exclusive policy and so there can by definition be no other policies.

vault/acl_test.go
```diff
@@ -182,6 +182,72 @@ func testACLRoot(t *testing.T, ns *namespace.Namespace) {
...
+	if !authResults.RootPrivs {
+		t.Fatal("expected root")
+	}
+	if !authResults.Allowed {
+		t.Fatal("expected permissions")
+	}
+}
+
+func TestACL_Namespace_Root_Fail(t *testing.T) {
+	t.Run("root-ns", func(t *testing.T) {
```
cipherboy, line 220: For simple tests you probably don't need `t.Run(...)` and splitting the core into a different method, but if you're going to test more complex scenarios, I could see it being useful. 

(fine leaving it though)

My 2c.

vault/core.go
```diff
@@ -1407,6 +1413,11 @@ func (c *Core) Sealed() bool {
 	return atomic.LoadUint32(c.sealed) == 1
 }
 
+// IsNSSealed checks if any namespace in namespace hierarchy is current sealed.
```
cipherboy: Is this right? Or is it the closest namespaces with per-namespace sealing that it checks if it is sealed?

I think we earlier said that in a hierarchy `ns1/ns2/ns3`; sealing ns1 results in sealing ns2/ns3 which requires manual intervention (recovery keys, restart)?

vault/token_store.go
```diff
@@ -979,16 +983,21 @@ func (ts *TokenStore) SaltID(ctx context.Context, id string) (string, error) {
...
+		Path:         view.Prefix() + "auth/token/root",
 		CreationTime: time.Now().Unix(),
-		NamespaceID:  namespace.RootNamespaceID,
+		NamespaceID:  ns.ID,
 		Type:         logical.TokenTypeService,
 	}
+	if ns.UUID == namespace.RootNamespaceUUID {
+		te.DisplayName = "root"
+	} else {
+		te.DisplayName = ns.ID + "_root"
```
cipherboy: You could probably just default to `DisplayName: "root"` above and go: `if ns.UUID != namespace.RootNamespaceUUID { te.DisplayName = ns.ID + "_" + te.DisplayName }` and that way you can modify the name in one place if we ever want to later.

cipherboy (commented):
I have a number of large files still left to look at, but these are some easy pieces right now.

command/namespace_generate_root.go
```diff
@@ -0,0 +1,521 @@
...
+	key = strings.TrimSpace(key)
+
+	// Verify we have a nonce value
+	if nonce == "" {
+		c.UI.Error("Missing nonce value: specify it via the -nonce flag")
+		return 1
+	}
+
+	// Provide the key, this may potentially complete the update
+	data := map[string]interface{}{"key": key, "nonce": nonce}
```
satoqz, line 339: This could reuse `GenerateRootUpdateRequest` from `http/sys_generate_root.go` if you also move that to `sdk/logical` like you did with `GenerateRootStatusResponse`.

Then again, not sure if those types should live in `sdk` in the first place, but certainly in some shared package.

vault/token_store.go
```diff
@@ -71,6 +71,10 @@ const (
 	// any namespace information
 	TokenLength = 24
 
+	// NSTokenLength is the size of tokens we are currently generating, for
+	// sealed Namespaces
+	NSTokenLength = 31
```
satoqz: If I understand correctly, this could be:

```suggestion
	NSTokenLength = TokenLength + MaxNSIdLength + 1
```

or

```suggestion
	NSTokenLength = TokenLength + namespaceIdLength + 1
```

vault/token_store.go
```diff
@@ -71,6 +71,10 @@ const (
 	// any namespace information
 	TokenLength = 24
 
+	// NSTokenLength is the size of tokens we are currently generating, for
+	// sealed Namespaces
```
satoqz:
```suggestion
	// sealable namespaces
```

?

vault/token_store.go
```diff
@@ -71,6 +71,10 @@ const (
...
+	NSTokenLength = 31
+
 	// MaxNsIdLength is the maximum namespace ID length (5 characters prepended by a ".")
 	MaxNsIdLength = 6
```
satoqz: Tangential, this seems outdated

vault/token_store.go
```diff
@@ -979,16 +983,21 @@ func (ts *TokenStore) SaltID(ctx context.Context, id string) (string, error) {
 }
 
 // rootToken is used to generate a new token with root privileges and no parent
-func (ts *TokenStore) rootToken(ctx context.Context) (*logical.TokenEntry, error) {
-	ctx = namespace.ContextWithNamespace(ctx, namespace.RootNamespace)
+func (ts *TokenStore) rootToken(ctx context.Context, ns *namespace.Namespace) (*logical.TokenEntry, error) {
+	ctx = namespace.ContextWithNamespace(ctx, ns)
```
satoqz: I think you can just keep this at `rootToken(ctx context.Context)` and take the namespace from `ctx`.

vault/namespace_store.go
```diff
@@ -614,8 +619,9 @@ func (ns *NamespaceStore) ModifyNamespaceByPath(ctx context.Context, path string
 }
 
 // ListAllNamespaces lists all available namespaces, optionally including the
-// root namespace.
-func (ns *NamespaceStore) ListAllNamespaces(ctx context.Context, includeRoot bool) ([]*namespace.Namespace, error) {
+// root namespace. skipSealed flag dictates whether sealed namespaces are
+// filtered out from the result slice.
+func (ns *NamespaceStore) ListAllNamespaces(ctx context.Context, includeRoot, skipSealed bool) ([]*namespace.Namespace, error) {
```
satoqz: Having one parameter where `true = more` and another where `false = more` is not optimal and harder to reason about for the caller.

```suggestion
// root namespace and sealed namespaces.
func (ns *NamespaceStore) ListAllNamespaces(ctx context.Context, includeRoot, includeSealed bool) ([]*namespace.Namespace, error) {
```

vault/namespace_store.go
```diff
@@ -430,22 +438,19 @@ func (ns *NamespaceStore) initializeNamespace(ctx context.Context, entry *namesp
...
+	return ns.createMounts(nsCtx)
 }
 
 // initializeNamespacePolicies loads the default policies for the namespace store.
-func (ns *NamespaceStore) initializeNamespacePolicies(ctx context.Context) error {
+func (ns *NamespaceStore) initializeNamespacePolicies(ctx context.Context, namespace *namespace.Namespace) error {
 	if err := ns.core.policyStore.loadDefaultPolicies(ctx); err != nil {
 		return fmt.Errorf("error creating default policies: %w", err)
 	}
+	ns.core.policyStore.policyTypeMap.Store(ns.core.policyStore.cacheKey(namespace, "root"), PolicyTypeACL)
```
satoqz: Hmm, this touches the policy store internals a bit too much for my taste. I think with some other minor changes in the policy store setup this step could be part of `loadDefaultPolicies`.

vault/namespace_store.go
```diff
@@ -871,22 +906,65 @@ func (ns *NamespaceStore) UnsealNamespace(ctx context.Context, path string, key
...
+	if ns.core.IsNSSealed(namespaceToUnseal) {
+		return nil
+	}
+
+	// Then proceed to load mounts and credentials
+	if err := ns.core.loadMountsForNamespace(ctx, namespaceToUnseal); err != nil {
+		return err
+	}
+
+	postUnsealFuncs := make([]func(), 0)
```
satoqz: Is there a reason this didn't work?

```suggestion
	var postUnsealFuncs []func()
```

vault/namespace_store.go
```diff
@@ -871,22 +906,65 @@ func (ns *NamespaceStore) UnsealNamespace(ctx context.Context, path string, key
...
 	if namespaceToUnseal.ID == namespace.RootNamespaceID {
 		return errors.New("unable to unseal root namespace")
 	}
 
-	return ns.core.sealManager.UnsealNamespace(ctx, namespaceToUnseal.Path, key)
+	if err := ns.core.sealManager.UnsealNamespace(ctx, namespaceToUnseal, key); err != nil {
+		return err
+	}
+	releaseNsLock = false
+	unlock()
```
satoqz: Maybe it's possible to split this function into two helpers, one that does the unsealing and uses the lock, and one that does the post-unseal loading.

satoqz (commented):
14/39 files viewed, this covers changes to:

- CLI
- Token Store
- Namespace Store

vault/barrier.go
wslabosz-reply: going to revert the change, although it hurts me as even the comments explicitly say "barrier", not "Vault", and the error message being uppercased 😢

vault/acl.go
wslabosz-reply: will add a comment, please check if it's correct

vault/acl_test.go
wslabosz-reply: let's leave it as it is for now

vault/core.go
wslabosz-reply:
```
// IsNSSealed checks if there's a namespace in upwards namespace
// hierarchy that is currently sealed.
```
does it make more sense?

> Or is it the closest namespaces with per-namespace sealing that it checks if it is sealed?

we omit namespaces that are not sealable by default (they do not have their own barrier).

> I think we earlier said that in a hierarchy ns1/ns2/ns3; sealing ns1 results in sealing ns2/ns3 which requires manual intervention (recovery keys, restart)?

yes, this is still true.

vault/token_store.go
wslabosz-reply: agreed.

helper/namespace/namespace.go
wslabosz-reply: it might be a leftover of some other endpoint format that I've used before, will adjust

command/namespace_generate_root.go
wslabosz-reply: sealable imo

command/namespace_generate_root.go
wslabosz-reply: it is

command/namespace_generate_root.go
wslabosz-reply: so for the respectively mirrored `operator_generate_root.go` it is done using functions on the api client, which I'd consider also replicating, but not sure if I want to do that in the scope of this PR.

vault/token_store.go
wslabosz-reply: yes

vault/token_store.go
wslabosz-reply: yes, but actually no.

vault/token_store.go
wslabosz-reply: 😢

vault/token_store.go
wslabosz-reply: yeah, not only there, but will adjust.

vault/namespace_store.go
wslabosz-reply: not sure, will adjust

vault/namespace_store.go
wslabosz-reply: will adjust

vault/barrier.go
cipherboy: @wslabosz-reply Yeah, me too.

Aha! https://github.com/openbao/openbao/pull/1188#issuecomment-2779189730

vault/core.go
cipherboy: Oohh, am I misunderstanding what `IsSealed(...)` means in this context? I was expecting this to return the first namespace which has a barrier which is marked as sealed (not accessible for read/writes, due to missing/withheld encryption keys).

Not, find me the first namespace which has a barrier?

wslabosz-reply: > I was expecting this to return the first namespace which has a barrier which is marked as sealed (not accessible for read/writes, due to missing/withheld encryption keys).

I mean, that is right, apart from that it does not return the namespace/barrier, but the boolean flag (basically using the `barrier.Sealed()` method underhood)

wslabosz-reply: okay, I think I understand the confusion, yes this method is redundant as we should always propagate the seal status, so if a ns is unsealed, then you know that every namespace in the "upward" hierarchy has to be unsealed, and if ns is sealed then it cannot be unsealed before all it's parents are unsealed (in sequential order).

wslabosz-reply: okay, to go even deeper into this, I think the reason we are using the `sm.barrierByNamespace.LongestPrefix` here is that the namespace we are currently requesting a resource from doesn't have to be sealable. Meaning that it won't be sealed when its parent is being sealed. This prompts us to answer a question whether we should reject any requests coming to (technically unsealed) namespaces which do have a sealed namespace as their ancestor.

command/namespace_generate_root.go
satoqz: Makes sense, let's do that another time

vault/core.go
cipherboy: Right, so I think we agree the comment is incorrect? :-) 

It is still not **any** namespace, but the closest encrypted namespace is sealed?

vault/namespace_store.go
cipherboy: Note that I just removed this map in #1613, so we'll eventually need to rebase the `namespace-seal` branch on top of main and reconcile the changes.

vault/generate_root.go
```diff
@@ -110,30 +127,42 @@ func (c *Core) GenerateRootConfiguration() (*GenerateRootConfig, error) {
...
-	// Copy the config if any
-	var conf *GenerateRootConfig
-	if c.generateRootConfig != nil {
-		conf = new(GenerateRootConfig)
-		*conf = *c.generateRootConfig
-		conf.OTP = ""
-		conf.Strategy = nil
+	namespaceRootGen, ok := c.namespaceRootGens[ns.UUID]
+	if !ok {
+		return nil, nil
```
wslabosz-reply: don't know, should we return an error here?

satoqz: I think it's better than not to

wslabosz-reply: this will introduce changes to existing behavior of root generation, but I think we are okay with that

sdk/logical/response.go
```diff
@@ -333,3 +333,16 @@ func ResolveRoleResponse(roleName string) (*Response, error) {
 		},
 	}, nil
 }
+
+type GenerateRootStatusResponse struct {
```
satoqz: Still wondering if this shouldn't be in `api/` instead.

vault/acl.go
```diff
@@ -102,16 +101,13 @@ func NewACL(ctx context.Context, policies []*Policy) (*ACL, error) {
 			return nil, errors.New("unable to parse policy (wrong type)")
 		}
 
-		// Check if this is root
+		// Check if this policy name is 'root'
```
satoqz: Nit: Bit of an unnecessary comment now, I think the one further below suffices.

vault/request_handling.go
```diff
@@ -606,9 +606,15 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
+		}
+
+		isRestrictedSysAPI := ns.ID != namespace.RootNamespaceID &&
+			strings.HasPrefix(req.Path, "sys/") &&
+			restrictedSysAPIs.HasPathSegments(req.Path[len("sys/"):])
 
-		if strings.HasPrefix(req.Path, "sys/") &&
-			restrictedSysAPIs.HasPathSegments(req.Path[len("sys/"):]) {
+		if isRestrictedSysAPI {
 			return nil, logical.CodedError(http.StatusBadRequest, "operation unavailable in namespaces")
```
```suggestion
			return nil, logical.CodedError(http.StatusBadRequest, "operation only available in root namespace")
```

I know that's probably my bad, but since we're already here

vault/logical_system_namespaces_test.go
```diff
@@ -17,23 +17,30 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func testCreateNamespace(t *testing.T, ctx context.Context, b logical.Backend, name string, customMetadata map[string]string) *namespace.Namespace {
+// testCreateNamespace returns details of created namespace, and keyshares of the namespace seal
+func testCreateNamespace(t *testing.T, ctx context.Context, b logical.Backend, name string, reqBody map[string]interface{}) (*namespace.Namespace, map[string][]string) {
```
satoqz: I don't see a single case where the key shares return value is used

vault/core.go
```diff
@@ -811,6 +810,14 @@ type CoreConfig struct {
 	UnsafeCrossNamespaceIdentity bool
 }
 
+// NamespaceRootGeneration manages the configuration and progress of an ongoing
+// root token generation process for a namespace.
+type NamespaceRootGeneration struct {
+	Config   *GenerateRootConfig
+	Progress [][]byte
+	Lock     sync.Mutex
```
satoqz: I think you don't need this lock and the global one on core is more than enough?

vault/core.go
```diff
@@ -1413,11 +1419,18 @@ func (c *Core) GetContext() (context.Context, context.CancelFunc) {
...
-// Sealed checks if the Vault is current sealed
+// Sealed checks if the Vault is currently sealed
 func (c *Core) Sealed() bool {
 	return atomic.LoadUint32(c.sealed) == 1
 }
 
+// IsNSSealed checks if there's a namespace in upwards namespace
+// hierarchy that is currently sealed, which should only happen
+// if the current namespace is not sealable, but its ancestors are.
+func (c *Core) IsNSSealed(ns *namespace.Namespace) bool {
```
```suggestion
func (c *Core) NamespaceSealed(ns *namespace.Namespace) bool {
```

We have enough space to spell it out, and omitting `Is` is more in line with existing names.

vault/core.go
```diff
@@ -1413,11 +1419,18 @@ func (c *Core) GetContext() (context.Context, context.CancelFunc) {
...
 
-// Sealed checks if the Vault is current sealed
+// Sealed checks if the Vault is currently sealed
 func (c *Core) Sealed() bool {
 	return atomic.LoadUint32(c.sealed) == 1
 }
 
+// IsNSSealed checks if there's a namespace in upwards namespace
+// hierarchy that is currently sealed, which should only happen
+// if the current namespace is not sealable, but its ancestors are.
```
satoqz: I think in this case it's actually "sealed"? Otherwise I don't fully understand the remark.

vault/logical_system.go
```diff
@@ -5854,14 +5856,27 @@ This path responds to the following HTTP methods.
...
+This path responds to the following HTTP methods.
 
 	POST /<name>/seal
 		Seal a namespace.
 
 	POST /<name>/unseal
 		Unseal a namespace.
+
+    GET /<name>/seal-status
+        Returns the seal status of the namespace.
```
satoqz: Indent wrong

vault/logical_system.go
```diff
@@ -5854,14 +5856,27 @@ This path responds to the following HTTP methods.
 
 	DELETE /<name>
 		Delete a namespace.
+		`,
+	},
+	"namespaces-seal": {
```
satoqz: Since we have specific files for these paths (`logical_system_namespaces.go`, `logical_system_namespaces_seals.go` etc.) I think it's worth putting the help strings in a separate map that lives in those files. See https://github.com/openbao/openbao/blob/50a1260cd0eae7c03ef0972ac2f93cc75d72a409/vault/logical_system_external_keys.go#L264 for inspiration.

vault/auth.go
```diff
@@ -670,6 +675,110 @@ func (c *Core) loadTransactionalCredentials(ctx context.Context, barrier logical
...
+	}
+	return c.loadTransactionalCredentialsForNamespace(ctx, ns)
+}
+
+// loadLegacyCredentialsForNamespace is used to load the auth mounts of a namespace on a legacy single-entry
+// format from a non-transactional storage to the mount table
+func (c *Core) loadLegacyCredentialsForNamespace(ctx context.Context, ns *namespace.Namespace) error {
+	if ns == nil {
+		return fmt.Errorf("namespace is nil")
+	}
```
satoqz: This is sus, do you expect that to ever happen from the caller's side?

vault/seal_manager.go
```diff
@@ -134,45 +167,281 @@ func (sm *SealManager) SealNamespace(ns *namespace.Namespace) error {
...
+	return sm.NamespaceBarrierByLongestPrefix(parentPath)
 }
 
-func (sm *SealManager) NamespaceBarrier(ns *namespace.Namespace) SecurityBarrier {
-	_, v, _ := sm.barrierByNamespace.LongestPrefix(ns.Path)
-	barrier := v.(SecurityBarrier)
+// NamespaceBarrierByLongestPrefix returns a barrier of a namespace matching
+// the longest prefix of the provided path, going up to root namespace.
+func (sm *SealManager) NamespaceBarrierByLongestPrefix(nsPath string) SecurityBarrier {
+	// this should acquire a lock
```
satoqz: Yes it should

vault/logical_system_namespaces_seals.go
```diff
@@ -0,0 +1,562 @@
...
+		err = b.Core.namespaceStore.UnsealNamespace(ctx, name, decodedKey)
+		if err != nil {
+			invalidKeyErr := &ErrInvalidKey{}
+			switch {
+			case errors.As(err, &invalidKeyErr):
+			case errors.Is(err, ErrBarrierInvalidKey):
+			case errors.Is(err, ErrBarrierNotInit):
+			case errors.Is(err, ErrBarrierSealed):
+			default:
+				return logical.RespondWithStatusCode(logical.ErrorResponse(err.Error()), req, http.StatusInternalServerError)
```
satoqz:
```suggestion
				return handleError(logical.CodedError(http.StatusInternalServerError, err.Error()))
```

vault/namespace_store.go
```diff
@@ -862,34 +887,77 @@ func (ns *NamespaceStore) SealNamespace(ctx context.Context, path string) error
...
 	if err != nil {
 		return err
 	}
+
+	// If namespace is still sealed meaning we do not have enough shards yet, return early
+	if ns.core.IsNSSealed(unsealedNamespace) {
+		return nil
+	}
+
+	return ns.postNamespaceUnseal(ctx, unsealedNamespace)
```
satoqz: What if the namespace is already unsealed? Currently this will happily continue and then fail at mounting cubbyhole twice.

vault/namespace_store.go
```diff
@@ -862,34 +887,77 @@ func (ns *NamespaceStore) SealNamespace(ctx context.Context, path string) error
 		return errors.New("unable to seal tainted namespace")
 	}
 
-	return ns.core.sealManager.SealNamespace(namespaceToSeal)
+	return ns.core.sealManager.SealNamespace(ctx, namespaceToSeal)
 }
 
 // UnsealNamespace unseals namespace with a given path, using provided key
 // TODO(wslabosz): track the unsealing progress in a SealManager
```
satoqz: Is this still TODO?

vault/seal_manager.go
```diff
@@ -116,14 +133,30 @@ func (sm *SealManager) StorageAccessForPath(path string) StorageAccess {
 }
 
 // SealNamespace seals the barriers of the given namespace and all of its children.
-func (sm *SealManager) SealNamespace(ns *namespace.Namespace) error {
+func (sm *SealManager) SealNamespace(ctx context.Context, ns *namespace.Namespace) error {
```
satoqz, line 149: This also needs a lock

vault/auth.go
```diff
@@ -670,6 +675,110 @@ func (c *Core) loadTransactionalCredentials(ctx context.Context, barrier logical
...
+
+// loadTransactionalCredentialsForNamespace is used to load the auth mounts of a namespace on a transactional
+// storage to the mount table
+func (c *Core) loadTransactionalCredentialsForNamespace(ctx context.Context, ns *namespace.Namespace) error {
+	if ns == nil {
+		return fmt.Errorf("namespace is nil")
+	}
+
+	if c.IsNSSealed(ns) {
+		return fmt.Errorf("namespace is sealed")
```
```suggestion
		return ErrNamespaceSealed
```

Probably in some other places too

vault/seal_manager.go
```diff
@@ -134,45 +167,281 @@ func (sm *SealManager) SealNamespace(ns *namespace.Namespace) error {
...
+
+// SecretProgress returns the number of keys provided so far. Lock
+// should only be false if the caller is already holding the read
+// statelock (such as calls originating from switchedLockHandleRequest).
+func (sm *SealManager) SecretProgress(ns *namespace.Namespace, lock bool) (int, string) {
+	if lock {
+		sm.lock.RLock()
+		defer sm.lock.RUnlock()
+	}
+	switch sm.unlockInformationByNamespace[ns.UUID]["default"] {
```
satoqz: I don't see this being safe when only core is state locked?

vault/seal_manager.go
```diff
@@ -134,45 +167,281 @@ func (sm *SealManager) SealNamespace(ns *namespace.Namespace) error {
...
+		if storedKeys == nil && err == nil && allowMissing {
+			return nil, nil
+		}
+
+		if err == nil && len(storedKeys) != 1 {
+			err = fmt.Errorf("expected exactly one stored key, got %d", len(storedKeys))
+		}
+		if err != nil {
+			return nil, fmt.Errorf("unable to retrieve stored keys: %w", err)
+		}
```
satoqz: Why not handle `err` right away 😭

satoqz (commented):
Next round, 31/38 viewed, mainly got all the small changes out of the way... lots more stuff to wrap my head around

vault/seal_manager.go
wslabosz-reply: true, probably was copied from other place, will look into it.

wslabosz-reply: collapsed `(*SealManager).unsealKeyToRootKey()` and `(*Core).unsealKeyToRootKey()` into one on `SealManager` struct.

vault/auth.go
wslabosz-reply: that's a sdk exported variable, should I use that?

vault/seal_manager.go
wslabosz-reply: I don't think I understand the question, the lock here is the one on `SealManager`

vault/namespace_store.go
wslabosz-reply: nope, done

vault/core.go
wslabosz-reply: sure, you are right.

vault/logical_system_namespaces_test.go
wslabosz-reply: I can either remove it now (and adjust multiple references), and add it later or leave it as it is, wdyt?

vault/seal_manager.go
satoqz: I'm referring to the doc comment of the method, which states that not locking the seal manager is fine if core is read-locked. I don't see how in that case reading from `unlockInformationByNamespace` is safe.

vault/auth.go
satoqz: Ah, yeah that one

vault/core.go
wslabosz-reply: `NamespaceSealed` return true if the current namespace is sealed or any of the ancestors is sealed:
```
ns1 -> ns2 -> ns3 -> ns4
// NamespaceSealed(ns4) -> if any ns from ns1, ns2, ns3 or ns4 is sealed then it will return false.
```

vault/logical_system.go
wslabosz-reply: I cannot find wrong indentation here

vault/logical_system.go
wslabosz-reply: sure

vault/auth.go
wslabosz-reply: no, it's not possible

vault/logical_system_namespaces_test.go
satoqz: Honestly not a big fan of this function in general, it should be replaced with `TestCoreCreateNamespaces` in most instances, but I won't ask you to do that here...

I would:
- Revert this one
- Use `TestCoreCreateSealedNamespaces()` in most places
- Do a manual `logical.TestRequest()` for testing the actual `key_shares` response

vault/namespace_store.go
wslabosz-reply: good spot.

vault/logical_system.go
satoqz: Set tabs to display as two spaces (the morally superior choice) and you'll see it 😄
This indent uses actual space characters, while it should use tabs.

vault/core.go
satoqz: > NamespaceSealed return true if the current namespace is sealed or any of the ancestors is sealed:

That makes perfect sense.

> `// NamespaceSealed(ns4) -> if any ns from ns1, ns2, ns3 or ns4 is sealed then it will return false.`

You mean `true`?

> `// which should only happen if the current namespace is not sealable, but its ancestors are.`

I still don't understand what this is trying to say then :D

vault/auth.go
wslabosz-reply: yeah, I don't think I should use that tbh, maybe I could introduce a separate error var like that, (like how `ErrBarrierSealed` is copied by sdk `ErrSealed`)

vault/seal_manager.go
wslabosz-reply: yea, that doesn't make sense.

vault/core.go
wslabosz-reply: yes, meant true*

yeah, I don't know where I was going with that comment...

vault/logical_system.go
wslabosz-reply: you really went here with `"hey this should be two tabs and not four spaces"` smh....

vault/logical_system_namespaces_test.go
wslabosz-reply: adjusted.

vault/logical_system_namespaces_generate_root.go
```diff
@@ -0,0 +1,288 @@
...
+		if ns == nil {
+			return nil, fmt.Errorf("namespace %q doesn't exist", name)
+		}
+
+		err = b.Core.GenerateRootCancel(ns)
+		if err != nil {
+			return handleError(err)
+		}
+
+		return &logical.Response{}, nil
```
satoqz:
```suggestion
		return nil, nil
```

Else this doesn't return `204`, but `200`, which isn't in line with how the endpoint behaves for the root namespace.

vault/namespace_store.go
```diff
@@ -862,34 +887,70 @@ func (ns *NamespaceStore) SealNamespace(ctx context.Context, path string) error
...
+		return nil
+	}
+
+	if err := ns.core.sealManager.UnsealNamespace(ctx, namespaceToUnseal, key); err != nil {
 		return err
 	}
 
-	if namespaceToUnseal == nil {
+	// If namespace is still sealed meaning we do not have enough shards yet, return early
+	if ns.core.NamespaceSealed(namespaceToUnseal) {
```
satoqz: I'd much prefer if the seal manager just propagated this information by `bool, error` return in the call above. Seems less racy.

vault/logical_system_namespaces_seals.go
```diff
@@ -0,0 +1,599 @@
...
+This path responds to the following HTTP methods.
+
+	POST /<name>/seal
+		Seal a namespace.
+
+	POST /<name>/unseal
+		Unseal a namespace.
+
+    GET /<name>/seal-status
+    	Returns the seal status of the namespace.
```
satoqz: Indent still wrong

vault/logical_system_namespaces_seals.go
satoqz, line 1: Most fields in this file are missing descriptions, but that can also be done later.

http/sys_generate_root.go
```diff
@@ -54,14 +56,18 @@ func handleSysGenerateRootAttemptGet(core *vault.Core, w http.ResponseWriter, r
...
-	if err != nil {
+	generationConfig, err := core.GenerateRootConfiguration(namespace.RootNamespace)
+	switch err {
+	// we return the progress as 0 in this case, root generation has not started
+	case nil:
+	case vault.ErrNoRootGeneration:
+	default:
 		respondError(w, http.StatusInternalServerError, err)
 		return
 	}
```
satoqz:
```suggestion
	switch {
	// we return the progress as 0 in this case, root generation has not started
	case errors.Is(err, vault.ErrNoRootGeneration):
	case err != nil:
		respondError(w, http.StatusInternalServerError, err)
		return
	}
```

vault/generate_root.go
```diff
@@ -33,11 +38,23 @@ type GenerateRootStrategy interface {
...
+	nsSeal, found := c.sealManager.sealsByNamespace[ns.UUID]["default"]
+	if !found {
+		return fmt.Errorf("no seal found for namespace")
+	}
+	rootKey, err := c.sealManager.unsealKeyToRootKey(ctx, nsSeal, combinedKey, false, false)
 	if err != nil {
 		return fmt.Errorf("unable to authenticate: %w", err)
 	}
-	if err := c.barrier.VerifyRoot(rootKey); err != nil {
+	nsBarrier, ok := c.sealManager.barrierByNamespace.Get(ns.Path)
```
satoqz: Needs lock, better suited with an actual method call to seal manager.

vault/generate_root.go
```diff
@@ -209,13 +259,17 @@ func (c *Core) GenerateRootUpdate(ctx context.Context, key []byte, nonce string,
 		return nil, &ErrInvalidKey{fmt.Sprintf("key is longer than maximum %d bytes", max)}
 	}
 
+	seal, found := c.sealManager.sealsByNamespace[ns.UUID]["default"]
```
satoqz: Needs lock

vault/auth.go
satoqz, line 1: This looks fine, but I wonder if the "load" and "setup" steps can't be refactored to share even more code between per-namespace handling and on-root-unseal handling. Same for `mount.go` ofc.

vault/seal_manager.go
```diff
@@ -134,45 +170,285 @@ func (sm *SealManager) SealNamespace(ns *namespace.Namespace) error {
...
+func (sm *SealManager) SecretProgress(ns *namespace.Namespace) (int, string) {
+	sm.lock.RLock()
+	defer sm.lock.RUnlock()
+
+	switch sm.unlockInformationByNamespace[ns.UUID]["default"] {
+	case nil:
+		return 0, ""
+	default:
+		return len(sm.unlockInformationByNamespace[ns.UUID]["default"].Parts), sm.unlockInformationByNamespace[ns.UUID]["default"].Nonce
+	}
```
```suggestion
	info, ok := sm.unlockInformationByNamespace[ns.UUID]["default"]
	if !ok {
		return 0, ""
	}

	return len(info.Parts), info.Nonce
```

vault/seal_manager.go
```diff
@@ -134,45 +170,285 @@ func (sm *SealManager) SealNamespace(ns *namespace.Namespace) error {
...
+			return false, err
+		}
+		sm.unlockInformationByNamespace[ns.UUID]["default"] = &unlockInformation{
+			Nonce: uuid,
+		}
+	}
+
+	// Store this key
+	sm.unlockInformationByNamespace[ns.UUID]["default"].Parts = append(sm.unlockInformationByNamespace[ns.UUID]["default"].Parts, key)
+	return true, nil
```
satoqz: You don't need all those map lookups, even if it's less concise.

```suggestion
	byNamedSeal, ok := sm.unlockInformationByNamespace[ns.UUID]
	if !ok {
		return false, fmt.Errorf("some error you'll need to come up with")
	}

	// Check if we already have this piece
	info, ok := byNamedSeal["default"]
	if ok {
		for _, existing := range info.Parts {
			if subtle.ConstantTimeCompare(existing, key) == 1 {
				return false, nil
			}
		}
	} else {
		uuid, err := uuid.GenerateUUID()
		if err != nil {
			return false, err
		}
		info = &unlockInformation{Nonce: uuid}
		byNamedSeal["default"] = info
	}

	// Store this key
	info.Parts = append(info.Parts, key)
	return true, nil
```

vault/seal_manager.go
```diff
@@ -134,45 +170,285 @@ func (sm *SealManager) SealNamespace(ns *namespace.Namespace) error {
...
+}
+
+// getUnsealKey uses key fragments recorded by recordUnsealPart and
+// returns the combined key if the key share threshold is met.
+// If the key fragments are part of a recovery key, also verify that
+// it matches the stored recovery key on disk.
+func (sm *SealManager) getUnsealKey(ctx context.Context, seal Seal, ns *namespace.Namespace) ([]byte, error) {
```
satoqz, line 402: Please fix map usage 😭

vault/seal_manager.go
```diff
@@ -134,45 +170,285 @@ func (sm *SealManager) SealNamespace(ns *namespace.Namespace) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Attempt to unlock
+	if err := barrier.Unseal(ctx, rootKey); err != nil {
+		return err
+	}
+
+	sm.logger.Debug("namespace is unsealed")
```
satoqz: I know it's a debug leftover, but would turn this into a proper info with some more context.

vault/seal_manager.go
```diff
@@ -134,45 +170,285 @@ func (sm *SealManager) SealNamespace(ns *namespace.Namespace) error {
...
+	case nil:
+		return 0, ""
+	default:
+		return len(sm.unlockInformationByNamespace[ns.UUID]["default"].Parts), sm.unlockInformationByNamespace[ns.UUID]["default"].Nonce
+	}
+}
+
+func (sm *SealManager) GetSealStatus(ctx context.Context, ns *namespace.Namespace) (*SealStatusResponse, error) {
+	// Verify that any kind of seal exists for a namespace
+	seal, ok := sm.sealsByNamespace[ns.UUID]["default"]
```
satoqz: Probably needs lock

satoqz (commented):
Suppose that's all of it for now, I can't give a super accurate review on all the sealing logic (`seal_manager.go`) as I'm a bit clueless there.

vault/generate_root.go
wslabosz-reply: right.

vault/seal_manager.go
wslabosz-reply: I've copied this part from `(*Core)` without giving it too much thought

vault/seal_manager.go
wslabosz-reply: yeah.....

vault/logical_system_namespaces_seals.go
```diff
@@ -0,0 +1,599 @@
...
+		return handleError(err)
+	}
+
+	if ns == nil {
+		return nil, fmt.Errorf("namespace %q doesn't exist", name)
+	}
+
+	barrier := b.Core.sealManager.NamespaceBarrier(ns.Path)
+	if barrier == nil {
+		return nil, fmt.Errorf("namespace %q doesn't have a barrier setup", ns.Path)
```
satoqz: For consistency with the other errors thrown in this file.

```suggestion
		return nil, fmt.Errorf("namespace %q is not a sealable namespace", ns.Path)
```

vault/seal_manager.go
```diff
@@ -1,60 +1,79 @@
...
 
-	sealsByNamespace     map[string][]Seal
-	barrierByNamespace   *radix.Tree
-	barrierByStoragePath *radix.Tree
+	// this additional map[string] layer on both seals and
+	// unlockInformation is a map of distinct (named) seals
+	sealsByNamespace             map[string]map[string]Seal
+	unlockInformationByNamespace map[string]map[string]*unlockInformation
+	barrierByNamespace           *radix.Tree
+	barrierByStoragePath         *radix.Tree
```
satoqz, line 47: I'm just going to say it: These should all be a single `*radix.Tree` that maps to a `sealManagerEntry` of sorts that bundles the relevant per-namespace info. I believe that would make the seal manager feel less awkward and error prone.

wslabosz-reply: @satoqz 
single `*radix.Tree` wouldn't suit a need of more than two different entries on the same path right? Like for a parallel sealing we would need to persist `x` seal configs for a namespace, and somehow manage that, or am I missing something?

satoqz: @wslabosz-reply I was thinking of something like this:


```go
type SealManager struct {
  // ...

  // entries maps namespace path -> *sealManagerEntry.
  entries *radix.Tree
}

type sealManagerEntry struct {
  barrier           SecurityBarrier
  seals             map[string]Seal
  unlockInformation map[string]*unlockInformation
}
```

so `sealManagerEntry` includes all of what is currently:
- `sealsByNamespace`
- `barrierByNamespace`
- `unlockInformationByNamespace`

As I think you've mentioned yourself, `barrierByStoragePath` is just convenience and can be replicated based on `barrierByNamespace`.

vault/logical_system.go
```diff
@@ -138,40 +138,42 @@ func NewSystemBackend(core *Core, logger log.Logger) *SystemBackend {
...
-	b.Backend.Paths = append(b.Backend.Paths, b.pprofPaths()...)
-	b.Backend.Paths = append(b.Backend.Paths, b.remountPaths()...)
-	b.Backend.Paths = append(b.Backend.Paths, b.metricsPath())
-	b.Backend.Paths = append(b.Backend.Paths, b.monitorPath())
-	b.Backend.Paths = append(b.Backend.Paths, b.inFlightRequestPath())
-	b.Backend.Paths = append(b.Backend.Paths, b.hostInfoPath())
-	b.Backend.Paths = append(b.Backend.Paths, b.quotasPaths()...)
-	b.Backend.Paths = append(b.Backend.Paths, b.loginMFAPaths()...)
-	b.Backend.Paths = append(b.Backend.Paths, b.introspectionPaths()...)
+	b.Paths = append(b.Paths, b.configPaths()...)
```
cipherboy, line 141: Neat, though I guess this is one of the refactors that would've made this PR smaller if we left it out ;-)

vault/namespace_store.go
```diff
@@ -614,8 +609,9 @@ func (ns *NamespaceStore) ModifyNamespaceByPath(ctx context.Context, path string
 }
 
 // ListAllNamespaces lists all available namespaces, optionally including the
-// root namespace.
-func (ns *NamespaceStore) ListAllNamespaces(ctx context.Context, includeRoot bool) ([]*namespace.Namespace, error) {
+// root namespace. includeSealed flag dictates whether sealed namespaces are
+// included in the result slice
+func (ns *NamespaceStore) ListAllNamespaces(ctx context.Context, includeRoot, includeSealed bool) ([]*namespace.Namespace, error) {
```
cipherboy, line 614: We should probably return a boolean here (`[]ns, partial bool, error`), where `partial` indicates one or more sealed namespaces were encountered and thus the results are not authoritative.

vault/seal_manager.go
cipherboy: Yeah, I think also using just a single radix might also help with locking.

cipherboy (approved):
I feel comfortable enough approving and merging and we'll continue to iterate on this in follow-up PRs.

cipherboy:
I opted for rebase and merge. Thanks @wslabosz-reply, @satoqz, @rencooo, @phyrog, @Huy-Dinh-Reply and others!

vault/seal_manager.go
```diff
@@ -116,63 +146,390 @@ func (sm *SealManager) StorageAccessForPath(path string) StorageAccess {
...
+		}
+		if err := sm.core.namespaceStore.UnloadNamespaceCredentials(ctx, descendantNamespace); err != nil {
+			errs = errors.Join(errs, err)
+		}
+		if err := sm.core.namespaceStore.UnloadNamespaceMounts(ctx, descendantNamespace); err != nil {
+			errs = errors.Join(errs, err)
+		}
+		if err = s.Seal(); err != nil {
+			errs = errors.Join(errs, err)
+		}
```
satoqz, line 177: @wslabosz-reply in retrospect I'm not convinced this works? Doesn't this go over just the sealable child namespaces, but not _all_ child namespaces? Hence we'd not necessarily unload all policies/mounts. I suppose this operation should be driven by the namespace store instead?

wslabosz-reply: @satoqz 
I don't think I thought about that, you are right, the (non-sealable) namespaces won't be cleaned up by this. To unload them (and load after unseal) it would have to be done by the namespace store, have to revisit...

cipherboy: Identity store would also need to be cleaned up fwiw.

Issue #1600, opened 2025-07-29
Improve git hooks


phil9909:
In the repo, we have some git hooks, which are installed during `make prep`: https://github.com/openbao/openbao/blob/b33e5a059797fc052db4994a5eaf3642b91edefd/Makefile#L150

This includes the following hooks:
- `pre-commit`
  - check go formating
  - run UI linter
- `pre-push`
  - ensure no file named `command/version_ent.go` is pushed

I propose the following changes:
- drop the `pre-push` hook: this is a leftover from HashiCorps code split.
- add a `commit-msg` hook, which ensures that the commit is signed-off.

PR #1601, opened 2025-07-29
improve git hooks

phil9909:
Resolves https://github.com/openbao/openbao/issues/1600


.hooks/commit-msg
```diff
@@ -0,0 +1,51 @@
...
+# Add all check functions to this space separated list.
+# They are executed in this order (see end of file).
+CHECKS="check_signoff"
+
+# ensure the commit message contains a Signed-off-by trailer
+check_signoff() {
+    if ! grep --silent '^Signed-off-by: ' "$COMMIT_MESSAGE_FILE" > /dev/null ; then
+        block "Commit is not signed-off. Hint use: git commit --signoff"
+    fi
+}
```
satoqz: Personally I'd prefer if this was pre-push, so I don't have to care about it with any WIP commits that aren't bound for merge in a PR yet.

phil9909: A common workflow I use:

```bash
git branch -c some-thing

# do some work

git add
git commit -m wip

# do more work

git commit  --amend --no-edit

# finalize work

git commit --amend # use vi to write proper commit message
git push
```

With the `commit-msg` hook, the initial `git commit -m wip` would fail and I need to press <kbd>UP</kbd> and append `--signoff`.

With the `pre-push` hook, the `git push` would fail and IMHO fixing things is more complex now, especially if I have multiple commits.

That being said, the `pre-push` hook is less intrusive and I will just keep the `commit-msg` hook locally for me and update the PR accordingly.

satoqz: > With the pre-push hook, the git push would fail and IMHO fixing things is more complex now, especially if I have multiple commits.

```sh
git rebase --signoff HEAD~[number of commits]
```

should work well for multiple commits

phil9909: I changed to a `pre-push` hook.

I replaced the existing `pre-push` hook, which was a leftover from HashiCorp (ensuring they don't accidentally push the enterprise code to the community edition remote) with a new hook based on `.git/hooks/pre-push.sample`. Therefore looking at the diff does not make a lot of sense. Instead look at the file directly: https://github.com/adfinis-forks/openbao/blob/improve-git-hooks/.hooks/pre-push

Or my changes compared to the sample provided by git:

```diff
--- .git/hooks/pre-push.sample	2025-06-25 14:22:33.500544570 +0200
+++ .hooks/pre-push	2025-10-30 12:55:32.669719561 +0100
@@ -1,28 +1,6 @@
 #!/bin/sh
 
-# An example hook script to verify what is about to be pushed.  Called by "git
-# push" after it has checked the remote status, but before anything has been
-# pushed.  If this script exits with a non-zero status nothing will be pushed.
-#
-# This hook is called with the following parameters:
-#
-# $1 -- Name of the remote to which the push is being done
-# $2 -- URL to which the push is being done
-#
-# If pushing without using a named remote those arguments will be equal.
-#
-# Information about the commits which are being pushed is supplied as lines to
-# the standard input in the form:
-#
-#   <local ref> <local oid> <remote ref> <remote oid>
-#
-# This sample shows how to prevent push of commits where the log message starts
-# with "WIP" (work in progress).
-
-remote="$1"
-url="$2"
-
-zero=$(git hash-object --stdin </dev/null | tr '[0-9a-f]' '0')
+zero=$(git hash-object --stdin </dev/null | tr '0-9a-f' '0')
 
 while read local_ref local_oid remote_ref remote_oid
 do
@@ -40,11 +18,15 @@
 			range="$remote_oid..$local_oid"
 		fi
 
-		# Check for WIP commit
-		commit=$(git rev-list -n 1 --grep '^WIP' "$range")
+		commit=$(git rev-list --reverse --extended-regexp --invert-grep --grep '^Signed-off-by: ' --grep '\(#[0-9]+\)$' "$range" | head -n 1)
+		# find commits which                                ^ are not            ^ signed off              ^ or merge commits (happens on rebases)
+
 		if test -n "$commit"
 		then
-			echo >&2 "Found WIP commit in $local_ref, not pushing"
+			rebase_commit=$(git rev-parse --short "$commit"^)
+			echo >&2 "Found commit(s) which are not signed-off."
+			echo >&2 "You can sign-off the commits using: git rebase --signoff --autostash $rebase_commit"
+			echo >&2 "or rerun the push with --no-verify"
 			exit 1
 		fi
 	fi
```

satoqz: Thanks @phil9909! Makes sense to me.

phil9909:
I tested this hook for some months now with just one problem, which I fixed here 4a6969f3435918413ecd6ca6f8323290c7fb906d (when pushing a new branch, it finds old, non-signed-of commits, which are now explicitly ignored).

It saved me a few minutes here and there already.

Maybe someone from @openbao/openbao-org-maintainers can do a quick review and we can finally merge this.

satoqz:
@phil9909 does this imply that commits pushed together with new branch will never be checked by the hook?

.hooks/pre-push
```diff
@@ -0,0 +1,36 @@
...
+do
+	if test "$local_oid" = "$zero"
+	then
+		# Handle delete
+		:
+	else
+		if test "$remote_oid" = "$zero"
+		then
+			# New branch, examine all commits
+			range="$local_oid"
```
satoqz, line 16: Perhaps a possible hack is to set a known starting commit that has only signed-off commits following it? (Around the fork point or so)

phil9909:
> @phil9909 does this imply that commits pushed together with new branch will never be checked by the hook?

No, this hook looks at the list of commits to be pushed and finds the first that is neither signed-off of nor a merge commit<sup>1</sup>. It expects this to be either empty string or "ed3069f01d4ee023d296212a3521d3354f109194"

--- 

<sup>1</sup> using this regex: `\(#[0-9]+\)$` matching the `(#1234)` that GitHub will append to PR merges. These are not necessarily merge commits by the definition git uses (having two or more parents), but the commits created by GitHub when merging a PR, therefore "merge commits".

satoqz:
Ah gotcha @phil9909, I misinterpreted.

.hooks/pre-push
phil9909: I guess this would prevent you from pushing an orphan branch (which probably isn't a big problem) , other than that I guess it would work. But what would be the advantage here compared to my current solution?

satoqz:
Turns out I can now click the merge button. Thanks @phil9909!

PR #1602, opened 2025-07-29
Prune unused repo-level files from internalized plugins

satoqz:
Remove directories such as `.github/`, `scripts/` and files such as `.gitignore`, `.go-version`, `CHANGELOG.md`, `LICENSE` from packages that were originally separate repos that have been internalized.

- All `.gitignore` files removed should already be covered by our top-level `.gitignore`.
- All `LICENSE` files removed were MPL, I assume this is okay and they would now be covered by the top-level `LICENSE`.

I've kept the list of files to remove very aggressive for starters, you tell me if there's something you'd prefer to keep :)


driif (approved):

cipherboy:
@satoqz One of the few things I didn't get around to doing when re-adding/merging plugins was making sure there was no plugin-specific testing that was missed from these. Mind doing a quick review?

satoqz:
@cipherboy the following have additional integration test scripts that we likely don't currently test in CI:

- `builtin/credential/kerberos`
- `builtin/credential/kubernetes`
- `builtin/logical/kubernetes`
- `builtin/logical/openldap`

I've decided to revert removing any `scripts/*` dirs as these tend to tie into said integration tests and the Makefiles, it's probably better to clean those up in another pass.

PR #1607, opened 2025-07-31
Add release of distribution source tarball

DrDaveD:
This adds an `openbao-dist-<version>.tar.xz` release asset for use by packages that need to be built without access to the internet.  This will be used for example by Fedora/EPEL packaging and was [requested by a reviewer there](https://bugzilla.redhat.com/show_bug.cgi?id=2376217).

This doesn't make any changes to the openbao code so I don't think it needs a changelog entry.

I wasn't able to test a combination release of this along with the goreleaser assets because of some credentials missing from my repo, or something like that.  This creates a draft release so hopefully goreleaser will just complete that rather than complaining that it is already there.  I couldn't figure out how to make the release notes completely blank so this instead puts in a short placeholder to be overwritten by goreleaser.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,16 @@ on:
         required: true
         default: true
         type: boolean
+      dodistrelease:
+        description: "Release distribution source tarball"
+        required: true
+        default: true
+        type: boolean
+      dogorelease:
+        description: "Run goreleaser"
+        required: true
+        default: true
+        type: boolean
 
 permissions:
   contents: write
@@ -36,7 +46,7 @@ jobs:
             exit 1
           fi
 
-  build-ui:
+  build-ui-dist:
     runs-on: ubuntu-latest
     needs:
       - setup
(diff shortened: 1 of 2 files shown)


DrDaveD:
I decided that it would be better to upload the new distribution tarball after goreleaser in any case, and so added a commit to do that by using upload-artifact in the build-ui-dist job and download-artifact in a new job after the goreleaser part.

.github/workflows/release.yml
```diff
@@ -36,13 +46,15 @@ jobs:
...
-  build-ui:
+  build-ui-dist:
     runs-on: ubuntu-latest
     needs:
       - setup
+    outputs:
+      version: ${{ steps.make-distribution-tarball.outputs.version }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v4.1.7
```
cipherboy, line 45: Ugh, dependabot is not dependable >:(

(Sorry @DrDaveD, I started reviewing this and got sidetracked on this, before closing up for Friday...)

DrDaveD:
@cipherboy Please review.

.github/workflows/release.yml
```diff
@@ -96,11 +108,37 @@ jobs:
...
+        run: |
+          go mod vendor
+
+      - name: Make distribution source tarball
+        if: inputs.dodistrelease
+        id: make-distribution-tarball
+        run: |
+          version="$(git describe --tags --exclude "api/*" --exclude "sdk/*" --abbrev=0 | sed 's/^v//')"
+          echo "version=$version" | tee -a "$GITHUB_OUTPUT"
+          tar --exclude .git --exclude '*.gz' --exclude '*/ui/.yarn/cache' -cJf ../openbao-dist-$version.tar.xz --xform="s,^\.,openbao-dist-$version," .
```
cipherboy: If we've already built the UI and it builds into `http/ui`, should we exclude `ui/node_modules` as well?

LICENSE_DEPENDENCIES.tpl
```diff
@@ -0,0 +1,18 @@
+# Dependency Licenses
```
cipherboy:
```suggestion
# Go Dependency Licenses
```

cipherboy: as I don't believe this covers the UI depdencies?

.github/workflows/release.yml
DrDaveD: Oh that would be another great savings in the size of the tarball! I will try it.

LICENSE_DEPENDENCIES.tpl
DrDaveD: It's true, it doesn't include the UI dependencies.  If we don't include the `node_modules` in the source bundle, then maybe we don't technically have the dependencies?

.github/workflows/release.yml
DrDaveD: And it worked.  Updated, and now down to 35M.

LICENSE_DEPENDENCIES.tpl
DrDaveD: I made the suggested change.

cipherboy:
Hearing no objections and needing to get a release out, this has been merged.

Issue #1591, opened 2025-07-25
Feature Request: Add a bao operator subcommand for syntax-only config validation


pmcconnell-tc:
**Is your feature request related to a problem? Please describe.**
We are using SaltStack to manage OpenBao configuration, and we'd like to validate the OpenBao config file before starting or reloading the service. The closest existing feature is `bao operator diagnose`, which does parse the configuration file. However, it also attempts to initialize the storage backend, which can be unsafe or misleading when run via automation, especially if the service is not yet running or the backend is locked. This makes it less suitable for use in pre-start validation.

**Describe the solution you'd like**
A dedicated subcommand that performs only configuration parsing and validation, such as:

```
bao operator validate-config -config /etc/openbao/openbao.hcl
```

This command should:
- Return non-zero exit codes on parse errors.
- Perform only config file parsing/validation.
- Avoid initializing storage backends, telemetry systems, or listeners.
- Be safe to run on stopped or uninitialized systems, as well as running instances.
- Allow the config file to be specified as the last argument in the command line, so that it works with a temp file passed by SaltStack's check_cmd requisite.

**Describe alternatives you've considered**

- Using `bao operator diagnose -skip=storage`: this still attempts to initialize storage.
- Wrapping `bao server -dev`: this seems brittle, and may behave differently in the future.
- Wrapping `bao operator diagnose` and skipping validation entirely when the service isn't running: this works but sacrifices safety and/or functionality in first-time or immutable deployments.

**Explain any additional use-cases**

- CI/CD pipelines validating configuration changes before deployment
- Pre-flight checks during provisioning of new clusters or cluster nodes
- GitOps workflows that want to dry-run HCL validation before applying config
- Admins validating config changes without disrupting live OpenBao instances

**Additional context**
Output of `bao operator diagnose` on a running instance of OpenBao 2.3.1 with a valid config file:
```
pmcconnell@openbao001:~$ sudo -u openbao bao operator diagnose -config /etc/openbao/openbao.hcl
OpenBao v2.3.1 (e3cdbbde8eb48fb7ae92b1d34afb63012e805233), built 2025-06-25T19:20:49Z

Results:
[ failure ] Vault Diagnose
  [ success ] Check Operating System
    [ success ] Check Open File Limits: Open file limits are set to 1048575.
    [ success ] Check Disk Usage: / usage ok.
  [ success ] Parse Configuration
[...]

PR #1609, opened 2025-07-31
Add bao operator subcommand for config validation

phil9909:
Resolves https://github.com/openbao/openbao/issues/1591

Useage:

```command
$ bao operator validate-config -config node-000.json
OpenBao v2.0.0-HEAD ('3918fbaa775aaaf228fbffdf4176960047a2a873+CHANGES'), built 2025-07-31T12:58:01Z
[ warning ] Validate Config: Unknown or unsupported field unknown found in configuration at node-000.json.
```

### DIFF
--- a/changelog/1609.txt
+++ b/changelog/1609.txt
@@ -1,5 +1,3 @@
 ```release-note:improvement
-cli: add new subcommand "bao operator validate-config" to validate a config.
-It is similar to "bao operator diagnose", but it only validates the config and not the system it is running on (e.g. it will not check that the raft directory is writable).
-This allows it to be used on an operators Laptop or in CI validations.
+cli: add new subcommand "bao operator validate-config" to validate a configuration file syntax
 ```
--- a/changelog/1609.txt
+++ b/changelog/1609.txt
@@ -1,0 +1,5 @@
+```release-note:feature
+cli: add new subcommand "bao operator validate-config" to validate a config.
+It is similar to "bao operator diagnose", but it only validates the config and not the system it is running on (e.g. it will not check that the raft directory is writable).
+This allows it to be used on an operators Laptop or in CI validations.
+```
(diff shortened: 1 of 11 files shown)


changelog/1609.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:feature
```
cipherboy: This could probably just be an improvement :-)

command/operator_validate_config_test.go
```diff
@@ -0,0 +1,83 @@
...
+			},
+			diagnose.Result{
+				Name:    "Validate Config",
+				Status:  diagnose.ErrorStatus,
+				Message: "expected: IDENT | STRING got: LBRACE",
+			},
+		}, {
+			"unknown_property",
+			[]string{
+				"-config", "./server/test-fixtures/diagnose_unknown_property.hcl",
```
cipherboy: We should have a few more existing test fixtures it might not be bad to try?

cipherboy (commented):
Two minor improvement ideas:

1. I think systemd unit files have a config validation step if I recall correctly? Do we want to hook this there?
2. Various subsystems (storage, KMS, ...) lack interfaces for validating options. Should we define one without having to construct the whole system? Don't need to _implement_ it for everything, just thinking that'd add to the validation capabilities of this place while we're thinking of it.

command/operator_validate_config_test.go
phil9909: changed the test to run against all files

changelog/1609.txt
phil9909: done

changelog/1609.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:improvement
+cli: add new subcommand "bao operator validate-config" to validate a config.
+It is similar to "bao operator diagnose", but it only validates the config and not the system it is running on (e.g. it will not check that the raft directory is writable).
+This allows it to be used on an operators Laptop or in CI validations.
```
cipherboy:
```suggestion
cli: add new subcommand "bao operator validate-config" to validate a configuration file syntax
```

Only features are multi-line.

cipherboy (approved):

cipherboy:
Thanks @phil9909!

Issue #1577, opened 2025-07-23
ineffassign (assigning values, but never using them) - 79 findings


phyrog:
Findings:

```
builtin/credential/approle/backend_test.go:197:3: ineffectual assignment to secretIDAccessor (ineffassign)
		secretIDAccessor = resp.Data["secret_id_accessor"]
builtin/credential/approle/backend_test.go:256:3: ineffectual assignment to secretIDAccessor (ineffassign)
		secretIDAccessor = resp.Data["secret_id_accessor"]
builtin/credential/approle/backend_test.go:315:3: ineffectual assignment to secretIDAccessor (ineffassign)
		secretIDAccessor = resp.Data["secret_id_accessor"]
builtin/credential/jwt/path_cel_role.go:246:2: ineffectual assignment to failurePolicy (ineffassign)
	failurePolicy := "Deny" // Default value
builtin/credential/kubernetes/path_config.go:221:9: ineffectual assignment to data (ineffassign)
	block, data := pem.Decode(data)
builtin/credential/kubernetes/path_config_test.go:250:2: ineffectual assignment to cert (ineffassign)
	cert, err := parsePublicKeyPEM([]byte(testRSACert))
builtin/credential/kubernetes/path_config_test.go:382:2: ineffectual assignment to cert (ineffassign)
	cert, err = parsePublicKeyPEM([]byte(testRSACert))
builtin/credential/kubernetes/path_login_test.go:985:2: ineffectual assignment to req (ineffassign)
	req = &logical.Request{
builtin/logical/database/credentials.go:316:4: ineffectual assignment to keyBits (ineffassign)
			keyBits = 2048
builtin/logical/database/credentials.go:319:4: ineffectual assignment to signatureBits (ineffassign)
			signatureBits = 256
builtin/logical/database/credentials.go:327:5: ineffectual assignment to signatureBits (ineffassign)
				signatureBits = 256
builtin/logical/database/credentials.go:329:5: ineffectual assignment to signatureBits (ineffassign)
				signatureBits = keyBits
builtin/logical/database/secret_creds.go:43:13: ineffectual assignment to ok (ineffassign)
		username, ok := usernameRaw.(string)
builtin/logical/database/secret_creds.go:117:13: ineffectual assignment to ok (ineffassign)
		username, ok := usernameRaw.(string)
builtin/logical/kubernetes/integrationtest/helpers.go:106:2: ineffectual assignment to canListPods (ineffassign)
[...]

PR #1611, opened 2025-08-01
fix some ineffassign warnings

phil9909:

This PR fixes the first few.

### DIFF
--- a/changelog/1611.txt
+++ b/changelog/1611.txt
@@ -1,0 +1,3 @@
+```release-note:change
+`certutil.ParsePublicKeyPEM` of the package `github.com/openbao/openbao/sdk/v2/helper/certutil` will now return a `crypto.PublicKey` instead of `any`. You might need to remove type assertions from your code.
+```


builtin/credential/kubernetes/path_config.go
```diff
@@ -216,10 +216,14 @@ type kubeConfig struct {
...
 
-// PasrsePublicKeyPEM is used to parse RSA and ECDSA public keys from PEMs
+// parsePublicKeyPEM is used to parse RSA and ECDSA public keys from PEMs
 func parsePublicKeyPEM(data []byte) (crypto.PublicKey, error) {
 	block, data := pem.Decode(data)
 	if block != nil {
+		if len(data) > 0 {
+			return nil, errors.New("unexpected trailing data")
+		}
+
```
phil9909: Alternatively we could just ignore trailing data

```suggestion
	block, _ := pem.Decode(data)
	if block != nil {
```

builtin/credential/kubernetes/path_login_test.go
```diff
@@ -982,16 +982,6 @@ func TestLoginIssValidation(t *testing.T) {
...
 
-	req = &logical.Request{
-		Operation: logical.UpdateOperation,
-		Path:      "login",
-		Storage:   storage,
-		Data:      data,
-		Connection: &logical.Connection{
-			RemoteAddr: "127.0.0.1",
-		},
-	}
```
phil9909, line 993: This looks like a left-over copy&paste

builtin/logical/database/secret_creds.go
```diff
@@ -36,23 +36,22 @@ func (b *databaseBackend) secretCredsRenew() framework.OperationFunc {
 		defer txRollback()
 
 		// Get the username from the internal data
-		usernameRaw, ok := req.Secret.InternalData["username"]
+		username, ok := req.Secret.InternalData["username"].(string)
 		if !ok {
-			return nil, errors.New("secret is missing username internal data")
+			return nil, errors.New(`secret is missing "username" string in internal data`)
```
satoqz: Nit:

```suggestion
			return nil, errors.New(`secret is missing "username" field in internal data`)
```

builtin/logical/database/secret_creds.go
```diff
@@ -36,23 +36,22 @@ func (b *databaseBackend) secretCredsRenew() framework.OperationFunc {
...
 		}
-		username, ok := usernameRaw.(string)
 
-		roleNameRaw, ok := req.Secret.InternalData["role"]
+		roleName, ok := req.Secret.InternalData["role"].(string)
 		if !ok {
-			return nil, fmt.Errorf("could not find role with name: %q", req.Secret.InternalData["role"])
+			return nil, errors.New(`secret is missing "role" string in internal data`)
```
satoqz:
```suggestion
			return nil, errors.New(`secret is missing "role" field in internal data`)
```

builtin/credential/kubernetes/path_config.go
satoqz: I wonder if we can't fully replace this one with

https://github.com/openbao/openbao/blob/9739e1854bc484a5cd59b3274c12a1ce3a9c08bc/sdk/helper/certutil/helpers.go?plain=1#L479

?

This one already handles trailing data. It also does ED25519 while the private Kubernetes one only does ECDSA/RSA, but I'm not sure if that would be unexpected...

Sidenote, the function I linked above should return `crypto.PublicKey` rather than `interface{}`.

phil9909: Agreed, I don't think supporting additional key types is a problem. Worst case they are never used.

satoqz:
Thanks @phil9909! I believe this actually deserves a change log entry now because we've changed the public API of `sdk`, even if the change should be quite compatible.

phil9909:
> I believe this actually deserves a change log entry now because we've changed the public API of `sdk`, even if the change should be quite compatible.

Makes sense, done.

satoqz (approved):

JanMa (approved):
LGTM

PR #1612, opened 2025-08-01
Define KMS interfaces (WIP)

### LINKED ISSUE #1467: OpenBao kms (crypto provider) internal interface
### Summary

This RFC aims to define a unified, internal interface dealing with key management and cryptographic primitives. This interface should be  evolutive, flexible and  it should allow to plug-in third parties crypto providers (as KMS, HSM, etc.).

### Problem Statement

Currently, there are many discussions and working groups related to HSM/PKCS#11 support, External_Keys, etc. : #1320, #1319, https://github.com/openbao/go-kms-wrapping/pull/33.

In order to address all these needs, I think we should define our own internal, lightweight  kms-like interface.
This interface should be used for implementing HSM/PKCS#11 support or third parties KMS support.
But also, at some point in time, the current, internal  Bao key management and cryptography should migrate to this interface: keyring for the Bao RootKey, unseal, barrear, PKI etc.

### User-facing description

The main benefit for the end-user would  be the ability to eventually use "External_Keys", ie. to easily delegate  some cryptographic operations (signing, for example)  to third parties KMS/HSM.

### Technical Description

**Note:** The following interface proposal is not exhaustive and , for simplicity, it uses "pseudo-code".

This the Bao internal interface that should be implemented by all the crypto provider that we want support in OpenBao (PKCS#11, HSM, cloud KMS, etc.)

### Init a crypto provider

This is the current, default Bao crypto provider (based on the go crypto library):
kms = KmsManager.getKMS("bao");

But we could have also:

kms = KmsManager.getKMS("pkcs#11"); // Generic PKCS#11 provider (an HSM, for example)
kms = KmsManager.getKMS("aws");   // AWS KMS
kms = KmsManager.getKMS("gcp"); // GCP KMS

### Login / get access to crypto provider

kms.Login(credentials); // // PIN code, user/pwd, token etc.

The current approach tries to capitalize on the best choices of all different cryptographic APIs/interfaces existing in the field. 
For instance, PKCS#11 remains a de facto standard and definitely influenced most of the current cloud KMS.
However, there are some aspects of the  PKCS#11 that are more related to the historical specificities of the "old" HSM (slot, token,  probably even the session management, etc.) . Therefore, our Bao interface should rather try to "hide" these "relics".
Furthermore, whether we like or not Java, IMHO the [JCA](https://docs.oracle.com/javase/8/docs/technotes/guides/security/crypto/CryptoSpec.html) is still a good pattern to follow when it comes to deal with different crypto providers/KMS/HSM , etc. 

### Downsides

_No response_

### Security Implications

_No response_

### User/Developer Experience

_No response_

### Unresolved Questions

_No response_

### Related Issues

_No response_

### Proof of Concept

_No response_

DanGhita:
Resolves #1467


satoqz:
Notes from yesterday's WG meeting:

- We'll add the interfaces to `go-kms-wrapping` rather than `openbao`
- Alex will do initial polishing and refinement of the scope
- Then we'll continue with further discussion on any remaining details over in go-kms-wrapping

satoqz:
This has moved to https://github.com/openbao/go-kms-wrapping/tree/main/kms, so closing here. Thanks @DanGhita!

Issue #1606, opened 2025-07-30
Policies in namespaces not working after a fail-over in HA


barbirict:
**Describe the bug**
When using HA and a fail-over happens, policies in namespaces don't work, you have to manually trigger the reload by accessing the policy as root. This does not happen in the root namespaces only in created ones. Auth method doesn't matter - I've tried this with both userpass and kubernetes auth and the outcome is the same, policies aren't "loaded", not even the default policy. This basically makes HA useless in a namespace context, since you have to manually intervene every time. 

**To Reproduce**
1. Deploy a 3 node HA openbao cluster using helm chart.
2. Create namespace
3. Create userpass auth method in the new namespace
4. Create a user in userpass method
5. Create a policy (I used this one: 
```hcl
path "/" {
  capabilities = ["create","read","update","patch","delete","list","scan","sudo"]
}
```
6. Assign policy to user entity
7. Login with user and verify policy works
8. Destroy the active node
9. Wait for the standby to become active
10. Try to access anything covered by the policy (doesn't work)
11. Log out and Log back in with root token
12. Go to policies > yourpolicy and open it (If you have multiple policies you have to repeat this step for every single one)
13. Log out and Log back in as previously created user
14. Now policy is correctly applied and works as intended 

**Expected behavior**
Expected behavior should be the same as in the root namespace; when a fail over happens policies should be "loaded" also for namespaces automatically.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.3.1
* OpenBao CLI Version (retrieve with `bao version`): OpenBao v2.3.1 (e3cdbbde8eb48fb7ae92b1d34afb63012e805233), built 2025-06-25T19:20:49Z
* Server Operating System/Architecture: Talos Linux v1.10.1 / Kubernetes v1.33.0 
* Helm chart version: openbao:0.16.2

OpenBao server configuration file(s):

```hcl
    log_requests_level = "trace"
    log_level = "trace"
    ui = true

    listener "tcp" {
      tls_disable = 1
      address = "[::]:8200"
      cluster_address = "[::]:8201"
      # Enable unauthenticated metrics access (necessary for Prometheus Operator)
      #telemetry {
      #  unauthenticated_metrics_access = "true"
      #}
    }

    storage "raft" {
      path = "/openbao/data"
    }

    service_registration "kubernetes" {}
```

**Additional context**

cipherboy:
[...]

PR #1613, opened 2025-08-01
Remove policyTypeMap from Policy Store

cipherboy:
Legacy Vault Enterprise had two types of policies: ACL and RGP; the later was a type of Sentinel policy tied to tokens and entities. We do not need to retain support for these policies (as they never existed in OpenBao) and thus the token map is redundant: every entry, if present, is of type ACL. This map is mostly used to resolve token policy types too ACL or RGP (which, as mentioned above are always ACL here), but also during EGP evaluation, to avoid evaluating missing policies. For this, we use a regular policy lookup to check if the policy actually exists or not.

In the future, if another policy type is added later (such as CEL policies), we will need to reintroduce this map; however, its use is made more complex by sealing of namespaces and thus should continue to fallback to attempting to load the policy from disk instead of maintaining a full list of all possible policies.

Resolves: #1606

### DIFF
--- a/changelog/1613.txt
+++ b/changelog/1613.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/policies: fix bug with missing existing policies in namespaces during failover, startup
+```


cipherboy:
Ok, test cases are fixed locally.

vault/policy_store.go
```diff
@@ -372,22 +354,18 @@ func (ps *PolicyStore) setPolicyInternal(ctx context.Context, p *Policy, casVers
...
+	policy, err := ps.GetPolicy(ctx, name, PolicyTypeACL)
+	if err != nil {
+		return nil, err
 	}
 
-	policyType, ok := pt.(PolicyType)
-	if !ok {
-		return nil, fmt.Errorf("unknown policy type for: %v", index)
+	if policy == nil {
+		return nil, ErrPolicyNotExistInTypeMap
```
satoqz: I believe this error makes little sense now

vault/policy_store_test.go
```diff
@@ -863,7 +869,7 @@ func TestPolicyStore_NestedNamespaces(t *testing.T) {
 
 	// Test parent namespace policy access from child namespace
 	parentPolicyInChild, err := ps.GetPolicy(childCtx, "../test-nested-policy", PolicyTypeToken)
-	require.NoError(t, err)
+	require.Error(t, err) // Bad relative policies now error
 	assert.Nil(t, parentPolicyInChild, "child namespace should not access parent policy via relative path")
```
satoqz: This gave me a Monday morning heart attack at first sight 😆

satoqz (approved):

vault/policy_store_test.go
cipherboy: @satoqz Any particular reason why?

vault/policy_store.go
cipherboy: @satoqz w.r.t. TypeMap, yes, but I believe it is a strictly internal error so we could perhaps just `ErrPolicyNotExist` and update the text accordingly. Thoughts?

satoqz: @cipherboy yeah, that's what I was thinking of

vault/policy_store_test.go
satoqz: @cipherboy sleepy me initially assumed that accessing policies in the parent namespace by relative path was something that somehow previously worked, inverting my world view for a solid 5 seconds. I missed that we're just moving from `nil, nil` to `nil, error` 😄

cipherboy: I've clarified the wording of the comment.

cipherboy:
Thanks @satoqz!

PR #1621, opened 2025-08-04
Authenticated rotation (rekey) tweaks

wslabosz-reply:
Tweaks to #1518
This PR is a part-one of two parts addition of new authenticated rekeys documentation update (second one will introduce the documentation itself)

Adds:
- `operator rotate-keys` CLI command (equivalent of `operator rekey`)

Fixed two bugs resulting in broken behavior:
- `(*Core) pgpEncryptShares()` saving both recovery and root keys to the same location (backup)
- CLI Rotate request with backup missing json struct tag for `backup` property

Tweaks:
- Changed multiple variables/comments use of `Rekey` verb to `Rotate` (`rekeying` -> `rotating`)
- Changed `(b *Barrier) Rekey` -> `(b *Barrier) RotateRootKey`
- Changed `barrierRekeyConfig` -> `barrierRotationConfig` (as well as `rekeyLock` and `recoveryRekeyConfig`)
- Changed `(Seal).RekeyProgress` -> `(Seal).RotationProgress`


wslabosz-reply:
@cipherboy 
Let me know if we want to split this one up in two, one fixing the bugs and the other one just as a refactor of the `rekey` word usage

vault/rotate.go
```diff
@@ -524,9 +524,13 @@ func (c *Core) pgpEncryptShares(ctx context.Context, rotationConfig *SealConfig,
...
 
 		pe := &physical.Entry{
-			Key:   coreRecoveryUnsealKeysBackupPath,
+			Key:   coreBarrierUnsealKeysBackupPath,
 			Value: buf,
 		}
+		if recovery {
+			pe.Key = coreRecoveryUnsealKeysBackupPath
+		}
+
```
wslabosz-reply, line 533: This is one of the two "essential" changes introduced with the PR.

api/sys_rekey.go
```diff
@@ -460,59 +460,7 @@ func (c *Sys) RekeyRecoveryKeyVerificationUpdateWithContext(ctx context.Context,
...
 	err = resp.DecodeJSON(&result)
 	return &result, err
 }
-
-type RekeyInitRequest struct {
-	SecretShares        int      `json:"secret_shares"`
-	SecretThreshold     int      `json:"secret_threshold"`
-	StoredShares        int      `json:"stored_shares"`
-	PGPKeys             []string `json:"pgp_keys"`
-	Backup              bool
```
wslabosz-reply, line 473: second one essential change in the PR - addition of the `json` struct tag for the `Backup` field (struct renamed to `RotateInitRequest`, moved to `api/sys_rotate.go`)

wslabosz-reply:
looking into failed `vault/external_tests/raft.TestRaft_SnapshotAPI_Rotate_Forward/rotate` test

api/sys_rekey.go
```diff
@@ -12,12 +12,12 @@ import (
 )
 
 // Deprecated: use RotateRootStatus instead.
-func (c *Sys) RekeyStatus() (*RekeyStatusResponse, error) {
+func (c *Sys) RekeyStatus() (*RotateStatusResponse, error) {
```
cipherboy: I think changing the types would be a breaking change, no?

cipherboy: Articulating differently, I think the client and CLIs should have a separate, longer deprecation period than 2.4.0 for the UI; there's nothing inherently wrong with the Client/CLI supporting the `rekey` terminology and the deprecation comment here gives the correct alternative.

api/sys_rotate.go
```diff
@@ -29,17 +87,17 @@ func (c *Sys) RotateRootStatusWithContext(ctx context.Context) (*RekeyStatusResp
...
 	return result.Data, err
 }
 
-func (c *Sys) RotateRecoveryStatus() (*RekeyStatusResponse, error) {
+func (c *Sys) RotateRecoveryStatus() (*RotateStatusResponse, error) {
 	return c.RotateRecoveryStatusWithContext(context.Background())
 }
 
-func (c *Sys) RotateRecoveryStatusWithContext(ctx context.Context) (*RekeyStatusResponse, error) {
+func (c *Sys) RotateRecoveryStatusWithContext(ctx context.Context) (*RotateStatusResponse, error) {
```
cipherboy, line 100: For these, it is fine to change.

helper/logging/logfile.go
```diff
@@ -25,7 +25,7 @@ type LogFile struct {
 	// Path to the log file
 	logPath string
 
-	// duration between each file rotation operation
+	// duration between each file rotation
```
cipherboy: This change seems unrelated. It is weird, but I also don't think it is that bad to call it an operation fwiw.

vault/core.go
```diff
@@ -324,9 +324,9 @@ type Core struct {
 	// These variables holds the config and shares we have until we reach
 	// enough to verify the appropriate root key. Note that the same lock is
 	// used; this isn't time-critical so this shouldn't be a problem.
-	barrierRekeyConfig  *SealConfig
-	recoveryRekeyConfig *SealConfig
-	rekeyLock           sync.RWMutex
+	barrierRotationConfig  *SealConfig
```
cipherboy:
```suggestion
	rootRotationConfig  *SealConfig
```

This isn't the configuration for the actual barrier keyring rotation (see #1021's terminology), but instead for the root key. While the barrier keyring needs to be involved (the root key is written encrypted via the barrier for standby upgrades, and the keyring needs to be re-encrypted with the new root key), the barrier isn't what is being rotated.

I know this will be a tough change, but I think it'd help with clarity.

cipherboy (commented):
Thanks, @wslabosz-reply! I really appreciate the work towards clarity in the internals so it matches our outwards docs. It'll be a lot easier for users and developers to understand, especially with the consistency in #1622.

helper/logging/logfile.go
wslabosz-reply: that was a bulk change, can revert this one

api/sys_rekey.go
wslabosz-reply: I will bring back the types as aliases to the new ones.

vault/barrier_aes_gcm.go
```diff
@@ -724,8 +724,8 @@ func (b *AESGCMBarrier) ActiveKeyInfo() (*KeyInfo, error) {
 	return info, nil
 }
 
-// Rekey is used to change the root key used to protect the keyring
-func (b *AESGCMBarrier) Rekey(ctx context.Context, key []byte) error {
+// Rotate is used to change the root key used to protect the keyring
```
satoqz:
```suggestion
// RotateRootKey is used to change the root key used to protect the keyring
```

command/operator_rekey.go
```diff
@@ -57,6 +52,10 @@ func (c *OperatorRekeyCommand) Help() string {
 	helpText := `
 Usage: bao operator rekey [options] [KEY]
 
+  WARNING: this method is being deprecated, please use:
+        $ bao operator rotate-keys
```
satoqz:
```suggestion
      $ bao operator rotate-keys 
```

satoqz (approved):
Nice, LGTM.

cipherboy (approved):
This looks great, thanks @wslabosz-reply !

PR #1624, opened 2025-08-04
Add negative test for HCSEC-2025-18

cipherboy:
When we fixed HCSEC-2024-05 (3300aa171fa983a47cb5f12e9287f60fd2891ff7), we correctly addressed the root cause: certificates should be compared with full wire form and any deviation should likely not be tolerated. Theoretically two certificates could have different encoded forms due to differences in parsers, however, this is unlikely as the signature would differ if the TBSCertificate differed which means it can only be in a few areas (the heading + length or in the signature field).

Our fix thus prevented us from being affected by this vulnerability.

See also: https://discuss.hashicorp.com/t/hcsec-2025-18-vault-certificate-auth-method-did-not-validate-common-name-for-non-ca-certificates/76037

---

cc:


builtin/credential/cert/backend_test.go
```diff
@@ -2687,4 +2723,17 @@ func TestBackend_RegressionDifferentTrustedLeaf(t *testing.T) {
...
 	}
+
+	// Create a new API client with the tampered leaf; it should fail.
+	tamperedAClient := getAPIClient(cores[0].Listeners[0].Address.Port, cores[0].TLSConfig(), leafCertATamperedFile.Name(), leafCertAKeyFile.Name())
+
+	secret, err = tamperedAClient.Logical().Write("auth/cert/login", map[string]interface{}{
+		"name": "trusted-leaf",
+	})
+	if err == nil {
+		t.Fatalf("when logging in with different leaf from trusted, expected err but got none: err=%v / secret=%v", err, secret)
```
satoqz, line 2765: Nit: Logging `err` here has no use, it's `nil`.

satoqz:
@cipherboy isn't the attack vector that you would set the CN to be something that's actually valid and precisely matches what's stored inside a cert registered with the engine (in this case, `a.cipherboy.com`) but isn't actually the same cert? Both would fail at the same check, but I'm not sure this tests the right issue?

satoqz (commented):
Looks good

builtin/credential/cert/backend_test.go
cipherboy: Ah yeah, I copied the earlier version I wrote, which also logs err. :-D 

I concur about the test, I'll add a second. But the raw comparison should be sufficient in either case.

cipherboy:
@satoqz Addressed; I confirmed these both fail pre-https://github.com/openbao/openbao/commit/3300aa171fa983a47cb5f12e9287f60fd2891ff7

JanMa (approved):

PR #1625, opened 2025-08-05
Fix TOTP verification reuse bypass

cipherboy:
The underlying TOTP library used, github.com/pquerna/otp, transparently removes whitespace from codes prior to processing. However, it does not internally have a code reuse cache; OpenBao maintained this itself without processing whitespace removal. We now opt to follow Vault's lead and reject invalid codes.

See also: https://discuss.hashicorp.com/t/hcsec-2025-17-vault-totp-secrets-engine-code-reuse/76036/1


builtin/logical/totp/backend_test.go
```diff
@@ -340,7 +344,11 @@ func TestBackend_keyCrudDefaultValues(t *testing.T) {
 			testAccStepValidateCode(t, "test", code, true, false),
 			// Next step should fail because it should be in the used cache
 			testAccStepValidateCode(t, "test", code, false, true),
-			testAccStepValidateCode(t, "test", invalidCode, false, false),
+			// Next step should fail because it should be in the used cache;
```
satoqz: Doesn't it just not reach the cache in the first place now?

cipherboy: @satoqz right, I'll update the comment. I suppose we could keep compatibility and strip whitespace before hitting the cache but IMO stricter validation isn't a bad thing.

satoqz (approved):

cipherboy:
Thanks all, merging!

PR #1626, opened 2025-08-05
drop legacy mapstructure hook in favor of upstream alternative 

satoqz:
We should be able to drop our own `LegacyStringToSliceHookFunc` now that a version of it was added upstream in https://github.com/go-viper/mapstructure/releases/tag/v2.4.0.


wslabosz-reply (approved):
good

satoqz:
We did release the function as part of `sdk`'s public API, so I'm not quite sure on the release note/deprecation process we go with here... I severely doubt that anyone's used it, is it fine to just remove + `release-note:change`?

cipherboy (approved):

PR #1627, opened 2025-08-05
Ensure lowercase policy in entities to forbid root policy

cipherboy:
When validating a policy is not the root policy, ensure we first lowercase policy names before comparing. This ensures a malicious entity administrator cannot directly escalate to the root policy. While this is nominally bad, any access to identity subsystem is highly privileged and should be adequately restricted as it allows granting access to additional policies. These policies may in turn allow writing new or modifying existing policies, which means practically root-like access anyways.

See also: https://discuss.hashicorp.com/t/hcsec-2025-13-vault-root-namespace-operator-may-elevate-token-privileges/76032

### DIFF
--- a/changelog/1627.txt
+++ b/changelog/1627.txt
@@ -1,0 +1,3 @@
+```release-note:security
+core/identity: Correctly lowercase policy names to prevent root policy assignment. HCSEC-2025-13 / CVE-2025-5999.
+```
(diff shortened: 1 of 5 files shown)


wslabosz-reply (approved):

cipherboy:
Thanks all!

PR #1628, opened 2025-08-05
Fix timing leak in userpass auth method

wslabosz-reply:
When trying to login using `userpass` auth method, we've attempted to mitigate the risk of timing-based information leaks by invoking bcrypt with a placeholder `"dummy"` string if the provided username did not exist in storage. Unfortunately `CompareHashAndPassword` exits early if the hash format provided is invalid.

This PR:
- fixes the format of the dummy hash
- enforces storage read on login (no cache retrieval)
- adds a test verifying retrieval through backend and not cache

See also: https://discuss.hashicorp.com/t/hcsec-2025-15-timing-side-channel-in-vault-s-userpass-auth-method/76034

### DIFF
--- a/builtin/credential/userpass/path_login.go
+++ b/builtin/credential/userpass/path_login.go
@@ -91,7 +91,7 @@ func (b *backend) pathLogin(ctx context.Context, req *logical.Request, d *framew
 	// for the bcrypt check so as not to have a timing leak.
 	if user != nil && userError == nil {
 		if len(user.PasswordHash) == 0 {
-			return nil, errors.New("invalid user entry: refusing to process pre-Vault v0.2 record")
+			return logical.ErrorResponse("invalid username or password"), nil
 		}
 
 		userPassword = user.PasswordHash
--- a/builtin/credential/userpass/path_login_test.go
+++ b/builtin/credential/userpass/path_login_test.go
@@ -59,12 +59,10 @@ func TestPathLogin_TimingLeak(t *testing.T) {
 
 	start := time.Now()
 	resp, err := b.HandleRequest(ctx, req)
-	existingUserTime := time.Since(start)
+	// ensuring we actually hit the storage
+	require.Greater(t, time.Since(start).Seconds(), 2.01*time.Second.Seconds())
 	require.Equal(t, resp.Data["error"], "invalid username or password")
 	require.ErrorIs(t, err, logical.ErrInvalidCredentials)
-
-	// ensuring we actually hit the storage
-	require.Greater(t, existingUserTime.Seconds(), 2.01*time.Second.Seconds())
 
 	req = &logical.Request{
 		Operation: logical.UpdateOperation,
@@ -77,15 +75,8 @@ func TestPathLogin_TimingLeak(t *testing.T) {
 
 	start = time.Now()
 	resp, err = b.HandleRequest(ctx, req)
-	notExistingUserTime := time.Since(start)
+	// ensuring we actually hit the storage
+	require.Greater(t, time.Since(start).Seconds(), 2.01*time.Second.Seconds())
 	require.Equal(t, resp.Data["error"], "invalid username or password")
 	require.Nil(t, err)
-
-	// ensuring we actually hit the storage
-	require.Greater(t, existingUserTime.Seconds(), 2.01*time.Second.Seconds())
-
-	// verify that login attempt of existing user takes the
-	// same amount of time as login of non-existing user
-	// are times within 100ms of each other - accounting for randomness.
-	require.InDelta(t, existingUserTime.Seconds(), notExistingUserTime.Seconds(), 0.1*time.Second.Seconds())
 }


builtin/credential/userpass/path_login_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	start = time.Now()
+	resp, err = b.HandleRequest(ctx, req)
+	notExistingUserTime := time.Since(start)
+	require.Equal(t, resp.Data["error"], "invalid username or password")
+	require.Nil(t, err)
+
+	// verify that read of existing user takes the
+	// same amount of time as read of non-existing user
+	require.InDelta(t, existingUserTime.Abs().Milliseconds(), notExistingUserTime.Abs().Milliseconds(), float64(time.Millisecond))
```
wslabosz-reply: @cipherboy not sure if it's too naive of a testcase

changelog/1628.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:security
+core/auth: Prevent timing-based leak in userpass auth method. HCSEC-2025-15 / CVE-2025-6011.
```
cipherboy:
```suggestion
auth/userpass: Prevent timing-based leak in userpass auth method. HCSEC-2025-15 / CVE-2025-6011.
```

builtin/credential/userpass/path_login_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"testing"
+	"time"
+
+	"github.com/openbao/openbao/helper/namespace"
+	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPathLogin_TimingLeak(t *testing.T) {
+	storage := &logical.InmemStorage{}
```
cipherboy:
```suggestion
	logger := logging.NewVaultLogger(log.Trace)
	inm  := inmem.NewInmem(nil, logger)
	latency := physical.NewLatencyInjector(inm, 2*time.Second, 0, logger)
	cache  := physical.NewCache(latency, 0, logger, &metrics.BlackholeSink{})
	storage := logical.NewLogicalStorage(cache)
```

would let us prove the physical cache busting works and we're more consistent across latency.

builtin/credential/userpass/path_login_test.go
cipherboy: I suggested an improvement that should let this work better. These timings will likely need to be adjusted (+/- expected jitter), but should hopefully work nicely.

cipherboy (commented):
Just an idea for improving the test case, the fix looks great. I didn't realize the cache bypass already existed when I suggested it. :laughing:

builtin/credential/userpass/path_login_test.go
```diff
@@ -0,0 +1,91 @@
...
+	require.Equal(t, resp.Data["error"], "invalid username or password")
+	require.Nil(t, err)
+
+	// ensuring we actually hit the storage
+	require.Greater(t, existingUserTime.Seconds(), 2.01*time.Second.Seconds())
+
+	// verify that login attempt of existing user takes the
+	// same amount of time as login of non-existing user
+	// are times within 100ms of each other - accounting for randomness.
+	require.InDelta(t, existingUserTime.Seconds(), notExistingUserTime.Seconds(), 0.1*time.Second.Seconds())
```
wslabosz-reply: @cipherboy 
I'm inclined to remove this check, as it's too unpredictable within CI.

(maybe it could be done with https://tip.golang.org/doc/go1.25#new-testingsynctest-package)

cipherboy: @wslabosz-reply That's fine. It is understandable, CI has a lot of oversubscription that may yield poor results.

builtin/credential/userpass/path_login.go
```diff
@@ -74,24 +81,25 @@ func (b *backend) pathLogin(ctx context.Context, req *logical.Request, d *framew
...
 	var userPassword []byte
-	// If there was an error or it's nil, we fake a password for the bcrypt
-	// check so as not to have a timing leak. Specifics of the underlying
-	// storage still leaks a bit but generally much more in the noise compared
-	// to bcrypt.
+	// If there was an error or it's nil, we fake a password hash
+	// for the bcrypt check so as not to have a timing leak.
 	if user != nil && userError == nil {
 		if len(user.PasswordHash) == 0 {
 			return nil, errors.New("invalid user entry: refusing to process pre-Vault v0.2 record")
```
cipherboy: @wslabosz-reply Do you mind adjusting this to the same:

> ```go
> logical.ErrorResponse("invalid username or password")
> ```

and logging this instead? Looking back at:

https://github.com/openbao/openbao/blob/8993802145833ab01d49c6070d787a9eccb81546/builtin/credential/userpass/path_login.go#L99-L116

I wonder if the issue in the brand-new HCSEC-2025-21 is that they still support legacy credentials and I was happy to remove them. That would emit a timing difference that's harder to resolve (without dropping support).

But in removing support, we _essentially_ leaked that the user existed, though, also forbid them from logging in altogether.

wslabosz-reply: sure!

cipherboy (approved):
Thanks again @wslabosz-reply!

cipherboy:
(in case you're wondering why I double-reviewed, apparently my work computer and personal laptop don't sync state and this one said "changes since last review..." even though you hadn't pushed anything new..., lol)

PR #1631, opened 2025-08-06
Update critical dependencies

cipherboy:
The OTP library was out of date and is also arguably critical and so should be updated as well.


wslabosz-reply (approved):

cipherboy:
Thanks all, merging!

PR #1632, opened 2025-08-06
Address login plugin lookahead contract failures

cipherboy:
```patch
commit 94657267e2730a00e83427b012a648fecbec4952
Author: Alexander Scheel <ascheel@gitlab.com>
Date:   Wed Aug 6 13:26:47 2025 -0500

    Address login plugin lookahead contract failures
    
    As part of HCSEC-2025-16, various plugins lacked did not conform to the
    entity alias lookahead contract; a different alias was returned in the
    login request as what is ultimately used at runtime. We have two ways of
    solving this:
    
    1. Only take the value at lookahead time. This means that for two users
       X, Y with different input data but similar aliases (e.g., subtly
       different LDAP bindings which result in accessing the same account),
       lockout will be incorrect.
    2. Or, we could strictly enforce that these two values match. This is
       the approach we opt for and subsequently enforce through
       authentication failures and core log messages in vault/.
    
    This is fundamentally a plugin (programming) failure that it does not
    match the expected interface contract.
    
    In the case of LDAP, this change is rather significant: we now incur a
    full roundtrip to the LDAP server (with authentication) in certain
    code paths.
    
    See also: https://discuss.hashicorp.com/t/hcsec-2025-16-vault-userpass-and-ldap-user-lockout-bypass/76035
    See also: https://discuss.hashicorp.com/t/hcsec-2025-20-vault-ldap-mfa-enforcement-bypass-when-using-username-as-alias/76092
    
commit d5ffa61442dc73c22d925133715441cb165be1ea
Author: Alexander Scheel <ascheel@gitlab.com>
Date:   Wed Aug 6 13:26:54 2025 -0500

    Refactor user lockout contract
    
    As part of HCSEC-2025-16, the general pattern of expecting the auth
    engine to be consistent between alias lookahead (strictly used for uesr
    lockout cache to prevent brute-forcing) and the final entity (which is
    used for the identity system and login MFA enforcement) is difficult.
    
    Plugins SHOULD return consistent data between the two, but as the
    lookahead needs to be cheap and ideally constant time to avoid user
    enumeration, it is hard to be in systems like LDAP which return
    different data at login time.
    
    This refactor ensures that user lockout behaves consistently regardless
    of whether the plugin responds consistently. In all scenarios, the
    initial response from lookahead is subsequently used to mark either a
    failed login or a valid authentication attempt. However, this first look
    is not used externally (in plugin<->identity entity aliasing) and thus
    is strictly an internal change. Thus, it may not necessarily align with
    effective user lockout.
    
    Plugins must on their own attempt to return consistently in login
    aliases; two inputs with equivalent lookahead values (e.g., spaces
    trimmed in usernames) SHOULD have equivalent alias values. However,
    any discrepancies between lookahead and alias values will not cause
    issues for lockout bypass now.
    
    See also: https://discuss.hashicorp.com/t/hcsec-2025-16-vault-userpass-and-ldap-user-lockout-bypass/76035
    
```

### DIFF
--- a/builtin/credential/jwt/path_cel_role.go
+++ b/builtin/credential/jwt/path_cel_role.go
@@ -99,20 +99,20 @@ func pathCelRole(b *jwtAuthBackend) *framework.Path {
 		},
 		"expiration_leeway": {
 			Type: framework.TypeSignedDurationSecond,
-			Description: `Duration in seconds of leeway when validating expiration of a token to account for clock skew. 
-Defaults to 150 (2.5 minutes) if set to 0 and can be disabled if set to -1.`,
+			Description: `Duration in seconds of leeway when validating expiration of a token to account for clock skew.
+ Defaults to 150 (2.5 minutes) if set to 0 and can be disabled if set to -1.`,
 			Default: claimDefaultLeeway,
 		},
 		"not_before_leeway": {
 			Type: framework.TypeSignedDurationSecond,
-			Description: `Duration in seconds of leeway when validating not before values of a token to account for clock skew. 
-Defaults to 150 (2.5 minutes) if set to 0 and can be disabled if set to -1.`,
+			Description: `Duration in seconds of leeway when validating not before values of a token to account for clock skew.
+ Defaults to 150 (2.5 minutes) if set to 0 and can be disabled if set to -1.`,
 			Default: claimDefaultLeeway,
 		},
 		"clock_skew_leeway": {
 			Type: framework.TypeSignedDurationSecond,
-			Description: `Duration in seconds of leeway when validating all claims to account for clock skew. 
-Defaults to 60 (1 minute) if set to 0 and can be disabled if set to -1.`,
+			Description: `Duration in seconds of leeway when validating all claims to account for clock skew.
+ Defaults to 60 (1 minute) if set to 0 and can be disabled if set to -1.`,
 			Default: jwt.DefaultLeeway,
 		},
 		"bound_audiences": {
@@ -148,20 +148,20 @@ Defaults to 60 (1 minute) if set to 0 and can be disabled if set to -1.`,
 			},
 			"expiration_leeway": {
 				Type: framework.TypeSignedDurationSecond,
-				Description: `Duration in seconds of leeway when validating expiration of a token to account for clock skew. 
-Defaults to 150 (2.5 minutes) if set to 0 and can be disabled if set to -1.`,
+				Description: `Duration in seconds of leeway when validating expiration of a token to account for clock skew.
+ Defaults to 150 (2.5 minutes) if set to 0 and can be disabled if set to -1.`,
 				Default: claimDefaultLeeway,
 			},
 			"not_before_leeway": {
 				Type: framework.TypeSignedDurationSecond,
-				Description: `Duration in seconds of leeway when validating not before values of a token to account for clock skew. 
-Defaults to 150 (2.5 minutes) if set to 0 and can be disabled if set to -1.`,
+				Description: `Duration in seconds of leeway when validating not before values of a token to account for clock skew.
+ Defaults to 150 (2.5 minutes) if set to 0 and can be disabled if set to -1.`,
 				Default: claimDefaultLeeway,
 			},
 			"clock_skew_leeway": {
 				Type: framework.TypeSignedDurationSecond,
-				Description: `Duration in seconds of leeway when validating all claims to account for clock skew. 
-Defaults to 60 (1 minute) if set to 0 and can be disabled if set to -1.`,
+				Description: `Duration in seconds of leeway when validating all claims to account for clock skew.
+ Defaults to 60 (1 minute) if set to 0 and can be disabled if set to -1.`,
 				Default: jwt.DefaultLeeway,
 			},
 			"bound_audiences": {
@@ -413,14 +413,14 @@ func validateCelRoleCreation(b *jwtAuthBackend, entry *celRoleEntry, ctx context
 
 func (b *jwtAuthBackend) validateCelProgram(program celhelper.CelProgram) (bool, error) {
 	// adding a minimal jwtClaims collection here, for validating usages in CEL expression
-	_, err := b.celEvalProgram(program, map[string]any{"sub": "email@example.com", "aud": "audience", "iss": "issuer"})
+	_, err := b.celEvalProgram(program, logical.UpdateOperation, map[string]any{"sub": "email@example.com", "aud": "audience", "iss": "issuer"})
 	if err != nil {
 		return false, fmt.Errorf("failed to validate CEL program: %w", err)
 	}
 	return true, nil
 }
 
-func (b *jwtAuthBackend) celEvalProgram(program celhelper.CelProgram, jwtClaims map[string]any) (any, error) {
+func (b *jwtAuthBackend) celEvalProgram(program celhelper.CelProgram, operation logical.Operation, jwtClaims map[string]any) (any, error) {
 	env, err := b.celEnv(program)
 	if err != nil {
 		return nil, err
@@ -429,8 +429,9 @@ func (b *jwtAuthBackend) celEvalProgram(program celhelper.CelProgram, jwtClaims 
 	// The "request" key allows CEL expressions to access and evaluate against input fields.
 	// Additional variables and evaluated results will be added dynamically during processing.
 	evaluationData := map[string]interface{}{
-		"claims": jwtClaims,
-		"now":    time.Now(),
+		"claims":    jwtClaims,
+		"now":       time.Now(),
+		"operation": string(operation),
 	}
 
 	// Evaluate all variables
--- a/builtin/credential/ldap/path_login.go
+++ b/builtin/credential/ldap/path_login.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/openbao/openbao/sdk/v2/framework"
 	"github.com/openbao/openbao/sdk/v2/helper/cidrutil"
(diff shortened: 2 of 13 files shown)


builtin/credential/jwt/path_cel_role.go
```diff
@@ -429,8 +429,9 @@ func (b *jwtAuthBackend) celEvalProgram(program celhelper.CelProgram, jwtClaims
 	// The "request" key allows CEL expressions to access and evaluate against input fields.
 	// Additional variables and evaluated results will be added dynamically during processing.
 	evaluationData := map[string]interface{}{
-		"claims": jwtClaims,
-		"now":    time.Now(),
+		"claims":    jwtClaims,
+		"now":       time.Now(),
+		"operation": string(operation),
```
cipherboy, line 434: @suprjinx @DrDaveD @fatima2003 FYI, I've stubbed through the operation so CEL programs can differentiate login from alias lookahead and respond more quickly in the latter if they desire.

builtin/credential/ldap/path_login.go
```diff
@@ -87,6 +110,13 @@ func (b *backend) pathLogin(ctx context.Context, req *logical.Request, d *framew
 	username := d.Get("username").(string)
 	password := d.Get("password").(string)
 
+	// See notes in pathLoginAliasLookahead(...) for more information. This
+	// is a hack specifically for HCSEC-2025-20 / CVE-2025-6013.
+	username = strings.TrimSpace(username)
+	if !cfg.CaseSensitiveNames {
+		username = strings.ToLower(username)
+	}
```
cipherboy: @openbao/openbao-core-committers @openbao/openbao-org-maintainers This change scares me. It is the only way I can think of to fix [HCSEC-2025-20](https://discuss.hashicorp.com/t/hcsec-2025-20-vault-ldap-mfa-enforcement-bypass-when-using-username-as-alias/76092) but IMHO it is really on the backend LDAP server to give us a useful error message (user does not exist) if we send them half-baked garbage rather than silently trimming spaces.

**Especially** with the invocation of `strings.ToLower(...)`, I suspect it will break users in production, if they've inadvertently set aliases with mixed-case alias references.

builtin/credential/radius/path_login.go
```diff
@@ -65,7 +65,10 @@ func pathLogin(b *backend) *framework.Path {
 func (b *backend) pathLoginAliasLookahead(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
 	username := d.Get("username").(string)
 	if username == "" {
-		return nil, errors.New("missing username")
+		username = d.Get("urlusername").(string)
+		if username == "" {
+			return nil, errors.New("missing username")
```
cipherboy, line 70: Interestingly Radius does not support user lockout but I noticed this was missing anyways.

builtin/credential/ldap/path_login.go
cipherboy: In particular:

https://github.com/openbao/openbao/blob/61a73e5732c8a1276c063ce4b44d21776cb48669/builtin/credential/ldap/backend.go#L172-L176

vs

https://github.com/openbao/openbao/blob/61a73e5732c8a1276c063ce4b44d21776cb48669/builtin/credential/ldap/backend.go#L211-L213

means that we _will_ be returning different entity aliases now. So it entirely depends on how the customer set up the entity earlier.

SamJoan: Hi @cipherboy 

~~Looking at HCSEC-2025-20, it does not seem to make any mention of "case" I.e. the `ToLower` portion of this PR could be skipped if you're only concerned about resolving HCSEC-2025-20.~~ Ah I see that it is related to HCSEC-2025-16.

Windows' LDAP implementation [is case insensitive](https://stackoverflow.com/questions/29897684/is-ldap-dn-case-insensitive) by default, so it would not break those integrations. It seems that in [most cases case insensitivity is the default as well](https://jumpcloud.com/it-index/what-is-case-sensitivity-in-ldap-filters).  The [RFC](https://www.ietf.org/rfc/rfc4512.txt) also states that

>   For example, a 'givenName' attribute can have more than one value, they must be Directory Strings, and they are case insensitive.  A 'givenName' attribute cannot hold both "John" and "JOHN", as these are equivalent values per the equality matching rule of the attribute type.

I'm not absolutely sure if this applies in this context, but it would seem that in most cases it should not impact LDAP users. The configuration can be changed as well, so that is a workaround that is suitable for users that may be affected. 🤔

cipherboy: Thanks @SamJoan! Yeah, I think on the whole this is the least-bad approach to fix just the stated issue with this method... But I'm not convinced it is globally optimal.

In general, I think we probably want warnings in the documentation _not_ to use this particular bodge and to instead use the canonical value returned from the LDAP server. The original commit doesn't give much context on why it was added in the first place.

vault/external_tests/userpass_binary/ip_token_binding_test.go
```diff
@@ -135,7 +135,10 @@ func Test_StrictIPBinding(t *testing.T) {
 
 		"-H", "Content-Type: application/json",
 		"--data", `{"password": "password"}`,
-		"https://" + vaultAddr + ":8200/v1/auth/userpass/login/testing",
+		// We switch the username to Testing to ensure case validation
+		// does not affect user lockout attribution. This is a test
+		// to validate our fix for HCSEC-2025-16 / CVE-2025-6004.
+		"https://" + vaultAddr + ":8200/v1/auth/userpass/login/Testing",
```
suprjinx, line 141: Should we also test the leading-spaces case (%20 in url? or move user name to data payload?)

cipherboy: @suprjinx This only does userpass login, not the affected ldap auth (which took username as an API parameter, not a path component); spaces in userpass are treated as routing errors, since the path component doesn't match:

```go
$ bao write 'auth/userpass/login/ admin ' password=admin
Error writing data to auth/userpass/login/ admin: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/auth/userpass/login/%20admin
Code: 404. Errors:

* unsupported path
```

https://github.com/openbao/openbao/blob/b9deb5de0a08c38d553ef35f575f4605d2f199dd/builtin/credential/userpass/path_login.go#L28

https://github.com/openbao/openbao/blob/b9deb5de0a08c38d553ef35f575f4605d2f199dd/sdk/framework/path.go#L18-L20

(where `%s` is not space but the string replacement for `Sprintf(...)` to add the desired parameter name in).


So I think it is fine to avoid testing that, in this case.

I don't have a good reproducer for LDAP; if anyone wants to write one, I'm more than happy to merge.

builtin/credential/ldap/path_login.go
suprjinx: it seems like entityId -> alias relationship is kind of loosey-goosey and should be an uuid or the like.

vault/external_tests/userpass_binary/ip_token_binding_test.go
cipherboy: I've pushed a reproducer for LDAP that confirms the specified fix does indeed work and normalize the alias name as expected in both the username-as-alias and default (server) case.

builtin/credential/ldap/path_login.go
cipherboy: @suprjinx Yeah, definitely I wish there was a better way of doing this... We'll think about this long-term, but for now this is the best we can do.

satoqz (approved):
Also going to ack FWIW, this has begun making sense to my poor brain.

driif (approved):
@cipherboy Thank You, looks good to me

builtin/credential/ldap/path_login.go
```diff
@@ -55,6 +56,28 @@ func (b *backend) pathLoginAliasLookahead(ctx context.Context, req *logical.Requ
...
 	}
 
+	// _Some_ LDAP backends will silently trim spaces, leading to an
+	// authentication lockout bypass if not adjusted in the alias.
+	// We normalize the username to avoid this, as _presumably_ users
+	// do not normally begin/end with leading space.
+	//
+	// See HCSEC-2025-16 / CVE-2025-6004 for more information.
+	username = strings.TrimSpace(username)
+
```
codergs, line 66: I am bit worried about this implicit new normalization. As per https://www.rfc-editor.org/rfc/rfc4514.html#section-4, 

> "If that UTF-8-encoded Unicode string does not have any of the following characters that need escaping, then that string can be used as the string representation of the value.
> - a space (' ' U+0020) or number sign ('#' U+0023) occurring at the beginning of the string;
> - a space (' ' U+0020) character occurring at the end of the string;

What this reads to me is as below:
- Leading spaces ARE valid in LDAP usernames (need escaping as \ John)
- Trailing spaces ARE valid in LDAP usernames (need escaping as John\ )
AND strings.TrimSpace() WILL BREAK these legitimate usernames

Is this correct assumption?

cipherboy: @codergs Yes, I do worry about this as well, as you see here: https://github.com/openbao/openbao/pull/1632#pullrequestreview-3093912814

I think we have two options:

1. Limit this to only applications where the parameter was supplied in the first place.
2. Deny this parameter altogether and strictly use the reliance on the LDAP lookup.

Preferences?

(Note that the comment place is incorrect; this has very limited scope to just the User Lookahead method which _at worst_ causes user lockout when two whitespace-identical users share the same normalized alias, whereas the change on lines 113...119 materially impact entity association and are more risky/critical.)

cipherboy: @suprjinx @JanMa @codergs I guess theoretically this would allow an attacker to lock out with `alex ` a user `alex` by sending the former. But they could do the same by just directly spamming `alex`, so I'm not really sure its a net-new DoS issue.

codergs: I was thinking more about this - let me walk through it

```
# Target user: "john.doe"
# Normal failed login attempts:
curl -X POST vault.com/v1/auth/ldap/login/john.doe -d '{"password":"wrong1"}'  # Attempt 1
curl -X POST vault.com/v1/auth/ldap/login/john.doe -d '{"password":"wrong2"}'  # Attempt 2  
curl -X POST vault.com/v1/auth/ldap/login/john.doe -d '{"password":"wrong3"}'  # Attempt 3
# User gets locked out after 3 attempts

# BUT, attacker can bypass lockout with variations:
curl -X POST vault.com/v1/auth/ldap/login/john.doe -d '{"password":"wrong4"}'    # Still locked
curl -X POST vault.com/v1/auth/ldap/login/ john.doe -d '{"password":"wrong5"}'   # Leading space - NEW USER in lockout system!
curl -X POST vault.com/v1/auth/ldap/login/john.doe  -d '{"password":"wrong6"}'   # Trailing space - NEW USER in lockout system!
curl -X POST vault.com/v1/auth/ldap/login/John.Doe -d '{"password":"wrong7"}'    # Different case - NEW USER in lockout system!
```

I wonder if we can add a new field to `FailedLoginUser`

```
type FailedLoginUser struct {
	aliasName         string  // ← Keep original 
	lockoutKeyHash    string  // ← ADD: Hash for consistent lockout (original username | aliasName | mount)
	mountAccessor     string  // ← Keep original
}
```

```
func generateLockoutHash(inputUserName, alias, mountAccessor string) string {
    // DON'T normalize inputUserName - use raw values
    // aliasName is normalized or whatever
    // create composite string with RAW values
    composite := fmt.Sprintf("%s|%s|%s", input, alias, mountAccessor)
    // generate SHA256 hash and return
}
```

cipherboy: @codergs Doesn't this mean though that an attacker can make even more requests? They only get locked out of one variant, not of the normalized form of the user.

codergs: Lets chat :) Will connect

JanMa (approved):
LGTM.

cipherboy:
From various discussions, we think this is a best effort that should remediate this issue. If there are any follow ups please contact our security mailing list.

PR #1634, opened 2025-08-06
Disable audit mount creation via API + disable audit log prefixing

satoqz:
Audit mount creation via the API places too much control over the underlying host in the operator's hands, as [^1] has shown. This introduces two new server configuration options:

- `unsafe_allow_api_audit_creation (default: false)` controls the ability to create audit mounts via the API
- `allow_audit_log_prefixing (default: false)` controls the availability of the `prefix` audit mount option

In the future (v2.4) we plan to move audit mount management almost entirely into the configuration file, providing an alternative to the now deprecated management via API. For now, the creation of new audit mounts will require setting `unsafe_allow_api_audit_creation = true`. This option can be reloaded at runtime via SIGHUP. Existing audit mounts are unaffected, unless they use the `prefix` option, in which case `allow_audit_log_prefixing = true` is required.

[^1]: https://discuss.hashicorp.com/t/hcsec-2025-14-privileged-vault-operator-may-execute-code-on-the-underlying-host/76033

### DIFF
--- a/builtin/logical/transit/path_keys_test.go
+++ b/builtin/logical/transit/path_keys_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/openbao/openbao/api/v2"
 	"github.com/openbao/openbao/audit"
 	"github.com/openbao/openbao/builtin/audit/file"
+	"github.com/openbao/openbao/command/server"
 	vaulthttp "github.com/openbao/openbao/http"
 	"github.com/openbao/openbao/sdk/v2/logical"
 	"github.com/openbao/openbao/vault"
@@ -24,6 +25,9 @@ import (
 
 func TestTransit_Issue_2958(t *testing.T) {
 	coreConfig := &vault.CoreConfig{
+		RawConfig: &server.Config{
+			UnsafeAllowAPIAuditCreation: true,
+		},
 		LogicalBackends: map[string]logical.Factory{
 			"transit": Factory,
 		},
--- a/changelog/1634.txt
+++ b/changelog/1634.txt
@@ -1,0 +1,5 @@
+```release-note:security
+audit: Add server configuration options to disable audit mount creation via the API and to disable audit log prefixing.
+  - `unsafe_allow_api_audit_creation (default: false)` controls the ability to create audit mounts via the API
+  - `allow_audit_log_prefixing (default: false)` controls the availability of the prefix audit mount option
+```
--- a/command/server/config.go
+++ b/command/server/config.go
@@ -113,6 +113,9 @@ type Config struct {
 
 	UnsafeCrossNamespaceIdentity bool `hcl:"unsafe_cross_namespace_identity"`
 
+	UnsafeAllowAPIAuditCreation bool `hcl:"unsafe_allow_api_audit_creation"`
+	AllowAuditLogPrefixing      bool `hcl:"allow_audit_log_prefixing"`
+
 	// Initialization is a configuration object that helps to initialize
 	// OpenBao. It can be specified multiple times and each instance can
 	// contain one or more `request` objects. This is used by the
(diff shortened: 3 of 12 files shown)


vault/testing.go
```diff
@@ -1642,6 +1642,7 @@ func NewTestCluster(t testing.T, base *CoreConfig, opts *TestClusterOptions) *Te
 	if coreConfig.RawConfig == nil {
 		c := new(server.Config)
 		c.SharedConfig = &configutil.SharedConfig{LogFormat: logging.UnspecifiedFormat.String()}
+		c.UnsafeAllowAPIAuditCreation = true
```
satoqz, line 1645: Just going to default the legacy behavior for anything that uses `NewTestCluster` for now, otherwise 14/16 test shards go poof

vault/audit.go
```diff
@@ -479,6 +480,14 @@ func (c *Core) newAuditBackend(ctx context.Context, entry *MountEntry, view logi
 		Location: salt.DefaultLocation,
 	}
 
+	prefix, hasPrefix := conf["prefix"]
+	badPrefixing := hasPrefix && prefix != "" &&
+		!c.rawConfig.Load().(*server.Config).AllowAuditLogPrefixing
+
+	if badPrefixing {
+		return nil, fmt.Errorf("audit log prefixing is not allowed")
+	}
```
satoqz: Unfortunately this one can't really be reloaded at runtime with much effect because it would already irrecoverably fail at unseal 🤷

satoqz:
So many test cases to patch up...

satoqz:
Cool, tests green, but still want to refactor the way I got there

vault/audit.go
cipherboy: Yes, I think that's perfectly fine.

satoqz:
Working on updates to documentation separately 😄

changelog/1634.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:feature
```
cipherboy: I'd love for these all to be grouped under security; I think reusing multi-line is fine for this.

```suggestion
```release-note:security
```

vault/logical_system.go
```diff
@@ -3137,6 +3138,10 @@ func (b *SystemBackend) handleEnableAudit(ctx context.Context, req *logical.Requ
 	description := data.Get("description").(string)
 	options := data.Get("options").(map[string]string)
 
+	if !b.Core.rawConfig.Load().(*server.Config).UnsafeAllowAPIAuditCreation {
```
cipherboy: I think this is fine, though I'm slightly curious why this wasn't in `newBackend(...)`? It seems like both this and prefix should be enforced in the same place for consistency.

cipherboy (commented):
One minor nit for curiosity.

vault/logical_system.go
satoqz: @cipherboy `newBackend` is also called on unseal/loading existing audits, so we would break existing setups. This way nothing changes if the audits already exist, but you have to set the config option for any future ones.

satoqz: Now I wonder why I've gone the stricter route with prefix enforcement, I suppose it's less likely to cause breakage there... LMKWYT.

satoqz:
New failures are just RabbitMQ + Docker being who they are.

builtin/logical/transit/path_keys_test.go
```diff
@@ -24,6 +25,9 @@ import (
 
 func TestTransit_Issue_2958(t *testing.T) {
```
wslabosz-reply, line 26: oh no, not the transit issue 2958

command/server/config.go
```diff
@@ -419,6 +422,21 @@ func (c *Config) Merge(c2 *Config) *Config {
 		result.EnableResponseHeaderRaftNodeID = c2.EnableResponseHeaderRaftNodeID
 	}
 
+	result.UnsafeCrossNamespaceIdentity = c.UnsafeCrossNamespaceIdentity
+	if c2.AllowAuditLogPrefixing {
```
wslabosz-reply: what does it have with your change tbh?
```suggestion
	if c2.UnsafeCrossNamespaceIdentity {
```

command/server/config.go
```diff
@@ -419,6 +422,21 @@ func (c *Config) Merge(c2 *Config) *Config {
...
+		result.UnsafeCrossNamespaceIdentity = c2.UnsafeCrossNamespaceIdentity
+	}
+
+	result.UnsafeAllowAPIAuditCreation = c.UnsafeAllowAPIAuditCreation
+	if c2.AllowAuditLogPrefixing {
```
wslabosz-reply:
```suggestion
	if c2.UnsafeAllowAPIAuditCreation {
```

command/server/config.go
satoqz: Oopsie, fixed. I added `UnsafeCrossNamespaceIdentity` as well because it was missing here.

builtin/logical/transit/path_keys_test.go
satoqz: Hate that one.

cipherboy: https://github.com/hashicorp/vault/issues/2958

vault/logical_system.go
cipherboy: > I suppose it's less likely to cause breakage there... 

More likely, perhaps? I think existing audit devices should be one-time auditable and then new devices created with prefixes would be forbidden, is my take. 

Otherwise, you risk erring startup (== panic during unseal) on upgrade. Though, it would tell us how many people actually use audit devices with prefixes :laughing: Likewise, you have to enable this option to allow your existing audit devices, which means you have no way to disable it later.


Making prefix behavior consistent with this option (`UnsafeAllowAPIAuditCreation`) would mean that you could one-time enable to set a prefix and then disable prefixes going forward so an attacker couldn't set one. 

My 2c.

satoqz: Yeah, agreed. Will update in a bit.

cipherboy (approved):
Thank you @satoqz! Appreciate the updates.

cipherboy:
Test failure unrelated, usual OIDC key rotation test failure. Merging...

PR #1638, opened 2025-08-06
feat(sys/seal-status): send both barrier and recovery seal types

pree:
Currently the `type` return on `v1/sys/seal-status` will always return `shamir`, as it references the barrier seal on manual unsealed clusters and the recovery seal on auto unsealed clusters.

With this PR it will be changed, so that the `type` will always return the barrier seal type and on auto unsealed clusters an additional `recovery_seal_type` will be sent.

This aligns with how Vault responds:
```
curl http://localhost:8200/v1/sys/seal-status | jq
{
  "type": "azurekeyvault",
  "initialized": true,
  "sealed": false,
  "t": 3,
  "n": 5,
  "progress": 0,
  "nonce": "",
  "version": "1.19.5",
  "build_date": "2024-08-06T20:45:58Z",
  "migration": false,
  "cluster_name": "vault-cluster-1acf9bdd",
  "cluster_id": "c43a6842-8e9d-ef17-d31c-0edc588956d1",
  "recovery_seal": true,
  "storage_type": "inmem",
  "recovery_seal_type": "shamir"
}
```

Resolves https://github.com/openbao/openbao-helm/issues/79

### DIFF
--- a/api/sys_seal.go
+++ b/api/sys_seal.go
@@ -96,21 +96,22 @@ func sealStatusRequestWithContext(ctx context.Context, c *Sys, r *Request) (*Sea
 }
 
 type SealStatusResponse struct {
-	Type         string   `json:"type"`
-	Initialized  bool     `json:"initialized"`
-	Sealed       bool     `json:"sealed"`
-	T            int      `json:"t"`
-	N            int      `json:"n"`
-	Progress     int      `json:"progress"`
-	Nonce        string   `json:"nonce"`
-	Version      string   `json:"version"`
-	BuildDate    string   `json:"build_date"`
-	Migration    bool     `json:"migration"`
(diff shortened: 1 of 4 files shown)


changelog/1638.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+api/sys/seal-status: type always returns the barrier seal type, explicitly adds recovery seal type
```
satoqz:
```suggestion
core: sys/seal-status endpoint now always returns the barrier seal type, explicitly adds recovery seal type
```

An `api` scope would imply that this is mainly a change to the `api` package, maybe this is more consistent.

pree: agreed, I've added your suggestion :)

PR #1641, opened 2025-08-07
fix(oidc-provider): correctly lowercase identity IDs (#1539)

cipherboy:
Backport of https://github.com/openbao/openbao/pull/1539.


cipherboy:
Setting DCO to pass as DCO passed on the original PR. Thanks all!

PR #1647, opened 2025-08-07
Fix timing leak in userpass auth method (#1628)

cipherboy:
Backport of https://github.com/openbao/openbao/pull/1628.


JanMa (approved):

cipherboy:
DCO set to pass because this is a backport and DCO passed on the original PR. Thanks all!

PR #1651, opened 2025-08-07
Refuse setting irregular/executable file modes in file audit

satoqz:
Additional hardening around HCSEC-2025-14, limiting the range of accepted values for `mode` in the file audit backend.

This could cause breakage in theory -- on the other hand, I'm not sure why you would set an executable audit file without questionable or malicious intent. So, open to discarding this suggestion if we think it's not worth it.

### DIFF
--- a/builtin/audit/file/backend.go
+++ b/builtin/audit/file/backend.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -104,8 +105,23 @@ func Factory(ctx context.Context, conf *audit.BackendConfig) (audit.Backend, err
 		default:
 			mode = os.FileMode(m)
 
-		}
+			// Refuse setting an irregular file mode.
+			if !mode.IsRegular() {
+				return nil, errors.New("file mode does not represent a regular file")
+			}
 
+			const execBitMask = uint32(1<<32-1) ^ 0o111
+			//                  ---------------   -----
+			//                         ^            ^
+			//                     all ones         |
+			//                                      |
+			//                       set all 3 exec bits to zeroes
+
+			// Strip executable bits. Part of the exploit for HCSEC-2025-14 /
+			// CVE-2025-6000 / CVE-2025-54997 consists of abusing the ability
+			// to create executable files.
+			mode = mode & fs.FileMode(execBitMask)
+		}
 	}
 
 	b := &Backend{
--- a/builtin/audit/file/backend_test.go
+++ b/builtin/audit/file/backend_test.go
@@ -5,6 +5,7 @@ package file
 
 import (
 	"context"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -15,10 +16,11 @@ import (
 	"github.com/openbao/openbao/helper/namespace"
 	"github.com/openbao/openbao/sdk/v2/helper/salt"
 	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/stretchr/testify/require"
 )
 
 func TestAuditFile_fileModeNew(t *testing.T) {
-	modeStr := "0777"
+	modeStr := "0644"
 	mode, err := strconv.ParseUint(modeStr, 8, 32)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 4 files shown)


JanMa (approved):
LGTM

driif (approved):

builtin/audit/file/backend_test.go
```diff
@@ -135,6 +136,30 @@ func TestAuditFile_fileMode0000(t *testing.T) {
 	}
 }
 
+func TestAuditFile_fileModeExecutable(t *testing.T) {
+	path, err := os.MkdirTemp("", "test")
+	require.NoError(t, err)
+	defer os.RemoveAll(path)
```
wslabosz-reply: linter wants you to check the error :(

phil9909: You can use [`t.TempDir()`](https://pkg.go.dev/testing#T.TempDir) instead.

> TempDir returns a temporary directory for the test to use. The directory is automatically removed when the test and all its subtests complete. Each subsequent call to t.TempDir returns a unique directory; if the directory creation fails, TempDir terminates the test by calling Fatal.

satoqz: Ah, thanks @phil9909. Copied the harness from one of the existing tests, presumably these are a bit dated.

builtin/audit/file/backend.go
```diff
@@ -103,9 +103,17 @@ func Factory(ctx context.Context, conf *audit.BackendConfig) (audit.Backend, err
 			}
 		default:
 			mode = os.FileMode(m)
-
+			// Refuse setting an irregular file mode.
```
cipherboy, line 108: This too probably warrants a call-out in the docs.

phil9909: How about stripping the executable flags instead of refusing them?

satoqz: I've considered this, but I'm not sure how I feel about silently altering input.

phil9909: For filemodes this is actually pretty common: the [`umask`](https://linux.die.net/man/2/umask) systemcall does exactly that.

phil9909: Also, I don't think there is any valid use-case for having your audit file executable. So, ignoring this seems fine.

wslabosz-reply (approved):

builtin/audit/file/backend.go
satoqz: If I understand correctly `umask` won't fully cover this case because we have explicit `chmod`s as well, changing the mode of an existing audit log file. I'm open to manually stripping the exec bits though, any opinions @cipherboy?

cipherboy: :shrug: I guess 777 is probably common enough that we probably don't want to globally deny it, but also don't think there's ever a valid use case for _actually_ granting execute bits. So yeah, I'm fine silently ignoring them.

Could you add a warning, though, on the response body if they do set execute anywhere?

satoqz: Adding warnings to `sys/audit` is a refactor rabbit hole both on the server and the client side, is it fine skipping that for now? For instance, audit device factories would need a way to emit warnings up to the handler, and once you've got that, the CLI won't even display them because `EnableAuditWithOptions` doesn't expose them. Since we will be deprecating this endpoint anyways, not sure if worth the hassle.

cipherboy: Oh, I see, that's nasty. It doesn't even have a regular `logger` :-(  Let's ignore it for now then.

builtin/audit/file/backend.go
```diff
@@ -104,8 +105,23 @@ func Factory(ctx context.Context, conf *audit.BackendConfig) (audit.Backend, err
...
+			if !mode.IsRegular() {
+				return nil, errors.New("file mode does not represent a regular file")
+			}
 
+			const execBitMask = uint32(1<<32-1) ^ 0o111
+			//                  ---------------   -----
+			//                         ^            ^
+			//                     all ones         |
+			//                                      |
+			//                       set all 3 exec bits to zeroes
```
wslabosz-reply, line 118: great ascii art

satoqz:
@phil9909 @cipherboy updated, LMKWYT! We silently strip exec bits now and keep erroring for irregular files, since the latter really really don't make sense.

website/content/docs/audit/file.mdx
```diff
@@ -57,7 +57,15 @@ these device-specific options:
   the bit pattern for the file mode, similar to `chmod`. Set to `"0000"` to
   prevent OpenBao from modifying the file mode.
 
+:::warning
 
+Note: Starting with OpenBao v2.4, any executable bits set in `mode` are zeroed
```
cipherboy:
```suggestion
Note: Starting with OpenBao v2.4.0, any executable bits set in `mode` are zeroed
```

cipherboy (approved):

satoqz:
Thanks everyone, merging!

PR #1652, opened 2025-08-07
change UI message for sealed to 'OpenBao'

### LINKED ISSUE #1650: OpenBao UI should show "OpenBao is sealed" instead of "Vault is sealed"
Image

l-with:
Resolves #1650

### DIFF
--- a/changelog/1652.txt
+++ b/changelog/1652.txt
@@ -1,2 +1,3 @@
-CHANGES:
+```release-note:improvement
 UI shows 'OpenBao is sealed' instead 'Vaul is sealed' for HTTP status 503
+```
--- a/changelog/1652.txt
+++ b/changelog/1652.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-UI shows 'OpenBao is sealed' instead 'Vaul is sealed' for HTTP status 503
+ui: Message changed to 'OpenBao is sealed' from 'Vault is sealed' for HTTP status 503.
 ```
--- a/changelog/1652.txt
+++ b/changelog/1652.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-ui: change the error text to 'OpenBao is Sealed'
+ui: change the message 'Vault is sealed to 'OpenBao is Sealed' by changing the title of the unseal template
 ```
--- a/changelog/1652.txt
+++ b/changelog/1652.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-ui: make UI only depend on error code 503 and no not depend on the error text
+ui: make UI only depend on error code 503 (not depend on the error text) and change the error text to 'OpenBao is Sealed'
 ```
--- a/changelog/1652.txt
+++ b/changelog/1652.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-ui: Message changed to 'OpenBao is sealed' from 'Vault is sealed' for HTTP status 503.
+ui: make UI only depend on error code 503 and no not depend on the error text
 ```
--- a/changelog/1652.txt
+++ b/changelog/1652.txt
@@ -1,0 +1,2 @@
+CHANGES:
+UI shows 'OpenBao is sealed' instead 'Vaul is sealed' for HTTP status 503
--- a/changelog/1652.txt
+++ b/changelog/1652.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-ui: make UI only depend on error code 503 (not depend on the error text) and change the error text to 'OpenBao is Sealed'
+ui: change the error text to 'OpenBao is Sealed'
 ```
--- a/ui/app/routes/vault/cluster.js
+++ b/ui/app/routes/vault/cluster.js
@@ -130,7 +130,7 @@ export default Route.extend(ModelBoundaryRoute, ClusterRoute, {
 
   actions: {
     error(e) {
-      if (e.httpStatus === 503 && e.errors[0] === 'Vault is sealed') {
+      if (e.httpStatus === 503 && e.errors[0] === 'OpenBao is sealed') {
         this.refresh();
       }
       return true;
--- a/ui/app/routes/vault/cluster.js
+++ b/ui/app/routes/vault/cluster.js
@@ -131,7 +131,6 @@ export default Route.extend(ModelBoundaryRoute, ClusterRoute, {
   actions: {
     error(e) {
       if (e.httpStatus === 503 && e.errors[0] === 'Vault is sealed') {
-        e.errors[0] = 'OpenBao is sealed'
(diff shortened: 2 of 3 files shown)


changelog/1652.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+UI shows 'OpenBao is sealed' instead 'Vaul is sealed' for HTTP status 503
```
wslabosz-reply: nit: cosmetic adjustment to the changelog
```suggestion
ui: Message changed to 'OpenBao is sealed' from 'Vault is sealed' for HTTP status 503.
```

cipherboy:
Hey @l-with -- appreciate the PR, but sadly we cannot accept it. As discussed in #1188 and #1599, various automation relies on this exact error message (`Vault is ....`) and thus changing the server-returned error string would fail.

So, the API check you've added isn't sufficient; the string returned from the server will indefinitely be `Vault is...` which should trigger the refresh, today. If I'm missing something, please let me know!

l-with:
Hi @cipherboy I thought a change in JavSacript code inside the /ui folder would only affect the web UI and nothing else.
1. Does this also affect parts outside the UI?
2. Does anything rely on text displayed in the UI?

cipherboy:
@l-with As I understand what you've changed, the check is taking in the error message from the server (that's `e`) and comparing that the error code with `OpenBao is sealed`.

Yes, you've not changed the server error.

But, you've also not... changed the server error... so this check will now no longer trigger and will instead fail _because_ the server is still sending `Vault is sealed`... and so the `isSealed` check will now no longer ever return true (except on someone's custom build).

Now, you could leave the check alone, but adjust the error message in the body of this if block to be the new value and maybe that'll fix the presentation issue you see. :-) But not entirely sure which instance you're trying to change.

ui/app/routes/vault/cluster.js
```diff
@@ -130,7 +130,7 @@ export default Route.extend(ModelBoundaryRoute, ClusterRoute, {
 
   actions: {
     error(e) {
-      if (e.httpStatus === 503 && e.errors[0] === 'Vault is sealed') {
+      if (e.httpStatus === 503 && e.errors[0] === 'OpenBao is sealed') {
```
driif: This statement is always false because the server never returns an error with message 'OpenBao is sealed'.

l-with:
@cipherboy I worked to fast, because I thought the UI would create the text based on the return code and I did not look sharp enough on the code.

wslabosz-reply:
To be honest, I also looked just at the changelog, and assumed the ui change was what was in the changelog, but yeah, this is just based on error message sent by the server (which as @cipherboy) pointed, cannot be changed as there are many parties dependent on that exact string.

ui/app/routes/vault/cluster.js
cipherboy: Right, I think if you add a new line 134, you could see if that fixes the issue (`e.errors[0] = 'OpenBao is sealed'` -- in the body of this if statement), but I'm not entirely convinced it is or isn't going to fix it. :-)

l-with:
@wslabosz-reply My personal opinion is that tools should only depend on the HTTP status response (403 for API /sys/health if sealed) or the error code of the CLI (3 for `bao status` if sealed). (This is how I build my OpenBao (formerly Vault) deployments). But maybe the reality of the world is different.

ui/app/routes/vault/cluster.js
```diff
@@ -130,7 +130,8 @@ export default Route.extend(ModelBoundaryRoute, ClusterRoute, {
 
   actions: {
     error(e) {
-      if (e.httpStatus === 503 && e.errors[0] === 'Vault is sealed') {
+      if (e.httpStatus === 503) {
+        e.errors[0] = 'OpenBao is sealed'
```
driif: I don't think we should remove the error message check from this statement.

l-with: Why? Doesn't this make the UI depend on the error strings of the backend instead just using the error code.

driif: @l-with The HTTP status code 503 means that the server is currently unable to handle the incoming requests, It doesn’t necessarily mean that OpenBao is sealed.

l-with: @driif Isn't the API endpoint /sys/health used for this? The [documentation](https://openbao.org/api-docs/system/health/#read-health-information) states that 503 is delivered when sealed. (I did not dive deeper in the UI structure.)

driif: @l-with Thank you for sharing this. In my opinion, if a 503 always means "is sealed," I would still check both for software maturity and rely also on the error message. I'm also confident that, for compatibility reasons, we are not going to change the error message in this case.

satoqz (approved):
LGTM, I think it makes sense to update the message when it is truly UI only and keep the legacy constant in the background.

There are also a few locations within the `commands/` "UI" that produce similar messages that should be safe to replace.

ui/app/routes/vault/cluster.js
```diff
@@ -131,6 +131,7 @@ export default Route.extend(ModelBoundaryRoute, ClusterRoute, {
   actions: {
     error(e) {
       if (e.httpStatus === 503 && e.errors[0] === 'Vault is sealed') {
+        e.errors[0] = 'OpenBao is sealed'
```
driif: I really appreciate your willingness to contribute to OpenBao, but after testing these changes, I didn't notice that it fixes the issue described.

driif: minor nitpick - the linter also suggests adding a ';' at the end of the line.

satoqz: Right... This is probably just an edge case for a specific route that happens to check the literal string, but other places in the UI forward the error as-is? My bad, I assumed this was tested.

l-with:
I dived deeper into the UI structure. The string “Vault” in the message "Vault is sealed" results from the [template for unseal](https://github.com/openbao/openbao/blob/c52795c1ef746c7f2c510f9225aa8ccbbd44f9fc/ui/app/templates/vault/cluster/unseal.hbs#L10)
`{{capitalize this.model.name}}`.
I simply replaced this. A more sophisticated approach would be to refactor the routing and ... in the UI (openbao instead vault), but simply capitalize the model name would not be sufficient, since OpenBao ist camel case.

image

l-with:
BTW, I apologize for my frequent rebase and force push because of my wrong handling of signed off.

JanMa (approved):
LGTM. Thanks @l-with

cipherboy:
Yeah, thanks @l-with! Appreciate you taking the time to figure this one out :-)

driif:
@l-with Thank you for fixing that

PR #1656, opened 2025-08-07
RFC - Efficient Search Components  (WIP)

Gabrielopesantos:
RFC

Resolves #

### DIFF
--- a/website/content/docs/rfcs/efficient-search-components.mdx
+++ b/website/content/docs/rfcs/efficient-search-components.mdx
@@ -184,14 +184,14 @@ The B+ tree implementation uses nodes as the fundamental building blocks. Each n
 
 ```go
 type Node struct {
-	ID     string `json:"id"`
-	IsLeaf bool   `json:"isLeaf"`
-	Keys []string `json:"keys"`
-	Values      [][]string `json:"values"`      // Only for leaf nodes
-	ChildrenIDs []string   `json:"childrenIDs"` // Only for internal nodes
-	ParentID    string     `json:"parentID"`
-	NextID      string     `json:"nextID"`     // ID of the next leaf node
-	PreviousID  string     `json:"previousID"` // ID of the previous leaf node
+	ID     		string 		`json:"id"`
+	IsLeaf 		bool   		`json:"isLeaf"`
+	Keys 		[]string 	`json:"keys"`
+	Values      [][]string	`json:"values"`      // Only for leaf nodes
+	ChildrenIDs []string	`json:"childrenIDs"` // Only for internal nodes
+	ParentID    string		`json:"parentID"`
+	NextID      string		`json:"nextID"`     // ID of the next leaf node
+	PreviousID  string		`json:"previousID"` // ID of the previous leaf node
 }
 
 // Core node operations (Does not match exactly the implementation)
@@ -280,6 +280,9 @@ As for the indexes available, they should be provided through a specific endpoin
 
 ## Rationale and Alternatives
 
+### Node Keys
+The current implementation uses a simple string-based key for each node, which is sufficient for the initial use case. However, this could be extended to support more complex key structures in the future. Strings keys has the limitation of not ordering integer values.
+
 ### Concurrency Control
 
 The current implementation uses tree-level locking with a single `RWMutex` per B+ tree instance. This approach was chosen for its simplicity and correctness guarantees, but it can become a performance bottleneck under high concurrent workloads.
@@ -299,7 +302,10 @@ The tree-level locking approach was selected for the initial implementation to e
 
 ### Storage
 
-The current storage system has the limitation of a one-to-one mapping between keys and nodes. This means that each key can only have one node associated with it, which can be a bottleneck when dealing with large amounts of data. To overcome this limitation, a new storage system that supports multiple nodes per key should be implemented. A page-based storage system can be used to achieve this.
+The current storage system has the limitation of a one-to-one mapping between keys and nodes. This means that each key can only have one node associated with it, which can be a bottleneck when dealing with large amounts of data. To overcome this limitation, a new storage system that supports multiple nodes per key should be implemented.
+
+**Alternative Considered: Page-Based Storage**
+A page-based storage system would allow multiple nodes to be associated with a single key, enabling more efficient storage and retrieval of data. This would require a more complex implementation but would provide better performance for large datasets. The current implementation uses a simple key-value store, which is sufficient for the initial use case but may need to be replaced in the future as the system scales.
 
 ## Downsides
 Code complexity and additional load on the system and storage when building indexes. It is suggested to make index building opt-in and also tunable so that system administrators can disable index building for mounts if desired.
--- a/website/content/docs/rfcs/efficient-search-components.mdx
+++ b/website/content/docs/rfcs/efficient-search-components.mdx
@@ -1,0 +1,326 @@
+# RFC - Components for efficient search over encrypted data
+
+## Summary
+
+OpenBao’s current search functionality is constrained by its key-value storage layer, which only supports constant-time retrieval when using an item’s identifier. As a result, searching based on other metadata (such as keys in the PKI engine or secrets in the KV engine) requires a linear scan through the dataset, which becomes inefficient at scale. This RFC proposes the introduction of a set of constructs designed to build indexes on this metadata, thereby enabling more efficient search operations. Specifically, the solution leverages a B+ tree index structure, which reduces the number of elements that need to be examined during a query through its balanced, and logarithmic search time.
+
+Specific endpoints will be introduced to supporting querying over the indexed data.
+
+## Problem Statement
+
+OpenBao’s storage layer provides two primary data retrieval operations:
+
+- get – Fetches an item by its identifier.
+- list – Retrieves all items or batches of a given type.
+
+However, neither operation is optimized for searching items based on non-identifier attributes. This limitation is manageable for engine backends that store minimal data but becomes problematic for others, such as KV and PKI, where efficient attribute-based lookups are desired.
+An issue with larger deployments of upstream and now OpenBao has always been performing searches across stored data. Common use cases include:
+Key Management (PKI) – Searching for cryptographic keys based on metadata (e.g., expiration date, issuer, subject components, or usage constraints).
+Static Secrets (KV) - Searching for secrets by content or metadata (e.g., finding stored passwords or secrets matching some metadata criteria)
+The lack of an efficient search mechanism forces users to retrieve large datasets and filter results client-side, leading to performance bottlenecks, increased API load, and degraded user experience. Addressing this limitation would improve scalability and usability across multiple storage backends, albeit by shifting some of the load to the server.
+An alternative is leveraging third party tools to create an external index(es) via the audit logs. This approach does, however, leak sensitive information ([not directly](https://openbao.org/docs/audit/#sensitive-information), unless log_raw=true is enabled).
+
+## User-facing Description
+
+Plugin authors would be responsible for creating new API endpoints to perform search operations and control user experience around search parameters. In conjunction with [LIST filtering](https://github.com/openbao/openbao/issues/769), this would allow searches over only user-accessible paths. Operations on engines which use this search engine would incur incrementally more cost to update the B+ trees, which can be done transactionally inline with the request or subsequently via dedicated indexing. But, retrieval of matches via search would be much more efficient than performing LIST or SCAN operations and fetching all matching results.
+
+This would implement a new custom operation, SEARCH, which could be separately ACL'd on paths which implement search APIs. Searching should respect pagination like paginated list endpoints, to reduce large volumes of queries.
+
+Separate RFCs would discuss how to implement this in given plugins. This RFC only discusses internal design semantics.
+
+## Technical Description
+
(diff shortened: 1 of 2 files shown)


website/content/docs/rfcs/efficient-search-components.mdx
```diff
@@ -0,0 +1,341 @@
...
+	…
+}
+
+func NewIndexManager(...) (*IndexManager, error)
+
+// Initialize loads into memory the existing indexes.
+func (im *IndexManager) Initialize(ctx context.Context, storage bptree.Storage) error
+// DeleteIndex marks an index for deletion, purging all data bound to it.
+func (im *IndexManager) DeleteIndex(ctx context.Context, storage bptree.Storage, indexName string) error
+// GetIndex retrieves the index with the provide identifier
```
cipherboy, line 59: Seems like we might also want `ListIndex(...)` to clean up any trailing indices?

website/content/docs/rfcs/efficient-search-components.mdx
```diff
@@ -0,0 +1,341 @@
...
+	TreeID  string `json:"tree_id"` // Unique identifier for multi-tree storage
+	Order   int    `json:"order"`   // Maximum number of children per node (default: 32) (Can't be updated...)
+	Version int    `json:"version"` // Configuration version for schema evolution
+}
+```
+
+**Attributes:**
+
+- `TreeID`: Unique identifier enabling multiple named trees within a single storage backend. Used to namespace storage paths as `<TreeID>/nodes/<nodeID>`, `<TreeID>/root`, and `<TreeID>/config`
+- `Order`: Defines the maximum number of children per internal node (minimum 3, default 32). This directly affects tree height and performance characteristics
```
cipherboy, line 103: I forget, `32` comes from bbolt or somewhere else?

website/content/docs/rfcs/efficient-search-components.mdx
```diff
@@ -0,0 +1,341 @@
...
+{
+  "op": "intersect",
+  "queries": [
+    { "key": "...", "value": "..." },
+    { "key": "...", "value": "..." }
+  ]
+}
+```
+
+Each query may optionally include a `match` or `prefix` field to control how values are compared. "exact" is the default and can be omitted.
```
cipherboy:
```suggestion
Each query may optionally include a `match` field to control how values are compared. "exact" is the default and can be omitted, else `prefix` can be used to signify that the value is a prefix of desired matches.
```

website/content/docs/rfcs/efficient-search-components.mdx
```diff
@@ -0,0 +1,341 @@
...
+No security implications were identified.
+
+## User/Developer Experience
+
+With these constructs available, an OpenBao developer will be able to dynamically configure and build indexes on the data persisted in storage.
+No user facing changes are expected with this RFC as it only introduces the constructs to build the indexes.
+
+## Unresolved Questions
+
+- How to keep track of the current progress of an index that is being incrementally built;
```
cipherboy, line 334: I think this would have to be in the plugin. They'd add an `indexed: <bool>` on the items themselves or an `indexed: <int>` for versioned items (with async re-indexing). Linear scan is going to be the only approach, +/- a queue system for handling async updates.

website/content/docs/rfcs/efficient-search-components.mdx
```diff
@@ -0,0 +1,341 @@
...
+- Purging or removing all nodes associated with a key from multiple trees;
```
cipherboy, line 335: I think you have to just use a lock/transaction across the index to do this?

website/content/docs/rfcs/efficient-search-components.mdx
```diff
@@ -0,0 +1,341 @@
...
+- Index rebuilding if any configuration is changed, such as order (currently not supported); Might be easier to simply rebuild the index from scratch.
+- Removal of an index;
```
cipherboy, line 337: I think the `ClearKeyView(...)` helper will be useful here (if I got the name correct from memory).

website/content/docs/rfcs/efficient-search-components.mdx
```diff
@@ -0,0 +1,341 @@
...
+## User/Developer Experience
+
+With these constructs available, an OpenBao developer will be able to dynamically configure and build indexes on the data persisted in storage.
+No user facing changes are expected with this RFC as it only introduces the constructs to build the indexes.
+
+## Unresolved Questions
+
+- How to keep track of the current progress of an index that is being incrementally built;
+- Purging or removing all nodes associated with a key from multiple trees;
+- Index rebuilding if any configuration is changed, such as order (currently not supported); Might be easier to simply rebuild the index from scratch.
```
cipherboy, line 336: I concur here.

website/content/docs/rfcs/efficient-search-components.mdx
```diff
@@ -0,0 +1,341 @@
...
+- Removal of an index;
+
+## Related Issues
+
+OpenBao:
+- https://github.com/orgs/openbao/discussions/1400
+
+GitHub:
```
cipherboy:
```suggestion
Vault:
```

cipherboy (approved):
Really excited to see this! Mostly just minor nits.

The only other question I have is have you considered substring? Looking at the serial number example, sometimes people are lazy and type e.g., just the middle of a serial number.

website/content/docs/rfcs/index.mdx
```diff
@@ -44,3 +44,4 @@ Steering Committee.
    into OpenBao.
  - [Using CEL Roles For JWT/OIDC Auth](/docs/rfcs/cel-jwt), adds support for CEL roles in JWT Auth
    Landed in [769](https://github.com/openbao/openbao/pull/769)
+ - [Efficient Search Components](/docs/rfcs/efficient-search-components), adds the necessary components to build indexes on data ...
```
cipherboy: This should go towards the top in the `Accepted` section, not the `Landed` section yet.

draggeta:
This would fix the biggest gripe I think my colleagues currently have with OB. Thank you for looking into this.

website/content/docs/rfcs/efficient-search-components.mdx
Gabrielopesantos: Not really and to be honest I didn't think much about it. From what I have read it is somewhat dependent on the storage layer used. For example, page/block based implementations is max number of children that fit in a page size.

If I understood correctly, bbolt doesn't operate with "Order" values but with page sizes directly, 4 KiB (by default). Then anode can have as many elements as will fit into that 4 Kib page.

website/content/docs/rfcs/efficient-search-components.mdx
Gabrielopesantos: Yes, we could also have something like that. This part hasn't been too flushed yet but some kind of list operations might be useful.
Going to add.

website/content/docs/rfcs/efficient-search-components.mdx
Gabrielopesantos: Not sure if  I understood what you are proposing here. The idea is for the plugin to track the items indexed but wouldn't that be iterating over all items and check if they have already been indexed (by checking the label, tag)? I was thinking more along the lines of having a milestone but don't know how would that be feasible.

website/content/docs/rfcs/efficient-search-components.mdx
Gabrielopesantos: Yes, now that I think better about it something like that would work. It's somewhat the same as querying multiple trees and joining their results.

cipherboy:
@Gabrielopesantos Can you add this to `sidebars.ts` as well? ty!

website/content/docs/rfcs/efficient-search-components.mdx
cipherboy: I think my point is, this RFC doesn't really need to answer this. It is on the plugin (which uses search) to handle indexing, not on the generic search library IMHO.

website/content/docs/rfcs/efficient-search-components.mdx
Gabrielopesantos: Well, I actually want some feedback here and probably on the whole storage proposal. Basically, the current implementation is wrapper around the `logical.Storage` interface that is required to be passed in on all bptree functions instead of only being set initializing a tree instance. I'm not quite sure I like this approach but was the best I could think of for supporting multiple operations on multiple trees on a single transaction. It works well for this use cases but introduces some issues, like shared cache for all indexes and harder to keep track changed nodes on an operation requiring all changes to be committed.

Gabrielopesantos:
Right, so that's what was missing. I thought it was added automatically or required some kind of match between filename and labels.

website/content/docs/rfcs/efficient-search-components.mdx
cipherboy: @Gabrielopesantos Right, but you can create a single `logical.Transaction` and pass it into multiple index operations (as the `logical.Storage` interface) to do removals of a single key across all transactions.

Gabrielopesantos: Can you clarify what do you mean with that?

cipherboy: Say I had multiple indices over a KVv2 entry:

```json
{
 "value": "...",
 "author": "...",
 "security_level": "..."
}
```

and I call `DELETE /data/my_entry`, causing me to want to remove `my_entry` from all three indices above. I'd do:

```go
txn, err := req.Storage.BeginTransaction()
storage, err := bptree.NewStorage(ctx, txn)

for indexName, indexValue := range getIndicesInEntry("my-entry") {
  idx := b.idxMgr.GetIndex(ctx, storage, indexName)
  idx.Delete(ctx, storage, indexValue)
  // or
  idx.Delete(ctx, storage, "my-entry")
}

txn.Commit()
```

Gabrielopesantos: Yes, the current implementation would work something like that but has its limitations. If operating with an "external" transaction, we don't know whether or not an operation was successful until is effectively committed, which is out of scope of the tree storage. For that reason, I have decided to disable caching if operating on an logical.storage transaction. I'm working on some storage improvements, that will be pushing the PoC branch soon. Worth updating the spec? If not, we should be good to go.

Answering the following here too:
> The only other question I have is have you considered substring? Looking at the serial number example, sometimes people are lazy and type e.g., just the middle of a serial number.

Yes, I have and adding support shouldn't be hard. Can be added for the final version.

cipherboy: @Gabrielopesantos Right, I had really wished the physical cache system allowed snapshotting, but alas. :(  What we ended up doing was creating a new (smaller) cache for the transaction, so at least inside the transaction you can take advantage of the benefits. 

Which cache library are you using? If you can get a soft-clone, that'd be ideal. Even if you discard the state afterwards, that'd be fine.

Gabrielopesantos: > Which cache library are you using?

The one created in  `sdk/lru/lru.go`, its just a wrapper around `github.com/hashicorp/golang-lru/v2`.

> If you can get a soft-clone, that'd be ideal. Even if you discard the state afterwards, that'd be fine.

Yes, something like that should be feasible. Going to look into that.

Gabrielopesantos: Ok, I have finished the changes in the storage layer. Besides the stuff to make the NodeStorage work with external transactions, I have introduced a new component in the storage to keep track of changed nodes, dirty tracker, on operations. With this, we aren't hitting the logical storage until a transaction (internal) is commited or we programatically do so. Not sure if this was confusing, the changes are in the PR if you want to have a look.

Also, there are still a few pending TODO but the PR is probably ready to start to be reviewed, there's is quite a lot in there.

cipherboy: @Gabrielopesantos said:

> The one created in `sdk/lru/lru.go`, its just a wrapper around `github.com/hashicorp/golang-lru/v2`.

I think we'll need to fork this to get it to behave how we want. Do you want to add it into SDK?

Gabrielopesantos: Can you clarify the best phrase? With the update approach we already get a per external transaction empty cache, it doesn't get anything from any existing storage (which could be useful) and will not be merged back into the "parent" at the end. I'm thinking of making a full copy but that might be too expensive. We can probably make it work in some other way where we get a reference to the existing storage cache but don't modify it, then operations just check if something is or isn't there and fallback to the transaction cache.

What would be the benefit of fork the LRU implementation here?  And yes, I could do it but probably on a separate PR.

cipherboy: @Gabrielopesantos Ah, actually I might take a stab at this. It is more complicated than I thought, will send you a PR to review in a bit.

Gabrielopesantos: If you are took busy with other stuff, feel free to leave me the details and I can take a  look.

cipherboy:
@Gabrielopesantos Terribly sorry about the delay. Could you rebase and we'll get this merge and start working on the feature? Thanks!

Gabrielopesantos:
> @Gabrielopesantos Terribly sorry about the delay. Could you rebase and we'll get this merge and start working on the feature? Thanks!

Hey @cipherboy, updated and CI should be passing now. Had some trouble fixing the absolute links check.

cipherboy:
Thanks @satoqz! Good to have buy-in from others. I'll go ahead and resolve the open discussions and we can go and merge. :D Thank you @Gabrielopesantos for the patience!

PR #1657, opened 2025-08-07
Fix license dependencies generation to work with the project go version

DrDaveD:
Fix go license dependency generation to work with the currently selected go version in `.git-version`.  This fixes errors that were happening when testing it after a recent update to go 1.24.6.  Apparently it just happened to have a matching go version during previous testing.


cipherboy (approved):
This should fix it. Thanks @DrDaveD!

For the benefit of others, in https://github.com/openbao/openbao-nightly/actions/runs/16808192332, the build failed. From what I can tell, the _default_ Go toolchain installed by GitHub on actions differed from what we just updated to (because ours was released 2 days ago to fix other CVEs and GitHub hasn't yet updated their base images yet).

This meant the `go-license` tool was compiling with an older one, causing these stdlib version mismatches which caused issues. By pulling Go version explicitly, we should ensure the compiled `go-license` matches our preferred one and should allow it to succeed.

In the event this doesn't work, https://github.com/openbao/openbao-nightly/actions/runs/16809151283 validated that we can disable the tarball conditionally and the release will still work. So we can always re-run from the same tag and just not build this tarball if things go haywire this time around.

PR #1666, opened 2025-08-08
Always do git checkout in release-dist job

DrDaveD:
This is to correct a [failure to upload](https://github.com/openbao/openbao/actions/runs/16821517603/job/47650673183) that happened during the 2.3.2 release.


cipherboy:
Thanks @DrDaveD !

PR #1667, opened 2025-08-08
Fix some ineffassign warnings

phil9909:

### DIFF
--- a/vault/seal_autoseal_test.go
+++ b/vault/seal_autoseal_test.go
@@ -210,7 +210,7 @@ func TestAutoSeal_HealthCheck(t *testing.T) {
 
 	asu := strings.Join(autoSealUnavailableDuration, ".") + ";cluster=" + core.clusterName
 	tries := 10
-	for tries = 10; tries > 0; tries-- {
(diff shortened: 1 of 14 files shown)


vault/seal_autoseal_test.go
```diff
@@ -210,7 +210,7 @@ func TestAutoSeal_HealthCheck(t *testing.T) {
 
 	asu := strings.Join(autoSealUnavailableDuration, ".") + ";cluster=" + core.clusterName
 	tries := 10
-	for tries = 10; tries > 0; tries-- {
+	for ; tries > 0; tries-- {
```
satoqz, line 213: Can the initial declaration be moved into the for statement?

phil9909: No, after the loop, we do:

```go
if tries == 0 {
		t.Fatalf("Expected value metric %s to be non-zero", asu)
}
```

I was inclined to refactor this to use `require.Eventually`, but I wanted to keep the changes minimal.

driif (approved):

satoqz:
Merging, thanks @phil9909!

PR #1668, opened 2025-08-08
Skip Container tests by default

cipherboy:
Many environments lack container execution environments; let's disable them by default and conditionally re-enable with the `BAO_RUN_CONTAINER_TESTS=true` variable.

### DIFF
--- a/sdk/helper/docker/testhelpers.go
+++ b/sdk/helper/docker/testhelpers.go
@@ -18,6 +18,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"testing"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
@@ -33,6 +34,7 @@ import (
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/docker/go-connections/nat"
 	"github.com/hashicorp/go-uuid"
+	"github.com/openbao/openbao/api/v2"
 )
 
 const DockerAPIVersion = "1.40"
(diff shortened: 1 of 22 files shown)


sdk/helper/docker/testhelpers.go
```diff
@@ -909,3 +911,9 @@ func (d *Runner) GetNetworkAndAddresses(container string) (map[string]string, er
 
 	return ret, nil
 }
+
+func CheckSkipContainerTests(t *testing.T) {
+	if value := api.ReadBaoVariable("BAO_RUN_CONTAINER_TESTS"); value != "true" {
```
satoqz: Thoughts on using `strconv.ParseBool` here?

> ParseBool returns the boolean value represented by the string. It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False. Any other value returns an error.

cipherboy:
@satoqz Sorry, that was a mess, fixed. I'll file an issue for `BAO_ACC=1` test fixing.

satoqz (approved):
Ran `go test ./...` for all the modules with

- No container images present locally
- `BAO_RUN_CONTAINER_TESTS` unset

and confirmed that tests do not pull any container images.

I do wonder if it wouldn't be easier to centrally skip in `NewTestDockerCluster(...)` since that already takes a `testing.T`?

cipherboy:
@satoqz Several of the places execute when `NewTestDockerCluster(...)` is not called, such as `prepareLDAPTestContainer` and variants. That helper is only used to spin up a new OpenBao cluster, it doesn't help with the various ancillary tests directly using other containers. My 2c.

satoqz:
@cipherboy totally, I was just thinking that we can simplify it for those cases that do use `NewTestDockerCluster(...)`, so you're just calling one test helper. Fine with either approach, merge if you'd prefer it remains explicit in all cases.

driif (approved):

cipherboy:
@JanMa Mind ACKing this one on behalf of @openbao/openbao-org-maintainers? Thanks!

cipherboy:
I'm closing this one as its conflicted again. I think it provides some value (in that container tests don't run by default), but also is kinda useless. :shrug: 

It seems best carried internally if this is a problem for downstream organizations.

PR #1675, opened 2025-08-11
Drop pre-v1.0 Vault database upgrade check

wslabosz-reply:
Drops `(*databaseBackend) roleAtPath()` function upgrade check introduced in `v0.10` of Vault due to `statments` typo in user role schema, also removes the upgrade test.

### DIFF
--- a/builtin/logical/database/backend.go
+++ b/builtin/logical/database/backend.go
@@ -13,13 +13,11 @@ import (
 
 	"github.com/armon/go-metrics"
 	log "github.com/hashicorp/go-hclog"
-	"github.com/hashicorp/go-secure-stdlib/strutil"
 	"github.com/hashicorp/go-uuid"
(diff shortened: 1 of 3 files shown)


cipherboy:
@wslabosz-reply Could you add a changelog prior to merging? Thanks!

changelog/1675.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:change
+bultin/logical/database: Dropped obsolete upgrade check in `roleAtPath()` function introduced in `v0.10` of Vault.
```
satoqz:
```suggestion
database: Drop obsolete upgrade check in `roleAtPath()` function introduced in `v0.10` of Vault.
```

Historically we've used just the `database` scope for this package.

satoqz:
CI failure unrelated, merging!

PR #1676, opened 2025-08-11
Lessen dependency on strutil library with removal of deprecated sdk helpers

wslabosz-reply:
This PR:
- changes every usage of `strutil.StrListContains()` to `slices.Contains()`
- removes `Initialize(context.Context, map[string]interface{}, bool) error` from `ConnectionProducer` as as it was deprecated with v0.12 of Vault
   - hence removing`(c *SQLConnectionProducer) Initialize` method
- removes `sdk/helper/base62`, `sdk/helper/mlock`, `sdk/helper/parseutil`, `sdk/helper/password`, `sdk/helper/strutil`, `sdk/helper/tlsutil` and changes their usage to equivalents from `"github.com/hashicorp/go-secure-stdlib/xxx"` packages


cipherboy (approved):
Just since we are changing `sdk` mind throwing a changelog on this one too? Thanks!

wslabosz-reply:
@JanMa 
I think it does make sense, but I'd first look if our fork got outdated since it started

satoqz:
This fails in a non-obvious way once merged (`git fetch upstream pull/1676/merge`), I'd hope that a rebase fixes it @wslabosz-reply.

satoqz:
https://github.com/openbao/openbao/pull/1667/files#diff-35b96bbe34071593de67e53b4523d7b54e7cf002981e203418fa344305693221 is the cause

satoqz (approved):
Looks good now, merging

PR #1678, opened 2025-08-12
nilnil refactor

KrzysztofKornalewski-Reply:
Related Issues:
https://github.com/openbao/openbao/issues/1579

### DIFF
--- a/api/kv_v2.go
+++ b/api/kv_v2.go
@@ -543,7 +543,7 @@ func extractVersionMetadata(secret *Secret) (*KVVersionMetadata, error) {
 	var metadata *KVVersionMetadata
 
 	if secret.Data == nil {
-		return nil, fmt.Errorf("metadata: fot found")
+		return nil, fmt.Errorf("metadata: not found")
 	}
 
 	// Logical Writes return the metadata directly, Reads return it nested inside the "metadata" key
--- a/api/logical.go
+++ b/api/logical.go
@@ -180,16 +180,16 @@ func (c *Logical) ListWithContext(ctx context.Context, path string) (*Secret, er
 		switch parseErr {
 		case nil:
 		case io.EOF:
-			//nolint:nilnil
-			return nil, nil
+			// nil, nil is fine because the caller doesn't expect a particular error in this case. Handling it without significant refactoring of the test and logic can lead to a loss of backward compatibility.
+			return nil, nil //nolint:nilnil
 		default:
 			return nil, parseErr
 		}
 		if secret != nil && (len(secret.Warnings) > 0 || len(secret.Data) > 0) {
 			return secret, nil
 		}
-		//nolint:nilnil
-		return nil, nil
+		// nil, nil is fine because the caller doesn't expect a particular error in this case. Handling it without significant refactoring of the test and logic can lead to a loss of backward compatibility.
+		return nil, nil //nolint:nilnil
 	}
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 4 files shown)


api/kv_v2.go
```diff
@@ -543,7 +543,7 @@ func extractVersionMetadata(secret *Secret) (*KVVersionMetadata, error) {
 	var metadata *KVVersionMetadata
 
 	if secret.Data == nil {
-		return nil, nil
+		return nil, fmt.Errorf("metadata: fot found")
```
satoqz:
```suggestion
		return nil, fmt.Errorf("metadata: not found")
```

api/logical.go
```diff
@@ -180,13 +180,15 @@ func (c *Logical) ListWithContext(ctx context.Context, path string) (*Secret, er
 		switch parseErr {
 		case nil:
 		case io.EOF:
+			//nolint:nilnil
 			return nil, nil
```
satoqz: You can inline these like so:

```suggestion
			return nil, nil //nolint:nilnil
```

It would be nice to have an explanatory comment on why `nil, nil` is fine in those cases that are ignored via `//nolint`, something like:

```go
// nil, nil is fine because callers expect... historically this has been...
return nil, nil //nolint:nilnil
```

api/kv_v2.go
satoqz, line 1: These seem fine to me and rather low-risk, but certainly a minor behavior change. Would want to hear other's thoughts here.

vault/barrier_aes_gcm.go
satoqz, line 1: This looks like a pure improvement to me. I don't see any callers that expect to handle a `nil` cipher here :+1:

satoqz (approved):

vault/barrier_aes_gcm.go
phyrog: The only nit-picks I have here are:
- unnecessarily using `fmt.Errorf` instead of `errors.New`, there is no formatting going on here
- the errors could be stored in global vars, so they can be used with `errors.Is` and `errors.As`

api/kv_v2.go
phyrog: Not sure if the behavior change is minor here. I checked the callsites of `extractVersionMetadata`, and there are early returns if the `err` is not `nil`, so instead of returning a `*KVSecret` with `VersionMetadata` set to `nil`, `*KVv2.Put` would return `nil` and an error.

https://github.com/openbao/openbao/blob/2578bbf0485c92e02f1d5e5fc934df3b2f7c0af2/api/kv_v2.go#L247-L256

That seems like a larger change to me (though it might actually be the intended behavior here).

The change for `extractFullMetadata` looks ok to me, there is a place where the returned value is accessed without checking it for nil (only checking the error), which would result in a panic: https://github.com/openbao/openbao/blob/2578bbf0485c92e02f1d5e5fc934df3b2f7c0af2/api/kv_v2.go#L178-L183

Overall I am a bit concerned about the lack of test failures due to these changes.

vault/barrier_aes_gcm.go
cipherboy: I don't know if the latter will ever be compared with that but I concur.

api/kv_v2.go
cipherboy: I concur fwiw; I'd mark this as lint-only and keep `nil, nil`

cipherboy:
@KrzysztofKornalewski-Reply If you address these, happy to review and merge.

cipherboy:
Closing based on discussion on Zulip: [#openssf-openbao-wg > PRs to review @ 💬](https://linuxfoundation.zulipchat.com/#narrow/channel/574533-openssf-openbao-wg/topic/PRs.20to.20review/near/581881966)

`nil, nil` linter has been pain and I concur Vault wasn't originally designed with it in mind, so OpenBao would have a hard time getting `nil, nil` clean. :-D

PR #1680, opened 2025-08-13
use "bit clear" to simplify executable bit strip

phil9909:
#1651 was merged before I could mention this.

@satoqz  creating an all ones int and xor it with a mask is the same as the "bitwise complement" `^` unary operator (in go it even uses the same symbol as xor, most other languages use `~`)
Go even has a "bit clear" binary operator `&^`, a combination of "bitwise complement" (`^`) and "bitwise and" (`&`)

But I like your ASCII art :+1:

### DIFF
--- a/builtin/audit/file/backend.go
+++ b/builtin/audit/file/backend.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/fs"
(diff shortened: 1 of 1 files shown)


satoqz:
Merging as this is a minor change to the PR we passed yesterday, with no effect on behavior.

PR #1682, opened 2025-08-14
rework cache invalidation error handling

phil9909:
Refactor error handling for cache invalidation and actually implement core restarting in case of fatal errors (e.g. audit table reload failure).

As discussed here: https://github.com/openbao/openbao/pull/1596#discussion_r2243339441

This PR is best reviewed with [white-space ignored](./files?w=1)

See #1550
See #1528 


vault/audit.go
```diff
@@ -410,6 +410,20 @@ func (c *Core) setupAudits(ctx context.Context) error {
...
+func (c *Core) invalidateAudits() {
+	go func() {
+		err := c.reconcileAudits(reconcileAuditsRequests{
+			ctx:       c.activeContext,
+			readonly:  true,
+			isInitial: false,
+		})
+		if err != nil {
+			c.logger.Error("unable to invalidate audits, restarting core", "error", err.Error())
+			c.restart()
```
suprjinx, line 422: I assume this is rare, but core restart seems like an intensive process. I'm wondering if there's something short of this to try first -- like just restarting the cache layer in question?

vault/ha.go
```diff
@@ -388,87 +388,102 @@ func (c *Core) StepDown(httpCtx context.Context, req *logical.Request) (retErr e
...
-			c.waitForLeadership(newLeaderCh, manualStepDownCh, leaderStopCh)
-			return nil
-		}, func(error) {
-			close(leaderStopCh)
-			c.logger.Debug("shutting down leader elections")
-		})
+		// Start all the actors
+		err := g.Run()
+		if err != nil {
+			c.logger.Debug("unexpected error in runStandby", "error", err.Error())
```
suprjinx: Should this be Error level log?

vault/audit.go
phil9909: I'm not sure if it is worth the added complexity.
What might make sense would be a simple retry (e.g. to compensate network hiccups when using PostgreSQL storage)

vault/ha.go
phil9909: yes, done.

vault/core_cache_invalidate.go
```diff
@@ -12,18 +12,14 @@ import (
...
-
 func (c *Core) Invalidate(key string) {
 	ctx, cancel := context.WithTimeout(c.activeContext, 2*time.Second)
 	defer cancel()
 
 	err := c.invalidateInternal(ctx, key)
 	if err != nil {
-		c.processInvalidationFailure(key, err)
+		c.logger.Error("cache invalidation failed, restarting core", "key", key, "error", err.Error())
+		c.restart()
```
cipherboy, line 22: Remember, we only want critical subsystems to restart. We don't want to do this more generally. I'm confused as well why `invalidateAudits(...)` has its own call to `c.restart(...)`?

vault/audit.go
cipherboy: @suprjinx In this case, we want to restart the core because the _audit_ subsystem failed to invalidate correctly. That's a much bigger failure than a random plugin and suggests we might miss e.g., new HMAC keys or changed config or something. So I think restarting the core is an appropriate response here. 

This should only occur on standby nodes, which shouldn't get leadership after restart (assuming the existing leader which wrote this change is stable).

vault/core_cache_invalidate.go
```diff
@@ -12,18 +12,14 @@ import (
 	"github.com/openbao/openbao/vault/quotas"
 )
 
-func (c *Core) processInvalidationFailure(key string, err error) {
-	c.logger.Error("cache invalidation failed", "key", key, "error", err.Error())
-	// TODO(phil9909): It's not save to continue, restart the core
-}
-
 func (c *Core) Invalidate(key string) {
```
cipherboy, line 15: Thinking about it some more, maybe we just want `core.Invalidate(...)` to run in a goroutine. While the one in the Raft layer is cute, having Core handle it means that it can wait to grab locks as necessary (in case someone holds a write lock when we need a read lock)... though we're back to the out-of-order invalidation discussion... Hm. :thinking: 

Discussions for later, not relevant to this PR.

vault/ha.go
```diff
@@ -388,87 +388,102 @@ func (c *Core) StepDown(httpCtx context.Context, req *logical.Request) (retErr e
 	return retErr
 }
 
+func (c *Core) restart() {
+	c.restartCh <- struct{}{}
```
cipherboy: We want this to be non-blocking, right? So we should wrap this in a switch with a default statement that just warns about an active restart? 

Otherwise, if we get multiple invalidation which fail in critical systems, we'll be in a restart cycle as this will block until the restart is processed...

vault/ha.go
```diff
@@ -388,87 +388,102 @@ func (c *Core) StepDown(httpCtx context.Context, req *logical.Request) (retErr e
...
+func (c *Core) runStandby(doneCh, manualStepDownCh, stopCh chan struct{}, restartCh chan struct{}) {
 	defer close(doneCh)
 	defer close(manualStepDownCh)
 	c.logger.Info("entering standby mode")
 
-	// wipe any existing mount tables
-	if err := c.preSeal(); err != nil {
-		c.logger.Error("pre-seal teardown failed", "error", err)
-	}
+	for restart := true; restart; {
```
cipherboy, line 429: I do think this is better than `standbyDoneCh`, but it is probably worth writing a note as to why.

vault/core_cache_invalidate.go
phil9909: `invalidateInternal` should only return fatal errors (as all other errors are handled by the subsystems).

Currently, this is iif `namespaceStore.GetNamespace` fails, which currently will fail iff it was marked as invalid/stale and reloading the namespaces from storage failed.
I tried to get rid of all potential errors from `invalidateInternal`, but I just don't know how to better handle namespace lookup errors.

vault/ha.go
phil9909: Sorry, I don't quite understand.
`standbyDoneCh` is used to signal that `runStandby` has exited, correct?

You like this better than which alternative? Returning and calling `runStandby` again from the outside?

vault/ha.go
phil9909: Makes sense: 3d446d0394eac8c8d1740ae61b72ce1b2440d117

vault/core_cache_invalidate.go
cipherboy: I think namespace lookup errors are probably fine. It probably goes back to out-of-order invalidation, tbh.

vault/ha.go
cipherboy: re: 

> `standbyDoneCh` is used to signal that `runStandby` has exited, correct?

I concur, and we should comment that.

> You like this better than which alternative? Returning and calling `runStandby` again from the outside?

Right, `runStandby` could be flatter if it returns a `restart` boolean iff it should be restarted and this loop moved up to its caller.

vault/core_cache_invalidate.go
cipherboy: @phil9909 If we start hitting out-of-order invalidations, I think we could re-queue the earlier invalidations with the hope that namespace gets added if it is a namespace missing error. Could be risky, though, if we have lots of namespaces being added and lots of out-of-order invalidations... 

Luckily namespace additions should be fairly rare. :thinking:

vault/ha.go
phil9909: Sorry, I don't really see how this is worth documenting.

We would just change the call site to a loop `for c.runStandby(...) {}`, move the channel `close` operations outside of `runStandby` and return a restart boolean. This seems like a minor implementation detail to me. Maybe I just don't get it :laughing: 

Could you propose a comment, so I might understand your point.

vault/ha.go
```diff
@@ -388,87 +388,122 @@ func (c *Core) StepDown(httpCtx context.Context, req *logical.Request) (retErr e
...
+		default:
+			return
+		}
+	}
+}
+
 // runStandby is a long running process that manages a number of the HA
 // subsystems.
-func (c *Core) runStandby(doneCh, manualStepDownCh, stopCh chan struct{}) {
+func (c *Core) runStandby(doneCh, manualStepDownCh, stopCh chan struct{}, restartCh chan struct{}) {
```
cipherboy:
```suggestion
//
// doneCh is used to indicate that the standby has finished operating in this invocation
// manualStepDownCh is a channel passed to the leadership acquisition process, to allow interrupting leadership acquisition by this node
// stopCh can be used to stop the standby and shutdown the process
// restartCh can be used to gracefully reset standby node state, reloading all cached information
//
// stopCh and restartCh differ in that the former is terminal and the latter is re-entrant.
func (c *Core) runStandby(doneCh, manualStepDownCh, stopCh chan struct{}, restartCh chan struct{}) {
```

@phil9909 WDYT? In particular, I'm trying to make sure we have some inline documentation in useful places for what these channels do and the differences between them, rather than just accumulating more.

phil9909: Ok, documenting the channels purpose makes sense to me. I understood you wanted to document why we loop inside `runStandby` and not outside of it. 

I took your proposal and added some more details to it (mainly about how the channel will be used i.e. does `runStandby` close it and/or read from it) and added directions to the channels were possible (`chan` vs `<-chan` vs `chan<-`) to use the type system as documentation. See https://github.com/openbao/openbao/commit/fedeb2455c198e954111f0674e8c3b7d0dbecea4

cipherboy:
@phil9909 Looks like there's a data race and then we can merge?

```

2025-09-09T11:19:41.3021028Z Read at 0x00c04c288eb8 by goroutine 622772:
2025-09-09T11:19:41.3021400Z   github.com/openbao/openbao/vault.testCoreRestart.func1()
2025-09-09T11:19:41.3021937Z       /home/runner/work/openbao/openbao/vault/ha_test.go:130 +0x105
2025-09-09T11:19:41.3022375Z   github.com/stretchr/testify/assert.EventuallyWithT.func1()
2025-09-09T11:19:41.3023030Z       /home/runner/go/pkg/mod/github.com/stretchr/testify@v1.10.0/assert/assertions.go:2044 +0xb7
2025-09-09T11:19:41.3023376Z 
2025-09-09T11:19:41.3023522Z Previous write at 0x00c04c288eb8 by goroutine 622085:
2025-09-09T11:19:41.3023903Z   github.com/openbao/openbao/vault.(*Core).postUnseal()
2025-09-09T11:19:41.3024378Z       /home/runner/work/openbao/openbao/vault/core.go:2399 +0x16f
2025-09-09T11:19:41.3024774Z   github.com/openbao/openbao/vault.(*Core).runStandby()
2025-09-09T11:19:41.3025223Z       /home/runner/work/openbao/openbao/vault/ha.go:441 +0x3b0
2025-09-09T11:19:41.3025778Z   github.com/openbao/openbao/vault.(*Core).loadIdentityStoreArtifacts()
2025-09-09T11:19:41.3026380Z       /home/runner/work/openbao/openbao/vault/identity_store_util.go:65 +0xe3
2025-09-09T11:19:41.3026891Z   github.com/openbao/openbao/vault.readonlyUnsealStrategy.unseal()
2025-09-09T11:19:41.3027693Z       /home/runner/work/openbao/openbao/vault/core.go:2372 +0x45a
2025-09-09T11:19:41.3028151Z   github.com/openbao/openbao/vault.readonlyUnsealStrategy.unseal()
2025-09-09T11:19:41.3028671Z       /home/runner/work/openbao/openbao/vault/core.go:2361 +0x3e4
2025-09-09T11:19:41.3029134Z   github.com/openbao/openbao/vault.(*readonlyUnsealStrategy).unseal()
2025-09-09T11:19:41.3029623Z       <autogenerated>:1 +0x6e
2025-09-09T11:19:41.3029987Z   github.com/openbao/openbao/vault.(*Core).postUnseal()
2025-09-09T11:19:41.3030452Z       /home/runner/work/openbao/openbao/vault/core.go:2428 +0x4b9
2025-09-09T11:19:41.3030839Z   github.com/openbao/openbao/vault.(*Core).postUnseal()
2025-09-09T11:19:41.3031290Z       /home/runner/work/openbao/openbao/vault/core.go:2424 +0x464
2025-09-09T11:19:41.3031678Z   github.com/openbao/openbao/vault.(*Core).runStandby()
2025-09-09T11:19:41.3032267Z       /home/runner/work/openbao/openbao/vault/ha.go:441 +0x3b0
2025-09-09T11:19:41.3032708Z   github.com/openbao/openbao/vault.(*Core).unsealInternal.gowrap1()
2025-09-09T11:19:41.3033226Z       /home/runner/work/openbao/openbao/vault/core.go:1924 +0x6b
```

vault/ha.go
```diff
@@ -509,7 +521,7 @@ func (c *Core) runStandby(doneCh, manualStepDownCh, stopCh chan struct{}, restar
 // waitForLeadership is a long running routine that is used when an HA backend
 // is enabled. It waits until we are leader and switches this Vault to
 // active.
-func (c *Core) waitForLeadership(newLeaderCh chan func(), manualStepDownCh, stopCh chan struct{}) {
+func (c *Core) waitForLeadership(newLeaderCh chan func(), manualStepDownCh, stopCh <-chan struct{}) {
```
cipherboy, line 526: I've never seen the arrow used in this way, what does it do?

vault/ha_test.go
```diff
@@ -85,3 +88,46 @@ func TestGrabLockOrStop(t *testing.T) {
...
+	select {
+	case <-activeContextBeforeRestart.Done():
+	case <-time.NewTimer(10 * time.Second).C:
+		t.Fatal("timeout while waiting for context to be cancelled")
+	}
+
+	// a new context should be started
+	require.EventuallyWithT(t, func(t *assert.CollectT) {
+		require.NotEqual(t, activeContextBeforeRestart, c.Cores[core].activeContext)
+		require.NotNil(t, c.Cores[0].activeContext)
```
cipherboy: This won't really work unless you grab locks to read the active context. 

I _think_ `stateLock` is the correct lock.

vault/ha.go
phil9909: It restricts a channel

> The optional <- operator specifies the channel direction, send or receive. If a direction is given, the channel is directional, otherwise it is bidirectional.
> https://go.dev/ref/spec#Channel_types

A receiving channel `<-chan T` can only be read from
A sending channel `chan<- T` can only be sent to or closed

You can assign (or pass as argument) a normal channel `chan T` to either `<-chan T` or `chan<- T`, see https://go.dev/play/p/JUUbVK4fZXu

cipherboy: Oh neat

vault/ha_test.go
phil9909: This was the cause of the race

vault/ha_test.go
```diff
@@ -85,3 +88,50 @@ func TestGrabLockOrStop(t *testing.T) {
...
 	workerWg.Wait()
 }
+
+func TestCoreRestart(t *testing.T) {
+	t.Parallel()
+
+	t.Run("active", func(t *testing.T) {
+		t.Skip("there is a data-race in waitForLeadership: https://github.com/openbao/openbao/blob/5a93ec0549a88516a8a15e96ef74dadac8ed506f/vault/ha.go#L676-L687")
+		testCoreRestart(t, 0)
+	})
```
phil9909, line 98: For now, the restart test is skipped because of  a data-race in the `waitForLeadership` implementation https://github.com/openbao/openbao/blob/5a93ec0549a88516a8a15e96ef74dadac8ed506f/vault/ha.go#L676-L687

If `l.lockOrStop()` returns `false` the code still modifies the state. I tried to fix this, by changing it to always lock and return in the end, if the context was cancelled. But other parts of the shutdown logic seem to rely on the unlocked state and it caused a dead-lock. So fixing this would be a bigger change.
As for now `core.restart()` is only called on standby nodes and the race only happens on an active node this PR does not make the situation worse. Therefore I decided to fix this in a separate PR (targeting main)

cipherboy:
Let's iterate on this in future PRs to unblock other work :-)

PR #1684, opened 2025-08-14
Update PKCS#11 seal docs

satoqz:
- Remove old `default_key_label` option, this never existed in OpenBao
- Clarify that `mechanism` can also be a named mechanism
- Document new `disable_software_encryption` option (see
  https://github.com/openbao/go-kms-wrapping/pull/41)


JanMa (approved):
LGTM

PR #1685, opened 2025-08-15
fix some ineffassign warnings

phil9909:


http/logical_test.go
```diff
@@ -975,7 +975,7 @@ func TestLogical_AuditEnabled_ShouldLogPluginMetadata_Auth(t *testing.T) {
 
 		auditResponse := map[string]interface{}{}
 		if req, ok := auditRecord["response"]; ok {
-			auditRequest = req.(map[string]interface{})
+			auditResponse = req.(map[string]interface{})
```
wslabosz-reply:
```suggestion
if res, ok := auditRecord["response"]; ok {
			auditResponse = res.(map[string]interface{})
```

http/logical_test.go
```diff
@@ -1065,7 +1065,7 @@ func TestLogical_AuditEnabled_ShouldLogPluginMetadata_Secret(t *testing.T) {
...
```
wslabosz-reply:
```suggestion
if res, ok := auditRecord["response"]; ok {
			auditResponse = res.(map[string]interface{})
```

vault/external_tests/plugin/external_plugin_test.go
```diff
@@ -975,7 +975,7 @@ func TestExternalPlugin_AuditEnabled_ShouldLogPluginMetadata_Secret(t *testing.T
 
 		auditResponse := map[string]interface{}{}
 		if req, ok := auditRecord["response"]; ok {
-			auditRequest = req.(map[string]interface{})
+			auditResponse = req.(map[string]interface{})
```
wslabosz-reply:
```suggestion
if res, ok := auditRecord["response"]; ok {
			auditResponse = res.(map[string]interface{})
```

wslabosz-reply (commented):
LGTM, but I see tests failing to some incorrect usage of `Fatalf` in tests or unused variables?

phil9909:
> LGTM, but I see tests failing to some incorrect usage of `Fatalf` in tests or unused variables?

Should be fixed now, sorry.

cipherboy:
@phil9909 Actually, the failure looks new to me? `TestSystemBackend_PoliciesDetailedAcl`: 

```
2025-09-05T13:38:28.6104692Z     logical_system_test.go:2546: 
2025-09-05T13:38:28.6105436Z         	Error Trace:	/home/runner/work/openbao/openbao/vault/logical_system_test.go:2546
2025-09-05T13:38:28.6106100Z         	Error:      	Expected value not to be nil.
2025-09-05T13:38:28.6106591Z         	Test:       	TestSystemBackend_PoliciesDetailedAcl
```

vault/logical_system_test.go
```diff
@@ -2543,6 +2543,7 @@ func TestSystemBackend_PoliciesDetailedAcl(t *testing.T) {
 
 	resp, err := b.HandleRequest(namespace.RootContext(nil), req)
 	require.NoError(t, err)
+	require.NotNil(t, resp)
```
cipherboy: Weird, policy updates should return data I believe, due to CAS.

phil9909: Oh, missed that one.

Created and Update seem to not return data in general.

I'm not sure, if it is required to make CAS useful, the usual workflow would look like this:
- while true:
  - read policy (including version)
  - modify policy locally
  - write policy
    - on success: break
 

The new version should be irrelevant (it would be a nice optimization, if a write would return the new data on a conflict to spare a round-trip, but not really required). 

For now, I will just change the test. If you think we should change the API here, I guess this should be a separate PR.

cipherboy: > Created and Update seem to not return data in general.

It is entirely dependent on the API endpoint. :-) 

But yes, I think we'll leave it as a future PR. Thanks @phil9909!

PR #1691, opened 2025-08-18
Bump goid for Go 1.25 compatibility

satoqz:
To future-proof and avoid hiccups for anyone building on Go 1.25 assuming backwards compatibility, even though we're currently still at Go 1.24.

For context, an outdated `goid` causes `go-deadlock` to report false positives.


cipherboy:
Thanks @satoqz!

PR #1696, opened 2025-08-18
Fix bug in tidy leading to infinite loop

cipherboy:
In `logical.HandleListPage(...)`, after was being redefined each loop, leading to the previous loop iteration limits not being tracked. Add a test case for this and correctly keep track of the last seen storage entry.

This also adjusts `logical.HandleListPage(...)` to take a context and check for cancellation between outer paginated calls and before processing each entry in the batch. This will ensure we have bounded iteration in the event of another issue with `ListPage(...)`. We also validate context cancellation.

### DIFF
--- a/changelog/1696.txt
+++ b/changelog/1696.txt
@@ -1,0 +1,9 @@
+```release-note:bug
+secrets/pki: Prevent infinite loop in tidy stemming from incorrect list pagination
+```
+```release-note:bug
+core/namespaces: Prevent infinite loop in namespace loading due to incorrect list pagination
+```
+```release-note:change
+sdk/logical: Introduce context to logical.HandleListPage(...).
+```
--- a/changelog/1696.txt
+++ b/changelog/1696.txt
@@ -2,7 +2,7 @@
 secrets/pki: Prevent infinite loop in tidy stemming from incorrect list pagination
 ```
 ```release-note:bug
-core/namespaces: Prevent infinite loop in namespace loading due to incorrect list pagination
+core/namespaces: Prevent infinite loop in namespace loading due to incorrect list pagination when more than 100 sibling namespaces exist under a given parent
 ```
 ```release-note:change
 sdk/logical: Introduce context to logical.HandleListPage(...).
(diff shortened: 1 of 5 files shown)


sdk/logical/storage.go
```diff
@@ -309,18 +309,23 @@ func ClearViewWithoutPagination(ctx context.Context, view ClearableView, logger
...
 	prefix string,
 	limit int,
 	itemCallback func(page int, index int, entry string) (cont bool, err error),
 	batchCallback func(page int, entries []string) (cont bool, err error),
 ) error {
-	page := 0
+	var page int
+	var after string
 	for {
-		var after string
```
cipherboy, line 320: This is the actual bug.

vault/namespace_store.go
```diff
@@ -215,7 +215,7 @@ func (ns *NamespaceStore) loadNamespacesRecursive(
 	ctx context.Context, barrier, view logical.Storage,
 	callback func(*namespace.Namespace) error,
 ) error {
-	return logical.HandleListPage(view, "", 100, func(page int, index int, entry string) (bool, error) {
+	return logical.HandleListPage(ctx, view, "", 100, func(page int, index int, entry string) (bool, error) {
```
cipherboy, line 218: You would need to have more than 100 namespaces within the given scope (root or parent namespace) to trigger this fwiw.

changelog/1696.txt
```diff
@@ -0,0 +1,9 @@
+```release-note:bug
+secrets/pki: Prevent infinite loop in tidy stemming from incorrect list pagination
+```
+```release-note:bug
+core/namespaces: Prevent infinite loop in namespace loading due to incorrect list pagination
```
cipherboy:
```suggestion
core/namespaces: Prevent infinite loop in namespace loading due to incorrect list pagination when more than 100 sibling namespaces exist under a given parent
```

sdk/logical/storage.go
fatima2003: No idea how this slipped past, good catch!

Issue #1517, opened 2025-07-01
Add CEL Helpers


fatima2003:
This is a follow up of #794, #869. 

In the CEL PKI RFC (#753), we mention that CEL supports injecting custom functions into the execution environment. This issue attempts to list all CEL helpers required for CEL in `pki` and `auth`. Some helpers are shared and others are specific to the individual engines. 

### Shared Helpers
- [ ] `decode_hex`; `encode_hex`
- [x] `decode_JSON`; `encode_JSON` #1549 
- [ ] `decode_ans1`; `encode_ans1`
- [ ] `http_request` + `allow_http` flag

### PKI Helpers
- [ ] `decode_PEM`; `encode_PEM`
- [ ] `decode_Base64`; `encode_Base64`
- [x] `check_valid_email`
- [ ] `check_valid_dns`
- [ ] `filter_list` - applies a function over a list, e.g., `check_valid_email` over a list.

#### JSON

```
"extensions": {
    "subject_alt_name": "{
        "dns":["example.com", "www.example.com"], 
        "emails":["admin@example.com"],
        "ips":["203.0.113.7"] }"
},
```

```
decode_JSON(request.extensions["subject_alt_name"])
```

#### Filtering Lists

```
"emails": ["alice@example.com","carol@domain.com"]
```

```
filter_list(check_valid_email, request.emails)
```

** The list is not exhausted, any suggestions are more than welcome :))

cipherboy:
@fatima2003 Thanks!


Some other candidates:

- `decode_hex` would be good, plus encode variants of the above. 
- `http_request` would be good but need to be limited by mount tunable + namespace tunable parameter and so would be more complex.

FrankSpitulski:
suggesting pulling in the ext package in cel-go https://github.com/google/cel-go/blob/master/ext/README.md. it has string, list, and math functions. base64 as well.

PR #1697, opened 2025-08-19
feat(cel): ext helpers

FrankSpitulski:
pulls in functions from the cel-go/ext package. these should be safe to include as they are packaged with the official cel library.

they contain useful functions for string manipulation and regex among others

this change also unifies the helpers between jwt and x509 cel programs. jwt users may find it useful to validate an email, not just x509 users.

Relates to #1517


driif:
Please add changelog.

driif (changes requested):
Please add changelog

driif (approved):

FrankSpitulski:
@cipherboy could you have a look at this one please?

suprjinx (approved):
Very helpful!

JanMa (approved):
Lgtm thanks @FrankSpitulski

PR #1700, opened 2025-08-19
Add declarative audit device creation mechanism

cipherboy:
In solving HCSEC-2025-14 / CVE-2025-6000 / CVE-2025-54997, we opted to introduce a new flag, unsafe_allow_api_audit_creation=false, which disables audit device creation via API. This is because creating audit devices is fundamentally a service operator (and not, an internal root-token) permission: many of these audit devices allow escaping the API sandbox of the service and allow e.g., writing to arbitrary files or opening arbitrary TCP or Unix sockets.

This change introduces a new server configuration block, audit, which takes two parameters (device plugin type and desired device path), plus the regular request body of a sys/audit enable request. This allows creating in a fully-declarative manner configuration.

When device configuration fails on startup/leadership election, the server will err and fail to load. Detailed error are written including on reload allowing configurations to be debugged.

---

Early look at the proof of concept for this. Will add RFC, docs, and tests before opening for broader review.

### DIFF
--- a/command/server/config.go
+++ b/command/server/config.go
@@ -122,6 +122,11 @@ type Config struct {
 	// declarative self-initialization subsystem on non-dev-mode instances
 	// and is part of the profile system.
 	Initialization []*profiles.OuterConfig `hcl:"-"`
+
+	// Audit specifies declaratively defined audit devices; these are created
+	// on the active node. Updates cannot occur, only additions or deletions,
+	// but can be modified through SIGHUP on a running server.
+	Audits []*AuditDevice `hcl:"-"`
 }
 
 const (
@@ -138,6 +143,9 @@ func (c *Config) Validate(sourceFilePath string) []configutil.ConfigError {
 	}
 	for _, l := range c.Listeners {
 		results = append(results, l.Validate(sourceFilePath)...)
+	}
+	for _, a := range c.Audits {
+		results = append(results, a.Validate(sourceFilePath)...)
 	}
 	results = append(results, c.validateEnt(sourceFilePath)...)
 	return results
@@ -772,6 +780,17 @@ func ParseConfig(d, source string) (*Config, error) {
 		result.Initialization = init
 	}
 
+	// Parse audit device stanzas.
+	if o := list.Filter("audit"); len(o.Items) > 0 {
+		delete(result.UnusedKeys, "audit")
+		audits, err := parseAuditDevices("audit", result.Audits, o)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing 'audit': %w", err)
+		}
+
+		result.Audits = audits
+	}
+
 	// Remove all unused keys from Config that were satisfied by SharedConfig.
 	result.UnusedKeys = configutil.UnusedFieldDifference(result.UnusedKeys, nil, append(result.FoundKeys, sharedConfig.FoundKeys...))
 	// Assign file info
@@ -1161,6 +1180,19 @@ func (c *Config) Sanitized() map[string]interface{} {
 		result["service_registration"] = sanitizedServiceRegistration
 	}
 
+	if len(c.Audits) > 0 {
+		var sanitizedAudits []map[string]interface{}
+		for _, a := range c.Audits {
+			cfg := map[string]interface{}{
+				"path":        a.Path,
+				"type":        a.Type,
+				"description": a.Description,
+			}
+			sanitizedAudits = append(sanitizedAudits, cfg)
+		}
+		result["audits"] = sanitizedAudits
+	}
+
 	return result
 }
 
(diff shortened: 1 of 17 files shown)


cipherboy:
@satoqz This now has a parsing test. I think someone else would be able to write a config test based on the listener test, but it was more involved than I had hoped for.

Some tests I've done locally:

1. Create audit device via SIGHUP or restart.
2. Delete audit device via SIGHUP or restart.
3. Restart with audit devices unchanged.
4. Try creating duplicate (two config or API+config or config+API).

website/content/docs/rfcs/config-audit-devices.mdx
```diff
@@ -0,0 +1,112 @@
...
+After [Vault HCSEC-2025-14 / CVE-2025-6000](https://discuss.hashicorp.com/t/hcsec-2025-14-privileged-vault-operator-may-execute-code-on-the-underlying-host/76033)
+and [OpenBao CVE-2025-54997](https://github.com/openbao/openbao/security/advisories/GHSA-xp75-r577-cvhp),
+it became apparent that _any_ API-driven audit device creation is unsafe.
+We remediate this fully by allowing full management of audit devices from
+the configuration file, throughout the life cycle of the device.
+
+## Problem Statement
+
+API-driven audit device creation is fundamentally unsafe. Even without control
+over `prefix`, audit devices such as `file` allow writing to arbitrary files
```
wslabosz-reply: nit:
```suggestion
over `prefix`, audit devices of type `file` allow writing to arbitrary files
```

website/content/docs/rfcs/config-audit-devices.mdx
```diff
@@ -0,0 +1,112 @@
...
+and `socket` allow writing to arbitrary Unix or network sockets. These
```
wslabosz-reply: nit:
```suggestion
and type `socket` allows writing to arbitrary Unix or network sockets. These
```

wslabosz-reply (approved):

website/content/docs/rfcs/config-audit-devices.mdx
```diff
@@ -0,0 +1,112 @@
+---
+sidebar_label: Audit Device Configuration
```
satoqz:
```suggestion
sidebar_label: Declarative Audit Devices
```

I suppose we've been able to configure audit devices before 😄

website/content/docs/rfcs/config-audit-devices.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+## Technical Description
+
+This change is largely self-contained to the audit system and server
+configuration parsing, with the exception of two necessary hooks:
+
+1. The standard unseal strategy now updates audit devices based on the
+   configuration.
+2. The reload handlers triggered on `SIGHUP` now will reload the seal as
+   well.
```
satoqz: seal?

command/server/config.go
```diff
@@ -1221,3 +1265,59 @@ func checkSkipPaths(path string, allPaths []string) (bool, error) {
...
+			return result, fmt.Errorf("%v.%d: %v type must be specified: %#v", name, index, name, item)
+		}
+
+		switch {
+		case i.Path != "":
+		case len(item.Keys) == 2:
+			i.Path = item.Keys[1].Token.Value().(string)
+		default:
+			return result, fmt.Errorf("%v.%d: %v path must be specified: %#v", name, index, name, item)
+		}
```
satoqz, line 1317: I'm not sure why we check `i.Type != ""` and `i.Path != ""` assuming they may somehow already be set, couldn't it just be:

```suggestion
		switch len(item.Keys) {
		case 0:
			return result, fmt.Errorf("%v.%d: %v type must be specified: %#v", name, index, name, item)
		case 1:
			return result, fmt.Errorf("%v.%d: %v path must be specified: %#v", name, index, name, item)
		case 2:
			i.Type = item.Keys[0].Token.Value().(string)
			i.Path = item.Keys[1].Token.Value().(string)
		default:
			return result, fmt.Errorf("%v.%d: %v: expected two keys: %#v", name, index, name, item)
		}
```

Probably missing something.

satoqz (approved):
LGTM, minor comments below.

command/server/config.go
cipherboy: @satoqz Honestly, not sure the full context, but this is used by listener parsing: 

https://github.com/openbao/openbao/blob/5e4fa283853c5580774b6caf589e42b4c37b4844/internalshared/configutil/listener.go#L195-L201

In:

https://github.com/openbao/openbao/commit/91b09c09b50dc5f008ee442aa6eac8302cd3cfd9#diff-0b1642f753dfe53bd60b0d3f731e4e7728f66e550b8820013a46e48db9c317b5L951-L953

You can see that the original behaved like you expected, simply conditioned around the length, but the new copy introduced the linked behavior. :shrug: 

Consistency is probably good?

website/content/docs/rfcs/config-audit-devices.mdx
cipherboy: Ah audit devices.

```suggestion
2. The reload handlers triggered on `SIGHUP` now will reload the audit devices as
   well.
```

website/content/docs/configuration/audit.mdx
```diff
@@ -0,0 +1,74 @@
...
+sidebar_label: audit
+description: |-
+  The audit stanza allows definition of audit devices from the configuration
+  file, reloaded on server restart and on SIGHUP.
+---
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+
+# Audit Devices
```
cipherboy:
```suggestion
# Declarative Audit Devices
```

website/content/docs/rfcs/config-audit-devices.mdx
```diff
@@ -0,0 +1,112 @@
+---
+sidebar_label: Declarative Audit Devices
+description: |-
+  An OpenBao RFC for defining audit devices in the server configuration
+  file through the complete audit device life cycle.
+---
+
+# Audit Device Configuration
```
cipherboy:
```suggestion
# Declarative Audit Device Configuration
```

command/server/config.go
satoqz: Hmm, sure. Not that it matters much.

PR #1702, opened 2025-08-20
Call umask(0o111) in server startup

### LINKED ISSUE #1661: Call `umask` in main

Prevent issues like [HCSEC-2025-14 Privileged Vault Operator May Execute Code on the Underlying Host](https://discuss.hashicorp.com/t/hcsec-2025-14-privileged-vault-operator-may-execute-code-on-the-underlying-host/76033) aka [CVE-2025-6000](https://www.cve.org/CVERecord?id=CVE-2025-6000) in the future (at least on all OSs that support `umask`, basically all but Windows afaik)

We should call `syscall.Umask(0o111)` as one of the first thing in the `main` method.

This will prevent `open` calls from creating files with executable permissions (silently removing them). See example program below. It does not prevent a `chmod` operation. So if executable permissions are required, they can still be assigned.

- Don't do it.

None

**Additional context**

See also: https://man.archlinux.org/man/umask.2

Example program:

```go
package main

import (
	"os"
	"syscall"
)

func main() {
	file, err := os.OpenFile("before-umask", os.O_CREATE, 0o755)
	if err != nil {
		panic(err)
	}
	file.Close()

	syscall.Umask(0o111)

	file, err = os.OpenFile("after-umask", os.O_CREATE, 0o755)
	if err != nil {
		panic(err)
	}
	file.Close()

	file, err = os.OpenFile("after-umask-explict-chmod", os.O_CREATE, 0o755)
	if err != nil {
		panic(err)
	}
	file.Chmod(0o755)
	file.Close()

}
```

Results in the following files

```console
$ ls -l
total 4
-rw-r--r-- 1 philipp philipp   0 Aug  8 11:16 after-umask
-rwxr-xr-x 1 philipp philipp   0 Aug  8 11:16 after-umask-explict-chmod
-rwxr-xr-x 1 philipp philipp   0 Aug  8 11:16 before-umask
-rw-r--r-- 1 philipp philipp 441 Aug  8 11:15 main.go
```

satoqz:
Resolves #1661

As suggested by @cipherboy, we only do this in the server "main" for now, and may later expand it to agent/proxy.

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -23,6 +23,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	systemd "github.com/coreos/go-systemd/v22/daemon"
@@ -938,6 +939,8 @@ func (c *ServerCommand) Run(args []string) int {
 		c.UI.Error(err.Error())
 		return 1
 	}
+
+	syscall.Umask(0o111)
 
 	// Don't exit just because we saw a potential deadlock.
 	deadlock.Opts.OnPotentialDeadlock = func() {}


changelog/1702.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+command/server: Call [`umask(2)`](https://man.archlinux.org/man/umask.2) at startup to avoid settings executable bits on file creation.
```
wslabosz-reply: nit?
```suggestion
command/server: Call [`umask(2)`](https://man.archlinux.org/man/umask.2) at startup to disallow setting executable bits on file creation.
```

satoqz: That is not a nit that is a fatal spelling error sir 😄

satoqz: Addressed, but still prefer "avoid" over "disallow" as it's a silent trim, and no errors are thrown.

cipherboy: Ah good, the update is good. I was going to say, we don't _disallow_ setting executable bits (and err), we _avoid_ it (in that, the call succeeds, they're just silently elided). 

e:f,b. :laughing:

wslabosz-reply: that's just one extra letter, code did not panic.

changelog/1702.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
cipherboy:
```suggestion
```release-note:change
```

Let's also make this a change so people are aware of it. I don't think feature is quite correct (improvement would be better), but it might break people.

satoqz: To my defense, I put `change` at first and then looked at #1661 and saw you tagged it `feature`... I will blame this on not trusting my own judgement enough :^)

cipherboy: We don't actually have `feature` vs `improvement` labels; that one was assigned by the default issue template fwiw. :-)

cipherboy:
@satoqz Oh interesting, this fails tests that create temporary directories:

https://github.com/openbao/openbao/actions/runs/17101608394/job/48499715504?pr=1702#step:7:12056

> ```
> === FAIL: command TestServer_DevTLS (0.00s)
>     server_test.go:313: 
>         	Error Trace:	/home/runner/work/openbao/openbao/command/server_test.go:313
>         	Error:      	Not equal: 
>         	            	expected: 0
>         	            	actual  : 1
>         	Test:       	TestServer_DevTLS
>         	Messages:   	open /tmp/vault-tls679893717/vault-ca.pem: permission denied
>         	            	remove /tmp/vault-tls679893717/vault-ca.pem: permission denied
>         	            	remove /tmp/vault-tls679893717/vault-cert.pem: permission denied
>         	            	remove /tmp/vault-tls679893717/vault-key.pem: permission denied
> ```

This means it likely also breaks ACME-enabled TLS certificate listeners. 

Which, I guess it does, but much earlier, because the Raft directory is also impacted:

>   2025-08-20T14:43:23.451Z [TRACE] Test_ACME.core-0:  Error initializing storage of type raft: open /openbao/file/raft/raft.db: permission denied

cipherboy (changes requested):
Marking request changes so we can get tests addressed. I think this may mean `Umask` is not a viable approach.

satoqz:
@cipherboy I had expected that this might trigger... _interesting_ errors. I'll investigate tomorrow.

cipherboy:
@satoqz No worries, we can also drop it for a future release.

satoqz:
Upon some further investigation, `umask`ing execute bits is completely infeasible. As `umask` applies to both files and directories, it'll unconditionally shave exec bits off any directories that OpenBao creates. Now the problem here is that execute bits on directories serve a different purpose that we would very much like to preserve: https://superuser.com/questions/168578/why-must-a-folder-be-executable. TIL, and fair enough.

PR #1707, opened 2025-08-21
Per-Namespace-Sealing: Remove barrierByStoragePath

satoqz:
This PR removes `barrierByStoragePath` and related functionality from `SealManager`. This is mainly due to the realization that the same information is usually already available without the need to lock and access centralized state, with the exception of the raw backend where we can just parse the storage path, similar to what horizontal scalability does.

### DIFF
--- a/vault/logical_raw.go
+++ b/vault/logical_raw.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	log "github.com/hashicorp/go-hclog"
+	"github.com/openbao/openbao/helper/namespace"
 	"github.com/openbao/openbao/sdk/v2/framework"
 	"github.com/openbao/openbao/sdk/v2/helper/compressutil"
 	"github.com/openbao/openbao/sdk/v2/logical"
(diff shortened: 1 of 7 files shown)


satoqz:
I'll leave this where it is right now, I wanted to go a bit further with the refactors... but I've just been getting burnt by the curse of core.

vault/logical_raw.go
```diff
@@ -48,6 +49,33 @@ func NewRawBackend(core *Core) *RawBackend {
...
 
+func (b *RawBackend) storageByPath(ctx context.Context, path string) (StorageAccess, error) {
+	ns, rest, err := b.core.NamespaceByStoragePath(ctx, path)
+	if err != nil {
+		return nil, err
+	}
+
+	// These paths use the "upper" barrier, which is the direct physical layer
+	// for the root namespace.
+	specialPath := rest == sealConfigPath || rest == recoverySealConfigPath
```
satoqz, line 60: FWIW `/sys/raw` did not support reading/writing these paths before, but it seemed like a QoL improvement to add that.

satoqz:
@cipherboy FYI this would be the next minor increment in the namespace sealing adventure.

wslabosz-reply (approved):
looks good to me, but I'm also a contributor to this PR, so don't consider my approval as a green light.

I wonder if the "canary" paths code does anything, I've never found a function that reads into the store to find these entries...

vault/logical_raw.go
cipherboy: Yeah, I mean, `sys/raw` is already highly unsafe. It is not like there's much more to be done to make it safer. 😆

vault/namespaces_oss.go
```diff
@@ -38,3 +39,22 @@ func NamespaceView(barrier logical.Storage, ns *namespace.Namespace) BarrierView
 func (c *Core) NamespaceByPath(ctx context.Context, path string) (*namespace.Namespace, string) {
 	return c.namespaceStore.GetNamespaceByLongestPrefix(ctx, path)
 }
+
+func (c *Core) NamespaceByStoragePath(ctx context.Context, path string) (*namespace.Namespace, string, error) {
```
cipherboy: Just since I'm picky about this, let's move this to `namespace_store.go`, without the HashiCorp copyright header.

cipherboy (approved):
Works for me, thanks @satoqz!

satoqz:
`bitnami/cassandra` is sabotaging tests, I'll ignore and merge. @wslabosz-reply we probably want to rebase the `namespaces-seal` branch again :^)

cipherboy:
@wslabosz-reply @satoqz Do you want me to do the rebase?

wslabosz-reply:
@cipherboy 
If you feel like it, otherwise I'm gonna tackle it tomorrow.

PR #1709, opened 2025-08-21
Add http audit method

cipherboy:
This adds a new HTTP audit method, allowing collection of audit events by a remote server.

The HTTP audit device behaves like other audit devices, failing closed on failure to send the specified audit message. This is in contrast to a retry queue or batching audit device mechanism, which would also be possible but would violate the audit device contract. This keeps the implementation fairly simple. 

See also: https://gitlab.com/gitlab-org/gitlab/-/issues/470461
See also: https://gitlab.com/gitlab-com/content-sites/handbook/-/merge_requests/15328

### DIFF
--- a/builtin/audit/http/backend.go
+++ b/builtin/audit/http/backend.go
@@ -1,0 +1,314 @@
+// Copyright (c) The OpenBao Contributors
+// SPDX-License-Identifier: MPL-2.0
+
+package http
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strconv"
+	"sync"
+	"sync/atomic"
+
+	"github.com/hashicorp/go-secure-stdlib/parseutil"
+	"github.com/openbao/openbao/audit"
+	"github.com/openbao/openbao/sdk/v2/helper/salt"
+	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/openbao/openbao/version"
+)
+
+// Backend is the audit backend for the http-based audit store.
+//
+// This presently has little logic to handle retrying failed requests.
+type Backend struct {
+	uri     string
+	headers http.Header
+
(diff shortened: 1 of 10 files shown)


website/content/docs/audit/http.mdx
satoqz, line 1: This needs an entry in `sidebars.ts`

builtin/audit/http/backend.go
```diff
@@ -0,0 +1,308 @@
...
+
+	b.clientLock.Lock()
+	defer b.clientLock.Unlock()
+
+	if b.client != nil {
+		return client, nil
+	}
+
+	// TODO (ascheel): use a different HTTP client.
+	b.client = http.DefaultClient
```
satoqz, line 290: Not sure what your plans are here, but I suppose it'd be nice to be a good citizen and attach a custom user agent, sth like `fmt.Sprintf("OpenBaoAuditor/%s", version.VersionNumber())`

satoqz (approved):

builtin/audit/http/backend.go
cipherboy: @satoqz Added! Also fixed a bug with headers not being set properly :man_facepalming:. Thanks @JAYKRISHNAN for reporting :-D

satoqz (approved):

builtin/audit/http/backend.go
```diff
@@ -0,0 +1,314 @@
...
+
+	_, err := b.getClient()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (b *Backend) Invalidate(_ context.Context) {
```
cipherboy, line 309: @wslabosz-reply Was there more you were looking for in invalidation? This is the same as other backends I believe.

wslabosz-reply: no, that's fine.

cipherboy:
Thanks all!

PR #1711, opened 2025-08-21
Fix creation of nested namespaces in self-init

cipherboy:
As reported by Jan Martens on Matrix, the following configuration snippet fails:

> initialize "namespaces" {
>   request "dev" {
>     operation = "create"
>     path = "sys/namespaces/dev"
>   }
>   request "nested" {
>     operation = "create"
>     path = "dev/sys/namespaces/nested"
>   }
> }

This is due to using the same context created in Initialize (which needs the root namespace attached) in the subsequent call to the profile system.

Resolves #

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -1791,13 +1791,13 @@ func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) erro
 	}
 
 	// Now perform the component requests of self-initialization.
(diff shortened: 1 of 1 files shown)


JanMa (approved):
Thanks for the quick fix @cipherboy . I can confirm it now works as expected and you can create nested namespaces :+1:

PR #1714, opened 2025-08-22
add SchemaName to Response struct as optional name of the schema for …

### LINKED ISSUE #1713: Add SchemaName to response to optionally name response schemas in OpenAPI specs.
**Problem**: We currently name response schemas based on the operation ID. This creates a new response schema for each operation, even when they share the same schema.

**Example**: handleLookup (Self, ByAccessor, GET, and POST) returns the same response, with the ID field optional—it isn’t returned for token lookups by accessor.

With framework.Path operations, there’s no way to generate a shared response schema for the client. This leads to code duplication, since a new response type is created for each operation.

**Proposal**: Add SchemaName to the Response type. This lets us assign an optional shared name to the response schema in OpenAPI, reducing generated code and client boilerplate.

driif:
**Problem**: We currently name response schemas based on the operation ID. This creates a new response schema for each operation, even when they share the same schema.

**Example**: handleLookup (Self, ByAccessor, GET, and POST) returns the same response, with the ID field optional—it isn’t returned for token lookups by accessor.

With framework.Path operations, there’s no way to generate a shared response schema for the client. This leads to code duplication, since a new response type is created for each operation.

**Proposal**: Add SchemaName to the Response type. This lets us assign an optional shared name to the response schema in OpenAPI, reducing generated code and client boilerplate.

Resolves #1713

### DIFF
--- a/changelog/1714.txt
+++ b/changelog/1714.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+**sdk/framework**: add Response.SchemaName to allow custom response schema names in the generated OpenAPI spec.
+```


wslabosz-reply (approved):

sdk/framework/openapi.go
```diff
@@ -519,8 +519,17 @@ func documentPath(p *Path, specialPaths *logical.Paths, requestResponsePrefix st
...
-						doc.Components.Schemas[responseName] = responseSchema
+						// Use custom schema name if provided, otherwise generate from operationID
+						responseName := resp.SchemaName
+						if responseName == "" {
+							responseName = hyphenatedToTitleCase(operationID) + "Response"
+						}
+
+						// Only add to schemas if not already present (allows schema reuse)
+						if _, exists := doc.Components.Schemas[responseName]; !exists {
+							doc.Components.Schemas[responseName] = responseSchema
```
cipherboy, line 530: Should we make `SchemaName` exclusive with `responseSchema`? :thinking:

driif: Hey @cipherboy , Could you clarify what you mean by making SchemaName “exclusive” with responseSchema. Are you suggesting "SchemaName" not to be a part of Response or validate that any reuse matches the first definition?

cipherboy: @driif Oh, I see, I was confused about the intent of this PR. I thought we were doing this as a reuse mechanism (response schema names would let us respond to multiple endpoints with the same response values without having to re-specify all fields). There, I would want exclusivity where you'd use a named response schema only if you don't specify a regular (fat) schema in the first place. More like a symbolic link.

This is just providing a way to name response schemas, not as a linking mechanism. My bad!

Issue #1708, opened 2025-08-21
Crash in `countEntitiesByNamespace` when `unsafe_cross_namespace_identity=true` (`MemDB.Txn` on nil receiver)


evsasha:
**Describe the bug**

When running OpenBao with `unsafe_cross_namespace_identity=true`, creating an additional namespace causes the server to crash during periodic metrics collection.  
The panic occurs in `IdentityStore.countEntitiesByNamespace`, which tries to open a transaction on a `nil` `MemDB`.  

Excerpt from stack trace:
```bash
2025-08-21T14:15:09.001Z [INFO]  core: successful mount: namespace=test/ path=cubbyhole/ type=ns_cubbyhole version=""
2025-08-21T14:15:09.002Z [INFO]  core: successful mount: namespace=test/ path=sys/ type=ns_system version=""
2025-08-21T14:15:09.002Z [INFO]  core: successful mount: namespace=test/ path=identity/ type=ns_identity version=""
2025-08-21T14:15:09.003Z [INFO]  core: enabled credential backend: namespace=test/ path=token/ type=ns_token version=""
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xcff0cd]

goroutine 333 [running]:
github.com/hashicorp/go-memdb.(*MemDB).getRoot(...)
        /home/runner/go/pkg/mod/github.com/hashicorp/go-memdb@v1.3.4/memdb.go:65
github.com/hashicorp/go-memdb.(*MemDB).Txn(0x0, 0x0)
        /home/runner/go/pkg/mod/github.com/hashicorp/go-memdb@v1.3.4/memdb.go:78 +0x4d
github.com/openbao/openbao/vault.(*IdentityStore).countEntitiesByMountAccessor.func1({0x3c77ce0?, 0xc0014723d0?}, {0x3b43940?, 0xc00147a5a0?})
        /home/runner/work/openbao/openbao/vault/identity_store_util.go:2517 +0xa6
internal/sync.(*HashTrieMap[...]).iter(0x5b67820, 0xc000e36460, 0xc00074cce8)
        /opt/hostedtoolcache/go/1.24.6/x64/src/internal/sync/hashtriemap.go:512 +0xea
internal/sync.(*HashTrieMap[...]).Range(0x476a6001?, 0xc0015e88c0?)
[...]

PR #1715, opened 2025-08-22
Fix nil panic with unsafe_cross_namespace_identity

cipherboy:
The only place where IdentityStore.db(...) is not called is when ranging over all views to collect metrics. This did not handle the unsafe cross-namespace identity behavior of using a single memdb (only available on the root namespace's view entry), leading to a nil panic.

Resolves: #1708

### DIFF
--- a/vault/identity_store_test.go
+++ b/vault/identity_store_test.go
@@ -185,10 +185,10 @@ func TestIdentityStore_UnsealingWhenConflictingAliasNames(t *testing.T) {
 func TestIdentityStore_EntityIDPassthrough(t *testing.T) {
 	// Enable AppRole auth and initialize
 	ctx := namespace.RootContext(nil)
-	is, ghAccessor, core := testIdentityStoreWithAppRoleAuth(ctx, t)
+	is, approleAccessor, core := testIdentityStoreWithAppRoleAuth(ctx, t)
 	alias := &logical.Alias{
 		MountType:     "approle",
-		MountAccessor: ghAccessor,
+		MountAccessor: approleAccessor,
 		Name:          "approleuser",
 	}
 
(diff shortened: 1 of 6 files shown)


cipherboy:
Accidentally pushed a scratch version of the tests; fixed this.

vault/identity_store_test.go
```diff
@@ -706,7 +715,15 @@ func testIdentityStoreWithAppRoleUserpassAuth(ctx context.Context, t *testing.T)
...
-	c, _, _ := TestCoreUnsealed(t)
+	conf := &CoreConfig{
+		BuiltinRegistry:              corehelpers.NewMockBuiltinRegistry(),
+		UnsafeCrossNamespaceIdentity: unsafeShared,
+		AuditBackends: map[string]audit.Factory{
+			"file": auditFile.Factory,
+		},
+	}
+	core := TestCoreWithSealAndUI(t, conf)
+	c, _, _ := testCoreUnsealed(t, core)
```
satoqz:
```suggestion
	c, _, _ := TestCoreUnsealedWithConfig(t, conf)
```

satoqz (approved):
LGTM

vault/identity_store_test.go
cipherboy: Ah, completely missed this one. I knew there had to be a helper here. :laughing:

cipherboy:
I'm merging based on reviews here. It looks like a container image we rely on has disappeared:

```
2025-08-28T12:29:02.9793775Z     connection_producer_test.go:44: Could not start docker cassandra: container create failed: Error response from daemon: No such image: bitnami/cassandra:3.11.11
```

PR #1719, opened 2025-08-22
Remove github.com/dustin/go-humanize

cipherboy:
This removes one dependency on `github.com/dustin/go-humanize` and removes an outdated usage of `github.com/mitchellh/go-homedir`, though leaves the dependency due to other issues.

### DIFF
--- a/vault/diagnose/os_common.go
+++ b/vault/diagnose/os_common.go
@@ -38,7 +38,7 @@ partLoop:
 					Advice("It is recommended to have more than five percent of the partition free."))
 			} else if usage.Free < 1<<30 {
 				quantity := usage.Free
-				suffix := " bytes"
+				suffix := "bytes"
 				switch {
 				case quantity > 1024*1024*1024:
 					quantity /= 1024 * 1024 * 1024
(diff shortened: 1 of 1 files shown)


vault/diagnose/os_common.go
```diff
@@ -38,7 +37,21 @@ partLoop:
...
+				case quantity > 1024*1024:
+					quantity /= 1024 * 1024
+					suffix = "MB"
+				case quantity > 1024:
+					quantity /= 1024
+					suffix = "KB"
+				}
+
+				SpotWarn(ctx, testName, fmt.Sprintf(partition.Mountpoint+" has %s%s free.", quantity, suffix),
 					Advice("It is recommended to have at least 1 GB of space free per partition."))
```
phil9909:
```suggestion
				suffix := "bytes"
				switch {
				case quantity > 1024*1024*1024:
					quantity /= 1024 * 1024 * 1024
					suffix = "GB"
				case quantity > 1024*1024:
					quantity /= 1024 * 1024
					suffix = "MB"
				case quantity > 1024:
					quantity /= 1024
					suffix = "KB"
				}

				SpotWarn(ctx, testName, fmt.Sprintf(partition.Mountpoint+" has %d %s free.", quantity, suffix),
					Advice("It is recommended to have at least 1 GB of space free per partition."))
```

Two things:
- `quantity` should be formatted using `%d`  otherwise you get something like "/ has %!s(uint64=42)GB free."
- Was it on purpose to put a space before "bytes", but not before the other units? My suggestion removes it from `" bytes"` and adds a space into the format string instead.

satoqz: Yeah, it seems that while "100MB" is common and "100 bytes" is obviously correct, "100 MB" is _more correct_ in that case, too.

cipherboy:
Accepted your suggestion, thanks @phil9909!

PR #1728, opened 2025-08-25
Fix profile field selector for lists, configs

cipherboy:
When parsing from a configuration file, the field selector is of type `[]interface`, not `[]string`. This also gives us the opportunity to handle lists through heterogeneously typed lists (e.g., `["data", "keys", 0]` for instance).

---

cc @KrzysztofKornalewski-Reply @JanMa

### DIFF
--- a/helper/profiles/history.go
+++ b/helper/profiles/history.go
@@ -95,7 +95,7 @@ func (eh *EvaluationHistory) GetResponse(outerBlock string, requestBlock string)
 	return eh.getValue(eh.Responses, outerBlock, requestBlock)
 }
 
-func (eh *EvaluationHistory) GetRequestField(outerBlock string, requestBlock string, fieldSelector interface{}) (interface{}, error) {
+func (eh *EvaluationHistory) GetRequestField(outerBlock string, requestBlock string, fieldSelector []interface{}) (interface{}, error) {
 	values, err := eh.getValue(eh.Requests, outerBlock, requestBlock)
 	if err != nil {
 		return nil, err
@@ -109,7 +109,7 @@ func (eh *EvaluationHistory) GetRequestField(outerBlock string, requestBlock str
 	return val, nil
 }
 
-func (eh *EvaluationHistory) GetResponseField(outerBlock string, requestBlock string, fieldSelector interface{}) (interface{}, error) {
+func (eh *EvaluationHistory) GetResponseField(outerBlock string, requestBlock string, fieldSelector []interface{}) (interface{}, error) {
 	values, err := eh.getValue(eh.Responses, outerBlock, requestBlock)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


helper/profiles/history.go
```diff
@@ -172,8 +176,58 @@ func (eh *EvaluationHistory) getField(obj map[string]interface{}, fieldSelector
 			}
 		}
 
+		return nil, errors.New("selector had zero length")
+	case []interface{}:
```
phil9909: new code should use `any` instead of an empty interface, not sure why the linter didn't catch this.

helper/profiles/history.go
```diff
@@ -146,20 +146,24 @@ func (eh *EvaluationHistory) getValue(block map[string]map[string]map[string]int
 	return req, nil
 }
 
+// getField operates on the premise that the outer request object is always a
+// map; this is true even of list responses as they're contained in a regular
+// response map. However, inner items may be lists; in this case, a selector
+// of type []interface{} must be used to index arrays.
 func (eh *EvaluationHistory) getField(obj map[string]interface{}, fieldSelector interface{}) (interface{}, error) {
 	switch typed := fieldSelector.(type) {
```
phil9909: How do you feel about converting all possible types to a `[]any` first and only implement the access logic once: https://play.golang.com/p/PCsbQituCPd

helper/profiles/history.go
```diff
@@ -172,8 +176,58 @@ func (eh *EvaluationHistory) getField(obj map[string]interface{}, fieldSelector
 			}
 		}
 
+		return nil, errors.New("selector had zero length")
+	case []interface{}:
+		var base interface{} = obj
+
+		for i, rawSelector := range typed {
+			switch selector := rawSelector.(type) {
```
phil9909: should we add a `default` case?

helper/profiles/request_source_test.go
```diff
@@ -171,3 +171,31 @@ func TestRequestSource_Evaluate_WithFieldSelector_String(t *testing.T) {
 		t.Fatalf("Evaluate result = %#v; want %#v", result, expected)
 	}
 }
+
+func TestRequestSource_Evaluate_WithFieldSelector_Interface(t *testing.T) {
+	ctx := context.Background()
```
phil9909, line 178:
```suggestion
	ctx := t.Context()
```

helper/profiles/request_source_test.go
```diff
@@ -171,3 +171,31 @@ func TestRequestSource_Evaluate_WithFieldSelector_String(t *testing.T) {
...
+	history := &EvaluationHistory{}
+	source := &RequestSource{
+		field: map[string]interface{}{"request_name": "mount-userpass", "field_selector": []interface{}{"userPass", 0}},
+	}
+
+	if _, _, err := source.Validate(ctx); err != nil {
+		t.Fatalf("Validate error: %v", err)
+	}
```
phil9909: Do we want to use the [testify](https://pkg.go.dev/github.com/stretchr/testify) library for assertions where possible?

```suggestion
    require.NoError(t, source.Validate(ctx), "Validate failed")
```

phil9909 (changes requested):

helper/profiles/history.go
cipherboy: @phil9909 Sounds great to do in a later PR. ;-)

helper/profiles/history.go
cipherboy: @phil9909 There's so much use of `interface{}` that I think it doesn't really matter and we can address later:

```
$ grep -ir '\]any' | wc -l
53
```

vs

```
$ grep -ir '\]interface{' | wc -l
7362
```

helper/profiles/request_source_test.go
cipherboy: This also has very little use. I want to address this later, but for now especially for this use (which doesn't use ctx at all), it shouldn't matter.

helper/profiles/request_source_test.go
cipherboy: Yes, will address when I'm fixing the above.

helper/profiles/history.go
cipherboy: Ah yes, this is missing. Will address.

helper/profiles/history.go
cipherboy: I ended up pulling this into this PR as well. as it lets us handle the `int` and `[]int` cases too.

cipherboy:
@openbao/openbao-core-committers This should be ready for review. Test failures unrelated, see comment in #1715, I'll open a PR shortly.

phil9909 (approved):

cipherboy:
Thanks @satoqz, I'll open that as a follow-up PR to not block this release too long.

PR #1729, opened 2025-08-25
add custom_metadata to key in transit plugin

Pacman29:
Add custom metadata to transit keys operations


cipherboy:
@Pacman29 Apologies for the delay; I don't think there's broad interest in this. With things like Profiles coming soon, it should be possible to manage both keys in Transit and metadata in KVv2 at the same time at a single endpoint.  This would also work for PKI and SSH engines as well.

PR #1732, opened 2025-08-25
feat(OIDC-Provider): Add clientCredentialsFlow to Oidc Provider

### LINKED ISSUE #1727: [OIDC Provider]: Support grant_type=client_credentials
Currently, only the [[authorization code flow](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth)](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) is supported. This limits OpenBao to interactive, user-centric authentication scenarios.
However, for many modern infrastructures, we also need **non-interactive service-to-service authentication**. For example, microservices, monitoring agents, or collectors often need to authenticate with each other securely without a user being involved.

Support for `grant_type=client_credentials` in the token endpoint, so OpenBao can be used as the central provider for machine-to-machine authentication.
The goal is to issue valid access tokens for services based on their client credentials, enabling secure communication between distributed components without deploying additional infrastructure.

* Installing and maintaining a full-fledged OIDC server such as Keycloak or Dex.
* Using external identity providers that support client credentials, which adds complexity and reduces the lightweight benefits of OpenBao.

**Additional context**
My main use case is to connect multiple instances of [[OpenTelemetry](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/oidcauthextension)](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/oidcauthextension) across different infrastructure components.
Having client credentials support in OpenBao would allow a consistent, secure, and lightweight authentication solution for service-to-service traffic in observability pipelines and other distributed setups.

Syndlex:
Currently, only the [authorization code flow](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) is supported. This limits OpenBao to interactive, user-centric authentication scenarios.
However, for many modern infrastructures, we also need non-interactive service-to-service authentication. For example, microservices, monitoring agents, or collectors often need to authenticate with each other securely without a user being involved.

Resolves #1727

Should this approach be accepted I will change the documentation.

### DIFF
--- a/ui/app/models/oidc/client.js
+++ b/ui/app/models/oidc/client.js
@@ -73,7 +73,7 @@ export default class OidcClientModel extends Model {
   idTokenTtl;
 
   @attr('boolean', {
-    defaultValue: false,
+    defaultValue: true,
     label: 'Allow authorization code flow',
   })
   authorizationCode;
--- a/ui/app/models/oidc/client.js
+++ b/ui/app/models/oidc/client.js
@@ -72,6 +72,18 @@ export default class OidcClientModel extends Model {
   })
   idTokenTtl;
 
+  @attr('boolean', {
+    defaultValue: false,
+    label: 'Allow authorization code flow',
+  })
+  authorizationCode;
+
+  @attr('boolean', {
+    defaultValue: false,
+    label: 'Allow client credential flow',
+  })
+  clientCredentials;
+
   // >> END MORE OPTIONS TOGGLE <<
 
   @attr('array', { label: 'Assign access' }) assignments; // no editType because does not use form-field component
--- a/vault/identity_store_oidc_provider.go
+++ b/vault/identity_store_oidc_provider.go
@@ -559,7 +559,7 @@ func oidcProviderPaths(i *IdentityStore) []*framework.Path {
 				},
 				"grant_type": {
 					Type:        framework.TypeString,
-					Description: "The authorization grant type. The following grant types are supported: 'authorization_code'.",
+					Description: "The authorization grant type. The following grant types are supported: 'authorization_code','client_credentials'.",
 					Required:    true,
 				},
 				"redirect_uri": {
@@ -570,6 +570,11 @@ func oidcProviderPaths(i *IdentityStore) []*framework.Path {
 				"code_verifier": {
 					Type:        framework.TypeString,
 					Description: "The code verifier associated with the authorization code.",
+				},
+				"scope": {
+					Type:        framework.TypeString,
+					Description: "A space-delimited, case-sensitive list of scopes to be requested. The 'openid' scope is required. This is used when using in with 'client_credentials' flow",
+					Required:    true,
 				},
 				// For confidential clients, the client_id and client_secret are provided to
 				// the token endpoint via the 'client_secret_basic' or 'client_secret_post'
@@ -1994,10 +1999,99 @@ func (i *IdentityStore) pathOIDCToken(ctx context.Context, req *logical.Request,
 	if grantType == "" {
 		return tokenResponse(nil, ErrTokenInvalidRequest, "grant_type parameter is required")
 	}
-	if grantType != "authorization_code" {
+	switch grantType {
+	case "authorization_code":
+		return i.authorizationCodeFlow(ctx, req, d, ns, clientID, name, client, key, provider)
+	case "client_credentials":
+		return i.clientCredentialsFlow(ctx, req, d, ns, clientID, name, client, key, provider)
+	default:
 		return tokenResponse(nil, ErrTokenUnsupportedGrantType, "unsupported grant_type value")
 	}
 
+}
+
+func (i *IdentityStore) clientCredentialsFlow(ctx context.Context, req *logical.Request, d *framework.FieldData, ns *namespace.Namespace, id string, name string, client *client, key *namedKey, provider *provider) (*logical.Response, error) {
+	// Since Credentials Flow does not use authorize endpoint Scopes come direcly from Token endpoint
+	// Validate that a scope parameter is present and contains the openid scope value
+	requestedScopes := strutil.ParseDedupAndSortStrings(d.Get("scope").(string), scopesDelimiter)
+	if len(requestedScopes) == 0 || !slices.Contains(requestedScopes, openIDScope) {
+		return tokenResponse(nil, ErrAuthInvalidRequest, fmt.Sprintf("scope parameter must contain the %q value", openIDScope))
+	}
+
+	// Scope values that are not supported by the provider should be ignored
+	scopes := make([]string, 0)
+	for _, scope := range requestedScopes {
+		if slices.Contains(provider.ScopesSupported, scope) && scope != openIDScope {
+			scopes = append(scopes, scope)
+		}
+	}
+
+	// The access token is a Vault batch token with a policy that only
+	// provides access to the issuing provider's userinfo endpoint.
+	accessTokenIssuedAt := time.Now()
+	accessTokenExpiry := accessTokenIssuedAt.Add(client.AccessTokenTTL)
+	accessToken := &logical.TokenEntry{
+		Type:               logical.TokenTypeBatch,
+		NamespaceID:        ns.ID,
+		Path:               req.Path,
+		TTL:                client.AccessTokenTTL,
+		CreationTime:       accessTokenIssuedAt.Unix(),
+		NoIdentityPolicies: true,
+		Meta: map[string]string{
+			"oidc_token_type": "access token",
+		},
+		InternalMeta: map[string]string{
+			accessTokenClientIDMeta: client.ClientID,
+			accessTokenScopesMeta:   strings.Join(scopes, scopesDelimiter),
+		},
+	}
+	err := i.tokenStorer.CreateToken(ctx, accessToken, true)
+	if err != nil {
+		return tokenResponse(nil, ErrTokenServerError, err.Error())
+	}
+
+	// Set the ID token claims
+	idTokenIssuedAt := time.Now()
+	idTokenExpiry := idTokenIssuedAt.Add(client.IDTokenTTL)
+	idToken := idToken{
+		Namespace: ns.ID,
+		Issuer:    provider.effectiveIssuer,
+		Subject:   client.ClientID,
+		Audience:  client.ClientID,
+		Expiry:    idTokenExpiry.Unix(),
+		IssuedAt:  idTokenIssuedAt.Unix(),
+	}
+
+	// Populate each of the requested scope templates
+	templates, conflict, err := i.populateScopeTemplatesWithoutEntity(ctx, req.Storage, ns, scopes...)
+	if !conflict && err != nil {
+		return tokenResponse(nil, ErrTokenServerError, err.Error())
+	}
+	if conflict && err != nil {
+		return tokenResponse(nil, ErrTokenInvalidRequest, err.Error())
+	}
+
+	// Generate the ID token payload
+	payload, err := idToken.generatePayload(i.Logger(), templates...)
+	if err != nil {
+		return tokenResponse(nil, ErrTokenServerError, err.Error())
+	}
+
+	// Sign the ID token using the client's key
+	signedIDToken, err := key.signPayload(payload)
+	if err != nil {
+		return tokenResponse(nil, ErrTokenServerError, err.Error())
+	}
+
+	return tokenResponse(map[string]interface{}{
+		"token_type":   "Bearer",
+		"access_token": accessToken.ID,
+		"id_token":     signedIDToken,
+		"expires_in":   int64(accessTokenExpiry.Sub(accessTokenIssuedAt).Seconds()),
+	}, "", "")
+}
+
+func (i *IdentityStore) authorizationCodeFlow(ctx context.Context, req *logical.Request, d *framework.FieldData, ns *namespace.Namespace, clientID string, name string, client *client, key *namedKey, provider *provider) (*logical.Response, error) {
 	// Validate the authorization code
 	code := d.Get("code").(string)
 	if code == "" {
@@ -2433,6 +2527,53 @@ func (i *IdentityStore) populateScopeTemplates(ctx context.Context, s logical.St
 			String:      template,
 			Entity:      identity.ToSDKEntity(entity),
 			Groups:      identity.ToSDKGroups(groups),
+			NamespaceID: ns.ID,
+		})
+		if err != nil {
+			i.Logger().Warn("error populating OIDC token template", "scope", scope,
+				"template", template, "error", err)
+		}
+
+		if populatedTemplate != "" {
+			claimsMap := make(map[string]interface{})
+			if err := json.Unmarshal([]byte(populatedTemplate), &claimsMap); err != nil {
+				i.Logger().Warn("error parsing OIDC template", "template", template, "err", err)
+			}
+
+			// Check top-level claim keys for conflicts with other scopes
+			for claimKey := range claimsMap {
+				if conflictScope, ok := claimsToScopes[claimKey]; ok {
+					return nil, true, fmt.Errorf("found scopes with conflicting top-level claim: claim %q in scopes %q, %q",
+						claimKey, scope, conflictScope)
+				}
+				claimsToScopes[claimKey] = scope
+			}
+
+			populatedTemplates = append(populatedTemplates, populatedTemplate)
+		}
+	}
+
+	return populatedTemplates, false, nil
+}
+
+// populateScopeTemplates populates the templates for each of the passed scopes.
+// Returns a slice of the populated JSON template strings and a bool to indicate
+// if a conflict in scope template claims occurred.
+func (i *IdentityStore) populateScopeTemplatesWithoutEntity(ctx context.Context, s logical.Storage, ns *namespace.Namespace, scopes ...string) ([]string, bool, error) {
+	// Gather the templates for each scope
+	templates, err := i.getScopeTemplates(ctx, s, scopes...)
+	if err != nil {
+		return nil, false, err
+	}
+
+	claimsToScopes := make(map[string]string)
+	populatedTemplates := make([]string, 0)
+	for scope, template := range templates {
+		// Parse and integrate the populated template. Structural errors with the template
+		// should be caught during configuration. Errors found during runtime will be logged.
+		_, populatedTemplate, err := identitytpl.PopulateString(identitytpl.PopulateStringInput{
+			Mode:        identitytpl.JSONTemplating,
+			String:      template,
 			NamespaceID: ns.ID,
 		})
 		if err != nil {
--- a/vault/identity_store_oidc_provider.go
+++ b/vault/identity_store_oidc_provider.go
@@ -86,13 +86,6 @@ const (
 	ErrAuthMaxAgeReAuthenticate = "max_age_violation"
 )
 
-type GrantType string
-
-const (
-	AuthorizationCode GrantType = "authorization_code"
-	ClientCredentials GrantType = "client_credentials"
-)
-
 type assignment struct {
 	GroupIDs  []string `json:"group_ids"`
 	EntityIDs []string `json:"entity_ids"`
@@ -103,22 +96,58 @@ type scope struct {
 	Description string `json:"description"`
 }
 
+type GrantType string
+
+const (
+	AuthorizationCode GrantType = "authorization_code"
+	ClientCredentials GrantType = "client_credentials"
+)
+
+func parseGrantType(s string) (GrantType, error) {
+	switch s {
+	case "authorization_code":
+		return AuthorizationCode, nil
+	case "client_credentials":
+		return ClientCredentials, nil
+	default:
+		return "", errors.New("unknown color: " + s)
+	}
+}
+
 type client struct {
 	// Used for indexing in memdb
 	Name        string `json:"name"`
 	NamespaceID string `json:"namespace_id"`
 
 	// User-supplied parameters
-	RedirectURIs   []string      `json:"redirect_uris"`
-	Assignments    []string      `json:"assignments"`
-	Key            string        `json:"key"`
-	IDTokenTTL     time.Duration `json:"id_token_ttl"`
-	AccessTokenTTL time.Duration `json:"access_token_ttl"`
-	Type           clientType    `json:"type"`
+	RedirectURIs      []string      `json:"redirect_uris"`
+	Assignments       []string      `json:"assignments"`
+	Key               string        `json:"key"`
+	IDTokenTTL        time.Duration `json:"id_token_ttl"`
+	AccessTokenTTL    time.Duration `json:"access_token_ttl"`
+	Type              clientType    `json:"type"`
+	AuthorizationCode bool          `json:"authorization_code"`
+	ClientCredentials bool          `json:"client_credentials"`
 
 	// Generated values that are used in OIDC endpoints
 	ClientID     string `json:"client_id"`
 	ClientSecret string `json:"client_secret"`
+}
+
+func (c *client) isAuthorizationFlow() bool {
+	return c.AuthorizationCode
+}
+
+// isFlowAllowed returns true if the given client ID is allowed to use the client flow
+func (c *client) isFlowAllowed(credentials GrantType) bool {
+	switch credentials {
+	case AuthorizationCode:
+		return c.AuthorizationCode
+	case ClientCredentials:
+		return c.ClientCredentials
+	default:
+		return false
+	}
 }
 
 type clientType int
@@ -335,6 +364,16 @@ func oidcProviderPaths(i *IdentityStore) []*framework.Path {
 					Type:        framework.TypeString,
 					Description: "The client type based on its ability to maintain confidentiality of credentials. The following client types are supported: 'confidential', 'public'. Defaults to 'confidential'.",
 					Default:     "confidential",
+				},
+				"authorization_code": {
+					Type:        framework.TypeBool,
+					Default:     true,
+					Description: "Whether or not to authorization code flow is allowed in this provider",
+				},
+				"client_credentials": {
+					Type:        framework.TypeBool,
+					Default:     false,
+					Description: "Whether or not to client credentials flow is allowed in this provider",
 				},
 			},
 			Operations: map[logical.Operation]framework.OperationHandler{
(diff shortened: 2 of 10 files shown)


DrDaveD:
This looks like a very useful feature, thank you.  Client credentials flow is very powerful so this is going to have to be reviewed carefully for security.  

There are quite a few references to oidc-provider in the ui.  I'm not sure but some of that may need to be updated.  I haven't used the oidc-provider feature so I'm not sure what the ui is capable of now.

The relevant documentation that will need to be updated is at least the [concepts documentation](https://github.com/openbao/openbao/blob/main/website/content/docs/concepts/oidc-provider.mdx) and the [api documentation](https://github.com/openbao/openbao/blob/main/website/content/docs/secrets/identity/oidc-provider.mdx) and I'm not sure if there might be other places.

Syndlex:
@DrDaveD Thanks for the links! added the Docs.

:) 
I was a little afraid to implement this... But the Flow is supper simple. :D

One think that I didn't know was what value I should use for.
`Audience`
 I used `client.ClientID` 
This is relatively important but since openbao is not implementing token exchange etc its not to bad I think.

Keycloak is using the clientID in the basic cases I found/remember.

DrDaveD:
That doesn't sound right for audience.  The [`aud` claim](https://mojoauth.com/glossary/jwt-audience/) is supposed to be a string used to identify the resource provider that accepts the JWT for authorization, not the client that requests the token.  Typically it is a [URL for the service](https://stackoverflow.com/questions/28418360/jwt-json-web-token-audience-aud-versus-client-id-whats-the-difference).

DrDaveD:
Oh I guess an audience of the client ID can make sense for an ID token, which is only intended to go back to the client.  Maybe that's what you meant.

Syndlex:
The bitnami images have moved a few days ago.

Do you fix this up steam?

DrDaveD:
> The bitnami images have moved a few days ago.
> 
> Do you fix this up stream?

I had never heard of bitnami until this.  However, researching it, it appears they postponed their moves until next month.  In any case, what do you see as the impact on openbao?  It does appear that there's one or two references to bitnami in the code.

vault/identity_store_oidc_provider.go
```diff
@@ -1991,10 +1996,99 @@ func (i *IdentityStore) pathOIDCToken(ctx context.Context, req *logical.Request,
...
+	case "client_credentials":
+		return i.clientCredentialsFlow(ctx, req, d, ns, clientID, name, client, key, provider)
+	default:
 		return tokenResponse(nil, ErrTokenUnsupportedGrantType, "unsupported grant_type value")
 	}
 
+}
+
+func (i *IdentityStore) clientCredentialsFlow(ctx context.Context, req *logical.Request, d *framework.FieldData, ns *namespace.Namespace, id string, name string, client *client, key *namedKey, provider *provider) (*logical.Response, error) {
+	// Since Credentials Flow does not use authorize endpoint Scopes come direcly from Token endpoint
```
DrDaveD: typo `direcly` -> `directly`

DrDaveD (changes requested):
It appears we don't have active reviewers who have much experience with the Oidc provider so we're going to have to rely a lot on your knowledge and testing and only superficial code reviews from others.

I would like to see more effort done on reducing duplication of code.  It looks like the `clientCredentialsFlow` and `populateScopeTemplatesWithoutEntity` functions are largely copied without making any common functions for the parts that don't change.

Syndlex:
@DrDaveD Bitnami has a lot of production ready Container images. They were bought by Qualcomm with the Citrix merger.

All container Images will move into another repository and will not be maintained anymore. You need to find new images. Or ask if a open source project can use them somehow.

---
Yes I can merge those methods. They are somewhat ugly to merge because the de deduplication will result in ugly if cases.

DrDaveD:
The bitnami discussion is off-topic and I suggest putting it into a separate issue.  (But in case someone looks at it here, according to what I have read it is owned by Broadcom, not Qualcomm, via VMWare, not Citrix.)

Syndlex:
@DrDaveD I added the requested changes.
I really like how I divided the two flows, because you can clearly see how the tokens are filled at the end.
If you’d like, I can try to use the same initialization technique I used in `populateScopeTemplates`, but I think that would lead to some insecure code.

:D Sry replied on my phone. That's correct!
The Bitnami think is a other Issue. The Tests fail because the cassandra image is missing `No such image: bitnami/cassandra:3.11.11`

vault/identity_store_oidc_provider.go
```diff
@@ -1988,13 +1991,91 @@ func (i *IdentityStore) pathOIDCToken(ctx context.Context, req *logical.Request,
...
+	// Populate each of the requested scope templates
+	templates, conflict, err := i.populateScopeTemplates(ctx, ClientCredentials, req.Storage, ns, nil, scopes...)
+	if !conflict && err != nil {
+		return tokenResponse(nil, ErrTokenServerError, err.Error())
 	}
-	if grantType != "authorization_code" {
-		return tokenResponse(nil, ErrTokenUnsupportedGrantType, "unsupported grant_type value")
+	if conflict && err != nil {
+		return tokenResponse(nil, ErrTokenInvalidRequest, err.Error())
+	}
```
DrDaveD: Suggest to replace that with
```
    if err != nil {
        if conflict {
            return tokenResponse(nil, ErrTokenInvalidRequest, err.Error())
        } else {
            return tokenResponse(nil, ErrTokenServerError, err.Error())
        }
    }
```
because it is easier to read at a glance.  Also below in authorizationCodeFlow.

Syndlex:
> I don't know what initialization technique you're referring to or where you're suggesting to use it. Please elaborate and tell me why you think that would be insecure.

at the moment I create the Struct in one shot.

in `populateScopeTemplates` I created the struct, filled half of the it in the initialization and the other half later depending on the Grant Type. We could remove `clientCredentialsFlow` and `authorizationCodeFlow` and do everything "in the flow".

vault/identity_store_oidc_provider.go
Syndlex: nice Change! Looks like some KPI optimization from the old Vault code :D

DrDaveD (changes requested):
There's still a gofumpt linter error.

> in `populateScopeTemplates` I created the struct, filled half of the it in the initialization and the other half later depending on the Grant Type. We could remove `clientCredentialsFlow` and `authorizationCodeFlow` and do everything "in the flow".

That would be OK with me if it eliminated more duplication of code.  You think that would be insecure?  How?

Speaking of security, I haven't spotted anything that would prevent any client id from using the client credentials flow.  Did I miss it?  I think that should only be allowed with a client explicitly configured for it, not just any client id.

Syndlex:
> There's still a gofumpt linter error. 

Wubs

> That would be OK with me if it eliminated more duplication of code. You think that would be insecure? How?

In my opinion filling the struct part by part is convoluted and hart to maintain. It will produce a long method with  a lot of code paths. A lot of code in the  `authorizationCodeFlow` is not needed in the `clientCredentialsFlow`.
Since the token is the most importend part is the token at the end I would let initialize at once so you can see in one place what is in the token without a lot of mental load.

> Speaking of security, I haven't spotted anything that would prevent any client id from using the client credentials flow. Did I miss it? I think that should only be allowed with a client explicitly configured for it, not just any client id.

This is in the caller Method  Currently line [1973](https://github.com/openbao/openbao/pull/1732/files#diff-bb220ba21b3773423ab4f14b05e8693ffdcc9cdeb9c928e66716ba0dd0986712R1973)

DrDaveD:
> In my opinion filling the struct part by part is convoluted and hart to maintain. It will produce a long method with a lot of code paths. A lot of code in the `authorizationCodeFlow` is not needed in the `clientCredentialsFlow`. Since the token is the most importend part is the token at the end I would let initialize at once so you can see in one place what is in the token without a lot of mental load.

Ok.

> > Speaking of security, I haven't spotted anything that would prevent any client id from using the client credentials flow. Did I miss it? I think that should only be allowed with a client explicitly configured for it, not just any client id.
> 
> This is in the caller Method Currently line [1973](https://github.com/openbao/openbao/pull/1732/files#diff-bb220ba21b3773423ab4f14b05e8693ffdcc9cdeb9c928e66716ba0dd0986712R1973)

That looks like it is just checking for if a client ID is valid or not.  What prevents a valid client from using the credentials flow if it is only supposed to use the authorization code flow?

Syndlex:
> That looks like it is just checking for if a client ID is valid or not. What prevents a valid client from using the credentials flow if it is only supposed to use the authorization code flow?

Oh that's correct. Added validation.

vault/identity_store_oidc_provider.go
```diff
@@ -93,24 +93,60 @@ type scope struct {
...
-	IDTokenTTL     time.Duration `json:"id_token_ttl"`
-	AccessTokenTTL time.Duration `json:"access_token_ttl"`
-	Type           clientType    `json:"type"`
+	RedirectURIs      []string      `json:"redirect_uris"`
+	Assignments       []string      `json:"assignments"`
+	Key               string        `json:"key"`
+	IDTokenTTL        time.Duration `json:"id_token_ttl"`
+	AccessTokenTTL    time.Duration `json:"access_token_ttl"`
+	Type              clientType    `json:"type"`
+	AuthorizationCode bool          `json:"authorization_code"`
```
cipherboy:
```suggestion
	AuthorizationCode *bool          `json:"authorization_code"`
```

Shouldn't this default to `true` since it is the existing behavior? On upgrade of an existing role, I think you'd have this set to `false` if I'm not mistaken. The logic elsewhere will need to be adjusted to default `nil` to `true`.

vault/identity_store_oidc_provider.go
```diff
@@ -1708,6 +1766,9 @@ func (i *IdentityStore) pathOIDCAuthorize(ctx context.Context, req *logical.Requ
 	if client == nil {
 		return authResponse("", state, ErrAuthInvalidClientID, "client with client_id not found")
 	}
+	if client.isAuthorizationFlow() {
+		return authResponse("", state, ErrTokenInvalidGrant, "code credential flow is not vor client allowed")
```
cipherboy:
```suggestion
		return authResponse("", state, ErrTokenInvalidGrant, "client not allowed to perform authorization code credential flow")
```

perhaps?

vault/identity_store_oidc_provider.go
```diff
@@ -1987,14 +2039,99 @@ func (i *IdentityStore) pathOIDCToken(ctx context.Context, req *logical.Request,
...
-		return tokenResponse(nil, ErrTokenInvalidRequest, "grant_type parameter is required")
+	grantType, err := parseGrantType(d.Get("grant_type").(string))
+	if err != nil {
+		return tokenResponse(nil, ErrTokenUnsupportedGrantType, "unsupported grant_type or parameter is empty value")
 	}
-	if grantType != "authorization_code" {
-		return tokenResponse(nil, ErrTokenUnsupportedGrantType, "unsupported grant_type value")
+	// Validate that the client allows auth flow to use the provider
+	if !client.isFlowAllowed(grantType) {
+		return tokenResponse(nil, ErrTokenInvalidGrant, "client is not authorized to use client credentials flow")
```
cipherboy:
```suggestion
		return tokenResponse(nil, ErrTokenInvalidGrant, "client is not authorized to use given credentials flow")
```

vault/identity_store_oidc_provider.go
Syndlex: :D Thanks didn't know this about Go. :( 

```go
type Config struct {
    Enabled bool `json:"enabled"`
}

func main() {
    cfg := Config{
        Enabled: true, // Default
    }

    // Überschreibt nur, wenn "enabled" im JSON vorhanden ist
    jsonData := []byte(`{}`)
    _ = json.Unmarshal(jsonData, &cfg)

    fmt.Println(cfg.Enabled) // true
}
```

I also looked if something like this would be a good idea. But I think this is hart to handle.

Another solution would be to read all clients on startup and add the value. 
Do you have some migration infrastructure for changes within the data model?

vault/identity_store_oidc_provider.go
```diff
@@ -1147,6 +1203,20 @@ func (i *IdentityStore) pathOIDCCreateUpdateClient(ctx context.Context, req *log
...
+	} else if req.Operation == logical.CreateOperation {
+		b := d.Get("authorization_code").(bool)
+		client.AuthorizationCode = &b
+	}
+
+	if clientCredentials, ok := d.GetOk("client_credentials"); ok {
+		client.ClientCredentials = clientCredentials.(bool)
+	} else if req.Operation == logical.CreateOperation {
+		client.ClientCredentials = d.Get("client_credentials").(bool)
+	}
```
DrDaveD, line 1222: This doesn't look right.  What's going on there?

vault/identity_store_oidc_provider.go
```diff
@@ -105,12 +123,35 @@ type client struct {
 	IDTokenTTL     time.Duration `json:"id_token_ttl"`
 	AccessTokenTTL time.Duration `json:"access_token_ttl"`
 	Type           clientType    `json:"type"`
+	// We want AuthorizationCode to be true by default if it is not set already.
+	AuthorizationCode *bool `json:"authorization_code"`
+	ClientCredentials bool  `json:"client_credentials"`
```
DrDaveD, line 132: @cipherboy is this what you intended?  It seems odd to have two parallel variables where one is a pointer and one isn't.

ui/app/models/oidc/client.js
```diff
@@ -72,6 +72,18 @@ export default class OidcClientModel extends Model {
   })
   idTokenTtl;
 
+  @attr('boolean', {
+    defaultValue: false,
```
DrDaveD: This should probably default to true.

vault/identity_store_oidc_provider.go
Syndlex: Json serialization with a missing value is false. Since default of this value ist default. Using bool is ok.

The oder bool default should be true when reading the DB for the first time after an upgrade. So we can treat  a missing value as true.
It would be nicer to set in the database and not having the pointer here.

vault/identity_store_oidc_provider.go
Syndlex: I copied this from somewhere else. It is used in a lot of places.

DrDaveD: Ok.  I guess I just haven't seen many CreateUpdate functions.

Syndlex:
@DrDaveD @cipherboy 
I will not be available to make changes for 2-3 week starting Saturday. Is there any chance this can land in this Timeframe?

vault/identity_store_oidc_provider.go
cipherboy: @DrDaveD Yeah, it looks odd in the diff but it makes the value correct. You see the pattern in PKI and Transit and elsewhere for this type of upgrade path. The alternative is negating the variables (`DisallowAuthorizationCode`, `DisallowClientCredentials`) where default false implies true, but I don't think we _actually_ want the default to be false, just the `nil` value to be false.

My 2c.

vault/identity_store_oidc_provider.go
```diff
@@ -1987,14 +2054,99 @@ func (i *IdentityStore) pathOIDCToken(ctx context.Context, req *logical.Request,
...
+		CreationTime:       accessTokenIssuedAt.Unix(),
+		NoIdentityPolicies: true,
+		Meta: map[string]string{
+			"oidc_token_type": "access token",
+		},
+		InternalMeta: map[string]string{
+			accessTokenClientIDMeta: client.ClientID,
+			accessTokenScopesMeta:   strings.Join(scopes, scopesDelimiter),
+		},
+	}
```
cipherboy: Am I completely off base or should this have an `InlinePolicy` directive to allow access? :thinking:

website/content/api-docs/secret/identity/oidc-provider.mdx
```diff
@@ -750,14 +751,17 @@ for an OIDC provider.
   `code`. Required for authorization codes that were granted using [PKCE](https://datatracker.ietf.org/doc/html/rfc7636).
   Required for `public` clients.
 
+- `scope` `(string: <required for client_credentials>)` - A space-delimited list of scopes to be requested. 
+  The `openid` scope is required when using the `client_credentials` grant type.
```
cipherboy, line 755: Is this optional or forbidden when using authorization code flows?

vault/identity_store_oidc_provider_test.go
```diff
@@ -607,6 +610,206 @@ func TestOIDC_Path_OIDC_Token(t *testing.T) {
 	}
 }
 
+func TestOIDC_Path_OIDC_Token_Client_Credentials_Flow(t *testing.T) {
```
cipherboy, line 611: @DrDaveD I'll delegate review of test to you fwiw.

cipherboy (commented):
I'll approve and merge once @DrDaveD approves; a few comments inline but this is looking fairly good to me. Thanks!

vault/identity_store_oidc_provider.go
cipherboy: @Syndlex Typically we do upgrades manually on read. In this case, identity store is not your friend so you'd likely need an `updateClient` called from `storageClientByName`. Then if the role is modified, the new value will be persisted. We don't typically in-place/force upgrade for fear of storage pressure. 

Also, `storageClientByID` is a nasty function, we should file that as a tech-debt label if we keep this identity store.

website/content/api-docs/secret/identity/oidc-provider.mdx
Syndlex: I think OIDC leaves this undefined:

> In a pure OIDC authorization code flow, adding scope to the Token Request is not required and the OIDC spec doesn’t define or assign any meaning to it.
scope is critical in the Authorization Request to determine what claims/resources you're asking for (openid, profile, email, etc.), but it plays no role when exchanging the code for tokens—unless a specific platform explicitly supports it.
If you’re working with a particular identity provider, check their docs—some may allow or ignore scope in the Token Request. But to be safe and interoperable, it's best to omit it unless documentation indicates otherwise.

vault/identity_store_oidc_provider.go
DrDaveD: There is one for the corresponding code in the authorization code flow that enables read/write access to userinfo, but perhaps it isn't needed here.  @Syndlex what do you think?

vault/identity_store_oidc_provider_test.go
```diff
@@ -607,6 +610,206 @@ func TestOIDC_Path_OIDC_Token(t *testing.T) {
...
+				tokenReq: func() *logical.Request {
+					req := testClientCredentialsTokenReq(s, "openid", clientID, clientSecret)
+					req.Data["grant_type"] = ""
+					return req
+				}(),
+			},
+			wantErr: ErrTokenUnsupportedGrantType,
+		},
+		{
+			name: "valid token request",
```
DrDaveD:
```suggestion
			name: "token request with client credentials disabled",
```

vault/identity_store_oidc_provider_test.go
```diff
@@ -1317,17 +1556,19 @@ func testAssignmentReq(s logical.Storage, entityID, groupID string) *logical.Req
...
-			"assignments":      []string{"test-assignment"},
-			"id_token_ttl":     "24h",
-			"access_token_ttl": "24h",
+			"key":                "test-key",
+			"redirect_uris":      []string{"https://localhost:8251/callback"},
+			"assignments":        []string{"test-assignment"},
+			"id_token_ttl":       "24h",
+			"access_token_ttl":   "24h",
+			"client_credentials": clientCredentials,
+			"authorization_code": authorizationCode,
```
DrDaveD: Rather than directly setting these, I would like to see verification of the default settings.  Please leave these unset when you initially create a data map, and only set the values if they are not the default setting.

DrDaveD (changes requested):

vault/identity_store_oidc_provider.go
Syndlex: This is a good question. The User info endpoint is used to retrieve information about a Entity I can understand the Policy there. (claims about the authenticated end-user (e.g., sub, email, name))

Since the Client Flow is meant for Service to Service coms the userinfo endpoint is not needed and shout reject the token.

vault/identity_store_oidc_provider_test.go
Syndlex: I have implemented this but it has one Downside. In `TestOIDC_Path_OIDC_Authorize` the clients are not reset or deleted. This makes perfect sense for testing speed. 

But when I want to test `invalid authorize provider does not allow authorization code flow` `authorization_code` is set to false the state is never set back to the default value. 

I put this test at the end so the test is working.
And noted the reason at the test.

vault/identity_store_oidc_provider.go
cipherboy: @Syndlex I guess then, what does this batch token achieve? It lacks both the default policy and inline policies so it shouldn't access... Anything? Do you have a spec reference as why we need to create it? AFACIT, I think this means we could respond with an invalid/garbage token, unless it is passed to an unauthenticated endpoint or something and used for identity lookup? 🤔

website/content/api-docs/secret/identity/oidc-provider.mdx
cipherboy: Cool, just making sure.

Syndlex:
I am not near a PC for the next 3 weeks. Would be nice if you could do small changes.

If there is anything bigger It has to wait. :) 

Thanks.

vault/identity_store_oidc_provider.go
Syndlex: I think the Batch token is in the `Auth engine` to access Vault. Not a function within the a OIDC Provider. The OIDC Provider is "miss using" existing JWT code from the Auth engine. It allows using existing RBAC infrastructure and reusing code. So we do not have to create extra token validation infrastructure to make the userinfo entpoint save.

AFAIK OIDC and Auth2 do not have a Concept of batch tokens.
This is just a 20 Minute Lookup in the code so I still could be wrong.

> it shouldn't access... Anything?

The token should not have access to anything Vault related. :) for this we have the JWT/OIDC Auth engine.
But it should grant access to the userinfo endpoint.

> AFACIT, I think this means we could respond with an invalid/garbage token, unless it is passed to an unauthenticated endpoint or something and used for identity lookup?

I do not fully understand what you are asking here. 
In case of auth code flow the userinfo endpoint needs a valid id token.
This Token is granted for a specific entity. The userinfo endpoint returns more information about that specific entity and nothing more. It is mostly used so the id token size can be small (header size limits) and private information is behind an extra endpoint. 

For example you could get a id token only containing the entity id without the users real name or email. With a request to the userinfo endpoint any service that has the id token can fetch more information about a user.

> Do you have a spec reference as why we need to create it?

This is all descripted in the spec for OIDC and oAuth2.
https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

cipherboy: @Syndlex Right, but see here:

https://github.com/openbao/openbao/blob/a193c3ada662c071a66da088cbdfd3651b42d46d/vault/identity_store_oidc_provider.go#L2082-L2102

In the existing application token issuance flow, this token is granted an inline policy to access the `userinfo` endpoint (line 2098).

This token that you've created for the client credentials flow cannot access the `userinfo` endpoint _I believe_, because it is:

1. An authenticated endpoint, as it is not present in:
   https://github.com/openbao/openbao/blob/a193c3ada662c071a66da088cbdfd3651b42d46d/vault/identity_store.go#L88-L92
2. This batch token has no policy attached, named or inline.

So while I've not tested it, I'm fairly confident to say that this issued batch token is useless and will 404 on all requests, including to the expected `userinfo` endpoint.

vault/identity_store_oidc_provider_test.go
```diff
@@ -1090,6 +1302,30 @@ func TestOIDC_Path_OIDC_Authorize(t *testing.T) {
 				}(),
 			},
 		},
+		{
+			// This needs to be last because we do not reset authorization_code explicitly so it stays on the set false value after and messes with tests.
```
DrDaveD: I was confused by this.  I think you mean
```suggestion
			// This needs to be last because we do not reset authorization_code explicitly in testAuthorizeReq so
			//  it stays on the set false value after and messes with tests.
```

DrDaveD: I do not see the corresponding test for the default value of `client_credentials`.  Did I miss it?

DrDaveD (changes requested):

vault/identity_store_oidc_provider.go
Syndlex: Oh yes. Now I understand. For bao it's useless.

Client flow tokens do not need access to the user info endpoint. Since they do not hold user information. Are not representing a user.

I don't fully know if we can delete it entirely because you can still use the access token since it contains scopes and can be used for service to service access.

DrDaveD: @Syndlex it seems to me that the only way the token could be useful service-to-service is if it is a JWT, and it isn't clear to me that that is what is being created.  If it is an openbao token, that would be opaque so it would be useless if it can't be used to access any bao APIs.

vault/identity_store_oidc_provider_test.go
Syndlex: Authorize endpoint: default behavior is tested implicitly every call.

Token endpoint has tests in. `TestOIDC_Path_OIDC_Token_Client_Credentials_Flow` unless i am not understanding you correctly here.

vault/identity_store_oidc_provider.go
Syndlex: Ohh yes you are correct. 

I don't need this code here. Creating a ID token is also wrong. I forgot that Keycloak is strange in that case.

DrDaveD:
Please rebase on the current main branch so CI checks can run.

vault/identity_store_oidc_provider_test.go
DrDaveD: What I mean is, where is `client_credentials` left unset and verified that it has defaulted to false?

Syndlex:
@DrDaveD rebased... uff runnig `make fmt` is changing 37 files. linting has some errors in files that I didn't touch.

vault/identity_store_oidc_provider_test.go
Syndlex: oh ok.

I can't find any angle how to test this.
At least I haven't found any other test that checks default Values. 

Since false is the default behavior missing json values I wouldn't test this.

DrDaveD: It's not enough to just leave `client_credentials` unset in `testAuthorizationCodeTokenReq`?  That was my original suggestion, but I'm not really understanding the full flow.  My point was to make sure that client credentials is not turned on unless somebody explicitly sets it to true.

Syndlex: ok. I hope `token request with client credentials disabled (by default)` will be sufficient

DrDaveD:
@cipherboy This is waiting on you.

PR #1733, opened 2025-08-26
Implement cache invalidation for mount tables

phil9909:
Implement cache invalidation for mount tables

See #1550
See #1528 

### DIFF
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -2706,9 +2706,9 @@ func (c *Core) reloadMountInternal(ctx context.Context, table, uuid string, desi
 
 type mountInvalidationWorker struct {
 	l                      sync.Mutex
-	invalidations          map[mountInvalidation]string
-	namespaceInvalidations map[string]string
-	legacyInvalidations    map[string]string
+	invalidations          map[mountInvalidation]struct{}
+	namespaceInvalidations map[string]struct{}
+	legacyInvalidations    map[string]struct{}
 	notify                 chan struct{}
 
 	core *Core
@@ -2716,9 +2716,9 @@ type mountInvalidationWorker struct {
 
 func newMountInvalidationWorker(c *Core) *mountInvalidationWorker {
 	return &mountInvalidationWorker{
-		invalidations:          map[mountInvalidation]string{},
-		namespaceInvalidations: map[string]string{},
-		legacyInvalidations:    map[string]string{},
+		invalidations:          map[mountInvalidation]struct{}{},
+		namespaceInvalidations: map[string]struct{}{},
+		legacyInvalidations:    map[string]struct{}{},
 		core:                   c,
 		notify:                 make(chan struct{}, 1),
 	}
@@ -2746,9 +2746,9 @@ func (w *mountInvalidationWorker) dispatchReloads(ctx context.Context) error {
 	invalidations := w.invalidations
 	namespaceInvalidations := w.namespaceInvalidations
 	legacyInvalidations := w.legacyInvalidations
-	w.invalidations = map[mountInvalidation]string{}
-	w.namespaceInvalidations = map[string]string{}
-	w.legacyInvalidations = map[string]string{}
+	w.invalidations = map[mountInvalidation]struct{}{}
+	w.namespaceInvalidations = map[string]struct{}{}
+	w.legacyInvalidations = map[string]struct{}{}
 	w.l.Unlock()
 
 	for invalidation := range invalidations {
@@ -2795,7 +2795,7 @@ func (w *mountInvalidationWorker) invalidateMount(namespaceUUID, key string) {
 	w.invalidations[mountInvalidation{
 		namespaceUUID: namespaceUUID,
 		key:           key,
-	}] = ""
+	}] = struct{}{}
 	w.trigger()
 }
 
@@ -2803,7 +2803,7 @@ func (w *mountInvalidationWorker) invalidateNamespaceMounts(namespaceUUID string
 	w.l.Lock()
 	defer w.l.Unlock()
 
-	w.namespaceInvalidations[namespaceUUID] = ""
+	w.namespaceInvalidations[namespaceUUID] = struct{}{}
 	w.trigger()
 }
 
@@ -2811,6 +2811,6 @@ func (w *mountInvalidationWorker) invalidateLegacyMounts(key string) {
 	w.l.Lock()
 	defer w.l.Unlock()
 
-	w.legacyInvalidations[key] = ""
+	w.legacyInvalidations[key] = struct{}{}
 	w.trigger()
 }


vault/core_cache_invalidate.go
```diff
@@ -87,6 +93,13 @@ func (c *Core) invalidateInternal(ctx context.Context, key string) error {
 			}
 		}()
 
+	case namespacedKey == coreMountConfigPath || namespacedKey == coreLocalMountConfigPath:
```
cipherboy: Just a thought I had, should `c.router.Invalidate(...)` be last?

vault/mount.go
```diff
@@ -2386,3 +2386,217 @@ func (c *Core) mountEntryView(me *MountEntry) (BarrierView, error) {
...
+		err := c.invalidateMountInternal(ctx, key)
+		if err != nil {
+			c.logger.Error("unable to invalidate mount, restarting core", "key", key, "error", err.Error())
+			// TODO(phil9909) call c.restart()
+			return
+		}
+	}()
+}
+
+func (c *Core) invalidateNamespaceMounts(ctx context.Context, uuid string) {
```
cipherboy: Is this called from somewhere? :eyes:

phil9909: Yes, here https://github.com/openbao/openbao/blob/a9d73af5c46fcb9e11dcd24adfb40d8588dcf01a/vault/core_cache_invalidate.go#L55-L69


Otherwise this test would fail: https://github.com/openbao/openbao/blob/a9d73af5c46fcb9e11dcd24adfb40d8588dcf01a/vault/core_cache_invalidate_test.go#L118-L124

vault/core_cache_invalidate.go
phil9909: Yes, it should be. Thanks, i changed it.

vault/mount.go
cipherboy: @phil9909 Ah, I had missed that. Already clicked "mark file as viewed" when I went searching so I didn't find it. :facepalm:

cipherboy:
@phil9909 Where are we at with this one? Do you want me to rebase the base branch on top of latest main?

phil9909:
> @phil9909 Where are we at with this one? Do you want me to rebase the base branch on top of latest main?

Yes, I'm still working on this. Currently trying to implement it for non-transactional storage, but that could be a different follow up PR.
Anyway, feel free to rebase.

vault/core_cache_invalidate.go
```diff
@@ -9,12 +9,12 @@ import (
...
 
 	"github.com/openbao/openbao/helper/namespace"
+	"github.com/openbao/openbao/sdk/v2/physical"
 	"github.com/openbao/openbao/vault/quotas"
 )
 
 func (c *Core) Invalidate(key string) {
-	ctx, cancel := context.WithTimeout(c.activeContext, 2*time.Second)
-	defer cancel()
+	ctx, _ := context.WithTimeout(c.activeContext, 2*time.Second) //nolint:govet // we can't cancel this context, because invalidation is asynchronous
```
cipherboy: Sorry, why is _invalidation_ not synchronous?

We don't want the actual reload to be synchronous, but the invalidation could be synchronous here, if we're doing asynchronicity in the physical layers to kick off invalidation.

I seem to recall that in PKI we had be careful not to use the invalidate context for anything because of it being cancelled after returning from the PKI handler.  

My 2c.

vault/core_cache_invalidate.go
```diff
@@ -52,7 +52,13 @@ func (c *Core) invalidateInternal(ctx context.Context, key string) error {
 	switch {
 	case strings.HasPrefix(namespacedKey, namespaceStoreSubPath):
 		c.namespaceStore.invalidate(ctx, "")
-		c.policyStore.invalidateNamespace(ctx, strings.TrimPrefix(namespacedKey, namespaceStoreSubPath))
+
+		ctx := physical.CacheRefreshContext(ctx, true)
```
cipherboy, line 68: Why would we add the refresh context, when we've invalidated the key? _If_ we were to do this, and I'm not sure it is a good idea to do it here (though, it maybe a good idea to do it in each of the actual _reload_ functions), we should probably add a comment and apply it more broadly and not just to this particular operation (e.g., apply it consistently in all areas of reload). 

And like I said above, I thought we had agreed to the actual invalidation being storage-less. 

Maybe I'm missing something though :-D

vault/mount.go
```diff
@@ -691,6 +692,10 @@ func (c *Core) mountInternal(ctx context.Context, entry *MountEntry, updateStora
 	}
 	defer unlock()
 
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
 	ns, err := namespace.FromContext(ctx)
```
cipherboy, line 700: Wonder if this should just check for context cancellation? 🤔 

Curious why you added the context check.

vault/mount.go
```diff
@@ -2378,3 +2383,343 @@ func (c *Core) mountEntryView(me *MountEntry) (BarrierView, error) {
 
 	return nil, errors.New("invalid mount entry")
 }
+
+func (c *Core) invalidateMount(ctx context.Context, key string) {
+	go func() {
```
cipherboy: I see, you're running this asynchronously from the invalidate thread rather than through some other mechanism. Hmm. 🤔 

I still go back to wanting a consistent, global pattern for this.

The existing pattern already in this repo was:

1. Invalidate was ordered (impossible, due to changes in ).
2. Invalidate was synchronous.
3. Invalidate marked entries as invalid, it did not reload them.
4. A later process checked for stale and reloaded. 

It seems like we're changing all of the stages at once, which I think is rather risky. I'm happy changing 1, it is a necessary design change. But 2-4 may be hard to adjust right now.

(I push on this because I want there to be a very consistent pattern for all other plugin developers to follow. Mixing patterns especially w.r.t. 3 & 4 will cause issues when it comes to plugins &c, as there's no storage reference there.)

phil9909: 3 is what I wanted to avoid for the mount table, because in some cases (non transactional storage) we would have to mark the whole mount table and router as dirty and stall the next incoming request (and all parallel requests) for the reload to happen or we trigger a asynchronous reload on the next access, but that IMHO just complicates things and delays the "correction" of the mount table.

Also, let's say you have deleted a mount, now we flag it in the mount table as stale. When would it actually get unmounted? Once it is accessed the next time? Which might be never (after all you deleted it, which you usually do, if you don't need it anymore). Or would you always mark the whole table as stale?

One alternative I see would be a dedicated worker, that reads from a channel and does the invalidations. But would this be a buffered channel? Probably you would need a queue instead and use the channel only to trigger a run.

If we need a global pattern, why not change the plugin interface to include access to the storage? Not having access to storage is an SDK only limitation right? so plugins can already do it. If you want it compatible without a `InvalidateWithStorageFunc` we can throw the storage into the context.  
But we might just admit that we shouldn't try to tar everyone with the same brush, maybe there is enough difference between the key that stores the mount table and the key that stores a single kv entry to justify having different patterns.

vault/core_cache_invalidate.go
phil9909: We invalidated the namespace path, but not the policy store and mounts. That's why this is specific to this case, but I could push that down to the individual methods.

vault/mount.go
phil9909: > Wonder if this should just check for context cancellation? 🤔

Huh? That's exactly what I'm doing here.

> Curious why you added the context check.

Sometimes the test cleanup crashed. If I recall correctly, this only happened when a test case failed (and the cleanup started while there were still invalidations running in the background). But it could happen in production.

cipherboy: @phil9909 Gotcha. Sorry, I wasn't precise: I wonder if you could just check for context cancellation in the `namespace.FromContext(...)` helper. It seems redundant, but adds another check for context cancellation in many places where we don't have it (which, may be a good thing), and simplifies a bunch of code that would otherwise do both.

vault/mount.go
cipherboy: @phil9909 said:

> 3 is what I wanted to avoid for the mount table, because in some cases (non transactional storage) we would have to mark the whole mount table and router as dirty and stall the next incoming request (and all parallel requests) for the reload to happen or we trigger a asynchronous reload on the next access, but that IMHO just complicates things and delays the "correction" of the mount table.

But aren't we anyways blocking the whole mount table for the update?

Reloading the mount table acquires an exclusive write lock, which requires stalling the next request. We know (through Rollback manager, which executes every minute) that even at complete idle we have requests flowing through the mount table, and so we'll always have a bounded time-to-invalidate.

The current code is not safe either: we don't acquire an exclusive lock before we read the entry in storage, which means we violate the ordering set forth everywhere else (except for very first initialization, before requests can be routed anyways). So we need to change the ordering (making lock acquisition happen before storage read), leading to even more equivalencies between the two.

In the current code, this means that in the time it takes to acquire an exclusive lock, we might have performed another exclusive lock operation and modified this mount table and subsequently (after invalidation got the lock), overwrote it in memory with the stale version.

My 2c., but this is a well-established pattern that we should likely continue. If we want to refactor this globally, we should consider doing it elsewhere.

cipherboy: @phil9909 said:

> One alternative I see would be a dedicated worker, that reads from a channel and does the invalidations. But would this be a buffered channel? Probably you would need a queue instead and use the channel only to trigger a run.

The current implementation incurs a goroutine per-mount. In the case of a large volume of new mounts (initial provisioning?), we can flood the standbys with lots of goroutines as they try to catch up to the new mount table. 

https://github.com/adfinis-forks/openbao/blob/e3f987d7a8e69cfc056d8bb3925ad8d3722b831e/vault/mount.go#L2387-L2395

This means we can likely bring a standby node down (as goroutines consume resources -- mount process is otherwise serial iirc) by creating or deleting many mounts very quickly. (`benchmark-openbao` supports creation at any rate).

vault/mount.go
```diff
@@ -2729,3 +2703,114 @@ func (c *Core) invalidateMountInternal(ctx context.Context, table, uuid string,
...
+	notify                 chan struct{}
+
+	core *Core
+}
+
+func newMountInvalidationWorker(c *Core) *mountInvalidationWorker {
+	return &mountInvalidationWorker{
+		invalidations:          map[mountInvalidation]string{},
+		namespaceInvalidations: map[string]string{},
+		legacyInvalidations:    map[string]string{},
```
cipherboy:
```suggestion
		namespaceInvalidations: map[string]struct{}{},
		legacyInvalidations:    map[string]struct{}{},
```

phil9909:
> The diff is rather hard to follow

Sorry about that

cipherboy:
Thanks @phil9909!

PR #1742, opened 2025-08-28
Bump critical & go-kms-wrapping

cipherboy:
This runs `make bump-critical` with extensions to also bump go-kms-wrapping dependencies. These had to be pulled manually in some cases as the latest module was not automatically detected for these new packages.

See also: https://github.com/openbao/go-kms-wrapping/tags


sdk/go.mod
```diff
@@ -5,7 +5,9 @@ module github.com/openbao/openbao/sdk/v2
 // automatically track the Go version used to build Vault itself.  Many projects import
 // the sdk module and we don't want to impose a newer version on them any more than we
 // have to.
-go 1.24
+go 1.24.0
+
+toolchain go1.24.6
```
satoqz: Intentional?

cipherboy: @satoqz Hmm, it appears to be because of the bump to `golang.org/x/exp@v0.0.0-20250819193227-8b4c13bb791b`:

https://github.com/golang/exp/commit/42675adae3e697041d19b22cb15349e9c361f430#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6

cipherboy: Namely, putting a `.0` line for `go` seems to also include the `toolchain` directive, which then results in the latest go toolchain being pulled as the default value.

I could manually overwrite this to 1.24.3 to match `api/`, but not sure that buys us much over 1.24.6?

cipherboy: @satoqz I'm rather confused though why we end up with `toolchain` directives and `x/exp` doesn't.

satoqz: @cipherboy hmm it must be something else, a standalone `make bump-critical` on current `main` does effect the `.0`, but not the `toolchain` directive.

FWIW, you can just remove the toolchain directive and tidy won't complain.

satoqz: Ha, I _do_ get the toolchain directive when I `make bump-critical` with Go 1.24 in path, but not with Go 1.25 😄 (tidy doesn't complain after removal on either version)

cipherboy: @satoqz Interesting. On my work computer with Go **1.23.6** selected (last release, random minor version), the toolchain is not added. On my personal laptop with 1.24.6, it injects the toolchain line fairly reliably.

JanMa (approved):

PR #1745, opened 2025-08-28
Fix release-dist checkout failure

DrDaveD:
The [2.4.0 release-dist step](https://github.com/openbao/openbao/actions/runs/17307089336/job/49135074501) failed and this should prevent it from happening again.


cipherboy (approved):
We'll give this a go next time!

cipherboy:
Thanks @DrDaveD and @wslabosz-reply!

PR #1749, opened 2025-08-31
Update goreleaser to latest, add license

cipherboy:
After the previous issues fixed in goreleaser v2.8, we had pinned to an earlier v2.5.x series, which was stable but no longer supported. This updates the pinning to the latest version.

We also introduce the goreleaser license key.

---

Running: https://github.com/openbao/openbao-nightly/actions/runs/17357488329

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -270,8 +270,7 @@ jobs:
         uses: goreleaser/goreleaser-action@v6
         with:
           install-only: true
-          # TODO: remove version pinning when Goreleaser 2.8 is released
-          version: v2.5.1
+          version: v2.11.2
 
       - name: Install C compiler for arm64 CGO cross-compilation
         if: inputs.dogorelease && matrix.release_os == 'hsm'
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -294,6 +293,7 @@ jobs:
           GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
           NFPM_DEFAULT_PASSPHRASE: ${{ secrets.GPG_PASSWORD }}
           NIGHTLY_RELEASE: ${{ inputs.nightly }}
+          GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
```
satoqz, line 296: I guess this has no effect yet since the workflow doesn't download the pro distribution? https://github.com/goreleaser/goreleaser-action?tab=readme-ov-file#workflow

cipherboy:
It looks like v2.11.x has issues with Docker builds and they're working on rewriting the functionality in an upcoming v3 release, so we'll wait for that before upgrading I guess.

Issue #1752, opened 2025-09-01
TOTP-enabled Userpass authentication issues "bad token" / "Not authorized"


AcronicaOpica:
**Describe the bug**
I am attempting to use the userpass authentication method with enforced TOTP. The initial login process (via both CLI and GUI) appears successful and returns a token. However, any subsequent API calls (including bao token lookup for the issued token) fail with a bad token or permission denied error (Ember Data Request GET /v1/sys/internal/ui/mounts returned a 403 Payload (application/json) [object Object]), even when the token is assigned a full-access policy.
It worked on version 2.3.1.

**To Reproduce**
1. Install, initialize and log in to OpenBao 2.4.0
2. Enable method Username & Password and create new user
3. Create new policy with full rights: path "*" { capabilities = ["create", "read", "update", "delete", "list", "sudo"]} and assign it to the user
4. Log in as a user to the gui
5. Enable MFA method TOTP with Period 30, Key size 20, QR size 200, Algorithm SHA1, Digits 6, Skew 0, Max attempts 5 with enforcement for the entity of new user
6. Run MFA setup and finish it (in my case Microsoft Authenticator on Android)
7. Log out and try logging in with your name, password and TOTP code. It will be successful but with error: Ember Data Request GET /v1/sys/internal/ui/mounts returned a 403 Payload (application/json) [object Object]

**Expected behavior**
Upon successful authentication with TOTP, a valid token should be issued that has all specified policies assigned and allows for performing operations in accordance with those policies. bao token lookup should successfully display the details of this token. The GUI should be fully functional.

**Environment:**
* OpenBao Server Version: 2.4.0
* OpenBao CLI Version: v2.4.0 (f9407a27fbee45f62ba521bba2bf1e2360c61031), built 2025-08-28T20:33:05Z
* Server Operating System/Architecture: Windows Server 2022 Standard (21H2)

OpenBao server configuration file(s):

```hcl
cluster_addr  = "https://[redacted].eu"
api_addr       = "https://[redacted].eu"
disable_mlock = true
ui                    = true

storage "file" {  
  path = "vault-data"  
} 

listener "tcp" {
  address            = "[redacted]"
  tls_disable        = 0
  tls_cert_file      = "D:\\Vault\\certs\\[redacted].eu-crt.pem"
  tls_key_file       = "D:\\Vault\\certs\\[redacted].eu-key.pem"
  tls_client_ca_file = "D:\\Vault\\certs\\[redacted].eu-ca.pem"
}

[...]

PR #1753, opened 2025-09-01
Fix persistence of tokens created through LoginMFA

cipherboy:
In creating inline authentication, I went through renaming the persistToken vs isInlineAuth, failing to correctly update the call in core.LoginMFACreateToken(...) to the latter instead of the negated former. This meant tokens created through the two-step Login MFA flow were not persistent.

This should not occur in the one-step login MFA flow. It was not caught by the existing test in
vault/external_tests/identity/login_mfa_totp_test.go, because this does not use the token after creation.

Resolves: #1752

### DIFF
--- a/changelog/1753.txt
+++ b/changelog/1753.txt
@@ -1,3 +1,6 @@
 ```release-note:bug
 auth/mfa: Correctly persist tokens created through two-step MFA login enforcement.
 ```
+```release-note:bug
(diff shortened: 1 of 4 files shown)


satoqz:
@cipherboy do you want to add a token lookup to the test you've mentioned?

cipherboy:
@satoqz Pushed, and fixed another bug while I was at it.

satoqz (approved):

PR #1754, opened 2025-09-01
Recovery rotation init CLI operation returns the recovery keys

wslabosz-reply:
Resolves #1751

While implementing the "initialization without generation of recovery keys" feature, we've missed to implement the support of returning newly generated keys with the new CLI operation (`operator rotate-keys`). This PR introduces support of that, while also allowing for the `operator init` command to specify the `recovery_shares` and `recovery_threshold` values as 0.

```
// run bao server
bao server -log-level debug -config <config>

// init the server
bao operator init -recovery-shares=0 -recovery-threshold=0

// get the root token and provide it with the command
bao operator rotate-keys -init -target=recovery -key-shares=3 -key-threshold=2
```

### DIFF
--- a/command/operator_init.go
+++ b/command/operator_init.go
@@ -34,10 +34,8 @@ type OperatorInitCommand struct {
 }
 
 const (
-	defKeyShares         = 5
-	defKeyThreshold      = 3
-	defRecoveryShares    = 5
-	defRecoveryThreshold = 3
+	defKeyShares    = 5
+	defKeyThreshold = 3
 )
 
 func (c *OperatorInitCommand) Synopsis() string {
@@ -143,7 +141,7 @@ func (c *OperatorInitCommand) Flags() *FlagSets {
 		Target:     &c.flagRecoveryShares,
 		Completion: complete.PredictAnything,
 		Usage: "Number of key shares to split the recovery key into. " +
-			"This is only used in auto-unseal mode.",
+			"This is only used in Auto Unseal mode.",
 	})
 
 	f.IntVar(&IntVar{
@@ -212,15 +210,7 @@ func (c *OperatorInitCommand) Run(args []string) int {
 	client.SetOutputCurlString(currentOutputCurlString)
 	client.SetOutputPolicy(outputPolicy)
 
-	switch sealInfo.RecoverySeal {
-	case true:
-		if c.flagRecoveryShares == 0 {
-			c.flagRecoveryShares = defRecoveryShares
-		}
-		if c.flagRecoveryThreshold == 0 {
-			c.flagRecoveryThreshold = defRecoveryThreshold
-		}
-	default:
+	if !sealInfo.RecoverySeal {
 		if c.flagKeyShares == 0 {
 			c.flagKeyShares = defKeyShares
 		}
--- a/command/operator_rotate_keys.go
+++ b/command/operator_rotate_keys.go
@@ -291,7 +291,7 @@ func (c *OperatorRotateKeysCommand) init(client *api.Client) int {
 	}
 
 	// Make the request
-	status, err := fn(&api.RotateInitRequest{
+	resp, err := fn(&api.RotateInitRequest{
 		SecretShares:        c.flagKeyShares,
 		SecretThreshold:     c.flagKeyThreshold,
 		PGPKeys:             c.flagPGPKeys,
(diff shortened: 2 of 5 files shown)


website/content/docs/commands/operator/init.mdx
```diff
@@ -86,7 +86,7 @@ flags](/docs/commands) included on all commands.
   generated root token will be encrypted and base64-encoded with the given
   public key.
 
-- `-status` `(bool": false)` - Print the current initialization status. An exit
+- `-status` `(bool: false)` - Print the current initialization status. An exit
   code of 0 means the OpenBao is already initialized. An exit code of 1 means an
   error occurred. An exit code of 2 means the OpenBao is not initialized.
```
satoqz:
```suggestion
- `-status` `(bool: false)` - Print the current initialization status. An exit
  code of 0 means OpenBao is already initialized. An exit code of 1 means an
  error occurred. An exit code of 2 means OpenBao is not initialized.
```

Drive by: I think "the" made sense when it was "the Vault", but "the OpenBao" sounds wrong to me.

command/operator_init.go
```diff
@@ -212,15 +210,7 @@ func (c *OperatorInitCommand) Run(args []string) int {
...
-	switch sealInfo.RecoverySeal {
-	case true:
-		if c.flagRecoveryShares == 0 {
-			c.flagRecoveryShares = defRecoveryShares
-		}
-		if c.flagRecoveryThreshold == 0 {
-			c.flagRecoveryThreshold = defRecoveryThreshold
-		}
-	default:
+	if !sealInfo.RecoverySeal {
```
satoqz, line 219: I think this warrants a comment.

command/operator_rotate_keys.go
```diff
@@ -303,35 +303,76 @@ func (c *OperatorRotateKeysCommand) init(client *api.Client) int {
...
-					"OpenBao's core. If you lose the encrypted keys after they are "+
-					"returned, you will not be able to recover them. Consider canceling "+
-					"this operation and re-running with -backup to allow recovery of the "+
-					"encrypted unseal keys in case of emergency. You can delete the "+
-					"stored keys later using the -delete flag.", strings.ToLower(keyTypeRequired))))
-			c.UI.Output("")
+		if len(c.flagPGPKeys) > 0 && !c.flagBackup {
+			if Format(c.UI) == "table" {
+				c.UI.Warn(wrapAtLength(
+					fmt.Sprintf("WARNING! You are using PGP keys for encrypted the resulting %s "+
```
satoqz:
```suggestion
					fmt.Sprintf("WARNING! You are using PGP keys for encryption of the resulting %s "+
```

command/operator_rotate_keys.go
```diff
@@ -303,35 +303,76 @@ func (c *OperatorRotateKeysCommand) init(client *api.Client) int {
...
+			KeysB64:         resp.KeysB64,
+			Backup:          resp.Backup,
+			PGPFingerprints: resp.PGPFingerprints,
+		})
 	}
 
-	// Provide the current status
-	return c.printStatus(status)
+	// otherwise provide the current status
+	return c.printStatus(resp)
```
satoqz: Logically it'd be nice if this was part of the `if !resp.Complete` branch above now.

command/operator_init.go
```diff
@@ -212,15 +210,9 @@ func (c *OperatorInitCommand) Run(args []string) int {
...
-	switch sealInfo.RecoverySeal {
-	case true:
-		if c.flagRecoveryShares == 0 {
-			c.flagRecoveryShares = defRecoveryShares
-		}
-		if c.flagRecoveryThreshold == 0 {
-			c.flagRecoveryThreshold = defRecoveryThreshold
-		}
-	default:
+	// for barrier seals use the default value of key shares and key threshold,
```
cipherboy, line 217: @wslabosz-reply Isn't this a behavior change with the default invocation? I think we'd want to explicitly allow specifying zero as a value, but continue defaulting to 3/5 if no parameters specified on CLI.

wslabosz-reply: can we differ between explicit zero and no parameter specification?

wslabosz-reply: okay, I've found a way how to do that, fixed.

PR #1756, opened 2025-09-02
reject JSON bodies which consume large amounts of RAM

phil9909:
This adds a new configuration `max_request_json_complexity` to the `listener` stanza with a default value of `10000`.

The configuration limits the amount of tokens a request JSON Body might contain. The amount of tokens in a JSON is used to approximate the memory overhead when unmarshaling a JSON into a map:

| Name                        | Allocation (Bytes)  | Input Size (Bytes)  | Factor | Tokens  |
|-----------------------------|--------------------:|--------------------:|-------:|--------:|
| low nesting, long data      |              43 273 |              10 078 |    4.3 |      24 |
| medium nesting, medium data |              73 452 |              10 050 |    7.3 |   1 250 |
| high nesting, short data    |             155 844 |               9 803 |   15.9 |   2 340 |
| array full of strings       |             189 699 |              10 009 |   19.0 |   3 340 |
| array full of arrays        |             269 694 |              10 009 |   26.9 |   6 673 |
| array full of objects       |             349 688 |              10 009 |   34.9 |   6 673 |

This should not hurt legitimate users. When quickly searing for a complex request JSON in our docs, the [most complex I found](https://openbao.org/api-docs/auth/jwt/#sample-payload-1) had token count 29. You can check the token count of your JSON using this [go playground](https://goplay.tools/snippet/VBNTFNOe9HC):

```go
package main

import (
	"encoding/json"
	"errors"
	"fmt"
	"io"
	"strings"
)

const example = `{
  "your json": "goes here"
}`

func main() {
	var tokenCount int64

	dec := json.NewDecoder(strings.NewReader(example))

	for {
		token, err := dec.Token()
		if err != nil {
			if errors.Is(err, io.EOF) {
				break
			}
			panic(err)
		}
		fmt.Printf("%14T | %v\n", token, token)
		tokenCount++
	}
	fmt.Println("\ntoken count", tokenCount)
}
```

FYI 10000 other than being a round number was an arbitrary choice. Probably we could get away with an even lower default.

I could not measure an impact on performance: https://gist.github.com/phil9909/a3d1e17aaa7db493c2c9f96e1c0d744a

### DIFF
--- a/changelog/1756.txt
+++ b/changelog/1756.txt
@@ -1,3 +1,3 @@
 ```release-note:security
-Put a limit on complexity of JSON in http request body. HCSEC-2025-24 / CVE-2025-6203.
+Limit the complexity of JSON in HTTP request bodies. HCSEC-2025-24 / CVE-2025-6203.
 ```
--- a/website/content/docs/configuration/listener/tcp.mdx
+++ b/website/content/docs/configuration/listener/tcp.mdx
@@ -88,9 +88,9 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
   request duration allowed before OpenBao cancels the request. This overrides
   `default_max_request_duration` for this listener.
 
-- `max_request_json_complexity` `(int: 10000)` – Specifies a hard maximum allowed
-  in JSON tokens. Defaults to 10000 if not set or set to `0`.
-  Specifying a number less than `0` turns off limiting altogether.
+- `max_request_json_complexity` `(int: 10000)` – Specifies the hard maximum
+  of JSON tokens allowed in the request body. Defaults to `10000` if not set or `0`.
+  A value lower than 0 disables this option altogether.
 
 - `proxy_protocol_behavior` `(string: "")` – When specified, enables a PROXY
   protocol version 1 behavior for the listener.
--- a/website/content/docs/internals/limits.mdx
+++ b/website/content/docs/internals/limits.mdx
@@ -248,7 +248,7 @@ The maximum number of tokens in a request body JSON is limited by the
 `max_request_json_complexity` option in the [listener
 stanza](/docs/configuration/listener/tcp). It defaults to 10000. This limit is
 in place to prevent attackers from sending specially crafted JSON bodies that,
-once unmarshaled, consume much more memory than in JSON form (circumventing the
+once unmarshaled, consume significantly more memory than the original request body (circumventing the
 `max_request_size` limit). This limit should not affect valid requests, as even
 large requests use a small number of tokens (a large string still counts as a
 single token).


phyrog (commented):
Since this touches the main request handling, I would like to see some tests for this functionality. Ideally also some benchmark to see what the impact is for small vs large vs too large/complex payloads.

phil9909:
> Since this touches the main request handling, I would like to see some tests for this functionality.

Added some. The code should be called by many other tests transitively, but having a few explicit tests - especially one that hits the limit - won't hurt for sure.

changelog/1756.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:security
+Put a limit on complexity of JSON in http request body. HCSEC-2025-24 / CVE-2025-6203.
```
satoqz:
```suggestion
Limit the complexity of JSON in HTTP request bodies. HCSEC-2025-24 / CVE-2025-6203.
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -88,6 +88,10 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
   request duration allowed before OpenBao cancels the request. This overrides
   `default_max_request_duration` for this listener.
 
+- `max_request_json_complexity` `(int: 10000)` – Specifies a hard maximum allowed
+  in JSON tokens. Defaults to 10000 if not set or set to `0`.
+  Specifying a number less than `0` turns off limiting altogether.
```
```suggestion
- `max_request_json_complexity` `(int: 10000)` – Specifies the hard maximum
  of JSON tokens allowed in the request body. Defaults to `10000` if not set or `0`.
  A value lower than 0 disables this option altogether.
```

website/content/docs/internals/limits.mdx
```diff
@@ -242,6 +242,19 @@ operation on a remote service, the OpenBao client will see a failure.
 The environment variable [`VAULT_CLIENT_TIMEOUT`](/docs/commands#vault_client_timeout) sets a client-side maximum duration as well,
 which is 60 seconds by default.
 
+### Request JSON Complexity
+
+The maximum number of tokens in a request body JSON is limited by the
+`max_request_json_complexity` option in the [listener
+stanza](/docs/configuration/listener/tcp). It defaults to 10000. This limit is
+in place to prevent attackers from sending specially crafted JSON bodies that,
+once unmarshaled, consume much more memory than in JSON form (circumventing the
```
satoqz:
```suggestion
once unmarshaled, consume significantly more memory than the original request body (circumventing the
```

website/content/docs/internals/limits.mdx
```diff
@@ -242,6 +242,19 @@ operation on a remote service, the OpenBao client will see a failure.
...
+`max_request_size` limit). This limit should not affect valid requests, as even
+large requests use a small number of tokens (a large string sill counts as a
```
satoqz:
```suggestion
large requests use a small number of tokens (a large string still counts as a
```

satoqz (approved):

changelog/1756.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:security
+Limit the complexity of JSON in HTTP request bodies. HCSEC-2025-24 / CVE-2025-6203.
```
cipherboy, line 2: We should acquire a new CVE ID for ours as well, do you wish to do so?

internalshared/configutil/listener.go
```diff
@@ -49,14 +49,16 @@ type Listener struct {
...
-	MaxRequestSizeRaw       interface{}   `hcl:"max_request_size"`
-	MaxRequestDuration      time.Duration `hcl:"-"`
-	MaxRequestDurationRaw   interface{}   `hcl:"max_request_duration"`
-	RequireRequestHeader    bool          `hcl:"-"`
-	RequireRequestHeaderRaw interface{}   `hcl:"require_request_header"`
+	Address                     string        `hcl:"address"`
+	ClusterAddress              string        `hcl:"cluster_address"`
+	MaxRequestSize              int64         `hcl:"-"`
+	MaxRequestSizeRaw           interface{}   `hcl:"max_request_size"`
+	MaxRequestJsonComplexity    int64         `hcl:"-"`
```
cipherboy, line 56: I like that this is per-listener fwiw.

http/handler.go
```diff
@@ -730,11 +738,74 @@ func parseQuery(values url.Values) map[string]interface{} {
...
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return nil, err
+		}
+		tokenCount++
+		if tokenCount > maxJsonTokens {
+			return nil, errors.New("failed to parse JSON input: too many tokens")
+		}
+	}
```
cipherboy, line 796: I might try a different approach to this, but if we want to merge this now, I'm fine with it. 

I'm thinking about also adding a `max_string_count` in addition to `max_token_count`, as each string incurs another HMAC invocation, which brings more overhead. Fewer HMACs the better, IMHO.

cipherboy (approved):
I'm not against these changes, but don't yet have time to propose the alternative implementation approach I've been suggested internally. 

It'd be great to have some benchmarking of upper limits on token count vs performance/memory usage &c.

changelog/1756.txt
phil9909: Sure, what is the process for this? I never did this before.

cipherboy:
Based on late review stage of #1784, I'm going ahead and merging this. Thank you @phil9909 for all your work, investigation, and discussions around this issue!!

PR #1757, opened 2025-09-02
Add RFC on OpenTelemetry

phil9909:
[Preview](https://github.com/adfinis-forks/openbao/blob/rfc-opentelemetry/website/content/docs/rfcs/opentelemetry.mdx)

### DIFF
--- a/website/content/docs/rfcs/opentelemetry.mdx
+++ b/website/content/docs/rfcs/opentelemetry.mdx
@@ -1,0 +1,98 @@
+# Switch to OpenTelemetry
+
+## Summary
+
+Should we drop [`go-metrics`][] library and thereby support for the metrics
+protocols of:
+
+- Circonus
+- Datadog
+- Stackdriver
+- Statsd
+- Statssite
+
+in favor of [`opentelemetry-go`][]?
+
+## Problem Statement
+
+Supporting vendor specific protocols only really makes sense, if we actually
+make use of their distinctive features. But as we are supporting them through a
+generic API, we don't do this. The downside would be, that this is breaking
+change.
+
+It is probably save to say that OpenTelemetry won the format war on telemetry
+(at least for the time being, until the next big thing comes). At some point
+most - if not all - metrics consumers/sink will either support it or become
+irrelevant.
+
+We could also use the chance to clean up the existing metrics:
+
+- use labels instead of dynamic names ([as discussed on GitHub][metric names
+  discussion])
+- define an API for plugins to expose metrics, with proper namespacing
+- etc.
+
+We can also mark the new metrics as experimental at first, which would give us
+the flexibility to iterate on them for a few releases while gathering feedback.
+
+## User Migration Path
+
+We should first add the OpenTelemetry based metrics and deprecated the
+[`go-metrics`][] based metrics in one release and in a later release remove the
+old metrics.
+
+### Circonus
+
+[Circonus was acquired by Apica][circonus was acquired] in February 2024.
+
+Citing Apica:
+
+> As a leader in this space, Apica recognizes that OpenTelemetry isn’t just a
+> trend—it’s becoming the foundation of modern observability strategies.
+> <https://www.apica.io/blog/opentelemetry-the-foundation-of-modern-observability-strategy/>
+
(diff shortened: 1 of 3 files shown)


website/content/docs/rfcs/opentelemetry.mdx
```diff
@@ -0,0 +1,98 @@
...
+in favor of [`opentelemetry-go`][]?
+
+## Problem Statement
+
+Supporting vendor specific protocols only really makes sense, if we actually
+make use of their distinctive features. But as we are supporting them through a
+generic API, we don't do this. The downside would be, that this is breaking
+change.
+
+It is probably save to say that OpenTelemetry won the format war on telemetry
```
wslabosz-reply:
```suggestion
It is probably safe to say that OpenTelemetry won the format war on telemetry
```

website/content/docs/rfcs/opentelemetry.mdx
```diff
@@ -0,0 +1,98 @@
...
+- etc.
+
+We can also mark the new metrics as experimental at first, which would give us
+the flexibility to iterate on them for a few releases while gathering feedback.
+
+## User Migration Path
+
+We should first add the OpenTelemetry based metrics and deprecated the
+[`go-metrics`][] based metrics in one release and in a later release remove the
+old metrics.
```
wslabosz-reply:
```suggestion
We should first add the OpenTelemetry based metrics and deprecate the
[`go-metrics`][] in one release and in later releases remove them all together.
```

website/content/docs/rfcs/opentelemetry.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+## Raft metrics
+
+The [raft library][] used by OpenBao uses [`go-metric` directly][go-metric in go
+mod]. To preserve its metrics we can register a custom sink and re-expose them
+via the OpenTelemetry API.
+
+[circonus was acquired]: https://www.apica.io/news/apica-acquires-telemetry-data-management-pioneer-circonus-and-lands-new-funding/
+[datadog-otel]: https://docs.datadoghq.com/opentelemetry/
+[`go-metrics`]: https://github.com/hashicorp/go-metrics
```
phyrog, line 88: Small note: We are currently mostly using the old version of go-metrics from [armon/go-metrics](https://github.com/armon/go-metrics)

We include [hashicorp/go-metrics](https://github.com/hashicorp/go-metrics) only in builtin/logical/openldap (for whatever reason).

wslabosz-reply:
@phil9909 
DCO fails as you applied my suggestion as-is, this uses wrong email, you probably need to manually change that.

website/content/docs/rfcs/opentelemetry.mdx
```diff
@@ -0,0 +1,96 @@
...
+irrelevant.
+
+We could also use the chance to clean up the existing metrics:
+
+- use labels instead of dynamic names ([as discussed on GitHub][metric names
+  discussion])
+- define an API for plugins to expose metrics, with proper namespacing
+- etc.
+
+We can also mark the new metrics as experimental at first, which would give us
```
cipherboy, line 35: I think this is a good idea, it is hard to get feedback on metrics without running them in production.

website/content/docs/rfcs/opentelemetry.mdx
```diff
@@ -0,0 +1,96 @@
+# Switch to OpenTelemetry
+
+## Summary
+
+Should we drop [`go-metrics`][] library and thereby support for the metrics
```
cipherboy, line 5: This is a level of sorcery I've not yet seen :laughing: Didn't realize you could leave the link blank and have it auto-fill to the text value.

cipherboy (approved):
I like this! 

Have we given any thought to metric formats/requirements/... or would that be a follow-up RFC?

phil9909:
> Have we given any thought to metric formats/requirements/... or would that be a follow-up RFC?

Follow up. For now I just wanted to check with the community, if OpenTelemetry would be an welcome change and if the idea of having two metric stacks for some time, so we can experiment more freely, is appreciated.
Now that this is clarified, I will start the real work :laughing:.

PR #1761, opened 2025-09-03
Automatically seal newly created sealable namespaces

wslabosz-reply:
This PR changes the behavior on the creation of a sealable namespace.

Changed:
- namespaces created with a seal config (sealable) are now sealed by default after creation
- moved `StorageAccess` interface to separate file `storage_access`
- tweaked tests related to namespace sealing

### DIFF
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -282,6 +282,32 @@ func (ns *NamespaceStore) invalidate(ctx context.Context, path string) error {
 	return nil
 }
 
+// SetNamespaceSealed is used to create or update a given sealable namespace.
+// Note that you cannot change the seal config of a namespace with this operation.
+func (ns *NamespaceStore) SetNamespaceSealed(ctx context.Context, entry *namespace.Namespace, sealConfig *SealConfig) ([][]byte, error) {
+	unlock, err := ns.lockWithInvalidation(ctx, true)
+	if err != nil {
+		return nil, err
+	}
+	defer unlock()
+
+	new, err := ns.setNamespaceLocked(ctx, entry)
+	if err != nil {
+		return nil, err
+	}
+
+	if new && sealConfig != nil {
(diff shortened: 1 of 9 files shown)


phyrog (approved):

vault/namespace_store.go
```diff
@@ -282,6 +282,32 @@ func (ns *NamespaceStore) invalidate(ctx context.Context, path string) error {
 	return nil
 }
 
+// SetNamespaceSealed is used to create or update a given sealable namespace.
+// Note that you cannot change the seal config of a namespace with this operation.
```
cipherboy, line 286: Is this comment not contradicted by line 301? 

I'm rather confused, the comment makes me believe there's a concrete difference between this and `SetNamespace` but the function name makes me think we're modifying sealed (encrypted, with no decryption keys presently held, barrier sealed that is) namespaces.

wslabosz-reply: @cipherboy
the difference is that you can create a namespace with sealConfig with `SetNamespaceSealed` (note that `Set` doesn't mean a "setter" in this context, like the same applies for the `SetNamespace`, where it is not a "setter" just a CreateOrUpdate, like a "patch"), and you cannot do that with `SetNamespace` as that one won't initialize the barrier (technically they could be collapsed into one), in the end what matters here is that there's no functionality that will let you edit a sealConfig of a namespace by updating namespace. Don't know if we want to implement something like that (with parallel unsealing it would be: add another config -> seal with another config -> remove the first one) but currently we do not support a migration of a seal in context of a namespace.

well technically we don't support namespace sealing at all, as this is just a feature branch.

cipherboy: Ahh I see. I'm with you now. Maybe:

> Note that you cannot change the Seal config of a namespace with this; only add a seal config to a net-new namespace.

Perhaps @wslabosz-reply ?

Issue #1739, opened 2025-08-28
Triage Bitnami container usage


cipherboy:
See: https://github.com/bitnami/containers/issues/83267

> After evaluating the impact and community feedback, the Bitnami team has postponed the deletion of the Bitnami public catalog (docker.io/bitnami) until September 29th to give users more time to adapt to the upcoming changes.
>
> To raise awareness before the registry deletion, we will run a series of brownouts over the coming weeks. During each brownout, a set of 10 container images from docker.io/bitnami will be temporarily unavailable for 24 hours. The scheduled brownouts are:
> 
>     August 28, 08:00 UTC → August 29, 08:00 UTC
>     September 2, 08:00 UTC → September 3, 08:00 UTC
>     September 17, 08:00 UTC → September 18, 08:00 UTC
> 
> The list of affected applications will be published on the day of each brownout via our usual channels.
> 
> As previously announced, since August 28th, we have not published new Bitnami container images or Helm charts to Docker Hub in OCI format. The source code for containers and Helm charts remains available on GitHub under the Apache 2.0 license.

This brownout affected us in the cassandra image, see comment in #1715.

JanMa:
See https://github.com/search?q=org%3Aopenbao%20bitnami&type=code for the affected code

satoqz:
Not resolved, there's still

```
sdk/helper/testhelpers/postgresql/postgresqlhelper.go
80:     return prepareTestContainer(t, name, "docker.mirror.hashicorp.services/bitnami/postgresql-repmgr", version, "secret", false, true, true, env, true)
102:    if repo == "bitnami/postgresql-repmgr" {
```

but this doesn't run in CI, so it's lower priority.

PR #1762, opened 2025-09-03
Remove dependency on bitnami/cassandra container image

satoqz:
This updates `TestSelfSignedCA` to use the official Cassandra image (which is used by all other tests) instead of the one by Bitnami. This test was using the Bitnami image because it can set the required `cassandra.yaml` options via environment variables, for which the official image only provides limited support.

The new `with_tls/cassandra.yaml` file is copied 1:1 from the existing `no_tls/cassandra.yaml` file, with the `truststore`, `keystore` and `client_encryption.enabled` parameters adapted to what was previously achieved via environment variables. The alternative to providing this giant file ourselves would be a bunch of terribly hacky `sed` commands injected into the entrypoint somehow... I think duplicating the config file is still better than that.

I know pretty much _nothing_ about operating Cassandra, but this passed the tests locally... Let's see what CI says :^)

Partially resolves https://github.com/openbao/openbao/issues/1739.

### DIFF
--- a/plugins/database/cassandra/test-fixtures/with_tls/cassandra.yaml
+++ b/plugins/database/cassandra/test-fixtures/with_tls/cassandra.yaml
@@ -1,0 +1,1150 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+# Cassandra storage config YAML
+
+# NOTE:
+#   See http://wiki.apache.org/cassandra/StorageConfiguration for
+#   full explanations of configuration directives
+# /NOTE
+
+# The name of the cluster. This is mainly used to prevent machines in
+# one logical cluster from joining another.
+cluster_name: 'Test Cluster'
+
+# This defines the number of tokens randomly assigned to this node on the ring
+# The more tokens, relative to other nodes, the larger the proportion of data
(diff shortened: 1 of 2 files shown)


plugins/database/cassandra/test-fixtures/with_tls/cassandra.yaml
```diff
@@ -0,0 +1,1150 @@
+# Copyright (c) HashiCorp, Inc.
```
wslabosz-reply, line 1: change to openbao please

satoqz: Honestly it's quite ridiculous for this to have a copyright header in the first place. It's literally the default `cassandra.yaml`, with 3 values or so changed.

wslabosz-reply: or if it's a 1:1 copy, then idk

cipherboy: Yeah, I think this is mostly the same as the original from HashiCorp, so at best we'd add our own header if we wanted to clarify copyright over these three lines. :laughing: 

But in most jurisdictions, I don't know that these three lines are copyrightable in the first place lol.

PR #1763, opened 2025-09-03
Drop "go.uber.org/atomic" package

### LINKED ISSUE #1456: Drop go.uber.org/atomic
`go.uber.org/atomic` existed before `sync/atomic` got nice helpers; we should try to drop this dependency completely and see if we can get strictly to the stdlib again.

wslabosz-reply:
Resolves #1456 

This PR:
- removes all usage of "go.uber.org/atomic"
- adjusts few lint issues
- adds named returns (enforced by `gofumpt`)

### DIFF
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -12,8 +12,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	atomic2 "go.uber.org/atomic"
-
 	"github.com/armon/go-metrics"
 	"github.com/hashicorp/go-multierror"
 	"github.com/openbao/openbao/helper/metricsutil"
@@ -270,9 +268,9 @@ func Backend(conf *logical.BackendConfig) *backend {
 	b.lastTidy = time.Now()
 
 	// Metrics initialization for count of certificates in storage
-	b.certCountEnabled = atomic2.NewBool(false)
-	b.publishCertCountMetrics = atomic2.NewBool(false)
-	b.certsCounted = atomic2.NewBool(false)
+	b.certCountEnabled = &atomic.Bool{}
+	b.publishCertCountMetrics = &atomic.Bool{}
+	b.certsCounted = &atomic.Bool{}
 	b.certCountError = "Initialize Not Yet Run, Cert Counts Unavailable"
 	b.certCount = &atomic.Uint32{}
 	b.revokedCertCount = &atomic.Uint32{}
@@ -296,11 +294,11 @@ type backend struct {
 	tidyStatus     *tidyStatus
 	lastTidy       time.Time
 
-	certCountEnabled                    *atomic2.Bool
-	publishCertCountMetrics             *atomic2.Bool
+	certCountEnabled                    *atomic.Bool
+	publishCertCountMetrics             *atomic.Bool
 	certCount                           *atomic.Uint32
 	revokedCertCount                    *atomic.Uint32
-	certsCounted                        *atomic2.Bool
+	certsCounted                        *atomic.Bool
 	certCountError                      string
 	possibleDoubleCountedSerials        []string
 	possibleDoubleCountedRevokedSerials []string
--- a/command/agentproxyshared/cache/lease_cache.go
+++ b/command/agentproxyshared/cache/lease_cache.go
@@ -16,6 +16,7 @@ import (
 	"net/url"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
@@ -33,7 +34,6 @@ import (
 	"github.com/openbao/openbao/sdk/v2/helper/locksutil"
 	"github.com/openbao/openbao/sdk/v2/logical"
 	gocache "github.com/patrickmn/go-cache"
-	"go.uber.org/atomic"
 )
 
 const (
@@ -122,7 +122,7 @@ type inflightRequest struct {
 func newInflightRequest() *inflightRequest {
 	return &inflightRequest{
 		ch:        make(chan struct{}),
-		remaining: atomic.NewUint64(0),
+		remaining: &atomic.Uint64{},
 	}
 }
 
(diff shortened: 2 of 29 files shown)


vault/barrier_aes_gcm.go
```diff
@@ -1151,7 +1151,7 @@ func (b *AESGCMBarrier) ConsumeEncryptionCount(consumer func(int64) error) error
 		err := consumer(c)
 		if err == nil && c > 0 {
 			// Consumer succeeded, remove those from local encryptions
-			b.UnaccountedEncryptions.Sub(c)
+			b.UnaccountedEncryptions.Store(c - 1)
```
satoqz: I don't think this is equal?

phyrog:
```suggestion
			b.UnaccountedEncryptions.Add(-c)
```

vault/barrier_aes_gcm.go
```diff
@@ -1238,18 +1238,19 @@ func (b *AESGCMBarrier) persistEncryptions(ctx context.Context) error {
...
+			// we need another autoRotateCheckInterval later.
 			newEncs := upe + 1
 			activeKey.Encryptions += uint64(newEncs)
 			newKeyring := b.keyring.Clone()
 			err := b.persistKeyringBestEffort(ctx, newKeyring)
 			if err != nil {
 				return err
 			}
-			b.UnaccountedEncryptions.Sub(newEncs)
+			b.UnaccountedEncryptions.Store(upe - newEncs)
```
phyrog: I don't think this would be atomic, `UnaccountedEncryptions` may have changed since loading it into `upe`.


```suggestion
			b.UnaccountedEncryptions.Add(-newEncs)
```

command/agentproxyshared/cache/lease_cache.go
```diff
@@ -250,8 +250,8 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
 	if found {
 		idLock.Unlock()
 		inflight = inflightRaw.(*inflightRequest)
-		inflight.remaining.Inc()
-		defer inflight.remaining.Dec()
+		newVal := inflight.remaining.Add(1)
+		defer inflight.remaining.Store(newVal - 1)
```
phyrog:
```suggestion
		defer inflight.remaining.Add(math.MaxUint64) // remaining is Uint64, so we can't add -1
```

command/agentproxyshared/cache/lease_cache.go
```diff
@@ -262,8 +262,8 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
 		}
 	} else {
 		inflight = newInflightRequest()
-		inflight.remaining.Inc()
-		defer inflight.remaining.Dec()
+		newVal := inflight.remaining.Add(1)
+		defer inflight.remaining.Store(newVal - 1)
```
phyrog:
```suggestion
		defer inflight.remaining.Add(math.MaxUint64) // remaining is Uint64, so we can't add -1
```

vault/barrier_aes_gcm.go
wslabosz-reply: probably saw that and thought it was `.Desc()`

command/agentproxyshared/cache/lease_cache.go
wslabosz-reply: I see, makes sense, but now understanding this I think I would write it as `^uint64(0)` just to not import `math`

command/agentproxyshared/cache/lease_cache.go
```diff
@@ -250,8 +250,8 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
 	if found {
 		idLock.Unlock()
 		inflight = inflightRaw.(*inflightRequest)
-		inflight.remaining.Inc()
-		defer inflight.remaining.Dec()
+		inflight.remaining.Add(1)
+		defer inflight.remaining.Store(^uint64(0)) // equivalent of decrementing by 1
```
phyrog:
```suggestion
		defer inflight.remaining.Add(^uint64(0)) // equivalent of decrementing by 1
```

command/agentproxyshared/cache/lease_cache.go
```diff
@@ -262,8 +262,8 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
 		}
 	} else {
 		inflight = newInflightRequest()
-		inflight.remaining.Inc()
-		defer inflight.remaining.Dec()
+		inflight.remaining.Add(1)
+		defer inflight.remaining.Store(^uint64(0)) // equivalent of decrementing by 1
```
phyrog:
```suggestion
		defer inflight.remaining.Add(^uint64(0)) // equivalent of decrementing by 1
```

satoqz:
I think in a lot of cases you don't need a `*atomic.X` and an `atomic.X` would suffice (because the struct holding the atomic is already referenced by pointer). But we can do that in another PR.

vault/core.go
```diff
@@ -3558,7 +3557,8 @@ func (c *Core) FinalizeInFlightReqData(reqID string, statusCode int) {
 	}
 
 	c.inFlightReqData.InFlightReqMap.Delete(reqID)
-	c.inFlightReqData.InFlightReqCount.Dec()
+	inFlightReqCount := c.inFlightReqData.InFlightReqCount.Load()
+	c.inFlightReqData.InFlightReqCount.Store(inFlightReqCount - 1)
```
satoqz: Should be a single call

satoqz (approved):

builtin/logical/pki/backend.go
```diff
@@ -296,11 +294,11 @@ type backend struct {
 	tidyStatus     *tidyStatus
 	lastTidy       time.Time
 
-	certCountEnabled                    *atomic2.Bool
-	publishCertCountMetrics             *atomic2.Bool
+	certCountEnabled                    *atomic.Bool
```
cipherboy, line 297: Yeah I concur with @satoqz, next revision should probably drop the pointers. But it is fine for now.

builtin/logical/pki/backend.go
```diff
@@ -12,8 +12,6 @@ import (
 	"sync/atomic"
```
cipherboy, line 12: Find it absolutely hilarious we imported both in the same package :laughing:

cipherboy:
Thank you all, appreciate the reviews and changes!

wslabosz-reply:
merging
Thanks @phyrog @satoqz @cipherboy

PR #1764, opened 2025-09-03
Add {LICENSE|SPDX|VERSION}_NODE_MODULES.txt to the release dist tarball

DrDaveD:
This collects licensing and version information for the node modules to put into the release distribution tarball, as required by the Fedora/EPEL package reviewers.


DrDaveD:
> If you want, we could also export the UI as a separate build artifact as well.

It is already included in the dist tarball.

PR #1772, opened 2025-09-04
Remove unnecessary `*_util.go` files

wslabosz-reply:
This PR:
- removes `*_util.go` files where appropriate, like: code is referenced just by the "main" file, code is a stub, code is unused, code is few lines long and doesn't make sense to encapsulate to another file.
- moves used code from `*_util.go` files to "main" file when required.
- removes `GroupUpdater` and `EntityCreator` interfaces and all related code (dead code)
- renamed `util.go` to `memzero.go`
- moved `namespace_oss.go` code to `namespace_store.go`


vault/memzero.go
```diff
@@ -3,11 +3,6 @@
 
 package vault
 
-import (
-	"crypto/rand"
-	"fmt"
-)
-
 // memzero is used to zero out a byte buffer. This specific format is optimized
```
phyrog: Pretty sure we can get rid of this and use the builtin [`clear()`](https://pkg.go.dev/builtin#clear) for this.

wslabosz-reply: 1. could be true as the original code is pre go v1.21 (`clear` introduction)
2. don't know how it impacts the actual security considerations (does the memory get released immediately? is GC prompted to release?)

phyrog: I don't think the GC is involved at all with `clear`, it just zeros the elements of a slice using memclr internally.

wslabosz-reply: @cipherboy 
what do you think? I assume this is crucial security-wise, so I wouldn't do that in the scope of this PR, but if it seems like a good replacement, then I will make another PR with that change

satoqz: I don't think it's as crucial security-wise because this is best-effort after all, and not covered by the threat model.

Based on the documentation for `clear()`, its effect on a `[]byte` should be 1:1 to what `memzero` does.

wslabosz-reply: okay, based on the comments, I will make the change to replace `memzero` with `clear`

cipherboy:
@wslabosz-reply I just worry this will conflict with all three feature branches. Are you OK sitting on this until after they merge to make the merge easier? Otherwise I suppose we could conflict all of our existing work and rebase it.

Regardless, I'd keep namespace stuff unmerged, the files have different copyright headers.

satoqz:
@cipherboy count external keys out, that already needs to be rebased anyways 😄

wslabosz-reply:
> @wslabosz-reply I just worry this will conflict with all three feature branches. Are you OK sitting on this until after they merge to make the merge easier? Otherwise I suppose we could conflict all of our existing work and rebase it.

Yeah, sure.

> Regardless, I'd keep namespace stuff unmerged, the files have different copyright headers.

I'd rather have the functions collapsed into their `namespaceStore` sources, returning back the file doesn't make sens to me, we've (open bao contributors) touched every single line in this file, so I don't know if the copyright header applies to that? (but IANAL)

cipherboy:
@wslabosz-reply Yeah, other than the stupid interface that identity store uses, I think I agree with you that the first two functions (original to HashiCorp) are thin wrappers and could be moved and the latter two, while also thin wrappers, are ours.

I just want to maintain separate files so that it is clear where copyright lies; in existing files, while we modify them, it isn't strictly clear whether or not we've (in the eyes of especially a U.S. court) made substantial contributions that lead to also having copyright (e.g., small linting fixes likely do not qualify). But, in net-new files, we (the contributors) are collectively the copyright owners and so HashiCorp would need to notate that were they to seek to incorporate our contributions into their own distribution.

So when we add large features (namespace store, per-namespace sealing, external keys, transactions, paginated lists, ...), having new files gives us a defensible position IMHO.

My 2c.

wslabosz-reply:
@cipherboy 

- deleted the `NamespaceByPath`
- moved the `NamespaceView` to `namespace_store.go`
- left `NamespaceByID` and `ListNamespaces` in the renamed (namespaces->namespace) _oss.go file.

cipherboy:
Thanks @wslabosz-reply.

PR #1775, opened 2025-09-05
Fix remaining ineffassign linter errors

phil9909:


builtin/credential/jwt/path_cel_role_test.go
```diff
@@ -59,11 +56,6 @@ func TestJwt_CelRoleCreate(t *testing.T) {
 				"cel_program": tc.CelProgram,
 			}
 
-			// Add failure_policy only if it's provided in the test case
-			if tc.FailurePolicy != "" {
-				roleData["failure_policy"] = tc.FailurePolicy
-			}
-
```
phil9909, line 66: `FailurePolicy` seems entirely unused. Is this some left-over code that should have been removed @suprjinx?

cipherboy: We decided against failure policy I believe based on memory @suprjinx @phil9909.

builtin/logical/ssh/path_roles.go
```diff
@@ -665,7 +665,6 @@ func (b *backend) checkUpgrade(ctx context.Context, s logical.Storage, n string,
 		}
 
 	SKIPVERSION2:
-		err = nil
```
cipherboy, line 668: Oh I see, this is tricky because we `return nil` below and not `return err`, but there's paths above where `err != nil`, so we clear the assignment. Keeping the clear probably makes some sense, tbh, with a comment we're intentionally clearing the error above, even though it isn't used.

vault/mount.go
```diff
@@ -1140,7 +1140,6 @@ func (c *Core) remountSecretsEngine(ctx context.Context, src, dst namespace.Moun
 	if c.rollback != nil && c.router.MatchingBackend(ctx, srcRelativePath) != nil {
 		if err := c.rollback.Rollback(rCtx, srcRelativePath); err != nil {
 			c.logger.Error("ignoring rollback error during remount", "error", err, "path", src.Namespace.Path+src.MountPath)
-			err = nil
```
cipherboy, line 1137: As above.

cipherboy (approved):
Minor things, really.

builtin/logical/ssh/path_roles.go
phil9909: The variable goes out of scope in the next line, because it is defined within the `if` https://github.com/openbao/openbao/blob/04e6a75db12142d7dcbeb25414ca784c2f89671e/builtin/logical/ssh/path_roles.go#L642

So setting it to `nil` would be purely for documentational propose. Don't get me wrong, this is fine, but we should be clear about this.
If we keep this, I'd prefer to move the `err = nil` assignment closer to the "decision point", i.e. right before the `goto`.

I'm not into the details of this code, but should we also log the error that can happen here (as we do for the other errors that are swallowed): https://github.com/openbao/openbao/blob/04e6a75db12142d7dcbeb25414ca784c2f89671e/builtin/logical/ssh/path_roles.go#L652-L653

WDYT?

cipherboy: @phil9909 Yeah, though I imagine this would be rarely hit, though if it was, would be very verbose (if there's multiple roles). It would also imply signing keys would fail with the configured keys, which means we'll probably get a good error message through other paths.

But yes, I'd probably document why we're clearing it and leave it with a no-lint comment. IIRC certain scanners complained when a err was not explicitly nil'd after ignoring it.

command/agent_test.go
```diff
@@ -863,7 +863,7 @@ auto_auth {
 							continue
 						}
 						if string(c) != templateRendered(i)+suffix {
-							err = fmt.Errorf("expected=%q, got=%q", templateRendered(i)+suffix, string(c))
+							err = fmt.Errorf("expected=%q, got=%q", templateRendered(i)+suffix, string(c)) //nolint:ineffassign // false positive: after a timeout the last error is returned
```
phyrog:
```suggestion
							err = fmt.Errorf("expected=%q, got=%q", templateRendered(i)+suffix, string(c)) //nolint:ineffassign,staticcheck // false positive: after a timeout the last error is returned
```

cipherboy:
@phil9909 Was this good to merge? I think DCO has failed to execute for some reason and I'm unable to trigger it. Mind rebasing this and seeing if that fixes it?

phil9909:
> @phil9909 Was this good to merge? 

From my side, yes.

> I think DCO has failed to execute for some reason and I'm unable to trigger it. Mind rebasing this and seeing if that fixes it?

Indeed this worked :man_shrugging: 

After the rebase we still get:

```bash
$ golangci-lint run --enable-only ineffassign ./...
0 issues.
```

satoqz (approved):
Thanks @phil9909! I'll merge.

PR #1771, opened 2025-09-04
Remove unused code

phyrog:
Resolves #1575 

This deletes all the unused code found by golangci-lint (or in some cases adds `//nolint:unused` with additional explanation why it is unused).

PR #1776, opened 2025-09-05
Fix login renewal with CEL

cipherboy:
OpenBao only allows a single login renewal handler. This works fine for the most part, but we need to begin differentiating native roles from CEL program roles, handling renewal separately.

This adds a new internal data field which we can use to disambiguate the two and wires up the pathLoginRenew handler.

---

Related: https://github.com/openbao/openbao/pull/1771#discussion_r2325072988

### DIFF
--- a/builtin/credential/jwt/path_login.go
+++ b/builtin/credential/jwt/path_login.go
@@ -180,7 +180,8 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d 
 		Alias:        alias,
 		GroupAliases: groupAliases,
 		InternalData: map[string]interface{}{
-			"role": roleName,
+			"role":      roleName,
+			"role_type": "native",
 		},
 		Metadata: tokenMetadata,
 	}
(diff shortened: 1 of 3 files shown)


cipherboy:
FYI @suprjinx and @fatima2003 -- we'll want to keep this in mind for cert auth with cel.

wslabosz-reply (approved):

cipherboy:
@DrDaveD I think this one could use your review. :-)

builtin/credential/jwt/path_login.go
```diff
@@ -204,6 +205,37 @@ func (b *jwtAuthBackend) pathLoginRenew(ctx context.Context, req *logical.Reques
...
 
+	roleType := req.Auth.InternalData["role_type"].(string)
+
+	switch roleType {
+	case "cel":
+		return b.pathCelLoginRenew(ctx, req, data)
+	case "native":
+		return b.pathNativeLoginRenew(ctx, req, data)
+	case "":
+		// Check if this is a native role.
```
DrDaveD: I assume this case is for support of internal data created with older versions of the software.  It would probably be good to note that in a comment.

cipherboy: Ah yes. 

```suggestion
		// Check if this is a native role. This is a fallback for earlier versions which didn't write role type.
```

Issue #1630, opened 2025-08-06
Evaluate https://github.com/arp242/zcache


cipherboy:
https://github.com/arp242/zcache looks to be a maintained fork of https://github.com/patrickmn/go-cache which appears to be unmaintained. We should consider updating and contributing if necessary. This appears to be used in login-mfa, jwt, and other critical places. 

https://github.com/arp242/zcache/issues/6 would be great to have in a fork.

cipherboy:
I'm leaving the `bug` label; based on discussion today, this code path:

https://github.com/openbao/openbao/blob/a768f0b8982ea125dcae43ab353dd1fa6af7da70/vault/login_mfa.go#L2375-L2390

May (with very rare chance) deschedule the immediate `.Set(... 1 ...)` and thus call `Increment( ... )` on other threads first and we'll overwrite those values.

PR #1777, opened 2025-09-05
Replace go-cache with zcache

satoqz:
Resolves https://github.com/openbao/openbao/issues/1630

### DIFF
--- a/vault/external_tests/identity/login_mfa_totp_test.go
+++ b/vault/external_tests/identity/login_mfa_totp_test.go
@@ -129,6 +129,7 @@ func TestLoginMfaGenerateTOTPTestAuditIncluded(t *testing.T) {
 	userClient1, entityID1, _ := testhelpers.CreateEntityAndAlias(t, client, mountAccessor, entity1, testuser1)
 	userClient2, entityID2, _ := testhelpers.CreateEntityAndAlias(t, client, mountAccessor, entity2, testuser2)
 	waitPeriod := 5
+	maxValidationAttempts := 3
 	totpConfig := map[string]interface{}{
 		"issuer":                  "yCorp",
 		"period":                  waitPeriod,
@@ -137,7 +138,7 @@ func TestLoginMfaGenerateTOTPTestAuditIncluded(t *testing.T) {
 		"skew":                    1,
 		"key_size":                10,
 		"qr_size":                 100,
-		"max_validation_attempts": 3,
+		"max_validation_attempts": maxValidationAttempts,
 		"method_name":             "foo",
 	}
 
@@ -354,9 +355,9 @@ func TestLoginMfaGenerateTOTPTestAuditIncluded(t *testing.T) {
 	}
 
 	var maxErr error
-	maxAttempts := 6
-	i := 0
-	for i = 0; i < maxAttempts; i++ {
+	// We should get the max validation attempts error message exactly at
+	// attempt n + 1, not before.
+	for i := range maxValidationAttempts + 1 {
 		_, maxErr = userClient1.Logical().WriteWithContext(context.Background(), "sys/mfa/validate", map[string]interface{}{
 			"mfa_request_id": secret.Auth.MFARequirement.MFARequestID,
 			"mfa_payload": map[string][]string{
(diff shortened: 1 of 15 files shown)


vault/external_tests/identity/login_mfa_totp_test.go
```diff
@@ -366,9 +367,14 @@ func TestLoginMfaGenerateTOTPTestAuditIncluded(t *testing.T) {
...
-	if !strings.Contains(maxErr.Error(), "maximum TOTP validation attempts") {
-		t.Fatalf("unexpected error message when exceeding max failed validation attempts: %s", maxErr.Error())
+
+		rateLimited := strings.Contains(maxErr.Error(), "maximum TOTP validation attempts")
+		switch {
+		case i < maxValidationAttempts && rateLimited:
+			t.Fatalf("got max validation attempts error message too early: %s", maxErr.Error())
+		case i == maxValidationAttempts && !rateLimited:
+			t.Fatalf("unexpected error message when exceeding max failed validation attempts: %s", maxErr.Error())
+		}
```
satoqz, line 377: I've made this test a bit stricter to ensure that I don't introduce an off by one error in TOTP rate limiting compared to existing behavior. I've ensured that this test passes both on main just as this branch.

vault/login_mfa.go
```diff
@@ -2372,21 +2372,20 @@ func (c *Core) validateTOTP(ctx context.Context, mfaFactors *MFAFactor, entityMe
...
-		}
+	// Ensure the rate limit is initialized; ignore error if it already is.
+	_ = c.loginMFABackend.rateLimits.AddWithExpire(rateLimitID, 0, passcodeTTL)
+	// Now increment the attempt counter.
+	numAttempts, ok := c.loginMFABackend.rateLimits.Modify(rateLimitID, func(n uint32) uint32 { return n + 1 })
+
+	if !ok {
+		// TODO: This (unlikely) edge case could be avoided entirely if we had
+		// a ModifySetWithExpire(key, ttl, func) API, such that the above can be
+		// performed in a single step.
```
satoqz, line 2383: I've filed https://github.com/arp242/zcache/issues/10 for this case. Either way, the present workaround should already be sufficient to avoid the race condition described in https://github.com/openbao/openbao/issues/1630.

vault/login_mfa.go
```diff
@@ -2372,21 +2372,20 @@ func (c *Core) validateTOTP(ctx context.Context, mfaFactors *MFAFactor, entityMe
...
+		return errors.New("failed to increment the TOTP code counter")
+	}
+
+	if numAttempts > maximumValidationAttempts {
+		return fmt.Errorf("maximum TOTP validation attempts %d exceeded the allowed attempts %d. Please try again in %v seconds", numAttempts, maximumValidationAttempts, passcodeTTL)
```
satoqz, line 2388: Previously, this error would not reflect the actual number of attempts, as the counter would not be incremented if it already exceeded the maximum. I'm not sure if this was intentional or not - if we have reason to believe it was, I'll add that behavior back.

cipherboy: @satoqz Are there really `n > maximumValidationAttempts` attempts? Can it really be called an attempt if nothing is performed?

(I'm happy either way)

vault/login_mfa.go
cipherboy: Ah I see, I was mistakenly under the impression that we'd have a single TTL but I see I was incorrect.

vault/login_mfa.go
satoqz: Logically, two approaches follow:

1. We count attempts only if they "do something". In this case, our error message would always cap out at `limit + 1`, hence there's no value in including _both_ the limit and the number of attempts that have breached the limit in the message, as one implies the other.

2. We count every processed request as an attempt. Only in this case does it make sense to report the precise number of attempts back to the user. However: Should the user even be able to retrieve this number, from a security perspective? I suppose the advantage would be that a legitimate user can debug why their validation is failing, e.g. because an attacker is simultaneously flooding it with attempts?

Thoughts?

cipherboy: @satoqz I guess the second approach is slightly more clear. I concur with the first point; an attacker only gains the information that someone else attempted in that window, which they could otherwise (sans, number of attempts), by first discovering the upper bound and then sending `bound - 1` requests and periodically sending a new one once the last one would've expired from cache. If they get permission denied, someone else tried auth on the same account in the same window.

satoqz: In that case we can leave the diff where it is right now :)

PR #1674, opened 2025-08-11
Handle read requests locally on standby nodes

fatima2003:
Resolves #

PR #1784, opened 2025-09-08
Limit maximum number of strings, memory in JSON input

cipherboy:
While slightly imprecise due to the simple nature of the parser,
limiting the total number of strings ensures that we have a bounded
number of calls to HMAC within the request body.
    
This could be improve by not counting strings which are keys in maps, at
the cost of added complexity in the tracking state.
    
Similarly, by adjusting our use of the tokenizer, we can estimate total
memory usage in the parsed representation of the request body; we want
to limit this to reasonable levels.
    
This could be improved by adding bucketing overhead for map items,
estimating based on the total number of items in the map.

---

`Co-authored-by: Arran Walker <ajwalker@gitlab.com>`, cc: @saracen

If we like this, I propose merging #1756 first and then landing this on top of it.  Only the top commit is new.

### DIFF
--- a/changelog/1756.txt
+++ b/changelog/1756.txt
@@ -1,3 +1,3 @@
 ```release-note:security
-http: Limit the complexity of JSON in HTTP request bodies through max_request_json_memory and max_request_json_strings. HCSEC-2025-24 / CVE-2025-6203.
+http: Limit the complexity of JSON in HTTP request bodies through max_request_json_memory and max_request_json_strings. HCSEC-2025-24 / CVE-2025-6203 / CVE-2025-59043.
 ```


http/json.go
```diff
@@ -0,0 +1,155 @@
...
+		case safeJSONModeNumber:
+			switch {
+			case (char >= '0' && char <= '9') || char == '.' || char == 'e' || char == 'E' || char == '+' || char == '-':
+				continue
+			default:
+				r.mode = safeJSONModeNone
+			}
+		}
+
+		// Ignore whitespace, commas, colons, and beginning/ending brackets.
```
satoqz:
```suggestion
		// Ignore whitespace, commas, colons, and ending brackets.
```

http/json.go
```diff
@@ -0,0 +1,155 @@
...
+		switch char {
+		case ' ', '\t', ',', ':', ']', '}':
```
satoqz: Should `}` and `]` count as tokens? I believe they would in https://github.com/openbao/openbao/pull/1756.

http/json.go
```diff
@@ -0,0 +1,155 @@
...
+	maxStrings int64
+}
+
+type safeJSONMode uint8
+
+const (
+	safeJSONModeNone   safeJSONMode = iota
+	safeJSONModeString safeJSONMode = iota
+	safeJSONModeNumber safeJSONMode = iota
+	safeJSONModeConst  safeJSONMode = iota
```
satoqz: This is just tangential random curiosity, but I believe I've seen you use this pattern several times now - Is there a reason you prefer this over the shorthand (just one initial iota)?

satoqz (commented):
Ah, very cool. I spent a while thinking about a similar solution that wouldn't require buffering all input, but still using `encoding/json`'s tokenizer. I hadn't considered straight up manually tokenizing though 😄

I think this can be a good approach, but fuzzing would definitely be nice.

http/json_test.go
```diff
@@ -0,0 +1,83 @@
...
+		{"\"he[]llo\"", 1, 1, false, 1, 1},
+		{"\"he]llo\"", 1, 1, false, 1, 1},
+		{"[\"hello\\\"\"]", 2, 1, false, 2, 1},
+		{"[\"hello\\'\"]", 2, 1, false, 2, 1},
+		{`{ "hello" : "world" }`, 3, 2, false, 3, 2},
+		{`{ "hello" : "world" }`, 2, 2, true, -1, -1},   // 2 tokens allowed but have 3
+		{`{ "hello" : "world" }`, 2, 0, true, -1, -1},   // 2 tokens allowed but have 3
+		{`{ "hello" : "world" }`, 2, 100, true, -1, -1}, // 2 tokens allowed but have 3
+		{`{ "hello" : "world" }`, 0, 1, true, -1, -1},   // 1 strings allowed but have 2
+		{`{ "hello" : "world" }`, 100, 1, true, -1, -1}, // 1 strings allowed but have 2
```
phil9909: I would use backticked strings for all test cases

http/json.go
```diff
@@ -0,0 +1,155 @@
...
+func (r *safeJSONReader) Read(input []byte) (int, error) {
+	n, err := r.r.Read(input)
+	if err != nil {
+		return n, err
+	}
+
+	// Short-circuit if bypassing limits.
+	if r.maxTokens < 0 && r.maxStrings < 0 {
+		return n, nil
+	}
```
phil9909: Could be done in `NewSafeJSONReader` instead.

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -88,6 +88,10 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
   request duration allowed before OpenBao cancels the request. This overrides
   `default_max_request_duration` for this listener.
 
+- `max_request_json_complexity` `(int: 10000)` – Specifies the hard maximum
+  of JSON tokens allowed in the request body. Defaults to `10000` if not set or `0`.
+  A value lower than 0 disables this option altogether.
+
```
phil9909: Missing documentation for string count

website/content/docs/internals/limits.mdx
```diff
@@ -242,6 +242,19 @@ operation on a remote service, the OpenBao client will see a failure.
...
+`max_request_json_complexity` option in the [listener
+stanza](/docs/configuration/listener/tcp). It defaults to 10000. This limit is
+in place to prevent attackers from sending specially crafted JSON bodies that,
+once unmarshaled, consume significantly more memory than the original request body (circumventing the
+`max_request_size` limit). This limit should not affect valid requests, as even
+large requests use a small number of tokens (a large string still counts as a
+single token).
+
+The definition for "JSON complexity" might change in the future.
+
```
phil9909: Missing documentation for string count

http/json.go
cipherboy: @satoqz I think in #1756, they would but `:` and `,` would not count. Which, makes me a little confused as the definition of token is fuzzy. Internally we were having a lot of discussions around `encoding/json/v2` and https://pkg.go.dev/encoding/json/jsontext#Token, which also doesn't have a token for `,` (which is probably unnecessary) or `:` (which arguably from a parser definition is a token but isn't fully necessary if you modify your state machine).  

Here, the definition of token is more "object", so maybe we want to reword them? :thinking:

http/json.go
cipherboy: @satoqz :shrug: consistency? :laughing: I think I came too much from a C background where everything would complain and err on uninitialized values, which this looks way too much like for my safety ;-)

http/json.go
satoqz: Given the _actual_ goal we have here it probably does make more sense to operate on an "object" basis rather than a token basis, yeah. I'd suggest rewording to "value" as an object is only `{}` in my head (ignoring JavaScript's opinion on the matter), but won't complain.

phil9909 (commented):
While this is a neat implementation and much shorter than I would have feared (probably because it exploits the fact that in can expect the JSON to be valid), it is still doing what I was trying to avoid: Add a second parser implementation into the mix, this might cause interoperability problems similar to [these](https://bishopfox.com/blog/json-interoperability-vulnerabilities). To me the code seems fine as long as go rejects any invalid json (I think there are more permissive parsers out there, that will try to recover from syntax errors, which could lead to token counting bypasses, but not the go stdlib parser), but I could very well have missed something.

I'm not strictly against this, but I'm unsure, if it is worth the risk (and complexity) 

What would be the advantages of this implementation? "Only" that request bodies don't need to be buffered, right?
Which might still be enough, but maybe I'm missing something even more convincing. 
BTW: we buffer any requests below 4096 bytes anyhow https://github.com/openbao/openbao/blob/788536bd3e10818a7b4fb00aac6affc23388e5a9/http/logical.go#L121-L122

> This parser would likely be a good candidate for fuzzing :-)

Couldn't agree more with this.

Again kudos to this nice implementation, which seems perfectly fine to me.

http/json.go
```diff
@@ -0,0 +1,155 @@
...
+			return 0, ErrJSONExceededTokens
+		}
+
+		switch {
+		case char == '"':
+			r.mode = safeJSONModeString
+			r.strings += 1
+
+			if r.maxStrings >= 0 && r.strings > r.maxStrings {
+				return 0, ErrJSONExceededStrings
```
saracen: In every case an error is returned, `n` should be returned, rather than zero, because at this point, we've read it.

It probably doesn't matter much in this use-case, but a good reader in Go will return what it read:

> When Read encounters an error or end-of-file condition after successfully reading n > 0 bytes, it returns the number of bytes read. 

https://pkg.go.dev/io#Reader

http/json_test.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+func TestSafeJSONReader(t *testing.T) {
+	tests := []struct {
+		input         string
+		maxTokens     int64
+		maxStrings    int64
+		expectedError bool
+		actualTokens  int64
+		actualStrings int64
```
saracen: Maybe `actualTokens` and `actualStrings` should be `expectedTokens` and `expectedStrings`? 🤔

It's what we expect and the values returned from the reader is what I'd call "actual"

http/json_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+		output, err := io.ReadAll(r)
+
+		if test.expectedError {
+			require.Error(t, err, "case: %v / input: %q", index, test.input)
+		} else {
+			require.Equal(t, output, []byte(test.input))
+			require.NoError(t, err, "case: %v / input: %q", index, test.input)
+			require.Equal(t, sjr.tokens, test.actualTokens, "case: %v / input: %q", index, test.input)
+			require.Equal(t, sjr.strings, test.actualStrings, "case: %v / input: %q", index, test.input)
```
saracen: These actual/expected should be swapped. This change would be better if `actualTokens`/`actualStrings` was renamed with `expectedTokens`/`expectedStrings`.

```suggestion
			require.Equal(t, test.actualTokens, sjr.tokens, "case: %v / input: %q", index, test.input)
			require.Equal(t, test.actualStrings, sjr.strings, "case: %v / input: %q", index, test.input)
```

http/json.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+		case char == '-' || (char >= '0' && char <= '9'):
+			r.mode = safeJSONModeNumber
+		case char == 't' || char == 'f' || char == 'n':
+			r.mode = safeJSONModeConst
+		}
+	}
```
saracen: A slightly more concise, and probably more performant version would be this:

We only need to parse the beginning of any JSON grammar identifier. Only numbers and strings need special handling, to find the end of them. Everything else can safely be ignored (so no need to match on whitespace etc.)

```suggestion
	for idx := range n {
		char := input[idx]

		switch r.mode {
		case safeJSONModeString:
			switch {
			case r.escaped:
				r.escaped = false
			case char == '\\':
				r.escaped = true
			case char == '"':
				r.mode = safeJSONModeNone
			}
			continue
		case safeJSONModeNumber:
			switch char {
			case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '.', 'e', 'E', '+', '-':
				continue
			default:
				r.mode = safeJSONModeNone
			}
		}

		switch char {
		case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '-':
			r.tokens += 1
			r.mode = safeJSONModeNumber

		case '{', '[', 'n', 'f', 't':
			r.tokens += 1

		case '"':
			r.tokens += 1
			r.strings += 1
			r.mode = safeJSONModeString

		default:
			continue
		}

		if r.maxStrings >= 0 && r.strings > r.maxStrings {
			return n, ErrJSONExceededStrings
		}

		if r.maxTokens >= 0 && r.tokens > r.maxTokens {
			return n, ErrJSONExceededTokens
		}
	}
```

cipherboy: @saracen I think we might end up dropping this in favor of the existing `encoding/json/v1` path from @phil9909 in #1756. It turns out the documentation on the method is incorrect and it _doesn't_ validate that `{` and `[` are matched before returning the initial token, so it'd behave similarly to this without needing to ourselves build a parser (though, with a little less control over buffer).

If we really do not want to allocate, that has the benefit of not doing JSON object construction inline with token limits.

http/json.go
```diff
@@ -0,0 +1,163 @@
...
+	}*/
+
+	reader, ok := parent.(io.ReadSeeker)
+	if !ok {
+		body, err := io.ReadAll(parent)
+		reader = bytes.NewReader(body)
+		if err != nil {
+			return reader, -1, -1, fmt.Errorf("failed to read JSON input: %w", err)
+		}
+	}
```
satoqz: IIUC we never actually have a `ReadSeeker` in request handling? So we always buffer right away.
What stops us from using `TeeReader` so we only have to buffer what we actually have to read, like so:

<details>

<summary>Patch</summary>

```diff
diff --git a/http/json.go b/http/json.go
index ba93fa0977..6ecc34cc84 100644
--- a/http/json.go
+++ b/http/json.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 )
 
@@ -100,20 +99,8 @@ func NewSafeJSONReader(ctx context.Context, parent io.Reader) (io.Reader, int64,
 		return parent, -1, -1, nil
 	}*/
 
-	reader, ok := parent.(io.ReadSeeker)
-	if !ok {
-		body, err := io.ReadAll(parent)
-		reader = bytes.NewReader(body)
-		if err != nil {
-			return reader, -1, -1, fmt.Errorf("failed to read JSON input: %w", err)
-		}
-	}
-
-	// Save our current position in the input stream.
-	pos, err := reader.Seek(0, io.SeekCurrent)
-	if err != nil {
-		return reader, -1, -1, fmt.Errorf("failed to read JSON input: %w", err)
-	}
+	buf := &bytes.Buffer{}
+	reader := io.TeeReader(parent, buf)
 
 	var memory int64
 	var strings int64
@@ -127,14 +114,14 @@ func NewSafeJSONReader(ctx context.Context, parent io.Reader) (io.Reader, int64,
 				break
 			}
 
-			return reader, memory, strings, err
+			return buf, memory, strings, err
 		}
 
 		// Limit our estimated cost to parse the input.
 		tokenCost := memoryForToken(token)
 		memory += tokenCost
 		if maxMemory >= 0 && memory > maxMemory {
-			return reader, memory, strings, ErrJSONExceededMemory
+			return buf, memory, strings, ErrJSONExceededMemory
 		}
 
 		// Separately limit the total number of strings to reduce cost
@@ -142,22 +129,15 @@ func NewSafeJSONReader(ctx context.Context, parent io.Reader) (io.Reader, int64,
 		if _, ok := token.(string); ok {
 			strings += 1
 			if maxStrings >= 0 && strings > maxStrings {
-				return reader, memory, strings, ErrJSONExceededStrings
+				return buf, memory, strings, ErrJSONExceededStrings
 			}
 		}
 	}
 
 	// Check for context cancellation.
-	err = ctx.Err()
-	if err != nil {
-		return reader, memory, strings, err
-	}
-
-	// Restore our position in the input stream.
-	_, err = reader.Seek(pos, io.SeekStart)
-	if err != nil {
-		return reader, memory, strings, fmt.Errorf("failed to read JSON input: %w", err)
+	if err := ctx.Err(); err != nil {
+		return buf, memory, strings, err
 	}
 
-	return reader, memory, strings, nil
+	return buf, memory, strings, nil
 }
```

</details>

http/json.go
```diff
@@ -0,0 +1,163 @@
...
+
+func memoryForToken(rawToken json.Token) int64 {
+	switch token := rawToken.(type) {
+	case json.Delim:
+		switch token {
+		case '[':
+			return safeJSONCostBase + safeJSONCostArray
+		case '{':
+			return safeJSONCostBase + safeJSONCostObject
+		}
```
satoqz: Any reason `}` and `]` still incur a `safeJSONCostBase`? I'd assume that their cost is accounted for by the opening pair.

cipherboy: Not really, fixed.

http/json.go
cipherboy: This is probably fine; I wish there was a `Seek(...)`-able variant of this, though, so that we could use it in horizontal scalability as well: https://github.com/openbao/openbao/pull/1674/files#diff-a19758ad27315bebf84e1ad2fc1d00b0dcfb556e576cb33c13b6b6db5a43fa98R735

http/json.go
saracen: 👍 

> without needing to ourselves build a parser

@cipherboy 

I saw there was a discussion of risks. Maybe rather than reader/parser, it'd have made more sense to call this an inspector, to better convey what it was. With minimum state, zero buffering and just counting, then the only risk was incrementing the count too much, or too little. Hopefully both easily testable.

The new solution of teeing into a new buffer and decoding twice feels like an odd choice for something that's protecting against memory use. It might work for this specific use-case, but hopefully nobody uses it elsewhere as a generic solution. I do wonder if this carries more risk :shrug:

cipherboy: Hey @saracen!

> The new solution of teeing into a new buffer and decoding twice feels like an odd choice for something that's protecting against memory use.

There was a discussion about this [on Matrix](https://chat.lfx.linuxfoundation.org/#/room/#openbao-development:chat.lfx.linuxfoundation.org), but despite this being called a "decoder", we aren't going all the way to objects here; it is just another tokenizer since we're only calling `Token(...)`. It is rather similar to yours (with or without my adjustments), except that it'll also validate JSON syntax as the same time.

So we incur a buffer cost (up to `max_request_size`, which is expected anyways as we'll later be forwarding requests via GRPC and thus taking advantage of this buffered copy of the body) but we:

1. Don't incur the full object construction cost. This keeps memory overhead low.
2. We fully validate JSON syntax in this restricted manner, rather than validating + parsing objects at the same time; this means if there is a syntax error, we'll get that issue faster with less memory overhead.
3. Using the `tee` with the decoder means we'll only buffer up to the first memory/string/json syntax/... error. So invalid payloads will not cause _parsing/decoding/unmarshalling_ memory pressure (only _buffering_), but obviously an attacker can toe up to the line and still cause issues.

On the whole, I think the risk is rather well managed.

I just wished object size correlated more to parse size; with a full decoding, I definitely see higher peak memory usage than final object size. We could overestimate by some percentage (25% seems to be experimentally confirmed), though I wish we had concrete data either way.

http/json.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return maxJsonStrings.(int64)
+}
+
+func addMaximumJsonStringsToContext(ctx context.Context, limit int64) context.Context {
+	return context.WithValue(ctx, ctxKeyMaxRequestJsonStringCount{}, limit)
+}
+
+const (
+	safeJSONCostBase   int64 = 8 + 8     // pointer to a value + type of pointier
```
wslabosz-reply:
```suggestion
	safeJSONCostBase   int64 = 8 + 8     // pointer to a value + type of pointer
```

website/content/docs/internals/limits.mdx
```diff
@@ -242,6 +242,29 @@ operation on a remote service, the OpenBao client will see a failure.
...
+### Request JSON Complexity
+
+The estimated maximum number of memory in the parsed representation of the
+JSON request body is limited by the `max_request_json_memory` option in the
+[listener stanza](/docs/configuration/listener/tcp). It defaults to
+32MB + 512KB. This limit is in place to prevent attackers from sending
+specifically crafted JSON bodies that, once unmarshaled, consume significantly
+more memory than the original request body (circumventing the
+`max_request_size` limit). This limit should not affect valid requests, as
+even large requests should a restricted total amount of memory due to using
```
satoqz:
```suggestion
even large requests should use a bounded total amount of memory due to using
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -88,6 +88,15 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
   request duration allowed before OpenBao cancels the request. This overrides
   `default_max_request_duration` for this listener.
 
+- `max_request_json_memory` `(int: 34078720)` – Specifies the maximum estimated
+  memory used while parsing the JSON request body. Defaults to 32MB + 512KB if
```
```suggestion
  memory used for the parsed JSON request body. Defaults to 32MB + 512KB if
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -88,6 +88,15 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
...
+  not set or `0`. A value lower than `0` disables this option altogether.
+
+- `max_request_json_strings` `(int: 1000)` – Specifies the maximum number of
+  strings (keys + values) while parsing the JSON request body. Defaults to
```
```suggestion
  strings (keys + values) allowed in the JSON request body. Defaults to
```

http/json_test.go
```diff
@@ -0,0 +1,256 @@
...
+		output, err := json.Marshal(test.input)
+		require.NoError(t, err)
+
+		buf := bytes.NewBuffer(output)
+
+		ctx := context.Background()
+		ctx = addMaximumJsonMemoryToContext(ctx, -1)
+		ctx = addMaximumJsonStringsToContext(ctx, -1)
+
+		_, memoryEstimate, _, err := NewSafeJSONReader(ctx, buf)
```
satoqz: `memoryEstimate` is always -1 here because `NewSafeJSONReader` chooses the fast path.

http/json_test.go
```diff
@@ -0,0 +1,256 @@
...
+	for range size {
+		list = append(list, map[string]interface{}{})
+	}
+	return list
+}
+
+func makeLongString(size int) interface{} {
+	var x string
+	for i := range size {
+		x += fmt.Sprintf("%d", i)
```
satoqz: I think you meant to only add a single digit here? This makes the test never finish on my end :D

http/json_test.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	return data
+}
+
+func makeRandomMap(size int) interface{} {
+	data := map[string]interface{}{}
+	for i := range size / 10 {
+		switch i % 3 {
+		case 0:
+			data[fmt.Sprintf("%d", i)] = makeLongString(size * 10)
```
satoqz:
```suggestion
			data[fmt.Sprintf("%d", i)] = makeLongString(size)
```

In conjunction with above comment, test takes infinitely long otherwise (I waited 5 minutes, it's 25 seconds now)

http/json_test.go
cipherboy: @satoqz Ah right. Hmmmm 🤔 I might need an alternative "slow" pass.

http/json_test.go
cipherboy:
```suggestion
		x += fmt.Sprintf("%d", i % 10)
```

http/json_test.go
cipherboy: (I had been testing with a modified version of the parser which didn't quick exit.)

Thoughts?

satoqz: I just passed `math.MaxInt64` here to get it to work for testing purposes.

cipherboy: Ah, I like that.

http/json.go
```diff
@@ -0,0 +1,137 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"io"
+)
+
+var (
+	ErrJSONExceededMemory  = errors.New("input JSON exceeded estimated memory limits")
+	ErrJSONExceededStrings = errors.New("input JSON exceeded number of strings")
```
satoqz: I feel like these are missing a word

```suggestion
	ErrJSONExceededMemory  = errors.New("input JSON exceeded maximum estimated memory limits")
	ErrJSONExceededStrings = errors.New("input JSON exceeded maximum number of strings")
```

cipherboy: @satoqz I'm fine with that, though I guess `exceeded` to me means it has to be over the maximum, and can't be under the minimum? :shrug:

changelog/1756.txt
```diff
@@ -1,3 +1,3 @@
 ```release-note:security
-Limit the complexity of JSON in HTTP request bodies. HCSEC-2025-24 / CVE-2025-6203.
+http: Limit the complexity of JSON in HTTP request bodies through max_request_json_memory and max_request_json_strings. HCSEC-2025-24 / CVE-2025-6203.
```
phil9909: I guess we should mention our own CVE number in the changelog.

```suggestion
http: Limit the complexity of JSON in HTTP request bodies through max_request_json_memory and max_request_json_strings. HCSEC-2025-24 / CVE-2025-6203 / CVE-2025-59043.
```

PR #1792, opened 2025-09-10
[Per-NS-Sealing] Seal Manager lifecycle adjustment

wslabosz-reply:
This PR changes:
- with a core sealing (root namespace) all namespaces are being sealed, and all their related entries in `SealManager` get removed (`(*SealManager) RemoveNamespace` and `(*SealManager) Reset` methods)
- moved `unlockInformation` (and related methods) from core to `SealManager` for streamline of the unlock process
- barrier key rotation process streamlined into one logical handler for both root and other namespaces
- multiple tweaks to `SealManager`
- `(*Core) NamespaceView()` usage refactored in places where the returned view wasn't used, instead a path was only needed - added a `namespaceLogicalStoragePath(*namespace)` function

Open question:
- `(*ExpirationManager) Restore()` runs once on the unseal of the core, we probably should implement similar feature for `Unseal()` of other namespaces.

### DIFF
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -415,16 +415,7 @@ func (c *Core) setupExpiration(e ExpireLeaseStrategy) error {
 			c.logger.Error("error shutting down core", "error", err)
 		}
 	}
-
-	// Accumulate existing leases
-	c.logger.Debug("collecting leases")
-	leases, leaseCount, err := expMgr.collectLeases()
-	if err != nil {
-		return err
-	}
-	c.logger.Debug("leases collected", "num_existing", leaseCount)
-
-	go c.expiration.Restore(leases, leaseCount, errorFunc)
+	go c.expiration.Restore(errorFunc)
 
 	quit := c.expiration.quitCh
 	go func() {
@@ -685,7 +676,7 @@ func (m *ExpirationManager) Tidy(ctx context.Context) error {
 
 // Restore is used to recover the lease states when starting.
 // This is used after starting the vault.
-func (m *ExpirationManager) Restore(leases map[*namespace.Namespace][]string, leaseCount int, errorFunc func()) {
+func (m *ExpirationManager) Restore(errorFunc func()) {
 	var retErr error
 	defer func() {
 		// Turn off restore mode. We can do this safely without the lock because
@@ -711,6 +702,16 @@ func (m *ExpirationManager) Restore(leases map[*namespace.Namespace][]string, le
 			}
 		}
 	}()
+
+	// Accumulate existing leases
+	m.logger.Debug("collecting leases")
+	var leases map[*namespace.Namespace][]string
+	var leaseCount int
+	leases, leaseCount, retErr = m.collectLeases()
+	if retErr != nil {
+		return
+	}
+	m.logger.Debug("leases collected", "num_existing", leaseCount)
 
 	// Make the channels used for the worker pool
 	type lease struct {
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -415,7 +415,16 @@ func (c *Core) setupExpiration(e ExpireLeaseStrategy) error {
 			c.logger.Error("error shutting down core", "error", err)
 		}
 	}
-	go c.expiration.Restore(errorFunc)
+
+	// Accumulate existing leases
+	c.logger.Debug("collecting leases")
+	leases, leaseCount, err := expMgr.collectLeases()
+	if err != nil {
+		return err
+	}
+	c.logger.Debug("leases collected", "num_existing", leaseCount)
+
+	go c.expiration.Restore(leases, leaseCount, errorFunc)
 
 	quit := c.expiration.quitCh
 	go func() {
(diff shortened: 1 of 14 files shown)


vault/seal_manager.go
```diff
@@ -59,39 +66,54 @@ func NewSealManager(core *Core, logger hclog.Logger) *SealManager {
...
 	sealLogger := c.baseLogger.Named("seal")
 	c.AddLogger(sealLogger)
 	c.sealManager = NewSealManager(c, sealLogger)
-	c.sealManager.barrierByNamespace.Insert("", c.barrier)
-	c.sealManager.barrierByStoragePath.Insert("", c.barrier)
-	c.sealManager.barrierByStoragePath.Insert(sealConfigPath, nil)
+	c.sealManager.setup()
+}
+
+// setup is used to initialize the internal structsof a sealManager.
```
phyrog:
```suggestion
// setup is used to initialize the internal structs of a sealManager.
```

vault/seal_manager.go
```diff
@@ -59,39 +66,54 @@ func NewSealManager(core *Core, logger hclog.Logger) *SealManager {
...
 }
 
-// teardownSealManager is used to remove seal manager
-// when the vault is being sealed.
-func (c *Core) teardownSealManager() error {
-	// seal all namespaces
-	// TODO: this probably does not work out of the box
-	// c.sealManager.SealNamespace(namespace.RootNamespace)
-	c.sealManager = nil
+// Reset seals all namespaces (beside root) and calls the
```
phyrog:
```suggestion
// Reset seals all namespaces (except root) and calls the
```

"beside" can be confusing here (could also be interpreted as "along with root")

vault/seal_manager.go
```diff
@@ -148,6 +171,21 @@ func (sm *SealManager) SetSeal(ctx context.Context, sealConfig *SealConfig, ns *
...
+// RemoveNamespace removes the given namespace from the SealManager's internal state.
+func (sm *SealManager) RemoveNamespace(ns *namespace.Namespace) {
+	nsSeal := sm.NamespaceSeal(ns.UUID)
+	if nsSeal == nil {
+		return
+	}
+
+	sm.sealsByNamespace[ns.UUID] = nil
+	sm.unlockInformationByNamespace[ns.UUID] = nil
+	sm.rotationConfigByNamespace[ns.UUID] = nil
```
phyrog:
```suggestion
	delete(sm.sealsByNamespace, ns.UUID)
	delete(sm.unlockInformationByNamespace, ns.UUID)
	delete(sm.rotationConfigByNamespace, ns.UUID)
```

vault/seal_manager.go
```diff
@@ -247,8 +291,17 @@ func (sm *SealManager) namespaceSeal(nsUUID string) Seal {
...
 
-// NamespaceUnlockInformation acquires a read lock and returns an unlock
-// information of a namespace with provided uuid.
+// ResetUnsealProcess removes the current unlock parts from memory,
+// to reset the unsealing process of a specified namespace.
+func (sm *SealManager) ResetUnsealProcess(nsUUID string) {
+	sm.lock.Lock()
+	defer sm.lock.Unlock()
+
+	sm.unlockInformationByNamespace[nsUUID]["default"] = nil
```
phyrog:
```suggestion
	delete(sm.unlockInformationByNamespace[nsUUID], "default")
```

Otherwise, the next call to `recordUnsealPart` will go into the `exists` branch and panic on accessing `info.Parts`.

Side question: Do we need to check in `recordUnsealPart` if `sm.unlockInformationByNamespace[ns.UUID]` exists before accessing `["default"]` on it to avoid panics for normal namespaces? Or is that caught earlier in the request?

vault/seal_manager.go
```diff
@@ -447,7 +457,7 @@ func (sm *SealManager) unsealFragment(ctx context.Context, ns *namespace.Namespa
 // recordUnsealPart takes in a key fragment, and returns true if it's a new fragment.
 func (sm *SealManager) recordUnsealPart(ns *namespace.Namespace, key []byte) (bool, error) {
 	info, exists := sm.unlockInformationByNamespace[ns.UUID]["default"]
-	if exists {
+	if exists && info != nil {
```
phyrog: That is unnecessary when using `delete`, as proposed in the other comments.

vault/expiration.go
```diff
@@ -810,8 +812,8 @@ LOOP:
 
 	// Let all go routines finish
 	wg.Wait()
-	if err != nil {
-		return err
+	if retErr != nil {
+		return
```
phyrog, line 817: No need to return the error anymore?

satoqz: Yeah, core just calls

```go
go c.expiration.Restore(errorFunc)
```

and never checks return values. Instead the function is supposed to filter its result (in the defer all the way at the top) and call `errorFunc` only on fatal errors, which will then shut the entire server down.

vault/expiration_test.go
```diff
@@ -446,10 +443,7 @@ func TestExpiration_Tidy(t *testing.T) {
 	testCoreUnsealed(t, testCore)
 
 	exp := testCore.expiration
-
-	if err := exp.Restore(nil); err != nil {
-		t.Fatal(err)
-	}
+	exp.Restore(nil)
```
satoqz: These should probably actually take an `errorFunc` that would fail the test.

vault/seal_manager.go
wslabosz-reply: > Side question: Do we need to check in recordUnsealPart if sm.unlockInformationByNamespace[ns.UUID] exists before accessing ["default"] on it to avoid panics for normal namespaces? Or is that caught earlier in the request?

we initialize that entry with namespace creation (sealable namespace), so I don't think a panic can occur there

vault/expiration_test.go
wslabosz-reply: added
```
	exp.Restore(func() {
		// we do not expect an error here, so fail immediately if it happens
		t.FailNow()
	})
```
to every occurence

satoqz (approved):

PR #1793, opened 2025-09-11
feat: Add support for forwarded Tls-Client-Cert

ochnerd:
Kudos to https://github.com/hashicorp/vault/pull/17272

We are evaluating openbao right now an I found this is a missing feature for us in openbao.

This PR implements the support for forwarded Tls-Client-Certs behind Proxies.

Resolves https://github.com/openbao/openbao/issues/1794

ATTENTION: Please tell me if it is ok to use this Code, because this is part of the vault codebase since 2024 and it don't know if it is allowed because of the BSL license!

### DIFF
--- a/changelog/1793.txt
+++ b/changelog/1793.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+auth/cert: Adds support for TLS certificate authenticaion through a reverse proxy that terminates the SSL connection
+```


satoqz:
Hmm, IANAL, but I guess

https://github.com/JasonN3/hashicorp-vault/tree/3b8c20c3f3b22714d9f71dd38159660feb5a0fba

is still a perfectly MPL-licensed version with this code? Since the PR was created in 2022, only the later commits when main was merged back into the branch once again would be BUSL.

cipherboy:
@ochnerd I have opted not to look at the code, but please **DO NOT** contribute code from HashiCorp. A quick look at our [CONTRIBUTING.md](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md) file could have told you as much:

> **It is the intention of this policy that when developing for feature parity, bug fixes, CVEs, etc. in comparison to Hashicorp Vault that a clean room methodology be used and all deliverables made original.**

It also goes against the DCO, which is required for all PRs: 

> OpenBao adheres to the Linux Foundation's [DCO](https://developercertificate.org/). All contributors are **required to confirm their legal rights to the source they contribute**. This is done through a "DCO Sign Off". Note that this is different from commit signing, such as using PGP or gitsign.

If you have copy/pasted this code from HashiCorp's main Vault repository, without being the original contributor, you are unable to attest to the DCO and we cannot and will not accept your PR.

ochnerd:
@cipherboy , thanks for the clarification. After looking at [CONTRIBUTING.md](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md) and the DCO more thorough I totally understand the decision. 
Because of the DCO and license a came to ask if this would be ok. I totally understand the reason why not! Have a nice day!

cipherboy:
> Hmm, IANAL, but I guess
> 
> https://github.com/JasonN3/hashicorp-vault/tree/3b8c20c3f3b22714d9f71dd38159660feb5a0fba
> 
> is still a perfectly MPL-licensed version with this code? Since the PR was created in 2022, only the later commits when main was merged back into the branch once again would be BUSL.

@satoqz I think the issue is, we'd still need @JasonN3's to open a PR under the DCO. I think the hard part becomes their original version of the patch could theoretically be MPL, but IIUC, it was developed under the CLA, which makes it harder and less clear from a legal perspective.

If it was merged to a MPL-licensed repository, it would be much clearer.

That said, as I just mentioned in #1794, I'm very much in favor of this in general, but I think I recall shortcomings with the original patch that I'm not sure if Steve or Scott addressed before merge.

PR #1795, opened 2025-09-11
Drop github.com/natefinch/atomic

### LINKED ISSUE #1717: Consider dropping github.com/natefinch/atomic
The comment is relevant:

```go
    // We don't care so much about atomic writes here.  We're using this package
    // because we don't have a portable way of verifying that the target file
    // is owned by the correct user.  The simplest way of ensuring that is
    // to simply re-write it, and the simplest way to ensure that we don't
    // damage an existing working file due to error is the write-rename pattern.
    // os.Rename on Windows will return an error if the target already exists.
    return atomic.ReplaceFile(tmpFile, i.tokenPath)
```

We likely could reimplement the capability fairly easily.

satoqz:
Resolves #1717

This code is ultimately quite similar to the original package, relevant files:

- https://github.com/natefinch/atomic/blob/master/file_unix.go
- https://github.com/natefinch/atomic/blob/master/file_windows.go
- https://github.com/natefinch/atomic/blob/master/zfile_windows.go

but slightly modernized. I decided not to use `mksyscall_windows` as the hand-written code is very simple, while the generated output was unnecessarily verbose.

I ran some tests on Windows via GitHub Actions in a temporary repo here:
https://github.com/satoqz/replacefile/actions/runs/17643358072


satoqz:
Actually, I need to investigate more, I've just assumed I need to maintain the same behavior 1:1, but I'm confused about the initial decision to use the atomic package.

The standard library `os.Rename` already sets `MOVEFILE_REPLACE_EXISTING` on windows, the only difference we have here is that we additionally set `MOVEFILE_WRITE_THROUGH` -- but that's for atomicity I would guess, which the original comment says doesn't really matter. Is there something `MOVEFILE_WRITE_THROUGH` does that's specific to permissions?

satoqz:
Ok, `MOVEFILE_REPLACE_EXISTING` has been in std for 10 years. That can't have been the issue? https://github.com/golang/go/commit/92c57363e0b4d193c4324e2af6902fe56b7524a0

satoqz:
Welp, I'm not sure. At least if we keep adding `MOVEFILE_WRITE_THROUGH` we can be 100% sure no breakage is introduced. OTOH I'm inclined to believe there was never a reason to use this package in the first place.

phyrog:
> OTOH I'm inclined to believe there was never a reason to use this package in the first place.

I tend to agree. The reason stated in the original commit (https://github.com/openbao/openbao/commit/124572bf80150689e747dbe84db0b76f4a1a6ad0) is that windows does not overwrite the file if it already exists. Since that does not seem to be the case (`MOVEFILE_REPLACE_EXISTING` being set by `os.Remove`), I don't see why we shouldn't just use `os.Remove`.

Edit: Also see https://github.com/natefinch/atomic/issues/23

satoqz:
Any final opinion here @cipherboy?

I take @JanMa's approval as a vote for the reserved approach, and @phyrog's comment as a vote for dropping everything and using `os.Rename`. Hence we need a tie breaker :^)

cipherboy:
@satoqz Ah, I thought `os.Rename` seemed as good as anything. 

> OTOH I'm inclined to believe there was never a reason to use this package in the first place.

was what I was :+1: ing.

satoqz:
Very well, updated.

wslabosz-reply (approved):

satoqz:
I think we've had enough between-the-lines approval for this approach, I won't beat a dead horse and ask for another round :)

PR #1796, opened 2025-09-11
fix release-dist checkout failure (#1745)

satoqz:
Backport of #1745

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -315,7 +315,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v5
         with:
-          fetch-tags: true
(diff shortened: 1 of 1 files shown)


satoqz:
Adjusted commit author to match Dave's signoff line to fix DCO.

cipherboy:
> Adjusted commit author to match Dave's signoff line to fix DCO.

@satoqz As a general FYI, I can mark explicitly overwrite DCO and set it to pass (tbh, you might be able to as well, not sure what the process is there).

satoqz:
> As a general FYI, I can mark explicitly overwrite DCO and set it to pass

I'm aware, just figured it'd be a bit cleaner to patch it up at the root.

PR #1798, opened 2025-09-11
Recovery rotation init CLI operation returns the recovery keys (#1754)

satoqz:
Backport of #1754


satoqz:
Adjusted commit author to match Wojciech's signoff line to fix DCO.

PR #1824, opened 2025-09-15
add plugin distribution via OCI images

### LINKED ISSUE #1476: Plugin Distribution via OCI Images
### Summary

This RFC proposes adopting Open Container Initiative (OCI) images as the primary
method for distributing OpenBao plugins. Users will define required plugins
within a new `plugins` configuration block, specifying the OCI image URL, the
plugin binary's name within the image, and its SHA256 checksum. On startup,
OpenBao will check for locally cached binaries, verify their integrity, and only
download and extract new or updated plugin binaries from the specified OCI
images if necessary. This approach uses existing container ecosystem tooling and
infrastructure for robust, secure, and efficient plugin management.

### Problem Statement

Currently, OpenBao requires plugins to be manually placed into a designated
directory. This manual process lacks built-in mechanisms for:

  * **Automated Discovery and Installation:** Users must manually acquire plugin
    binaries, often from disparate sources, and correctly place them.
  * **Integrity Verification:** There's no inherent, automated way to verify the
    authenticity and integrity of downloaded plugins beyond manual checksum
    checks.
  * **Version Management:** Managing specific plugin versions across different
    OpenBao instances is a manual and error-prone process.
  * **Supply Chain Security:** Without a standardized, verifiable distribution
    mechanism, ensuring the security of plugin binaries from their source to
    deployment is challenging.

A more robust, automated, and secure distribution mechanism is needed to
streamline plugin management, enhance security, and improve the overall user and
developer experience for OpenBao.

### User-facing description

As an OpenBao operator, you will now manage your desired plugins directly within
OpenBao's configuration file using a new `plugins` block. For each plugin, you
will provide:

  * The **OCI image URL** where the plugin is hosted (e.g., on Docker Hub,
    Artifactory, or your private registry).
  * The **exact name of the plugin binary** inside that image.
  * The **version of the plugin**.
  * The **SHA256 checksum** of the plugin binary to ensure its integrity.

**Example `plugins` configuration block:**

```hcl
plugins {
  "my-secret-plugin" {
    url         = "oci://myregistry.com/openbao/plugins/my-secret-plugin:1.0.0@sha256:abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890"
    binary_name = "my-secret-plugin"
    version     = "1.0.0"
    sha256sum   = "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
  }
  "another-auth-plugin" {
    url         = "oci://anotherregistry.io/openbao/plugins/auth-plugin:2.1.0@sha256:fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210"
    binary_name = "auth-plugin"
    version     = "2.1.0"
    sha256sum   = "fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210"
  }
}
```

On OpenBao startup, the system will automatically check if these plugins are
already available locally with the correct checksum. If not, it will download
the specified OCI image, extract the plugin binary, verify its checksum, and
place it into the designated plugin directory. This ensures your OpenBao
instance always has the correct and verified versions of your required plugins.

### Technical description

The proposed solution involves the following technical components and logic:

1.  **Configuration Parsing:** OpenBao's configuration loader will be extended
    to parse a new top-level `plugins` block. This block will be a map where
    keys are logical plugin names (e.g., `"my-secret-plugin"`) and values are
    objects containing `url` (string), `binary_name` (string), and `sha256sum`
    (string).
2.  **Plugin Directory Management:** OpenBao will use a configurable plugin
    directory (e.g., via `VAULT_PLUGIN_DIR` environment variable or a
    configuration parameter). This directory will be where extracted plugin
    binaries are stored.
3.  **Startup Plugin Loading Logic (Per Plugin):**
      * **Local Cache Check:** For each plugin defined in the configuration,
        OpenBao will first construct the expected path to the plugin binary
        within the plugin directory (e.g.,
        `${VAULT_PLUGIN_DIR}/${binary_name}`). It will attempt to read this
        local file.
      * **Local Checksum Verification:** If the local file exists and is
        readable, OpenBao will compute its SHA256 checksum. This local checksum
        will be compared against the `sha256sum` provided in the `plugins`
        configuration block.
      * **Conditional Download:**
          * If the local file's checksum matches the configured `sha256sum`,
            OpenBao will consider the plugin correctly loaded from the local
            cache and proceed to initialize it. No network operation for this
            plugin will occur.
          * If the local file does not exist, is unreadable, or its checksum
            *does not match* the configured `sha256sum`, OpenBao will initiate
            the download process.
      * **OCI Image Pull:** OpenBao will use an OCI client library (e.g.,
        `go-containerregistry` or similar) to pull the OCI image specified by
        the `url`. This URL is expected to support standard OCI registry
        authentication mechanisms (e.g., Docker Hub, private registries).
      * **Binary Extraction:** Once the OCI image is pulled, OpenBao will
        inspect its contents. It will identify and extract the file specified by
        `binary_name` from the image's layers. The image should ideally be a
        single-file artifact or a simple tarball containing the plugin binary.
      * **Extracted Binary Checksum Verification:** The SHA256 checksum of the
        *extracted* plugin binary will be computed and compared against the
        `sha256sum` from the configuration. This is a crucial security step.
      * **Installation:** If the checksum matches, the extracted binary will be
        moved/copied into the designated plugin directory, overwriting any
        previous version if one existed. It should be made executable.
      * **Error Handling:** Robust error handling will be implemented at each
        step (network errors, OCI image not found, extraction failures, checksum
        mismatches, file system errors). Plugin loading failures will be logged
        with sufficient detail, potentially preventing OpenBao from fully
        starting if a critical plugin fails.

The choice of OCI images for plugin distribution is based on several key
rationales:

  * **Industry Standard:** OCI is the de facto standard for container image
    distribution, benefiting from widespread adoption, mature tooling, and
    established security practices.
  * **Existing Infrastructure:** Leveraging existing OCI registries means
    OpenBao doesn't need to build, host, and maintain a separate, dedicated
    registry service, significantly reducing operational overhead for the
    project.
  * **Security Features:** OCI registries and image formats support features
    like content-addressable storage (via digests), which inherently ensures
    immutability, and often integrate with image signing (e.g., Notary, Cosign)
    for enhanced supply chain security.
  * **Familiarity for Users:** Many OpenBao users are already familiar with OCI
    images and registries from their containerized deployments, reducing the
    learning curve.

**Alternatives:**

  * **Dedicated OpenBao Plugin Registry:** While offering full control, this
    approach introduces significant complexity in terms of development, hosting,
    maintenance, and security (e.g., building a CDN, handling authentication,
    DDoS protection). The benefits are largely outweighed by the overhead
    compared to leveraging OCI.
  * **Direct HTTP/S Downloads:** This would be simpler to implement but lacks
    the inherent integrity verification (beyond a checksum file), rich metadata,
    and established security infrastructure provided by OCI registries. It would
    also require OpenBao to manage more complex download logic.
  * **Package Managers (e.g., APT, YUM, Homebrew):** While good for system-level
    binaries, this would tie OpenBao to specific operating system ecosystems and
    introduce external dependencies beyond the application's control for plugin
    management. It's less portable and flexible for cross-platform plugin
    distribution.

### Downsides

  * **Dependency on OCI Registries:** OpenBao's plugin distribution becomes
    dependent on the availability and performance of external OCI registries.
    Private registries mitigate this but still require external infrastructure.
  * **Initial Download Size:** The initial download of an OCI image might be
    larger than just the plugin binary if the image contains other files or
    layers, though well-crafted plugin images should be minimal.
  * **Client Library Complexity:** Integrating an OCI client library adds a new
    dependency and complexity to the OpenBao codebase.
  * **Authentication Complexity:** While standard, OCI registry authentication
    (e.g., for private registries) might require additional configuration within
    OpenBao, potentially involving credentials management.

### Security Implications

This proposal significantly improves OpenBao's plugin security posture:

  * **Mandatory SHA256 Verification:** By requiring and verifying the SHA256 sum
    of the plugin binary, we ensure its integrity. Any tampering during
    transmission or within the OCI image will be detected, preventing the
    execution of malicious or corrupted plugins.
  * **Immutable Image Digests:** Encouraging the use of OCI image digests in the
    `url` (e.g., `oci://...@sha256:digest`) provides strong content
    addressability. This ensures that the exact, same image bits are always
    retrieved, eliminating mutable tag vulnerabilities.
  * **Leveraging OCI Registry Security:** OCI registries often provide features
    like access control, vulnerability scanning (e.g., Trivy, Clair), and audit
    logs, contributing to a more secure supply chain.
  * **Future Cryptographic Signing:** This design lays the groundwork for future
    integration of more robust cryptographic signature verification of OCI
    images (e.g., using Notary or Cosign). This would allow OpenBao to verify
    that the image was indeed published by a trusted entity.
  * **Reduced Manual Risk:** Automating the plugin acquisition and verification
    process reduces the human element, minimizing the risk of manual errors or
    accidental download of malicious binaries.

### User/Developer Experience

**User Experience:**

  * **Simplified Configuration:** Users define plugins directly in the HCL
    configuration, making plugin dependencies explicit and declarative.
  * **Automated Management:** OpenBao automatically handles downloading,
    verifying, and placing plugins, significantly reducing manual effort.
  * **Reproducibility:** The combination of OCI image URLs (ideally with
    digests) and binary SHA256 sums ensures that OpenBao deployments are highly
    reproducible across different environments.

**Developer Experience:**

  * **Standardized Packaging:** Plugin developers will package their plugins
    into OCI images, a familiar process for many.
  * **Simplified Distribution:** Developers can leverage public or private OCI
    registries for distributing their plugins without needing to set up custom
    download servers.
  * **Tooling Consistency:** Developers can use standard OCI tools (e.g.,
    `oras`, `docker build`) to create and push their plugin images.

### Unresolved questions

  * **Image Structure Requirements:** What are the strict requirements for the
    OCI image structure? Will it be a single tarball with the binary, or do we
    need to define a custom Manifest structure similar to [OpenTofu](https://opentofu.org/docs/main/cli/oci_registries/provider-mirror/#manifest-structure)?
  * **Plugin registration:** Should the downloaded plugins also automatically be
    registered in the plugin catalog?
  * **Multiple Plugin Versions:** How to best handle mulitple versions of the same plugin?

JanMa:
Add a new plugin distribution system enabling declarative plugin management in OpenBao server configuration files. 
This feature allows automatic downloading, caching and validation of plugins distributed as OCI images.

> [!NOTE]
> This feature has been created with the help of GitHub Copilot using Claude Sonnet 4.
> All AI generated code has been thoroughly reviewed as well as edited and refactored where necessary.

## Example configuration

```hcl
plugin_directory = "/tmp/test-plugins"
plugin_download_behavior = "fail"

plugins {
  "secrets-aws" {
    url = "ghcr.io/openbao/openbao-plugin-secrets-aws:v0.0.1"
    binary_name = "openbao-plugin-secrets-aws"
    sha256sum = "c8d23e6d31be2a59d0c269bb7243158c4c61c5073f7ba50ce6f1a0050e023e2d"
  }

  "secrets-nomad" {
    url = "ghcr.io/openbao/openbao-plugin-secrets-nomad:v0.1.3-dev2"
    binary_name = "openbao-plugin-secrets-nomad"
    sha256sum = "b7f2db3cbba3df7b4c2868c7c18e43956862b7cd69f90a96bdbba6a3013e400b"
  }
}
```

Resolves #1476


website/content/docs/plugins/plugin-management.mdx
```diff
@@ -18,6 +18,16 @@ will automatically attempt to determine the `api_addr` as well.
 Detailed information regarding the plugin system can be found in the
 [internals documentation](/docs/plugins).
 
+## Plugin distribution methods
+
+OpenBao supports two methods for distributing external plugins:
+
+1. **Manual management** - Manual download of plugin binaries and registration via CLI/API
+2. **OCI-based distribution** - Automatic download from container registries with declarative configuration and registation via CLI/API
```
wslabosz-reply:
```suggestion
2. **OCI-based distribution** - Automatic download from container registries with declarative configuration and registration via CLI/API
```

command/server/config.go
```diff
@@ -270,6 +300,92 @@ func (b *ServiceRegistration) GoString() string {
...
+			Problem: "plugin_oci_auth: at least one of username, password, or token must be provided",
+		})
+	}
+
+	// If username is provided, password should also be provided
+	if p.Username != "" && p.Password == "" {
+		results = append(results, configutil.ConfigError{
+			Problem: "plugin_oci_auth: password must be provided when username is specified",
+		})
+	}
```
wslabosz-reply: what about the exclude case? If username&password combination is provided, token should not be, and the reverse also applies right?

sdk/helper/pluginutil/oci/downloader.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	// Use username/password auth if available
+	if authConfig.Username != "" && authConfig.Password != "" {
+		logger.Debug("using basic authentication for registry", "registry", registry)
+		return &authn.Basic{
+			Username: authConfig.Username,
+			Password: authConfig.Password,
+		}, nil
+	}
```
wslabosz-reply: if we won't validate the use of both username&password and token as a invalid case, then it should be noted in documentation that `token` takes precedence.

sdk/helper/pluginutil/oci/downloader.go
satoqz, line 1: Is there a reason this is in `/sdk/helper` rather than `/helper`? We can avoid adding the new dependencies to `sdk` otherwise.

website/content/docs/plugins/plugin-management.mdx
```diff
@@ -18,6 +18,16 @@ will automatically attempt to determine the `api_addr` as well.
...
+
+See the [OCI Plugin](/docs/configuration/oci-plugins) configuration reference
+for complete documentation.
```
satoqz: This section does not make it very clear that the remaining content of the file explains method 1, while method 2 requires you to visit the page linked in the paragraph below the enumeration.

website/content/docs/configuration/oci-plugins.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+- `"fail"` `(default)` - Server startup fails if any plugin download fails
+- `"warn"` - Log warnings for failed downloads but continue startup
+
+## `plugin_oci_auth`
+
+Configures authentication for private OCI registries.
+
+```hcl
+plugin_oci_auth {
```
satoqz: I'd just call this `plugin_registries` tbh.

website/content/docs/configuration/oci-plugins.mdx
```diff
@@ -0,0 +1,131 @@
...
+}
+```
+
+### Registry parameters
+
+- `username` `(string: optional)` - Username for basic authentication
+- `password` `(string: optional)` - Password for basic authentication  
+- `token` `(string: optional)` - Bearer token for token-based authentication
+
+Use either `username`/`password` combination OR `token`, not both.
```
satoqz:
```suggestion
Use either a pair of `username`/`password` OR just `token`, not both.
```

command/server/config.go
```diff
@@ -147,6 +153,30 @@ func (c *Config) Validate(sourceFilePath string) []configutil.ConfigError {
...
+		if authConfig != nil {
+			results = append(results, authConfig.Validate(sourceFilePath)...)
+		}
+	}
+
+	// Validate plugin_download_behavior
+	if c.PluginDownloadBehavior != "" {
+		if c.PluginDownloadBehavior != oci.PluginDownloadFailStartup && c.PluginDownloadBehavior != oci.PluginDownloadContinue {
+			results = append(results, configutil.ConfigError{
+				Problem: fmt.Sprintf("plugin_download_behavior must be either '%s' or '%s', got %q", oci.PluginDownloadFailStartup, oci.PluginDownloadContinue, c.PluginDownloadBehavior),
```
satoqz: either
```suggestion
				Problem: fmt.Sprintf("plugin_download_behavior must be either %q or %q, got %q", oci.PluginDownloadFailStartup, oci.PluginDownloadContinue, c.PluginDownloadBehavior),
```
or
```suggestion
				Problem: fmt.Sprintf("plugin_download_behavior must be either '%s' or '%s', got '%s'", oci.PluginDownloadFailStartup, oci.PluginDownloadContinue, c.PluginDownloadBehavior),
```

command/server/config.go
```diff
@@ -1321,3 +1461,38 @@ func parseAuditDevices(name string, list *ast.ObjectList) ([]*AuditDevice, error
...
+		if authConfig != nil {
+			result[registry] = &oci.PluginOCIAuthConfig{
+				Username: authConfig.Username,
+				Password: authConfig.Password,
+				Token:    authConfig.Token,
+			}
+		}
+	}
+	return result
+}
```
satoqz: This seems like unnecessary indirection caused by the new `oci` package being in `sdk`.

website/content/docs/commands/plugin/init.mdx
```diff
@@ -0,0 +1,71 @@
+---
+description: |-
+  The "plugin init" command downloads OCI-based plugins configured in 
+  OpenBao server configuration files.
+---
+
+# plugin init
```
satoqz, line 7: Should it be `install` instead? I think `terraform init` makes sense in comparison because it initializes against a storage backend, but `init` never really intuitively reads out as "download" or "install" to me. Lots of package managers (`go mod`, `cargo`, `npm/pnpm/yarn`, etc.) use the `init` verb to create an initial configuration file, but not for network-related operations.

command/plugin_init.go
```diff
@@ -0,0 +1,225 @@
...
+	return "Initialize and download OCI-based plugins"
+}
+
+func (c *PluginInitCommand) Help() string {
+	helpText := `
+Usage: bao plugin init [options]
+
+  Downloads and initializes OCI-based plugins configured in the OpenBao server
+  configuration. This command reads the plugin configuration from the server
+  config files and downloads the specified OCI images containing plugin binaries.
```
satoqz: Continuing https://github.com/openbao/openbao/pull/1824/files#r2355325643, like, I'm not sure what "initializes" is even supposed to mean here. It just downloads.

command/plugin_init.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	// Check if plugins are configured
+	if len(config.Plugins) == 0 {
+		c.UI.Warn("No OCI plugins configured in the configuration files.")
+		return 0
+	}
+
+	hclog.Default().Info(fmt.Sprintf("Plugin directory: %s", pluginDir))
+	hclog.Default().Info(fmt.Sprintf("Found %d OCI plugin(s) configured", len(config.Plugins)))
```
satoqz:
```suggestion
	hclog.Default().Info(fmt.Sprintf("Found %d OCI plugin(s) in configuration", len(config.Plugins)))
```

It's technically fine, but I had to read it ~4 times to realize that it is :D

command/plugin_init.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// Ensure plugin directory exists
+	if err = os.MkdirAll(pluginDir, 0o755); err != nil {
+		hclog.Default().Error(fmt.Sprintf("Failed to create plugin directory: %v", err))
+		return 1
+	}
+
+	// Initialize plugins using the existing vault reconciliation logic
```
satoqz: The existing what? 😄

sdk/helper/pluginutil/oci/downloader.go
```diff
@@ -0,0 +1,380 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/hashicorp/go-hclog"
+)
+
+// Plugin download error behavior constants
+const (
+	PluginDownloadFailStartup = "fail"
+	PluginDownloadContinue    = "continue"
```
satoqz:
```suggestion
	PluginDownloadFail     = "fail"
	PluginDownloadContinue = "continue"
```

website/content/docs/commands/plugin/index.mdx
```diff
@@ -12,6 +12,15 @@ the plugin catalog
 
 ## Examples
 
+Download OCI-based plugins configured in server configuration:
+
+```shell-session
+$ bao plugin init -config=/etc/openbao/openbao.hcl
+Plugin directory: /opt/openbao/plugins
+Found 2 OCI plugin(s) configured
+[INFO] plugins downloaded successfully
```
satoqz: I can't find the last log line in the code, I think you've changed them?

website/content/docs/commands/plugin/init.mdx
```diff
@@ -0,0 +1,71 @@
...
+- `-config` `(string: required)` - Path to configuration file or directory. Can be specified multiple times.
+
+- `-directory` `(string: optional)` - Plugin directory override. Uses `plugin_directory` from config if not specified.
+
+- `-timeout` `(duration: "5m")` - Timeout for downloading all plugins.
+
+## Configuration requirements
+
+The command requires:
+- At least one plugin defined in a `plugins` block
```
satoqz, line 48: This implies that the command would fail if the plugin list is empty, which is not the case.

command/plugin_init.go
```diff
@@ -0,0 +1,225 @@
...
+		return 1
+	}
+
+	return c.runPluginInit()
+}
+
+func (c *PluginInitCommand) runPluginInit() int {
+	// Require config flags to be specified
+	if len(c.flagConfigs) == 0 {
+		c.UI.Error("No configuration specified. Use -config flag to specify configuration files or directories.")
```
satoqz, line 126: This currently doesn't match the error message given by the equivalent code path in `bao server`, but I think this one is better than the existing one. Might make sense to sync them?

website/content/docs/commands/plugin/init.mdx
```diff
@@ -0,0 +1,71 @@
...
+description: |-
+  The "plugin init" command downloads OCI-based plugins configured in 
+  OpenBao server configuration files.
+---
+
+# plugin init
+
+The `plugin init` command downloads OCI-based plugins configured in OpenBao server configuration files. It reads plugin definitions from the configuration and downloads the specified OCI images containing plugin binaries.
+
+Downloaded plugins are placed in the plugin directory but are **not** automatically registered in the plugin catalog. Use [`plugin register`](/docs/commands/plugin/register) to register downloaded plugins.
```
satoqz: This should link to `oci-plugins.mdx` somewhere.

sdk/helper/pluginutil/oci/downloader.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		closeErr := file.Close()
+		// in case closing the file returns an error
+		// make calculateSHA256 exit with that error
+		if err == nil && closeErr != nil {
+			err = closeErr
+		}
```
satoqz:
```suggestion
		err = errors.Join(err, file.Close())
```

sdk/helper/pluginutil/oci/downloader.go
```diff
@@ -0,0 +1,380 @@
...
+			// Copy the binary data
+			_, err = io.Copy(outFile, tarReader)
+			outFile.Close() //nolint:errcheck
+			if err != nil {
+				removeErr := os.Remove(targetPath)
+				if removeErr != nil {
+					return errors.Join(fmt.Errorf("failed to extract binary data: %w", err), fmt.Errorf("failed to remove %s: %w", targetPath, err))
+				}
+				return fmt.Errorf("failed to extract binary data: %w", err)
+			}
```
satoqz:
```suggestion
			if _, copyErr := io.Copy(outFile, tarReader); copyErr != nil {
				err = fmt.Errorf("failed to extract binary data: %w", copyErr)
			}

			if closeErr := outFile.Close(); closeErr != nil {
				err = errors.Join(err, fmt.Errorf("failed to close %s: %w", targetPath, closeErr))
			}

			if err != nil {
				if removeErr := os.Remove(targetPath); removeErr != nil {
					err = errors.Join(err, fmt.Errorf("failed to remove %s: %w", targetPath, removeErr))
				}
				return err
			}
```

sdk/helper/pluginutil/oci/downloader.go
```diff
@@ -0,0 +1,380 @@
...
+
+// IsPluginCacheValid checks if the plugin already exists in the plugin directory
+// and matches the expected SHA256 hash (fast path)
+func (d *PluginDownloader) IsPluginCacheValid(pluginName string, config *PluginConfig) bool {
+	if d.pluginDirectory == "" {
+		return false
+	}
+
+	// Check if the symlink exists in the plugin directory
+	symlinkPath := filepath.Join(d.pluginDirectory, pluginName)
```
satoqz: Should this include the version as Alex suggested in https://github.com/openbao/openbao/issues/1476#issuecomment-3229250024?

website/content/docs/configuration/oci-plugins.mdx
satoqz, line 1: Generally it would be nice to list some use cases / scenarios for the feature here. Personally, while I understand parts of the goal, am still lacking understanding on what the intended practical workflow (in a production deployment scenario) is.

website/content/docs/configuration/oci-plugins.mdx
```diff
@@ -0,0 +1,131 @@
...
+  "ghcr.io" {
+    username = "${GITHUB_USERNAME}"
+    password = "${GITHUB_TOKEN}"
+  }
+}
+```
+
+## Environment variables
+
+Configuration supports environment variable interpolation using `${VARIABLE_NAME}` syntax.
```
satoqz: Is this actually implemented? Maybe I missed it in the code. Either way, I'm pretty sure this would add a third format of parsing environment variables to the config file, which is not great.

We already have:
- Profile env/file sources system (Can be used here, maybe)
- `parseutil.ParsePath` (used for seal configs via go-kms-wrapping, see https://github.com/hashicorp/go-secure-stdlib/blob/594f3ad4e7ec0d6dfa4fd86481001ed6cab284ff/parseutil/parsepath.go#L46)

satoqz (commented):
A handful of thoughts!

command/server/config.go
JanMa: You're right, this is missing a case :+1:

sdk/helper/pluginutil/oci/downloader.go
JanMa: No, to be honest I wasn't sure where to put it. Since there was already a `pluginutil` in the `sdk` I put the code there. If you have a strong preference to move it to `/helper` I'd be open for that

website/content/docs/configuration/oci-plugins.mdx
JanMa: Yeah, I was unsure about the naming of this parameter, it's structure and wehter to add it in the first place. The alternative I thought of was to just use the standard Docker `config.json` as described here:
https://github.com/google/go-containerregistry/tree/v0.20.6/pkg/authn#tldr-for-consumers-of-this-package
This is also more or less what Kubernetes uses:
https://github.com/google/go-containerregistry/tree/v0.20.6/pkg/authn#tldr-for-consumers-of-this-package

website/content/docs/configuration/oci-plugins.mdx
JanMa: No you're right this is nonsense. I'll remove it.

sdk/helper/pluginutil/oci/downloader.go
JanMa: I chose to use the map key as the symlink target, so this will allow you to specify versions like so:
```hcl
plugins {
  "secrets-aws-0.0.1" {
    url = "ghcr.io/openbao/openbao-plugin-secrets-aws:v0.0.1"
    binary_name = "openbao-plugin-secrets-aws"
    sha256sum = "<sum>"
  }
  
  "secrets-aws-0.1.0" {
    url = "ghcr.io/openbao/openbao-plugin-secrets-aws:v0.1.0"
    binary_name = "openbao-plugin-secrets-aws"
    sha256sum = "<sum>"
  }
}
```

Otherwise you'd have to also specify the version twice, once in the URL and then in an additional version field :thinking:

website/content/docs/commands/plugin/init.mdx
JanMa: I have no strong preference, `install` would also be fine I guess. I choose `init` because of the comparison with Terraform/OpenTofu where it downloads and extracts the providers if you run the `init` command.

command/plugin_init.go
JanMa: > Downloads and initializes

It means downloads the OCI image, extracts the plugin from the OCI image into a cache and creates a symlink to the plugin directory. 

> Downloads and extracts OCI-based plugins configured in the OpenBao server configuration.

Would also work I think

command/plugin_init.go
JanMa: :speak_no_evil:

website/content/docs/configuration/oci-plugins.mdx
JanMa: Sure, let me add some more context :+1:

command/server/config.go
JanMa: Yes, otherwise `sdk` would need to depend on the internal `server` package. See my comment above, I'd be open to move the code somewhere else and get rid of this interface

website/content/docs/commands/plugin/index.mdx
JanMa: Good catch, I removed it because this was too verbose

website/content/docs/commands/plugin/init.mdx
JanMa: You're right, I updated the command so it fails

website/content/docs/configuration/oci-plugins.mdx
satoqz: I mean it would be good to have the ability to inject secrets here.

sdk/helper/pluginutil/oci/downloader.go
satoqz: I was thinking you could automatically extract the "version" from the tag (as in, just use the literal tag, without the potential `@sha256...` suffix). But leaving it up to the user entirely seems fine, too.

sdk/helper/pluginutil/oci/downloader.go
satoqz: Not really based on preferences, my point is that we'd add dependencies to the sdk module and force consumers to depend on those for no reason, as the main module is the only one using the package for the foreseeable future.

satoqz: I suppose `pluginutil` is for _both_ the main module and plugin authors, but `pluginutil/oci` isn't.

website/content/docs/commands/plugin/init.mdx
satoqz: Yeah, I understand the inspiration but I'm really not sure since we already use the verb in `bao operator init` which does _actual one-time initialization_. IMO this was always an odd choice in Terraform/OpenTofu that seemed to stem less from the provider installation part but the state initialization instead.

sdk/helper/pluginutil/oci/downloader.go
JanMa: I'm convinced, I'll move the package :-)

website/content/docs/configuration/oci-plugins.mdx
JanMa: After thinking it over for some days I decided to make use of the builtin auth support of `go-containerregistry`.

satoqz: @JanMa yeah, the diff speaks for itself :D

command/server/config.go
```diff
@@ -266,6 +290,63 @@ func (b *ServiceRegistration) GoString() string {
...
+					Problem: fmt.Sprintf("plugin %q: sha256sum contains invalid character '%c' at position %d, must be hexadecimal", pluginName, c, i),
+				})
+				break
+			}
+		}
+	}
+
+	return results
+}
+
```
JanMa, line 394: AI Generated code

command/server/config.go
```diff
@@ -483,6 +564,21 @@ func (c *Config) Merge(c2 *Config) *Config {
...
+	for name, plugin := range c2.Plugins {
+		result.Plugins[name] = plugin // c2 takes precedence
+	}
+
+	// Merge plugin download behavior
+	result.PluginDownloadBehavior = c.PluginDownloadBehavior
+	if c2.PluginDownloadBehavior != "" {
+		result.PluginDownloadBehavior = c2.PluginDownloadBehavior
+	}
+
```
JanMa: AI generated code

command/plugin_init.go
```diff
@@ -0,0 +1,225 @@
...
+		return 1
+	}
+
+	if len(f.Args()) > 0 {
+		c.UI.Error(fmt.Sprintf("Too many arguments (expected 0, got %d)", len(f.Args())))
+		return 1
+	}
+
+	return c.runPluginInit()
+}
```
JanMa, line 121: AI generated code

command/plugin_init.go
```diff
@@ -0,0 +1,225 @@
...
+			c.UI.Warn(fmt.Sprintf("WARNING: ignoring duplicate configuration found in directory: %v", path))
+		}
+	}
+
+	if config == nil {
+		return nil, configErrors, fmt.Errorf("no valid configuration found")
+	}
+
+	return config, configErrors, nil
+}
```
JanMa: AI generated code

helper/pluginutil/oci/downloader.go
```diff
@@ -0,0 +1,321 @@
...
+
+// shouldFailOnPluginError determines whether plugin download errors should fail startup
+func (d *PluginDownloader) shouldFailOnPluginError() bool {
+	behavior := d.config.PluginDownloadBehavior
+	if behavior == "" {
+		behavior = server.PluginDownloadFail
+	}
+
+	return behavior == server.PluginDownloadFail
+}
```
JanMa, line 102: AI generated code

helper/pluginutil/oci/downloader.go
```diff
@@ -0,0 +1,321 @@
...
+
+	hasher := sha256.New()
+	_, err = io.Copy(hasher, file)
+	if err != nil {
+		return "", err
+	}
+
+	result = hex.EncodeToString(hasher.Sum(nil))
+	return result, err
+}
```
JanMa: AI generated code

vault/plugin_catalog_oci_test.go
```diff
@@ -0,0 +1,299 @@
...
+	layer := static.NewLayer(buf.Bytes(), "application/vnd.docker.image.rootfs.diff.tar")
+
+	// Start with an empty image and add our layer
+	img, err := mutate.AppendLayers(empty.Image, layer)
+	if err != nil {
+		t.Fatalf("failed to create image: %v", err)
+	}
+
+	return img
+}
```
JanMa, line 64: AI generated code

command/plugin_init.go
JanMa: Will do it in a separate PR

command/server/config.go
```diff
@@ -267,6 +291,63 @@ func (b *ServiceRegistration) GoString() string {
...
+		for i, c := range p.SHA256Sum {
+			//nolint:staticcheck // applying De Morgan's law does not make this any easier to understand
+			if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F')) {
+				results = append(results, configutil.ConfigError{
+					Problem: fmt.Sprintf("plugin %q: sha256sum contains invalid character '%c' at position %d, must be hexadecimal", pluginName, c, i),
+				})
+				break
+			}
+		}
+	}
```
cipherboy, line 390: I might suggest you attempt to just hex decode it and augment the err from the hex decoder handle the rest.

command/plugin_init.go
```diff
@@ -0,0 +1,224 @@
...
+	if err = os.MkdirAll(pluginDir, 0o755); err != nil {
+		hclog.Default().Error(fmt.Sprintf("Failed to create plugin directory: %v", err))
+		return 1
+	}
+
+	// Initialize plugins using the existing reconciliation logic
+	return c.reconcilePlugins(config, pluginDir)
+}
+
+func (c *PluginInitCommand) parseConfig() (*server.Config, []error, error) {
```
cipherboy: Can we just refactor this to use the same logic? I think we can pass `*BaseCommand` or `UI` directly as arguments tbh. And then ours can just add the additional empty-config check.

helper/pluginutil/oci/downloader.go
```diff
@@ -0,0 +1,321 @@
...
+	logger.Info("successfully downloaded and validated plugin",
+		"cached_path", cachedPluginPath,
+		"symlink_path", symlinkPath,
+		"hash", actualHash)
+
+	return nil
+}
+
+// calculateSHA256 computes the SHA256 hash of a file
+func (d *PluginDownloader) calculateSHA256(filePath string) (result string, err error) {
```
cipherboy: Wonder if we can refactor this out of other places as well. Thoughts? It is very similar to code in command/server.go doing the same thing.

command/server/config.go
cipherboy: `command/server/config.go`'s AI generation is trivial configuration parsing and representative of code already in the repository under the MPL. 

This file is an existing HashiCorp copyrighted file and so we believe under clause (b) of the DCO this is fine.

command/server/config.go
cipherboy: `command/server/config.go`'s AI generation is trivial configuration parsing and representative of code already in the repository under the MPL. 

This file is an existing HashiCorp copyrighted file and so we believe under clause (b) of the DCO this is fine.

command/plugin_init.go
cipherboy: The initial AI generated block in `command/plugin_init.go` (`Flags()` through `Run()`) is likewise trivial boilerplate and representative of code already in the repository.

@JanMa Can you add a HashiCorp copyright to the header? Because this is a derived work, we need to indicate that it came from HashiCorp under the MPL (like most existing commands are). Then it would be eligible for DCO under clause `(b)`.

command/plugin_init.go
cipherboy: This second AI generated block in `command/plugin_init.go` (`parseConfig()`) is a near copy/paste of the existing `parseConfig()` in command/server.go, with one additional modification (erring on empty configuration -- this makes since  because we expect to read a configuration to parse plugin information and not globally). 

As stated below though: https://github.com/openbao/openbao/pull/1824#discussion_r2523345643

I'd like to see this refactored and ultimately removed.

helper/pluginutil/oci/downloader.go
cipherboy: The first AI generated block in helper/pluginutil/oci/downloader.go (shouldFailOnPluginError()) is very trivial and likely not copyrightable.

helper/pluginutil/oci/downloader.go
cipherboy: This second AI generated block in `helper/pluginutil/oci/downloader.go` (`calculateSHA256()`) is already in the project (`command/server.go`, in `addPlugin()`), it is just extracted and the logic trivially transformed.

@JanMa Like https://github.com/openbao/openbao/pull/1824#discussion_r2523953364, I believe we should add HashiCorp to the copyright header in recognition of the derived nature of this and to satisfy DCO clause `(b)`.

---

If we wanted to, this could probably also be refactored and used through common code, though it is lower on my priority list than `parseConfig()`. This is my comment here as well: https://github.com/openbao/openbao/pull/1824#discussion_r2523372732

If we do that, we wouldn't need to add HashiCorp to the copyright header.

vault/plugin_catalog_oci_test.go
cipherboy: Lastly, the final AI generated block in `vault/plugin_catalog_oci_test.go` is the only immediately non-trivial part:

  - We have various places like `command/debug.go` (`compress()`) and `sdk/helper/docker/testhelpers.go` (`ToTarball()`) which perform nearly the same compression logic, +/- handling of directories.
  - But we don't have anything which works with OCI images (last 2 lines of code + error handling). This is used in test code and given it is only 2 function calls + error handling, is nearly trivial. Writing them by hand will almost certainly produce the same results.

@JanMa Like https://github.com/openbao/openbao/pull/1824#discussion_r2523953364, I believe we should add HashiCorp to the copyright header in recognition of the derived nature of this (tar processing) and to satisfy DCO clause `(b)`.

command/plugin_init.go
JanMa: Added the header

helper/pluginutil/oci/downloader.go
JanMa: Added the header

vault/plugin_catalog_oci_test.go
JanMa: Added the header

command/server/config.go
JanMa: done

changelog/1824.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+Add plugin distribution via OCI images
```
cipherboy:
```suggestion
Add **declarative plugin distribution via OCI images**: using the `plugin` configuration keyword.
```

command/server/config.go
```diff
@@ -74,6 +79,9 @@ type Config struct {
 	PluginFilePermissions    int         `hcl:"-"`
 	PluginFilePermissionsRaw interface{} `hcl:"plugin_file_permissions,alias:PluginFilePermissions"`
 
+	Plugins                map[string]*PluginConfig `hcl:"plugins"`
```
cipherboy: @JanMa What do you think about moving this back to a simple `[]*PluginConfig` like we do for `[]*Initialize` blocks with a `Type` and `Name` field? In particular, the way plugin versioning works, people might want one or more plugin versions around at a time:

```
plugin "auth" "<name>" {
  version = 1.0.0
}

plugin "auth" "<name>" {
  version = 2.0.0
}
```

When you enable a new plugin (`SIGHUP`), you download it and can enable it for new plugins (e.g., testing) while existing plugins continue operating on the existing plugin. Likewise, https://openbao.org/api-docs/system/mounts/#parameters lets you explicitly select an old version.

I think we should mirror this to make partial upgrades possible.

(This also matches what we talked about in Matrix, though I think differs from what is in the proposal -- I think it was an oversight in #1476 that we missed the auth vs secret plugin type keyword as well.)

Thoughts?

command/server/config.go
```diff
@@ -267,6 +292,60 @@ func (b *ServiceRegistration) GoString() string {
 	return fmt.Sprintf("*%#v", *b)
 }
 
+// PluginConfig represents the configuration for a single OCI-based plugin
+type PluginConfig struct {
```
cipherboy, line 296: I would've expected type to be here, even in map format? 🤔

website/content/docs/commands/plugin/init.mdx
cipherboy: @satoqz It is weird, since `bao plugin init` just _fetches_ the plugin and doesn't install it into the plugin catalog (as, barrier would be sealed). 

I think `install` would imply that. Maybe `download`? But I'm happy with `init`. 🚲 🏠

website/content/docs/commands/plugin/init.mdx
```diff
@@ -0,0 +1,69 @@
...
+description: |-
+  The "plugin init" command downloads OCI-based plugins configured in 
+  OpenBao server configuration files.
+---
+
+# plugin init
+
+The `plugin init` command downloads OCI-based plugins configured in OpenBao server configuration files. It reads [plugin definitions](/docs/configuration/oci-plugins) from the configuration and downloads the specified OCI images containing plugin binaries.
+
+Downloaded plugins are placed in the plugin directory but are **not** automatically registered in the plugin catalog. Use [`plugin register`](/docs/commands/plugin/register) to register downloaded plugins.
```
cipherboy: @JanMa is this hallucinated? 😆 I think there's no way to support this:

https://github.com/openbao/openbao/pull/1824/files#r2529014328

it'd be on the other side of the barrier, so you'd have to specify unseal somehow. I think just putting it part of SIGHUP would be better, IMHO.  Mostly `plugin init` means server doesn't have to download them and can just verify + register them, which is fine.

website/content/docs/plugins/plugin-management.mdx
```diff
@@ -18,6 +18,18 @@ will automatically attempt to determine the `api_addr` as well.
...
+
+OpenBao supports two methods for distributing external plugins:
+
+1. **Manual management** - Manual download of plugin binaries
+2. **OCI-based distribution** - Automatic download from container registries
+   with declarative configuration. See the [OCI
+   Plugin](/docs/configuration/oci-plugins) configuration reference for complete
+   documentation.
+
+Both methods require the plugin to be registered in OpenBao before it can be used.
```
cipherboy, line 31: I kinda feel based on discussion in #1476 as well (https://github.com/openbao/openbao/issues/1476#issuecomment-3229250024 + https://github.com/openbao/openbao/issues/1476#issuecomment-2999543561) that we might as well make it auto-register. I don't think it'd be that hard to do, tbh, and declarative audit devices provide an approach for it (two different `types` to denote sources).

cipherboy (commented):
@JanMa Random thoughts!

I'm happy picking up auto-registration if you're in favor but don't have time to do it, but I feel more strongly about config format if we do go down that route (as plugin type will be necessary, though I guess it is kinda somewhat unnecessary now).

helper/pluginutil/oci/downloader.go
JanMa: Moved it into a helper in `osutil`.

command/server/config.go
JanMa: As described in https://github.com/openbao/openbao/pull/1824#discussion_r2356507953 versioning plugins is possible by giving them a different key in the map. At this time we have no need to introduce an argument for the plugin type (secret, auth,..) because it doesn't matter for downloading.

JanMa: By letting users just set whatever as the map key and then using this for the symlink name we give users a lot of flexibility to do whatever they want. But of course this also has drawbacks and a more structured approach like you suggest could make things a bit easier.

I'm open to change it of more people from @openbao/openbao-core-committers and @openbao/openbao-org-maintainers think this might be the better approach :+1:

command/server/config.go
JanMa: Why do we need ot right now? We just download the binary, the plugin type doesn't matter for this

website/content/docs/commands/plugin/init.mdx
JanMa: This is not hallucinated, I wrote this :smile:  I haven't looked if auto-registration is actually possible, I thought it would be good to give a disclaimer anyways

website/content/docs/plugins/plugin-management.mdx
JanMa: I'm happy to investigate this, but I'd prefer to implement it in a subsequent PR.

command/server/config.go
JanMa: @cipherboy I understand now your proposed config schema follows the one we introduced for declarative audit devices. I will adapt it also for plugins so we use the same schema :+1:

command/base.go
```diff
@@ -726,3 +728,31 @@ func printFlagDetail(w io.Writer, f *flag.Flag) {
 	indented := wrapAtLengthWithPadding(usage, 6)
 	fmt.Fprintf(w, "%s\n\n", indented)
 }
+
+func (c *BaseCommand) ParseConfig(configFiles []string) (*server.Config, []configutil.ConfigError, error) {
```
cipherboy, line 732: Nit for Alex: I'll rename this to `ParseServerConfig`

command/server/config.go
```diff
@@ -74,6 +79,10 @@ type Config struct {
 	PluginFilePermissions    int         `hcl:"-"`
 	PluginFilePermissionsRaw interface{} `hcl:"plugin_file_permissions,alias:PluginFilePermissions"`
 
+	// Plugins specifies declaratively defined external plugins
+	Plugins                []*PluginConfig `hcl:"-"`
+	PluginDownloadBehavior string          `hcl:"plugin_download_behavior"`
```
cipherboy, line 84: @JanMa What do you think about moving this into `plugins`?

I'm thinking of two additional parameters here:

```
plugin "<type>" "<name>" {
  download_behavior = ... // continue / fail
  auto_download = ... // false --> only bao plugin init will download it.
  auto_register = ... // false --> requires continue/fail on download_behavior. 
}
```

`continue` / `fail` will continue to determine server startup + plugin init behavior. But server downloading can be skipped conditionally and registration can be set independently: if the file exists (and matches the config), we'll register it regardless of whether or not server downloaded it or plugin init downloaded it.

This way we get granular behavior on a per-plugin basis. Critical plugins/updates can be marked with `fail` (security?), `auto_download=true`. While `auto_register=true/false` allows controlling _when_ exactly plugins get registered.

Thoughts?

helper/osutil/fileinfo.go
```diff
@@ -82,3 +85,23 @@ func OwnerPermissionsMatchFile(file *os.File, uid int, permissions int) error {
 
 	return nil
 }
+
+// FileSha256Sum calculates the Sha256 hash of a file
+func FileSha256Sum(path string) (result string, err error) {
```
cipherboy, line 90: Long term, if we intend to use this for plugins, I wonder if we should try to avoid a TOCTOU vulnerability and see if we can do something like return the full plugin contents here, compare hashes, and then write it back out to a temporary location and execute from there (lacking the fancy form of exec from file handle). 

But that's neither here nor there for now. 

Thoughts?

helper/pluginutil/oci/downloader.go
```diff
@@ -0,0 +1,297 @@
...
+	// Download the image
+	img, err := remote.Image(ref, remote.WithContext(ctx), remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	if err != nil {
+		return fmt.Errorf("failed to download OCI image: %w", err)
+	}
+
+	// Extract the plugin binary from the image using hidden cache path
+	// Format: <plugin_directory>/.oci-cache/<plugin_slug>/<sha256_prefix>/<binary_name>
+	sha256Prefix := config.SHA256Sum[:8]
+	cacheDir := filepath.Join(d.pluginDirectory, PluginCacheDir, config.Slug(), sha256Prefix)
```
cipherboy, line 176: @JanMa What do you think about simplifying this? Plugins must be unique across the three-tuple `(type, name, version)`; we could switch this to `slug=<type>-<name>` and name binaries `<binary_name>-<version>` and drop `sha256_prefix` in the middle for a flatter hierarchy.

cipherboy: @JanMa This is actually breaking:

```
[cipherboy@x1c 165]$ bao plugin register -sha256=b7f2db3cbba3df7b4c2868c7c18e43956862b7cd69f90a96bdbba6a3013e400b secret secret-nomad-v0.1.3-dev2 
Error registering plugin secret-nomad-v0.1.3-dev2: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/sys/plugins/catalog/secret/secret-nomad-v0.1.3-dev2
Code: 500. Errors:

* 1 error occurred:
	* cannot execute files outside of configured plugin directory
```

per:

https://github.com/openbao/openbao/blob/36746e0d6385fc159587b35fdbb63b2281698303/vault/plugin_catalog.go#L891-L906

So I think we have one of two options:

1. Flat cache in the plugin directory, perhaps `oci-<type>-<name>-<version>`, or:
2. Allow cache again and make sure the plugin lives in that and the sha256 matches the config value.

Thoughts?

JanMa: :facepalm:  I completely missed that until now. I kind of like the dedicated cache directory, so I'd try to allow executing plugins from there.

command/server/config.go
JanMa: This somehow feels too granular for me. Personally I would prefer to set this behavior globally and not on the individual plugin. :thinking:

helper/osutil/fileinfo.go
JanMa: I think as long as this is only used for validating the downloaded binary without executing it it's fine. You're completely right that when using it to check the SHA256 sum of a binary before executing it, this might cause a TOCTOU vulnerability.

command/server/config.go
cipherboy: @JanMa Perhaps a global default with a per-plugin override? 

I'm thinking the use case is for rolling updates:

1. First upgrade base bao binary. If everything is modular and out-of-tree, this really just updates core.
2. Then update individual plugins, starting on read-enabled standby nodes and then going to the active node.
3. Test the one plugin upgrade.

If we push `auto_register` to the plugin, we can granularly update individual plugins, while downloading all plugins at once up front or in batches (if you keep downloading granular as well). 

Thoughts?

cipherboy: @JanMa I'll start with the coarse behavior and add granular behavior later. But I'll keep the proposed set of options.

vault/plugin_catalog.go
```diff
@@ -901,7 +929,10 @@ func (c *PluginCatalog) setInternal(ctx context.Context, name string, pluginType
 		return nil, fmt.Errorf("error while validating the command path: %w", err)
 	}
 
-	if symAbs != c.directory {
+	// Format: <plugin_directory>/.oci-cache/<plugin_slug>/<sha256_prefix>
+	shaPrefix := hex.EncodeToString(sha256)[:8]
```
cipherboy, line 933: This fails in tests, but I'm planning on a quick follow-up to allow registration, so I'll go ahead and merge this and address in a follow-up PR.

cipherboy: (Remaining tests are flakes; the very rare rabbitmq one and the one talked about in [#openssf-openbao-wg-scalability > Test flakes](https://linuxfoundation.zulipchat.com/#narrow/channel/532998-openssf-openbao-wg-scalability/topic/Test.20flakes/with/558747811)).

PR #1830, opened 2025-09-19
Add Recovery Seal RFC

Huy-Dinh-Reply:
Add RFC for Recovery Seal.
[Preview](https://github.com/Ki-Reply-GmbH/openbao/blob/rfc-recoveryseal/website/content/docs/rfcs/recovery-seal.mdx)


klaus-sap:
I would like to vote for the Alternative (Two Seal Stanzas), i.e. the Recovery Keys should remain as they are today and we should create new key material for the emergency use case.

website/content/docs/rfcs/recovery-seal.mdx
```diff
@@ -0,0 +1,302 @@
...
+  enable_recovery_seal = true
+  ...
+}
+```
+
+If `enable_recovery_seal` is set to `true`, a recovery seal will be
+autogenerated on OpenBao when an auto seal's recovery key is generated. This
+recovery seal will be named `<seal name>-recovery`. For example, if the auto
+seal is named `default`, its recovery seal will be named `default-recovery`. To
+prevent naming conflicts, OpenBao will prohibit manually onfigured seals from
```
wslabosz-reply:
```suggestion
prevent naming conflicts, OpenBao will prohibit manually configured seals from
```

website/content/docs/rfcs/recovery-seal.mdx
```diff
@@ -0,0 +1,302 @@
...
+seal "hsm" {
+  name = "default"
+  enable_recovery_seal = true
+  ...
+}
+```
+
+If `enable_recovery_seal` is set to `true`, a recovery seal will be
+autogenerated on OpenBao when an auto seal's recovery key is generated. This
+recovery seal will be named `<seal name>-recovery`. For example, if the auto
```
wslabosz-reply: I'm not really sure how to phrase that, but I feel like it's unclear as it is now.
```suggestion
If `enable_recovery_seal` is set to `true`, a recovery seal will created during
generation of a specified auto seal's recovery key. This
recovery seal will be named `<seal name>-recovery`. For example, if the auto
```

Huy-Dinh-Reply: Thanks, yeah it did sound unclear, I changed it now, let me know what you think.

Huy-Dinh-Reply:
@cipherboy I did some thinking after our conversation on horizontal scalability seal status propagation. Learning that recovery keys can be used to prompt auto seals to resume their attempts to unseal gave me a new perspective on this RFC.
I now feel like with the proposal I'm overloading the recovery keys with 2 different jobs: a "resume" button and a "break glass" button. That's why I'm leaning a bit towards the alternative of having the recovery seal in a separate stanza, which gives us a separate key materials for each purpose.
When you have a spare moment I would love to get your thoughts here as well. Really appreciate it! 😃

cipherboy:
@Huy-Dinh-Reply Yeah, I agree, separate keys seems best. 

Can we change all the terminology so `recovery` becomes a different, more descriptive word? It'll require a lot of re-education and aliasing of endpoints but will be clearer long-term.

Or, make it clear recovery is in very limited scenarios and not for unseal? Make the alternative Shamir's take a different name?

(If we do that, we're really most of the way to parallel unseal anyways, minus 2->3+).

Thoughts?

klaus-sap:
Hi, for the "break glass" case I would like to vote for the terminology "emergency keys" - wdyt?

Huy-Dinh-Reply:
@klaus-sap @cipherboy Good point, for the name I'm also having something like `emergency-seal`/`backup-seal` in mind, it's more of a concept, umbrella name because the user would be able to name their alternative Shamir's seal to their liking in the seal stanza.

A full rename of the `recovery` stuff sounds like a big project, and probably a separate effort for later. For now I'm inclined towards the idea of stressing the use cases in our documentation. I noticed we already touch on this [here](https://github.com/openbao/openbao/blob/17a15121d531e050e1f29d2d6c8ead31c92a1800/website/content/docs/concepts/seal.mdx?plain=1#L117-L118) but we can definitely make it clearer. 😄 

And you're right, this takes us most of the way to parallel unseal. However, I did intentionally leave out the ability to rotate the root keys when the primary auto seal is offline. My thinking is that it's a reasonable trade off for now to keep things simple. The catch up logic feels more critical when we have a setup with multiple auto seals, and a bit less so in a setup with one auto seal and one Shamir's seal. We can build this on top of this foundation once we move on to the full Parallel Unseal implementation.

How does this plan sound to you? If it sounds reasonable I'll update the RFC with this.

klaus-sap:
LGTM

cipherboy:
@Huy-Dinh-Reply Works for me.

> And you're right, this takes us most of the way to parallel unseal. However, I did intentionally leave out the ability to rotate the root keys when the primary auto seal is offline. My thinking is that it's a reasonable trade off for now to keep things simple. The catch up logic feels more critical when we have a setup with multiple auto seals, and a bit less so in a setup with one auto seal and one Shamir's seal. We can build this on top of this foundation once we move on to the full Parallel Unseal implementation.

This is fair, I think that is a bit more work.

Huy-Dinh-Reply:
@klaus-sap @cipherboy @wslabosz-reply thanks for the feedback! I've updated the proposal to go with the alternative, using an explicit seal stanza for the break-glass seal. I'm calling it "Emergency Seal" for now, but am very open to other suggestions.

Once you have a moment I would really appreciate it if you could have another look at the RFC to see if this direction feels right.

klaus-sap:
LGTM 👍

Huy-Dinh-Reply:
@cipherboy I've updated the RFC to reflect the latest discussion about using separate key material. Whenever you have a moment, could you take a look and see if it can be ready to merge? Thanks.

website/content/docs/rfcs/index.mdx
```diff
@@ -26,6 +26,8 @@ Steering Committee.
    metrics based on `go-metrics` with OpenTelemetry based metrics.
  - [Components for efficient search over encrypted data](efficient-search-components.mdx), adds the
    necessary constructs to build indexes on data efficiently allowing for faster search operations.
+ - [Emergency Seal](emergency-seal.mdx), for adding a secondary Shamir's
```
cipherboy, line 29: @Huy-Dinh-Reply Sorry, it looks like I dropped this from the Accepted section during a refactor, mind re-adding it?

Huy-Dinh-Reply: thanks for the heads up, but I noticed that this entry is still there in this file on main, is there another place where I should look?

PR #1743, opened 2025-08-28
Fix bump-critical-deps & tidy-all

cipherboy:
Several of our release automation steps left off the new api/auth/jwt module along with the tools/ module. This switches to uses find statements to locate all go.mod files and adds a variant of sync-deps to update an out-of-tree go-kms-wrapping to the versions of dependencies pinned here.

PR #1832, opened 2025-09-19
Bump critical dependencies and to Go 1.25.1

cipherboy:
This runs `make bump-critical` and swaps to Go 1.25.1 for the OpenBao v2.5.0 release series.

### DIFF
--- a/scripts/sync-deps.sh
+++ b/scripts/sync-deps.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-modules=$(find . -name go.mod | grep -v '^./go.mod')
+modules=$(find . -name go.mod -mindepth 2)
 
 grep '^	' go.mod | while read line; do
 	pkg="$(awk '{print $1}' <<< "$line")"


satoqz:
CI failures seem to imply that golangci-lint needs a bump: https://github.com/golangci/golangci-lint/issues/5873#issuecomment-3183482574 (we use 2.1 currently)

satoqz:
Okay, I've taken the liberty to push some additional commits here since I've already done _too much_ investigation, let's see if we can get out of this mess...

scripts/sync-deps.sh
```diff
@@ -1,15 +1,17 @@
...
+
 grep '^	' go.mod | while read line; do
 	pkg="$(awk '{print $1}' <<< "$line")"
 	version="$(awk '{print $2}' <<< "$line")"
 
-	find . -name go.mod | while read go_mod; do
+	while read go_mod; do
 		dir="$(dirname "$go_mod")"
-		if grep -q "^	$pkg" "$go_mod" && grep "$pkg" "$go_mod" | grep -qv "$pkg $version"; then
+		if grep -q "^	$pkg " "$go_mod" && grep "$pkg" "$go_mod" | grep -qv "$pkg $version"; then
```
satoqz, line 11: Note the additional space here. The lack thereof was causing a loop between `make sync-deps` and `make tidy-all`. The script would see that `go.mod` has `google.golang.org/genproto` and `sdk/go.mod` has `google.golang.org/genproto/googleapis/rpc`, then run `go get` for `google.golang.org/genproto` in `sdk/`, which `make tidy-all` would remove again.

Basically, this was never _really_ an issue, but created confusing output.

satoqz (approved):
Cool, green now

satoqz:
Ok, hmm, well. Should probably move to v2 instead?

`physical/raft/msgpack.go`
```
// If we downgrade msgpack from v1.1.5 to v0.5.5, everything will still
// work, but any pre-existing raft clusters will break on upgrade.
// This file exists so that the Vault project has an explicit dependency
// on the library, which allows us to pin the version in go.mod.
```

Just based on the notes at https://github.com/hashicorp/go-msgpack/releases, it would seem ok to upgrade. Let me do that...

satoqz:
Ugh, this is a disaster. If we stop pinning to `go-msgpack@v1.1.5` and go back to `v0.5.5`, we lose compatibility. If we move to `go-msgpack/v2`, we can't indirectly pin `raft-boltdb` to use a newer version of `go-msgpack` than `v0.5.5` anymore -- the fix would be to upgrade `raft-boltdb`, which depends on `go-msgpack/v2` in a newer version, but that circles back to https://github.com/openbao/openbao/pull/920.

Unless I've missed something, the most practical thing we can do right now is stick to the retracted version?

Excuse me @cipherboy for all the intrusion - feel free to drop commits as you see fit. I was just hoping to quickly patch this up to move dependency PRs forward 😄

scripts/sync-deps.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-modules=$(find . -name go.mod)
+modules=$(find . -name go.mod | grep -v '^./go.mod')
```
satoqz: FWIW, the only reason I've visibly been obsessed with the `go-msgpack` warning is that it started showing up following https://github.com/openbao/openbao/pull/1743 because `make sync-deps` now compares `/go.mod` to `/go.mod`. I'm not sure we want that in the context of that script, even if it surfaces some fair-ish warnings.

cipherboy:
> Unless I've missed something, the most practical thing we can do right now is stick to the retracted version?

Yep, that's what I think we're stuck with. Your analysis is where I ended up as well. I wonder if there's a route towards dropping old BoltDB (pre-etcd/bbolt) support

scripts/sync-deps.sh
cipherboy: @satoqz Ah, I see. Yes, let's skip the root `go.mod` file.

.github/workflows/code-checker.yml
```diff
@@ -36,36 +36,36 @@ jobs:
       - uses: ./.github/actions/set-up-go
       - uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
         with:
-          version: v2.1
+          version: v2.5
```
cipherboy, line 39: Ah, I guess Dependabot won't update this for us?

cipherboy (commented):
Thanks @satoqz!

.github/workflows/code-checker.yml
satoqz: No, that's a parameter completely internal to that particular action. The hash above is already at latest, it has nothing to do with the actual version of the golangci-lint binary.

cipherboy:
@JanMa Any chance I could get a review on this? Thanks!

satoqz:
@DrDaveD the main bump in this PR is that we'd build our own releases & test in CI with Go 1.25 (see `.go-version`). Nothing changes in terms of Go compatibility, this is already at Go 1.24.6 as of https://github.com/openbao/openbao/pull/1637 😄

https://github.com/openbao/openbao/blob/3e7d811721ae5ff9dbe097f410791fb733bf307c/go.mod#L12

None of the additional dependency bumps in this PR change that.

When I `dnf install go` inside of https://hub.docker.com/r/redhat/ubi10, I seem to get 1.24.6? Perhaps we are fine?

DrDaveD:
> @DrDaveD the main bump in this PR is that we'd build our own releases & test in CI with Go 1.25 (see `.go-version`).

Oh, well that should be OK.

> Nothing changes in terms of Go compatibility, this is already at Go 1.24.6 as of #1637 😄
> 
> https://github.com/openbao/openbao/blob/3e7d811721ae5ff9dbe097f410791fb733bf307c/go.mod#L12
> 
> None of the additional dependency bumps in this PR change that.

I had to apply a packaging patch to change that to 1.24.4.

> When I `dnf install go` inside of https://hub.docker.com/r/redhat/ubi10, I seem to get 1.24.6? Perhaps we are fine?

You're right, that is now the current version on el8/9/10, I missed that.

So it's OK to bump .go-version, but please don't increase go.mod over 1.24.6 for now.  I should probably add a CI check to verify that the version on el8 is no older than what go.mod requires.

satoqz:
Merging this one because I feel the pressure from every new PR hitting the testify issue.
FWIW, locally merging this one before merging e.g. https://github.com/openbao/openbao/pull/1834 did not cause conflicts, but don't hold me to that :^)

PR #1833, opened 2025-09-20
PKI: Add `allowed_ip_sans_cidr` to roles

### LINKED ISSUE #1815: Add option for ACME server to limit IP SANs to within certain CIDR(s)
When hosting my own CA, I try to use EAB to tie every single device to a single set of domains and IPs that belong to that device. In this case, even if the acme key is compromised, an attacker can only MITM the device it compromises, which does not realistically add any security risk as the device is already compromised and the attacker probably already had access to the TLS private key. However, OpenBao only support `allow_ip_sans` without any granular control, in the previous case, the attacker would be able to request certificates for other IPs if combined with other attacks (e.g. ARP spoofing) or simply wait the target to be offline and assign a static IP within the lan, and impersonate that device with a valid certificate.

Add an `allowed_ip_sans_cidr` or `allowed_ip_cidr` parameter for PKI engine.

Another option would be securing the IP assignment/lan access, but this complicates the network for other users/devices that does not need a certificate.

invis-z:
Add `allowed_ip_sans_cidr` parameter to PKI role system, to provide additional checks for IP SANs.

Resolves #1815

### DIFF
--- a/builtin/logical/pki/path_roles.go
+++ b/builtin/logical/pki/path_roles.go
@@ -164,9 +164,10 @@ CN and DNS SANs, and the host part of email addresses. Defaults to true.`,
 		},
 
 		"allow_ip_sans": {
-			Type:        framework.TypeBool,
-			Required:    true,
-			Description: `If set, IP Subject Alternative Names are allowed.`,
+			Type:     framework.TypeBool,
+			Required: true,
+			Description: `If set, IP Subject Alternative Names are allowed.
+Allowed IP ranges can be further restricted using allowed_ip_sans_cidr.`,
 		},
 
 		"allowed_ip_sans_cidr": {
@@ -555,9 +556,10 @@ CN and DNS SANs, and the host part of email addresses. Defaults to true.`,
 			},
 
 			"allow_ip_sans": {
-				Type:        framework.TypeBool,
-				Default:     true,
-				Description: `If set, IP Subject Alternative Names are allowed.`,
+				Type:    framework.TypeBool,
+				Default: true,
+				Description: `If set, IP Subject Alternative Names are allowed.
+Allowed IP ranges can be further restricted using allowed_ip_sans_cidr.`,
 				DisplayAttrs: &framework.DisplayAttributes{
 					Name:  "Allow IP Subject Alternative Names",
 					Value: true,
--- a/website/content/api-docs/secret/pki.mdx
+++ b/website/content/api-docs/secret/pki.mdx
@@ -3808,8 +3808,8 @@ allowing update of specific fields. Note that `bao write` uses `POST`.
   allowed for CNs, DNS SANs, and the host part of email addresses.
 
 - `allow_ip_sans` `(bool: true)` - Specifies if clients can request IP Subject
-  Alternative Names. No authorization checking is performed except to verify
-  that the given values are valid IP addresses.
+  Alternative Names. Allowed IP ranges can be further restricted if
+  allowed_ip_sans_cidr is set.
 
 - `allowed_ip_sans_cidr` `(string: "")` - Specifies the IP CIDRs this role is
   allowed to issue certificates for. This is used with the allow_ip_sans to 


invis-z:
@cipherboy I tried to cobble this together. Writing code is not my stronghold, and some of the solutions might be kind of crude. Please review and provide feedback. Thanks!

builtin/logical/pki/path_roles.go
```diff
@@ -163,10 +164,19 @@ CN and DNS SANs, and the host part of email addresses. Defaults to true.`,
...
+		"allowed_ip_sans_cidr": {
+			Type:     framework.TypeCommaStringSlice,
 			Required: true,
-			Description: `If set, IP Subject Alternative Names are allowed.
-Any valid IP is accepted and No authorization checking is performed.`,
+			Description: `Specifies the IP CIDRs this role is allowed
+to issue certificates for. This is used with the allow_ip_sans to 
+determine matches for the common name, and IP-typed SAN entries of
+certificates. See the documentation for more information. This parameter
+accepts a comma-separated string or list of CIDR notated IPs.`,
```
cipherboy:
```suggestion
accepts a comma-separated string or list of CIDR notated IPs.
Roles with no value allow all IP ranges; to disable IP sans entirely
set allow_ip_sans=false instead.`,
```

Could you also add these to the documentation (`website/content/api-docs/secret/pki.mdx`)? Thanks!

(also relevant below)

builtin/logical/pki/cert_util.go
```diff
@@ -1389,7 +1389,24 @@ func generateCreationBundle(b *backend, data *inputBundle, caSign *certutil.CAIn
...
+								break
+							}
+						}
+						if valid {
+							ipAddresses = append(ipAddresses, parsedIP)
+							continue
+						} else {
+							return nil, nil, errutil.UserError{Err: fmt.Sprintf(
+								"the IP address %q is not allowed in this role", v)}
+						}
```
cipherboy: Just a style nit.


```suggestion

						if !valid {
							return nil, nil, errutil.UserError{Err: fmt.Sprintf(
								"the IP address %q is not allowed in this role", v)}
						}
						
						ipAddresses = append(ipAddresses, parsedIP)
```

builtin/logical/pki/backend_test.go
```diff
@@ -3549,6 +3549,85 @@ func TestBackend_URI_SANs(t *testing.T) {
...
+
+	var err error
+
+	_, err = CBWrite(b, s, "root/generate/internal", map[string]interface{}{
+		"ttl":         "40h",
+		"common_name": "example.com",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
```
cipherboy, line 3564: If you want, there's a few useful helpers like `requireSuccessNonNilResponse(...)` and you can use `require.*` functions. But happy to take this as-is as well, I think it is sufficient.

invis-z: I basiclly copied this function from `TestBackend_URI_SANs`, I will need to really understand the other functions before I can use the helpers. For this PR, I think I will keep it as-is for now.

website/content/api-docs/secret/pki.mdx
```diff
@@ -3811,6 +3811,14 @@ allowing update of specific fields. Note that `bao write` uses `POST`.
   Alternative Names. No authorization checking is performed except to verify
   that the given values are valid IP addresses.
```
cipherboy: Sorry to nitpick; I missed this in the last iteration!

Should this be updated to clarify allowed ranges can be specified by `allowed_ip_sans_cidr`?

builtin/logical/pki/path_roles.go
```diff
@@ -543,16 +555,24 @@ CN and DNS SANs, and the host part of email addresses. Defaults to true.`,
 			},
 
 			"allow_ip_sans": {
-				Type:    framework.TypeBool,
-				Default: true,
-				Description: `If set, IP Subject Alternative Names are allowed.
-Any valid IP is accepted and No authorization checking is performed.`,
+				Type:        framework.TypeBool,
+				Default:     true,
+				Description: `If set, IP Subject Alternative Names are allowed.`,
```
cipherboy: As above about `allowed_ip_sans_cidr`.

cipherboy (approved):
@invis-z If you don't want to clean up the documentation, let me know! Otherwise, this looks good to me. Thank you very much @invis-z!

cipherboy:
Not entirely sure why I didn't merge this sooner. Thanks @invis-z!

PR #1834, opened 2025-09-21
feat(core): Add metrics_only and disallow_metrics listener options

### LINKED ISSUE #1704: disallow_metrics and metrics_only listener options

Metrics have two modes:

1. Authenticated via regular request token, subject to ACL evaluation
2. Unauthenticated conditionally on `unauthenticated_metrics_access`.

However, like audit and rotation, there's some potential information disclosure here. Why should API operators be able to scrape metrics, other than metrics _happening_ to look like another OpenBao endpoint? 

Sometimes metrics need to be exposed beyond the normal cluster; a custom listener with `metrics_only=true` and `unauthenticated_metrics_access=true` would be great for this use case.

Additionally, on the normal cluster listener (where client traffic comes in), `disallow_metrics=true` would be good to set, so anyone receiving API access cannot necessarily scrape audit information.

These should be per-listener. 

n/a.

This came up internally. 

**Additional context**

The related improvements to `unsafe_allow_api_audit_creation` and `disable_unauthed_rekey_endpoints` are a good basis. Behavior shouldn't change by default, though, as nothing is yet proven to be fundamentally insecure as a basis of this. Metrics should be non-sensitive data only.

kyounghunJang:
This PR resolves #1704 by introducing two new per-listener configuration options to provide finer-grained control over metrics exposure, enhancing security.

- `metrics_only=true`: When set, the listener will only serve the `/sys/metrics` endpoint. This is ideal for exposing metrics to a monitoring system on a dedicated port without exposing other API endpoints.

- `disallow_metrics=true`: When set, the listener will explicitly forbid access to the `/sys/metrics` endpoint. This can be used to harden the primary API listeners by preventing API clients from scraping metrics.

Resolves: #1704


eyenx:
Hi @kyounghunJang thanks for your contribution.

How does the second option differ from just using the `unauthenticated_metrics_access` option in the telemetry parameter of the listener? Usually you'd want to only allow authorized access to metrics.

kyounghoonJang:
@eyenx 
Thanks for the question!

`unauthenticated_metrics_access` in the telemetry block controls who can access /sys/metrics 
(authenticated vs unauthenticated).

The proposed `disallow_metrics` option instead controls whether the endpoint is available at all on a given listener, regardless of authentication.

- unauthenticated_metrics_access → governs authentication requirements (token required vs open access).
- disallow_metrics → disables the /sys/metrics endpoint entirely on that listener, even for authenticated clients.

This allows a setup such as:
- Primary API listener: disallow_metrics=true (no metrics exposure)
- Dedicated metrics listener: metrics_only=true (only metrics exposed, optionally unauthenticated)

cipherboy:
@eyenx @kyounghunJang Yes, exactly. In particular, if you have a system with privileged tokens and they leak (but, you can't get into the internal control plane where metrics are scraped), the `disallow_metrics=true` option would prevent this token from being used to scrape metrics on a global/external interface.

http/metrics.go
```diff
@@ -0,0 +1,41 @@
+// http/metrics.go
```
satoqz:
```suggestion
// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
// SPDX-License-Identifier: MPL-2.0
```

satoqz (commented):
Code looks good to me, additions to documentation in `website/content/docs/configuration/listener/{tcp,unix}.mdx` would be nice!

eyenx:
> @eyenx @kyounghunJang Yes, exactly. In particular, if you have a system with privileged tokens and they leak (but, you can't get into the internal control plane where metrics are scraped), the `disallow_metrics=true` option would prevent this token from being used to scrape metrics on a global/external interface.

Usually we try to use short lived tokens to access the metrics endpoint. But good for me.  I find this a good option, but feel like people will then just make the metrics endpoint available on another port (only internally) which will be better, but making it then accessible w/o authentication which is IMO not the best approach. But as long as we can still enable authentication there, I'm fine.

kyounghoonJang:
> Code looks good to me, additions to documentation in `website/content/docs/configuration/listener/{tcp,unix}.mdx` would be nice!

@satoqz 
Thanks for the suggestion! I’ve added the docs files to the commit as well. 
Looking forward to your feedback.

internalshared/configutil/listener.go
```diff
@@ -147,6 +147,8 @@ type Listener struct {
 	// an authenticated variant with new semantics is available on a new
 	// endpoint, this will be set to true (disabling request handling).
 	DisableUnauthedRekeyEndpoints bool `hcl:"disable_unauthed_rekey_endpoints"`
+	DisallowMetrics               bool `hcl:"disallow_metrics"`
+	MetricsOnly                   bool `hcl:"metrics_only"`
```
satoqz: Should these go into the `telemetry` section together with the `UnauthenticatedMetricsAccess` field? See `ListenerTelemetry` struct.

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -119,6 +119,12 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
   `/sys/rekey-recovery-key/*`). These are a security risk to leave exposed on
   public listeners.
 
+- `disallow_metrics` `(bool: false)` - Specifies if the `/v1/sys/metrics` endpoint is disabled on this listener. 
+  This is useful for hardening a listener that handles general client traffic, preventing API clients from scraping metrics.
+
+- `metrics_only` `(bool: false)` - Specifies if the listener should *only* serve the `/v1/sys/metrics` endpoint, blocking all other API requests. 
+  This is useful for creating a dedicated, secure listener for monitoring systems.
```
satoqz: If we honor the above comment, we should group these under the "telemetry parameters" section below, which already contains `unauthenticated_metrics_access`.

website/content/docs/configuration/listener/unix.mdx
```diff
@@ -32,6 +32,12 @@ multiple sockets.
   `/sys/rekey-recovery-key/*`). These are a security risk to leave exposed on
   public listeners.
 
+- `disallow_metrics` `(bool: false)` - Specifies if the `/v1/sys/metrics` endpoint is disabled on this listener. 
+  This is useful for hardening a listener that handles general client traffic, preventing API clients from scraping metrics.
+
+- `metrics_only` `(bool: false)` - Specifies if the listener should *only* serve the `/v1/sys/metrics` endpoint, blocking all other API requests. 
+  This is useful for creating a dedicated, secure listener for monitoring systems.
```
satoqz, line 44: This file doesn't seem to document `unauthenticated_metrics_access` at all, do you mind creating a new section here to keep parity with `tcp.mdx`?

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -389,6 +395,29 @@ listener "tcp" {
...
+  address            = "127.0.0.1:8200"
+  disallow_metrics   = true
+}
+
+# Dedicated listener for monitoring systems (e.g., Prometheus)
+# Only serves metrics and blocks all other requests.
+listener "tcp" {
+  address                        = "127.0.0.1:9101"
+  metrics_only                   = true
+  unauthenticated_metrics_access = true
```
```suggestion
  telemetry {
    unauthenticated_metrics_access = true
  }
```

(At least) this one would be nested in a `telemetry` block. Also see the example slightly above: https://github.com/kyounghunJang/openbao/blob/1066ee126c2744ee0062d4da8af97f63f89fc5a9/website/content/docs/configuration/listener/tcp.mdx?plain=1#L379-L381

internalshared/configutil/listener.go
kyounghoonJang: I agree with your point — it makes sense that metrics-related items should be grouped and managed under ListenerTelemetry. I think it’s correct to update it that way.

website/content/docs/configuration/listener/unix.mdx
kyounghoonJang: I’ll update the docs and add it in as well while making the changes.

kyounghoonJang: @satoqz 

In my opinion, it doesn't seem necessary to add documentation for the unauthenticated_metrics_access setting specifically for Unix listeners. Access to Unix sockets is already controlled by filesystem permissions (socket_mode, socket_user, socket_group), so I believe this setting is redundant in that context.

On the other hand, the disallow_metrics and metrics_only options could still be very useful for Unix listeners. Since they explicitly control which endpoints are served, I think it would be appropriate to add them to the Telemetry documentation.

What do you think of this approach?

satoqz: @kyounghunJang you are not wrong to doubt the use case, but that's besides the point: `unauthenticated_metrics_access` _is_ an available option for unix listeners _already_, so it should be documented either way, IMO. You don't have to do that in this PR though, it was just a drive by suggestion :)

kyounghoonJang: @satoqz 
Thanks for your feedback. 
I’ll add documentation for disallow_metrics and metrics_only under the Telemetry section in the Unix docs, and organize it while doing so.

website/content/docs/configuration/listener/unix.mdx
```diff
@@ -77,3 +88,32 @@ listener "unix" {
...
+```hcl
+# Primary listener for general API traffic via a UNIX socket
+# Metrics are disabled for security.
+listener "unix" {
+  path                 = "/var/run/openbao.sock"
+  socket_mode          = "0660" 
+  telemetry {
+    disallow_metrics   = true
+  }
+}
```
satoqz, line 108:
```suggestion
listener "unix" {
  path        = "/var/run/openbao.sock"
  socket_mode = "0660"
  telemetry {
    disallow_metrics = true
  }
}
```

website/content/docs/configuration/listener/unix.mdx
```diff
@@ -77,3 +88,32 @@ listener "unix" {
...
+
+# Dedicated listener for monitoring systems (e.g., Prometheus)
+# Only serves metrics and blocks all other requests.
+listener "unix" {
+  path                             = "/var/run/openbao-metrics.sock"
+  socket_mode                      = "0660" 
+  telemetry{
+    metrics_only                   = true
+  }
+}
```
satoqz, line 118:
```suggestion
listener "unix" {
  path        = "/var/run/openbao-metrics.sock"
  socket_mode = "0660"
  telemetry {
    metrics_only = true
  }
}
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -389,6 +395,33 @@ listener "tcp" {
...
+
+# Dedicated listener for monitoring systems (e.g., Prometheus)
+# Only serves metrics and blocks all other requests.
+listener "tcp" {
+  address                          = "127.0.0.1:9101"
+  telemetry{
+    metrics_only                   = true
+    unauthenticated_metrics_access = true
+  }
+}
```
satoqz, line 422:
```suggestion
# Primary listener for general API traffic
# Metrics are disabled for security.
listener "tcp" {
  address = "127.0.0.1:8200"
  telemetry {
    disallow_metrics = true
  }
}

# Dedicated listener for monitoring systems (e.g., Prometheus)
# Only serves metrics and blocks all other requests.
listener "tcp" {
  address = "127.0.0.1:9101"
  telemetry {
    metrics_only                   = true
    unauthenticated_metrics_access = true
  }
}
```

kyounghoonJang:
@satoqz 
Thank you for kindly providing feedback and pointing out the code changes!

satoqz (approved):

changelog/1834.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
cipherboy:
```suggestion
```release-note:improvement
```

http/handler.go
```diff
@@ -256,8 +256,8 @@ func handler(props *vault.HandlerProperties) http.Handler {
 	corsWrappedHandler := wrapCORSHandler(helpWrappedHandler, core)
 	quotaWrappedHandler := rateLimitQuotaWrapping(corsWrappedHandler, core)
 	genericWrappedHandler := genericWrapping(core, quotaWrappedHandler, props)
-	wrappedHandler := wrapMaxRequestSizeHandler(genericWrappedHandler, props)
-
+	metricsWrappedHandler := wrapMetricsListenerHandler(genericWrappedHandler, props)
```
cipherboy, line 259: @kyounghunJang This is an elegant solution and one I hadn't thought of. Originally I was thinking just special casing all of the listeners, but I think I prefer this approach. Thank you!

cipherboy:
I'll merge once CI passes after my minor CL update. Thanks @kyounghunJang!

PR #1844, opened 2025-09-24
Adjust audit device creation on active node only

cipherboy:

### DIFF
--- a/vault/audit.go
+++ b/vault/audit.go
@@ -750,7 +750,7 @@ func (c *Core) handleAuditLogSetup(ctx context.Context) error {
 
 func (c *Core) addAuditFromConfig(ctx context.Context, auditConfig *server.AuditDevice) error {
 	if c.standby {
-		c.logger.Warn("audit device present in standby but not active node", "path", auditConfig.Path)
+		c.logger.Warn("audit device present in local configuration but not in the configuration of the active node", "path", auditConfig.Path)
 		return nil
 	}
 


cipherboy:
@phil9909 I adjusted this one so it'd behave more like expectations; lmkwyt.

vault/audit.go
```diff
@@ -747,6 +749,11 @@ func (c *Core) handleAuditLogSetup(ctx context.Context) error {
 }
 
 func (c *Core) addAuditFromConfig(ctx context.Context, auditConfig *server.AuditDevice) error {
+	if c.standby {
+		c.logger.Warn("audit device present in standby but not active node", "path", auditConfig.Path)
+		return nil
+	}
+
```
phil9909, line 756: I guess this will led to false positives when you have just added a new audit device and are syncing the config across your nodes, but that might be fine.

Also, I had a hard time understanding the sentence. I think we should add a little more details to make it easier for end-users to understand.

```suggestion
	if c.standby {
		c.logger.Warn("audit device present in local configuration but not in the configuration of the active node", "path", auditConfig.Path)
		return nil
	}

```

cipherboy: Ah, that's much better. Thanks @phil9909!

cipherboy:
@phil9909 Yes, I had mentioned this in the commit message: 

```
When called in the new read-enabled standby, log warning but do not
treat them as fatal. This is because the order w.r.t. replication and
config processing between nodes may be different and the configuration
reload has not yet happened (or, the storage corresponding to the audit
device has yet to be replicated).
```

I think it is unfortunate not to know where we are at w.r.t. both `SIGHUP` and physical storage (really, it is a cross-storage synchronization problem with a manual update procedure so it is not entirely unexpected).

But, it at least provides something for administrators to look out for when updating configuration. Luckily these can be added by SIGHUP so in a well-controlled environment, need not require a full pod/service restart.

phil9909:
> @phil9909 Yes, I had mentioned this in the commit message:

Sorry, missed that.

> Luckily these can be added by SIGHUP so in a well-controlled environment, need not require a full pod/service restart.

Agree, the proper prosedure would be:

1. Update config on all nodes
2. Send `SIGHUP` to active node

But I fear many (most?) people will:

1. Update config on all nodes
2. Restart node (either active or all)

phil9909 (approved):

PR #1779, opened 2025-09-05
fix(backend): Missing OIDCFlow

mase-informaticon:
The OIDC auth method case was missing when constructing the JWK validator

Fix:
- Add OIDCFlow

Resolves #1778

PR #1845, opened 2025-09-24
Change gomod check to run on el8

DrDaveD:
This changes the gomod linter check to run in an almalinux:8 container so it can verify that openbao will work with the golang version that is current there.  This is needed for EPEL packaging to be successful.

### DIFF
--- a/.github/workflows/code-checker.yml
+++ b/.github/workflows/code-checker.yml
@@ -85,11 +85,20 @@ jobs:
         run: semgrep ci --include '*.go' --config 'tools/semgrep/ci'
 
   gomod:
+    # Run in an almalinux:8 container to make sure it works with the
+    # native golang version there. That is needed for EPEL packaging.
     name: Go mod checks
     runs-on: ubuntu-latest
+    container: ghcr.io/almalinux/8-base
     steps:
+      - name: Install required packages
+        run: dnf -y install golang git make
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-      - uses: ./.github/actions/set-up-go
+      - name: Set safe.directory
+        # This is supposed to be done by default in the checkout action
+        # but it doesn't work within a container.
+        # See https://github.com/actions/checkout/issues/2031
+        run: git config --global --add safe.directory $(pwd)
       - name: Run go mod tidy to check for differences
         run: make ci-tidy-all
       - name: Run go.mod synchronization to check for differences


.github/workflows/code-checker.yml
```diff
@@ -85,11 +85,20 @@ jobs:
         run: semgrep ci --include '*.go' --config 'tools/semgrep/ci'
 
   gomod:
+    # Run in an almalinux:8 container to make sure it works with the
+    # native golang version there. That is needed for EPEL packaging.
     name: Go mod checks
     runs-on: ubuntu-latest
+    container: almalinux:8
```
satoqz:
```suggestion
    container: ghcr.io/almalinux/8-micro
```

Seems they also publish to ghcr.io, so that's less likely to run into rate limits and is maybe faster to pull. One of the smaller variants (`-micro`, `-minimal`) may also suffice?

DrDaveD:
> Personally not a big fan of bending to downstream's rules, just because it's a popular one. I'd much prefer to move Go versions up as we like (and benefit the numerous ecosystems that will package latest Go just fine) and leave ecosystems that won't ship recent compiler versions on similarly-dated OpenBao versions, too.

I don't like that at all.  That puts a lot of strain on me as the maintainer, having to backport any security fixes that come out during that period or figure out how to convert the newer version to the older go, and asking the users who want the newer version to be patient.

satoqz:
> I don't like that at all. That puts a lot of strain on me as the maintainer, having to backport any security fixes that come out during that period or figure out how to convert the newer version to the older go

I totally understand that concern, hence I suggested we apply a different policy for patch releases. I think there are several scenarios we need to consider separately:

1. We release a patch version (e.g., 2.4.2) for a minor version (2.4.x) that was already packaged downstream because the Go version available downstream (1.24.6) initially supported 2.4.0. In this case, you should not have to do any additional work, any backports done upstream should remain compatible with the Go version initially used for 2.4.0.

2. We release a minor version (e.g., 2.5.0), potentially with a higher minimum Go version than the previous minor version. In this case, whether you can package it depends on whether the new required Go version is already available downstream. Yes, users will have to wait. This is how those distributions work, after all you must also wait for the newer Go version to be available. While I'm not fully familiar with Fedora's process, I doubt that it makes sense for a Fedora release cycle to bump OpenBao to 2.5.0 but _not bump_ Go to 1.25.x when Go 1.25.x was released _before_ OpenBao 2.5.0.

3. We release a patch version (e.g., 2.5.1) for a minor version (2.5.x) that _has not_ been packaged downstream. This patch version fixes issues that are only present in 2.5.x, not in previous releases. In this case, you don't have to do anything.

4. We release a patch version (e.g., 2.5.1) for a minor version (2.5.x) that _has not_ been packaged downstream. (But!) this patch version fixes issues that are also relevant for previous releases _that have been packaged downstream_ (e.g., 2.4.1). This is the only case that truly hurts, and you'd need to backport. You can do that either downstream only, or propose an additional patch version upstream (e.g., 2.4.2) to benefit everyone else. Other maintainers could then also help with that process. Is this not the standard? I'm not sure the default solution for fixed-release distributions is to bump minor versions (which, in OpenBao, primarily comprise features and potentially breaking changes!) just to get the latest security patches.

> asking the users who want the newer version to be patient.

Is this a real concern for users of the Enterprise Linux family of distributions (besides security patches)? Most software in those repositories is not at the latest version, this is what you opt into as their user. If you really must have the latest version, or have arbitrary control over the version in the first place, containers should be a convenient escape hatch, or perhaps the correct initial choice.

cipherboy:
@DrDaveD and @satoqz Before I wade back into Fedora packaging, I think its good to get things sorted out upstream first. :-) Then we can present a unified picture downstream.

---

> I don't like that at all. That puts a lot of strain on me as the maintainer, having to backport any security fixes that come out during that period or figure out how to convert the newer version to the older go, and asking the users who want the newer version to be patient.

The biggest reason I'm personally itchy about newer toolchains is that it is the only place they'll ship native FIPS and Post-Quantum updates. Native FIPS isn't a concern for downstream distros and one might say that PQC falls antithetical to EPEL's FIPS desires. Ironically though, Go 1.24 saw more progress in both of those areas, though Go 1.25 saw great performance improvements in native FIPS mode. 

It is the same strain that I am under internally, fwiw: https://gitlab.com/gitlab-org/govern/secrets-management/openbao-internal/-/blob/main/Makefile#L50 

I internally remove the go version and will be stuck patching for any incompatibility that way. I'm happy to share patches.

---

That said, only @DrDaveD's approach truly solves Fedora/EPEL packaging right now.

@satoqz said:

>> asking the users who want the newer version to be patient.
>
> Is this a real concern for users of the Enterprise Linux family of distributions (besides security patches)? Most software in those repositories is not at the latest version, this is what you opt into as their user. If you really must have the latest version, or have arbitrary control over the version in the first place, containers should be a convenient escape hatch, or perhaps the correct initial choice.

For EPEL, perhaps not. But even the latest Fedora only has Go 1.24.7 at the moment... Strictly, from what I recall of packaging guidelines, I think a proper Fedora strategy would only ever ship minor version updates. So your proposed minor release tied to Go minor release, while attractive, may not outlive the Fedora release. Fedora really assumes a multi-versioned support model but we only support the latest releases. And we only support a single one. 

(Historically, I think we're averaging around 3-4 months per minor release but Fedora is on a 6mo window, so that's 3mos of @DrDaveD doing manual bug/security backports, which I'm not sure is kind either.)


---


@satoqz I believe scenario 3 is very rare 

> 3. We release a patch version (e.g., 2.5.1) for a minor version (2.5.x) that has not been packaged downstream. This patch version fixes issues that are only present in 2.5.x, not in previous releases. In this case, you don't have to do anything.

v2.4.0 and v2.4.1 has several fixes which were present in v2.3.x series:

v2.4.0:

1. `sys/health` in #1730, 
2. `BAO_ADDR` typo in #1527 
3. Namespace/PKI infinite loop fix in #1696 
4. Identity<->namespace metrics panic in #1715 
5. PostgreSQL parallel table creation bug fix in #1506 

v2.4.1:

1. Security issue in #1756 
2. OIDC Flow fix in #1779. 

IMHO, all of these under a Fedora packaging guideline would be eligible for backport. 

> I'm not sure the default solution for fixed-release distributions is to bump minor versions (which, in OpenBao, primarily comprise features and potentially breaking changes!) just to get the latest security patches.

@satoqz and @DrDaveD The single-stream release process was really to encourage downstream paid-support places (like, well, Red Hat) to come join the community if they wanted to have upstream releases supported through the entire paid downstream release cycle. With our community doing first-party packaging in that ecosystem, it may just be time to revisit the support model.

IMHO, I'm happy doing security patch backports to the last supported version and try to aim for a 6mo support window for security patches for a given patch release, but we have enough bug fixes that I'd want there to be other users capable of doing full releases without requiring Jan or myself to be available before I suggest more than security patches.

My 2c.

DrDaveD:
> > > asking the users who want the newer version to be patient.
> > 
> > 
> > Is this a real concern for users of the Enterprise Linux family of distributions (besides security patches)? Most software in those repositories is not at the latest version, this is what you opt into as their user. If you really must have the latest version, or have arbitrary control over the version in the first place, containers should be a convenient escape hatch, or perhaps the correct initial choice.
> 
> For EPEL, perhaps not. But even the latest Fedora only has Go 1.24.7 at the moment... Strictly, from what I recall of packaging guidelines, I think a proper Fedora strategy would only ever ship minor version updates. 

Actually I think EPEL has more of an expectation of upgrades than Fedora.  Since new Fedora releases come out so frequently, the older but still active releases are expected to be more stable and pretty much only get patch releases unless there are security problems identified.  Yes RHEL itself is also expected to be mostly stable, but EPEL contains a lot of external packages which get frequent updates over the many years that RHEL versions are supported and they can happen at any time.  RHEL itself also has exceptions to the general policy against progressing minor versions, for example golang which is updated along with each of the RHEL minor releases every 6 months.  I think the excuse there is that golang has so many security fixes that it needs frequent updates; major security fixes get updated right away, and medium & low are updated with each RHEL minor release.

.github/workflows/code-checker.yml
DrDaveD: I switched to using ghcr.io/almalinux/8-base.  The micro and minimal variants appear to not have dnf so there's no way to install the additional needed packages.

DrDaveD:
I propose that we accept this PR for now to prevent unnecessary updates to newer golang versions and if we come across a really strong reason to update golang to a newer version, then come up with a strategy on how to handle that case and change this check.

cipherboy:
@DrDaveD Thoughts from community call (10/9/2025):

- GitLab [overrides the Go version](https://gitlab.com/gitlab-org/govern/secrets-management/openbao-internal/-/blob/main/Makefile?ref_type=heads#L50-54)
- Would be good for a second, dedicated pass on oldest-supported RHEL compiler. (Optional)
- Cannot do just release check as betas are cut from main, which may mean we don't know that GA is infeasible until later.

Can deviate from it, but as it gets closer to beta/patch releases, should validate that we're in compliance with it.

DrDaveD:
Replaced by #1937

PR #1846, opened 2025-09-25
Fix missing pipe character in table column

songlim327:
There is a missing pipe character (`|`) in the documentation. Added the character and give it a header.


satoqz (approved):
Thanks @songlim327! That looks a lot better :)

JanMa (approved):

PR #1847, opened 2025-09-26
Read replication: fix body cloning and runStandby race

cipherboy:
---

cc @fatima2003 @phil9909 

This is the last part of the rebase on main.


helper/buffer/buffer.go
```diff
@@ -0,0 +1,56 @@
...
+		return nil
+	}
+
+	rCloser, ok := b.rOrig.(io.Closer)
+	if !ok {
+		return nil
+	}
+
+	return rCloser.Close()
+}
```
phil9909: Alternative implementation: 
- if the original reader is a closer, we call `Close()` directly after `ReadAll(orig)`, no need to keep it open, if we consumed it completely, this allows to simplify the code so that we only need one wrapper type. 
- change the return type of `NewSeekableReader` to `io.ReadSeekCloser`

```suggestion
// nopSeekableReader implements ReadSeekCloser by forwarding read and seek to
// the underlying ReaderSeeker and implementing Close as a noop. Due to
// net/http.Request.Body contracts, we need to add close support while
// preserving Seek capability.
type nopSeekableReader struct {
	io.ReadSeeker
}

func (n nopSeekableReader) Close() error { return nil }

func NewSeekableReader(orig io.Reader) (io.ReadSeekCloser, error) {
	if seekReaderCloser, ok := orig.(io.ReadSeekCloser); ok {
		return seekReaderCloser, nil
	}

	if seekReader, ok := orig.(io.ReadSeeker); ok {
		return nopSeekableReader{seekReader}, nil
	}

	data, err := io.ReadAll(orig)
	if err != nil {
		return nil, err
	}

	if closer, ok := orig.(io.Closer); ok {
		err = closer.Close()
		if err != nil {
			return nil, err
		}
	}

	return &nopSeekableReader{
		bytes.NewReader(data),
	}, nil
}
```

http/json.go
```diff
@@ -85,17 +84,15 @@ func memoryForToken(rawToken json.Token) int64 {
 // or number of distinct strings (currently defined to be all strings,
 // including keys in a map, which aren't HMAC'd but will contribute to
 // overhead).
-func NewSafeJSONReader(ctx context.Context, parent io.Reader) (io.Reader, int64, int64, error) {
+func NewSafeJSONReader(ctx context.Context, reader io.Reader) (int64, int64, error) {
```
phil9909: I guess the name of this function (and the doc string) no longer make sense

```suggestion
func EnforceJSONComplexityLimits(ctx context.Context, reader io.Reader) (int64, int64, error) {
```

vault/ha.go
```diff
@@ -946,9 +1004,7 @@ func (c *Core) periodicCheckKeyUpgrades(ctx context.Context, stopCh chan struct{
 				lopCount := opCount
 
 				// Only check if we are a standby
-				c.stateLock.RLock()
-				standby := c.standby
-				c.stateLock.RUnlock()
+				standby := c.standby.Load()
 				if !standby {
```
phil9909, line 1008:
```suggestion
				if !c.standby.Load() {
```

vault/ha.go
```diff
@@ -408,6 +410,58 @@ func (c *Core) drainPendingRestarts() {
...
+		case <-timeout.C:
+			canceled = true
+		case <-stopCh:
+			canceled = true
+		case <-acquiredCh:
+		}
+
+		if !timeout.Stop() {
+			<-timeout.C
+		}
```
phil9909, line 438: I don't get this part, why do we need this?

http/util.go
```diff
@@ -43,15 +60,52 @@ func wrapMaxRequestSizeHandler(handler http.Handler, props *vault.HandlerPropert
...
 		if maxRequestSize > 0 {
-			r.Body = http.MaxBytesReader(w, r.Body, maxRequestSize)
+			clonedReq.Body = http.MaxBytesReader(w, clonedReq.Body, maxRequestSize)
 		}
+
 		ctx = logical.CreateContextOriginalBody(ctx, originalBody)
-		r = r.WithContext(ctx)
+		clonedReq = clonedReq.WithContext(ctx)
+
+		// Now that we've rated-limited the body, copy it into a buffer: we
```
phil9909:
```suggestion
		// Now that we've size-limited the body, copy it into a buffer: we
```

http/logical.go
```diff
@@ -116,49 +96,53 @@ func buildLogicalRequestNoAuth(w http.ResponseWriter, r *http.Request) (*logical
...
 			// a form or as JSON. The amount to look ahead (512 bytes) is arbitrary
 			// but extremely tolerant (i.e. allowing 511 bytes of leading whitespace
 			// and an incorrect content-type).
-			head, err := bufferedBody.Peek(512)
-			if err != nil && err != bufio.ErrBufferFull && err != io.EOF {
+			head := make([]byte, 0, 512)
+			n, err := io.ReadFull(r.Body, head)
+			if n >= 0 && n < 512 {
+				head = head[0:n]
+			}
```
phil9909: `ReadFull` will return an `ErrUnexpectedEOF`, if the buffer could only be filled partially (i.e. 0 and n bytes are ok, but 1, ..., n-1 is "unexpected"). 

```suggestion
			head, err := io.ReadAll(io.LimitReader(r.Body, 512))
```

cipherboy:
> Turning the standby bool into an atomic instead of protecting it with the stateLock means we can no longer modify it atomically with the other state, not sure if this is a problem.

@phil9909 We've not though. Yes, we've added the atomic but everywhere that cares about it remaining constant through a function still acquired the state lock.

helper/buffer/buffer.go
cipherboy: The only concern I have is around this:

> if the original reader is a closer, we call `Close()` directly after `ReadAll(orig)`, no need to keep it open, if we consumed it completely, this allows to simplify the code so that we only need one wrapper type.

It may be that calling close now causes other errors, but hopefully tests will pick that up for us.

vault/ha.go
cipherboy: Timers will leak channels if you don't stop them. When the channel reports that it is already stopped, you need to consume the value out of the channel as nobody else will.

http/json.go
cipherboy: Applied manually.

cipherboy:
> After merging this, we should really make sure to quickly merge the feature branch into main, because this has some nasty merge conflict potential (or cherry-pick some changes over to main and drop them from the feature branch).

I'm happy resolving merge conflicts until we're confident in the correctness of this feature branch. We need a more involved test suite, plus the legacy behavior change, before I'm confident merging to `main`.

cipherboy:
@phil9909 This has your changes applied. It looks like closing the request early is OK, though I'm having a hard time tracing that down in the go source code.

phil9909 (approved):

PR #1849, opened 2025-09-26
Add feature flag for read replication and fix some forwarding bugs

phil9909:
This PR:

- ~Adds a `feature_flags` stanza to the configuration~ Add a top level configuration `disable_standby_reads` (and documentation)
- ~Adds `ForwardPerformanceStandby` to kv engine "CUD" (CRUD without R) operations.~ change error wrapping in KVv1 put to support forwarding on standby.
- Fix a panic when invalidations arrive before an `activeContext` is created
- Fix requests are not forwarded, if the standby node does not have an `activeContext` (either because read replication is disabled or it is not done with post unseal).
- Fix potential nil pointer during shutdown in raft cache invalidation logic
- Actually call `HookInvalidate` (upps)

See #1550
See #1528 

### DIFF
--- a/builtin/logical/kv/passthrough.go
+++ b/builtin/logical/kv/passthrough.go
@@ -229,7 +229,7 @@ func (b *PassthroughBackend) handleWrite() framework.OperationFunc {
 			Value: buf,
 		}
 		if err := req.Storage.Put(ctx, entry); err != nil {
-			return nil, fmt.Errorf("failed to write: %v", err)
+			return nil, fmt.Errorf("failed to write: %w", err)
 		}
 
 		return nil, nil
--- a/physical/raft/raft.go
+++ b/physical/raft/raft.go
@@ -209,7 +209,15 @@ type RaftBackend struct {
 // HookInvalidate implements physical.CacheInvalidationBackend.
 func (r *RaftBackend) HookInvalidate(hook physical.InvalidateFunc) {
 	r.fsm.hookInvalidate(func(key string) {
-		_, leaderId := r.raft.LeaderWithID()
+		r.l.RLock()
+		raft := r.raft
+		r.l.RUnlock()
+
+		if raft == nil {
+			return
+		}
+		_, leaderId := raft.LeaderWithID()
+
 		if r.localID != string(leaderId) {
 			hook(key)
 		}
--- a/.github/workflows/test-go.yml
+++ b/.github/workflows/test-go.yml
@@ -310,6 +310,10 @@ jobs:
           path: test-results/go-test
           pattern: test-results-*
           merge-multiple: true
+      - name: Test Summary
+        uses: test-summary/action@v2
+        with:
+          paths: "test-results/go-test/results-*.xml"
       - env:
           EXPECTED_IDS: ${{ needs.test-matrix.outputs.matrix_ids }}
         run: |
--- a/http/logical.go
+++ b/http/logical.go
@@ -364,6 +364,18 @@ func handleLogicalRecovery(raw *vault.RawBackend, token *atomic.Value) http.Hand
 // toggles. Refer to usage on functions for possible behaviors.
 func handleLogicalInternal(core *vault.Core, injectDataIntoTopLevel bool, noForward bool) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !core.StandbyReadsEnabled() {
+			standby, err := core.Standby()
+			if err != nil {
+				respondError(w, 503, err)
+				return
+			}
+			if standby {
+				forwardRequest(core, w, r)
+				return
+			}
+		}
+
 		req, statusCode, err := buildLogicalRequest(core, w, r)
 		if err != nil || statusCode != 0 {
 			respondError(w, statusCode, err)
--- a/internalshared/configutil/config.go
+++ b/internalshared/configutil/config.go
@@ -57,6 +57,8 @@ type SharedConfig struct {
 	ClusterName string `hcl:"cluster_name"`
 
 	AdministrativeNamespacePath string `hcl:"administrative_namespace_path"`
+
+	DisableStandbyReads bool `hcl:"disable_standby_reads"`
 }
 
 func ParseConfig(d string) (*SharedConfig, error) {
--- a/vault/core.go
+++ b/vault/core.go
@@ -1076,6 +1076,10 @@ func coreInit(c *Core, conf *CoreConfig) error {
 		c.physical = physical.NewStorageEncoding(c.physical)
 	}
 
+	if cib, ok := c.underlyingPhysical.(physical.CacheInvalidationBackend); ok && !c.rawConfig.Load().(*server.Config).DisableStandbyReads {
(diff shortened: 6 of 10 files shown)


cipherboy:
@phil9909 I would not modify KV in this way. As discussed on past horizontal scalability working group calls, we want the handler to hit ErrReadOnly before forwarding. 

Also, I'd probably recommend this be based on top of #1847.

internalshared/configutil/feature_flags.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/hashicorp/hcl"
+	"github.com/hashicorp/hcl/hcl/ast"
+)
+
+// FeatureFlags is a set of toggles to enable or disable experimental features
+type FeatureFlags struct {
+	FoundKeys  []string     `hcl:",decodedFields"`
+	UnusedKeys UnusedKeyMap `hcl:",unusedKeyPositions"`
+
+	DisableStandbyReads bool `hcl:"disable_standby_reads"`
```
cipherboy: I think it is fine to add this as a top-level key in the configuration. My 2c.

phil9909:
> @phil9909 I would not modify KV in this way. As discussed on past horizontal scalability working group calls, we want the handler to hit ErrReadOnly before forwarding.

I will revert this and check why it doesn't work.

> Also, I'd probably recommend this be based on top of #1847.

Yeah, I saw the potential for conflict :laughing:, I will just rebase this one if #1847 is merged first. Should be simple.

internalshared/configutil/feature_flags.go
phil9909: I thought about this as well. IMHO it would be nice to have a dedicated section for experimental features, with an explicit disclaimer that the flag defaults might change or the flags get removed entirely. On the other hand, if we have just one or two feature flags in that section per release, it might be overkill and we can just add the disclaimer right next to the individual flags.  

Should we talk about this in the Community Meeting or do you want to take this decision as our BDFL?

cipherboy: _BDF **Election**_ ;-) 

I agree with you; it'd be nice to have a section to feature flag experimental features. But at least according to GitLab's usage (which, I feel is standardized):

1. These should be of time-boxed scope. Eventually the experiment either fails or gets rolled into the product.
2. These should default to negative (disabled) and be manually turned on. When they land in the product, they should default to enabled.

I think what we discussed in the last horizontal scalability WG meeting is that there might be a long-term usage for returning to the legacy HA node behavior. Users who want strict consistency guarantees, for one, or who have limited utilization and want simple redundancy. Thus, this flag would live indefinitely. (And, it is negated from expectations of a feature flag.)

From a feature flag perspective, if we were to want to keep this, I'd do the opposite: 

```hcl
feature_flags {
  experimental_read_scalability = true
}
```

to enable what we've built, with the intention of default enabling it in v2.6.0 and dropping the feature flag.

I think either approach is valid, but in v2.6.0, we'd still want that other _behavior_ flag, unless we disagree with our earlier assessment.

But, maybe GitLab's definition of feature flag isn't as industry standard as I think. 

If I'm a raving lunatic, tell me and we can discuss in the community call or sooner. ;-)

phil9909: In my world view, feature flags can be opt-in and opt-out. Roughly this lifecycle:

```mermaid
stateDiagram-v2
    state "Opt-In"  as OptIn
    state "Opt-Out" as OptOut
    OptIn --> OptOut: promte
    [*] --> OptIn
    [*] --> OptOut
    OptIn --> [*]: remove (failure)
    OptOut --> [*]: remove (failure or promte)
```

- a feature-flag might be added as opt-in or opt-out
- a feature-flag might be removed at any time, if the feature has "failed"
- a feature-flag might be promoted from opt-in to opt-out
- a opt-out feature-flag might be removed and simply turned on

I would start opt-in for features which can cause irreversible and severe damage (data loss), e.g. automatically switching from bbolt to something else on upgrade. Features you can simply opt-out of as soon as you notice problems may start as opt-out (e.g. a few of your clients got 500 responses, other than that no harm). 

--- 

Anyway, thanks for reminding me of that discussion we had in the working group i.e. that we want to have an opt-out switch indefinitely. This changes things, as I agree feature flags should have limited lifetime. I will turn this into a top-level setting and drop the feature flags struct for now.

cipherboy:
@phil9909 I've rebased on top of `main` one more time :-)

physical/raft/raft.go
```diff
@@ -209,7 +209,15 @@ type RaftBackend struct {
...
 	r.fsm.hookInvalidate(func(key string) {
-		_, leaderId := r.raft.LeaderWithID()
+		r.l.RLock()
+		raft := r.raft
+		r.l.RUnlock()
+
+		if raft == nil {
+			return
+		}
+		_, leaderId := raft.LeaderWithID()
```
cipherboy, line 219: @phil9909 🤔 Do we think it is safe to call `LeaderWithID()` if `raft` changes under us? Or should we hold the `r.l` lock through this entire thing? 

I don't really see any other callers of `LeaderWithID()` but for things like `State()`, they hold the lock throughout the call... 🤔

cipherboy (commented):
The only other thing I think I expected to see here was an adjustment to `http.handleLogicalInternal(...)` to always forward requests if this constant is set?

physical/raft/raft.go
phil9909: The raft object itself has it's own lock. So yes, this should be safe.

builtin/logical/kv/passthrough.go
```diff
@@ -229,7 +229,7 @@ func (b *PassthroughBackend) handleWrite() framework.OperationFunc {
 			Value: buf,
 		}
 		if err := req.Storage.Put(ctx, entry); err != nil {
-			return nil, fmt.Errorf("failed to write: %v", err)
+			return nil, fmt.Errorf("failed to write: %w", err)
```
phil9909, line 232: @cipherboy BTW: this is why forwarding didn't work

phil9909:
I guess this should be good to go now. I updated the PR description to match the changes made.

builtin/logical/kv/passthrough.go
cipherboy: @phil9909 Ahh right. We should probably do `strings.Contains(err.Error(), logical.ErrReadOnly.Error())` perhaps.

cipherboy:
@phil9909 Ah, did you see my comment above? 

> The only other thing I think I expected to see here was an adjustment to `http.handleLogicalInternal(...)` to always forward requests if this constant is set?

phil9909:
> Ah, did you see my comment above?

Saw it, but had other things to do first and forgot :laughing: sorry

Should be fixed now.

builtin/logical/kv/passthrough.go
phil9909: > @phil9909 Ahh right. We should probably do `strings.Contains(err.Error(), logical.ErrReadOnly.Error())` perhaps.

As much as I hate this: You are probably right :disappointed:

Is it fine to have this in a separate PR? Would love to get this merged.

PR #1850, opened 2025-09-26
Bump go-msgpack & raft-boltdb, workaround legacy bolt dependency

satoqz:
A potential approach to get over https://github.com/openbao/openbao/pull/920#issuecomment-2611549791 and https://github.com/openbao/openbao/pull/1832#issuecomment-3318080373, and avoid falling behind upstream indefinitely without compromising platform support.

Open to dropping this if we think it's too cursed, personally I find it acceptable because the stub is quite small.


helper/stubbolt/stubbolt.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+// Package bolt holds stubs for github.com/boltdb/bolt functionality required
+// by github.com/hashicorp/raft-boltdb/v2. Bolt is unmaintained and a dependency
+// on it would reduce OpenBao's platform support (namely, we would have to
+// drop riscv64). raft-boltdb depends on the old bolt package only to provide a
+// function (MigrateToV2) to migrate to bbolt, which OpenBao does not use.
```
satoqz, line 8: This comment is the spicy bit.

go.mod
```diff
@@ -207,6 +208,7 @@ require (
 	github.com/aws/aws-sdk-go v1.55.6 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/bgentry/speakeasy v0.1.0 // indirect
+	github.com/boltdb/bolt v1.3.1 // indirect
```
satoqz, line 211: While we can't avoid this line, any code pulled in here is never used due to the `replace` directive above.

go.sum
```diff
@@ -436,9 +436,8 @@ github.com/hashicorp/go-memdb v1.3.5 h1:b3taDMxCBCBVgyRrS1AZVHO14ubMYZB++QpNhBg+
 github.com/hashicorp/go-memdb v1.3.5/go.mod h1:8IVKKBkVe+fxFgdFOYxzQQNjz+sWCyHCdIC/+5+Vy1Y=
 github.com/hashicorp/go-metrics v0.5.4 h1:8mmPiIJkTPPEbAiV97IxdAGNdRdaWwVap1BU6elejKY=
 github.com/hashicorp/go-metrics v0.5.4/go.mod h1:CG5yz4NZ/AI/aQt9Ucm/vdBnbh7fvmv4lxZ350i+QQI=
+github.com/hashicorp/go-msgpack v0.5.5 h1:i9R9JSrqIz0QVLz3sz+i3YJdT7TTSLcfLLzJi9aZTuI=
 github.com/hashicorp/go-msgpack v0.5.5/go.mod h1:ahLV/dePpqEmjfWmKiqvPkv/twdG7iPBM1vqhUKIvfM=
-github.com/hashicorp/go-msgpack v1.1.5 h1:9byZdVjKTe5mce63pRVNP1L7UAmdHOTEMGehn6KvJWs=
-github.com/hashicorp/go-msgpack v1.1.5/go.mod h1:gWVc3sv/wbDmR3rQsj1CAktEZzoz1YNK9NfGLXJ69/4=
 github.com/hashicorp/go-msgpack/v2 v2.1.2 h1:4Ee8FTp834e+ewB71RDrQ0VKpyFdrKOjvYtnQ/ltVj0=
 github.com/hashicorp/go-msgpack/v2 v2.1.2/go.mod h1:upybraOAblm4S7rx0+jeNy+CWWhzywQsSRV5033mMu4=
```
satoqz, line 442: v0.5.5 is only pulled in because it is used in https://github.com/hashicorp/raft-snapshot 's tests. This makes v2.1.2 the only version that ever makes its way into our binary.

physical/raft/raft.go
```diff
@@ -420,8 +420,9 @@ func NewRaftBackend(conf map[string]string, logger log.Logger) (physical.Backend
 		dbPath := filepath.Join(path, "raft.db")
 		opts := boltOptions(dbPath)
 		raftOptions := raftboltdb.Options{
-			Path:        dbPath,
-			BoltOptions: opts,
+			Path:                    dbPath,
+			BoltOptions:             opts,
+			MsgpackUseNewTimeFormat: true,
```
satoqz, line 425: Since we were on go-msgpack v1.1.5 before, this _should_ be the correct option to set. go-msgpack v0.x uses the old time format, v1.x uses the new time format, v2.x requires that we explicitly opt into the new time format.

PR #1851, opened 2025-09-26
website: fix warning/note/danger indentations

phil9909:
Fix indentation of `:::warning`, `:::info` and `:::danger` blocks in lists.

We have many places on our website were we do something like this:

```mdx
- `parameter_a` ...
- `parameter_b` ...

:::warning

`parameter_b` has the limitation ...

:::

- `parameter_c` ...
```

In the generated HTML this will be two unordered lists (`parameter_a`, `parameter_b`) (`parameter_c`) interrupted by a block for the warning.

If we instead indent the warning block

```mdx
- `parameter_a` ...
- `parameter_b` ...

  :::warning

  `parameter_b` has the limitation ...

  :::

- `parameter_c` ...
```

We will get one list and the warning block will be part of the second list item, which in most cases is semantically correct.

Screenshots:

Old:
image

New:

image

Source Page: https://openbao.org/docs/secrets/transit/

This has several advantages:

- It looks better (IMHO)
- It is better for accessibility as the HTML is semantically more correct. 
- For ordered lists (as seen in the screenshot) the numbering of the points will be reset to `1.` if the list is interrupted.

This PR fixes a few places with a similar problem with code blocks <code>```</code>

### DIFF
--- a/website/content/docs/agent-and-proxy/proxy/index.mdx
+++ b/website/content/docs/agent-and-proxy/proxy/index.mdx
@@ -112,12 +112,14 @@ These are the currently-available general configuration options:
 
 - `listener` <code>([listener][listener]: \<optional>)</code> - Specifies the addresses and ports on which the Proxy will respond to requests.
 
-:::warning
+  :::warning
 
-**Note:** On `SIGHUP` (`kill -SIGHUP $(pidof bao)`), OpenBao Proxy will attempt to reload listener TLS configuration.
-This method can be used to refresh certificates used by OpenBao Proxy without having to restart its process.
+  **Note:** On `SIGHUP` (`kill -SIGHUP $(pidof bao)`), OpenBao Proxy will
+  attempt to reload listener TLS configuration. This method can be used to
+  refresh certificates used by OpenBao Proxy without having to restart its
+  process.
 
-:::
+  :::
 
 - `pid_file` `(string: "")` - Path to the file in which the Proxy's Process ID
 (PID) should be stored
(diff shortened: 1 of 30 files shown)


website/content/docs/agent-and-proxy/proxy/index.mdx
```diff
@@ -112,12 +112,14 @@ These are the currently-available general configuration options:
 
 - `listener` <code>([listener][listener]: \<optional>)</code> - Specifies the addresses and ports on which the Proxy will respond to requests.
 
-:::warning
+  :::warning
 
-**Note:** On `SIGHUP` (`kill -SIGHUP $(pidof bao)`), OpenBao Proxy will attempt to reload listener TLS configuration.
-This method can be used to refresh certificates used by OpenBao Proxy without having to restart its process.
+  **Note:** On `SIGHUP` (`kill -SIGHUP $(pidof bao)`), OpenBao Proxy will
```
satoqz, line 117: I feel like an additional `**Note:**` prefix inside of a `:::warning` block is an antipattern. Just an observation.

website/content/docs/secrets/transit/index.mdx
```diff
@@ -185,14 +185,14 @@ the proper permission, it can use this secrets engine.
...
-    configuration.
+     OpenBao HTTP API imposes a maximum request size of 32MB to prevent a denial
+     of service attack. This can be tuned per [`listener`
+     block](/docs/configuration/listener/tcp) in the OpenBao server
+     configuration.
 
-:::
+     :::
 
 1.  Decrypt a piece of data using the `/decrypt` endpoint with a named key:
```
satoqz: I think this is one space too many?

satoqz (approved):
Looks good. Thanks for rewrapping some of the paragraphs too, big fan.

driif (approved):

website/content/docs/secrets/transit/index.mdx
phil9909: It renders correctly, but you are right, should be one less. Fixed.

PR #1853, opened 2025-09-26
Support custom metrics paths

cipherboy:
In certain environments, utilities scraping metrics only respond on standardized paths. In the OpenTelemetry world, `/metrics` is one such common path. When we're on a metrics-only endpoint, allow rewriting the `/v1/sys/metrics` path to a standardized variant to help operators in these environments.

See also: https://docs.runway.gitlab.com/reference/observability/#cloud-run
See also: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config

---

This works both with and without authentication so I didn't think it needed to be called out separately.

fyi: @phil9909 @kyounghunJang


internalshared/configutil/listener.go
```diff
@@ -29,6 +29,7 @@ type ListenerTelemetry struct {
 	DisallowMetricsRaw              interface{}  `hcl:"disallow_metrics,alias:DisallowMetrics"`
 	MetricsOnly                     bool         `hcl:"-"`
 	MetricsOnlyRaw                  interface{}  `hcl:"metrics_only,alias:MetricsOnly"`
+	MetricsPath                     string       `hcl:"metrics_path,alias:MetricsPath"`
```
cipherboy, line 32: @phil9909 In the context of OTEL paths, we'll probably add a `otel_metrics_path` and then later merge the two once otel takes over.

website/content/docs/configuration/listener/unix.mdx
```diff
@@ -32,6 +32,13 @@ multiple sockets.
   `/sys/rekey-recovery-key/*`). These are a security risk to leave exposed on
   public listeners.
 
+   :::warning
+
+   **In OpenBao v2.4.0, this parameter will default to true, forbidding any calls
+to the unauthenticated rekey endpoints. This will be a breaking change.
```
satoqz: Either

```suggestion
   In OpenBao v2.4.0, this parameter will default to true, forbidding any calls
   to the unauthenticated rekey endpoints. This will be a breaking change.
```

or

```suggestion
   **In OpenBao v2.4.0, this parameter will default to true, forbidding any calls
   to the unauthenticated rekey endpoints. This will be a breaking change.**
```

http/metrics.go
```diff
@@ -21,21 +21,39 @@ func wrapMetricsListenerHandler(handler http.Handler, props *vault.HandlerProper
 		}
 
 		listenerConfig := props.ListenerConfig
-		isMetricsPath := r.URL.Path == "/v1/sys/metrics"
+
+		isMetricsOnly := listenerConfig.Telemetry.MetricsOnly
+		disallowsMetrics := listenerConfig.Telemetry.DisallowMetrics
+
+		realMetricsPath := "/v1/sys/metrics"
+		metricsPath := "/v1/sys/metrics"
```
satoqz:
```suggestion
		metricsPath := realMetricsPath
```

satoqz (approved):
Just two nits

PR #1854, opened 2025-09-28
Improve ergonomics of CEL helpers

cipherboy:
Calling into CEL should largely be standardized: CEL programs, as viewed by OpenBao have two main components:

1. Some environment provided by the plugin, and
2. Dynamic execution by the user.

Notably the two are not mixed: the plugin's environment is fully defined and initialized before the user's code executes.

From a CEL execution environment perspective, the user cannot influence the environment except through defining custom variables. While the plugin should have full control over the CEL environment, in reality, there's a few common options all CEL programs will likely comply with:

1. Whether to define a custom container,
2. Whether to extend the standard library,
3. Whether to add email and other OpenBao-specific helpers,
4. Whether to inject identity helpers in, and
5. Any custom definitions and types this helper might provide.

From an API perspective, OpenBao plugins are concerned about two operations:

1. Type checking, to give reasonable assurance to the role administrator that their CEL program will execute correctly, and
2. Execution at evaluation time.

All other specifics of evaluation can be hidden from plugin authors, limiting their exposure to CEL library internals.

In addressing this, we've cleaned up several behaviors:

1. JWT will now properly type check the role.
2. Consistent evaluation ordering now occurs between type checking and full program evaluation, with variables exactly immediately preceding the evaluation of the final expression.
3. Variables are now added incrementally in the order they are defined, preventing a potential circular reference.
4. Standardized parsing and framework schema definitions of the program field.

Type checking is a point of contention: we frequently use ternary conditional operators in a context where it can return one of two different types rather than one of two objects of the same type as implied by the definition in the CEL Go library. This required a workaround to supplement the definition of the ternary conditional operator, though parsing remains unaffected.

See also: https://github.com/google/cel-go/issues/1221

fyi: @fatima2003 @suprjinx

### DIFF
--- a/changelog/1854.txt
+++ b/changelog/1854.txt
@@ -1,0 +1,9 @@
+```release-note:improvement
+auth/jwt: Add type checking to role
+```
+```release-note:bug
+auth/jwt: Fix ordering of variable declarations in CEL program roles
+```
+```release-note:bug
+secrets/pki: Fix ordering of variable declarations in CEL program roles
+```


builtin/credential/jwt/path_cel_login_test.go
```diff
@@ -28,7 +28,7 @@ func Test_runCelProgram(t *testing.T) {
 		{
 			name: "Boolean false will return error",
 			celRole: celRoleEntry{
-				CelProgram: celhelper.CelProgram{
+				Program: &celhelper.Program{
```
suprjinx, line 31: good rename +1

builtin/credential/jwt/path_cel_role.go
```diff
@@ -399,72 +390,19 @@ func (b *jwtAuthBackend) getCelRole(ctx context.Context, s logical.Storage, role
...
-	}
-
-	// Add "pb.Auth" return type to environment
-	envOptions = append(envOptions, cel.Types(&pb.Auth{}))
-	env, err := cel.NewEnv(envOptions...)
-	if err != nil {
-		return nil, err
+func (b *jwtAuthBackend) celEvalConfig() *celhelper.EvalConfig {
+	return &celhelper.EvalConfig{
+		WithExtLib: true,
```
suprjinx, line 379: nice

suprjinx:
excellent refactor -- do you think existing tests are covering all of the changed code?

cipherboy:
Thanks @suprjinx!

Yes, I think we've largely just refactored the existing integration and not added or removed features. The only surprise was really the change w.r.t. ternary expression over CEL, but the workaround I think is fine based on what I know of the CEL internals.

cipherboy:
@openbao/openbao-org-maintainers I think this will require review from you all due to changes to PKI. Thanks!

DrDaveD (commented):
The jwt part looks fine to me.  I didn't check the pki part.

builtin/logical/pki/path_cel.go
```diff
@@ -18,7 +17,7 @@ type CELRoleEntry struct {
 	// Required, the name of the role
 	Name string `json:"name"`
 	// Required, defines validation logic
-	CelProgram celhelper.CelProgram `json:"validation_program"`
+	Program celhelper.Program `json:"validation_program"`
 }
```
wslabosz-reply, line 21: is this struct even marshaled at some point, so that it makes for the struct tag to appear?

wslabosz-reply: EDIT: It is persisted to the storage using that name, but the field (name) throughout its lifecycle is never used as it's always returned as `cel_program`

cipherboy: Ah, a quirk of storage.

Yeah, in general we've opted to separate out storage representations from API representations, which leads to quirks like this. Probably more work than it is worth at this point, moving this to `cel_program`.

wslabosz-reply: I was just confused overall, I think it's fine, with maybe additional comment saying that that field is never used in the outbound responses, changing it to `cel_program` is also fine!

cipherboy:
@openbao/openbao-org-maintainers This will require a review as we don't have codeowners for PKI.

cipherboy:
@phyrog Do you mind taking a look at this? I ended up rejecting most of your suggestions from the earlier deprecations PR and keeping what was in this one. Thanks!

cipherboy:
Thanks all! Merging so this doesn't conflict again. :-D

PR #1861, opened 2025-09-29
Delay updates for dependencies to counter supply chain attacks

driif:
I propose to follow the rule: **don’t touch it if it works**.

Minor and major updates often include refactors, API changes, and new features. We rely on our dependencies for stable functions, so we prefer only bug fixes and critical security updates.

This PR updates Dependabot to block minor and major updates and allow only patch updates that are at least 7 days old. Per the docs, critical security updates will still come through.

Since the website is not a critical part of our app, I configured rules to allow minor and major updates there, but only after a set minimum age.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,12 +5,11 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-    ignore:
-      - dependency-name: "*"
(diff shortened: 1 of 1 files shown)


cipherboy:
Thanks @driif, I've filed something for this week's supply chain security and Dev WG meetings. I'll see if I can make it.

cipherboy:
@driif If you want to update the title, I'll merge :-)

PR #1871, opened 2025-09-30
Add markdownlint to website and fix some broken links

phil9909:
We have some broken internal links on the website, therefore I figured it would be good to have a linter against this.

I tried a few eslint integrations, but nothing really worked, in the end `markdownlint` worked perfectly.

So this PR:
- Adds `markdownlint` to `package.json`, `Makefile` and GitHub Actions with many checks disabled for now (see `website/.markdownlint-cli2.jsonc`)
- Fixes all broken internal links
  - some were dropped (e.g. features that were removed when forking)
  - many just needed simple fixes like missing `s` or to many `-` (Docusaurus will turn `## My Sub-Heading` into an anchor `my-subheading` and not `my-sub-heading`, which I guess the tool that HashiCorp used did)
  - Some content got moved around, etc.
- Adds the [Concepts > CEL page](https://openbao.org/docs/concepts/cel/) to the sidebar.

### DIFF
--- a/website/.markdownlint-cli2.jsonc
+++ b/website/.markdownlint-cli2.jsonc
@@ -28,7 +28,7 @@
     "commands-show-output": false,
     "no-trailing-punctuation": false,
 
-    // "Parmanent" config
+    // "Permanent" config
     "first-line-h1": false, // to support partials
     "no-inline-html": false, // this is fine with mdx/Docusaurus
     "no-bare-urls": false, // some markdown parsers don't support this, our does so ignore it


website/.markdownlint-cli2.jsonc
```diff
@@ -0,0 +1,39 @@
...
+    "list-indent": false,
+    "single-trailing-newline": false,
+    "single-h1": false,
+    "no-emphasis-as-heading": false,
+    "descriptive-link-text": false,
+    "emphasis-style": false,
+    "commands-show-output": false,
+    "no-trailing-punctuation": false,
+
+    // "Parmanent" config
```
wslabosz-reply: (don't apply my suggestion through gh UI)
```suggestion
    // "Permanent" config
```

PR #1872, opened 2025-09-30
KV engine returns timestamp strings in RFC3339Nano format

### LINKED ISSUE #1841: iso8601 Dates in KeyValue engine
**Describe the bug**

The date format in the key/Value engine is different than Vault and other engines in OpenBao: 
- OpenBao: `iso8601`. Example: `created_time = 2025-09-22T18:09:34Z`
- Vault: uses `iso8601` with fractional seconds Example: `created_time = 2025-09-22T18:38:02.637633225Z`

Is this intended? I think fractional seconds has more precision. I wonder why is this different here?

The documentation of write secret in the keyValue secret engine shows what we get from Vault and not from OpenBao. On the other hand,

https://openbao.org/api-docs/secret/databases/#get-static-credentials gives fractional seconds on `last_vault_rotation`.

**To Reproduce**

```shell-session
$ curl \
    --request POST \
    --data @payload.json \
    https://127.0.0.1:8200/v1/secret/data/my-secret
```

Gives 

```json
{
  "data": {
    "created_time": "2025-09-22T18:09:34Z",
    "custom_metadata": {
      "owner": "jdoe",
      "mission_critical": "false"
    },
    "deletion_time": "",
    "destroyed": false,
    "version": 1
  }
}
```

**Expected behavior**

created_time has fractional seconds.

**Environment:**
```
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    1
Threshold       1
Version         2.4.0
Build Date      2025-08-28T20:33:02Z
Storage Type    inmem
Cluster Name    vault-cluster-25671780
Cluster ID      07827538-3e52-bf3b-9cd0-ef453be8bbe4
HA Enabled      false
```

**Additional context**

wslabosz-reply:
Resolves #1841

`ptypesTimestampToString` now returns back the timestamp string in `RFC3339Nano` format, as it was done, (but wrongly indicated [here](https://github.com/golang/protobuf/blob/75de7c059e36b64f01d0dd234ff2fff404ec3374/ptypes/timestamp.go#L76-L86)) by the previous version of this code fragment.

### DIFF
--- a/builtin/logical/kv/backend.go
+++ b/builtin/logical/kv/backend.go
@@ -439,7 +439,7 @@ func ptypesTimestampToString(t *timestamppb.Timestamp) string {
 		return ""
 	}
 
-	return t.AsTime().Format(time.RFC3339)
+	return t.AsTime().Format(time.RFC3339Nano)
 }
 
 var backendHelp string = `
--- a/changelog/1872.txt
+++ b/changelog/1872.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+secrets/kv: KV entries timestamps are now correctly formatted in `RFC3339Nano`, as previously done so.
+```


wslabosz-reply:
@cipherboy 
fyi: I also see some timestamps in identity subsystem using the `AsTime().Format(time.RFC3339)`

edit:
looked at our [docs](https://openbao.org/api-docs/secret/identity/entity/#sample-response-1) which indicate these should also be in `RFC3339Nano` format, let me adjust

PR #1875, opened 2025-09-30
Default to disable_unauthed_rekey_endpoints=true

cipherboy:
As discussed with the new authenticated rotation endpoints available for a full release, now we can complete the deprecation and default to disabling the unsafe legacy rekey endpoints.


wslabosz-reply:
@cipherboy
I know it's in draft, but just a reminder that we need to update the `website/content/docs/deprecation/index.mdx` with this one.

wslabosz-reply:
closed, new PR [here](https://github.com/openbao/openbao/pull/2125)

PR #1877, opened 2025-10-01
Fix UI tests following #1732

satoqz:
UI tests started failing on main following #1732, see https://github.com/openbao/openbao/commits/main. This makes it harder to spot legit CI failures because, well, green check vs red cross.

The issue seems to be incorrect formatting only, resolved via `yarn fmt:js`.

### DIFF
--- a/ui/app/models/oidc/provider.js
+++ b/ui/app/models/oidc/provider.js
@@ -46,7 +46,13 @@ export default class OidcProviderModel extends Model {
   _attributeMeta = null; // cache initial result of expandAttributeMeta in getter and return
   get formFields() {
     if (!this._attributeMeta) {
-      this._attributeMeta = expandAttributeMeta(this, ['name', 'issuer', 'scopesSupported','authorizationCode','clientCredentials']);
+      this._attributeMeta = expandAttributeMeta(this, [
+        'name',
+        'issuer',
+        'scopesSupported',
+        'authorizationCode',
+        'clientCredentials',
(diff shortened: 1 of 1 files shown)


cipherboy (approved):
D'oh. Thanks @satoqz!

satoqz:
@cipherboy tbh, while this resolves the "deterministically errors due to formatting" issue, it seems we're stuck flaking out on other (varying) tests more often than not (see CI history, this PR is at run number 4 right now, and I can't get a pass). IMO we should consider fully disabling UI tests for merges to main.

cipherboy:
@satoqz Yeah, especially as we are in the swing of a rewrite of the UI in React. Eventually that'll take over and we won't need it. I'm fine with that now I suppose, though I still like knowing if the failure is legitimate or not.

I've found unless you put `ui` in the branch name (and few contributors do, as we saw here), UI tests won't run on PRs... Maybe that's something else we should fix while removing `main` execution. My 2c.

wslabosz-reply (approved):

satoqz:
@cipherboy the `ui` label also suffices as per

https://github.com/openbao/openbao/blob/ed043f8c1a05c3d8608cbd3e586724fc622243a3/.github/workflows/ci.yml#L126

and as I've done with this PR 😄

satoqz:
Will merge this as-is, let's think about locking UI tests down more if the flakes get too annoying.

Issue #1864, opened 2025-09-29
Drop github.com/klauspost/compress


cipherboy:
Static GZIP compression should be easily achievable with a `sync.Once` and a simple handler to detect the gzip accept-encoding type.

This is only used in the HTTP handlers. Arguably, `robots.txt` shouldn't need to be compressed, either, and perhaps the UI build should come pre-GZIP'd to save binary size. This would require we store it in-memory decompressed as well as compressed, but this is the current state of things.

kyounghoonJang:
@cipherboy 
Hi, what do you think about using a` sync.RWMutex` for this issue?  
I think it would give us a bit more flexibility compared to compressing everything once up front.

cipherboy:
@kyounghunJang Hmmm.... I'm not sure the flexibility is useful in this particular case?

1. These are static assets compiled into the server. Compressing (or decompressing) would happen once anyways. 
2. A sufficient number of clients are liable to fetch all assets, so there's no point in granular locking of the compression routine.

Maybe I'm missing something? :-)

kyounghoonJang:
@cipherboy 

I was considering the case where, with only a small number of users, some assets might go unused for a long time. But if in the end all assets are going to be fetched anyway, then compressing them once does make more sense! Would it be okay if I take on this issue?

Thanks!

cipherboy:
For anyone interested in picking this up, there should be a very simple approach to do this during the build process: https://github.com/openbao/openbao/pull/1880#issuecomment-3944009185

satoqz:
Issue should stay open no?

PR #1880, opened 2025-10-02
refactor: Drop github.com/klauspost/compress

kyounghunJang:
Replace runtime gzip compression (klauspost/compress) with startup-time pre-compression and in-memory caching for UI assets.

- Add gzip.go for in-memory asset caching and compression
- Serve pre-compressed or raw based on Accept-Encoding
- Remove dependency on github.com/klauspost/compress

Fixes: #1864

### DIFF
--- a/http/gzip.go
+++ b/http/gzip.go
@@ -1,0 +1,186 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package http
+
+import (
+	"bytes"
+	"compress/gzip"
+	"io"
+	"mime"
+	"net/http"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/openbao/openbao/vault"
+)
+
+type uiAsset struct {
+	raw []byte
+	gz  []byte
+	typ string
+}
+
+var (
+	uiAssets     = map[string]uiAsset{}
+	uiAssetsOnce sync.Once
+)
+
+// compressibleTypes defines MIME types that should be compressed
+var compressibleTypes = map[string]struct{}{
+	"text/html":              {},
+	"text/css":               {},
+	"text/plain":             {},
+	"text/xml":               {},
+	"application/javascript": {},
+	"application/json":       {},
+	"application/xml":        {},
+	"application/xhtml+xml":  {},
+	"image/svg+xml":          {},
+}
+
+// isCompressible checks if the MIME type should be compressed
+func isCompressible(mimeType string) bool {
+	// Remove charset and other parameters from MIME type
+	if idx := strings.Index(mimeType, ";"); idx != -1 {
+		mimeType = mimeType[:idx]
+	}
+	mimeType = strings.TrimSpace(mimeType)
+	_, ok := compressibleTypes[mimeType]
+	return ok
+}
+
+func loadUIAsset(core *vault.Core, fs http.FileSystem, name string) (uiAsset, error) {
+	f, err := fs.Open(name)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


driif:
@kyounghunJang Thank you for continuing to clean up and remove dependencies. 

I don't think, we need to store both raw and compressed files in memory, since all browsers support gzip. 

What do you think about over-optimising this solution: compress only the following MIME types and store only them in mem:

- text/html
- text/css
- text/plain
- text/xml
- application/javascript
- application/json
- application/xml
- image/svg+xml

If we support compression, return the compressed version, otherwise - return from fs. For example, PNG files can come straight from fs. 

---

PS: What if we drop compression? We currently save around 6 MB of network traffic, but browsers cache aggressibely nowerdays. Is it still necessary to keep compression?

cipherboy:
@driif If browser support is so good... I was actually wondering the opposite: if we wanted to optimize heavily, we should compress the assets during the web UI build processes to minimize file size. What do you think?

Then the embedding of the UI in the binary minimizes overhead.

kyounghoonJang:
@cipherboy @driif 

I think compression is necessary for optimization, even if it’s only about 6MB.
I’ve also confirmed that we should only compress the MIME types you mentioned and that they are well-supported by browsers. It seems we don’t need to keep the uncompressed files in memory!

Thank you for your feedback!!

driif:
@cipherboy @kyounghunJang 

> @driif If browser support is so good... I was actually wondering the opposite: if we wanted to optimize heavily, we should compress the assets during the web UI build processes to minimize file size. What do you think?
> 
> Then the embedding of the UI in the binary minimizes overhead.

I like compressing assets at build time and serving only compressed static files. If a client doesn’t support gzip, it’s probably very old, so we don’t need that extra complexity in the code. Let’s keep it simple and only revisit if we run into an issue.

kyounghoonJang:
@driif 

I’ve updated the code as requested: only MIME types are compressed and stored in memory, while the rest remain uncompressed.

 Thank you!

driif:
Hey @kyounghunJang, thanks for the update, but I think implementing the approach @cipherboy proposed will be the best solution here.

Let’s pause this until we discuss it in the community or the dev WG sync and make a final decision.

satoqz:
TBH I'm in favor of merging in this PR with @kyounghunJang's current approach and perhaps reconsidering the origin of compression later on, but this seems like a net improvement overall if it lets us drop the dependency?

http/gzip.go
```diff
@@ -0,0 +1,186 @@
+// Copyright (c) HashiCorp, Inc.
```
satoqz, line 1:
```suggestion
// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
```

Assuming this code wasn't copied from HashiCorp in any way?

http/handler.go
```diff
@@ -27,7 +27,6 @@ import (
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
 	"github.com/hashicorp/go-sockaddr"
 	"github.com/hashicorp/go-uuid"
-	gziphandler "github.com/klauspost/compress/gzhttp"
```
satoqz, line 30: Can we drop this from go.mod, too?

driif:
@satoqz 

Dropping a library is good, and it would be great to reduce the code too. The proposed solution adds unnecessary complexity.

Like @cipherboy replied, we can build the gzip file during the web UI build process, and then we’d only need to change few lines of code to serve gzip.

cipherboy:
Hey @kyounghunJang -- feel free to reopen if you're interested in picking up the revised approach :-) Thank you!

PR #1882, opened 2025-10-03
Fix: Use JobManager instead of github.com/gammazero/workerpool

### LINKED ISSUE #1863: Use JobManager instead of github.com/gammazero/workerpool
github.com/gammazero/workerpool is used for the rollback manager, limiting the number of parallel jobs that we take on. However, we already have a built-in alternative, JobManager, which we could use instead (and is used in ExpirationManager).

We should thus drop workerpool.

vr-varad:
Resolves #1863

### DIFF
--- a/vault/rollback.go
+++ b/vault/rollback.go
@@ -87,7 +87,6 @@ type rollbackJob struct {
 
 func (r *rollbackJob) Execute() error {
 	if err := r.manager.attemptRollback(r.ctx, r.fullPath, r.rs, r.grabStatelock); err != nil {
-		r.manager.logger.Error("rollback job failed", "path", r.fullPath, "error", err)
 		return err
 	}
 	select {


vr-varad:
I have a few doubts
1. Should I remove the runner from the struct?
2. How should I use the job runner here?
https://github.com/openbao/openbao/blob/5cba43284c39322390d52e84fc299f6166e1a70e/vault/rollback.go#L225

// @cipherboy

cipherboy:
@vr-varad Yes, definitely remove the item from the struct; we want to remove the dependency on the package. 

I'd probably call `AddJob` to add the job: https://github.com/openbao/openbao/blob/5cba43284c39322390d52e84fc299f6166e1a70e/helper/fairshare/jobmanager.go#L104-L127

vr-varad:
@cipherboy Kindly review when convenient.

vr-varad:
@cipherboy Kindly review when convenient.

cipherboy:
HEy @vr-varad -- it looks like two CI passes are failing; I think you need to run `make tidy-all` locally.

vault/rollback.go
```diff
@@ -195,7 +218,7 @@ func (m *RollbackManager) triggerRollbacks() {
 func (m *RollbackManager) startOrLookupRollback(ctx context.Context, fullPath string, grabStatelock bool) *rollbackState {
 	m.inflightLock.Lock()
 	defer m.inflightLock.Unlock()
-	defer metrics.SetGauge([]string{"rollback", "queued"}, float32(m.runner.WaitingQueueSize()))
+	defer metrics.SetGauge([]string{"rollback", "queued"}, float32(len(m.jobManager.GetWorkQueueLengths())))
```
cipherboy: Don't we want to call:


```suggestion
	defer metrics.SetGauge([]string{"rollback", "queued"}, float32(m.jobManager.GetPendingJobCount()))
```

perhaps? `len` seems wrong, given the current call returns a map of `worker->number of job`, we would want a `sum(...)` operation, which this does.

vr-varad:
running `make tidy-all` makes chenges in `go.mod` and `go.sum` and I don't think we want that. // @cipherboy

wslabosz-reply:
@vr-varad 
we definitely want `make tidy-all` changing `go.mod`, as the issue you are working on is about removal of `github.com/gammazero/workerpool` dependency, so at least that one would disappear from the `go.(mod|sum)`

vr-varad:
Made the changes @cipherboy @wslabosz-reply

vr-varad:
Made the changes, @satoqz actually the err was not handling inside one of the functions so fixed it.

vault/rollback.go
```diff
@@ -76,6 +76,31 @@ type rollbackState struct {
...
+	ctx           context.Context
+	fullPath      string
+	rs            *rollbackState
+	grabStatelock bool
+	manager       *RollbackManager
+}
+
+func (r *rollbackJob) Execute() error {
+	if err := r.manager.attemptRollback(r.ctx, r.fullPath, r.rs, r.grabStatelock); err != nil {
+		r.manager.logger.Error("rollback job failed", "path", r.fullPath, "error", err)
```
satoqz:
```suggestion
```

IIUC this error will automatically flow to 

```go
func (r *rollbackJob) OnFailure(err error) {
```

vr-varad: yes u are right, I logged there only, this could be another approach.

vr-varad: thanks @satoqz for help

vault/rollback.go
```diff
@@ -222,13 +247,14 @@ func (m *RollbackManager) startOrLookupRollback(ctx context.Context, fullPath st
...
-			}
-		})
+		job := &rollbackJob{
+			ctx:           ctx,
+			fullPath:      fullPath,
+			rs:            rs,
+			grabStatelock: grabStatelock,
+			manager:       m,
+		}
+		m.jobManager.AddJob(job, fullPath)
```
cipherboy, line 257: @satoqz @wslabosz-reply We'll have to keep an eye on this later. I think what this does is places each rollback job in a separate queue; we might want to have one per-namespace or similar, to avoid starving other namespaces when a single namespace has a lot of mounts.

But we can deal with that later.

cipherboy (approved):
Thanks @vr-varad! Appreciate your patience on this :-)

vr-varad:
thanks @vr-varad !! would love to contribute more!!

PR #1883, opened 2025-10-03
Namespace seal/unseal CLI commands

wslabosz-reply:
This PR introduces the `bao namespace seal <ns>` and `bao namespace unseal <ns>` CLI commands

Changes:
- added `api/sys_namespace.go` file with api operation code to create a net new namespaces, generate root tokens for namespaces and seal/unseal (or retrieve seal status)
- tweaked `command/namespace_generate_root.go` file, and added unit tests
- added `command/namespace_seal.go`, `command/namespace_unseal.go` and its corresponding tests

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,10 +3,6 @@ linters:
   enable:
     - nilnil
     - misspell
-  settings:
-    errcheck:
-      exclude-functions:
-        - (io.ReadCloser).Close
 formatters:
   enable:
     - gofumpt
--- a/api/sys_generate_root.go
+++ b/api/sys_generate_root.go
@@ -9,15 +9,16 @@ import (
 )
 
 type GenerateRootStatusResponse struct {
-	Nonce          string `json:"nonce"`
-	Started        bool   `json:"started"`
-	Progress       int    `json:"progress"`
-	Required       int    `json:"required"`
-	Complete       bool   `json:"complete"`
-	EncodedToken   string `json:"encoded_token"`
-	PGPFingerprint string `json:"pgp_fingerprint"`
-	OTP            string `json:"otp"`
-	OTPLength      int    `json:"otp_length"`
+	Nonce            string `json:"nonce"`
+	Started          bool   `json:"started"`
+	Progress         int    `json:"progress"`
+	Required         int    `json:"required"`
+	Complete         bool   `json:"complete"`
+	EncodedToken     string `json:"encoded_token"`
+	EncodedRootToken string `json:"encoded_root_token"`
+	PGPFingerprint   string `json:"pgp_fingerprint"`
+	OTP              string `json:"otp"`
+	OTPLength        int    `json:"otp_length"`
 }
 
 func (c *Sys) GenerateRootStatus() (*GenerateRootStatusResponse, error) {
--- a/sdk/helper/roottoken/decode.go
+++ b/sdk/helper/roottoken/decode.go
@@ -6,21 +6,37 @@ package roottoken
 import (
 	"encoding/base64"
 	"fmt"
+	"strings"
 
+	"github.com/hashicorp/go-uuid"
 	"github.com/openbao/openbao/sdk/v2/helper/xor"
 )
 
 // DecodeToken will decode the root token returned by the Vault API
 // The algorithm was initially used in the generate root command
 func DecodeToken(encoded, otp string, otpLength int) (string, error) {
-	tokenBytes, err := base64.RawStdEncoding.DecodeString(encoded)
-	if err != nil {
-		return "", fmt.Errorf("error decoding base64'd token: %v", err)
-	}
+	switch otpLength {
+	case 0:
+		// Backwards compat
+		tokenBytes, err := xor.XORBase64(encoded, otp)
+		if err != nil {
+			return "", fmt.Errorf("error xoring token: %s", err)
+		}
 
-	tokenBytes, err = xor.XORBytes(tokenBytes, []byte(otp))
-	if err != nil {
-		return "", fmt.Errorf("error xoring token: %v", err)
+		uuidToken, err := uuid.FormatUUID(tokenBytes)
+		if err != nil {
+			return "", fmt.Errorf("error formatting base64 token value: %s", err)
+		}
+		return strings.TrimSpace(uuidToken), nil
+	default:
+		tokenBytes, err := base64.RawStdEncoding.DecodeString(encoded)
+		if err != nil {
+			return "", fmt.Errorf("error decoding base64'd token: %v", err)
+		}
+		tokenBytes, err = xor.XORBytes(tokenBytes, []byte(otp))
+		if err != nil {
+			return "", fmt.Errorf("error xoring token: %v", err)
+		}
+		return string(tokenBytes), nil
 	}
-	return string(tokenBytes), nil
 }
--- a/sdk/helper/roottoken/otp.go
+++ b/sdk/helper/roottoken/otp.go
@@ -1,0 +1,43 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package roottoken
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+
+	"github.com/hashicorp/go-secure-stdlib/base62"
+)
+
+// DefaultBase64EncodedOTPLength is the number of characters that will be randomly generated
+// before the Base64 encoding process takes place.
+const defaultBase64EncodedOTPLength = 16
+
+// GenerateOTP generates a random token and encodes it as a Base64 or as a Base62 encoded string.
+// Returns 0 if the generation completed without any error, 2 otherwise, along with the error.
+func GenerateOTP(otpLength int) (string, error) {
+	switch otpLength {
+	case 0:
+		// This is the fallback case
+		buf := make([]byte, defaultBase64EncodedOTPLength)
+		readLen, err := rand.Read(buf)
+		if err != nil {
+			return "", fmt.Errorf("error reading random bytes: %s", err)
+		}
+
+		if readLen != defaultBase64EncodedOTPLength {
+			return "", fmt.Errorf("read %d bytes when we should have read 16", readLen)
+		}
+
+		return base64.StdEncoding.EncodeToString(buf), nil
+	default:
+		otp, err := base62.Random(otpLength)
+		if err != nil {
+			return "", fmt.Errorf("error reading random bytes: %w", err)
+		}
+
+		return otp, nil
+	}
+}
--- a/api/sys_namespaces.go
+++ b/api/sys_namespaces.go
@@ -46,6 +46,8 @@ func (c *Sys) CreateNamespaceWithContext(ctx context.Context, name string, req *
 	if err != nil {
 		return nil, err
 	}
+
+	//nolint:errcheck // ignoring the error here as its only fulfilling the signature and not returning any sensible error
 	defer resp.Body.Close()
 	var result struct {
 		Data *CreateNamespaceResponse
@@ -85,6 +87,8 @@ func (c *Sys) NamespaceSealStatusWithContext(ctx context.Context, name string) (
 	if err != nil {
 		return nil, err
 	}
+
+	//nolint:errcheck // ignoring the error here as its only fulfilling the signature and not returning any sensible error
 	defer resp.Body.Close()
 
 	var result struct {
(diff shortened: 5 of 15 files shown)


api/sys_generate_root.go
```diff
@@ -9,16 +9,15 @@ import (
 )
 
 type GenerateRootStatusResponse struct {
-	Nonce            string `json:"nonce"`
-	Started          bool   `json:"started"`
-	Progress         int    `json:"progress"`
-	Required         int    `json:"required"`
-	Complete         bool   `json:"complete"`
-	EncodedToken     string `json:"encoded_token"`
-	EncodedRootToken string `json:"encoded_root_token"`
```
wslabosz-reply, line 18: I couldn't find any usage of this field.

sdk/helper/roottoken/otp.go
```diff
@@ -11,27 +11,44 @@ import (
...
+		var length int
 		// This is the fallback case
-		buf := make([]byte, defaultBase64EncodedOTPLength)
+		if namespace {
+			length = namespaceDefaultBase64EncodedOTPLength
+		} else {
+			length = defaultBase64EncodedOTPLength
+		}
+
+		buf := make([]byte, length)
```
wslabosz-reply: So I've been looking at it, and I don't know if trimming the generated base64 string makes sense... also from [docs](https://openbao.org/api-docs/system/generate-root/#sample-response):

```
The OTP is a base62 string, with length of otp_length. The raw bytes (char codes) of the token will be XOR'd with this value before being returned as a response to the final unseal key, encoded as base64.
```
but we actually generate OTP as a base64 string, so I don't know where does the discrepancy come from

vault/external_tests/identity/identity_test.go
```diff
@@ -757,7 +757,9 @@ func TestLDAPNamespaceMultipleCreates(t *testing.T) {
 	client.SetToken(token)
 
 	for _, ns := range []string{"ns1", "ns2", "ns3"} {
-		resp, err := client.Logical().Write("sys/namespaces/"+ns, nil)
+		resp, err := client.Sys().CreateNamespace(&api.CreateNamespaceRequest{
+			Name: ns,
+		})
```
satoqz: I wonder if we should have

```go
func (s *Sys) CreateNamespace(string)
func (s *Sys) CreateNamespaceWithContext(context.Context, string)
func (s *Sys) CreateNamespaceWithOptions(context.Context, *CreateNamespaceRequest)
```

just to have a shorthand for the common case?

vault/namespace_store.go
```diff
@@ -930,6 +930,18 @@ func (ns *NamespaceStore) SealNamespace(ctx context.Context, path string) error
...
 	}
 
+	barrier := ns.core.sealManager.NamespaceBarrier(namespaceToSeal.Path)
+	if barrier == nil {
+		return ErrNotSealable
+	}
+
+	if barrier.Sealed() {
+		return nil
+	}
```
satoqz: Would it be better to perform this check in `sealManager.SealNamespace`? Saves a read lock.

api/sys_namespace.go
satoqz, line 1: Tbh I'd just have a single `sys_namespaces.go` (`s` at the end as the API endpoint is `/sys/namespaces`) for all that's currently `sys_namespace*.go`. Seems to mirror existing `sys_*` files more, and these are reasonably short.

vault/logical_system_namespaces_seals.go
```diff
@@ -60,29 +48,6 @@ func (b *SystemBackend) namespaceSealPaths() []*framework.Path {
...
-		},
-		"cluster_name": {
-			Type: framework.TypeString,
-		},
-		"cluster_id": {
-			Type: framework.TypeString,
-		},
-		"recovery_seal": {
-			Type: framework.TypeBool,
-		},
```
satoqz, line 82: Don't we need this one later?

sdk/helper/roottoken/otp.go
satoqz: I agree the entire function is cursed. Realistically I'd hope that the `default` branch is the one that gets used in practice, but the client doesn't validate that the server returns a non-zero otp length.

So preferably we'd be able to remove the entire `case 0` branch? Unless I'm missing something?

command/command_test.go
```diff
@@ -71,6 +72,42 @@ func testVaultServer(tb testing.TB) (*api.Client, func()) {
 	return client, closer
 }
 
+// testVaultServerWithNamespace creates a test vault cluster (with an existing namespace
+// sealed or unsealed depending on the [sealed] flag provided to the function) and returns
```
satoqz:
```suggestion
// sealed or unsealed depending on the sealed flag provided to the function) and returns
```

Doc links don't work for arguments.

api/sys_seal.go
```diff
@@ -90,19 +90,13 @@ func sealStatusRequestWithContext(ctx context.Context, c *Sys, r *Request) (*Sea
...
-	Sealed           bool     `json:"sealed"`
-	T                int      `json:"t"`
-	N                int      `json:"n"`
-	Progress         int      `json:"progress"`
-	Nonce            string   `json:"nonce"`
+type CoreSealStatusResponse struct {
+	*SealStatusResponse
 	Version          string   `json:"version"`
 	BuildDate        string   `json:"build_date"`
 	Migration        bool     `json:"migration"`
```
satoqz, line 108: Seems like an unnecessary breaking change for `api`, I'd probably just duplicate the full definitions for core/namespaced.

api/sys_namespace_generate_root.go
```diff
@@ -0,0 +1,112 @@
...
+
+package api
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+func (c *Sys) NamespaceGenerateRootStatus(namespacePath string) (*GenerateRootStatusResponse, error) {
```
satoqz:
```suggestion
func (c *Sys) NamespaceGenerateRootStatus(name string) (*GenerateRootStatusResponse, error) {
```

should this be `name`?

api/sys_namespace_seal.go
```diff
@@ -0,0 +1,93 @@
...
+	Type        string `json:"type"`
+	Initialized bool   `json:"initialized"`
+	Sealed      bool   `json:"sealed"`
+	T           int    `json:"t"`
+	N           int    `json:"n"`
+	Progress    int    `json:"progress"`
+	Nonce       string `json:"nonce"`
+}
+
+type NamespaceUnsealRequest struct {
```
satoqz:
```suggestion
type UnsealNamespaceRequest struct {
```

api/sys_namespace_seal.go
```diff
@@ -0,0 +1,93 @@
...
+	defer resp.Body.Close()
+
+	var result struct {
+		Data *SealStatusResponse
+	}
+	err = resp.DecodeJSON(&result)
+	return result.Data, err
+}
+
+func (c *Sys) NamespaceSeal(namespacePath string) error {
```
satoqz:
```suggestion
func (c *Sys) SealNamespace(namespacePath string) error {
```

I think for these we can afford placing the verb in front. I see how that's difficult with the status and root generation endpoints though.

api/sys_namespace_seal.go
```diff
@@ -0,0 +1,93 @@
...
+	defer cancelFunc()
+
+	_, err := c.c.rawRequestWithContext(ctx, r)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Sys) NamespaceUnseal(req NamespaceUnsealRequest) (*SealStatusResponse, error) {
```
satoqz:
```suggestion
func (c *Sys) UnsealNamespace(req UnsealNamespaceRequest) (*SealStatusResponse, error) {
```

vault/external_tests/identity/identity_test.go
wslabosz-reply: tbh, makes sense, will add that

api/sys_namespace.go
wslabosz-reply: I worry about the length of the file, but fine, will make it one file for now.

vault/logical_system_namespaces_seals.go
wslabosz-reply: we will need it in the future, yes.

api/sys_namespace_generate_root.go
wslabosz-reply: in my mind `name` == `namespacePath` in this context, as we cannot specify namespaces as a hierarchy here (e.g. "ns1/ns2" won't work as you need to pass the "higher order" namespace in the context)

satoqz: Hence I was thinking `name` would be more true to this API (And since we've already made the distinction on the backend).

.golangci.yml
```diff
@@ -3,6 +3,10 @@ linters:
   enable:
     - nilnil
     - misspell
+  settings:
+    errcheck:
+      exclude-functions:
+        - (io.ReadCloser).Close
```
phyrog: I'm curious about the reason for this

sdk/helper/roottoken/decode.go
```diff
@@ -6,38 +6,21 @@ package roottoken
...
 	"fmt"
-	"strings"
 
-	uuid "github.com/hashicorp/go-uuid"
 	"github.com/openbao/openbao/sdk/v2/helper/xor"
 )
 
 // DecodeToken will decode the root token returned by the Vault API
 // The algorithm was initially used in the generate root command
 func DecodeToken(encoded, otp string, otpLength int) (string, error) {
```
phyrog, line 17: `otpLength` seems to be unused now, and with the removal of `GenerateOTP` we're already introducing a breaking change to the sdk.

api/sys_generate_root.go
phyrog: Yeah it's weird. It is either not set or it is the same as `EncodedToken`. Removing the field in the API is a breaking change though, and IMO would need to be documented (ideally deprecated first) -> might want to pull this change out and bring it to main asap.

phyrog (commented):
Looks good overall, just a few comments

sdk/helper/roottoken/decode.go
wslabosz-reply: gonna revert changes to the sdk, and introduce them to main then (I hate this.)

.golangci.yml
wslabosz-reply: the motivation here is that we've got 150+ occurrences of a `defer resp.Body.Close()` which fail due to `errcheck` linter rule. The issue with this is that in very very very rare cases it will fail, we still cannot do anything about it, hence my proposal here to exclude it from the rule.

api/sys_generate_root.go
wslabosz-reply: @cipherboy 
I'd like your insight here, do we need to deprecate this completely unused field before removing it alltogether?

.golangci.yml
phyrog: Doesn't this pose a danger for memory leaks? I can see where you're coming from, but this would ignore **all** `(io.ReadCloser).Close` calls everywhere.

I would rather have a bunch of `//nolint` directives to make the conscious decision which returns to ignore, than to flat-out ignore all of them. But that's just my opinion.

wslabosz-reply: what memory leaks? It's just ignoring an error case that never happens (for `net/http Response.Body.Close()` method)

phyrog: Other (non-`Response.Body`) instances of `io.ReadCloser` elsewhere in the code. Not saying there are any such memory/resource leaks in the code currently, but this option would hide future potential leaks.

wslabosz-reply: right, I'd have to limit the omit rule to just `net/http.(Response.Body)` struct... or disable the rule with `nolint`, let me try former, and if it won't work then I'll move to latter.

api/sys_generate_root.go
cipherboy: @phyrog @wslabosz-reply I think, based on GitHub code search, there's many references to this. I'd like, if possible, to keep allowing them to continue using it, even if it is empty as many of them are guarded by non-empty checks. Including in bank-vaults. So if we want more people to adopt our client library, I think it best to keep the field even though upstream and us have kept it empty for a while.

wslabosz-reply:
@cipherboy 
will need your approval to proceed with this PR

cipherboy (approved):
I confessed I only skimmed the commands, but no point holding this up more. 


Many thanks @wslabosz-reply!

PR #1889, opened 2025-10-06
fix memory leak in raft

phil9909:
Supersedes: #1878

Fixes a memory leak in rafts transaction handling on standby nodes.

`completeTransaction` was never called on the standby nodes, therefore the map that tracks the transactions (for the fast apply optimization) grows until OOM. 

This change add an optional `lowest_active_index` to the protobuf raft protocol to inform standbys which entries they can remove. As this is an optional field, it should be completely compatible, but if either the primary or standby does not support the field yet, the standby will continue to leak memory.

EDIT: `vault.raft-storage.txn_fast_apply_read_hit` show a hit ration of 100% when benchmarking with 60 kvv2 writes per second

### DIFF
--- a/physical/raft/transaction.go
+++ b/physical/raft/transaction.go
@@ -627,6 +627,10 @@ func (t *RaftTransaction) Commit(ctx context.Context) error {
 	defer func() {
 		if t.writable {
 			t.b.fsm.fastTxnTracker.completeTransaction(t.index)
+
+			// in "Rollback" we call fastTxnTracker.clearOldEntries(...) at this
+			// We don't do this in "Commit" because it will be called from the fsm.
+			// This ensures the cleanup also happens on standby nodes
 		}
 
 		t.b.fsm.l.RUnlock()


physical/raft/transaction.go
```diff
@@ -715,6 +720,7 @@ func (t *RaftTransaction) Commit(ctx context.Context) error {
 	// the transaction application in Raft, applyLog will gather it for us
 	// and return it as a proper error.
 	t.b.l.RLock()
+	log.LowestActiveIndex = pointerutil.Ptr(min(lowestActiveIndex, t.b.raft.AppliedIndex()-1)) // we need to cap the lowest active index, otherwise we might miss transaction started later
```
phil9909, line 727: No sure, if we really need this. Seems to be a pathological case, especially on clusters with some load (as this will only affect transactions that are committed within their start index and every single log bumps the index).

I guess it is the more "correct" way to handle it, but it has the downside, that it might keep entries longer in the tracker than needed. If we on the other hand evict entries to early, we will have additional reads. As this should only happen on clusters without much load, it probably does not matter to have a few additional reads.

physical/raft/transaction.go
cipherboy, line 629: @phil9909 Actually, looking at this a touch more, why don't we apply the logic below here as well and remove all older transactions on commit as well?

phil9909: Because the logic is triggered in `ApplyBatch` of the fsm, which runs also on standbys.

cipherboy: @phil9909 Hmm, but why do we apply it in `rollback`? I think `commit` vs `rollback` should be the same.

phil9909: Well, rollback does not happen on the standbys, it's not part of the raft.
The cleanup on the standbys is deferred until the next successful commit. We could do this on the primary as well (i.e. delete the call from rollback), if you wish.

physical/raft/transaction.go
driif, line 727:
```suggestion
	appliedIndex := t.b.raft.AppliedIndex()
	// Prevent underflow: only set if AppliedIndex > 0
	if appliedIndex > 0 {
		log.LowestActiveIndex = pointerutil.Ptr(min(lowestActiveIndex, appliedIndex-1)) // we need to cap the lowest active index, otherwise we might miss transaction started later
	}
```

Should we check here at first if appliedIndex > 0 ?

physical/raft/transaction.go
```diff
@@ -735,6 +741,12 @@ func (t *RaftTransaction) Rollback(ctx context.Context) error {
 	// Also unlock the read lock on the underlying fsm.
 	defer func() {
 		if t.writable {
+			lowestActiveIndex := t.b.fsm.fastTxnTracker.lowestActiveIndexAfterCommit(t.index)
+			t.b.l.RLock()
+			lowestActiveIndex = min(lowestActiveIndex, t.b.raft.AppliedIndex()-1) // we need to cap the lowest active index, otherwise we might miss transaction started later
```
driif, line 750:
```suggestion
			appliedIndex := t.b.raft.AppliedIndex()
			// Prevent underflow: only cap if AppliedIndex > 0
			if appliedIndex > 0 {
				lowestActiveIndex = min(lowestActiveIndex, appliedIndex-1) // we need to cap the lowest active index, otherwise we might miss transaction started later
			}
```

Also here

physical/raft/transaction.go
phil9909: I don't think so. 0 is a valid index in raft (read: in the implementation of raft we are using)

driif: But does this mean, if appliedIndex == 0, we will clean up everything, as we will get the LowestActiveIndex max value of uint64?

driif: @phil9909 I've checked https://go.dev/play/p/a0dASDoA3Jn - okay in this case we will get back lowestActiveIndex.

physical/raft/transaction.go
cipherboy, line 750: @phil9909 This line, no? https://github.com/openbao/openbao/pull/1889/files#r2408076728 

This happens on the active node, in Rollback. Why is this not also in Commit?

physical/raft/transaction.go
cipherboy: @phil9909 See https://github.com/openbao/openbao/pull/1889/files#r2410530359

physical/raft/transaction.go
phil9909: You mean the call to `min`? It is: https://github.com/openbao/openbao/pull/1889/files#diff-2b934c9e2debbca87bb8b4c1d94caac8fb77ed5ee82fe04773d9a424c16ad4d8R723

cipherboy: And `
			t.b.fsm.fastTxnTracker.clearOldEntries(lowestActiveIndex)`.

phil9909: Is called from within `ApplyBatch` of the fsm: https://github.com/openbao/openbao/pull/1889/files#diff-ff728caf2c3176f15a2f7c7ce5e4390f8f7a7448a66a78cc25b5c761a4d8bf4fR922-R924

Which will make sure it runs on standbys. We could also run it in the `Commit` method, but then we will scan the map twice on the primary and the second run should not find any additional items.

cipherboy: Ahhh, yes. Do you want to add a comment about this?

phil9909: Added.

cipherboy: Perfect, thanks! I think it just helps to clarify for anyone looking at why `rollback` and `commit` differ in the future.

phil9909:
Can we merge this?

satoqz (approved):

satoqz:
@phil9909 yes!

PR #1893, opened 2025-10-07
Switch to singleton pattern regexes

cipherboy:
In scenarios with large numbers of homogeneous mounts, the path cache generates new regexes for every backend, despite regexes being thread-safe. Thus a global singleton cache allows us to greatly reduce memory usage as seen in the profiling discussions.

See also: https://github.com/orgs/openbao/discussions/1865

---

fyi @stormshield-gt

### DIFF
--- a/sdk/framework/backend.go
+++ b/sdk/framework/backend.go
@@ -30,6 +30,10 @@ import (
 	"github.com/openbao/openbao/sdk/v2/logical"
 )
 
+// regexSingletonCache is used to reduce memory of mounting of multiple
+// builtin plugins.
+var regexSingletonCache sync.Map
+
 // Backend is an implementation of logical.Backend that allows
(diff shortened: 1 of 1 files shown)


cipherboy:
> LGTM. I suppose the we'd never need to purge the cache as its size is bounded by the amount of built in engines (a single one for plugin processes) times their paths.

I think it is strictly bounded above by number of unique regexes in the process context (builtin or external). Some different engines reuse regexes (e.g., I'm sure `GenericNameRegex` is called with `path` or `name` or `role` in a lot of different engines), saving us memory there.

It would be hard to implement purging; we'd need a ref counter or something to support cleaning up when all mounts using that particular pattern unmount, not just all mounts of a particular type. But since the number of regexes is bounded, it seems unlikely to cause a memory leak in practice, unless an intentionally malicious plugin was loaded.

PR #1899, opened 2025-10-08
Namespace sealing should also cleanup non-sealable NS resources

wslabosz-reply:
This PR introduces following changes:
- refactored `(*SealManager).SealNamespace()` into `(*NamespaceStore).SealNamespace()` (and `(*NamespaceStore).sealNamespaceLocked()`, enabling cleanup of the namespace hierarchy even if a part of it is not sealable
- refactored affected `handleNamespacesSeal()` handler
- reordered and removed unnecessary code in `(*SealManager).InitializeBarrier()`
- tweaked affected tests

### DIFF
--- a/vault/logical_system_namespaces_seals.go
+++ b/vault/logical_system_namespaces_seals.go
@@ -275,25 +275,7 @@ func (b *SystemBackend) handleNamespacesSeal() framework.OperationFunc {
 			return nil, errors.New("name must not contain /")
 		}
 
-		namespaceToSeal, err := b.Core.namespaceStore.GetNamespaceByPath(ctx, name)
-		if err != nil {
-			return nil, err
-		}
-
-		if namespaceToSeal == nil {
-			return nil, errors.New("namespace doesn't exist")
-		}
-
-		if namespaceToSeal.ID == namespace.RootNamespaceID {
-			return nil, errors.New("unable to seal root namespace")
-		}
-
-		if namespaceToSeal.Tainted {
-			return nil, errors.New("unable to seal tainted namespace")
-		}
-
-		err = b.Core.namespaceStore.SealNamespace(ctx, namespaceToSeal)
-		if err != nil {
+		if err := b.Core.namespaceStore.SealNamespace(ctx, name); err != nil {
 			return handleError(err)
 		}
 
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -899,7 +899,7 @@ func namespaceMatchPredicate(targetNS *namespace.Namespace) func(*MountEntry) bo
 
 // SealNamespace acquires a read lock, and seals provided namespace,
 // cleaning up namespace resources.
-func (ns *NamespaceStore) SealNamespace(ctx context.Context, namespaceToSeal *namespace.Namespace) error {
+func (ns *NamespaceStore) SealNamespace(ctx context.Context, path string) error {
 	defer metrics.MeasureSince([]string{"namespace", "seal_namespace"}, time.Now())
 
 	unlock, err := ns.lockWithInvalidation(ctx, false)
@@ -907,6 +907,23 @@ func (ns *NamespaceStore) SealNamespace(ctx context.Context, namespaceToSeal *na
 		return err
 	}
 	defer unlock()
+
+	namespaceToSeal, err := ns.getNamespaceByPathLocked(ctx, path, false)
+	if err != nil {
+		return err
+	}
+
+	if namespaceToSeal == nil {
+		return errors.New("namespace doesn't exist")
+	}
+
+	if namespaceToSeal.ID == namespace.RootNamespaceID {
+		return errors.New("unable to seal root namespace")
+	}
+
+	if namespaceToSeal.Tainted {
+		return errors.New("unable to seal tainted namespace")
+	}
 
 	return ns.sealNamespaceLocked(ctx, namespaceToSeal)
 }
@@ -918,7 +935,7 @@ func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSe
 
 	var errs error
 	ns.namespacesByPath.PostOrderTraversal(namespaceToSeal.Path, func(namespaceEntry *namespace.Namespace) {
-		if namespaceEntry.UUID == namespace.RootNamespaceUUID || ns.core.NamespaceSealed(namespaceEntry) {
+		if namespaceEntry.ID == namespace.RootNamespaceID || ns.core.NamespaceSealed(namespaceEntry) {
 			return
 		}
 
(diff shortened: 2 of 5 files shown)


vault/namespace_store_tree.go
```diff
@@ -81,6 +90,31 @@ func (nt *namespaceTree) LongestPrefix(path string) (string, *namespace.Namespac
...
+
+	for len(nodes) > 0 {
+		node := nodes[len(nodes)-1]
+		nodes = nodes[:len(nodes)-1]
+		ordNodes = append(ordNodes, node)
+
+		keys := slices.Collect(maps.Keys(node.children))
+		for _, key := range keys {
+			nodes = append(nodes, node.children[key])
+		}
```
Huy-Dinh-Reply: nit: I think this would also work
```suggestion
		for _, childNode := range node.children {
			nodes = append(nodes, childNode)
		}
```
though I'm wondering if sibling order matters here, if yes we may need to get the keys, sort them, and append. For our use case I think sibling order does not matter?

vault/seal_manager.go
```diff
@@ -623,7 +587,6 @@ func (sm *SealManager) AuthenticateRootKey(ctx context.Context, ns *namespace.Na
 
 func (sm *SealManager) InitializeBarrier(ctx context.Context, ns *namespace.Namespace) ([][]byte, error) {
 	sm.lock.RLock()
-	defer sm.lock.RUnlock()
```
Huy-Dinh-Reply, line 626: Would removing this cause the sm lock to stay locked if any of the operation fails and `SealManager.InitializeBarrier` returns before we call `sm.lock.RUnlock()`?

vault/seal_manager.go
```diff
@@ -667,37 +630,32 @@ func (sm *SealManager) InitializeBarrier(ctx context.Context, ns *namespace.Name
...
 
-	if err := sm.SealNamespace(ctx, ns); err != nil {
-		return nil, fmt.Errorf("failed to seal namespace barrier: %w", err)
-	}
+	// unlock as namespaceStore.sealNamespace will acquire a lock on itself
+	sm.lock.RUnlock()
 
-	results := &InitResult{
-		SecretShares: [][]byte{},
+	if err := sm.core.namespaceStore.sealNamespaceLocked(ctx, ns); err != nil {
```
Huy-Dinh-Reply: Should this have been

```suggestion
	if err := sm.core.namespaceStore.SealNamespace(ctx, ns); err != nil {
```

to ensure we have the `NamespaceStore` lock?

vault/namespace_store_tree.go
wslabosz-reply: yeah, that works.

sibling order shouldn't matter

vault/seal_manager.go
wslabosz-reply: you are right, missed that

vault/seal_manager.go
wslabosz-reply: well, not that easy, `InitalizeBarrier()` is called in `SetNamespaceLocked()` which itself holds a lock, so that's why I've used `sealNamespaceLocked()`... have to rethink it once again

vault/namespace_store.go
```diff
@@ -318,13 +318,14 @@ func (ns *NamespaceStore) SetNamespaceSealed(ctx context.Context, entry *namespa
 	if err != nil {
 		return nil, err
 	}
-	defer unlock()
 
 	new, err := ns.setNamespaceLocked(ctx, entry)
 	if err != nil {
 		return nil, err
```
Huy-Dinh-Reply, line 327: should we also unlock before returning here?
```suggestion
		unlock()
		return nil, err
```
I wonder if it makes sense to create a Mediator for `NamespaceStore` and `SealManager` to break up the mutual dependency between them. So methods where the 2 structs interact should be implemented there like (very roughly)

```
func (m *Mediator) sampleMethod(ctx context.Context) (error) {
	_, nsUnlock := m.ns.lockWithInvalidation(ctx,true)
    defer nsUnlock()

    m.sm.lock.Lock()
    defer m.sm.lock.Unlock()
    
    // Interactions go here
    m.ns.sampleMethodLocked()
    m.sm.sampleMethodLocked()
}
```
so we have to worry less about recursive deadlocks. What do you think? 🤔

wslabosz-reply: not sure how feasible is that, I worry that few patterns wouldn't be easy to "straighten" into sequential method

Huy-Dinh-Reply: good point. I was thinking we may keep the same logics, just use the `...Locked()` versions of the methods in the Mediator calls, but this may get complicated and is anyway out of scope for the PR. Maybe it's worth having a look later.

vault/namespace_store.go
```diff
@@ -914,35 +897,54 @@ func namespaceMatchPredicate(targetNS *namespace.Namespace) func(*MountEntry) bo
...
+		ctx = namespace.ContextWithNamespace(ctx, namespaceEntry)
+		if err := ns.clearNamespacePolicies(ctx, namespaceEntry, false); err != nil {
+			errs = errors.Join(errs, err)
+		}
+		if err := ns.UnloadNamespaceCredentials(ctx, namespaceEntry); err != nil {
+			errs = errors.Join(errs, err)
+		}
+		if err := ns.UnloadNamespaceMounts(ctx, namespaceEntry); err != nil {
+			errs = errors.Join(errs, err)
+		}
```
satoqz, line 959: Reminder for https://github.com/openbao/openbao/pull/1599#discussion_r2394649089, but certainly fits into another PR, too.

vault/namespace_store.go
```diff
@@ -914,35 +897,54 @@ func namespaceMatchPredicate(targetNS *namespace.Namespace) func(*MountEntry) bo
...
+// cleaning up namespace resources.
+func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSeal *namespace.Namespace) error {
+	defer metrics.MeasureSince([]string{"namespace", "seal_namespace"}, time.Now())
 
-	if namespaceToSeal.ID == namespace.RootNamespaceID {
-		return errors.New("unable to seal root namespace")
-	}
+	var errs error
+	ns.namespacesByPath.PostOrderTraversal(namespaceToSeal.Path, func(namespaceEntry *namespace.Namespace) {
+		if namespaceEntry.UUID == namespace.RootNamespaceUUID || ns.core.NamespaceSealed(namespaceEntry) {
```
satoqz:
```suggestion
		if namespaceEntry.ID == namespace.RootNamespaceID || ns.core.NamespaceSealed(namespaceEntry) {
```

vault/logical_system_namespaces_seals.go
```diff
@@ -275,7 +275,24 @@ func (b *SystemBackend) handleNamespacesSeal() framework.OperationFunc {
...
+			return nil, errors.New("namespace doesn't exist")
+		}
+
+		if namespaceToSeal.ID == namespace.RootNamespaceID {
+			return nil, errors.New("unable to seal root namespace")
+		}
+
+		if namespaceToSeal.Tainted {
+			return nil, errors.New("unable to seal tainted namespace")
+		}
```
satoqz: These checks happen too early, before `SealNamespace` re-fetches the entry, which might have changed by that point (e.g., it might be tainted on second fetch). Is there really a need for the initial `GetNamespaceByPath`? Preferably all would happen in the `SealNamespace` call only.

vault/namespace_store.go
wslabosz-reply: forgot about it (again)

vault/logical_system_namespaces_seals.go
wslabosz-reply: I've moved it there so (*SealManager).Reset() would omit the fetching and checks, as we forbid root sealing, I've got to think about something else there then.

PR #1900, opened 2025-10-08
Fix incorrect references of `-plugin-version` arg in the documentation

### LINKED ISSUE #1890: `bao plugin register` with `-version` results in: "flag provided but not defined: -plugin-version"
**Describe the bug**
The [documentation](https://openbao.org/docs/commands/plugin/register/#command-options) describes the option `-version`:

```text
-plugin-version (string: "") - Semantic version of the plugin to run from the catalog. If unspecified, refers to the unversioned plugin registered with the same name and type, or the built-in plugin, in that order of precedence.
```

But that option is not processed:

```bash
$ bao plugin register -sha256=some_sh256_sum -plugin-version 0.1.0 secret openbao-plugin-secrets-aws 
flag provided but not defined: -plugin-version
```

**To Reproduce**
1. Run `bao plugin register -sha256=some_sh256_sum -plugin-version 0.1.0 secret openbao-plugin-secrets-aws`
2. See error: `flag provided but not defined: -plugin-version`

**Expected behavior**
Either the documentation should not describe this option, or (better) bao should allow me to set a version for the plugin.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.4.1 (and 2.4.0)
* OpenBao CLI Version (retrieve with `bao version`): 2.4.1 (and 2.4.0)

```hcl
ui = true


storage "raft" {
  path = "/opt/openbao/raft"
  node_id = "openbao-1"
}

cluster_addr = "http://127.0.0.1:8201"

listener "tcp" {
  address = "0.0.0.0:8200"
  tls_disable = true
}
```

wslabosz-reply:
Resolves #1890

### DIFF
--- a/command/plugin_info.go
+++ b/command/plugin_info.go
@@ -39,6 +39,10 @@ Usage: bao plugin info [options] TYPE NAME
 
       $ bao plugin info database mysql-database-plugin
 
+  Get info about a specific version of a plugin:
+
+      $ bao plugin info -version=v1.0.0 database mysql-database-plugin
+
 ` + c.Flags().Help()
 
 	return strings.TrimSpace(helpText)
--- a/website/content/docs/commands/auth/enable.mdx
+++ b/website/content/docs/commands/auth/enable.mdx
@@ -71,7 +71,6 @@ flags](/docs/commands) included on all commands.
   mount in the UI-specific listing endpoint. Valid values are `"unauth"` or `"hidden"`,
   with the default `""` being equivalent to `"hidden"`.
 
-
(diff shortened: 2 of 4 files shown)


command/plugin_info.go
```diff
@@ -39,6 +39,10 @@ Usage: bao plugin info [options] TYPE NAME
 
       $ bao plugin info database mysql-database-plugin
 
+  Get info about a specific version of plugin:
```
satoqz:
```suggestion
  Get info about a specific version of a plugin:
```

satoqz (approved):

JanMa (approved):

PR #1902, opened 2025-10-08
Cache logical backend field information

stormshield-gt:
This MR caches logical backend fields to decrease memory usage when there are a lot of mounts.

I tested it with 10k PKI mounts,  I go from 4,5 GiB to 2,84 GiB, a 37% improvement. 

<details>

<summary>Before</summary>

without-metrics-patch-cipherboy-10k

without-metrics-patch-cipherboy-global

without-metrics-patch-cipherboy-pki

</details>

<details>

<summary>After</summary>

10k-patch-claud-v2

10k-patch-claud-v2-global

10k-patch-claud-v2-pki

</details>

Here is a diff flamegraph to better understand where allocation shifted:

<details>

<summary>Diff flamegraph</summary>

image

</details>

> [!WARNING] 
> AI tooling was used to help me generate the patch, as I'm not a go developer

Related to https://github.com/orgs/openbao/discussions/1865

### DIFF
--- a/sdk/framework/backend.go
+++ b/sdk/framework/backend.go
@@ -37,6 +37,58 @@ import (
 // on the resulting regex.
 var regexSingletonCache sync.Map
 
+// fieldSingletonCache is a shared cache of Fields maps across all backend instances.
+// Fields maps contain FieldSchema definitions which are immutable metadata (Type, Description, etc.).
+// By deduplicating identical Fields maps, we avoid storing the same schema definitions
+// thousands of times when mounting many instances of the same backend type.
+// This is safe because FieldSchemas are read-only and contain no backend-specific state.
+var fieldSingletonCache sync.Map // map[string]map[string]*FieldSchema
+
+// fieldsCacheKey generates a stable cache key for a Fields map.
+// We build a key from sorted field names combined with their type and description prefix.
+// Since field definitions are immutable, identical field sets will produce identical keys.
+func fieldsCacheKey(fields map[string]*FieldSchema) string {
+	if len(fields) == 0 {
+		return ""
+	}
+
+	// Collect and sort field names for stable key generation
+	names := make([]string, 0, len(fields))
+	for name := range fields {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+
+	// Build key from sorted names and type signatures
+	// Use first 50 chars of description to keep key length reasonable while maintaining uniqueness
+	var keyBuilder strings.Builder
+	for _, name := range names {
+		schema := fields[name]
+		descLen := min(len(schema.Description), 50)
+		desc := schema.Description[:descLen]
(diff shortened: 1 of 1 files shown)


suprjinx:
Looks very reasonable -- I'd definitely appreciate some unit tests for the new functions, mainly as documentation.

cipherboy:
@stormshield-gt Just a note: 

> AI tooling was used to help me generate the patch, as I'm not a go developer

I'd suggest reading CONTRIBUTING.md's call out about this.

----

Ahh I see. I think this is fine, but there's also an outstanding bug where PKI modifies schema directly (whoops), but that's fixable with some random code I have lying on my laptop. I might take a pass at an alternative version of this.

I think it'd be hard to de-duplicate `PathOperation`. I think that's where a lot of the remaining memory usage is... :thinking: 

I think what ultimately needs to happen is `PathOperation` would need to be split into two parts (a common/static portion and just the handler) and the former is de-duplicated.

sdk/framework/backend.go
```diff
@@ -37,6 +37,58 @@ import (
 // on the resulting regex.
 var regexSingletonCache sync.Map
 
+// fieldSingletonCache is a shared cache of Fields maps across all backend instances.
+// Fields maps contain FieldSchema definitions which are immutable metadata (Type, Description, etc.).
+// By deduplicating identical Fields maps, we avoid storing the same schema definitions
+// thousands of times when mounting many instances of the same backend type.
+// This is safe because FieldSchemas are read-only and contain no backend-specific state.
+var fieldSingletonCache sync.Map // map[string]map[string]*FieldSchema
```
cipherboy, line 45: How much memory do we save by de-duplicating the whole map versus just the `FieldSchema` instances? I think the latter would likely be sufficient and be easier to cache, perhaps.

sdk/framework/backend.go
```diff
@@ -37,6 +37,58 @@ import (
...
+	names := make([]string, 0, len(fields))
+	for name := range fields {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+
+	// Build key from sorted names and type signatures
+	// Use first 50 chars of description to keep key length reasonable while maintaining uniqueness
+	var keyBuilder strings.Builder
+	for _, name := range names {
```
cipherboy, line 65: I don't think this works the way you want it to. Range order over maps is [non-deterministic](https://nathanleclaire.com/blog/2014/04/27/a-surprising-feature-of-golang-that-colored-me-impressed/#iteration-order) so you'd at minimum have to sort the keys first.

You might also use `sha256.Sum()` or something to improve performance on high-cardinality endpoints like PKI.

stormshield-gt:
Thanks for the detailed explanation, I understand and respect the decision.

I will submit an issue then

PR #1905, opened 2025-10-08
Ensure raft voter operations are handled locally

cipherboy:
On standby nodes, calls to change voter status are forwarded to the primary, which doesn't help when only non-voters remain. This makes leadership elections unwinnable without modifying `peers.json` directly. This adjust the handler to support being called on standby nodes, to allow directly interacting with the current standby's state.

See also: https://github.com/openbao/openbao/pull/996#issuecomment-3382139174

---

@pree @eyenx Let me know if this helps. I'm not entirely convinced if this will work on standby nodes (do you know off the top of your head @JanMa?), but if it does, it might solve some things. You might need to call this multiple times on each non-voter node:

```
BAO_ADDR=standby1 bao write sys/storage/raft/promote server_id=standby1
BAO_ADDR=standby2 bao write sys/storage/raft/promote server_id=standby1
BAO_ADDR=standby3 bao write sys/storage/raft/promote server_id=standby1

BAO_ADDR=standby1 bao write sys/storage/raft/promote server_id=standby2
BAO_ADDR=standby2 bao write sys/storage/raft/promote server_id=standby2
BAO_ADDR=standby3 bao write sys/storage/raft/promote server_id=standby2

&c
```

to get elections to fully work.

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -204,6 +204,8 @@ func handler(props *vault.HandlerProperties) http.Handler {
 
 		mux.Handle("/v1/sys/storage/raft/bootstrap", handleSysRaftBootstrap(core))
 		mux.Handle("/v1/sys/storage/raft/join", handleSysRaftJoin(core))
+		mux.Handle("/v1/sys/storage/raft/promote", handleLogicalNoForward(core))
+		mux.Handle("/v1/sys/storage/raft/demote", handleLogicalNoForward(core))
 		mux.Handle("/v1/sys/internal/ui/feature-flags", handleSysInternalFeatureFlags(core))
 
 		for _, path := range injectDataIntoTopRoutes {
--- a/website/content/api-docs/system/storage/raft.mdx
+++ b/website/content/api-docs/system/storage/raft.mdx
@@ -145,6 +145,10 @@ This endpoint removes a node from the raft cluster.
 | :----- | :------------------------------ |
 | `POST` | `/sys/storage/raft/remove-peer` |
 
+### Parameters
+
+- `server_id` `(string: <required>)`: id of the server to promote or demote
+
 ### Sample payload
 
 ```json
@@ -172,6 +176,10 @@ may be provided if the node is in a DR secondary cluster.
 |:-------|:----------------------------|
 | `POST` | `/sys/storage/raft/promote` |
 
+### Parameters
+
+- `server_id` `(string: <required>)`: id of the server to promote or demote
+
 ### Sample payload
 
 ```json
@@ -198,6 +206,10 @@ may be provided if the node is in a DR secondary cluster.
 | Method | Path                       |
 |:-------|:---------------------------|
 | `POST` | `/sys/storage/raft/demote` |
+
+### Parameters
+
+- `server_id` `(string: <required>)`: id of the server to promote or demote
 
 ### Sample payload
 


JanMa:
@cipherboy from my current understanding this will not help, as promotions/demotions in a cluster always need an active leader. By making the standby handle this request, you can update the map of permanent non-voters in the nodes local Raft state, but it will not cause the voter state to change.

cipherboy:
@JanMa Hmm, does this

> By making the standby handle this request, you can update the map of permanent non-voters in the nodes local Raft state, but it will not cause the voter state to change.

mean then that in a cluster of only non-voters (and thus leaderless), there is no way to do a promotion of any node to active? 

Sounds like something we'd have to poke the Raft library to allow.

JanMa:
> @JanMa Hmm, does this
> 
> > By making the standby handle this request, you can update the map of permanent non-voters in the nodes local Raft state, but it will not cause the voter state to change.
> 
> mean then that in a cluster of only non-voters (and thus leaderless), there is no way to do a promotion of any node to active?

That's exactly what the `peers.json` file is for. Not sure if we can make this happen in a running cluster, as you said we need to look at the Raft library if it's possible.

cipherboy:
@JanMa Hmm, this means our approach to disaster recovery and using API to change voting node status in the event of a failure isn't workable. We'd have to have a manual promotion process (using `peers.json`) instead, which is less useful and might push us to some other form of DR... 🤔

pree:
I think in a normally working cluster having it handled by the leader is the right way to go.
Still it would be nice to have the possibility to forcefully promote a follower in a leaderless cluster.

With `peers.json` you need to rebuild the desired state of nodes for a clean failover. IMHO a disaster recovery process will always be a manual step to prevent the issue of introducing a split brain situation in case of a network issue between the clusters.

Another thing we need to be aware of: if we promote a node on the standby cluster, will it just be added to the existing (& dead) voters and thus unable to perform quorum on it's own? This would also be an issue if we promote the same number of nodes as the active cluster had, as we're not reaching n/2+1 nodes.

pree:
> Still it would be nice to have the possibility to forcefully promote a follower in a leaderless cluster.

For example a "takeover" command, which makes the targeted standby node the only leader node. Afterwards the other standby nodes can be easily promoted.

cipherboy:
@pree said:

> With `peers.json` you need to rebuild the desired state of nodes for a clean failover. IMHO a disaster recovery process will always be a manual step to prevent the issue of introducing a split brain situation in case of a network issue between the clusters.

Right. I wasn't arguing that the step should be automatically: I just meant you should be able to achieve DR failover without requiring manual access to the box. It'd be great if this were fully API driven with no additional direct system access requirements.


> Another thing we need to be aware of: if we promote a node on the standby cluster, will it just be added to the existing (& dead) voters and thus unable to perform quorum on it's own? This would also be an issue if we promote the same number of nodes as the active cluster had, as we're not reaching n/2+1 nodes.

There's already two APIs for this `demote` and `remove-peer`. Which you use depends on how hard the primary zone failed. If it is a network glitch you can't wait to recover from, demote the existing voters on all standbys first and then promote your new subset of DR voters on all standbys. This gives you known quorum.

pree:
Thanks for clarification @cipherboy 

I guess this now depends on how feasible it is to have raft change the voter mode?

cipherboy:
@pree Yeah, do you want to investigate that? I'm also really curious if this just happens to work magically somehow.

JanMa:
> Thanks for clarification @cipherboy
> 
> I guess this now depends on how feasible it is to have raft change the voter mode?

AFAIK a node in an existing Raft cluster/quorum can't change it's own voter status. This can only be done by the elected leader or by rejoining from scratch. So essentially we'd need to find a non-disruptive way to recover the cluster state from the remaining non-voter nodes in the Standby cluster. The Raft library exposes an API for this, which is used when loading the `peers.json` file, but I don't know if this can be done in a "live" cluster.

https://pkg.go.dev/github.com/hashicorp/raft#RecoverCluster

pree:
@JanMa isn't the override we do using the `peers.json` also disruptive? From what I understand it's overriding all members with the ones defined in `peers.json`.

As this is a DR operation it could also trigger an restart imho, having to unseal it afterwards wouldn't be that bad.

@cipherboy I'm happy to have a look, but I'm pretty busy currentl, so it might take 2-3 weeks until I find some time for this.

cipherboy:
> https://pkg.go.dev/github.com/hashicorp/raft#RecoverCluster

@JanMa @pree My read of this is that it is exactly what we want, though, may be less safe to do live in the general case. Since all remaining nodes in a DR scenario are _presumably_ non-voter (or, the active could be marked non-voter and step-down), I think it is safe to say that there's no in flight commits and thus that we should be able to globally adjust peer information.

What's missing, I think, is that all peers would need to be updated, so we'll either need a peer-to-peer GRPC (currently we only have peer->active) or we'd need to call this on every non-voting node.

Thoughts?

pree:
I think calling on every node would be fine, but not ideal.

There are two different setups I would envision:
A) 
1. The DR cluster consists of 5 non-voters connected to the active node in a cluster of 5 voters. 
2. :zap: Disaster struck :zap: and all voters are unavailable.
3. We "promote/detach" one of the non-voters
4. The 4 other voters realize there is a new active node in town and connect to it.
5. Either automatically (after detach) or manually the other 4 non-voters are promoted using `sys/storage/raft/promote`

B)
1. The DR cluster consists of 1 non-voter connected to the active node in a cluster of 5 voters.
    > This has the benefit, that we don't need fancy networking like a peered VNET or Service Mesh
2. :zap: Disaster struck :zap: and all voters are unavailable.
3. We "promote/detach" the non-voter
4. We scale up the DR cluster to 5 nodes.
5. As the 4 new nodes are uninitialized they join the single node and form a new raft cluster.

pree:
Both options are then pretty close to HashiCorp's implementation of DR. As it's pulling the WAL (iirc) it's way easier to configure then non-voters imho. Especially if we think of complex setups like multi-cloud or different kubernetes clusters.

JanMa:
> Both options are then pretty close to HashiCorp's implementation of DR. As it's pulling the WAL (iirc) it's way easier to configure then non-voters imho. Especially if we think of complex setups like multi-cloud or different kubernetes clusters.

I think long term we need to think about introducing some sort of replication to a backup cluster as well for proper DR. In theory this should already be doable with PostgreSQL logical replication today, but of course something built in on top of Raft would be better.

cipherboy:
> As it's pulling the WAL (iirc) it's way easier to configure then non-voters imho. Especially if we think of complex setups like multi-cloud or different kubernetes clusters.

@pree Sorry, could you elaborate on the multi-cloud/cluster issues with Raft non-voter?

I think in the non-voter case (regardless of backend), we could loosen the restrictions here around upgrade order: https://openbao.org/docs/upgrading/ha-upgrade/ -- and state that while it is _preferable_ non-voters upgrade before voters, it is OK to delay them as long as they are never promoted to voting member while on an older version than the last active server.

w.r.t. InfraSec between these, wouldn't the requirements be the same regardless of mechanism? WAL shipping is something not currently supported by the public Raft library and rather involved mechanism, and so would be prone to failure in other ways. This means higher operator burden (monitor two different processes with different failure modes, rather than just one).

But, if the performance of approach 1 is not great and approach 2 has too high of recovery time, I could see how WAL shipping becomes our only option.

---

> In theory this should already be doable with PostgreSQL logical replication today, but of course something built in on top of Raft would be better.

@JanMa Sorry, could you expand on the latter half? Just meaning if you'd like a hybrid Raft+PostgreSQL solution for DR with PostgreSQL backends? Or if PostgreSQL would remain a "pure" DR strategy and you're suggesting that Raft should have its own native DR strategy?

pree:
The issue we see with using PostgreSQL as a backend when it's only required for DR is that you need to managed a HA OpenBao cluster and a HA PostgreSQL cluster, which introduces operational complexity.

> Sorry, could you elaborate on the multi-cloud/cluster issues with Raft non-voter?

With the testing @eyenx did we noticed that you cannot join a non-voter from a different network, especially if DNS is not resolving or the direct IP address is not reachable. This is true for normal kubernetes clusters, or different subnets which are not peered with each other (for example hosting on two different cloud providers).

Example architecture:
image


Bao-2 (green) is the active node. Bao-3 wants to join the Raft cluster and contacts the active node through the LoadBalancers (Multi-DR and Regional). It does not have a direct path to the active node directly (10.0.1.3).
This does work, the active node does receive the request and now trys to contact the non-voter back directly using it's advertised local IP, which is `10.0.2.1`. This does not work as the two networks are not peered with each other.

This issue does not happen if only a single non-voter lives inside Cluster 2, as it can advertise the Cluster 2 LoadBalancer IP as it's own and the traffic gets routed correctly.

So there needs to be some path for each node to be able to contact each node, regardless of in which cluster it lives.

pree:
Another thing to consider:

As we are already putting a lot of work in making the Raft DB transactional safe we could think of also writing a transactional log, which can be stored and replayed from the latest snapshot. IIRC this is how some databases implement DR capabilities.

Issue #1895, opened 2025-10-07
Large CRL


idenkov-tc:
**Describe the bug**
We have openbao cluster with 3 nodes and around 22 namespaces and we are seeing this error:

```
2025-10-06T12:30:20.814Z [ERROR] expiration: failed to revoke lease: lease_id=pki/issue/app_certificates/hT2iyhhjSUKoaxFx7PXR06F3t.3QnP3y error="failed to revoke entry: resp: (*logical.Response)(nil) err: error encountered during CRL building: error building CRLs: error building CRLs: unable to build CRL for issuer (b66a5802-986d-3bfb-3a37-40ada2157719): error storing CRL: put failed due to value being too large; got 2960551 bytes, max: 1048576 bytes" attempts=1 next_attempt=22.101195528s
```

Auto-tidy is not enabled so I tried to run tidy on namespaces that have this pki role(as the namespace is not in the logs so not sure where it comes from exactly), but that didn't worked until we increased max_entry_size to 4MB(4194304).

Now the errors are gone and we can run tidy and possibly enable auto-tidy, but it seems the CRL is still abnormally large, I have tried to export it to check the size, and while it works on some namespaces on others fails as it is too big for audit log to process and right now disabling audit logging is not an option.

```
[ERROR] audit: backend failed to log response: backend=syslog/ error="write unixgram @->/run/systemd/journal/dev-log: write: message too long"

[ERROR] core: failed to audit response: request_path=pki/crl/pem 
```

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.4.1
* OpenBao CLI Version (retrieve with `bao version`): 2.4.1
* Server Operating System/Architecture:  Debian VMs

OpenBao server configuration file(s):

```hcl
ui = true

# Specifies the address to advertise to other OpenBao servers
cluster_addr  = "https://openbao001:8201"

# Specifies the address (full URL) to advertise to other OpenBao servers in the cluster for client redirection
api_addr = "https://openbao.lb.dev-openbao.dc.systems"

enable_response_header_hostname = true

storage "raft" {
  path = "/data/raft"
  node_id = "openbao001"
  retry_join {
    leader_api_addr = "https://openbao001:8200"
  }
  retry_join {
    leader_api_addr = "https://openbao002:8200"
  }
  retry_join {
    leader_api_addr = "https://openbao003:8200"
  }
  max_entry_size = 4194304
}

# Configures how OpenBao is listening for API requests.
listener "tcp" {
[...]

PR #1906, opened 2025-10-09
feat: return namespace, path and revoke error for lease lookup

phil9909:
Example Output:

```
$ bao lease lookup certificates/issue/example-dot-com/JPmYJPmnB1RgGLN5FpjMGEtj.ZoG8sM
Key               Value
---               -----
expire_time       2025-10-12T10:31:14.000106371+02:00
id                certificates/issue/example-dot-com/JPmYJPmnB1RgGLN5FpjMGEtj.ZoG8sM
issue_time        2025-10-09T10:31:14.34655468+02:00
last_renewal      <nil>
namespace_path    ns1/ns2/
path              certificates/issue/example-dot-com
renewable         false
ttl               71h59m45s
```

Security consideration: Is it save to display the revoke error? It could contain sensitive information (while it should not, because it will get logged).

See also https://github.com/openbao/openbao/issues/1895

### DIFF
--- a/changelog/1906.txt
+++ b/changelog/1906.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-lease lookup will now include `path`, `namespace_path` and `revoke_error`
+core/leases: lease lookup will now include `path`, `namespace_path` and `revoke_error`
 ```


changelog/1906.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+lease lookup will now include `path`, `namespace_path` and `revoke_error`
```
cipherboy:
```suggestion
core/leases: lease lookup will now include `path`, `namespace_path` and `revoke_error`
```

phil9909:
> Arguably, you should probably already know the path and namespace, as they would be part of the lease path, but 🤷

You know the namespace id, but I don't think there is an easy way for an user to lookup a namespace by id.

Also, extracting these information relies on undocumented implementation details. Or is this documented somewhere?

wslabosz-reply (approved):
I think it's useful to have the exact namespace name (path) on some of the responses, like the lease here

satoqz:
I don't think there's a reason to hold this any longer, merging...

PR #1907, opened 2025-10-09
docs: add warning to mount and auth move

phil9909:
add warning to not migrate mounts that are actively used.

See #1550
See #1528 

### DIFF
--- a/website/content/docs/commands/auth/move.mdx
+++ b/website/content/docs/commands/auth/move.mdx
@@ -8,16 +8,28 @@ description: |-
 
 # auth move
 
-The `auth move` command moves an existing auth method to a new path. Any
-leases from the old auth method are revoked, but all configuration associated
-with the engine is preserved. The command can be issued for a move within or across
+The `auth move` command moves an existing auth method to a new path. Any leases
+from the old auth method are revoked, but all configuration associated with the
+engine is preserved. The command can be issued for a move within or across
 namespaces, using namespace prefixes in the arguments.
 
-The command will trigger a remount operation and uses the returned migration ID to poll the 
-status of the operation until a terminal state of `success` or `failure` is reached.
+The command will trigger a remount operation and uses the returned migration ID
+to poll the status of the operation until a terminal state of `success` or
+`failure` is reached.
 
-**Moving an existing auth method will revoke any leases from the old
-method.**
+:::warning
+
+Moving an existing auth method will revoke any leases from the old method.
+
+:::
+
+:::warning
+
+Before moving an auth method to a new path, one should ensure it is currently
+not actively used. Using an auth method while it's being moved will lead to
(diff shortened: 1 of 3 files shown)


satoqz:
@phil9909 is there a particular reason why this is suddenly scarier than before with read replication?

website/content/docs/commands/auth/move.mdx
```diff
@@ -8,16 +8,24 @@ description: |-
...
 
-The command will trigger a remount operation and uses the returned migration ID to poll the 
-status of the operation until a terminal state of `success` or `failure` is reached.
+The command will trigger a remount operation and uses the returned migration ID
+to poll the status of the operation until a terminal state of `success` or
+`failure` is reached.
 
-**Moving an existing auth method will revoke any leases from the old
-method.**
+::::warning
```
satoqz: Should this be 3 colons? Or am I missing something?

phil9909: 4 seem to work as well, but the other documents all use 3, so I changed it to 3, thanks.

cipherboy:
@satoqz said:

> @phil9909 is there a particular reason why this is suddenly scarier than before with read replication?

Standby nodes with read replication will be able to respond to requests before the mount is invalidated. This means you can see (from standby PoV) this order:

1. Mount table storage updated to taint mount
2. Storage started to be removed
3. (invalidation occurs here, not yet processed)
4. Request comes in that reads or list, on data that is not cached, and thus sees partially deleted storage state
   -  This can fail with API errors most commonly. (warning is about this)
5. Invalidation is processed tainting mount.  

4 is what we want this warning to apply to, as standby behavior differs from active node in this case. Luckily it is not data _corruption_ (standbys still can't write to data), just bad results. 

(3 and 4 can be swapped order, but the main point is that storage update isn't synchronous with mount tainting and invalidation order may not be linearized/serial fwiw)

satoqz:
@phil9909 @cipherboy I'd modify the warning to be significantly more specific in that case. It currently sounds really, really scary 😄

- Mention that this only applies when using read replicas (and _on_ read replicas only)
- Mention that the "undefined behavior" equals inconsistent reads, not corruption of writes

phil9909:
> Mention that this only applies when using read replicas (and _on_ read replicas only)

read replication will be on by default.

How about this:

Before migrating a mount to a new path, one should ensure it is currently not actively used. Using a mount while it's being moved will lead to inconsistent reads on standby nodes.

phil9909:
> Before migrating a mount to a new path, one should ensure it is currently not actively used. Using a mount while it's being moved will lead to inconsistent reads on standby nodes.

Updated the three warnings to this phrasing.

cipherboy (approved):
I think the new language is less alarmist. Thanks @phil9909!

PR #1910, opened 2025-10-10
Sunset aead seal usage

### LINKED ISSUE #1876: Deprecate AEAD seal mechanism
As far as I can tell historically, nobody has used the `seal "aead"` mechanism. This is undocumented (both upstream and here in OpenBao) and behaves like our new `static` seal mechanism, except it lacks native key rotation support. This means that in order to rotate keys, one must use seal migration between two different `seal "aead"` stanzas, performing a manual unseal in the process using recovery keys.

This is less than ideal UX and thus we should prefer the documented `static` seal mechanism instead.

Given the lack of documentation and usage on GitHub (unlike every other seal mechanism), we can likely sunset this in v2.5.0 without notice.

wslabosz-reply:
Resolves #1876 

- Removed `aead` seal config parsing case from `configureWrapper` and the parsing function `GetAEADKMSFunc()`

### DIFF
--- a/changelog/1910.txt
+++ b/changelog/1910.txt
@@ -1,0 +1,3 @@
+```release-note:deprecation
+core/seal: Remove the undocumented "aead" seal mechanism. Consider switching to the static seal instead as a replacement.
+```


changelog/1910.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:deprecation
+core/seal: remove support for 'aead' seal wrapper.
```
satoqz:
```suggestion
core/seal: Remove the undocumented "aead" seal mechanism. Consider switching to the static seal if the aead seal was in use for unlikely reasons.
```

to be more user-facing, and explain why + alternatives.

PR #1913, opened 2025-10-11
deps(lz4): Update to mod compatible version

mikelolasagasti:
SDK was importing a non-versioned module incompatible with Go modules.


satoqz:
@cipherboy I wonder if we can drop LZ4 altogether? The only real reference is in the `logical_raw.go` read path. It seems the only subsystem that ever _wrote_ LZ4 was the activity log (removed 1.5 years ago)?

LZW also seems like a candidate for removal, the only code path it is used in (`helper/forwarding/util.go`) appears to be dead code.

cipherboy:
@satoqz Works for me.

cipherboy:
@satoqz works for me.

Thanks @mikelolasagasti!

PR #1914, opened 2025-10-12
Add ML-DSA to sdk and transit

f-hc:
Add ML-DSA (also known as CRYSTALS-Dilithium, and FIPS 204) to sdk and transit

## How to create a key:
payload.json:
```json
{"type": "ml-dsa", "derived": "false", "parameter_set": "87"}
```

run:
```bash
curl --header "X-Vault-Token: root" --request POST --data @payload.json http://127.0.0.1:8200/v1/transit/keys/my-key
```

Part of #496

### DIFF
--- a/website/content/api-docs/secret/transit.mdx
+++ b/website/content/api-docs/secret/transit.mdx
@@ -65,6 +65,7 @@ values set here cannot be changed after key creation.
   - `rsa-3072` - RSA with bit size of 3072 (asymmetric)
   - `rsa-4096` - RSA with bit size of 4096 (asymmetric)
   - `hmac` - HMAC (HMAC generation, verification)
+  - `ml-dsa` - ML-DSA (asymmetric)
 
 - `key_size` `(int: "0", optional)` - The key size in bytes for algorithms
   that allow variable key sizes.  Currently only applicable to HMAC, where
@@ -74,6 +75,9 @@ values set here cannot be changed after key creation.
   this key should be rotated automatically. Setting this to "0" (the default)
   will disable automatic key rotation. This value cannot be shorter than one
   hour. Uses [duration format strings](/docs/concepts/duration-format).
+
+- `parameter_set` `(string:"")` - The parameter set to use for ML-DSA. Required
+  for ML-DSA. Valid values are `44`, `65`, and `87`.
 
 ### Sample payload
 


sdk/helper/keysutil/policy.go
```diff
@@ -177,6 +188,15 @@ func (kt KeyType) AssociatedDataSupported() bool {
...
 }
 
+func (kt KeyType) IsPQC() bool {
+	switch kt {
+	case KeyType_ML_DSA:
+		return true
+	default:
+		return false
+	}
+}
```
satoqz: This appears unused, did you have a use case in mind?

f-hc:
> This looks good to me FWIW, but my experience with Transit and PQC is limited at best.
> 
> I do think some tests would be nice, likely in these locations:
> 
> * `builtin/logical/transit/path_keys_test.go`
> * `helper/keysutil/policy_test.go`

Working on it

sdk/helper/keysutil/policy.go
f-hc: I missed to remove it, it was part of my POC for the Hybrid (ECDSA + PQC) but I won't use it

cipherboy:
Hey @f-hc! I've expressed this in a few places, though, apparently not on #496 which is unfortunate... but I think waiting until Go toolchain natively supports ML-DSA & friends is perhaps our best approach.

As much as I appreciate the work of the Cloudflare team, ultimately the [Go team will only face pressure to add it](https://github.com/golang/go/issues/64537) when the community declines to use third-party libraries for it. Ultimately, using a third-party crypto library isn't feasible, as `crypto/x509` isn't pluggable and so many places require it externally. So we _need_ the Go standard library to support these algorithms as they're unlikely willing to make x509 externally pluggable.

While we could enable thus merge this for Transit, I think Transit has ultimately best served as a Go standard crypto library package.

(There's also a tangential discussion of if we should be using circl or liboqs, a fellow Linux Foundation project and consortium of many other vendors.)

All this to say, I don't know what I think. I was hoping, at the time I wrote that RFC initially, that PQC would be within the next year or so. It has now been over a year and we've not seen positive progress.

---

The actionable part, though, is ML-KEM if you're interested in that instead! That's already part of the standard library, and some discussion of key exchange has occurred, e.g., in the context of #655.

f-hc:
Hello @cipherboy, thank you for the clarification.
I'll convert this PR to draft and be waiting for the RFC.

cipherboy:
> Hello @cipherboy, thank you for the clarification. I'll convert this PR to draft and be waiting for the RFC.

@f-hc @DanGhita If there's general appetite for a PQC ahead of Go standard library... We do have `openbao-plugins`. :thinking: 

We could fork Transit+keysutil into a PQC-only (experimental?) plugin, that could ultimately be folded into this version, depending freely on liboqs or circl or whatever we want.

My biggest worry is that key storage might differ between Go and liboqs/... (and I've not been overly happy with how keysutil does this today anyways for that matter), so there may not be a clean upgrade path.

At any rate, it'd let us iterate towards a long-term solution, though perhaps requiring an import+export operation once this version is improved.

Thoughts?

Issue #1917, opened 2025-10-14
Explicity specifying BAO_NAMESPACE=root result in a 404


Nerkho:
**Describe the bug**

When explicitly setting the namespace to `root` (e.g. with env var `BAO_NAMESPACE=root`), the request results in a 404.

**To Reproduce**
1. `export BAO_NAMESPACE=root`
2. `bao login`

```bash
❯ bao login
Token (will be hidden): 
Error authenticating: error looking up token: Error making API request.

URL: GET http://127.0.0.1:8200/v1/auth/token/lookup-self
Code: 404. Errors:

* namespace not found
```

**Expected behavior**

Setting the namespace to `root` should redirect requests to the root namespace.

**Workaround**

Using `BAO_NAMESPACE=/` or `BAO_NAMESPACE=""` instead works as expected.

cipherboy:
@Nerkho @phil9909 Is this supported by Vault?

Nerkho:
@cipherboy Yup, setting `VAULT_NAMESPACE=root` works in Vault.

cipherboy:
This is confirmed by this documentation: https://support.hashicorp.com/hc/en-us/articles/16772116543507-Prevent-Vault-from-Brute-Force-Attack-User-Lockout

PR #1918, opened 2025-10-14
Support BAO_NAMESPACE=root

phil9909:
Fixes: https://github.com/openbao/openbao/issues/1917

This PR also add some tests for behavior inherited from the usage of `path.Clean` (e.g. turning `a//////b/` into `a/b/`). If we later want to change the implementation, we should make sure we don't change quirks like that, because someone somewhere will rely on it (e.g. because the string concatenate two namespaces).

Also, `Canonicalize` is additionally tested for idempotency (by calling it a second time for every testcase, checking the result does not change).

### DIFF
--- a/changelog/1918.txt
+++ b/changelog/1918.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-Setting the `X-Vault-Namespace` Header (or the `BAO_NAMESPACE` environment variable when using the cli) to "root" now maps to the root namespace.
+core/namespaces: Setting the `X-Vault-Namespace` Header (or the `BAO_NAMESPACE` environment variable when using the cli) to "root" now maps to the root namespace.
 ```


changelog/1918.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+Setting the `X-Vault-Namespace` Header (or the `BAO_NAMESPACE` environment variable when using the cli) to "root" now maps to the root namespace.
```
cipherboy:
```suggestion
core/namespaces: Setting the `X-Vault-Namespace` Header (or the `BAO_NAMESPACE` environment variable when using the cli) to "root" now maps to the root namespace.
```

helper/namespace/namespace.go
```diff
@@ -257,16 +257,16 @@ func HeaderFromContext(ctx context.Context) string {
...
-
 	// Canonicalize the path to not have a '/' prefix
 	nsPath = strings.TrimPrefix(nsPath, "/")
 
 	// Remove duplicate slashes and any ../ values if present.
 	nsPath = path.Clean(nsPath)
 
+	if nsPath == "." || nsPath == "root" {
+		return ""
+	}
```
satoqz, line 268: How does upstream behave for

```
X-Vault-Namespace: root/foo/bar/
```

?

We would parse `root/foo/bar/`, but should it be `foo/bar/`?

Other cases that I'm unsure about: `root/root/`, `root/foo/root/`, ...

driif: In OpenBao - "root" is a reserved path and cannot be used as a namespace name - which I find in general optimal, even if it's not compatible with upstream.

satoqz (approved):

PR #1919, opened 2025-10-14
Fix some errcheck lint findings

klaus-sap:
Resolves one file from https://github.com/openbao/openbao/issues/1574


cipherboy (approved):
Thanks @klaus-sap!

PR #1920, opened 2025-10-14
Add tlsdebug build tag to log TLS session keys

phil9909:
This enables decrypting TLS traffic e.g. with Wireshark (see https://wiki.wireshark.org/TLS#using-the-pre-master-secret), which can be useful for debugging.

It can be enabled with `dev-tlsdebug` or by setting the `tlsdebug` build tag using another method.

The path of the tls session key log file will be printed to the log.

It tested it for request forwarding and raft internal communication. Any connection that wants to make use of this has to call `tlsdebug.Inject` passing its `tls.Config` as a parameter (which is a no-op when tlsdebug is disabled). I didn't do this for other tls contexts like the ACME plugin as I have no experience on how to test this.

### DIFF
--- a/helper/tlsdebug/inject.go
+++ b/helper/tlsdebug/inject.go
@@ -1,0 +1,31 @@
+//go:build tlsdebug
+
+package tlsdebug
+
+import (
+	"crypto/tls"
+	"os"
+	"sync"
+
+	"github.com/hashicorp/go-hclog"
+)
+
+var openKeyLogFile = sync.OnceValues(func() (*os.File, error) {
+	return os.CreateTemp("", "openbao_tls_session_key_log_")
+})
+
+func Inject(logger hclog.Logger, config *tls.Config) *tls.Config {
+	if config == nil {
+		return nil
+	}
+
+	file, err := openKeyLogFile()
+	if err != nil {
+		logger.Error("could not open TLS key log file", "error", err)
+	}
+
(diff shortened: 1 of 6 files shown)


helper/tlsdebug/inject.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"crypto/tls"
+	"os"
+	"sync"
+
+	"github.com/hashicorp/go-hclog"
+)
+
+var openKeyLogFile = sync.OnceValues(func() (*os.File, error) {
+	return os.CreateTemp("", "openbao_tls_session_key_log_")
```
cipherboy, line 14: Do you want to make this an environment variable? 🤔 It seems like `pwd` may not be writable in all circumstances.

(thinking, debugging TLS issues in a k8s environment)

JanMa: This will not create a file in `pwd` but most likely at `/tmp`. When running OpenBao in K8S this will usually be writable unless the user explicitly enables a read-only root FS. And then they can still mount a volume at `/tmp` to make it writable

JanMa (approved):
LGTM

helper/tlsdebug/inject.go
cipherboy: Ah, d'oh, my bad. So it will!

JanMa:
@cipherboy do you think this needs a changelog?

PR #1923, opened 2025-10-15
feat(valkey): Support connection_url for configuration

### LINKED ISSUE #1812: Add connection_url to Valkey create connection parameters

In the database secrets engine API some connection strings could be unified in order to simplify the API for creating a connection. Both PostgreSQL and MySQL/MariaDB offer a `connection_url` as parameter. This parameter is also described in the API overview as a [common field](https://openbao.org/api-docs/secret/databases/#common-fields). In PostgreSQL, for example, the connection has the following form

```
postgres[ql]://[username[:password]@][host[:port],]/database[?parameter_list]

\_____________/\____________________/\____________/\_______/\_______________/
     |                   |                  |          |            |
     |- schema           |- userspec        |          |            |- parameter list
                                            |          |
                                            |          |- database name
                                            |
                                            |- hostspec
```

Other databases could adopt this common field with their respective server URI. For example, Valkey's uri format is 

```
valkey://user:password@host:port/dbnum
```

or 

```
valkeys://user:password@host:port/dbnum
```

Currently, Valkey requires you to pass the `host` and `port` separately as mandatory parameters, and does not offer a connection URL template like PostgreSQL does. In Postgres you pass the user and password as template parameters:

- `connection_url` `(string: <required>)` - Specifies the PostgreSQL DSN. This field can be templated and supports passing the username and password parameters in the following format `{{field_name}}`. Certificate authentication can be used by setting `?sslmode=` to be any of the applicable values as outlined in the [Postgres SQL documentation](https://www.postgresql.org/docs/11/libpq-ssl.html#LIBPQ-SSL-PROTECTION) and giving the SSL credentials in the `sslrootcert`, `sslcert` and `sslkey` credentials. A templated connection URL is required when using root credential rotation. This field supports both format string types, URI and keyword/value. Both formats support multiple host connection strings. Due to how `pgx` works, parameters such as `sslrootcert`, `sslcert`, `sslkey` are treated as paths on the OpenBao server.

Offer a `connection_url` parameter in Valkey instead of passing `host` and `port` and separately. 

Keep it like it is 

Unifying the creation of connection between openbao and databases helps openbao-clients to implement fewer changes based on different databases.  This could also help improve the interface of the database plugin API.

**Additional context**

I'm developing a [swift client for OpenBao](https://github.com/vault-courier/vault-courier) and while adding support to Valkey, I noticed this.

kyounghunJang:
Adds the ability to configure the Valkey database connection using a single `connection_url` parameter. This provides a unified and convenient alternative to specifying individual parameters like `host`, `port`, `username`, and `password`.

A new test case is added to validate the initialization via `connection_url`.

Resolves #1812

### DIFF
--- a/changelog/1923.txt
+++ b/changelog/1923.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+database/valkey: Adds the ability to configure the Valkey database connection using a single connection_url parameter.
+```
--- a/plugins/database/valkey/connection_producer.go
+++ b/plugins/database/valkey/connection_producer.go
@@ -9,6 +9,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"net"
+	"net/url"
 	"strconv"
 	"sync"
 
@@ -18,13 +19,14 @@ import (
 )
 
 type valkeyDBConnectionProducer struct {
-	Host        string `json:"host"`
-	Port        int    `json:"port"`
-	Username    string `json:"username"`
-	Password    string `json:"password"`
-	TLS         bool   `json:"tls"`
-	InsecureTLS bool   `json:"insecure_tls"`
-	CACert      string `json:"ca_cert"`
+	Host          string `json:"host"`
+	Port          int    `json:"port"`
+	Username      string `json:"username"`
+	Password      string `json:"password"`
+	TLS           bool   `json:"tls"`
+	InsecureTLS   bool   `json:"insecure_tls"`
+	CACert        string `json:"ca_cert"`
+	ConnectionURL string `json:"connection_url"`
 
 	Initialized bool
 	rawConfig   map[string]interface{}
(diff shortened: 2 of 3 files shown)


kyounghoonJang:
@wslabosz-reply 
Thanks for the feedback! I’ve added the changelog entry as requested.

cipherboy (approved):
@jcubit Do you want to give this a try?

Thanks @kyounghunJang!

jcubit:
@cipherboy @kyounghunJang thanks! Yes, I'd love to. I'll give you a feedback at latest during the weekend

cipherboy:
Since v2.5.0 is far enough out, I'll merge this in the interim. Thanks again!

jcubit:
One idea to explore in the future would be if it makes sense to also use "templated parameters" in the url_connection as done in the postgres plugin:

```
vault write database/config/my-postgresql-database \
    plugin_name="postgresql-database-plugin" \
    allowed_roles="my-role" \
    connection_url="postgresql://{{username}}:{{password}}@localhost:5432/database-name" \
    username="vaultuser" \
    password="vaultpass" \
    password_authentication="scram-sha-256"
```

Right now, we error with "cannot specify both connection_url and individual connection parameters". I'm not sure if it's an  enhancement or it's more secure with the template parameters.

Thanks again for your contribution @kyounghunJang !

PR #1926, opened 2025-10-15
KV engine returns timestamp strings in RFC3339Nano format (#1872)

phil9909:
Backport of #1872 by @wslabosz-reply 

Clean cherry-pick


satoqz:
Manually set DCO to pass.

PR #1927, opened 2025-10-15
Identity subsystem returns timestamp strings in RFC3339Nano format (#1873)

phil9909:
Backport of #1873 by @wslabosz-reply 

Clean cherry-pick


satoqz:
Manually set DCO to pass.

satoqz:
Ah, Cassandra. @phil9909 thoughts on backporting https://github.com/openbao/openbao/pull/1762? Not strictly necessary as we can just bypass, but gets the failures out of the way by default.

PR #1929, opened 2025-10-15
Move armon/go-metrics -> hashicorp/go-metrics/compat

satoqz:
Second part of my attempt to unblock dependabot, i.e., resolve this error: https://github.com/openbao/openbao/actions/runs/18524597614/job/52792268572

Changes:

- Replace `github.com/armon/go-metrics` with `github.com/hashicorp/go-metrics`
- Replace `cloud.google.com/go/monitoring/apiv3` with `cloud.google.com/go/monitoring/apiv3/v2`, this was required such that the above change compiles again
- Some additional formatting noise because `make fmt` felt like it

As per https://github.com/hashicorp/go-metrics/releases/tag/v0.5.0, we should ensure that all code uses the same metrics sink. I believe we are good here - the only remaining (indirect) reference to `armon/go-metrics` runs through `hashicorp/go-metrics`'s compat module.

```
$ go mod why github.com/armon/go-metrics
# github.com/armon/go-metrics
github.com/openbao/openbao/helper/testhelpers
github.com/hashicorp/raft
github.com/hashicorp/go-metrics/compat
github.com/armon/go-metrics
```


satoqz:
> As per https://github.com/hashicorp/go-metrics/releases/tag/v0.5.0, we should ensure that all code uses the same metrics sink. I believe we are good here - the only remaining (indirect) reference to `armon/go-metrics` runs through `hashicorp/go-metrics`'s compat module.
> 
> ```
> $ go mod why github.com/armon/go-metrics
> # github.com/armon/go-metrics
> github.com/openbao/openbao/helper/testhelpers
> github.com/hashicorp/raft
> github.com/hashicorp/go-metrics/compat
> github.com/armon/go-metrics
> ```

Mhh, nevermind, I don't think this would route correctly according to https://github.com/hashicorp/go-metrics?tab=readme-ov-file#backwards-compatibility, as the default compat sink is still the `armon` namespace. Could lean on the build tag, but that'll be an inconvenience to literally every packager ever.

wslabosz-reply (approved):
looks clean to me 🟢 
EDIT: I see the compatibility issue, to be resolved then...

phyrog:
> A subsequent release after that will eventually remove the compatibility layer all together. The rough timeline for this will be mid-2025 for changing the default behavior and then the end of 2025 for removal of the compatibility layer.

So we'll just have to wait for them to change the default and/or remove the compat layer.

satoqz:
> So we'll just have to wait for them to change the default and/or remove the compat layer.

Yup, that's what this PR is effectively tracking now.

satoqz:
As per suggestion by @phyrog I've updated this to point at `hashicorp/go-metrics/compat` instead of going to `hashicorp/go-metrics` right away.

This gets us an intermediate step towards removing `armon/go-metrics` (it becomes an indirect dependency), and is compatible _right now_. Once the default behavior of the `compat` package changes, we're unaffected as all components would use the new sink automatically. Once the `compat` package is removed (and there are no remaining dependencies that import it, hopefully) we can just update our imports again.

satoqz:
bump @JanMa @cipherboy!

cipherboy (approved):
Apologies about the delay, LGTM!

PR #1931, opened 2025-10-15
Fix backup keys retrieval for namespace seals

wslabosz-reply:
This PR:
- removes `namespace` field from openapi schema and actual responses to the `rotate` adjacent operations for namespaces
- adds missing `namespace` field to the `"namespaces/(?P<namespace>.+)/rotate/(root|recovery)/backup"` endpoint, which made the endpoint unresponsive
- fixes `pgpEncryptShares` always saving the backup keys to the `unseal` path (essentially rendering the backup of recovery shares unusable)
- renames `command/rotate.go` to `command/operator_rotate.go` for consistency

### DIFF
--- a/api/sys_rotate.go
+++ b/api/sys_rotate.go
@@ -23,7 +23,6 @@ type RotateInitRequest struct {
 }
 
 type RotateStatusResponse struct {
-	Namespace            string   `json:"namespace"`
 	Complete             bool     `json:"complete,omitempty"`
 	Keys                 []string `json:"keys,omitempty"`
 	KeysB64              []string `json:"keys_base64,omitempty"`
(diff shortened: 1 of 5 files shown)


api/sys_rotate.go
```diff
@@ -23,6 +23,7 @@ type RotateInitRequest struct {
 }
 
 type RotateStatusResponse struct {
+	Namespace            string   `json:"namespace"`
```
satoqz: Repeating the question I asked internally: Why do we want the namespace to be included in the response?

wslabosz-reply: I think it was an unnecessary addition, will look into removal

wslabosz-reply:
merging to feature branch based on approvals from @satoqz & @cipherboy

PR #1937, opened 2025-10-15
Add check that builds bao with the go on el8

DrDaveD:
This adds an optional linter check to verify that openbao will work with the golang version that is current on el8, even if go.mod has been set to a later go version. This is needed for EPEL packaging to be successful.

Replaces #1845.

### DIFF
--- a/.github/workflows/code-checker.yml
+++ b/.github/workflows/code-checker.yml
@@ -97,6 +97,8 @@ jobs:
 
   el8gobuild:
     # make sure that bao can still be built with the golang version on el8
+    # unless there is a `no-el8gobuild` label
+    if: "!contains(github.event.pull_request.labels.*.name, 'no-el8gobuild')"
     name: EL8 Go build checks
     runs-on: ubuntu-latest
     container: ghcr.io/almalinux/8-base


satoqz:
@DrDaveD @cipherboy What happens when we _do_ introduce changes that cause builds on EL8's Go version to fail, but potentially weeks or months before releasing to GA? E.g., for a beta release, or even way before the beta release when merging new features? It seems to me we would be stuck with a red CI check to ignore for quite some time, which isn't great at all -- even if at final GA release time, EL8 may be compatible again.

I'm not sure what the best solution would be either, but it seems to me this should be tied closer to the actual release cycles _somehow_.

cipherboy:
@satoqz said:

> I'm not sure what the best solution would be either, but it seems to me this should be tied closer to the actual release cycles _somehow_.

Given our release is... _flexibly_ scheduled, I think the red check is an important sign: we can't release (now) without breaking RHEL. :shrug: 

I don't like red CI any more than the next person, but I think it is accurate at least.

DrDaveD:
Maybe we need a label to ignore it like we do with changelog?

satoqz:
@DrDaveD good call, maybe the inverse? Such that the label would trigger the check, and one can just add it whenever there's uncertainty about compatibility. Either way around would be a good solution I think.

DrDaveD:
I think having an inverse label would defeat the purpose of the check so I don't go for that.

DrDaveD:
I have added a `no-el8gobuild` optional label to skip the check.  Sound good?

cipherboy:
Based on above, merging. Thanks @DrDaveD!

PR #1950, opened 2025-10-16
read replication: fix some tests

phil9909:
This PR fixes some test failures caused by read replication. See also:  #1932

See #1550
See #1528 


cipherboy:
Thanks @suprjinx for the review!

PR #1958, opened 2025-10-16
fix:  False “existing mount” conflict when mounting a backend at a path equal to the current namespace name.

### LINKED ISSUE #1957: Mount path equal to namespace name incorrectly conflicts with namespace sys/ (“existing mount at …”
# Description
Mounting a secrets engine inside a namespace at a path equal to the namespace name incorrectly returns a 409 conflict: “existing mount at <namespace>/”. This is a false positive caused by the router’s namespace-conflict detection treating the namespace’s own sys/ sentinel as a conflicting mount.

# To Reproduce

- Create a namespace:
`bao write -force sys/namespaces/team1`
- In that namespace, try to mount a secrets engine at a path equal to the namespace name: `BAO_NAMESPACE=team1 bao secrets enable -path=team1 kv`

See error:
```
Error enabling: Error making API request.

Namespace: team1/
URL: POST https://localhost:8200/v1/sys/mounts/team1
Code: 400. Errors:

* existing mount at team1/
```
Note: A similar conflict may appear when enabling auth at -path=team1.

# Expected behavior
Mounting a backend at a path equal to the namespace name (within that same namespace) should succeed.
Only true conflicts (e.g., overlapping an existing mount or colliding with a child namespace path) should be rejected.

# Environment:
OpenBao Server Version (retrieve with bao status): Version 2.4.1
OpenBao CLI Version (retrieve with bao version): OpenBao v2.4.1 (refs/tags/v2.4.1) (cgo)
Server Operating System/Architecture: Docker (arm64)
OpenBao server configuration file(s): nil

# Probable Root cause: 
`vault/router.go`: matchingNamespaceInternal checks for namespace conflicts without scoping the candidate path to the current namespace and does not ignore the current namespace’s own sys/ mount. This leads to a mount path equal to the namespace name being mistaken for a conflict with the namespace sentinel.

dc-tec:
# Summary
Fix false “existing mount” conflict when mounting a backend at a path equal to the current namespace name. Align namespace conflict detection with other router checks by scoping the input path to the current namespace and ignoring the current namespace’s own sys/ sentinel.

# Changes
`vault/router.go`
- Prefix candidate path with the current namespace (ns.Path + path) in matchingNamespaceInternal.
- Ignore the current namespace’s own sys/ entry when scanning for conflicts to avoid false positives.

`vault/router_test.go`
Add TestRouter_NamespaceNameMount_NoConflict to cover:
- No conflict for a mount path equal to the namespace name within that namespace.
- Conflict for a child namespace path.

# Rationale
- Other matching helpers already compare against namespace-scoped absolute paths; matchingNamespaceInternal was inconsistent.
- The sys/ mount is a sentinel for namespace presence and should not be considered a conflicting mount against a mount path that equals the namespace name.

# Behavioral Impact
- Fixes false conflicts when the mount path equals the namespace name in that namespace.
- No change to behavior for valid conflicts (e.g., mounting under an existing mount or conflicting with child namespaces).

Resolves #1957

### DIFF
--- a/vault/router.go
+++ b/vault/router.go
@@ -419,17 +419,17 @@ func (r *Router) matchingNamespaceInternal(ctx context.Context, path string) str
 	// locked as we're trying to mount required mounts for a new namespace.
 	var existing string
 	fn := func(existingPath string, v interface{}) bool {
-		if !strings.HasSuffix(existingPath, "sys/") {
+		nsPath, ok := strings.CutSuffix(existingPath, "sys/")
+		if !ok {
 			return false
 		}
 
 		// Ignore the current namespace's own sys mount; we only want to
 		// detect conflicts with child namespace prefixes.
-		if existingPath == ns.Path+"sys/" {
+		if nsPath == ns.Path {
 			return false
 		}
 
-		nsPath := strings.TrimSuffix(existingPath, "sys/")
 		if strings.HasPrefix(path, nsPath) {
 			existing = nsPath
 			return true


dc-tec:
Tested it, and it now works as intented:

image

changelog/1958.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+vault/router: Fix namespace-aware mount conflict detection to ignore namespace's own sys mount
```
satoqz:
```suggestion
core/namespaces: Fix mount creation failing if mount name is equal to the name of the containing namespace
```

I think this would be more helpful to a user that's not familiar with internals.

satoqz (commented):
~~I can't seem to find a scenario where `matchingNamespaceInternal` is even called by `MountConflict` when there is a legit conflict, as `matchingPrefixInternal` should return a result first, no?~~

Edit: Ah, found the scenario, looking at `TestNamespaceMount_Exclusion` helped.

vault/router.go
```diff
@@ -417,6 +423,12 @@ func (r *Router) matchingNamespaceInternal(ctx context.Context, path string) str
 			return false
 		}
 
+		// Ignore the current namespace's own sys mount; we only want to
+		// detect conflicts with child namespace prefixes.
+		if existingPath == ns.Path+"sys/" {
+			return false
+		}
+
 		nsPath := strings.TrimSuffix(existingPath, "sys/")
```
satoqz:
```suggestion
		nsPath, ok := strings.CutSuffix(existingPath, "sys/")
		if !ok {
			return false
		}

		// Ignore the current namespace's own sys mount; we only want to
		// detect conflicts with child namespace prefixes.
		if nsPath == ns.Path {
			return false
		}
```

This could be simpler with `strings.CutSuffix`.

satoqz (approved):
Makes sense to me overall.

vault/router.go
satoqz: (Suggestion won't apply as-is, this is meant to be inserted from the start of the closure's body)

changelog/1958.txt
dc-tec: Agree, thanks for the suggestion.

PR #1959, opened 2025-10-16
fix(valkey): Correctly parse creation_statements as a string array

### LINKED ISSUE #1813: Wrong format in creation_statements parameter in Valkey plugin
**Describe the bug**

The `creation_statements` parameter of the AppRole statements in Valkey has a strange format which might be an oversight from the redis implementation (?). If you send an array of strings as input it will fail on decoding if the string itself is not a string of an array.

Sending an array of strings of arrays work, but this seems to be an unnecessary format (and kind of differs from the SQL databases as well). The problem seems here

https://github.com/openbao/openbao/blob/9082e0eb9fa15cd71935d4b645e0e322e882bbdb/plugins/database/valkey/valkey.go#L142

As you can see only the first parameter of the input is taken. Valkey allows for multi-value ACL rules and are read from left to right as the valkey.conf templates states:

```
# ACL users are defined in the following format:
#
#   user <username> ... acl rules ...
#
# For example:
#
#   user worker +@list +@connection ~jobs:* on >ffa9203c493aa99
#
# 
# ACL rules can be specified in any order: for instance you can start with
# passwords, then flags, or key patterns. However note that the additive
# and subtractive rules will CHANGE MEANING depending on the ordering.
# For instance see the following example:
#
#   user alice on +@all -DEBUG ~* >somepassword
#
# This will allow "alice" to use all the commands with the exception of the
# DEBUG command, since +@all added all the commands to the set of the commands
# alice can use, and later DEBUG was removed. However if we invert the order
# of two ACL rules the result will be different:
#
#   user alice on -DEBUG +@all ~* >somepassword
#
# Now DEBUG was removed when alice had yet no commands in the set of allowed
# commands, later all the commands are added, so the user will be able to
# execute everything.
#
# Basically ACL rules are processed left-to-right.
```

An array of ACL strings as input for `creation_statements` should be enough

**To Reproduce**
1. Create a valkey 
2. With this payload 
```json
{
  "db_name": "valkey_connection_name",
  "creation_statements": [
    "+@admin"
  ],
  "default_ttl": "1h",
  "max_ttl": "24h"
}
```
create a valkey dynamic role

```
curl \
    --request POST \
    --data @payload.json \
    http://127.0.0.1:8200/v1/database/roles/my-role

```
   
2. Run a generation of credentials (this triggers the user creation)
   
```
curl \
http://127.0.0.1:8200/v1/database/creds/my-role
```
   
3. See error
   
```json
{"errors":["1 error occurred:\n\t* error unmarshalling VALKEY rules in the creation statement JSON: invalid character '+' looking for beginning of value\n\n"]}
```
   

**Expected behavior**

An array of ACL strings as input for `creation_statements` should work or a more descriptive error message.

**Environment:**
```
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    1
Threshold       1
Version         2.4.0
Build Date      2025-08-28T20:33:02Z
Storage Type    inmem
Cluster Name    vault-cluster-9630e143
Cluster ID      dd438a08-003e-b4a1-3af6-d93a32ba1e1f
HA Enabled      false

### PR DESCRIPTION
**Pull request by kyounghunJang**
This commit fixes the issue by implementing a fallback mechanism:
1. It first attempts to unmarshal the statement as JSON to maintain backward compatibility for any configurations using the old format.
2. If unmarshaling fails, it logs a warning and treats the entire `creation_statements` input as a raw array of ACL rules, which is the new, expected behavior.

Resolves #1813

### DIFF
--- a/plugins/database/valkey/valkey_test.go
+++ b/plugins/database/valkey/valkey_test.go
@@ -274,7 +274,9 @@ func testValkeyDBCreateUser(t *testing.T, address string, port int) {
 		t.Fatalf("err: %s", err)
 	}
 
-	db.Close()
+	if err := db.Close(); err != nil {
+		t.Fatalf("failed to close db: %v", err)
+	}
 
 	if err := checkCredsExist(t, userResp.Username, password, address, port); err != nil {
 		t.Fatalf("Could not connect with new credentials: %s", err)
@@ -317,6 +319,7 @@ func testValkeyDBCreateUser_WithCreationStatements(t *testing.T, address string,
 	}
 
 	db := new()
+
 	_, err := db.Initialize(context.Background(), initReq)
 	if err != nil {
 		t.Fatalf("Failed to initialize database: %s", err)
@@ -333,13 +336,19 @@ func testValkeyDBCreateUser_WithCreationStatements(t *testing.T, address string,
 			RoleName:    "test",
 		},
 		Statements: dbplugin.Statements{
-			Commands: []string{"+@read"},
+			Commands: []string{"+@read", "~*"},
 		},
 		Password:   password,
 		Expiration: time.Now().Add(time.Minute),
 	}
 
 	userResp, err := db.NewUser(context.Background(), createReq)
+
+	defer func() {
+		if err := db.Close(); err != nil {
+			t.Fatalf("failed to close DB connection during cleanup: %v", err)
+		}
+	}()
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}

### CODE REVIEW

File: plugins/database/valkey/valkey_test.go [hunk 1]
```diff
@@ -273,14 +327,13 @@ func testValkeyDBCreateUser(t *testing.T, address string, port int) {
...
-
 	createReq := dbplugin.NewUserRequest{
 		UsernameConfig: dbplugin.UsernameMetadata{
 			DisplayName: "test",
 			RoleName:    "test",
 		},
 		Statements: dbplugin.Statements{
-			Commands: []string{},
+			Commands: []string{"+@read"},
 		},
```
wslabosz-reply, line 340: wasn't the point of this test to verify that `["+@read", "~*"]` (array of statements, parsed as is from left to right) is also allowed?

plugins/database/valkey/valkey_test.go
```diff
@@ -214,29 +214,83 @@ func testValkeyDBInitialize_TLS(t *testing.T, host string, port int) {
...
-func testValkeyDBCreateUser(t *testing.T, address string, port int) {
+func testValkeyDBCreateUser_WithCreationStatements(t *testing.T, address string, port int) {
 	if api.ReadBaoVariable("BAO_ACC") == "" {
 		t.SkipNow()
 	}
-
-	t.Log("Testing CreateUser()")
+	t.Log("Testing CreateUser() with creation statements")
 
 	connectionDetails := map[string]interface{}{
```
wslabosz-reply, line 297: don't know why the linter shows an error here... but it's probably correct to close the db connection (and check the error while at it)

plugins/database/valkey/valkey_test.go
kyounghoonJang: @wslabosz-reply 
Thanks for the feedback! I assumed that once array handling was supported, multiple values would naturally be processed as well, so I wrote the test that way. I realize that was a bit careless on my part. I’ve updated the test to make it clearer.
Thank you!

kyounghoonJang: @wslabosz-reply 
Thanks for the feedback! I assumed that once array handling was supported, multiple values would naturally be processed as well, so I wrote the test that way. I realize that was a bit careless on my part. I’ve updated the test to make it clearer.
Thank you!

wslabosz-reply: no problem, thanks for clarifying!

plugins/database/valkey/valkey_test.go
kyounghoonJang: @wslabosz-reply 
I added error checking for the Close() call, but the linter error still seems to appear. I’m not sure why this is happening… Should I also update other Close() calls as well?

kyounghoonJang: @wslabosz-reply 

I added error checking for the Close() call, but the linter error still seems to appear. I’m not sure why this is happening… Should I also update other Close() calls as well?

plugins/database/valkey/valkey_test.go
```diff
@@ -214,29 +214,83 @@ func testValkeyDBInitialize_TLS(t *testing.T, host string, port int) {
...
+		Expiration: time.Now().Add(time.Minute),
+	}
+
+	userResp, err := db.NewUser(context.Background(), createReq)
 	if err != nil {
-		t.Fatalf("Testing Init() with connection_url failed: error: %s", err)
+		t.Fatalf("err: %s", err)
+	}
+
+	db.Close()
```
wslabosz-reply: @kyounghunJang 
so the linter is actually unhappy with this line (found it using following command thanks @satoqz)

```
git fetch upstream pull/1959/merge
git checkout FETCH_HEAD
```

I've got no idea why this is the case, but if you edit this one, this should (hopefully) work.

kyounghoonJang: @wslabosz-reply 
Thank you for finding it!!

plugins/database/valkey/valkey.go
```diff
@@ -138,13 +137,18 @@ func newUser(ctx context.Context, db radix.Client, username string, req dbplugin
 
 	aclargs := []string{"SETUSER", username, "ON", ">" + req.Password}
 
-	var args []string
-	err := json.Unmarshal([]byte(statements[0]), &args)
+	logger := hclog.New(&hclog.LoggerOptions{})
```
cipherboy, line 140: This is strange, I understand it is the pattern of the plugin, but also probably not the best idea to always create a new logger like this.

cipherboy (approved):

Issue #1580, opened 2025-07-23
staticcheck (anything flagged by staticcheck; general code improvements, simplifications, ...) - 406 findings


phyrog:
Findings:

```
```

phyrog:
Continuing:

```
vault/barrier.go:18:21: ST1005: error strings should not be capitalized (staticcheck)
	ErrBarrierSealed = errors.New("Vault is sealed")
vault/barrier.go:22:26: ST1005: error strings should not be capitalized (staticcheck)
	ErrBarrierAlreadyInit = errors.New("Vault is already initialized")
vault/barrier.go:26:22: ST1005: error strings should not be capitalized (staticcheck)
	ErrBarrierNotInit = errors.New("Vault is not initialized")
vault/barrier_test.go:145:58: ST1008: error should be returned as the last argument (staticcheck)
func testInitAndUnseal(t *testing.T, b SecurityBarrier) (error, *logical.StorageEntry, []byte) {
vault/barrier_view_test.go:265:12: SA4010: this result of append is never used, except maybe in other appends (staticcheck)
		expect = append(expect, e.Key)
vault/cluster_test.go:340:57: SA1029: should not use built-in type string as key for value; define your own type to avoid collisions (staticcheck)
	req = req.WithContext(context.WithValue(req.Context(), "original_request_path", req.URL.Path))
vault/core.go:1895:62: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
		ctx, ctxCancel := context.WithCancel(namespace.RootContext(nil))
vault/core.go:1953:58: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
	ctx, cancel := context.WithCancel(namespace.RootContext(nil))
vault/core.go:1987:48: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
	return c.sealInitCommon(namespace.RootContext(nil), req)
vault/core.go:3588:35: S1012: should use time.Since instead of time.Now().Sub (staticcheck)
		"duration", fmt.Sprintf("%dms", time.Now().Sub(reqData.StartTime).Milliseconds()),
vault/core_metrics_test.go:121:4: QF1003: could use tagged switch on l.Name (staticcheck)
			if l.Name == "mount_point" {
vault/deadlock_test.go:43:3: SA2001: empty critical section (staticcheck)
[...]

PR #1962, opened 2025-10-17
Fix some staticcheck findings

phyrog:
This fixes some of the low hanging fruits of #1580 

The issue was a bit misleading in that we had the implicit limit on same issues enabled in golangci-lint. With that disabled, after this PR, there are now about 1700 findings left (many of which are the same issue over and over).

I recommend reviewing commit by commit.

### DIFF
--- a/.golangci.deprecations.yml
+++ b/.golangci.deprecations.yml
@@ -9,3 +9,4 @@ linters:
         - SA1019
 issues:
   max-issues-per-linter: 0
+  max-same-issues: 0
--- a/.golangci.fmt.yml
+++ b/.golangci.fmt.yml
@@ -10,3 +10,4 @@ formatters:
       module-path: github.com/openbao/openbao
 issues:
   max-issues-per-linter: 0
+  max-same-issues: 0
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,3 +12,4 @@ formatters:
       module-path: github.com/openbao/openbao
 issues:
   max-issues-per-linter: 0
+  max-same-issues: 0


builtin/logical/transit/path_export_test.go
```diff
@@ -343,7 +343,7 @@ func TestTransit_Export_KeysDoesNotExist_ReturnsNotFound(t *testing.T) {
 	}
 	rsp, err := b.HandleRequest(context.Background(), req)
 
-	if !(rsp == nil && err == nil) {
+	if rsp != nil || err != nil {
 		t.Fatal("Key does not exist but does not return not found")
```
wslabosz-reply: nit: this message hurts me, can we change that

command/operator_diagnose.go
```diff
@@ -572,16 +573,17 @@ SEALFAIL:
...
 			for _, ln := range lns {
-				ln.Listener.Close()
+				errs = errors.Join(errs, ln.Close())
 			}
 		}
 
 		c.cleanupGuard.Do(listenerCloseFunc)
 
-		return nil
+		return errs
```
wslabosz-reply: I wonder if this changes the behavior of the command, can you double check?

plugins/database/mysql/mysql-legacy-database-plugin/main.go
```diff
@@ -21,8 +21,7 @@ func main() {
 
 // Run instantiates a MySQL object, and runs the RPC server for the plugin
 func Run() error {
-	var f func() (interface{}, error)
-	f = mysql.New(mysql.DefaultLegacyUserNameTemplate)
+	f := mysql.New(mysql.DefaultLegacyUserNameTemplate)
 	dbType, err := f()
```
wslabosz-reply:
```suggestion
dbType, err := mysql.New(mysql.DefaultLegacyUserNameTemplate)
```

plugins/database/mysql/mysql-database-plugin/main.go
```diff
@@ -21,8 +21,7 @@ func main() {
 
 // Run instantiates a MySQL object, and runs the RPC server for the plugin
 func Run() error {
-	var f func() (interface{}, error)
-	f = mysql.New(mysql.DefaultUserNameTemplate)
+	f := mysql.New(mysql.DefaultUserNameTemplate)
 
 	dbplugin.ServeMultiplex(f)
```
wslabosz-reply:
```suggestion
dbplugin.ServeMultiplex(mysql.New(mysql.DefaultUserNameTemplate))
```

command/operator_diagnose.go
phyrog: It changes the behavior in the sense that it will now also error when it fails to close the listeners. Given that this command is used to find potential problems, I'd say that is a good thing.

I did change this code a bit again, so we're now not failing the whole check, but instead just issue a warning; I think this might be more appropriate if this fails.

What do you think @wslabosz-reply?

wslabosz-reply: I think that's correct approach, I wonder if a change like this would still require a changelog, but apart from that LGTM.

phyrog:
Rebased on main

wslabosz-reply (approved):

PR #1963, opened 2025-10-17
feat: add initial OSSF security scorecard workflow

weberval:
In order to reach the “become incubating” stage at OpenSSF we need to integrate the OpenSSF security scorecard.

This PR adds a “close to upstream” GitHub workflow and integrated the badge to the README.md.

This is related to https://github.com/openbao/dev-wg/issues/55

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -36,10 +36,6 @@ jobs:
         with:
           results_file: results.sarif
           results_format: sarif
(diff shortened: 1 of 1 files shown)


weberval:
This workflow was tested against my fork: https://scorecard.dev/viewer/?uri=github.com/weberval-forks/openbao

weberval:
@cipherboy is that what you had in mind?

cipherboy (approved):
Makes sense to me, thanks @weberval!

PR #1224, opened 2025-04-15
Detailed metadata in policy endpoint

ShabiniRajadas:
This PR allows Openbao to list detailed information about capabilities in policies endpoint, avoiding N+1 while trying to retrieve the policies and their capabilities.

Issue link: https://github.com/openbao/openbao/issues/1156

PR #1965, opened 2025-10-17
Allow detailed listing for a subset of policies

cipherboy:
In adding documentation, I noticed that listing a subset of policies was not supported. We should likely support this as it is a generally useful feature: few people with structured policies will want to see all policies at once and will likely want just a subset of data.

See also: https://github.com/openbao/openbao/pull/1224


cipherboy:
fyi @ShabiniRajadas!

ShabiniRajadas (approved):
Thanks @cipherboy

PR #1966, opened 2025-10-19
Run make sync-deps on Dependabot PRs

JanMa:
To automatically keep dependencies across modules in sync

### DIFF
--- a/.github/workflows/sync-deps.yaml
+++ b/.github/workflows/sync-deps.yaml
@@ -1,0 +1,45 @@
+name: Sync Dependencies
+
+on:
+  push:
+    branches:
+      - "dependabot/**"
+    paths:
+      - go.mod
+
+permissions:
+  contents: write
+
+jobs:
+  sync-deps:
+    name: Sync Dependencies
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-deps.yaml
```diff
@@ -0,0 +1,45 @@
+name: Sync Dependencies
+
+on:
+  pull_request:
+    branches:
+      - "dependabot/**"
```
satoqz, line 6: Doesn't this say "run this workflow if target branch matches `dependabot/**`" instead of "run this workflow if source branch matches `dependabot/**`"?

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request

Maybe there's a better way, but this should probably be `branches: [main]` and then check if the PR author is dependabot as the first step?

JanMa: Right :facepalm: Thanks for catching this, I got confused by the docs and thought this means the _source_ branch needs to match this pattern

JanMa: When using the `push` target it works as intended

cipherboy (approved):
Very cool! I was looking into this when `make sync-deps` was first introduced and the recommended advice was to replace the way Dependabot runs and execute it manually in CI. This seems much cleaner. Thanks @JanMa!

PR #1972, opened 2025-10-20
Remove make.bat

satoqz:
`make.bat` calls into `scripts/windows/build.bat`, which was removed in https://github.com/openbao/openbao/pull/168, and is broken as a result.

I think we have several good reasons to remove `make.bat` as well:

- We are missing `build.bat`
- I'm not aware of any frequent contributor that uses Windows for development
- Nobody has complained about `make.bat` being broken (outside of https://github.com/orgs/openbao/discussions/1922, which pertains to packaging rather than development, which is what these scripts are for)
- `make.bat` functionality lags years behind `Makefile` functionality, with nobody interested in maintaining it


wslabosz-reply (approved):
🟢

PR #1973, opened 2025-10-20
Add Best Practices badge for Scorecard

cipherboy:
It appears not to have been linked in this repository as code, only from issues.


cipherboy:
Merging based on @wslabosz-reply's review; this only adds a new badge to the README.

PR #1979, opened 2025-10-21
Fix false positive namespace conflict

### LINKED ISSUE #1957: Mount path equal to namespace name incorrectly conflicts with namespace sys/ (“existing mount at …”

phyrog:
Resolves #1957 by ignoring the current namespace when finding conflicting namespaces during a mount.

### DIFF
--- a/vault/router.go
+++ b/vault/router.go
@@ -418,7 +418,8 @@ func (r *Router) matchingNamespaceInternal(ctx context.Context, path string) str
 		}
 
 		nsPath := strings.TrimSuffix(existingPath, "sys/")
-		if strings.HasPrefix(path, nsPath) {
+		// ignore the namespace we are operating in
+		if strings.HasPrefix(path, nsPath) && nsPath != ns.Path {
 			existing = nsPath
 			return true
 		}
--- a/vault/router_test.go
+++ b/vault/router_test.go
@@ -4,6 +4,7 @@
 package vault
 
 import (
+	"context"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


phyrog:
Duplicate of #1958

PR #1982, opened 2025-10-21
fix(core/namespaces): only return namespaces which the provided token has access to

### LINKED ISSUE #1981: UI namespace endpoint doesn't filter only accessible Namespaces
**Describe the bug**
The API Endpoint `sys/internal/ui/namespaces` will always return a list of all namespaces instead of only returning the namespaces the token has access to.
I'd expect that only namespaces where the user has a policy attached to will be returned (This would be the same behavior as in HCV Ent).

Additionally side note: The [docs page](https://openbao.org/api-docs/system/internal-ui-namespaces/) has incorrect information in it: `and is an unauthenticated endpoint.` - It  requires a token, so it's authenticated.

**To Reproduce**
1. Create a bunch of namespaces (e.g. `test1`, `test2`)
2. Create an userpass auth method and user
3. Attach a policy which only grants access to namespace `test1` (example `path "test1/* { capabilities = ["read"] }`)
4. `bao read sys/internal/ui/namespaces`
```
Key     Value
---     -----
keys    [test1/ test2/]
```

**Expected behavior**
```
Key     Value
---     -----
keys    [test1/]
```

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.4.1
* OpenBao CLI Version (retrieve with `bao version`): 2.4.1

pree:
Resolves #1981

### DIFF
--- a/changelog/1982.txt
+++ b/changelog/1982.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-core/namespaces: only return namespaces which the provided token has access to
+core/namespaces: only report namespaces which the provided token has access to from `sys/internal/ui/namespaces`
 ```


changelog/1982.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+core/namespaces: only return namespaces where the token has access
```
wslabosz-reply:
```suggestion
core/namespaces: only return namespaces which the provided token has access to
```

pree: Thanks, added :)

pree:
Tested as follows:

```
#test.hcl
path "test1/*" {
  capabilities = ["read"]
}
```

```
$ bao server -dev
$ bao namespace create test1
$ bao namespace create test2
$ bao auth enable userpass
$ bao write auth/userpass/users/test password=test policies=test
$ bao policy write test ./test.hcl

$ bao login -method=userpass username=test
$ bao read sys/internal/ui/namespaces
Key     Value
---     -----
keys    [test1/]
```

vault/logical_system_helpers.go
satoqz: While we're at it, mind turning this into a statically declared method and moving it to `logical_system.go` together with other UI-internal handlers such as `pathInternalUIMountRead`? Maybe also worth a separate PR that fully removes `logical_system_helpers.go`.

pree: Wouldn't the method fit more into `logical_system_namespaces.go`?

pree: I see, the other path UI methods are also inside `logical_system.go`, so I moved it there.

pree:
@driif thanks for the hint! You're absolute right, due to using `relativePath` I've checked the ACL always compared to the root ns. I've now chainged to the full path and in my testing this is fixed now.

changelog/1982.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+core/namespaces: only return namespaces which the provided token has access to
```
cipherboy:
```suggestion
core/namespaces: only report namespaces which the provided token has access to from `sys/internal/ui/namespaces`
```

perhaps?

cipherboy (approved):

cipherboy:
Thanks @pree and everyone for the reviews!

PR #1985, opened 2025-10-21
Remove any usage of LZ4 and LZW compression

satoqz:
The `compressutil` helper package supports Gzip, Snappy, LZ4 and LZW. LZ4 and LZW have not been in use by any component for a long time, so drop support. In particular, dropping LZ4 lets us drop the associated dependency.

The only risk I can think of is that any residual data that was LZ4 or LZW-compressed cannot be read via `sys/raw` anymore. Thoughts?

Also see: https://github.com/openbao/openbao/pull/1913#issuecomment-3396530980

### DIFF
--- a/sdk/helper/compressutil/compress.go
+++ b/sdk/helper/compressutil/compress.go
@@ -24,6 +24,13 @@ const (
 
 	CompressionTypeSnappy        = "snappy"
 	CompressionCanarySnappy byte = 'S'
+
+	// These compression types were in use in previous versions of the project,
+	// but are not supported anymore. We keep the canaries around to provide
+	// clearer error messages in the unlikely event that residual, unsupported
+	// data is attempted to be read.
+	CompressionCanaryLZ4 byte = '4'
+	CompressionCanaryLZW byte = 'L'
 )
 
 // SnappyReadCloser embeds the snappy reader which implements the io.Reader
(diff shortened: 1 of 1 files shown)


vault/mount.go
```diff
@@ -1700,7 +1700,7 @@ func (c *Core) persistMounts(ctx context.Context, barrier logical.Storage, table
 
 	// Handle writing the legacy mount table by default.
 	writeTable := func(mt *MountTable, path string) (int, error) {
-		// Encode the mount table into JSON and compress it (lzw).
+		// Encode the mount table into JSON and compress it (Gzip).
```
satoqz, line 1703: Image

sdk/helper/compressutil/compress.go
```diff
@@ -107,10 +93,6 @@ func Compress(data []byte, config *CompressionConfig) ([]byte, error) {
 		buf.Write([]byte{CompressionCanarySnappy})
 		writer = snappy.NewBufferedWriter(&buf)
 
-	case CompressionTypeLZ4:
-		buf.Write([]byte{CompressionCanaryLZ4})
-		writer = lz4.NewWriter(&buf)
-
 	default:
 		return nil, errors.New("unsupported compression type")
```
cipherboy:
```suggestion
		return nil, fmt.Errorf("unsupported compression type: %v", config.Type)
```

perhaps, and likewise with decoding, keeping the constants? I agree, that way anyone who happens to have weird entries lying around can very clearly see why.

sdk/helper/compressutil/compress.go
cipherboy, line 181: I'm going to merge this, but it is very weird that this returns it as decompressed but there's no fallthrough for when you do see a canary and have uncompressed data. 😆

cipherboy:
Thanks @satoqz!

PR #1986, opened 2025-10-21
Merge Read Replication Support to Main

cipherboy:
This is the work of the `scalability` working group under the Dev WG.

A huge thanks to @fatima2003 and @phil9909 for their work on this! Thanks to Adfinis for funding their time on this effort.

The intention is to rebase merge this to keep authorship; I've made sure each commit builds cleanly, though tests will only pass at the end.

----

This introduces read scalability support to OpenBao in a flat (single-cluster) architecture:

- Standby nodes are automatically read-enabled unless explicitly disabled via `disable_standby_reads=true` in the config.
- Reads in this context refer to storage operations: all requests are fully processed on the standby node they come in to (unlike before, where they were immediately forwarded) and only are forwarded if the processing incurs a storage write operation.
  -  For now, this means two audit log entries, but this could change if we introduce forwarding at the `logical.Request` level.
- Usual Vault forwarding semantics apply, that is, a plugin's `ForwardPerformanceStandby` is now consulted.
- This is not Performance Secondary: nodes/mounts exist in a flat topology and there are no split-cluster scalability. Each mount strictly has a single writer, even for (Performance Secondary) cluster-local data.
- The existing invalidation mechanism was preserved: storage on secondaries indicates to Core when new storage writes have been applied.
- Currently this branch only has Raft support. Ideally PostgreSQL support will come separately, but soon.

The general usability of this feature is expected to follow existing HA support: https://openbao.org/docs/concepts/ha/ + https://openbao.org/docs/internals/high-availability/

**Plugin authors** will need to implement `logical.Backend.Invalidate` to be notified of any invalidations which occur on cached data. This _should_ just mark the cache as stale and _should not_ incur any storage operations. Instead, the next regular request (or, `logical.Backend.PeriodicFunc` invocation) should handle refreshing the stale cache.

### DIFF
--- a/http/help.go
+++ b/http/help.go
@@ -14,8 +14,7 @@ import (
 
 func wrapHelpHandler(h http.Handler, core *vault.Core) http.Handler {
 	return http.HandlerFunc(func(writer http.ResponseWriter, req *http.Request) {
-		// If the help parameter is not blank, then show the help. We request
-		// forward because standby nodes do not have mounts and other state.
+		// If the help parameter is not blank, then show the help.
 		if v := req.URL.Query().Get("help"); v != "" || req.Method == "HELP" {
 			http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				handleHelp(core, w, r)
--- a/http/logical.go
+++ b/http/logical.go
@@ -369,7 +369,7 @@ func handleLogicalInternal(core *vault.Core, injectDataIntoTopLevel bool, noForw
 		if core.HAEnabled() && !core.StandbyReadsEnabled() {
 			standby, err := core.Standby()
 			if err != nil {
-				respondError(w, 503, err)
+				respondError(w, http.StatusServiceUnavailable, err)
 				return
 			}
 			if standby {
--- a/sdk/logical/error.go
+++ b/sdk/logical/error.go
@@ -140,12 +140,8 @@ func ShouldForward(err error) bool {
 	}
 
 	errMsg := err.Error()
-	if strings.Contains(errMsg, ErrPerfStandbyPleaseForward.Error()) ||
+	return strings.Contains(errMsg, ErrPerfStandbyPleaseForward.Error()) ||
 		strings.Contains(errMsg, ErrReadOnly.Error()) ||
 		strings.Contains(errMsg, "node is not the leader") ||
-		strings.Contains(errMsg, consts.ErrStandby.Error()) {
-		return true
-	}
-
-	return false
+		strings.Contains(errMsg, consts.ErrStandby.Error())
 }
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -2499,9 +2499,9 @@ func (c *Core) reloadLegacyMounts(ctx context.Context, keys ...string) error {
 	invalidations := map[string]invalidation{}
 
 	for _, path := range keys {
-		table := "mounts"
+		table := mountTableType
 		if path == coreAuthConfigPath && path != coreLocalAuthConfigPath {
-			table = "auth"
+			table = credentialTableType
 		}
 
 		raw, err := c.barrier.Get(ctx, path)
@@ -2576,12 +2576,12 @@ func (c *Core) reloadMount(ctx context.Context, key string) error {
 	prefix, uuid := path.Split(key)
 	prefix = path.Clean(prefix)
 
-	table := "mounts"
+	table := mountTableType
 	if prefix != coreLocalMountConfigPath && prefix != coreMountConfigPath {
 		if prefix != coreAuthConfigPath && prefix != coreLocalAuthConfigPath {
 			return fmt.Errorf("invalid path prefix %q", prefix)
 		}
-		table = "auth"
+		table = credentialTableType
 	}
 
 	ns, err := namespace.FromContext(ctx)
@@ -2603,7 +2603,7 @@ func (c *Core) reloadMount(ctx context.Context, key string) error {
 
 func (c *Core) reloadMountInternal(ctx context.Context, table, uuid string, desiredMountEntry *MountEntry) error {
 	switch table {
-	case "auth", "mounts":
+	case credentialTableType, mountTableType:
 	default:
 		return fmt.Errorf("invalid mount table type passed: %q", table)
 	}
@@ -2619,7 +2619,7 @@ func (c *Core) reloadMountInternal(ctx context.Context, table, uuid string, desi
 	case desiredMountEntry == nil && actualMountEntry != nil: // mount was deleted
 		c.logger.Debug("cache invalidation: mount was deleted", "type", table, "uuid", uuid)
 
-		if table == "auth" {
+		if table == credentialTableType {
 			err = c.removeCredEntry(ctx, actualMountEntry.Path, false)
 		} else {
 			err = c.removeMountEntry(ctx, actualMountEntry.Path, false)
@@ -2629,8 +2629,8 @@ func (c *Core) reloadMountInternal(ctx context.Context, table, uuid string, desi
 		}
 
 		routerPath := actualMountEntry.Path
-		if table == "auth" {
-			routerPath = path.Join("auth", routerPath) + "/"
+		if table == credentialTableType {
+			routerPath = path.Join(credentialRoutePrefix, routerPath) + "/"
 		}
 		if err := c.router.Unmount(ctx, routerPath); err != nil {
 			return err
@@ -2646,7 +2646,7 @@ func (c *Core) reloadMountInternal(ctx context.Context, table, uuid string, desi
 	case desiredMountEntry != nil && actualMountEntry == nil: // mount was created
 		c.logger.Debug("cache invalidation: mount was created", "type", table, "uuid", uuid)
 
-		if table == "auth" {
+		if table == credentialTableType {
 			err = c.enableCredentialInternal(ctx, desiredMountEntry, false)
 			if err != nil {
 				return err
@@ -2663,7 +2663,7 @@ func (c *Core) reloadMountInternal(ctx context.Context, table, uuid string, desi
 		c.logger.Debug("cache invalidation: mount was modified", "type", table, "uuid", uuid)
 
 		lock := &c.mountsLock
-		if table == "auth" {
+		if table == credentialTableType {
 			lock = &c.authLock
 		}
 
@@ -2671,8 +2671,8 @@ func (c *Core) reloadMountInternal(ctx context.Context, table, uuid string, desi
 		defer lock.Unlock()
 
 		routerPath := actualMountEntry.Path
-		if table == "auth" {
-			routerPath = path.Join("auth", routerPath) + "/"
+		if table == credentialTableType {
+			routerPath = path.Join(credentialRoutePrefix, routerPath) + "/"
 		}
 
 		if desiredMountEntry.Tainted != actualMountEntry.Tainted {
@@ -2697,7 +2697,7 @@ func (c *Core) reloadMountInternal(ctx context.Context, table, uuid string, desi
 		}
 
 		if desiredMountEntry.Options["version"] != actualMountEntry.Options["version"] {
-			err = c.reloadBackendCommon(ctx, desiredMountEntry, table == "auth")
+			err = c.reloadBackendCommon(ctx, desiredMountEntry, table == credentialTableType)
 			if err != nil {
 				return err
 			}


physical/raft/raft.go
```diff
@@ -1598,6 +1617,11 @@ func (b *RaftBackend) RestoreSnapshot(ctx context.Context, metadata raft.Snapsho
 
 // Delete inserts an entry in the log to delete the given path
 func (b *RaftBackend) Delete(ctx context.Context, path string) error {
+	// Return early if follower node is attempting to write to storage
+	if b.raft.State() != raft.Leader {
```
cipherboy: @phil9909 I wonder if we should do this or what I proposed in #1987 and add Raft-specific constants to the forwarding hook?

Alternatively, we could actually provision another layer into `physical` which allows us to set `logical.ErrReadOnly` on HA standby nodes (on writes) and then reset to `nil` error on active nodes. Thoughts?

cipherboy: The only reason why I didn't do this already was because it seemed weird to duplicate `NewBarrierView`, which has the logic as well. But there's a lot of places where `c.barrier` is accessed directly, that we'd probably want to fail, too.

vault/auth.go
```diff
@@ -96,6 +96,10 @@ func (c *Core) enableCredentialInternal(ctx context.Context, entry *MountEntry,
 	}
 	defer unlock()
 
+	if err := ctx.Err(); err != nil {
```
cipherboy: @phil9909 I am entertained with this; I'm not sure if I want to keep it or not. See #1990 for some examples. 

At minimum, I don't think we can do this when `updateStorage=false`, and we definitely cannot do it on the `disableCredentialInternal` version.... maybe this means the context cancellation check needs to occur elsewhere? 🤔

vault/core.go
```diff
@@ -2273,9 +2287,14 @@ type UnsealStrategy interface {
 	unseal(context.Context, log.Logger, *Core) error
 }
 
-type standardUnsealStrategy struct{}
+type standardUnsealStrategy struct {
+	// Inherit read-only unseal methods
+	readonlyUnsealStrategy
+}
 
 func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c *Core) error {
```
cipherboy, line 2309: Do we need to set replication constants? 🤔

vault/dynamic_system_view.go
```diff
@@ -181,6 +181,9 @@ func (d dynamicSystemView) LocalMount() bool {
 // in read mode.
 func (d dynamicSystemView) ReplicationState() consts.ReplicationState {
 	state := d.core.ReplicationState()
+	if d.core.standby.Load() {
```
cipherboy: Which is to say, should https://github.com/openbao/openbao/pull/1986/files#r2450003133 handle this rather than setting it here.

vault/core.go
cipherboy: X-link: https://github.com/openbao/openbao/pull/1986/files#r2450004654

vault/request_handling.go
```diff
@@ -2397,6 +2416,9 @@ func (c *Core) PopulateTokenEntry(ctx context.Context, req *logical.Request) err
 	// decodes the SSCT, and it may need the original SSCT to check state.
 	te, err := c.LookupToken(ctx, token)
 	if err != nil {
+		if errors.Is(err, logical.ErrPerfStandbyPleaseForward) {
```
cipherboy, line 2417: @phil9909 Should we use `logical.ShouldForward` for consistency?

vault/ha.go
```diff
@@ -388,79 +383,223 @@ func (c *Core) StepDown(httpCtx context.Context, req *logical.Request) (retErr e
...
+func (c *Core) runStandby(doneCh chan<- struct{}, manualStepDownCh chan struct{}, stopCh, restartCh <-chan struct{}) {
+	defer close(doneCh)
+	defer close(manualStepDownCh)
+
+	for restart := true; restart; {
+		c.logger.Info("entering standby mode")
+		restart = false
+
+		var perfCancel context.CancelFunc
+		if c.StandbyReadsEnabled() {
```
cipherboy: @phil9909 Do we think this should be part of the rungroup? 🤔

phil9909: I guess this would create races with `waitForLeadership`

vault/ha.go
```diff
@@ -388,79 +383,223 @@ func (c *Core) StepDown(httpCtx context.Context, req *logical.Request) (retErr e
...
+				c.logger.Error("runStandby: unable to grab state lock", "err", err)
+				return
+			}
+
+			// wipe any existing mount tables
+			if err := c.preSeal(); err != nil {
+				c.logger.Error("pre-seal teardown failed", "error", err)
+			}
+
+			c.drainPendingRestarts()
```
phil9909: I guess this needs to be moved out of the `if` block.

Currently, only standby nodes will use the `restart` functionality, but still.

vault/ha.go
cipherboy: Ah, that's fair.

cipherboy:
@cipherboy to fix individual commit build issues on rebase.

vault/ha.go
```diff
@@ -1150,3 +1297,17 @@ func (c *Core) SetNeverBecomeActive(on bool) {
...
+func (c *Core) StandbyReadsEnabled() bool {
+	if _, ok := c.underlyingPhysical.(physical.CacheInvalidationBackend); !ok {
+		return false
+	}
+
+	conf := c.rawConfig.Load()
+	if conf == nil {
+		return false
+	}
+	return !conf.(*server.Config).DisableStandbyReads
```
cipherboy, line 1327: @phil9909 This should probably be restructured. There's no way to swap this at runtime currently, so we shouldn't read it from `rawConfig` as that is dynamic and subject to change, but instead read it once on startup and set it on the `core` object directly.

physical/raft/raft.go
cipherboy: This has been resolved by adding the constant to the forwarding hook and marking the barrier read-only so no writes can go past it.

vault/core.go
cipherboy: Fixed.

vault/ha.go
cipherboy: @phil9909 If we create this as `make(chan struct{}, 1)`, why would we need to drain? :thinking: We'd only ever have a single pending restart. 

Let's get others' thoughts on the channel here.

sdk/logical/error.go
```diff
@@ -128,3 +131,21 @@ func (e *KeyNotFoundError) WrappedErrors() []error {
...
+
+	errMsg := err.Error()
+	if strings.Contains(errMsg, ErrPerfStandbyPleaseForward.Error()) ||
+		strings.Contains(errMsg, ErrReadOnly.Error()) ||
+		strings.Contains(errMsg, "node is not the leader") ||
+		strings.Contains(errMsg, consts.ErrStandby.Error()) {
+		return true
+	}
+
+	return false
```
wslabosz-reply:
```suggestion
	return strings.Contains(errMsg, ErrPerfStandbyPleaseForward.Error()) ||
		strings.Contains(errMsg, ErrReadOnly.Error()) ||
		strings.Contains(errMsg, "node is not the leader") ||
		strings.Contains(errMsg, consts.ErrStandby.Error())
```

http/logical.go
```diff
@@ -379,7 +366,19 @@ func handleLogicalRecovery(raw *vault.RawBackend, token *atomic.Value) http.Hand
 // toggles. Refer to usage on functions for possible behaviors.
 func handleLogicalInternal(core *vault.Core, injectDataIntoTopLevel bool, noForward bool) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		req, origBody, statusCode, err := buildLogicalRequest(core, w, r)
+		if core.HAEnabled() && !core.StandbyReadsEnabled() {
+			standby, err := core.Standby()
+			if err != nil {
+				respondError(w, 503, err)
```
wslabosz-reply:
```suggestion
				respondError(w, http.StatusServiceUnavailable, err)
```

vault/mount.go
```diff
@@ -2337,3 +2378,331 @@ func (c *Core) mountEntryView(me *MountEntry) (BarrierView, error) {
...
+		}
+		desiredMountEntry = nil
+	}
+
+	return c.reloadMountInternal(ctx, table, uuid, desiredMountEntry)
+}
+
+func (c *Core) reloadMountInternal(ctx context.Context, table, uuid string, desiredMountEntry *MountEntry) error {
+	switch table {
+	case "auth", "mounts":
```
wslabosz-reply:
```suggestion
	case credentialTableType, mountTableType:
```

vault/mount.go
```diff
@@ -2337,3 +2378,331 @@ func (c *Core) mountEntryView(me *MountEntry) (BarrierView, error) {
...
+				invalidations[entry.UUID] = invalidation{
+					Table:             entry.Table,
+					DesiredMountEntry: nil,
+					Namespace:         entry.Namespace(),
+				}
+			}
+		}
+	}
+	c.authLock.RUnlock()
+	c.mountsLock.RUnlock()
```
wslabosz-reply, line 2563: instead of looping over [`c.mounts`, `c.auth`], could we do one at a time, the benefit of that would be only acquiring a lock of the mount table we are currently iterating over. That would also be cleaner if we declared a separate function which could be reused providing mount tables. /My 2c.

cipherboy: @wslabosz-reply I agree for the record, but was really hoping to refactor much of this after merge and with your changes.

wslabosz-reply: @cipherboy 
that's completely fair, will refrain from commenting on the "mounts" portion of the feature then.

cipherboy: @wslabosz-reply If you notice a bad correctness problem, more than happy to fix it sooner. But for code structuring and lock ordering, I think that's a different conversation.

http/help.go
```diff
@@ -17,10 +17,9 @@ func wrapHelpHandler(h http.Handler, core *vault.Core) http.Handler {
 		// If the help parameter is not blank, then show the help. We request
 		// forward because standby nodes do not have mounts and other state.
```
satoqz: Outdated comment

sdk/physical/inmem/inmem_ha.go
```diff
@@ -46,6 +49,13 @@ func (i *InmemHABackend) LockWith(key, value string) (physical.Lock, error) {
 	return l, nil
 }
 
+func (i *InmemHABackend) HookInvalidate(hook physical.InvalidateFunc) {
+	i.l.Lock()
+	defer i.l.Unlock()
+
+	i.invalidators = append(i.invalidators, hook)
```
satoqz, line 56: I'm a bit surprised that Raft only stores a single `InvalidateFunc` at a time, but here we support several?

http/handler.go
```diff
@@ -772,47 +779,32 @@ func parseFormRequest(r *http.Request) (map[string]interface{}, error) {
...
-		}
-
-		forwardRequest(core, w, r)
-	})
-}
-
 func forwardRequest(core *vault.Core, w http.ResponseWriter, r *http.Request) {
+	// Reset our body before forwarding to ensure an accurate location.
+	if err := resetBody(r); err != nil {
+		respondStandby(core, w, r.URL)
```
driif, line 784: Should we respond here with Error if we can't reset Body?

```suggestion
		respondError(w, http.StatusInternalServerError, err)
```

cipherboy: @driif Standby is accurate so that we forward the request by 307 redirect (client-side forward).

If we respond with err, we don't even forward.

cipherboy: Additionally, I think this err is nearly impossible to happen here, given that we've wrapped it in an bytes buffer. Outside of OOB issues, this one doesn't err, and we always seek to the beginning so there's no OOB issues to be had.

vault/ha.go
driif: @cipherboy @phil9909 I agree, changing to `c.standbyRestartCh.Store(make(chan struct{}, 1))` seems to be an elegant solution in this case which makes `*Core.drainPendingRestarts()` not necessary anymore.

We can also change `c.standbyStopCh.Store(make(chan struct{}, 1))` for consistancy.

http/handler.go
phil9909: Because it is an error that should not happen under "normal" operation, I tend to agree with @driff and we should `respondError` here. Forwarding should be reserved to failures caused by the fact that the node is in standby, shouldn't it?

vault/ha.go
```diff
@@ -69,10 +63,7 @@ func (c *Core) ActiveTime() time.Time {
 // StandbyStates is meant as a way to avoid some extra locking on the very
 // common sys/health check.
 func (c *Core) StandbyStates() (standby bool) {
-	c.stateLock.RLock()
-	standby = c.standby
-	c.stateLock.RUnlock()
-	return standby
+	return c.standby.Load()
 }
```
driif, line 68: This function can be deleted, no longer used intentionally.

http/handler.go
cipherboy: @driif @phil9909 Keep in mind the context here.

We're in `forwardRequest`. We _**want**_ to be forwarding this request. The only case where we shouldn't, is if we encountered a fatal error that prevents us from forwarding (unable to get leadership, &c). Any error doesn't prevent us from forwarding via 307, but does prevent us from forwarding via GRPC client, since we cannot reset the body to beginning to forward the body ourselves.

My 2c.

http/handler.go
```diff
@@ -175,42 +175,42 @@ func handler(props *vault.HandlerProperties) http.Handler {
 		mux.Handle("/v1/sys/init", handleSysInit(core))
 		mux.Handle("/v1/sys/seal-status", handleSysSealStatus(core))
 		mux.Handle("/v1/sys/seal", handleSysSeal(core))
-		mux.Handle("/v1/sys/step-down", handleRequestForwarding(core, handleSysStepDown(core)))
+		mux.Handle("/v1/sys/step-down", handleSysStepDown(core))
```
driif, line 178: Seems like it can have some unexpected behavior. 

in `ha.go` we in `*Core.stepDown` we check if it's standby node and return nil, wich should return `200 OK`, but we aren't forwarding request to leader.


https://github.com/openbao/openbao/blob/981a3ff590021c5c45d8de0ce511ef71364a1150/vault/ha.go#L272-L274

cipherboy: Agreed, I think this is an oversight.

satoqz: Note this commit:

```
commit 20eadd350b4f94d7615131f9550dbde53eedd68b
Author: Jeff Mitchell <jeffrey.mitchell@gmail.com>
Date:   Thu May 25 11:55:40 2017 -0400

    Have step-down request forward.

    Unlike seal, this command has no meaning other than on the active node,
    so when issuing it the expected behavior would be for whichever node is
    currently active to step down.
```

Yet docs at https://openbao.org/api-docs/system/step-down say:

> This endpoint forces the node to give up active status. If the node does not have active status, this endpoint does nothing.

vault/audit.go
```diff
@@ -413,33 +473,33 @@ func (c *Core) setupAudits(ctx context.Context) error {
...
-		successCount++
-	}
-
-	if len(c.audit.Entries) > 0 && successCount == 0 {
-		return errLoadAuditFailed
+		// Register the backend
+		c.auditBroker.Register(entry.Path, backend, view, entry.Local)
+		if c.logger.IsInfo() {
+			c.logger.Info("enabled audit backend", "path", entry.Path, "type", entry.Type)
+		}
```
satoqz:
```suggestion
	  c.logger.Info("enabled audit backend", "path", entry.Path, "type", entry.Type)
```

Nonsensical pattern that probably warrants a separate issue.

sdk/physical/inmem/cache_test.go
```diff
@@ -103,6 +103,63 @@ func TestCache_Purge(t *testing.T) {
...
+	require.NoError(cache.Put(context.Background(), &physical.Entry{
+		Key:   "foo",
+		Value: []byte("bar"),
+	}))
+
+	// Start a transaction
+	tx, err := cache.(physical.TransactionalBackend).BeginTx(context.Background())
+	require.NoError(err)
+
+	// Modify in under
```
satoqz, line 127: ...lying backend?

sdk/physical/inmem/inmem_ha.go
cipherboy: @satoqz That's because inmem HA is a test-only backend and kinda garbage.

Unlike every other storage backend with multiple copies of data, inmem HA literally uses a single inmem backend instance and just... Passes it to multiple cores. It has no replication mechanism and no single-writer behavior. It's just a single map masquerading as a proper storage backend.

As a result, multiple (test core) callers will register invalidation hooks on the same actual struct instance. Nowhere else will this happen in reality.

We can put more work into making this behave more like a real database, but perhaps there's diminishing returns at some point, when we can use an actual supported database and the goal is to be fast. 🤷

vault/ha.go
phil9909: Currently, it is still used in `sys_health.go`

But I agree that we could probably remove this method as `Core.Standby()` and `Core.StandbyStates()` are essentially equal (`Standby` returns an additional error that is always `nil`)

A lot of places that call `Standby` simply ignore the error. My suggestion:
- remove `StandbyStates`
- drop the error from `Standby`

But as this would require quite some call-sites to be changed, I would defer this to a separate PR.

sdk/logical/error.go
phil9909: Will be addressed in https://github.com/openbao/openbao/pull/2060

http/logical.go
phil9909: Will be addressed in https://github.com/openbao/openbao/pull/2060

vault/mount.go
phil9909: Will be addressed in https://github.com/openbao/openbao/pull/2060

http/help.go
phil9909: Will be addressed in https://github.com/openbao/openbao/pull/2060

vault/core.go
```diff
@@ -2361,26 +2429,10 @@ func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c
...
-	defer c.clusterParamsLock.Unlock()
-
-	c.metricsCh = make(chan struct{})
-	go c.emitMetricsActiveNode(c.metricsCh)
-
-	// Establish version timestamps at the end of unseal on active nodes only.
-	if err := c.handleVersionTimeStamps(ctx); err != nil {
-		return err
+	// Finally, start processing invalidations. We'll have cleared the queue
+	// when we started this, but any
```
satoqz, line 2433: but any?

vault/core.go
```diff
@@ -2301,6 +2327,48 @@ func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c
...
+
+	// Establish version timestamps at the end of unseal on active nodes only.
+	if err := c.handleVersionTimeStamps(ctx); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// readonlyUnsealStrategy
```
satoqz, line 2359: readonlyUnsealStrategy!

vault/core_cache_invalidate.go
```diff
@@ -0,0 +1,525 @@
...
+			pending := im.pending
+			im.pending = nil
+			im.pendingLock.Unlock()
+
+			im.core.metricSink.SetGauge([]string{dispatcherName, "pending-dequeue-size"}, float32(len(pending)))
+
+			for _, key := range pending {
+				job, queue := im.buildInvalidateJobForKey(quitCh, quitContext, key)
+				im.dispatcher.AddJob(job, queue)
+			}
```
satoqz, line 171: I wonder if we should deduplicate keys here?

vault/ha.go
driif: @phil9909 Yes, I've seen, that it's used. With `not intentionally` I'd like to say - we are not going to return different states of standby, we are wrapping only the function `c.standby.Load()`.


I agree that a separate PR is okay here for a small refactor.

vault/core_cache_invalidate.go
cipherboy: @satoqz I'm a little hesitant to do so. We could theoretically, but if a caller expects to write a number of entries and have it be semantically meaningful... Or more likely, we risk seeing some partial update and having more entries gives us a higher chance of dispatching one properly. :shrug:

vault/core.go
```diff
@@ -2361,26 +2429,10 @@ func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c
...
-
-	c.metricsCh = make(chan struct{})
-	go c.emitMetricsActiveNode(c.metricsCh)
-
-	// Establish version timestamps at the end of unseal on active nodes only.
-	if err := c.handleVersionTimeStamps(ctx); err != nil {
-		return err
+	// Finally, start processing invalidations. We'll have cleared the queue
+	// when we started this, but any
+	if c.standby.Load() {
```
cipherboy, line 2434: I'm fairly sure this doesn't work, because we're never marked active until later and so we'll always be standby here even on active nodes, but I have a commit to clean this up in my PostgreSQL invalidation branch. I think its fine to deal with later as we have other safeguards against populating this map.

vault/core.go
cipherboy: > but any invalidations that occurred during startup will now be processed.

i think is what I meant to say.

vault/mount.go
cipherboy: (I'll leave this for #2040 and successors)

cipherboy:
https://github.com/cipherboy/openbao/pull/new/post-merge-fixes should address feedback from above.

cipherboy:
@openbao/openbao-org-maintainers I realized that `builtin/logical/pki/ca_test.go` is the only file strictly requiring org-maintainer review. The rest falls into @openbao/openbao-core-committers group.

vault/barrier_aes_gcm.go
```diff
@@ -66,8 +66,9 @@ var (
 type AESGCMBarrier struct {
 	backend physical.Backend
 
-	l      sync.RWMutex
-	sealed bool
+	l        sync.RWMutex
+	sealed   bool
+	readOnly atomic.Bool
```
DanGhita, line 71: I suppose we need the `readOnly` attribute in the Barrier itself, even if this is more an attribute of the storage/backend...

cipherboy: @DanGhita Yeah, the issue is different backends don't have consistent semantics:

1. `InmemHA` as discussed above is a _really_ weird backend, which only has a single data store and is blindly passed to multiple Cores (active/standby). This is somewhat the same with (removed) backends like S3, GCS, &c.
2. PostgreSQL will read-only err _only_ if talking to a PostgreSQL replica; the PostgreSQL primary DB instance will always accept writes and doesn't differentiate between active/standby nodes (to some extent, that's what #1571 addresses).
3. Raft implements strict single-writer.
4. File isn't HA so this doesn't matter.

So without writing a whole new layer, Barrier is kinda the best place to put this? :shrug: 

A new layer would be >100 LoC (to conditionally support transactions), which seems excessive, given how much smaller and more cleanly this works.

We still allow direct writes to the database of course, but they just can't go through Barrier, which in practice means it is just root key writes that occur this way.

---

@satoqz / @wslabosz-reply: I also like this from a namespace seal / multi-active (single-active namespaces, spread across the cluster): if SealManager sets up multiple Barrier instances, each Barrier can be `readOnly` or not, potentially independently of it the primary/root namespace is read-only or not.

In the interim, though, it might be more coordination work.

Thoughts?

JanMa (approved):
LGTM ✔️

cipherboy:
Thank you all for your reviews and work on the feature! #2089 will be a follow-up.

PR #1987, opened 2025-10-21
Small improvements to read replication

cipherboy:
cc: @phil9909

### DIFF
--- a/vault/ha.go
+++ b/vault/ha.go
@@ -487,6 +487,8 @@ func (c *Core) runStandby(doneCh chan<- struct{}, manualStepDownCh chan struct{}
 
 		var perfCancel context.CancelFunc
 		if c.StandbyReadsEnabled() {
+			c.logger.Info("enabling horizontal scalability (reads)")
+
 			if err := c.runStandbyGrabStateLock(stopCh); err != nil {
 				c.logger.Error("runStandby: unable to grab state lock", "err", err)
 				return


vault/core.go
```diff
@@ -3995,3 +3998,22 @@ func (c *Core) DetectStateLockDeadlocks() bool {
 	}
 	return false
 }
+
+func (c *Core) invalidateKeyrings(path string) {
+	go func() {
+		c.stateLock.Lock()
+		defer c.stateLock.Unlock()
```
cipherboy, line 4006: I'm not actually sure if we need the write state lock for this.

vault/core_cache_invalidate.go
```diff
@@ -90,6 +90,16 @@ func (c *Core) invalidateInternal(ctx context.Context, key string) error {
...
 
+	case namespacedKey == rootKeyPath ||
+		namespacedKey == legacyRootKeyPath ||
+		namespacedKey == keyringPath ||
+		namespacedKey == shamirKekPath ||
+		namespacedKey == StoredBarrierKeysPath ||
+		strings.HasPrefix(namespacedKey, keyringUpgradePrefix):
+
+		// Invalidating keyring uses the same logic as the HA startup code,
+		// just started in a background goroutine and using a
```
cipherboy:
```suggestion
		// just started in a background goroutine.
```

sdk/logical/error.go
```diff
@@ -142,6 +142,7 @@ func ShouldForward(err error) bool {
 	errMsg := err.Error()
 	if strings.Contains(errMsg, ErrPerfStandbyPleaseForward.Error()) ||
 		strings.Contains(errMsg, ErrReadOnly.Error()) ||
+		strings.Contains(errMsg, "node is not the leader") ||
```
phil9909, line 145: Why not like this?

```suggestion
		strings.Contains(errMsg, raft.ErrNotLeader.Error()) ||
```

vault/core.go
```diff
@@ -1079,6 +1083,7 @@ func coreInit(c *Core, conf *CoreConfig) error {
 	}
 
 	if c.StandbyReadsEnabled() {
+		c.logger.Info("enabling horizontal scalability (reads)")
```
phil9909: Should we move this to `runStandby`?

sdk/logical/error.go
cipherboy: @phil9909 Then we incur a whole new dependency in the SDK on Raft, which has been problematic due to architecture build changes (though, that may be limited to the boltdb module).

I don't think importing the package just for one error constant is useful, given this is unlikely to change? We could add a test assertion to the base package, if we want, to ensure it doesn't change ever.

My 2c.

vault/core.go
cipherboy: @phil9909 We can; I think it is just a question of how many times we want to execute this. I'll move it.

cipherboy:
@phil9909 updated!

physical/raft/raft_test.go
```diff
@@ -520,6 +520,12 @@ func TestRaft_Backend_PutTxnMargin(t *testing.T) {
 	}
 }
 
+func TestRaft_LeaderConstant(t *testing.T) {
+	// We check against this exact string in sdk/logical/error.go, to avoid
+	// a full dependency on the hashicorp/raft module.
+	require.Equal(t, raft.ErrNotLeader.Error(), "node is not the leader")
```
phil9909: How about this? (the package already imports `logical`, but not in this file)

```suggestion
	// We check have copied the error string into ShouldForward, to avoid a full
	// dependency on the hashicorp/raft module.
	require.True(logical.ShouldForward(raft.ErrNotLeader))
```

sdk/logical/error.go
phil9909: That's what I suspected. Thanks for clarifying. Should we add a comment here?

```suggestion
		strings.Contains(errMsg, "node is not the leader") || // message copied from hashicorp/raft.ErrNotLeader to avoid importing the package
```

cipherboy:
@phil9909 Revised, I also removed the extraneous check in `Put` and `Delete` now.

phil9909:
Did you see this? https://github.com/openbao/openbao/pull/1987/files#r2451797181

The comment is already marked as resolved and I can't "unresolve" it

cipherboy:
@phil9909 Fairly sure I applied something similar to that, though it didn't build so I repushed: https://github.com/openbao/openbao/pull/1987/commits/ec7857d8f4a863fc6905e7af2f9a36f413747b08

vault/core.go
cipherboy: I confirmed in #2016 that we needed this write lock.

PR #1990, opened 2025-10-21
Enable transactional namespace creation

cipherboy:
Before this, when namespace creation was interrupted, restarting OpenBao would fail with a panic in identity store. This was caused by the namespace entry existing (causing it to show up in ListNamespaces(...)), but the sys/ mount table entry for this namespace not existing. This meant that IdentityStore.AddNamespaceView(...) was never called on restart, so entityPacker(partialNs) would fail, causing unseal to fail.

The only remediation procedure would be to use recovery mode to manually remove the namespace or to create new mount tables manually.

This transactional namespace creation ensures that if namespace creation fails, we do not have any remnants lying around and lets us rollback any in-memory items as well.

This was tested by running something like:

> ```
> $ dist/linux/amd64/benchmark-openbao run \
>     -config=docs/examples/sys/namespaces.hcl -duration=1s
> ```

and hitting ctrl+C randomly. Comparing:

> ```
> $ bao namespace list | wc -l
> ```

before and after restarting the instance will tell you if any in-progress namespaces were not properly deleted.

### DIFF
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -627,6 +627,10 @@ func (ns *NamespaceStore) undoCreateMounts(nsCtx context.Context, namespaceToDel
 		for _, me := range authMountEntries {
 			err := ns.core.disableCredentialInternal(nsCtx, me.Path, false)
 			if err != nil {
+				if errors.Is(err, errNoMatchingMount) {
+					continue
+				}
+
 				ns.logger.Error(fmt.Sprintf("failed to unmount %q", me.Path), "namespace", namespaceToDelete.Path, "error", err.Error())
 				continue
 			}
@@ -964,6 +968,9 @@ func (ns *NamespaceStore) clearNamespaceResources(nsCtx context.Context, namespa
 	for _, me := range authMountEntries {
 		err := ns.core.disableCredentialInternal(nsCtx, me.Path, true)
 		if err != nil {
+			if errors.Is(err, errNoMatchingMount) {
+				continue
+			}
 			ns.logger.Error(fmt.Sprintf("failed to unmount %q", me.Path), "namespace", namespaceToDelete.Path, "error", err.Error())
 			return false
 		}
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -247,7 +247,7 @@ func (c *Core) disableCredentialInternal(ctx context.Context, path string, updat
 	// Verify exact match of the route
 	match := c.router.MatchingMount(ctx, path)
 	if match == "" || ns.Path+path != match {
-		return errors.New("no matching mount")
+		return errNoMatchingMount
 	}
 
 	// Get the view for this backend


vault/namespace_store.go
```diff
@@ -498,14 +593,65 @@ func (ns *NamespaceStore) createMounts(ctx context.Context) error {
...
+	mountEntries, err := ns.core.mounts.findAllNamespaceMounts(nsCtx)
+	if err != nil {
+		ns.logger.Error("failed to retrieve namespace mounts", "namespace", namespaceToDelete.Path, "error", err.Error())
+	} else {
+		for _, me := range mountEntries {
+			err := ns.core.unmountInternal(nsCtx, me.Path, false)
+			if err != nil {
+				if errors.Is(err, errNoMatchingMount) {
+					continue
+				}
```
wslabosz-reply, line 650: I see it's just to not log an error when we cannot find the mount. Any reason we don't do the same for auth mounts just before?

cipherboy: @wslabosz-reply Ahh, I see, we missed this in https://github.com/openbao/openbao/commit/ed3b3fb9482; this is the same logic as `clearNamespaceMounts`, just limited to things which don't write to storage as `ctx` has (potentially) been cancelled and _most_ everything else is in a transaction.

cipherboy: @wslabosz-reply Updated!

wslabosz-reply (approved):
thanks @cipherboy

vault/namespace_store.go
```diff
@@ -381,9 +386,74 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
...
+	if txnable, ok := storage.(logical.TransactionalStorage); ok {
+		ns.logger.Info("using transaction to manage storage")
+
+		txn, err := txnable.BeginTx(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to begin transaction: %w", err)
+		}
+
+		defer func() {
+			if err := txn.Rollback(ctx); err != nil {
```
chrom9k, line 438: What is the reason for handling the error resulting from executing Rollback method?

cipherboy: I think we should perhaps ignore `errors.Is(err, physical.ErrTransactionAlreadyCommitted)`, but in general it is a signal of a failing storage backend.

satoqz: > I think we should perhaps ignore `errors.Is(err, physical.ErrTransactionAlreadyCommitted)`

Definitely, this throws quite the unnerving log message, which I suppose is why @chrom9k arrived here 😄 FWIW @cipherboy most other code doesn't handle the error returned by `txn.Rollback` at all.

cipherboy: @satoqz tbf I'm fairly sure I only handled it because the new linter complained 😆 

I'll open a PR to clean this up and add `nolint` given most everywhere else ignores it.

PR #1991, opened 2025-10-22
Pin many package verisons to hashes

cipherboy:
First step in addressing many of the Scorecard requirements around package pinning.

- All workflows should now be pinned.
- Removed some unnecessary shell scripts
- Pinned the first of the Dockerfiles; however, it seems like ones under `scripts/` are currently being ignored by Dependabot (le sigh). 

---

cc: @weberval @karras @openbao/openbao-org-maintainers

### DIFF
--- a/website/Dockerfile
+++ b/website/Dockerfile
@@ -1,7 +1,7 @@
 ## Dockerfile adapted from https://docusaurus.community/knowledge/deployment/docker/.
 
 ## Start with a base image containing NodeJS so we can build Docusaurus.
-FROM node:lts
+FROM node:lts@sha256:915acd9e9b885ead0c620e27e37c81b74c226e0e1c8177f37a60217b6eabb0d7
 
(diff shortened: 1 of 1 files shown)


JanMa (approved):

cipherboy:
> Would be nice to make the version comments behind the action hashes match the exact semver version for more transparency (instead of just "v5"), but LGTM.

@satoqz Happy to update, though I expecting Dependabot to eventually do that on next updates?

satoqz:
> though I expecting Dependabot to eventually do that on next updates

@cipherboy I'm not sure, hence I always make sure to make it as easy as possible on Dependabot to take off from the string I've given it 😄 But we can just leave it as-is, and run the experiment, then adjust if needed.

cipherboy:
@satoqz Let's experiment with dependabot and see how newly dependable it is after everyone's fixes to it. :laughing:

PR #1994, opened 2025-10-22
Add check that builds bao with the go on el8 (#1937)

cipherboy:
Backport of #1937 to `release/2.4.x`.

---

* add check that builds bao with the go on el8

* skip el8gobuild if there is a "no-el8gobuild" label

---------

Resolves #


satoqz (approved):

cipherboy:
Merging based on review from @satoqz: as this is a backport of the check, I think this is sufficient rather than waiting for an @openbao/openbao-org-maintainers review in this case.

PR #2002, opened 2025-10-22
Refactor audit log formatting due to copy

cipherboy:
When cloning the to be logged structure, the audit subsystem attempted
to walk the original copy, modifying values in the cloned. This resulted
in a few cases where the subsystem would reject response auditing due to
an inability to correctly walk both in parallel.
    
Additionally, the HMAC'd HTTPRawBody would not be correctly HMAC'd as
the original copy is usually of type []byte, a non-HMAC'd type.

---

Appreciate @phil9909 and @satoqz for reporting and @wslabosz-reply for pairing with me while I wrote the fix and adding tests. 

(fyi @suprjinx)

### DIFF
--- a/audit/format.go
+++ b/audit/format.go
@@ -572,20 +572,23 @@ func NewTemporaryFormatter(format, prefix string) *AuditFormatter {
 	return ret
 }
 
-// doElideListResponseData performs the actual elision of list operation response data, once surrounding code has
-// determined it should apply to a particular request. The data map that is passed in must be a copy that is safe to
-// modify in place, but need not be a full recursive deep copy, as only top-level keys are changed.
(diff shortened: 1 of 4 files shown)


changelog/2002.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:security
+audit: redact `HTTPRawBody` response parameter in audit logs; CVE-2025-62513 / GHSA-ghfh-fmx4-26h8.
+```
+```release-note:security
+audit: redact `[]byte` type response parameters in audit logs; GHSA-rc54-2g2c-g36g.
```
cipherboy, line 5: As I noted in the GHSA, I'm waiting on @github-staff for CVE assignment; we'll update these on the website &c after it is present.

PR #1870, opened 2025-09-30
feat(goreleaser): publish packages to S3

pree:
Create proper release repositories for APT and RPM releases.

* changes .deb package names from `bao(-hsm)` to `openbao(-hsm)` with `bao` in `provides/replaces/conflicts`. This should be removed in a couple of releases (for example 2.7.0)
  * Packages are usually named `openbao` on other repositories: https://repology.org/project/openbao/versions

PR #2010, opened 2025-10-22
Clean go build cache after each Linux architecture

cipherboy:
Since v2.4.2 failed in nightly due to out-of-disk (and sadly, I wasn't patient enough to wait for the full Linux build to succeed before pushing the full tag)...

...we're going to v2.4.3. Luckily we have a fairly simple fix right now: `go cache -clean` after each build.

I can't actually find documentation that this is safe to call in parallel with running builds, but we don't really have much choice given how few hooks Goreleaser has for this. 

This passed: https://github.com/openbao/openbao-nightly/actions/runs/18727547006/job/53416615762


pree:
You can also delete some stuff from the runner before running goreleaser. I've added this to the package repositories PR for my testings, as the runner is out of storage otherwise: https://github.com/openbao/openbao/pull/1870/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R196

satoqz:
> `sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /opt/hostedtoolcache/CodeQL`

beautiful 😄

PR #2012, opened 2025-10-23
grab state lock before closing standbyStop channel

phil9909:
in `sealInternalWithOptions` this was removed here: https://github.com/openbao/openbao/pull/1847/commits/cef50884967e184e2e9a6ca6322dbba29c79f8db#diff-a0a40b411f03b3c927d6a2eafaf8a17693cc05a44e81bfe1362045a9372b9cfeL2217-L2220 by @cipherboy

Was this an oversight? I think it would probably fix the same issue as adding the lock grab timeout in `runStandby` (but causes a race in `TestSysMonitorStreamingLogs`). So probably we don't need both changes.

See #1550
See #1528 

### DIFF
--- a/vault/core.go
+++ b/vault/core.go
@@ -2224,6 +2224,10 @@ func (c *Core) sealInternalWithOptions(grabStateLock, keepHALock, performCleanup
 		if keepHALock {
 			atomic.StoreUint32(c.keepHALockOnStepDown, 1)
 		}
+		if grabStateLock {
+			cancelCtxAndLock()
+			defer c.stateLock.Unlock()
+		}
 
 		// If we are trying to acquire the lock, force it to return with nil so
 		// runStandby will exit
--- a/vault/ha.go
+++ b/vault/ha.go
@@ -384,6 +384,9 @@ func (c *Core) StepDown(httpCtx context.Context, req *logical.Request) (retErr e
 }
 
 func (c *Core) stopStandby() {
+	c.stateLock.Lock()
+	defer c.stateLock.Unlock()
+
 	select {
 	case c.standbyStopCh.Load().(chan struct{}) <- struct{}{}:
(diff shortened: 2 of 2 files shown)


vault/ha.go
```diff
@@ -384,6 +384,9 @@ func (c *Core) StepDown(httpCtx context.Context, req *logical.Request) (retErr e
 }
 
 func (c *Core) stopStandby() {
+	c.stateLock.Lock()
```
cipherboy, line 387: If the `standbyStopCh` is atomic, why do we need `stateLock` here? It seems both redundant and latency-inducing. 

Should we instead split this into two?

```go
stopCh := c.standbyStopCh.Load()

if stopCh == nil {
  return 
}

select {
  ...
}
```

vault/core.go
```diff
@@ -2224,6 +2224,10 @@ func (c *Core) sealInternalWithOptions(grabStateLock, keepHALock, performCleanup
 		if keepHALock {
 			atomic.StoreUint32(c.keepHALockOnStepDown, 1)
 		}
+		if grabStateLock {
+			cancelCtxAndLock()
+			defer c.stateLock.Unlock()
```
cipherboy, line 2229: My memory was that this caused deadlocks in the shutdown process, but I wasn't sure _why_. I think it happened on standby nodes? 🤔 

But I would find that surprising. Let me see what #2016 yields.

vault/ha.go
cipherboy: It also doesn't seem to cause problems, though 😆 

I wonder if RLock is sufficient?

vault/core.go
cipherboy: It looks fine from this PoV, maybe I had other bugs at the time. 😆

PR #2014, opened 2025-10-23
fix nil issues and race condtions on standby shutdown

phil9909:
- add nil checks to the policy store invalidation and `reloadLegacyMounts`
- make the restart channel in core atomic 

See #1550
See #1528 

### DIFF
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -2543,6 +2543,9 @@ func (c *Core) reloadLegacyMounts(ctx context.Context, keys ...string) error {
 	c.mountsLock.RLock()
 	c.authLock.RLock()
 	for _, table := range []*MountTable{c.mounts, c.auth} {
+		if table == nil {
+			continue
+		}
 		for _, entry := range table.Entries {
 			if ns.ID != namespace.RootNamespaceID && ns.ID != entry.NamespaceID {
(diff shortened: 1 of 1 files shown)


cipherboy:
@phil9909 Besides the test failures, it makes me wonder if we should instead just add that physical layer. Thoughts?

phil9909:
> @phil9909 Besides the test failures, it makes me wonder if we should instead just add that physical layer. Thoughts?

Which physical layer? A decorator that handles<sup>1</sup> the read-only error?

<sup>1</sup> injects

cipherboy:
@phil9909 Yeah, exactly.

phil9909:
> @phil9909 Yeah, exactly.

I guess that would make life easier for PostgreSQL, which would need something similar to what I've built here, right?

Sure, why not.




> Besides the test failures

questioning my life choices meme

cipherboy:
> I guess that would make life easier for PostgreSQL, which would need something similar to what I've built here, right?

Well, PostgreSQL will yield a lock error I think actually, but yes.

But it avoids the expansion of `ShouldForward` with random physical backend's stuff and allows us to get to a place where the backend is pluggable. My 2c.

Issue #1774, opened 2025-09-05
Refactor: audit and secrets mount tables duplication


phil9909:
While working on the mount tables for cache invalidation I noticed a fair bit of duplication:

1. Between the logic for audits and secrets mounts, e.g. `setupAudits` and `setupMounts`  https://github.com/openbao/openbao/blob/43bc4b117a245847fd24ca8f7ed4ae63f6b8d9e3/vault/mount.go#L1870-L1986 https://github.com/openbao/openbao/blob/43bc4b117a245847fd24ca8f7ed4ae63f6b8d9e3/vault/audit.go#L393-L443
1. Between loading mounts on startup and loading mounts via API call.

So the logic for mounting is duplicated 4 times, albeit with major differences. The open question here is, which of this differences are important and which are accidental. 

Maybe we can unify the common logic and add hooks for the specifics.

I acknowledge that often times duplicating code is fine and we shouldn't make [`DRY`](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) a religion. But in this case I think the logic is complex enough and already starts to drift apart, that the added complexity is worth it.

satoqz:
@phil9909 have you looked at auth mounts, too? Try comparing `auth.go` to `mount.go`, see if you notice anything 😆

cipherboy:
💯 , I'd love to see this happen.

wslabosz-reply:
I'm gonna start working on this one, just out of spite of the complexity in `vault` package.

phil9909:
> I'm gonna start working on this one, just out of spite of the complexity in `vault` package.

Nice, please keep the cache invalidation topic in mind that we are developing on [feature branch `read-replication`](https://github.com/openbao/openbao/tree/read-replication)

If you want some feedback on early drafts, I'm happy to look at half finished PoC implementations or napkin drawings.

cipherboy:
@wslabosz-reply Have you given any thought to unifying `audit` into the regular plugin architecture as well? I'm not entirely sure if it is a separate ask or if it just kinda makes sense to do now to get the three tables (auth, audit, and mount) down to one fundamental type (all being pluggable).

PR #2016, opened 2025-10-24
Clean up write enablement

cipherboy:
This applies three PRs for CI testing:

1. #1987 
2. #2012 
3. #2014, minus the inmem HA read-only parts, and replaced with a small change to Barrier to apply globally. 

---

On top of this, it introduces:

1. A major refactor to the invalidation system, to execute invalidations in a worker system, allowing `vault.Core`-associated jobs to continue past a short invalidation time.
2. Allow batching of invalidations, requiring fewer lock acquisitions when e.g., using a Transaction or Log Bundle with Raft. 
3. Allow marking the entire barrier read-only in one shot, preventing arbitrary writes.
4. Fix race during sealing of active HA node.

As a result of 3, this fixes various test cases that depended on the old In-Mem logic (i.e., that standbys could write to storage, which was unique to that backend). 

And various other small cleanups.

### DIFF
--- a/vault/external_tests/response/allowed_response_headers_test.go
+++ b/vault/external_tests/response/allowed_response_headers_test.go
@@ -65,7 +65,8 @@ func TestIdentityStore_EntityDisabled(t *testing.T) {
 		},
 	}
 	cluster := vault.NewTestCluster(t, coreConfig, &vault.TestClusterOptions{
-		HandlerFunc: vaulthttp.Handler,
+		HandlerFunc:         vaulthttp.Handler,
+		DisableStandbyReads: true,
 	})
 	cluster.Start()
 	defer cluster.Cleanup()


vault/core.go
```diff
@@ -651,12 +649,20 @@ type Core struct {
...
+	invalidations *invalidationManager
 }
 
 // c.stateLock needs to be held in read mode before calling this function.
 func (c *Core) HAState() consts.HAState {
 	switch {
 	case c.standby.Load():
+		if c.StandbyReadsEnabled() {
+			return consts.PerfStandby
+		}
```
phil9909, line 664: I think we initially did want to distinguish between "normal" and performance standby (as performance standby / standby reads is not "paywalled")

I'm fine with adding this distinction, since we added a flag to explicitly disable standby reads.

vault/core_cache_invalidate.go
```diff
@@ -5,97 +5,484 @@ package vault
...
+
+	im.pending = append(im.pending, key)
+
+	select {
+	case im.pendingNotify <- struct{}{}:
+	default:
+	}
+}
+
+func (im *invalidationManager) splitNamespaceFromKey(key string) (string, string) {
```
phil9909, line 504: Should we move this to the namespace package?

vault/core_cache_invalidate.go
```diff
@@ -5,97 +5,484 @@ package vault
...
 	"github.com/openbao/openbao/helper/namespace"
 	"github.com/openbao/openbao/sdk/v2/physical"
 	"github.com/openbao/openbao/vault/quotas"
 )
 
+const (
+	dispatcherName    = "invalidate-dispatch"
+	refresherName     = "invalidate-cache-refresh"
+	maxInvalidateTime = 30 * time.Second
+	maxDispatchers    = 128
```
phil9909: For mounts. this means we reintroduce the possibility to have an old and slow invalidation overwrite data of a newer and faster invalidation. It's only possible, if the land in different queues, but I think the namespace invalidation will land in a different one than the mount table itself.

vault/core_cache_invalidate.go
```diff
@@ -5,97 +5,484 @@ package vault
...
+		return ij.legacyMountInvalidation(ctx)
+	case isTransactionalMountPath(key):
+		ij.fatal = true
+		return ij.transactionalMountInvalidation(ctx)
+	case isKeyringPath(key):
+		// Invalidating keyring uses the same logic as the HA startup code,
+		// just started in a background goroutine. It isn't fatal since the
+		// goroutine handles fatal errors.
+		return ij.keyringInvalidation(ctx)
+	case ij.im.core.router.Invalidate(ctx, key):
```
phil9909:
```suggestion
	case ij.im.core.router.Invalidate(ctx, ij.key):
```

the router reads the namespace from the context

vault/core_cache_invalidate_test.go
```diff
@@ -169,7 +172,7 @@ func TestCore_Invalidate_Namespaces(t *testing.T) {
 	}, 10*time.Second, 10*time.Millisecond)
 }
 
-func TestCore_Invalidate_Namespaces_NonTransactional(t *testing.T) {
+/*func TestCore_Invalidate_Namespaces_NonTransactional(t *testing.T) {
```
phil9909:
```suggestion
func TestCore_Invalidate_Namespaces_NonTransactional(t *testing.T) {
```

Just ran this test locally, it works.

vault/core_cache_invalidate_test.go
```diff
@@ -250,7 +253,7 @@ func TestCore_Invalidate_Namespaces_NonTransactional(t *testing.T) {
 		resp = testCore_Invalidate_handleRequest(collect, namespace.RootContext(t.Context()), c, req)
 		require.NotNil(collect, resp)
 	}, 10*time.Second, 10*time.Millisecond)
-}
+}*/
```
phil9909:
```suggestion
}
```

vault/core_cache_invalidate_test.go
```diff
@@ -408,8 +411,8 @@ func TestCore_Invalidate_Plugin(t *testing.T) {
 			uuid := resp.Data["uuid"].(string)
 
 			// 4. Invalidate Paths
-			c.Invalidate(nsPrefix + "logical/" + uuid + "/foo")
-			c.Invalidate(nsPrefix + "logical/" + uuid + "/bar/bazz")
+			c.invalidateSynchronous(nsPrefix + "logical/" + uuid + "/foo")
+			c.invalidateSynchronous(nsPrefix + "logical/" + uuid + "/bar/bazz")
 
 			// 5. Check callback was called
 			assert.Equal(t, invalidatedKey, []string{"foo", "bar/bazz"})
```
phil9909:
```suggestion
			assert.Equal(t, []string{"foo", "bar/bazz"}, invalidatedKey)
```

`Equal` expects "expected" first and "actual" second. My mistake, sorry.

vault/core_cache_invalidate.go
cipherboy: > For mounts. this means we reintroduce the possibility to have an old and slow invalidation overwrite data of a newer and faster invalidation.

@phil9909 Why do you say this? I agree w.r.t. ordering of invalidations (an older invalidation may run after a newer invalidation).

But, invalidation does not depend on the _data_: if a newer invalidation runs before an older one, it menas data has caught up to the newer point. So yes, the "old" invalidation still runs, but it sees an even newer storage state, and so will not "revert".

phil9909: The problem is invalidations are not atomic

Suppose this:
We have two invalidations A and B:
- A starts being processed, reads data from storage and blocks on waiting for a lock
- B starts being processed, reads data from storage, wins the battle for the lock and applies updates to the mount table.
- A finally is able to get the lock and reverts changes done by B

cipherboy: @phil9909 The general rules about _plugin_ invalidations hold; I'll document these in https://openbao.org/docs/plugins/plugin-authors-guide/ While we've modified core to allow long-lived invalidations, plugins should still use the atomic bool operation.

---

W.r.t. the mount table specifically, I think this goes back to my original assertion in #1733 --- we should hold locks for the mount table longer. In Namespace store, we will hold locks longer (once the invalidation is processed--it uses the atomic bool approach at the moment). 

But, with this design, we've fixed the issue of long-running changes in general: physical backend's invalidation hook does grab locks...

Let's give this a go and see if we need to adjust locking.

cipherboy:
@phil9909 I applied your changes and also others which should help with the tests. If we're in a similar state as other branches, I'll start merging the PR queues down.

vault/core_cache_invalidate.go
phil9909: Yes, I remember your comment in #1733. And with the change to a single threaded worker to process mount invalidations we "accidentally" also fixed that problem (because on standbys invalidations should be only piece modfiing the mount table after start-up).

I agree that we ignore this for now and see, if this turns out to be problematic. Maybe we can fix this in #1774.
Fixing it probably requires "lifting" the lock from methods that modify the mount table up into the invalidation logic.

vault/core_cache_invalidate.go
cipherboy: @phil9909 I think we should leave this here; namespaces package isn't concerned about storage layout at all at the moment.

vault/core.go
cipherboy: @phil9909 HAState is only called by the metric subsystem, which I think is useful to have a differentiation between "active", "read-enabled standby", and "standby" at a glance.

physical/raft/fsm.go
```diff
@@ -933,19 +933,20 @@ func (f *FSM) ApplyBatch(logs []*raft.Log) []interface{} {
...
+					switch op.OpType {
+					case putOp, deleteOp:
+						keys = append(keys, op.Key)
 					}
 				}
 			}
-		}()
+		}
+
+		go f.invalidateHook(keys...)
```
cipherboy, line 949: @phil9909 I was not able to fully make this a blocking call. With the lock acquisition, I think there was a moderately circular dependency _somewhere_, so I opted to keep this as-is. But, with the batched variant, we can acquire fewer locks.

physical/raft/raft.go
```diff
@@ -208,19 +208,8 @@ type RaftBackend struct {
...
-		r.l.RUnlock()
-
-		if raft == nil {
-			return
-		}
-		_, leaderId := raft.LeaderWithID()
-
-		if r.localID != string(leaderId) {
-			hook(key)
-		}
```
cipherboy, line 223: This logic is all handled in the hook now, @phil9909, making it more robust to misuse.

Should I even bother with this or should I just have `raft.HookInvalidate` call `r.fsm.hookInvalidate(hook)`?

vault/core_cache_invalidate.go
```diff
@@ -5,120 +5,484 @@ package vault
...
+		key == StoredBarrierKeysPath ||
+		key == barrierSealConfigPath ||
+		key == recoverySealConfigPath ||
+		key == recoveryKeyPath ||
+		strings.HasPrefix(key, keyringUpgradePrefix)
+}
+
+func (ij *invalidationJob) Execute() error {
+	ij.im.dispacherLogger.Trace("issuing invalidation", "key", ij.key, "ns", ij.nsUUID)
+	defer ij.im.dispacherLogger.Trace("finished invalidation", "key", ij.key, "ns", ij.nsUUID)
```
cipherboy: Should add metrics here as well.

physical/raft/raft.go
cipherboy: I opted for the latter.

vault/ha.go
```diff
@@ -485,126 +493,123 @@ func (c *Core) runStandby(doneCh chan<- struct{}, manualStepDownCh chan struct{}
...
-		if c.StandbyReadsEnabled() {
-			c.logger.Info("enabling horizontal scalability (reads)")
+	c.logger.Info("runStandby stopped")
+}
 
-			if err := c.runStandbyGrabStateLock(stopCh); err != nil {
-				c.logger.Error("runStandby: unable to grab state lock", "err", err)
-				return
-			}
+func (c *Core) runStandbyOnce(doneCh chan<- struct{}, manualStepDownCh chan struct{}, stopCh, restartCh <-chan struct{}) bool {
```
cipherboy, line 502: @phil9909 I refactored this so it was easier to diff against `main`.

cipherboy:
@phil9909 Welllll, got it CI clean, but we still have that race on shutdown between audit log disabling and pending requests.

phil9909:
Failing tests:

```sh
go test github.com/openbao/openbao/vault/external_tests/plugin -run="TestExternalPlugin_AuthMethod/parallel_execution_group/vault-plugin-auth-approle-2"
go test github.com/openbao/openbao/vault/external_tests/plugin -run="TestExternalPlugin_AuthMethod/parallel_execution_group"
go test github.com/openbao/openbao/vault/external_tests/plugin -run="TestExternalPlugin_AuthMethod"
go test github.com/openbao/openbao/vault -run="TestCore_Standby_Seal" -race
go test github.com/openbao/openbao/command -run="TestAgent_Cache_DynamicSecret" -race
go test github.com/openbao/openbao/command -run="TestAgent_Logging_ConsulTemplate" -race
# summary: 6 failures of 13372 cases (3 failures with race detector)
```

3 of these failures are the same test, so we are down to 4 distinct failures

http/util.go
```diff
@@ -34,8 +34,13 @@ var (
 	adjustResponse = func(core *vault.Core, w http.ResponseWriter, req *logical.Request) {}
 )
 
+func isSnapshotRequest(req *http.Request) bool {
+	return strings.HasSuffix(req.URL.Path, "/sys/storage/raft/snapshot") ||
+		strings.HasSuffix(req.URL.Path, "/sys/storage/raft/snapshot-force")
```
phil9909: CVE incoming :laughing:

vault/core_cache_invalidate.go
```diff
@@ -5,120 +5,509 @@ package vault
...
+	// should be sufficient to prevent races here, we want to ensure no core
+	// state changes during processing of the request. We bind this to our
+	// time-limited channel to ensure it processes in a reasonable amount of
+	// time.
+	l := newLockGrabber(ij.im.core.stateLock.RLock, ij.im.core.stateLock.RUnlock, lockCtx.Done())
+	go l.grab()
+
+	if stopped := l.lockOrStop(); stopped {
+		// Failure to grab a lock is never fatal: HA state change will mean
+		// that we'll restart the invalidation manager anyways.
```
phil9909, line 309: To me, this feels a bit brittle.

vault/external_tests/identity/userlockouts_test.go
```diff
@@ -36,7 +36,8 @@ func TestIdentityStore_DisableUserLockoutTest(t *testing.T) {
 		},
 	}
 	cluster := vault.NewTestCluster(t, coreConfig, &vault.TestClusterOptions{
-		HandlerFunc: vaulthttp.Handler,
+		HandlerFunc:         vaulthttp.Handler,
+		DisableStandbyReads: true,
```
phil9909, line 40: Something we should document I guess.

vault/logical_system_paths.go
```diff
@@ -1306,6 +1306,10 @@ func (b *SystemBackend) statusPaths() []*framework.Path {
 									Type:     framework.TypeInt64,
 									Required: false,
 								},
+								"grpc_connected": {
+									Type:     framework.TypeString,
+									Required: true,
+								},
```
phil9909: I don't like the name. `request_forwarding_connection_established` seems a bit long.

http/util.go
cipherboy: @phil9909 Meaning?

vault/logical_system_paths.go
cipherboy: @phil9909 We can drop this, it was for the test case only.

vault/core_cache_invalidate.go
cipherboy: @phil9909 Why? It is an ordering problem, active context is not cancelled until after state lock is grabbed.

http/util.go
phil9909: This has false-positives like `/v1/mymount/sys/storage/raft/snapshot`

phil9909: I'm not sure, if this is exploitable somehow, but it puts our code into unexpected paths, which is a good start for finding an exploit:

```bash
$ bao kv put -mount=secret sys/storage/raft/snapshot hello=world
Error writing data to secret/data/sys/storage/raft/snapshot: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/secret/data/sys/storage/raft/snapshot
Code: 400. Errors:

* no data provided
```

vault/core_cache_invalidate.go
phil9909: The assumption: If we can't lock within two seconds, there must be some HA state change in progress. It might be true today, but will this be true next year?

http/util.go
cipherboy: @phil9909 Fixed to align with the existing check here:

https://github.com/openbao/openbao/blob/1b56d6d39f604e308e91e4c3ebf3864dd6c5dedc/http/logical.go#L128

vault/core_cache_invalidate.go
cipherboy: @phil9909 I don't see why it wouldn't be; grabbing a write state lock is a rare occurrence. If it isn't true in two years, we've added a new form of replication and would need to modify invalidation entirely.

cipherboy: @phil9909 More concretely, if you have proposals post-merge, I'm happy to see them.

phil9909: The thing is: if this assumption breaks, we don't necessarily look at this part of the code and notice we have to adapt it. This is just something I try to avoid (i.e. hard-coded assumptions, that are hard to verify via tests or compiler). I'm not saying this is a bug, but it might become one (which is probably true for most code :laughing:, it just seem that this has higher than average probability of becoming problematic)

Anyway, we can still merge this. Maybe I can come up with something better.

cipherboy:
Merging based on approval in [#openssf-openbao-wg-scalability > Cache Invalidation @ 💬](https://linuxfoundation.zulipchat.com/#narrow/channel/532998-openssf-openbao-wg-scalability/topic/Cache.20Invalidation/near/547691275)

Issue #2015, opened 2025-10-23
segfault crash during step down


rbastiaans-tc:
**Describe the bug**

We were performing a "rolling restart" of 1 of our OpenBao clusters, due to a config file update to enable proxy protocol support.

This is a cluster of 3 OpenBao nodes
* we first restarted both standby OpenBao nodes
* then performed a raft step-down of the active leader (to prepare for the last node restart)
* the leader then crashed with a segmentation violation while it was "stepping down"

```
2025-10-23T08:24:29.033Z [ERROR] storage.raft: failed to appendEntries to: peer="{Voter openbao-xxx003 openbao-xxx003:8201}" error="remote error: tls: internal error"
2025-10-23T08:24:29.121Z [ERROR] storage.raft: failed to appendEntries to: peer="{Voter openbao-xxx003 openbao-xxx003:8201}" error="dial tcp 1.2.3.139:8201: connect: connection refused"
2025-10-23T08:24:29.168Z [ERROR] storage.raft: failed to heartbeat to: peer=openbao-xxx003:8201 backoff time=10ms error="dial tcp 1.2.3.139:8201: connect: connection refused"
2025-10-23T08:24:29.227Z [ERROR] storage.raft: failed to appendEntries to: peer="{Voter openbao-xxx003 openbao-xxx003:8201}" error="dial tcp 1.2.3.139:8201: connect: connection refused"
2025-10-23T08:24:29.270Z [ERROR] storage.raft: failed to appendEntries to: peer="{Voter openbao-xxx003 openbao-xxx003:8201}" error="dial tcp 1.2.3.139:8201: connect: connection refused"
2025-10-23T08:24:29.450Z [ERROR] storage.raft: failed to appendEntries to: peer="{Voter openbao-xxx003 openbao-xxx003:8201}" error="dial tcp 1.2.3.139:8201: connect: connection refused"
2025-10-23T08:24:29.671Z [INFO]  storage.raft: pipelining replication: peer="{Voter openbao-xxx001 openbao-xxx001:8201}"
2025-10-23T08:24:29.694Z [ERROR] storage.raft: failed to appendEntries to: peer="{Voter openbao-xxx003 openbao-xxx003:8201}" error="dial tcp 1.2.3.139:8201: connect: connection refused"
[...]

PR #2019, opened 2025-10-24
Fix potential segfault during step down

phil9909:
This does two things:

- check for context cancellation in the namespace store. The context should be canceled before setting the namespaceStore to `nil`. If a context is passed, that is not derived from the activeContext, a nil pointer might still occur.
- check for context cancellation  in attemptIrrevocableLeasesRevoke loop (also adds a call to `stateLock` because previously this was racey).

~~Resolves~~ See #2015

### DIFF
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -425,8 +425,10 @@ func (c *Core) setupExpiration(e ExpireLeaseStrategy) error {
 			case <-quit:
 				return
 			case <-t.C:
-				c.expiration.attemptIrrevocableLeasesRevoke()
-				t.Reset(24 * time.Hour)
+				c.expiration.jobManager.AddJob(fairshare.SimpleJob(func() {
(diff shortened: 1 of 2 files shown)


vault/expiration.go
```diff
@@ -949,14 +954,22 @@ func (m *ExpirationManager) lazyRevokeInternal(ctx context.Context, leaseID stri
 
 // should be run on a schedule. something like once a day, maybe once a week
 func (m *ExpirationManager) attemptIrrevocableLeasesRevoke() {
+	m.core.stateLock.RLock()
```
cipherboy: I think we now need to use the pattern of a regular job here, and check the quitCh in addition here.

cipherboy: https://github.com/openbao/openbao/blob/45983e33b556bd407a3a925c2d8a3ced78eaa8c8/vault/expiration.go

In particular.

PR #2020, opened 2025-10-24
use deploy key for dependency synchronization

JanMa:
In https://github.com/openbao/openbao/pull/1966 we added a workflow to run `make sync-deps` on Dependabot PRs to automatically keep dependencies in sync across all modules. As it turns out using the default `GITHUB_TOKEN` provided to a workflow run when committing changes will _not_ trigger any new workflow runs for the new commit. To do that, you need to use a PAT, a Deploy Key or a GitHub App to make the commit.

This PR introduces the use of a deploy key in the sync deps workflow to ensure the required workflows are triggered when `make sync-deps` commits any changes.


cipherboy (approved):
This is strange to me that checkout action influences a later `git push` but I've seen stranger things. Thanks @JanMa for your patience and experimentation!

PR #2021, opened 2025-10-26
Update hashicorp/hcl to 1.0.1-vault-7

JanMa:
This bumps `hashicorp/hcl` to the latest version and adjusts our HCL parsing code to deal with no longer allowed duplicate Keys in HCL.

### DIFF
--- a/helper/random/parser.go
+++ b/helper/random/parser.go
@@ -7,14 +7,10 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-	"strings"
 	"unicode/utf8"
 
 	"github.com/go-viper/mapstructure/v2"
 	"github.com/hashicorp/hcl"
-	"github.com/hashicorp/hcl/hcl/ast"
-	hclParser "github.com/hashicorp/hcl/hcl/parser"
-	jsonParser "github.com/hashicorp/hcl/json/parser"
 )
 
 // ParsePolicy is a convenience function for parsing HCL into a StringGenerator.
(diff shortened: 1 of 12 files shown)


cipherboy:
@JanMa When we merge this we should probably add a deprecation to align ourselves with newer Vault, which dropped support for duplicate keys.

helper/random/parser.go
```diff
@@ -157,3 +161,24 @@ func stringToRunesFunc(from reflect.Kind, to reflect.Kind, data interface{}) (in
 	}
 	return []rune(raw), nil
 }
+
+// detectConfigFormat determines if the input is JSON or HCL format
+func detectConfigFormat(data []byte) bool {
```
wslabosz-reply: can we rename this to `detectJSONConfigFormat`? right now it doesn't imply that `true` value returned from this function means the input is a JSON (or HCL)

JanMa:
> I feel like it's confusing how `ParseConfig([]byte)` is imported as `random`, suggest that there's something random about parsing a config, but it's where the function is declared (`random` package), can we move/rename the function/package, or alias the import?

I initially put it into `helper/random` because it already contains parsing related functions. But your're right the naming is a bit confusing so I moved the helper function to `sdk/helper/hclutil`, this might be a better fit.

api/ssh_agent.go
```diff
@@ -145,10 +148,29 @@ func LoadSSHHelperConfig(path string) (*SSHHelperConfig, error) {
 	return ParseSSHHelperConfig(string(contents))
 }
 
+// isJson determines if the input is JSON, i.e if it starts with '{'
+func isJson(data []byte) bool {
+	// Trim whitespace and check if it starts with '{'
+	trimmed := strings.TrimSpace(string(data))
+	return strings.HasPrefix(trimmed, "{")
+}
```
wslabosz-reply, line 156: is there a reason this function is here? couldn't we use `hclutil.isJson()` (after importing it) at this point?

JanMa: That would introduce a dependency between `api` and `sdk` which we currently don't have

wslabosz-reply: okay, so I'm fine with leaving it as it is

cipherboy: Right, we can't depend on SDK from the API, otherwise we'll get a circular dependency (SDK depends on API).

PR #2029, opened 2025-10-28
Read replication fix races

phil9909:
All this races seem to occur on teardown. I'm wondering why this is not happening on main

See #1550
See #1528 


vault/audit.go
```diff
@@ -169,7 +169,7 @@ func (c *Core) enableAudit(ctx context.Context, entry *MountEntry, updateStorage
 	c.audit = newTable
 
 	// Register the backend
-	c.auditBroker.Register(entry.Path, backend, view, entry.Local)
+	c.auditBroker.Load().Register(entry.Path, backend, view, entry.Local)
```
cipherboy: I think this change is superseded in #2016 but the rest are fair game to pick up still I think.

PR #2030, opened 2025-10-28
Final cleanup to read replication branch

cipherboy:
~This is based on #2029; will be rebased when that merges.~

----

1. Additional invalidation skips based on CI results.
2. Simplify when invalidations get tracked.
3. Expand locking on routes to further prevent additional race conditions.

### DIFF
--- a/vault/router.go
+++ b/vault/router.go
@@ -241,6 +241,9 @@ func (r *Router) Unmount(ctx context.Context, prefix string) error {
 
 	// Call backend's Cleanup routine
 	re := raw.(*routeEntry)
+	re.l.Lock()
+	defer re.l.Unlock()
+
 	if re.backend != nil {
 		re.backend.Cleanup(ctx)
 	}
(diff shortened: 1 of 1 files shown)


vault/core_cache_invalidate.go
```diff
@@ -406,6 +394,15 @@ func (ij *invalidationJob) Execute() error {
...
 		// the invalidation is therefore dispatched.
+	case isMissedMountKey(ij.key):
+		// router.Invalidate(...) may return false when a matching plugin was
+		// not yet loaded, even though this was under a path we'd expect to
+		// be a mount key (auth/, audit/, or logical/) prefix. Ignoring it is
+		// fine: a later change to a subsequent entry will actually load the
+		// mount, loading any data this entry would've contained for the first
+		// time.
+		//
+		// This is true in reverse for deletions.
```
phil9909, line 405: And for mount moves I guess

PR #2034, opened 2025-10-29
NS wrapper

KrzysztofKornalewski-Reply:
Main problem:
`The SealManager (SM) and NamespaceStore (NS) have their own locks to control concurrent data access. These distributed locks (that are normally locked and released in each method) however can make it hard to track lock states in cross-component calls, and operation that requires mutiple calls to the namespace-seal subsystems.
`
`Therefore it's desirable to eliminate these races/deadlock. Streamlining the lock process would be ideal to prevent such situations from happening.`

Plan how to fix potential deadlocks and races between the SealManager and the NamespaceStore. 

1. Wrapper (seal into namespace - eliminates SM lock)
2. API Layer refactor (resolve syntax errors)
3. Rotation refactor 
4. SealOrchestrator Mediator - Enforces lock order (NS -> SM)

The main goal is to move the seal information related to the namespace from the SealManager to the NamespaceStore in order to resolve potential deadlocks and races between them (first point)

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -575,7 +575,7 @@ func (c *Core) loadTransactionalCredentials(ctx context.Context, barrier logical
 	globalEntries := make(map[string][]string, len(allNamespaces))
 	localEntries := make(map[string][]string, len(allNamespaces))
 	for index, ns := range allNamespaces {
-		view := c.NamespaceView(ns)
+		view := c.NamespaceView(UnwrapNamespace(ns))
 
 		nsGlobal, nsLocal, err := c.listTransactionalCredentialsForNamespace(ctx, view)
 		if err != nil {
@@ -607,7 +607,7 @@ func (c *Core) loadTransactionalCredentials(ctx context.Context, barrier logical
 		}
 
 		for nsIndex, ns := range allNamespaces {
-			view := c.NamespaceView(ns)
+			view := c.NamespaceView(UnwrapNamespace(ns))
 			for index, uuid := range globalEntries[ns.ID] {
 				entry, err := c.fetchAndDecodeMountTableEntry(ctx, view, coreAuthConfigPath, uuid)
 				if err != nil {
@@ -624,7 +624,7 @@ func (c *Core) loadTransactionalCredentials(ctx context.Context, barrier logical
 
 	if len(localEntries) > 0 {
 		for nsIndex, ns := range allNamespaces {
-			view := c.NamespaceView(ns)
+			view := c.NamespaceView(UnwrapNamespace(ns))
 
 			for index, uuid := range localEntries[ns.ID] {
 				entry, err := c.fetchAndDecodeMountTableEntry(ctx, view, coreLocalAuthConfigPath, uuid)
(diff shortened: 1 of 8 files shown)


cipherboy:
@KrzysztofKornalewski-Reply @phyrog @wslabosz-reply @satoqz @Huy-Dinh-Reply Wanted to make sure I understand this, given how large of a PR this already is...

> The SealManager (SM) and NamespaceStore (NS) have their own locks to control concurrent data access. These distributed locks (that are normally locked and released in each method) however can make it hard to track lock states in cross-component calls, and operation that requires mutiple calls to the namespace-seal subsystems. 

In thinking about this problem w.r.t. per-namespace storage segments as well, I wondered if we're not relying on `namespace.Clone()` enough.

Rather than dynamically looking up (and dealing with lock ordering), we could have a message passing interface:

 - NS load/unload (whether create/delete, seal/unseal, lock/unlock, &c) results in a message to SealManager from NS Store with the updated namespace, cloned.
 - SM maintains these cloned copies.

Memory usage is a touch higher, but we fully avoid needing cross-subsystem locking; SM wouldn't need to acquire the NS lock.

satoqz:
I wasn't involved in all internal communication around this patch, but IIRC this was meant as an experiment to see what a largely different architecture would look like?

Either way, to keep things simple given all the existing code on this branch, IMO the best approach would be to enforce a consistent calling direction and thereby consistent lock ordering:

- `NamespaceStore` can call into `SealManager`, but `SealManager` _must not_ call into `NamespaceStore`
- A `NamespaceStore` lock, if needed by an operation, is always acquired _before_ acquiring the `SealManager` lock.

This would force operations to be driven by `NamespaceStore` every time both components are involved, which I think is good. We already have prior work going towards this approach with https://github.com/openbao/openbao/pull/1899, we should apply the same refactor to remaining `SealManager` -> `NamespaceStore` calls.

KrzysztofKornalewski-Reply:
> @KrzysztofKornalewski-Reply @phyrog @wslabosz-reply @satoqz @Huy-Dinh-Reply Wanted to make sure I understand this, given how large of a PR this already is...
> 
> > The SealManager (SM) and NamespaceStore (NS) have their own locks to control concurrent data access. These distributed locks (that are normally locked and released in each method) however can make it hard to track lock states in cross-component calls, and operation that requires mutiple calls to the namespace-seal subsystems.
> 
> In thinking about this problem w.r.t. per-namespace storage segments as well, I wondered if we're not relying on `namespace.Clone()` enough.
> 
> Rather than dynamically looking up (and dealing with lock ordering), we could have a message passing interface:
> 
> * NS load/unload (whether create/delete, seal/unseal, lock/unlock, &c) results in a message to SealManager from NS Store with the updated namespace, cloned.
> * SM maintains these cloned copies.
> 
> Memory usage is a touch higher, but we fully avoid needing cross-subsystem locking; SM wouldn't need to acquire the NS lock.

I agree that  that this PR in this aprroach leads to the massive changes so 
I understand that instead of enforcing strict lock ordering (NS -> SM) across components, we could adopt a message passing pattern where:  
NamespaceStore clones the updated namespace data after each state change  
NamespaceStore sends these clones to SealManager via message
SealManager maintains its own internal cache of these cloned namespaces  
Both components operate independently on their own data, eliminating cross-component locks entirely

Are there any existing patterns in OpenBao for this kind of component communication?
Should messages be cloned every time we perform operations such as creating, deleting, sealing or unsealing?
Would you suggest cloning the whole object from NS to SM or just particular variables depending on the action that triggered it?

KrzysztofKornalewski-Reply:
@cipherboy , We had a call to discuss the current situation in this task/PR, and we agreed that we would like to try @satoqz idea. This should simplify the implementation, and we hope it will resolve the race and deadlock problems. 
Cloning and messaging as you proposed will solve the deadlock problem, but the race problem will remain, as we understand it.

KrzysztofKornalewski-Reply:
I am closing this PR. Future work will  be continue in another PR

PR #2036, opened 2025-10-30
Update or remove deprecated code

phyrog:
This is part of #1580.

~There are several changes that I expect to prompt some discussion:~
- ~Removal of CRLv1 support for the TLS auth method~
  - ~CRLv1 is no longer supported by the non-deprecated Go std functions~
  - ~CRLv2 MUST be generated by CAs conforming to RFC3280 (2003) and its successor RFC5280 (2008)~
  - ~OpenBao itself only generates CRLv2~
  - ~Counterpoint: OpenSSL CLI by default (without adding at least one extension to the CRL) will still produce a CRLv1, see discussion [here](https://github.com/openssl/openssl/issues/23109)~
- ~Removal of DSA signatures for cert signing in the SSH secrets engine~
  - ~The `crypto/dsa` Go package is deprecated~
  - ~DSA support was removed from OpenSSH in version 1.10, being disabled by default since 2015~
  - ~Counterpoint: There may still be legacy systems around that need DSA for SSH~

~We might want to not merge these two changes right away, but deprecate them first?~

Some other notable parts of this PR:
- We'll still be using a bunch of our own deprecated functions, especially `client.RawRequest`. I replaced it with `client.Logical().ReadRaw` where possible, but we are lacking corresponding methods for Delete, Post, ..., as well as options to set headers for the requests.
- `elliptic.Marshal` can't be replaced easily at the moment. Go 1.25 introduced `ecdsa.PublicKey.Bytes()`, which can be used as a direct replacement, but we are still bound to 1.24 at the moment (especially the EL8 runner). Converting to `ecdh.PublicKey` (as recommended by the Go docs) also does not work, as it does not support the curve P224, which we support. I opted to add a comment with the code we can replace it with in 1.25 and leave it as is otherwise.
- I removed the usage of `OASDocument.CreateOperationIDs` and could not detect any change in behavior in the automated tests or my manual tests. It is documented as being there for legacy plugins, but it was not clear to me if the legacy plugins would call this, or if we need to call it for the legacy plugins.
- After this PR, there are two deprecation warnings left: one for the mentioned `elliptic.Marshal`, the other one for usage of `cloud.google.com/go/monitoring/apiv3`, which is tied to #1929.

I recommend to review commit by commit.

### DIFF
--- a/builtin/credential/jwt/path_cel_role.go
+++ b/builtin/credential/jwt/path_cel_role.go
@@ -12,7 +12,8 @@ import (
 
 	"github.com/go-jose/go-jose/v4/jwt"
 	"github.com/google/cel-go/cel"
-	"github.com/google/cel-go/checker/decls"
+	"github.com/google/cel-go/common/decls"
+	"github.com/google/cel-go/common/types"
 	"github.com/openbao/openbao/sdk/v2/framework"
 	celhelper "github.com/openbao/openbao/sdk/v2/helper/cel"
 	"github.com/openbao/openbao/sdk/v2/logical"
@@ -439,7 +440,7 @@ func (b *jwtAuthBackend) celEnv(program celhelper.CelProgram) (*cel.Env, error) 
 
 	// Add all variable declarations to the CEL environment.
 	for _, variable := range program.Variables {
-		envOptions = append(envOptions, cel.Declarations(decls.NewVar(variable.Name, decls.Dyn)))
+		envOptions = append(envOptions, cel.VariableDecls(decls.NewVariable(variable.Name, types.DynType)))
 	}
 
 	// Add "pb.Auth" return type to environment
--- a/builtin/logical/pki/test_helpers.go
+++ b/builtin/logical/pki/test_helpers.go
@@ -114,7 +114,7 @@ func getSelfSigned(t *testing.T, subject, issuer *x509.Certificate, key *rsa.Pri
 func getCrlCertificateList(t *testing.T, client *api.Client, mountPoint string) *x509.RevocationList {
 	t.Helper()
 
-	path := fmt.Sprintf("/v1/%s/crl", mountPoint)
+	path := fmt.Sprintf("%s/crl", mountPoint)
 	return getParsedCrlAtPath(t, client, path)
 }
 
@@ -155,15 +155,14 @@ func requireSerialNumberInCRL(t *testing.T, revokeList *x509.RevocationList, ser
 func getParsedCrl(t *testing.T, client *api.Client, mountPoint string) *x509.RevocationList {
 	t.Helper()
 
-	path := fmt.Sprintf("/v1/%s/crl", mountPoint)
+	path := fmt.Sprintf("%s/crl", mountPoint)
 	return getParsedCrlAtPath(t, client, path)
 }
 
 func getParsedCrlAtPath(t *testing.T, client *api.Client, path string) *x509.RevocationList {
 	t.Helper()
 
-	req := client.NewRequest("GET", path)
-	resp, err := client.RawRequest(req)
+	resp, err := client.Logical().ReadRaw(path)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/changelog/2036.txt
+++ b/changelog/2036.txt
@@ -1,0 +1,3 @@
+```release-note:change
+api: The deprecated api.MountConfigOutput.PluginName field was removed. This was already always empty.
+```
--- a/api/client.go
+++ b/api/client.go
@@ -1254,9 +1254,9 @@ func (c *Client) NewRequest(method, requestPath string) *Request {
 // a Vault server not configured with this client. This is an advanced operation
 // that generally won't need to be called externally.
 //
-// Deprecated: RawRequest exists for historical compatibility and should not be
-// used directly. Use client.Logical().ReadRaw(...) or higher level methods
-// instead.
+// RawRequest exists for historical compatibility and should not be used
+// directly. Use client.Logical().ReadRaw(...) or higher level methods instead
+// if possible.
 func (c *Client) RawRequest(r *Request) (*Response, error) {
 	return c.RawRequestWithContext(context.Background(), r)
 }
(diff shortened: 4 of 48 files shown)


builtin/logical/kubernetes/integrationtest/integration_test.go
wslabosz-reply: I think this function is also fine to remove, as afaik there's no entry as `sys/license/status`

builtin/logical/pki/test_helpers.go
```diff
@@ -162,8 +162,7 @@ func getParsedCrl(t *testing.T, client *api.Client, mountPoint string) *x509.Rev
 func getParsedCrlAtPath(t *testing.T, client *api.Client, path string) *x509.RevocationList {
 	t.Helper()
 
-	req := client.NewRequest("GET", path)
-	resp, err := client.RawRequest(req)
+	resp, err := client.Logical().ReadRaw(strings.TrimPrefix(path, "/v1/"))
```
wslabosz-reply: nit: I understand the need for trim of the path, but would rather change all callers to not provide the path with `v1` prefix

wslabosz-reply (approved):
not considering the "should we remove it before officially deprecating question", everything looks good!

builtin/logical/pki/test_helpers.go
phyrog: Done

builtin/logical/kubernetes/integrationtest/integration_test.go
phyrog: Done

wslabosz-reply (approved):

changelog/2036.txt
```diff
@@ -0,0 +1,12 @@
+```release-note:change
+The TLS auth plugin no longer accepts CRLv1
+```
+
+```release-note:change
+The deprecated api.MountConfigOutput.PluginName field was removed. This was already always empty.
+```
+
+```release-note:change
+The SSH secrets engine no longer supports DSA for signatures
```
cipherboy:
```suggestion
secrets/ssh: Remove support for signing user-supplied DSA keys.
```

builtin/credential/jwt/path_cel_role.go
```diff
@@ -439,7 +440,7 @@ func (b *jwtAuthBackend) celEnv(program celhelper.CelProgram) (*cel.Env, error)
 
 	// Add all variable declarations to the CEL environment.
 	for _, variable := range program.Variables {
-		envOptions = append(envOptions, cel.Declarations(decls.NewVar(variable.Name, decls.Dyn)))
+		envOptions = append(envOptions, cel.VariableDecls(decls.NewVariable(variable.Name, types.DynType)))
```
cipherboy, line 443: Just a friendly reminder to @openbao/openbao-core-committers that https://github.com/openbao/openbao/pull/1854 is still open. :-)

builtin/logical/pki/path_acme_test.go
```diff
@@ -502,6 +502,7 @@ func TestAcmeNonce(t *testing.T) {
 				case "header":
 					req = client.NewRequest(http.MethodHead, "/v1/"+tc.directoryUrl)
 				}
+				//nolint:staticcheck // currently there is no other way to perform this specific request
```
cipherboy: If there's no other way, should we just undeprecate it though still suggst `ReadRaw...`?

changelog/2036.txt
```diff
@@ -0,0 +1,12 @@
+```release-note:change
+The TLS auth plugin no longer accepts CRLv1
+```
+
+```release-note:change
+The deprecated api.MountConfigOutput.PluginName field was removed. This was already always empty.
```
cipherboy:
```suggestion
api: The deprecated api.MountConfigOutput.PluginName field was removed. This was already always empty.
```

changelog/2036.txt
```diff
@@ -0,0 +1,12 @@
+```release-note:change
+The TLS auth plugin no longer accepts CRLv1
```
cipherboy:
```suggestion
auth/cert: No longer accept CRLv1.
```

cipherboy: This is the change that I think will be most breaking. I'm fine removing it, but we likely have to go through a deprecation window for this one.

changelog/2036.txt
cipherboy: This one might also be breaking, but I don't really care tbh.

cipherboy (commented):
Some thoughts, let me know what you think!

builtin/logical/pki/path_acme_test.go
phyrog: I like that

changelog/2036.txt
phyrog: Yeah I figured. I will pull this and the DSA change out of this PR and will create another one to deprecate these two. Should the api calls return warnings when using either of those features? Not sure what the procedure is for plugins.

cipherboy: @phyrog We don't have a policy or precedent for warning ahead of deprecations. I'm not against making it so, typically upstream just published a deprecations table and removed it afterwards.

Most of these endpoints probably only get hit once by a human and then forevermore by automation, so not really sure warnings are visible in output. Maybe a warn-once in server logs? :shrug:

cipherboy:
> as it does not support the curve P224, which we support.

For some reason I was (without having looked at code/docs) thinking we didn't support this, as we had followed Go's precedence there. But I see too many references to P224 for that to be a coincidence. Maybe only Go FIPS didn't support P224?

I wonder if its worth removing altogether. Fewer weird algorithms as we go into a PQC future wouldn't be the worst thing.

What's the community's thoughts? Probably worth filing an issue to discuss for a bit before we do deprecate it.

satoqz:
@cipherboy while not the most secure, I don't really see a reason to drop P224 as long as it is well-supported by Go? This is the case for `crypto/ecdsa + crypto/elliptic + crypto/x509`, and isn't for `crypto/ecdh`.

Go 1.24 is only in an awkward spot because several functions in `crypto/elliptic` were deprecated and don't have a true replacement yet. Go 1.25 [has addressed this with new APIs in the `crypto/ecdsa` package](https://go.dev/doc/go1.25#cryptoecdsapkgcryptoecdsa). Note that this applies to _all_ curves in `crypto/elliptic`, not just P224. In terms of FIPS, the package doc in `crypto/ecdsa` links [this FIPS document from 2023](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf) which seems to include P224 for ECDSA purposes. Neither does the package seem to drop support for P224 in FIPS mode.

cipherboy:
@satoqz I stand corrected. Per https://github.com/golang/go/blob/master/src/crypto/internal/boring/Dockerfile, https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4735 is the certificate used for FIPS mode currently with BoringSSL, which relies on this Security Policy: https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp4735.pdf which states P-224 was indeed certified by this module.

I was thinking historically they mostly only certified P-256 and P-384 and didn't certify P-224 or P-521. 

I know P-521 changed recently, but I was surprised to see P-224 supported.

phyrog:
Rebased

cipherboy (approved):
Sorry about the delay @phyrog but thank you!

cipherboy:
This conflicted nastily with #1854 just for the record.

PR #2037, opened 2025-10-30
forward login attempt result to primary

phil9909:
Forward Login Attempts from standby to primary to implement user lockout

Currently, only failed attempts are forwarded.

TODO:
- [x] implement successful attempt forwarding
- [x] improve testing
- [x] should forwarding be async?
- [x] should we use mount accessors instead of uuids?

See #1550
See #1528 

### DIFF
--- a/vault/request_handling.go
+++ b/vault/request_handling.go
@@ -14,6 +14,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -1964,18 +1965,30 @@ func (c *Core) LoginCreateToken(ctx context.Context, ns *namespace.Namespace, re
 	return leaseGenerated, resp, nil
 }
 
-// failedUserLoginProcess updates the userFailedLoginMap with login count and  last failed
+var failedUserLoginProcessNotSupportedWarning sync.Once
+
+// failedUserLoginProcess updates the userFailedLoginMap with login count and last failed
 // login time for users with failed login attempt
 // If the user gets locked for current login attempt, it updates the storage entry too
 func (c *Core) failedUserLoginProcess(ctx context.Context, mountEntry *MountEntry, userLockoutInfo *FailedLoginUser) error {
 	if c.standby.Load() {
-		_, err := c.rpcForwardingClient.ForwardLoginAttempt(ctx, &forwarding.LoginAttempt{
+		c.requestForwardingConnectionLock.RLock()
+		rpcForwardingClient := c.rpcForwardingClient
+		c.requestForwardingConnectionLock.RUnlock()
+
+		if rpcForwardingClient == nil {
+			return nil
+		}
+
+		_, err := rpcForwardingClient.ForwardLoginAttempt(ctx, &forwarding.LoginAttempt{
 			NamespaceUuid: mountEntry.Namespace().UUID,
 			MountAccessor: mountEntry.Accessor,
 			UserAliasName: userLockoutInfo.aliasName,
-		}) // TODO: should we make this async?
-		if s, ok := status.FromError(err); ok && s.Code() == codes.Unimplemented {
-			c.logger.Warn("can not forward failed login attempt to primary, please update your primary")
+		})
+		if s, ok := status.FromError(err); ok && s.Code() == codes.Unimplemented { // remove in OpenBao 4.0.0
+			failedUserLoginProcessNotSupportedWarning.Do(func() {
+				c.logger.Warn("can not forward failed login attempt to primary, please update your primary")
+			})
 			err = nil
 		}
 		return err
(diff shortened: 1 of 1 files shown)


vault/request_handling.go
```diff
@@ -1964,7 +1967,20 @@ func (c *Core) LoginCreateToken(ctx context.Context, ns *namespace.Namespace, re
...
 // login time for users with failed login attempt
 // If the user gets locked for current login attempt, it updates the storage entry too
-func (c *Core) failedUserLoginProcess(ctx context.Context, mountEntry *MountEntry, req *logical.Request, userLockoutInfo *FailedLoginUser) error {
+func (c *Core) failedUserLoginProcess(ctx context.Context, mountEntry *MountEntry, userLockoutInfo *FailedLoginUser) error {
+	if c.standby.Load() {
+		_, err := c.rpcForwardingClient.ForwardLoginAttempt(ctx, &forwarding.LoginAttempt{
+			NamespaceUuid: mountEntry.Namespace().UUID,
+			MountAccessor: mountEntry.Accessor,
+			UserAliasName: userLockoutInfo.aliasName,
+		}) // TODO: should we make this async?
```
cipherboy:
```suggestion
		})
```

No, I think synchronous is fine.

vault/request_handling.go
```diff
@@ -1964,7 +1967,20 @@ func (c *Core) LoginCreateToken(ctx context.Context, ns *namespace.Namespace, re
...
+		if s, ok := status.FromError(err); ok && s.Code() == codes.Unimplemented {
+			c.logger.Warn("can not forward failed login attempt to primary, please update your primary")
```
cipherboy: This should probably be a warn-once given the volume this would potentially be sent.

vault/request_handling.go
```diff
@@ -1964,7 +1967,20 @@ func (c *Core) LoginCreateToken(ctx context.Context, ns *namespace.Namespace, re
 // failedUserLoginProcess updates the userFailedLoginMap with login count and  last failed
 // login time for users with failed login attempt
 // If the user gets locked for current login attempt, it updates the storage entry too
-func (c *Core) failedUserLoginProcess(ctx context.Context, mountEntry *MountEntry, req *logical.Request, userLockoutInfo *FailedLoginUser) error {
+func (c *Core) failedUserLoginProcess(ctx context.Context, mountEntry *MountEntry, userLockoutInfo *FailedLoginUser) error {
+	if c.standby.Load() {
+		_, err := c.rpcForwardingClient.ForwardLoginAttempt(ctx, &forwarding.LoginAttempt{
```
cipherboy: Two things here:

1. This needs a lock. 
2. rpcForwardingClient can be nil, so we should ignore it then.

vault/request_handling.go
```diff
@@ -2287,10 +2303,23 @@ func (c *Core) RegisterAuth(ctx context.Context, tokenTTL time.Duration, path st
...
 		}
-		// Successful login, remove any entry from userFailedLoginInfo map
-		// if it exists. This is done for service tokens (for oss) here.
-		// For ent it is taken care by registerAuth RPC calls.
-		if userLockoutInfo != nil {
+
+		if c.standby.Load() {
+			// If we are a standby, notify the primary about the successful login
+			// The primary will use this to reset the user lockout counter
+			c.logger.Trace("forwarding successful login attempt", "mount_accessor", userLockoutInfo.mountAccessor, "user_alias", userLockoutInfo.aliasName)
```
cipherboy: As above.

cipherboy (commented):
@phil9909 A few minor things for you hopefully!

vault/request_handling.go
```diff
@@ -2287,10 +2316,30 @@ func (c *Core) RegisterAuth(ctx context.Context, tokenTTL time.Duration, path st
...
+					NamespaceUuid: ns.UUID,
+					MountAccessor: userLockoutInfo.mountAccessor,
+					UserAliasName: userLockoutInfo.aliasName,
+					Successful:    true,
+				})
+				if err != nil {
+					c.logger.Warn("failed to forward successful login attempt", "error", err, "mount_accessor", userLockoutInfo.mountAccessor, "user_alias", userLockoutInfo.aliasName)
+				}
+			}
+		} else if userLockoutInfo != nil {
```
cipherboy, line 2340:
```suggestion
		}

		if userLockoutInfo != nil {
```

Shouldn't these be separate?

vault/request_handling.go
```diff
@@ -1961,10 +1965,35 @@ func (c *Core) LoginCreateToken(ctx context.Context, ns *namespace.Namespace, re
...
+
+// failedUserLoginProcess updates the userFailedLoginMap with login count and last failed
 // login time for users with failed login attempt
 // If the user gets locked for current login attempt, it updates the storage entry too
-func (c *Core) failedUserLoginProcess(ctx context.Context, mountEntry *MountEntry, req *logical.Request, userLockoutInfo *FailedLoginUser) error {
+func (c *Core) failedUserLoginProcess(ctx context.Context, mountEntry *MountEntry, userLockoutInfo *FailedLoginUser) error {
+	if c.standby.Load() {
+		c.requestForwardingConnectionLock.RLock()
+		rpcForwardingClient := c.rpcForwardingClient
+		c.requestForwardingConnectionLock.RUnlock()
```
cipherboy, line 1977: @phil9909 Is this locking sufficient? 🤔  It differs from the pattern in `Core.ForwardRequest`.

vault/request_handling.go
```diff
@@ -1961,10 +1965,35 @@ func (c *Core) LoginCreateToken(ctx context.Context, ns *namespace.Namespace, re
...
+
+		if rpcForwardingClient == nil {
+			return nil
```
cipherboy, line 1980: @phil9909 Same here as in https://github.com/openbao/openbao/pull/2037/files#r2557715682. This means a standby operating without a forwarding client will not maintain a local lockout cache and we will never lockout users.

This seems like the opposite what we want. We want GRPC, but we also want a local cache (which is much cheaper than GRPC) for the GRPC-down case.

cipherboy:
@phil9909 I'm wondering if we want to leave this out of beta, and let it be a future feature enhancement, opt-in at that point in time. Thoughts?

phil9909:
> @phil9909 I'm wondering if we want to leave this out of beta, and let it be a future feature enhancement, opt-in at that point in time. Thoughts?

leaving it out of the beta would imply leaving it out of 2.5.0?

I'm fine with this. Only downside is it will create some overhead: We should document this as a known limitation and later revisit that.

cipherboy:
> > @phil9909 I'm wondering if we want to leave this out of beta, and let it be a future feature enhancement, opt-in at that point in time. Thoughts?
> 
> leaving it out of the beta would imply leaving it out of 2.5.0?
> 
> I'm fine with this. Only downside is it will create some overhead: We should document this as a known limitation and later revisit that.

Yeah that was my thought. Let it remain open, link it to an issue, and see if anyone comes back and really wants it. 

Then we risk beta/GA less. My 2c.

cipherboy:
Closing for now; in the event of interest in this feature, feel free to reopen.

Issue #1681, opened 2025-08-13
"static" seal not recognized


Tblue:
**Describe the bug**
OpenBao v2.3.2 does not recognize the `static` seal in its configuration file, contrary to the [documentation](https://openbao.org/docs/configuration/seal/static/).

**To Reproduce**
1. Add the `static` seal to the config file as [documented](https://openbao.org/docs/configuration/seal/static/):
   ```hcl
   seal "static" {
     current_key_id = "0D5D9911-593C-4AA3-B594-09382AE85879"
     current_key = "file:///opt/secrets/seal-secret/currentKey"
   }
   ```
2. Start OpenBao.
3. Observe error log message:
   ```
   Error parsing Seal configuration: Unknown KMS type "static"
   ```

**Expected behavior**
OpenBao should recognize the `static` seal in the config file, as per the documentation.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): v2.3.2
* OpenBao CLI Version (retrieve with `bao version`): v2.3.2
* Server Operating System/Architecture: Kubernetes v1.30.6

OpenBao server configuration file(s):

```hcl
ui = true

listener "tcp" {
  tls_disable = 1
  address = "[::]:8200"
  cluster_address = "[::]:8201"
}

storage "postgresql" {
  # Config injected using env vars
}

service_registration "kubernetes" {}

seal "static" {
  current_key_id = "0D5D9911-593C-4AA3-B594-09382AE85879"
  current_key = "file:///opt/secrets/seal-secret/currentKey"
}
```

**Additional context**
I suspect that this is not yet part of the current release, but it's already documented. Maybe the documentation should note in which version of OpenBao features became available.

Tblue:
Recapping discussion from over at https://github.com/openbao/go-kms-wrapping/issues/42 regarding issues with the documentation:

- The documentation is not versioned at the moment. PRs are welcome.
- A workaround might be adding a note like "Feature available in OpenBao vX.X.X" to the docs.

satoqz:
> A workaround might be adding a note like "Feature available in OpenBao vX.X.X" to the docs.

I think this is going to be the best solution for now.

Related, coincidentally I came up with similar concerns yesterday: https://github.com/openbao/openbao/pull/1622#pullrequestreview-3114339355

wslabosz-reply:
Ideally we would have our docsite have a "version picker", so that the documentation could be also versioned and tied to specific release (at least parts of it).

cc: @cipherboy

[...]

PR #2039, opened 2025-10-30
Docs: change all absolute links to relative ones

JanMa:
Inside the docs and api-docs directories. This is a necessary step to
build multi-version docs and ensures links are always pointing to the
same version.

We also need to backport those changes to 2.4.x and 2.3.x so users will stay in
the docs for the same version when browsing older versioned docs.
Positive side effect of this change is that once we release 2.6.x all docs in
2.5.x have relative links and work as expected :-)

Part-Of: #1681


website/content/api-docs/auth/jwt.mdx
```diff
@@ -383,7 +383,7 @@ which should be shown to the user to enter at the `auth_url`.
   provider. See [Redirect URIs](/docs/auth/jwt#redirect-uris) for more information.
 - `client_nonce` `(string: <optional>)` - Optional client-provided nonce that
   must match the `client_nonce` value provided during a subsequent request to the
-  [callback](/api-docs/auth/jwt#oidc-callback) API.
+  [callback](jwt.mdx#oidc-callback) API.
```
cipherboy, line 386: Just curious, is the `.mdx` suffix required?

JanMa: Yes, this is the recommended way by Docusaurus. See: 
https://docusaurus.io/docs/versioning#link-docs-by-file-paths

satoqz (approved):
LGTM. I sanity checked that the Docusaurus build would fail if there were any dead links 😄

PR #2040, opened 2025-10-30
Partition code in `mount.go` in preparation of refactor

wslabosz-reply:
This PR is a groundwork preparation for the mounts refactor.

### Changes:
- moved code out of `mount.go` file into: `mount_table.go` (with definition of `mountTable` struct and its methods), `mount_entry.go` (with definition of `mountEntry` struct and its methods)
- replaced `(*Core).namespaceMountEntryView()` wrapper usage with direct use of `NamespaceView()`
- changed `(*Core).tableMetrics()` method signature
- removed `MountState` property from `MountEntry`
- `(*MountTable).setTaint` code changed (no longer returns an error, and cannot set the `MountState` field)

### DIFF
--- a/vault/mount_table.go
+++ b/vault/mount_table.go
@@ -1,0 +1,278 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package vault
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"sort"
+	"strings"
+
+	metrics "github.com/hashicorp/go-metrics/compat"
+	uuid "github.com/hashicorp/go-uuid"
+	"github.com/openbao/openbao/helper/metricsutil"
+	"github.com/openbao/openbao/helper/namespace"
+)
+
+// MountTable is used to represent the internal mount table
+type MountTable struct {
+	Type    string        `json:"type"`
+	Entries []*MountEntry `json:"entries"`
+}
+
+// shallowClone returns a copy of the mount table that
+// keeps the MountEntry locations, so as not to invalidate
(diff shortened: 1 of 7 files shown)


vault/mount_table.go
```diff
@@ -0,0 +1,241 @@
...
+// if modifying entries rather than modifying the table itself
+func (t *MountTable) shallowClone() *MountTable {
+	return &MountTable{
+		Type:    t.Type,
+		Entries: slices.Clone(t.Entries),
+	}
+}
+
+// setTaint is used to set the taint on given mount entry. Accepts either the mount
+// entry's path or namespace + path, i.e. <ns-path>/secret/ or <ns-path>/token/
```
phyrog: Is this comment still correct? Was it still correct before?

wslabosz-reply: it looks correct to me? what about it doesn't look correct to you?

EDIT: I guess you can argue that it's not "*either* mount entry path or namespace", but both

phyrog: It's the
- either the mount entry's path
- or namespace + path

that irritates me. I think this might be left-over from not having namespaces? From the code it seems to me like it would always require namespace + path.

wslabosz-reply: adjusted

wslabosz-reply:
@cipherboy

all down for the introduction of mount package, I'm afraid it will require more than easily reviewable amount of changes, but yeah, could be introduced with future PRs.

cipherboy:
@wslabosz-reply Oh yeah, didn't mean in this PR, thought it could be a separate PR :-)

wslabosz-reply:
merging @cipherboy Thanks!

PR #2043, opened 2025-10-31
Fix some errcheck lint findings

klaus-sap:
Resolves 3 files from https://github.com/openbao/openbao/issues/1574

### DIFF
--- a/builtin/credential/cert/path_login.go
+++ b/builtin/credential/cert/path_login.go
@@ -519,7 +519,6 @@ func (b *backend) matchesCertificateExtensions(clientCert *x509.Certificate, con
 		_, err := asn1.Unmarshal(ext.Value, &parsedValue)
 		if err != nil {
 			b.Logger().Trace("failed to unmarshal client certificate extension", "error", err)
-			continue
 		}
 		clientExtMap[ext.Id.String()] = parsedValue
 	}
@@ -561,7 +560,6 @@ func (b *backend) certificateExtensionsMetadata(clientCert *x509.Certificate, co
 			_, err := asn1.Unmarshal(ext.Value, &parsedValue)
 			if err != nil {
 				b.Logger().Trace("failed to unmarshal metadata certificate extension", "error", err)
-				continue
 			}
 			metadata[metadataKey] = parsedValue
 		}


builtin/credential/cert/path_login.go
```diff
@@ -516,7 +516,11 @@ func (b *backend) matchesCertificateExtensions(clientCert *x509.Certificate, con
 	clientExtMap := make(map[string]string, len(clientCert.Extensions))
 	for _, ext := range clientCert.Extensions {
 		var parsedValue string
-		asn1.Unmarshal(ext.Value, &parsedValue)
+		_, err := asn1.Unmarshal(ext.Value, &parsedValue)
+		if err != nil {
+			b.Logger().Error("failed to unmarshal client certificate extension", "error", err)
```
cipherboy: Right, I remember this context now. 

I don't think we can log here; if we do, we'll spam the server with mostly expected failures: extensions like AKID are required on the certificate but are not strings, so we're intending to ignore the return code. Logging would be much too verbose and open us up to allowing attackers to create completely arbitrary log messages even at high logging levels. 

I think we can still continue, thoug.

klaus-sap: @cipherboy thanks, you are right. Should I change to Debug log level to enable logging when analyzing cert parsing errors, or do not log at all here?

cipherboy: @klaus-sap How about `trace`? I know very few people who run servers at that level for long periods of time, and only when debugging issues.

klaus-sap: Changed log level to Trace

builtin/credential/cert/path_login.go
```diff
@@ -554,7 +558,11 @@ func (b *backend) certificateExtensionsMetadata(clientCert *x509.Certificate, co
 			// including its ASN.1 type tag bytes. For the sake of simplicity, assume string type
 			// and drop the tag bytes. And get the number of bytes from the tag.
 			var parsedValue string
-			asn1.Unmarshal(ext.Value, &parsedValue)
+			_, err := asn1.Unmarshal(ext.Value, &parsedValue)
+			if err != nil {
+				b.Logger().Trace("failed to unmarshal metadata certificate extension", "error", err)
+				continue
```
klaus-sap: @cipherboy I realize that this "continue" changes the logic. In case of an error, we don't create an entry in metadata anymore. Is this ok?

cipherboy: @klaus-sap Best to drop the `continue` I think. We could have a boolean authorization decision over a binary (non-string) extension like:

```
required_extensions[1.2.3.4] = ""
```

and certs having this extension would be allowed today, but rejected with this change (due to the logic on line 530 above, but below the other `continue`). Certs without this extension would be rejected in both cases.

For the metadata (this line you actually commented on) there's no way to condition on a metadata field existing, only to use it in templating, with an exception/error occurring if it didn't exist, so I'm not quite sure what I think, but best to keep the two places consistent, IMHO. 

We should probably codify the behavior with a test so users know they can rely on it, though arguably such users are better served by CEL support (with support for parsing custom extension types).

klaus-sap: removed 2x "continue"

cipherboy:
Thank you @klaus-sap, much appreciated!

PR #2045, opened 2025-10-31
Remove Enterprise license and feature_flag references 

### LINKED ISSUE #2044: Remove `sdk/v2/helper/license` and all Enterprise license references
After the fork, we've ended up with multiple references to the Enterprise version of Vault. Some of them implemented over time, few waiting for their turn, and few completely redundant. One of the redundant references is a `sdk/v2/helper/license` package and its usage of `feature_flags` which we've never supported. This would cleanup some dead code, and could possibly remove potential confusion to the end-user reading about OpenBao supporting feature flags, which we really do not do in a way it is described. All opt-in/out features introduced to OpenBao are optional config values provided directly to the server config or specific subsystem configs.

I'd propose removing the package, and all its usage/references. 

There's still a topic of integration tests of `credential/kerberos`, which reference LICENSE as one of the required inputs, but that could be adjusted on its own.

wslabosz-reply:
Resolves #2044 

- Removes `sdk/v2/helper/license` package
- Removes `FeatureFlags` struct and `persistFeatureFlags()` method from `Core`
- Removes `http/sys_feature_flags.go` file code with the ability to retrieve feature_flags
- Removes `internal/ui/feature-flags` endpoint
- Adjusts UI code for removed feature_flag endpoint


sdk/helper/consts/replication.go
```diff
@@ -19,7 +19,9 @@ const (
 	// manager.  It should contain a character that is not allowed in secondary
 	// ids to ensure it doesn't collide.
 	CurrentReplicatedSecondaryIdentifier = ".current"
-	CoreFeatureFlagPath                  = "core/cluster/feature-flags"
+
+	// For replication compability we are leaving it here, but this location won't be used.
+	CoreFeatureFlagPath = "core/cluster/feature-flags"
```
cipherboy: I think this is safe to drop actually. Not sure we'll use it and doesn't appear to be used elsewhere.

audit/format.go
```diff
@@ -400,7 +400,7 @@ type AuditRequestEntry struct {
 	Auth          *AuditAuth    `json:"auth,omitempty"`
 	Request       *AuditRequest `json:"request,omitempty"`
 	Error         string        `json:"error,omitempty"`
-	ForwardedFrom string        `json:"forwarded_from,omitempty"` // Populated in Enterprise when a request is forwarded
+	ForwardedFrom string        `json:"forwarded_from,omitempty"`
```
cipherboy, line 403: @phil9909 @wslabosz-reply We currently don't use this as we use the existing `http.Request`-level forwarding and thus incur a double-audit. This currently doesn't populate that field. 🤔

cipherboy (commented):
One really minor note while you're here @wslabosz-reply.

audit/format.go
wslabosz-reply: is it safe to remove?

cipherboy: @wslabosz-reply @phil9909 I don't know, I think maybe we want to set this actually in the context of #1986...? So let's leave it for now.

wslabosz-reply: looking at this field, and it's counterpart of `AuditResponseEntry` `Forwarded`, they seem to be unused, but will leave as you say.

cipherboy (approved):
Haven't tested the UI, though.

.hooks/pre-push
satoqz, line 1: This will conflict https://github.com/openbao/openbao/pull/1601, but I'm sure Philipp won't mind.

satoqz (approved):
Based on @wslabosz-reply's internal comment on his displeasure with Ember, I trust he's tested the UI 😄

wslabosz-reply:
@cipherboy 
I'm struggling making UI work, let me inform you when I finish testing.

.hooks/pre-push
phil9909: Fine with me. Will be an easy rebase.

audit/format.go
phil9909: This could be helpful. @cipherboy Will you add it to #1986?

wslabosz-reply:
@cipherboy
UI CI passes, and by manually clicking through UI I haven't seen anything broken.

PR #2047, opened 2025-10-31
Deprecate JSONX audit format

### LINKED ISSUE #1874: Deprecation candidate: JSONX audit logging
Audit logging supports a `format` parameter, which takes `jsonx` as a value.

See: https://openbao.org/docs/audit/#common-configuration-options 

We should see if we can deprecate the jsonx format option; it is unlikely anyone is consuming audit logs in XML format and it adds another dependency and overhead to the log format.

wslabosz-reply:
Resolves #1874

This PR removes `"jsonx"` audit format parsing code, and deprecates the `"jsonx"` as a valid audit format option. To keep the compatibility with existing audit configs, the `format` option is still parsed, but will error on `jsonx` value (with proper messaging), and still default to `json` when provided empty string.


builtin/audit/file/backend.go
```diff
@@ -50,7 +50,9 @@ func Factory(ctx context.Context, conf *audit.BackendConfig) (audit.Backend, err
 		format = "json"
 	}
 	switch format {
-	case "json", "jsonx":
+	case "json":
+	case "jsonx":
+		return nil, errors.New(`jsonx format was discontinued, remove "format" config option, or provide "json" value`)
```
satoqz:
```suggestion
		return nil, errors.New(`jsonx formatting has been removed, consider switching to json by omitting "format" or setting it to "json"`)
```

Not sure, just a different shot at wording.

website/content/docs/audit/index.mdx
satoqz, line 1: Should also add a deprecation notice to `website/content/docs/deprecation`

builtin/audit/file/backend.go
cipherboy: Either are fine by me 🤷

changelog/2047.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:deprecation
+core/audit: deprecated `jsonx` as a output format option for audit mounts
```
cipherboy:
```suggestion
```release-note:change
core/audit: removed `jsonx` as a output format option for audit mounts
```

phyrog: Shouldn't we have a deprecation period before actually removing it?

satoqz: See discussion here: https://github.com/openbao/openbao/issues/1874#issuecomment-3473465661

wslabosz-reply:
merging based on approvals from @satoqz & @cipherboy

PR #2056, opened 2025-11-03
Remove set-up-gotestsum action

satoqz:
The `set-up-gotestsum` action downloads `gotestsum` from GitHub releases. Listing GitHub releases is an API call that fails rather frequently (compared to, say, the Go module mirror), so replace the action by reverting to `go install`, which may take a few seconds longer but should be more reliable.

Also see:
- https://github.com/openbao/openbao/actions/runs/19029557302/job/54340212536?pr=2055
- https://github.com/cli/cli/issues/7533


satoqz:
> Ultimately we'll need to pin tool hashes, but `@latest` is fine for now

My thoughts exactly.

PR #2060, opened 2025-11-03
address read-replication PR comments

phil9909:
Addresses the comments by @wslabosz-reply in #1986


cipherboy (approved):
Thanks @phil9909!

PR #2061, opened 2025-11-04
PKI health check returning proper mount path

### LINKED ISSUE #2057: pki_health_check appears to not be templating a value correctly
This requires an incorrect config to be supplied so shouldn't be visible when Openbao is correctly configured and working correctly.

**Describe the bug**
When using pki_health_check and acme endpoint is incorrectly configured the enable_acme_issuance message returns a path that looks like it's meant to template the mount into the path but instead prints {{mount}}.

```
$bao pki health-check pki
...
enable_acme_issuance
--------------------
status      endpoint                  message
------      --------                  -------
warning     /pki/config/cluster       ACME enabled in config but not functional: using configured 'path' field ('https://bao.example.com/v1/pki') in /{{mount}}/config/cluster failed to reach the ACME directory: https://bao.example.com/v1/pki/acme/directory: Get "https://bao.example.com/v1/pki/acme/directory": dial tcp 127.0.0.1:8200: connect:connection refused
...
```

**To Reproduce**
1. `bao secrets enable -path test/ pki`.
2. `bao write test/root/generate/internal common_name=example.com`
3. `bao write test/config/cluster aia_path=http://example.com/v1/test path=https://example.com/v1/test`
4. `bao write test/config/acme enabled=true`
5. `bao pki health-check test`, you should see a {{mount}} in the path
 
**Expected behavior**I'd expect the {{mount}} to be replaced with the name of the mount that was issued in the bao pki health-check command

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.4.3
* OpenBao CLI Version (retrieve with `bao version`): v2.4.3
* Server Operating System/Architecture: x86_64 Rocky Linux 10

wslabosz-reply:
Resolves #2057

This PR introduces a change to the PKI health_check operations, so that it will now correctly return mount path instead of it's template `{{mount}}` value.

Using the source issue reproduction instruction:
1. bao secrets enable -path test pki
2. bao write test/root/generate/internal common_name=example.com
3. bao write test/config/cluster aia_path=http://example.com/v1/test path=https://example.com/v1/test
4. bao write test/config/acme enabled=true
5. bao pki health-check test // no longer {{mount}}, but actual path of a mount - `test`

### DIFF
--- a/changelog/2061.txt
+++ b/changelog/2061.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-secrets/pki: health check now returns proper mount path instead of templated value.
+command: pki health check error now contains actual mount path instead of a template placeholder.
 ```
--- a/command/healthcheck/pki_allow_acme_headers.go
+++ b/command/healthcheck/pki_allow_acme_headers.go
@@ -140,9 +140,9 @@ func (h *AllowAcmeHeaders) Evaluate(e *Executor) ([]*Result, error) {
 func craftInsufficientPermissionResult(e *Executor, path, errorMsg string) []*Result {
 	var retMsg string
 	if e.Client.Token() == "" {
-		retMsg = fmt.Sprintf("No token available so unable to call the tune endpoint for this mount. %s", errorMsg)
+		retMsg = fmt.Sprintf("No token available, unable to read the tune endpoint for this mount. %s", errorMsg)
 	} else {
-		retMsg = fmt.Sprintf("This token lacks permission to call the tune endpoint for this mount. %s", errorMsg)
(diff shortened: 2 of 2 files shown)


changelog/2061.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+secrets/pki: health check now returns proper mount path instead of templated value.
```
satoqz:
```suggestion
command/pki: Health check error now contains actual mount path instead of a template placeholder.
```

satoqz: Ah, maybe the scope should be commands, this isn't an issue with the actual secrets engine.

wslabosz-reply:
merging based on approvals from @satoqz & @cipherboy

PR #2062, opened 2025-11-04
Fix certificate expiration time calculation in operator diagnose

### LINKED ISSUE #2059: bao operator diagnose listener TLS issues
**Describe the bug**
When an openbao listener is configured to use tls the `bao operator diagnose` command fails to build the trust chain correctly using system roots. When a CA file is not set it reports an incorrect end date which changes when re-run.

Using the listener config shown at the bottom with the certificate being valid for the localhost domain:

If `/opt/openbao/tls/tls.crt` contains a leaf cert and the root certificate is installed in the OS trust store and there's no intermediate authority, when the diagnose command is run the result is:

```bash
[ failure ] Vault Diagnose
  [ failure ] Start Listeners
    [ failure ] Check Listener TLS: bao.example.com:8200: Failed to verify intermediate certificate: x509: certificate signed by unknown authority.
      [ warning ] bao.example.com:8200: Leaf certificate ... is expired or near expiry. Time to expire is: 70h4m10s.
```

The same result is observed (except the time to expire is different) when `/opt/openbao/tls/tls.crt` contains a leaf cert followed by a branch/intermediate certificate and the system store contains the root CA.

When `/opt/openbao/tls/tls.crt` contains the entire trust chain including the root CA at the end the result of diagnose is:

```bash
[ failure ] Vault Diagnose
  [ warning ] Start Listeners
    [ warning ] Check Listeners TLS:
      [ warning ] bao.example.com:8200: Leaf certificate ... is expired or near expiry. Time to expire is: 70h15m40.46s.
```

Running `openssl x509 -noout -dates -in /opt/openbao/tls/tls.crt` it reports a not after value of Nov 15 2025. So warning about the near expiry and the time till expiry is incorrect.

If you wait a few minutes before re-running the `bao operator diagnose` command the Time to expire will have changed. It's not consistent with the time passed and for me it reported 70h35m25.40s, so somehow it's increased its lifetime.

I'm not sure what the number is after the `Leaf certificate` is in the warning, it doesn't match the Subject Key Identifier, serial number or any other value I can see in any of the certificates I've used in the test.

**Expected behavior**
I'd expect in the diagnose command to correctly use the system trust store the to build the trust chain as openbao server does. Instead of requiring the root to be appended to the end of `tls_cert_file`.

I'd also expect the Leaf certificate check to correctly report the time till expiry.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.4.3
* OpenBao CLI Version (retrieve with `bao version`): v2.4.3
* Server Operating System/Architecture: Rocky 10 x86_64

```hcl
listener "tcp" {
  address            = "localhost:8200"
  tls_cert_file      = "/opt/openbao/tls/tls.crt"
  tls_key_file       = "/opt/openbao/tls/tls.key"
}
```

**Additional context**
This issue has also existed in Vault since v1.12 but I'm not sure it's ever been reported there.

satoqz:
Resolves #2059

Two issues resolved here:

- Serial number formatting is switched from decimal to hex
- The calculation of the time to expiration was `(now + 30 days - not after)`, not `(not after - now)`

### DIFF
--- a/vault/diagnose/tls_verification.go
+++ b/vault/diagnose/tls_verification.go
@@ -258,32 +258,37 @@ func TLSFileWarningChecks(leafCerts, interCerts, rootCerts []*x509.Certificate) 
 
 	for _, c := range leafCerts {
 		if willExpire, timeToExpiry := NearExpiration(c); willExpire {
-			warnings = append(warnings, fmt.Sprintf("Leaf certificate %d is expired or near expiry. Time to expire is: %s.", c.SerialNumber, timeToExpiry))
+			warnings = append(warnings, fmt.Sprintf("Leaf certificate (serial %x) is expired or near expiry. Time to expire is: %s.", c.SerialNumber, timeToExpiry))
 		}
 	}
 	for _, c := range interCerts {
 		if willExpire, timeToExpiry := NearExpiration(c); willExpire {
-			warnings = append(warnings, fmt.Sprintf("Intermediate certificate %d is expired or near expiry. Time to expire is: %s.", c.SerialNumber, timeToExpiry))
+			warnings = append(warnings, fmt.Sprintf("Intermediate certificate (serial %x) is expired or near expiry. Time to expire is: %s.", c.SerialNumber, timeToExpiry))
 		}
 	}
 	for _, c := range rootCerts {
 		if willExpire, timeToExpiry := NearExpiration(c); willExpire {
-			warnings = append(warnings, fmt.Sprintf("Root certificate %d is expired or near expiry. Time to expire is: %s.", c.SerialNumber, timeToExpiry))
+			warnings = append(warnings, fmt.Sprintf("Root certificate (serial %x) is expired or near expiry. Time to expire is: %s.", c.SerialNumber, timeToExpiry))
 		}
 	}
 
 	return warnings, nil
 }
 
-// NearExpiration returns a true if a certficate will expire in a month and false otherwise
+// NearExpiration returns (true, time to expiration) if less than 15% of c's
+// validity period remains.
+//
+// For   7-day certs, that's     ~1 day.
+// For  30-day certs, that's   ~4.5 days.
+// For  90-day certs, that's  ~13.5 days.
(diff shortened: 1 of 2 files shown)


wslabosz-reply (approved):

vault/diagnose/tls_verification.go
```diff
@@ -258,32 +258,30 @@ func TLSFileWarningChecks(leafCerts, interCerts, rootCerts []*x509.Certificate)
...
-// NearExpiration returns a true if a certficate will expire in a month and false otherwise
+// NearExpiration returns (true, time to expiration) if c expires in less than 30 days.
 func NearExpiration(c *x509.Certificate) (bool, time.Duration) {
-	oneMonthFromNow := time.Now().Add(30 * 24 * time.Hour)
-	var timeToExpiry time.Duration
-	if oneMonthFromNow.After(c.NotAfter) {
-		timeToExpiry := oneMonthFromNow.Sub(c.NotAfter)
-		return true, timeToExpiry
+	now := time.Now()
+	if now.Add(30 * 24 * time.Hour).After(c.NotAfter) {
```
cipherboy: See comment in #2059  -- what if we make this something like:

`c.NotAfter.Sub(c.NotBefore) * 100 / 15`

or similar.

For 7-day certs, that's ~1 day. 
For 30-day certs, that's ~4.5 days
For 90-day certs, that's ~13.5 days
For 365-day certs, that's ~54.75 days

Per current certbot docs, they use 1/3 (33%) remaining: https://eff-certbot.readthedocs.io/en/stable/using.html#renewing-certificates

So by placing it at about half of that, we align with guidance but also alert operators that automation has likely failed. 

Thoughts?

EschewObfuscaton: Should it be `c.NotAfter.Sub(c.NotBefore) * 15 / 100` if you want 15% of the lifetime?

But using the percentage makes sense to me.

satoqz: Added! 15% seems good.

wslabosz-reply (approved):

PR #2065, opened 2025-11-05
Fix operator diagnose certificate verification

satoqz:
## Description

Two problems addressed here:

1. When verifying the certificate chain, `operator diagnose` treats a standalone leaf certificate as a root certificate and thus verifies it successfully. However, the same strategy was not applied to intermediate certificates, hence their verification would fail if no root is supplied.
   - @EschewObfuscaton note that none of this ever takes the the system roots into account, by design. They are fully ignored in either case. Whether or not `operator diagnose` should ensure that listener certificates are trusted by the host system itself is another discussion, but IMO no, at least by default.

2. PEM decoding in `operator diagnose` would treat trailing non-PEM data as an error. This is triggered by as little as an additional trailing newline, and doesn't match the behavior of `tls.LoadX509Certificate` (hence, this has not been an error when running the actual server). The standard library simply breaks out of the loop whenever the result of `pem.Decode` is `nil`, so I've adjusted the implementation here to do the same.

## Rationale

Also see the standard library's PEM decoding loop: https://github.com/golang/go/blob/9f3a108ee0d2f08b4994c4201e54e5a53acbc216/src/crypto/tls/tls.go#L272-L277

Resolves: #2063

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/diagnose/tls_verification.go
+++ b/vault/diagnose/tls_verification.go
@@ -215,14 +215,15 @@ func TLSErrorChecks(leafCerts, interCerts, rootCerts []*x509.Certificate) error 
 	}
 
 	if len(rootCerts) == 0 {
-		// This is a self signed server certificate, or the root is just not
-		// provided. In any case, we need to bypass the root verification step
-		// by adding the intermediate (or leaf, if there is no intermediate)
-		// itself to the root pool.
+		// No root is provided, so:
 		switch {
 		case len(interCerts) > 0:
-			rootPool.AddCert(interCerts[0])
-		case len(leafCerts) > 0:
+			// ... we'll treat the highest intermediate cert in the chain as a
+			// root certificate.
+			rootPool.AddCert(interCerts[len(interCerts)-1])
+		default:
+			// ... we'll treat the leaf itself as a root certificate if there
+			// are no intermediates.
 			rootPool.AddCert(leafCerts[0])
 		}
 	}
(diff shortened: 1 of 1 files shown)


EschewObfuscaton:
Thanks for the clarification, I'd naively assumed it was using the system roots to verify but, I guess, obviously it doesn't need to do that to offer the certs to other clients.

I don't think `operator diagnose` should ensure that listener certificates are trusted by the host system as `bao server` won't either.

Thanks for the fixes!

cipherboy (approved):
I _think_ it should be safe to add just a single intermediate/root.

satoqz:
@cipherboy eh I should probably be adding all intermediate certs to the pool no?

EDIT: Ah I see what you mean. I hadn't even thought of that tbh, but yeah, I guess it should just work like that. I'll give it a test with multiple intermediates before I merge.

cipherboy:
@satoqz Well, theoretically according to the TLS RFC, you're really only supposed to send intermediates related to validating the leaf. So if you select any random intermediate, it should validate the leaf.

I'm not sure what behavior we have for selecting leaf certificate; you could perhaps have an out-of-order file and have (int a, int b, leaf b, leaf a), where (int a -> leaf a) &c, and thus fail to validate.

This is enough of an edge case I don't think it is really worth supporting in diagnose, so I'd probably skip it.

satoqz:
@cipherboy I wasn't worried so much about the leaf but rather validating the chain of intermediates when only one of them is selected as root. I've updated this to select the _last_ intermediate which should be the highest one in the chain. Before, it would add the one closest to the leaf as the root which would then fail to verify all higher intermediates.

From https://datatracker.ietf.org/doc/html/rfc8446:

>  Note: Prior to TLS 1.3, "certificate_list" ordering required each
   certificate to certify the one immediately preceding it; however,
   some implementations allowed some flexibility.  Servers sometimes
   send both a current and deprecated intermediate for transitional
   purposes, and others are simply configured incorrectly, but these
   cases can nonetheless be validated properly.  For maximum
   compatibility, all implementations SHOULD be prepared to handle
   potentially extraneous certificates and arbitrary orderings from any
   TLS version, with the exception of the end-entity certificate which
   MUST be first.

So this is still not perfect but probably good enough and will error if you don't follow the best practice of just ordering the chain correctly (At the same time we don't even have a check that ensures the leaf comes first if there is any, eugh).

Issue #2035, opened 2025-10-29
Trim OIDC discovery URL if present in UI config


RingOfStorms:
**Is your feature request related to a problem? Please describe.**
When pasting in a discovery URL with the `/.well-known/*` suffix the config page errors when saving. The config expects the URL without the suffix. Many other apps require it for this setting and the error is not clear as to why it is failing.

<img width="470" height="186" alt="Image" src="https://github.com/user-attachments/assets/10275b55-4abb-44a3-a15b-794ebbc92246" />

[See zulip convo](https://linuxfoundation.zulipchat.com/#narrow/channel/530381-openssf-openbao-support/topic/Zitadel.20OIDC.20auth/with/547796223)[#openssf-openbao-support > Zitadel OIDC auth](https://linuxfoundation.zulipchat.com/#narrow/channel/530381-openssf-openbao-support/topic/Zitadel.20OIDC.20auth/with/547796223)

**Describe the solution you'd like**
If openbao see's the suffix it could handle it gracefully and trim when saving/using it.

Issue #1204, opened 2025-04-09
Error checking OIDC discovery url with Keycloack


gabriel-suela:
I'm trying to make oidc login via keycloak, but I get an error when trying to add on the UI and via CLI.

```
openbao write auth/oidc/config \
    oidc_client_id="openbao" \
    oidc_client_secret="${KEYCLOACK_SECRET}" \
    oidc_discovery_url="https://${URL}/realms/hq/.well-known/openid-configuration"

Error writing data to auth/oidc/config: Error making API request.

URL: PUT https://openbao.incentive.land/v1/auth/oidc/config
Code: 400. Errors:

* error checking oidc discovery URL
```
I have no clue why this says that oidc discovery URL is wrong because im using the same URL in other applications and works flawlessly.

cipherboy:
@gabriel-suela You do not include the `/.well-known/openid-configuration` in the `oidc_discovery_url`.

```
openbao write auth/oidc/config \
    oidc_client_id="openbao" \
    oidc_client_secret="${KEYCLOACK_SECRET}" \
    oidc_discovery_url="https://${URL}/realms/hq"
```

gabriel-suela:
Ty @cipherboy

PR #2066, opened 2025-11-05
Return an error on well-known component presence in OIDCDiscoveryURL value

wslabosz-reply:
Resolves: #2035

This PR introduces a change to the `jwtAuthBackend`, which now returns an error on encountering a `.well-known/` component passed with `OIDCDiscoveryURL`.

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/credential/jwt/path_config.go
+++ b/builtin/credential/jwt/path_config.go
@@ -48,7 +48,7 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 		Fields: map[string]*framework.FieldSchema{
 			"oidc_discovery_url": {
 				Type:        framework.TypeString,
-				Description: `Either the base URL of the OIDC Discovery endpoint or the full URL with ".well-known/openid-configuration" path suffix. Cannot be used with "jwks_url" or "jwt_validation_pubkeys".`,
+				Description: `The base URL of the OIDC Discovery URL without ".well-known/openid-configuration" component. Cannot be used with "jwks_url" or "jwt_validation_pubkeys".`,
 			},
 			"oidc_discovery_ca_pem": {
 				Type:        framework.TypeString,
@@ -384,16 +384,8 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 		return logical.ErrorResponse("both 'oidc_client_id' and 'oidc_client_secret' must be set for OIDC"), nil
 
 	case config.OIDCDiscoveryURL != "":
-		// due to `go-oidc` issuer matching logic, we cannot handle base urls
-		// not ending with a trailing slash, but that's been always the case
-
-		// trim '.well-known/openid-configuration', due to jwt.NewOIDCDiscoveryKeySet()
-		// not accounting for its presence.
-		// config is saved to the storage without '.well-known' component
-		trimmedDiscoveryURL, _, _ := strings.Cut(config.OIDCDiscoveryURL, ".well-known/openid-configuration")
-		if trimmedDiscoveryURL != config.OIDCDiscoveryURL {
-			config.OIDCDiscoveryURL = trimmedDiscoveryURL
-			resp.AddWarning(fmt.Sprintf("modifying provided `OIDCDiscoveryURL` to %q", trimmedDiscoveryURL))
+		if strings.Contains(config.OIDCDiscoveryURL, ".well-known/openid-configuration") {
+			return logical.ErrorResponse("'oidc_discovery_url' contains '.well-known' component"), nil
 		}
 
 		var err error
--- a/builtin/credential/jwt/path_config_test.go
+++ b/builtin/credential/jwt/path_config_test.go
@@ -386,45 +386,20 @@ func TestConfig_OIDC_Write(t *testing.T) {
 		t.Fatal(diff)
 	}
 
-	discoveryURLTests := []struct {
-		name         string
-		discoveryURL string
-	}{
-		{
-			"base url with trailing slash",
-			"https://team-vault.auth0.com/",
-		},
-		// verify that the .well-known component gets stripped
-		{
-			"full url with trailing slash",
-			"https://team-vault.auth0.com/.well-known/openid-configuration",
-		},
-		{
-			"full url without trailing slash",
-			"https://team-vault.auth0.com/.well-known/openid-configuration/",
+	// verify that specifying the '.well-known' component gets rejected with appriopriate err message
+	req = &logical.Request{
+		Operation: logical.UpdateOperation,
+		Path:      configPath,
+		Storage:   storage,
+		Data: map[string]interface{}{
+			"oidc_discovery_url": "https://team-vault.auth0.com/.well-known/openid-configuration",
+			"oidc_client_id":     "abc",
+			"oidc_client_secret": "def",
 		},
 	}
-
-	for _, test := range discoveryURLTests {
-		t.Run(test.name, func(t *testing.T) {
-			req := &logical.Request{
-				Operation: logical.UpdateOperation,
-				Path:      configPath,
-				Storage:   storage,
-				Data: map[string]interface{}{
-					"oidc_discovery_url": test.discoveryURL,
-					"oidc_client_id":     "abc",
-					"oidc_client_secret": "def",
-				},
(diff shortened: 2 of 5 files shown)


website/content/api-docs/auth/jwt.mdx
```diff
@@ -37,7 +37,7 @@ set.
 
 ### Parameters
 
-- `oidc_discovery_url` `(string: <optional>)` - The OIDC Discovery URL, without any .well-known component (base path). Cannot be used with "jwks_url" or "jwt_validation_pubkeys".
+- `oidc_discovery_url` `(string: <optional>)` - The OIDC Discovery URL, any .well-known component will be trimmed by default. Cannot be used with "jwks_url" or "jwt_validation_pubkeys".
```
satoqz: I think it should read more like "with or without .well-known component", as in "we'll add it if missing".

builtin/credential/jwt/path_config.go
```diff
@@ -387,6 +384,8 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 		return logical.ErrorResponse("both 'oidc_client_id' and 'oidc_client_secret' must be set for OIDC"), nil
 
 	case config.OIDCDiscoveryURL != "":
+		// trim '.well-known'
+		config.OIDCDiscoveryURL, _, _ = strings.Cut(config.OIDCDiscoveryURL, ".well-known/")
```
satoqz: I think `TrimSuffix` is more appropriate?

builtin/credential/jwt/path_config.go
```diff
@@ -387,6 +384,8 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 		return logical.ErrorResponse("both 'oidc_client_id' and 'oidc_client_secret' must be set for OIDC"), nil
 
 	case config.OIDCDiscoveryURL != "":
+		// trim '.well-known'
```
satoqz: I'd probably extend the comment to mention how plays together with the behavior of `github.com/hashicorp/cap/jwt`.

website/content/api-docs/auth/jwt.mdx
satoqz, line 47: Isn't this one also a `.well-known` URL?

https://github.com/search?q=repo%3Ahashicorp%2Fcap%20.well-known&type=code

wslabosz-reply: haven't looked at it (as the referenced issue only mentiones `OIDCDiscoveryURL`, but looking at it, in contrast to `jwt.NewOIDCDiscoveryKeySet()`, it does not craft a url by manually appending the `.well-known/jwks.json` to the url...

satoqz: Ah probably fine to leave as-is then.

builtin/credential/jwt/path_config.go
```diff
@@ -387,6 +384,10 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 		return logical.ErrorResponse("both 'oidc_client_id' and 'oidc_client_secret' must be set for OIDC"), nil
 
 	case config.OIDCDiscoveryURL != "":
+		// trim '.well-known/openid-configuration', due to jwt.NewOIDCDiscoveryKeySet()
+		// assuming the `OIDCDiscoveryURL` doesn't account for its presence.
+		// config is saved to the storage without '.well-known' component
+		config.OIDCDiscoveryURL = strings.TrimSuffix(strings.TrimSuffix(config.OIDCDiscoveryURL, "/"), ".well-known/openid-configuration")
```
DrDaveD: If it does end up trimming off `.well-known/openid-configuration` here, it will end up with a trailing `/` which otherwise it tries to trim (for whatever reason).  I don't know how important that is, but in any case it's inconsistent.  Probably it's enough to add a `/` before `.well-known`.

builtin/credential/jwt/path_config_test.go
```diff
@@ -338,7 +338,8 @@ func TestConfig_OIDC_Write(t *testing.T) {
 	// First we provide an invalid CA cert to verify that it is in fact paying
 	// attention to the value we specify
 	data := map[string]interface{}{
-		"oidc_discovery_url":    "https://team-vault.auth0.com/",
+		// since https://github.com/openbao/openbao/pull/2066 .well-known component gets stripped
+		"oidc_discovery_url":    "https://team-vault.auth0.com/.well-known/openid-configuration",
```
DrDaveD: Is this necessary?  Why add it in just to be stripped?  Oh, this is a test for the stripping?  If that's the case I don't think it's helpful to include the PR number but it would be helpful to say something like
```
    // verify that the .well-known component gets stripped
```
Maybe this test was also supposed to verify that having a trailing `/` was handled, in which case this test should also probably add a trailing slash.

builtin/credential/jwt/path_config_test.go
```diff
@@ -370,7 +371,7 @@ func TestConfig_OIDC_Write(t *testing.T) {
 		JWTSupportedAlgs:           []string{},
 		OIDCResponseTypes:          []string{},
 		OverrideAllowedServerNames: []string{},
-		OIDCDiscoveryURL:           "https://team-vault.auth0.com/",
+		OIDCDiscoveryURL:           "https://team-vault.auth0.com",
```
DrDaveD: Why this change?

changelog/2066.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+auth/jwt: Support OIDCDiscoveryURL with specified '.well-known/' component, which gets stripped while parsed.
```
DrDaveD:
```suggestion
auth/jwt: Allow OIDCDiscoveryURL to optionally include '.well-known/openid-configuration' path suffix.
```

cipherboy:
> I'm not really convinced that this feature is all that necessary, that maybe the person that made the original issue should just change their code, but I guess it doesn't hurt.

@DrDaveD We've had this happen n>3 times in the OpenBao project itself; while yes, they can change their submission, I think this is common enough (and easy enough to deal with) that we probably want to just handle it rather than keep returning weird errors.

Edit: some references include https://github.com/openbao/openbao/issues/1204 and I recall at least one on Matrix.  https://discuss.hashicorp.com/t/configuring-a-dex-oidc-auth-provider/23124 is one such thread on the upstream Vault discussion forum as well.

builtin/credential/jwt/path_config.go
wslabosz-reply: turns out we need a trailing slash, because that's how the response from `go-oidc.NewProvider()` is shaped, and we have to match against that

builtin/credential/jwt/path_config_test.go
wslabosz-reply: reverted

wslabosz-reply:
To surface private discussions with @satoqz and @cipherboy:

the issue with the introduction of graceful handling of `.well-known` component in the `oidcDiscoveryURL` is that by trimming the `.well-known/openid-configuration` part we assume correct intention of the operator (that being whatever he passes is able to respond to the discovery queries).

With the caveat being that the `coreos/go-oidc` lib verifies that the `issuer` of the created provider config (through `hashicorp/cap/oidc` lib) matches the `issuerURL` which `go-oidc` [will do an exact equal string matching](https://github.com/coreos/go-oidc/issues/442) (e.g. if you provide `https://example.com` it *will* be matched against `https://example.com/` - this is **dependent** on the provider response).

So back to the proposed solution:
`trimmedDiscoveryURL, _, _ := strings.Cut(config.OIDCDiscoveryURL, ".well-known/openid-configuration")`
will modify provided URL in a way that ignores whatever trailing characters there are after the well-known component (`https://example.com/.well-known/openid-configuration/whatever////` -> `https://example.com/`)

Instead of modifying operator's input @satoqz proposes returning a custom error or logging an informative warning/error when `.well-known` component is present.

@cipherboy @satoqz @DrDaveD thoughts?

wslabosz-reply:
After debating the matter with @satoqz & @cipherboy, opted to instead return an error on encountering `.well-known` component with the discovery url, for clear guidelines to the operator, how to proceed.

Issue #2050, opened 2025-11-02
Memory leak 2.4.3


venkatamutyala:
**Describe the bug**
High memory usage on openbao pods using raft storage. ~4GB/pod

**To Reproduce**
I have Openbao deployed using raft storage and i'm using v2.4.3 with the openbao helm chart v0.19.1:
After migrating from hashicorp vault 1.14.X to openbao 2.4.3 i found that within 48 hours the new openbao environment is ussing about 4000% more memory usage across all my pods. I have 3 replicas running with the helm chart. Prior to this my memory usage was only ~180MB/pod currently it's ~4GB/pod

**Expected behavior**
Lower memory usage

**Environment:**
* OpenBao Server Version (retrieve with `bao status`):
```
Key                     Value
---                     -----
Seal Type               shamir
Initialized             true
Sealed                  false
Total Shares            1
Threshold               1
Version                 2.4.3
Build Date              2025-10-22T20:21:42Z
Storage Type            raft
Cluster Name            vault-cluster-c5377a83
Cluster ID              384aad12-e1e6-ff62-ea4f-82ca21eb9c84
HA Enabled              true
HA Cluster              https://vault-1.vault-internal:8201
HA Mode                 standby
Active Node Address     https://10.42.7.30:8200
Raft Committed Index    137432872
Raft Applied Index      137432871
/ $ 
```
* OpenBao CLI Version (retrieve with `bao version`):
```
/ $ bao version
OpenBao v2.4.3 (ef854342df72dba6ecfbdd3f130e251941ba7dca), built 2025-10-22T20:21:42Z
/ $ 
``` 
 
* Server Operating System/Architecture:
Running the official docker image in k8s

OpenBao server configuration file(s):

```hcl
                ui = true

                listener "tcp" {
                  address            = "0.0.0.0:8200"
                  cluster_address    = "0.0.0.0:8201"
                  tls_cert_file      = "/openbao/userconfig/vault-tls/tls.crt"
                  tls_key_file       = "/openbao/userconfig/vault-tls/tls.key"
                  tls_client_ca_file = "/openbao/userconfig/vault-tls/ca.crt"
                  tls_min_version    = "tls12"
                  telemetry {
                    unauthenticated_metrics_access = true
                  }
                }

                storage "raft" {
                  path = "/openbao/data"

                  retry_join {
                    # Keep label selector/DNS stable by using name/fullname override = "vault"
[...]

PR #2067, opened 2025-11-05
raft: send lowestActiveIndex also for put and delete operations

phil9909:
## Description

Fixes a memory leak in the raft storage layer, which happens when only `Put` and `Delete` log are applies (but no transactions). The fsm has some cache to speed up transaction validation, which it prunes based on the `lowestActiveIndex`, which was previously only sent along with transaction commits. If there are never transactions committed, every write and delete will cause that cache to grow without limit.

See this graph, based on the reproduction posted here: https://github.com/venkatamutyala/debug-openbao-issues/tree/main

pod_usage

With the fix, we still see some memory growth, which I will need to investigate.

## Rationale

Resolves: #2050

## Acknowledgements

       also allowed.

### DIFF
--- a/physical/raft/raft_test.go
+++ b/physical/raft/raft_test.go
@@ -11,6 +11,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"path/filepath"
 	"strings"
@@ -24,6 +25,7 @@ import (
 	"github.com/hashicorp/raft"
 	"github.com/openbao/openbao/sdk/v2/helper/jsonutil"
 	"github.com/openbao/openbao/sdk/v2/physical"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	bolt "go.etcd.io/bbolt"
 	"google.golang.org/protobuf/proto"
@@ -118,6 +120,8 @@ func TestRaft_TransactionalBackend(t *testing.T) {
 	defer os.RemoveAll(dir)
 
 	physical.ExerciseTransactionalBackend(t, b)
+
+	testRaft_assertFastTxnTrackerCleanup(t, b)
 }
 
 func TestRaft_ParseAutopilotUpgradeVersion(t *testing.T) {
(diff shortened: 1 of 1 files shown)


phil9909:
After leaving it running for longer, the memory consumption has stabilized.

pod_usage

driif (approved):

physical/raft/raft.go
```diff
@@ -1608,6 +1609,7 @@ func (b *RaftBackend) Delete(ctx context.Context, path string) error {
 				Key:    path,
 			},
 		},
+		LowestActiveIndex: pointerutil.Ptr(b.fsm.fastTxnTracker.lowestActiveIndex()),
```
cipherboy: Why apply this here and not globally in `applyLog(...)`? IMHO given we want to send it on every single log regardless of type, it makes the most sense.

physical/raft/raft.go
```diff
@@ -1734,6 +1737,13 @@ func (b *RaftBackend) applyLog(ctx context.Context, command *LogData) error {
 		return err
 	}
 
+	if command.LowestActiveIndex != nil { // we need to cap the lowest active index - 1, otherwise we might miss transaction started concurrently
```
cipherboy: If we do that, we can simplify this as well; command should always set `LowestActiveIndex` and apply it here as `min(b.raft.AppliedIndex() - 1, b.fsm.fastTxnTracker.lowestActiveIndex())`.

cipherboy (commented):
Just a thought to simplify our lives for later @phil9909!

physical/raft/raft.go
phil9909: We need to distinguish between commit and anything else.

To figure out the `LowestActiveIndex` for commit we ask "what will be the lowest active after we have commit this id" and for other its "what is currently the lowest active"

But we could check in `applyLog` if it has be set already and if not fill it with the default, so `Commit` can overwrite this.

cipherboy: @phil9909 Ah, I see, `completeTransaction` is called in the defer, after `applyLog(...)`, which makes sense. Yeah, I think I concur that having `Commit(...)` do it manually and then have `applyLog` do the standard case is probably best.

cipherboy:
@phil9909 I think you can also build a test case modeled off of `TestRaft_Backend_ThreeNode`, which shows that commit indices are cleaned up post-apply on all nodes.

PR #2074, opened 2025-11-07
Make raft tests faster

phil9909:
Add some `t.Parallel()` calls to the raft tests.

~`TestRaft_TransactionalBackend` did fail with `Parallel` on, I need to look into this.~

Also use `t.TempDir()` instead of `os.MkdirTemp()` + `defer os.RemoveAll()` to simplify tests.

On my machine:

```bash
$ git switch main
Switched to branch 'main'
Your branch is up to date with 'upstream/main'.

$ go test ./physical/raft/ -count=1
ok  	github.com/openbao/openbao/physical/raft	189.846s

$ git switch make-raft-tests-faster
Switched to branch 'make-raft-tests-faster'
Your branch is up to date with 'origin/make-raft-tests-faster'.

$ go test ./physical/raft/ -count=1
ok  	github.com/openbao/openbao/physical/raft	39.017s
```

Obviously this speedup depends on the machine and what other (tests) are already running, but as some of the tests call `time.Sleep(10 * time.Second)` even very slow machines should see a speedup.

## Acknowledgements

       also allowed.

### DIFF
--- a/physical/raft/chunking_test.go
+++ b/physical/raft/chunking_test.go
@@ -183,12 +183,16 @@ func TestFSM_Chunking_TermChange(t *testing.T) {
 }
 
 func TestRaft_Chunking_AppliedIndex(t *testing.T) {
-	t.Parallel()
-
+	// t.Parallel() this can't be parallel, because we modify a global variable "raftchunking.ChunkSize"
 	raft, _ := GetRaft(t, true, false)
 
 	// Lower the size for tests
+	originalChunkSize := raftchunking.ChunkSize
 	raftchunking.ChunkSize = 1024
+	t.Cleanup(func() {
+		raftchunking.ChunkSize = originalChunkSize
+	})
+
 	val, err := uuid.GenerateRandomBytes(3 * raftchunking.ChunkSize)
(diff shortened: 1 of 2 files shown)


cipherboy:
@phil9909 `TestRaft_Snapshot_Loading` and `TestRaft_Chunking_AppliedIndex` look like they conflict with data races.

phil9909:
> @phil9909 `TestRaft_Snapshot_Loading` and `TestRaft_Chunking_AppliedIndex` look like they conflict with data races.

Interestingly `"github.com/hashicorp/go-raftchunking"` defines the chunk size as a global variable, which we lower in `TestRaft_Chunking_AppliedIndex` to force chunking without sending large amounts of data. Made that test non parallel and also added a cleanup hook to restore the original chunk size after the test. `TestRaft_TransactionalBackend` now also works in parallel. Probably just failed because of the concurrent modification of the chunk size.

cipherboy:
We shall give this a shot. I think in some scenarios the tests were too expensive to run in parallel, but if we see issues with that we can always adjust as needed later. Thanks @phil9909!

PR #2079, opened 2025-11-07
Allow cache-less OCSP handling

cipherboy:
## Description

When the OCSP server does not indicate that it caches responses, we should not cache responses either to always ensure up-to-date information. This fixes a bug where certificate authentication would fail, yielding an error like:

> invalid validity:
>   producedAt: 2025-11-05 16:04:33 +0000 UTC,
>   thisUpdate: 2025-11-05 16:04:33 +0000 UTC,
>   nextUpdate: 0001-01-01 00:00:00 +0000 UTC

## Rationale

Resolves: #2068

## Acknowledgements

       also allowed.

### DIFF
--- a/sdk/helper/ocsp/client.go
+++ b/sdk/helper/ocsp/client.go
@@ -167,7 +167,7 @@ func validateOCSP(ocspRes *ocsp.Response) (*ocspStatus, error) {
 	if ocspRes == nil {
 		return nil, errors.New("OCSP Response is nil")
 	}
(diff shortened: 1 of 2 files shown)


cipherboy:
Pushed, thanks @wslabosz-reply!

cipherboy:
@wslabosz-reply Ok, this should properly fix test cases now. :laughing:

wslabosz-reply (approved):

Issue #1794, opened 2025-09-11
Add support for forwarded Tls-Client-Cert


ochnerd:
**Is your feature request related to a problem? Please describe.**
We want to use vault with client auth behind an nginx ingress.
It is possible to passtrough Client Certs with x-forwarded-for headers but the listener needs to accept them.

**Describe the solution you'd like**
Allow the specific x-forwarded-for headers to be accepted.

**Explain any additional use-cases**
Our Vault runs behind a Nginx Ingress and we use client auth to authentik our machines. Hashicorp Vault allows this feature but as of now it is missing. It was implemented with this PR: https://github.com/hashicorp/vault/pull/17272

cipherboy:
@ochnerd Apologies; I'm very much in favor of this topic in general, but copying HashiCorp's implementation without JasonN3 being the one to contribute it isn't the right approach.

I do remember that I had some misgivings about the original PR; alas, time has muddled my memory and I don't recall what they are.

Sadly, having looked at the Vault implementation, I think we'll have to disqualify you from contributing in this area. But if anyone else is interested, I'm more than happy to discuss changes in this area.

ochnerd:
@cipherboy no problem and as mentioned in my PR totally understable.

scj643:
I have a working implementation that is a clean room implementation.

PR #2080, opened 2025-11-08
Add support for forwarded Tls-Client-Cert

scj643:
## Description
Add support for handling `XForwardedForClientCertHeader`, and `XForwardedForClientCertDecoders`.
This implementation works at the handler level that stuffs a base64 encoded DER certificate into `X-Processed-TLS-Client-Certificate` while making sure to prevent it from being set externally.
This is then used in the cert auth plugin to handle logging in.

## Rationale
Implement a mostly upstream compatible method of processing `XForwardedForClientCertHeader`
This implementation also removes the requirement for a tls connection for using client certs. (One use case for this is having bao on a unix socket)
The one incompatability is the renaming of `DER` decoder to `PEM` due to upstream having the `DER` decoder actually mean take PEM and encode it as DER

## TODO
- [x] Add tests for new codepaths.
- [ ] Decide if we want the mount to have a config option to use the connection or proxy header.
- [x] Documentation
- [ ] Secure the upstream header so that only the proxy can set it. (Most likely with a static header that is shared between OpenBao and the proxy)

Resolves: #1794

## Acknowledgements

 - [ x ] By contributing this change, I certify I have not used generative AI
 - [ x ] By contributing this change, I certify I have signed-off on the
       also allowed.

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -2966,8 +2966,13 @@ func startHttpServers(c *ServerCommand, core *vault.Core, config *server.Config,
 		if len(ln.Config.XForwardedForAuthorizedAddrs) > 0 {
 			handler = vaulthttp.WrapForwardedForHandler(handler, ln.Config)
 		}
+		// Small handler that always removes `X-Processed-Tls-Client-Certificate`
+		handler = vaulthttp.WrapRemoveProcessedCertificateHandler(handler)
 		if len(ln.Config.XForwardedForClientCertHeader) > 0 {
 			handler = vaulthttp.WrapClientCertificateHandler(handler, ln.Config)
+		} else {
+			// Remove Client-Cert if we are not using them
+			handler = vaulthttp.WrapRemoveDefaultClientCertificateHandler(handler)
 		}
 
 		// server defaults
--- a/http/handler.go
+++ b/http/handler.go
@@ -584,6 +584,24 @@ func WrapForwardedForHandler(h http.Handler, l *configutil.Listener) http.Handle
 	})
 }
 
+func WrapRemoveProcessedCertificateHandler(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Strip any X-Processed-Tls-Client-Certificate headers.
+		r.Header.Del("X-Processed-Tls-Client-Certificate")
+		h.ServeHTTP(w, r)
+		return
+	})
+}
+
+func WrapRemoveDefaultClientCertificateHandler(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Remove any Client-Cert if not configured to use them. To prevent injection of client certs in plugins.
+		r.Header.Del("Client-Cert")
+		h.ServeHTTP(w, r)
+		return
+	})
+}
+
 func WrapClientCertificateHandler(h http.Handler, l *configutil.Listener) http.Handler {
 	clientCertificateHeader := l.XForwardedForClientCertHeader
 	// Iterate through the processors to handle the header.
--- a/http/forwarding_certificate_test.go
+++ b/http/forwarding_certificate_test.go
@@ -29,7 +29,21 @@ func getTestHandler(listenerConfig *configutil.Listener) func(props *vault.Handl
 			w.Header().Set("X-Processed-Tls-Client-Certificate-Resp", r.Header.Get("X-Processed-Tls-Client-Certificate"))
 			w.WriteHeader(http.StatusOK)
 		})
-		return WrapClientCertificateHandler(origHandler, listenerConfig)
+		// This mirrors how the handlers would be set in server.go
+		return WrapRemoveProcessedCertificateHandler(WrapClientCertificateHandler(origHandler, listenerConfig))
+	}
+}
+
+// Test for configs that have WrapClientCertificate disabled.
+func getTestHandlerNoWrapping(listenerConfig *configutil.Listener) func(props *vault.HandlerProperties) http.Handler {
+	return func(props *vault.HandlerProperties) http.Handler {
+		origHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Write out what we internally processed.
+			w.Header().Set("X-Processed-Tls-Client-Certificate-Resp", r.Header.Get("X-Processed-Tls-Client-Certificate"))
+			w.WriteHeader(http.StatusOK)
+		})
+		// This mirrors how the handlers would be set in server.go
+		return WrapRemoveProcessedCertificateHandler(origHandler)
 	}
 }
 
@@ -346,6 +360,36 @@ func TestHandler_XForwardedForClientCert(t *testing.T) {
 	t.Run("inject_processed_header", func(t *testing.T) {
 		t.Parallel()
 		testHandler := getTestHandler(getDefaultListenerConfigForClientCerts([]string{}))
+		cluster := vault.NewTestCluster(t, nil, &vault.TestClusterOptions{
+			HandlerFunc: HandlerFunc(testHandler),
+		})
+		cluster.Start()
+		defer cluster.Cleanup()
+		client := cluster.Cores[0].Client
+		req := client.NewRequest("GET", "/")
+		req.Headers = make(http.Header)
+		// Intentionally use the wrong canonical name.
+		req.Headers.Add("X-Processed-TLS-Client-Certificate", clientCertBase64)
+		resp, err := client.RawRequest(req)
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer func(Body io.ReadCloser) {
+			err := Body.Close()
+			if err != nil {
+				t.Fatal("failed to close response body")
+			}
+		}(resp.Body)
+		if resp.StatusCode != http.StatusOK {
+			t.Fatalf("bad status: %d", resp.StatusCode)
+		}
+		if resp.Header.Get("X-Processed-Tls-Client-Certificate-Resp") != "" {
+			t.Fatal("client certificate header should not have been set")
+		}
+	})
+
+	t.Run("inject_processed_header_no_handler", func(t *testing.T) {
+		testHandler := getTestHandlerNoWrapping(&configutil.Listener{})
 		cluster := vault.NewTestCluster(t, nil, &vault.TestClusterOptions{
 			HandlerFunc: HandlerFunc(testHandler),
 		})


scj643:
@cipherboy when you get a chance can you take a look at this. The code for handling the headers seems to be set. Just need to write tests for header processing.

http/handler.go
```diff
@@ -583,6 +585,61 @@ func WrapForwardedForHandler(h http.Handler, l *configutil.Listener) http.Handle
...
+		if forwardedHeaderOK || len(forwardedHeader) > 0 {
+			r.Header.Del("X-Processed-TLS-Client-Certificate")
+		}
+		clientCertHeaders, clientCertHeadersOK := r.Header[clientCertificateHeader]
+		// Short circuit if no client cert header
+		if !clientCertHeadersOK || len(clientCertHeaders) == 0 {
+			h.ServeHTTP(w, r)
+			return
+		}
+		// TODO: Handle multiple clientCertHeaders
```
cipherboy: While I understand this is true of how headers work, is this likely in practice, for a load-balancer to send multiple client certs? Do we have an example of one? IIRC TLS only negotiates down to a single leaf cert? 

Would it send along intermediates here? 🤔

http/handler.go
```diff
@@ -583,6 +585,61 @@ func WrapForwardedForHandler(h http.Handler, l *configutil.Listener) http.Handle
...
+				if decoders[i] == "PEM" {
+					block, _ := pem.Decode([]byte(headerValue))
+					if block == nil || block.Type != "CERTIFICATE" {
+						respondError(w, http.StatusBadRequest, errors.New("failed to decode PEM certificate"))
+						return
+					}
+					headerValue = base64.StdEncoding.EncodeToString(block.Bytes)
+				}
+			}
+		}
```
cipherboy, line 636: I'd suggest doing a `DER -> *x509.Certificate` parse operation and reject the request with a malformed header err early, in the event the certificate is not parsable. That saves us a lot of overhead.

http/logical.go
```diff
@@ -227,6 +245,11 @@ func buildLogicalRequestNoAuth(w http.ResponseWriter, r *http.Request) (*logical
 		Headers:    r.Header,
 	}
 
+	headerCert := handleCertHeader(r)
+	if headerCert != nil {
+		req.ClientHeaderCert = headerCert
```
cipherboy, line 250: (Just cross-linking to discussion for others reviewing:  https://linuxfoundation.zulipchat.com/#narrow/channel/529890-openssf-openbao-discussion/topic/Forwarded.20Tls-Client-Cert/near/554761022)

command/server.go
```diff
@@ -2966,6 +2966,9 @@ func startHttpServers(c *ServerCommand, core *vault.Core, config *server.Config,
 		if len(ln.Config.XForwardedForAuthorizedAddrs) > 0 {
 			handler = vaulthttp.WrapForwardedForHandler(handler, ln.Config)
 		}
+		if len(ln.Config.XForwardedForClientCertHeader) > 0 {
```
cipherboy, line 2971: Ah, I see the issue. We need listener config, but `buildLogicalRequestNoAuth` wouldn't have it, so we might be stuck with the logic in `WrapClientCertificateHandler` (+/- moving it into `http/handler.go`) and then "deleting" the header later.

http/handler.go
cipherboy: Ah, never mind, I see you do that in `handleCertHeader`; might be worth adding a comment there.

builtin/credential/cert/path_login.go
```diff
@@ -693,9 +699,8 @@ func parsePEM(raw []byte) (certs []*x509.Certificate) {
 // by at trusted certificate. Most of this logic is lifted from the client
 // verification logic here:  http://golang.org/src/crypto/tls/handshake_server.go
 // The trusted chains are returned.
-func validateConnState(roots *x509.CertPool, cs *tls.ConnectionState) ([][]*x509.Certificate, error) {
-	certs := cs.PeerCertificates
-	if len(certs) == 0 {
+func validateConnState(roots *x509.CertPool, cert *x509.Certificate) ([][]*x509.Certificate, error) {
```
cipherboy, line 702:
```suggestion
func validateConnState(roots *x509.CertPool, certs ...*x509.Certificate) ([][]*x509.Certificate, error) {
```

perhaps?

Then all the logic can remain and you can call it with what you have (either `cs.PeerCertificates` or `logical.ClientHeaderCert`). 

Just a thought.

sdk/logical/request.go
```diff
@@ -108,6 +109,9 @@ type Request struct {
 	// authentication/protection.
 	Connection *Connection `json:"connection" structs:"connection" mapstructure:"connection"`
 
+	// ClientHeaderCert is the x509 client certificate passed from a header. nil if unset.
+	ClientHeaderCert *x509.Certificate
```
cipherboy, line 113:
```suggestion
	ClientHeaderCert []*x509.Certificate
```

might future-proof it and allow us to handle chains in the future.

Perhaps clarifying `certs[0]` is the leaf and `certs[1...]` are the intermediates, if any.

builtin/credential/cert/backend_test.go
```diff
@@ -1169,18 +1220,24 @@ func TestBackend_basic_CA(t *testing.T) {
 		CredentialBackend: testFactory(t),
 		Steps: []logicaltest.TestStep{
 			testAccStepCert(t, "web", ca, "foo", allowed{}, false),
-			testAccStepLogin(t, connState),
+			testAccStepLogin(t, connState, false),
+			testAccStepLogin(t, connState, true),
```
cipherboy, line 1226: Not every function needs to do both, though I think this is fine if it doesn't take too long to run. 😆

builtin/credential/cert/path_login.go
```diff
@@ -60,6 +59,23 @@ func pathLogin(b *backend) *framework.Path {
 	}
 }
 
+// Handle headers and client certs
+func (b *backend) handleCerts(req *logical.Request) (*x509.Certificate, error) {
+	if req.ClientHeaderCert != nil {
+		return req.ClientHeaderCert, nil
```
cipherboy, line 65: The above also keeps return typing the same fwiw. Both would now be lists.

http/handler.go
scj643: I need to test that with NGINX. Just haven't gotten around to it because I have been testing against a Yubikey which only sends the leaf certificate.

scj643: So at least with NGINX it seems it will only return the leaf certificate.

scj643: RFC9440 requires the server to only return a single certificate so I think we should not handle multiple certs.

builtin/credential/cert/backend_test.go
scj643: Got to have that code coverage though :)

command/server.go
cipherboy: @scj643 I still maintain this should be moved into `http/handler.go` and applied with the logic inverted:

the `ln.Config != nil && len(ln.Config.XForwardedForClientCertHeader) > 0` check needs to be inside `WrapClientCertificateHandler` so that we can always remove headers if not explicitly configured.

That way `handleCertHeader` correctly ignores forwarded client certs when the listener is not configured to do forwarded cert processing.

Otherwise, right now, I'm fairly certain if we do not set the header, we can set arbitrary client certs from a non-proxy direct client and it silently works, without needing to prove possession of the key material.

http/handler.go
```diff
@@ -583,6 +585,71 @@ func WrapForwardedForHandler(h http.Handler, l *configutil.Listener) http.Handle
...
+		clientCertHeaders, clientCertHeadersOK := r.Header[textproto.CanonicalMIMEHeaderKey(clientCertificateHeader)]
+		// Short circuit if no client cert header
+		if !clientCertHeadersOK || len(clientCertHeaders) == 0 {
+			h.ServeHTTP(w, r)
+			return
+		}
+		// Do not handle multiple certs.
+		// This is out of scope and if a proxy is following RFC 9440 the chain excluding the end entity would be a separate header
+		headerValue := clientCertHeaders[0]
+		for i := range decoders {
```
cipherboy: nit: I think this could be a little more cleanly written as a `switch` plus a bunch of helper methods:

```go
for _, decoder := range decoders {
  var err error
  switch decoder {
  case "RFC9440":
    headerValue, err = rfc9440Decode(headerValue)
  case "URL":
    headerValue, err = urlDecode(headerValue)
  case "PEM":
    headerValue, err = pemDecode(headerValue)
  }

  if err != nil {
    respondError(w, http.StatusBadRequest, err)
    return
  }
}
```

http/logical.go
cipherboy: @scj643 I do kinda like something like this:

<details>

<summary> proposed diff </summary>

```diff
diff --git a/http/logical.go b/http/logical.go
index 7f0131ea75..ba3698e7f5 100644
--- a/http/logical.go
+++ b/http/logical.go
@@ -245,11 +245,6 @@ func buildLogicalRequestNoAuth(w http.ResponseWriter, r *http.Request) (*logical
                Headers:    r.Header,
        }
 
-       headerCert := handleCertHeader(r)
-       if headerCert != nil {
-               req.ClientHeaderCert = headerCert
-       }
-
        if passHTTPReq {
                req.HTTPRequest = r
        }
@@ -621,5 +616,17 @@ func getConnection(r *http.Request) (connection *logical.Connection) {
                RemotePort: remotePort,
                ConnState:  r.TLS,
        }
+
+       if r.TLS != nil {
+               connection.PeerCertificates = r.TLS.PeerCertificates
+       }
+
+       headerCert := handleCertHeader(r)
+       if headerCert != nil {
+               connection.ProxiedCertificates = []*x509.Certificate{headerCert}
+       }
+
+       connection.PopulateRawCerts()
+
        return connection
 }
diff --git a/sdk/logical/connection.go b/sdk/logical/connection.go
index e590e6f59a..069e48390e 100644
--- a/sdk/logical/connection.go
+++ b/sdk/logical/connection.go
@@ -5,6 +5,8 @@ package logical
 
 import (
        "crypto/tls"
+       "crypto/x509"
+       "fmt"
 )
 
 // Connection represents the connection information for a request. This
@@ -18,4 +20,57 @@ type Connection struct {
 
        // ConnState is the TLS connection state if applicable.
        ConnState *tls.ConnectionState `sentinel:""`
+
+       // ProxiedCertificate is a certificate verified by a proxy header.
+       ProxiedCertificates    []*x509.Certificate `json:"-"`
+       ProxiedCertificatesRaw [][]byte            `json:"proxied_certificates"`
+
+       // PeerCertificate is a certificate verified by this server through a TLS
+       // connection.
+       PeerCertificates    []*x509.Certificate `json:"-"`
+       PeerCertificatesRaw [][]byte            `json:"peer_certificates"`
+}
+
+func (c *Connection) PopulateRawCerts() {
+       if len(c.ProxiedCertificates) > 0 {
+               c.ProxiedCertificatesRaw = make([][]byte, len(c.ProxiedCertificates))
+               for i, cert := range c.ProxiedCertificates {
+                       c.ProxiedCertificatesRaw[i] = cert.Raw
+               }
+       }
+
+       if len(c.PeerCertificates) > 0 {
+               c.PeerCertificatesRaw = make([][]byte, len(c.PeerCertificates))
+               for i, cert := range c.PeerCertificates {
+                       c.PeerCertificatesRaw[i] = cert.Raw
+               }
+       }
+}
+
+func (c *Connection) PopulateParsedCerts() error {
+       if len(c.ProxiedCertificatesRaw) > 0 {
+               c.ProxiedCertificates = make([]*x509.Certificate, len(c.ProxiedCertificatesRaw))
+               for i, rawCert := range c.ProxiedCertificatesRaw {
+                       cert, err := x509.ParseCertificate(rawCert)
+                       if err != nil {
+                               return fmt.Errorf("[cert %d]: failed to parse: %w", i, err)
+                       }
+
+                       c.ProxiedCertificates[i] = cert
+               }
+       }
+
+       if len(c.PeerCertificatesRaw) > 0 {
+               c.PeerCertificates = make([]*x509.Certificate, len(c.PeerCertificatesRaw))
+               for i, rawCert := range c.PeerCertificatesRaw {
+                       cert, err := x509.ParseCertificate(rawCert)
+                       if err != nil {
+                               return fmt.Errorf("[cert %d]: failed to parse: %w", i, err)
+                       }
+
+                       c.PeerCertificates[i] = cert
+               }
+       }
+
+       return nil
 }
diff --git a/sdk/logical/request.go b/sdk/logical/request.go
index 977569bbb2..404c34b947 100644
--- a/sdk/logical/request.go
+++ b/sdk/logical/request.go
@@ -5,7 +5,6 @@ package logical
 
 import (
        "context"
-       "crypto/x509"
        "fmt"
        "io"
        "net/http"
@@ -109,9 +108,6 @@ type Request struct {
        // authentication/protection.
        Connection *Connection `json:"connection" structs:"connection" mapstructure:"connection"`
 
-       // ClientHeaderCert is the x509 client certificate passed from a header. nil if unset.
-       ClientHeaderCert *x509.Certificate
-
        // ClientToken is provided to the core so that the identity
        // can be verified and ACLs applied. This value is passed
        // through to the logical backends but after being salted and
```

</details>

Ultimately it lets us drop _most_ usage of the direct TLS connection state which is good for two reasons:

1. Forwarded requests get easier, though these already worked.
2. External auth methods can now use it, as we can protobuf encode/decode `logical.Connection`'s other fields and _just_ drop the connection state. Fewer plugins care about the remaining TLS details, we mostly just care about leaf certificate (+ chain).
3. This reverts a lot of the "single cert" handling that you had, which would break backwards compatibility.

I think the main downside is that a lot of `backend_test.go` &c will have to change, to create `logical.Connection` and ignore the `ConnectionState`, but IMHO this is a good thing and probably is a big net-simplification for the test suite.

LMKWYT!

sdk/logical/request.go
cipherboy: (proposed alternative in https://github.com/openbao/openbao/pull/2080/files#r2511574637)

website/content/docs/configuration/listener/index.mdx
```diff
@@ -12,5 +12,8 @@ respond to requests. At this time, there are two listeners:
 - [TCP][tcp]
 - [Unix Domain Socket][unix]
 
+There is a [shared config][shared] that is common to all listeners.
```
cipherboy, line 15: This will be nice to have! What do you think about instead using the [partials system](https://github.com/openbao/openbao/tree/main/website/content/partials)? See e.g., https://github.com/openbao/openbao/blob/main/website/content/docs/internals/telemetry/index.mdx?plain=1#L18.

website/content/docs/configuration/listener/shared.mdx
```diff
@@ -0,0 +1,79 @@
...
+# Shared listener parameters
+
+There is config that is common between all listeners.
+
+## Client certificate header parameters
+
+- `x_forwarded_for_client_cert_header` `(string: "")` -
+  Specifies the header to be used for client certificate authentication. This is required
+  if [Client Certificate Auth](/docs/auth/cert.mdx) is used and your server is behind a reverse
+  proxy. It is recommended to have this set to the canonical `Client-Cert` header as defined in
```
cipherboy, line 16: In our guarding logic, I wonder if its worth proactively removing this header as well?

cipherboy (commented):
Few comments! Looks like we're on the right path :-) Thank you!

website/content/docs/configuration/listener/index.mdx
scj643: That sounds good

website/content/docs/configuration/listener/shared.mdx
scj643: It gets removed after it's been processed.

website/content/docs/configuration/listener/index.mdx
scj643: We might want to split it out into two. A partial with the parameters but a link to a full page of documentation on the client cert header functionality.

website/content/docs/configuration/listener/shared.mdx
cipherboy: @scj643 Sorry, this was also about https://github.com/openbao/openbao/pull/2080#discussion_r2528743263. I think we should always remove both this header but especially the processed header if this config flag is not set

http/handler.go
scj643: Sounds good. I'll put the functions in `http/util.go`

command/server.go
scj643: From my understanding if it's unset the `WrapClientCertificateHandler` never gets called.

scj643: Oh I see what you mean now. We can make it so that if it's unset we just add a stub handler that always removes the header.

scj643:
```go
func WrapRemoveProcessedCertificateHandler(h http.Handler, _ *configutil.Listener) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		// Strip any X-Processed-Tls-Client-Certificate headers.
		r.Header.Del("X-Processed-Tls-Client-Certificate")
		h.ServeHTTP(w, r)
		return
	})
}
```

cipherboy: @scj643 Yes, exactly. And also remove the default `Client-Cert` header. That way it doesn't accidentally get processed either by some downstream plugin and treated as legitimate in an environment where we aren't using proxies.

(Technically there's `passthrough_request_headers` to prevent that, but that's tenant controllable, and not controlled by the infra operator like this is.)

http/handler.go
```diff
@@ -583,6 +584,52 @@ func WrapForwardedForHandler(h http.Handler, l *configutil.Listener) http.Handle
...
 
+func WrapClientCertificateHandler(h http.Handler, l *configutil.Listener) http.Handler {
+	clientCertificateHeader := l.XForwardedForClientCertHeader
+	// Iterate through the processors to handle the header.
+	decoders := l.XForwardedForClientCertDecoders
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Strip any X-Processed-Tls-Client-Certificate headers.
+		r.Header.Del("X-Processed-Tls-Client-Certificate")
+		// Use the CanonicalMIMEHeaderKey so that if the user has the header not canonicalized in the config.
+		clientCertHeaders, clientCertHeadersOK := r.Header[textproto.CanonicalMIMEHeaderKey(clientCertificateHeader)]
```
cipherboy, line 613: nit:

```suggestion
		clientCertHeaders, clientCertHeadersOK := r.Header.Get(clientCertificateHeader)
```

http/logical.go
```diff
@@ -38,6 +39,23 @@ func newBufferedReader(r io.ReadCloser) *bufferedReader {
 	}
 }
 
+func handleCertHeader(req *http.Request) *x509.Certificate {
+	clientCertHeader, clientCertHeaderOk := req.Header["X-Processed-Tls-Client-Certificate"]
```
cipherboy, line 43: nit:

```suggestion
	clientCertHeader, clientCertHeaderOk := req.Header.Get("X-Processed-Tls-Client-Certificate")
```

http/handler.go
```diff
@@ -583,6 +584,52 @@ func WrapForwardedForHandler(h http.Handler, l *configutil.Listener) http.Handle
...
+				respondError(w, http.StatusBadRequest, err)
+				return
+			}
+		}
+		// Validate that the processed cert is valid StdEncoding base64
+		_, err := base64.StdEncoding.DecodeString(headerValue)
+		if err != nil {
+			respondError(w, http.StatusBadRequest, errors.New("error decoding client certificate header as base64"))
+			return
+		}
```
cipherboy, line 642: We should probably also call `ParseCertificate` here so that we can return a proper server err rather than silently ignoring the error. Then `logical.go` essentially cannot error unless it is a non-deterministic parse bug.

command/server.go
scj643: Removing the default header like that will require another handler but sure.

scj643: https://github.com/openbao/openbao/pull/2080/commits/29c90a783fb460b09cf95379b8c2359725217e78

cipherboy:
Closing; I will provide a follow-up PR which squashes this one and replaces it with updates I want to see before landing. 

Thank you for the contribution @scj643!

Issue #2097, opened 2025-11-13
PostgreSQL backend - does not detect lock removal


cipherboy:
**Describe the bug**

As a variant of #2096, the lock code does not detect truncation/removal of the lock. Instead it is silently upserted as if the lock had been held the entire time.

Query: https://github.com/openbao/openbao/blob/17a15121d531e050e1f29d2d6c8ead31c92a1800/physical/postgresql/postgresql.go#L218-L225

Maintenance loop: https://github.com/openbao/openbao/blob/17a15121d531e050e1f29d2d6c8ead31c92a1800/physical/postgresql/postgresql.go#L605-L614

Query exec: https://github.com/openbao/openbao/blob/17a15121d531e050e1f29d2d6c8ead31c92a1800/physical/postgresql/postgresql.go#L620-L625

Renew interval: https://github.com/openbao/openbao/blob/17a15121d531e050e1f29d2d6c8ead31c92a1800/physical/postgresql/postgresql.go#L34-L35

**To Reproduce**

Same as #2096, just truncate `openbao_ha_locks` instead.

**Expected behavior**

I'd expect the active node to step down, as if it had lost the lock, rather than silently re-acquiring it.

**Environment:**

Same as #2096.

**Additional context**

n/a

PR #2100, opened 2025-11-13
Ensure PostgreSQL writes fail on lock loss

cipherboy:
Based on #1571; will be rebased when that merges. 

## Description

When a lock is manually removed from the PostgreSQL backend (say with
`TRUNCATE`), PostgreSQL will silently renew this lock without informing
the instance that the lock was lost and re-created. In the case of all
data being removed, this leads to the node not noticing.

While this partly is caused by the lack of fencing on writes (which this
is built on top of, and thus remediating), it also needs to be a
property of the lock itself: upsert should only be used for initial
creation (and strictly limited to other identities) and renewal should
not attempt to steal the lock from an another or insert it if missing.

This latter case was also a failure mode: in the total ordering:

- A acquires lock
- B steals the lock, perhaps because A failed to renew.
- B goes stale.
- A's renewal finally executes, taking the lock from B.

A will silently think it has been the active node this whole time and
continue operations, ignoring that B previously held the lock. The total
time would be on the order of 30s based on our lock holding periods, which
isn't an unreasonably long delay (say, if A was overloaded).

## Rationale

Resolves: #2097

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2100.txt
+++ b/changelog/2100.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+physical/postgresql: ensure underlying HA lock removal from database causes lock loss, write failures
+```


cipherboy:
@openbao/openbao-core-committers this one should be ready for review now.

physical/postgresql/postgresql.go
```diff
@@ -704,12 +720,32 @@ func (l *PostgreSQLLock) periodicallyRenewLock(done chan struct{}) {
 // evaluate the TTL.  Returns true if the lock was obtained, false if not.
 // If false error may be nil or non-nil: nil indicates simply that someone
 // else has the lock, whereas non-nil means that something unexpected happened.
-func (l *PostgreSQLLock) writeItem() (bool, error) {
+//
```
satoqz, line 723:
```suggestion
```

cipherboy: @satoqz Why wouldn't we break this into two paragraphs?

satoqz: Ah, no, ignore. Reading diffs is hard.

PR #2101, opened 2025-11-14
plugins: make sure all database plugins ignore not found on revoke

phil9909:
## Description

- [x] Cassandra: The CQL (cassandra query language) supports `DROP USER IF EXISTS`, that default revoke statement was adapted accordingly
- [x] InfluxDB: there is no support in the query language, but we can match on the `"user not found"` error
- [x] MySQL: ??? see below
- [x] PostgreSQL: no action required, in `defaultDeleteUser` https://github.com/openbao/openbao/blob/17a15121d531e050e1f29d2d6c8ead31c92a1800/plugins/database/postgresql/postgresql.go#L411-L419 we already check, if the users exits.
- [x] ValKey: The [`ACL DELUSER`](https://valkey.io/commands/acl-deluser/) operation ignores missing user by default. I adapted the test to verify this.

## MySQL

For MySQL we could do

```SQL
REVOKE ALL PRIVILEGES, GRANT OPTION FROM '{{name}}'@'%' IGNORE UNKNOWN USER;
DROP USER IF EXISTS '{{name}}'@'%'
```

But MariaDB (a fork of MySQL which shares the plugin) does not support the `IGNORE UNKNOWN USER` modifier on `REVOKE` (`DROP USER IF EXISTS` is supported).

We could:

- Implement a similar approach to PostgreSQL https://github.com/openbao/openbao/blob/17a15121d531e050e1f29d2d6c8ead31c92a1800/plugins/database/postgresql/postgresql.go#L348-L357 there the plugin checks, if custom statements are provided. If not deletion uses a custom go logic, which queries for the user first existence first.
- detect if we are connected to MySQL or MariaDB (`SELECT VERSION();` can be used for this) and use different defaults based on that (which would mean that missing users are not ignored for MariaDB)
- ignore the error based on string matching, but the message is rather generic `Can't revoke all privileges for one or more of the requested users`, which might lead to false positives.

Opinions?

## Rationale

See https://openbao.org/docs/plugins/plugin-authors-guide/#revoke-operations-should-ignore-not-found-errors

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2101.txt
+++ b/changelog/2101.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+database: all database plugins now ignore "not found" errors on revoke by default. See [Plugin Author Guide](https://openbao.org/docs/plugins/plugin-authors-guide/#revoke-operations-should-ignore-not-found-errors) for rationale.
+```


cipherboy:
@phil9909 Following the PostgreSQL logic for consistency on MySQL &co seem fine. Thanks!

cipherboy:
@JanMa While less complicated due to the lack of permission grants, this follows the example of PostgreSQL: 

https://github.com/openbao/openbao/blob/b65dcb13d58d750c4208e3bdcedbc04c36fb77c0/plugins/database/postgresql/postgresql.go#L404-L419

which later calls the deletion logic like this does.

cipherboy:
@phil9909 This needs a changelog, please! Thank you!

cipherboy:
Race failure unrelated, rerunning:

```
2025-11-18T13:41:28.1745898Z ##[endgroup]
2025-11-18T13:41:28.3761522Z ##[group][31mFAIL[0m vault/external_tests/plugin.TestExternalPlugin_AuthMethod/parallel_execution_group/vault-plugin-auth-approle-4 (1.90s)
2025-11-18T13:41:28.3762847Z     external_plugin_test.go:346: Error making API request.
2025-11-18T13:41:28.3763339Z         
2025-11-18T13:41:28.3764601Z         URL: PUT https://127.0.0.1:35739/v1/auth/vault-plugin-auth-approle-4/role/role1/secret-id
2025-11-18T13:41:28.3765494Z         Code: 404. Errors:
2025-11-18T13:41:28.3765827Z         
2025-11-18T13:41:28.3766224Z         * role "role1" does not exist
```

PR #2114, opened 2025-11-18
Add Kubernetes JWT provider that authenticates with pod Service Account token

### LINKED ISSUE #2113: Support authenticated OIDC discovery for Kubernetes-based deployments

When running OpenBao on Kubernetes, there are two available authentication methods for validating Kubernetes Service Accounts:

* JWT/OIDC authentication method
* Kubernetes authentication method

Both currently require modifications to Kubernetes RBAC configuration:

* JWT/OIDC: Requires making the Kubernetes API server’s OIDC discovery endpoint accessible without authentication.
* Kubernetes: Requires granting OpenBao’s Service Account permission to call the TokenReview API.

This raises concerns from cluster administrators, who are understandably hesitant to do cluster-wide RBAC changes for any application, especially ones that look like they could reduce the security. As a result, this leads to repeated discussions where the need for these modifications must be justified.

The JWT/OIDC authentication method should support authenticating to the Kubernetes API server when fetching the OIDC discovery document and JWKS file.

If OpenBao can authenticate these requests, it would remove the need for RBAC changes, improving API server security and the experience of running OpenBao on Kubernetes.

tsaarni:
## Description 

This change implements a Kubernetes-specific JWT provider that uses the pod's Service Account token and CA certificate for authenticating requests to the Kubernetes API server's OIDC endpoints.

This approach avoids requiring RBAC policy changes on Kubernetes, that would otherwise be needed to allow anonymous access to the OIDC discovery and JWKS endpoints.

The provider is configured in following way

```console
$ bao write auth/jwt/config - <<EOF
{
    "oidc_discovery_url": "https://kubernetes.default.svc.cluster.local",
    "provider_config": {
        "provider": "kubernetes"
    }
}
EOF
```

The provider uses the default mount paths `/var/run/secrets/kubernetes.io/serviceaccount/token` and `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` which removes burden from user to configure anything further than setting the discovery URL and selecting the Kubernetes provider. The default automounted token is sufficient, no RBAC permissions need to be created.

## Rationale 

Resolves: #2113

## Detailed Description

The JWT/OIDC provider interface is extended to allow providers to supply a custom `http.Client` instance to be used for HTTP requests. Currently, this is used for the OIDC discovery document and JWKS endpoints, which are the only endpoints supported by the Kubernetes API server when acting as OIDC provider.

The interface is implemented by the new Kubernetes provider. It creates `http.Client` that uses a custom `http.RoundTripper` which automatically adds the pod’s Service Account token to the `Authorization` header for all requests. This idea is taken from Kubernetes client-go library, which uses the same mechanism to attach the token ([link](https://github.com/kubernetes/client-go/blob/45e0decafa9b847c983f55c84b4f6ce5617f8f69/transport/round_trippers.go#L310-L324)).

HTTP requests are handled in `github.com/hashicorp/cap`,  `github.com/coreos/go-oidc` and `golang.org/x/oauth2`. These libraries support injecting a custom `http.Client` using `context` with a variable with specific context key ([link](https://github.com/golang/oauth2/blob/f28b0b5467dda26d56f1240381158f7c334654d1/oauth2.go#L339-L341)). This PR uses that capability to supply the provider-specific client. This enables authentication without requiring explicit support from the dependency libraries.

Note that `github.com/hashicorp/cap` creates its own client if a CA certificate is provided to it . Fortunately, if no CA certificate is passed, it does not override the context, allowing the provider-created `http.Client` to remain in effect ([link](https://github.com/hashicorp/cap/blob/70a13e3d0b472b4d6c12aec92361724741e6a451/jwt/keyset.go#L206-L227)). For this reason, the Kubernetes provider introduced in this PR does not allow user to specify `oidc_discovery_ca_pem` or `jwks_ca_pem` in the authentication method parameters.

The provider reads the Service Account token and Kubernetes CA certificate directly from pod instead of asking user to provide the content of the files and storing that in the backend configuration. This avoids problems when the files are rotated.  When first JWT/OIDC login is  validated, the `http.Client` is brought to memory via keyset validator and then reused in next logins ([link](https://github.com/openbao/openbao/blob/05457927fb4a8edd7d2c8e527cb2c891f06e9459/builtin/credential/jwt/backend.go#L143-L178)). If the Kubernetes signing keys are rotated  `github.com/coreos/go-oidc` will automatically re-download JWKS with the same client.

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/credential/jwt/backend.go
+++ b/builtin/credential/jwt/backend.go
@@ -151,19 +151,20 @@ func (b *jwtAuthBackend) jwtValidator(config *jwtConfig) (*jwt.Validator, error)
 	var err error
 	var keySet jwt.KeySet
 
-	pConfig, err := NewProviderConfig(b.providerCtx, config, ProviderMap())
-	if err != nil {
-		return nil, fmt.Errorf("error initializing provider_config: %w", err)
-	}
-
 	// Configure the key set for the validator
 	switch config.authType() {
 	case JWKS:
-		keySet, err = jwt.NewJSONWebKeySet(injectProviderHTTPClient(pConfig, b.providerCtx), config.JWKSURL, config.JWKSCAPEM)
+		keySet, err = jwt.NewJSONWebKeySet(b.providerCtx, config.JWKSURL, config.JWKSCAPEM)
 	case StaticKeys:
 		keySet, err = jwt.NewStaticKeySet(config.ParsedJWTPubKeys)
 	case OIDCDiscovery, OIDCFlow:
-		keySet, err = jwt.NewOIDCDiscoveryKeySet(injectProviderHTTPClient(pConfig, b.providerCtx), config.OIDCDiscoveryURL, config.OIDCDiscoveryCAPEM)
+		keySet, err = jwt.NewOIDCDiscoveryKeySet(b.providerCtx, config.OIDCDiscoveryURL, config.OIDCDiscoveryCAPEM)
+	case CustomProviderDiscovery:
+		pConfig, initErr := NewProviderConfig(b.providerCtx, config, ProviderMap())
+		if initErr != nil {
+			return nil, initErr
+		}
+		keySet, err = pConfig.(KeySetDiscovery).NewKeySet(b.providerCtx)
 	default:
 		return nil, errors.New("unsupported config type")
 	}
--- a/builtin/credential/jwt/backend.go
+++ b/builtin/credential/jwt/backend.go
@@ -151,14 +151,19 @@ func (b *jwtAuthBackend) jwtValidator(config *jwtConfig) (*jwt.Validator, error)
 	var err error
 	var keySet jwt.KeySet
 
+	pConfig, err := NewProviderConfig(b.providerCtx, config, ProviderMap())
+	if err != nil {
+		return nil, fmt.Errorf("error initializing provider_config: %w", err)
+	}
+
 	// Configure the key set for the validator
 	switch config.authType() {
 	case JWKS:
-		keySet, err = jwt.NewJSONWebKeySet(b.providerCtx, config.JWKSURL, config.JWKSCAPEM)
+		keySet, err = jwt.NewJSONWebKeySet(injectProviderHTTPClient(pConfig, b.providerCtx), config.JWKSURL, config.JWKSCAPEM)
 	case StaticKeys:
 		keySet, err = jwt.NewStaticKeySet(config.ParsedJWTPubKeys)
 	case OIDCDiscovery, OIDCFlow:
-		keySet, err = jwt.NewOIDCDiscoveryKeySet(b.providerCtx, config.OIDCDiscoveryURL, config.OIDCDiscoveryCAPEM)
+		keySet, err = jwt.NewOIDCDiscoveryKeySet(injectProviderHTTPClient(pConfig, b.providerCtx), config.OIDCDiscoveryURL, config.OIDCDiscoveryCAPEM)
 	default:
 		return nil, errors.New("unsupported config type")
 	}
--- a/builtin/credential/jwt/provider_kubernetes.go
+++ b/builtin/credential/jwt/provider_kubernetes.go
@@ -1,0 +1,112 @@
+// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package jwtauth
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/hashicorp/go-cleanhttp"
+)
+
+// Global variables instead of const to allow test cases to overwrite paths.
+var (
+	// localJWTPath is the path to the Kubernetes Service Account JWT token file.
+	localJWTPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+
+	// localCACertPath is the path to the Kubernetes API server CA certificate.
+	localCACertPath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+)
+
+type KubernetesProvider struct{}
+
+func (k *KubernetesProvider) Initialize(_ context.Context, jc *jwtConfig) error {
+	// Ensure JWKSCAPEM and OIDCDiscoveryCAPEM are not set. These options conflict with
+	// the Kubernetes provider because they configure a custom HTTP client via github.com/hashicorp/cap
+	// when calling jwt.NewJSONWebKeySet() or jwt.NewOIDCDiscoveryKeySet(), preventing
+	// the Kubernetes provider from injecting its own HTTP client with proper CA certificate
+	// and Authorization header.
+	if jc.JWKSCAPEM != "" {
+		return errors.New("jwks_ca_pem must not be set when using the kubernetes provider")
+	}
+
+	if jc.OIDCDiscoveryCAPEM != "" {
+		return errors.New("oidc_discovery_ca_pem must not be set when using the kubernetes provider")
+	}
+
+	// Verify that the Service Account token and CA certificate files are accessible.
+	_, err := os.ReadFile(localJWTPath)
+	if err != nil {
+		return fmt.Errorf("error reading service account token file: %w", err)
+	}
+
+	_, err = os.ReadFile(localCACertPath)
+	if err != nil {
+		return fmt.Errorf("error reading CA certificate file: %w", err)
+	}
+
+	return nil
+}
+
+func (k *KubernetesProvider) SensitiveKeys() []string {
+	return []string{}
+}
+
+func (k *KubernetesProvider) GetHTTPClient() *http.Client {
+	// The HTTP client is created only once by JWT/OIDC authentication method.
+	// Note: CA certificate rotation is not supported without restarting or re-creating backend instance.
+	certPool := x509.NewCertPool()
+	caCert, err := os.ReadFile(localCACertPath)
+	if err != nil {
+		return cleanhttp.DefaultPooledClient()
+	}
+
+	certPool.AppendCertsFromPEM([]byte(caCert))
+
+	tlsConfig := &tls.Config{
+		RootCAs: certPool,
+	}
+
+	baseTransport := cleanhttp.DefaultPooledTransport()
+	baseTransport.TLSClientConfig = tlsConfig
+
+	saAuthTransport := &bearerAuthRoundTripper{
+		baseTransport:      baseTransport,
+		kubernetesProvider: k,
+	}
+
+	return &http.Client{
+		Transport: saAuthTransport,
+	}
+}
+
+// bearerAuthRoundTripper is an http.RoundTripper that adds an Authorization header
+// containing the Kubernetes Service Account token as a bearer token.
+type bearerAuthRoundTripper struct {
+	baseTransport      http.RoundTripper
+	kubernetesProvider *KubernetesProvider
+}
+
+func (rt *bearerAuthRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if len(req.Header.Get("Authorization")) != 0 {
+		return rt.baseTransport.RoundTrip(req)
+	}
+
+	// Clone the request to avoid modifying the original.
+	req = req.Clone(req.Context())
+
+	// Read the token from disk on each request.
+	// Since discovery and JWKS downloads are infrequent, caching in memory has minimal benefit.
+	token, err := os.ReadFile(localJWTPath)
+	if err == nil {
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+	}
+
+	return rt.baseTransport.RoundTrip(req)
+}
--- a/builtin/credential/jwt/provider_kubernetes.go
+++ b/builtin/credential/jwt/provider_kubernetes.go
@@ -7,12 +7,17 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
+	"strings"
 
+	"github.com/hashicorp/cap/jwt"
 	"github.com/hashicorp/go-cleanhttp"
+	"golang.org/x/oauth2"
 )
 
 // Global variables instead of const to allow test cases to overwrite paths.
@@ -24,20 +29,27 @@ var (
 	localCACertPath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 )
 
-type KubernetesProvider struct{}
+type KubernetesProvider struct {
+	// oidcDiscoveryURL is the OIDC discovery URL for the Kubernetes API server.
+	oidcDiscoveryURL string
+}
 
 func (k *KubernetesProvider) Initialize(_ context.Context, jc *jwtConfig) error {
-	// Ensure JWKSCAPEM and OIDCDiscoveryCAPEM are not set. These options conflict with
-	// the Kubernetes provider because they configure a custom HTTP client via github.com/hashicorp/cap
-	// when calling jwt.NewJSONWebKeySet() or jwt.NewOIDCDiscoveryKeySet(), preventing
-	// the Kubernetes provider from injecting its own HTTP client with proper CA certificate
-	// and Authorization header.
-	if jc.JWKSCAPEM != "" {
-		return errors.New("jwks_ca_pem must not be set when using the kubernetes provider")
+	// Verify that no conflicting configuration is set.
+	if jc.OIDCDiscoveryURL != "" {
+		return errors.New("oidc_discovery_url must not be set when using the kubernetes provider")
 	}
 
 	if jc.OIDCDiscoveryCAPEM != "" {
 		return errors.New("oidc_discovery_ca_pem must not be set when using the kubernetes provider")
+	}
+
+	if jc.JWKSURL != "" {
+		return errors.New("jwks_url must not be set when using the kubernetes provider")
+	}
+
+	if jc.JWKSCAPEM != "" {
+		return errors.New("jwks_ca_pem must not be set when using the kubernetes provider")
 	}
 
 	// Verify that the Service Account token and CA certificate files are accessible.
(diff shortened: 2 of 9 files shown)


builtin/credential/jwt/backend.go
```diff
@@ -151,14 +151,19 @@ func (b *jwtAuthBackend) jwtValidator(config *jwtConfig) (*jwt.Validator, error)
...
+	pConfig, err := NewProviderConfig(b.providerCtx, config, ProviderMap())
+	if err != nil {
+		return nil, fmt.Errorf("error initializing provider_config: %w", err)
+	}
+
 	// Configure the key set for the validator
 	switch config.authType() {
 	case JWKS:
-		keySet, err = jwt.NewJSONWebKeySet(b.providerCtx, config.JWKSURL, config.JWKSCAPEM)
+		keySet, err = jwt.NewJSONWebKeySet(injectProviderHTTPClient(pConfig, b.providerCtx), config.JWKSURL, config.JWKSCAPEM)
```
cipherboy: nit: I'd probably split this into two lines, but eh 😆

builtin/credential/jwt/provider_kubernetes.go
```diff
@@ -0,0 +1,112 @@
...
+	"net/http"
+	"os"
+
+	"github.com/hashicorp/go-cleanhttp"
+)
+
+// Global variables instead of const to allow test cases to overwrite paths.
+var (
+	// localJWTPath is the path to the Kubernetes Service Account JWT token file.
+	localJWTPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
cipherboy, line 27: Wonder if we can pull this out of a k8s library somewhere? I suppose by definition, we know we're on Linux 😆

builtin/credential/jwt/provider_kubernetes.go
```diff
@@ -0,0 +1,112 @@
...
+	// Verify that the Service Account token and CA certificate files are accessible.
+	_, err := os.ReadFile(localJWTPath)
+	if err != nil {
+		return fmt.Errorf("error reading service account token file: %w", err)
+	}
+
+	_, err = os.ReadFile(localCACertPath)
+	if err != nil {
+		return fmt.Errorf("error reading CA certificate file: %w", err)
+	}
```
cipherboy, line 65: My issue is that we don't have ways for instance operators to disallow this behavior. 🤔 

OpenBao's biggest difference, I'd say, is that we're trying to be multi-tenant. Which admittedly is a bit of a long process, but means we probably need to (by default) restrict plugins from reading storage, or have mechanisms to do so.

E.g., if GitLab runs OpenBao on a k8s platform, (though, admittedly we don't allow directly mounting/configuring auth plugins anyways...), we want to deny users from pulling our k8s creds.

Long-term, that may require a plugin sandboxing strategy (running on separate containers/hosts, ...) but for a controlled/_managed_ setup, that should be easier.

---

I'm not really sure what a path to prohibit this would look like in this instance. Thoughts?

If we instead moved to a model where we supported additional parameters, you'd have to rotate the creds, but at least they'd be explicitly known by the user. Thoughts?

tsaarni: Ok, I see the issue. I wasn't aware of the plans for managed multitenant setup. Assuming auth plugin configuration would eventually be allowed per tenant, there's at least one other spot to consider: I implemented similar change a few years back into Kubernetes auth method [hashicorp/vault-plugin-auth-kubernetes#122](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/122).

Personally I don't think passing the token as an additional parameter is practical. It would be quite cumbersome, as the token is renewed on every pod start, and its renewal interval can be as low as 600 seconds. 

One alternative might be to require service account token path (and CA as well) to be configured out-of-band before mounting the auth method, for example via environment variables. In that case, it would be up to the deployment to allow the use of local files, while by default it would not work. Though I'm not sure we have any precedent for that kind of approach?

larssb:
Does this PR make it easier to onboard auth for OpenBao clients on Kubernetes clusters, that are **NOT** the local cluster? My intend being to use jwt auth when integrating with OpenBao to get a token that can then be used as part of injecting a Secret stored in OpenBao via the OpenBao provider to the secrets store csi driver.

cipherboy:
@larssb No, this is currently designed strictly for the local cluster.

It's an interesting thought, I wonder if there's a mutual solution here? 🤔 

@tsaarni Ultimately the ability to execute arbitrary GET requests with this container's credentials (from the jwks_url you get full path control, though cannot convert this to POST or set headers) is perhaps worrying even from a single-tenant service provider perspective.

builtin/credential/jwt/provider_kubernetes.go
DrDaveD: At minimum I think the paths should be configurable.

larssb:
Thank you for responding @cipherboy ... much appreciated.

So doing something like:

```bash
bao write auth/jwt-clusterA/config \
  oidc_discovery_url="https://EXTERNAL_OR_INTERNAL_API_ENDPOINT" \
  oidc_discovery_ca_pem=@clusterA-ca.crt
```

- path per cluster jwt integration and thereby an OpenBao conf.
- inject the crt to trust in some way. Specifically looking into injecting the ca.crt that OpenBao needs to trust for non-local Kubernetes clusters. This line `oidc_discovery_ca_pem=@/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. I'm to test out the viability of storing this cert in OpenBao and then pointing to this ca.crt internally.

Would such an approach be possible?

I hope I make sense @cipherboy 

N.B. I did look into the CNCF OpenBao Slack channel ... to see if posting questions like the above would be better over there. However, the channel looks very very inactive.

Thank you so much

cipherboy:
@larssb OpenBao is not actually under the CNCF and is under the OpenSSF. That slack channel was an experiment until the LF had a proper replacement for their unsupported matrix bridge with unlimited history.

Check out https://github.com/openbao/#contact for actual chat channels. I should go see if I can delete that room.

tsaarni:
When running inside a Kubernetes pod, the following approach could work:

If the new Kubernetes-specific JWT/OIDC provider introduced in this PR is selected:

* The provider would automatically use the Kubernetes API server URL from pod environment variables: `https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT` ([link](https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#directly-accessing-the-rest-api)).
  * For security reasons, the `oidc_discovery_url` and `jwks_url` parameters ([link](https://openbao.org/api-docs/auth/jwt/#parameters)) parameters should not be configurable. This avoids scenarios where a malicious administrator could redirect requests to a server they control to gain access to the service account credentials mounted in the OpenBao pod.
* The provider would use the default mount paths for the service account token and CA certificate: `/var/run/secrets/kubernetes.io/serviceaccount/token` `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, 
  * These paths should not be configurable, again for security reasons, to prevent administrators from having OpenBao read arbitrary files from the pod filesystem.
If configurability is required, it could be done through out-of-band mechanisms such as environment variables, to make sure control remains with the deployment operator, but I'm not sure if there is precedent for such configuration so far in any other case.

tsaarni:
> @tsaarni Ultimately the ability to execute arbitrary GET requests with this container's credentials (from the jwks_url you get full path control, though cannot convert this to POST or set headers) is perhaps worrying even from a single-tenant service provider perspective.

This scenario assumes the malicious administrator also controls the CA certificate. Otherwise, the HTTP client would fail to connect to the server, unless the admin somehow obtained a server certificate signed by the Kubernetes CA.  But it will be of course good to avoid this scenario too. 

I wanted to highlight this, because the current code in this PR has a limitation (noted in the comments) that prevents administrator from setting `jwks_ca_pem` and ` oidc_discovery_ca_pem`.

cipherboy:
> This scenario assumes the malicious administrator also controls the CA certificate. Otherwise, the HTTP client would fail to connect to the server, unless the admin somehow obtained a server certificate signed by the Kubernetes CA. But it will be of course good to avoid this scenario too.

@tsaarni Sorry, I'm confused. Don't we set the Kubernetes CA pem in this PR from disk? So connecting at least to Kubernetes URLs should succeed? 

https://github.com/openbao/openbao/pull/2114/files#diff-78646e947d647130834d6d1a3dc6fe711983383016e7a086a539d32260b69cd6R64-R70 

in particular.

Edit: I see, it'd be in conjunction with your proposal in https://github.com/openbao/openbao/pull/2114#issuecomment-3563368438, to hard-code k8s addresses when using the k8s provider. :thinking:

tsaarni:
> Edit: I see, it'd be in conjunction with your proposal in [#2114 (comment)](https://github.com/openbao/openbao/pull/2114#issuecomment-3563368438), to hard-code k8s addresses when using the k8s provider

With that we'd avoid the whole problem.

| So connecting at least to Kubernetes URLs should succeed?

It would successfully connect to the API server and any other servers signed by the Kubernetes CA. In general, I don’t think it would be trivial for an attacker to obtain a server certificate signed by the Kubernetes CA. They would need significant privileges in the Kubernetes API to request such a certificate. And if they already have that level of Kubernetes administrative access, they likely wouldn’t even need to steal the service account token anymore.

Of course, mileage may vary. For example, if someone configured that CA to be OpenBao itself ;)

tsaarni:
It might be clearer to explicitly describe the threat scenario, as it can get confusing otherwise.

An attacker who is an OpenBao administrator could do the following:

1. Set up a web server that logs incoming `Authorization` headers.
2. Configure the JWT/OIDC auth method with `oidc_discovery_url` pointing to `https://my-malicious-server.example.com`.
3. Wait for a user to initiate login.
4. Capture the OpenBao Service Account token from the server logs.
5. Use the captured token to access whatever privileges the token grants.

By controlling the discovery URL, the attacker can effectively obtain the token.

I haven't confirmed whether HTTPS is strictly required in this setup. If HTTPS is mandatory, the attacker would also need a valid TLS certificate for `https://my-malicious-server.example.com`, which comes back to https://github.com/openbao/openbao/pull/2114#issuecomment-3563496042. If TLS is not enforced, they could redirect the discovery request to any HTTP endpoint. However, that would likely violate [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-1) which states _"TLS is mandatory to implement and use with this specification"_.

I’d be in favor of (partially) hardcoding the Kubernetes API server URL using`https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT` https://github.com/openbao/openbao/pull/2114#issuecomment-3563368438. This is the officially recommended way to access the API server from within a pod, and it also frees administrators from having to manually determine and configure the address. It would improve the user experience when running OpenBao inside a Kubernetes pod.

cipherboy:
Thanks @tsaarni!

> It might be clearer to explicitly describe the threat scenario, as it can get confusing otherwise.
> 
> An attacker who is an OpenBao administrator could do the following:
> 
>     1. Set up a web server that logs incoming `Authorization` headers.
> 
>     2. Configure the JWT/OIDC auth method with `oidc_discovery_url` pointing to `https://my-malicious-server.example.com`.
> 
>     3. Wait for a user to initiate login.
> 
>     4. Capture the OpenBao Service Account token from the server logs.
> 
>     5. Use the captured token to access whatever privileges the token grants.
> 
> 
> By controlling the discovery URL, the attacker can effectively obtain the token.

Hmm:

https://github.com/Nordix/openbao/blob/e7e661ac055fdf82b122e267ee82cc1691f6053c/builtin/credential/jwt/path_config.go#L397-L410 

https://github.com/Nordix/openbao/blob/e7e661ac055fdf82b122e267ee82cc1691f6053c/builtin/credential/jwt/path_config.go#L310-L317

I _think_ a valid attack based on the code here also would be:

1. Attacker sets `jwks_url` and performs a `READ auth/jwt/config`.

This _might_ disclose enough of the GET response to be useful to the attacker. 

> I haven't confirmed whether HTTPS is strictly required in this setup. If HTTPS is mandatory, the attacker would also need a valid TLS certificate for `https://my-malicious-server.example.com`, which comes back to [#2114 (comment)](https://github.com/openbao/openbao/pull/2114#issuecomment-3563496042). If TLS is not enforced, they could redirect the discovery request to any HTTP endpoint. However, that would likely violate [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-1) which states _"TLS is mandatory to implement and use with this specification"_.

I believe it isn't, based on past experience. Or rather, [in GDK](https://gitlab.com/gitlab-org/gitlab-development-kit/-/blob/df493eca9dd4341943c7801c207033144223eab3/support/templates/openbao/config.hcl.erb#L73-75) we configure discovery URL with HTTP by default, so I think I can confidently say it is optional.

> I’d be in favor of (partially) hardcoding the Kubernetes API server URL using`https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT` [#2114 (comment)](https://github.com/openbao/openbao/pull/2114#issuecomment-3563368438). This is the officially recommended way to access the API server from within a pod, and it also frees administrators from having to manually determine and configure the address. It would improve the user experience when running OpenBao inside a Kubernetes pod.

Right, I think this is a safe option.

builtin/credential/jwt/provider_kubernetes.go
tsaarni: I don’t think so, even Kubernetes’ client-go has the default path hardcoded and not exposed publicly: https://github.com/kubernetes/client-go/blob/a296bd743c2dd954f56664f6ee7bc8558f7d8e32/rest/config.go#L544-L547

builtin/credential/jwt/provider_kubernetes.go
tsaarni: I’m not sure there’s a way to fix this if we want to avoid letting an OpenBao admin point at any file in pod and have it read as a token. The only other option I can think of is setting those paths during deployment instead of through the REST API, but I’m not aware of any precedent for that kind of configuration elsewhere in the code?

tsaarni:
Sorry for the delay, I got pulled into other tasks.

I’ve updated the PR according to our previous discussion. The discovery URL is no longer configurable and is now derived from the environment variables that Kubernetes sets in the pod (similar to how client-go [handles](https://github.com/kubernetes/client-go/blob/a296bd743c2dd954f56664f6ee7bc8558f7d8e32/rest/config.go#L548) it). The motivation is security: if the URL were configurable, a malicious OpenBao administrator could theoretically point OpenBao to a server they control and capture the pod’s service account token from OpenBao’s request.

While implementing this change, I realized I couldn’t use HashiCorp cap’s `jwt.NewOIDCDiscoveryKeySet()` for fetching the discovery document because it requires the `issuer` to match the hostname in the URL ([link](https://github.com/hashicorp/cap/blob/b72d477d7f92c08b2dadd2ce2a9c97e45f1ec94c/jwt/keyset.go#L96-L100)). In Kubernetes `KUBERNETES_SERVICE_HOST` environment variable might be e.g. `10.96.0.1` while discovery document might have issuer like `https://kubernetes.default.svc.cluster.local`.  

Rather than trying to send pull request to https://github.com/hashicorp/cap, I re-implemented the discovery document fetch logic without enforcing issuer validation. As far as I can tell, this does not introduce any security issues. 

To support this, I’m updating my proposed new provider interface. Previously, I suggested an interface that allowed providers to supply a custom `http.Client` (for injecting the `Authorization` header). Now, I’m proposing an interface that instead allows providers to implement their own discovery logic.

builtin/credential/jwt/provider_kubernetes.go
cipherboy: Let's table the discussion of multi-tenancy; @satoqz has put forward a proposal in https://github.com/openbao/openbao/pull/2207 which would likely have broad impacts on many other plugins. I'm happy with this continuing avoiding the conversation for now.

satoqz: @cipherboy note that https://github.com/openbao/openbao/pull/2207 is designed mainly with built-in constant mounts in mind (`/sys` and friends) as you can't yet write policies that apply to "all mounts of type x".

It would however enable you to generally block enabling `jwt` mounts and their configuration within a tenant's space, and delegate that to administrators in an upper namespace (perhaps to the root namespace adminstrators).

tsaarni:
> Rather than trying to send pull request to https://github.com/hashicorp/cap, I re-implemented the discovery document fetch logic without enforcing issuer validation. As far as I can tell, this does not introduce any security issues.

As discussed in yesterday’s community meeting: even if we don’t block and wait for an solution in `hashicorp/cap` for issuer validation, I’ve submitted a feature request [hashicorp/cap#176](https://www.github.com/hashicorp/cap/issues/176).

xref: community call [recording](https://zoom.us/rec/share/LBfxoCOAeQzHd8cHbhXsBscfMEy2ewV8wGHEKuOZNfU6H9ajzgauw2olKKmWvDfu.-fNiMsLBLvBxNot0) on Dec 18, 2025.

tsaarni:
Just checking in on this PR. Please let me know if you need any changes or if I can help with the review process by any other means.

DrDaveD:
@cipherboy do you have any remaining concerns?

tsaarni:
Made some minor improvements: clarified errors, used `net.JoinHostPort()` to handle potential IPv6 addresses, and reordered parameters to place `Context` first as recommended in the [docs](https://pkg.go.dev/context). 

EDIT: Also rebased, as the branch point was old enough that CI reported vulnerabilities.

tsaarni:
@cipherboy Friendly ping on this PR!  Let me know if any further changes are needed.

I rebased again to include https://github.com/openbao/openbao/pull/2426 to avoid govulncheck failure.

tsaarni:
Hi! Following up on this. It is ready for review whenever you have time.

JanMa:
Sorry for the delay @tsaarni , I will have a look at it this week

JanMa (approved):
LGTM, thanks a lot @tsaarni.
@openbao/openbao-k8s-committers please also take a look

cipherboy (approved):
I think this makes sense, thanks @tsaarni!

eyenx:
Thank you @tsaarni this is a good change!

tsaarni:
Thanks for the review and merge!

PR #2122, opened 2025-11-20
Simplify handling of Sealable Namespaces when Core is sealed

KrzysztofKornalewski-Reply:
## Main problem
`The SealManager (SM) and NamespaceStore (NS) have their own locks to control concurrent data access. These distributed locks (that are normally locked and released in each method) however can make it hard to track lock states in cross-component calls, and operation that requires mutiple calls to the namespace-seal subsystems.`

## Description

Refactors `Reset()` and `InitializeBarrier()` to eliminate lock ordering violations between NS and SM. It moves orchestration logic from SM into preSeal(), provides consistent lock ordering that NS lock acquired before SM lock throughout the codebase.
This is related with a previous PR that had a different approach to same problem.

This is a spirit continuation of work for this PR:
[https://github.com/openbao/openbao/pull/2034](https://github.com/openbao/openbao/pull/2034)
Same problem, different approach 

TO DO:

- [x] Refactor SealManager.Reset()
SealManager only handles its internal state reset. 
- [x]  Refactor SealManager.InitializeBarrier()
Remove NamespaceStore callbacks from InitializeBarrier(). The method should only handle seal and barrier initialization. Orchestration is moved to NamespaceStore.SetNamespace().

About SealManager.InitializeBarrier(), at this point, changes are not necessary IMO

## Rationale

The `Reset()` and `InitializeBarrier()` methods calls into NS methods, creating inconsistent lock ordering based on this agreement:
[https://github.com/openbao/openbao/pull/2034#issuecomment-3533205883](https://github.com/openbao/openbao/pull/2034#issuecomment-3533205883)

To enforce a consistent calling direction and thereby consistent lock ordering.

## Acknowledgements

       also allowed.
[At this point, changes are not necessary IMO.]

### DIFF
--- a/vault/seal_manager_test.go
+++ b/vault/seal_manager_test.go
@@ -4,7 +4,6 @@
 package vault
 
 import (
-	"context"
 	"testing"
 
 	"github.com/openbao/openbao/helper/namespace"


vault/core.go
```diff
@@ -2500,8 +2500,24 @@ func (c *Core) preSeal() error {
...
 	}
-	if err := c.sealManager.Reset(context.Background()); err != nil {
-		result = multierror.Append(result, fmt.Errorf("error reseting seal manager: %w", err))
+
+	// Lock ordering: NS lock acquired first, then SM lock acquired inside ResetInternal
+	unlock, err := c.namespaceStore.lockWithInvalidation(context.Background(), false)
+	if err != nil {
+		result = multierror.Append(result, fmt.Errorf("error during acquire namespace lock during preSeal: %w", err))
+	} else {
+		defer unlock()
```
wslabosz-reply: probably should be immediately unlocked after the `ns.sealNamespaceLocked` invocation, to make it more nicely, you could take both `ns.lockWithInvalidation` and `ns.sealNamespaceLocked`, and write a namespace store method running both in a single func

vault/seal_manager.go
```diff
@@ -94,27 +94,12 @@ func (sm *SealManager) setup() {
 	}
 }
 
-// Reset seals all namespaces (except root) and calls the
-// `(*SealManager) setupSealManager` to overwrite the internal
-// storage of a sealManger with a default values for root namespace.
-func (sm *SealManager) Reset(ctx context.Context) error {
-	if sm.core.namespaceStore == nil {
-		return nil
-	}
```
wslabosz-reply, line 103: I think you should retain this check

vault/seal_manager_test.go
```diff
@@ -0,0 +1,37 @@
...
+		t.Fatalf("ResetInternal failed: %v", err)
+	}
+
+	// Verify that SM state was reset and setup() reinitializes maps
+	if c.sealManager.barrierByNamespace == nil {
+		t.Fatal("expected barrierByNamespace to be initialized after reset")
+	}
+
+	// Verify that it's a new map
+	if c.sealManager.barrierByNamespace == originalBarriers {
```
wslabosz-reply: you'd probably want `reflect.DeepEqual()` for this

vault/namespace_store.go
```diff
@@ -315,12 +315,29 @@ func (ns *NamespaceStore) invalidate(ctx context.Context, path string) {
...
+		ns.logger.Error("set namespace failed", "error", err)
 		return nil, err
 	}
 
-	return ns.setNamespaceLocked(ctx, entry, sealConfig)
+	// Lock was released by setNamespaceLocked, so mark it
+	lockReleased = true
+
+	return result, nil
 }
```
satoqz, line 334: Not sure I understand why this is needed, `setNamespaceLocked` always unlocks even if it fails.

vault/namespace_store.go
```diff
@@ -459,20 +475,26 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
...
 	} else {
 		// While we don't have transactions to aid us, we still shouldn't
-		// leave partial namespaces lying around in memory.
-		defer cleanupFailed()
+		// leave partial namespaces lying around in memory on error.
+		defer func() {
+			if err != nil {
+				cleanupFailed()
+			}
+		}()
```
satoqz: I don't understand these changes either. We already track failure via `failed` (which is checked by `cleanupFailed()`). Neither will checking `err` catch all errors as we don't always assign the returned error to `err`.

vault/namespace_store.go
KrzysztofKornalewski-Reply: During refactoring I faced panic during test:

```
 Error: -24T14:38:17.089Z [ERROR] TestACL_Namespace_Root/root-ns.namespace: failed to rollback transaction: error="transaction has been committed or rolled back"
  fatal error: sync: Unlock of unlocked RWMutex
  
  goroutine 43 [running]:
  sync.fatal({0x44e105f?, 0xc0006e0480?})
  	/opt/hostedtoolcache/go/1.25.3/x64/src/runtime/panic.go:1026 +0x18
  sync.(*RWMutex).Unlock(0xc0006e0498)
```

So it's possible that it can try to unlock the unlocked.
That's why I put it here 
But yeah maybe if I re-structure the code differently, it might not need this flag

satoqz: @KrzysztofKornalewski-Reply I think you are causing that panic in this function:

```go
func (ns *NamespaceStore) SealAllNamespacesLocked(ctx context.Context) error {
	unlock, err := ns.lockWithInvalidation(ctx, false)
	if err != nil {
		return err
	}
	defer unlock()

	return ns.sealNamespaceLocked(ctx, namespace.RootNamespace)
}
```

This will always double-unlock.

vault/namespace_store.go
KrzysztofKornalewski-Reply: Yeah, you are right it should be like:

```
defer func() {
			if failed {
				if rollbackErr := txn.Rollback(ctx); rollbackErr != nil {
					ns.logger.Error("failed to rollback transaction", "error", rollbackErr)
				}
				cleanupFailed()
			}
		}()
```

satoqz: @KrzysztofKornalewski-Reply `cleanupFailed()` already checks `failed`, and `Rollback()` is okay to call at all times. So I think it's fine as-is on the `namespaces-seal` branch.

vault/namespace_store.go
KrzysztofKornalewski-Reply: Hmm, this `defer unlock()` should be removed from this func and keep `_, err := ns.lockWithInvalidation(ctx, false)` but I guess lockReleased flag is still useful for a lock tracking in SetNamespace.

satoqz: @KrzysztofKornalewski-Reply I think `SealNamespace` behaves the same way, all you need is `_, err := ns.lockWithInvalidation(ctx, false)` in either case.

vault/namespace_store.go
```diff
@@ -1405,3 +1415,11 @@ func (c *Core) NamespaceByStoragePath(ctx context.Context, path string) (*namesp
 
 	return ns, rest, nil
 }
+
+func (ns *NamespaceStore) SealAllNamespacesLocked(ctx context.Context) error {
+	_, err := ns.lockWithInvalidation(ctx, false)
+	if err != nil {
+		return err
+	}
+	return ns.sealNamespaceLocked(ctx, namespace.RootNamespace)
```
satoqz:
```suggestion
	if _, err := ns.lockWithInvalidation(ctx, false); err != nil {
		return err
	}
	return ns.sealNamespaceLocked(ctx, namespace.RootNamespace)
```

satoqz:
@KrzysztofKornalewski-Reply can you revert the unneeded changes to `setNamespaceLocked` and `SetNamespace` we've discussed above? I see no reason to change anything in those functions.

KrzysztofKornalewski-Reply:
> @KrzysztofKornalewski-Reply can you revert the unneeded changes to `setNamespaceLocked` and `SetNamespace` we've discussed above? I see no reason to change anything in those functions.

Done

vault/namespace_store.go
```diff
@@ -1405,3 +1404,10 @@ func (c *Core) NamespaceByStoragePath(ctx context.Context, path string) (*namesp
 
 	return ns, rest, nil
 }
+
+func (ns *NamespaceStore) SealAllNamespacesLocked(ctx context.Context) error {
```
satoqz:
```suggestion
func (ns *NamespaceStore) SealAllNamespaces(ctx context.Context) error {
```

vault/seal_manager.go
```diff
@@ -94,27 +94,15 @@ func (sm *SealManager) setup() {
 	}
 }
 
-// Reset seals all namespaces (except root) and calls the
-// `(*SealManager) setupSealManager` to overwrite the internal
-// storage of a sealManger with a default values for root namespace.
-func (sm *SealManager) Reset(ctx context.Context) error {
+// Lock ordering: This method acquires SM lock. The caller (Core.preSeal)
+// should already hold NS lock, maintaining NS -> SM lock ordering.
+func (sm *SealManager) ResetInternal(ctx context.Context) error {
```
satoqz: No need to call this `ResetInternal`, and no need to document the lock ordering as this is fully independent of the namespace store anyways. So please revert this, too.


```suggestion
// Reset clears all internal state, leaving only the root namespace's seal.
func (sm *SealManager) Reset(ctx context.Context) error {
```

vault/seal_manager.go
```diff
@@ -94,27 +94,15 @@ func (sm *SealManager) setup() {
...
 	if sm.core.namespaceStore == nil {
 		return nil
 	}
```
satoqz: Not needed

```suggestion
```

vault/seal_manager.go
```diff
@@ -94,27 +94,15 @@ func (sm *SealManager) setup() {
...
-	// providing root namespace to the function, but it will be omitted from actual sealing
-	if err := sm.core.namespaceStore.sealNamespaceLocked(ctx, namespace.RootNamespace); err != nil {
-		return err
-	}
-
 	sm.lock.Lock()
 	defer sm.lock.Unlock()
+
 	sm.setup()
 	return nil
```
satoqz: This has no reason to return an error anymore.
```suggestion
```

vault/seal_manager_test.go
```diff
@@ -0,0 +1,42 @@
...
+	// Verify that SM state exists before reset
+	if c.sealManager.barrierByNamespace == nil {
+		t.Fatal("expected barrierByNamespace to be initialized")
+	}
+
+	// Store reference to original barrier map
+	originalBarriers := c.sealManager.barrierByNamespace
+
+	// Call Reset
+	c.sealManager.Reset()
```
satoqz: I think this test is currently a bit too trivial to be worthwhile (Just `Reset()`), what if we test that seal manager fields are reset by actually sealing/unsealing core and creating some sealable namespaces inbetween?

KrzysztofKornalewski-Reply: You are right, thx for pointing it out. I'll improve this test!

vault/seal_manager_test.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/openbao/openbao/helper/namespace"
+)
+
+func TestSealManager_Resets(t *testing.T) {
+	t.Helper()
+
```
satoqz: Remove, this is not a test helper.

vault/seal_manager_test.go
```diff
@@ -0,0 +1,112 @@
...
+	nsSealCfg := &SealConfig{
+		Type:            "shamir",
+		SecretShares:    1,
+		SecretThreshold: 1,
+		StoredShares:    1,
+	}
+
+	nsA, keySharesA := createSealableNamespace(t, c, ctx, "a/", nsSealCfg)
+	if nsA == nil || len(keySharesA) == 0 {
+		t.Fatal("failed to create namespace a/")
```
satoqz: It would be nice if you could use the `testify/require` package.

vault/seal_manager_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	nsB, keySharesB := createSealableNamespace(t, c, ctx, "b/", nsSealCfg)
```
satoqz: You can use `NamespaceStore.SetNamespace` or `TestCoreCreateSealedNamespaces`, no need to define your own helper function.

vault/seal_manager_test.go
```diff
@@ -0,0 +1,112 @@
...
+	preBarriers := c.sealManager.barrierByNamespace.Len()
+	c.sealManager.lock.RUnlock()
+
+	if preSeals < 3 {
+		t.Fatalf("expected 3 namespaces (root, a and b) before reset, got %d", preSeals)
+	}
+
+	// Call Reset() with write lock held
+	c.sealManager.lock.Lock()
+	c.sealManager.Reset()
```
satoqz: I was thinking that instead of calling `Reset()` directly, you could seal and unseal the actual test core using `Core.Seal` and `TestCoreUnseal` (see `namespace_store_test.go` for similar tests).

vault/seal_manager_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+	"testing"
+
+	"github.com/openbao/openbao/helper/namespace"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSealManager_Resets(t *testing.T) {
+	c, _, rootToken := TestCoreUnsealed(t)
+	ctx := namespace.ContextWithNamespace(context.Background(), namespace.RootNamespace)
```
satoqz:
```suggestion
	ctx := namespace.RootContext(t.Context())
```

vault/core.go
```diff
@@ -2500,8 +2500,8 @@ func (c *Core) preSeal() error {
 	if err := c.stopExpiration(); err != nil {
 		result = multierror.Append(result, fmt.Errorf("error stopping expiration: %w", err))
 	}
-	if err := c.sealManager.Reset(context.Background()); err != nil {
-		result = multierror.Append(result, fmt.Errorf("error reseting seal manager: %w", err))
+	if err := c.sealAllNamespaces(context.Background()); err != nil {
```
cipherboy, line 2503: @KrzysztofKornalewski-Reply @satoqz @wslabosz-reply I think this goes to my discussion with Wojciech yesterday on the horizontal scalability WG (and before on Zulip), but do we want to seal all namespaces at `preSeal(...)` or should this be called in place of `c.barrier.Seal()`? IMHO it needs to be the latter to allow proper standby->active changeover, otherwise we lose all Shamir's keys. 

Regardless, not really relevant to this PR, as it is not the focus of it.

cipherboy:
Thanks @KrzysztofKornalewski-Reply!

vault/core.go
satoqz: I lack some context as I've only been tangentially involved in the discussion but this makes sense to me. I suppose keeping SealManager around outside of `preSeal` has also set us up to do that rather easily.

PR #2126, opened 2025-11-21
Add forwarded TLS certificate support

### LINKED ISSUE #1794: Add support for forwarded Tls-Client-Cert

cipherboy:
## Description

See information in #2080.

I've picked up the PR, squashed the original commits, and applied a bunch of review feedback. 

## Rationale

Resolves: #1794

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/credential/cert/backend_test.go
+++ b/builtin/credential/cert/backend_test.go
@@ -570,6 +570,17 @@ path "kv/ext/{{identity.entity.aliases.%s.metadata.2-1-1-1}}" {
 }
 
 func TestBackend_NonCAExpiry(t *testing.T) {
+	testbackendNoncaexpiryFunc(t, false)
+}
+
+func TestBackend_NonCAExpiryHeader(t *testing.T) {
+	testbackendNoncaexpiryFunc(t, true)
+}
+
+func testbackendNoncaexpiryFunc(t *testing.T, useHeader bool) {
+	var resp *logical.Response
+	var err error
+
 	// Create CA and issue a short-lived leaf certificate.
 	shortExpiry := 3 * time.Second
 	ca := &certyaml.Certificate{
@@ -605,7 +616,7 @@ func TestBackend_NonCAExpiry(t *testing.T) {
 		Data:      certData,
 	}
 
-	resp, err := b.HandleRequest(t.Context(), certReq)
+	resp, err = b.HandleRequest(t.Context(), certReq)
 	if err != nil || (resp != nil && resp.IsError()) {
 		t.Fatalf("err:%v resp:%#v", err, resp)
 	}
@@ -625,6 +636,12 @@ func TestBackend_NonCAExpiry(t *testing.T) {
 		},
 	}
 
+	// If we are testing header handling yank the peer certificate and put it in the ClientHeaderCert
+	if useHeader {
+		loginReq.Connection.ConnState = nil
+		loginReq.ClientHeaderCert = connState.PeerCertificates[0]
+	}
+
 	// Login when the certificate is still valid. Login should succeed.
 	resp, err = b.HandleRequest(t.Context(), loginReq)
 	if err != nil || (resp != nil && resp.IsError()) {
@@ -642,8 +659,15 @@ func TestBackend_NonCAExpiry(t *testing.T) {
 }
 
 func TestBackend_RegisteredNonCA_CRL(t *testing.T) {
-	tc := setupTestCerts(t)
+	testbackendRegisteredNonCACRLFunc(t, false)
+}
 
+func TestBackend_RegisteredNonCA_CRLHeader(t *testing.T) {
+	testbackendRegisteredNonCACRLFunc(t, true)
+}
+
+func testbackendRegisteredNonCACRLFunc(t *testing.T, useHeader bool) {
+	tc := setupTestCerts(t)
 	config := logical.TestBackendConfig()
 	storage := &logical.InmemStorage{}
 	config.StorageView = storage
@@ -685,6 +709,11 @@ func TestBackend_RegisteredNonCA_CRL(t *testing.T) {
 		Connection: &logical.Connection{
 			ConnState: &connState,
 		},
+	}
+	// If we are testing header handling yank the peer certificate and put it in the ClientHeaderCert
+	if useHeader {
+		loginReq.Connection.ConnState = nil
+		loginReq.ClientHeaderCert = connState.PeerCertificates[0]
 	}
 	// Login should succeed.
 	resp, err = b.HandleRequest(t.Context(), loginReq)
@@ -739,8 +768,15 @@ func TestBackend_RegisteredNonCA_CRL(t *testing.T) {
 }
 
 func TestBackend_CRLs(t *testing.T) {
-	tc := setupTestCerts(t)
+	testbackendCrlsfunc(t, false)
+}
 
+func TestBackend_CRLsHeader(t *testing.T) {
+	testbackendCrlsfunc(t, true)
+}
+
+func testbackendCrlsfunc(t *testing.T, useHeader bool) {
+	tc := setupTestCerts(t)
 	config := logical.TestBackendConfig()
 	storage := &logical.InmemStorage{}
 	config.StorageView = storage
@@ -785,6 +821,14 @@ func TestBackend_CRLs(t *testing.T) {
 		},
 	}
 	resp, err = b.HandleRequest(t.Context(), loginReq)
+
+	// If we are testing header handling yank the peer certificate and put it in the ClientHeaderCert
+	if useHeader {
+		loginReq.Connection.ConnState = nil
+		loginReq.ClientHeaderCert = connState.PeerCertificates[0]
+	}
+
+	resp, err = b.HandleRequest(t.Context(), loginReq)
 	if err != nil || (resp != nil && resp.IsError()) {
 		t.Fatalf("err:%v resp:%#v", err, resp)
 	}
@@ -796,6 +840,12 @@ func TestBackend_CRLs(t *testing.T) {
 		t.Fatalf("error testing connection state: %v", err)
 	}
 	loginReq.Connection.ConnState = &connState
+
+	// If we are testing header handling yank the peer certificate and put it in the ClientHeaderCert
+	if useHeader {
+		loginReq.Connection.ConnState = nil
+		loginReq.ClientHeaderCert = connState.PeerCertificates[0]
+	}
 
 	// Attempt login with the updated connection
 	resp, err = b.HandleRequest(t.Context(), loginReq)
@@ -847,6 +897,11 @@ func TestBackend_CRLs(t *testing.T) {
 		t.Fatalf("error testing connection state: %v", err)
 	}
 	loginReq.Connection.ConnState = &connState
+	// If we are testing header handling yank the peer certificate and put it in the ClientHeaderCert
+	if useHeader {
+		loginReq.Connection.ConnState = nil
+		loginReq.ClientHeaderCert = connState.PeerCertificates[0]
+	}
 
 	// Attempt login with the updated connection
 	resp, err = b.HandleRequest(t.Context(), loginReq)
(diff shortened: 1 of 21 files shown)


cipherboy:
@openbao/openbao-core-committers I'm happy pushing this to v2.6.0 if that's preferable, feel free to adjust the milestone as appropriate.

website/content/docs/configuration/listener/shared.mdx
```diff
@@ -0,0 +1,88 @@
...
+  List of processors to handle `x_forwarded_for_client_cert_header`
+  These are processed in the order of the array.
+
+  Available values are
+  - URL - Urldecode the supplied value
+  - PEM - Convert a PEM certificate to base64 encoded DER
+  - RFC9440 - Use base64 encoded DER as defined in [RFC 9440 2.1][RFC 9440 2.1]
+
+  The result of this processing should end up with a base64 encoded DER certificate.
+  If the header value is already base64 encoded DER then no value needs to be set.
```
JanMa, line 38: RFC9440 is a great standard for forwarding client certificate information, but the reality is, that it is not used a lot (yet). In Kubernetes based environments that use a service mesh based on Envoy proxy, the [X-Forwarded-Client-Certificate](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert) header is what's being used.

If we intend this feature to be usable to authenticate at OpenBao using the inside a service mesh (wich we should IMO), then we must add support for the XFCC header as well.

A good library for parsing the XFCC header, which is also used by Istio Internally is:
https://github.com/alecholmes/xfccparser

JanMa (changes requested):
Not done with reviewing the code, but this is definitely a blocker IMO

website/content/docs/configuration/listener/shared.mdx
cipherboy: Hey @JanMa , apologies, dropped the ball here.

As long as `cert` is sent in the key field, it should be feasible to parse. We can support that in this implementation by adding an Envoy decoder (or similar named) to extract just the cert field from the XFCC format. If `cert` is not sent, we'll have to reject the request as the cert auth method cannot support non-full-cert authentication. 

In the interest of keeping this moving could we continue on without it and add it as a follow-up PR? I don't see anything stopping us from implementing it, with or without the library.

cipherboy: @JanMa Bumping this if I may? :-D

tsaarni: I can contribute the Envoy support. I anyways work with envoy a lot :)

JanMa: Fine by me, let's add this in a follow-up PR :+1:

JanMa (approved):
LGTM

cipherboy:
@JanMa @satoqz This should now be rebased on top of @tsaarni's cert generation helpers and @wslabosz-reply's `t.Context()` refactor.

cipherboy:
(fixed the missing PEM file in the http test)

cipherboy:
@JanMa @satoqz Just for my clarity, did someone want to re-review prior to merge or are we good here? :-D Not trying to rush, just trying to understand if I should click the button or not. :laughing:

satoqz:
I'll hand in a review next week, sorry for not being clear. This was at the bottom of my long review stack :D

website/content/docs/configuration/listener/shared.mdx
satoqz, line 1: There are other shared parameters such as the `telemetry` block. It would be good to (at least eventually) consolidate that into this page to avoid confusion.

command/server.go
```diff
@@ -2935,8 +2935,8 @@ func startHttpServers(c *ServerCommand, core *vault.Core, config *server.Config,
 			RecoveryMode:          c.flagRecovery,
 		})
 
-		if len(ln.Config.XForwardedForAuthorizedAddrs) > 0 {
-			handler = vaulthttp.WrapForwardedForHandler(handler, ln.Config)
+		if ln.Config != nil {
+			handler = vaulthttp.WrapHttpServerHandler(handler, ln.Config)
 		}
```
satoqz, line 2941: Why doesn't the above call to `Handler` already do this?

http/handler.go
```diff
@@ -559,6 +606,105 @@ func WrapForwardedForHandler(h http.Handler, l *configutil.Listener) http.Handle
...
+		if len(clientCertificateHeaderName) == 0 {
+			// Nothing to do; listener configuration does not set the
+			// x_forwarded_for_client_cert_header option so we should
+			// continue on.
+			//
+			// Remove the standardized client cert headers for safety,
+			// even though we don't explicitly process them; they'd be
+			// untrusted.
+			r.Header.Del(RFC9440ClientCertHeader)
+			r.Header.Del(RFC9440ClientChainHeader)
```
satoqz, line 632: Should this be deferred unconditionally?

satoqz (approved):
Looks reasonable, some small notes.

command/server.go
cipherboy: @satoqz That's an interesting question.

I think it is because `runRecoveryMode(...)` doesn't need these: we're not _actually_ doing any real authentication (other than recovery token creation), and thus don't need to handle request forwarding. Similarly, for `SetClusterHandler` (for request forwarding) it isn't necessary to have these as the accepting node would already have resolved and modified the `X-Forwarded-For` headers on the original copy of the request before performing the forward on the modified copy. 

Thoughts?

cipherboy: @satoqz I do wonder if `wrapClientCertificateHandler` should be applied conditionally in this too or if it should be applied globally. I kinda feel conditionally is the best preference tbh.

http/handler.go
cipherboy: @satoqz I don't really have opinions here I don't think. In the event the `X-Forwarded-For` handler is bound+called, we already prune this and other headers if the request was not forwarded.

In the event that a non-standard name certificate header is used, we don't delete the standard ones, but we also attach the actual certificates in the correct place so we don't need to care. Similarly the headers are hidden from plugins by default, so... maybe? Thoughts?

builtin/credential/cert/backend_test.go
```diff
@@ -2219,6 +2226,18 @@ func TestBackend_RegressionDifferentTrustedLeaf(t *testing.T) {
 		t.Fatal("expected a successful authentication")
 	}
 
+	// Putting an incorrect leaf into the Processed headers should be ignored.
+	//
+	// This would fail if we actually used our tampered certificate.
+	newAClient.AddHeader("X-Processed-Tls-Client-Certificate-Resp", url.QueryEscape(base64.StdEncoding.EncodeToString(certBTampered)))
```
tsaarni, line 2233: Should this be `X-Processed-Tls-Client-Certificate` instead of `X-Processed-Tls-Client-Certificate-Resp`?
```suggestion
	newAClient.AddHeader(vaulthttp.ProcessedForwardedClientCertHeader, url.QueryEscape(base64.StdEncoding.EncodeToString(certBTampered)))
```

helper/configutil/listener.go
```diff
@@ -119,6 +119,16 @@ type Listener struct {
 	XForwardedForRejectNotAuthorized    bool                          `hcl:"-"`
 	XForwardedForRejectNotAuthorizedRaw interface{}                   `hcl:"x_forwarded_for_reject_not_authorized,alias:XForwardedForRejectNotAuthorized"`
 
+	// This is different from upstream by using array of strings to allow easier ordering of processing.
+	// Internally all certs are passed as headers that are base64 encoded DER.
+	// Available decoders: PEM, URL, RFC9440
+	XForwardedForClientCertHeader              string      `hcl:"x_forwarded_for_client_cert_header,alias:XForwardedForClientCertHeader"`
+	XForwardedForClientCertDecoders            []string    `hcl:"x_forwarded_for_client_cert_decoders,alias:XForwardedForClientCertificateProcessing"`
```
tsaarni, line 126: The `XForwardedForClientCertDecoders` configuration lacks validation, allowing invalid values to pass:

```
listener "tcp" {
  x_forwarded_for_client_cert_decoders = ["FOO"]
}
```

Additionally, `http/handler.go:wrapClientCertificateHandler()`  lacks a `default` branch for handling invalid decoder names, so requests that fail due to an invalid config return an unclear HTTP error 

```http
HTTP/1.1 400 Bad Request

{
    "errors": [
        "error decoding client certificate header as base64"
    ]
}
```

Though of course it is better to catch the issue early at startup by having validation in  `helper/configutil/listener.go:ParseListeners()`, not in `wrapClientCertificateHandler()`  .

http/handler.go
tsaarni: I guess same applies to this branch as well https://github.com/cipherboy/openbao/blob/8d1da829713e552aefe9673e0ada2744e96a75d3/http/handler.go#L640-L643. But to be honest, I do not understand why RFC 9440 has specific handling? Is it because it uses two separate headers, and removing `Client-Chain` would not happen automatically when `Client-Cert` is set in the configuration?

BTW regarding plugins: I don't have full picture, but I wanted to note that plugins still seem to receive peer certificates from `tls.ConnectionState`:

https://github.com/cipherboy/openbao/blob/8d1da829713e552aefe9673e0ada2744e96a75d3/sdk/plugin/pb/translation.go#L683

I guess that could cause some confusion when the plugins rely on that field while forwarding client certs is enabled?

cipherboy: @tsaarni @satoqz Ah, I think I understand better now.

We strip headers:

1. When they're not validated; if no `X-Forwarded-For` was set, we shouldn't trust a forwarded certificate. Similarly, if we don't have an explicit name of a header to take a forwarded cert from, we shouldn't ever use forwarded certs. That ends up in a bunch of different headers, but the only standardized one is RFC so we encourage RFC compliance. The RFC doesn't have specific handling; if we have a named header, we remove that too when it gets time for it to be passed to the OpenBao internals (see line 696 below).
2. We don't care about error cases, as nothing makes it to the server.  So we could defer it, but we remove these headers in both code paths through this function as it is :shrug: 

In your case that you point out,  @tsaarni, we don't have any values, so removing nothing doesn't really matter as much. I'm more worried about an untrusted value being sent along. 

---

Good find about external plugin support @tsaarni -- I don't think we currently/previously supported running cert auth externally but there's no reason why we shouldn't with this change.

builtin/credential/cert/backend_test.go
cipherboy: All should be ignored.

helper/configutil/listener.go
cipherboy: Revised, thanks @tsaarni!

builtin/credential/cert/backend_test.go
tsaarni: It was just bit confusing to me since the the production code never processes the `X-Processed-Tls-Client-Certificate-Resp` header.  It seemed like any header name, such as `x-test-that-nobody-ever-reads-cert-from-here` would have worked the same way. :) However, the server does read `X-Processed-Tls-Client-Certificate` so testing that header made sense to me.

command/server.go
satoqz: `http.handler()` knows about `props.RecoveryMode`, no? Can't we apply both conditionally that way?

cipherboy: @satoqz It does, but it doesn't know about `SetClusterHandler` / forwarded requests.

satoqz: Ahh gotcha, I understand now. Then this is best perhaps.

tsaarni (approved):
Looks good to me!

sdk/plugin/pb/backend.proto
```diff
@@ -705,6 +705,13 @@ message Connection {
 	// ConnectionState is the marshalled tls.ConnectionState from the original
 	// request
 	ConnectionState connection_state = 2;
+
+    // ProxiedCertificate is a certificate verified by a proxy header.
+    repeated string proxied_certificates_raw = 4;
+
+    // PeerCertificate is a certificate verified by this server through a TLS
+    // connection.
+    repeated string peer_certificates_raw = 5;
```
satoqz: I think this can just use `repeated bytes` and spare us the b64 roundtrip?

cipherboy: Ah, d'oh. I use proto so infrequently.

cipherboy: @satoqz Fixed!

satoqz (approved):

cipherboy:
@scj643 Many thanks for your contribution and patience while we get this landed!! :rocket: 

Thanks @JanMa, @satoqz, and @tsaarni for your feedback. :-)

Valgueiro:
@cipherboy Is there any timeline on releasing this change? I mean even a 2.6.0-beta version.

I couldnt find a release date for 2.6.0 anywhere, sorry if this isnt the correct place to ask this!

PR #2127, opened 2025-11-21
Fix read scalability test flakes

cipherboy:
## Description

This fixes two test flakes:

1. One where an eventually was required so that we could ensure token revocation had propagated fully.
2. One where mount table locking ordering meant that we could have inconsistencies when applying non-transactional mount table invalidations. 

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -685,7 +685,14 @@ func (c *Core) mountInternal(ctx context.Context, entry *MountEntry, updateStora
 			locked = false
 		}
 	}
-	defer unlock()
+
+	return c.mountInternalWithLock(ctx, entry, updateStorage, unlock)
+}
+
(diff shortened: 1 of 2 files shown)


vault/mount.go
```diff
@@ -685,7 +685,14 @@ func (c *Core) mountInternal(ctx context.Context, entry *MountEntry, updateStora
 			locked = false
 		}
 	}
-	defer unlock()
+
+	return c.mountInternalWithLock(ctx, entry, updateStorage, unlock)
+}
+
+func (c *Core) mountInternalWithLock(ctx context.Context, entry *MountEntry, updateStorage bool, unlock func()) error {
```
cipherboy, line 692: @phil9909 @wslabosz-reply Not sure that I like this as a potential fix, plus it causes other failures.

Thoughts?

cipherboy: I think we'll pull this and let #2040 merge instead, for GA.

PR #2132, opened 2025-11-23
Auto register plugins

cipherboy:
## Description

This introduces a second form to the plugin configuration file, using
directly downloaded plugins but specified with a command to execute,
which is still a base name of an executable in the plugin_directory
path.
    
We also introduce two new global configuration options,
plugin_auto_download, whether or not plugins are automatically
downloaded, and plugin_auto_register, for whether or not plugins are
automatically registered into the catalog. Both trigger on server
restart and/or SIGHUP. `bao plugin init` is not affected by the value
of plugin_auto_download and will always download all defined plugins.
    
In this way, the configuration file can be used to fully manage the
lifecycle of plugins from downloading, to registration, to removal.
Note that OpenBao will not automatically remove stale plugins; this
is on the operator to handle.
    
Like audit devices, plugins manually registered via the API cannot be
shadowed in the configuration file. This means that if OCI plugins are
registered manually, plugin_auto_register cannot be set to true.
    
Future enhancements could be removing previously declared (but not
currently registered or declared) plugins and disabling API-driven
plugin registration.

## Rationale

Related: #1476
See also: #1824

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/plugin_catalog.go
+++ b/vault/plugin_catalog.go
@@ -260,7 +260,7 @@ func (c *Core) registerOCIPlugins(ctx context.Context, logger log.Logger, config
 		return nil
 	}
 
-	// Create OCI plugin downloader
+	// Register declarative plugins
 	logger.Info("starting OCI plugin registration")
 	defer logger.Info("OCI plugin registration completed")
 


command/plugin_init.go
```diff
@@ -40,7 +40,12 @@ func (c *PluginInitCommand) Help() string {
 Usage: bao plugin init [options]
 
   This command reads the plugin configuration from the server config files,
-  downloads the specified OCI images and extracts the contained plugin binaries.
+  downloads the specified OCI images, and extracts the contained plugin
+  binaries. This command does not automatically register the plugin to the
+  server, which is handled automatically via the server on startup and SIGHUP
+  if plugin_auto_register=true is set in the configuration file. When the
```
JanMa:
```suggestion
  if `plugin_auto_register=true` is set in the configuration file. When the
```

command/plugin_init.go
```diff
@@ -40,7 +40,12 @@ func (c *PluginInitCommand) Help() string {
...
+  server's configuration includes plugin_auto_download=true, plugins will be
```
JanMa:
```suggestion
  server's configuration includes `plugin_auto_download=true`, plugins will be
```

vault/audit.go
```diff
@@ -658,7 +658,7 @@ func (c *Core) ReloadAuditLogs() {
 	c.stateLock.RLock()
 	defer c.stateLock.RUnlock()
 
-	if c.Sealed() {
+	if c.Sealed() || (c.standby.Load() && !c.StandbyReadsEnabled()) || c.activeContext == nil {
```
JanMa, line 661: How is this change relevant to this PR?

command/server/config.go
JanMa: @cipherboy you need to allow this to be empty I think

JanMa (approved):
Some small remarks, otherwise LGTM

vault/audit.go
cipherboy: @JanMa I was re-evaluating SIGHUP reload strategy when I was adding support for it here, and realized the conditions here in audit was incorrect. I can change it if you want.

command/server/config.go
cipherboy: Great catch, thanks @JanMa!

cipherboy:
@JanMa Fixed and pushed, thanks! LMK what you think on the last item, but I'll also give a chance for @openbao/openbao-core-committers to review tomorrow morning.

command/plugin_init.go
cipherboy: @JanMa Changed to single quotes, they're in a backtick block already.

vault/plugin_catalog.go
```diff
@@ -207,18 +225,48 @@ func (c *Core) reconcileOCIPlugins(ctx context.Context) error {
...
 	return nil
 }
 
+func (c *Core) registerOCIPlugins(ctx context.Context, logger log.Logger, config *server.Config, standby bool) error {
+	if !config.PluginAutoRegister {
+		logger.Info("skipping OCI plugin registration")
+		return nil
+	}
+
+	// Create OCI plugin downloader
```
wslabosz-reply: wrong comment?

website/content/docs/configuration/plugins.mdx
```diff
@@ -44,21 +52,41 @@ plugin "type" "name" {
...
+```hcl
+plugin "type" "name" {
+  command     = "plugin"
+  version     = "v0.0.0"
+  binary_name = "plugin-binary-name"
+  sha256sum   = "sha256-checksum"
+  env         = ["MY_ENV=value"]
+  args        = ["-some-argument"]
+}
+
```
satoqz, line 66:
```suggestion
```

command/server/config.go
```diff
@@ -983,6 +1022,22 @@ func ParseConfig(d, source string) (*Config, error) {
...
+		result.PluginAutoDownload = autoDownload
+	}
+
+	if result.PluginAutoRegisterRaw != nil {
+		autoRegister, err := parseutil.ParseBool(result.PluginAutoRegisterRaw)
+		if err != nil {
+			return nil, err
+		}
+		result.PluginAutoRegister = autoRegister
+	}
```
satoqz, line 1039: Any reason to manually parse these instead of directly via `hcl` tag?

satoqz (approved):

command/server/config.go
cipherboy: @satoqz This happens everywhere in OpenBao, the `parseutil.ParseBool` supports many more boolean variants than `hcl` does natively.

vault/plugin_catalog.go
cipherboy:
```suggestion
	// Register declarative plugins
```

website/content/docs/configuration/plugins.mdx
cipherboy: GitHub doesn't let me apply this one, so I'll pick it up with changes that #2141 had.

PR #2143, opened 2025-11-24
Correctly lowercase policies in identity groups

cipherboy:
## Description

Confusingly, while strutil.RemoveDuplicates and
strutil.RemoveDuplicatesStable both take a second parameter to perform case-insensitive comparison, the latter (as originally used by the group subsystem) returned the original input strings (performing comparisons that were case-insensitive) while the former returns normalized (lowercase) forms. This oversight meant that initial analysis of the vulnerability meant the identity group subsystem was incorrectly deemed not affected, so only the identity entity system was patched.

## Rationale

In HCSEC-2025-13 / CVE-2025-5999 as fixed upstream, the fix was correctly applied to both the identity entity and identity group subsystems.

Resolves: GHSA-7ff4-jw48-3436

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2143.txt
+++ b/changelog/2143.txt
@@ -1,0 +1,3 @@
+```release-note:security
+core/identity: Correctly lowercase policy names on identity groups to prevent root policy assignment. CVE-2025-64761 / GHSA-7ff4-jw48-3436. Second part of upstream's HCSEC-2025-13 / CVE-2025-5999.
+```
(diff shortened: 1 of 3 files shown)


satoqz (approved):

cipherboy:
Test failures expected as #2132 was merged after this was opened.

PR #2154, opened 2025-11-25
docs: fix rendering of namespaces page

phil9909:
## Description

Fix this rendering issue:

image

on https://openbao.org/docs/concepts/namespaces/

## Rationale

There are two problems:

-  A missing colon to get a proper info box
- The `@include` syntax is a [ReMark plugin from HashiCorp](https://github.com/hashicorp/remark-plugins/tree/main/plugins/include-markdown), which seems to include files verbatim instead of rendering it to HTML. I replaced this with the more verbose Docusaurus way of doing things as documented here: https://docusaurus.io/docs/markdown-features/react#importing-markdown

I can create another PR to replace all uses of  `@include` with the `import` syntax.

BTW: should we just drop that info box? It seems rather obvious information to me.

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/concepts/namespaces/index.mdx
+++ b/website/content/docs/concepts/namespaces/index.mdx
@@ -123,4 +123,6 @@ endpoints. But, for security reasons, OpenBao restricts access to some of the
 system backend endpoints to calls from the root namespace or calls that use a
 token in the root namespace with elevated permissions.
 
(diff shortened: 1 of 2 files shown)


satoqz (approved):
I'd vote to drop the info box.

satoqz:
Merging this before I branch off to 2.5.x to avoid another backport. Thanks @phil9909!

PR #2163, opened 2025-11-27
Ignore returned error for tx.Rollback and Body.Close

phyrog:
## Description

Disable lint rule `errcheck` for lines not handling return values of http.Response.Body.Close and tx.Rollback

## Rationale

- tx.Rollback returns an error if tx has already been committed, see previous discussion here: https://github.com/openbao/openbao/pull/1990#discussion_r2527711000
- Body.Close only returns an error in very rare cases, and even then, there is nothing actionable for us to do

Part of: #1574 

## Acknowledgements

       also allowed.

### DIFF
--- a/api/auth_token.go
+++ b/api/auth_token.go
@@ -35,7 +35,7 @@ func (c *TokenAuth) CreateWithContext(ctx context.Context, opts *TokenCreateRequ
 	if err != nil {
 		return nil, err
 	}
-	defer resp.Body.Close()
+	defer resp.Body.Close() //nolint:errcheck
 
 	return ParseSecret(resp.Body)
 }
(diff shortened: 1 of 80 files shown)


satoqz:
@phyrog do you want to ignore the namespace store case, too? Then we can backport this one to 2.5 (though that particular fix could be in a different PR, it doesn't really matter IMO).

satoqz (approved):

cipherboy (approved):
While browsing, I came across https://github.com/securego/gosec/pull/935 which seems to have removed this check from one type of linter. Wonder if we can configure this linter similarly?

JanMa (approved):

phyrog:
> While browsing, I came across [securego/gosec#935](https://github.com/securego/gosec/pull/935) which seems to have removed this check from one type of linter. Wonder if we can configure this linter similarly?

Unfortunately that's not possible as far as I can tell. You can ignore functions completely (e.g. `fmt.Println`), but you can't ignore `defer`ed errors[^1]. Since `Response.Body` is a `io.ReadCloser` and not a concrete type, I don't think we can even use the config to ignore it.

[^1]: https://golangci-lint.run/docs/linters/configuration/#errcheck

satoqz:
I think no need to backport this anymore given #2226 removes the relevant namespace store code anyways.

Issue #2168, opened 2025-11-27
Declarative audit stanzas ignored at boot on 2.5.0-beta (audit device only loads after SIGHUP)


robertdebock:
### Describe the bug

[Declarative audit stanzas](https://openbao.org/docs/rfcs/config-audit-devices/) defined in the server configuration are not enabled when the cluster boots for the first time (observed on `openbao/openbao:2.5.0-beta20251125`). The `audit` block is parsed (warnings appear in the server log), but `bao audit list` shows `No audit devices are enabled` and no logfile is created. Sending `SIGHUP` to the leader or follower immediately causes OpenBao to log `core: adding new audit device`, enables `file-audit/`, and creates `/openbao/logs/audit.log`. The same configuration works as expected on `openbao/openbao:2.4.4` and enables the audit device at boot without any manual intervention.

### To Reproduce

1. Start from a clean slate:

   ```bash
   docker compose down -v
   docker compose up -d
   ```

> Note: Docker compose file is on the bottom of this file.

2. Authenticate on the leader (`openbao-1`):

   ```bash
   docker exec openbao-1 bao login -address=http://127.0.0.1:8200 \
     -method=userpass username=admin password=admin
   ```

3. List audit devices:

   ```bash
   docker exec openbao-1 bao audit list -address=http://127.0.0.1:8200
   # Output: "No audit devices are enabled."
   ```

4. Check the configured logfile path:

   ```bash
   docker exec openbao-1 ls -l /openbao/logs
   # Output: "total 0"
   ```

5. Send `SIGHUP` to the leader:

> Note: This "fixes" the issue, but that's not the point of this issue.

   ```bash
   docker kill -s HUP openbao-1
   ```

6. Repeat step 3–4:

   ```bash
   docker exec openbao-1 bao audit list -address=http://127.0.0.1:8200
   # file-audit/ is now listed

   docker exec openbao-1 ls -l /openbao/logs
   # audit.log now exists and grows
   ```

### Expected behavior

Per the [declarative audit documentation](https://openbao.org/docs/rfcs/config-audit-devices/), the `audit "<type>" "<path>"` stanza should create or remove audit devices on the active node during startup and whenever `SIGHUP` is sent, while standby nodes simply ignore the stanza but receive the replicated device state from the leader. A fresh deployment should therefore start with `file-audit/` already enabled and writing to `/openbao/logs/audit.log`.

### Actual behavior

[...]

PR #2169, opened 2025-11-27
Fix audit log configuration on active node

cipherboy:
## Description

On active nodes, standby=false during unseal and is only set to true once the active has fully finished initialization. This still makes some sense, as we don't really want to accept requests or handle background jobs until after unseal is finished.

However, in the course of HA, standby status needs to be passed through on the active core to allow audit devices to be configured during startup. This was already handled correctly for OCI plugins.

## Rationale

Resolves: #2168

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/audit.go
+++ b/vault/audit.go
@@ -662,12 +662,12 @@ func (c *Core) ReloadAuditLogs() {
 		return
 	}
 
-	if err := c.handleAuditLogSetup(c.activeContext); err != nil {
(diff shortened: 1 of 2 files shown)


cipherboy:
(Since this was introduced in v2.5.0 beta and fixed before GA, I don't think this needs a changelog entry.)

JanMa (approved):

PR #2171, opened 2025-11-28
Add versioned docs

JanMa:
This adds a new Makefile target in the `website` folder to build versioned docs.
The versions can be configured in the `doc-versions.json` file. During build we checkout the matching `release/*` branches for the desired versions and extract the docs from them. 

Resolves: #1681

## Acknowledgements

       also allowed.

### Version Dropdown
image

### Old version warning
image

### Development warning
image


website/Makefile
```diff
@@ -32,3 +33,46 @@ install: node_modules
 docker-start:
 	$(DOCKER) build -t openbao-docs .
 	$(DOCKER) run -p 3000:3000 --mount "type=bind,source=$(CURDIR),destination=/opt/docusaurus$(DOCKER_MOUNT_ARGS)" -it openbao-docs
+
+# Versioned docs
+GIT_REMOTE := $(shell git remote get-url origin)
```
phil9909: I call my remote `upstream` (`origin` in my case is the fork)

```suggestion
GIT_REMOTE ?= $(shell git remote get-url origin)
```

website/docusaurus.config.ts
```diff
@@ -3,10 +3,28 @@ import type { Config } from "@docusaurus/types";
 import type * as Preset from "@docusaurus/preset-classic";
 import { includeMarkdown } from "@hashicorp/remark-plugins";
 import path from "path";
+import pluginSidebarJson from "./src/docusaurus-plugin-sidebar-json";
+
+const docVersions = {
+  current: { label: "Development" },
+  "2.4.x": { label: "Version 2.4.x" },
+  "2.3.x": { label: "Version 2.3.x" },
+};
```
phil9909: Maybe we could move this information to a `versions.json` file, so we can more easily consume it from the `Makefile`

EDIT: I see you actually generate the `version.json` from grepping this file. I would do the other way around, just read the json file here and check it in. IMHO this is less error prone.

.github/workflows/docs-ci.yml
```diff
@@ -44,7 +44,11 @@ jobs:
         name: build-docs
         working-directory: ./website
         run: |
-          make build
+          make build-versioned
+      - uses: actions/upload-artifact@v4
```
satoqz: Should be pinned by hash - though, did you want to keep this or is it just for testing?

website/content/docs/policies/release.mdx
```diff
@@ -71,6 +71,7 @@ a better release process going forward.
       - [ ] If doing a patch release off an existing major or minor release
             branch, update the `main` repository's `CHANGELOG.md` file as
             well, ordering it according to version number.
+      - [ ] Update docs versions in `docusaurus.config.ts` if necessary
```
satoqz: I suppose since this maps to branches, we should add this to the last step down below:

>  For new major or minor versions, create a `release/<version>` branch
       after the first release on the minor series (including beta and other
       pre-release branches). Future features should target the next release
       (`main` branch) but bug and security fixes can be backported.

website/docusaurus.config.ts
```diff
@@ -3,10 +3,28 @@ import type { Config } from "@docusaurus/types";
 import type * as Preset from "@docusaurus/preset-classic";
 import { includeMarkdown } from "@hashicorp/remark-plugins";
 import path from "path";
+import pluginSidebarJson from "./src/docusaurus-plugin-sidebar-json";
+
+const docVersions = {
+  current: { label: "Development" },
+  "2.4.x": { label: "Version 2.4.x" },
```
satoqz:
```suggestion
  "2.5.x": { label: "Version 2.5.x" },
  "2.4.x": { label: "Version 2.4.x" },
```

Probably makes sense to have even if we're still in beta.

website/Makefile
```diff
@@ -32,3 +33,46 @@ install: node_modules
...
+	@mkdir -p $@
+
+api-docs_versioned_sidebars:
+	@mkdir -p $@
+
+api-docs_versioned_docs/version-%: node_modules | api-docs_versioned_docs api-docs_versioned_sidebars
+	@git init $@
+	@git -C $@ remote add origin $(GIT_REMOTE)
+	@git -C $@ sparse-checkout set --no-cone website/content/api-docs sidebarsApi.ts
+	@git -C $@ pull --depth 1 origin $(subst api-docs_versioned_docs/version-,release/,$@)
```
satoqz, line 72: I wonder if we could gracefully handle missing branches such that when a new major/minor is released, we don't need an additional final PR to main updating version information once the release branch was created (or, alternatively, push the release branch ahead of time and then push it up to the actual release commit later)

website/README.md
```diff
@@ -8,6 +8,7 @@ Website](https://openbao.org).  It is built using
 
 - [Contributions](#contributions-welcome)
 - [Running the Site Locally](#running-the-site-locally)
+- [Versioned documentation](#versioned-documentation)
```
satoqz:
```suggestion
- [Versioned Documentation](#versioned-documentation)
```

nit

website/Makefile
satoqz: Eh I guess you'd need to retrigger the docs-deploy workflow either way. Though probably we could trigger docs-deploy to run _on main_ whenever main _or_ a release branch was pushed. Thoughts?

JanMa: >  Though probably we could trigger docs-deploy to run on main whenever main or a release branch was pushed. 

Yes I think this should be possible with some sort of trigger :+1:

.github/workflows/docs-ci.yml
JanMa: For now It's just for testing the docs build properly, but I think it doesn't hurt to keep this

website/docusaurus.config.ts
JanMa: The problem is, that Docusaurus is very picky about the contents of a `versions.json` file and you can only include a list of version numbers. I would like us to be able to configure some more options, like the label that is displayed for a given version.

As a compromise I introduced  `doc-versions.json`

website/Makefile
JanMa: I adjusted the docs deployment workflow so it also triggers on release branches.

website/doc-versions.json
```diff
@@ -0,0 +1,11 @@
+{
+    "current": {
+        "label": "Development"
+    },
+    "2.4.x": {
+        "label": "Version 2.4.x"
+    },
```
satoqz, line 10:
```suggestion
    "2.5.x": {
        "label": "Version 2.5.x"
    },
    "2.4.x": {
        "label": "Version 2.4.x"
    },
```

website/README.md
```diff
@@ -28,6 +29,29 @@ installed](https://nodejs.org/en/) you can run:
 
 ...and then visit `http://localhost:3000/openbao`.
 
+## Versioned documentation
```
satoqz:
```suggestion
## Versioned Documentation
```

website/Makefile
```diff
@@ -32,3 +33,43 @@ install: node_modules
 docker-start:
 	$(DOCKER) build -t openbao-docs .
 	$(DOCKER) run -p 3000:3000 --mount "type=bind,source=$(CURDIR),destination=/opt/docusaurus$(DOCKER_MOUNT_ARGS)" -it openbao-docs
+
+# Versioned docs
+GIT_REMOTE ?= $(shell git remote get-url origin)
+DOCS_VERSIONS := $(shell jq -r 'keys_unsorted - ["current"] | .[]' doc-versions.json)
+
+build-versioned: node_modules $(addprefix versioned_docs/version-,$(DOCS_VERSIONS)) $(addprefix api-docs_versioned_docs/version-,$(DOCS_VERSIONS)) versions.json api-docs_versions.json
```
satoqz, line 41: I'm slightly out of the loop; what reads `versions.json` and `api-docs_versions.json`? I see docusaurus itself just reads `doc-versions.json`.

website/README.md
```diff
@@ -28,6 +29,29 @@ installed](https://nodejs.org/en/) you can run:
...
+
+The doc versions can be configured in the `docusaurus.config.ts` file:
+
+``` typescript
+const docVersions = {
+  current: { label: "Development" },
+  "2.4.x": { label: "Version 2.4.x" },
+  "2.3.x": { label: "Version 2.3.x" },
+};
+```
```
satoqz, line 60: This is outdated now that it's moved to `doc-versions.json`.

website/content/docs/policies/release.mdx
```diff
@@ -95,3 +96,4 @@ a better release process going forward.
        after the first release on the minor series (including beta and other
        pre-release branches). Future features should target the next release
        (`main` branch) but bug and security fixes can be backported.
+      - [ ] Update docs versions in `doc-versions.json` if necessary
```
satoqz:
```suggestion
```

I think this would always be part of the `CHANGELOG.md` commit (as described above). The flow would then be:

- `CHANGELOG.md` + `doc-versions.json` change is pushed to `main` -> docs-deploy workflow fails as branch is not found
- New `release/x.x.x` branch is pushed, docs-deploy workflow triggers again, succeeds this time around as branch is available

Accepting that one intermediate failure may be the path of least resistance? Not sure.

website/README.md
JanMa: Thanks, I missed this

website/Makefile
JanMa: Docusaurus depends on the `version.json` and `api-docs_versions.json` file to determine which versions it should include in the build. Therefore the Makefile has to generate them as a prerequisite to running the build command. The Docusaurus Docs have an overview of all the files it expects to be in place:

https://docusaurus.io/docs/versioning#overview

Unfortunately you need to adjust multiple places in the config to render the versioned docs correctly. 

- `versions.json` and `api-docs_versions.json` needs a list of the versions that should be included
- the docs plugin needs the versions and settings passed in its [config](https://github.com/openbao/openbao/pull/2171/files#diff-1de336ccbebeecf69018f0a9d5559e7f033215d5f0c399933cfd686186b5b0fbR75)
- the dropdown menu needs the versions and labels passed in its [config](https://github.com/openbao/openbao/pull/2171/files#diff-1de336ccbebeecf69018f0a9d5559e7f033215d5f0c399933cfd686186b5b0fbR191-R201)

Because there is no single place where I can set all config, I introduced the `docs-versions.json` file which is read by the docusaurus config and Makefile to adjust the different settings.

cipherboy:
@JanMa Wow, the new versioned docs build is _slow_! Looks great once built and it is fast locally, really like the switcher and the notices. I wonder if we can cache unchanged portions?

But either way, I think this is good enough to merge now and we'll iterate later.  Bypassing glitched DCO.

PR #2185, opened 2025-12-01
detect raft transaction leaks

phil9909:
## Description

Clean up leaked (meaning calling `BeginTx`, but never `Commit` or `Rollback`) raft transactions using `runtime.AddCleanup`

See also: [Zulip Thread](https://linuxfoundation.zulipchat.com/#narrow/channel/530381-openssf-openbao-support/topic/Raft.20Improvements/near/554338410)

## Rationale

Leaking a raft transaction has some negative effects:
- the fast transaction tracking will pile up memory
- the permit is not returned
- a read lock will be held indefinitely
- the underlying `bbolt` transaction will not be freed

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2185.txt
+++ b/changelog/2185.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+raft: detect, log, and rollback transactions that have never been committed or rolled-back. If you see the message "transaction was leaked" in your logs, please open an issue.
+```


physical/raft/transaction.go
```diff
@@ -738,19 +779,22 @@ func (t *RaftTransaction) Rollback(ctx context.Context) error {
...
+
 	// The transaction is done; release the permit pool entry when we're done
 	// here.
 	//
 	// Also unlock the read lock on the underlying fsm.
 	defer func() {
 		if t.writable {
-			lowestActiveIndex := t.b.fsm.fastTxnTracker.lowestActiveIndexAfterCommit(t.index)
+			t.b.fsm.fastTxnTracker.completeTransaction(t.index)
+			lowestActiveIndex := t.b.fsm.fastTxnTracker.lowestActiveIndex()
```
phil9909, line 795: `lowestActiveIndex()` was added after this code was written. I took the chance to clean this up.

physical/raft/transaction.go
```diff
@@ -738,19 +779,22 @@ func (t *RaftTransaction) Rollback(ctx context.Context) error {
...
+
 			t.b.l.RLock()
-			lowestActiveIndex = min(lowestActiveIndex, t.b.raft.AppliedIndex()-1) // we need to cap the lowest active index, otherwise we might miss transaction started later
+			lowestActiveIndex = min(lowestActiveIndex, t.b.raft.AppliedIndex()) // we need to cap the lowest active index, otherwise we might miss transaction started later
```
phil9909, line 798: The `-1` is not needed as `clearOldEntries` will already clean only entries older than the given value. This was already changed for the commit logic in #2067

physical/raft/transaction.go
```diff
@@ -216,16 +219,52 @@ func (b *RaftBackend) newTransaction(ctx context.Context, writable bool) (*RaftT
...
 		index:    index,
 		started:  time.Now(),
-	}, nil
+	}
+
+	txn.cleanup = runtime.AddCleanup(txn, func(startIndex uint64) {
+		// make sure we don't use txn directly in the function,
+		// otherwise we prevent it from ever being garbage collected.
+
+		b.logger.Error("transaction was leaked",
```
cipherboy:
```suggestion
		b.logger.Debug("transaction was leaked",
```

I think Error is too verbose.

cipherboy:
@phil9909 Apologies, dropped the ball here. Where are we with wanting this?

physical/raft/transaction.go
phil9909: If I remember correctly, a single leaked transaction currently equals unbounded memory growth (at least for raft), so `Error` seemed appropriate. But this PR at least prevents the leak, so maybe `Error` is to verbose.

On the other hand: A leaked transaction means there is a bug, so `Debug` is not prefect either IMO.

How about printing at `Error` level once and then downgrading to `Debug`?

phil9909:
> Where are we with wanting this?

I want to measure the runtime cost of this. I will have some overhead, but I suspect not too much.

physical/raft/transaction.go
cipherboy: That last one works for me.

I was just thinking certain flows were expected to leak transactions (e.g., slow call which resulted in context timeout), so we should be mindful of that and the potential to spam logs as a result. Or, if a heavily used endpoint does leak transactions, again, operator would see a lot of spam.

phil9909:
> I want to measure the runtime cost of this. I will have some overhead, but I suspect not too much.

I created a micro benchmar:

<details>
<summary>benchmark code</summary>

```go
package raft

import (
	"context"
	"testing"
	"time"

	"github.com/hashicorp/go-hclog"
	"github.com/hashicorp/go-uuid"
)

func BenchmarkRaftLeak(b *testing.B) {
	r := getBenchRaft(b)

	b.ResetTimer()

	b.RunParallel(func(p *testing.PB) {
		for p.Next() {
			tx, err := r.BeginTx(b.Context())
			require.NoError(b, err)
			require.NoError(b, tx.Rollback(b.Context()))
		}
	})
}

func getBenchRaft(t testing.TB) *RaftBackend {
	raftDir := t.TempDir()

	id, err := uuid.GenerateUUID()
	if err != nil {
		t.Fatal(err)
	}

	logger := hclog.NewNullLogger()

	conf := map[string]string{
		"path":                  raftDir,
		"trailing_logs":         "100",
		"node_id":               id,
		"doNotStoreLatestState": "",
	}

	backendRaw, err := NewRaftBackend(conf, logger)
	if err != nil {
		t.Fatal(err)
	}
	backend := backendRaw.(*RaftBackend)

	err = backend.Bootstrap([]Peer{
		{
			ID:      backend.NodeID(),
			Address: backend.NodeID(),
		},
	})
	if err != nil {
		t.Fatal(err)
	}

	err = backend.SetupCluster(context.Background(), SetupOpts{})
	if err != nil {
		t.Fatal(err)
	}

	for backend.raft.AppliedIndex() < 2 {
		time.Sleep(time.Millisecond)
	}

	backend.DisableAutopilot()

	return backend
}
```

</details>

Here are my results:

  | [`c9dde8c0eb`](https://github.com/openbao/openbao/commit/c9dde8c0eb747be2d0a452a010c9eda34f61d7e4) | [`0ff2b33014`](https://github.com/adfinis-forks/openbao/commit/0ff2b3301405e52bad57727dd9262a2900b21a6d) | slowdown
-- | -- | -- | --
BenchmarkRaftLeak | 1239 | 1309 | 5.65%
BenchmarkRaftLeak-4 | 1411 | 1431 | 1.42%
BenchmarkRaftLeak-16 | 2099 | 2190 | 4.34%
BenchmarkRaftLeak-64 | 2662 | 2721 | 2.22%
BenchmarkRaftLeak-256 | 2686 | 2804 | 4.39%



```console
$ git rev-parse HEAD; go test -cpu 1,4,16,64,256 -benchtime 10s -bench BenchmarkRaftLeak ./physical/raft/ -run '^$'
0ff2b3301405e52bad57727dd9262a2900b21a6d
goos: linux
goarch: amd64
pkg: github.com/openbao/openbao/physical/raft
cpu: Intel(R) Core(TM) Ultra 7 155H
BenchmarkRaftLeak         	 9547681	      1309 ns/op
BenchmarkRaftLeak-4       	10296885	      1431 ns/op
BenchmarkRaftLeak-16      	 6948542	      2190 ns/op
BenchmarkRaftLeak-64      	 6359874	      2721 ns/op
BenchmarkRaftLeak-256     	 6659652	      2804 ns/op
PASS
ok  	github.com/openbao/openbao/physical/raft	277.989s


$ git rev-parse HEAD; go test -cpu 1,4,16,64,256 -benchtime 10s -bench BenchmarkRaftLeak ./physical/raft/ -run '^$'
c9dde8c0eb747be2d0a452a010c9eda34f61d7e4
goos: linux
goarch: amd64
pkg: github.com/openbao/openbao/physical/raft
cpu: Intel(R) Core(TM) Ultra 7 155H
BenchmarkRaftLeak         	10129227	      1239 ns/op
BenchmarkRaftLeak-4       	10005015	      1411 ns/op
BenchmarkRaftLeak-16      	 7430812	      2099 ns/op
BenchmarkRaftLeak-64      	 6738628	      2662 ns/op
BenchmarkRaftLeak-256     	 6828820	      2686 ns/op
PASS
ok  	github.com/openbao/openbao/physical/raft	276.528s
```

So creating (and rollback) of a single transaction gets around 5% slower.

Setting this in the macro context: I started a bao dev server `bao server -dev -dev-root-token-id=root` and ran `curl -X POST -H "X-Vault-Token: root" -d '{"data":{"bar":"bazz"}}' -w '%{time_total}s' "http://127.0.0.1:8200/v1/secret/data/foo"`.
This report around 0.001s (or 1_000_000 ns).

Adding a 100 or 200ns to this should not be notifiable, especially when you consider that this was a dev server (using an in memory storage). I don't think it is worth the time to tests this with a real raft cluster, I don't expect this to be measurable difference.

phil9909:
> Likely should add it to the GRPC layer (for external plugins, though I'm not aware of any that use transactions yet) and PostgreSQL. Want to take a stab at those?

I was under the assumption that external plugins should be covered (if the cluster uses raft).

For PostgreSQL, I can add it as well.

Might be worth adding it to the `InmemBackend` as well, so we can catch leaks earlier.

Should we add this to this PR, or follow up?

physical/raft/transaction.go
phil9909: > I was just thinking certain flows were expected to leak transactions (e.g., slow call which resulted in context timeout),

IMHO: In a timeout, it should still call rollback.

Change to "error once then debug"

cipherboy: @phil9909 Ideally yes, though I think it comes down to how context cancellation is handled. If we're passed a cancelled context, can we still rollback?

cipherboy:
@phil9909 Happy doing separate PRs for these. I do think this warrants a changelog entry, though.

phil9909:
> I do think this warrants a changelog entry, though.

Definitely.
Added that

PR #2188, opened 2025-12-03
Fix standby namespace deletion

cipherboy:
## Description

This fixes two bugs with namespace deletion:

1. On standby nodes, the wrong context was used for invalidation, leading to the standby's root namespace's `sys/` mount being unmounted (along with any other equivalent mounts). As this is only in beta, it doesn't have a changelog entry. This did not affect actual storage and was recoverable by restarting the standby node.
2. On all nodes, namespace deletion could cause startup failures if incomplete. We marked the namespace as tainted but did not ignore tainted namespaces in several loops (lease cleanup, ...), leading to a loss of ability to start up. This affects both standby nodes (starting when the leadership is in inconsistent state) and active nodes (restarted when storage currently has a partially deleted namespace). The expected behavior is to re-delete the partially deleted namespace. 

## Rationale

Reported by @wslabosz-reply on Zulip.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2188.txt
+++ b/changelog/2188.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/namespaces: improve recovery from partial deletion of namespaces, preventing server startup failure
+```


cipherboy:
Thanks @wslabosz-reply, merging now that I've added changelog and updated description! No other changes were made.

vault/namespace_store.go
```diff
@@ -484,6 +480,10 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
 		}
 	}
 
+	if err := ns.writeNamespace(ctx, storage, entry); err != nil {
+		return fmt.Errorf("failed to persist namespace: %w", err)
+	}
```
satoqz, line 485: I think now that we unlock early only if the namespace is new and perform the write _afterwards_, a subsequent storage write against the existing namespace could race against it, losing the second write?

cipherboy: @satoqz @wslabosz-reply This is interesting, but I think I agree.

_Technically_ from a storage PoV, we don't need a lock held. But we need this from a consistency w.r.t. multiple writes ordering.

I think we could get around this by re-locking here and then cross-comparing the to-be-written entry with the version still in storage.

That would open us up to a race condition between namespace creation and deletion then.

My suggestion would then be a write flow similar to deletion, wherein we first write it with tainted status (though, in-memory mark it not-tainted), and finish with a follow-up write to clear tainted, after re-grabbing the lock and verifying we haven't modified the namespace since.

Alternatively, by creating the namespace as tainted, we could actually mark it tainted in memory and hope that we can still call `initializeNamespace(...)`, thus preventing further modifications to this namespace. We can then re-acquire lock (knowing we'd at most be racing with deletion) and write the non-tainted variant. We could avoid this race via a separate tainted-like (`creating: true`) status as well.

Thoughts?

satoqz: @cipherboy perhaps we should revisit whether `initializeNamespace` really requires an intermediate unlock? IIUC we don't have a deep technical reason besides "mount/unmount code may want to read-lock the namespace store in certain cases". Perhaps this could also be eased by @wslabosz-reply's mounts refactor.

cipherboy: @satoqz I recall we had deadlocks as subsystems referenced namespaces. But that's my recollection.

PR #2189, opened 2025-12-03
Standby nodes namespace seal status propagation grpc definition

wslabosz-reply:
Waiting on the https://github.com/openbao/openbao/pull/2204 merge.

Note: this PR targets [namespaces-seal](https://github.com/openbao/openbao/tree/namespaces-seal) branch

## Acknowledgements

       also allowed.


vault/namespace_store.go
```diff
@@ -515,18 +543,18 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
...
+		if err := ns.sealNamespaceLocked(ctx, storage, entry); err != nil {
 			return nil, fmt.Errorf("failed to seal namespace: %w", err)
 		}
 	}
 
+	// Finally commit the changes into storage.
+	if err := commit(); err != nil {
+		return nil, err
+	}
+
```
wslabosz-reply, line 536: under review, this one is complicated because:
a) we cannot write the entry after committing (transaction), as this function will finish with `cleanup()`
b) we cannot seal before writing as the entry will be written with wrong barrier

EDIT: adjusted

vault/namespace_seal_service.proto
```diff
@@ -0,0 +1,29 @@
...
+    string namespace_uuid = 1;
+ };
+
+message SealStateStatus {
+    // namespace_uuid is the UUID of the namespace
+    string namespace_uuid = 1;
+    // root_key contains the namespace root key if it is unsealed. 
+    // It is unset otherwise.
+    optional bytes root_key = 2;
+};
```
phyrog, line 21: I think we'll want to batch these, to avoid sending n requests for n namespaces.

phyrog: Side question: why do we request one namespace here, but return a list of SealStateStatus?

cipherboy: I suspect that can be an optimization later, if/when traffic warrants it, but I agree about the asymmetry here.

wslabosz-reply:
Changes introduced with this PR will be directly open towards main later on, closing for now.

PR #2193, opened 2025-12-04
Resolve some more errcheck findings

phyrog:
## Description

Resolves a number of errcheck findings for not handling `os.RemoveAll` in tests to delete temporary directories by just not doing that and instead using `t.TempDir`, which automatically handles temp dir removal. I also cleaned up now unused return values of a couple of functions/methods.

I also had to update mitchellh/go-testing-interface, since the latest tagged version did not include `TempDir` in the `T` interface. I just updated it to the latest commit before the repo was archived.

## Rationale

Part of #1574 

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/audit/file/backend_test.go
+++ b/builtin/audit/file/backend_test.go
@@ -26,12 +26,7 @@ func TestAuditFile_fileModeNew(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	path, err := os.MkdirTemp("", "vault-test_audit_file-file_mode_new")
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	defer os.RemoveAll(path)
+	path := t.TempDir()
 
 	file := filepath.Join(path, "auditTest.txt")
 
--- a/builtin/credential/cert/backend_test.go
+++ b/builtin/credential/cert/backend_test.go
@@ -67,10 +67,8 @@ const (
 
 func generateTestCertAndConnState(t *testing.T, template *x509.Certificate) (string, tls.ConnectionState, error) {
 	t.Helper()
-	tempDir, err := os.MkdirTemp("", "vault-cert-auth-test-")
-	if err != nil {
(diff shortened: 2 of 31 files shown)


go.sum
```diff
@@ -637,8 +637,8 @@ github.com/mitchellh/copystructure v1.2.0/go.mod h1:qLl+cE2AmVv+CoeAwDPye/v+N2HK
 github.com/mitchellh/go-homedir v1.1.0 h1:lukF9ziXFxDFPkA1vsr5zpc1XuPDn/wFntq5mG+4E0Y=
 github.com/mitchellh/go-homedir v1.1.0/go.mod h1:SfyaCUpYCn1Vlf4IUYiD9fPX4A5wJrkLzIz1N1q0pr0=
 github.com/mitchellh/go-testing-interface v1.0.0/go.mod h1:kRemZodwjscx+RGhAo8eIhFbs2+BFgRtFPeD/KE+zxI=
-github.com/mitchellh/go-testing-interface v1.14.1 h1:jrgshOhYAUVNMAJiKbEu7EqAwgJJ2JqpQmpLJOu07cU=
-github.com/mitchellh/go-testing-interface v1.14.1/go.mod h1:gfgS7OtZj6MA4U1UrDRp04twqAjfvlZyCfX3sDjEym8=
+github.com/mitchellh/go-testing-interface v1.14.2-0.20210821155943-2d9075ca8770 h1:drhDO54gdT/a15GBcMRmunZiNcLgPiFIJa23KzmcvcU=
```
cipherboy, line 635: @phyrog I wonder if we can drop this in the future?

cipherboy (approved):
Sorry about the delay, and thanks as always for this @phyrog!  Will let you merge.

go.sum
phyrog: Would be nice! I believe the major part we're using it for is three-node dev clusters? Would also help with (but not solve) #2305

PR #2204, opened 2025-12-08
Fix shadowing of the transaction with barrier retrieval in mounts

wslabosz-reply:
Fixes the shadowing of the transaction storage by using a method that retrieves the barrier from the in-memory struct (SealManager).

Note: this PR targets [namespaces-seal](https://github.com/openbao/openbao/tree/namespaces-seal) branch

- adjusted barrier passing/retrieval in `mount.go` and `auth.go` 
- adjusted `(*Core).moveStorage()` method
- adjusted test cases related to (re)mounting in `mount_test.go` and `auth_test.go`
- adjusted tests cases in `vault/rollback_test.go` and `vault/core_cache_invalidate_test.go`
- (unrelated cleanup) removed `mountStateUnmounting` and `MountState` from `MountEntry`

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -65,11 +65,7 @@ func (c *Core) enableCredential(ctx context.Context, entry *MountEntry) error {
 	}
 
 	// Enable credential internally
-	if err := c.enableCredentialInternal(ctx, entry, MountTableUpdateStorage); err != nil {
-		return err
-	}
-
-	return nil
+	return c.enableCredentialInternal(ctx, entry, MountTableUpdateStorage)
 }
 
 // enableCredential is used to enable a new credential backend
@@ -194,6 +190,7 @@ func (c *Core) enableCredentialInternalWithLock(ctx context.Context, entry *Moun
 	newTable := c.auth.shallowClone()
 	newTable.Entries = append(newTable.Entries, entry)
 	if updateStorage {
+		// barrier nil is fine here, as the namespace is embedded in the context
 		if err := c.persistAuth(ctx, nil, newTable, &entry.Local, entry.UUID); err != nil {
 			c.logger.Error("failed to update auth table", "error", err)
 			return fmt.Errorf("failed to update auth table: %w", err)
@@ -350,6 +347,7 @@ func (c *Core) removeCredEntryWithLock(ctx context.Context, path string, updateS
 
 	if updateStorage {
 		// Update the auth table
+		// barrier nil is fine here, as the namespace is embedded in the context
 		if err := c.persistAuth(ctx, nil, newTable, &entry.Local, entry.UUID); err != nil {
 			return fmt.Errorf("failed to update auth table: %w", err)
 		}
@@ -374,14 +372,13 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
 		return fmt.Errorf("cannot remount auth mount to non-auth mount %q", dst.MountPath)
 	}
 
-	for _, auth := range protectedAuths {
-		if strings.HasPrefix(src.MountPath, auth) {
+	// Prevent protected paths from being remounted, or target mounts being in protected paths
+	for _, p := range protectedAuths {
+		if strings.HasPrefix(src.MountPath, p) {
 			return fmt.Errorf("cannot remount %q", src.MountPath)
 		}
-	}
 
-	for _, auth := range protectedAuths {
-		if strings.HasPrefix(dst.MountPath, auth) {
+		if strings.HasPrefix(dst.MountPath, p) {
 			return fmt.Errorf("cannot remount to %q", dst.MountPath)
 		}
 	}
@@ -397,11 +394,6 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
 
 	if match := c.router.MountConflict(ctx, dstRelativePath); match != dst.Namespace.Path && match != "" {
 		return fmt.Errorf("path in use at %q", match)
-	}
-
-	srcBarrierView, err := c.mountEntryView(mountEntry)
-	if err != nil {
-		return err
 	}
 
 	// Mark the entry as tainted
@@ -434,13 +426,8 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
 	srcPath := mountEntry.Path
 	mountEntry.Path = strings.TrimPrefix(dst.MountPath, credentialRoutePrefix)
 
-	dstBarrierView, err := c.mountEntryView(mountEntry)
-	if err != nil {
-		return err
-	}
-
 	// Update the mount table
-	if err := c.persistAuth(ctx, nil, c.auth, &mountEntry.Local, mountEntry.UUID); err != nil {
+	if err := c.persistAuth(ctx, c.NamespaceView(mountEntry.namespace), c.auth, &mountEntry.Local, mountEntry.UUID); err != nil {
 		mountEntry.namespace = src.Namespace
 		mountEntry.NamespaceID = src.Namespace.ID
 		mountEntry.Path = srcPath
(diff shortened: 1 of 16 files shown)


vault/seal_manager.go
```diff
@@ -168,7 +168,9 @@ func (sm *SealManager) RemoveNamespace(ns *namespace.Namespace) {
 
 // NamespaceView finds the correct barrier to use for the namespace
 // and returns BarrierView restricted to the logical.Storage space
-// of the given namespace.
+// of the given namespace. As this function retrieves the barrier
+// from the seal manager, remember to __not__ shadow the transaction
+// by retrieving the view using this method.
```
```suggestion
// of the given namespace. As this function retrieves the barrier
// from the seal manager, remember to __not__ shadow transactions
// by retrieving the view using this method.
```

vault/auth.go
```diff
@@ -598,7 +598,7 @@ func (c *Core) loadTransactionalCredentials(ctx context.Context, barrier logical
 	}
 
 	if len(globalEntries) == 0 {
-		// TODO(ascheel) Assertion: globalEntries is empty iff there is only
+		// TODO(ascheel) Assertion: globalEntries is empty if there is only
```
```suggestion
		// TODO(ascheel) Assertion: globalEntries is empty iff there is only
```

https://en.wikipedia.org/wiki/iff

wslabosz-reply: lol, my bad

vault/seal_manager.go
wslabosz-reply: correct

vault/seal_manager.go
```diff
@@ -166,9 +166,13 @@ func (sm *SealManager) RemoveNamespace(ns *namespace.Namespace) {
...
-// NamespaceView finds the correct barrier to use for the namespace
-// and returns BarrierView restricted to the logical.Storage space
-// of the given namespace.
+// NamespaceView returns BarrierView restricted to the logical.Storage
+// space of the given namespace (if it exists, otherwise returning parent's
+// barrier, going up to root namespace barrier).
+// As this function retrieves __different__ barrier, instead of returning
+// subview of an another barrier, remember to __not__ shadow transactions
+// initialized with separate barrier instance by retrieving the barrier view
+// using this method.
```
satoqz: I think we can keep it simple:

```suggestion
// NamespaceView returns the BarrierView that applies to the given namespace.
// Remember that this method does not take in an existing storage type and is
// likely wrong to call within the context of a transaction.
```

vault/rollback_test.go
```diff
@@ -90,6 +90,7 @@ func TestRollbackManager_ManyWorkers(t *testing.T) {
 	ran := make(chan string)
 	release := make(chan struct{})
 	core, _, _ = testCoreUnsealed(t, core)
+	ctx := namespace.RootContext(context.Background())
```
satoqz:
```suggestion
	ctx := namespace.RootContext(t.Context())
```

and more below.

vault/auth.go
```diff
@@ -925,7 +912,11 @@ func (c *Core) persistAuth(ctx context.Context, barrier logical.Storage, table *
 	// Sometimes we may not want to explicitly pass barrier; fetch it if
 	// necessary.
 	if barrier == nil {
-		barrier = c.barrier
+		ns, err := namespace.FromContext(ctx)
+		if err != nil {
+			return err
+		}
+		barrier = c.NamespaceView(ns)
 	}
```
satoqz, line 916: I'd prefer if we did the opposite and enforced that an explicit `barrier` is passed. That seems more robust w.r.t. the awareness that's required by the caller regardless now that barriers must be selected carefully.

vault/namespace_store.go
```diff
@@ -433,7 +433,7 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
 	// when a namespace entry existed but the corresponding sys/ mount table
 	// entry wasn't created.
 	failed := true
-	storage := ns.storage
```
satoqz: Do we still need this field?

vault/namespace_store.go
```diff
@@ -545,29 +545,28 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
...
+	}
 
-		return nil
-	}); err != nil {
+	if err := logical.WithTransaction(ctx, storage.SubView(namespaceStoreSubPath),
+		func(s logical.Storage) error {
+			item, err := logical.StorageEntryJSON(entry.UUID, &entry)
+			if err != nil {
+				return fmt.Errorf("error marshalling storage entry: %w", err)
+			}
```
satoqz: This can be marshalled before starting the transaction.

vault/namespace_store.go
```diff
@@ -545,29 +545,28 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
...
+
+			if err := s.Put(ctx, item); err != nil {
+				return err
+			}
```
satoqz: ...and then there's no reason left to even have a transaction since we only have a single operation.

vault/namespace_store.go
```diff
@@ -609,16 +608,16 @@ func (ns *NamespaceStore) initializeNamespace(ctx context.Context, storage logic
 		return fmt.Errorf("error creating default policies: %w", err)
 	}
 
-	return ns.createMounts(nsCtx, storage)
+	return ns.createMounts(nsCtx, ns.core.NamespaceView(entry))
 }
 
-// createMounts handles creation of sys/ and token/ mounts for this new
```
satoqz, line 624: identity too, so maybe we'll want to just call these "required mounts".

vault/namespace_store.go
```diff
@@ -518,7 +518,7 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
 		unlocked = true
 
 		// Create sys/, token/ mounts and policies for the new namespace.
-		if err := ns.initializeNamespace(ctx, storage, entry); err != nil {
+		if err := ns.initializeNamespace(ctx, entry); err != nil {
```
satoqz, line 531: I think there are some additional remains that must be cleaned up:

- Default policy (This was previously cleaned up by rolling back the transaction)
- Anything written by `SetSeal()` (both in-memory and storage).

satoqz (commented):
I think this will create code drift over the fix to https://github.com/openbao/openbao/issues/2214. IMO we should address main first and then see what the ramifications are for this branch.

vault/auth.go
```diff
@@ -428,13 +421,8 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
...
 	mountEntry.Path = strings.TrimPrefix(dst.MountPath, credentialRoutePrefix)
 
-	dstBarrierView, err := c.mountEntryView(mountEntry)
-	if err != nil {
-		return err
-	}
-
 	// Update the mount table
-	if err := c.persistAuth(ctx, nil, c.auth, &mountEntry.Local, mountEntry.UUID); err != nil {
+	if err := c.persistAuth(ctx, c.NamespaceView(mountEntry.namespace), c.auth, &mountEntry.Local, mountEntry.UUID); err != nil {
```
cipherboy, line 427: Yeah, it is a little larger change but I think I agree with @satoqz that if we're going to adjust this, we should make sure we globally pass barrier. It is more difficult, but single barrier view only really made sense before namespaces anyways (even ignoring the matter of multiple sealed namespace barriers, there's still many views over the same barrier today).

vault/rollback_test.go
wslabosz-reply: ahh forgot about the `t.Context()`, will adjust

vault/namespace_store.go
wslabosz-reply: will wait on rebase on main with changes related to `NamespaceStore`

vault/namespace_store.go
wslabosz-reply: will wait on rebase on main with changes related to `NamespaceStore`

vault/namespace_store.go
wslabosz-reply: will wait on rebase on main with changes related to `NamespaceStore`

vault/namespace_store.go
wslabosz-reply: will wait on rebase on main with changes related to `NamespaceStore`

vault/auth.go
wslabosz-reply: adjusted!

wslabosz-reply:
Changes introduced with this PR will be directly open towards main later on, closing for now.

PR #2207, opened 2025-12-08
Add RFC on ACL Boundaries

satoqz:
## Summary

This RFC proposes a new policy type, "ACL Boundary", that is based on existing ACL Policies. Unlike traditional ACL Policies, ACL Boundaries are attached to namespaces (rather than tokens or identity mappings) and allow operators and administrators to globally limit the base token capabilities granted to a tenant's subtree of namespaces.

[Rendered](https://github.com/Ki-Reply-GmbH/openbao/blob/rfc-acl-boundaries/website/content/docs/rfcs/acl-boundaries.mdx)

---

cc @cipherboy this mirrors some of the initial thoughts we've had around tenant capability management (that is: many, though not all, it is hard to squash everything in one strike). Ultimately this is quite different in design than what we had discussed so far. Let me know what you think!

Thank you @Huy-Dinh-Reply for playing rubber duck and leading me to the ultimate realization that reusing the ACL Policy format may be a good fit.

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/rfcs/acl-boundaries.mdx
+++ b/website/content/docs/rfcs/acl-boundaries.mdx
@@ -1,0 +1,327 @@
+---
+sidebar_label: ACL Boundaries
+description: |-
+  An OpenBao RFC on adding ACL Boundaries to layer global access controls across
+  namespaces to better control capabilities in multi-tenant scenarios.
+---
+
+# ACL Boundaries
+
+## Summary
+
+This RFC proposes a new policy type, "ACL Boundary", that is based on existing
+ACL Policies. Unlike traditional ACL Policies, ACL Boundaries are attached
+to namespaces (rather than tokens or identity mappings) and allow operators
+and administrators to globally limit the base token capabilities granted to a
+tenant's subtree of namespaces.
+
+## Problem Statement
+
+With the introduction of namespaces (and, in the future, [Per-Namespace
+Sealing](namespace-sealing.mdx)), OpenBao now presents a strong foundation for
+multi-tenant use cases by delegating administration of a namespace or an entire
+subtree of namespaces to a separate tenant. While a tenant should generally
+receive full control over their namespace, the instance operator (or tenant of
+a parent namespace) may want to globally restrict certain capabilities within a
+sub-tenant's space.
+
+Thinkable use cases include:
+
+- Limiting the types of secrets and auth engines that can be mounted in a
+  namespace. For example, allow a tenant to create secrets engine mounts of type
+  `kv` and `pki` in their namespaces, but not `transit`.
+
+- Limiting the KMS providers that a tenant can consume via [External
+  Keys](external-keys.mdx) or via an Auto Seal used with [Per-Namespace
+  Sealing](namespace-sealing.mdx). For example, the operator may want to allow
+  only certain namespaces to access PKCS#11 modules.
+
+- Generally limiting available endpoints and their allowed parameters, like ACL
+  Policies do for tokens. For example, forbidding the creation of additional
+  child namespaces even when the operation is driven by a namespace-level root
+  token generated in a sealable namespace.
+
+Presently, there is no way to achieve any of the above unless tenants are not
+given the capability to manage policies in their namespaces or cannot generate
+their own root tokens given a sealable namespace.
+
+### Prior Work
+
+The accepted [External Keys RFC](external-keys.mdx) previously proposed to solve
+the respective subset of the problem described above by adding new configuration
+options specific to External Keys to both namespaces via the API and the server
+configuration file, storing what is effectively an allowlist of KMS providers
+per namespace. While workable, this design is feature-specific and does not
+generically handle many of the possible concerns around tenant restriction.
+This RFC proposes an alternative design that would supersede what was previously
+proposed for External Keys (Note that External Keys has not been released at the
+time of writing, hence the design is still free to pivot).
+
+## User-facing Description
+
+A new policy type, `acl-boundary`, will be available besides `/sys/policies/acl`
+under `/sys/policies/acl-boundary`. ACL Boundary Policies (and related
+endpoints) share the same parameters and syntax that present ACL Policies
+do - The additional endpoints are just provide an additional bucket of named
+policies. ACL Boundary Policies cannot be attached to tokens, which continue to
+refer to the existing ACL Policies only.
+
+A new field, `boundary_policies`, will be available under
+`/sys/namespaces/:name`. This is a list of policy names that refer to ACL
+Boundary Policies within the parent namespace (i.e., the namespace that hosts
+the `sys/` mount that is interacted with). For example:
+
+```shell-session
+$ bao write sys/policies/acl-boundary/foo policy=@foo.hcl
+$ bao write sys/namespaces/bar boundary_policies=foo
+```
+
+The policy is attached to the `bar/` namespace, but hosted in its parent
+namespace (the root namespace in this case).
+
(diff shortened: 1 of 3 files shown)


pree:
I like this idea a lot! As mentioned in [#openssf-openbao-discussion > inherit policies from parent namespace](https://linuxfoundation.zulipchat.com/#narrow/channel/529890-openssf-openbao-discussion/topic/inherit.20policies.20from.20parent.20namespace/with/568776152) I would also love to see a similar inherit mechanism to be present for normal policies, could ofc also be a different type of policy, like `ACL Inherited Policy`

website/content/docs/rfcs/acl-boundaries.mdx
```diff
@@ -0,0 +1,327 @@
+---
+sidebar_label: ACL Boundaries
+description: |-
+  An OpenBao RFC on adding ACL Boundaries to layer global access controls across
+  namespaces to better control capabilities in multi-tenant scenarios.
+---
+
+# ACL Boundaries
```
cipherboy, line 8: I don't like calling them ACLs; I think we should reserve the ACL terminology for token-adjacent policies. Otherwise we risk confusion between ACL policies and ACL boundary policies, &c.

cipherboy: Especially since these are namespace-attached rather than path-attached like ACLs are today.

cipherboy: How about calling them "Namespace Policies"? Then `sys/policies/namespace` makes a lot of sense.

website/content/docs/rfcs/acl-boundaries.mdx
```diff
@@ -0,0 +1,327 @@
...
+- Limiting the types of secrets and auth engines that can be mounted in a
+  namespace. For example, allow a tenant to create secrets engine mounts of type
+  `kv` and `pki` in their namespaces, but not `transit`.
+
+- Limiting the KMS providers that a tenant can consume via [External
+  Keys](external-keys.mdx) or via an Auto Seal used with [Per-Namespace
+  Sealing](namespace-sealing.mdx). For example, the operator may want to allow
+  only certain namespaces to access PKCS#11 modules.
+
+- Generally limiting available endpoints and their allowed parameters, like ACL
```
cipherboy, line 39: This is the only moderately token-bound restriction. Everything else is more of a system restriction that wouldn't really be evaluated in the ACL policy evaluation flow... Even this likely wouldn't fit with the existing policy engine and would be better as a post-policy follow-up.

website/content/docs/rfcs/acl-boundaries.mdx
```diff
@@ -0,0 +1,327 @@
...
+
+A new field, `boundary_policies`, will be available under
+`/sys/namespaces/:name`. This is a list of policy names that refer to ACL
+Boundary Policies within the parent namespace (i.e., the namespace that hosts
+the `sys/` mount that is interacted with). For example:
+
+```shell-session
+$ bao write sys/policies/acl-boundary/foo policy=@foo.hcl
+$ bao write sys/namespaces/bar boundary_policies=foo
+```
```
cipherboy, line 77: Out of curiosity, why not make `sys/policies/namespace/:name` be the name of the namespace it applies to? 

While we prevent reuse across sibling namespaces (outside perhaps of a default policy for all children), we allow clear nested overrides by administrators (`sys/policies/namespace/foo/bar` can grant PKI usage even if `sys/policies/namespace/foo` forbids it, assuming `foo/sys/policies/namespace/bar` doesn't also forbid it).

It also avoids having to apply a two-way association of policies. Policy evaluation can be done statically on the namespace in the store --- we (recursively on parent namespace) load policies for this given namespace and then retain it until invalidated by a write.

Thoughts?

website/content/docs/rfcs/acl-boundaries.mdx
```diff
@@ -0,0 +1,327 @@
...
+
+The policy is attached to the `bar/` namespace, but hosted in its parent
+namespace (the root namespace in this case).
+
+When OpenBao handles a request, it then evaluates ACL Boundaries against the
+request starting from the root namespace walking down to:
```
cipherboy, line 83: This evaluation I think is why the other storage/association format would make a lot of sense. While you could pre-compute them there as well, you have to handle invalidation differently as you'd need to associate which policy with which namespace(s). 

Though, you'd only have to deal with a single source of invalidation (direct parent) and not distant ancestors which is nice in other ways.  And we could always start with that in the other approach (only allow `:direct-child` and not `:direct-child/:indirect-child`.

website/content/docs/rfcs/acl-boundaries.mdx
```diff
@@ -0,0 +1,327 @@
...
+while the root policy always bypass any additional rules. To be exact, neither a
+root policy nor a default policy exist in the context of ACL Boundary Policies.
+
+If ACL evaluation fails, the request is rejected. If the list of Boundary
+Policies in the namespace is empty (the default), the request is never rejected
+as only the virtual allow-all policy is included.
+
+Note that this process is disjoint from token/identity ACL policy evaluation,
+which occurs after ACL Boundary evaluation. Either step may fail and reject
+the request; if token policy evaluation previously rejected a request it will
```
cipherboy, line 111: For some reason I was thinking the reverse, but this is fine.

website/content/docs/rfcs/acl-boundaries.mdx
```diff
@@ -0,0 +1,327 @@
...
+continue to do so regardless of ACL Boundaries as they cannot grant additional
+capabilities.
+
+Additionally, a new top-level boolean field `inheritable` (default `false`) is
```
cipherboy, line 115: Interesting, I never thought of this... 🤔 

I was assuming we'd be _always_ namespace-insensitive by default. 

E.g., denying `sys/mounts/:asdf` (in a root-level policy applied to `bar`) would mean it'd deny it in `bar/sys/mounts/:asdf`, `bar/foo/sys/mounts/:asdf`, &c.

website/content/docs/rfcs/acl-boundaries.mdx
```diff
@@ -0,0 +1,327 @@
...
+accepted in policy configuration files. If set to `true`, the policy's paths are
+always matched against the request's target namespace path rather than the path
+of the policy's namespace. This can be used to restrict well-known endpoints
+such as system backend endpoints across all child namespaces without needing to
+know or specify all child namespace paths:
+
+```hcl
+inheritable = true
+
+path "sys/mounts/*" {
```
cipherboy, line 125: Wait, now I'm confused. Is this in an _existing_ ACL policy (`sys/policies/acl/my-policy`) or within a "boundary" policy?

website/content/docs/rfcs/acl-boundaries.mdx
```diff
@@ -0,0 +1,327 @@
...
+
+`inheritable` will only be an accepted field within the context of ACL Boundary
+Policies, though extending support to traditional ACL Policies further down the
+line is considerable.
+
+For completeness, Boundary Policies may also be attached to the root namespace
+by adding `boundary_policy` stanzas to the server configuration file.
+
+```hcl
+boundary_policy "<name>" {
```
cipherboy, line 144: I'm kinda missing two key points here, which are the first two items in the "thinkable use cases" --- how do we limit the types of secrets engines, and how do we limit the seal mechanisms?

satoqz:
After some talks with @cipherboy, I'm putting this on hold until we find time and space to have a more focused design round on the problem at hand and get to think about some alternative approaches, so closing for now.

Issue #2224, opened 2025-12-15
Managed Keys feature reference should be removed


wslabosz-reply:
`Managed-Keys` Vault Enterprise abstraction for third party key management system is not present in OpenBao. For clarity, we should remove any code references to the feature existing in the codebase.

### additional context
Currently there's an ongoing development of the OpenBao's replacement of the feature called [External-keys](https://github.com/openbao/openbao/blob/main/website/content/docs/rfcs/external-keys.mdx), which will come with it's own custom support in the kms and main repository.

PR #2225, opened 2025-12-15
Remove "managed-keys" feature references

wslabosz-reply:
## Description
One of the features of the Vault Enterprise, "managed-keys" are going to be replaced by the OpenBao's own take on the matter - [external-keys](https://github.com/openbao/openbao/blob/main/website/content/docs/rfcs/external-keys.mdx), this PR removes any mentions of the feature that is not present in the OpenBao ecosystem.

## Rationale
Similar to: [removal of ent license references](https://github.com/openbao/openbao/pull/2045), this deletes obsolete code.

Resolves: #2224 

## Acknowledgements

       also allowed.


satoqz:
`sdk/logical/managed_key.go` might be a good (open-source) design reference for what the External Keys SystemView API could look like, but no reason to keep it in main :+1:

cipherboy:
@wslabosz-reply I would not remove them from API or SDK fwiw

wslabosz-reply:
@cipherboy 
sdk I might understand, what's blocking us for doing it in `api`?

cipherboy:
@wslabosz-reply I'd expect API client to continue to work across Vault and OpenBao, there's no point removing features specific to Vault pre-fork, unless they materially change and break afterwards. Especially if we want to drive adoption of OpenBao API, where users can download it without also downloading BUSL licensed proprietary code, we'll want to ensure API compatibility for open-source users publishing libraries &c which integrate with either Vault or OpenBao.

My 2c.?

wslabosz-reply:
@cipherboy 
I guess I get the point of the "broken compatibility", but I don't really understand a scenario where someone publishes a library that uses the managed keys sdk/api, and we assume that should work with OpenBao? Or are you saying we should allow for the these libraries to exists (and be usable), not really "working"?

wslabosz-reply:
@cipherboy 
I've reverted changes in `/api` and `/sdk` for now.

Issue #2214, opened 2025-12-11
Deadlock in mount tables


cipherboy:
I think now that we unlock early only if the namespace is new and perform the write _afterwards_, a subsequent storage write against the existing namespace could race against it, losing the second write?

_Originally posted by @satoqz in https://github.com/openbao/openbao/pull/2188#discussion_r2589388847_

----

Per conversation on Zulip, this seems to cause deadlock. 

[#openssf-openbao-wg-namespaces > Namespace Sealing / Read Scalability Release @ 💬](https://linuxfoundation.zulipchat.com/#narrow/channel/532995-openssf-openbao-wg-namespaces/topic/Namespace.20Sealing.20.2F.20Read.20Scalability.20Release/near/562524361)

I believe this should deadlock with the following [benchmark](https://github.com/openbao/benchmark-openbao) config:

```
test "namespace" "ns_create" {
    # Column Q
    weight = 33
    config {
        namespace_prefix = "benchmark-project-"
    }
}

test "mount" "kvv2_mount" {
    # Column R
    weight = 33
    config {
        mount_type = "secret"
        plugin = "kv-v2"
    }
}

test "mount" "jwt_mount" {
    # Column S
    weight = 34
    config {
        mount_type = "auth"
        plugin = "jwt"
    }
}
```

phil9909:
Can't reproduce with this steps:

```bash
~/openbao $ git rev-parse HEAD
b81bd96afcfeae87ebeb024d18ae59dcf826f3cf
~/openbao $ make dev && !!
```

```bash
~/benchmark-openbao $ go run . run -config /dev/stdin <<HCL
test "namespace" "ns_create" {
    # Column Q
    weight = 33
    config {
        namespace_prefix = "benchmark-project-"
    }
}

test "mount" "kvv2_mount" {
    # Column R
    weight = 33
    config {
        mount_type = "secret"
        plugin = "kv-v2"
    }
}

test "mount" "jwt_mount" {
    # Column S
    weight = 34
    config {
        mount_type = "auth"
        plugin = "jwt"
    }
}
HCL

2025-12-11T15:51:08.257+0100 [INFO]  vault-benchmark: setting up targets
2025-12-11T15:51:08.257+0100 [INFO]  vault-benchmark: starting benchmarks: duration=10s
2025-12-11T15:51:18.351+0100 [INFO]  vault-benchmark: benchmark complete
Target: http://127.0.0.1:8200
[...]

Issue #2210, opened 2025-12-09
ACL modification - per-namespace locking


cipherboy:
**Describe the bug**

ACL grabs a single global modification lock:

```
    // This is used to ensure that writes to the store (acl) or to the egp
    // path tree don't happen concurrently. We are okay reading stale data so
    // long as there aren't concurrent writes.
    modifyLock *sync.RWMutex
```

We should move this over to a lock set with per-namespace differentiation. 

**To Reproduce**

Run a [benchmark](https://github.com/openbao/benchmark-openbao/blob/main/docs/examples/sys/acl-write.hcl).

**Expected behavior**

**Environment:**
* OpenBao Server Version (retrieve with `bao status`):
* OpenBao CLI Version (retrieve with `bao version`):
* Server Operating System/Architecture:

OpenBao server configuration file(s):

```hcl
# Paste your OpenBao config here.
# Be sure to scrub any sensitive values
```

**Additional context**

satoqz:
Could lock sharding à la `locksutil` be a good enough solution compared to true per-namespace locking?

cipherboy:
Yeah that's what I was thinking. But we likely want to lock at the namespace level not at the policy level for behavior party (and to make lock acquisition in the namespace invalidation case easier).

I think the LRU cache is concurrent friendly presently.

PR #2226, opened 2025-12-17
Fix namespace locking, transaction issues around creation, deletion

cipherboy:
## Description

This completely reworks namespace store's creation and deletion logic:

- Creation now properly handles lock acquisition before transactions to avoid deadlocking (consistent transaction vs lock ordering -- as some transactions on PostgreSQL internally grab locks on the database server and thus can cause vacuous deadlocks). Cleanup is also properly handled now, leaving any partially created namespaces in a tainted (deletable) state.
- Deletion is now reworked to happen in a jobmanager since it is already asynchronous.
- Policy store parallelism improves creation speed
- Fix `WithTransaction` to actually use the created transaction. 
- Fix JobManager's metrics to be actually useful. 
- Ensure identity store's periodic func for OIDC issuer rotation operates on only the given namespace. 

## Rationale

Resolves: #2210
Resolves: #2214

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2226.txt
+++ b/changelog/2226.txt
@@ -1,0 +1,21 @@
+```release-note:bug
+core/namespaces: Fix deadlock on namespace creation, deletion due to transaction/lock ordering.
+```
+```release-note:improvement
+core/namespaces: Use JobManager for namespace deletion, decreasing lock contention.
+```
+```release-note:bug
+core/namespaces: Ensure namespace creation is interruptable, allowing namespace deletion for cleanup.
+```
+```release-note:improvement
+core/policies: Use per-namespace write lock, improving parallelism.
+```
+```release-note:bug
+sdk/logical: Use created transaction for WithTransaction callback.
+```
+```release-note:bug
+helper/jobmanager: Fix queue length metrics to report as gauges.
+```
+```release-note:bug
+core/identity: Ensure periodic func only operates on a single namespace at a time, decreasing storage contention.
+```


vault/policy_store.go
```diff
@@ -381,7 +391,7 @@ func (ps *PolicyStore) getBarrierView(ns *namespace.Namespace, _ PolicyType) Bar
 
 // GetPolicy is used to fetch the named policy
 func (ps *PolicyStore) GetPolicy(ctx context.Context, name string, policyType PolicyType) (*Policy, error) {
-	return ps.switchedGetPolicy(ctx, name, policyType, true)
+	return ps.switchedGetPolicy(ctx, name, policyType, false)
```
cipherboy: We could leave this as `true`, but I don't understand why it is set to true in the first place; at best it should be `RLock()` rather than `Lock()` as the latter is exclusive and unnecessary. The lock is also released before returning so there's no point to it except cache<->storage consistency, which seems unnecessary. 

`switchedGetPolicy(..., false)` is not called anywhere from what I can see; everyone comes in through `GetPolicy(...)`.

vault/namespace_store.go
```diff
@@ -914,65 +1007,31 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
...
-			ns.logger.Error("failed to delete namespace storage", "namespace", namespaceToDelete.Path, "error", err.Error())
-		}
-	}()
+	ns.deletionDispatcher.AddJob(ns.newNamespaceDeletionJob(parent, namespaceToDelete), parent.UUID)
 
 	return "in-progress", nil
 }
 
-func (ns *NamespaceStore) clearNamespaceResources(nsCtx context.Context, namespaceToDelete *namespace.Namespace) bool {
+func (ns *NamespaceStore) clearNamespaceResources(nsCtx context.Context, parent *namespace.Namespace, namespaceToDelete *namespace.Namespace) error {
```
wslabosz-reply, line 999: `parent` is unused in this method

vault/namespace_store.go
```diff
@@ -841,7 +929,12 @@ func (ns *NamespaceStore) ListNamespaces(ctx context.Context, includeParent bool
 	}
 	defer unlock()
 
-	return ns.namespacesByPath.List(parent.Path, includeParent, recursive)
+	entries, err := ns.namespacesByPath.List(parent.Path, includeParent, recursive, ns.creationDeletionMap)
+	if err != nil {
+		return nil, err
+	}
+
+	return entries, nil
```
wslabosz-reply: won't `return ns.namespacesByPath.List(parent.Path, includeParent, recursive, ns.creationDeletionMap)` be the equivalent?

vault/namespace_store.go
```diff
@@ -417,78 +447,107 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
...
+		ns.lock.Lock()
+		unlocked = false
 	}
 
-	// Finally commit the changes into storage.
-	if err := commit(); err != nil {
-		return err
+	// Finally, write the non-tainted namespace or perform our only write if
+	// we are modifying an existing entry.
+	if err := ns.writeNamespace(ctx, nil, entry); err != nil {
```
wslabosz-reply: we call `writeNamespace` 3 times, and every time it is with `nil` as a storage, which then retrieves the `ns.storage`, thinking that it would be simpler to pass with `ns.storage` for clarity? (per-ns-sealing will change that of course...)

vault/namespace_store.go
```diff
@@ -558,7 +611,7 @@ func (ns *NamespaceStore) initializeNamespace(ctx context.Context, storage logic
 		return err
 	}
 
-	if err := ns.createMounts(nsCtx, storage); err != nil {
+	if err := ns.createMounts(nsCtx, ns.storage); err != nil {
```
```suggestion
	return ns.createMounts(nsCtx, ns.storage)
```

wslabosz-reply: also `storage` in the context of this method is now unused, but we already pass `ns.storage` to this function, so I'd use the passed var.

vault/namespace_store.go
```diff
@@ -591,35 +644,47 @@ func (ns *NamespaceStore) createMounts(ctx context.Context, storage logical.Stor
...
+
+		for _, mount := range credentials.Entries {
+			if err := ns.core.persistAuth(ctx, txn, ns.core.auth, &mount.Local, mount.UUID); err != nil {
+				return fmt.Errorf("failed to persist auth mount (path=%v): %w", mount.Path, err)
+			}
 		}
+
+		return nil
+	}); err != nil {
+		return err
```
wslabosz-reply: you can also just return the `logical.WithTransaction()` outright

vault/namespace_store_tree.go
```diff
@@ -104,7 +104,7 @@ func (nt *namespaceTree) WalkPath(path string, predicate func(namespace *namespa
 // List lists child Namespace entries at a given path, optionally including the
 // namespace at the given path, optionally recursing down into all child
 // namespaces.
-func (nt *namespaceTree) List(path string, includeParent bool, recursive bool) ([]*namespace.Namespace, error) {
+func (nt *namespaceTree) List(path string, includeParent bool, recursive bool, tainted map[string]bool) ([]*namespace.Namespace, error) {
```
satoqz, line 107: I think this invades the namespace tree a little, thoughts on leaving it as is (plus the clone fix) and just iterating and updating the result on the call site?

Edit: I see Philipp shares this feeling, tbh I don't even think an extra helper method is warranted.

vault/identity_store_oidc.go
```diff
@@ -2020,12 +2022,23 @@ func (i *IdentityStore) oidcKeyRotation(ctx context.Context, s logical.Storage)
...
+	}
+
 	var nextRun time.Time
 	now := time.Now()
 
-	v, ok, err := i.oidcCache.Get(noNamespace, "nextRun")
+	v, ok, err := i.oidcCache.Get(ns, nextRunKey)
 	if err != nil {
-		i.Logger().Error("error reading oidc cache", "err", err)
+		i.Logger().Error("error reading oidc cache", "ns", ns.UUID, "err", err)
```
satoqz:
```suggestion
		i.Logger().Error("error reading oidc cache", "namespace", ns.Path, "err", err)
```

Likely more useful to users? (A few more cases below, too)

vault/namespace_store_tree.go
```diff
@@ -127,12 +127,23 @@ func (nt *namespaceTree) List(path string, includeParent bool, recursive bool) (
...
 	for idx := 0; idx < len(nodes); idx++ {
 		node = nodes[idx]
 		for _, child := range node.children {
-			entries = append(entries, child.entry.Clone(false))
+			entry := child.entry.Clone(false)
+			if value := tainted[entry.UUID]; value {
+				entry.Tainted = value
+			}
+
+			entries = append(entries, entry)
```
phil9909, line 146: This feels weird. How about moving this outside of list, into its own loop:

```diff
diff --git a/vault/namespace_store.go b/vault/namespace_store.go
index 065646a311..94ba9dcf68 100644
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -929,12 +929,26 @@ func (ns *NamespaceStore) ListNamespaces(ctx context.Context, includeParent bool
 	}
 	defer unlock()
 
-	entries, err := ns.namespacesByPath.List(parent.Path, includeParent, recursive, ns.creationDeletionMap)
+	entries, err := ns.namespacesByPath.List(parent.Path, includeParent, recursive)
 	if err != nil {
 		return nil, err
 	}
 
-	return entries, nil
+	return ns.overrideTaintStatus(entries), nil
+}
+
+func (ns *NamespaceStore) overrideTaintStatus(namespaces []*namespace.Namespace) []*namespace.Namespace {
+	if len(ns.creationDeletionMap) == 0 {
+		return namespaces
+	}
+
+	for i := range namespaces {
+		if value := ns.creationDeletionMap[namespaces[i].UUID]; value {
+			namespaces[i].Tainted = value
+		}
+	}
+
+	return namespaces
 }
 
 // taintNamespace is used to taint the namespace designated to be deleted
@@ -988,7 +1002,7 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
 	}
 
 	// checking whether namespace has child namespaces
-	childNS, err := ns.namespacesByPath.List(namespaceToDelete.Path, false, false, map[string]bool{})
+	childNS, err := ns.namespacesByPath.List(namespaceToDelete.Path, false, false)
 	if err != nil {
 		return "", err
 	}
diff --git a/vault/namespace_store_test.go b/vault/namespace_store_test.go
index ae257cd569..b0eea3442e 100644
--- a/vault/namespace_store_test.go
+++ b/vault/namespace_store_test.go
@@ -538,11 +538,11 @@ func TestNamespaceTree(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, beforeSize-1, tree.size)
 
-	entries, err := tree.List("", false, false, map[string]bool{})
+	entries, err := tree.List("", false, false)
 	require.NoError(t, err)
 	require.Equal(t, 2, len(entries))
 
-	entries, err = tree.List("", true, true, map[string]bool{})
+	entries, err = tree.List("", true, true)
 	require.NoError(t, err)
 	require.Equal(t, tree.size, len(entries))
 
diff --git a/vault/namespace_store_tree.go b/vault/namespace_store_tree.go
index 0b272e6af2..7ca9cbf64f 100644
--- a/vault/namespace_store_tree.go
+++ b/vault/namespace_store_tree.go
@@ -104,7 +104,7 @@ func (nt *namespaceTree) WalkPath(path string, predicate func(namespace *namespa
 // List lists child Namespace entries at a given path, optionally including the
 // namespace at the given path, optionally recursing down into all child
 // namespaces.
-func (nt *namespaceTree) List(path string, includeParent bool, recursive bool, tainted map[string]bool) ([]*namespace.Namespace, error) {
+func (nt *namespaceTree) List(path string, includeParent bool, recursive bool) ([]*namespace.Namespace, error) {
 	path = namespace.Canonicalize(path)
 	var segments []string
 	if path != "" {
@@ -127,23 +127,12 @@ func (nt *namespaceTree) List(path string, includeParent bool, recursive bool, t
 	var entries []*namespace.Namespace
 	if includeParent {
 		entries = make([]*namespace.Namespace, 0, len(node.children)+1)
-
-		entry := node.entry.Clone(false)
-		if value := tainted[entry.UUID]; value {
-			entry.Tainted = value
-		}
-
-		entries = append(entries, entry)
+		entries = append(entries, node.entry.Clone(false))
 	}
 	for idx := 0; idx < len(nodes); idx++ {
 		node = nodes[idx]
 		for _, child := range node.children {
-			entry := child.entry.Clone(false)
-			if value := tainted[entry.UUID]; value {
-				entry.Tainted = value
-			}
-
-			entries = append(entries, entry)
+			entries = append(entries, child.entry.Clone(false))
 			if recursive {
 				nodes = append(nodes, child)
 			}
```

vault/namespace_store.go
```diff
@@ -417,78 +447,107 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
...
+		// initial mounts can succeed.
+		entry.Tainted = false
 
-		storage = txn
-		commit = func() error { return txn.Commit(ctx) }
-	} else {
-		// While we don't have transactions to aid us, we still shouldn't
-		// leave partial namespaces lying around in memory.
-		defer cleanupFailed()
+		ns.creationDeletionMap[entry.UUID] = true
```
phil9909, line 479: This can run in parallel for different namespaces as locking is on per-namespace basis, right?

Go `map`s are not safe to use concurrently.

vault/namespace_store.go
```diff
@@ -392,22 +422,22 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
...
+	// 4. Finalize the namespace by re-writing the durable storage entry
+	//    as not tainted.
 	//
-	// In particular, the failure mode before was a panic in identity store
-	// when a namespace entry existed but the corresponding sys/ mount table
-	// entry wasn't created.
+	// Cleanup thus needs to handle partial deletion and remove any storage
+	// using the background context, not our primary context. We delete the
+	// storage entry while holding the lock, but Tainted status ensures
+	// nobody else attempts to mutate until we're done.
```
satoqz, line 441: Given cleanup happens with a longer-lived context than the requesting context, perhaps it should also be a background job like deletion? Any errors that occur in cleanup also don't affect the request's error anymore.

vault/namespace_store.go
```diff
@@ -591,35 +644,47 @@ func (ns *NamespaceStore) createMounts(ctx context.Context, storage logical.Stor
...
 	}
 
-	for _, credential := range credentials.Entries {
-		if err := ns.core.enableCredentialInternal(ctx, credential, MountTableNoUpdateStorage); err != nil {
+	// Grab all locks in the correct order. We hold these locks over updating
+	// both the in-memory mount table and the transaction.
+	ns.core.mountsLock.Lock()
+	ns.core.authLock.Lock()
+	defer ns.core.authLock.Unlock()
+	defer ns.core.mountsLock.Unlock()
```
satoqz, line 617: What's sub-optimal is that we keep holding these even during `persistMounts`/`persistAuth`, just because they read `ns.core.mounts`/`ns.core.auth`. If `mountInternal`/`enableCredential` returned the final mounted entry, we could likely build a minimal virtual copy of the mount table that has just the required entries and call persist on that, without a lock?

vault/namespace_store.go
```diff
@@ -628,11 +693,14 @@ func (ns *NamespaceStore) createMounts(ctx context.Context, storage logical.Stor
 // undoCreateMounts handles commit failure for in-memory resources. Note
 // that we do not modify storage here as our context has (usually) been
 // canceled and so we'd need activeContext or similar.
-func (ns *NamespaceStore) undoCreateMounts(nsCtx context.Context, namespaceToDelete *namespace.Namespace) {
+func (ns *NamespaceStore) undoCreateMounts(nsCtx context.Context, namespaceToDelete *namespace.Namespace) bool {
+	success := true
+
 	// clear auth mounts
 	authMountEntries, err := ns.core.auth.findAllNamespaceMounts(nsCtx)
```
satoqz, line 657: Needs a read lock, or alternatively, since we don't write to storage below, we could just write lock auth and mounts all over and have `disableCredentialInternalWithLock` etc. like you've done with the `enable` counterparts.

satoqz: Alternatively, in `undoCreateMounts` we technically already know the mounts we created and shouldn't need to query them. But this doesn't need to be optimized in this PR.

vault/namespace_store_tree.go
phil9909: > tbh I don't even think an extra helper method is warranted.

You are right. Only after I wrote the helper method, I noticed that only one call-site actually passes a non empty map :laughing:

vault/namespace_store.go
cipherboy: @phil9909 Isn't the lock still held until new line 523 or am I missing a code path?

I think you might be confusing it with the deferred release of the lock in the cleanup handler.

vault/namespace_store.go
cipherboy: @satoqz Do we not need the final full mount table as well? 🤔 

But that could be true and a candidate for the mount refactor logic we have outstanding. Trying to keep this minimal-ish while not forgetting anything I've discovered lol!

wslabosz-reply: definitely a mounts refactor candidate

vault/namespace_store.go
cipherboy: @wslabosz-reply This is true but will be not true for reasons made apparent in the future. :-)

wslabosz-reply: @cipherboy 
I'm not aware of a need for it, could you elaborate a bit?

vault/namespace_store_tree.go
wslabosz-reply: also in favor of modifying the result in the body of the accessing method (`ListNamespaces`) instead

cipherboy: @satoqz @phil9909 @wslabosz-reply I'm fine with this, but will note the performance trade-off we're missing here: re-iterating the response after return means an additional `O(n)` iteration versus an in-list-build `O(1)` lookup. As we look to scale up the number of namespaces (>10k), this becomes rather noticeable.

satoqz: Hmm, I guess I didn't expect it to be that much more expensive.

@cipherboy going one step back, when exactly is this true again? It's not immediately obvious to me:

```go
// A namespace may not be marked tainted (in memory) if it is in this
// list, so results need to be combined when namespaces are externally
// exposed.
```

cipherboy: @satoqz Namespace creations should be mark tainted until completed. It's a side-effect of adding them to the lists early.

satoqz: @cipherboy wait... so we pass the namespace to `mountInternalWithLock`/`enableCredentialInternalWithLock` via context, but that just stores the original pointer, never cloning it. When we then choose to update the `Tainted` field later on, it'll either be a data race or never see the update anyways, no? I think code that grabs `Tainted` from the namespace embedded in the mount needs a more live check.

Other than that initializing a namespace while keeping it tainted in namespace store memory until done did not cause problems based on quick local testing?

vault/namespace_store.go
cipherboy: @satoqz We could, I'm not sure what the impacts of this would be... I think it'd be fine, but I wonder if it'll be more confusing than blocking? :thinking: 

Given we use the long-lived context, maybe its better to do this.

satoqz: @cipherboy I agree it might be slightly confusing as you might not be able to immediately retry the operation. My case is based on the long-lived context though, yeah.

What about starting a background job, but trying our best to wait for its completion synchronously before responding to the request? Essentially, the job runs on the long-lived context, but we may wait for it to complete as long as our short-lived context hasn't expired?

phyrog: How would you make sure to send the response shortly before the request context expires? I don't think you can send a response when the request context is already done.

vault/namespace_store.go
phyrog: > locking is on per-namespace basis

The whole NamespaceStore is locked (per namespace locks would simplify some things though IMO).

vault/namespace_store_tree.go
phyrog: Alternative suggestion to decouple this more: pass a map function to `List()` that gets called for every entry. No additional O(n) iteration and no coupling of the namespace tree to tainting.

vault/namespace_store.go
satoqz: @phyrog I think you just wouldn't get to see that response then, no different from what would happen given the current state of the PR? It is a best effort synchronous delete that then falls back to an asynchronous delete if the request times out.

phyrog: Ah I thought you wanted to wait almost until the request times out and then quickly send a response along the lines of "tried waiting, but it took too long, will continue in the background".

vault/namespace_store.go
```diff
@@ -711,6 +790,7 @@ func (ns *NamespaceStore) GetNamespaceByLongestPrefix(ctx context.Context, path
 	combinedPath := ctxNs.Path + path
 	ns.lock.RLock()
 	prefix, entry, _ := ns.namespacesByPath.LongestPrefix(combinedPath)
+	entry.Tainted = entry.Tainted || ns.creationDeletionMap[entry.UUID]
```
Huy-Dinh-Reply, line 779: I'm guessing we would also want to clone the entry here or acquire a write lock if we're modifying the original pointer?
```suggestion
	prefix, item, _ := ns.namespacesByPath.LongestPrefix(combinedPath)
	entry := item.Clone(false)
	entry.Tainted = entry.Tainted || ns.creationDeletionMap[entry.UUID]
```
Edit: maybe it even makes sense to clone inside `LongestPrefix`, would also prevent `ResolveNamespaceForRequest` from returning the original pointer in the tree out.

vault/namespace_store.go
cipherboy: @wslabosz-reply We should probably discuss a unified style guide for this. I go back and forth which I feel is better -- the present code allows easy modifications (adding `fmt.Sprintf(...)`, more calls, &c), though I agree is strictly unused in this case.

vault/namespace_store_tree.go
cipherboy: @satoqz @phyrog Hmm...

My understanding here is that namespace store is re-entrant (locked creation -> unlocked mounting -> relocked finalization) for a reason. Certain default mounts (probably looking at `identity/` and `sys/`) are re-entrant and call into namespace store to resolve accessors/... and we're not purely doing initialization based on the context-passed namespace indefinitely.

(Aside: whether it should be or not, could be discussed -- perhaps we could fix this by relying on namespace-in-context more heavily. But it probably requires refactoring portions of shared backends like `sys/` and `identity/` to accept both context-namespace and the regular `NamespaceStore.List`'s results and merging them).

But for now, this necessitates both including the namespace into list (so we don't get failures to _find_ namespaces that we know to exist because it is part of initialization) and IMHO justifies that tainting is an integral part of listing of namespaces. (Now, whether that's at the NamespaceStore vs NamespaceTree level, I guess could be debated. But either way, we know we want to modify the results and we already have a local pointer copy so we might as well handle tainting here inline. My 2c).

Whether we want this to be a proper filter/modification function or not, I suppose we can discuss. I don't immediately see another use case for such, but if we get more than just tainted, we can definitely refactor.

---

> I think code that grabs Tainted from the namespace embedded in the mount needs a more live check.

`initializeNamespace(...)` explicitly calls `Clone()` on the namespace before attaching it to the context. This means that while a single namespace context is used for all newly created mounts, it is unique and not shared with the globally mutable copy actually stored in the NamespaceStore object.

I'm having a hard time actually coming up where `MountEntry.namespace.Tainted` is used tbh :thinking: 

Most places either come in from a list or from a namespace-in-context... But there's a few places using the pattern `namespace.ContextWithNamespace(ctx, src.Namespace)`:

- Expiration revocation --> this doesn't check tainted status because its presumed something is tainted anyways (either the mount and/or or the namespace)
- Initialization, which takes the tainted namespace. Again, this is during mount creation so I think it is safe to assume this namespace could potentially be tainted and ignore it.
- Reloading a namespace during invalidation; this again mostly calls cleanup+initialization so again, makes sense namespace could be tainted.

Or rather, I'm reasonably confident that tainted status persisting on namespace, while perhaps an oversight, isn't actually problematic.

Thoughts?

vault/namespace_store.go
cipherboy: @satoqz Isn't the current state of the PR that the deletion is asynchronous with deletion on creation failure? :thinking:

satoqz: @cipherboy yes, but the request could exceed `DefaultMaxRequestDuration` by switching to the long-lived context on the request's Goroutine.

vault/namespace_store.go
cipherboy: @satoqz Let me know if this adjustment make sense or if it was too heavy-handed. Part of the refactor should be better locking semantics for the mount table... :thinking:

vault/namespace_store.go
cipherboy: Thanks @Huy-Dinh-Reply, good catch!

vault/namespace_store.go
cipherboy: I'm going to mark this as resolved; I think the global locking is fine.

vault/namespace_store_tree.go
cipherboy: I had a longer discussion here: https://github.com/openbao/openbao/pull/2226#discussion_r2691407133 

Feel free to add comments there.

vault/namespace_store.go
wslabosz-reply: @cipherboy definitely all down for style guide (if it's gonna be enforceable at sufficient level). I am of stance that, it's not a problem if you gonna modify it and it will result in higher addition numbers, I definitely tend to yagni, but I'm also agreeable enough to just accept "I think it's better this way" as a response /my 2c.

vault/namespace_store.go
cipherboy: @satoqz Oh, I see now. We use `defer`, not `go`. :facepalm: 

Yes, I agree, proper queuing would be better than what we have now. Will adjust. :-D

cipherboy: @satoqz Revised, LMKWYT!

vault/identity_store_oidc.go
```diff
@@ -2020,12 +2022,23 @@ func (i *IdentityStore) oidcKeyRotation(ctx context.Context, s logical.Storage)
...
+	ns, err := namespace.FromContext(ctx)
+	if err != nil {
+		i.Logger().Error("identity store periodic function called without namespace", "error", err)
+		return
+	}
+
+	if ns.Tainted {
+		i.logger.Info("skipping OIDC periodic function for tainted namespace", "namespace", ns.Path)
+		return
+	}
```
satoqz, line 2034: > I'm having a hard time actually coming up where MountEntry.namespace.Tainted is used tbh 🤔

@cipherboy you're using it right here 😄

This will never see the update to the `Tainted` field written by `DeleteNamespace` unless the namespace was loaded from storage in a tainted state. Conversely, if we made the namespace store's namespace and the mount table's namespace one and the same pointer, we'd need to guard around races in the mount table whenever we update a namespace's fields.

vault/namespace_store_tree.go
satoqz: Yeah, I think you are right, I had missed the clone in `initializeNamespace()`. I like @phyrog's suggestion though will not nitpick this any further :)

satoqz:
https://github.com/openbao/openbao/blob/cefeb84b27b69e90bd0870e2b0f9783415dd7404/vault/namespace_store.go#L389-L393

Should we turn this into a late check inside of `createMounts`, or perhaps check again later on with the mount table locked? The equivalent codepaths for mounting secrets and auth engines always lock the mount table before calling `MountConflict()`, but here I believe we're risking a stale read. I think either way the current section lacks a read lock into the router.

vault/identity_store_oidc.go
cipherboy: @satoqz Ahhh I see. I had missed that Rollback manager did not refresh the namespace. :thinking: 

That seems a little unfortunate... Not that custom metadata is really _utilized_ but it could be out-of-date as well as Tainted status.

I almost wonder if every call to `SetNamespace(...)` should push it to the corresponding mount table entries? :thinking: 

Thoughts?

cipherboy:
> I think either way the current section lacks a read lock into the router.

I've added this one.

I don't think the late check in checkMounts is ideal as we've already started populating namespace objects in memory... :thinking: 

And lock ordering is fine I believe with this.

cipherboy:
@satoqz Update, LMKWYT!

vault/namespace_store.go
```diff
@@ -860,11 +938,12 @@ func (ns *NamespaceStore) taintNamespace(ctx context.Context, namespaceToTaint *
...
 
 	nsCopy := namespaceToTaint.Clone(true /* preserve unlock */)
-	if err := ns.writeNamespace(ctx, nil, nsCopy); err != nil {
+	if err := ns.writeNamespace(ctx, ns.storage, nsCopy); err != nil {
 		return fmt.Errorf("failed to persist namespace taint: %w", err)
 	}
 
-	return nil
+	// Push the update to all mounts.
+	return ns.pushToMounts(ctx, namespaceToTaint.Clone(false))
```
satoqz, line 942: I think we can skip this if the namespace doesn't exist yet and save some time on namespace creation.

vault/namespace_store.go
```diff
@@ -660,10 +690,43 @@ func (ns *NamespaceStore) undoCreateMounts(nsCtx context.Context, namespaceToDel
...
+	defer ns.core.authLock.Unlock()
+
+	authMountEntries, err := ns.core.auth.findAllNamespaceMounts(nsCtx)
+	if err != nil {
+		return err
+	}
+
+	for _, mount := range authMountEntries {
+		mount.namespace = entry
+	}
```
satoqz: Could we just operate on the table directly without creating a new temporary list?

vault/namespace_store.go
cipherboy: @satoqz But wasn't the issue that on namespace creation, we mark tainted at first? I think we need to do this on creation.

vault/namespace_store.go
cipherboy: @satoqz Ah, and condition around `namespace`? That's better, I like that. Thanks!

satoqz: I think we can just condition around the namespace ID written to the mount entry next to the namespace, no need to read the live namespace field at all.

vault/namespace_store.go
satoqz: Ah you are right. Again it would be nicer if we kept the subset of the mount table (that we just created anyways) around for a second in that case so we don't need to search the whole thing.

wslabosz-reply:
@JanMa 
could you take a look at this?

cipherboy:
@JanMa I've fixed merge conflicts.

JanMa (approved):
LGTM ✔️ 

Some feedback: I found this quite difficult to review because the PR fixes multiple issues at one. IMO splitting this in 3 PRs would have made review a lot easier (and faster, sorry for the delay) for me. 

@cipherboy If you don't mind, can you please clean up the commit history a bit, to at least fixup bbec7f0d89 and then merge this with a rebase instead of squash. I think it's valuable to have the individual commits on `main`

cipherboy:
@JanMa @satoqz I've pushed a cleaned up rebase proposal, let me know what you think.

PR #2229, opened 2025-12-17
Fix some more unhandled errors

phyrog:
## Description

Fixes (or ignores) some more unhandled errors, this time focusing on non-test files under `command/` and `backend.Setup()`.

## Rationale

Part of #1574 

## Acknowledgements

       also allowed.

### DIFF
--- a/command/agent/config/config.go
+++ b/command/agent/config/config.go
@@ -475,12 +475,14 @@ func LoadConfig(path string) (*Config, error) {
 }
 
 // LoadConfigDir loads the configuration at the given path if it's a directory
-func LoadConfigDir(dir string) (*Config, error) {
+func LoadConfigDir(dir string) (cfg *Config, err error) {
 	f, err := os.Open(dir)
 	if err != nil {
 		return nil, err
 	}
-	defer f.Close()
+	defer func() {
+		err = errors.Join(err, f.Close())
+	}()
 
 	fi, err := f.Stat()
 	if err != nil {
--- a/command/server.go
+++ b/command/server.go
@@ -620,7 +620,11 @@ func (c *ServerCommand) runRecoveryMode() int {
 			ErrorLog:          c.logger.StandardLogger(nil),
 		}
 
-		go server.Serve(ln.Listener)
+		go func(ln net.Listener) {
+			if err := server.Serve(ln); err != nil && !errors.Is(err, http.ErrServerClosed) {
+				c.UI.Error(fmt.Sprintf("HTTP server (listening on %s) exited with error: %v", ln.Addr().String(), err))
+			}
+		}(ln.Listener)
 	}
 
 	if sealConfigError != nil {
@@ -1538,7 +1542,7 @@ func (c *ServerCommand) Run(args []string) int {
 
 		case <-c.SigUSR2Ch:
 			logWriter := c.logger.StandardWriter(&hclog.StandardLoggerOptions{})
-			pprof.Lookup("goroutine").WriteTo(logWriter, 2)
+			_ = pprof.Lookup("goroutine").WriteTo(logWriter, 2)
 
 			if api.ReadBaoVariable("BAO_STACKTRACE_WRITE_TO_FILE") != "" {
 				c.logger.Info("Writing stacktrace to file")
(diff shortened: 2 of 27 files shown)


builtin/logical/openldap/backend_test.go
```diff
@@ -45,7 +46,9 @@ func getBackend(throwsErr bool) (*backend, logical.Storage) {
 	}
 
 	b := Backend(&fakeLdapClient{throwErrs: throwsErr})
-	b.Setup(context.Background(), config)
+	if err := b.Setup(context.Background(), config); err != nil {
```
wslabosz-reply:
```suggestion
	if err := b.Setup(t.Context(), config); err != nil {
```

command/server/listener_tcp.go
```diff
@@ -92,7 +93,7 @@ func (ln TCPKeepAliveListener) Accept() (c net.Conn, err error) {
 	if err != nil {
 		return c, err
 	}
-	tc.SetKeepAlive(true)
-	tc.SetKeepAlivePeriod(3 * time.Minute)
-	return tc, nil
+	err = errors.Join(err, tc.SetKeepAlive(true))
+	err = errors.Join(err, tc.SetKeepAlivePeriod(3*time.Minute))
+	return tc, err
```
wslabosz-reply:
```suggestion
	err = errors.Join(err, tc.SetKeepAlive(true), tc.SetKeepAlivePeriod(3*time.Minute))
	return tc, err
```

command/agent/config/config.go
```diff
@@ -475,12 +475,14 @@ func LoadConfig(path string) (*Config, error) {
...
 // LoadConfigDir loads the configuration at the given path if it's a directory
-func LoadConfigDir(dir string) (*Config, error) {
+func LoadConfigDir(dir string) (cfg *Config, err error) {
 	f, err := os.Open(dir)
 	if err != nil {
 		return nil, err
 	}
-	defer f.Close()
+	defer func() {
+		err = errors.Join(err, f.Close())
```
cipherboy, line 484: 🤔 I thought we decided to ignore `.Close()` failures somewhere in one of the past conversations? Am I misremembering?

command/server.go
```diff
@@ -1541,7 +1545,7 @@ func (c *ServerCommand) Run(args []string) int {
 
 		case <-c.SigUSR2Ch:
 			logWriter := c.logger.StandardWriter(&hclog.StandardLoggerOptions{})
-			pprof.Lookup("goroutine").WriteTo(logWriter, 2)
+			_ = pprof.Lookup("goroutine").WriteTo(logWriter, 2)
```
cipherboy, line 1545: Curious your thoughts on the `_` vs `//nolint:errcheck` approach?

command/agent/config/config.go
wslabosz-reply: I've rollbacked on it, as you cannot ignore specific .Close() error signatures, just all, and some of them might be actually valid....

cipherboy: Gotcha, that's fine then.

command/server.go
phyrog: I think generally I tend to use `_` if I don't care about an error, and `//nolint` if there is nothing that can be done about the error (usually with a comment explaining). I haven't formalized this though, and it's probably not always consistent.

command/server.go
```diff
@@ -2953,7 +2964,11 @@ func startHttpServers(c *ServerCommand, core *vault.Core, config *server.Config,
 			continue
 		}
 
-		go server.Serve(ln.Listener)
+		go func(ln net.Listener) {
+			if err := server.Serve(ln); err != nil && !errors.Is(err, http.ErrServerClosed) {
+				c.UI.Error(fmt.Sprintf("HTTP server (listening on %s) exited with error: %v", ln.Addr().String(), err))
+			}
+		}(ln.Listener)
```
satoqz, line 2971: @phyrog this causes alarming logs on shutdown because listeners are closed directly instead of via `http.Server.Close`.

```
==> OpenBao shutdown triggered
HTTP server (listening on 127.0.0.1:8200) exited with error: accept tcp 127.0.0.1:8200: use of closed network connection
```

PR #1760, opened 2025-09-02
RFC control groups for access control

suprjinx:
This PR adds an RFC document for a control group implementation

Relates #1669

Issue #1669, opened 2025-08-09
Add control group authorization workflows approval process for secrets


nolon:
Simple use case: 

create a Intermediate CA-Certificate that needs an approval by another entity.

Basically the shamir shared keys unseal approach but for secrets not for startup

cipherboy:
Hey @nolon -- I'm not sure how control groups would interact with some of the new features we've added (like server-side list filtering, inline authentication and other things). However, if someone wants to do an investigation, propose an RFC, and implement it, I suspect it'd be interesting.

I'm curious if there's alternatives as well. To be clear, this isn't about a "sharmir shared key" approach: this is a multi-party, human-in-the-loop _approval_ flow. There is nothing cryptographic backing this. Could we make a cryptographic backing for certain areas? Perhaps and maybe that'd be stronger, but control groups (as implemented upstream) work with just about any API endpoint and thus are weaker. 

As @suprjinx reminds me, we had some discussion earlier around CEL policies possibly augmenting this to add human-in-the loop based reviews, and also they should be able to trigger control group enforcement conditionally on paths (if added anyways).

My 2c.

suprjinx:
I'm going to take a stab at writing up an RFC for this.

PR #2241, opened 2025-12-19
Control Groups

suprjinx:
## Description

This PR implements control-group policy factors for protected paths. A control group can be specified for a path and operation in the ACL policy. When a requesting user encounters this, they receive a wrapped response instead of the data at the path. The access token for the wrapped response can be used to authorize the request. Once authorized, the original user can complete the request by unwrapping with the wrapping token.

For more information, see the Control Groups RFC, merged in PR #1760 

## Rationale

Relates: #1669
Relates: #1760 

## Acknowledgements

 - [ x] By contributing this change, I certify I have not used generative AI
 - [x ] By contributing this change, I certify I have signed-off on the
       also allowed.


suprjinx:
@cipherboy Do you think this is the right approach here? In particular, storing control-group authorizations in the metadata of the token is the path of least resistance but may not be correct. If you think ok, I can "upgrade" to a full PR.  Thanks!

cipherboy:
@suprjinx Hmm, looks like the rebase was a little messy. If you clean that up, I can start looking later today or tomorrow!

suprjinx:
Merge/rebase is not working out so applying a clean diff as one commit

suprjinx:
reopening -- rebased to main and applied work as single diff

suprjinx:
fixed the rebase issue @cipherboy

vault/request_handling.go
```diff
@@ -1018,7 +1021,23 @@ func (c *Core) handleCancelableRequest(ctx context.Context, req *logical.Request
...
-		cubbyResp, cubbyErr := c.wrapInCubbyhole(ctx, req, resp, auth)
+
+		extraData := map[string]string{}
+		// For controlgroup, store the original request and control group details with the cubbyhole data
+		if auth.PolicyResults != nil && auth.PolicyResults.ControlGroup != nil {
+			reqJson, err := jsonutil.EncodeJSON(req)
+			if err != nil {
+				return resp, err
+			}
+			extraData["request"] = string(reqJson)
```
cipherboy: `extraData` going to `logical.TokenEntry.Meta` is slightly entertaining, as my understanding is that it'll always be plaintext audited. 

In the case of Control Group, I'm not sure it matters (auditing might be slightly beneficial so you know what the control group policy is), but I'm not sure if the request should be; the request will be stored anyways.

vault/token_store.go
```diff
@@ -2671,6 +2704,64 @@ func (ts *TokenStore) handleTidy(ctx context.Context, req *logical.Request, data
 	return logical.RespondWithStatusCode(resp, req, http.StatusAccepted)
 }
 
+// handleUpdateApproveAccessor handles the auth/token/approve-accessor path for authorizing
+// a wrapped response token governed by a control-group policy.
+func (ts *TokenStore) handleUpdateApproveAccessor(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+	accessor := data.Get("accessor").(string)
```
cipherboy: Sorry, I just realized this. We're slightly out of spec with Vault Enterprise here, right? 

https://developer.hashicorp.com/vault/api-docs/system/control-group  --> `sys/control-group/authorize` vs `auth/token/approve-accessor`?

cipherboy (commented):
Proof of concept is looking fairly good code wise, I still need to play with it and try it out!

Some small thoughts here.

vault/request_handling.go
suprjinx: I was thinking that the approver would want to see the request, and this is where it could be easily seen with just the accessor token, but let me know if there's a better approach!

vault/token_store.go
suprjinx: I guess so! I was trying to align with the existing `lookup-accessor` endpoint used in response wrapping. You think move to a `sys/control-group` api?

vault/request_handling.go
cipherboy: @suprjinx I'd write request information to `InternalMeta`.

suprjinx: Edit: moving to InternalMeta

vault/token_store.go
suprjinx: I'm moving this API to match Vault Enterprise

luke-clifton:
I left a comment on the RFC PR, but will post it here again for visibility.

But, the no-self-authorizion rule does not exist upstream.

https://github.com/openbao/openbao/pull/1760#pullrequestreview-4242000225

EDIT:

I actually see that this implementation allows self-authorization, so, cool. Though, I can't seem to succesfully unwrap afterwards.

<details><summary>Output</summary>
<pre>
~/tmp/cgdemo$ ~/src/openbao/bin/bao read kv/some
Key                 Value
---                 -----
refresh_interval    768h
a                   b
~/tmp/cgdemo$ ~/src/openbao/bin/bao login -token-only -method=userpass username=alice password=alice123
s.wCy5E4H1WeaB8yRG7FJJomex
~/tmp/cgdemo$ ~/src/openbao/bin/bao login -token-only -method=userpass username=bob password=bob123
s.ni2yrZqbxPNU3ePt3OttUTQT
~/tmp/cgdemo$ BAO_FORMAT=json BAO_TOKEN=s.wCy5E4H1WeaB8yRG7FJJomex ~/src/openbao/bin/bao read kv/some
{
  "request_id": "",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": null,
  "warnings": null,
  "wrap_info": {
    "token": "s.TbnSoGQQwcKKSNMtNb2ktLfr",
    "accessor": "f6ICkXZVKfXTYjXNEu90Iqws",
    "ttl": 600,
    "creation_time": "2026-05-07T16:36:07.093308+08:00",
    "creation_path": "kv/some",
    "wrapped_accessor": ""
  }
}
~/tmp/cgdemo$ BAO_TOKEN=s.ni2yrZqbxPNU3ePt3OttUTQT ~/src/openbao/bin/bao write sys/control-group/authorize accessor=f6ICkXZVKfXTYjXNEu90Iqws
Key         Value
---         -----
approved    true
~/tmp/cgdemo$ BAO_FORMAT=json BAO_TOKEN=s.wCy5E4H1WeaB8yRG7FJJomex ~/src/openbao/bin/bao unwrap s.TbnSoGQQwcKKSNMtNb2ktLfr
{
  "request_id": "41efa70d-59ff-01fb-f88d-024343a008b7",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": null,
  "warnings": null
}
</pre></details>

cipherboy:
Hey @luke-clifton!

>  But, the no-self-authorizion rule does not exist upstream.

In general, we've been happy deviating from upstream's proprietary extensions where it makes sense. Given the intent of control groups is to enforce a second review, self-review seems a little excessive. I wonder if we could have a boolean flag (`control_group_self_review = ...`) which defaults to false but could be enabled on the Policy on a case-by-case basis? That'd give us a more secure baseline than Vault Enterprise.

Thoughts @suprjinx?

luke-clifton:
From an enterprise perspective, self review can make sense. In my organisation, we just need to ensure two qualified people have looked at it, so we set `approvals = 2`.

Yes, it's tedious approving your own request if you are one of the qualified people! Maybe, auto-self-approval would be nice 🤔

suprjinx:
Thanks for the feedback, @luke-clifton! I'll take a look at the unwrapping issue. I think the "no self authorization" rule came from the way my org uses control groups with Vault Enterprise (they enforce with a Sentinel policy). But definitely makes sense to add an attribute to the policy declaration for that instead of making it mandatory.

suprjinx:
@cipherboy @luke-clifton I fixed the error with unwrapping after authorized, and added a new SelfAuthorizationAllowed attribute to the ControlGroup struct -- default false. Ready for another look!

luke-clifton:
Hi, just got the time to test this. I can indeed  unwrap now! But, I can also self-authorise. This is regardless of if I include `self_auth_allowed = false` or `true` just beneath the `ttl` paramter of the control group.

```
#! /usr/bin/env bash
set -x

bao version

bao auth enable userpass
bao secrets enable -path=kv kv-v2
bao kv put kv/some-secret key=value

bao write auth/userpass/users/alice password=alice123
bao write auth/userpass/users/bob password=bob123

up_accessor=$(bao read -field accessor sys/auth/userpass)

bao policy write thepolicy - <<EOF
path "sys/control-group/*" {
    capabilities = ["read", "create", "update", "list"]
}

path "kv/*" {
    capabilities = ["read"]
    control_group = {
        ttl = "600s"
        factor "admin" {
            name = "admin-approval"
            controlled_capabilities = ["read"]
            identity {
                group_names = ["admin"]
                approvals = 1
            }
        }
    }
}
EOF

bao write identity/entity-alias name=alice mount_accessor=$up_accessor
bao write identity/entity-alias name=bob mount_accessor=$up_accessor
alice=$(bao write -field=id identity/lookup/entity alias_mount_accessor=$up_accessor alias_name=alice)
bob=$(bao write -field=id identity/lookup/entity alias_mount_accessor=$up_accessor alias_name=bob)


bao write identity/group/name/admin member_entity_ids=$alice,$bob policies=thepolicy

ALICE_TOKEN=$(bao login -token-only -method=userpass username=alice password=alice123)

CG_RESPONSE=$(BAO_FORMAT=json BAO_TOKEN=$ALICE_TOKEN bao kv get kv/some-secret)
CG_TOKEN=$(jq .wrap_info.token -r <<< "$CG_RESPONSE")
CG_ACCESSOR=$(jq .wrap_info.accessor -r <<< "$CG_RESPONSE")

BAO_TOKEN="$ALICE_TOKEN" bao write sys/control-group/authorize accessor=$CG_ACCESSOR
BAO_TOKEN="$ALICE_TOKEN" bao unwrap "$CG_TOKEN"
```

Which ultimately ends up allowing Alice to unwrap.

```
+ bao unwrap s.oKMyixIcEIMh8z4y6xZTEekN
Key         Value
---         -----
data        map[key:value]
metadata    map[created_time:2026-05-18T01:16:57.146977Z custom_metadata:<nil> deletion_time: destroyed:false version:4]
```

luke-clifton:
Additionally, there seems to be some confusion about which namespace things are being run in, at least when using `UnsafeCrossNamespaceIdentity`.

vault/request_handling.go
```diff
@@ -879,6 +887,31 @@ func (c *Core) handleCancelableRequest(ctx context.Context, req *logical.Request
...
+					return nil, err
+				}
+				if deferredReq != nil {
+					valid, err = c.validateControlGroup(ctx, tokenEntry, deferredReq.Operation)
+					if err != nil {
+						return nil, err
+					}
+					if valid {
+						deferredReq.ForwardedFrom = forwardedFromDeferral
+						return c.handleCancelableRequest(ctx, deferredReq)
```
luke-clifton: Yeah, maybe this is where it's using the wrong context?  `c.handleCancelableRequest()` is passed in the `ctx` for _this_ request, not the deferred request?

suprjinx:
@luke-clifton thanks for spotting -- wrt self-authorization problem, I did some testing and in fact your whole control group factor was just ignored by the validate function since it expects a slice of factors rather than one (alice token could unwrap even before self authorization :-)). But with a slice of factors it in fact works currently. I'm adding support for the single factor format of your policy.

Edit: in fact, the upstream only shows examples with "name in key" format, so I'll drop the name attribute and require it in the key.

Edit 2: changed deferred request to use NamespaceID from wrapping token entry in call to `handleCancelableRequest`

vault/request_handling.go
suprjinx: switched to use ctx with NamespaceID from the wrapping token

suprjinx:
@luke-clifton @cipherboy calling for comments, previous are addressed!

luke-clifton:
This passes all the tests I have on my end!

My only remaining comment is on the default value of `self_auth_allowed`. I'm glad that we have the option to go either way, but, it would be nice to be able to just migrate over existing policies from Vault, and have them do the right thing.

I'm not going to complain though! Really glad to see this, and would be happy to have it merged either way.

vault/request_handling.go
```diff
@@ -879,6 +887,36 @@ func (c *Core) handleCancelableRequest(ctx context.Context, req *logical.Request
 			if !valid {
 				return nil, consts.ErrInvalidWrappingToken
 			}
+			// Handle deferred requests for approved control-group unwrap
+			if req.Path == "sys/wrapping/unwrap" {
+				token, _ := req.Data["token"].(string)
+				if token == "" {
+					token = req.ClientToken
+				}
+				tokenEntry, err := c.tokenStore.lookupTainted(ctx, token)
```
luke-clifton, line 896: `lookupTainted` can return `(nil, nil)`. But there is no check for `tokenEntry` being `nil`.

luke-clifton: https://github.com/openbao/openbao/blob/ca892ba62340b0d093ab17d3de9bc0b28537326a/vault/logical_system.go#L3152-L3158

suprjinx: added check for nil token entry

sdk/helper/hclutil/hcl.go
```diff
@@ -13,6 +13,28 @@ import (
 	jsonParser "github.com/hashicorp/hcl/json/parser"
 )
 
+// WhenHCLKeyPresent will execute the provided function with each matching child node as argument
+func WhenHCLKeyPresent(node ast.Node, key string, fn func(*ast.ObjectItem) error) error {
```
cipherboy, line 17: Random thought: Wonder if this would be useful as a more general pattern for config stanza parsing elsewhere (`command/server` being the big one). Given parent and key, apply parse function to each match. 

Absolutely no action here, but perhaps @phyrog has opinions as our resident pattern/refactor expert.

vault/policy/policy.go
```diff
@@ -162,6 +163,27 @@ type ACLPermissions struct {
...
+type ControlGroup struct {
+	TTL                      time.Duration        `hcl:"-"`
+	TTLHCL                   interface{}          `hcl:"ttl"`
+	Factors                  []ControlGroupFactor `hcl:"-"`
+	SelfAuthorizationAllowed bool                 `hcl:"self_auth_allowed"`
+}
+
+type ControlGroupFactor struct {
+	Name                      string              `hcl:"-"`
+	ControlledCapabilitiesHCL []string            `hcl:"controlled_capabilities"`
```
cipherboy, line 178: No action required here either, but I'm entertained this file doesn't follow the pattern elsewhere of `Raw` as the suffix for parsed attributes rather than `HCL`. 😆

vault/policy/acl_test.go
```diff
@@ -38,6 +38,69 @@ func testNewACL(t *testing.T, ns *namespace.Namespace) {
...
+	controlGroupHCL := `
+path "test/control_group" {
+	capabilities = ["create", "list"]
+	control_group = {
+		ttl = "48m"
+		factor "admin-approval" {
+		        controlled_capabilities = ["create"]
+			identity = {
+				group_names = ["admin"]
+				approvals = 1
```
cipherboy:
```suggestion
			controlled_capabilities = ["create"]
			identity = {
				group_names = ["admin"]
				approvals = 1
```

vault/request_handling.go
```diff
@@ -879,6 +887,39 @@ func (c *Core) handleCancelableRequest(ctx context.Context, req *logical.Request
...
+				deferredReq, err := c.getRequestFromTokenEntry(ctx, tokenEntry)
+				if err != nil && err != ErrDeferredRequestNotFound {
+					return nil, err
+				}
+				if deferredReq != nil {
+					valid, err = c.validateControlGroup(ctx, tokenEntry, deferredReq.Operation)
+					if err != nil {
+						return nil, err
+					}
+					if valid {
```
cipherboy, line 912: I think we should forward to the active node here if we're on a standby node. Otherwise we risk double request execution as we (should) always incur a write anyways with this.

Something like:

```go
if c.Standby() {
    return nil, logical.ErrPerfStandbyPleaseForward
}
```

vault/request_handling.go
```diff
@@ -1240,7 +1322,13 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
 	}
 
 	// Route the request
-	resp, routeErr := c.doRouting(ctx, req)
+	var routeErr error
```
cipherboy: @suprjinx Do you mind refactoring this into another method? I think the same call needs to happen in `handleLoginRequest(...)` ideally: plugins could differentiate between auth/non-authed users in login requests and behave differently, and thus should be ControlGroup-able.

cipherboy (commented):
@suprjinx Some random thoughts from last time I looked, haven't yet finished reviewing 😆

vault/policy/policy.go
suprjinx: The other structs in this file use the HCL suffix for the raw values (eg, PathRules), so I'll leave it to be consistent with the proximate examples.

vault/request_handling.go
suprjinx: added this here

vault/control_group.go
```diff
@@ -0,0 +1,444 @@
...
+	out, err := c.tokenStore.lookupInternal(ctx, aEntry.TokenID, false, true)
+	if err != nil {
+		return logical.ErrorResponse(err.Error()), logical.ErrInvalidRequest
+	}
+
+	if out == nil {
+		return logical.ErrorResponse("bad token"), logical.ErrPermissionDenied
+	}
+
+	approved, err := c.validateControlGroup(ctx, out, logical.ReadOperation)
```
cipherboy: With this versus the implementation of `validateControlGroup`, I'm confused how `logical.ReadOperation` is chosen? 

I think it implies that the operation is only ever a read operation that is authorized, which is confusing; once created, I'd assume the control group applies indefinitely, as the response wrapping token is formally bound to the given request type? 🤔 

Otherwise, if the control group doesn't apply on read operations (only writes) don't we get an approved here even though we shouldn't? 🤔

cipherboy: (`vault/request_handling.go:908:` handles this correctly as far as I'm concerned)

suprjinx: Updated to match `vault/request_handling.go:908`

vault/request_handling.go
suprjinx: added new function for this, `doRoutingIfApproved`

cipherboy (approved):
Perfect, thanks for the revisions @suprjinx!

Let's merge this and we can keep testing this in main for v2.7.0. 🥳 

Appreciate your patience and contributions again!!

cipherboy:
@suprjinx (And, we'll add a changelog with the docs which can come later.)

Issue #1809, opened 2025-09-12
SystemBackend could be per-namespace


cipherboy:
Currently sys backend is a singleton; I think this might be a holdover of Vault Enterprise and we may consider if it makes sense to split it:

1. We can implement per-namespace versions of certain handlers. Thinking the future policy system for one, which has no need to be cross-namespace. Potentially other system-internal interface could also be SystemBackend objects rather than be singletons.
2. ~I believe the `SystemBackend.db` field is unused entirely today, so can be removed.~

Many things in system will still be global and make sense to be on Core rather than the SystemBackend, but it might give us some tools to improve performance. For example, PolicyStore, with a lot of work, could be split which would give us a lock-per-namespace rather than a single lock globally thus decreasing lock contention on writes. 

And likely other reasons I'm missing.

wslabosz-reply:
@cipherboy 
I will take a look at this one!

satoqz:
> Currently sys backend is a singleton

Turns out it is not, though I also expected it to be:

https://github.com/openbao/openbao/blob/623ab9d97f8e7ff8fb975466c22970947f6040be/vault/core.go#L1313-L1323

Agreed on the performance aspect (e.g., policy store, per-namespace locking &c), but wouldn't the performant choice w.r.t. `SystemBackend` be to _make it a singleton_?

In fact, with this patch:

```diff
diff --git a/vault/core.go b/vault/core.go
index 643c795783..d247940ad5 100644
--- a/vault/core.go
+++ b/vault/core.go
@@ -1320,7 +1320,12 @@ func (c *Core) configureLogicalBackends(backends map[string]logical.Factory, log
 		}
 		return b, nil
 	}
-	logicalBackends[mountTypeNSSystem] = logicalBackends[mountTypeSystem]
+	logicalBackends[mountTypeNSSystem] = func(ctx context.Context, bc *logical.BackendConfig) (logical.Backend, error) {
+		if c.systemBackend != nil {
+			return c.systemBackend, nil
+		}
+		return nil, errors.New("system backend does not exist")
+	}
 
 	// Identity
 	logicalBackends[mountTypeIdentity] = func(ctx context.Context, config *logical.BackendConfig) (logical.Backend, error) {
```

the memory usage of an idle `bao server -dev` with 1000 empty namespaces drops from 1.3 GB to 900 MB on my system.

PR #2290, opened 2026-01-09
Remove memdb from `SystemBackend` struct

wslabosz-reply:
## Description
As per #1809 this PR removes the `memdb` from `SystemBackend` struct.

This PR doesn't address the following note from issue description:
`We can implement per-namespace versions of certain handlers. Thinking the future policy system for one, which has no need to be cross-namespace. Potentially other system-internal interface could also be SystemBackend objects rather than be singletons.`

requirements and details should be further refined in (possibly separate) issue.

Before this PR `SystemBackend` was already scoped to the namespace (although there are no specific namespace handlers defined).

## Rationale
`memdb` field on the `SystemBackend` struct was entirely unused, so was `administrativeNamespace` property also removed with this PR.

Resolves part of: #1809

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/core.go
+++ b/vault/core.go
@@ -1273,7 +1273,6 @@ func (c *Core) configureCredentialsBackends(backends map[string]logical.Factory,
 	credentialBackends[mountTypeToken] = func(ctx context.Context, config *logical.BackendConfig) (logical.Backend, error) {
 		tsLogger := logger.Named("token")
 		c.AddLogger(tsLogger)
-
 		return NewTokenStore(ctx, tsLogger, c, config)
 	}
 	credentialBackends[mountTypeNSToken] = func(ctx context.Context, config *logical.BackendConfig) (logical.Backend, error) {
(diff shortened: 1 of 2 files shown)


vault/core.go
```diff
@@ -1271,24 +1271,26 @@ func (c *Core) configureCredentialsBackends(backends map[string]logical.Factory,
...
 		if c.tokenStore != nil {
 			return c.tokenStore, nil
 		}
-		return nil, errors.New("token store does not exist")
+		return NewTokenStore(ctx, tsLogger, c, config)
 	}
+	credentialBackends[mountTypeNSToken] = credentialBackends[mountTypeToken]
 
 	c.credentialBackends = credentialBackends
 }
```
satoqz, line 1286: This doesn't really check out unfortunately because you will be creating the token store once only, with a given logger, but create a logger _each time_ the factory is called but never use it beyond the first one.

That being said, turns out `Core.allLoggers` is never cleared between seal cycles and presents a minor memory leak 😄

wslabosz-reply: yeah, the order is wrong, need to move the existence check to the beginning

EDIT: actually reverted the whole thing

wslabosz-reply: `Core.allLoggers` being a memory leak is a different story

Issue #2027, opened 2025-10-27
Switch to atomic types


cipherboy:
There's a lot of usage of `atomic.Store*(...)` where proper atomic types exist:

```
builtin/credential/approle/path_tidy_user_id.go:71:	defer atomic.StoreUint32(b.tidySecretIDCASGuard, 0)
builtin/logical/pki/path_tidy.go:841:		atomic.StoreUint32(b.tidyCancelCAS, 0)
builtin/logical/pki/path_tidy.go:842:		defer atomic.StoreUint32(b.tidyCASGuard, 0)
builtin/logical/kv/upgrade.go:101:					atomic.StoreUint32(b.upgrading, 0)
builtin/logical/kv/upgrade.go:115:			atomic.StoreUint32(b.upgrading, 0)
builtin/logical/kv/upgrade.go:280:		atomic.StoreUint32(b.upgrading, 0)
command/agentproxyshared/sink/file/sink_test.go:83:		atomic.StoreUint32(&b.tryCount, 0)
command/monitor_test.go:78:				atomic.StoreInt64(&code, int64(cmd.Run(tc.args)))
helper/testhelpers/testhelpers.go:462:	atomic.StoreUint32(&vault.TestingUpdateClusterAddr, 1)
http/forwarding_test.go:429:	atomic.StoreUint32(numWorkers, num)
vault/external_tests/raft/raft_autopilot_test.go:749:	atomic.StoreUint32(&vault.TestingUpdateClusterAddr, 1)
vault/external_tests/raft/raft_test.go:161:	atomic.StoreUint32(&vault.TestingUpdateClusterAddr, 1)
vault/external_tests/raft/raft_test.go:209:	atomic.StoreUint32(&vault.TestingUpdateClusterAddr, 1)
vault/external_tests/raft/raft_test.go:272:	atomic.StoreUint32(&vault.TestingUpdateClusterAddr, 1)
vault/external_tests/raft/raft_test.go:1072:	atomic.StoreUint32(&vault.TestingUpdateClusterAddr, 1)
vault/external_tests/raftha/raft_ha_test.go:58:	atomic.StoreUint32(&vault.TestingUpdateClusterAddr, 1)
vault/external_tests/raftha/raft_ha_test.go:188:		atomic.StoreUint32(&vault.TestingUpdateClusterAddr, 1)
vault/external_tests/sealmigration/seal_migration_test.go:39:			atomic.StoreUint32(&vault.TestingUpdateClusterAddr, 1)
[...]

PR #2291, opened 2026-01-09
Atomic boolean type refactor

wslabosz-reply:
## Description
- Changed the prevalent use of `int32` with `atomic.Store/Load/CompareAndSwap()` methods with concrete types (usually `atomic.Bool`). 
- Changed representation of the atomic field from pointer to value whenever it felt unnecessary for the pointer to be used. 
- Changed representation of the `*atomic.Value` to `atomic.Pointer[Type]` whenever pointer was better suited
- Removed (unused) `neverBecomeActive` core property.
- Removed (unused) `replicationFailure` core property.
- Removed (unused) `licensingStopCh` core property.
- Removed `tools/semgrep/ci/atomic.yml` semgrep rule.

## Rationale
The use of `int32` and `atomic` package could be safely replaced by concrete atomic types (in almost all cases `atomic.Boolean`), this removes complexity and improves consistency, as the `int32` value representing either `true` (1) or `false` (0) is not the greatest implementation.

Resolves: #2027

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/core.go
+++ b/vault/core.go
@@ -300,8 +300,8 @@ type Core struct {
 
 	standby              atomic.Bool
 	standbyDoneCh        chan struct{}
-	standbyStopCh        *atomic.Value
-	standbyRestartCh     *atomic.Value
+	standbyStopCh        atomic.Value
+	standbyRestartCh     atomic.Value
 	manualStepDownCh     chan struct{}
 	keepHALockOnStepDown atomic.Bool
 	heldHALock           physical.Lock
@@ -432,7 +432,7 @@ type Core struct {
 	physicalCache physical.ToggleablePurgemonster
 
 	// logRequestsLevel indicates at which level requests should be logged
-	logRequestsLevel *atomic.Int32
+	logRequestsLevel atomic.Int32
 
 	// reloadFuncs is a map containing reload functions
 	reloadFuncs map[string][]reloadutil.ReloadFunc
@@ -492,8 +492,8 @@ type Core struct {
 
 	// replicationState keeps the current replication state cached for quick
 	// lookup; activeNodeReplicationState stores the active value on standbys
-	replicationState           *atomic.Uint32
-	activeNodeReplicationState *atomic.Uint32
+	replicationState           atomic.Uint32
+	activeNodeReplicationState atomic.Uint32
 
 	// uiConfig contains UI configuration
 	uiConfig *UIConfig
@@ -541,12 +541,6 @@ type Core struct {
 	// recovery mode
 	postRecoveryUnsealFuncs []func() error
 
-	// replicationFailure is used to mark when replication has entered an
-	// unrecoverable failure.
-	replicationFailure *uint32
-
-	licensingStopCh chan struct{}
-
 	// Stores loggers so we can reset the level
 	allLoggers     []log.Logger
 	allLoggersLock sync.RWMutex
@@ -592,7 +586,7 @@ type Core struct {
 
 	// KeyRotateGracePeriod is how long we allow an upgrade path
 	// for standby instances before we delete the upgrade keys
-	keyRotateGracePeriod *atomic.Int64
+	keyRotateGracePeriod atomic.Int64
 
 	autoRotateCancel context.CancelFunc
 
@@ -929,8 +923,6 @@ func CreateCore(conf *CoreConfig) (*Core, error) {
 		seal:                conf.Seal,
 		stateLock:           stateLock,
 		router:              NewRouter(),
-		standbyStopCh:       new(atomic.Value),
-		standbyRestartCh:    new(atomic.Value),
 		baseLogger:          conf.Logger,
 		logger:              conf.Logger.Named("core"),
 		logLevel:            conf.LogLevel,
@@ -945,9 +937,6 @@ func CreateCore(conf *CoreConfig) (*Core, error) {
 		rawEnabled:                     conf.EnableRaw,
 		introspectionEnabled:           conf.EnableIntrospection,
 		shutdownDoneCh:                 new(atomic.Value),
-		replicationState:               new(atomic.Uint32),
-		activeNodeReplicationState:     new(atomic.Uint32),
-		replicationFailure:             new(uint32),
 		allLoggers:                     conf.AllLoggers,
 		builtinRegistry:                conf.BuiltinRegistry,
 		metricsHelper:                  conf.MetricsHelper,
@@ -957,7 +946,6 @@ func CreateCore(conf *CoreConfig) (*Core, error) {
 		raftJoinDoneCh:                 make(chan struct{}),
 		pendingRaftPeerChallengeKey:    make([]byte, 32),
 		clusterHeartbeatInterval:       clusterHeartbeatInterval,
-		keyRotateGracePeriod:           new(atomic.Int64),
 		numExpirationWorkers:           conf.NumExpirationWorkers,
 		raftFollowerStates:             raft.NewFollowerStates(),
 		disableAutopilot:               conf.DisableAutopilot,
@@ -1209,8 +1197,6 @@ func (c *Core) configureListeners(conf *CoreConfig) error {
 
 // configureLogRequestsLevel configures the Core with the supplied log requests level.
 func (c *Core) configureLogRequestsLevel(level string) {
-	c.logRequestsLevel = &atomic.Int32{}
-
 	lvl := log.LevelFromString(level)
 
 	switch {
@@ -3583,7 +3569,7 @@ func (c *Core) StoreInFlightReqData(reqID string, data InFlightReqData) {
 // request from the inFlightReqMap and decrement the number of in-flight
 // requests by one.
 func (c *Core) FinalizeInFlightReqData(reqID string, statusCode int) {
-	if c.logRequestsLevel != nil && c.logRequestsLevel.Load() != 0 {
+	if c.logRequestsLevel.Load() != 0 {
 		c.LogCompletedRequests(reqID, statusCode)
 	}
 
--- a/vault/ha.go
+++ b/vault/ha.go
@@ -1220,9 +1220,12 @@ func (c *Core) advertiseLeader(ctx context.Context, uuid string, leaderLostCh <-
 		D:    key.D,
 	}
 
-	locCert := *c.localClusterCert.Load()
-	localCert := make([]byte, len(locCert))
-	copy(localCert, locCert)
+	locCert := c.localClusterCert.Load()
+	if locCert == nil {
+		return errors.New("couldn't load local cluster cert")
+	}
+	localCert := make([]byte, len(*locCert))
+	copy(localCert, *locCert)
 	adv := &activeAdvertisement{
 		RedirectAddr:     c.redirectAddr,
 		ClusterAddr:      c.ClusterAddr(),
--- a/vault/testing.go
+++ b/vault/testing.go
@@ -1002,10 +1002,6 @@ func (c *TestClusterCore) stop() error {
 		}
 		c.Logger().Info("listeners successfully shut down")
 	}
-	if c.licensingStopCh != nil {
-		close(c.licensingStopCh)
-		c.licensingStopCh = nil
-	}
 
 	if err := c.Shutdown(); err != nil {
 		return err


builtin/credential/approle/backend.go
```diff
@@ -93,7 +94,7 @@ func Backend(conf *logical.BackendConfig) (*backend, error) {
 		// Create locks to modify the generated SecretIDAccessors
 		secretIDAccessorLocks: locksutil.CreateLocks(),
 
-		tidySecretIDCASGuard: new(uint32),
+		tidySecretIDCASGuard: atomic.Bool{},
```
satoqz: No need to even construct these if the default is false.

wslabosz-reply: yeah, that's actually valid, will review the use then (although I kinda like the explicitness just a tad bit, but will adjust)

vault/request_forwarding_rpc.go
```diff
@@ -173,7 +172,7 @@ func (c *forwardingClient) startHeartbeat() {
 			case <-c.echoContext.Done():
 				c.echoTicker.Stop()
 				c.core.logger.Debug("forwarding: stopping heartbeating")
-				atomic.StoreUint32(c.core.activeNodeReplicationState, uint32(consts.ReplicationUnknown))
+				c.core.activeNodeReplicationState.Store(uint32(consts.ReplicationUnknown))
```
satoqz, line 175:
```suggestion
				c.core.activeNodeReplicationState.Store(consts.ReplicationUnknown)
```

builtin/logical/pki/backend.go
```diff
@@ -266,12 +262,9 @@ func Backend(conf *logical.BackendConfig) *backend {
...
 
 	// Metrics initialization for count of certificates in storage
-	b.certCountEnabled = &atomic.Bool{}
-	b.publishCertCountMetrics = &atomic.Bool{}
-	b.certsCounted = &atomic.Bool{}
 	b.certCountError = "Initialize Not Yet Run, Cert Counts Unavailable"
-	b.certCount = &atomic.Uint32{}
-	b.revokedCertCount = &atomic.Uint32{}
+	b.certCount = atomic.Uint32{}
+	b.revokedCertCount = atomic.Uint32{}
```
satoqz:
```suggestion
```

command/agentproxyshared/sink/sink.go
```diff
@@ -54,19 +54,17 @@ type SinkServer struct {
...
 
 func NewSinkServer(conf *SinkServerConfig) *SinkServer {
-	ss := &SinkServer{
+	return &SinkServer{
 		logger:        conf.Logger,
 		client:        conf.Client,
 		random:        rand.New(rand.NewSource(int64(time.Now().Nanosecond()))),
 		exitAfterAuth: conf.ExitAfterAuth,
-		remaining:     new(int32),
+		remaining:     atomic.Int32{},
```
satoqz:
```suggestion
```

helper/metricsutil/gauge_process_test.go
```diff
@@ -77,13 +77,13 @@ func startSimulatedTime() *SimulatedTime {
...
 type SimulatedCollector struct {
-	numCalls    uint32
+	numCalls    atomic.Uint32
 	callBarrier chan uint32
 }
 
 func newSimulatedCollector() *SimulatedCollector {
 	return &SimulatedCollector{
-		numCalls:    0,
+		numCalls:    atomic.Uint32{},
```
satoqz:
```suggestion
```

helper/metricsutil/gauge_process_test.go
```diff
@@ -100,8 +100,8 @@ func (s *SimulatedCollector) waitForCall(t *testing.T) {
 }
 
 func (s *SimulatedCollector) EmptyCollectionFunction(ctx context.Context) ([]GaugeLabelValues, error) {
-	atomic.AddUint32(&s.numCalls, 1)
-	s.callBarrier <- s.numCalls
+	s.numCalls.Add(1)
+	s.callBarrier <- s.numCalls.Load()
```
satoqz:
```suggestion
	s.callBarrier <- s.numCalls.Add(1)
```

helper/metricsutil/gauge_process_test.go
```diff
@@ -545,8 +545,8 @@ func TestGauge_MeasurementError(t *testing.T) {
 	}
 
 	f := func(ctx context.Context) ([]GaugeLabelValues, error) {
-		atomic.AddUint32(&c.numCalls, 1)
-		c.callBarrier <- c.numCalls
+		c.numCalls.Add(1)
+		c.callBarrier <- c.numCalls.Load()
```
satoqz:
```suggestion
		c.callBarrier <- c.numCalls.Add(1)
```

sdk/logical/response.go
```diff
@@ -206,27 +206,27 @@ func RespondWithStatusCode(resp *Response, req *Request, code int) (*Response, e
 // write directly to the HTTP response writer.
 type HTTPResponseWriter struct {
 	http.ResponseWriter
-	written *uint32
+	written *atomic.Bool
```
satoqz:
```suggestion
	written atomic.Bool
```

sdk/logical/response.go
```diff
@@ -206,27 +206,27 @@ func RespondWithStatusCode(resp *Response, req *Request, code int) (*Response, e
...
 }
 
 // NewHTTPResponseWriter creates a new HTTPResponseWriter object that wraps the
 // provided io.Writer.
 func NewHTTPResponseWriter(w http.ResponseWriter) *HTTPResponseWriter {
 	return &HTTPResponseWriter{
 		ResponseWriter: w,
-		written:        new(uint32),
+		written:        new(atomic.Bool),
```
satoqz:
```suggestion
```

vault/expiration_test.go
```diff
@@ -2368,12 +2368,12 @@ func TestExpiration_revokeEntry_rejected_fairsharing(t *testing.T) {
 	core, _, _ := TestCoreUnsealed(t)
 	exp := core.expiration
 
-	rejected := new(uint32)
+	rejected := atomic.Bool{}
```
satoqz:
```suggestion
	var rejected atomic.Bool
```

vault/ha.go
```diff
@@ -517,7 +517,7 @@ func (c *Core) runStandbyOnce(doneCh chan<- struct{}, manualStepDownCh chan stru
 		defer readStandbyCancel()
 
 		// Unseal, holding the state lock.
-		atomic.StoreUint32(c.replicationState, uint32(consts.ReplicationDRDisabled|consts.ReplicationPerformanceStandby))
+		c.replicationState.Store(uint32(consts.ReplicationDRDisabled | consts.ReplicationPerformanceStandby))
```
satoqz, line 517:
```suggestion
		c.replicationState.Store(consts.ReplicationDRDisabled | consts.ReplicationPerformanceStandby)
```

vault/ha.go
```diff
@@ -798,7 +791,7 @@ func (c *Core) waitForLeadership(manualStepDownCh, stopCh <-chan struct{}) {
 
 		// Handle a failure to unseal
 		if err != nil {
-			atomic.StoreUint32(c.replicationState, uint32(consts.ReplicationDRDisabled|consts.ReplicationPerformanceStandby))
+			c.replicationState.Store(uint32(consts.ReplicationDRDisabled | consts.ReplicationPerformanceStandby))
```
satoqz, line 791:
```suggestion
			c.replicationState.Store(consts.ReplicationDRDisabled | consts.ReplicationPerformanceStandby)
```

vault/ha.go
```diff
@@ -786,7 +779,7 @@ func (c *Core) waitForLeadership(manualStepDownCh, stopCh <-chan struct{}) {
 		}
 
 		// Attempt the post-unseal process
-		atomic.StoreUint32(c.replicationState, uint32(consts.ReplicationDRDisabled|consts.ReplicationPerformancePrimary))
+		c.replicationState.Store(uint32(consts.ReplicationDRDisabled | consts.ReplicationPerformancePrimary))
```
satoqz, line 779:
```suggestion
		c.replicationState.Store(consts.ReplicationDRDisabled | consts.ReplicationPerformancePrimary)
```

vault/expiration.go
```diff
@@ -347,15 +347,12 @@ func NewExpirationManager(c *Core, e ExpireLeaseStrategy, logger log.Logger, det
...
-		// new instances of the expiration manager will go immediately into
-		// restore mode
-		restoreMode:  new(int32),
 		restoreLocks: locksutil.CreateLocks(),
 		quitCh:       make(chan struct{}),
 
 		coreStateLock:     c.stateLock,
 		quitContext:       c.activeContext,
-		leaseCheckCounter: new(uint32),
+		leaseCheckCounter: atomic.Uint32{},
```
satoqz:
```suggestion
```

vault/cors.go
```diff
@@ -37,18 +32,18 @@ var StdAllowedHeaders = []string{
 type CORSConfig struct {
 	sync.RWMutex   `json:"-"`
 	core           *Core
-	Enabled        *uint32  `json:"enabled"`
-	AllowedOrigins []string `json:"allowed_origins,omitempty"`
-	AllowedHeaders []string `json:"allowed_headers,omitempty"`
+	Enabled        *atomic.Bool `json:"enabled"`
+	AllowedOrigins []string     `json:"allowed_origins,omitempty"`
+	AllowedHeaders []string     `json:"allowed_headers,omitempty"`
 }
```
satoqz, line 39: Unsure about this one as it's serialized to storage. Does it load an older state that wrote enabled as an integer correctly?

physical/raft/fsm.go
```diff
@@ -110,8 +110,8 @@ type FSM struct {
 
 	// latestIndex and latestTerm are the term and index of the last log we
 	// received
-	latestIndex *uint64
-	latestTerm  *uint64
+	latestIndex *atomic.Uint64
+	latestTerm  *atomic.Uint64
```
satoqz: Do these need to be pointers?

sdk/physical/inmem/inmem.go
```diff
@@ -377,12 +361,8 @@ func (i *InmemBackend) listPaginatedInternal(ctx context.Context, prefix string,
 	return out, nil
 }
 
-func (i *InmemBackend) FailList(fail bool) {
```
satoqz, line 380: I'm unsure about changing these APIs. There may not currently be a test case in our tree that sets `FailX(false)`, but I don't see why we wouldn't want that in the future. It also breaks `sdk/` compatibility.

physical/raft/fsm.go
wslabosz-reply: probably not

vault/request_forwarding_rpc.go
wslabosz-reply: unfortunately that doesn't work, we need to cast those.

vault/ha.go
wslabosz-reply: won't work

vault/ha.go
wslabosz-reply: won't work

vault/ha.go
wslabosz-reply: won't work

vault/cors.go
satoqz: It doesn't.

vault/request_forwarding_rpc.go
satoqz: :clash-royale-cry:

http/cors.go
```diff
@@ -29,7 +29,7 @@ func wrapCORSHandler(h http.Handler, core *vault.Core) http.Handler {
 		// If CORS is not enabled or if no Origin header is present (i.e. the request
 		// is from the Vault CLI. A browser will always send an Origin header), then
 		// just return a 204.
-		if !corsConf.IsEnabled() {
```
satoqz, line 32: I'd probably add this back.

vault/cors.go
phyrog: I believe a custom Marshal method like this should work:

```go
func (c *CORSConfig) MarshalJSON() ([]byte, error) {
    type Alias CORSConfig
    var enabled uint32
    if c.Enabled.Load() {
        enabled = 1
    }
    return json.Marshal(&struct{
        Enabled uint32 `json:"enabled"`
        *Alias
    }{
        Enabled: enabled,
        Alias:   (*Alias)(c),
    })
}
```

Same for Unmarshal in reverse.

command/agentproxyshared/auth/jwt/jwt.go
```diff
@@ -182,9 +180,9 @@ func (j *jwtMethod) runWatcher() {
 			return
 
 		case <-j.ticker.C:
-			latestToken := j.latestToken.Load().(string)
+			latestToken := *j.latestToken.Load()
 			j.ingressToken()
-			newToken := j.latestToken.Load().(string)
+			newToken := *j.latestToken.Load()
```
satoqz: Are we sure these cannot nil deref?

vault/cors.go
satoqz: I'm just not sure the complexity is worth it at that point.

wslabosz-reply: I would rather introduce a change that will enable backwards compatibility in a separate PR.

command/agentproxyshared/auth/jwt/jwt.go
wslabosz-reply: good question, will nil guard

vault/cors.go
cipherboy: Ugh, I really don't like using marshaled representations as the in-memory ones when it comes to atomic booleans. It would also be a ton of work to split.

If we want to do this in a separate PR though (which I'm fine with), let's leave the current behavior as-is then? Thanks!

sdk/physical/inmem/inmem.go
cipherboy: Eh, I don't think we have a lot of usage of `Fail*`. I think it is fine. 

I am slightly confused though, why we don't take `bool` here, still?

vault/cors.go
wslabosz-reply: @cipherboy sure, makes sense to me, to be clear - currently I've only changed the field type from `*uint32` to `*atomic.Uint32`

sdk/physical/inmem/inmem.go
satoqz: That is my point, I'd like to retain the parameter.

wslabosz-reply: @cipherboy 
so before it was a toggle of sorts, but there was no "toggle use" it was only used as "set the field to true", never setting to false, so I just opted for streamlined implementation, can revert and make it a toggle again

vault/cors.go
```diff
@@ -70,30 +65,26 @@ func (c *Core) saveCORSConfig(ctx context.Context) error {
 
 // This should only be called with the core state lock held for writing
 func (c *Core) loadCORSConfig(ctx context.Context) error {
-	view := c.systemBarrierView.SubView("config/")
-
 	// Load the config in
-	out, err := view.Get(ctx, "cors")
+	out, err := c.systemBarrierView.SubView("config/").Get(ctx, "cors")
```
satoqz:
```suggestion
	out, err := c.systemBarrierView.Get(ctx, "config/cors")
```

sdk/physical/inmem/inmem.go
wslabosz-reply: reverted

vault/cors.go
cipherboy: @wslabosz-reply I think even that might have serialization differences over `*uint32`? So I'm not sure even keeping `*atomic.Uint32` is compatible.

phyrog: The atomic types only contain private fields and do not have any custom serialization code, so essentially they all serialize to `{}`. Hence my proposal to serialize the underlying value to uint32.

wslabosz-reply: yeah, my bad, didn't notice it doesn't marshal, I can go either with a revert to non-atomic type or with @phyrog suggestion.

vault/core.go
```diff
@@ -492,8 +492,8 @@ type Core struct {
 
 	// replicationState keeps the current replication state cached for quick
 	// lookup; activeNodeReplicationState stores the active value on standbys
-	replicationState           *uint32
-	activeNodeReplicationState *uint32
+	replicationState           *atomic.Uint32
+	activeNodeReplicationState *atomic.Uint32
```
phyrog:
```suggestion
	replicationState           atomic.Uint32
	activeNodeReplicationState atomic.Uint32
```

vault/core.go
```diff
@@ -596,7 +592,7 @@ type Core struct {
 
 	// KeyRotateGracePeriod is how long we allow an upgrade path
 	// for standby instances before we delete the upgrade keys
-	keyRotateGracePeriod *int64
+	keyRotateGracePeriod *atomic.Int64
```
phyrog:
```suggestion
	keyRotateGracePeriod atomic.Int64
```

vault/core.go
```diff
@@ -296,14 +296,14 @@ type Core struct {
 
 	// stateLock protects mutable state
 	stateLock locking.RWMutex
-	sealed    *uint32
+	sealed    atomic.Bool
 
 	standby              atomic.Bool
 	standbyDoneCh        chan struct{}
 	standbyStopCh        *atomic.Value
 	standbyRestartCh     *atomic.Value
```
phyrog:
```suggestion
standbyStopCh        atomic.Value
standbyRestartCh     atomic.Value
```

vault/core.go
```diff
@@ -950,29 +945,19 @@ func CreateCore(conf *CoreConfig) (*Core, error) {
...
 		introspectionEnabled:           conf.EnableIntrospection,
 		shutdownDoneCh:                 new(atomic.Value),
-		replicationState:               new(uint32),
-		localClusterPrivateKey:         new(atomic.Value),
-		localClusterCert:               new(atomic.Value),
-		localClusterParsedCert:         new(atomic.Value),
-		activeNodeReplicationState:     new(uint32),
-		keepHALockOnStepDown:           new(uint32),
+		replicationState:               new(atomic.Uint32),
+		activeNodeReplicationState:     new(atomic.Uint32),
```
phyrog:
```suggestion

```

vault/core.go
```diff
@@ -920,25 +920,20 @@ func CreateCore(conf *CoreConfig) (*Core, error) {
...
+		physical:            conf.Physical,
+		serviceRegistration: conf.GetServiceRegistration(),
+		underlyingPhysical:  conf.Physical,
+		storageType:         conf.StorageType,
+		redirectAddr:        conf.RedirectAddr,
+		seal:                conf.Seal,
+		stateLock:           stateLock,
+		router:              NewRouter(),
+		standbyStopCh:       new(atomic.Value),
+		standbyRestartCh:    new(atomic.Value),
```
phyrog:
```suggestion

```

vault/core.go
```diff
@@ -950,29 +945,19 @@ func CreateCore(conf *CoreConfig) (*Core, error) {
...
 		secureRandomReader:             conf.SecureRandomReader,
-		rawConfig:                      new(atomic.Value),
 		recoveryMode:                   conf.RecoveryMode,
-		postUnsealStarted:              new(uint32),
-		raftInfo:                       new(atomic.Value),
 		raftJoinDoneCh:                 make(chan struct{}),
 		pendingRaftPeerChallengeKey:    make([]byte, 32),
 		clusterHeartbeatInterval:       clusterHeartbeatInterval,
-		keyRotateGracePeriod:           new(int64),
+		keyRotateGracePeriod:           new(atomic.Int64),
```
phyrog:
```suggestion

```

vault/ha.go
```diff
@@ -1240,7 +1220,7 @@ func (c *Core) advertiseLeader(ctx context.Context, uuid string, leaderLostCh <-
 		D:    key.D,
 	}
 
-	locCert := c.localClusterCert.Load().([]byte)
+	locCert := *c.localClusterCert.Load()
```
phyrog: How sure are we that we can never load `nil` here? I'd prefer an explicit check here.

Issue #1453, opened 2025-06-17
Move to github.com/cenkalti/backoff/v5


cipherboy:
`github.com/cenkalti/backoff/v5` is the latest version of backoff, which introduces some new API related changes. This would be good to upgrade to so that `dependabot` can keep us on the latest version. AFAIK, this library is used in a few places but is relatively non-critical.

nithinputhenveettil:
@cipherboy : I would like to work on this. Could you please assign this to me? Thanks.

cipherboy:
@nithinputhenveettil Thanks! Let me know if you have any questions. :-)

wslabosz-reply:
Hey @nithinputhenveettil are you still working on this issue?

nithinputhenveettil:
Very sorry @wslabosz-reply, I missed this earlier. I’ll update you on the progress by this weekend.

satoqz:
Hi @nithinputhenveettil, are you still working on this issue? I can take it over if you don't mind.

PR #2301, opened 2026-01-14
Move to github.com/cenkalti/backoff/v5

satoqz:
Resolves: #1453

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/logical/kubernetes/integrationtest/wal_rollback_test.go
+++ b/builtin/logical/kubernetes/integrationtest/wal_rollback_test.go
@@ -207,29 +207,29 @@ func checkObjects(t *testing.T, roleConfig map[string]interface{}, isClusterBind
 	}
 
 	// Check the k8s objects that should have been created (all but the ServiceAccount)
-	operation := func() (any, error) {
+	operation := func() (none struct{}, err error) {
 		if existingRole == "" {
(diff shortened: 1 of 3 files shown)


builtin/logical/kubernetes/integrationtest/wal_rollback_test.go
```diff
@@ -207,29 +207,29 @@ func checkObjects(t *testing.T, roleConfig map[string]interface{}, isClusterBind
 	}
 
 	// Check the k8s objects that should have been created (all but the ServiceAccount)
-	operation := func() (any, error) {
+	operation := func() (none struct{}, err error) {
```
phyrog, line 210: I'd prefer a `//nolint:nilnil` instead (or disabling the rule alltogether). We're never using the result of any of these operations.

satoqz: Didn't really like the `any` regardless, so I figured using an empty struct might be more idiomatic. Dunno 🤷

wslabosz-reply (approved):

builtin/logical/kubernetes/integrationtest/wal_rollback_test.go
phyrog: Doesn't seem like anyone else has opinions in either direction, so fine with me 🙂

PR #2307, opened 2026-01-15
Remove some additional enterprise license references

satoqz:
As title states, just some more leftovers. Following this, I believe all remaining enterprise license references are in Makefiles and bats tests of plugins that we've internalized, I'm not sure what their status is or if they are even used anymore.

## Acknowledgements

       also allowed.


.copywrite.hcl
satoqz, line 1: Not related to enterprise licensing but references mostly enterprise paths in the ignores. I also don't think we use this tool, but if we wanted to could trivially add this file back.

satoqz:
Some tests to look at and and I believe something about versioned docs and partials is broken, I'll get back to this.

satoqz:
Ah so https://github.com/hashicorp/remark-plugins is playing into my cards. Perhaps I can just get rid of that one.

satoqz:
Nevermind, removing or even updating partials is a giant can of worms I'm not willing to open here 😓

cc @JanMa, we've left partials unversioned (and I'm not sure how we would best version them retroactively since they are not a concept native to Docusaurus, but achieved via remark plugin), hence you cannot really remove or update them without breaking the docs of older versions.

satoqz:
Merging as I've only made a minor change to fix the docs build post-approval.

JanMa:
@satoqz i think long term it makes sense to get rid of partials, especially as they are now causing issues with the versioned docs. TBH this is kind of tedious and I wasn't motivated to do this so far 😅

PR #2298, opened 2026-01-12
Add Securosys to the integrators page

thgoebel:
## Description

Add Securosys to the integrators page.

Logo: https://github.com/openbao/openbao-ecosystem-logos/pull/9 --> I'll update the references here once that is merged.

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #2308, opened 2026-01-15
Fix broken Randomize

thgoebel:
## Description

The "fix" in 1ed7b7365fc1bde7ce86205e568e54482b89a1c2 (https://github.com/openbao/openbao/pull/2298) to prevent crashes when there is a single child results in no randomization happening when there are multiple children.

So revert the "fix". ~~And comment out Randomize on the integrators page. But keep it around, so that future integrators add it.~~
EDIT: After sleeping over it, I changed this to instead explicitly handle the single-Object case. That seems better, since it allows applying `<Randomize>` to single items, which seems like the better UX for writers.

Sorry for missing this earlier 🙈 

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/ecosystem/lib/Randomize.tsx
+++ b/website/content/ecosystem/lib/Randomize.tsx
@@ -1,5 +1,7 @@
 import shuffle from "lodash/shuffle";
+import { Children } from "react";
 
-export default function Randomize(children) {
-  return shuffle(children);
+export default function Randomize({children}) {
+  const items = Children.toArray(children);
+  return shuffle(items);
 }


website/content/ecosystem/lib/Randomize.tsx
```diff
@@ -1,5 +1,8 @@
 import shuffle from "lodash/shuffle";
 
-export default function Randomize(children) {
-  return shuffle(children);
+export default function Randomize({children}) {
+  // When <Randomize> is applied to a single child, `children` is a single
+  // Object instead of an Array, but shuffle() expects an Array.
+  const items = Array.isArray(children) ? children : [children];
```
phyrog: Might want to use `Children.toArray` (https://react.dev/reference/react/Children#children-toarray) here instead. According to the React docs, you should treat children as an opaque thing:

> The difference is that the children data structure is considered opaque. This means that even if it’s sometimes an array, you should not assume it’s an array or any other particular data type.

satoqz: The funny part is that `Children.Array` is a "legacy API" yet it remains the recommended solution.

thgoebel: Good point, done!

JanMa (approved):

website/content/ecosystem/lib/Randomize.tsx
```diff
@@ -1,5 +1,7 @@
 import shuffle from "lodash/shuffle";
+import Children from "react";
```
phyrog:
```suggestion
import { Children } from "react";
```

phyrog (changes requested):

website/content/ecosystem/lib/Randomize.tsx
thgoebel: Should I then also change it to 

```
import { shuffle } from "lodash/shuffle";
```

for consistency?

thgoebel: Nvm, it would need to be

```
import { shuffle } from "lodash";
```

In which case I don't think it makes a big difference.

satoqz:
Thanks @thgoebel!

PR #2309, opened 2026-01-15
Opt-in to Docusaurus Faster

thgoebel:
## Description

Make the website build faster.

See:

- https://docusaurus.io/blog/releases/3.6
- https://docusaurus.io/blog/releases/3.8

On my laptop, the speedup is ~50% (as measured by `npm run clear && time npm run build`):

```
BEFORE:

real    2m57.519s
user    6m17.808s
sys     1m4.498s

AFTER:

real    1m32.886s
user    3m10.598s
sys     1m10.958s
```

I clicked through most of the pages, but please also double check that this doesn't break any existing pages.

And note from the [Docusaurus 3.6 blog](https://docusaurus.io/blog/releases/3.6):

> Don't be afraid to turn this feature on. What's experimental is the config options.
>
> The new infrastructure is robust and well-tested by our CI pipeline. The [Docusaurus site](https://docusaurus.io/) already uses it in production, and we plan to use it on other Meta docs sites as well.

## Acknowledgements

       also allowed.

### DIFF
--- a/website/docusaurus.config.ts
+++ b/website/docusaurus.config.ts
@@ -49,6 +49,13 @@ const config: Config = {
   },
   staticDirectories: ["public"],
 
+  future: {
+    v4: {
+      removeLegacyPostBuildHeadAttribute: true, // needed by experimental_faster
+    },
+    experimental_faster: true,
+  },
+
   markdown: {
     mermaid: true,
   },
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@docusaurus/core": "^3.9.2",
+        "@docusaurus/faster": "^3.9.2",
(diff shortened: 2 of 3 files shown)


phyrog (approved):
Speed-up seems to be legit, 6min vs 3min now in CI.

phyrog:
> I'm personally not really convinced it is worth enabling an experimental feature to improve a build we're hardly bottlenecked by 🤷 But perhaps others feel differently.

It was introduced in 2024, so I don't think it's that experimental any more 🙂

cipherboy:
@satoqz n=1 but I found the versions docs build to be rather slow locally. I agree it's not a bottleneck for 90% of contributions, but every so often it might be nice.

satoqz (approved):
I mean we can revert if there are problems, so I don't want to stand in the way :)

cipherboy:
Thank you @thgoebel!

Issue #2311, opened 2026-01-17
MySQL database plugin multi-host connection failover support


cropalato:
**Is your feature request related to a problem? Please describe.**
When using MariaDB in a clustered setup (such as Galera Cluster or MySQL master-master replication), there is no way to specify multiple hosts in the connection URL for automatic failover. If the primary host becomes unavailable, OpenBao cannot automatically connect to an alternative host, resulting in credential generation failure until manual intervention.

**Describe the solution you'd like**
Support multiple hosts in the MySQL database secrets engine connection URL:

connection_url="{{username}}:{{password}}@tcp(host1:3306,host2:3306,host3:3306)/dbname"

When connecting, OpenBao should try each host sequentially until one succeeds. This would be implemented using `go-sql-driver/mysql`'s `Config.DialFunc` with `mysql.NewConnector()` and `sql.OpenDB()`.

**Describe alternatives you've considered**
- **External load balancer (HAProxy/ProxySQL):** Works but adds infrastructure complexity and another point of failure.
- **DNS-based failover:** Relies on DNS TTL and can have delayed failover.

**Explain any additional use-cases**

**Additional context**
- PostgreSQL multi-host support was added via migration to `pgx` driver (hashicorp/vault#7247).
  - The `go-sql-driver/mysql` v1.5.0+ supports custom `DialFunc` which enables this without changing drivers.
  - This feature would be backward compatible - single-host URLs would work unchanged.

wslabosz-reply:
@cropalato 
Hello! First of all, glad to have you here.

I've looked at the PR you put forward and verified it with both documentations of mysql and the driver we use. I'm not happy with the driver being unable to pass multiple addresses for the multi-host connection failover overall, but your proposal seems sounds. I'd also consider adding the `priority` property passing, and handling as per [documentation](https://dev.mysql.com/doc/x-devapi-userguide/en/connecting-to-one-of-multiple-hosts.html).

PR #2312, opened 2026-01-17
database/mysql: Add multi-host connection failover support

cropalato:
## Description

  Add support for multiple hosts in MySQL connection URL for automatic failover.

  Format: `tcp(host1:3306,host2:3306)/dbname`

## Rationale

  Resolves: #2311

## Acknowledgements

 - [ X ] By contributing this change, I certify I have not used generative AI (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes.
 - [ X ] By contributing this change, I certify I have signed-off on the [DCO ownership](https://developercertificate.org/) statement and this change did not use post-BUSL-licensed code from HashiCorp. Existing MPL-licensed code is still allowed, subject to attribution. Code authored by yourself and submitted to HashiCorp for inclusion is also allowed.

### DIFF
--- a/changelog/2312.txt
+++ b/changelog/2312.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+database/mysql: Add multi-host connection failover support. Connection URLs can now specify multiple hosts (e.g., `tcp(host1:3306,host2:3306)`) for automatic failover when a host becomes unavailable.
+```
--- a/plugins/database/mysql/connection_producer.go
+++ b/plugins/database/mysql/connection_producer.go
@@ -10,12 +10,16 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"net"
 	"net/url"
+	"regexp"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/go-sql-driver/mysql"
 	"github.com/go-viper/mapstructure/v2"
+	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
 	"github.com/hashicorp/go-uuid"
 	"github.com/openbao/openbao/sdk/v2/database/helper/connutil"
@@ -38,6 +42,9 @@ type mySQLConnectionProducer struct {
 
 	// tlsConfigName is a globally unique name that references the TLS config for this instance in the mysql driver
 	tlsConfigName string
+
+	// hosts holds parsed host:port pairs for multi-host failover
+	hosts []string
 
 	RawConfig             map[string]interface{}
 	maxConnectionLifetime time.Duration
@@ -75,6 +82,11 @@ func (c *mySQLConnectionProducer) Init(ctx context.Context, conf map[string]inte
 		"username": url.PathEscape(c.Username),
 		"password": password,
 	})
+
+	// Parse multi-host DSN for failover support
+	if err := c.parseMultiHostDSN(); err != nil {
+		return nil, fmt.Errorf("error parsing multi-host DSN: %w", err)
+	}
 
 	if c.MaxOpenConnections == 0 {
 		c.MaxOpenConnections = 4
@@ -143,15 +155,29 @@ func (c *mySQLConnectionProducer) Connection(ctx context.Context) (interface{}, 
 		c.db.Close()
 	}
 
-	connURL, err := c.addTLStoDSN()
+	// Parse the DSN into a Config struct
+	config, err := mysql.ParseDSN(c.ConnectionURL)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to parse connectionURL: %w", err)
 	}
 
-	c.db, err = sql.Open("mysql", connURL)
-	if err != nil {
-		return nil, err
+	// Set TLS config if configured
+	if c.tlsConfigName != "" {
+		config.TLSConfig = c.tlsConfigName
 	}
+
+	// Enable multi-host failover if multiple hosts are configured
+	if len(c.hosts) > 1 {
+		config.DialFunc = c.dialWithFailover
+	}
+
+	// Create connector and open DB using the connector-based approach
+	connector, err := mysql.NewConnector(config)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create mysql connector: %w", err)
+	}
+
+	c.db = sql.OpenDB(connector)
 
 	// Set some connection pool settings. We don't need much of this,
 	// since the request rate shouldn't be high.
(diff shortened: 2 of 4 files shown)


cropalato:
@wslabosz-reply Thanks for the review, but I have a question. You are proposing to let parseDSN take care of the url. But if I understand correctly, what will follow the following flow:
Init() -> Connection() -> mysql.ParseDSN("user:pass@tcp(host1:3306,host2:3307)/db") -> ensureHavePort("host1:3306,host2:3307"). and that will return "[host1:3306,host2:3307]:3306"
Please let me know if I missed something.

About priority, I was not able to find if go-sql-driver/mysql supports that.

Please, point me in the direction openbao wants to go with that. I really want to have this feature available in the official version (that is a real use case for my usage)

wslabosz-reply:
> @wslabosz-reply Thanks for the review, but I have a question. You are proposing to let parseDSN take care of the url. But if I understand correctly, what will follow the following flow: Init() -> Connection() -> mysql.ParseDSN("user:pass@tcp(host1:3306,host2:3307)/db") -> ensureHavePort("host1:3306,host2:3307"). and that will return "[host1:3306,host2:3307]:3306" Please let me know if I missed something.

Yeah, in "EDIT" I've mentioned that this will conflict with tls specification, and it conflicts with the address as you said, you'd have to manually edit this property (`Addr`) which doesn't sound like a great pattern.

> About priority, I was not able to find if go-sql-driver/mysql supports that.

Yes,`go-sql-driver/mysql` doesn't support that (like specifying multiple addresses), it's about what is passed as a 'connection string' value, which should work downstream.
 
> Please, point me in the direction openbao wants to go with that. I really want to have this feature available in the official version (that is a real use case for my usage)

For me the best direction would be to go to `go-sql-driver/mysql` repo and ask about support for the multi-host specification, but I know it's not something to be done in a week, hence I'm okay with your proposal as is.

plugins/database/mysql/connection_producer.go
```diff
@@ -143,16 +154,30 @@ func (c *mySQLConnectionProducer) Connection(ctx context.Context) (interface{},
...
+	config, err := mysql.ParseDSN(c.ConnectionURL)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to parse connectionURL: %w", err)
+	}
+
+	// Set TLS config if configured
+	if c.tlsConfigName != "" {
+		config.TLSConfig = c.tlsConfigName
+	}
```
wslabosz-reply, line 167: this logic is done in `(mySQLConnectionProducer).addTLStoDSN()` method that you've removed the use of, would be better to either adjust that method or remove it all together (with associated test, which should be replaced by an equivalent then) / my 2c.

cropalato: Indeed, it is a duplication. I'll remove. Thanks for the comment @wslabosz-reply. I removed the code.

cropalato: @wslabosz-reply please let me know if you detect any other problem.

plugins/database/mysql/connection_producer.go
```diff
@@ -218,16 +243,76 @@ func (c *mySQLConnectionProducer) getTLSAuth() (tlsConfig *tls.Config, err error
...
+
+	if len(c.hosts) == 0 {
+		return errors.New("no valid hosts found in connection URL")
+	}
+
+	// Normalize the ConnectionURL to use only the first host
+	// This is required because mysql.ParseDSN() doesn't support multiple hosts
+	normalizedAddr := "tcp(" + c.hosts[0] + ")"
+	c.ConnectionURL = re.ReplaceAllString(c.ConnectionURL, normalizedAddr)
```
wslabosz-reply:
```suggestion
	c.ConnectionURL = re.ReplaceAllString(c.ConnectionURL, fmt.Sprintf("tcp(%s)", c.hosts[0]))
```

plugins/database/mysql/connection_producer.go
```diff
@@ -218,16 +243,76 @@ func (c *mySQLConnectionProducer) getTLSAuth() (tlsConfig *tls.Config, err error
...
+
+	return nil
+}
+
+// dialWithFailover tries each host in sequence until one succeeds.
+func (c *mySQLConnectionProducer) dialWithFailover(ctx context.Context, network, _ string) (conn net.Conn, err error) {
+	var lastErr error
```
wslabosz-reply: you can also make use of `multierror.Error` from `"github.com/hashicorp/go-multierror"`

cropalato: Thanks. I just did

plugins/database/mysql/connection_producer.go
```diff
@@ -218,16 +244,75 @@ func (c *mySQLConnectionProducer) getTLSAuth() (tlsConfig *tls.Config, err error
...
+		merr = multierror.Append(merr, fmt.Errorf("failed to connect to %s: %w", host, err))
+	}
+
+	if merr != nil {
+		return nil, merr.ErrorOrNil()
 	}
 
-	connURL = config.FormatDSN()
-	return connURL, nil
+	return nil, errors.New("no hosts configured")
```
wslabosz-reply: I'm not sure you need that as on line 170, you just set the custom dial func if there is at least one host specified, so knowing that nothing can modify the `c.hosts` between connection setup and the host parsing, this looks like a unnecessary guard

cropalato: nice catch. I just removed. thanks.

changelog/2312.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
cipherboy: This isn't in our `feature` format; is this large enough to be a top-level feature call out (happy to suggest structure if so), or should this just be an `improvement`?

cropalato:
@wslabosz-reply The failing check is not related to my code. Would you like me to do something about that?

plugins/database/mysql/connection_producer.go
```diff
@@ -218,16 +244,71 @@ func (c *mySQLConnectionProducer) getTLSAuth() (tlsConfig *tls.Config, err error
...
+}
+
+// dialWithFailover tries each host in sequence until one succeeds.
+func (c *mySQLConnectionProducer) dialWithFailover(ctx context.Context, network, _ string) (conn net.Conn, err error) {
+	var merr *multierror.Error
+
+	for _, host := range c.hosts {
+		var d net.Dialer
+
+		conn, err := d.DialContext(ctx, network, host)
```
JanMa: This accidentally shadows the named return values. It doesn't actually do any harm but I think you should either remove the named return or switch this to a plain assignment
```suggestion
		conn, err = d.DialContext(ctx, network, host)
```

plugins/database/mysql/connection_producer.go
```diff
@@ -218,16 +244,71 @@ func (c *mySQLConnectionProducer) getTLSAuth() (tlsConfig *tls.Config, err error
...
 }
 
-func (c *mySQLConnectionProducer) addTLStoDSN() (connURL string, err error) {
-	config, err := mysql.ParseDSN(c.ConnectionURL)
-	if err != nil {
-		return "", fmt.Errorf("unable to parse connectionURL: %s", err)
+// parseMultiHostDSN extracts multiple hosts from connection URL for failover.
+func (c *mySQLConnectionProducer) parseMultiHostDSN() error {
+	// Match tcp(...) in the connection URL
+	re := regexp.MustCompile(`tcp\(([^)]+)\)`)
```
JanMa, line 250: This will match IPv6 addresses e.g. `[2001:db8:0:0:0:ff00:42:8329]` which will break your port splitting logic. Not sure if anyone actually runs their MySQL servery IPv6 only, but if they do this will not work properly if they fail to specify a port 😄

plugins/database/mysql/connection_producer_test.go
```diff
@@ -316,3 +263,69 @@ func writeFile(t *testing.T, filename string, data []byte, perms os.FileMode) {
...
+		"multiple hosts without ports": {
+			connectionURL:         "user:password@tcp(host1,host2)/test",
+			expectedHosts:         []string{"host1:3306", "host2:3306"},
+			expectedConnectionURL: "user:password@tcp(host1:3306)/test",
+		},
+		"unix socket": {
+			connectionURL:         "user:password@unix(/var/run/mysqld/mysqld.sock)/test",
+			expectedHosts:         nil,
+			expectedConnectionURL: "user:password@unix(/var/run/mysqld/mysqld.sock)/test",
+		},
```
JanMa, line 289: I think this we should add a test here that TLS still works with multiple hosts

plugins/database/mysql/connection_producer.go
cropalato: fair.

plugins/database/mysql/connection_producer.go
cropalato: good point

plugins/database/mysql/connection_producer_test.go
cropalato: All done. Thanks for the feedback.

cropalato:
@JanMa anything else?

JanMa (approved):
Sorry for the delay, LGTM @cropalato

cropalato:
@JanMa and @wslabosz-reply , Do you need me to do anything else?

wslabosz-reply:
@cropalato 
Will look at it this week, in the meantime, can you rebase on top of main? Thanks!

wslabosz-reply:
Summary from my local tests:

(running with 3 mysql instances)

`connection_url="{{username}}:{{password}}@tcp(127.0.0.1:3306,127.0.0.1:3307,127.0.0.1:3308)/"`

while taking down currently first-point-of-contact instance, running
```
bao read database/creds/my-role
```
results with following log:
```
[mysql] 2026/04/01 11:57:36 connection.go:706 closing bad idle connection: EOF
```
which makes sense, the idle connection to instance taken down should be closed.


Taking down all database instances results in following error msg:
```
Error reading database/creds/my-role: Error making API request.

URL: GET http://127.0.0.1:8200/v1/database/creds/my-role
Code: 500. Errors:
* 1 error occurred:
        * 3 errors occurred:
        * failed to connect to 127.0.0.1:3306: dial tcp 127.0.0.1:3306: connect: connection refused
        * failed to connect to 127.0.0.1:3307: dial tcp 127.0.0.1:3307: connect: connection refused
        * failed to connect to 127.0.0.1:3308: dial tcp 127.0.0.1:3308: connect: connection refused
```
which probably could be improved, but not in scope of the PR.

satoqz:
The behavior seems fine to me, one thing that's still missing (and sorry for not bringing this up earlier) is a small update to documentation at `website/content/docs/secrets/databases/mysql-maria.mdx`.

plugins/database/mysql/connection_producer_test.go
```diff
@@ -311,3 +258,89 @@ func writeFile(t *testing.T, filename string, data []byte, perms os.FileMode) {
...
+		})
+	}
+}
+
+func Test_dialWithFailover(t *testing.T) {
+	producer := &mySQLConnectionProducer{
+		hosts: []string{"invalid-host-1:3306", "invalid-host-2:3306"},
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
satoqz:
```suggestion
	ctx, cancel := context.WithTimeout(t.Context(), 2*time.Second)
```

cropalato: changed

changelog/2312.txt
cropalato: replaced.

cropalato:
@wslabosz-reply @satoqz @cipherboy thanks for the comments. All suggestions were applied, and the code rebased.

Issue #2319, opened 2026-01-19
Using cross namespace identiy causes incosistent data in storage


phil9909:
**Describe the bug**

When using cross namespace identity, corrupt data might be written to the storage (while the in-memory representation is correct until restart or leadership change). 

**To Reproduce**

Script

```bash
#!/usr/bin/env bash

set -exuo pipefail

bao namespace create test

# create child group in root
bao write identity/group name=external_group_root type=external
EXTERNAL_GROUP_ROOT_ID=$(bao read -format=json identity/group/name/external_group_root | jq -r .data.id)

# create parent group in test namespace
bao write -namespace=test identity/group name=internal_group_namespaced type=internal member_group_ids=${EXTERNAL_GROUP_ROOT_ID}

# verify: everything looks correct
bao read -format=json                 identity/group/name/external_group_root       | jq .data.parent_group_ids
bao read -format=json -namespace=test identity/group/name/internal_group_namespaced | jq .data.member_group_ids

# transfer leadership
bao operator step-down -address=$(bao operator members -format=json | jq -r '.Nodes[] | select(.active_node).api_address')
sleep 10

# verify: inconsistent data
bao read -format=json                 identity/group/name/external_group_root       | jq .data.parent_group_ids
bao read -format=json -namespace=test identity/group/name/internal_group_namespaced | jq .data.member_group_ids
```

Output:
```raw
+ bao namespace create test
Key                Value
---                -----
custom_metadata    map[]
id                 iv6GtF
locked             false
path               test/
tainted            false
uuid               458b7df0-63a1-9cd8-78db-4ec72cb328a1
+ bao write identity/group name=external_group_root type=external
Key     Value
---     -----
id      fe318448-e174-10e3-e6b3-389d3857136e
name    external_group_root
++ bao read -format=json identity/group/name/external_group_root
++ jq -r .data.id
+ EXTERNAL_GROUP_ROOT_ID=fe318448-e174-10e3-e6b3-389d3857136e
+ bao write -namespace=test identity/group name=internal_group_namespaced type=internal member_group_ids=fe318448-e174-10e3-e6b3-389d3857136e
Key     Value
---     -----
id      4645eea3-2427-18ba-2902-6949b652b53d.iv6GtF
name    internal_group_namespaced
+ bao read -format=json identity/group/name/external_group_root
+ jq .data.parent_group_ids
[
  "4645eea3-2427-18ba-2902-6949b652b53d.iv6GtF"
]
[...]

PR #2321, opened 2026-01-20
Fix cross namespace identiy corrupts data

phil9909:
## Description

This changes two things:
- `UpsertGroupInTxn` will return an error if the context namespace does not match the group namespace
- `sanitizeAndUpsertGroup` will (temporarily) update the context namespace, if the member group has a different namespace.

in #2319 I had a simpler fix proposed (In `UpsertGroupInTxn` overwrite the context namespace with whatever is configured in the group), but I guess this masks programming errors and I like this more explicit approach better (it's also easier to test).

There is also separate commit adding an e2e test, this was useful for local testing, but I guess we can/should drop this before merging.

## Rationale

See https://github.com/openbao/openbao/issues/2319 

We should keep the issue open. In a follow up PR I will propose some logic to cleanup the storage after upgrade.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/external_tests/identity/cross_namespace_test.go
+++ b/vault/external_tests/identity/cross_namespace_test.go
@@ -78,7 +78,6 @@ func TestUnsafeCrossNamespaceIdentity(t *testing.T) {
 
 			childClient, err := client.Clone()
 			require.NoError(t, err)
-			require.NoError(t, err)
 			if tc.childNamespaced {
 				childClient.SetNamespace("test")
 			}


phil9909:
Test passes locally:

```bash
$ go test ./builtin/logical/rabbitmq/ -run 'TestBackend_basic'
ok  	github.com/openbao/openbao/builtin/logical/rabbitmq	50.894s
```

looks like a flake to me

suprjinx (approved):
LGTM

vault/external_tests/identity/cross_namespace_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+			parentClient, err := client.Clone()
+			require.NoError(t, err)
+			if tc.parentNamespaced {
+				parentClient.SetNamespace("test")
+			}
+
+			childClient, err := client.Clone()
+			require.NoError(t, err)
+			require.NoError(t, err)
```
JanMa: I think requiring no error once is sufficient 🙂 
```suggestion
			require.NoError(t, err)
```

phil9909: Better safe than sorry, maybe the first check gets influenced by cosmic rays :laughing: 

Fixed, thanks.

phil9909:
@openbao/openbao-auth-jwt-committers @openbao/openbao-k8s-committers Can I get a review on this?

We discussed in the DEV WG that this should land before 2.5.0 GA release.

phil9909:
rebased to fix a merge conflict in `testing.go`

cipherboy (approved):
This makes sense to me.

Issue #651, opened 2024-10-22
Feature Request: Support for mysql storage backend


ajaygk95:
**Is your feature request related to a problem? Please describe.**
This is a new feature request to add support for mysql as a storage backend

**Describe the solution you'd like**
Vault had support for mysql backend, and we were looking into the same options in OpenBao. 
Going over https://github.com/orgs/openbao/discussions/64 and https://github.com/openbao/openbao/pull/467 OpenBao currently supports in-mem, file, raft and postgres backend only.

https://github.com/orgs/openbao/discussions/64 - Describes additional storage backends can be added as plugin to OpenBao ? How do we do this for a new storage backend ?

cipherboy:
Hello @ajaygk95 -- the [recording from last week](https://lf-edge.atlassian.net/wiki/spaces/OP/pages/15211863/OpenBao+Meetings) was posted btw.

From a _community_ PoV I think there's two components that need to be done:

1. Write [a RFC](https://github.com/openbao/openbao/issues/new?assignees=&labels=rfc%2Cpending-decision&projects=&template=rfc.yml) detailing the proposed changes. You can use the template there but post it here (either in a follow-up comment or in the main text of the issue).

2. We also need to determine a support policy for external storage plugins. My thought is best-effort: we can't yet rule out more changes to the storage protocol, which might break external plugins or might require more changes for them to reach parity with the two built-in storage backends. 

I'd ask what the need for MySQL is over PostgreSQL, given they are otherwise fairly equivalent databases.


From a _technical_ PoV: look at the [database](https://github.com/openbao/openbao/tree/main/sdk/database/dbplugin), [auth/secret engine](https://github.com/openbao/openbao/tree/main/sdk/plugin), and even [go-kms-wrapping](https://github.com/openbao/go-kms-wrapping/tree/main/plugin) plugin interfaces built on [go-plugin](https://github.com/hashicorp/go-plugin).

In particular, you'll end up creating a GRPC client/server model. The external storage engine will need to be built against this SDK and the output artifact will be a binary which runs a GRPC client. OpenBao will be the GRPC server and will start the binary based on information stored in the server config. A small third portion, under `physical/external` will need to be crafted to actually use the GRPC server interface to translate `sdk/physical/` `Storage` and `TransactionalStorage` APIs into GRPC calls.

The MySQL storage engine would be third-party; we can consider including it under the OpenBao organization later for others to use (once it is built and stable), but would likely never be added back to the main binary as it is redundant.

[...]

Issue #2184, opened 2025-12-01
Physical storage on Azure Table (working fork)


axi92:
**Is your feature request related to a problem? Please describe.**
We are using vault right now and want to move over to openbao but want an cloud data storage as the physical storage engine.

**Describe the solution you'd like**
We wrote a feature that makes openbao connect to Azure Table Storage.

**Explain any additional use-cases**
Its an additional physical storage engine.

**Additional context**
Now we have some questions on how to proceed further.

I haven't opened a PR because I first want to make sure that feature could be merged.
That is the fork right now: https://github.com/evva-sfw/openbao

Then I have questions about the tests.
I saw that the postgres tests are using a docker helper. Is it ok to write a helper to start an azurite (azure service mocking) container during the tests?

cipherboy:
Hey @axi92! I don't know that we're likely to accept this in the short term. There's still a lot of things up in the air about the storage layer and while it looks deceptively simple, we've learned a lot of edge conditions just reintroducing PostgreSQL alongside integrated storage.

Long term, I think this will look like the auth/secret engine grpc interface (as discussed in the original roadmap: #569), but for now every couple of releases are adding changes to storage which will be hard to stabilize for external engines. 

Assuming we get full parity between PostgreSQL, why would you prefer Azure Data Table over a managed PSQL offering?


I'm happy to chat more, if you're interested. The horizontal scalability working group would likely be the best venue for that.

dwimberger:
Hey there,

>Assuming we get full parity between PostgreSQL, why would you prefer Azure Data Table over a managed PSQL offering?

Because Data Tables are
- easy for us to get (corporate already has mechanisms and policies in place for Azure storage service usage)
- super simple (Key-Value store, no running instances) and fast
- really cheap, storage based pricing (compared to current Postgres offerings; let's see how Horizon DB will be priced)
- easily scalable and resilient (to needs, [LRS, ZRS, GRS, GZRS](https://learn.microsoft.com/en-us/azure/storage/common/storage-redundancy))

>while it looks deceptively simple, we've learned a lot of edge conditions

We would be really interested in understanding these issues and try to address them. 

>The horizontal scalability working group would likely be the best venue for that.

Where is this discussion going on? 
[Here?](https://linuxfoundation.zulipchat.com/#narrow/channel/532998-openssf-openbao-wg-scalability)

cipherboy:
Hey @dwimberger -- few replies:

[...]

PR #2329, opened 2026-01-22
add mysql backend

silenceper:
## Description

This PR adds MySQL as a physical storage backend for OpenBao, enabling seamless migration from HashiCorp Vault 1.14.1 to OpenBao. Users can migrate by simply replacing the binary and restarting the service without any data migration.

## Rationale

Resolves: #651

## Acknowledgements

       also allowed.

### DIFF
--- a/command/commands.go
+++ b/command/commands.go
@@ -37,6 +37,7 @@ import (
 	logicalDb "github.com/openbao/openbao/builtin/logical/database"
 	logicalKv "github.com/openbao/openbao/builtin/logical/kv"
 
+	physMysql "github.com/openbao/openbao/physical/mysql"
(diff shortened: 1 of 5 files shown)


JanMa:
Hello @silenceper ,

thank you for your PR but we're not going to be able to accept it as is. After a quick comparison it seems to me this re-adds the code for the MySQL storage backend from Vault 1.14x (before BUSL). Since then we made a lot of improvements to OpenBao's storage engine which are not supported in the code you add (Transactional storage for example). 

The long term direction of how we want to handle additional storage engines is also not completely clear yet. The discussion in https://github.com/openbao/openbao/issues/2184 goes into some details about the discussion we're having internally.

Therefore I'm going to close this PR, please open an issue if you're interested in discussing this further.

Issue #2325, opened 2026-01-21
Raft configuration API invalid response in 2.5.x


rfwatson:
**Describe the bug**

In 2.5.x, `bao operator list-peers` returns invalid information about whether a peer is a leader or a follower.

Specifically, the output always claims that the **current node** is the leader, even when it's not. 

This is caused by this code which assumes that the request is always being served from the leader. This was a valid assumption in 2.4.x and earlier, but no longer holds in 2.5.x.

https://github.com/openbao/openbao/blob/e42d18d6ad312e07075bf9ed6ed49e4c500d50f6/physical/raft/raft.go#L1408-L1410

Note that setting `disable_standby_reads=true` in config also fixes the issue.

**To Reproduce**

1. Spin up a multi-node cluster
2. Run `bao operator raft list-peers` from any follower
3. Note that the current node (a follower) is listed as the leader

**Expected behavior**
The actual leader node should be listed as the leader, not the follower.

**Environment:**
N/A

PR #2335, opened 2026-01-23
Fix timing issue in `runAutoseal` and `runShamir` test functions

phyrog:
## Description

Fixes a timing issue in `runAutoseal` and `runShamir`. One test affected by this is `TestSealMigration_ShamirToTransit_Post14` (currently failing on main). The underlying issue has been present for longer, but hidden by #2325, as each node reported its own node as leader, regardless of raft cluster state. With #2331 this is no longer the case.

With this PR, we will no longer wait a fixed period of time (15 seconds), but periodically check the result of `VerifyRaftConfiguration` until either no error is returned, or a timeout of 3 minutes has been reached.

## Acknowledgements

       also allowed.


phyrog:
Not sure if we want to backport this to 2.5.x to fix the tests on that branch or if we're fine with them failing there

Issue #2364, opened 2026-01-29
Loading times for downloads page on website are confusing users


gianklug:
**Describe the bug**
When visiting the OpenBao homepage, the downloads page takes a few seconds to load the releases; currently it looks like the page is empty/unfinished while that is happening.
It happened to me multiple times already that I did not realize I was already on the right page, just had to wait, but then navigated away again.

Additionally, unauthenticated requests to the GitHub API are rate-limited, so it might happen that the request fails.

**To Reproduce**

- Open a private browsing window / clear your cache
- Visit the OpenBao homepage
- Click the "Downloads" button
- Be greeted with a mostly empty page for a few seconds

**Expected behavior**
- The user is informed that the page is still loading

PR #2365, opened 2026-01-29
feat: add loading and error indicator to downloads page

gianklug:
## Description

This PR improves the user experience when trying to download OpenBao from the official website.

When navigating to the `/downloads` page from the home page, the GitHub API takes a few seconds to return the available releases and leaves the impression to the user that the actual downloads page is located somewhere else.

Loading:

image

Error:

image

## Rationale

Resolves: #2364

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2365.txt
+++ b/changelog/2365.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+website: Add loading and error indicators to downloads page
+```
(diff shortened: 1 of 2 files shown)


satoqz (approved):
Very nice, thanks @gianklug!

pree:
I think with versioned docs we could write this information into the gh-pages branch, this would also solve the IPv4 only issue of api.github.com.

satoqz:
Yes, imho, this would optimally just be pregenerated at build time.

PR #2371, opened 2026-01-29
Clean up started external plugins when an error occurs later in the request

phyrog:
## Description

This calls `backend.Cleanup` if an error occurs during `enableCredentialInternal` or `mountInternal`. This will stop and clean up any external plugins that were created during this request.

## Rationale

Forwarding requests from read replicas works currently by trying to write to local storage first and then, if an error occurs, forward it to the leader. If a read replica receives a mount request, it will start the plugin, and if an error occurs, forwards the request. That leaves the plugin process still running, and can be observed in tests by running

```
gotestsum --format=testname ./vault/external_tests/plugin -run '^TestExternalPlugin_AuthMethod$'
```

After finishing (even successfully), you can still see the zombie processes running:

```
ps | grep vault-plugin
51410 ttys000    0:00.05 /private/var/folders/qt/k902my_s6y19_fkrzf_yd83r0000gn/T/633506120/vault-plugin-auth-approle
51411 ttys000    0:00.05 /private/var/folders/qt/k902my_s6y19_fkrzf_yd83r0000gn/T/633506120/vault-plugin-auth-approle
51412 ttys000    0:00.06 /private/var/folders/qt/k902my_s6y19_fkrzf_yd83r0000gn/T/633506120/vault-plugin-auth-approle
51413 ttys000    0:00.05 /private/var/folders/qt/k902my_s6y19_fkrzf_yd83r0000gn/T/633506120/vault-plugin-auth-approle
```

While it's easy to reproduce this using read replica request forwarding, I believe it is not restricted to that scenario.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -73,7 +73,7 @@ func (c *Core) enableCredential(ctx context.Context, entry *MountEntry) error {
 }
 
 // enableCredential is used to enable a new credential backend
-func (c *Core) enableCredentialInternal(ctx context.Context, entry *MountEntry, updateStorage bool) error {
+func (c *Core) enableCredentialInternal(ctx context.Context, entry *MountEntry, updateStorage bool) (err error) {
 	// Ensure we end the path in a slash
 	if !strings.HasSuffix(entry.Path, "/") {
 		entry.Path += "/"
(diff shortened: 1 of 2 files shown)


satoqz:
I must share that Tom had ~1200 zombie processes running a standalone approle auth plugin lingering on his computer when I pointed out that plugin processes hadn't been killed following the debugging session for https://github.com/openbao/openbao/pull/2361 😄

wslabosz-reply (approved):
I think it's fine, as long as the cleanup never takes more than the request context timespan (30s by default), but I guess if the mount (plugin or else) failed to mount then it shouldn't take too long to clean it up

PR #2390, opened 2026-02-03
Skip OCI plugin test on hosts that are not linux/amd64

satoqz:
## Description

Skip `TestReconcileOCIPlugins` if the host system is not linux/amd64.

## Rationale

The test expects static hashes for linux/amd64 binaries and fails if executed on any other type of host. We could add additional hashes for other platforms or dynamically prefetch the expected hashes, though I find that unnecessarily complex, so I suggest we just skip it on other systems. 

## Acknowledgements

       also allowed.


JanMa (approved):
Good catch, thanks @satoqz

Issue #2395, opened 2026-02-03
Plugin Reload returns "Success!" for non-existent plugin


mariusholm-sb1u:
**Describe the bug**
`bao plugin reload -plugin <name>` does not validate if the plugin actually exists and returns `Success! Reloaded plugin: <name>` even if it does not exist.

<img width="717" height="38" alt="Image" src="https://github.com/user-attachments/assets/5375696d-0c5e-4f9a-882e-660fdc5c8850" />

**To Reproduce**
1. Authenticate to OpenBao with CLI
2. Run `bao plugin reload -plugin does-not-exist`
3. See the false log: `Success! Reloaded plugin: does-not-exist`

**Expected behavior**
`bao plugin reload -plugin <name>` should fail if the plugin does not exist. 

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.4.4
* OpenBao CLI Version (retrieve with `bao version`): `OpenBao v2.4.4 (4bfd70723d4f9b82be00e87b8c018ac661dd9b99), built 2025-11-24T19:54:48Z`
* Server Operating System/Architecture: AWS ECS Fargate Linux/X86_64

OpenBao server configuration file(s):

```hcl
ui               = true
cluster_addr     = "https://${PRIVATE_DNS}:8201"
api_addr         = "https://openbao.${HOSTED_ZONE_NAME}:8200"
plugin_directory = "/opt/openbao/plugins"

storage "postgresql" {
  connection_url = "${BAO_PG_CONNECTION_URL}"
  max_parallel   = "40"
  ha_enabled     = "true"
}

listener "tcp" {
  address       = "0.0.0.0:8200"
  tls_cert_file = "/etc/openbao/openbao.crt"
  tls_key_file  = "/etc/openbao/openbao.key"
}

seal "awskms" {
  kms_key_id = "${BAO_AWSKMS_SEAL_KEY_ID}"
}
```

PR #2398, opened 2026-02-04
Return error if no matching plugin is found when reloading plugin

Huy-Dinh-Reply:
Resolves: #2395

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/logical_system.go
+++ b/vault/logical_system.go
@@ -621,12 +621,12 @@ func (b *SystemBackend) handlePluginReloadUpdate(ctx context.Context, req *logic
 	if pluginName != "" {
 		err := b.Core.reloadMatchingPlugin(ctx, pluginName)
 		if err != nil {
-			return nil, err
+			return handleError(err)
 		}
 	} else if len(pluginMounts) > 0 {
 		err := b.Core.reloadMatchingPluginMounts(ctx, pluginMounts)
 		if err != nil {
-			return nil, err
+			return handleError(err)
 		}
 	}
 


wslabosz-reply:
> @Huy-Dinh-Reply I would expect this to check if the plugin exists, not if it is in use anywhere. So IMO this would need a lookup into the plugin catalog (similar to what actually creating a new mount needs to do) and fail/succeed based on that. Thoughts?

I think this is valid as an additional check, like after we conclude that not a single mount that we have is the plugin mount, we can call into registry to verify if it exists there, but not really what the issue was about? But overall I agree on that it would be nice enhancement of the operator exp.

satoqz:
> Expected behavior
> bao plugin reload -plugin <name> should fail if the plugin does not exist.

The issue is very much about ensuring the plugin exists in the system, not that it is used. So a reload of a plugin that exists but is not in use anywhere should still succeed.

Huy-Dinh-Reply:
@satoqz @wslabosz-reply thanks, yeah I think it makes sense to check the plugin catalog and if found, succeed anyway (as "nothing to do") even if the plugin isn't used anywhere. I'll put that in and push again.

Huy-Dinh-Reply:
@wslabosz-reply @satoqz I opted to add another `HasPlugin` function to the catalog to check for availability, because `Get` would require a version, and it looks like `List` would get more expensive the more plugins we have. Let me know what you think.

wslabosz-reply:
@Huy-Dinh-Reply 
quickly glancing over previously existing `plugin_catalog.go` and the plugin registry code, I'm not fond of that code, and you are probably right with direction of proposed changes.

changelog/2398.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+vault/plugin: Fix plugin reload returning success for non-existent plugin
```
satoqz:
```suggestion
sys/plugin: Fix plugin reload returning success for non-existent plugin.
```

vault/plugin_reload.go
```diff
@@ -84,6 +84,15 @@ func (c *Core) reloadMatchingPlugin(ctx context.Context, pluginName string) erro
...
 	}
 
+	// Check if the plugin is in the plugin catalog
+	pluginFound, err := c.pluginCatalog.HasPlugin(ctx, pluginName)
+	if err != nil {
+		return err
+	}
+	if !pluginFound {
+		return fmt.Errorf("%w: %q", ErrPluginNotFound, pluginName)
+	}
```
satoqz: Note that the plugin catalog only holds external binary plugins and dynamically registered ones (kv, database), but `bao plugin reload` also applies to built-in "plugins", e.g., "system". This now fails when trying to reload those plugins.

@Huy-Dinh-Reply I think the best way to get this right is to trace the exact plugin resolution step for mount calls, starting from `mountInternal()`. IIRC this only calls into the plugin catalog as a last resort, and resolves built-in plugins (and any legacy aliases, which we can think about handling) earlier.

satoqz: I believe we also don't need to (or even want to) check for all possible types, as this endpoint relates to reloading mountable plugins (auth, secret) only.

satoqz: Aha! Note that Vault changed their API here, and supports all of auth/secret/database with a new "type" parameter. I think we should eventually match that design: https://developer.hashicorp.com/vault/docs/commands/plugin/reload.

cipherboy: @satoqz I think this is probably enabled by new `sdk` changes that might've broken plugin compatibility. We should see about that... I still want to try and build a "clean-room" (MPL-only!) sdk/api mirror of Vault.

Huy-Dinh-Reply: Thanks!
From our earlier internal discussion: the "singleton" plugins are skipped when trying to reload

```Go
 	singletonMounts = []string{
		mountTypeCubbyhole,
		mountTypeNSCubbyhole,
		mountTypeSystem,
		mountTypeNSSystem,
		mountTypeToken,
		mountTypeNSToken,
		mountTypeIdentity,
		mountTypeNSIdentity,
	}
```

```Go
func (c *Core) reloadBackendCommon(ctx context.Context, entry *MountEntry, isAuth bool) error {
	// Make sure our cache is up-to-date. Since some singleton mounts can be
	// tuned, we do this before the below check.
	entry.SyncCache()

	// We don't want to reload the singleton mounts. They often have specific
	// inmemory elements and we don't want to touch them here.
	if slices.Contains(singletonMounts, entry.Type) {
		c.logger.Debug("skipping reload of singleton mount", "type", entry.Type)
		return nil
	}
```

Therefore, I decided to check for this and fail early if the user is trying to reload mounts of these plugins. Also added the handling for the case when a plugin is in the catalog but should not be reloaded. LMKWYT.

Huy-Dinh-Reply: Looks like there's a test that tries to reload a database plugin and fails, I'll have a look

Huy-Dinh-Reply: After checking I think database reloading is not being supported anyway currently, the call to reload fell through without doing anything. So I opted to skip that test case for now. What do you think?

vault/plugin_reload.go
```diff
@@ -84,6 +84,24 @@ func (c *Core) reloadMatchingPlugin(ctx context.Context, pluginName string) erro
...
+
+	// Check if the plugin is in the plugin catalog and is of a reloadable type
+	types, err := c.pluginCatalog.TypesFromName(ctx, pluginName)
+	if err != nil {
+		return err
+	}
+	if len(types) == 0 {
+		return fmt.Errorf("%w: %q", ErrPluginNotFound, pluginName)
+	}
+	// Only plugin of types auth and secret and secrets can be reloaded
```
satoqz:
```suggestion
	// Only plugin of types auth and secret can be reloaded
```

vault/plugin_reload.go
```diff
@@ -84,6 +84,24 @@ func (c *Core) reloadMatchingPlugin(ctx context.Context, pluginName string) erro
 		return err
 	}
 
+	// Fail early if the plugin is a singleton plugin
+	if slices.Contains(singletonMounts, pluginName) {
+		return fmt.Errorf("cannot reload singleton plugin %q", pluginName)
+	}
+
```
satoqz, line 92: I'd prefer to just modify the original counterpart in `reloadBackendCommon()` to error. This would make `reloadMatchingPluginMounts()` behave consistently with this change. There are some other subsystems that call into `reloadBackendCommon()`, but I don't think any of them do so on a singleton mount.

Huy-Dinh-Reply: Thanks, good point with `reloadMatchingPluginMounts`. I modified `reloadBackendCommon` to err when called on a singleton mount. It probably still makes sense to keep this check here though, otherwise the catalog existence check will fail first and the user will get "not found in catalog" instead of the singleton being reloading error. Thoughts on this?

When trying to trace calls to `reloadBackendCommon`, I noticed that one of the callers, `reloadMountInternalWithLock` (called by `transactionalMountInvalidation`) looks like it may technically be called on a singleton mount (e.g. when user tunes a singleton but not "untunable" mount, like `token`). We are saved from this by this check here within `reloadMountInternalWithLock `
https://github.com/openbao/openbao/blob/cafcaec92b2ef0114de0a3eebe14569b8a3a6910/vault/mount.go#L2190-L2195
since singleton mounts are not defined with `version` in their `Options`, this check also prevents `reloadBackendCommon` from being called on singleton plugins. While this works, this seems we're checking a side effect to me, not sure if it's worth adding an explicit singleton check here or if this existing check is clear enough?

vault/plugin_reload.go
```diff
@@ -97,7 +97,7 @@ func (c *Core) reloadMatchingPlugin(ctx context.Context, pluginName string) erro
 	if len(types) == 0 {
 		return fmt.Errorf("%w: %q", ErrPluginNotFound, pluginName)
 	}
-	// Only plugin of types auth and secret and secrets can be reloaded
+	// Only plugin of types auth and secret can be reloaded
 	if !slices.Contains(types, consts.PluginTypeCredential) && !slices.Contains(types, consts.PluginTypeSecrets) {
 		return fmt.Errorf("plugin %q cannot be reloaded, only plugins of type auth and secret can be reloaded", pluginName)
 	}
```
satoqz: Actually, should we error here? I was thinking we'd just ignore any found types that aren't auth and secrets and pretend they don't exist within the scope of this endpoint. I.e., we'd match like so:

```
[] => error. not found
[auth] => ok
[other] => error, not found
[secret] => ok
[auth, secret] => ok
[auth, secret, ...other] => ok, but ignore other
```

vault/plugin_reload.go
satoqz: > It probably still makes sense to keep this check here though, otherwise the catalog existence check will fail first and the user will get "not found in catalog" instead of the singleton being reloading error. Thoughts on this?

Yeah, let's keep it then.

> While this works, this seems we're checking a side effect to me, not sure if it's worth adding an explicit singleton check here or if this is clear enough?

I agree it's imperfect but I think the version check is sufficient tbh.

vault/plugin_reload.go
Huy-Dinh-Reply: hmm I think this is already what we're already doing, right? if the returned `types` doesn't contain `PluginTypeCredential` nor `PluginTypeSecrets`, we err, otherwise we keep going.

satoqz: I guess just the error message is different, yes. Perhaps that's fine.

wslabosz-reply:
@JanMa 
given your history with plugins, would you like to take a look?

cipherboy (approved):
Thanks @Huy-Dinh-Reply! This makes sense to me. 🎻

satoqz:
@Huy-Dinh-Reply rebased this for you since I caused conflicts with #2586.

PR #2399, opened 2026-02-04
Fix data race when cloning API client

satoqz:
## Description

Cloning the API client was not thread-safe as intended as it did not acquire the read locks before reading the `config.CloneHeaders` field.

## Rationale

See CI run: https://github.com/openbao/openbao/actions/runs/21664187879/job/62455545174.

## Acknowledgements

       also allowed.


wslabosz-reply (approved):
😢

Issue #2413, opened 2026-02-04
expire.num_leases metric not decremented after revoking irrevocable leases


driif:
The issue was detected while using the Nomad plugin due to a bug in the plugin. The bug has already been fixed - https://github.com/openbao/openbao-plugins/pull/48

The Nomad plugin created irrevocable leases. After upgrading the plugin:
- the number of irrevocable leases goes down
- the total number of leases goes down on the backup node
- the total number of leases does not go down on the main cluster

The bug only affects the metric: after revoking irrevocable leases, `metric.num_leases` was not decremented.

**To Reproduce**
```bash
# start docker with postgres

bao secrets enable database

bao write database/config/mydb \
  plugin_name=postgresql-database-plugin \
  allowed_roles="myrole" \
  connection_url="postgresql://{{username}}:{{password}}@localhost:5432/postgres sslmode=disable" \
  username="postgres" \
  password="secret"

bao write database/roles/myrole \
  db_name=mydb \
  creation_statements="CREATE ROLE \"{{name}}\" WITH LOGIN PASSWORD '{{password}}' VALID UNTIL '{{expiration}}';" \
  revocation_statements="DROP ROLE IF EXISTS \"{{name}}\";" \
  default_ttl="30m" \
  max_ttl="60m"

LEASE_1=$(bao read -format=json database/creds/myrole | jq -r '.lease_id')
LEASE_2=$(bao read -format=json database/creds/myrole | jq -r '.lease_id')
LEASE_3=$(bao read -format=json database/creds/myrole | jq -r '.lease_id')

# STOP DOCKER with PostgreSQL !!!

bao lease revoke "$LEASE_1"

# wait 5–15 minutes for 6 failed attempts to connect to the DB
# after that, the leases will be marked as irrevocable

# check metrics
curl -s "$BAO_ADDR/v1/sys/metrics" | jq '.Gauges[] | select(.Name | contains("num_leases"))'
curl -s "$BAO_ADDR/v1/sys/metrics" | jq '.Gauges[] | select(.Name | contains("irrevocable"))'

# Force revoke LEASE_1
./bao lease revoke -force -prefix $LEASE_1

./bao list sys/leases/lookup/database/creds/myrole
```

Check metrics - irrevocable leases should be 0 and num_leases 2 (But returns 3)

PR #2414, opened 2026-02-04
fix decrement of total lease count after irrevocable revocation and a…

driif:
After revocation of irrevocable leases also decrease total lease number in metrics.

Resolves: #2413

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2414.txt
+++ b/changelog/2414.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/expiration: Fix total lease count not being decremented when revoking irrevocable leases
+```


JanMa (approved):

driif:
@cipherboy
> One comment before I merge though @driif:
> 
> https://github.com/openbao/openbao/blob/b72657616acb3b7a37b2317639d2eb232434cfd1/vault/expiration.go#L553-L558
> 
> mirrors this logic, but I wonder if `revokeCommon` needs `m.leaseCount--` outside of this if check? My understanding is `revokeCommon` is called on both paths (irrevokable or regular revoke) and I don't see where `m.leaseCount--` is called otherwise.

https://github.com/openbao/openbao/blob/b72657616acb3b7a37b2317639d2eb232434cfd1/vault/expiration.go#L1065-L1075

It should be inside this check. Otherwise, we use `removeFromPending`, which decrements lease count.

Issue #2418, opened 2026-02-05
ssh subcommand should search for SSH keys if unspecified


ecksun:
**Is your feature request related to a problem? Please describe.**

    bao ssh -mode=ca -role=myrole user@example.com

Will only try using ~/.ssh/id_rsa. However I'd prefer it to have the same logic as ssh normally does, which is to try ed25519 first, followed by ecdsa, followed by rsa lsat.

**Describe the solution you'd like**
If the flags -private-key-path and -public-key-path is unspecified the following files should be tried in order:

- ~/.ssh/id_ed25519{,.pub}
- ~/.ssh/id_ecdsa{,.pub}
- ~/.ssh/id_rsa{,.pub}

**Describe alternatives you've considered**
Using a shell alias works but requires manual work for users.

**Explain any additional use-cases**
I'm not sure if the _sk files should also be included

PR #2419, opened 2026-02-05
Find SSH keys with ssh subcommand

ecksun:
This patch will search for ed25519, ecdsa and rsa keys in order if the -public-key-path or -private-key-path flags are not specified

This should make using the CLI slightly more convenient and more similar to how ordinary SSH works

Resolves: #2418

## Acknowledgements

       also allowed.

### DIFF
--- a/command/ssh.go
+++ b/command/ssh.go
@@ -12,11 +12,13 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"path/filepath"
 	"strings"
 	"syscall"
 
 	"github.com/go-viper/mapstructure/v2"
 	"github.com/hashicorp/cli"
+	"github.com/mitchellh/go-homedir"
 	"github.com/openbao/openbao/api/v2"
 	"github.com/openbao/openbao/builtin/logical/ssh"
 	"github.com/posener/complete"
@@ -156,20 +158,20 @@ func (c *SSHCommand) Flags() *FlagSets {
 	f.StringVar(&StringVar{
 		Name:       "public-key-path",
 		Target:     &c.flagPublicKeyPath,
-		Default:    "~/.ssh/id_rsa.pub",
+		Default:    "",
 		EnvVar:     "",
 		Completion: complete.PredictFiles("*"),
-		Usage:      "Path to the SSH public key to send to OpenBao for signing.",
+		Usage:      "Path to the SSH public key to send to OpenBao for signing. If not set, ~/.ssh/id_ed25519.pub, ~/.ssh/id_ecdsa.pub and ~/.ssh/id_rsa.pub will be searched in order.",
 	})
 
 	f.StringVar(&StringVar{
 		Name:       "private-key-path",
 		Target:     &c.flagPrivateKeyPath,
-		Default:    "~/.ssh/id_rsa",
+		Default:    "",
 		EnvVar:     "",
 		Completion: complete.PredictFiles("*"),
 		Usage: "Path to the SSH private key to use for authentication. This must " +
-			"be the corresponding private key to -public-key-path.",
+			"be the corresponding private key to -public-key-path. If not set, ~/.ssh/id_ed25519, ~/.ssh/id_ecdsa and ~/.ssh/id_rsa will be searched in order.",
 	})
 
 	f.StringVar(&StringVar{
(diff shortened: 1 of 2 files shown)


command/ssh.go
```diff
@@ -246,6 +248,36 @@ func (c *SSHCommand) Run(args []string) int {
...
+			for _, p := range []string{"id_ed25519", "id_ecdsa", "id_rsa"} {
+				keyPath := filepath.Join(home, ".ssh", p)
+				if _, err := os.Stat(keyPath); err == nil {
+					c.flagPrivateKeyPath = keyPath
+					break
+				}
+			}
+		}
+		if c.flagPublicKeyPath == "" {
+			c.UI.Error("Could not find a SSH public key. Specify with -public-key-path")
```
satoqz:
```suggestion
			c.UI.Error("Could not find an SSH public key. Specify with -public-key-path.")
```

command/ssh.go
```diff
@@ -246,6 +248,36 @@ func (c *SSHCommand) Run(args []string) int {
...
+			return 1
+		}
+		if c.flagPrivateKeyPath == "" {
+			c.UI.Error("Could not find a SSH private key. Specify with -private-key-path")
```
satoqz:
```suggestion
			c.UI.Error("Could not find an SSH private key. Specify with -private-key-path.")
```

command/ssh.go
```diff
@@ -156,20 +158,20 @@ func (c *SSHCommand) Flags() *FlagSets {
 	f.StringVar(&StringVar{
 		Name:       "public-key-path",
 		Target:     &c.flagPublicKeyPath,
-		Default:    "~/.ssh/id_rsa.pub",
+		Default:    "",
 		EnvVar:     "",
 		Completion: complete.PredictFiles("*"),
-		Usage:      "Path to the SSH public key to send to OpenBao for signing.",
+		Usage:      "Path to the SSH public key to send to OpenBao for signing. If not set ~/.ssh/id_ed25519.pub, ~/.ssh/id_ecdsa.pub and ~/.ssh/id_rsa.pub will be searched in order.",
```
satoqz:
```suggestion
		Usage:      "Path to the SSH public key to send to OpenBao for signing. If not set, ~/.ssh/id_ed25519.pub, ~/.ssh/id_ecdsa.pub and ~/.ssh/id_rsa.pub will be searched in order.",
```

command/ssh.go
```diff
@@ -156,20 +158,20 @@ func (c *SSHCommand) Flags() *FlagSets {
...
 	f.StringVar(&StringVar{
 		Name:       "private-key-path",
 		Target:     &c.flagPrivateKeyPath,
-		Default:    "~/.ssh/id_rsa",
+		Default:    "",
 		EnvVar:     "",
 		Completion: complete.PredictFiles("*"),
 		Usage: "Path to the SSH private key to use for authentication. This must " +
-			"be the corresponding private key to -public-key-path.",
+			"be the corresponding private key to -public-key-path. If not set ~/.ssh/id_ed25519, ~/.ssh/id_ecdsa and ~/.ssh/id_rsa will be searched in order.",
```
satoqz:
```suggestion
			"be the corresponding private key to -public-key-path. If not set, ~/.ssh/id_ed25519, ~/.ssh/id_ecdsa and ~/.ssh/id_rsa will be searched in order.",
```

command/ssh.go
```diff
@@ -246,6 +248,36 @@ func (c *SSHCommand) Run(args []string) int {
...
+		}
+		if home != "" && c.flagPrivateKeyPath == "" {
+			for _, p := range []string{"id_ed25519", "id_ecdsa", "id_rsa"} {
+				keyPath := filepath.Join(home, ".ssh", p)
+				if _, err := os.Stat(keyPath); err == nil {
+					c.flagPrivateKeyPath = keyPath
+					break
+				}
+			}
+		}
```
satoqz: Suppose we find `id_ed25519.pub`, I think we should directly assert that `id_ed25519` exists and use that private key path deterministically rather than searching paths again. You wouldn't want to introduce confusing results when you find an `id_ed25519.pub` but no `id_ed25519` and then fall back to a present `id_ecdsa` or `id_rsa` in combination with the Ed25519 public key.

cipherboy: I concur fwiw.

ecksun: Yeah, make sense.

That would mean that if one flag is given, say -private-key-path, then we should guess the public key based on that?

i.e. for `-private-key-path foo` we use `foo` as the private key and `foo.pub` as the public key

and for `-public-key-path bar.pub` we use `bar` as the private key

I will fix that

ecksun: alright, should be done now. Does that look like what you were thinking about?

satoqz: Yes, that looks great. Thanks!

satoqz (approved):
Many thanks @ecksun!

PR #2425, opened 2026-02-06
Refactor barrier into own package

phyrog:
## Description

This puts all barrier related code into its own package and renames a couple of types to avoid `barrier.Barrier...` name stutter.

Some previously private constants, vars and functions have to be public now, so they can be used from the `vault` package.

## Rationale

The vault package is too large and interconnected to effectively reason about it. This is part 1 of splitting up `vault` into more manageable and focused packages.

`barrier` is the first one, since it has the least dependencies on other parts of `vault`.

## Acknowledgements

       also allowed.


vault/barrier/view.go
```diff
@@ -0,0 +1,198 @@
...
+import (
+	"context"
+	"errors"
+	"sync"
+
+	"github.com/openbao/openbao/sdk/v2/logical"
+)
+
+// viewCore defines the core operations unique to BarrierView.
+type viewCore interface {
```
cipherboy, line 15: For a later refactor, but should this be refactored from logical iirc?

cipherboy (approved):
Love this, thanks @phyrog! 

Will let @wslabosz-reply &co also ACK in case he wants to hold off given namespace sealing conflicts.

wslabosz-reply (approved):
@phyrog 
feel free to merge, I will probably manage with the synthetic PRs.

Issue #2274, opened 2025-12-29
initialize stanza breaks retry_join in HA clusters


NoamY8:
**Describe the bug**

When using the `initialize` stanza in combination with `retry_join` in a Raft HA cluster (e.g., Kubernetes StatefulSet with 3 replicas), the `initialize` stanza causes a race condition where follower nodes fail to join the leader and instead self-initialize, creating multiple independent clusters.

The `initialize` stanza appears to extend the initialization window of the first node, during which time other nodes attempting `retry_join` fail to connect and fall back to self-initialization. This results in each pod having a different Cluster ID instead of forming a unified cluster.

**To Reproduce**

1. Deploy OpenBao as a Kubernetes StatefulSet with 3 replicas
2. Configure `retry_join` pointing to the headless service:
   ```hcl
   retry_join {
     leader_api_addr = "https://openbao-internal.openbao.svc.cluster.local:8200"
   }
   ```
3. Add an `initialize` stanza with any requests (e.g., enable userpass auth, create policy)
4. Deploy the StatefulSet - all 3 pods start within seconds of each other
5. After pods are running, check each pod's Cluster ID:
   ```bash
   kubectl exec -n openbao openbao-0 -- bao status | grep "Cluster ID"
   kubectl exec -n openbao openbao-1 -- bao status | grep "Cluster ID"
   kubectl exec -n openbao openbao-2 -- bao status | grep "Cluster ID"
   ```
6. **Result**: Each pod has a DIFFERENT Cluster ID - they formed independent clusters instead of joining

**Expected behavior**

All 3 pods should have the same Cluster ID and form a single Raft cluster:
- `openbao-0` initializes first (runs `initialize` stanza)
- `openbao-1` and `openbao-2` successfully join via `retry_join`
- `bao operator raft list-peers` shows all 3 nodes in one cluster

This is exactly what happens when the `initialize` stanza is **removed** - the cluster forms correctly every time.

**Environment:**
* OpenBao Server Version: OpenBao v2.4.4 (4bfd70723d4f9b82be00e87b8c018ac661dd9b99), built 2025-11-24T19:54:48Z
* OpenBao CLI Version: OpenBao v2.4.4
* Server Operating System/Architecture: Linux containers on Kubernetes (EKS), arm64

OpenBao server configuration file(s):

```hcl
ui = true
cluster_addr = "https://POD_IP:8201"
api_addr = "https://POD_IP:8200"
disable_mlock = true
log_level = "debug"

[...]

PR #2442, opened 2026-02-09
Make self-init failures fatal

### LINKED ISSUE #2190: Declarative Self-Init: Make failures fatal
**Describe the bug**

Running a self-init file with an obvious bug (e.g. `sys/policieZZZZZZZZZZZ/acl/magic`) results in a failure:

```
Beginning post-unseal configuration
...
initialize.[gitlab-auth (0)]: request.[manager-role (3)]: failed to evaluate request: 1 error occurred:
	* unsupported path
```

However, subsequent restart is OK:

```
2025-12-03T13:22:19.497-0600 [INFO]  core: usage gauge collection is disabled
2025-12-03T13:22:19.497-0600 [INFO]  core: post-unseal setup complete
```

This causes problems when executing under any auto-retry system (k8s, systemd, ...). If self-init fails, subsequent restart works.

We should probably have some self-init start/finished logic.

**To Reproduce**

1. Run `bao server ...` with bad config
2. Run `bao server ...` again without deleting data
3. See error

**Expected behavior**

OpenBao should fatally fail, forcing any task manager to enter into a retry-fail backoff loop. 

**Environment:**

`main`

```json
{
  "initialize":
  [
    {
      "gitlab-auth": {
        "request": [
          {
            "manager-policy": {
              "allow_failure": false,
              "data": {
                "policy": "path \"*\" { capabilities = [\"create\", \"read\", \"update\", \"delete\", \"list\", \"sudo\"] }"
              },
              "operation": "create",
              "path": "sys/policies/acl/secrets_manager"
            }
          }
        ]
      }
    }
  ]
}
```

**Additional context**

Discovered by @fcat

EnricoFusi:
## Description

This PR fixes a critical panic during the core self-initialization process by properly propagating `context.Context` through the `MarkSelfInitComplete` workflow, instead of relying on detached contexts or implicit backgrounds.

## Rationale

Resolves: #2190

The panic occurs because the physical backend requires a valid context, but the previous implementation didn't supply one correctly during self-init.

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI...
 - [x] By contributing this change, I certify I have signed-off on the DCO ownership...


cipherboy:
@EnricoFusi You can drop the Go 1.25.7 upgrade; that was handled in #2426.

vault/core.go
```diff
@@ -3962,3 +3955,32 @@ func (c *Core) DetectStateLockDeadlocks() bool {
...
+
+// MarkSelfInitComplete persists the success state.
+func (c *Core) MarkSelfInitComplete(ctx context.Context) error {
+	if c.physical == nil {
+		return fmt.Errorf("physical backend missing")
+	}
+	// Direct write to physical storage. Bypass barrier.
+	return c.physical.Put(ctx, &physical.Entry{
+		Key:   coreStatusSelfInit,
+		Value: []byte("true"),
```
cipherboy: I think we need two sentinel values: whether or not initialization has started and whether or not initialization has finished. If we only store finished status, we don't have a viable codepath for new line 1752: we never write false into storage.

I think order is:

1. Check status
2. Track initialization (started=true, finished=false)
3. Actually do self-init
4. Finalize initialization (started=true, finished=true)

vault/core.go
```diff
@@ -3962,3 +3955,32 @@ func (c *Core) DetectStateLockDeadlocks() bool {
...
+	coreStatusSelfInit = "core/status/self-init-completed"
+)
+
+// MarkSelfInitComplete persists the success state.
+func (c *Core) MarkSelfInitComplete(ctx context.Context) error {
+	if c.physical == nil {
+		return fmt.Errorf("physical backend missing")
+	}
+	// Direct write to physical storage. Bypass barrier.
+	return c.physical.Put(ctx, &physical.Entry{
```
cipherboy: I think it would be best to write these with barrier. We'll have it available if initialization is detected (new line 1752 code path), and can delay writing the check until after barrier initialization but before self-init is run.

.changelog/2190.txt
```diff
@@ -0,0 +1,2 @@
+release-note:bug
+CORE: Fix panic in self-init workflow (GH-2190)
```
cipherboy: This could perhaps look like:

    ```release-note:bug
    core: Fail self-initialization if a previous self-initialization failed to run
    ```
    ```release-note:bug
    core/profiles: Fix panic in self-init when a request succeeds with a nil response
    ```

Either way, formatting is slightly off :-)

cipherboy (commented):
Some thoughts @EnricoFusi!

vault/core.go
EnricoFusi: Thank you for the feedback.

Regarding the state tracking: I deliberately opted for a minimal implementation to align with the principles discussed in Issue #1340 (Declarative Self-Initialization). The goal there was to favor idempotency and a declarative approach over introducing complex state machines during the initialization phase, which can lead to stale states during partial failures.

However, if the project's direction or requirements have evolved since that RFC, I am happy to refactor this to include the two-state tracking (`started`/`finished`) as you suggested. Please let me know if I should proceed with that change.

vault/core.go
EnricoFusi: Wait, I see a potential issue regarding the read path on subsequent restarts.

If we write this check into the barrier (encrypted), we can only read it back AFTER the vault is unsealed.
However, the logic to determine "Should I run self-init?" usually needs to happen very early, potentially before unsealing is complete (or to decide if we need to initialize at all).

If I move this to `c.barrier.Put`, how do we check this state on a fresh restart of a sealed node? Won't `c.barrier.Get` fail or be inaccessible until unseal happens?
If the intention is to check this state *only* during the active/unsealed phase, then `barrier` is fine. But if this flag is needed to gate the initialization process itself on restart, don't we need it in `physical` (plaintext) so we can read it before having the keys?

vault/core.go
cipherboy: I think it is more of a bug, tbh, than that project requirements have changed.

I bet if you were to run self-init on v2.5.0 (before this fix) and then resume the cluster on this new version, you'd hit the error on new line 1753: the entry won't exist in storage because it wasn't written at self-init time.

Thus I think we need a "softer" sentinel than whether the storage entry exists.

My 2c.

vault/core.go
cipherboy: > However, the logic to determine "Should I run self-init?" usually needs to happen very early, potentially before unsealing is complete (or to decide if we need to initialize at all).

Note that for self-init to have run you're using auto-unseal and not using manual unseal. So you _should_ have keys available and auto-unseal _should've_ run. 

See e.g., existing line 1730 (`self-initialization requires auto-unseal ...`). 

I had tracked this through, once, as to how early auto-unseal ran, and the answer was really early :-D If we're concerned, though, we could grab the state lock which should prevent sealing during this check.

vault/core.go
EnricoFusi: I see your argoment, and I agree. Relying on a new explicit storage entry that wasn't present in previous versions would cause false negatives for existing clusters upgraded to this version.

I agree that a "softer" sentinel is better. I will refactor the check to rely on standard initialization artifacts (like the presence of the barrier keyring or simply `c.core.IsInitialized()` status) instead of a custom marker entry. This ensures that any previously initialized cluster is correctly recognized as "done" without needing the new storage key.

Updating the PR.

EnricoFusi:
I am closing this PR to supersede it with a cleaner implementation that better aligns with the current OpenBao initialization architecture and backward compatibility requirements.

### The Architectural Conflict
During the review and implementation of this fix, a fundamental conflict emerged between **Robust Failure Handling** and **Upgrade Compatibility**:

1.  **Partial Failure State:** The current self-initialization process is monolithic. It generates keys (Phase 1) and then applies configuration (Phase 2) using an ephemeral Root Token.
    *   If Phase 2 fails, the keys are already persisted (Phase 1 complete).
    *   On restart, OpenBao detects the keys (`core/keyring`) and assumes initialization is complete, skipping the failed configuration.
    *   Crucially, because the Root Token was ephemeral and lost on crash, **we cannot retry Phase 2** automatically on restart. The node starts in an "initialized but unconfigured" state.

2.  **The "Sentinel" Problem:** To detect and handle this partial state, we would need a specific persistence marker (e.g., `self-init-completed`).
    *   However, introducing this marker breaks backward compatibility for existing clusters (which lack the marker), causing them to panic or re-enter initialization logic upon upgrade.
    *   As discussed, we cannot assume a cluster is "broken" just because it lacks this new marker.

### Conclusion
We cannot solve the "Partial Failure Retry" problem within the scope of a bugfix PR without a larger RFC to restructure how bootstrap tokens are persisted or recovered.

### Next Steps
I will open a new, simplified PR that focuses **strictly on the original panic fix** for the first run, while accepting the "Keyring Check" as the source of truth for initialization status.

This means:
*   **Panic on First Fail:** We will still ensure that a configuration failure during the *initial* bootstrap causes a panic (preventing a silent failure on day 0).
*   **Softer Sentinel:** We will rely on `core/keyring` existence for restart checks, prioritizing upgrade safety over partial failure detection on restart.

Thank you @cipherboy for the detailed feedback that highlighted these edge cases.

EnricoFusi:
After some fruitful conversation with @cipherboy, I applied the solution discussed, then reopening the PR.

vault/core.go
EnricoFusi: Thanks for the context on auto-unseal timing. I stuck with the physical backend for the status markers intentionally to decouple the 'should I init?' logic from the unseal status entirely. This avoids any chicken-and-egg dependency on whether auto-unseal has finished or succeeded yet.

Regarding the state lock: Since this check happens sequentially in the startup flow before the server is fully active, I think the race condition is theoretical. Adding the lock would require exposing internal core locks which seems invasive for this specific check. Unless you strongly prefer adding it, I'd suggest keeping the current simpler implementation, with two states.

wslabosz-reply:
@EnricoFusi
can you please fix DCO? (how to [here](https://github.com/openbao/openbao/pull/2442/checks?check_run_id=63076309699)) Thanks!

command/server.go
```diff
@@ -1753,6 +1762,12 @@ func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) erro
 	barrierConfig := &vault.SealConfig{}
 	recoveryConfig = &vault.SealConfig{}
 
+	// 1. Mark that we are STARTING self-initialization.
+	// This creates the "started" marker on physical storage.
+	if err := core.MarkSelfInitStarted(ctx); err != nil {
+		return fmt.Errorf("failed to mark self-init started: %w", err)
+	}
+
```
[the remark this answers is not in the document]
EnricoFusi: Solved, code updated.

.changelog/2190.txt
EnricoFusi: Fixed.

vault/core.go
EnricoFusi: Implemented as requested.

command/server.go
```diff
@@ -1762,19 +1762,29 @@ func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) erro
...
-
+		core.Logger().Error("failed to initialize: unexpected error occurred")
 		return fmt.Errorf("self-initialization failed: %w", err)
 	}
 
+	// We decide init worked
+	if err := core.MarkSelfInitComplete(ctx); err != nil {
+		return fmt.Errorf("failed to mark self-init completed: %v", err)
+	}
+
```
JanMa: How can we decide init worked here if `c.doSelfInit` didn't run yet?

vault/core.go
```diff
@@ -3962,3 +3955,74 @@ func (c *Core) DetectStateLockDeadlocks() bool {
...
 	return false
 }
+
+const (
+	// coreStatusSelfInitStatus is the key for the initialization state machine.
+	// Values:
+	// - (missing): Legacy cluster or Manual Init (Success assumed)
+	// - "started": Auto-init attempted but not finished (Partial Failure)
+	// - "completed": Auto-init finished successfully
+	coreStatusSelfInitStatus = "core/status/self-init-status"
```
JanMa: Can we please: 
- call this `coreStatusSelfInit` and remove the superfluous `Status` suffix
- remove the superfluous `-status` suffix in the value
- move this block to the other `const` definitions at the top of the file
- introduce constants for the expected values (started, completed)

vault/core_self_init_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"sync"
+	"testing"
+
+	"github.com/openbao/openbao/sdk/v2/physical"
+)
+
+// --- CONSTANTS MOCK ---
+// Must match the key used in core.go implementation
+const (
+	mockStatusKey = "core/status/self-init-status"
```
JanMa: why do you introduce a new constant here and don't reuse the one defined in core.go?

vault/core_self_init_test.go
```diff
@@ -0,0 +1,223 @@
...
+	return nil
+}
+
+func (m *MockPhysicalBackend) List(ctx context.Context, prefix string) ([]string, error) {
+	return nil, nil //nolint:nilnil
+}
+
+func (m *MockPhysicalBackend) ListPage(ctx context.Context, prefix string, page string, limit int) ([]string, error) {
+	return nil, nil //nolint:nilnil
+}
```
JanMa: Why exactly do we need this mock backend? As far as I can tell your tests should work fine with the InMem backend that is used in all other core tests

vault/core_self_init_test.go
JanMa, line 1: Overall I get the impression that the tests added in this file don't add much value. They more or less just test if reading and writing a key to storage work, and not if the full self-init with the added logic does. Maybe it's worth a try to write a full external test for it in `vault/external_tests`?

vault/core.go
JanMa: @EnricoFusi how exactly did you implement this as requested? Your code is not going through the barrier as suggested by Alex?

JanMa (changes requested):
I think this needs some more work before we can merge. Also I get the strong impression that at least some of this code was AI generated. @EnricoFusi can you please elaborate on if and how you used AI to create this PR?

command/server.go
EnricoFusi: Hi Jan, yes, I think both calls must be in doSelfInit,
I was confused by the name  core.Initialize where I put the markers, not just the one you mentioned.

About AI, no, I am not using it, all was my personal mistake I confused doSelfINit with Initialize in my mind. 
AI makes bigger mistakes, AFAIK. 
I am going to fix it, thanks for the feedback.

vault/core_self_init_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+// openCore returns a *Core with an initialized, unsealed security barrier.
+// All state machine methods require the barrier to be open.
+//
+// Uses TestCoreNewSeal (auto-unseal capable) + TestCoreInit + UnsealWithStoredKeys.
+// TestCoreUnseal is NOT used here because it uses the Shamir path; auto-unseal
+// cores must be unsealed via UnsealWithStoredKeys which reads the stored keys
+// written by core.Initialize.
+func openCore(t *testing.T) *Core {
```
wslabosz-reply: this function can be replaced by `TestCoreUnsealed()`

vault/core_self_init_test.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+// rootCtx returns the context used throughout the OpenBao core for root-namespace
+// operations.
+func rootCtx() context.Context {
```
wslabosz-reply: not sure I understand, but the comment about this function is wrong, and it's currently working as an alias to `context.Background()`

vault/core_self_init_test.go
```diff
@@ -0,0 +1,249 @@
...
+//     c.barrier (not c.physical). A sealed or uninitialized barrier returns
+//     ErrBarrierSealed on any Get/Put.
+//   - Tests live in package vault (white-box) because the state machine
+//     operates on c.barrier, which is unexported.
+//   - Each test exercises exactly one observable behaviour.
+//   - Accuracy level 5+: every assertion has a comment explaining the invariant.
+//
+// NOTE: TestInitialize_SetsCompletedMarker was removed from this file.
+// Its coverage is provided by command/server_initialize_test.go which exercises
+// the full command.Initialize() path including MarkSelfInitStarted/Complete.
```
wslabosz-reply: was there a test like that before this PR?

command/server_initialize_test.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/openbao/openbao/sdk/v2/physical"
+	physInmem "github.com/openbao/openbao/sdk/v2/physical/inmem"
+	vault "github.com/openbao/openbao/vault"
+	"github.com/stretchr/testify/require"
+)
+
+// newTestServerCommand returns a minimal *ServerCommand suitable for calling
+// Initialize. Mirrors testServerCommand in server_test.go and adds a logger
+// to avoid nil panic in doSelfInit → profiles.WithLogger(c.logger.Named(...)).
+func newTestServerCommand(tb testing.TB) *ServerCommand {
```
wslabosz-reply: nit: you could update the `testServerCommand()` to have a logger and use that one instead

EnricoFusi: FIxed.

vault/core_self_init_test.go
EnricoFusi: COmment refers to a WIP test. Not there anymore.

vault/core_self_init_test.go
EnricoFusi: I corrected the comments and the implementation.

command/server.go
```diff
@@ -1731,50 +1732,67 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
...
 // or if the core is already initialized.
 func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) error {
-	// Skip initialize for dev server as it is handled in initDevCore
-	if len(config.Initialization) == 0 || c.flagDev {
+	if len(config.Initialization) == 0 {
 		return nil
 	}
 
+	// Skip the consistency check if the core is still sealed: we are a
+	// follower or standby node coming up after another node already ran
```
wslabosz-reply: Nomenclature here is hard, ~but I reckon that follower == standby nodes, the difference might be:~

~read-enabled standby nodes vs "passive" (non-handling any requests) nodes
voter vs non-voter nodes (raft leadership election consensus algorithm)~

but in all these cases, we are talking about standby nodes. I hope this explains a bit, so to my point, I'd remove "follower" from the comment.

EDIT: I've got to correct my misunderstanding here, so:
- follower is a node in a setup that uses consensus (raft)
- standby node exists in any HA setup

no need to separate "follower" as different case in this context

command/server.go
```diff
@@ -1731,50 +1732,67 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
...
+	// initialization. IsSelfInitComplete reads through the barrier and
+	// would return "Vault is sealed"; the normal unseal flow handles this.
+
```
wslabosz-reply: this comment appears again on line :1758, which I think is the correct place for it?

command/server.go
```diff
@@ -1731,50 +1732,67 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
...
+		if core.Sealed() {
+			return nil
+		}
+		complete, err := core.IsSelfInitComplete(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to verify self-init consistency: %w", err)
+		}
+		if !complete {
+			return fmt.Errorf("FATAL: Storage inconsistency detected. OpenBao is initialized (barrier exists) but Self-Initialization failed or was interrupted. Manual storage wipe required")
+		}
```
wslabosz-reply: I'd phrase it as such, I'm up for discussion, but I think overly "robust" explanation of the guard clauses make unclear cause -> effect  explanations
```suggestion
		// IsSelfInitComplete would fail at reading as barrier is sealed, skip the consistency check; we are a
		// standby coming up after another node ran initialization.

		if core.Sealed() {
			return nil
		}
		
		complete, err := core.IsSelfInitComplete(ctx)
		if err != nil {
			return fmt.Errorf("failed to verify self-init consistency: %w", err)
		}
		if !complete {
			return fmt.Errorf("FATAL: Storage inconsistency detected. OpenBao is initialized (barrier exists) but Self-Initialization failed or was interrupted. Manual storage wipe required")
		}
```

command/server.go
```diff
@@ -1731,50 +1732,67 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
...
 			return nil
 		}
-
+		core.Logger().Error("failed to initialize: unexpected error occurred")
 		return fmt.Errorf("self-initialization failed: %w", err)
 	}
-
+	// 1. Mark that we are STARTING self-initialization.
+	// This creates the "started" marker on physical storage.
+	// We know we have barriers 'cause we just initialized
```
wslabosz-reply: again up for discussion, but I think this comment could be shorten to
```suggestion
	// Write "started" marker to storage (self-init started).
```

command/server.go
```diff
@@ -1786,9 +1804,17 @@ func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) erro
 	if !isLeader {
 		return fmt.Errorf("initializing node did not win leadership election: ensure only one active, voting node during initialization")
 	}
-
 	// Now perform the component requests of self-initialization.
-	return c.doSelfInit(core, config, init.RootToken)
+	if err := c.doSelfInit(core, config, init.RootToken); err != nil {
+		return err // Original patch: Fail fast on config error
```
wslabosz-reply: comment like that lacks context, I'd remove that.
```suggestion
		return err // Fail fast on config error
```

command/server.go
```diff
@@ -1786,9 +1804,17 @@ func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) erro
...
+	}
+	// Mark self-init as fully complete only after doSelfInit succeeds.
+	// This is the single authoritative point of completion.
+	// We are 100% sure we are initialized
```
wslabosz-reply: To be consistent with earlier comment we should do something along the lines of:
```suggestion
    // Write "completed" marker to storage (self-init complete).
```

command/server_initialize_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+//go:build !race && !hsm
+
+package command
+
+// Tests for (*ServerCommand).Initialize in command/server.go.
+//
+// Design rationale:
+//   - Uses vault.TestCoreNewSeal(t) to get a Core with RecoveryKeySupported()==true,
+//     which is the precondition enforced by Initialize.
```
wslabosz-reply: I think this is a comment that describes code, but doesn't really explain anything to someone who haven't yet read the code /my 2c.

```suggestion
//   - Requires `Core` with an auto-unseal method setup, as self-init doesn't work with shamir seal.
```

command/server_initialize_test.go
```diff
@@ -0,0 +1,315 @@
...
+//   - Uses physInmem backend (embedded in TestCoreNewSeal via TestCoreWithSeal).
```
wslabosz-reply: I don't think type of backend is relevant to this test

command/server_initialize_test.go
```diff
@@ -0,0 +1,315 @@
...
+//   - Does NOT test doSelfInit internals (profile engine): that is out of scope here.
+//     We use an empty Initialization slice for happy-path tests so doSelfInit
+//     short-circuits via profiles.NewEngine with zero requests.
+//   - Each test exercises exactly one observable behaviour of Initialize.
+//   - Build tags match server_test.go: !race !hsm.
```
wslabosz-reply: is this information necessary?

command/server_initialize_test.go
```diff
@@ -0,0 +1,315 @@
...
+//
+// Pattern matches: command/server_test.go (testServerCommand helper)
+// Core construction: vault/testing.go TestCoreNewSeal()
```
wslabosz-reply: In my mind such comments might be useful during one's exploratory work on the matter, but I don't see the value of having those in the source material /my 2c.

command/server_initialize_test.go
```diff
@@ -0,0 +1,315 @@
...
+	// Simulate a previously successful self-init: initialize the core and
+	// write the completed marker.
+	vault.TestCoreInit(t, core)
+	require.NoError(t, core.UnsealWithStoredKeys(ctx))
+	require.NoError(t, core.MarkSelfInitComplete(ctx))
+
+	err := cmd.Initialize(core, emptyInitConfig())
+	require.NoError(t, err,
+		"already-initialized core with completed marker must return nil")
+}
```
wslabosz-reply, line 118:
```suggestion
	t.Parallel()

	_, cmd := testServerCommand(t)
	core := vault.TestCoreNewSeal(t)

	// Simulate a previously successful self-init: initialize the core and
	// write the completed marker.
	vault.TestCoreInit(t, core)
	require.NoError(t, core.UnsealWithStoredKeys(rootCtxCmd()))
	require.NoError(t, core.MarkSelfInitComplete(rootCtxCmd()))

	err := cmd.Initialize(core, emptyInitConfig())
	require.NoError(t, err,
		"already-initialized core with completed marker must return nil")
}
```

wslabosz-reply: applies to other tests also, no need to locally bind the ctx

command/server_initialize_test.go
wslabosz-reply, line 1: I like the descriptions of the tests, they are useful in understanding the intention behind every test.

vault/core_self_init.go
```diff
@@ -0,0 +1,99 @@
...
+// SPDX-License-Identifier: MPL-2.0
+package vault
+
+// Auto-init state machine persisted to the security barrier.
+// Tracks whether auto-initialization was attempted and completed successfully.
+//
+// All three methods require the barrier to be initialized and unsealed:
+//   - MarkSelfInitStarted  : called after core.Initialize(), barrier just created and open
+//   - MarkSelfInitComplete : called after doSelfInit(), barrier open
+//   - IsSelfInitComplete   : called in the "if inited" branch, auto-unseal already done
```
wslabosz-reply: These feel like specific function go docs

vault/core_self_init.go
```diff
@@ -0,0 +1,99 @@
...
+//
+// Pattern matches: vault/core.go (c.barrier.Get/Put with logical.StorageEntry)
```
wslabosz-reply: I don't know what's the use of this comment.

vault/core_self_init.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/openbao/openbao/sdk/v2/logical"
+)
+
+const (
+	// coreStatusSelfInitKey is the barrier key for the auto-init state machine.
```
wslabosz-reply: I feel like entry key makes more sense, "barrier key" in my mind implicates something like an encryption/root key 
```suggestion
	// coreStatusSelfInitKey is the entry key for the auto-init state machine.
```

vault/core_self_init.go
```diff
@@ -0,0 +1,99 @@
...
+	coreStatusSelfInitStarted = "started"
+
+	// coreStatusSelfInitCompleted is written after auto-init succeeds.
+	coreStatusSelfInitCompleted = "completed"
+)
+
+// MarkSelfInitStarted writes the "started" marker to the security barrier.
+// This must be called after core.Initialize() so the barrier exists and is open.
+// Its purpose is crash detection: if the process dies before MarkSelfInitComplete
+// is called, the next startup will find "started" and refuse to proceed.
```
wslabosz-reply:
```suggestion
// MarkSelfInitStarted is used to detect a self-init failure by writing the `coreStatusSelfInitStarted` marker to the storage. 
// Must be called after core.Initialize() as the barrier has to be unsealed.
```

vault/core_self_init.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	return c.barrier.Put(ctx, &logical.StorageEntry{
+		Key:   coreStatusSelfInitKey,
+		Value: []byte(coreStatusSelfInitStarted),
+	})
+}
+
+// MarkSelfInitComplete overwrites the "started" marker with "completed".
+// This must be called after auto-initialization finishes successfully.
+// The barrier must be open at this point.
```
wslabosz-reply:
```suggestion
// MarkSelfInitComplete overwrites the `coreStatusSelfInitStarted` marker with `coreStatusSelfInitCompleted`.
// Must be called after auto-initialization finishes successfully.
```

vault/core_self_init.go
```diff
@@ -0,0 +1,99 @@
...
+// and returns whether initialization can be considered complete.
+//
+// Precondition: the barrier must be unsealed (guaranteed by the "if inited"
+// branch in command.Initialize — auto-unseal has already run at that point).
+//
+// State transitions:
+//   - missing marker → true, nil  (backward compat: old cluster or manual init)
+//   - "completed"    → true, nil  (auto-init succeeded)
+//   - "started"      → false, err (crash detected mid-init: corrupt state)
+//   - anything else  → false, err (unknown value: corrupt state)
```
wslabosz-reply:
```suggestion
// IsSelfInitComplete reads the auto-init state machine from the storage
// used for self-init completion reporting.
//
// Precondition: the barrier must be unsealed (guaranteed by the "if inited"
// branch in command.Initialize — auto-unseal has already run at that point).
//
// State transitions:
//   - missing marker → true, nil  (backward compat: old cluster or manual init)
//   - "completed"    → true, nil  (auto-init succeeded)
//   - "started"      → false, err (crash detected mid-init: corrupt state)
//   - anything else  → false, err (unknown value: corrupt state)
```

vault/core_self_init_test.go
```diff
@@ -0,0 +1,219 @@
...
+//     c.barrier (not c.physical). A sealed or uninitialized barrier returns
+//     ErrBarrierSealed on any Get/Put.
+//   - Tests live in package vault (white-box) because the state machine
+//     operates on c.barrier, which is unexported.
+//   - Each test exercises exactly one observable behaviour.
+//   - Accuracy level 5+: every assertion has a comment explaining the invariant.
+//
+//
+// Pattern matches: vault/core_self_init.go (OpenBao MPLv2)
+// Test helpers from: vault/testing.go (OpenBao MPLv2)
```
wslabosz-reply: What's the motivation behind adding these comments?

vault/core_self_init_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+package vault
+
+// Tests for the self-init state machine in core_self_init.go.
+//
+// Design rationale:
+//   - All tests use TestCoreNewSeal(t) + TestCoreInit + unseal to get a Core
+//     with an open security barrier, since the state machine now writes to
+//     c.barrier (not c.physical). A sealed or uninitialized barrier returns
+//     ErrBarrierSealed on any Get/Put.
```
wslabosz-reply: this comment indicates a continuously changed state of the feature, but this PR introduced the mentioned "state machine" so it's not helpful to include details like `since the state machine now writes to c.barrier (not c.physical)`

vault/core_self_init_test.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/stretchr/testify/require"
+)
+
+// ---------------------------------------------------------------------------
+// MarkSelfInitStarted
+// ---------------------------------------------------------------------------
+
+// TestMarkSelfInitStarted_WritesStartedMarker verifies that after calling
+// MarkSelfInitStarted the barrier contains coreStatusSelfInitKey = "started".
```
wslabosz-reply: don't need to add the quotes, but imho it's redundant to write both the variable name and it's value
```suggestion
// TestMarkSelfInitStarted_WritesStartedMarker verifies that after calling
// MarkSelfInitStarted the barrier contains `coreStatusSelfInitKey`.
```

vault/core_self_init_test.go
```diff
@@ -0,0 +1,219 @@
...
+//
+// This is the first transition in the state machine (∅ → started).
+func TestMarkSelfInitStarted_WritesStartedMarker(t *testing.T) {
+	t.Parallel()
+
+	c, _, _ := TestCoreUnsealed(t)
+	ctx := context.Background()
+
+	err := c.MarkSelfInitStarted(ctx)
+	require.NoError(t, err, "MarkSelfInitStarted must not return an error on an open barrier")
```
wslabosz-reply:
```suggestion
	require.NoError(t, err, "MarkSelfInitStarted must not return an error when unsealed")
```

vault/core_self_init_test.go
```diff
@@ -0,0 +1,219 @@
...
+	entry, err := c.barrier.Get(ctx, coreStatusSelfInitKey)
+	require.NoError(t, err, "barrier.Get after MarkSelfInitStarted must not fail")
+	require.NotNil(t, entry, "barrier entry must exist after MarkSelfInitStarted")
+	require.Equal(t, coreStatusSelfInitStarted, string(entry.Value),
+		"stored value must be the const coreStatusSelfInitStarted, not a magic string")
+}
+
+// TestMarkSelfInitStarted_Idempotent verifies that calling MarkSelfInitStarted
+// twice does not fail and leaves the marker at "started".
+func TestMarkSelfInitStarted_Idempotent(t *testing.T) {
```
wslabosz-reply: looks like something that could be verified in the previous test, but I'm not against having it all together

wslabosz-reply (commented):
This PR introduces a correct approach to handling self-init silently failing on a second run, also fixing the panic bug, it is also well tested.

I've left a lot of comments, I think all of them are related to the comment usage, it shouldn't be treated as anything blocking, but I'm happy to discuss any of them.

vault/core_self_init_test.go
EnricoFusi: I rewrite completely the tests, so this is not present anymore.

vault/core.go
EnricoFusi: Done, with the new push.

vault/core_self_init_test.go
EnricoFusi: New push with code rewriting, it is no more there.

command/server_initialize_test.go
```diff
@@ -0,0 +1,313 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+//go:build !race && !hsm
```
satoqz: This isn't necessary, and I don't believe it is needed in `server_test.go` either.

vault/core_self_init_test.go
EnricoFusi: I am going to rewrite the test together with a new implementation which uses barrier to write. So this will disappear.

command/server.go
EnricoFusi: Actually the current comment in 1758 makes more sense, so I will just remove from there. I moved the check from the beginning of initialize, but not the comment.

vault/core_self_init_test.go
EnricoFusi: If you don't mind, I prefer having each test testing just one thing. It's kind of a mindset I like to keep, wo I will leave two tests.

vault/core_self_init_test.go
EnricoFusi: Summarized in the new push.  Just a note I put for myself.

vault/core_self_init_test.go
EnricoFusi: Well, at the beginning I was putting this notes to keep sure I took the code from opensource project, so I check everytime and noted. I should have polish the comments, for sure. I just removed them.

vault/core_self_init.go
EnricoFusi: Usually I check the code I put here comes from some openbao opensource code, so I am not sure to collide with proprietary code. But ist was just for my convenience, so I am polishing away.

command/server_initialize_test.go
EnricoFusi: Yes, that was exactly the point, those are things I wrote for myselfa. Polishing away for the next push.

command/server_initialize_test.go
EnricoFusi: Was useful while writing the file, polishing away for the next push.

command/server_initialize_test.go
EnricoFusi: At the beginning I used physical, then after review with colleagues i went to logical. This comment survived there, will polish for the next push.

vault/core_self_init.go
EnricoFusi: Also, notes I put there for myself. Polished away in next push.

command/server_initialize_test.go
EnricoFusi: I polished it away for the next push. Was useful during deve & testing.

wslabosz-reply (approved):
LGTM

command/server_initialize_test.go
EnricoFusi: Thank you. This is part of EAL5, something which comes from my past. I like it's appreciated.

satoqz:
@EnricoFusi I would guess that you can just remove the build tags from `server_test.go` as well, then you don't need to have them anywhere and the function will be available.

EnricoFusi:
> @EnricoFusi I would guess that you can just remove the build tags from `server_test.go` as well, then you don't need to have them anywhere and the function will be available.

the tags were  in server_test.go long time before I wrote another test file recalling it.... I don't know why they were in the code, honestly, so I just follow. I am a bit shy in changing old code... I can polish this too, anyhow.

EnricoFusi:
After a quick chat with @satoqz  , we decided to clean offending tags from  old code.

command/server.go
```diff
@@ -1731,50 +1732,60 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
...
-		// way of sidestepping authentication. At first startup there is no
-		// other authentication information but on subsequent startups
-		// presumably the admin has created an alternative mechanism we should
-		// defer to.
+		// We refuse to rerun self-initialization.
+		// HOWEVER, we must verify that the previous initialization actually finished.
+		// If barrier exists but self-init marker is missing, we are in a broken state.
+		//
+		// IsSelfInitComplete would fail at reading as barrier is sealed, skip the consistency check; we are a
+		// standby coming up after another node ran initialization.
```
cipherboy: Not sure I understand this comment @EnricoFusi @satoqz -- was a word dropped?

command/server.go
```diff
@@ -1731,50 +1732,60 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
...
 		BarrierConfig:  barrierConfig,
 		RecoveryConfig: recoveryConfig,
 	})
 	if err != nil {
-		core.Logger().Error("failed to initialize", "error", err)
 		if errors.Is(err, vault.ErrParallelInit) || errors.Is(err, vault.ErrAlreadyInit) {
 			return nil
 		}
-
+		core.Logger().Error("failed to initialize: unexpected error occurred")
```
cipherboy:
```suggestion
		core.Logger().Error("failed to initialize: unexpected error occurred", "error", err)
```

otherwise we may risk hiding this error, depending on what goes to the logger vs `ui.Error()`.

command/server.go
cipherboy: Just curious as well, when would `core.Initialized(ctx)` and `core.SealAccess().RecoveryKeySupported()` return true, but `core.Sealed()` return false?

vault/core_self_init.go
```diff
@@ -0,0 +1,90 @@
...
+	return c.barrier.Put(ctx, &logical.StorageEntry{
+		Key:   coreStatusSelfInitKey,
+		Value: []byte(coreStatusSelfInitStarted),
+	})
+}
+
+// MarkSelfInitComplete overwrites the `coreStatusSelfInitStarted` marker with `coreStatusSelfInitCompleted`.
+// Must be called after auto-initialization finishes successfully.
+func (c *Core) MarkSelfInitComplete(ctx context.Context) error {
+	if c.barrier == nil {
```
cipherboy: I go back to, is `c.barrier` ever actually nil? 🤔 Maybe I'm missing the code path.

cipherboy (commented):
I trust others have reviewed the tests on this one, but the actual implementation looks good sans nits.

vault/core_self_init.go
```diff
@@ -0,0 +1,90 @@
...
+// used for self-init completion reporting.
+//
+// Precondition: the barrier must be unsealed (guaranteed by the "if inited"
+// branch in command.Initialize — auto-unseal has already run at that point).
+//
+// State transitions:
+//   - missing marker → true, nil  (backward compat: old cluster or manual init)
+//   - "completed"    → true, nil  (auto-init succeeded)
+//   - "started"      → false, err (crash detected mid-init: corrupt state)
+//   - anything else  → false, err (unknown value: corrupt state)
```
satoqz: Can't we just remove the entry once complete if that is semantically equal anyway? We could write a "failed" state before self-init starts, then remove it once self-init completes successfully.

vault/core_self_init.go
```diff
@@ -0,0 +1,90 @@
...
+	if c.barrier == nil {
+		return fmt.Errorf("security barrier not available")
+	}
+	return c.barrier.Put(ctx, &logical.StorageEntry{
+		Key:   coreStatusSelfInitKey,
+		Value: []byte(coreStatusSelfInitCompleted),
+	})
+}
+
+// IsSelfInitComplete reads the auto-init state machine from the storage
```
satoqz: Why do you refer to "self-init" as "auto-init" several times?

vault/core_self_init.go
satoqz, line 1: Can we name this to `self_init.go`? Prefixing files with `core_*` is nonsensical to me, many files in `/vault` add methods to core.

vault/core_self_init_test.go
satoqz: I don't think this is resolved? I agree this file is superfluous, `server_initialize_test.go` is rather sufficient (if anything I'd like to see additional end-to-end tests over there or in `vault/external_tests`, esp w.r.t. https://github.com/openbao/openbao/pull/2442#discussion_r2983239670 which current tests don't capture at all).

To give some context @EnricoFusi, totally appreciate you giving much care to tests, OTOH this project has taught me that tests are a gigantic source of technical debt so I'd like to not overdo them and only add what realistically matters.

vault/core_self_init.go
satoqz: I don't think it is ever nil either.

command/server.go
satoqz: Auto Unseal attempts are started asynchronously and we aren't guaranteed to complete the unseal by this point in `ServerCommand.Run()` as far as I can tell. So I'm not sure this works at all? We'll race and skip the check.

EnricoFusi: @satoqz  I got your point. runUnseal is launched as a goroutine before Initialize is called, so core.Sealed() may still return true at that point. The consistency check would always be skipped. I'll fix in the next push.

cipherboy: @satoqz Ah, I see. In that case, we should probably wait for unseal to complete if we can, rather than racing.

vault/core_self_init_test.go
EnricoFusi: I've addressed all the other feedback. For the end-to-end test, could you point me to a specific scenario that is not covered by TestInitialize_AlreadyInitialized_CrashDetected in server_initialize_test.go? I want to make sure I implement what you actually mean. I've written the tests for those 3 function that I added, it seemed to me due. Anyhow, I removed them.

command/server.go
EnricoFusi: Fixed.

command/server.go
EnricoFusi: Fixed, thanks.

vault/core_self_init.go
EnricoFusi: Fixed.

vault/core_self_init.go
EnricoFusi: Done as requested.

vault/core_self_init.go
EnricoFusi: Mi mistake, ESL.

vault/core_self_init.go
EnricoFusi: Done.

Nerkho:
I tried to build from this branch as I'm having issues with self-init that could have been related.

For now, the build fails at some point during the process. The node just restarted itself but without fully completing the init process (I have an auth-method configuration in my self-init stanza that is not provisioned).

I have attached one of the pod's logs here for review.

[openbao-0.txt](https://github.com/user-attachments/files/26405955/openbao-0.txt)

cipherboy:
@EnricoFusi @satoqz hmm:

> openbao-0 openbao failed to mark self-init started: cannot write to readonly storage

(from @Nerkho's logs)

Suggests we may need to block writing self-init started state until Raft is ready. :thinking:

command/server.go
```diff
@@ -1731,50 +1733,59 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
...
 		}
-
+		core.Logger().Error("failed to initialize: unexpected error occurred", "error", err)
 		return fmt.Errorf("self-initialization failed: %w", err)
 	}
-
+	// Write "started" marker to storage (self-init started).
+	if err := core.MarkSelfInitStarted(ctx); err != nil {
+		return fmt.Errorf("failed to mark self-init started: %w", err)
+	}
```
cipherboy: @satoqz @Nerkho Yes, this addition needs to be moved down until new line 1799 (after the `!isLeader` check).

EnricoFusi: Done, and pushed.

Nerkho:
Thanks! Yes, I tested it again, and now the init goes fully through.

It doesn't address https://github.com/openbao/openbao/issues/2274, but that was a misunderstanding on my side that this was related.

cipherboy (approved):
Thanks @EnricoFusi!

command/server.go
```diff
@@ -1731,47 +1733,53 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
 // OpenBao core. This will exit early if there is no configuration for this
 // or if the core is already initialized.
 func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) error {
-	// Skip initialize for dev server as it is handled in initDevCore
-	if len(config.Initialization) == 0 || c.flagDev {
+	if len(config.Initialization) == 0 {
 		return nil
 	}
```
satoqz, line 1738: With this being here, we don't check for a self-init failure if the config was removed on the second run. Seems rare, but a valid edge case IMO.

command/server.go
```diff
@@ -1731,47 +1733,53 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
...
+		// If barrier exists but self-init marker is missing, we are in a broken state.
+		complete, err := core.IsSelfInitComplete(ctx)
+		if err != nil {
+			if errors.Is(err, barrier.ErrBarrierSealed) {
+				// Barrier not yet unsealed, we can skip the check.
+				// There are two cases in which the barrier is unsealed at this point:
+				// follower node in HA cluster, or primary node with unseal still in progress.
+				// In both cases skipping is correct: the node did not run self-init and the unseal flow will handle it.
+				return nil
+			}
```
satoqz, line 1761: Sorry, no, this is nonsense and breaks the entire thing.

> primary node with unseal still in progress, in both cases skipping is correct

How is this possibly correct? With this test configuration:

```hcl

listener "tcp" {
  address = "127.0.0.1:8200"
  tls_disable = true
}

storage "raft" {
  path = "hack/raft"
}

seal "static" {
   current_key = "542a91c2aff86797e8dfa16adf86dcbea5706410791b319798936a1b08e4ef8b"
   current_key_id = "20250630"
}

initialize "nonsense" {
  request "nonsense" {
    operation = "update"
    path = "foo/bar"
    data = {}
  }
}
```

I get the following on initial startup, as expected:

```
Beginning post-unseal configuration
initialize.[nonsense (0)]: request.[nonsense (0)]: failed to evaluate request: 1 error occurred:
        * unsupported path
```

and the "failed" marker is set in storage.

But on the next runs, OpenBao consistently starts up and unseals fine (when it should error, given the marker). This hits the branch above and skips the check because `barrier.ErrBarrierSeal` is hit, presumably because Auto Unseal just hasn't run yet.

IMO this check should be moved out of `ServerCommand.Initialize` and into `unsealInternal` right after `barrier.Unseal` so it is sync with Auto Seals. We don't need to do that for the initial write or the marker since `core.Initialize` has already called `UnsealWithStoredKeys` at that point.

satoqz (changes requested):

command/server.go
EnricoFusi: Hello , I will check and change the code as requested.

satoqz: @EnricoFusi no worries, I've linked a refactor with the changes I'd expect so we can get this shipped in 2.5.3.

satoqz: See #2908

klaus-sap: @satoqz so could you mark the requested change as resolved?

satoqz: No, it's not resolved in this PR, but I've proposed an alternative as linked.

EnricoFusi: Fixed. The consistency check has been moved into runUnseal right after UnsealWithStoredKeys succeeds, as suggested.

EnricoFusi:
Closing in favor of #2908, which implements the same fix with a cleaner approach. Thanks to everyone who reviewed and provided feedback

Issue #2302, opened 2026-01-15
Support auto-unseal for Open Telekom Cloud KMS


mrclki:
Hi OpenBao team,

I am currently deploying OpenBao on Open Telekom Cloud. I would like to utilize auto-unseal, but this provider uses a proprietary KMS API that is not compatible with the existing ones. Currently, there is no native integration to support this workflow.

I have already submitted a [Pull Request](https://github.com/openbao/go-kms-wrapping/pull/63) to the `go-kms-wrapping` repository adding the specific API logic. I also have the necessary changes for OpenBao ready locally, including adding the seal configuration stanza and wiring up the new wrapper.

I wanted to open this issue to raise awareness of the feature and track the dependency. Since the core integration depends on the upstream library changes in [PR](https://github.com/openbao/go-kms-wrapping/pull/63).

Thanks for your time and for maintaining this project!

mrclki:
This also addresses addresses outstanding [requests](https://github.com/openbao/openbao/issues/1054) from other users.

satoqz:
Hi @mrclki, thanks for the issue + PR, much appreciated!

We've been getting many requests (and ready-made PRs) to add support for auto-unsealing via new KMSes - While we totally want to support all of those, @cipherboy's and my stance is that we'll want to build a plugin infrastructure for auto-unseal mechanisms before shipping any additional providers to avoid dependency creep going forward. This would make auto-unseal wrappers distributable as separate binaries, much like current secret engine plugins &c work.

This would likely be part of a v2.6 release, hence a bit of a delay (we're currently working to release v2.5 the following week or so).

I hope this makes sense to you, I felt the need to provide some transparency around the project's plans before people get the feeling their work was being ignored or unappreciated :)

baurmatt:
@satoqz Thanks for the headups! We're also interested in contributing an auto-unsealing method ([OpenStack](https://github.com/openbao/openbao/issues/1114)). Is there already a ticket/design document/.. discussing the auto-unsealing plugin system?

satoqz:
@baurmatt nothing yet, but this will not look fundamentally different from the existing system (see `Wrapper` interface in https://github.com/openbao/go-kms-wrapping) as https://github.com/hashicorp/go-plugin lets us extend that across an RPC boundary rather cleanly.

mrclki:
@satoqz Thanks for the headups! 

Since the underlying change isn’t designed yet, would you be open to contributions on that effort? I’m happy to help drive this forward.

[...]

PR #2459, opened 2026-02-11
Add RFC on Auto Unseal plugins

satoqz:
## Description

Add an RFC as discussed in #2302.

[Rendered](https://github.com/Ki-Reply-GmbH/openbao/blob/rfc-auto-unseal-plugins/website/content/docs/rfcs/auto-unseal-plugins.mdx)

cc @cipherboy @mrclki

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/rfcs/auto-unseal-plugins.mdx
+++ b/website/content/docs/rfcs/auto-unseal-plugins.mdx
@@ -278,6 +278,8 @@ Additional PRs TBD
 > over dynamic library loading and since we've had it built and tested for
 > years, we'll continue to use it."
 
+...
+
 > "_Shared libraries_ have one major advantage over _our (plugin) system_
 > which is much higher performance. In real world scenarios across our various
 > tools, we've never required any more performance out of our _plugin system_


website/content/docs/rfcs/auto-unseal-plugins.mdx
```diff
@@ -0,0 +1,240 @@
...
+OpenBao ultimately seeks to support all of the proposed Auto Unseal providers.
+However, simply including all of these into the main binary seems naive:
+
+1. Each provider usually comes with its own set of dependencies, e.g., an SDK
+   package specific to the provider.
+2. An OpenBao instance usually only uses one (up to two with [Seal
+   Migration](../concepts/seal.mdx)) seal mechanisms at a time. This number
+   may increase with [Emergency Seals](./emergency-seal.mdx) or [Namespace
+   Sealing](./namespace-sealing.mdx), but it is unlikely for an instance to make
+   use of a large variety of mechanisms at a time.
```
cipherboy, line 69: I'd add 3 and reference the original plugin decision and say that our goal was for the main binary to have only OSI-licensed testable integrations with everything else being an external plugin. Many of these provider-specific ones rely on proprietary servers and thus the project and it's broader community can't reliably access them.

website/content/docs/rfcs/auto-unseal-plugins.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+OpenBao Auto Seal Plugins will be available as standalone
+binaries (and OCI images) just like current secrets/auth/database
+plugins are. The source code of these plugins will live in
+[go-kms-wrapping](https://github.com/openbao/go-kms-wrapping), which is also
+where they will be released separately from the main release schedule. Custom,
+3rd party plugins not upstreamed into go-kms-wrapping can also be built and used
+independently, though upstreaming will be encouraged.
+
+In contrast to existing plugin types, KMS plugins can only be registered
```
cipherboy, line 92: Are we using KMS and auto-unseal plugin types interchangeably here? Or is this a difference between two new types hinted at by this RFC? (Auto-unseal plugins now and KMS plugins later?)

satoqz: Ah, see 2-3 paragraphs later, I propose to have them be the same and eventually service External Keys as well. But I should probably revise the initial paragraph to clarify from the start.

website/content/docs/rfcs/auto-unseal-plugins.mdx
```diff
@@ -0,0 +1,240 @@
...
+  version   = "v0.0.1"
+  sha256sum = "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
+}
+
+seal "opentelekomcloudkms" {
+  // as usual...
+}
+```
+
+The value of `plugin_auto_register` is ignored and KMS plugins are never
```
cipherboy: What happens with `version`? Traditional plugins can have multiple versions registered and running. I assume we'd only allow one unseal plugin version and err on multiple?

satoqz: TBH I've only left it in there as I copied and adapted the example from the declarative plugins RFC, but I wonder if it could be useful to support once many seals are running concurrently (via Per-Namespace Sealing &c)?

cipherboy: Perhaps, I suppose. I think we'd have to assume plugins are compatible across versions though. 

I'd say we start simple and err tbh

satoqz: Yeah, totally fine with that.

website/content/docs/rfcs/auto-unseal-plugins.mdx
pree: Maybe we could name them `External KMS plugins` ?

website/content/docs/rfcs/auto-unseal-plugins.mdx
```diff
@@ -0,0 +1,240 @@
...
+Since Auto Seals naturally must initialize _before_ unsealing, following the
+legacy API/storage-driven plugin lifecycle is not feasible; at the same time,
+declarative plugin management is already generally preferred and recommended
+over API-driven plugin management for all plugin types following their recent
+introduction to OpenBao.
+
+An exemplary server config entry may look like the following:
+
+```hcl
+plugin "kms" "opentelekomcloudkms" {
```
pree, line 104: Would the plugins be able to read from the ENV's?
A lot of auto unseal mechanisms support setting authentication credentials set via ENV's (which is also needed when doing WIF for example).

As other plugins require you to specify the ENV variables (incl. values) inside the registration, I would love to see something like `allowed_envs` or `passthrough_envs` which only includes a list of the ENV variable keys that are passed through to the plugin.

satoqz: Good point, I had also wondered this but didn't get to it yet because it's a very deep codepath to traverse.

Looking at it now (without checking actual runtime behavior in OpenBao) it would seem that all plugin processes inherit the host environment by default?

https://github.com/hashicorp/go-plugin/blob/abeb978c3d6f630da9beaefa43fea3b7d2f5caef/client.go#L270-L272

We could add options (before handing over to go-plugin) to increase control over this behavior, though I am happy with passthrough being the default behavior as it mirrors current seal usage.

pree: passthrough of all env variables as the default would be fine for  me for now too. I remember that I had some issues with the ENV variables previously on upstream where not all ENV variables attached to the host process were inherited.

I think more fine-grained control of those values could be something picked up at a later stage as an addition.

website/content/docs/rfcs/auto-unseal-plugins.mdx
```diff
@@ -0,0 +1,240 @@
...
+- [KMIP](../configuration/seal/kmip.mdx)
+- [PKCS#11](../configuration/seal/pkcs11.mdx)
+- [Static Key](../configuration/seal/static.mdx)
+
+All of the currently available mechanisms are statically compiled into the
+released OpenBao binary, with the exception of PKCS#11, which is only available
+in the HSM distribution of OpenBao due to requiring CGO.
+
+Furthermore, requests for (usually along with a Pull Request ready for review)
+have been received to add additional vendor-specific mechanisms:
```
JanMa:
```suggestion
Furthermore, requests to add additional vendor-specific mechanisms have been received (usually along with a Pull Request ready for review):
```

website/content/docs/rfcs/auto-unseal-plugins.mdx
```diff
@@ -0,0 +1,240 @@
...
+increased complexity in release management. This RFC proposes to solve the
+presented problem by pluginizing Auto Unseal mechanisms using the same plugin
+mechanism that is already used for secrets, auth & database engines.
+
+### User-facing description
+
+OpenBao Auto Seal Plugins will be available as standalone
+binaries (and OCI images) just like current secrets/auth/database
+plugins are. The source code of these plugins will live in
+[go-kms-wrapping](https://github.com/openbao/go-kms-wrapping), which is also
```
JanMa: I would prefer if the source code can live in our [main plugin repo](https://github.com/openbao/openbao-plugins) to reuse the existing distribution mechanism we've built. If necessary we can add additional wrappers in g-k-w.

website/content/docs/rfcs/auto-unseal-plugins.mdx
```diff
@@ -0,0 +1,240 @@
...
+  complexity.
+- Requiring users to build soft-forked versions of OpenBao downstream to
+  statically include the plugins they require.
+
+Neither of these are in line with OpenBao's existing plugin-based philosophy.
+
+### Downsides
+
+Any kind of plugin requires operators to handle additional complexity. Based
+on unformal community observation, it appears that many instances are operated
```
JanMa:
```suggestion
on informal community observation, it appears that many instances are operated
```

website/content/docs/rfcs/auto-unseal-plugins.mdx
satoqz: Hmm, can you elaborate @JanMa? The majority of a plugin (for now) will be the wrapper code, plus the typical tiny main function that each plugin needs. Would you want to move all wrappers to openbao-plugins? Define just the main functions in openbao-plugins? Perhaps we could (even without placing any of the Go code in openbao-plugins) drive release builds from openbao-plugins and just reference remote Go packages?

JanMa: I'd definitely like to drive the release builds from openbao-plugins. I'm not sure yet if I prefer toadd the wrapper there as well, just the main function or just reference a remote package. All have their pro's and con's. The question we need to ask ourselves is if we want all those custom integrations in g-k-w or focus on the main encryption parts in there :thinking:

satoqz: As things stand most of the wrappers already present in go-kms-wrapping would eventually become external plugins 🤷 So to avoid an awkward split between old and new external integrations, we'd need to move them to openbao-plugins.

I'd vote for keeping all implementations (+ main function packages) in go-kms-wrapping together with the SDK, but driving releases from openbao-plugins.

satoqz: @JanMa fine, after reading your beautiful Makefile and workflows in openbao-plugins I now think the easiest path towards shipping this will be to add the main functions to openbao-plugins and treat go-kms-wrapping as a pure library.

After all that'll at least be equal in complexity to what is required to add a builtin wrapper to the main repo (or to update its version), you'd need to PR in both go-kms-wrapping and the main repo, we're shifting the latter to openbao-plugins.

JanMa: I'm flattered :blush: :smile:

website/content/docs/rfcs/auto-unseal-plugins.mdx
```diff
@@ -0,0 +1,245 @@
...
+publish guides around plugin operation in typical standard environments (e.g.,
+Kubernetes) on the OpenBao website. These could be similar to existing guides
+around HSM integration with PKCS#11.
+
+### Security Implications
+
+This change improves OpenBao's overall security posture by reducing the attack
+surface of the average OpenBao installation and by enabling plug-and-play custom
+Auto Unseal implementations if desired.
+
```
DanGhita, line 226: @satoqz , @cipherboy  The likelihood is very low (plugins are "signed")  and the vector attack probably not realistic, but if an attacker manages to compromise this kind of auto-unseal plugins, he/she could replace the KMS/HSM keys by a hard-coded (constant key known by the pirate) . That would compromise the whole Vault.

satoqz: @DanGhita I think that would be very similar to compromising the integrity of the main OpenBao binary right now. Do we have a reason to assume that the plugin-based attack vector is easier to exploit?

DanGhita: I would say the level of confidence we have in OpenBao is higher than the trust in the third party binaries. Without even talking about binaries, tempering a third party component is probably less visible than tempering the OpenBao code itself.
But I agree, there's not much we can do, maybe just a specific warning in the documentation saying that the auto-unseal makes use of very sensitive plugins and the user must be extremely careful  with the auto-unseal plugins that he/she authorizes.

satoqz: Yes, fair. Note that the goal is not to push auto unseal plugins to be maintained by 3rd parties (though we enable this if they deem it necessary) - we'd still attempt to maintain most plugins officially, as 1st-party OpenBao plugins. Either way, I'll make sure the risks are highlighted appropriately in the eventual documentation.

pree: But it would still make it possible for 3rd parties to create and sign their own plugins and load them into OpenBao, no?

satoqz: @pree yes, the point is that preferably you wouldn't _need to_, but you _could_.

cipherboy:
@DanGhita Are we OK to merge this one, or do you want more time to review? :-)

satoqz:
Alright, merging this then :)

For those curious, a lot of progress has already been made on the implementation:

- https://github.com/openbao/go-kms-wrapping/pull/71
- https://github.com/openbao/openbao-plugins/pull/55
- [`auto-unseal-plugins`](https://github.com/Ki-Reply-GmbH/openbao/tree/auto-unseal-plugins), to be opened as a PR once go-kms-wrapping has support merged in

Issue #2447, opened 2026-02-09
Initalize blocks in dev-mode server causes error


dbanetto:
**Describe the bug**

Running `bao server -config ./dev.hcl -dev -dev-root-token-id root` with the following `dev.hcl` exits on start up.

```hcl
initialize "pki-setup" {
  request "test-root" {
    operation = "update"
    path = "sys/mounts/pki/root/test"
    data = {
      type = "pki",
    }
  }
}
```

**To Reproduce**
1. Create `dev.hcl` with the content above
2. Start a local dev server with `bao server -config ./dev.hcl -dev -dev-root-token-id root` 
3. See `self-initialization requires auto-unseal as there is no way to persist the Shamir's keys` error

<details>

<summary>Sample log output from running the above steps</summary>

```
❯ bao server -config ./local.hcl -dev -dev-root-token-id root 
==> OpenBao server configuration:

Administrative Namespace: 
             Api Address: http://127.0.0.1:8200
                     Cgo: enabled
         Cluster Address: https://127.0.0.1:8201
   Environment Variables: COLORTERM, COMMAND_MODE, EDITOR, GHOSTTY_BIN_DIR, GHOSTTY_RESOURCES_DIR, GH
OSTTY_SHELL_FEATURES, GREP_COLOR, GREP_COLORS, HOME, HOMEBREW_CELLAR, HOMEBREW_PREFIX, HOMEBREW_REPOS
ITORY, INFOPATH, LANG, LESS_TERMCAP_mb, LESS_TERMCAP_md, LESS_TERMCAP_me, LESS_TERMCAP_se, LESS_TERMC
AP_so, LESS_TERMCAP_ue, LESS_TERMCAP_us, LOGNAME, LSCOLORS, MANPATH, MASON, MYVIMRC, NVIM, NVIM_LOG_F
ILE, NVIM_UNCEPTION_PIPE_PATH_HOST, OLDPWD, OSLogRateLimit, PATH, PWD, SHELL, SHLVL, SSH_AUTH_SOCK, S
TARSHIP_SESSION_KEY, STARSHIP_SHELL, TERM, TERMINFO, TERM_PROGRAM, TERM_PROGRAM_VERSION, TMPDIR, USER
, XDG_DATA_DIRS, XPC_FLAGS, XPC_SERVICE_NAME, _, __CFBundleIdentifier, __CF_USER_TEXT_ENCODING
              Go Version: go1.25.6
[...]

PR #2463, opened 2026-02-11
Allow initalize blocks in dev mode

dbanetto:
## Description

Allows `initalize` blocks to be used `bao server -dev`.

## Rationale

Resolves: #2447

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2463.txt
+++ b/changelog/2463.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+command/server: Allow self-initialization stanzas in development server mode.
+```
--- a/changelog/2571.txt
+++ b/changelog/2571.txt
@@ -1,3 +1,0 @@
-```release-note:improvement
-command: Allow initalization blocks in config for devlopment server.
-```


command/server.go
```diff
@@ -1732,7 +1732,7 @@ func (c *ServerCommand) waitForLeader(core *vault.Core) (bool, error) {
 // OpenBao core. This will exit early if there is no configuration for this
 // or if the core is already initialized.
 func (c *ServerCommand) Initialize(core *vault.Core, config *server.Config) error {
-	if len(config.Initialization) == 0 {
+	if len(config.Initialization) == 0 || !c.flagDevSkipInit {
```
dbanetto: Used `flagDevSkipInit` instead of `flagDev` as it will handle the case when the user wants to go through the regular initialisation flow.

dbanetto: 🤔  Now having another think about it.

Maybe it should be

```suggestion
	if len(config.Initialization) == 0 || !c.flagDev {
```

satoqz: I would expect `-dev` to trigger self-init if configured, and `-dev-skip-init` to skip both of self-init and dev mode init.

dbanetto: Thank you, I've updated it to be `flagDev`

satoqz: This skips self-initialization any time we're not in dev mode. Are you sure this is the right code to change?

dbanetto: Ah, I made a mistake with the keeping the `!` from when I had `!c.flagDevSkipInit` 🤦 

It should be `len(config.Initialization) == 0 || c.flagDev` so that self-initialization is skipped in the case of dev-mode as it is handled in the `initDevCore()` function instead of here.

[Luckily a tests did catch my mistake](https://github.com/openbao/openbao/actions/runs/22597247277/job/65482940161) though I can't quite figure out how to run them locally.

satoqz: Ah that seems reasonable :+1: I'll give this a manual test run tomorrow.

satoqz:
@dbanetto in the meantime, do you mind adding a small changelog entry? You'll find lots of examples in the `changelog/` directory. `release-note:improvement` is probably the best category for this one.

satoqz:
Gave this a test run and it works nicely :)

dbanetto:
Hi @satoqz, I've added an entry into `changelog/`

JanMa (approved):

satoqz:
Thanks @dbanetto!

Issue #2466, opened 2026-02-12
Cannot migrate from "file" to "raft" on v2.5.0


hylobates-agilis:
**Describe the bug**
Since v2.5.0, following [the official procedure](https://openbao.org/docs/commands/operator/migrate/), it is not possible to unseal after migration.

The procedure of migration itself is successful but you cannot unseal afterwards

**To Reproduce**
1. Start a fresh OpenBao v2.5.0
2. Start with a minimal "file" storage configuration
3. [Follow procedure](https://openbao.org/docs/commands/operator/migrate/) to migrate from "file" to "raft"
4. Change the configuration file
5. Restart
6. Unseal is impossible

**Expected behavior**
With v2.4.4, unseal was possible after migration. The data itself seems to be fine, but it’s impossible to unseal.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): v2.5.0
* OpenBao CLI Version (retrieve with `bao version`): v2.5.0
* Server Operating System/Architecture: Debian 13, Linux AMD64

OpenBao server configuration file(s):

```hcl
# Copyright (c) HashiCorp, Inc.
# SPDX-License-Identifier: MPL-2.0

# Full configuration options can be found at https://github.com/openbao/openbao/tree/main/website/content/docs/configuration

ui = true

storage "raft" {
  path = "/opt/openbao/data"
}

api_addr = "https://bao1.incus:8200"
cluster_addr = "https://bao1.incus:8201"

# HTTPS listener
listener "tcp" {
  address       = "0.0.0.0:8200"
  tls_cert_file = "/opt/openbao/tls/tls.crt"
  tls_key_file  = "/opt/openbao/tls/tls.key"
}
```

**Additional context**
Cannot reproduce within v2.4.4, so I guess there’s something wrong with v2.5.0

Corresponding debug log:
```
2026-02-12T10:17:48.469Z [INFO]  proxy environment: http_proxy="" https_proxy="" no_proxy=""
2026-02-12T10:17:48.470Z [DEBUG] storage.raft.fsm: time to open database: elapsed=1.073907ms path=/opt/openbao/raft/vault.db
[...]

PR #2467, opened 2026-02-12
Don't upgrade mount table in standby mode

satoqz:
## Description

Skip writing the results of a legacy mount table upgrade to disk if we're loading mounts as standby.

## Rationale

Standby nodes would attempt a legacy mount table upgrade e.g. when upgrading from an old, pre-transactional raft state or when migrating storage from a non-transactional backend. This would cause unseal failures even in single-node, freshly migrated clusters as nodes always unseal as standby first before pivoting to an active-node unseal.

Is there a cleaner way to achieve this? `core.Standby()` is false at all times while unsealing and is only set to true afterwards, and I didn't want to break that guarantee for this change.

Resolves: #2466

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/request_handling_test.go
+++ b/vault/request_handling_test.go
@@ -25,7 +25,7 @@ func TestRequestHandling_Wrapping(t *testing.T) {
 	core.logicalBackends["kv"] = PassthroughBackendFactory
 
 	meUUID, _ := uuid.GenerateUUID()
-	err := core.mount(namespace.RootContext(nil), &MountEntry{
+	err := core.mount(namespace.RootContext(t.Context()), &MountEntry{
 		Table: mountTableType,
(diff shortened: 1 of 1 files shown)


satoqz:
As a follow-up we could look at the behavior of invalidation, though that really only applies when live-upgrading from a pre-transactional raft cluster to a transactional + read-scalability-enabled one. So that is en _extreme_ edge case I'm not sure is worth solving if it happens to cause problems.

phil9909:
> Is there a cleaner way to achieve this? `core.Standby()` is false at all times while unsealing and is only set to true afterwards, and I didn't want to break that guarantee for this change.

We could just try persisting and ignore read only errors, not sure if this is really better.

PR #2468, opened 2026-02-12
Add Seal manager lifecycle

wslabosz-reply:
## Description
- Added `SealManager` setup on core creation
- Added `(*SealManager).sealAll` on `sealInternalWithOptions()`
- Removed `keepHALock` and `performCleanup` parameters from `sealInternalWithOptions` (always passed as `true`)
- Removed `keepHALockOnStepDown` from `Core` struct
- Finished implementation of `NamespaceSealed()` and `storageByPath()` for now without a possible sealed namespace creation
- Added seal namespace cleanup logic, cleaning up policies and mounts from memory

All logic changes are sourced from [namespaces-seal feature branch](https://github.com/openbao/openbao/tree/namespaces-seal)

## Rationale
Introduced changes complete the initialization of the seal manager, and the cleanup at the node step-down, seal or any other (not-fatal) failure.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/core.go
+++ b/vault/core.go
@@ -276,13 +276,12 @@ type Core struct {
 	stateLock locking.RWMutex
 	sealed    atomic.Bool
 
-	standby              atomic.Bool
-	standbyDoneCh        chan struct{}
-	standbyStopCh        atomic.Value
-	standbyRestartCh     atomic.Value
-	manualStepDownCh     chan struct{}
-	keepHALockOnStepDown atomic.Bool
-	heldHALock           physical.Lock
+	standby          atomic.Bool
+	standbyDoneCh    chan struct{}
+	standbyStopCh    atomic.Value
+	standbyRestartCh atomic.Value
+	manualStepDownCh chan struct{}
+	heldHALock       physical.Lock
 
 	// shutdownDoneCh is used to notify when core.Shutdown() completes.
 	// core.Shutdown() is typically issued in a goroutine to allow Vault to
@@ -1856,27 +1855,30 @@ func (c *Core) unsealInternal(ctx context.Context, rootKey []byte) error {
 
 	// Do post-unseal setup if HA is not enabled
 	if c.ha == nil {
-		// We still need to set up cluster info even if it's not part of a
-		// cluster right now. This also populates the cached cluster object.
-		if err := c.setupCluster(ctx); err != nil {
-			c.logger.Error("cluster setup failed", "error", err)
-			c.barrier.Seal()
-			c.logger.Warn("vault is sealed")
-			return err
+		runPostUnseal := func(context.Context) error {
+			// We still need to set up cluster info even if it's not part of a
+			// cluster right now. This also populates the cached cluster object.
+			if err := c.setupCluster(ctx); err != nil {
+				c.logger.Error("cluster setup failed", "error", err)
+				return err
+			}
+
+			if err := c.migrateSeal(ctx); err != nil {
+				c.logger.Error("seal migration error", "error", err)
+				return err
+			}
+
+			ctx, ctxCancel := context.WithCancel(namespace.RootContext(context.Background()))
+			if err := c.postUnseal(ctx, ctxCancel, standardUnsealStrategy{}); err != nil {
+				c.logger.Error("post-unseal setup failed", "error", err)
+				return err
+			}
+			return nil
 		}
 
-		if err := c.migrateSeal(ctx); err != nil {
-			c.logger.Error("seal migration error", "error", err)
-			c.barrier.Seal()
-			c.logger.Warn("vault is sealed")
-			return err
-		}
-
-		ctx, ctxCancel := context.WithCancel(namespace.RootContext(context.TODO()))
-		if err := c.postUnseal(ctx, ctxCancel, standardUnsealStrategy{}); err != nil {
-			c.logger.Error("post-unseal setup failed", "error", err)
-			c.barrier.Seal()
-			c.logger.Warn("vault is sealed")
+		if err := runPostUnseal(ctx); err != nil {
+			err = errors.Join(err, c.sealManager.sealAll())
+			c.logger.Warn("OpenBao is sealed")
 			return err
 		}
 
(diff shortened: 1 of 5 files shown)


vault/mount.go
```diff
@@ -1575,6 +1575,56 @@ func (c *Core) unloadMounts(ctx context.Context) error {
...
+	if c.mounts != nil {
+		mountTable := c.mounts.ShallowClone()
+		if err := c.cleanupNamespaceMounts(ctx, mountTable, ns.UUID); err != nil {
+			return err
+		}
+	}
+
+	if c.logger.IsInfo() {
+		c.logger.Info(fmt.Sprintf("successfully unmounted namespace %q mounts from mount table", ns.Path))
+	}
```
satoqz:
```suggestion
	c.logger.Info("unmounted all secrets engines", "namespace", ns.Path)
```

vault/auth.go
```diff
@@ -1176,6 +1176,25 @@ func (c *Core) teardownCredentials(ctx context.Context) error {
...
+	if c.auth != nil {
+		authTable := c.auth.ShallowClone()
+		if err := c.cleanupNamespaceMounts(ctx, authTable, ns.UUID); err != nil {
+			return err
+		}
+	}
+
+	if c.logger.IsInfo() {
+		c.logger.Info(fmt.Sprintf("successfully unmounted namespace %q mounts from auth table", ns.Path))
+	}
```
```suggestion
	c.logger.Info("unmounted all auth engines", "namespace", ns.Path)
```

vault/auth.go
```diff
@@ -1176,6 +1176,25 @@ func (c *Core) teardownCredentials(ctx context.Context) error {
...
 }
 
+// UnloadNamespaceCredentialMounts is used before we seal the namespace to reset
+// auth mounts to their unloaded state.
+func (c *Core) UnloadNamespaceCredentialMounts(ctx context.Context, ns *namespace.Namespace) error {
+	c.authLock.Lock()
+	defer c.authLock.Unlock()
+
+	if c.auth != nil {
+		authTable := c.auth.ShallowClone()
```
satoqz: Do we need to clone here?

vault/mount.go
```diff
@@ -1575,6 +1575,56 @@ func (c *Core) unloadMounts(ctx context.Context) error {
...
+	return nil
+}
+
+func (c *Core) cleanupNamespaceMounts(ctx context.Context, mountTable *routing.MountTable, nsUUID string) error {
```
satoqz: This is very similar to `undoCreateMounts` in `namespace_store.go`, even if the purpose is different as of now. Can we reuse?

vault/namespace_store.go
```diff
@@ -922,6 +932,40 @@ func (ns *NamespaceStore) ListNamespaces(ctx context.Context, includeParent bool
...
+			return
+		}
+
+		ctx = namespace.ContextWithNamespace(ctx, namespaceEntry)
+		errs = errors.Join(
+			ns.clearNamespacePolicies(ctx, namespaceEntry, false),
+			ns.core.identityStore.RemoveNamespaceView(namespaceEntry),
+			ns.core.UnloadNamespaceCredentialMounts(ctx, namespaceEntry),
+			ns.core.UnloadNamespaceSecretMounts(ctx, namespaceEntry),
+		)
```
satoqz: Similarly, I wonder if we can fully reuse `clearNamespaceResources` here if we make storage writes toggleable? Sealing a namespace is technically the same operation is deleting it in-memory.

vault/mount.go
wslabosz-reply: I don't think it's as similar, their internals are completely different, as `cleanupNamespaceMounts/Auths` is called on the sealing of the namespace, we don't taint the mount or route entry, we also do not clear the mount view.

wslabosz-reply: update: actually maybe additional operations like tainting of the mount/route is not that bad, will adjust.

vault/core.go
```diff
@@ -2201,21 +2198,19 @@ func (c *Core) sealInternalWithOptions(grabStateLock, keepHALock, performCleanup
...
+			c.logger.Error("error stopping storage cluster", "error", err)
+			return err
 		}
-
-		// Stop the cluster listener
-		c.stopClusterListener()
 	}
 
-	c.logger.Debug("sealing barrier")
-	if err := c.barrier.Seal(); err != nil {
```
satoqz, line 2217: There are several other call sites for `c.barrier.Seal()` - we should triage each one and see if they need to be updated to run a full seal, too.

vault/namespace_store.go
```diff
@@ -922,6 +922,40 @@ func (ns *NamespaceStore) ListNamespaces(ctx context.Context, includeParent bool
 	return ns.namespacesByPath.List(parent.Path, includeParent, recursive, ns.creationDeletionMap)
 }
 
+// sealNamespaceLocked assumes the read lock is hold, and seals provided namespace,
+// cleaning up namespace resources.
+func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSeal *namespace.Namespace) error {
```
satoqz, line 929: To recap internal discussion I've had with Wojciech for other reviewers:

This method will be used only to seal a sub-tree of namespaces via the API. To perform a root-level seal (usually on shutdown, but also possible the API) we just reuse the existing `preSeal` routine as that is far more performant given we know we need to wipe _everything_ from memory.

vault/core.go
wslabosz-reply: addressed

satoqz (approved):

vault/core.go
```diff
@@ -2172,9 +2172,6 @@ func (c *Core) sealInternalWithOptions(grabStateLock, keepHALock, performCleanup
 		// If we are keeping the lock we already have the state write lock
 		// held. Otherwise grab it here so that when stopCh is triggered we are
 		// locked.
-		if keepHALock {
-			c.keepHALockOnStepDown.Store(true)
```
cipherboy, line 2176: This must've been for a Vault Enterprise feature.

PR #2472, opened 2026-02-13
pki: fix "context canceled" issue when processing cache invalidation

phil9909:
## Description

This lead to pki returning 500 until reload.

## Rationale

The context is cancelled as soon as the invalidation returns, but the PKI engine processes the invalidation asynchronously. 

This causes newly created pki mount to be unusable on standbys (because the primary used storage version 1 and the standbys use storage version 0, causing issues like "could not fetch the CA certificate (was one set?): no legacy cert bundle exists").

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -5,6 +5,7 @@ package pki
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"strings"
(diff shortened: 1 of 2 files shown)


builtin/logical/pki/backend.go
```diff
@@ -470,7 +470,7 @@ func (b *backend) invalidate(ctx context.Context, key string) {
 		// within updatePkiStorageVersion.
 		go func() {
 			b.Logger().Info("Detected a migration completed, resetting pki storage version")
-			b.updatePkiStorageVersion(ctx, true)
+			b.updatePkiStorageVersion(context.Background(), true)
```
cipherboy: I wonder if we can/should bound this so we get cancellations in the case of plugin shutdown?

Either temporally or with a long-lived background context that `Shutdown()` cancels?

satoqz (approved):

builtin/logical/pki/backend.go
phil9909: changed this

Issue #2536, opened 2026-02-24
Failed to revoke lease after upgrading to >= v2.5.0


Schluggi:
**Describe the bug**
After upgrading to v2.5.0 our follower logs gone crazy with these messages:
```
[ERROR] expiration: failed to revoke lease: lease_id=auth/XXX/login/YYY error="failed to delete lease entry: cannot write to readonly storage" attempts=5 next_attempt=3m41.404979112s
```
Everything works fine so it seems more like a cosmetic. Could someone confirm?
We also upgraded to 2.5.1 this morning in hope #2467 would fix it, but nothing changed.

**To Reproduce**
Upgrading to >= v2.5.0.

**Expected behavior**
I would expect none of these errors appear? 

**Environment:**
* OpenBao Server Version: 2.5.1
* OpenBao CLI Version: 2.5.1
* Server Operating System/Architecture: RHEL9

OpenBao server configuration file(s):

```hcl
listener "tcp" {
  address          = "0.0.0.0:8200"
  cluster_address  = "192.168.24.2:8201"
  tls_cert_file    = "/etc/pki/tls/certs/vault.crt"
  tls_key_file     = "/home/openbao/vault.key"
  telemetry {
   unauthenticated_metrics_access="false"
  }
}

listener "tcp" {
  address                  = "0.0.0.0:8202"
  cluster_address          = "192.168.24.2:8203"
  tls_cert_file            = "/etc/pki/tls/certs/vault.crt"
  tls_key_file             = "/home/openbao/vault.key"
}

storage "raft" {
  node_id = "vault-prod2"
  path    = "/var/bao/data"
}

api_addr      = "https://vault-prod2.linuxdmz.de:8200"
cluster_addr  = "https://vault-prod2.linuxdmz.de:8201"
ui            = true
allow_audit_log_prefixing = true

telemetry {
   prometheus_retention_time="24h"
   disable_hostname=true
   enable_hostname_label=true
}
```

**Additional context**
If I understand it correctly, the message appears because the followers are trying to write to the raft storage, but only the leader can actually do this. Is that correct?

satoqz:
@Schluggi your description and understanding is correct, I've postponed fixing this one to v2.5.2 as it isn't blocking or fatal in any way and v2.5.1 already had other critical fixes that we wanted to get out soon. I agree that the logs are annoying and confusing though. Thanks for reporting, I hadn't gotten around to properly documenting this yet 😄

satoqz:
[...]

PR #2549, opened 2026-02-26
Skip lease restoration on standby nodes

wslabosz-reply:
## Description
This PR updates the setup logic of the expiration mgr. Skipping setup of the restore job manager run (and irrevocable leases revocation job).

## Rationale
Although not impactful, the restoration process takes some cpu cycles and introduces log noise, so definitely an experience enhancement for the bao administrator.

Resolves: #2536

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2549.txt
+++ b/changelog/2549.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Skip re-scheduling lease expiration jobs that need to write to storage when a node unseals in read-only mode.
+```


vault/expiration.go
```diff
@@ -384,27 +385,27 @@ func NewExpirationManager(c *Core, e ExpireLeaseStrategy, logger log.Logger, det
...
-	expMgr := NewExpirationManager(c, e, expLogger, detectDeadlocks)
+	expMgr := NewExpirationManager(c, e, expLogger, slices.Contains(c.detectDeadlocks, "expiration"))
 	c.expiration = expMgr
 
 	// Link the token store to this
 	c.tokenStore.SetExpirationManager(expMgr)
 
+	if standby {
+		c.logger.Info("skipping lease restoration - standby")
+		expMgr.restoreMode.Store(false)
```
wslabosz-reply: not sure if we should use the `restoreLock` at this point? @satoqz

satoqz: No I think this is correct.

vault/expiration.go
```diff
@@ -384,27 +385,27 @@ func NewExpirationManager(c *Core, e ExpireLeaseStrategy, logger log.Logger, det
...
+		expMgr.restoreLocks = nil
```
satoqz: Perhaps we can pass `standby` to `NewExpirationManager` and avoid creating them in the first place, plus setting the correct `restoreMode` value from the start too.

wslabosz-reply: hmm valid point, I wonder how that impacts (if it does) the first dry run (in standby) of the leader node

EDIT: adjusted

PR #2552, opened 2026-02-27
Fix expired test certificates

satoqz:
## Description

Fix test certificates that recently expired using a combination of arcane OpenSSL invocations & PKI. This required quite a few rotations because the root certificate used in the cert engine expired.

## Rationale

All new certificates expire in 100 years. Optimally though, we wouldn't need these fixtures and just generate certificates just in time in Go when tests run.

## Acknowledgements

       also allowed.


tsaarni:
On a side note, I maintain a project for generating test certificates and CRLs on the fly ([demo code](https://github.com/tsaarni/certy-tutorial/blob/main/go/README.md) and [project link](https://github.com/tsaarni/certyaml)) but it is not very widely used and I am the sole maintainer, so not the ideal dependency to add. I could implement similar functionality directly in OpenBao though, if that would be useful.

satoqz:
@tsaarni this looks nice. I'd certainly welcome a PoC; I think it would be ok to depend on your repo as it would be testonly. Though since we need the Go API parts only, perhaps it is cleaner to vendor those directly. But anything is better than the current workflow, a declarative/codified setup would also make the purpose of certain certificates much clearer (I had to poke around with `openssl` for quite a while to figure out what to do here...)

JanMa (approved):

tsaarni:
@satoqz, I created a poc #2554. Please let me know your thoughts.

Based on the files replaced in this PR,  `vault/diagnose/test-fixtures/twoRootCA.pem` might be the only remaining certificate to update. I saw plenty of others left in the project, but maybe they are not expired yet?

satoqz:
@tsaarni yes, there are plenty more certificates in other locations, but many of those have been renewed not long ago (This is not the first PR of its kind...). Regardless of the remaining validity period, it'd be nice to codify all of those if possible.

I'll merge this for now and have a closer look at your changes next week :)

PR #2554, opened 2026-02-28
PoC: replace hardcoded certs

tsaarni:
## Description

This PR is a proof-of-concept that aims to remove hardcoded test certificates, keys, and CRL files from the project. These files are replaced with on-demand generation using [certyaml](https://github.com/tsaarni/certyaml). Currently only `./builtin/credential/cert/` is covered. More directories still need to be addressed.

Note! I am the sole maintainer of `certyaml`, and the project is not widely used. I am open to contributing parts of that code to OpenBao as a test utility if preferred.

## Rationale

Hardcoding test certificates can be bad because:

* They can expire, as seen in #2552
* They can trigger false alarms in secret scanners.
* They are opaque binary blobs that may contain malicious content, see the xz incident report [here](https://yieldcode.blog/post/xz-backdoor/).

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/credential/cert/backend_test.go
+++ b/builtin/credential/cert/backend_test.go
@@ -318,11 +318,11 @@ func TestBackend_PermittedDNSDomainsIntermediateCA(t *testing.T) {
 	// This is used to configure TLS in the api client.
 	tempDir := t.TempDir()
 	caCertFile := filepath.Join(tempDir, "ca.pem")
-	os.WriteFile(caCertFile, []byte(cluster.CACertPEM), 0o600)
+	require.NoError(t, os.WriteFile(caCertFile, []byte(cluster.CACertPEM), 0o600))
 	leafCertFile := filepath.Join(tempDir, "leaf.pem")
-	os.WriteFile(leafCertFile, []byte(leafCertPEM), 0o600)
+	require.NoError(t, os.WriteFile(leafCertFile, []byte(leafCertPEM), 0o600))
 	leafCertKeyFile := filepath.Join(tempDir, "leaf-key.pem")
-	os.WriteFile(leafCertKeyFile, []byte(leafCertKeyPEM), 0o600)
+	require.NoError(t, os.WriteFile(leafCertKeyFile, []byte(leafCertKeyPEM), 0o600))
 
 	// This function is a copy-pasta from the NewTestCluster, with the
 	// modification to reconfigure the TLS on the api client with the leaf
@@ -411,7 +411,7 @@ func TestBackend_MetadataBasedACLPolicy(t *testing.T) {
 	tempDir := t.TempDir()
 	exCertFile := filepath.Join(tempDir, "extcert.pem")
 	exCertKeyFile := filepath.Join(tempDir, "extcert-key.pem")
-	extCert.WritePEM(exCertFile, exCertKeyFile)
+	require.NoError(t, extCert.WritePEM(exCertFile, exCertKeyFile))
 
 	// Start cluster with cert auth method enabled
 	coreConfig := &vault.CoreConfig{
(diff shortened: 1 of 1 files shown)


phil9909:
In general I like the idea, but I guess this will slow down the tests, right?

Could we:
- Automate the creation of the certificates, but still commit them (`go generate` could be used to update the certs).
- Cache them locally, but ignored via `.gitignore`, automatically regenerating the certs, if necessary, otherwise using cache (in CI we should then make sure to persist this cache).

In another project, I had certificates generated on-demand for unit tests. Especially for RSA4096 this really slowed the development cycle down.

satoqz:
> but I guess this will slow down the tests, right?
> Especially for RSA4096 this really slowed the development cycle down.

IIUC @tsaarni's library uses ECDSA by default, which is *A LOT* faster than RSA. This shows in the test times, which really don't change:

```
$ go test ./builtin/credential/cert
ok  	github.com/openbao/openbao/builtin/credential/cert	27.708s

$ gh pr checkout 2554
From https://github.com/openbao/openbao
 * [new ref]               refs/pull/2554/head -> poc-with-certyaml
Switched to branch 'poc-with-certyaml'

$ go test ./builtin/credential/cert
ok  	github.com/openbao/openbao/builtin/credential/cert	27.581s
```

In fact, ECDSA and Ed25519 are so fast at generating a key pair and signing a certificate they aren't even an order of magnitude slower than reading from disk:

```
BenchmarkGenerateAndSignCertRSA-11        	       2	 507542896 ns/op
BenchmarkGenerateAndSignCertECDSA-11      	   15544	     77292 ns/op
BenchmarkGenerateAndSignCertEd25519-11    	   17180	     70077 ns/op
BenchmarkReadCertFromDisk-11              	  108765	     10936 ns/op
```

I think the difference is negligible.

Though RSA really is bad. Luckily we don't have a reason to use it except for cases where it must be tested explicitly. I agree it might be better to just commit the cert there, we can just use certyamls _actual_ yaml files to manage those then I'd say.

tsaarni:
So far I have not needed to store certificates for performance reasons in any project, though I have moved away from RSA. The default in certyaml is EC 256 unless specified otherwise.

I ran benchmark for certificate generation using different algorithms and sizes ([link](https://gist.github.com/tsaarni/76df47cfabb7fc64a98cdc0951f5ed90))

```
$ go test -bench=. -benchmem 
goos: linux
goarch: amd64
pkg: example.com/m
cpu: 12th Gen Intel(R) Core(TM) i9-12900H
BenchmarkGenerateEC256-20          11551            106750 ns/op           17662 B/op        309 allocs/op
BenchmarkGenerateEC384-20           1538            702434 ns/op           18906 B/op        330 allocs/op
BenchmarkGenerateEC521-20            655           1813454 ns/op           21496 B/op        340 allocs/op
BenchmarkGenerateEd25519-20        14910             80108 ns/op            9559 B/op        221 allocs/op
BenchmarkGenerateRSA1024-20          126           9176774 ns/op          296021 B/op       3099 allocs/op
BenchmarkGenerateRSA2048-20           31          39983448 ns/op          604289 B/op       5616 allocs/op
BenchmarkGenerateRSA4096-20           12         454578581 ns/op         2595466 B/op      12971 allocs/op
PASS
ok      example.com/m   12.555s
```

On my machine, generating an EC 256 certificate took 0.106 ms, so performance is likely not a concern. I have not seen OpenBao tests that require a specific key type, so we can avoid using RSA 4096, which took 454 ms. Most tests focus on SANs and other attributes.

tsaarni:
Sorry, I did not see @satoqz just responded the same :-D

phil9909:
Just an idea: https://github.com/phil9909/certyaml/commit/b3317546ea47b0d93aafb90384ab14130c65f0fd

I changed certyaml to include 64 precomputed rsa4096 keys:

```console
~/workspace/certyaml $ git switch master
Switched to branch 'master'
Your branch is up to date with 'upstream/master'.

~/workspace/certyaml $ go test . --count=1
ok  	github.com/tsaarni/certyaml	0.662s

~/workspace/certyaml $ git switch precompute-rsa-4096
Switched to branch 'precompute-rsa-4096'
Your branch is up to date with 'origin/precompute-rsa-4096'.

~/workspace/certyaml $ go test . --count=1
ok  	github.com/tsaarni/certyaml	0.072s
```

The RSA keys themselves don't expire and therefore don't have to be recomputed periodically.

Maybe this needs explicit opt-in, e.g. by adding a `AllowPrecomputed bool` in the `Certifcate struct`

**EDIT:**
This is also something we could do later, if we feel we need a speedup.

tsaarni:
Thanks @phil9909! Yes, lets revisit this later if a speedup becomes necessary. 

I generally have thought that RSA days are counted and we should move towards other standards :) At the moment, it feels like high performance for RSA 4096 is quite rare requirement, and I'd like to avoid embedding precomputed keys into certyaml itself. But surely some way can be found when necessary!

satoqz:
@openbao/openbao-org-maintainers any strong feelings on this? Otherwise I'm inclined to merge.

satoqz:
Merging as per discussion in [#openssf-openbao-wg > PRs to review @ 💬](https://linuxfoundation.zulipchat.com/#narrow/channel/574533-openssf-openbao-wg/topic/PRs.20to.20review/near/581441661). Thanks @tsaarni!

tsaarni:
Thanks @satoqz, I did start removing hardcoded certificates from the remaining packages as well. I will send a follow-up once finished.

Issue #2548, opened 2026-02-26
HA standby nodes return 500 for ?help=1 requests when disable_standby_reads = true


Huy-Dinh-Reply:
**Describe the bug**
In an HA setup with OpenBao 2.5.0/2.5.1 with `disable_standby_reads = true`, API help requests that arrive at a standby node return:
- `500 {"errors":["please forward to the active node"]}`

**To Reproduce**
1. Run OpenBao HA (3+ nodes, Raft) with `disable_standby_reads = true` in the server config.
2. Port-forward one active node and one standby node.
3. Send the same help request to both nodes, for example:
   - `GET /v1/sys/health?help=1`
4. Observe:
   - standby returns `500` with `please forward to the active node`
   - active returns a normal help response (`200`)

Other paths that also produce this:
- `/v1/sys/internal/ui/mounts?help=1`
- `/v1/sys/policies/acl/?help=1`
- `/v1/auth/<path>/?help=1`

**Expected behavior**
Requests should succeed on standby nodes as well.

**Environment:**
* OpenBao Server Version (retrieve with bao status): 2.5.1 (also observed on 2.5.0)
* OpenBao CLI Version (retrieve with bao version): 2.5.1
* Server Operating System/Architecture: Kubernetes (kind), Linux

OpenBao server configuration file(s):

```hcl
ui = true
disable_standby_reads = true

listener "tcp" {
  address = "[::]:8200"
  cluster_address = "[::]:8201"
  tls_disable = "true"
}

storage "raft" {
  path = "/openbao/data"

  retry_join {
    leader_api_addr = "http://openbao-0.openbao-internal:8200"
  }
  retry_join {
    leader_api_addr = "http://openbao-1.openbao-internal:8200"
  }
  retry_join {
    leader_api_addr = "http://openbao-2.openbao-internal:8200"
  }
}

seal "static" {
  current_key_id = "1"
  current_key    = "file:///openbao/secrets/static-seal/unseal-1.key"
}
```

**Additional context**
I observed this on the UI while trying to look at auth mounts.

<img width="1300" height="348" alt="Image" src="https://github.com/user-attachments/assets/a940c55a-c3bd-4089-99f2-23994017d963" />

wslabosz-reply:
I've prepared a PR for this one.

PR #2572, opened 2026-03-02
Handle help requests on standby nodes when reads are disabled

wslabosz-reply:
## Description
Help requests couldn't be handled on the standby nodes (with `disable_standby_reads` prop being `true`), as the help request handler didn't propagate the request to the active node. This PR introduces propagation of such requests, still retaining local responses when the `disable_standby_reads` is `false`.

## Rationale
This fixes minor ux inconvenience making an user unable to read help requests while calling standby node (with disabled read requests).

Even though help requests do not read anything from storage, there's not a pattern of responding to any requests on standby nodes with `disable_standby_reads` set to `true` hence we need to forward. (This definitely could use of a refactor of a help request handling, as the forward introduced unnecessary traffic).

Resolves: #2548

## Acknowledgements

       also allowed.

### DIFF
--- a/http/help.go
+++ b/http/help.go
@@ -29,7 +29,7 @@ func wrapHelpHandler(h http.Handler, core *vault.Core) http.Handler {
 func handleHelp(core *vault.Core, w http.ResponseWriter, r *http.Request) {
 	if !strings.HasPrefix(r.URL.Path, "/v1/") {
 		//nolint:staticcheck // User facing error
-		respondError(w, http.StatusNotFound, errors.New("Missing /v1/ prefix in path. Use bao path-help command to retrieve API help for paths"))
+		respondError(w, http.StatusNotFound, errors.New("Missing /v1/ prefix in path. Use the bao path-help command to retrieve API help for paths"))
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


http/help.go
```diff
@@ -29,7 +29,12 @@ func wrapHelpHandler(h http.Handler, core *vault.Core) http.Handler {
 func handleHelp(core *vault.Core, w http.ResponseWriter, r *http.Request) {
 	if !strings.HasPrefix(r.URL.Path, "/v1/") {
 		//nolint:staticcheck // User facing error
-		respondError(w, http.StatusNotFound, errors.New("Missing /v1/ prefix in path. Use vault path-help command to retrieve API help for paths"))
+		respondError(w, http.StatusNotFound, errors.New("Missing /v1/ prefix in path. Use bao path-help command to retrieve API help for paths"))
```
satoqz:
```suggestion
		respondError(w, http.StatusNotFound, errors.New("Missing /v1/ prefix in path. Use the bao path-help command to retrieve API help for paths"))
```

satoqz (approved):

PR #2573, opened 2026-03-02
feat(docs): rework/refactor downloads page

gianklug:
## Description
    
I've taken some time to rework the downloads page. It should now be less cluttered and easier to work with.

Demo at https://openbao-dl-rework.pages.dev/downloads/

## Acknowledgements

       also allowed.

### DIFF
--- a/website/src/components/CodeBlockWrap/styles.module.css
+++ b/website/src/components/CodeBlockWrap/styles.module.css
@@ -23,8 +23,8 @@
     mask-image: none;
   }
 
-  &:checked ~ .codeBlockLabel {
-    display: none;
+  &:checked ~ .codeBlockLabel:after {
+    content: "show less ‹";
   }
 }
 


pree:
This looks really nice, thanks @gianklug!

I think it would be nice to have the long code blocks collapsible again after expanding.
As discussed via PM I think we can also remove the "Toggle word wrap" button, as it's not doing anything and is currently buggy.

gianklug:
Let's do that in another refactoring round and raise a follow-up issue for it. In my opinion we should make all of the downloads page code even more generic and move things into components.

satoqz:
Note that my goal is to remove HSM binaries by v2.7 and move PKCS#11 functionality into a plugin, so perhaps I wouldn't prioritize that too much :D

website/src/pages/downloads.tsx
```diff
@@ -172,40 +174,43 @@ const Asset = ({ urls }) => {
...
-          <p className="text--center">Version: {selectedItem}</p>
-        </div>
-        <div className="card__footer">
-          <div class="button-group button-group--block">
-            <a class="button button--primary" href={asset}>
-              <span>Download</span>
-            </a>
+      <div className="card__header">
+        <h3 className="download-card__header-title">
+          {AssetArchitecture(asset).toUpperCase()}
```
satoqz, line 183: WDYT about not uppercasing these? I think the lowercase versions are easier to parse and more familiar.

satoqz: Eh I'll merge for now and perhaps tweak this later.

Issue #2553, opened 2026-02-27
/v1/sys/storage/raft requests are not forwarded to the leader


rfwatson:
**Describe the bug**
Requests to remove a Raft peer are not forwarded from follower nodes to leader.

**To Reproduce**
Make the following request to a Raft follower.

```
$ curl -k -H "x-vault-token: $BAO_TOKEN" -X POST -d '{"server_id": "any_server_id"}' https://127.0.0.1:8200/v1/sys/storage/raft/remove-peer
{"errors":["1 error occurred:\n\t* raft storage autopilot is not initialized\n\n"]}
```

Note that autopilot is enabled, and making the same request directly to the leader is successful.

**Expected behavior**
Since OpenBao 2.5.0, the request should be forwarded to the leader. From reading the code I believe this is because the [ShouldForward](https://github.com/openbao/openbao/blob/6666b2a3027693d194a24482d46add790cc1aa63/sdk/logical/error.go#L135-L147) function does not match [this error string](https://github.com/openbao/openbao/blob/6666b2a3027693d194a24482d46add790cc1aa63/physical/raft/raft.go#L1229).

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.5.1
* OpenBao CLI Version (retrieve with `bao version`): 2.5.1
* Server Operating System/Architecture: Linux/amd64

OpenBao server configuration file(s):

```hcl
api_addr = "https://example.com:8200"
cluster_addr = "https://1.2.3.4:8201"

ui = true
log_level = "debug"
plugin_directory = "/opt/openbao/plugins"
plugin_auto_download = true
plugin_auto_register = true

storage "raft" {
  node_id = "i-075b3039a814ace17"
  path = "/opt/openbao/data"
  retry_join {
    auto_join           = "provider=aws region=eu-west-1 addr_type=private_v4 tag_key=openbao-role tag_value=node"
    leader_ca_cert_file = "ca.crt"
  }
}

listener "tcp" {
  address             = "0.0.0.0:8200"
  tls_cert_file       = "tls.crt"
  tls_key_file        = "tls.key"
}

seal "awskms" {
  region = "eu-west-1"
  kms_key_id = "foo"
}

plugin "auth" "aws" {
  image       = "ghcr.io/openbao/openbao-plugin-auth-aws"
  version     = "v0.1.1"
  sha256sum   = "7a77057e62973c1aae6035f52110e3302605a47b622756d954915b7b55eca10c"
[...]

PR #2574, opened 2026-03-02
Forward raft autopilot operations

wslabosz-reply:
## Description
This PR makes Raft storage handlers propagate the request to the active node by default for following operations:
- peer demote/promote
- peer remove/join
- autopilot config read/update

## Rationale
Standby nodes are unable to handle raft cluster management requests, and requests weren't properly propagated. Additionally snapshot writes are not handled on the standby nodes, so that also should be propagated. Finally Raft autopilot is only configured on the active node, so config reads/updates won't happen unless propagated.

Resolves: #2553

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2574.txt
+++ b/changelog/2574.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+raft: Propagate peer join/remove/promote/demote and autopilot read/update requests to active node.
+```


changelog/2574.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+raft: Fix raft peer add/remove/promote/demote operations on standby nodes not being propagated to active node.
```
satoqz:
```suggestion
raft: Fix peer add/remove/promote/demote operations on standby nodes not being propagated to active node.
```

needless repetition

physical/raft/raft.go
```diff
@@ -1226,7 +1226,10 @@ func (b *RaftBackend) RemovePeer(ctx context.Context, peerID string) error {
 	}
 
 	if b.autopilot == nil {
-		return errors.New("raft storage autopilot is not initialized")
+		if b.HAEnabled() && b.raft.State() == raft.Follower {
+			return fmt.Errorf("%w: %w", ErrRaftAutopilotNotInitialized, logical.ErrPerfStandbyPleaseForward)
```
cipherboy: I'd suggest perhaps adding `ForwardPerformanceStandby` to `storage/raft/autopilot/configuration` in `vault/logical_system_raft.go` on line 230-ish is fewer lines of code. 

Though, correct way of behavior should probably be with @JanMa.

cipherboy: Tagging @wslabosz-reply :-D

wslabosz-reply: that would have to be added to the `storage/raft/bootstrap/challenge`, `storage/raft/remove-peer`, `storage/raft/promote`, and `storage/raft/demote` also, if we want to remove it from raft backend code.

wslabosz-reply: @cipherboy adjusted

PR #2585, opened 2026-03-05
Add Generation, period, and expiration for TOTP codes

wslabosz-reply:
## Description
This PR adds `generated`, `expiry`, and `period` fields to totp code generation.

## Rationale
As described in the source issue, it's pretty useful to know all of these things at the stage of the code generation (without having to do a round trip to get details of the key).

Resolves: #2233 

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/logical/totp/backend_test.go
+++ b/builtin/logical/totp/backend_test.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"strings"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/go-viper/mapstructure/v2"
@@ -113,6 +114,65 @@ func TestBackend_KeyName(t *testing.T) {
 		require.NotEmpty(t, resp.Data["expire_time"].(int64))
 		require.NotEmpty(t, resp.Data["period"].(string))
 	}
+}
+
+func TestBackend_Timesteps(t *testing.T) {
+	config := logical.TestBackendConfig()
+	config.StorageView = &logical.InmemStorage{}
+	ctx := namespace.RootContext(t.Context())
+	b, err := Factory(ctx, config)
+	require.NoError(t, err)
+
+	resp, err := b.HandleRequest(ctx, &logical.Request{
+		Path:      "keys/test",
+		Operation: logical.UpdateOperation,
+		Storage:   config.StorageView,
+		Data: map[string]interface{}{
+			"generate":     true,
+			"account_name": "bao",
+			"issuer":       "openbao",
+		},
+	})
+	require.NoError(t, err)
+
+	generateCode := func(ctx context.Context) (int64, int64) {
+		resp, err = b.HandleRequest(ctx, &logical.Request{
+			Path:      "code/test",
+			Operation: logical.ReadOperation,
+			Storage:   config.StorageView,
+		})
+		require.NoError(t, err)
+		require.Len(t, resp.Data["code"].(string), 6)
+		require.Equal(t, "30s", resp.Data["period"].(string))
+		return resp.Data["generated"].(int64), resp.Data["expire_time"].(int64)
+	}
+
+	synctest.Test(t, func(t *testing.T) {
+		generated, expire_time := generateCode(t.Context())
+		t.Logf("generated: %v | expire_time: %v", generated, expire_time)
+		halfStepTime := 15 * time.Second
+
+		// fake clock in a bubble (synctest) starts at UTC 2000-01-01
+		start := time.Now()
+		require.Equal(t, start.Unix(), generated)
+		require.Equal(t, start.Add(30*time.Second).Unix(), expire_time)
+		require.WithinDuration(t, time.Unix(generated, 0), time.Unix(expire_time, 0), 2*halfStepTime)
+
+		time.Sleep(halfStepTime)
+
+		generated, expire_time = generateCode(t.Context())
+		t.Logf("generated: %v | expire_time: %v", generated, expire_time)
+		require.Equal(t, start.Add(halfStepTime).Unix(), generated)
+		require.Equal(t, start.Add(30*time.Second).Unix(), expire_time)
+
+		time.Sleep(halfStepTime)
+
+		generated, expire_time = generateCode(t.Context())
+		t.Logf("generated: %v | expire_time: %v", generated, expire_time)
+		require.Equal(t, start.Add(2*halfStepTime).Unix(), generated)
+		// next step
+		require.Equal(t, start.Add(4*halfStepTime).Unix(), expire_time)
+	})
 }
 
 func TestBackend_readCredentialsDefaultValues(t *testing.T) {
@@ -1227,9 +1287,7 @@ func testAccStepValidateCode(t *testing.T, name, code string, valid, expectError
 		},
 		ErrorOk: expectError,
 		Check: func(resp *logical.Response) error {
-			if resp == nil {
-				return fmt.Errorf("bad: %#v", resp)
-			}
+			require.NotNil(t, resp)
 
 			var d struct {
 				Valid bool `mapstructure:"valid"`
@@ -1238,18 +1296,7 @@ func testAccStepValidateCode(t *testing.T, name, code string, valid, expectError
 			if err := mapstructure.Decode(resp.Data, &d); err != nil {
 				return err
 			}
-
-			switch valid {
-			case true:
-				if d.Valid != true {
-					return fmt.Errorf("code was not valid: %s", code)
-				}
-
-			default:
-				if d.Valid != false {
-					return fmt.Errorf("code was incorrectly validated: %s", code)
-				}
-			}
+			require.Equal(t, valid, d.Valid)
 			return nil
 		},
 	}


wslabosz-reply:
@johnnybubonic
would be great if you could verify if this matches your expectations, thanks!

johnnybubonic:
will take a look in-depth when i can fit it in soon, but the analysis bot(?) above is spot-on - expiration *aligns to* a period, the period isn't *added on*.

closely related, see https://github.com/hashicorp/vault/issues/31690

wslabosz-reply:
@johnnybubonic 
Thank you, corrected the expiry calculation.

builtin/logical/totp/nanoepoch.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (n *NanoEpoch) UnmarshalJSON(b []byte) (err error) {
+	if len(b) == 0 {
+		return err
+	}
+
+	var sec int64
+	var nsec int64
+	spl := strings.SplitN(string(b), ".", 2)
```
phil9909: [`strings.Cut`](https://pkg.go.dev/strings#Cut) is designed exactly for this use case.

wslabosz-reply: adjusted (not gonna lie, haven't given it much thought, just used the baseline code snippet provided with the issue)

DanGhita: @phil9909 : I would definitely not implement this with a nano second precision ! First, I don't see why we would need this precision. Secondly, IMHO, this precision would provide too much insights about implementation and maybe (I know it's paranoîa and my scenario very unlikely) someone could use this for side channels attacks (time based attack) .

wslabosz-reply:
@DanGhita 
> Use the same precision as for "creation_time" . For the reasons provided in the Conversation, avoid nanosecond precision!

the reasons being:
1. > First, I don't see why we would need this precision.

well, I'd wait for @johnnybubonic input before deciding for/against.

2. > Secondly, IMHO, this precision would provide too much insights about implementation and maybe (I know it's paranoîa and my scenario very unlikely) someone could use this for side channels attacks (time based attack)

not really sure what you mean by "insights about implementation" in context of opensource codebase. Also don't know about "likeliness" of a scenario you haven't really described, I can imagine one would be able to tell the timespan between request <=> code generation <=> response, I don't think there's a path there, where time-based attack could be utilized for sensitive material extraction (basically the leak of the root key of the totp). Remember that access to this endpoint follows the security model of OpenBao (as access to any secret engine) and requires appropriate ACL permissions.

> For retrocompatibility and security reasons

It's not breaking in any way, so there's not a reason to introduce opt-in flag.
Security-wise, there's only implication of a leak of OpenBao version that you are running, but this would also apply for almost every other feature we've shipped across past releases, and I don't think this is covered in our threat model.

> the new returned fields should rather respect the naming used for the self [[tokens](https://openbao.org/api-docs/auth/token/#lookup-a-token-self)] lookup: `"creation_time"`, `"creation_ttl"`, `"expire_time"` etc.

tokens are OpenBao internals, generated totp code is either issued or tracked by OpenBao Secret Engine, so I don't think they *have* to match the format of OpenBao authorization engine, but up for debate.

DanGhita:
> @DanGhita
> 
> > Use the same precision as for "creation_time" . For the reasons provided in the Conversation, avoid nanosecond precision!
> 
> the reasons being:
> 
> 1. > First, I don't see why we would need this precision.
> 
> well, I'd wait for @johnnybubonic input before deciding for/against.
> 
> 2. > Secondly, IMHO, this precision would provide too much insights about implementation and maybe (I know it's paranoîa and my scenario very unlikely) someone could use this for side channels attacks (time based attack)
> 
> not really sure what you mean by "insights about implementation" in context of opensource codebase. Also don't know about "likeliness" of a scenario you haven't really described, I can imagine one would be able to tell the timespan between request <=> code generation <=> response, I don't think there's a path there, where time-based attack could be utilized for sensitive material extraction (basically the leak of the root key of the totp). Remember that access to this endpoint follows the security model of OpenBao (as access to any secret engine) and requires appropriate ACL permissions.

From the beginning, I agreed this was a highly unlikely scenario.  But why tempting the devil with a useless (and potentially dangerous) feature ? Just let's use the same precision as for the other OpenBao TTLs, leases etc. 
Because yes, theoretically, nanosecond precision might leak unwanted information about underlying hardware, the crypto library version (FIPS or not, for ex.), or even worse, maybe information about the key, etc. I suppose the administrator of an OpenBao cluster would want to keep this kind of information "hidden" from everyone (including employees, legitimately possessing OpenBao tokens).

> 
> > For retrocompatibility and security reasons
> 
> It's not breaking in any way, so there's not a reason to introduce opt-in flag. Security-wise, there's only implication of a leak of OpenBao version that you are running, but this would also apply for almost every other feature we've shipped across past releases, and I don't think this is covered in our threat model.
>
I cannot guarantee that current clients using OpenBao would not run into issues if the JSON response changes (they shouldn't, but who knows ?).
From a security point a view: you might have administrators wanting to "hide" the currently  defined  validity time window. If, for some reasons, they have to configure a longer period (let's say 30 minutes instead of 30 seconds), they won't necessarily want to shout it from the rooftops.
> > the new returned fields should rather respect the naming used for the self [[tokens](https://openbao.org/api-docs/auth/token/#lookup-a-token-self)] lookup: `"creation_time"`, `"creation_ttl"`, `"expire_time"` etc.
> 
> tokens are OpenBao internals, generated totp code is either issued or tracked by OpenBao Secret Engine, so I don't think they _have_ to match the format of OpenBao authorization engine, but up for debate.

Of course, the general idea here was to try to remain consistent with the existing API and re-use, as much as possible,  the namings already existing. The token was an example of OpenBao objects having this kind of attributes. For instance,  since we already have "`expire_time`", I suggested to use this instead of the initial proposal "`expiry`".

wslabosz-reply:
> From the beginning, I agreed this was a highly unlikely scenario. But why tempting the devil with a useless (and potentially dangerous) feature ?

If the feature is useless, we shouldn't debate adding it, just close up the PR.

> theoretically, nanosecond precision might leak unwanted information about the crypto library version (FIPS or not, for ex.), or even worse, maybe information about the key, etc.

I'd kindly ask for a explanation how would that work, how exactly *this* PR might leak the totp key or FIPS-enablement.

> I cannot guarantee that current clients using OpenBao would not run into issues if the JSON response changes (they shouldn't, but who knows ?).

I don't think there is a single json parsing client library that *doesn't* ignore additional fields by default. I don't see any value in a stance of "this shouldn't happen in all foreseeable cases, but what about cases we don't think of".

> From a security point a view: you might have administrators wanting to "hide" the currently defined validity time window. If, for some reasons, they have to configure a longer period (let's say 30 minutes instead of 30 seconds), they won't necessarily want to shout it from the rooftops.

The whole idea is that this exact information is available to the end-user,  but with an additional call to the `/validate` endpoint, it's not "shouting" it's just saving a round trip to other endpoint to get the same information you'd be able to get with one call. This is inherently not a sensitive information.

> Of course, the general idea here was to try to remain consistent with the existing API and re-use, as much as possible, the namings already existing. The token was an example of OpenBao objects having this kind of attributes. For instance, since we already have "expire_time", I suggested to use this instead of the initial proposal "expiry".

We do have:
- `expire_time` for token lookup, lease lookup (and in future ControlGroups) which are either RFC3339Nano or RFC3339
- `expiration` for Certificate data (UNIX)
- `expiry` for CRL parameter (string time, like `period` for totp)
- `exp` for OIDC (UNIX)
- `expiration_time` for AppRoles (RFC3339)

my point being, the naming exists, but is not at all consistent, in my mind, most suited choice would be `generation` and `expiration` (both being UNIX/UNIXNano)

builtin/logical/totp/nanoepoch.go
cipherboy: @DanGhita @wslabosz-reply Fairly sure precision does not matter here; per [TOTP RFC 6238, Section 4.2 Description](https://datatracker.ietf.org/doc/html/rfc6238#section-4.2):

>   Basically, we define TOTP as `TOTP = HOTP(K, T)`, where T is an integer
>   and represents the number of time steps between the initial counter
>   time T0 and the current Unix time.
>
>   More specifically, `T = (Current Unix time - T0) / X`, where the
>   default floor function is used in the computation.
>
>   For example, with T0 = 0 and Time Step X = 30, T = 1 if the current
>   Unix time is 59 seconds, and T = 2 if the current Unix time is
>   60 seconds.

So, not only is nanosecond precision unnecessary (from a client PoV), it is also security irrelevant. There should be no timing side-channel here.

cipherboy:
> * For retrocompatibility and security reasons, I would advise to add a new boolean parameter to the create TOTP key API for explicitly allowing to return the new fields. That would be something like `"allow_ttl_info"` and it would default to false for keeping unchanged the current behaviour (just return the code with no extra info).

@DanGhita I don't think we need to worry about compatibility here; this is a new response parameter; it doesn't need a new request parameter flag to enable. We do not have that level of API compatibility guarantee to date: adding new response fields is done frequently; we only promise not to break existing fields. My 2c.

Also, `allow_ttl_info` is not able to be properly/easily ACL'd; it would be better to add an entirely separate endpoint if this were a concern fwiw.

builtin/logical/totp/backend_test.go
cipherboy, line 1: Its hard in the step-driven test suite, but you could theoretically add a different test which validates that the values are useful and correct, not just non-empty fwiw. With mocked time, it might be possible without sleeps, though that gets perhaps harder to inject back into the server.  

Might be valuable as we're expecting people to rely on the accuracy of these values for what its worth. 🤷 

But I'm confident with the implementation.

wslabosz-reply:
Summarizing actionable items from multiple discussions over proposed changes:
- nanosecond precision is unnecessary and should be removed (unless @johnnybubonic has something to say about it).
- `generated` and `expiry` field names should be changed to respectively (my proposal) `generation` and `expiration`.

cipherboy:
> * `generated` and `expiry` field names should be changed to respectively (my proposal) `generation` and `expiration`.

Sorry, why change this? From an English PoV, the former seem more relevant. 

I'm also not fully sure that `expiry` is counted correctly, but I still have to dig through the implementation.

builtin/logical/totp/nanoepoch.go
johnnybubonic: > Two questions come to my mind:
> 
> * Do we really need nano second precision here? I guess we should be fine with seconds

for fidelity? iirc seconds should be fine, but nanoseconds were provided for unmarshaling. think of downstream consumers. if it's a nanosecond, consumers can use e.g. a `time.Time` directly for the field without needing a custom type simply to unmarshal.

> * Does the `generated` timestamp provide any value?

generally for viz, again on downstream consumers - providing this value gives UI consumers choice on how/what to render without additional calculation.

johnnybubonic:
> Please see here-below my remarks/requests.
> 
> * For retrocompatibility and security reasons, I would advise to add a new boolean parameter to the create TOTP key API for explicitly allowing to return the new fields. That would be something like `"allow_ttl_info"` and it would default to false for keeping unchanged the current behaviour (just return the code with no extra info).

these would be extra fields. the return would be backwards-compatible unless you're returning errors on extra fields. to my knowledge, there isn't a JSON parser in the common languages that does this by default.

additionally, expiration data is necessary because it is expected to be provided to the consumer. it does not present a security risk because otherwise *the consumer has no idea when the token expires*, and *when/if it would need to regen/refetch*. that's not "more secure", that's "unusable".

> * Use the same precision as for "`creation_time`" . For the reasons provided in the Conversation, avoid nanosecond precision!

unmarshaling, as mentioned.

johnnybubonic:
@wslabosz-reply
> 2. > Secondly, IMHO, this precision would provide too much insights about implementation and maybe (I know it's paranoîa and my scenario very unlikely) someone could use this for side channels attacks (time based attack)
> 
> not really sure what you mean by "insights about implementation" in context of opensource codebase. Also don't know about "likeliness" of a scenario you haven't really described, I can imagine one would be able to tell the timespan between request <=> code generation <=> response, I don't think there's a path there, where time-based attack could be utilized for sensitive material extraction (basically the leak of the root key of the totp). Remember that access to this endpoint follows the security model of OpenBao (as access to any secret engine) and requires appropriate ACL permissions.
@DanGhita 
> From the beginning, I agreed this was a highly unlikely scenario. But why tempting the devil with a useless (and potentially dangerous) feature ? Just let's use the same precision as for the other OpenBao TTLs, leases etc.
Because yes, theoretically, nanosecond precision might leak unwanted information about underlying hardware, the crypto library version (FIPS or not, for ex.), or even worse, maybe information about the key, etc. I suppose the administrator of an OpenBao cluster would want to keep this kind of information "hidden" from everyone (including employees, legitimately possessing OpenBao tokens).

i'd be curious to hear of a feasible attack in which you could derive key information from the algorithms outlined in the RFC, given that the `generated` value isn't an internal delta - it's an external one. the only start period the consumer has is when they made the request - which would require their clock guaranteed to be aligned to the server. (which it isn't.)

the proposed `generated` field is provided so the consumer can determine latency between request and code validity, allowing consumers to adjust request or backoff if there's no way to receive a valid TOTP token within the time period instead of continuously trying, risking an account lockout for no reason.

johnnybubonic:
@DanGhita
>From a security point a view: you might have administrators wanting to "hide" the currently defined validity time window. If, for some reasons, they have to configure a longer period (let's say 30 minutes instead of 30 seconds), they won't necessarily want to shout it from the rooftops.

This breaks the expectations in RFC.
In all normal non-centralized cases of TOTP, a consumer IS the generator, and *knows the expiry/validity of the token*. This is considered expected knowledge. Consumers of TOTP tokens **cannot function properly without this knowledge**, hence the original opening of the related issue.
A client cannot determine if or when they need to generate a new code unless the expiry is known. TOTP are not expected to be used at the exact same moment of generation (which is why they operate on *time periods* rather than a simple *timestamp with drift*), and when generation is centralized and **fetched** as it is in e.g. OpenBao, the consumer has even less visibility - it doesn't know when the code is generated, it doesn't know if it needs a new one because the time between generation and usage is invalid.

Given that a response from an API call has no way of "staying up to date/in-line with the time period" as local TOTP generators do, this means this information is in fact required - not a security risk (they're already fetching a token), and necessary for stable TOTP authentication.

johnnybubonic:
@cipherboy
> I'm also not fully sure that `expiry` is counted correctly, but I still have to dig through the implementation.

Expiration will always occur on the _next_ **time step**, not `<validity>+<generation>`, if it helps.
(Vault made the latter mistake in their web UI.)

This is why when you have a bunch of different 30s-period TOTPs, they all rollover at the same time, not at different times - any two TOTPs using the same validity period will **always** have the exact same time steps. (Assuming they're using UNIX epoch as their T0, though I don't think I've ever seen any that don't.)

builtin/logical/totp/path_code.go
```diff
@@ -77,10 +90,18 @@ func (b *backend) pathReadCode(ctx context.Context, req *logical.Request, data *
...
+	period := time.Duration(key.Period) * time.Second
+
+	// calculate time step
+	ts := time.Duration(tNow.Unix()/int64(key.Period)) * period
 	return &logical.Response{
 		Data: map[string]interface{}{
-			"code": totpToken,
+			"code":      totpToken,
+			"generated": NanoEpoch(tNow),
+			"expiry":    NanoEpoch(time.Unix(int64(ts.Seconds()), 0).Add(period)),
```
cipherboy: OK, to get back to nitpicking:

`expiry` and `ttl` are both durational fields in the response; I see now that this is a timestamp.

The closest I can find is in the token store, where `creation_time` and `expire_time` are relevant. Rather than using a nano second epoch (excessive, due to period rounding as I mentioned above), lets just use the `expire_time` format.

cipherboy: Alternatively, `"expiration"` is what is used in PKI with regular (non-nano) epoch. :shrug:

builtin/logical/totp/backend_test.go
wslabosz-reply: > With mocked time, it might be possible without sleeps, though that gets perhaps harder to inject back into the server.

synctest for the rescue

I've added something like that, lmk if it makes sense

builtin/logical/totp/path_code.go
wslabosz-reply: adjusted

PR #2586, opened 2026-03-05
Add Auto Unseal plugins

satoqz:
## Description

This implements pluginized Auto Unseals as proposed in https://github.com/openbao/openbao/pull/2459.

Dependencies:
- https://github.com/openbao/go-kms-wrapping/pull/71 for the plugin server SDK
- https://github.com/openbao/openbao-plugins/pull/55 to build compatible plugin server executables

## Quick Tutorial

1. Check out my branch at https://github.com/openbao/openbao-plugins/pull/55 to build a KMS plugin binary. For example, for PKCS#11:
   ```sh
   $ make kms-pkcs11
   building bin/openbao-plugin-kms-pkcs11_linux_amd64_v1
   ```

2. Create a configuration file with matching `plugin` and `seal` stanzas:
   ```hcl
   # Place the binary you built above in here.
   plugin_directory = "plugins"

   plugin "kms" "pkcs11" {
     # Whatever the binary you copied from openbao-plugins is named.
     command = "openbao-plugin-kms-pkcs11_linux_amd64_v1"
     # Get the value here via `sha256sum < plugins/openbao-plugin-kms-pkcs11_linux_amd64_v1`.
     sha256sum = "..."
   }

   seal "pkcs11" {
     # Depending on your HSM, see https://openbao.org/docs/configuration/seal/pkcs11.
     # This works as before.
     lib = "/usr/lib/softhsm/libsofthsm2.so"
     pin = "1234"
     token_label = "openbao"
     key_label = "aes-root"
   }
   ```

   Substitute `"pkcs11"` with the plugin you've built in step 1, e.g., `"awskms"`. The plugin name must match the seal type.

3. Build a `bao` binary from this branch and run the server as usual:

   ```sh
   $ go run . server -config=config.hcl
   ...
   Seal Metadata: pkcs11 (key_id: "", key_label: "aes-root", lib: "/usr/lib/softhsm/libsofthsm2.so", token_label: "openbao")
   ...
   2026-03-05T16:02:04.557+0100 [INFO]  kmsplugin.pkcs11: configuring client automatic mTLS
   2026-03-05T16:02:04.578+0100 [INFO]  kmsplugin.pkcs11.openbao-plugin-kms-pkcs11_linux_amd64_v1: configuring server automatic mTLS: timestamp="2026-03-05T16:02:04.578+0100"
   ```

4. You can also try killing the plugin process, and watch for it to respawn and recover state automatically when the auto seal healthcheck runs (every 10 minutes).

## Acknowledgements

       also allowed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,8 +23,6 @@ replace github.com/openbao/openbao/sdk/v2 => ./sdk
 
 replace github.com/boltdb/bolt => ./helper/stubbolt
 
-replace github.com/openbao/go-kms-wrapping/plugin/v2 => github.com/Ki-Reply-GmbH/go-kms-wrapping/plugin/v2 v2.1.0-development20240408.0.20260309151306-c1801988409b
-
 require (
 	cloud.google.com/go/monitoring v1.24.3
 	github.com/ProtonMail/go-crypto v1.4.0
@@ -102,7 +100,7 @@ require (
 	github.com/moby/moby/client v0.2.2
 	github.com/oklog/run v1.2.0
 	github.com/okta/okta-sdk-golang/v2 v2.20.0
-	github.com/openbao/go-kms-wrapping/plugin/v2 v2.0.0-00010101000000-000000000000
+	github.com/openbao/go-kms-wrapping/plugin/v2 v2.2.1-0.20260318145607-5d6196b11a58
 	github.com/openbao/go-kms-wrapping/v2 v2.7.0
 	github.com/openbao/go-kms-wrapping/wrappers/alicloudkms/v2 v2.2.0
 	github.com/openbao/go-kms-wrapping/wrappers/awskms/v2 v2.3.0
--- a/go.sum
+++ b/go.sum
@@ -76,8 +76,6 @@ github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2
 github.com/DataDog/datadog-go v2.2.0+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
 github.com/DataDog/datadog-go v3.2.0+incompatible h1:qSG2N4FghB1He/r2mFrWKCaL7dXCilEuNEeAn20fdD4=
 github.com/DataDog/datadog-go v3.2.0+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
-github.com/Ki-Reply-GmbH/go-kms-wrapping/plugin/v2 v2.1.0-development20240408.0.20260309151306-c1801988409b h1:J/QzV2elZddOyqqykrn6UFWr8kQ9mycSJ2XMr6V2W5w=
-github.com/Ki-Reply-GmbH/go-kms-wrapping/plugin/v2 v2.1.0-development20240408.0.20260309151306-c1801988409b/go.mod h1:E9m3iOkTDWYqru2ueyrP35cTPqZ10dKlhEoclk3CjaQ=
 github.com/Masterminds/goutils v1.1.1 h1:5nUrii3FMTL5diU80unEVvNevw1nH4+ZV4DSLVJLSYI=
 github.com/Masterminds/goutils v1.1.1/go.mod h1:8cTjp+g8YejhMuvIA5y2vz3BpJxksy863GQaJW2MFNU=
 github.com/Masterminds/semver/v3 v3.2.0/go.mod h1:qvl/7zhW3nngYb5+80sSMF+FG2BjYrf8m9wsX0PNOMQ=
@@ -699,6 +697,8 @@ github.com/onsi/ginkgo/v2 v2.27.2 h1:LzwLj0b89qtIy6SSASkzlNvX6WktqurSHwkk2ipF/Ns
 github.com/onsi/ginkgo/v2 v2.27.2/go.mod h1:ArE1D/XhNXBXCBkKOLkbsb2c81dQHCRcF5zwn/ykDRo=
 github.com/onsi/gomega v1.38.2 h1:eZCjf2xjZAqe+LeWvKb5weQ+NcPwX84kqJ0cZNxok2A=
 github.com/onsi/gomega v1.38.2/go.mod h1:W2MJcYxRGV63b418Ai34Ud0hEdTVXq9NW9+Sx6uXf3k=
+github.com/openbao/go-kms-wrapping/plugin/v2 v2.2.1-0.20260318145607-5d6196b11a58 h1:b8HxBAhwmf2AlSdCuKpwkuRXsa+KsuyaJ59RZJDMEIY=
+github.com/openbao/go-kms-wrapping/plugin/v2 v2.2.1-0.20260318145607-5d6196b11a58/go.mod h1:E9m3iOkTDWYqru2ueyrP35cTPqZ10dKlhEoclk3CjaQ=
 github.com/openbao/go-kms-wrapping/v2 v2.7.0 h1:/Vm3UF0bMLdeU2U1fXI0hLVuJbqBlqpAwXlmn88IJ2Q=
 github.com/openbao/go-kms-wrapping/v2 v2.7.0/go.mod h1:xoUdq1btm4MNgwBpMkVLvXxuvnH4gGzODozWGFpft8w=
 github.com/openbao/go-kms-wrapping/wrappers/alicloudkms/v2 v2.2.0 h1:evwGV3ojzXhno2AnCYzyj10HxGHujq6R93zSlaGRRNk=


changelog/2586.txt
```diff
@@ -0,0 +1,16 @@
+```release-note:feature
+**Auto Unseal plugins**: Add a new `kms` plugin type that enables pluggable Auto
+Unseal mechanisms.
+  - Declaratively register KMS plugins using the `plugin` configuration keyword.
```
wslabosz-reply: not a stanza?

satoqz: This is wording copied from the original declarative plugins changelog but we can do whichever.

hb9cwp:
Nice work. Will the "Quick Tutorial" above make it into `go-kms-wrapping/examples/plugin-cli/README.md`, or similar? Thank you.

satoqz:
@hb9cwp go-kms-wrapping is mostly just a generic Go package (that tries to remain as decoupled from OpenBao as possible) so all concerns local to that repo are really documented in Go doc already.

That said, of course there will be user-facing documentation at https://openbao.org on this change before it is released. I plan to author that once we're happy with the feature as implemented here :)

command/server/config.go
```diff
@@ -354,13 +345,29 @@ func (p *PluginConfig) Validate(sourceFilePath string) []configutil.ConfigError
...
 			Problem: fmt.Sprintf("plugin %q: %s", p.Name, err.Error()),
 		})
 	}
 
+	switch typ {
+	case consts.PluginTypeKMS:
+		// KMS plugins are unversioned.
+		if p.Version != "" {
+			results = append(results, configutil.ConfigError{
+				Problem: fmt.Sprintf("plugin %q: version of KMS plugins must be unset", p.Slug()),
```
JanMa: What's the reason behind this? The download of plugin images doesn't work if you're not allowed to define a version.

helper/kmsplugin/catalog_test.go
```diff
@@ -0,0 +1,175 @@
...
+	gkwplugin "github.com/openbao/go-kms-wrapping/plugin/v2"
+	wrapping "github.com/openbao/go-kms-wrapping/v2"
+	"github.com/openbao/go-kms-wrapping/wrappers/static/v2"
+	"github.com/openbao/openbao/command/server"
+	"github.com/openbao/openbao/helper/osutil"
+	"github.com/openbao/openbao/sdk/v2/helper/consts"
+	"github.com/stretchr/testify/require"
+)
+
+const TestPluginServerEnv = "OPENBAO_TEST_PLUGIN_SERVER"
```
JanMa:
```suggestion
const TestPluginServerEnv = "BAO_TEST_PLUGIN_SERVER"
```

Env vars are prefixed with `BAO_`

vault/plugin_catalog.go
```diff
@@ -205,67 +215,36 @@ func (c *Core) ReloadPlugins() {
...
-	}
+	c.pluginCatalog.lock.Lock()
+	defer c.pluginCatalog.lock.Unlock()
 
-	// Create OCI plugin downloader
-	logger.Info("starting OCI plugin downloading")
-	downloader := oci.NewPluginDownloader(c.pluginDirectory, config, logger)
-	err := downloader.ReconcilePlugins(ctx)
-	if err != nil {
-		return err
```
JanMa, line 249: Not sure if I'm missing something, but where are plugins downloaded now during startup? It seems to me that this logic is completely removed from startup and you have to run `bao plugin init` if you want OCI based plugins to be downloaded

JanMa (changes requested):
I have some open questions :-)

vault/plugin_catalog.go
JanMa: Nevermind, I missed something. Automated downloading and registering of plugins still works

helper/kmsplugin/catalog_test.go
satoqz: Fixed

command/server/config.go
satoqz: Ah, this is me simplifying https://github.com/openbao/openbao/pull/2459#discussion_r2793709871 too much, and great catch. I'll overthink this one a bit and come up with better handling :)

satoqz: @JanMa fixed in f13fd730d6aa89e2991fd09c7cdd41006905de0a, plus tested against an actual container registry this time.

JanMa (approved):
LGTM, great work @satoqz

command/server.go
```diff
@@ -1135,14 +1148,7 @@ func (c *ServerCommand) Run(args []string) int {
...
 
-	// prepare a secure random reader for core
-	secureRandomReader, err := configutil.CreateSecureRandomReaderFunc(config.SharedConfig, barrierWrapper)
-	if err != nil {
-		c.UI.Error(err.Error())
-		return 1
-	}
-
-	coreConfig := createCoreConfig(c, config, backend, configSR, barrierSeal, unwrapSeal, metricsHelper, metricSink, secureRandomReader)
+	coreConfig := createCoreConfig(c, config, backend, configSR, barrierSeal, unwrapSeal, metricsHelper, metricSink, rand.Reader)
```
DanGhita, line 1138: I might be wrong, but I think the initial intention of the `secureRandomReader` function  was to be able  to provide a more reliable random, typically provided by a KMS/HSM. Sure, it seems that it was an unfinished staff since the current implementation just calls `rand.Reader `, but is this a reason to get rid of `secureRandomReader` ? Maybe one day,  our new kms interface could provide True Random  for `secureRandomReader`. In any case, having a good source of random for sensitive operations is really important.

command/server.go
```diff
@@ -2472,42 +2492,48 @@ func setSeal(c *ServerCommand, config *server.Config, infoKeys *[]string, info m
...
+				//nolint:staticcheck // User-facing error.
+				return nil, nil, nil, nil, nil, fmt.Errorf("Error creating auto seal: %w", err)
 			}
+
+			infoKey := "seal metadata"
+			if configSeal.Disabled {
+				infoKey = "old seal metadata"
+			}
+
+			// Auto seals emit metadata that we can print at server startup.
```
DanGhita: Just to be sure:  I suppose this is not about printing potentially sensitive information (as an HSM PIN, for ex.) ?

satoqz: This mirrors the output of `Wrapper.SetConfig`, which conventionally contains non-sensitive metadata only. It would be a bug in the wrapper implementation to return secrets from that API.

To be fair, we previously had full control over the fields forwarded because all possible wrappers were known statically at compile time. This doesn't scale so I've refactored it to output the entire returned metadata in a structured format. I've cross-checked implementations go-kms-wrapping and haven't seen anything concerning.

command/server.go
satoqz: Yeah, I'm not against having this feature in the future but setting up the randomness source would need to be refactored either way to use the new `helper/kmsplugin` package, so for now it would be dead-end baggage.
We can easily add the tiny amount of plumbing I've removed here back if we choose to ultimately implement this in a similar way as hinted at by the pre-fork leftovers.

command/server.go
DanGhita: Well, if we are talking about a HSM PIN code (or any other credentials for authenticating against a third party KMS/HSM), IMHO those are sensitive information. They are certainly necessary in the OpenBao startup configuration, but still we should try to reduce their exposition (some users could implement an OpenBao bootstrapping process where those credentials are provided on the fly,  in a more "secure" way than just a configuration file stored on the disk) . In light of these considerations, I was just wondering if  it's a good idea to let OpenBao display this kind of information to the console.

satoqz: As I said, there is no sensitive information returned from `SetConfig`, including no PINs or credentials or anything alike. It is not a 1:1 echo of the incoming config, rather, it is metadata generated from applying the incoming (potentially sensitive) config.

changelog/2586.txt
```diff
@@ -0,0 +1,17 @@
...
+**Auto Unseal plugins**: Add a new `kms` plugin type that enables pluggable Auto
+Unseal mechanisms.
+  - Declaratively register KMS plugins using `plugin "kms" "<name>" { ... }`
+    configuration stanzas, making `<name>` available as `seal "<name>" { ... }`.
+  - Automatically restart and recover from crashes, avoiding a full instance
+    restart when a seal reaches a bad state (e.g., via a misbehaving PKCS#11
+    library).
+  - Pre-built plugins for many of the seals currently built into OpenBao are
+    available at https://github.com/openbao/openbao-plugins. We expect to phase
+    many of the built-in seals out of the `bao` binary by v2.7; the v2.6 series
```
cipherboy:
```suggestion
    many of the built-in, proprietary seals out of the `bao` binary by v2.7; the v2.6 series
```

command/server.go
```diff
@@ -471,29 +481,33 @@ func (c *ServerCommand) runRecoveryMode() int {
...
 		}
+
+		props := make([]string, 0, len(config.Metadata))
+		for k, v := range config.Metadata {
+			props = append(props, fmt.Sprintf("%s: %q", k, v))
+		}
+		sort.Strings(props)
+
+		info["seal metadata"] = fmt.Sprintf("%s (%s)", configSeal.Type, strings.Join(props, ", "))
+		infoKeys = append(infoKeys, "seal metadata")
```
cipherboy: Just for visibility, if this information is easy enough to extract, could we also record seal source? `builtin` vs `external`?

helper/kmsplugin/wrapper.go
```diff
@@ -0,0 +1,283 @@
...
+// getWrapper returns a new wrapping.Wrapper that is either builtin or
+// pluginized, in which case a new plugin process may be spawned.
+func (c *Catalog) getWrapper(name string) (wrapping.Wrapper, error) {
+	client, ok, err := c.getClient(name)
+	switch {
+	case err != nil:
+		return nil, err
+	case !ok:
+		// Try builtin wrappers.
+		if factory, ok := builtinWrappers[wrapping.WrapperType(name)]; ok {
```
cipherboy, line 101: (lazy reviewer) It looks like we prefer external plugins over builtin plugins; does this mean we also allow registering external plugins with the same name as builtin plugins?

changelog/2586.txt
cipherboy: In particular, I'd suggest we keep things like `transit` and `static` as built-in plugins, though there' a technical argument for why `pkcs11` should be moved externally.

command/server.go
satoqz: Good idea, yeah that should be doable

changelog/2586.txt
satoqz: Mhh. I guess the wording here will depend on when we'd like to move PKCS#11 out?

helper/kmsplugin/wrapper.go
satoqz: Yes. As far as I see, existing plugins work the same way? Though the main idea is that in 2.6, it should be easy to switch back and forth between external and builtin seals, that is, by just removing/adding the `plugin "kms"` stanza.

changelog/2586.txt
satoqz: I think "vendor-specific" is probably better than "proprietary", the latter sounds a bit too much as though we were linking proprietary code or so before.

helper/kmsplugin/wrapper.go
cipherboy: That's fair, works for me.

command/server.go
satoqz: Added a `builtin: true/false` key.

satoqz:
@cipherboy adjusted, just in case you've missed it: There's a tutorial on plugin acquisition in the PR description!

satoqz:
@cipherboy that's a race condition around how soon the context here gets canceled:

https://github.com/openbao/go-kms-wrapping/blob/5d6196b11a58817c866cd99bffee3f862d90054a/plugin/client.go#L31-L33

So sometimes, when calling right after the plugin crashes, we won't get `ErrPluginShutdown` yet but a bare gRPC error, and won't retry that. Perhaps we can synchronize that better though backend plugins follow the same pattern.

satoqz:
Given the large amount of approvals I'll go ahead and merge, plus start working on documentation. There remains plenty of time before a v2.6 release, so any additional feedback to do more ironing out if needed can be discussed in issues & follow-up PRs.

Thanks everyone!

Issue #2540, opened 2026-02-25
Openbao v2.5.0 alpine docker image cannot use /dev/stdout as logging


wsellitti:
I added audit logging to /dev/stdout in my config.

On restarting the container and trying to unseal I get this in the logs:
```
app-1    | 2026-02-25T02:21:41.368Z [ERROR] core: post-unseal setup failed: error="failed to create new audit device to-stdout/: sanity check failed; unable to open \"/dev/stdout\" for writing: open /dev/stdout: permission denied"
```

I'm using image `openbao/openbao@sha256:d24890f589bc57658a43c18faab2afc4fbedaadad3806d5a638e1adaed2dff8d`

Config
```
{
    "ui": true,
    "api_addr": "https://my.domainname.here/",
    "storage": {
        "file": {
            "path": "/openbao/file"
        }
    },
    "listener": [
        {
            "tcp": {
                "address": "0.0.0.0:8200",
                "tls_disable": true
            }
        }
    ],
    "audit": [
        {
            "file": {
                "to-stdout": {
                    "description": "This audit device should never fail.",
                    "options": {
                        "file_path": "/dev/stdout",
                        "log_raw": "true"
                    }
                }
            }
        }
    ]
}
```

Relevant compose service:
```
  app:
    hostname: vault
    image: openbao/openbao:2.5.0
    restart: always
    networks:
      - web
    cap_add:
      - IPC_LOCK
    volumes:
      - app-data:/openbao/file
      - app-logs:/openbao/logs
      - ./bao.json:/openbao/config/bao.json
    command: [server]
```

I'm not sure if this is related to the use of su-exec in the alpine docker image entrypoint script. See: https://github.com/ncopa/su-exec/issues/39

satoqz:
@wsellitti I would also conclude that `su-exec` is the problem. You can avoid calling it by setting

```yaml
user: openbao
```

in your compose file.

Sidenote, I really want to remove shell scripts and the shell generally from our containers, and lift most custom startup logic in the entrypoint script into OpenBao itself... We should have distroless containers.

wsellitti:
Neat.

Looking through that script/the docs it looks like those startup scripts do some initial setup that only needs to run the first time. If I was going to move this compose stack to a new host I’d want to remove the `user: openbao` line right? Then I could re-add it and down and up the stack.

satoqz:
[...]

PR #2589, opened 2026-03-06
Clean up Dockerfile, don't default to root on Alpine

satoqz:
## Description

This cleans up our Dockerfile by removing redundant code and correcting some of the comments.

Additionally, this sets a default non-root user for the Alpine image, like is already the case for the UBI image. I believe it is better to keep this consistent, even if minimally breaking, plus it resolves https://github.com/openbao/openbao/issues/2540 in the common case.

Nightly run for testing: https://github.com/openbao/openbao-nightly/actions/runs/22761224535

## Rationale

As a followup, I'd like to propose a new container image variant, `openbao-minimal`, that is distroless. This does some upfront housekeeping before I get to that.

## Acknowledgements

       also allowed.

### DIFF
--- a/goreleaser.hsm.yaml
+++ b/goreleaser.hsm.yaml
@@ -17,7 +17,7 @@ builds:
       - ui
       - hsm
     ldflags:
-      - -X github.com/openbao/openbao/version.fullVersion={{.Version}} -X github.com/openbao/openbao/version.GitCommit={{.Commit}} -X github.com/openbao/openbao/version.CommitDate={{ .Date }} -X github.com/openbao/openbao/version.VersionMetadata=hsm
+      - -X github.com/openbao/openbao/version.fullVersion={{ .Version }} -X github.com/openbao/openbao/version.GitCommit={{ .Commit }} -X github.com/openbao/openbao/version.CommitDate={{ .Date }} -X github.com/openbao/openbao/version.VersionMetadata=hsm
     env:
       - CGO_ENABLED=1
       - >-
@@ -130,7 +130,7 @@ dockers:
       - builds-hsm
     build_flag_templates:
       - "--pull"
-      - "--build-arg=BIN_NAME={{ .ProjectName}}"
+      - "--build-arg=BIN_NAME={{ .ProjectName }}"
       - "--build-arg=EXTRA_PACKAGES=gcompat"
       - "--platform=linux/amd64"
       - '--label=org.opencontainers.image.created={{ time "2006-01-02T15:04:05Z07:00" }}'
--- a/goreleaser.linux.yaml
+++ b/goreleaser.linux.yaml
@@ -16,7 +16,7 @@ builds:
     tags:
       - ui
     ldflags:
-      - -X github.com/openbao/openbao/version.fullVersion={{.Version}} -X github.com/openbao/openbao/version.GitCommit={{.Commit}} -X github.com/openbao/openbao/version.CommitDate={{ .Date }}
+      - -X github.com/openbao/openbao/version.fullVersion={{ .Version }} -X github.com/openbao/openbao/version.GitCommit={{ .Commit }} -X github.com/openbao/openbao/version.CommitDate={{ .Date }}
     env:
       - CGO_ENABLED=0
     goos:
@@ -215,7 +215,7 @@ dockers:
       - builds-linux
     build_flag_templates:
       - "--pull"
-      - "--build-arg=BIN_NAME={{.ProjectName}}"
+      - "--build-arg=BIN_NAME={{ .ProjectName }}"
       - "--platform=linux/ppc64le"
       - '--label=org.opencontainers.image.created={{ time "2006-01-02T15:04:05Z07:00" }}'
       - "--label=org.opencontainers.image.authors=OpenBao <openbao@lists.openssf.org>"


satoqz:
Also cc @openbao/openbao-k8s-committers, any concerns?

changelog/2589.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+packaging/container: The openbao & openbao-hsm container images now run under the openbao user rather than the root user by default. This matches the existing behavior of openbao-ubi variants.
```
cipherboy: I'm not against this, but fwiw, the historical distinction came from UBI being aligned with Red Hat's OpenShift Certification Requirements, which the alpine image wasn't required to/couldn't undergo (as, they only certified RHEL UBI based images).

That's why a lot of the weird patterns you see in the UBI image are the way they are.

Probably worth getting broader community feedback on this, IMHO. I'll add it to this week's community call.

satoqz: Regardless of the mentioned background I think the UBI container was following best practices by setting a default non-root user while the Alpine container wasn't. It is a minor breaking change though that can easily be accounted for downstream by manually setting the container user to root.

phyrog: Maybe add that to the changelog ("If you are relying on the old behavior, here is what you can do: ...")

satoqz: Adjusted the changelog to give more detail.

satoqz: ...if only I did that on the right branch.

changelog/2589.txt
```diff
@@ -0,0 +1,12 @@
+```release-note:improvement
```
cipherboy:
```suggestion
```release-note:change
```

Just to keep the right formatting

Dockerfile
```diff
@@ -1,52 +1,46 @@
...
       maintainer="OpenBao <openbao@lists.openssf.org>" \
       vendor="OpenBao" \
-      version=${PRODUCT_VERSION} \
-      release=${PRODUCT_REVISION} \
-      revision=${PRODUCT_REVISION} \
       summary="OpenBao is a tool for securely accessing secrets." \
-      description="OpenBao is a tool for securely accessing secrets. A secret is anything that you want to tightly control access to, such as API keys, passwords, certificates, and more. OpenBao provides a unified interface to any secret, while providing tight access control and recording a detailed audit log."
+      description="OpenBao is a tool for securely accessing secrets. A secret is anything that you want to tightly control access to, such as API keys, passwords, certificates, and more. OpenBao provides a unified interface to any secret, while providing tight access control and recording a detailed audit log." \
+      version=${VERSION} \
+      release=${VERSION} \
```
cipherboy: @satoqz Am I misreading this or is `release` now swapped between `version` and `revision`?

satoqz: Ah good catch. On second thought, do we even want to keep the labels here? We already set the following in Goreleaser:

```yaml
- "--label=org.opencontainers.image.title={{ .ProjectName }}"
- "--label=org.opencontainers.image.vendor=OpenBao"
- "--label=org.opencontainers.image.description=OpenBao {{ .Version }}"
- "--label=org.opencontainers.image.url=https://github.com/openbao/openbao"
- "--label=org.opencontainers.image.documentation=https://github.com/openbao/openbao/blob/main/README.md"
- "--label=org.opencontainers.image.source=https://github.com/openbao/openbao"
- "--label=org.opencontainers.image.licenses=MPL-2.0"
- "--label=org.opencontainers.image.version={{ .Version }}"
- "--label=org.opencontainers.image.revision={{ .FullCommit }}"
- '--label=org.opencontainers.image.created={{ time "2006-01-02T15:04:05Z07:00" }}'
```

cipherboy: Are these fully qualified later by something or bare labels? Seems like moving them to goreleaser is the right call either way.

satoqz:
```shell-session
$ docker inspect openbao/openbao | jq '.[0].Config.Labels'
{
  "description": "OpenBao is a tool for securely accessing secrets. A secret is anything that you want to tightly control access to, such as API keys, passwords, certificates, and more. OpenBao provides a unified interface to any secret, while providing tight access control and recording a detailed audit log.",
  "maintainer": "OpenBao <openbao@lists.openssf.org>",
  "name": "OpenBao",
  "org.opencontainers.image.created": "2026-02-23T17:49:22Z",
  "org.opencontainers.image.description": "OpenBao 2.5.1",
  "org.opencontainers.image.documentation": "https://github.com/openbao/openbao/blob/main/README.md",
  "org.opencontainers.image.licenses": "MPL-2.0",
  "org.opencontainers.image.revision": "e546fae8cbfe95d8f36a351deb2cd23bfb94119e",
  "org.opencontainers.image.source": "https://github.com/openbao/openbao",
  "org.opencontainers.image.title": "bao",
  "org.opencontainers.image.url": "https://github.com/openbao/openbao",
  "org.opencontainers.image.vendor": "OpenBao",
  "org.opencontainers.image.version": "2.5.1",
  "release": "2.5.1",
  "revision": "e546fae8cbfe95d8f36a351deb2cd23bfb94119e",
  "summary": "OpenBao is a tool for securely accessing secrets.",
  "vendor": "OpenBao",
  "version": "2.5.1"
}
```

Looks bare to me. I'd suggest we standardize on the opencontainers labels and migrate information that isn't given in those labels yet (such as maintainer email) to the respective standard labels.

satoqz: Ah, this also clarifies why the above revision/version was mixed up. Goreleaser actually set those the way I expected, and was overriding the Dockerfile's label in the image we ultimately release.

satoqz:
@cipherboy adjusted to use labels in https://github.com/opencontainers/image-spec/blob/main/annotations.md only, and rearranged all labels into the same order as given in the spec for easy comparison.

Dockerfile
```diff
@@ -142,14 +99,14 @@ VOLUME /openbao/file
 # OpenBao.
 EXPOSE 8200
 
+# Use the OpenBao user as the default user for starting this container.
+USER openbao
```
eyenx, line 103: Is the user able to execute /bin/bao?

satoqz: Yes. It's the same user that was executing /bin/bao before, we always eventually drop down to this user before starting OpenBao in https://github.com/openbao/openbao/blob/main/.release/docker/docker-entrypoint.sh if the container is started as root.

eyenx: Yes I tried it out in the live container right now.

satoqz:
Rebased to account for minor conflict with https://github.com/openbao/openbao/pull/2678.

PR #2592, opened 2026-03-06
Add minimal, distroless container image variant

satoqz:
## Description

Add a new container image variant, "openbao-distroless" to the existing Alpine-based and UBI-based container images officially maintained by OpenBao. openbao-distroless is based on Google's `distroless/static:nonroot` container images.

This is currently based on #2589.

## Rationale

OpenBao, being a Go application that compiles to a static binary on Linux, is perfectly capable of running in an extremely minimal environment with no conventional OS utilities available. However, container images currently released by the OpenBao project do not make any use of this property and require downstream users to accept that official images come with a full rootfs based on Alpine Linux or UBI at their option.

There are advantages to both "distroful" and "distroless" container packaging:

- distroful: Easy to add extra OS dependencies via package managers, easy to debug via the included shell and tools, helpful entrypoint scripts to take over some automatable confguration steps
- distroless: Reduced size, reduced dependencies, reduced attack surface, near-zero scanner noise caused by unrelated OS dependencies. Updating the base image is rarely required (really just to update ca-certificates) and can safely coincide with regular releases of the main software.

OpenBao is deployed across a variety of different container-based setups with different modifications such as plugin installation, PKCS#11 libraries, additional scripts &c. As a result, I believe it is best to make both distroful and distroless containers available.

Notably, there is no "distroless-hsm" variant. The reasoning behind its omission is two-fold:

1. I seek to entirely remove the HSM distribution of OpenBao in the future, moving CGO-based PKCS#11 functionality into plugin binaries via https://github.com/openbao/openbao/pull/2459.
2. I expect advanced users that set up PKCS#11 libraries with a distroless image to understand how to manually add the required runtime libraries to a distroless image, including some version of libc.

As a security-focused project, I believe it is a disservice to users not to offer a container image distribution that is optimized to have the lowest possible attack surface (There are further improvements we can make here, such as reproducible builds, but that is an undertaking for another day).

I also imagine https://github.com/openbao/openbao/security/advisories/GHSA-xp75-r577-cvhp would have been much more difficult to exploit on a system that does not have additional utilities available.

If we'd prefer an RFC on the topic, I can do that, however I believe this is fairly non-controversial and follows a pattern common in server-type open source software.

#### Why use Google's distroless images?

They're free, widely adopted & have dependency scanning metadata for the little included packages that they do have (by proxy of being based on Debian). Additionally, unlike distroful images, it is much easier to switch to a separate source of distroless base images (or to build our own) without breaking users.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2592.txt
+++ b/changelog/2592.txt
@@ -1,3 +1,6 @@
 ```release-note:feature
-packaging/container: Add new "openbao-distroless" container image variant based on [distroless/static](https://github.com/GoogleContainerTools/distroless). The only executable contained in these images is OpenBao itself.
+Add a **Distroless container image**:
+  - Based on Google's [distroless/static](https://github.com/GoogleContainerTools/distroless).
+  - Available as `openbao/openbao-distroless` on docker.io, ghcr.io and quay.io.
+  - The only executable in the image is OpenBao itself, limiting attack surface & dependencies.
 ```
--- a/changelog/2592.txt
+++ b/changelog/2592.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+packaging/container: Add new "openbao-distroless" container image variant based on [distroless/static](https://github.com/GoogleContainerTools/distroless). The only executable contained in these images is OpenBao itself.
+```
--- a/changelog/2592.txt
+++ b/changelog/2592.txt
@@ -1,6 +1,4 @@
 ```release-note:feature
-Add a **Distroless container image**:
-  - Based on Google's [distroless/static](https://github.com/GoogleContainerTools/distroless).
-  - Available as `openbao/openbao-distroless` on docker.io, ghcr.io and quay.io.
-  - The only executable in the image is OpenBao itself, limiting attack surface & dependencies.
+Add new **openbao-distroless** container image variant based on [distroless/static](https://github.com/GoogleContainerTools/distroless).
+  - The only executable contained in these images is OpenBao itself.
 ```
--- a/website/src/pages/downloads.tsx
+++ b/website/src/pages/downloads.tsx
@@ -299,7 +299,7 @@ const PackageRepo = ({ type }) => {
 const DockerList = ({ version, registry }) => {
   const dockerVersion = version.slice(1);
 
-  // For example: v2.6.0, v2.5.0-beta20251125.
+  // For example: 2.6.0, 2.5.0-beta20251125.
   // Consider replacing with a more complete semver parser once required.
   const [major, minor, _patch] = dockerVersion
     .match(/^(\d+)\.(\d+)\.(\d+)/).slice(1).map(v => parseInt(v));
--- a/website/src/pages/downloads.tsx
+++ b/website/src/pages/downloads.tsx
@@ -297,12 +297,19 @@ const PackageRepo = ({ type }) => {
   );
 };
 const DockerList = ({ version, registry }) => {
-  const dockerVersion = version.slice(1);
+  // For example: v2.6.0, v2.5.0-beta20251125.
+  // Consider replacing with a more complete semver parser once required.
+  const [major, minor, _patch] = version
+    .match(/^v(\d+)\.(\d+)\.(\d+)/).slice(1).map(v => parseInt(v));
+
   const dockerDistros = {
     "Alpine Image Distribution": "openbao/openbao",
     "Alpine Image Distribution with HSM Support": "openbao/openbao-hsm",
     "Red Hat Universal Base Image (UBI) Distribution": "openbao/openbao-ubi",
     "Red Hat Universal Base Image (UBI) Distribution with HSM support": "openbao/openbao-hsm-ubi",
+    ...(major >= 2 && minor >= 6 ? {
+      "Distroless Distribution": "openbao/openbao-distroless",
+    } : {}),
   }
   return (
       <>
--- a/website/src/pages/downloads.tsx
+++ b/website/src/pages/downloads.tsx
@@ -297,10 +297,12 @@ const PackageRepo = ({ type }) => {
   );
 };
 const DockerList = ({ version, registry }) => {
+  const dockerVersion = version.slice(1);
+
   // For example: v2.6.0, v2.5.0-beta20251125.
   // Consider replacing with a more complete semver parser once required.
-  const [major, minor, _patch] = version
-    .match(/^v(\d+)\.(\d+)\.(\d+)/).slice(1).map(v => parseInt(v));
+  const [major, minor, _patch] = dockerVersion
+    .match(/^(\d+)\.(\d+)\.(\d+)/).slice(1).map(v => parseInt(v));
 
   const dockerDistros = {
     "Alpine Image Distribution": "openbao/openbao",
@@ -317,7 +319,7 @@ const DockerList = ({ version, registry }) => {
           <div key={label}>
             <p>{label}</p>
             <CodeBlock language="shell">
-              {`docker pull ${registry}/${image}:${version}`}
+              {`docker pull ${registry}/${image}:${dockerVersion}`}
             </CodeBlock>
           </div>
         ))}
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,8 +59,7 @@ FROM registry.access.redhat.com/ubi10-minimal:10.1 AS ubi
 COPY LICENSE /licenses/mozilla.txt
 
 # Set up ca-certificates & base tooling.
-RUN set -eux; \
-    microdnf install -y ca-certificates gnupg openssl libcap tzdata procps shadow-utils util-linux
+RUN microdnf install -y ca-certificates gnupg openssl libcap tzdata procps shadow-utils util-linux
 
 # Create a non-root user to run the software.
 RUN groupadd --gid 1000 openbao && \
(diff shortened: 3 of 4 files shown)


phyrog (approved):

phyrog:
What do you think about using tags, instead of separate image repositories, like you see in other images?

Example: `golang:1.26.1-alpine`, `golang:1.26.1-trixie`, ...

satoqz:
@phyrog I would be in favor though it would imply an even greater breaking change as we'd want to collapse other existing repositories into one as well to avoid complete inconsistency. @cipherboy let's ask about this on the community call as well perhaps?

phyrog:
We could start with publishing both via tags and the existing repositories and at some later minor release stop updating the separate repositories + a release note. But yeah let's talk in the community call.

Updating the Helm chart and waiting for a while before stopping support for the "old" image repositories should probably catch most users of the image.

cipherboy:
@phyrog @satoqz With the distroless container, do we care as much about maintaining both Alpine vs UBI any more? It seems like three variants is a bit excessive; one of RHEL or Alpine is probably sufficient?

phyrog:
I like the pattern that some other projects follow with one minimal image variant and another debug image that works the same as minimal (same user in the container, etc), just with some more debugging tools.

In case of a problem you could just switch the image to the debug variant and debug away without having to worry about other image differences.

So with distroless being based on debian, maybe distroless + debian (+ ubi? not sure how important it is to keep this)

satoqz:
@cipherboy I think the main "benefits" of RHEL are:

- Glibc for compatibility with proprietary PKCS#11 libraries (Though a more neutral source for this type of image would be Debian)
- Compliance for RHEL shops? You tell me :D

OTOH, Alpine has:

- Image size much closer to the distroless variant
- Musl + glibc-compat which covers some PKCS#11 libraries, but certainly not all

Both of course also have a package manager and the docker-entrypoint.sh script that provides some convenience, which may be a reason not to switch to the distroless image.

So depending on the use case, I think all 3 images have valid use cases:

- Gnarly PKCS#11 library that can't handle musl -> UBI
- Just want an extra shell script you can exec? -> Alpine
- No special needs or expertise to set up C libraries and tooling manually -> Distroless

One (random-ish) inspiring factor for the layout proposed in this PR is the one chosen by https://github.com/vectordotdev/vector in https://hub.docker.com/r/timberio/vector/tags, where the default is Alpine, but both Debian and Distroless images are alternatively available under separate tags.

---

I think we can't easily conclude that dropping any of these doesn't break a common workflow, especially around PKCS#11. I'd suggest we go ahead as-is, then eventually get to removing `*-hsm` variants, and once that's done reconsider what to do with UBI & Alpine?

In terms of storage bloat/network usage, the layer with the OpenBao binary should be shared so I don't think a 3rd variant would explode those.

satoqz:
> I like the pattern that some other projects follow with one minimal image variant and another debug image that works the same as minimal (same user in the container, etc), just with some more debugging tools.
> 
> In case of a problem you could just switch the image to the debug variant and debug away without having to worry about other image differences.
> 
> So with distroless being based on debian, maybe distroless + debian (+ ubi? not sure how important it is to keep this)

The distroless project itself provides debug images with busybox, so perhaps Debian wouldn't be needed. There are also Glibc-based distroless images, so that covers some area as well.

The main complications that make me think it's better to keep distros around for now are:
- Installing additional libraries beyond Glibc that may be necessary for a PKCS#11 library to work
- [`docker-entrypoint.sh`](https://github.com/openbao/openbao/blob/main/.release/docker/docker-entrypoint.sh), which we depend on ourselves in test helpers.

cipherboy:
@satoqz I learned recently there's also multiple levels of UBI containers:

- "ubi" https://catalog.redhat.com/en/software/containers/ubi10/ubi/66f2b46b122803e4937d11ae  (76MB)
- "ubi-minimal" https://catalog.redhat.com/en/software/containers/ubi10/ubi-minimal/66f1504a379b9c2cf23e145c (33MB)
- "ubi-micro" https://catalog.redhat.com/en/software/containers/ubi10/ubi-micro/66f2b273b37c38851ee372f7 (8MB)

So we could likely save significantly on size by moving down the list.

satoqz:
@cipherboy we already use ubi-minimal. Perhaps it is worth stepping down to micro though I fear the average PKCS#11 user will end up reinstalling the removed packages yet again :D

I don't really care as much about base image size tbh, and it's not the point of the distroless images either - Alpine itself is already about 8MB, and that's still minimal overhead compared to the <1MB overhead of distroless when the bao binary is ~200MB (we can get down to 150MB or lower with go-discover & most wrappers removed, btw).

phyrog:
> @cipherboy we already use ubi-minimal. Perhaps it is worth stepping down to micro though I fear the average PKCS#11 user will end up reinstalling the removed packages yet again :D

Note that micro does not contain a package manager, which would make (re)installing the needed packages harder for users

gianklug:
Should we maybe also update https://github.com/openbao/openbao/blob/main/website/src/pages/downloads.tsx#L301 in the same PR?

satoqz:
@gianklug depends on how we'd like to merge this, the downloads page updates live but the images will be out at release time only. Perhaps the easiest path is to merge this shortly before release and include the website changes with it, yeah.

Alternatively we can build some logic into the website to only display the updated images when detecting a version >= 2.6 tbf. I'll have a look at what seems best.

pree:
> Note that micro does not contain a package manager, which would make (re)installing the needed packages harder for users

What packages are you thinking of? Generally the docker images should be scoped to be as minimal as possible and only include things required to run OpenBao (imho). 


---

In general I'm happy to move to distroless, in case this doesn't break current functionality (PKCS#11 being probably to worst offender). UBI might be nice for folks running on OpenShift, especially in environments that are highly regulated. 
Is it possible that plugins could run as their own container? Like having a separate unseal container which handles communication via PKCS#11. I think this could already be possible with existing plugins on Kubernetes via Sidecars.

phyrog:
> What packages are you thinking of?

That was mainly in reference to @satoqz mention of PKCS#11 users. Some HSM libraries seem to require other libraries, e.g. libsqlite3.

In general I agree with you that we should keep the main image minimal, but we should consider how how users can make use of the more advanced features (like PKCS#11 with custom libraries) and document these options (build their own image, mount volume, etc).

satoqz:
> Is it possible that plugins could run as their own container?

IIRC this is a Vault Enterprise feature; something we can look at though certainly a bit out of scope for now :D

---

This has evolved into several separate discussions that we shouldn't mix up too much. To provide a broad overview of proposed container-related changes:

1. Adding a distroless container image _besides_ existing Alpine and UBI images
2. Generally re-evaluating the base images we support
3. Moving from separate registry repositories into a single one, using tags to differentiate between bases
4. Dropping `*-hsm` images once Auto Seal plugins have been around for a while
5. Containerizing plugins

Do we mind focusing on 1. for this PR and moving the others, especially any breaking changes, to separate discussions? I'm in favor of most of the other proposals down the line though the focus of this one is to keep it simple and expand the available options first without assuming too much about our users' existing workflows :)

goreleaser.hsm.yaml
```diff
@@ -130,9 +130,9 @@ dockers:
       - builds-hsm
     build_flag_templates:
       - "--pull"
-      - "--build-arg=BIN_NAME={{.ProjectName}}"
-      - "--build-arg=REVISION={{.FullCommit}}"
-      - "--build-arg=VERSION={{.Version}}"
+      - "--build-arg=BIN_NAME={{ .ProjectName}}"
```
pree:
```suggestion
      - "--build-arg=BIN_NAME={{ .ProjectName }}"
```

pree (commented):
I agree on keeping it simple and just focusing on one thing at a time. But I would love it not being called `-minimal` but rather `-distroless` for the time being? I'm not quite sure about distroless, but with alpine other projects usually supply it via `:alpine` as a tag, so I think specifying the source in here would be better than just a generic `minimal`.

goreleaser.hsm.yaml
satoqz: Fixed at its source in #2589, need to rebase this anyway.

satoqz:
@pree agreed and changed.

satoqz:
> Should we maybe also update https://github.com/openbao/openbao/blob/main/website/src/pages/downloads.tsx#L301 in the same PR?

@gianklug added something for you.

changelog/2592.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+packaging/container: Add new "openbao-distroless" container image variant based on [distroless/static](https://github.com/GoogleContainerTools/distroless). The only executable contained in these images is OpenBao itself.
```
cipherboy:
```suggestion
Add new **openbao-distroless** container image variant based on [distroless/static](https://github.com/GoogleContainerTools/distroless).
  - The only executable contained in these images is OpenBao itself.
```

website/src/pages/downloads.tsx
```diff
@@ -298,19 +298,28 @@ const PackageRepo = ({ type }) => {
 };
 const DockerList = ({ version, registry }) => {
   const dockerVersion = version.slice(1);
+
+  // For example: v2.6.0, v2.5.0-beta20251125.
+  // Consider replacing with a more complete semver parser once required.
+  const [major, minor, _patch] = dockerVersion
+    .match(/^(\d+)\.(\d+)\.(\d+)/).slice(1).map(v => parseInt(v));
```
cipherboy:
```suggestion
  // For example: 2.6.0, 2.5.0-beta20251125.
  // Consider replacing with a more complete semver parser once required.
  const [major, minor, _patch] = dockerVersion
    .match(/^(\d+)\.(\d+)\.(\d+)/).slice(1).map(v => parseInt(v));
```

cipherboy (approved):

changelog/2592.txt
satoqz: So you were first after all... which do we like better, the one I've added in the meantime or your suggestion :D

satoqz: I think I'm going with this one.

satoqz:
I'll defer merging until @cipherboy gets to creating the new repos so this doesn't block nightly releases in the meantime.

cipherboy:
@satoqz Repos are now created!

Blacks-Army:
> I agree on keeping it simple and just focusing on one thing at a time. But I would love it not being called -minimal but rather -distroless for the time being? I'm not quite sure about distroless, but with alpine other projects usually supply it via :alpine as a tag, so I think specifying the source in here would be better than just a generic minimal.

Just a thought: while most projects expose Alpine image versions (and similar variants) via image tags, I think that’s more of a convention than a best practice.

In my opinion, tags should ideally only contain the semantic version (and related version tags), while the image name describes the variant. Something like openbao-distroless:vX.Y.Z therefore seems cleaner and more explicit than encoding the variant in the tag. 

Thankfully, that’s already how it’s done here.

satoqz:
@Blacks-Army good feedback, thanks! I also don't think we have a good argument to change the structure in retrospect. There aren't any problems with it maintenance-wise, either.

Issue #2613, opened 2026-03-10
/sys/rotate/root should not rotate the Shamir KEK


satoqz:
**Describe the bug**

Calling the plain `/sys/rotate/root` endpoint when using a Shamir seal results in both the Shamir KEK and the root key being rotated, without ever re-splitting the KEK and returning the shares back to the user. This effectively leads to loss of all key shares and the inability to unseal in the future.

**To Reproduce**
1. Run an instance using a Shamir Seal (e.g., `bao server -dev -dev-skip-init`.
2. `bao operator init` & `bao operator unseal` if needed.
3. Call `/sys/rotate/root`.
4. `bao operator seal` & attempt to `bao operator unseal` again, keys don't work anymore.

**Expected behavior**

`/sys/rotate/root` should only ever rotate the root key, just like it does correctly for Auto Seals.

**Environment:**

Should be present in any version `>= 2.4.0`.

PR #2619, opened 2026-03-11
Fix root key rotation endpoint rotating Shamir's KEK

wslabosz-reply:
## Description
This PR fixes the `POST /sys/rotate/root` endpoint, which for Shamir seals, not only rotates the root key, but also the "seal" key (unseal key split into shares). This can lead into losing the access to the instance as new unseal key wasn't returned to the end-user.

## Rationale
This was a complete oversight of the implementation of the RFC [documentation](https://openbao.org/docs/rfcs/authenticated-rekey/#post-sysrotateroot) which says: 

> Note that, for Shamir's seals, the sys/rotate/root/init endpoint performs both rotations (the root key and the Shamir shares), conflating concerns, but sys/rotate/root and sys/rotate/root/init endpoint behave the same for auto-unseal mechanisms (minus the difference in when the rotation occurs).

This update now makes the implementation match the described behavior.

### `POST /sys/rotate/root`
(while running shamir seal)

```
$ bao write -force /sys/rotate/root
Success! Data written to: sys/rotate/root

$ bao operator seal
$ bao operator unseal <--- the unseal key won't work
```

for auto seals there are no "unseal keys" just recovery keys, and they are not rotated within this scenario. This PR fixes the accidental unseal key rotation, so after root key rotation user can still unseal the bao instance using unseal shares he's holding. 

Resolves: #2613

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2619.txt
+++ b/changelog/2619.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/seal: Fix `/sys/rotate/root` call rotating both root key and unseal key when using a Shamir Seal, losing all key shares.
+```


changelog/2619.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bugfix
```
satoqz:
```suggestion
```release-note:bug
```

See

https://github.com/hashicorp/go-changelog/blob/e5ec3dbf6c8def8abe18be04f4d71c5ee11f1123/entry.go#L20-L33 for all available types and https://github.com/openbao/openbao/blob/main/changelog/changelog.tmpl for those we handle.

changelog/2619.txt
```diff
@@ -0,0 +1,3 @@
...
+core/seal: Fix root key rotation endpoint rotating both root key and unseal key for Shamir seals.
```
satoqz:
```suggestion
core/seal: Fix root `/sys/rotate/root` rotating both root key and unseal key when using a Shamir seal, rendering all key shares lost.
```

Best to name the specific endpoint.

vault/logical_system_rotate.go
```diff
@@ -506,23 +506,12 @@ func (b *SystemBackend) handleRotateRoot() framework.OperationFunc {
...
 		// Perform the rotation
-		if err := b.Core.performBarrierRekey(ctx, newKey); err != nil {
-			return nil, logical.CodedError(http.StatusInternalServerError, "failed to perform barrier rekey: %v", err)
+		if err := b.Core.RotateBarrierRootKey(ctx); err != nil {
+			return nil, logical.CodedError(http.StatusInternalServerError, "failed to rotate barrier root key: %v", err)
 		}
 
-		// Remove the rotation config
-		b.Core.rootRotationConfig = nil
+		b.Core.logger.Info("barrier root key rotated", "stored", existingConfig.StoredShares, "shares", existingConfig.SecretShares, "threshold", existingConfig.SecretThreshold)
```
satoqz: I don't think logging share count / threshold here makes sense since this endpoint has nothing to do with Shamir seal keys or Shamir recovery keys anymore.

vault/logical_system_rotate.go
```diff
@@ -506,23 +506,12 @@ func (b *SystemBackend) handleRotateRoot() framework.OperationFunc {
 			return handleError(ErrNotInit)
 		}
```
satoqz: Would a `core.Initialized()` check be equivalent?

wslabosz-reply: I think so.

changelog/2619.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+core/seal: Fix `/sys/rotate/root` call rotating both root key and unseal key when using a Shamir seal, losing all key shares.
+```
```
satoqz:
````suggestion
```release-note:bug
core/seal: Fix `/sys/rotate/root` call rotating both root key and unseal key when using a Shamir seal, losing all key shares.
```
```release-note:bug
core/seal: Fix `/sys/rotate/root/{init,update}` returning and persisting a random, unused unseal key when using an Auto Seal.
```
````

Two changelogs is probably clearest.

File: vault/rotate.go [hunk 6]
```diff
@@ -379,10 +376,9 @@ func (c *Core) updateBarrierRotation(ctx context.Context, config *SealConfig, ke
 		}
 	}
 
-	// Generate a new key: for AutoUnseal, this is a new root key; for Shamir,
-	// this is a new unseal key, and performBarrierRekey will also generate a
-	// new root key.
-	newKey, result, err := c.generateKey(c.rootRotationConfig, true)
+	// Generate new unseal/recovery keys;
+	// if we are rotating barrier root key running auto unseal, this is a no-op.
+	newSealKey, result, err := c.generateKey(c.rootRotationConfig, recovery)
```
satoqz: I'd prefer if we didn't even call this function if we don't need new key shares.

vault/rekey.go
```diff
@@ -527,23 +527,24 @@ func (c *Core) performBarrierRekey(ctx context.Context, newSealKey []byte) logic
...
+		return logical.CodedError(http.StatusInternalServerError, "failed to rotate barrier root key: %v", err)
 	}
+
 	if c.logger.IsInfo() {
-		c.logger.Info("security barrier rekeyed", "stored", c.rootRotationConfig.StoredShares, "shares", c.rootRotationConfig.SecretShares, "threshold", c.rootRotationConfig.SecretThreshold)
+		c.logger.Info("barrier root key rotated", "stored", c.rootRotationConfig.StoredShares, "shares", c.rootRotationConfig.SecretShares, "threshold", c.rootRotationConfig.SecretThreshold)
 	}
 
-	if len(newSealKey) > 0 {
+	if c.seal.StoredKeysSupported() != seal.StoredKeysSupportedGeneric {
```
satoqz: Can we do something like

```go
isShamir := c.seal.StoredKeysSupported() == seal.StoredKeysSupportedShamirRoot
```

at the start of the method?

vault/rotate.go
satoqz: I think the same code remains in `rekey.go` covering the original rekey endpoints, do we care to fix those?

wslabosz-reply: I think we can do that.

PR #197, opened 2024-03-10
Remove references to enterprise in vault/

cipherboy:
This removes enterprise stubs related to quotas, licenses, undo logs (a
Vault Enterprise feature related to performance secondaries and Raft),
WALs, authorization policies that only exist in Enterprise, handling
control groups, managed keys, and other nil functions.

---

Based on #195; will be rebased once that merges.

PR #2624, opened 2026-03-12
Remove keyring canary & poison pill

satoqz:
## Description

Removes writing to `core/canary-keyring` and reading from `core/poison-pill`.

## Rationale

These paths are either written and never read or read but never written. I'm confused each time I re-read HA key rotation code.

I do see that @cipherboy kept `core/poison-pill` around in https://github.com/openbao/openbao/pull/197, perhaps we still have a good reason to keep it there but for now I will naively suggest we just remove it :D

## Acknowledgements

       also allowed.

### DIFF
--- a/sdk/physical/cache.go
+++ b/sdk/physical/cache.go
@@ -31,15 +31,8 @@ const (
 // These paths don't need to be cached by the LRU cache. This should
 // particularly help memory pressure when unsealing.
 var cacheExceptionsPaths = []string{
-	"wal/logs/",
-	"index/pages/",
-	"index-dr/pages/",
 	"sys/expire/",
-	"core/poison-pill",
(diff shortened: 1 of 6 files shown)


sdk/physical/cache.go
```diff
@@ -31,15 +31,8 @@ const (
 // These paths don't need to be cached by the LRU cache. This should
 // particularly help memory pressure when unsealing.
 var cacheExceptionsPaths = []string{
-	"wal/logs/",
-	"index/pages/",
-	"index-dr/pages/",
```
satoqz, line 36: IIUC these paths also don't exist.

cipherboy: Yes, I suspect they're a remnant of partial support for Vault Enterprise.

Issue #2627, opened 2026-03-13
Incomplete HTTP Audit device configuration docs


JAYKRISHNAN:
- The [documentation for HTTP audit device configuration](https://openbao.org/docs/audit/http/#configuration) do not mention the `skip_test` option.

- The [skip_test option](https://github.com/openbao/openbao/blob/v2.5.1/vault/audit.go#L139) is implemented in the codebase such that if this configuration value is set as "true", Openbao does NOT do an initial test probe request to the destination of the HTTP to ensure that the audit log destination is healthy and accepting requests

- If `skip_test` value is NOT set as "true", Openbao makes a test probe request. If this fails, the mounting of the HTTP audit device fails and Openbao fails to unseal

- We faced this issue in our Kubernetes deployment setup where the Openbao pods were initialized and up and running before the HTTP audit device destination server was up and running causing the audit device mounting and unsealing in Openbao to fail

- We used the `skip_test` configuration flag as a workaround and it is working perfectly as expected.

- This configuration flag needs to be documented properly in the documentation

JAYKRISHNAN:
Opened a PR to add the missing docs - https://github.com/openbao/openbao/pull/2628

PR #2628, opened 2026-03-13
docs(docs): Add missing documentation on HTTP audit device

JAYKRISHNAN:
## Description

- The [documentation for HTTP audit device configuration](https://openbao.org/docs/audit/http/#configuration) do not mention the skip_test option.

- The [skip_test option](https://github.com/openbao/openbao/blob/v2.5.1/vault/audit.go#L139) is implemented in the codebase such that if this configuration value is set as "true", Openbao does NOT do an initial test probe request to the destination of the HTTP to ensure that the audit log destination is healthy and accepting requests

- If `skip_test` value is NOT set as "true", Openbao makes a test probe request. If this fails, the mounting of the HTTP audit device fails and Openbao fails to unseal

- This PR adds this missing information to the HTTP audit device documentation

## Rationale

- This option is useful in specific cases like the one we faced where in our Kubernetes deployment setup where the Openbao pods were initialized and up and running before the HTTP audit device destination server was up and running causing the audit device mounting and unsealing in Openbao to fail. So we should document this existing useful feature to make docs consistent with the actual code implementation

Resolves: #2627

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/audit/index.mdx
+++ b/website/content/docs/audit/index.mdx
@@ -172,10 +172,10 @@ docs](/api-docs/system/audit) for more details.
   actual log line.
 
 - `skip_test` `(string: "")` When set to `"true"`, OpenBao will not send an
-initial test probe request to the audit log destination when enabling the
-audit device. By default, OpenBao tests the connection to the destination server on
-mount, and if the destination is unreachable, the audit device fails to
-enable which can block unsealing.
+  initial test probe request to the audit log destination when enabling the
+  audit device. By default, OpenBao tests the connection to the destination
+  server on mount, and if the destination is unreachable, the audit device fails
+  to enable which can block unsealing.
 
   :::warning
 


satoqz:
@JAYKRISHNAN looking at `vault/audit.go`, this seems to apply to all audit devices generically - do you mind moving the docs to https://openbao.org/docs/audit/#common-configuration-options?

JAYKRISHNAN:
> @JAYKRISHNAN looking at `vault/audit.go`, this seems to apply to all audit devices generically - do you mind moving the docs to https://openbao.org/docs/audit/#common-configuration-options?

@satoqz That's a good point. I moved it to the docs of generic all audit devices in this commit - https://github.com/openbao/openbao/pull/2628/commits/737ec73265e188fcf49f7eb464aea10eba6dda61

website/content/docs/audit/index.mdx
```diff
@@ -171,6 +171,12 @@ docs](/api-docs/system/audit) for more details.
 - `prefix` `(string: "")` - A customizable string prefix to write before the
   actual log line.
 
+- `skip_test` `(string: "")` When set to `"true"`, OpenBao will not send an
+initial test probe request to the audit log destination when enabling the
+audit device. By default, OpenBao tests the connection to the destination server on
+mount, and if the destination is unreachable, the audit device fails to
+enable which can block unsealing.
```
satoqz:
```suggestion
- `skip_test` `(string: "")` When set to `"true"`, OpenBao will not send an
  initial test probe request to the audit log destination when enabling the
  audit device. By default, OpenBao tests the connection to the destination
  server on mount, and if the destination is unreachable, the audit device fails
  to enable which can block unsealing.
```

to match formatting in other bullet points.

PR #2629, opened 2026-03-13
Remove protoc-go-inject-tag & bump protoc

satoqz:
## Description

Removes `protoc-go-inject-tag` from `Makefile` and `tools/`, plus upgrades protoc to 34.0.

## Rationale

`protoc-go-inject-tag` was used to inject Sentinel-specific Go tags into protoc output. Sentinel does not exist in OpenBao, so we can drop the tags and the tool.

## Acknowledgements

       also allowed.


Makefile
```diff
@@ -255,16 +251,7 @@ influxdb-database-plugin:
 postgresql-database-plugin:
 	@CGO_ENABLED=0 $(GO_CMD) build -o bin/postgresql-database-plugin ./plugins/database/postgresql/postgresql-database-plugin
 
-mssql-database-plugin:
```
wslabosz-reply, line 258: I see you've deleted commands building non-existing database plugins, but left `mysql-legacy-database-plugin` (also included in `.PHONY`)

satoqz: That one exists though AFAICT?

wslabosz-reply: it does, missed it as it's located inside `mysql` dir

Issue #2612, opened 2026-03-09
Race condition between namespace updates and mount deletion


Nerkho:
**Describe the bug**

There appears to be a race condition between namespace updates and mount deletion.
This creates a situation where it's not possible anymore to disable a secrets engine mount until OpenBao is restarted.

**To Reproduce**

While the reproduction steps involve creating a huge amount of namespaces, the issue seems not to be related directly to the number of namespaces. It's just easier to trigger that way.

2000 namespaces were enough to trigger the issue on my kind deployment. Your mileage may vary.
The customer who reported the issue had just ~300 namespaces.

1. OpenBao deployed with a Raft storage backend.
2. Create a huge amount of namespaces with a KV secrets engine in each

```bash
for i in {1..2000} ; do
  name=$(LC_ALL=C tr -dc A-Za-z0-9 </dev/urandom | head -c 64)
  bao namespace create $name
  bao secrets enable -namespace=$name kv
done
```

3. Try to disable one of the KV secrets engines.
```bash
bao secrets disable -ns=ns31 kv
Error disabling secrets engine at kv/: context deadline exceeded
```

In our tests, we were still able to create new namespaces and enable new secrets engines after the race condition triggered. Only deletion was affected.

Additionally, there was nothing pointing to the issue in the logs even with `trace` enabled.

**Expected behavior**

Disabling the secrets engine should still work.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): tested on 2.5.0 and 2.5.1
* OpenBao CLI Version (retrieve with `bao version`): 2.5.1
* Server Operating System/Architecture: Kubernetes and VMs

**Additional context**

It might be in some ways related to https://github.com/openbao/openbao/issues/2555.

phil9909:
Looks like rollback manager got stuck

<img width="1182" height="1775" alt="Image" src="https://github.com/user-attachments/assets/62e11451-d2f6-4bf8-9d85-296c0e284bc8" />

[goroutines.pprof.zip](https://github.com/user-attachments/files/25940084/goroutines.pprof.zip)

phil9909:
How to get this graph:

Reproduce the issue, while the `bao secrets disable` command is stuck, run the following in another terminal to capture a profile:

```bash
curl --header "X-Vault-Token: ${BAO_TOKEN}" "$BAO_ADDR/v1/sys/pprof/goroutine" > goroutines.pprof
```

to view the profile, run this:

```bash
[...]

PR #2630, opened 2026-03-13
fix race condition in jobmanager

phil9909:
## Description

Fixes a race in JobManager that made `AddJob` potentially blocking (and - together with other locks - led to a deadlock in the RollbackManager)

When this was originally implemented, I guess the assumption was that this can't block, because of the `if len(j.queues) == 0 {` meaning only the scheduled job would write to the channel and this happens under `j.l.Lock()` so it can't race.

BUT

The inner loop of `assignWork` works like this: https://github.com/openbao/openbao/blob/bae41858140bb28c2b28d3ed7cea07988f4c7ab3/helper/fairshare/jobmanager.go#L279-L299

1. We drain `newWork` (which is good)
2. We call `getNextJob()` 
3. We dispatch the job

`getNextJob` does proper locking and will remove empty queues.

What I think happens is: `AddJob` gets called between step 1 and 2 as well as between 2 and 3 (with and otherwise empty queue).

The first time `len(j.queues)` was zero and a message was sent to `newWork`, but it does not block, because it's a buffered channel.
The second time `len(j.queues)` is zero again (because `getNextJob` was called in between and popped the first job) and now sending to `newWork` blocks. 

To fix this, I wrapped this in a `select` block with a `default` arm.

## Rationale

Resolves: #2612

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2630.txt
+++ b/changelog/2630.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: fix potential deadlock in JobManager, which can cause mount deletion timeouts
+```
--- a/helper/fairshare/jobmanager.go
+++ b/helper/fairshare/jobmanager.go
@@ -108,7 +108,10 @@ func (j *JobManager) AddJob(job Job, queueID string) {
 		defer func() {
 			// newWork must be buffered to avoid deadlocks if work is added
 			// before the job manager is started
-			j.newWork <- struct{}{}
+			select {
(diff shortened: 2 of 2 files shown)


satoqz (approved):
Makes sense to me.

Do you want to add a changelog entry?

cipherboy:
Very nice job @phil9909! Would not have expected that at all.

PR #2631, opened 2026-03-13
Fix expired test certificates (#2552)

phil9909:
Backport of https://github.com/openbao/openbao/pull/2552 (clean cherry-pick)


satoqz:
@phil9909 can we standardize on removing only `needs-backport` but keeping `backport-x.x.x` on the original PR? I think that's useful to quickly tell which release a change was backported to, if at all.

PR #2620, opened 2026-03-11
Remove log gating

satoqz:
## Description

This removes log gating during the startup phase of the server, agent, proxy & debug commands, including the associated undocumented `-disable-gated-logs` flag.

Log gating is a cosmetic feature that was introduced all the way back in 2015. During startup, any logs generated by the system are held back (buffered internally) and only flushed once we've printed:

```
==> OpenBao server started! Log data will stream in below:
```

This makes the server startup logs slightly less messy as output intended to be read by humans only is always pinned to the top of the log.

## Rationale

Log gating comes with the severe drawback that any logs produced during the gated phase will not be streamed to the output in real-time. This tradeoff is mostly acceptable as the initial gated phase during startup is typically very short and does not produce many logs anyway. Still, looking specifically at `command/server.go`, the gated phase covers setup code for several critical components, to name a few:

- Physical storage backend
- Auto Seals
- Service Discovery
- Metrics
- Listeners

Additionally, with https://github.com/openbao/openbao/pull/2586, this phase would include:

- Initial OCI plugin downloads
- Auto Seal / KMS plugin setup

Particularly OCI plugin downloads may lengthen the pre-core, gated setup phase significantly. A lack of logs makes this hard to debug; the logs generated by the OCI downloader are only visible once plugin downloads have finished and either failed or core has started.

We could come up with workarounds and expose only specific logs before releasing the gate, though I believe entirely removing log gating is the simplest & best solution that will cover all kinds of scenarios.

As a result, a small amount of logs may be displayed before the

```
==> OpenBao server configuration
```

header is reached, and the following

```
==> OpenBao server started!
```

header will appear once all logs that lead to successful startup have been displayed in real-time, not before them.

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #2657, opened 2026-03-17
Fix docker test cert injection race condition

satoqz:
## Description

This fixes a frequent test flake I introduced with https://github.com/openbao/openbao/pull/2620 that affected all tests using `NewTestDockerCluster`.

Some background first:

`NewTestDockerCluster` initially starts nodes with listeners configured against a dummy cert that is generated pre-startup. Once the container is up, we learn its real IP and write a new certificate that has the correct SAN into the container and send SIGHUP.

To do this, we must wait for the listener to read the initial dummy cert such that we don't swap the cert out in a racy way, e.g., by sending SIGHUP before the signal listener is registered or modifying files while they're still being loaded.

Previously, waiting for the listener to be ready was achieved by waiting for _any_ kind of log output from the node. This worked because the log gate would buffer structured logs (stderr) emitted pre listener startup and emit the first stdout line (UI messages) only once either listeners are ready or startup failed.

With https://github.com/openbao/openbao/pull/2620, stderr log lines can now appear _before_ listeners are ready. To fix this, we simply wait for the first stdout line only instead of both channels. (At least I think, CI will tell me...)

## Rationale

I've tried refactoring the readiness detection to be network-based but that was a rabbit hole that led me nowhere.

## Acknowledgements

       also allowed.

### DIFF
--- a/sdk/helper/testcluster/docker/environment.go
+++ b/sdk/helper/testcluster/docker/environment.go
@@ -694,16 +694,21 @@ func (n *DockerClusterNode) Start(ctx context.Context, opts *DockerClusterOption
 	wg.Add(1)
 	var seenLogs atomic.Bool
 	logConsumer := func(s string) {
+		n.Logger.Trace(s)
+	}
+	logStdout := &LogConsumerWriter{func(s string) {
+		// HACK: The first message printed to stdout implies that listeners are
+		// ready, or that startup failed. We use this to time the rotation of
+		// the initial certificate, such that:
+		// 1. SIGHUP is not sent too early.
+		// 2. We don't cause a race condition on the file system where OpenBao
+		//    will see either no cert or a mismatched cert/key.
 		if seenLogs.CompareAndSwap(false, true) {
 			wg.Done()
 		}
 		n.Logger.Trace(s)
-	}
-	logStdout := &LogConsumerWriter{logConsumer}
+	}}
 	logStderr := &LogConsumerWriter{func(s string) {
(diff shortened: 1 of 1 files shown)


sdk/helper/testcluster/docker/environment.go
```diff
@@ -694,16 +694,21 @@ func (n *DockerClusterNode) Start(ctx context.Context, opts *DockerClusterOption
...
 	logConsumer := func(s string) {
-		if seenLogs.CompareAndSwap(false, true) {
-			wg.Done()
-		}
 		n.Logger.Trace(s)
 	}
-	logStdout := &LogConsumerWriter{logConsumer}
-	logStderr := &LogConsumerWriter{func(s string) {
+	logStdout := &LogConsumerWriter{func(s string) {
+		// HACK: The first message printed to stdout implies that listeners are
```
suprjinx, line 700: maybe it's not a "hack" :-)

cipherboy: Makes sense to me too, tbh.

satoqz: I kinda hate it :D

If I was to redesign I'd probably want the failure case to be handled by a context that is canceled when the container happens to die in the background, and the success case to select on a more specific log line, preferably one with a well-known event tag or so.

But this is the real world, so fine.

satoqz:
3 CI runs without a single failure (well, besides actions/checkout getting stuck that one time), merging...

PR #2668, opened 2026-03-19
fix: close HTTP servers during shutdown to avoid alarming logs

### LINKED ISSUE #2667: Listeners error on shutdown
This causes alarming logs on shutdown because listeners are closed directly instead of via `http.Server.Close`.

```
==> OpenBao shutdown triggered
HTTP server (listening on 127.0.0.1:8200) exited with error: accept tcp 127.0.0.1:8200: use of closed network connection
```

_Originally posted by @satoqz in https://github.com/openbao/openbao/pull/2229#discussion_r2860911031_

Notably this change is unreleased, but we should certainly fix it before v2.6.

Deepak8858:
Fixes #2667

Previously, listeners were closed directly while the HTTP servers were still serving, causing error logs on shutdown. Now, servers are explicitly closed, ensuring a graceful exit.

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -547,7 +547,11 @@ func (c *ServerCommand) runRecoveryMode() int {
 		})
 	}
 
+	var servers []*http.Server
 	listenerCloseFunc := func() {
+		for _, srv := range servers {
+			srv.Close()
+		}
 		var errs error
 		for _, ln := range lns {
 			errs = errors.Join(errs, ln.Close())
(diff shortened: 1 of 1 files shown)


satoqz:
@Deepak8858 please have a look at [CONTRIBUTING.md](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md) and follow the [Pull Request template](https://github.com/openbao/openbao/blob/main/.github/pull_request_template.md). Thanks!

Issue #2664, opened 2026-03-18
oidcPeriodicFunc: OIDC named key rotation silently broken for all non-root namespaces in v2.5.1 (regression from PR #2226)


JAYKRISHNAN:
**Describe the bug**

I think there is a bug in `v2.5.1`.  OIDC named key rotation never runs for non-root namespaces. I noticed that `v2.4.4` rotates OIDC keys in all namespaces as expected. `v2.5.1` silently skips rotation in all non-root namespaces.

I think that this might be a regression introduced by [PR 2226 Fix namespace locking, transaction issues around creation, deletion](https://github.com/openbao/openbao/pull/2226)  [in the commit - Refactor identity store periodic func](https://github.com/openbao/openbao/commit/6b57835bb4a6dcf17ae4286726ac360b0672cd43) which was shipped in `v2.5.1`

We faced the issue which was mentioned in the commit message of  [this commit - Refactor identity store periodic func](https://github.com/openbao/openbao/commit/6b57835bb4a6dcf17ae4286726ac360b0672cd43) - i.e. 

```
This periodic function was running once per namespace, iterating over all namespaces. This causes significant read load on the cluster as the number of namespaces grows, triggering what appears to be a soft deadlock due to the large number of storage operations.
```

When trying to validate that upgrading to `v2.5.1` would resolve the significant read load issue, I could not reproduce the key rotation for all namespaces. 

**To Reproduce**

I  ran Openbao docker images for 2 versions - `v2.4.4` and `v2.5.1`. The rotation is happening as expected in `v2.4.4` whereas in `v2.5.1` the rotation is not happening.

## Setup

```bash
# Start container (-dev mode: inmem storage, auto-unsealed, root token = "root")

docker run -d --name openbao-<version> -p <port>:8200 \
  -e BAO_LOG_LEVEL=warn \
  quay.io/openbao/openbao:<version> \
  bao server -dev -dev-root-token-id=root

export BAO_ADDR=http://127.0.0.1:<port>
export BAO_TOKEN=root

# Create 1 namespace, 1 OIDC key (rotation_period=2m), 1 role

bao namespace create test-ns-1

bao write -namespace=test-ns-1 identity/oidc/key/test-key \
  rotation_period=2m verification_ttl=10m algorithm=RS256

bao write -namespace=test-ns-1 identity/oidc/role/test-role key=test-key ttl=5m

# T=0 snapshot
[...]

PR #2669, opened 2026-03-19
Fix regression in OIDC named key rotation

JAYKRISHNAN:
## Description

- We had a regression in `oidcPeriodicFunc` OIDC named key for non root namespaces  in v2.5.1 which caused OIDC named key rotation never runs for non-root namespaces.
- The regression introduced by [PR 2226 Fix namespace locking, transaction issues around creation, deletion](https://github.com/openbao/openbao/pull/2226)  [in the commit - Refactor identity store periodic func](https://github.com/openbao/openbao/commit/6b57835bb4a6dcf17ae4286726ac360b0672cd43) which was shipped in `v2.5.1`
- See issue - https://github.com/openbao/openbao/issues/2664 for more details
- The bug was because of manual prepending of `ns.Path` to `MatchingStorageByAPIPath` which was already prepending `ns.path` causing a double-prefixed path like `test-ns-1/test-ns-1/identity/oidc`
- This change fixes the issue by removing the manual prepending of `ns.Path`

## Rationale

- We need to fix the regression so that the `oidcPeriodicFunc` run is not broken and runs successfully for non root namespaces

Resolves: #2664

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/identity_store_oidc_test.go
+++ b/vault/identity_store_oidc_test.go
@@ -1160,7 +1160,7 @@ func TestOIDC_PeriodicFunc(t *testing.T) {
 // for non root namespaces.
 func TestOIDC_PeriodicFunc_NonRootNamespace(t *testing.T) {
 	c, _, _ := TestCoreUnsealed(t)
-	rootCtx := namespace.RootContext(context.TODO())
+	rootCtx := namespace.RootContext(t.Context())
 
 	ns := testCreateNamespace(t, rootCtx, c.systemBackend, "ns", nil)
 	nsCtx := namespace.ContextWithNamespace(rootCtx, ns)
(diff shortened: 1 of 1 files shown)


JAYKRISHNAN:
@cipherboy Could you re-run the failing test ? It looks like it is unrelated to the change made in this PR.

vault/identity_store_oidc_test.go
```diff
@@ -1155,6 +1155,75 @@ func TestOIDC_PeriodicFunc(t *testing.T) {
 	}
 }
 
+// TestOIDC_PeriodicFunc_NonRootNamespace tests that oidcPeriodicFunc rotates
+// OIDC named keys in non-root namespaces. This is a test for a regression that happened
+// for non root namespaces.
+func TestOIDC_PeriodicFunc_NonRootNamespace(t *testing.T) {
+	c, _, _ := TestCoreUnsealed(t)
+	rootCtx := namespace.RootContext(context.TODO())
```
satoqz:
```suggestion
	rootCtx := namespace.RootContext(t.Context())
```

vault/identity_store_oidc_test.go
```diff
@@ -1155,6 +1155,75 @@ func TestOIDC_PeriodicFunc(t *testing.T) {
...
+	key.NextRotation = time.Now().Add(key.RotationPeriod)
+
+	// Store namedKey
+	entry, _ := logical.StorageEntryJSON(ident.NamedKeyConfigPath+key.Name, key)
+	if err := storage.Put(nsCtx, entry); err != nil {
+		t.Fatal("writing to in mem storage failed")
+	}
+
+	// sleep for the rotation period
+	time.Sleep(key.RotationPeriod + 100*time.Millisecond)
```
satoqz: Can we make this shorter or instant?

vault/identity_store_oidc_test.go
JAYKRISHNAN: Made the change in this commit - https://github.com/openbao/openbao/pull/2669/changes/3af518c5f351a5966b6faf8c95ca147bf891a440

vault/identity_store_oidc_test.go
JAYKRISHNAN: Made the change in this commit - https://github.com/openbao/openbao/pull/2669/changes/3af518c5f351a5966b6faf8c95ca147bf891a440

Issue #2665, opened 2026-03-18
Removal of Entropy Augmentation hooks


cipherboy:
OpenBao inherited many hooks from Vault CE for Vault Enterprise's Entropy Augmentation. This used Go standard crypto library's RNG hooks with HSM provided RNG. Go recently decided not to support these hooks any longer (after previously going through a deprecation period that conditionally read additional randomness from the source to ensure they cannot be used for deterministic key generation from PRFs). 

As we do not use this behavior, we are not impacted but it gives us a reason to clean up the code. 

See also: https://github.com/golang/go/issues/76856

mwriter:
> Go recently decided not to support these hooks any longer

And how can you evaluate the adequacy of such changes?
Maybe, instead of dutifully deleting our own generations, we need to fight for the cancellation of these inadequate changes in the language?
You can't help but realize that they won't stop there.

cipherboy:
@mwriter Personally, I do not understand the need for entropy augmentation as a feature of this software. 

You very rarely have OpenBao running in a vacuum and thus supplementing or replacing `/dev/urandom` is very clearly the better choice for other applications.

Even if you are purely running OpenBao and nothing else needs hardware-guaranteed randomness, you run into the fundamental shortcoming that things like the TLS stack (through which nearly all secrets flow -- either directly for TLS listener or indirectly for GRPC request forwarding) or if not that, AES-GCM nonce generation for secret storage via barrier encryption, do not use Entropy Augmentation. So even if OpenBao internally used EA consistently, you'd still have a boundary that raw secret data touched that was not hardware generated.

Many of these are high throughput/latency sensitive as well, so hardware round-trips (e.g., networked HSM, admittedly mitigated by PCIe HSM) may not be appealing.

At any rate, EA is a value proposition that says certain keys matter more -- and a threat model with compromised software randomness does not hold up, independent of EA. 

In such a world (that certain keys matter more -- but not the TLS session keys), external key support (neé Managed Keys, in Vault Ent terminology) is much stronger guarantee and (will be) used nearly everywhere EA would be: keys are generated in hardware and never leave hardware. With EA, while hardware provides the randomness, but software stores the key and can thus be compromised as software would. But with external/managed keys, hardware can provide stronger storage guarantees which makes the threat model objectively better than EA. So, I'd recommend any org which has HSM to use them properly, for relevant sensitive-key crypto operations, and not trust just use an expensive RNG.

[...]

PR #2670, opened 2026-03-19
Remove Entropy Augmentation hooks from core

satoqz:
## Description

Removes `secureRandomReader` from core.

80% of the diff is going down the errcheck rabbit hole and not directly related, so this is not as big of a change as the +/- makes it appear.

## Rationale

This covers most of #2665, though `GetRandomReader` remains in the SDK on `framework.Backend`.

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/logical/ssh/path_config_ca.go
+++ b/builtin/logical/ssh/path_config_ca.go
@@ -14,7 +14,6 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
-	"io"
 	"net/http"
 
 	"github.com/openbao/openbao/sdk/v2/framework"
@@ -137,11 +136,7 @@ func caKey(ctx context.Context, storage logical.Storage, keyType string) (*keySt
 	return &keyEntry, nil
 }
 
-func generateSSHKeyPair(randomSource io.Reader, keyType string, keyBits int) (string, string, error) {
-	if randomSource == nil {
-		randomSource = rand.Reader
-	}
-
+func generateSSHKeyPair(keyType string, keyBits int) (string, string, error) {
 	var publicKey crypto.PublicKey
 	var privateBlock *pem.Block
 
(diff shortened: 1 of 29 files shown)


builtin/logical/ssh/path_config_ca.go
```diff
@@ -155,7 +150,7 @@ func generateSSHKeyPair(randomSource io.Reader, keyType string, keyBits int) (st
 			return "", "", fmt.Errorf("refusing to generate weak %v key: %v bits < 2048 bits", keyType, keyBits)
 		}
 
-		privateSeed, err := rsa.GenerateKey(randomSource, keyBits)
+		privateSeed, err := rsa.GenerateKey(rand.Reader, keyBits)
```
cipherboy, line 153: I wonder if they'll let us pass `nil` in the future? 😆 

Alas, looks like they don't let us do so now.

satoqz:
@wslabosz-reply reports that he read tests 😄

DanGhita:
@satoqz , @cipherboy :In term of code cleaning, it probably makes sense to have this PR. But do we want to completely give up the option  to add external entropy / True Random to OpenBao ?
As a recall, in the KMS interface we had defined a GetRandom() function supposed to return True Random to OpenBao.  IMHO, the day where the External Keys will be fully implemented, we'll still need a good source of Entropy/Random for internal OpenBao needs.

cipherboy:
@DanGhita Please see the linked issue; Go has opted to remove the `crypto.Rand` support from things like `rsa.GenerateKey` so we'd need to reimplement key generation from scratch for it to use KMS's random source. c.f. https://go.dev/doc/go1.26#cryptorsapkgcryptorsa

IMO, it makes a lot more sense to either push that entropy down to the system (kernel) level or to fully utilize HSM (like the new APIs do) for key generation and storage, and not to just use HSM entropy.

PR #2671, opened 2026-03-19
Add CEL source in profiles

cipherboy:
CEL is a useful tool in profiles as it gives more verbose control for operators than simple Go templating. This essentially gives us room to inject additional arbitrary functions into the CEL source engine. In the future, we can allow configuration of the evaluation config, though I don't think we'll immediately need it.

## Rationale

This will be useful for profiles in core (#1960) and part of my WIP branch, split off to make reviews easier and land this part sooner for declarative self-init users. 

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2671.txt
+++ b/changelog/2671.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+self-init: add CEL support to self-initialization, allowing finer control over structuring requests
+```


helper/profiles/cel_source_test.go
```diff
@@ -0,0 +1,84 @@
...
+func TestCelSourceBuilder_Constant(t *testing.T) {
+	ctx := t.Context()
+	engine := &ProfileEngine{sourceBuilders: map[string]SourceBuilder{}}
+
+	field := map[string]interface{}{"expression": "123"}
+	src := CELSourceBuilder(ctx, engine, field)
+	require.NotNil(t, src)
+
+	_, ok := src.(*CELSource)
+	require.True(t, ok, "expected *CELSource")
```
satoqz: You can use `require.IsType` and `require.Implements` for these types of checks.

website/content/docs/concepts/profiles.mdx
```diff
@@ -201,3 +202,38 @@ initialize "example" {
     data = {}
   }
 }
+
```
satoqz, line 206:
````suggestion
```

````

File: helper/profiles/cel_source.go [hunk 3]
```diff
@@ -0,0 +1,175 @@
...
+func (s *CELSource) Evaluate(ctx context.Context, eh *EvaluationHistory) (interface{}, error) {
+	data := map[string]interface{}{}
+	history := eh.AsMap()
+
+	if s.allowRequests {
+		data["requests"] = history["requests"]
+	}
+	if s.allowResponses {
+		data["responses"] = history["requests"]
+	}
```
satoqz: Do you need `AsMap()`?

helper/profiles/cel_source.go
```diff
@@ -0,0 +1,175 @@
...
+//   - expression (string): expression to evaluate
+//   - variables([]map[string]string): additional variables to evaluate;
+//     each entry in the list is a map string->string which has two keys:
+//     -> name, the name of the variable to inject into the CEL context,
+//     -> expression, the expression for this variable to equal.
+//
+// When allowed as sources, the CEL context already includes:
+//   - requests
+//   - responses
+func CELSourceBuilder(ctx context.Context, engine *ProfileEngine, field map[string]interface{}) Source {
```
satoqz, line 27: I wonder if `SourceBuilder` and `Validate` should really take contexts?

helper/profiles/cel_source.go
cipherboy: Ah, yes, I removed that in the next PR.

helper/profiles/cel_source.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (s *CELSource) Evaluate(ctx context.Context, eh *EvaluationHistory) (interface{}, error) {
+	data := map[string]interface{}{}
+	history := eh.AsMap()
+
+	if s.allowRequests {
+		data["requests"] = history["requests"]
+	}
+	if s.allowResponses {
+		data["responses"] = history["requests"]
```
wslabosz-reply:
```suggestion
		data["responses"] = history["responses"]
```

helper/profiles/cel_source.go
cipherboy: `SourceBuilder` arguably not, though I think that's a bit outside of this scope. I was originally expecting CEL to require a context when I designed it, though I think since it doesn't, we can likely drop it from `Validate` as well.

helper/profiles/cel_source.go
cipherboy: Ah thanks. I had fixed that in #2727, so I've moved it forward.

helper/profiles/cel_source.go
satoqz: Yeah we can move this to an "unused variable cleanup" type PR later.

cipherboy:
Thanks all! Will merge and clean up the second PR then!

Issue #2660, opened 2026-03-18
BuildDate ldflag definition is misleading


hashworks:
According to the [packaging docs](https://openbao.org/docs/contributing/packaging/) the `github.com/openbao/openbao/version.BuildDate` ldflag definition should reference the date of the actual build, but the date of the commit. The [`scripts/build.sh`](https://github.com/openbao/openbao/blob/main/scripts/build.sh#L31) does this using `git show --no-show-signature -s --format=%cd --date=format:"%Y-%m-%dT%H:%M:%SZ" HEAD`.

I think this is misleading. A user running `bao version` is shown:
```
OpenBao v2.5.1 (e546fae8cbfe95d8f36a351deb2cd23bfb94119e), built 2026-02-23T18:02:34Z
```

If the user doesn't know the packaging guide, he would reason that the date the binary was build at was `2026-02-23`, not that the commit is from this date.

I would recommend to either adjust the `bao version` output from `built` to `release date` or `commit date` or to add another definition `CommitDate` used for this date and keep the `BuildDate` for the actual build date.

cipherboy:
Hey @hashworks -- I think this is the only reasonable choice for reproducibility. Otherwise you end up with non-determinism in the build system. We can probably adjust the documentation rather than change the behavior, IMHO.

hashworks:
Fair. In this case changing
```
OpenBao v2.5.1 (e546fae8cbfe95d8f36a351deb2cd23bfb94119e), built 2026-02-23T18:02:34Z
```
to
```
OpenBao v2.5.1 (e546fae8cbfe95d8f36a351deb2cd23bfb94119e), commited 2026-02-23T18:02:34Z
```

would clear up any confusion and remain reproducibility? Maybe also adjust `BuildDate` to `CommitDate`.

PR #2678, opened 2026-03-20
Change BuildDate to CommitDate to avoid confusion

hashworks:
## Description

Changes the misleading `BuildDate` to `CommitDate`, since the date of the git commit is used for releases instead of the actual date of the built to remain reproducibility. The docs mention this, but the variable name and CLI/API output is confusing to users.

## Rationale

Resolves: #2660

## Acknowledgements

       also allowed.

### DIFF
--- a/command/version_history.go
+++ b/command/version_history.go
@@ -4,6 +4,7 @@
 package command
 
 import (
+	"cmp"
 	"fmt"
 	"strings"
 
@@ -122,7 +123,10 @@ func (c *VersionHistoryCommand) Run(args []string) int {
 			return 2
 		}
 
-		table = append(table, fmt.Sprintf("%s | %s | %s", version, versionInfo["timestamp_installed"], versionInfo["commit_date"]))
+		// Fallback: Use build_date for backwards-compatibility with OpenBao < v2.6.0.
+		commitDate := cmp.Or(versionInfo["commit_date"], versionInfo["build_date"])
+
+		table = append(table, fmt.Sprintf("%s | %s | %s", version, versionInfo["timestamp_installed"], commitDate))
 	}
 
 	c.UI.Warn("")
--- a/api/sys_seal.go
+++ b/api/sys_seal.go
@@ -113,6 +113,9 @@ type SealStatusResponse struct {
 	StorageType      string `json:"storage_type,omitempty"`
 	// unused in OpenBao, but present for compatibility
 	Warnings []string `json:"warnings,omitempty"`
+	// Deprecated: Use CommitDate instead.
+	// This field remains for backwards-compatibility with OpenBao < v2.6.0.
+	BuildDate string `json:"build_date"`
 }
 
 type UnsealOpts struct {


website/content/docs/commands/version-history.mdx
```diff
@@ -9,9 +9,9 @@ The `version-history` command prints the historical list of installed OpenBao ve
 
 ```shell-session
 Note: Version tracking was added in 1.9.0. Earlier versions have not been tracked.
-      The Build Date will only be available for versions 1.11.0 or greater.
+      The Commit Date will only be available for versions 1.11.0 or greater.
```
hashworks, line 16: Should there be a comment that versions prior to this PR used `Build Date`?

version/version_base.go
```diff
@@ -13,7 +13,7 @@ var (
 	GitDescribe string
 
 	// The compilation date. This will be filled in by the compiler.
-	BuildDate string
+	CommitDate string
```
hashworks, line 16: Should we keep `BuildDate` here in addition to `CommitDate` and set the latter if the former is used? This would keep other build scripts compatible, but it's also not hard to change them.

hashworks:
I don't think the `builtin/logical/database TestBackend_StaticRole_LockRegression` test failure is related to this change.

version/version_base.go
satoqz: Given this errors and forces packagers to adjust to fix the build, I'm fine breaking it.

website/content/docs/commands/version-history.mdx
satoqz: That would be nice, perhaps a tiny note box below via

```
:::note

foo bar

:::
```

changelog/2678.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:change
+Renamed misleading ldflags definition `BuildDate` to `CommitDate`. Build systems need to adjust their pipelines to reflect this change.
```
satoqz: This should also note that the schema for the seal status API has changed. I'd probably split the entry like so:

````
```release-note:change
packaging: Renamed ...
```
```release-note:change
sys/seal-status: Renamed ...
```
````

File: vault/version_store.go [hunk 4]
```diff
@@ -24,10 +24,10 @@ const (
 type VaultVersion struct {
 	TimestampInstalled time.Time
 	Version            string
-	BuildDate          string
+	CommitDate         string
```
satoqz: I think we'll need to pin this to serialize as "BuildDate" so state on disk remains compatible 🤔

website/content/api-docs/system/version-history.mdx
```diff
@@ -12,7 +12,7 @@ This endpoint returns the version history of the OpenBao. The response will cont
 
 - `keys`: a list of installed versions in chronological order based on the time installed
 - `key_info`: a map indexed by the versions found in the `keys` list containing the following subkeys:
-  - `build_date`: the time (in UTC) at which the OpenBao binary used to run the OpenBao server was built.
+  - `commit_date`: the time (in UTC) of the commit which the OpenBao binary used to run the OpenBao server was built with
```
satoqz: nit:
```suggestion
  - `commit_date`: the time (in UTC) of the commit which the OpenBao binary used to run the OpenBao server was built at
```

version/version_base.go
hashworks: AFAIK it won't error.
```go
package main

import (
	"fmt"
)

var CommitDate string

func main() {
	fmt.Println(CommitDate)
}
```

```
$ go run -ldflags "-X 'main.CommitDate=someday'" .
someday
$ go run -ldflags "-X 'main.BuildDate=someday'" .

$ echo $?
0
```

changelog/2678.txt
satoqz: Ah, and `sys/version-history`, too

version/version_base.go
satoqz: Oh, sorry, I mean this as in: If we don't include backwards-compatibility here, setting the old flag will error, which is good because it won't be silently dropped.

Edit: Did I test this wrong? Let's see..

satoqz: Ah you're right and I'm bad at reading errors. I'd prefer to keep it backwards-compatible then and have both 🙈

changelog/2678.txt
hashworks: Resolved in 8f26153743

website/content/api-docs/system/version-history.mdx
hashworks: I was referring to the commit, but I agree that referring to the time is better. Resolved in 50e4b2e004

satoqz: I like to think of commits as locations, so it's "at" for me 😄

website/content/docs/commands/version-history.mdx
hashworks: Resolved in 148ddc3c2b

website/content/docs/commands/version-history.mdx
```diff
@@ -7,11 +7,15 @@ description: |-
 
 The `version-history` command prints the historical list of installed OpenBao versions in chronological order.
 
+:::note
+Versions up to 2.5.1 used "Build Date" instead of "Commit Date".
```
satoqz: I think we won't backport this to 2.5.x as it is minimally breaking (on the API), so it'd be "Versions prior to v2.6". Sorry, should've mentioned that.

version/version_base.go
hashworks: Two options then:

### A) Init function

```go
package main

import (
	"fmt"
)

var CommitDate string
var BuildDate string

func init() {
	if CommitDate == "" && BuildDate != "" {
		CommitDate = BuildDate
	}
}

func main() {
	fmt.Println(CommitDate)
}
```

```
$ go run -ldflags "-X 'main.BuildDate=someday'" .
someday
```

I would prefer this: No need to change other files since it keeps stuff to `version_base.go` and we can simply drop the init function if we someday drop the deprecated `BuildDate`.

### B) Get function

```go
package main

import (
	"fmt"
)

var CommitDate string
var BuildDate string

func commitDate() string {
	if CommitDate != "" {
		return CommitDate
	}
	return BuildDate
}

func main() {
	fmt.Println(commitDate())
}
```

Which would you prefer?

version/version.go
```diff
@@ -13,7 +13,7 @@ type VersionInfo struct {
 	Version           string `json:"version,omitempty"`
 	VersionPrerelease string `json:"version_prerelease,omitempty"`
 	VersionMetadata   string `json:"version_metadata,omitempty"`
-	BuildDate         string `json:"build_date,omitempty"`
+	CommitDate        string `json:"commit_date,omitempty"`
```
satoqz: I don't think this struct is ever (de-)serialized? We can probably drop the tags.

website/content/docs/contributing/packaging.mdx
```diff
@@ -73,7 +73,7 @@ definitions](https://pkg.go.dev/cmd/link) as appropriate:
 | :-------- | :---- |
 | `github.com/openbao/openbao/version.fullVersion` | Version number of OpenBao; typically this matches the upstream version. Must be semver (`<major>.<minor>.<patch>`). Required. |
 | `github.com/openbao/openbao/version.GitCommit` | Git commit for reference; string. Required. |
-| `github.com/openbao/openbao/version.BuildDate` | Build timestamp string in the format `%Y-%m-%dT%H:%M:%SZ` (UTC). Usually this is the date of the commit. Required. |
+| `github.com/openbao/openbao/version.CommitDate` | Date of the commit in the format `%Y-%m-%dT%H:%M:%SZ` (UTC). Not the built date to preserve reproducibility. Required. |
```
satoqz:
```suggestion
| `github.com/openbao/openbao/version.CommitDate` | Date of the commit in the format `%Y-%m-%dT%H:%M:%SZ` (UTC). Not the build date to preserve reproducibility. Required. |
```

version/version_base.go
satoqz:
```go
func init() {
	if CommitDate == "" {
		CommitDate = BuildDate
	}
}
```

should be good.

satoqz: Plus a comment on `BuildDate` only remaining for backwards-compatibility.

hashworks:
> Sorry for all the back-and-forth @hashworks

It's fine, this changes a lot of files and I will add a commit for each change-request anyway.

website/content/docs/commands/version-history.mdx
hashworks: Resolved in 4baf037

version/version_base.go
hashworks: Resolved in 244067781a

website/content/docs/contributing/packaging.mdx
hashworks: Correct, thanks. Changed it to `date of the build` since it sounds better. Resolved in 5f2a96473b989b30b24635c587bdc669b459c045.

version/version.go
hashworks: Maybe at some point it was planned to add some `--json` flag for the version command. But yeah, I don't see any serialization either. I'll drop them.

hashworks: Resolved in db9ef20016.

vault/version_store.go
hashworks: Do you mean in https://github.com/openbao/openbao/blob/main/vault/logical_system.go#L4550 ? 

```go
		entry := map[string]interface{}{
			"timestamp_installed": v.TimestampInstalled.Format(time.RFC3339),
			"commit_date":         v.CommitDate,
			"previous_version":    nil,
		}
```

hashworks: Ah, nevermind there is a JSON marshal in `version_store.go`
```go
	marshalledVaultVersion, err := json.Marshal(vaultVersion)
	if err != nil {
		return false, err
	}
```

satoqz: Here:

https://github.com/openbao/openbao/blob/c973c7693e46b188e338a43384eb3875398f834b/vault/version_store.go#L42-L50

and here:

https://github.com/openbao/openbao/blob/c973c7693e46b188e338a43384eb3875398f834b/vault/version_store.go#L126-L137

this gets written to storage as JSON and needs to survive version upgrades by definition 😄

hashworks: Added in 9fd3ebef59. I'll try to verify that this works with a raft cluster before I resolve this.

scripts/build.sh
```diff
@@ -47,7 +47,7 @@ mkdir -p bin/
 echo "==> Building bao..."
 ${GO_CMD} build \
     -gcflags "${GCFLAGS}" \
-    -ldflags "${LD_FLAGS} -X github.com/openbao/openbao/version.GitCommit='${GIT_COMMIT}${GIT_DIRTY}' -X github.com/openbao/openbao/version.BuildDate=${BUILD_DATE}" \
+    -ldflags "${LD_FLAGS} -X github.com/openbao/openbao/version.GitCommit='${GIT_COMMIT}${GIT_DIRTY}' -X github.com/openbao/openbao/version.CommitDate=${commit_date}" \
```
satoqz:
```suggestion
    -ldflags "${LD_FLAGS} -X github.com/openbao/openbao/version.GitCommit='${GIT_COMMIT}${GIT_DIRTY}' -X github.com/openbao/openbao/version.CommitDate=${COMMIT_DATE}" \
```

vault/version_store.go
```diff
@@ -24,10 +24,10 @@ const (
 type VaultVersion struct {
 	TimestampInstalled time.Time
 	Version            string
-	BuildDate          string
+	CommitDate         string `json:"BuildDate"` // Pinned to BuildDate to remain backwards-compatibility
```
satoqz:
```suggestion
	CommitDate         string `json:"BuildDate"` // Pinned to BuildDate to remain backwards-compatible.
```

changelog/2678.txt
```diff
@@ -0,0 +1,11 @@
+```release-note:change
+packaging: Renamed misleading ldflags definition `BuildDate` to `CommitDate`. Build systems need to adjust their pipelines to reflect this change.
+```
+
+```release-note:change
+sys/seal-status: Renamed misleading schema attribute `BuildDate` to `CommitDate`
```
satoqz:
```suggestion
sys/seal-status: Renamed misleading `build_date` response field to `commit_date`.
```

changelog/2678.txt
```diff
@@ -0,0 +1,11 @@
...
+```
+
+```release-note:change
+sys/version-history: Renamed misleading schema attribute `BuildDate` to `CommitDate`
```
satoqz:
```suggestion
sys/version-history: Renamed misleading `build_date` response field to `commit_date`.
```

hashworks:
Client binaries with this patch will fail to read the version history from servers prior to this patch:
```
$ bao version
OpenBao main (c973c7693e46b188e338a43384eb3875398f834b), committed 2026-03-20T08:55:38Z
$ bao version-history

Note:
Use of this command requires a server running Vault 1.10.0 or greater.
Version tracking was added in 1.9.0. Earlier versions have not been tracked.


Version  Installation Time     Commit Date
-------  -----------------     -----------
[...]
2.0.0    2026-03-16T15:45:36Z  %!s(<nil>)
2.5.1    2026-03-20T15:00:48Z  %!s(<nil>)
```

Do we want a fallback here? Or just a nil-check that displays n/a?

Similarly, `bao status -format=json` will display `"commit_date": "",`.

When both the client and the server have this PR it works fine:
```
2.0.0    2026-03-16T15:45:36Z  2026-03-04T10:34:34Z
2.5.1    2026-03-20T15:00:48Z  2026-02-23T18:02:34Z
```
```
"commit_date": "2026-03-20T08:55:38Z",
```

satoqz:
@hashworks I'd suggest we compromise and:

- In this patch, update the CLI to check for both `commit_date` and `build_date`. So newer clients will work with older servers. That would mean keeping both fields in `api.SealStatusResponse` for `bao status`, while `bao version-history` uses a plain map anyway.

- Accept that older clients will not work perfectly with newer servers. I think this is far more acceptable as upgrading the server is more involved than upgrading the client, e.g., I might hold the the server in production for a bit but upgrade the client on my developer machine right away.

Thoughts?

We can then also mark it deprecated in Go:

```go
type SealStatusResponse struct {
	// Deprecated: Use CommitDate instead.
	// This field remains for backwards-compatibility with OpenBao < v2.6.0.
	BuildDate        string `json:"build_date"`
	CommitDate       string `json:"commit_date"`
}
```

vault/version_store.go
hashworks: The test was successful.

changelog/2678.txt
hashworks: Resolved in fd803d07b513088177ce1e0fa4a7500677916ac0

vault/version_store.go
hashworks: Resolved in 43859efe14ed99a99bce7485a87dc6703b1d5dbe

scripts/build.sh
hashworks: Resolved in c0ad66801aa1a2ddf561aa8a27ba6e73e9cea240

hashworks:
> Thoughts?

@satoqz Seems fine to me. In my test server-to-server communication wasn't affected, especially for cluster updates. But I should note that I didn't test any auto-unseal features - not sure if those use that in some way.

satoqz:
I don't think this interacts with Auto Unseals or clustering so that should be fine.

command/version_history.go
```diff
@@ -122,7 +122,13 @@ func (c *VersionHistoryCommand) Run(args []string) int {
 			return 2
 		}
 
-		table = append(table, fmt.Sprintf("%s | %s | %s", version, versionInfo["timestamp_installed"], versionInfo["build_date"]))
+		commitDate, ok := versionInfo["commit_date"].(string)
+		if !ok {
+			// Fallback: Use build_date for backwards-compatibility with OpenBao < v2.6.0.
+			commitDate = versionInfo["build_date"].(string)
+		}
```
satoqz: This would panic if we reach the fallback and `build_date` is not set either, I think this works:

```suggestion
		// Fallback: Use build_date for backwards-compatibility with OpenBao < v2.6.0.
		commitDate := cmp.Or(versionInfo["commit_date"], versionInfo["build_date"])
```

...and then just let Sprintf take care of figuring out what type it is underneath.

hashworks: Good catch, resolved in 7938d1f1fe92f0f59e788e2a5a69700664572c10

Issue #2658, opened 2026-03-17
We could invert this logic and simplify it.


AndrewCharlesHay:
_Originally posted by @cipherboy in https://github.com/openbao/openbao/pull/2491#discussion_r2948950685_

cipherboy:
@AndrewCharlesHay Just to be clear, I don't think #2681 actually solves this issue. In the comment, I was pointing more to the `if ... { } return nil` structure of the function. It should probably just be `if !cond { return nil }`, or even hoist that to the calling site. That way the entire function body isn't doubly indented.

PR #2681, opened 2026-03-20
vault: simplify resolveTokenPolicies default policy logic

AndrewCharlesHay:
## Description

Decouple addDefault from noDefaultPolicy in the switch cases so each branch purely expresses intent, then apply noDefaultPolicy as a single gate at the final SanitizePolicies call.

Previously, addDefault was set to !noDefaultPolicy in each branch and then a second pass with strutil.StrListDelete was needed to strip "default" that may have arrived via inheritance. Now the logic flows in one direction: branches set `addDefault=true` when appropriate, and noDefaultPolicy is applied once at the end via `addDefault && !noDefaultPolicy`, with `strutil.StrListDelete` still handling "default" that was already present in the policy list before SanitizePolicies runs.

## Rationale

Related: #2658

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/token_store.go
+++ b/vault/token_store.go
@@ -4314,7 +4314,7 @@ func (ts *TokenStore) resolveTokenPolicies(ctx context.Context, req *logical.Req
 	// We are creating a token from a parent namespace. We should only use the input
 	// policies.
 	case ns.ID != parent.NamespaceID:
-		addDefault = !noDefaultPolicy
+		addDefault = true
 
 	// No policies specified, inherit parent
 	case len(policies) == 0:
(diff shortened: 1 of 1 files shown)


cipherboy (approved):
I believe I've convinced myself by looking at the full function this is equivalent, but would appreciate a second set of eyes. 

Regardless, thanks @AndrewCharlesHay!

vault/token_store.go
```diff
@@ -4334,22 +4334,22 @@ func (ts *TokenStore) resolveTokenPolicies(ctx context.Context, req *logical.Req
 			return logical.ErrorResponse("child policies must be subset of parent"), nil, logical.ErrInvalidRequest
 		}
 
-		// If the parent has default, and they haven't requested not to get it,
-		// add it. Note that if they have explicitly put "default" in
-		// data.Policies it will still be added because NoDefaultPolicy
-		// controls *automatic* adding.
```
satoqz, line 4340: For understanding, this was incorrect, yes?

satoqz (approved):

vault/token_store.go
AndrewCharlesHay: Not incorrect in terms of observable behavior — the old code produced the same result. The `!noDefaultPolicy` guard here was *redundant* rather than wrong: the old flow checked `noDefaultPolicy` both inside each switch branch (to suppress setting `addDefault = true`) and again afterward via `strutil.StrListDelete` to strip any inherited "default". Both paths conspired to give the correct answer, but the intent was scattered across multiple sites.

The refactor removes the redundant per-branch checks so each branch purely expresses "should default be added based on this branch's own criteria?" (here: does the parent carry "default"?). The `noDefaultPolicy` opt-out is then applied exactly once at the end (`addDefault && !noDefaultPolicy`), plus the existing `StrListDelete` for any "default" that arrived via explicit policy input. Net behavior is identical, but the control flow is easier to reason about.

satoqz: Oh behavior was clear, I was asking about the original code comment which seemed to document the opposite of what the original code actually did.

wslabosz-reply (approved):

vault/token_store.go
AndrewCharlesHay: Ah, sorry, misread the question. Yes, the original comment was misleading. It said *"if they have explicitly put 'default' in data.Policies it will still be added because NoDefaultPolicy controls *automatic* adding"*, but the code path at the bottom of `resolveTokenPolicies`:

```go
if noDefaultPolicy {
    finalPolicies = strutil.StrListDelete(finalPolicies, "default")
}
```

…unconditionally strips "default" from the final list whenever `no_default_policy` is set, regardless of whether the caller supplied "default" explicitly in `data.Policies`. So the "will still be added" claim only held until that post-switch step ran.

The refactor collapses both gates into one:

```go
finalPolicies := policyutil.SanitizePolicies(policies, addDefault && !noDefaultPolicy)
```

and the updated comment now says the explicit "default" will be *stripped* below, which matches the actual behavior both before and after this change.

Good catch, the old comment really did document the opposite of what the code did.

PR #2682, opened 2026-03-20
Enable sealable namespace creation

wslabosz-reply:
## Description
- Added `seal` optional parameter to namespace creation
- Adjusted namespace creation logic in `NamespaceStore` to allow for namespace seal creation within `SealManager`
- Adjusted affected tests

All logic changes are sourced from [namespaces-seal feature branch](https://github.com/openbao/openbao/tree/namespaces-seal)

## Rationale
Introduced changes allow for sealable namespace creation through API, currently without a way to unseal them after creation.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/logical_system_namespaces.go
+++ b/vault/logical_system_namespaces.go
@@ -5,16 +5,24 @@ package vault
 
 import (
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"strings"
 
+	"github.com/go-viper/mapstructure/v2"
 	"github.com/openbao/openbao/helper/namespace"
 	"github.com/openbao/openbao/sdk/v2/framework"
 	"github.com/openbao/openbao/sdk/v2/logical"
 )
+
+var namespacePathSchema = &framework.FieldSchema{
+	Type:        framework.TypeString,
+	Required:    false,
+	Description: "Path of the namespace.",
+}
 
 func (b *SystemBackend) namespacePaths() []*framework.Path {
 	namespaceListSchema := map[string]*framework.FieldSchema{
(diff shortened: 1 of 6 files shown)


cipherboy (approved):
Will let @satoqz review, but this looks good from my side.

vault/logical_system_namespaces.go
```diff
@@ -339,15 +344,40 @@ func (b *SystemBackend) handleNamespacesSet() framework.OperationFunc {
...
+				return logical.ErrorResponse("invalid seal config data"), logical.ErrInvalidRequest
+			}
+
+			if err := sealConfig.Validate(); err != nil {
+				return logical.ErrorResponse("invalid seal config: %v", err), err
+			}
+		} else {
+			// fallback to nil
+			sealConfig = nil
+		}
```
```suggestion
		var sealConfig *SealConfig
		if ok {
		  sealConfig = &SealConfig{}
			if err := mapstructure.Decode(sealRaw, &sealConfig); err != nil {
				return logical.ErrorResponse("invalid seal config data"), logical.ErrInvalidRequest
			}

			if err := sealConfig.Validate(); err != nil {
				return logical.ErrorResponse("invalid seal config: %v", err), err
			}
		}
```

vault/namespace_store.go
```diff
@@ -482,28 +495,39 @@ func (ns *NamespaceStore) setNamespaceLocked(ctx context.Context, nsEntry *names
...
 	ns.namespacesByUUID[entry.UUID] = entry
 	ns.namespacesByAccessor[entry.ID] = entry
 
+	var sealKeyShares [][]byte
 	if !exists {
-		// unlock before initializeNamespace since that will re-acquire the lock
+		if sealConfig != nil {
+			if err := ns.core.sealManager.SetSeal(ctx, sealConfig, entry, true); err != nil {
+				return nil, fmt.Errorf("failed to set namespace seal: %w", err)
+			}
```
satoqz, line 516: This writes into the parent namespace and won't be cleaned up by a counterpart in `namespaceCreationFailureJob` if we fail below IIUC.

wslabosz-reply: I don't think I understand, sealConfig is written through parent's barrier - yes, but it's written to the newly created namespace storage, which will be cleaned up using `ClearView`, so not sure what you mean by "won't be cleaned up".

satoqz (approved):

PR #2701, opened 2026-03-24
fix: apply socket file mode when user/group are not set

### LINKED ISSUE #2594: Unix socket filesystem mode is not set  if user and / or group are null
**Issue**
When using a unix listener as such, without socket_user = "", socket_group = "", the file mode is not applied to the socket:
```
listener "unix" {
    address = "/run/sockets/openbao.sock"
    socket_mode = "0660"
}
```

**To Reproduce**
Use the above listener for a unix socket, then use ls -la in the listener dir to view permissions

**Expected behavior**
According to documentation at https://openbao.org/docs/2.4.x/configuration/listener/unix/:
```
unix listener parameters

    address (string: "/run/openbao.sock", <required>) – Specifies the address to bind the Unix socket.

    socket_mode (string: "", <optional>) – Changes the access permissions and the special mode flags of the Unix socket.

    socket_user (string: "", <optional>) – Changes the user owner of the Unix socket.

    socket_group (string: "", <optional>) – Changes the group owner of the Unix socket.

    disable_unauthed_rekey_endpoints (bool: false) - Whether to disable requests to the legacy unauthenticated rekey endpoints (under /sys/rekey/* and /sys/rekey-recovery-key/*). These are a security risk to leave exposed on public listeners.
```

This states that neither socket_user, socket_group or socket_mode are required. It does not state that to perform a mode change (And potentially a group change, untested) we need to set socket_user to a non nul value.

The code under openbao/helper/listenerutil/listener.go:195:setFilePermissions() doesn't have a case for only mode setting:
```
func setFilePermissions(path string, user, group, mode string) error {
	var err error
	uid, gid := os.Getuid(), os.Getgid()

	if user != "" {
		if uid, err = strconv.Atoi(user); err == nil {
			goto GROUP
		}

		// Try looking up the user by name
		u, err := osuser.Lookup(user)
		if err != nil {
			return fmt.Errorf("failed to look up user %q: %v", user, err)
		}
		uid, _ = strconv.Atoi(u.Uid)
	}

GROUP:
	if group != "" {
		if gid, err = strconv.Atoi(group); err == nil {
			goto OWN
		}

		// Try looking up the user by name
		g, err := osuser.LookupGroup(group)
		if err != nil {
			return fmt.Errorf("failed to look up group %q: %v", user, err)
		}
		gid, _ = strconv.Atoi(g.Gid)
	}

OWN:
	if err := os.Chown(path, uid, gid); err != nil {
		return fmt.Errorf("failed setting ownership to %d:%d on %q: %v",
			uid, gid, path, err)
	}

	if mode != "" {
		mode, err := strconv.ParseUint(mode, 8, 32)
		if err != nil {
			return fmt.Errorf("invalid mode specified: %v", mode)
		}
		if err := os.Chmod(path, os.FileMode(mode)); err != nil {
			return fmt.Errorf("failed setting permissions to %d on %q: %v",
				mode, path, err)
		}
	}

	return nil
}
```

**Environment:**
* OpenBao Server Version 2.5.1
* Server Operating System: AlmaLinux 10.1 x86_64
* Running under rootless podman in quadlets as an unprivileged user, which does have 2770 permissions on mounted volume. The volume mount to reference is: /mnt/persist/role/pki/var/run:/run/sockets:Z

Quadlet file:
```
[Unit]
Description=OpenBao Secrets Manager
After=local-fs.target
Requires=podman.socket

[Container]
Pod=openbao.pod
Image=ghcr.io/openbao/openbao:latest
AutoUpdate=registry

Environment=BAO_SKIP_DROP_ROOT=true
Environment=SKIP_CHOWN=true

Exec=bao server -config=/openbao/config

Volume=/mnt/persist/role/pki/var/run:/run/sockets:Z
Volume=/mnt/persist/role/pki/var/lib/openbao-data/config:/openbao/config:Z
Volume=/mnt/persist/role/pki/var/lib/openbao-data/file:/openbao/file:Z
Volume=/mnt/persist/role/pki/var/lib/openbao-data/logs:/openbao/logs:Z

[Service]
Restart=always

[Install]
WantedBy=multi-user.target default.target
```

```hcl
ui = true

api_addr = "unix:///run/sockets/openbao.sock"
#cluster_addr = "http://127.0.0.1:8201"
#cluster_name = "openbao.pki.core.lab.dttc.lan"

storage "file" {
    path = "/openbao/file/store"
#    node_id = "{{ ansible_facts['hostname'] }}.pki.core.lab.dttc.lan"
}

#seal "static" {
#    current_key_id = "20260303-1"
#    current_key = "file:///openbao/unseal.key"
#}

#listener "tcp" {
#    address = "127.0.0.1:8200"
#
#    // FIXME We can enable this as we're using Caddy as a terminator, but eventually we could/should fix this
#    tls_disable = "true"
#}

listener "unix" {
    address = "/run/sockets/openbao.sock"
    socket_mode = "0660"
}

```

**Fixes**
This is either a documentation issue, where we state we need to explicitly state the user and/or group, or we need to modify openbao/helper/listenerutil/listener.go:195:setFilePermissions() method to incorporate proper support for each single ACL operation.

In this case, I fixed it by updating me config to:
```
listener "unix" {
    address = "/run/sockets/openbao.sock"
    socket_mode = "0660"
    socket_user = "root"
    socket_group = "root"
}
```

Be careful with this fix if you're running outside of a rootless container.

majiayu000:
## Summary

Fixes #2594

The conditional in `unixListenerFactory` (`command/server/listener_unix.go:22-24`) used `&&` (AND), requiring **all three** of `SocketMode`, `SocketUser`, and `SocketGroup` to be non-empty before creating the `UnixSocketsConfig`. This caused `socket_mode` to be silently ignored when only mode was specified without user/group.

**Fix:** Change `&&` to `||` so the config is created when **any** of the three fields is set. The downstream `setFilePermissions()` already handles empty fields correctly:
- Empty user/group defaults to the current process uid/gid
- Empty mode skips `chmod`

## Test plan

- Added `mode_only` subtest to `helper/listenerutil/listener_test.go` — creates a `UnixSocketListener` with only `Mode="644"` and verifies permissions are applied
- Added `TestUnixListenerModeOnly` to `command/server/listener_unix_test.go` — verifies the factory function applies mode when only `SocketMode` is set
- All existing tests continue to pass

## Acknowledgements

 - [ ] By contributing this change, I certify I have not used generative AI (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes.
 - [ ] By contributing this change, I certify I have signed-off on the [DCO ownership](https://developercertificate.org/) statement and this change did not use post-BUSL-licensed code from HashiCorp. Existing MPL-licensed code is still allowed, subject to attribution. Code authored by yourself and submitted to HashiCorp for inclusion is also allowed.

### DIFF
--- a/helper/listenerutil/listener_test.go
+++ b/helper/listenerutil/listener_test.go
@@ -75,12 +75,9 @@ func TestUnixSocketListener(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		mode, err := strconv.ParseUint("644", 8, 32)
-		if err != nil {
-			t.Fatal(err)
-		}
-		if fi.Mode().Perm() != os.FileMode(mode) {
-			t.Fatalf("expected permissions %o, got %o", os.FileMode(mode), fi.Mode().Perm())
+		expectedMode := os.FileMode(0o644)
+		if fi.Mode().Perm() != expectedMode {
+			t.Fatalf("expected permissions %o, got %o", expectedMode, fi.Mode().Perm())
 		}
 	})
 	t.Run("names", func(t *testing.T) {


helper/listenerutil/listener_test.go
```diff
@@ -55,6 +55,34 @@ func TestUnixSocketListener(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		mode, err := strconv.ParseUint("644", 8, 32)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if fi.Mode().Perm() != os.FileMode(mode) {
+			t.Fatalf("expected permissions %o, got %o", os.FileMode(mode), fi.Mode().Perm())
+		}
```
phil9909: In go (as in many programing languages) you can directly write octal numbers with the `0o` prefix (just like the more commonly known `0x` prefix for hex)

```suggestion
		expectedMode := os.FileMode(0o644)
		if fi.Mode().Perm() != expectedMode {
			t.Fatalf("expected permissions %o, got %o", expectedMode, fi.Mode().Perm())
		}
```

phil9909: Oh, I see you even did that in the other test 👍

phil9909:
I saw you deleted the "Acknowledgements" section from the PR template. I have re-added them now, please read them and check the boxes, if they apply.

helper/listenerutil/listener_test.go
majiayu000: Good catch, thanks! Fixed in 411ee1a — now using `0o644` octal literal directly instead of `strconv.ParseUint`.

majiayu000: Thanks for the review! 🙏

cipherboy:
Closing based on a glance at user profile and ignoring our PR template.

Issue #2324, opened 2026-01-20
Deviation from Vault in the Userpass auth method: password_hash parameter


mak-surf:
**Describe the bug**
In Vault the Userpass auth method accepts a `password_hash` instead of a `password` parameter (only one of them is required). In OpenBao the Userpass auth method only supports the `password` parameter.

I was in doubt if this is a bug, or a feature request, but since you state that the aim of OpenBao is to be a drop in replacement for Vault I decided to file this as a bug. Please let me know if I should take another route to get this to you attention 🙂 

**To Reproduce**
Check the documentation of Vault here: https://developer.hashicorp.com/vault/api-docs/v1.21.x/auth/userpass
Check the documentation of OpenBao here: https://openbao.org/api-docs/auth/userpass/

**Expected behavior**
I expect the functionality of Userpass to be the same as Vault. 

**Environment:**
* OpenBao Server Version (retrieve with `bao status`):
```
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    1
Threshold       1
Version         2.5.0-beta20251125
Build Date      2025-11-25T19:29:04Z
Storage Type    inmem
Cluster Name    vault-cluster-7e64f4bf
Cluster ID      a0f77143-b8fd-54f3-da1f-7a2441825f9d
HA Enabled      false
```
* OpenBao CLI Version (retrieve with `bao version`):
```
OpenBao v2.5.0-beta20251125 (a258d870924dcca231999390030292587b5b4b51), built 2025-11-25T19:29:04Z
```
* Server Operating System/Architecture:
`docker: openbao/openbao:2.5.0-beta20251125`

OpenBao server configuration file(s):

`NA`

**Additional context**
The `password_hash` is for a requirement to be able to switch to OpenBao from Vault for us. We use this heavily.

The `password_hash` parameter was introduced in Vault 1.17, which was released on June 12, 2024. 

I tested this on OpenBao 2.4.x and 2.5.x

I was in doubt if this is a bug, or a feature request, but since you state that the aim of OpenBao is to be a drop in replacement for Vault I decided to file this as a bug. I leave it up to you how to triage this differently off course 🙂

JanMa:
[...]

PR #2702, opened 2026-03-24
Add password_hash parameter to userpass engine

EnricoFusi:
## Description

Adds password_hash field to the userpass auth method, allowing operators
to provide a pre-hashed bcrypt string instead of a plaintext password
when creating or updating credentials via API. Exactly one of password
or password_hash must be provided.

## Rationale

OpenBao aims to be a drop-in replacement for Vault. Vault 1.17 introduced
password_hash support in the userpass auth method. This change restores
parity.

Resolves: #2324

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/credential/userpass/path_user_password.go
+++ b/builtin/credential/userpass/path_user_password.go
@@ -7,6 +7,7 @@ package userpass
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"golang.org/x/crypto/bcrypt"
 
@@ -119,6 +120,19 @@ func (b *backend) updateUserPassword(password, passwordHash string, userEntry *U
 		return err, nil
 	}
 
+	const (
+		// minCost is the minimum bcrypt cost enforced when passing a pre-hashed
+		// password to avoid accidentally setting a weakly hashed password. 5 is
+		// picked as the lower bound as this is the default in `htpasswd -B`.
+		minCost = 5
+
+		// maxCost is the maximum bcrypt cost accepted when passing a pre-hashed
+		// password to avoid DoS. Notably, bcrypt.DefaultCost is 10, we pick 12
+		// to allow for some leeway. If this is insufficient, open an issue and
+		// suggest a configuration path.
+		maxCost = 12
+	)
+
 	if password != "" {
 		hash, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
 		if err != nil {
@@ -126,8 +140,14 @@ func (b *backend) updateUserPassword(password, passwordHash string, userEntry *U
 		}
 		userEntry.PasswordHash = hash
 	} else {
-		if _, err := bcrypt.Cost([]byte(passwordHash)); err != nil {
+		cost, err := bcrypt.Cost([]byte(passwordHash))
+		switch {
+		case err != nil:
 			return errors.New("password_hash is not a valid bcrypt hash"), nil
+		case cost < minCost:
+			return fmt.Errorf("bcrypt cost of %d is too low, want %d at minimum", cost, minCost), nil
+		case cost > maxCost:
+			return fmt.Errorf("bcrypt cost of %d exceeds maximum cost of %d", cost, maxCost), nil
 		}
 		userEntry.PasswordHash = []byte(passwordHash)
 	}


builtin/credential/userpass/path_users.go
```diff
@@ -73,6 +74,15 @@ func pathUsers(b *backend) *framework.Path {
 				},
 			},
 
+			"password_hash": {
+				Type:        framework.TypeString,
+				Description: "Pre-hashed bcrypt password for this user. Mutually exclusive with password.",
+				Deprecated:  true,
```
wslabosz-reply: wait, why?

builtin/credential/userpass/password_validation.go
satoqz, line 1: No need for a new file, keep this in path_user_password.go (that would also provide additional justification for adding the OpenBao copyright header in it)

changelog/2324.txt
satoqz, line 1: Remove

builtin/credential/userpass/path_user_password.go
```diff
@@ -81,18 +94,27 @@ func (b *backend) pathUserPasswordUpdate(ctx context.Context, req *logical.Reque
 	return nil, nil
 }
 
+// updateUserPassword sets PasswordHash on userEntry from either a plaintext
+// password hashed with bcrypt, or a pre-hashed bcrypt string stored directly.
 func (b *backend) updateUserPassword(req *logical.Request, d *framework.FieldData, userEntry *UserEntry) (error, error) {
```
wslabosz-reply: I'd rearrange code a bit, we are doing `d.GetOk("password/password_hash")` 3 times across this single request. One of the checks in `validatePasswordInput` verifies if both of them are missing, but they cannot as we only call that validation after we know that at least one of them exists.

I'd probably make `updateUserPassword` and `validatePasswordInput` accept already parsed string values and not the `*framework.FieldData` to parse
/my 2c.

builtin/credential/userpass/password_validation_test.go
```diff
@@ -0,0 +1,313 @@
...
+		"password":      "plaintextpassword",
+		"password_hash": string(hash),
+	})
+
+	err = validatePasswordInput(d)
+
+	require.Error(t, err,
+		"both password and password_hash present must be rejected")
+	require.Contains(t, err.Error(), "only one of",
+		"error message must indicate mutual exclusion")
```
wslabosz-reply: there's a `require.ErrorContains` method which encompasses these both

builtin/credential/userpass/password_validation_test.go
```diff
@@ -0,0 +1,313 @@
...
+		Raw:    raw,
+		Schema: passwordFieldSchema(),
+	}
+}
+
+// TestValidatePasswordInput_OnlyPassword verifies that providing only the
+// plaintext password field is accepted as valid input.
+func TestValidatePasswordInput_OnlyPassword(t *testing.T) {
+	t.Parallel()
+
```
satoqz:
```suggestion
```

None of these tests are good candidates for `t.Parallel()`, they're CPU-bound.

builtin/credential/userpass/password_validation_test.go
satoqz, line 1: Do you think this can be rearranged to use a table-driven test?

wslabosz-reply, line 1: I'd recommend writing `validatePasswordInput` tests with https://go.dev/wiki/TableDrivenTests approach, would be very easy to define the input/output pairs and would remove a lot of redundant code.

builtin/credential/userpass/path_user_password.go
```diff
@@ -63,13 +74,20 @@ func (b *backend) pathUserPasswordUpdate(ctx context.Context, req *logical.Reque
 		return nil, errors.New("username does not exist")
 	}
 
-	userErr, intErr := b.updateUserPassword(req, d, userEntry)
+	password, _ := d.GetOk("password")
+	passwordHash, _ := d.GetOk("password_hash")
+
+	passwordStr, _ := password.(string)
+	passwordHashStr, _ := passwordHash.(string)
```
wslabosz-reply:
```suggestion
	passwordStr := d.Get("password").(string)
	passwordHashStr := d.Get("password_hash").(string)
```

builtin/credential/userpass/path_users.go
```diff
@@ -251,8 +260,14 @@ func (b *backend) userCreateUpdate(ctx context.Context, req *logical.Request, d
 		return logical.ErrorResponse(err.Error()), logical.ErrInvalidRequest
 	}
 
-	if _, ok := d.GetOk("password"); ok {
-		userErr, intErr := b.updateUserPassword(req, d, userEntry)
+	password, _ := d.GetOk("password")
+	passwordHash, _ := d.GetOk("password_hash")
+
+	passwordStr, _ := password.(string)
+	passwordHashStr, _ := passwordHash.(string)
```
wslabosz-reply:
```suggestion
	passwordStr := d.Get("password").(string)
	passwordHashStr := d.Get("password_hash").(string)
```

builtin/credential/userpass/path_users.go
```diff
@@ -291,10 +307,18 @@ func (b *backend) userCreateUpdate(ctx context.Context, req *logical.Request, d
...
 
 func (b *backend) pathUserWrite(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
-	password := d.Get("password").(string)
-	if req.Operation == logical.CreateOperation && password == "" {
-		return logical.ErrorResponse("missing password"), logical.ErrInvalidRequest
+	password, _ := d.GetOk("password")
+	passwordHash, _ := d.GetOk("password_hash")
+
+	passwordStr, _ := password.(string)
+	passwordHashStr, _ := passwordHash.(string)
```
wslabosz-reply:
```suggestion
	passwordStr := d.Get("password").(string)
	passwordHashStr := d.Get("password_hash").(string)
```

builtin/credential/userpass/path_users.go
EnricoFusi: It seemed a way to prevent this to appear in the login page. I doublecheck, it doesn't appear in any case, so I removed.

changelog/2324.txt
EnricoFusi: As requested.

builtin/credential/userpass/path_user_password.go
EnricoFusi: It required to change func signature, but is done as requested.

builtin/credential/userpass/password_validation_test.go
EnricoFusi: I assumed multicore. Just removed.

builtin/credential/userpass/password_validation_test.go
EnricoFusi: Sure. Currently I am using EAL5+ tests, which is why at the beginning in the comments of tests there is Preconditions, PostConditions and Invariants. I can use gotests and rigenerate. In the next push.

EnricoFusi: Sure. Currently I am using EAL5+ tests, which is why at the beginning in the comments of tests there is Preconditions, PostConditions and Invariants. I can use gotests and rigenerate. In the next push.

builtin/credential/userpass/path_users.go
EnricoFusi: Changed and pushed.

builtin/credential/userpass/path_users.go
EnricoFusi: Is done as requested.

builtin/credential/userpass/path_users.go
```diff
@@ -291,10 +304,18 @@ func (b *backend) userCreateUpdate(ctx context.Context, req *logical.Request, d
...
+
+	if req.Operation == logical.CreateOperation {
+		if err := validatePasswordInput(passwordStr, passwordHashStr); err != nil {
+			return logical.ErrorResponse(err.Error()), logical.ErrInvalidRequest
+		}
 	}
+
```
satoqz: Can we just move this entire check into `userCreateUpdate` or, alternatively, inline `userCreateUpdate` since it has a single caller only?

satoqz:
@EnricoFusi is on vacation so I've taken this over.

builtin/credential/userpass/path_users.go
```diff
@@ -1,4 +1,5 @@
 // Copyright (c) HashiCorp, Inc.
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
```
cipherboy, line 2: I'm fine with this, though at some point, we should apply this more globally. We should have diff data from `fork-point` to be able to apply this consistently. 

We should probably also take out the year more globally, it seems unnecessary.

builtin/credential/userpass/path_user_password.go
```diff
@@ -81,17 +96,41 @@ func (b *backend) pathUserPasswordUpdate(ctx context.Context, req *logical.Reque
...
+	}
+
+	if password != "" {
+		hash, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
+		if err != nil {
+			return nil, err
+		}
+		userEntry.PasswordHash = hash
+	} else {
+		if _, err := bcrypt.Cost([]byte(passwordHash)); err != nil {
```
cipherboy: @satoqz Do we want to enforce a min/max cost? This seems rather expensive in the general case, that someone can send arbitrarily hard passwords to the server. 

Mount tunables seems excessive, but would let an operator set values here. 

Thoughts?

satoqz: I've set a minimum of 5 and maximum of 12 for now, which seemed reasonable, see comments.

If there are requirements out of this range in the future, someone can open an issue and propose a configuration path.

cipherboy (approved):
Thank you @EnricoFusi and @satoqz!

mak-surf:
Thanks a lot for implementing this feature. When it appears in the nightly builds I will take it for a spin 👍

PR #2705, opened 2026-03-24
Address Ember-Data array like deprecation

JanMa:
## Description

Update the UI Codebase to address the [Ember-Data Array Like](https://deprecations.emberjs.com/id/ember-data-deprecate-array-like) deprecation. This is a lot of search and replace, I recommend looking at it commit by commit.

## Rationale

Before we can upgrade EmberJS in the UI any further, we need to first address all the deprecations introduced in version `5.0.0`. See here for an overview: https://deprecations.emberjs.com

## Acknowledgements

       also allowed.

### DIFF
--- a/ui/app/components/mfa/mfa-form.js
+++ b/ui/app/components/mfa/mfa-form.js
@@ -78,6 +78,7 @@ export default class MfaForm extends Component {
   @task *validate() {
     try {
       this.error = null;
+      this.countdown = 0;
       const response = yield this.auth.totpValidate({
         clusterId: this.args.clusterId,
         ...this.args.authData,
--- a/ui/app/templates/components/mfa/mfa-form.hbs
+++ b/ui/app/templates/components/mfa/mfa-form.hbs
@@ -38,7 +38,7 @@
                 placeholder={{if (gt constraint.methods.length 1) "Enter passcode"}}
                 spellcheck="false"
                 autofocus="true"
-                disabled={{or this.validate.isRunning this.newCodeDelay.isRunning}}
+                disabled={{or this.validate.isRunning this.countdown}}
                 @value={{constraint.passcode}}
                 data-test-mfa-passcode={{index}}
               />
@@ -51,7 +51,7 @@
           {{/if}}
         {{/each}}
       </div>
-      {{#if this.newCodeDelay.isRunning}}
+      {{#if this.countdown}}
         <div>
           <AlertInline @type="danger" @sizeSmall={{true}} @message={{this.codeDelayMessage}} />
         </div>
@@ -59,13 +59,13 @@
       <button
         id="validate"
         type="submit"
-        disabled={{or this.validate.isRunning this.newCodeDelay.isRunning}}
+        disabled={{or this.validate.isRunning this.countdown}}
         class="button is-primary {{if this.validate.isRunning 'is-loading'}}"
         data-test-mfa-validate
       >
         Verify
       </button>
-      {{#if this.newCodeDelay.isRunning}}
+      {{#if this.countdown}}
         <Icon @name="delay" class="has-text-grey" />
         <span class="has-text-grey is-v-centered" data-test-mfa-countdown>{{this.countdown}}</span>
       {{/if}}


ui/app/components/identity/popup-policy.js
```diff
@@ -8,7 +8,7 @@ import { computed } from '@ember/object';
 import Base from './_popup-base';
 
 export default Base.extend({
-  model: alias('params.firstObject'),
+  model: alias('params[0]'),
```
satoqz: To be sure, indexing works even if inside a string?

ui/app/components/auth-form.js
```diff
@@ -163,7 +163,9 @@ export default Component.extend(DEFAULTS, {
     return templateName;
   }),
 
-  hasCSPError: alias('csp.connectionViolations.firstObject'),
+  hasCSPError: computed('csp.connectionViolations.[]', function () {
```
satoqz, line 166: Is this correct?

(I really know nothing about Ember after all so I'm just flagging anything that's suspicious 😄)

ui/tests/integration/components/policy-form-test.js
```diff
@@ -121,7 +121,9 @@ module('Integration | Component | policy-form', function (hooks) {
     await click(SELECTORS.uploadFileToggle);
     assert.dom(SELECTORS.policyUpload).exists({ count: 1 }, 'Policy upload is shown after toggle');
     assert.dom(SELECTORS.policyEditor).doesNotExist('Policy editor is not shown');
-    await triggerEvent(SELECTORS.policyUpload, 'change', { files: [this.file] });
+    const input = find(SELECTORS.policyUpload);
+    Object.defineProperty(input, 'files', { value: [this.file], configurable: true });
+    await triggerEvent(SELECTORS.policyUpload, 'change');
     assert.dom(SELECTORS.nameInput).hasValue('test-policy', 'it fills in policy name');
     await click(SELECTORS.saveButton);
```
satoqz, line 126: what.

ui/app/components/identity/popup-policy.js
JanMa: No it doesn't and I should have used the same pattern as `policyName` below. Good catch :-)

ui/app/components/auth-form.js
JanMa: Yes this is correct. Based on the value of `csp.connectionViolations.[]` we compute a new value for `hasCSPError`. Whenever the dependency changes, `hasCSPError` will also be updated and trigger a UI refresh.

ui/tests/integration/components/policy-form-test.js
JanMa: This was a "fun" one. After the upgrade to EmberJS 4.12 this  test was flaky. It is supposed to test a file upload via the policy form. In old EmberJS versions it worked by simply calling the event Trigger for a file upload. In newer versions this doesn't reliably update the DOM so the uploaded file was not visible to the running event. To fix this I use `Object.defineProperty` to manually manipulate the DOM so the test works as intended. 

A lot of talking to Gemini was involved in finding this out :joy:

ui/app/components/transit-key-actions.js
```diff
@@ -75,7 +75,7 @@ export default Component.extend(TRANSIT_PARAMS, {
       return;
     }
     // eslint-disable-next-line ember/no-get
-    set(this, 'selectedAction', get(this, 'key.supportedActions.firstObject'));
+    set(this, 'selectedAction', get(this, 'key.supportedActions[0]'));
```
satoqz, line 78: Missed?

ui/app/routes/vault/cluster/secrets/backend/actions.js
```diff
@@ -30,6 +30,6 @@ export default EditBase.extend({
   setupController(controller, model) {
     this._super(...arguments);
     const { selectedAction } = this.paramsFor(this.routeName);
-    controller.set('selectedAction', selectedAction || model.secret.get('supportedActions.firstObject'));
+    controller.set('selectedAction', selectedAction || model.secret.get('supportedActions[0]'));
```
satoqz, line 33: Missed?

satoqz:
Nevermind, GitHub bugged out...

PR #2706, opened 2026-03-25
Add support for BAO_TOKEN_PATH to override ~/.vault-token

### LINKED ISSUE #2696: Add support for VAULT_TOKEN_FILE environment variable or some sort of "profile" selector

Currently, vault login and similar do not play nice with having multiple vault instances. This feature would allow for more easily coexisting/switching between profiles.

Add support for VAULT_TOKEN_FILE to override the path used - instead of $HOME/.vault-token

Hack workarounds setting $HOME or similar.

Some sort of "profile" selector similar to what is done with the AWS CLI - where "if a profile was defined in VAULT_PROFILE it would add that as a suffix like .vault-token-$VAULT_PROFILE

This would also help when using the sidecar in k8s - to have it automatically see/use /run/secrets/vault/token just by setting that env var.

Additionally, for CI/CD injection, this would allow safer injection of a vault token by using the "env var references file on a ramdisk/secure store" instead of "env var has the token".

**Additional context**

This wouldn't unfortunately translate to other tools using vault via libraries, but would at least work for the base cli client tool.

nneul:
## Description

Allow overriding of ~/.vault-token file path. This will allow multiple improvements:

  * coexistence of multiple vault sessions
  * direct support of using jenkins, k8s, or other tooling that typically provide an env var to provide location of a secret
  * support to use other storage beyond home directory for sensitive content, such as /dev/shm

## Rationale

Resolves: #2696 

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2706.txt
+++ b/changelog/2706.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+command: Allow overriding the location of ~/.vault-token via the BAO_TOKEN_PATH environment variable.
+```
--- a/command/agent_generate_config.go
+++ b/command/agent_generate_config.go
@@ -203,9 +203,17 @@ func generateConfiguration(ctx context.Context, client *api.Client, flagExec str
 		execCommand = []string{"env"}
 	}
 
-	tokenPath, err := homedir.Expand("~/.vault-token")
-	if err != nil {
-		return nil, fmt.Errorf("could not expand home directory: %w", err)
+	var tokenPath string
+	var err error
+
+	baoTokenPath := api.ReadBaoVariable("BAO_TOKEN_PATH")
+	if baoTokenPath == "" {
+		tokenPath, err = homedir.Expand("~/.vault-token")
+		if err != nil {
+			return nil, fmt.Errorf("could not expand home directory: %w", err)
+		}
+	} else {
+		tokenPath = baoTokenPath
 	}
 
 	templates, err := constructTemplates(ctx, client, flagPaths)
--- a/command/agent_generate_config.go
+++ b/command/agent_generate_config.go
@@ -203,17 +203,13 @@ func generateConfiguration(ctx context.Context, client *api.Client, flagExec str
 		execCommand = []string{"env"}
 	}
 
-	var tokenPath string
-	var err error
-
-	baoTokenPath := api.ReadBaoVariable("BAO_TOKEN_PATH")
-	if baoTokenPath == "" {
+	tokenPath := api.ReadBaoVariable(api.EnvTokenPath)
+	if tokenPath == "" {
+		var err error
 		tokenPath, err = homedir.Expand("~/.vault-token")
 		if err != nil {
 			return nil, fmt.Errorf("could not expand home directory: %w", err)
 		}
-	} else {
-		tokenPath = baoTokenPath
 	}
 
 	templates, err := constructTemplates(ctx, client, flagPaths)
--- a/command/token/helper_internal.go
+++ b/command/token/helper_internal.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/openbao/openbao/helper/homedir"
+	"github.com/openbao/openbao/api/v2"
 )
 
 var _ TokenHelper = (*InternalTokenHelper)(nil)
@@ -35,7 +36,12 @@ func NewInternalTokenHelper() (*InternalTokenHelper, error) {
 // populateTokenPath figures out the token path using homedir to get the user's
 // home directory
 func (i *InternalTokenHelper) populateTokenPath() {
-	i.tokenPath = filepath.Join(i.homeDir, ".vault-token")
+	baoTokenPath := api.ReadBaoVariable("BAO_TOKEN_PATH")
+	if baoTokenPath == "" {
+	    i.tokenPath = filepath.Join(i.homeDir, ".vault-token")
+	} else {
+	    i.tokenPath = baoTokenPath
+	}
 }
 
 func (i *InternalTokenHelper) Path() string {
--- a/command/token/helper_internal.go
+++ b/command/token/helper_internal.go
@@ -9,11 +9,10 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"strings"
 
-	"github.com/openbao/openbao/helper/homedir"
 	"github.com/openbao/openbao/api/v2"
+	"github.com/openbao/openbao/helper/homedir"
 )
 
 var _ TokenHelper = (*InternalTokenHelper)(nil)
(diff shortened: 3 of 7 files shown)


changelog/2706.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+cli: Allow overriding location of .vault-token with VAULT_TOKEN_FILE environment variable
```
satoqz:
```suggestion
command: Allow overriding the location of ~/.vault-token via the BAO_TOKEN_FILE environment variable.
```

command/token/helper_internal.go
```diff
@@ -35,7 +35,12 @@ func NewInternalTokenHelper() (*InternalTokenHelper, error) {
 // populateTokenPath figures out the token path using homedir to get the user's
 // home directory
 func (i *InternalTokenHelper) populateTokenPath() {
-	i.tokenPath = filepath.Join(i.homeDir, ".vault-token")
+	vaultTokenFile := os.Getenv("VAULT_TOKEN_FILE")
+	if vaultTokenFile == "" {
+	    i.tokenPath = filepath.Join(i.homeDir, ".vault-token")
+	} else {
+	    i.tokenPath = vaultTokenFile
+	}
```
satoqz, line 33: Do we want to move all of this setup to `NewInternalTokenHelper()` and remove this method? We can then avoid the `homedir.Dir()` call if the environment variable is set.

command/agent_generate_config.go
```diff
@@ -203,9 +203,17 @@ func generateConfiguration(ctx context.Context, client *api.Client, flagExec str
 		execCommand = []string{"env"}
 	}
 
-	tokenPath, err := homedir.Expand("~/.vault-token")
-	if err != nil {
-		return nil, fmt.Errorf("could not expand home directory: %w", err)
+	var tokenPath string
+	var err error
+
+	vaultTokenFile := os.Getenv("VAULT_TOKEN_FILE")
```
satoqz:
```suggestion
	vaultTokenFile := api.ReadBaoVariable("BAO_TOKEN_FILE")
```

This way it will prefer `BAO_TOKEN_FILE` and fall back to `VAULT_TOKEN_FILE`.

We should also make the environment variable a constant somewhere, likely a private one in `command` first and later move that to `api`.

website/content/docs/commands/index.mdx
```diff
@@ -263,7 +263,8 @@ your session information as a cookie in the browser. Token helpers are
 customizable, and you can even build your own.
 
 The default token helper stores the token in `~/.vault-token`. You can delete
-this file at any time to "logout" of OpenBao.
+this file at any time to "logout" of OpenBao. You can override this location
+by setting the `VAULT_TOKEN_PATH` environment variable.
```
satoqz:
```suggestion
by setting the `VAULT_TOKEN_FILE` environment variable.
```

though, IMHO, I prefer `VAULT_TOKEN_PATH`.

website/content/docs/commands/index.mdx
satoqz, line 1: While we're at it, do you mind replacing `VAULT_*` with `BAO_*` across this entire file? We really prefer those to be the canonical names though `VAULT_*` is always supported as a fallback.

satoqz (commented):
Long-term I would prefer to see proper AWS-like profile support but I don't think that's mutually exclusive with this simpler feature.

command/agent_generate_config.go
nneul: Nice. Will do - with this change/addition - is the preference to document using the BAO prefixes, or the VAULT prefixes, or both?

command/token/helper_internal.go
nneul: Seems like it would be cleaner to me, wondering about the different 'expand home dir' in the other section as well - which is preferred usage/pattern?

website/content/docs/commands/index.mdx
nneul: Definitely.

website/content/docs/commands/index.mdx
nneul: It seems like _FILE is more commonly used for secret references - though I don't have any strong preference here. However, looking at the rest of THIS codebase, _PATH is defeinitely more consistently used. I'll update.

command/agent_generate_config.go
satoqz: I think particularly new options don't need their `VAULT` variant documented. The automatic fallback is mostly there to ease migrations. With this approach, arguably there's not much of a reason to support the `VAULT` variant in the first place, but I suppose it is more consistent and perhaps helps familiarity.

command/token/helper_internal.go
satoqz: Hmm, fair point. I think either is fine, `Expand` is a tiny bit more indirection but arguably more concise.

command/token/helper_internal.go
```diff
@@ -35,7 +36,12 @@ func NewInternalTokenHelper() (*InternalTokenHelper, error) {
 // populateTokenPath figures out the token path using homedir to get the user's
 // home directory
 func (i *InternalTokenHelper) populateTokenPath() {
-	i.tokenPath = filepath.Join(i.homeDir, ".vault-token")
+	baoTokenPath := api.ReadBaoVariable("BAO_TOKEN_PATH")
```
cipherboy: Almost needs a `WithDefault` variant, but that's out of scope from this PR.

changelog/2706.txt
cipherboy: The rest of the changelog format suggestions still stand as well, @nneul.

But looking fairly good! Thanks for the other cleanup :-)

website/content/docs/commands/index.mdx
```diff
@@ -367,35 +375,27 @@ each invocation of the `bao` CLI typically only makes a few requests,
...
 The draft specifies that the SRV record lookup is ignored if a port is given.
 
-### `VAULT_HTTP_PROXY`
-
-HTTP or HTTPS proxy location which should be used by all requests to access OpenBao.
- When present, this overrides the default proxy resolution behavior.
-Format should be `http://server:port` or `https://server:port`.
-
-(See `VAULT_PROXY_ADDR` below).
-
```
JanMa, line 389: Why are you removing the docs for `VAULT_HTTP_PROXY`?

changelog/2706.txt
nneul: Updated, must have read that like 5 times before I could actually see what was being suggested....

website/content/docs/commands/index.mdx
nneul: Fixed.

command/token/helper_internal.go
satoqz: This would be good to resolve fwiw, otherwise you can't override a "home directory can't be determined" error by setting the env var.

satoqz:
Hey @nneul - is there continued interest in getting this over the finish line? If you've lost interest or are short on time let us know and someone else can take over :)

nneul:
> Hey @nneul - is there continued interest in getting this over the finish line? If you've lost interest or are short on time let us know and someone else can take over :)

I could use some clarity on what is being asked additional to do/change in it.

nneul:
Note that I'm not really a go programmer, so some of the syntax/etc. in these is outside of my "yeah, that's obvious" multi-language support -- so if you have an approach on a piece of it - by all means please add to it.

satoqz:
No worries, I've pushed a commit to resolve the last suggestion for you. Apologies if I wasn't clear!

satoqz (approved):
ACKing but would like another review since I've mingled with this.

wslabosz-reply (approved):
LGTM

satoqz:
Merging, many thanks @nneul!

PR #2727, opened 2026-03-26
Add input, text/template based input sources, output sink to the profile system

cipherboy:
## Description

~Based on https://github.com/openbao/openbao/pull/2671; will be rebased when that is merged.~

This introduces new sources into the profile system and expands its
execution environment:
    
 - An input source, for handling framework.FieldData information;
 - A text/template source, for performing simpler manipulations than
   what CEL would involve; and
 - An output sink, for structuring the result of the profile execution.
    
When the input, request, and response sources are in use, they can be
utilized by the CEL and text/template sources.

We also make the profile system re-entrant, allowing more flexibility in evaluated parameter definitions by nesting/combining multiple source types. 

Of these, only the template source is relevant for the existing self-init server. 

## Rationale

Inputs are necessary for Profiles when executed in core. 

## Acknowledgements

       also allowed.

### DIFF
--- a/sdk/framework/field_type.go
+++ b/sdk/framework/field_type.go
@@ -107,6 +107,11 @@ func (t FieldType) String() string {
 	}
 }
 
+// ParseFieldType converts known string values (as used in configuration
+// files) to FieldType constants. When new values are added, assignment of
+// a string constant ensures they can be used within the profile system.
+//
+// When the string constant is not known, an error will be returned.
 func ParseFieldType(value string) (FieldType, error) {
 	switch value {
 	case "string":


cipherboy:
@satoqz @wslabosz-reply This should be ready for review!

website/content/docs/concepts/profiles.mdx
```diff
@@ -41,9 +41,9 @@ If request fails, the engine stops immediately and returns an error to the calle
 
 ## Request/Response history
 
-In profiles every API call template and automatically tracked it so that its inputs and outputs can be reused later in the same run. 
-The mechanism—implemented and integrated request just before it is issued and the corresponding response immediately after it completes, then stores both as JSON in an in-memory map keyed by the outer block and request name. 
-Any subsequent field may reference those stored JSON objects through the request or response source types, using a selector to pull out a specific value. 
+In profiles every API call template and automatically tracked it so that its inputs and outputs can be reused later in the same run.
```
wslabosz-reply, line 44: something strange with wording here

website/content/docs/concepts/profiles.mdx
```diff
@@ -41,9 +41,9 @@ If request fails, the engine stops immediately and returns an error to the calle
...
+The mechanism—implemented and integrated request just before it is issued and the corresponding response immediately after it completes, then stores both as JSON in an in-memory map keyed by the outer block and request name.
```
wslabosz-reply, line 45: can you also rewrite this paragraph? I'm having hard time understanding it

website/content/docs/concepts/profiles.mdx
```diff
@@ -142,8 +142,9 @@ initialize "ca-setup" {
 
 #### `request` source
 
-- `initialize_name` `(string: <required>)` - name of the initialize block to
-  reference.
+- `<outer-block>_name` `(string: <required>)` - name of the outer block block to
```
wslabosz-reply, line 145: if there are no plans to introduce more possible names, can we change the type declaration `(string:<required>)` to reflect that this accepts just two possible const names of `initialize_name` and `context_name`; also adding `_name` to the end points towards it being `initialize/context_name_name`

sdk/framework/field_type.go
```diff
@@ -104,3 +106,44 @@ func (t FieldType) String() string {
...
+		return TypeInt, nil
+	case "int64":
+		return TypeInt64, nil
+	case "bool":
+		return TypeBool, nil
+	case "map":
+		return TypeMap, nil
+	case "duration_second":
+		return TypeDurationSecond, nil
+	case "signed_uration_second":
```
wslabosz-reply:
```suggestion
	case "signed_duration_second":
```

sdk/framework/field_data.go
```diff
@@ -28,6 +28,24 @@ type FieldData struct {
 	Schema map[string]*FieldSchema
 }
 
+// Clone is used by the profile system to duplicate FieldData: Schema is a
+// map globally initialized once per routable path, while Raw changes
+// per-request and is thus safe to mutuate in certain scenarios. We wish to
+// mutate Schema, to add custom per-profile (per-request) fields that we
+// parse from raw. Notably, we also do not mutate existing schema entries,
+// so this is essentially a shallow clone except of the schema map.
```
wslabosz-reply: I feel like colon is not the right punctuation here 
```suggestion
// Clone is used by the profile system to duplicate FieldData; Schema field is a
// map globally initialized once per routable path, while Raw changes
// per-request and is thus safe to mutate in certain scenarios. We wish to
// mutate Schema, to add custom per-profile (per-request) fields that we
// parse from raw. Notably, we also do not mutate existing schema entries,
// so this is essentially a shallow clone except of the schema map.
```
also
> mutuate

has to be a typo of mutate right?

helper/profiles/template_source.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	s.template = templateStr
+
+	rawData, present := s.field["data"]
+	if !present {
+		rawData = map[string]interface{}{}
+	}
+
+	data, ok := rawData.(map[string]interface{})
```
wslabosz-reply, line 71: I think `data` in documentation is annotated as `[]map[string]interface{}` which doesn't match the implementation, but I'm not sure which is correct

website/content/docs/concepts/profiles.mdx
```diff
@@ -238,3 +241,65 @@ initialize "provision" {
   }
 }
 ```
+
+#### `template` source
+
+- `template` `(string: <required>)` - the [`template`](https://pkg.go.dev/github.com/openbao/openbao/sdk/v2/helper/template#StringTemplate)
+  expression to evaluate; errs if not present.
+- `data` `([]map[string]interface{}): <optional>)`: additional context for the
```
wslabosz-reply:
```suggestion
- `data` `([]map[string]interface{}: <optional>)`: additional context for the
```

website/content/docs/concepts/profiles.mdx
cipherboy: Yeah, that's true. This was from @KrzysztofKornalewski-Reply; I'll revise in a follow-up PR.

website/content/docs/concepts/profiles.mdx
cipherboy: I suspect cross-plugin communication might be another place where we introduce this. I'm not sure what we'd call things in this context... :thinking:

helper/profiles/template_source.go
cipherboy: This one is correct, I'll fix it. The other one was a copy/paste error of CEL.

sdk/framework/field_data.go
```diff
@@ -28,6 +28,24 @@ type FieldData struct {
...
+	clone := &FieldData{}
+	clone.Raw = d.Raw
+	clone.Schema = make(map[string]*FieldSchema, len(d.Schema))
+
+	for key, schema := range d.Schema {
+		clone.Schema[key] = schema
+	}
+
+	return clone
+}
```
satoqz: I'm not sure if I don't find these semantics to be too application-specific to be in sdk. I would 100% blindly expect a `Clone()` method to clone the raw data instead of shallow copying it.

sdk/framework/field_type.go
```diff
@@ -104,3 +106,44 @@ func (t FieldType) String() string {
 		return "unknown type"
 	}
 }
+
+func ParseFieldType(value string) (FieldType, error) {
```
satoqz, line 115: A doc string would be nice given this is sdk.

sdk/framework/field_data.go
cipherboy: @satoqz Out of curiosity, why? This will end up being called in PKI as well, where we inject a new parameter occasionally depending on which endpoint gets called and whether or not it exists.

Data is request bound, so you discard it afterwards  regardless of modifications and for the most part you're just injecting or modifying schema fields (and at worst, injecting data). But schema needs to be actually closed since it is not request bound but instead one global copy that cannot be mutated without mutating everybody. In particular, if you are modifying data, typically that would be either useful throughout the rest of the request or not relevant otherwise, IMO. 

That schema is global and cross-request means we save memory, which is nice, but does open us up to this weird condition.

satoqz: @cipherboy the application and goal is perfectly reasonable.

The type name is `FieldData`. The included data (`.Raw`) is the primary content to me while `.Schema` is additional metadata. I expect `FieldData.Clone()` to deep-clone data, not a schema. I'm not saying that we should do this, just that this is what I would expect the method to do based on its name.

I'm happy if we call it `CloneSchema` or similar.

cipherboy: Very good, will revise the name!

I'm really curious if we're happy with how framework behaves or if we'd have something more standardized we'd want to move to in the future, tbh.

satoqz: I'm largely happy with it and think the non-standard bits (very forgiving conversion semantics) make a good compromise for a tool that has fully pluggable endpoints that must be driven by a CLI easily (`bao write foo x=y`)

helper/profiles/config.go
```diff
@@ -34,6 +35,44 @@ type RequestConfig struct {
 	AllowFailure interface{} `hcl:"allow_failure"`
 }
 
+// InputConfig is an untyped configuration object that contains one or more
+// fields in a logical.FrameworkData format. This is used with the
```
satoqz:
```suggestion
// fields in a framework.FieldSchema format. This is used with the
```

cipherboy:
@satoqz updated, LMKWYT!

helper/profiles/config.go
```diff
@@ -34,6 +35,44 @@ type RequestConfig struct {
 	AllowFailure interface{} `hcl:"allow_failure"`
 }
 
+// InputConfig is an untyped configuration object that contains one or more
+// fields in a logical.Fieldschema format. This is used with the
```
satoqz:
```suggestion
// fields in a framework.FieldSchema format. This is used with the
```

I'm sorry to keep nitpicking on this, but it is `framework.FieldSchema`.

```shell-session
$ rg 'type FieldSchema'
sdk/framework/backend.go
683:type FieldSchema struct {
```

cipherboy: Ah, do'h, too fast typo :-D Fixed.

cipherboy:
Thanks all! I'll go ahead and merge after CI passes and we can always revise later :-)

PR #2728, opened 2026-03-26
Add workflows to core via profile engine

cipherboy:
## Description

This is based on https://github.com/openbao/openbao/pull/2671 and https://github.com/openbao/openbao/pull/2727; will be rebased when those are merged.

This introduces workflows into the core subsystem:

Introduce the workflow system to Core

Introducing workflow into Core allows us to build a facade over the
OpenBao APIs, combing multiple API calls into a single API invocation,
as controlled by an operator. We introduce two new API sections:

- `sys/workflows/manage/<path>`, for managing profiles; and
- `sys/workflows/execute/<path>`, `sys/workflows/unauthed-execute/<path>`,
  and `sys/workflows/trace/<path>`, for executing workflows with or
  without authentication and to trace all request/responses that are
  executed (for debugging).

Workflows have several properties:

- The workflow to be executed
- A description of the workflow
- Whether check-and-set is required to update the workflow
- The version of the workflow (starts at 1)
- Whether unauthenticated calls to the workflow are allowed

Unauthenticated workflows must be explicitly allowed by the operator via
the `allow_unauthenticated_workflows` configuration option.

The workflow subsystem builds on top of profile engine which already
underlays self-initialization.

## Rationale

Resolves: #

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/workflow_store.go
+++ b/vault/workflow_store.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"path/filepath"
 	"strings"
-	"sync"
 
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/hcl"
@@ -16,6 +15,7 @@ import (
 	"github.com/openbao/openbao/helper/profiles"
 	"github.com/openbao/openbao/sdk/v2/framework"
 	"github.com/openbao/openbao/sdk/v2/helper/consts"
+	"github.com/openbao/openbao/sdk/v2/helper/locksutil"
 	"github.com/openbao/openbao/sdk/v2/logical"
 	"github.com/openbao/openbao/vault/barrier"
 )
@@ -78,21 +78,48 @@ func (we *WorkflowEntry) Parse(ctx context.Context) (*profiles.InputConfig, []*p
 }
 
 type WorkflowStore struct {
-	core   *Core
-	lock   sync.RWMutex
-	logger log.Logger
+	core        *Core
+	modifyLocks []*locksutil.LockEntry
+	logger      log.Logger
 }
 
 func NewWorkflowStore(c *Core) *WorkflowStore {
 	logger := c.baseLogger.Named("workflow")
 	return &WorkflowStore{
-		core:   c,
-		logger: logger,
+		core:        c,
+		modifyLocks: locksutil.CreateLocks(),
+		logger:      logger,
 	}
 }
 
 func (c *Core) setupWorkflowStore(ctx context.Context) {
 	c.workflowStore = NewWorkflowStore(c)
+}
+
+func (ws *WorkflowStore) lockWithUnlock(ctx context.Context) func() {
+	ns, err := namespace.FromContext(ctx)
+	if err != nil || ns == nil {
+		ns = namespace.RootNamespace
+	}
+
+	lock := locksutil.LockForKey(ws.modifyLocks, ns.UUID)
+
+	ws.logger.Trace("acquiring lock for", "namespace", ns.UUID)
+	lock.Lock()
+	return lock.Unlock
+}
+
+func (ws *WorkflowStore) rLockWithUnlock(ctx context.Context) func() {
+	ns, err := namespace.FromContext(ctx)
+	if err != nil || ns == nil {
+		ns = namespace.RootNamespace
+	}
+
+	lock := locksutil.LockForKey(ws.modifyLocks, ns.UUID)
+
+	ws.logger.Trace("acquiring lock for", "namespace", ns.UUID)
+	lock.RLock()
+	return lock.RUnlock
 }
 
 // getView returns the storage view for the given namespace
@@ -101,9 +128,13 @@ func (ws *WorkflowStore) getView(ns *namespace.Namespace) barrier.View {
 }
 
 func (ws *WorkflowStore) Get(ctx context.Context, path string) (*WorkflowEntry, error) {
-	ws.lock.RLock()
-	defer ws.lock.RUnlock()
+	// Check namespace existence before calling RLock().
+	_, err := namespace.FromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
 
+	defer ws.rLockWithUnlock(ctx)()
 	return ws.getLocked(ctx, path)
 }
 
@@ -144,8 +175,7 @@ func (ws *WorkflowStore) Set(ctx context.Context, workflow *WorkflowEntry, casVe
 	path := ws.sanitizePath(workflow.Path)
 	view := ws.getView(ns)
 
-	ws.lock.Lock()
-	defer ws.lock.Unlock()
+	defer ws.lockWithUnlock(ctx)()
 
 	existing, err := ws.getLocked(ctx, workflow.Path)
 	if err != nil {
@@ -196,8 +226,7 @@ func (ws *WorkflowStore) Delete(ctx context.Context, path string) error {
 	path = ws.sanitizePath(path)
 	view := ws.getView(ns)
 
-	ws.lock.Lock()
-	defer ws.lock.Unlock()
+	defer ws.lockWithUnlock(ctx)()
 
 	return view.Delete(ctx, path)
 }
@@ -211,8 +240,7 @@ func (ws *WorkflowStore) List(ctx context.Context, prefix string, recursive bool
 	prefix = ws.sanitizePath(prefix)
 	view := ws.getView(ns).SubView(prefix)
 
-	ws.lock.RLock()
-	defer ws.lock.RUnlock()
+	defer ws.rLockWithUnlock(ctx)()
 
 	var keys []string
 	if !recursive {
@@ -248,9 +276,7 @@ func (ws *WorkflowStore) Execute(ctx context.Context, path string, unauthed bool
 	}
 
 	workflow, err := func() (*WorkflowEntry, error) {
-		ws.lock.RLock()
-		defer ws.lock.RUnlock()
-
+		defer ws.rLockWithUnlock(ctx)()
 		return ws.getLocked(ctx, path)
 	}()
 	if err != nil {


cipherboy:
A poll on Zulip has determined these will be called "workflows", though we'll potentially leave the name of the driving engine the same since it is generic and will be used in (at least) two other places besides workflows (already in self-init and likely in cross-plugin communication in the future).

satoqz:
@cipherboy there seems to be a deadlock, see test-go (0).

cipherboy:
@satoqz Fixed, I came in one level lower since we already have that lock.

website/content/docs/configuration/index.mdx
```diff
@@ -259,6 +259,9 @@ can have a negative effect on performance due to the tracking of each lock attem
 - `disable_standby_reads` `(bool: false)` - Disable the handling of read-only
   requests on standby nodes.
 
+- `allow_unauthenticated_profiles` `(bool: false)` - Whether to allow execution
+  of unauthenticated profiles.
```
wslabosz-reply:
```suggestion
- `allow_unauthenticated_workflows` `(bool: false)` - Whether to allow execution
  of unauthenticated workflows.
```

vault/logical_system_workflows.go
```diff
@@ -0,0 +1,392 @@
...
+			},
+
+			Fields: map[string]*framework.FieldSchema{
+				"after": {
+					Type:        framework.TypeString,
+					Description: "Optional entry to begin listing after; not required to exist.",
+				},
+				"limit": {
+					Type:        framework.TypeInt,
+					Description: "Optional number of entries to return; defaults to all entries.",
```
wslabosz-reply, line 70: I'm sure this won't cause trouble if users create workflows like some people do with 1k+ pki mounts

vault/logical_system_workflows.go
```diff
@@ -0,0 +1,392 @@
...
+					},
+					Summary: "List workflows.",
+				},
+				logical.ScanOperation: &framework.PathOperation{
+					Callback: b.handleWorkflowsList(true),
+					Responses: map[int][]framework.Response{
+						http.StatusOK: {{Description: "OK", Fields: workflowListSchema}},
+					},
+					Summary: "Scan (recursively list) workflows.",
+				},
```
wslabosz-reply, line 174: nit: seems redundant to have these callbacks defined separately on both `workflows/manage` and `workflows/manage/`, wouldn't be able to define one regex that would handle `/<path>` whether it exists or not?

vault/workflow_store.go
```diff
@@ -0,0 +1,342 @@
...
+	if len(profile) == 0 {
+		return nil, nil, nil, fmt.Errorf("workflow must have at least one %q block", workflowOuterName)
+	}
+
+	return input, profile, output, nil
+}
+
+type WorkflowStore struct {
+	core   *Core
+	lock   sync.RWMutex
```
wslabosz-reply: let's make it anonymous?

vault/workflow_store.go
```diff
@@ -0,0 +1,342 @@
...
+	logger log.Logger
+}
+
+func NewWorkflowStore(c *Core) *WorkflowStore {
+	logger := c.baseLogger.Named("profile")
```
wslabosz-reply: engine is the profile, but logger should probably be called `workflow`?
```suggestion
	logger := c.baseLogger.Named("workflow")
```

vault/workflow_store.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+
+	if err := engine.Evaluate(noNsCtx); err != nil {
+		return nil, fmt.Errorf("failed to evaluate namespace: %w", err)
+	}
+
+	return nil, nil
+}
+
+func (ws *WorkflowStore) sanitizePath(path string) string {
```
wslabosz-reply, line 370: not sure how I feel with it being a `WorkflowStore` method

vault/logical_system_workflows.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+}
+
+// handleWorkflowsRead handles the "/sys/workflows/manage/<path>" endpoints to read a
+// workflow.
+func (b *SystemBackend) handleWorkflowsRead() framework.OperationFunc {
+	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+		path := data.Get("path").(string)
+
+		pe, err := b.Core.workflowStore.Get(ctx, path)
```
wslabosz-reply, line 327: unfortunately as `workflowStore.Get()` can return `nil, nil` you've got to check for that otherwise we are dereferencing nil in `createWorkflowDataResponse`

vault/logical_system_workflows.go
```diff
@@ -0,0 +1,392 @@
...
+		if err != nil {
+			return handleError(err)
+		}
+
+		return &logical.Response{Data: createWorkflowDataResponse(pe)}, nil
+	}
+}
+
+// handleWorkflowSet handles the "/sys/workflows/manage/<path>" endpoint to
```
wslabosz-reply:
```suggestion
// handleWorkflowsUpdate handles the "/sys/workflows/manage/<path>" endpoint to
```

vault/logical_system_workflows.go
```diff
@@ -0,0 +1,392 @@
...
+	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+		path := data.Get("path").(string)
+
+		err := b.Core.workflowStore.Delete(ctx, path)
+		return nil, err
+	}
+}
+
+// handleWorkflowsExecute handles the "/sys/workflow/execute/<path>" and
+// "sys/workflow/unauthed-execute/<path>" endpoint to execute workflows.
```
wslabosz-reply:
```suggestion
// "/sys/workflow/unauthed-execute/<path>" endpoints to execute workflows.
```

website/content/api-docs/system/workflows.mdx
```diff
@@ -0,0 +1,181 @@
...
+  `allow_unauthenticated_workflows` configuration option.
+
+- `cas_required` `(bool: false)` - Whether check-and-set semantics are
+  required to update this workflow.
+
+- `description` `(string: "")` - Textual description of this workflow for
+  operators.
+
+- `cas` `(int: nil)` - Last version number for modifications. Set to `-1`
+  for creation of a new workflow.
```
satoqz: Should it default to -1 then? Either way no big fan of `int: nil` 😄

website/content/api-docs/system/workflows.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+This endpoint deletes the workflow with the given path.
+
+| Method   | Path                          |
+| :------- | :---------------------------- |
+| `DELETE` | `/sys/workflows/manage/:path` |
+
+#### Parameters
+
+- `path` `(string: <required>)` -- Specifies the path of the workflow to delete.
```
satoqz:
```suggestion
- `path` `(string: <required>)` - Specifies the path of the workflow to delete.
```

website/content/api-docs/system/workflows.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+| Method | Path                                    | Requires Authentication | Full Request Information |
+| :----- | :-------------------------------------- | :---------------------- | :----------------------- |
+| `POST` | `/sys/workflows/execute/:path`          | yes                     | no                       |
+| `POST` | `/sys/workflows/unauthed-execute/:path` | no                      | no                       |
+| `POST` | `/sys/workflows/trace/:path`            | yes                     | yes                      |
+
+#### Parameters
+
+- `path` `(string: <required>)` -- Specifies the path of the workflow to create.
```
satoqz:
```suggestion
- `path` `(string: <required>)` - Specifies the path of the workflow to create.
```

changelog/2728.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:feature
+Support **Workflows** in Core. This adds new endpoints under `sys/workflows` to allow operators to create workflows and users to execute them.
+  - Workflows allow the creation of simplified or managed interfaces over OpenBao's standard API.
+  - Use of the `allow_unauthenticated_workflows` server configuration value enables unauthenticated execution of workflows; any dispatched requests still require authentication but this can be provided as a request parameter.
+  - Workflows are built on the common profile engine powering declarative self-initialization.
```
satoqz:
```suggestion
  - Workflows are built on the common profile engine powering declarative self-initialization and use the same syntax.
```

vault/workflow_store.go
```diff
@@ -0,0 +1,342 @@
...
+
+func (ws *WorkflowStore) List(ctx context.Context, prefix string, recursive bool, after string, limit int) ([]*WorkflowEntry, error) {
+	ns, err := namespace.FromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	prefix = ws.sanitizePath(prefix)
+	view := ws.getView(ns)
+	view = view.SubView(prefix)
```
wslabosz-reply:
```suggestion
	view := ws.getView(ns).SubView(prefix)
```

vault/logical_system.go
```diff
@@ -5654,4 +5655,43 @@ This path responds to the following HTTP methods.
...
+
+	GET /<path>
+		Retrieve a workflow.
+
+	PUT /<path>
+		Create or update a workflow.
+
+	PATCH /<path>
+		Update a workflow.
+
```
satoqz:
```suggestion
```

I don't see this implemented anywhere.

vault/logical_system.go
```diff
@@ -5654,4 +5655,43 @@ This path responds to the following HTTP methods.
...
+	"workflows": {
+		"Create, read, update and delete workflows.",
+		`
+This path responds to the following HTTP methods.
+
+	GET /<path>
+		Retrieve a workflow.
+
+	PUT /<path>
+		Create or update a workflow.
```
satoqz, line 5679: The documentation says `POST`... I know we handle both, but let's either name both in all places or name only one and the same in all places.

vault/workflow_store.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+
+	// Prefer permission denied for missing workflows when unauthenticated.
+	if unauthed && (workflow == nil || !workflow.AllowUnauthenticated) {
+		return nil, logical.ErrPermissionDenied
+	}
+
+	if workflow == nil {
+		return nil, nil
+	}
```
wslabosz-reply, line 293: I think it might lead to questions "I've run execute and nothing happen", shouldn't we return an error of "workflow doesn't exist", I guess we would be leaking existence of a workflow, but at this point aren't we sure that user has permission to run the thing?

vault/workflow_store.go
```diff
@@ -0,0 +1,342 @@
...
+
+	if output != nil {
+		return engine.EvaluateResponse(noNsCtx)
+	}
+
+	if err := engine.Evaluate(noNsCtx); err != nil {
+		return nil, fmt.Errorf("failed to evaluate namespace: %w", err)
+	}
+
+	return nil, nil
```
wslabosz-reply: not sure but I think this will translate to `204 No Content` on the response, where you describe the endpoint as returning `200 OK`

vault/workflow_store.go
satoqz, line 1: I realize this is a small component that may not warrant a package, but I do want to just _mention_ that @phyrog's direction has been to move all of these "stores" and "managers" to a subpackage. Up to you.

vault/logical_system_workflows.go
satoqz: I need a workflow for each of my PKI engines, duh?

vault/logical_system_workflows.go
```diff
@@ -0,0 +1,392 @@
...
+		"allow_unauthenticated": pe.AllowUnauthenticated,
+		"description":           pe.Description,
+	}
+}
+
+func createWorkflowDataResponse(pe *WorkflowEntry) map[string]any {
+	base := createWorkflowListResponse(pe)
+	base["workflow"] = pe.Workflow
+	return base
+}
```
satoqz:
```suggestion
func createWorkflowListResponse(we *WorkflowEntry) map[string]any {
	return map[string]any{
		"path":                  we.Path,
		"version":               we.Version,
		"cas_required":          we.CASRequired,
		"allow_unauthenticated": we.AllowUnauthenticated,
		"description":           we.Description,
	}
}

func createWorkflowDataResponse(we *WorkflowEntry) map[string]any {
	base := createWorkflowListResponse(we)
	base["workflow"] = we.Workflow
	return base
}
```

we!

vault/logical_system_workflows.go
```diff
@@ -0,0 +1,392 @@
...
+		after := data.Get("after").(string)
+		limit := data.Get("limit").(int)
+
+		workflows, err := b.Core.workflowStore.List(ctx, parent, scan, after, limit)
+		if err != nil {
+			return nil, err
+		}
+
+		var keys []string
+		keyInfo := make(map[string]interface{})
```
satoqz:
```suggestion
		keys := make([]string, 0, len(workflows))
		keyInfo := make(map[string]any, len(workflows))
```

vault/logical_system_workflows.go
```diff
@@ -0,0 +1,392 @@
...
+}
+
+// handleWorkflowsExecute handles the "/sys/workflow/execute/<path>" and
+// "sys/workflow/unauthed-execute/<path>" endpoint to execute workflows.
+func (b *SystemBackend) handleWorkflowsExecute(unauthed bool, trace bool) framework.OperationFunc {
+	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+		path := data.Get("path").(string)
+		if unauthed {
+			trace = false
+		}
```
satoqz, line 391: Is this necessary, or just a precaution so that we don't mis-encode this in the paths above?

vault/logical_system_workflows.go
cipherboy: @wslabosz-reply @satoqz We support arbitrary path to the workflow, so you can organize them in a nested pattern following the mount itself if one wants.

This is just a standard paginated list endpoint, though, I realize we don't have SCAN support for this either.

vault/logical_system_workflows.go
cipherboy: @wslabosz-reply I think this is due to the way the handler engine works, but I might be wrong. :thinking: Will test it out.

vault/workflow_store.go
cipherboy: Yeah, this was the pattern of the profile store. Thoughts?

website/content/api-docs/system/workflows.mdx
cipherboy: CAS is optional; only if `cas_required=true` is this parameter required. So you can interact with the endpoint without passing `cas`, so `nil` is correct I believe.

vault/logical_system.go
cipherboy: Ah, d'oh. Copied from policy store again :laughing:

vault/logical_system_workflows.go
cipherboy: Precaution; I could also err if we want.

vault/logical_system_workflows.go
cipherboy: Hmm, my find/replace was lazy 😆

vault/workflow_store.go
cipherboy: Ah yes, let's err in that case.

vault/workflow_store.go
cipherboy: Hmm that's fair. I wonder how we want to handle this. Most profiles probably will return something... 🤔

website/content/api-docs/system/workflows.mdx
satoqz: Do we have precedent on optional parameters that have no default value / are unset by default? `nil` is similarly Go-specific to me as the `text/template` case.

vault/logical_system_workflows.go
satoqz: Fine with me

website/content/api-docs/system/workflows.mdx
cipherboy: @satoqz I think it is notated differently, but this pattern as a whole was original to KVv2 as inherited from Vault: https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#cas

So will update docs to align.

vault/workflow_store.go
```diff
@@ -0,0 +1,341 @@
...
+			return nil, fmt.Errorf("failed to fetch workflow (%d/%v) in list: %w", index, path, err)
+		}
+
+		results = append(results, entry)
+	}
+
+	return results, nil
+}
+
+func (ws *WorkflowStore) Execute(ctx context.Context, path string, unauthed bool, trace bool, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
```
satoqz, line 272: Should we have protections against infinite recursion via workflows calling workflows? I think the stack blows up far sooner than context cancellation will propagate :D

vault/logical_system_workflows.go
cipherboy: @wslabosz-reply No, I think this is correct.

You can call `workflows/manage` on the `LIST` endpoint with no path. But you can't write to it because it needs a name (only in URL); that's why you can't combine the two regexes. 

But you can do `LIST workflows/manage/some/prefix` and see all workflows under that prefix.

vault/workflow_store.go
cipherboy: @wslabosz-reply @satoqz What do we think of removing this lock? It only guarantees consistency of list/scan operations w.r.t. in-flight deletes, which you already don't know where you're at there vs ordering, and a transaction can handle instead (`ScanView` transparently is transactional). 

Thoughts?

vault/logical_system.go
cipherboy: Can we address this later? A ton of endpoints in the system document `PUT` via help and `POST` in the website documentation.

vault/workflow_store.go
cipherboy: I think in general we'll probably leave packages for larger things? Policies warrant one because there's several files, `logical_system` probably warrants one, same with expiration and identity and seals. But not sure I'd move single file subsystems... yet?

vault/workflow_store.go
cipherboy: @satoqz Hmmm, interesting idea. I'm not quite sure how we'd handle that. We could perhaps inject something in the context to prevent the recursion? :thinking:

vault/workflow_store.go
wslabosz-reply: I see the need for such function, and as there already exists `sanitizePath` (with different logic) in `vault` package; you've added it as `*WorkflowStore` method, this works fine, I'd contemplate renaming and ripping it out from `*WorkflowStore` but probably as it is specially designed to sanitize path of the workflow, maybe it's best to leave it as it is.

vault/logical_system.go
satoqz: Sure

vault/workflow_store.go
satoqz: Context works but we'd need to make sure we mutate the inner value of the passed context rather than layering a new context on each stage to avoid creating the context chain of doom. I'd just pass a `*int` that increments in `Execute` or is initialized at 0 if not present, and force stops at 100 or so. Probably also needs to be transferred from `httpCtx` -> `ctx` in `switchedLockHandleRequest`.

vault/workflow_store.go
satoqz: Matter of taste I'd say. I like small packages... but that's why I really just wanted to mention the option.

vault/workflow_store.go
satoqz: @cipherboy well, generally I'd agree it should be possible to drop the lock entirely, except for one detail that might force us into keeping it:

There is CAS, which if you compare with `PolicyStore`'s implementation, should hold a lock over the entirety of the operation, which we don't yet do here.

I wonder if transactions would allow us to drop the lock around a CAS write? Would have to test it with Raft and Postgres.

vault/workflow_store.go
satoqz: Alternatively, we can consider blocking recursive workflow calls entirely if that's not within scope to begin with. Though that'd kill the upcoming sub-community focused on functional programming via OpenBao APIs.

vault/workflow_store.go
cipherboy: @satoqz Hmm, in this case, we'll have to still support a lock for non-transactional backends. With different read/write orderings, I think you could end up with a deadlock on PostgreSQL, which is why we left locking in certain places like PKI... It might work, but we'd have to test it more and probably also want to drop non-transactional backends first. 

@wslabosz-reply Thinking about anonymous lock, wouldn't that result in `.Lock()` &c being exposed on the `WorkflowStore` struct rather than being an internal implementation detail?

vault/workflow_store.go
cipherboy: Yeah, we'll refactor some of these helper methods later I'm sure.

vault/workflow_store.go
cipherboy: @satoqz said:

> Context works but we'd need to make sure we mutate the inner value of the passed context rather than layering a new context on each stage to avoid creating the context chain of doom.

Sadly I think we end up with this anyways as we have to deal with both removing the namespace before executing `switchedLockHandleRequest` and the conversion of `httpCtx -> ctx`. 

> Alternatively, we can consider blocking recursive workflow calls entirely if that's not within scope to begin with. Though that'd kill the upcoming sub-community focused on functional programming via OpenBao APIs.

I wonder if we care about either? It seems like it'd very clearly be on the operator (this is a rather privileged API after all) to avoid this. And currently noisy-neighbor isn't a solved problem w.r.t. namespaces... Thoughts?

vault/workflow_store.go
cipherboy: Let's revisit later then. Other than `switchedLockedHandleRequest` being a new dependency, most of the rest of the calls should be consistent IMO.

vault/workflow_store.go
satoqz: @cipherboy lock sharding like you added to `PolicyStore` is of course entirely thinkable.

cipherboy: @satoqz That's fair, let me add that.

vault/workflow_store.go
satoqz: > I wonder if we care about either? It seems like it'd very clearly be on the operator (this is a rather privileged API after all) to avoid this. And currently noisy-neighbor isn't a solved problem w.r.t. namespaces... Thoughts?

@cipherboy yes but this hard crashes OpenBao. It might not even be intentional and just a misconfiguration. We can argue that "privileged operators shouldn't misuse their privilege", but it's much better to automatically stop them from firing a footgun in the first place.

```hcl
flow "test" {
  request "boom" {
    operation = "update"
    path = "sys/workflows/execute/boom"
  }
}
```

```sh
$ bao write sys/workflows/manage/boom workflow=@boom.hcl
$ bao write -force sys/workflows/execute/boom
```

... wait a few seconds ...

server crashes irrecoverably.

cipherboy: @satoqz Never mind, I have a cute hack for this, will send along shortly in a separate PR. :-D

vault/workflow_store.go
wslabosz-reply: @cipherboy 
yes, that's the point (also @satoqz kinda pointed out that these are called embedded fields rather than anonymous).

cipherboy:
@satoqz @wslabosz-reply Ok, I think that's all the immediately actionable feedback.

https://github.com/openbao/openbao/pull/2804 has a commit I can apply here if preferred that adds one attempt at workflow recursion prevention and improves auditing of workflows, so I thought it was logically separate.

vault/workflow_store.go
cipherboy: @wslabosz-reply Sorry, why would we want to expose `Lock()` for use outside of the internal implementation of WorkflowStore's methods?

Issue #2729, opened 2026-03-26
Address EmberJS 5.0 deprecations in UI


JanMa:
Before we can upgrade EmberJS further past version `5.0.0` we need to address a lot of deprecations:

## Mandatory

These deprecations are a hard blocker for upgrading.
| ID                                                                                                                                           | Status                                           |
|----------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------|
| ["ember-data:deprecate-array-like"](https://deprecations.emberjs.com/id/ember-data-deprecate-array-like)                                     | ✔️  https://github.com/openbao/openbao/pull/2705  |
| ["ember-data:deprecate-early-static"](https://deprecations.emberjs.com/id/ember-data-deprecate-early-static)                                 | 🚧 https://github.com/openbao/openbao/pull/2813                                                 |
| ["ember-data:deprecate-model-reopen"](https://deprecations.emberjs.com/id/ember-data-deprecate-model-reopen)                                 | ✔️ https://github.com/openbao/openbao/pull/2759  |
| ["ember-data:deprecate-model-reopenclass"](https://deprecations.emberjs.com/id/ember-data-deprecate-model-reopenclass)                       | ✔️ https://github.com/openbao/openbao/pull/2732   |
| ["ember-data:deprecate-promise-proxies"](https://deprecations.emberjs.com/id/ember-data-deprecate-promise-proxies)                           | ✔️ https://github.com/openbao/openbao/pull/2730   |
| ["ember-data:no-a-with-array-like"](https://deprecations.emberjs.com/id/ember-data-no-a-with-array-like)                                     | ✔️  https://github.com/openbao/openbao/pull/2770  |
| ["ember-data:deprecate-promise-many-array-behaviors"](https://deprecations.emberjs.com/id/ember-data-deprecate-promise-many-array-behaviors) | ✔️ https://github.com/openbao/openbao/pull/2768  |
| ["ember-data:deprecate-has-record-for-id"](https://deprecations.emberjs.com/id/ember-data-deprecate-has-record-for-id)                       |  ✔️ https://github.com/openbao/openbao/pull/2778 |
| ["setting-on-hash"](https://deprecations.emberjs.com/id/setting-on-hash)                                                                     | ✔️ https://github.com/openbao/openbao/pull/2811                                               |

## Nice to have

These are third-party EmberJS addon deprecations which would be nice to get rid of but are not required.
  
[...]

PR #2730, opened 2026-03-26
address ember-data-deprecate-promise-proxies deprecation

JanMa:
## Description

Make sure to await a PromiseObject before accessing it's methods.
Luckily for us there was only a single route left that didn't do this properly.

Part-Of: #2729

## Acknowledgements

       also allowed.


ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
```diff
@@ -263,15 +263,15 @@ export default Route.extend({
         throw err;
       }
     }
-    await capabilities;
+    const resolvedCapabilities = await capabilities;
     if (modelType === 'secret-v2') {
       // after the the base model fetch, kv-v2 has a second associated
       // version model that contains the secret data
       secretModel = await this.fetchV2Models(capabilities, secretModel, params);
```
satoqz: Should this take `resolvedCapabilities`?

JanMa: It seems this function never uses the `capabilities` argument that is passed in :smile: Let me refactor that

ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
```diff
@@ -263,15 +260,15 @@ export default Route.extend({
...
     if (modelType === 'secret-v2') {
       // after the the base model fetch, kv-v2 has a second associated
       // version model that contains the secret data
-      secretModel = await this.fetchV2Models(capabilities, secretModel, params);
+      secretModel = await this.fetchV2Models(secretModel, params);
     }
     return {
       secret: secretModel,
-      capabilities,
+      resolvedCapabilities,
```
satoqz:
```suggestion
      capabilities: resolvedCapabilities,
```

I assume this shouldn't be renamed?

JanMa: I _think_ this returns an anonymous object (I'm using the UI modernization as a means to better understand JS), so it doesn't make a difference which name we use here.

satoqz: It's not clear to me where this is accessed later on, but I was thinking that other code would later access this object using `.capabilities`, but we're breaking that by returning an object with a different key (`resolvedCapabilities`) here.

This syntax:

```js
return {
  resolvedCapabilities,
}
```

Is a shorthand for:

```js
return {
  resolvedCapabilities: resolvedCapabilities,
}
```

...which is not equal to

```js
return {
  capabilities: resolvedCapabilities,
}
```

PR #2731, opened 2026-03-26
website: enable shell-session syntax highlighting

### LINKED ISSUE #1768: `shell-session` tags on website don't do anything, actually
https://github.com/openbao/openbao/pull/1767#discussion_r2321207934 made me think:

"Wait, if `shell-session` is specific to Docusaurus... then why are none of the code blocks on the website actually highlighted?"

_Welp._

For starters, we don't actually load the `shell-session` Prism parser, which is [not enabled by default](https://github.com/FormidableLabs/prism-react-renderer/blob/e1c83a468b05df7f452b3ad7e4ae5ab874574d4e/packages/generate-prism-languages/index.ts#L9-L26). This is easily resolved like so:

```patch
diff --git a/website/docusaurus.config.ts b/website/docusaurus.config.ts
index 44452bdde0..f6c192b409 100644
--- a/website/docusaurus.config.ts
+++ b/website/docusaurus.config.ts
@@ -189,7 +189,7 @@ const config: Config = {
     prism: {
       theme: prismThemes.github,
       darkTheme: prismThemes.dracula,
-      additionalLanguages: ["hcl"],
+      additionalLanguages: ["hcl", "shell-session"],
     },
     metadata: [
       { name: 'keywords', content: 'openbao, secrets management, open source, linux foundation, encryption as a service, key management system, pki, transit, ssh, secret vault, database passwords' },
```

No big deal, right? This does indeed get the parser to work, and if you inspect the HTML Docusaurus generates now, `shell-session` codeblocks have different CSS classes per token. For reference, via inspect element:

Image

---

Nice. So why do I still see no color? Turns out the `prism-react-renderer` themes that we use:

- Light mode: https://github.com/FormidableLabs/prism-react-renderer/blob/e1c83a468b05df7f452b3ad7e4ae5ab874574d4e/packages/prism-react-renderer/src/themes/github.ts
- Dark mode: https://github.com/FormidableLabs/prism-react-renderer/blob/e1c83a468b05df7f452b3ad7e4ae5ab874574d4e/packages/prism-react-renderer/src/themes/dracula.ts

support a bunch of generic token types, but not the ones we get for `shell-session`, as seen above :D
Hence highlighting for e.g. `hcl` works, but not `shell-session`.

---

Presumably the solution is to extend the themes and add support for the tokens generated by `shell-session`. The grammar is defined here: https://github.com/PrismJS/prism/blob/v2/src/languages/shell-session.js.

We could try to upstream this, alternatively it's probably also possible to extend the theme locally.

anik7afk:
## Summary
- Load the `shell-session` Prism language parser by adding it to `additionalLanguages`
- Extend both the GitHub (light) and Dracula (dark) Prism themes with styles for `shell-symbol` and `output` tokens, which are not covered by the default themes
- Token colors are drawn from each theme's existing palette so they feel native (e.g., Dracula green for the prompt symbol, Dracula comment-gray for output)

## Details
The `shell-session` code blocks on the website were not being highlighted because:
1. The `shell-session` Prism parser was not loaded (it is not enabled by default in `prism-react-renderer`)
2. Even after loading, the GitHub and Dracula themes don't include styles for the `shell-session`-specific token types (`shell-symbol`, `important`, `output`)

This PR fixes both issues by adding a small `withShellSessionTokens()` helper that extends the base themes with the missing token styles.

Fixes #1768

## Test plan
- [ ] Build the website locally (`npm run build` in `website/`)
- [ ] Verify `shell-session` code blocks show syntax highlighting (prompt symbol in color, command highlighted via bash grammar, output in muted color)
- [ ] Verify both light and dark mode themes render correctly
- [ ] Verify existing `hcl` code block highlighting is unaffected

### DIFF
--- a/changelog/2731.txt
+++ b/changelog/2731.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+website: Enable shell-session syntax highlighting in code blocks
+```


satoqz:
I'm inclined to start closing PRs by first-time contributors that fail to consider CONTRIBUTING.md and the PR template without warning from now on, like I will with this one. In particular, fully ignoring the PR template only proves to me that zero human thought went into the creation of a pull request.

Taking copyright out of the picture, there is no value in purely LLM-generated contributions with so little human oversight that basic guidelines for contribution are ignored. If we liked the approach of agents automatically taking issues, this could be taken over by maintainers that will provide better instructions and local review passes, or by a dedicated bot.

Personally, I'm not willing to politely engage with and foster this new "scan for good first issue label and one-shot prompt a patch to farm GitHub karma" culture. Even if well-intended, it lacks any respect, both for the humans on the maintainer team, and for yourself who has chosen to be entirely replaced and represented by a fancy random number generator. There are ways to use this technology respectfully and productively; this is not one of them.

cc @openbao/openbao-org-maintainers

PR #2735, opened 2026-03-27
Namespace sealing mount storage refactor

wslabosz-reply:
## Description
- updates `auth.go` and `mount.go` mounts reading/writing logic to use barrier of the respective namespace
- updates legacy mount system to split mounts per namespace depending on the namespace "ownership"
- adjusts legacy mount invalidation logic

Introduced changes mostly based on #2204 PR 

## Rationale
Introduces writes/reads of mounts through namespace barriers.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/logical_system.go
+++ b/vault/logical_system.go
@@ -1856,11 +1856,15 @@ func (b *SystemBackend) handleTuneWriteCommon(ctx context.Context, path string, 
 		}
 	}
 
-	var err error
+	ns, err := namespace.FromContext(ctx)
+	if err != nil {
+		return handleError(err)
+	}
+
 	if isAuth {
-		err = b.Core.persistAuth(ctx, nil, b.Core.auth, &mountEntry.Local, mountEntry.UUID)
+		err = b.Core.persistAuth(ctx, b.Core.NamespaceView(ns), b.Core.auth, &mountEntry.Local, mountEntry.UUID)
 	} else {
-		err = b.Core.persistMounts(ctx, nil, b.Core.mounts, &mountEntry.Local, mountEntry.UUID)
+		err = b.Core.persistMounts(ctx, b.Core.NamespaceView(ns), b.Core.mounts, &mountEntry.Local, mountEntry.UUID)
 	}
 
 	if err != nil {
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -118,41 +118,49 @@ func (c *Core) generateMountAccessor(entryType string) (string, error) {
 	return accessor, nil
 }
 
-// DecodeMountTable is used for testing
-func (c *Core) DecodeMountTable(ctx context.Context, raw []byte) (*routing.MountTable, error) {
-	return c.decodeMountTable(ctx, raw)
-}
-
-func (c *Core) decodeMountTable(ctx context.Context, raw []byte) (*routing.MountTable, error) {
-	// Decode into mount table
-	mountTable := new(routing.MountTable)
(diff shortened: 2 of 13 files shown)


vault/mount.go
```diff
@@ -118,41 +118,45 @@ func (c *Core) generateMountAccessor(entryType string) (string, error) {
...
+	for _, entry := range entries {
+		// Decode into single mount table
+		mountTable := new(routing.MountTable)
+		if err := jsonutil.DecodeJSON(entry.Value, mountTable); err != nil {
+			return nil, 0, err
+		}
+		size += size + len(entry.Value)
+
+		// TODO(wslabosz): figure out if this makes sense.
+		mountType = mountTable.Type
```
cipherboy, line 137: I think this is probably fine. The likelihood of mixing entries like this seems low, though if we wanted to we could always guard properly.

vault/mount.go
```diff
@@ -805,32 +803,26 @@ func (c *Core) moveStorage(ctx context.Context, src namespace.MountPathDetails,
...
-			if err != nil {
+		if err := logical.WithTransaction(ctx, srcBarrier, func(s logical.Storage) error {
+			se, err := s.Get(ctx, entryKey)
+			if err != nil || se == nil {
 				return err
 			}
-			err = s.Delete(ctx, srcPrefix+key)
-			if err != nil {
+
+			if err := logical.WithTransaction(ctx, dstBarrier, func(s logical.Storage) error {
```
cipherboy: Single puts don't need to be in a transaction.

wslabosz-reply: fixed

vault/mount.go
wslabosz-reply: added guard clause proper.

vault/seal_manager.go
```diff
@@ -178,6 +178,10 @@ func (sm *SealManager) RemoveNamespace(ns *namespace.Namespace) {
 // Remember that this method does not take in an existing storage type and is
 // likely wrong to call within the context of a transaction.
 func (c *Core) NamespaceView(ns *namespace.Namespace) barrier.View {
+	// Root namespace fallback.
+	if ns == nil {
+		ns = namespace.RootNamespace
+	}
```
satoqz: What made this necessary? On one hand, sure, we avoid panics, on the other hand, we don't want to retrieve the wrong barrier ever.

vault/logical_system.go
```diff
@@ -1856,11 +1856,15 @@ func (b *SystemBackend) handleTuneWriteCommon(ctx context.Context, path string,
 		}
 	}
 
-	var err error
+	ns, err := namespace.FromContext(ctx)
+	if err != nil {
+		return handleError(err)
+	}
```
satoqz, line 1862: I really wish we had just made this a panicking operation...

wslabosz-reply: well, that would be one way to address it for sure

vault/seal_manager.go
wslabosz-reply: There is a`TestCore_MountEntryView` in `mount_test.go`, which had a test case of a mount entry without `Namespace` being populated, this resulted in a panic when `(*Core).mountEntryView()` was called. I've added a fallback in the `(*Core).NamespaceView()` but it definitely wasn't the right call, mount entries should always have a `Namespace` populated, so I've now updated the test (removed the "no namespace" test cases). I think it was made obsolete with the change to the storage introduced with this PR.

PR #2737, opened 2026-03-27
Revise profile use of ctx, documentation

cipherboy:
## Description

Based on review feedback in #2671 and #2727, this:

1. Revises the introductory documentation substantially.
2. Removes `ctx` from some unnecessary places, refactoring tests as well.
3. Adds support for setting headers on requests.

## Rationale

This enables the profile system to work with inline auth, which is important or useful for unauthenticated profiles.

cc: @satoqz @wslabosz-reply

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/concepts/profiles.mdx
+++ b/website/content/docs/concepts/profiles.mdx
@@ -11,8 +11,8 @@ entry point. This can be parsed from HCL or JSON as desired.
 
 Data sources are dependent on the profile call site: self-initialization has
 one set of allowed sources whereas API profiles have a different combination.
-Data sources can be nested; for instance, a `text/template` source can be used
-to combine an environment variable and the contents of a file.
+Data sources can be nested; for instance, a template source can be used to
+combine an environment variable and the contents of a file.
 
 :::info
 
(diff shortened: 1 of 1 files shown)


website/content/docs/concepts/profiles.mdx
```diff
@@ -99,42 +192,45 @@ For example, to reference an initial admin password from an environment
 variable:
 
 ```hcl
-initialize "identity" {
```
satoqz, line 102: A follow-up I'd suggest is to rework the self-init doc page. Previously, this page was effectively the better documentation for self-init by giving many more practical examples. The main page boasts a single example that doesn't even work anymore because we disabled API-driven audits by default.

website/content/docs/concepts/profiles.mdx
```diff
@@ -1,12 +1,18 @@
...
-The system implements the request/response pattern, supporting chained operations and accepting request handlers and parsed profile configurations.
-These are then transformed into templates for one or more API requests.
+The Profiles engine processes declarative API requests with dynamic data
+source evaluation. The system implements a request/response pattern,
+supporting chained operations, allowing multiple API calls from a single
+entry point. This can be parsed from HCL or JSON as desired.
+
+Data sources are dependent on the profile call site: self-initialization has
+one set of allowed sources whereas API profiles have a different combination.
+Data sources can be nested; for instance, a `text/template` source can be used
```
satoqz: Do we really want to call it `text/template` source or just "Template"? I feel as though this is confusing for non-Go programmers.

website/content/docs/concepts/profiles.mdx
cipherboy: Agreed, I'll work on that in a little bit.

website/content/docs/concepts/profiles.mdx
cipherboy: Fair point. Revised, thanks!

PR #2739, opened 2026-03-27
Add conditional execution to profiles

cipherboy:
## Description

Addition of the `when` keyword allows variability in profiles,
conditionally skipping requests which may be known to fail or which
are unnecessary based on various dynamic data. With the use of
`text/template` and CEL now added to profiles, this becomes especially
powerful.

## Rationale

While less necessary for self-init, there has been some discussion around how to handle conditional items. For API profiles though, I'd strongly suspect this would be of more interest.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2739.txt
+++ b/changelog/2739.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+self-init: Allow conditional execution of requests with `when` keyword.
+```


satoqz:
image

cipherboy:
@satoqz You're not wrong. But in the future, the same engine will be used for driving cross-plugin communication eventing, so some of these will be useful there as well, IMHO, and will look more like IFTTT rather than Ansible. ;-)

satoqz:
@cipherboy I didn't mean to criticize at all :D This is perfectly reasonable.

cipherboy:
@satoqz Looping will be my next pull request :laughing:

cipherboy:
Just want to make sure I comment on UX here: `when` is not a CEL expression, but another "value" expression that must be a boolean; so you can select between:

```hcl
{
  when = false
} 
```

or:

```hcl
{
  when = {
    eval_type = "bool"
    eval_source="template"
    template="{{ .my-value | truncate_sha256 }}"
  }
}
```

or similar.

satoqz (approved):
I agree with the linter, LGTM otherwise.

helper/profiles/profiles.go
```diff
@@ -370,7 +377,7 @@ func (p *ProfileEngine) evaluateRequest(ctx context.Context, history *Evaluation
 
 // buildRequest transforms an input configuration's request into a proper
 // output
-func (p *ProfileEngine) buildRequest(ctx context.Context, history *EvaluationHistory, outerIndex int, outerBlock *OuterConfig, requestIndex int, requestBlock *RequestConfig) (req *logical.Request, allowFailure bool, err error) {
+func (p *ProfileEngine) buildRequest(ctx context.Context, history *EvaluationHistory, outerIndex int, outerBlock *OuterConfig, requestIndex int, requestBlock *RequestConfig) (req *logical.Request, execute bool, allowFailure bool, err error) {
```
satoqz, line 380: It might be worth returning a struct here if the returns with many variables became too messy :^)

PR #2759, opened 2026-03-30
Remove reopen from attach-capabilities.js

tmccombs:
## Description

Refactor attach-capabilities, so that it doesn't use `reopen` or `reopenClass`.

This involves creating a new `capability` decorator  to track fields for related capabilities, and storing metadata on the class for how to dynamically generate the paths. And I changed the `relatedCapabilities` method to be a standalone function that can use that metadata to add relationships.

## Rationale

This is part of  #2729  for ["ember-data:deprecate-model-reopen"](https://deprecations.emberjs.com/id/ember-data-deprecate-model-reopen). 

It isn't (yet) all use cases of reopen, but figured I would start a draft with what I have so far. 

It's a bit more of a refactor than I originally attended. But also moves away from some "classic" syntax that I hope will help with upgrading it further.

## Acknowledgements

       also allowed.


ui/app/lib/capabilities.js
tmccombs, line 1: Maybe this file should be renamed? I don't know.

ui/app/lib/capabilities.js
```diff
@@ -4,82 +4,92 @@
...
-      const newData = data.map(this.relatedCapabilities);
-      return {
-        data: newData,
-        included,
-      };
-    }
-    const context = {
-      id: data.id,
-      ...data.attributes,
+export function addCapabilityRelationships(capabilities, jsonAPIDoc) {
```
tmccombs, line 60: Potentially this could take the Model class as an argument instead of the capabilities itself, and then look up the capabilities on the class. I wasn't entirely sure which way would be better.

ui/app/lib/attach-capabilities.js
```diff
@@ -4,82 +4,92 @@
...
+ * add capabilities relationships to the JSON-API document used by Ember Data
  */
-export default function attachCapabilities(modelClass, capabilities) {
-  const capabilityKeys = Object.keys(capabilities);
-  const newRelationships = capabilityKeys.reduce((ret, key) => {
-    ret[key] = belongsTo('capabilities', { async: true, inverse: null });
-    return ret;
-  }, {});
-
-  //TODO: move this to the application serializer and do it JIT instead of on app boot
```
tmccombs, line 41: I'm not really sure what was intended by this TODO, or if it is still relevant. 

Perhaps, instead of having anything for this in the model, the field is just marked as `@belongsTo('capabilities', { async: true, inverse: null })`, and we have a helper function in to add relationships that we call from the serializer for identity/entity and auth-method?

If this TODO is worth doing, I could make that part of this.

tmccombs (commented):
Would it be better to have a single PR for all the reopen changes, or split it up into smaller PRs?

JanMa:
> Would it be better to have a single PR for all the reopen changes, or split it up into smaller PRs?

I'm fine with a single PR :-)

ui/app/lib/capabilities.js
JanMa: Agreed, I think something like `capabilities.js` might fit better now.

ui/app/lib/attach-capabilities.js
JanMa: I don't think that TODO is still relevant. At best you could move `addCapabilityRelationships` out into a serializer, but I think having it in this file is nicer for readability.

ui/app/lib/capabilities.js
JanMa: I think passing `capabilities`  directly is the cleaner approach, otherwise you'd first have to check again if `model.relatedCapabilities` exists.

ui/app/deprecation-workflow.js
```diff
@@ -10,6 +10,7 @@ export const deprecationWorkflowConfig = {
   // deprecations that will not be removed until 5.0.0 are filtered by deprecation-filter initializer rather than silencing below
   workflow: [
     { handler: 'log', matchId: 'ember-data:deprecate-array-like' },
+    { handler: 'throw', matchId: 'ember-data:deprecate-model-reopen' },
```
JanMa:
```suggestion
    { handler: 'log', matchId: 'ember-data:deprecate-model-reopen' },
```

You don't need to use `throw` here, `log` is enough. I introduced a deprecation handler to forward the deprecation log to stdout and cause a test failure
https://github.com/openbao/openbao/blob/c85d2ce93654e57d2e7778d0ab7dae5b661509b9/ui/tests/test-helper.js#L19-L36

changelog/2759.txt
JanMa, line 1: You don't need to add a changelog entry for these refactorings in the next ones. We'll just tag the PR with `pr/no-changelog` :-)

JanMa (changes requested):
One tiny remark, otherwise LGTM. Thanks @tmccombs

changelog/2759.txt
tmccombs: ok, thanks

JanMa (approved):
LGTM, thanks @tmccombs

PR #2761, opened 2026-03-30
Add baseline dependency groups

wslabosz-reply:
## Description
Introduces dependabot groups, grouping updates of:
- golang.org/x/*
- k8s.io/*
- google.golang.org/*

dependencies.

## Rationale
Added dependency group should lower the number of open dependabot PRs every week, while also ensuring that these specific group shouldn't conflict within each one.

Resolves: no issue

## Acknowledgements

       also allowed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -34,6 +34,16 @@ updates:
       semver-patch-days: 7
       semver-minor-days: 14
       semver-major-days: 30
+    groups: 
+      golang-x:
+        patterns:
+          - "golang.org/x/*"
+      k8s:
+        patterns:
+          - "k8s.io/*"
+      google-golang:
+        patterns:
+          - "google.golang.org/*"
     ignore:
       # armon/go-metrics was moved to hashicorp/go-metrics, but we cannot
       # drop the indirect dependency on it just yet. This causes errors in


.github/dependabot.yml
```diff
@@ -34,6 +34,28 @@ updates:
       semver-patch-days: 7
       semver-minor-days: 14
       semver-major-days: 30
+    groups: 
+      golang-x:
+        applies-to: version-updates
```
wslabosz-reply: not sure if I should add the `security-updates` there also?

.github/dependabot.yml
```diff
@@ -34,6 +34,28 @@ updates:
...
+        patterns:
+          - "@golang.org/x/*"
+        update-types:
+          - "minor"
+          - "patch"
```
satoqz: Do we need to specify these explicitly?

.github/dependabot.yml
satoqz: Can we leave the field out to automatically apply to everything?

.github/dependabot.yml
wslabosz-reply: I think not? unless we want to have major updates separately, if we don't care about those, then not.

.github/dependabot.yml
wslabosz-reply: omitting the field will default to just version-updates

image

.github/dependabot.yml
satoqz: We weren't setting anything before, so I think it's okay to drop these. If we ever get to a new major version on, let's say, Kubernetes, we'd more likely than not want to upgrade in one batch.

.github/dependabot.yml
satoqz: Let's keep the default then.

.github/dependabot.yml
wslabosz-reply: so if we *want* major version upgrades in separate PRs (ungrouped) then we have to keep these listed separately

satoqz: We don't want them in separate PRs.

PR #2771, opened 2026-03-31
Switch website to pnpm

satoqz:
## Description

This switches the OpenBao website to use `pnpm` instead of `npm`. This is nearly a drop-in replacement as evident by the minimal changes required in `website/Makefile`.

## Rationale

In light of today's [compromise of the axios package](https://www.stepsecurity.io/blog/axios-compromised-on-npm-malicious-versions-drop-remote-access-trojan) which deployed a trojan that triggers via post-install scripts, I do no longer trust myself to use the `npm install` command without accidentally compromising my environment. `npm` runs post-install scripts by default, while `pnpm` blocks all post-install scripts and runs only those whitelisted (see `pnpm-workspace.yaml`). Additionally, `pnpm` is a nice performance boost over `npm` but security concerns are what really motivate me to ultimately make this switch. I believe this is the better choice to keep both maintainers and contributors safe from compromise.

## Acknowledgements

       also allowed.

### DIFF
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -37,7 +37,6 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5.0.0
         with:
-          version: latest
           package_json_file: website/package.json
           cache: true
           cache_dependency_path: website/pnpm-lock.yaml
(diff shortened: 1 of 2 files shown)


website/package.json
wslabosz-reply, line 1: not sure what caused the addition of these packages?

satoqz: See the description of my second commit.

satoqz:
@wslabosz-reply done

satoqz:
Thanks everyone! Let's hope I got the deploy workflow right...

space-aditya:
@satoqz, do you think this might be needed too? 
https://github.com/openbao/openbao/pull/2791

space-aditya:
https://github.com/openbao/openbao/pull/2792
and related linting changes since i tested the scripts

PR #2789, opened 2026-04-02
Apply several "go fix" analyzers

satoqz:
## Description

This is the result of:

```sh
for mod in $(find . -name go.mod) ; do
  (cd "$(dirname "$mod")" && go fix -minmax -fmtappendf -forvar -mapsloop -rangeint -slicescontains -slicessort -stringscut -stringscutprefix ./...)
done
make fmt
```

_twice_.

This diff should be reproducible via the above commands if made from the same head commit (094e5f6957afedc34d1ae4224fc4a88969af404b).

```console
$ git diff | sha256sum -
11b17ba5af58e53541207e68a0d55486fa07ce7e6420d79d532ca7b92dfd7ad8  -

$ curl -sL https://github.com/openbao/openbao/pull/2789.diff | sha256sum -
11b17ba5af58e53541207e68a0d55486fa07ce7e6420d79d532ca7b92dfd7ad8  -
```

Note that it takes 2 passes of the above to reach a stable hash, the second pass gets rid of one extra unnecessary loop variable.

I've omitted several analyzers that would lead to a very large diff (running all of them at once leads to a -8000 diff :D). Optimally we'd get all analyzers besides `any` out of the way, and keep that one for last, with a diff that's easy to automatically verify.

For a list of all passes, see https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize.

## Rationale

I'd like to get the codebase "modernized", i.e., apply Go modernizers until we can't no more, and then add a CI check to ensure this remains the case. This is step 1!

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/logical/kubernetes/client.go
+++ b/builtin/logical/kubernetes/client.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	maps0 "maps"
 	"strings"
 	"time"
 
@@ -263,9 +264,7 @@ func makeRoleType(roleType string) string {
 func combineMaps(maps ...map[string]string) map[string]string {
 	newMap := make(map[string]string)
 	for _, m := range maps {
-		for k, v := range m {
-			newMap[k] = v
-		}
+		maps0.Copy(newMap, m)
 	}
 	return newMap
 }
--- a/builtin/logical/kubernetes/integrationtest/helpers.go
+++ b/builtin/logical/kubernetes/integrationtest/helpers.go
@@ -6,6 +6,7 @@ package integrationtest
 import (
 	"context"
 	"fmt"
+	maps0 "maps"
 	"math/rand"
 	"os"
 	"strings"
@@ -405,9 +406,7 @@ func testK8sTokenAudiences(t *testing.T, expectedAudiences []interface{}, token 
 func combineMaps(maps ...map[string]string) map[string]string {
 	newMap := make(map[string]string)
 	for _, m := range maps {
-		for k, v := range m {
-			newMap[k] = v
-		}
+		maps0.Copy(newMap, m)
 	}
 	return newMap
 }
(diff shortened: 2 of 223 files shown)


builtin/logical/kubernetes/integrationtest/helpers.go
```diff
@@ -6,6 +6,7 @@ package integrationtest
 import (
 	"context"
 	"fmt"
+	maps0 "maps"
```
wslabosz-reply, line 9: why `maps0` is there a naming conflict?

builtin/logical/kubernetes/client.go
```diff
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	maps0 "maps"
```
wslabosz-reply, line 10: maps0

builtin/logical/kubernetes/integrationtest/helpers.go
satoqz: Yep there is, the affected function takes a `maps` parameter.

satoqz: I'd kind of like to avoid making manual fixups in this patch (at least in the initial review phase) to avoid making it harder to automatically verify.

builtin/logical/pki/acme_wrappers.go
```diff
@@ -354,11 +349,8 @@ func getAcmeRoleAndIssuer(sc *storageContext, data *framework.FieldData, config
 		if !allowAnyRole {
 
 			var foundRole bool
-			for _, name := range config.AllowedRoles {
-				if name == role.Name {
-					foundRole = true
-					break
-				}
+			if slices.Contains(config.AllowedRoles, role.Name) {
```
cipherboy, line 352: Out of scope, but I wonder if we should collapse this further to `foundRole := ...`

satoqz: Yeah, might want to do a later manual pass and fix up some of the rough edges we've observed here.

PR #2794, opened 2026-04-02
detect transaction leaks in inmem storage

phil9909:
## Description

If the leak is detected while running a test (via `go test` as reported by [`testing.Testing()`](https://pkg.go.dev/testing#Testing)) instead of logging we will `panic`.

For the inmem backend I keep to severity of the log at error even for the 2nd and all following leaks (for raft we decided to decrease to debug after the first error).

## Rationale

See #2185

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2794.txt
+++ b/changelog/2794.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+inmem: detect, log, and rollback transactions that have never been committed or rolled-back. If you see the message "transaction was leaked" in your logs, please open an issue.
+```
--- a/sdk/physical/inmem/inmem.go
+++ b/sdk/physical/inmem/inmem.go
@@ -8,10 +8,13 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"runtime"
+	"runtime/debug"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
+	"testing"
 
 	"github.com/armon/go-radix"
 	log "github.com/hashicorp/go-hclog"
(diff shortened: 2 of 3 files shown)


changelog/2794.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+inmem: detect, log, and rollback transactions that have never been committed or rolled-back. If you see the message "transaction was leaked" in your logs, please open an issue.
```
satoqz, line 2: nit:

```suggestion
sdk/physical/inmem: Detect, log, and rollback transactions that have never been committed or rolled-back. If you see the message "transaction was leaked" in your logs, please open an issue.
```

I think `raft` deserves top-level scope as it spans a few components, but inmem is a bit niche so let's use full path.

cipherboy: Do we care? I think other than dev server, inmem is mostly only used in tests. :thinking:

satoqz: It is in SDK so we might want to document changes out of principle.

sdk/physical/inmem/inmem.go
```diff
@@ -156,9 +161,24 @@ func NewInmem(conf map[string]string, logger log.Logger) (physical.Backend, erro
 		return b, nil
 	}
 
+	leakedTransactionHook := func(args []any) {
+		logger.Error("transaction was leaked", args...)
```
cipherboy, line 165:
```suggestion
		logger.Error("transaction was leaked", "stack", args)
```

Realized I forgot to look to see this is the case on #2185. vargs are for JSON formatting, not array support.

cipherboy: #2185 is good, but this one needs to be adjusted btw @phil9909.

Issue #2762, opened 2026-03-30
Support in-memory versions of ClientCert & ClientKey in api.TLSConfig


satoqz:
**Is your feature request related to a problem? Please describe.**

`api.TLSConfig` is declared as follows:

```go
// TLSConfig contains the parameters needed to configure TLS on the HTTP client
// used to communicate with Vault.
type TLSConfig struct {
	// CACert is the path to a PEM-encoded CA cert file to use to verify the
	// Vault server SSL certificate. It takes precedence over CACertBytes
	// and CAPath.
	CACert string

	// CACertBytes is a PEM-encoded certificate or bundle. It takes precedence
	// over CAPath.
	CACertBytes []byte

	// CAPath is the path to a directory of PEM-encoded CA cert files to verify
	// the Vault server SSL certificate.
	CAPath string

	// ClientCert is the path to the certificate for Vault communication
	ClientCert string

	// ClientKey is the path to the private key for Vault communication
	ClientKey string

	// Other fields...
```

Notably, it supports passing a custom CA cert in-memory via `CACertBytes`, but does not allow the same for the client cert and key, forcing them to be loaded from the filesystem. I have been unable to find a way around this as configuring the API client's TLS settings is a one-way road towards immutability.

**Describe the solution you'd like**

I would like to be able to pass client cert & key in-memory, just like is possible for the CA cert via `CACertBytes`. The new fields would likely be `ClientCertBytes` and `ClientKeyBytes` and use PEM encoding.

**Explain any additional use-cases**

Avoiding the file system is necessary when cert parameters must be controlled by operators that do not and should not have access the the filesystem of a machine, but need to pass custom certs via an API.

cipherboy:
@satoqz I view improved client auth favorably, happy to take contributions in this direction.

PR #2798, opened 2026-04-02
Support in-memory versions of ClientCert & ClientKey in api.TLSConfig

wslabosz-reply:
## Description
Introduces `ClientCertBytes` and `ClientKeyBytes` as `api.TLSConfig` struct fields  for in-memory passing.

No e2e were conducted due to (possible) high complexity.
(Initially unit tests were generated using Claude Opus 4.6, but heavily reworked before pushing).

## Rationale
(From the issue):
> Avoiding the file system is necessary when cert parameters must be controlled by operators that do not and should not have access the the filesystem of a machine, but need to pass custom certs via an API.

Resolves: #2762 

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2798.txt
+++ b/changelog/2798.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+api: Add `ClientCertBytes` and `ClientKeyBytes` as possible in-memory cert contents in `TLSConfig`.
+```


satoqz:
@wslabosz-reply I'm not sure we should add env vars - seems a little over the top - I'm happy with the fields being available on `TLSConfig` for programmatic consumption in Go.

cipherboy:
@satoqz Not to make @wslabosz-reply redo the work again, but why not for consistency?

satoqz:
My bad, I had missed that `EnvVaultCACertBytes` already exists. Well, I hate it, but it's probably better to keep it consistent then.

wslabosz-reply:
@satoqz @cipherboy 
commit still exists, I can revert and bring them back

api/client.go
```diff
@@ -257,26 +259,35 @@ type Config struct {
...
 
-	// CACertBytes is a PEM-encoded certificate or bundle. It takes precedence
-	// over CAPath.
+	// ClientCert is the path to the certificate for OpenBao communication.
+	ClientCert string
+
+	// ClientKey is the path to the private key for OpenBao communication.
+	ClientKey string
+
+	// CACertBytes is in-memory stored PEM-encoded certificate or bundle.
```
satoqz, line 279:
```suggestion
	// CACertBytes is an in-memory PEM-encoded certificate or bundle.
```

api/client.go
```diff
@@ -257,26 +259,35 @@ type Config struct {
...
 	CACertBytes []byte
 
-	// CAPath is the path to a directory of PEM-encoded CA cert files to verify
-	// the Vault server SSL certificate.
-	CAPath string
+	// ClientCertBytes is in-memory stored PEM-encoded certificate or bundle.
```
satoqz, line 282:
```suggestion
	// ClientCertBytes is an in-memory PEM-encoded certificate or bundle for
	// OpenBao communication.
```

api/client.go
```diff
@@ -257,26 +259,35 @@ type Config struct {
...
+	ClientCertBytes []byte
 
-	// ClientCert is the path to the certificate for Vault communication
-	ClientCert string
+	// ClientKeyBytes is in-memory stored PEM-encoded key.
```
satoqz, line 285:
```suggestion
	// ClientKeyBytes is an in-memory PEM-encoded key for OpenBao communication.
```

api/client.go
```diff
@@ -257,26 +259,35 @@ type Config struct {
...
+	ClientKeyBytes []byte
 
-	// ClientKey is the path to the private key for Vault communication
-	ClientKey string
+	// CAPath is the path to a directory of PEM-encoded CA cert files to verify
+	// the OpenBao server SSL certificate.
```
satoqz, line 289:
```suggestion
	// CAPath is the path to a directory of PEM-encoded CA cert files to verify
	// the OpenBao server's certificate.
```

Let's avoid the term "SSL".

api/client.go
```diff
@@ -335,39 +346,52 @@ func DefaultConfig() *Config {
 }
 
 // configureTLS is a lock free version of ConfigureTLS that can be used in
-// ReadEnvironment where the lock is already hold
+// ReadEnvironment where the lock is already hold.
```
satoqz, line 349:
```suggestion
// ReadEnvironment where the lock is already held.
```

While we're at it.

api/client.go
```diff
@@ -257,26 +259,35 @@ type Config struct {
...
 
 // TLSConfig contains the parameters needed to configure TLS on the HTTP client
-// used to communicate with Vault.
+// used to communicate with OpenBao.
 type TLSConfig struct {
+	// Order or precedence of loading certs is as follows:
+	// - CACert (filepath) -> CACertBytes (inmem) -> CAPath (directory)
+	// - ClientCert (filepath) -> ClientCertBytes (inmem)
+	// - ClientKey (filepath) -> ClientKeyBytes (inmem)
+
```
satoqz, line 268:
```suggestion
//
// The order or precedence for loading certificates & keys is as follows:
//  - For the CA cert: CACert -> CACertBytes -> CAPath
//  - For the client certificate: ClientCert -> ClientCertBytes
//  - For the client key: ClientKey -> ClientKeyBytes
type TLSConfig struct {
```

Otherwise this will be swallowed on pkg.go.dev

api/client.go
```diff
@@ -335,39 +346,52 @@ func DefaultConfig() *Config {
...
+		if err != nil {
+			return err
+		}
+		foundClientCert = true
+		c.curlClientCert = "passed-in-memory"
+		c.curlClientKey = "passed-in-memory"
 	case t.ClientCert != "" || t.ClientKey != "":
 		return errors.New("both client cert and client key must be provided")
+	case len(t.ClientCertBytes) != 0 || len(t.ClientKeyBytes) != 0:
+		return errors.New("both client cert and client key pem bundles must be provided")
```
satoqz, line 380:
```suggestion
	case t.ClientCert != "" || t.ClientKey != "", len(t.ClientCertBytes) != 0 || len(t.ClientKeyBytes) != 0:
		return errors.New("both client cert and client key must be provided")
```

Let's make the error the same - they're PEM bundles in either case.

api/client.go
```diff
@@ -335,39 +346,52 @@ func DefaultConfig() *Config {
...
 	}
 
 	if t.CACert != "" || len(t.CACertBytes) != 0 || t.CAPath != "" {
 		c.curlCACert = t.CACert
+		if t.CACert == "" {
+			c.curlCACert = "passed-in-memory"
+		}
```
satoqz, line 387: IMO this should remain empty if not passed as a filepath.

api/client.go
```diff
@@ -335,39 +346,52 @@ func DefaultConfig() *Config {
...
 		c.curlCAPath = t.CAPath
 		rootConfig := &certConfig{
 			CAFile:        t.CACert,
 			CACertificate: t.CACertBytes,
 			CAPath:        t.CAPath,
 		}
-		if err := configureTLS(clientTLSConfig, rootConfig); err != nil {
+		if err = configureTLS(clientTLSConfig, rootConfig); err != nil {
```
satoqz, line 394:
```suggestion
		if err := configureTLS(clientTLSConfig, rootConfig); err != nil {
```

I think it's better to keep error scopes where the language lets us confidently do so.

satoqz:
Oh. I had 10 comments queued... will open a follow-up.

api/client_test.go
```diff
@@ -364,37 +368,41 @@ func TestClientEnvSettings(t *testing.T) {
...
+	require.NoError(t, os.Setenv(EnvVaultCACert, cwd+"/test-fixtures/keys/cert.pem"))
+	require.NoError(t, os.Setenv(EnvVaultCACertBytes, string(caCertBytes)))
+	require.NoError(t, os.Setenv(EnvVaultCAPath, cwd+"/test-fixtures/keys"))
+	require.NoError(t, os.Setenv(EnvVaultClientCert, cwd+"/test-fixtures/keys/cert.pem"))
+	require.NoError(t, os.Setenv(EnvVaultClientCertBytes, string(oldClientCertBytes)))
+	require.NoError(t, os.Setenv(EnvVaultClientKey, cwd+"/test-fixtures/keys/key.pem"))
+	require.NoError(t, os.Setenv(EnvVaultClientKeyBytes, string(oldClientKeyBytes)))
+	require.NoError(t, os.Setenv(EnvVaultSkipVerify, "true"))
+	require.NoError(t, os.Setenv(EnvVaultMaxRetries, "5"))
+	require.NoError(t, os.Setenv(EnvVaultDisableRedirects, "true"))
```
satoqz, line 393: We should update this to use `t.Setenv`.

satoqz:
I'll collapse some of the proposed cleanup into a PR for https://github.com/openbao/openbao/issues/2820.

api/client_test.go
cipherboy: @satoqz Tbh, I wonder if we should drop the whole Setenv thing, but I suppose we want to test precedence order.

cipherboy:
@satoqz Ah, sorry. GitHub needs a lock mechanism if maintainers have opened the PR for review and have pending comments. 😆

api/client.go
cipherboy: @satoqz On the one hand I agree with you, though on the other, I wonder if it a useful sentinel for the curl command not being precise.

Issue #2725, opened 2026-03-26
Consider removing gox dependency


DrDaveD:
Part of the openbao build process installs https://github.com/mitchellh/gox.  That has been archived since 2023 so I think we should consider removing the dependency.  It appears to only be used in `builtin/logical/kv/scripts/build.sh`.

cipherboy:
@DrDaveD We don't use `build.sh` at all in kv, so I think it is fine to remove from the RPM distribution. 

We can probably remove the build.sh file as well.

PR #2801, opened 2026-04-02
Remove unnecessary plugin scripts

cipherboy:
## Rationale

Resolves: #2725 

## Acknowledgements

       also allowed.


cipherboy:
@wslabosz-reply I think we should migrate anything we're missing, yes. But some of the integration tests are inherent to the system, e.g., KV is already well integrated in other parts of the test suite, but downstream integration (e.g., with real/mocked k8s) should probably be added. 

Want to file a ticket for each area and see if we can get interest?

PR #2806, opened 2026-04-03
Simplify request context chaining

cipherboy:
## Rationale

While we want to avoid leaking most of the context placed in the `httpCtx`, we need to also ensure our context gets cancelled on the first of either http request cancellation or the active context cancellation. Using `AfterFunc` to bind the two avoids a long-lived goroutine while keeping the hierarchy as it is today.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/request_handling.go
+++ b/vault/request_handling.go
@@ -594,9 +594,7 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
 	ctx, cancel := context.WithCancel(c.activeContext)
 	defer cancel()
 
-	stop := context.AfterFunc(httpCtx, func() {
-		cancel()
-	})
+	stop := context.AfterFunc(httpCtx, cancel)
 	defer stop()
 
 	// A namespace was manually passed to HandleRequest, as can be the case with:


vault/request_handling.go
```diff
@@ -587,14 +587,17 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
-		case <-httpCtx.Done():
-			cancel()
-		}
-	}(ctx, httpCtx)
+	defer cancel()
+
+	stop := context.AfterFunc(httpCtx, func() {
+		cancel()
+	})
+	defer stop()
```
satoqz: Hold my confusing syntax:

```suggestion
	defer context.AfterFunc(httpCtx, cancel)()
```

I think this is probably cleanest though:

```suggestion
	stop := context.AfterFunc(httpCtx, cancel)
	defer stop()
```

vault/request_handling.go
```diff
@@ -587,14 +587,17 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
-		select {
-		case <-ctx.Done():
-		case <-httpCtx.Done():
-			cancel()
-		}
-	}(ctx, httpCtx)
+	defer cancel()
+
+	stop := context.AfterFunc(httpCtx, func() {
+		cancel()
```
cipherboy: This cancel still gets technically executed in a separate goroutine, but it is much shorter lived than the longer-lived goroutine. And most of the time we'll probably call `stop` before calling `AfterFunc`, except when we have a server-side long-lived request.

JavierLimon:
The stress test I'm running are passing!
Nice work!

cipherboy:
@satoqz @JanMa Do we think this requires a changelog and warrants a backport? :thinking:

satoqz:
I still doubt that this actually has much of an impact beyond having a couple less idle goroutines. IIUC, @wslabosz-reply had reported that the mentioned tests were already running fine against main before.

cipherboy:
> that the mentioned tests were already running fine against main before.

@satoqz OP's tests were for the jobmanager which this didn't change, but the load test screenshot showed increasing goroutines. I think you can probably reproduce similar with `benchmark-openbao`, though I don't have a good testing harness with metrics (something I wish I had added to `devbao` tbh).

JavierLimon:
Adding some info:

I reproduce the error in a small raspberry cluster with one remote machine running openbao-benchmark 8k with 90% writes and 10% this eventually start to leave orphan go routines, the leader node sometimes collapse or/and sometimes splits leadership, but the next one is the next node, until all nodes collapse, I still don't know very well what heppens at the end as I need to restart the hardware as it becomes irresponsible.

After applying the fix here this does not happened, once the openbao benchmark it lowers the go routines to the normal state, every time.

let me know if I can help with more information to reproduce it.

satoqz:
It would be good to find out why goroutines leak on main then since I don't see a direct connection between this patch and a leak being fixed. If we attach a release note in the form of "fixed goroutine leak under high load", it would be good to at least understand the root cause, not just observing a change in behavior. Nevertheless, given the supporting evidence, I'm not against backporting.

cipherboy:
@satoqz Sorry, not sure I follow.

I'm not arguing that there's a _leak_ of goroutines, just that we _incur_  an extra goroutine per request. This has been true since `fork-point`, and is easily reproducible:

```
$ benchmark-openbao run -config docs/examples/kv/kvv2-read.hcl -duration 60s -workers 10000
```

and

```
$ watch 'bash -c "bao read -format=json sys/metrics | grep -C 4 -i goroutines"'
```


Steady state looks like:

```
      {
       	"Labels": {},
        "Name": "vault.runtime.num_goroutines",
        "Value": 508
      },
```

but under load:

```
      {
       	"Labels": {},
        "Name": "vault.runtime.num_goroutines",
        "Value": 27613
      },
```

and with the patch applied:

```
      {
       	"Labels": {},
        "Name": "vault.runtime.num_goroutines",
        "Value": 17895
      },
```

which is a 1x decrease per worker, caused by having fewer background goroutines from the channel binding logic that `fork-point` to `main` has. 

As soon as the request stops, these go away, so I think I'm comfortable not calling it a leak.

satoqz:
@cipherboy I'm talking about a leak as @JavierLimon is implying so:

> this eventually start to leave orphan go routines

I totally agree that the expected outcome of this is to simply reduce goroutine allocations per request.

It wasn't clear to me whether we're standing by "let's backport this because it fixes a supposed but not entirely proven bug" or "let's backport because it is a clear performance improvement". The latter is fine, I would've liked to see more analysis if we were to go for the former.

cipherboy:
@satoqz Ah, no, I'm convinced this PR isn't a leak. I think it's just 2:1 vs 1:1 goroutine-to-request resource consumption drop. 

And from local testing, other than goroutine count, the rest of the performance metrics look the same between the existing and new code.

cipherboy:
I'll merging this without a changelog; we won't backport this. The memory/CPU consumption change is minimal, so this is just a decrease in internal goroutine contention mostly.

PR #2813, opened 2026-04-06
No early static

tmccombs:
## Description

Fix deprecation of ["ember-data:deprecate-early-static"](https://deprecations.emberjs.com/id/ember-data-deprecate-early-static)

This doesn't fix everywhere, but it's a start at least. I probably won't be able to get to finishing it for a little while, but if someone else wants to pick it up, go for it.

Part of #2729 

## Acknowledgements

       also allowed.

### DIFF
--- a/ui/app/services/path-help.js
+++ b/ui/app/services/path-help.js
@@ -10,6 +10,7 @@
 */
 import Model from '@ember-data/model';
 import Service from '@ember/service';
+import { inject as service } from '@ember/service';
 import { encodePath } from 'vault/utils/path-encoding-helpers';
 import { getOwner } from '@ember/application';
 import { assign } from '@ember/polyfills';
@@ -31,6 +32,7 @@ export function sanitizePath(path) {
 export default Service.extend({
   attrs: null,
   dynamicApiPath: '',
+  store: service(),
   ajax(url, options = {}) {
     const appAdapter = getOwner(this).lookup(`adapter:application`);
     const { data } = options;
@@ -41,9 +43,8 @@ export default Service.extend({
 
   getNewModel(modelType, backend, apiPath, itemType) {
     const owner = getOwner(this);
-    const modelName = `model:${modelType}`;
 
-    const modelFactory = owner.factoryFor(modelName);
+    const modelFactory = owner.factoryFor(`model:${modelType}`);
     let newModel, helpUrl;
     // if we have a factory, we need to take the existing model into account
     if (modelFactory) {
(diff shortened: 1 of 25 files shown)


ui/app/services/path-help.js
```diff
@@ -301,7 +302,8 @@ export default Service.extend({
 
   registerNewModelWithProps(helpUrl, backend, newModel, modelName) {
     return this.getProps(helpUrl, backend).then((props) => {
-      const { attrs, newFields } = combineAttributes(newModel.attributes, props);
+      const origAttrs = this.store.getSchemaDefinitionService().attributesDefinitionFor({ type: modelName });
```
tmccombs: This whole thing should probably be redesigned to use a custom SchemaDefinitionsService to handle the dynamic attributes, but that is probably a big refactor and I'm not entirely sure how that would work (I haven't found great documentation for it). 

But for now, this fixes this deprecation.

ui/app/services/path-help.js
```diff
@@ -301,7 +301,14 @@ export default Service.extend({
 
   registerNewModelWithProps(helpUrl, backend, newModel, modelName) {
     return this.getProps(helpUrl, backend).then((props) => {
-      const { attrs, newFields } = combineAttributes(newModel.attributes, props);
+      const schemas = this.store.getSchemaDefinitionService();
+      let origAttrs;
+      if (schemas.doesTypeExist(modelName)) {
+        origAttrs = schemas.attributesDefinitionFor({ type: modelName });
```
tmccombs, line 307: turns out this doesn't work the way I hoped.  Unfortunately, calling this populates the cache used by the store for the model class. And then the unregister/register at the end of this doesn't actually update the store's cache, so it continues using the old model. 

I'm not really sure what the best way to handle this is. The cleanest solution is probably to create a custom SchemaService, but I'm not really sure how to do that here in a way that takes advantage of existing model classes and merges that with dynamic schema information. And that also seems like a pretty big refactor. 

If anyone has any good ideas on how best to handle this deprecation, I'm open to them.

JanMa: I've taken a look at it and it turns out we can clean the cache after unregistering the model.

```diff
@@ -354,6 +354,9 @@ export default Service.extend({
       const regName = `model:${modelName}`;
       owner.unregister(regName);
       owner.register(regName, newModel);
+      delete schemas._attributesDefCache[modelName];
+      delete schemas._relationshipsDefCache[modelName];
+      delete this.store._modelFactoryCache[modelName];
     });
   },
   getFieldGroups(meta) {
```

This makes all tests pass when I run them locally. It's not ideal since this accesses private properties, but I think implementing a custom schema service is something that can wait until we've upgraded to Ember JS 5.x

ui/app/services/path-help.js
```diff
@@ -344,32 +351,36 @@ export default Service.extend({
         },
       });
       newModel.merged = true;
-      owner.unregister(modelName);
-      owner.register(modelName, newModel);
+      const regName = `model:${modelName}`;
+      owner.unregister(regName);
+      owner.register(regName, newModel);
```
JanMa, line 356:
```suggestion
      owner.register(regName, newModel);
      delete schemas._attributesDefCache[modelName];
      delete schemas._relationshipsDefCache[modelName];
      delete this.store._modelFactoryCache[modelName];
```

JanMa (changes requested):

JanMa (approved):
Taking the liberty to merge this because my proposed changes worked and all the test pass now. Thanks a lot for your work @tmccombs

PR #2836, opened 2026-04-08
Add Namespace Unsealing

wslabosz-reply:
## Description
- added [logical_system_namespaces_seals.go‎](https://github.com/openbao/openbao/pull/2836/changes/6c77d883892115374daf470b7da209826f0ecd2b#diff-a9b63d42c00ac8767e7754a43f927cdc488c23c693907016b0624dd1ff5feb64) with (for now) namespace unseal endpoint
- added load/setup(Mounts/Credentials)ForNamespace methods
- added namespace unseal handling logic in `namespace_store.go`

## Rationale
Implements namespace unsealing logic in similar way as the root namespace (core) unseal semantics. 

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -552,6 +552,19 @@ func (c *Core) loadCredentials(ctx context.Context, standby bool) error {
 	return nil
 }
 
+// loadCredentialsForNamespace is invoked as part of postNamespaceUnseal
+// to load the auth mounts of a namespace.
+func (c *Core) loadCredentialsForNamespace(ctx context.Context, ns *namespace.Namespace) error {
+	c.authLock.Lock()
+	defer c.authLock.Unlock()
+
+	// Check if we're on a non-transactional storage
+	if _, ok := c.barrier.(logical.TransactionalStorage); !ok {
+		return c.loadLegacyCredentialsForNamespace(ctx, ns)
+	}
+	return c.loadTransactionalCredentialsForNamespace(ctx, ns)
+}
+
 // loadTransactionalCredentials reads the transactional split auth (credential)
 // table, populates the storage if there are no existing entries.
 func (c *Core) loadTransactionalCredentials(ctx context.Context, barrier logical.Storage, standby bool) error {
@@ -1070,6 +1083,32 @@ func (c *Core) setupCredentials(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+// setupCredentialsForNamespace is invoked after we've loaded auth mounts
+// of a namespace to initialize the credential backends and update the router.
+func (c *Core) setupCredentialsForNamespace(ctx context.Context, ns *namespace.Namespace) ([]func(), error) {
+	c.authLock.Lock()
+	defer c.authLock.Unlock()
+
+	postUnsealFuncs := make([]func(), 0)
+	for _, entry := range c.auth.SortEntriesByPath().Entries {
+		// Only process entries with matching namespace ID
+		if entry.NamespaceID != ns.ID {
+			continue
+		}
+
+		postUnsealFunc, err := c.setupCredential(ctx, entry)
+		if err != nil {
+			return postUnsealFuncs, err
+		}
+
+		if postUnsealFunc != nil {
+			postUnsealFuncs = append(postUnsealFuncs, postUnsealFunc)
+		}
+	}
+
+	return postUnsealFuncs, nil
 }
 
 // setupCredential initializes the credential backend
--- a/vault/logical_system_namespaces_seals.go
+++ b/vault/logical_system_namespaces_seals.go
@@ -1,0 +1,178 @@
+// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package vault
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/openbao/openbao/helper/namespace"
+	"github.com/openbao/openbao/sdk/v2/framework"
+	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/openbao/openbao/vault/barrier"
+)
+
+func (b *SystemBackend) namespaceSealPaths() []*framework.Path {
+	sealStatusSchema := map[string]*framework.FieldSchema{
+		"type": {
+			Type:     framework.TypeString,
+			Required: true,
+		},
+		"initialized": {
+			Type:     framework.TypeBool,
+			Required: true,
+		},
+		"sealed": {
+			Type:     framework.TypeBool,
+			Required: true,
+		},
+		"t": {
+			Type:     framework.TypeInt,
+			Required: true,
+		},
+		"n": {
+			Type:     framework.TypeInt,
+			Required: true,
+		},
+		"progress": {
+			Type:     framework.TypeInt,
+			Required: true,
+		},
+		"nonce": {
+			Type:     framework.TypeString,
+			Required: true,
+		},
+	}
+
+	return []*framework.Path{
+		{
+			Pattern: "namespaces/(?P<path>.+)/unseal",
+			DisplayAttrs: &framework.DisplayAttributes{
+				OperationPrefix: "namespaces",
+			},
+			Fields: map[string]*framework.FieldSchema{
+				"path": namespacePathSchema,
+				"key": {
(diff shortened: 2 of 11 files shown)


vault/auth.go
```diff
@@ -762,6 +729,66 @@ func (c *Core) loadLegacyCredentials(ctx context.Context, barrier logical.Storag
...
+	}
+
+	if entry != nil {
+		mEntries, size, err := c.decodeMountTable(ctx, entry)
+		if err != nil {
+			c.logger.Error("failed to decompress and/or decode the legacy auth table", "error", err)
+			return err
+		}
+		c.auth.Entries = append(c.auth.Entries, mEntries...)
+		// TODO(wslabosz): metrics
```
wslabosz-reply: there are two (x2 mirrored by `mount.go`) occurrences of reporting legacy mount table size. With currently proposed changes, it's no longer easy to summarize all the namespaces tables to get the "final" table size (across all namespaces).

Should I:
- remove this metric altogether (legacy mount system)
- make it report per namespace
- write code summing the size over all namespaces

cipherboy: Per namespace seems best.

I've opted for removal as I don't see the value in having those on the `loadLegacyMounts/Credentials()` level. Transactional equivalent does not report the size of the mounts, but we do that during `persistMount/Auth`, which should be called on any change in underlying storage.

cipherboy: @wslabosz-reply That's fine, the only comment I'd make is that loading is more frequent than writing, so in some systems the metric may disappear.

that's fair, but we do not recommend using non-transactional storage right?
so in 99% (made up number) cases we still do not report the metric on load, just on change; we only did that for legacy mounts, which well, maybe let's create an open issue asking if that metric even makes sense from community pov, and if so we probably should revisit its reporting logic for transactional storage.

cipherboy: Right and with plans to deprecated and remove, I think it's fine not to care too much longer.

wslabosz-reply: created https://github.com/openbao/openbao/issues/2852

vault/mount.go
```diff
@@ -119,49 +119,32 @@ func (c *Core) generateMountAccessor(entryType string) (string, error) {
...
-			return nil, 0, errors.New("broken mount table encountered; mismatched table types between namespaces")
-		}
-		mountType = mountTable.Type
-
-		// Populate the namespace in memory
-		for _, entry := range mountTable.Entries {
-			if entry.NamespaceID == "" {
-				entry.NamespaceID = namespace.RootNamespaceID
-			}
+func (c *Core) decodeMountTable(ctx context.Context, entry *logical.StorageEntry) ([]*routing.MountEntry, error) {
```
phyrog:
```suggestion
func (c *Core) decodeMountEntries(ctx context.Context, entry *logical.StorageEntry) ([]*routing.MountEntry, error) {
```

vault/mount.go
```diff
@@ -1454,124 +1449,137 @@ func (c *Core) persistMounts(ctx context.Context, barrier logical.Storage, table
...
-				entry.RunningVersion = versions.GetBuiltinVersion(consts.PluginTypeSecrets, entry.Type)
-			}
+	// update the entry running version with the configured
+	// version, which was verified during registration.
+	entry.RunningVersion = entry.Version
+	if entry.RunningVersion == "" && entry.RunningSha256 == "" {
+		// don't set the running version to a builtin if it is running as an external plugin
+		entry.RunningVersion = versions.GetBuiltinVersion(consts.PluginTypeSecrets, entry.Type)
+	}
+
```
phyrog, line 1534:
```suggestion
var skipSetCoreBackend bool

```

vault/mount.go
```diff
@@ -1454,124 +1449,137 @@ func (c *Core) persistMounts(ctx context.Context, barrier logical.Storage, table
...
+	if versions.IsBuiltinVersion(entry.RunningVersion) {
+		_, err := c.handleDeprecatedMountEntry(ctx, entry, consts.PluginTypeSecrets)
+		if c.isMajorVersionFirstMount(ctx) && err != nil {
+			go c.ShutdownCoreError(fmt.Errorf("could not mount %q: %w", entry.Type, err))
+			return nil, errLoadMountsFailed
+		} else if err != nil {
+			c.logger.Error("skipping deprecated mount entry", "name", entry.Type, "path", entry.Path, "error", err)
+			backend.Cleanup(ctx)
+			backend = nil
+			goto ROUTER_MOUNT
```
phyrog:
```suggestion
			skipSetCoreBackend = true
```

Let's get rid of the goto.

vault/mount.go
```diff
@@ -1454,124 +1449,137 @@ func (c *Core) persistMounts(ctx context.Context, barrier logical.Storage, table
...
-			_, err := c.handleDeprecatedMountEntry(ctx, entry, consts.PluginTypeSecrets)
-			if c.isMajorVersionFirstMount(ctx) && err != nil {
-				go c.ShutdownCoreError(fmt.Errorf("could not mount %q: %w", entry.Type, err))
-				return errLoadMountsFailed
-			} else if err != nil {
-				c.logger.Error("skipping deprecated mount entry", "name", entry.Type, "path", entry.Path, "error", err)
-				backend.Cleanup(ctx)
-				backend = nil
-				goto ROUTER_MOUNT
+	{
```
phyrog:
```suggestion
if !skipSetCoreBackend {
```

vault/mount.go
```diff
@@ -1454,124 +1449,137 @@ func (c *Core) persistMounts(ctx context.Context, barrier logical.Storage, table
...
-			backendType := backend.Type()
+		c.setCoreBackend(entry, backend, view)
+	}
 
-			if backendType != logical.TypeLogical {
-				if err := knownMountType(entry.Type); err != nil {
-					return err
-				}
-			}
+ROUTER_MOUNT:
```
phyrog:
```suggestion
```

cipherboy: @phyrog I think this actually applies to #2854 -- but while I like the idea of removing the GOTO, should we keep refactoring it into more helper methods rather than keeping the `if { ... }` flow? :thinking:

vault/mount.go
```diff
@@ -904,6 +904,19 @@ func (c *Core) loadMounts(ctx context.Context, standby bool) error {
...
+// load the mounts of a namespace.
+func (c *Core) loadMountsForNamespace(ctx context.Context, ns *namespace.Namespace) error {
+	c.mountsLock.Lock()
+	defer c.mountsLock.Unlock()
+
+	// Check if we're on a non-transactional storage
+	if _, ok := c.barrier.(logical.TransactionalStorage); !ok {
+		return c.loadLegacyMountsForNamespace(ctx, ns)
+	}
+	return c.loadTransactionalMountsForNamespace(ctx, ns)
```
cipherboy, line 909: @wslabosz-reply I think my comment on #2854 does actually apply then -- we need to ensure we don't duplicate entries ever. Sealing _should_ remove entries for a namespace but I'd hate for `loadMountsForNamespace` to be called twice and chaos to ensue afterwards.

cipherboy (commented):
One nit on this one.

vault/mount.go
phyrog: Oops, yes, I wanted to add the review there instead of here 😄 I'll resolve these comments here and add them in the other PR

wslabosz-reply:
@phyrog resolved your comments as they are addressed in https://github.com/openbao/openbao/pull/2854

vault/logical_system_namespaces_seals.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	return []*framework.Path{
+		{
+			Pattern: "namespaces/(?P<path>.+)/unseal",
+			DisplayAttrs: &framework.DisplayAttributes{
+				OperationPrefix: "namespaces",
+			},
+			Fields: map[string]*framework.FieldSchema{
+				"path": namespacePathSchema,
```
phyrog, line 60: Should this be called `path` or `name`?

wslabosz-reply: I think we want to change that to `name` but for now I want to keep it consistent across all namespace operations and stay with `path`, and change it all in one PR.

vault/mount.go
wslabosz-reply: `loadMountsForNamespace` is only invoked as a part of `postNamespaceUnseal` which right now can only happen on manual unseal of the namespace, said namespace have had to be sealed first, so all its mounts were removed from memory.

cipherboy (approved):

PR #2851, opened 2026-04-10
agent: allow changing the uid and gid of a file sink

### LINKED ISSUE #2850: Allow changing the ownership of the bao token created by a file sink

When using the file sink from the openbao agent, it is currently only possible to change the mode of the resulting token file but not the ownership.
This is an issue if the service that requires the token isn't running as root or as the same user as the openbao-agent as the service won't have permissions to read the token.

The file sink config should accept a `uid` and `gid` option so that the ownership of the token file can be changed.

Something like:
```hcl
sink {
  type = "file"
  config = {
    path = "/tmp/token"
    uid  = 1001
    gid  = 1001
    mode = 0660
  }
}
```

The only "work-around" I've found so far was setting the mode to `0666` but that doesn't seem like a secure solution.

jon4hz:
## Description

This PR introduces a new config option for the file sink of the openbao agent to change the UID and GID of the token file that the agent creates.

This makes it possible to create token files that can be read by services that are running as unprivileged users.

## Rationale

Resolves: #2850

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2851.txt
+++ b/changelog/2851.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+agent: Add uid and gid configuration options for the file sink
+```
--- a/changelog/2851.txt
+++ b/changelog/2851.txt
@@ -1,3 +1,3 @@
-```release-note:feature
+```release-note:improvement
 agent: Add uid and gid configuration options for the file sink
 ```
--- a/command/agentproxyshared/sink/file/file_sink.go
+++ b/command/agentproxyshared/sink/file/file_sink.go
@@ -19,6 +19,8 @@ import (
 type fileSink struct {
 	path   string
 	mode   os.FileMode
+	uid    int
+	gid    int
 	logger hclog.Logger
 }
 
@@ -33,6 +35,8 @@ func NewFileSink(conf *sink.SinkConfig) (sink.Sink, error) {
 	f := &fileSink{
 		logger: conf.Logger,
 		mode:   0o640,
+		uid:    -1,
+		gid:    -1,
 	}
 
 	pathRaw, ok := conf.Config["path"]
@@ -59,6 +63,28 @@ func NewFileSink(conf *sink.SinkConfig) (sink.Sink, error) {
 
 		f.logger.Debug("overriding default file sink", "mode", mode)
 		f.mode = os.FileMode(mode)
+	}
+
+	if uidRaw, ok := conf.Config["uid"]; ok {
+		f.logger.Debug("verifying override for default file sink uid")
+		uid, typeOK := uidRaw.(int)
+		if !typeOK {
+			return nil, errors.New("could not parse 'uid' as integer")
+		}
+
+		f.logger.Debug("overriding default file sink", "uid", uid)
+		f.uid = uid
+	}
+
+	if gidRaw, ok := conf.Config["gid"]; ok {
+		f.logger.Debug("verifying override for default file sink gid")
+		gid, typeOK := gidRaw.(int)
+		if !typeOK {
+			return nil, errors.New("could not parse 'gid' as integer")
+		}
+
+		f.logger.Debug("overriding default file sink", "gid", gid)
+		f.gid = gid
 	}
 
 	if err := f.WriteToken(""); err != nil {
@@ -119,6 +145,11 @@ func (f *fileSink) WriteToken(token string) error {
 			return fmt.Errorf("error removing temp file %s during write check: %w", tmpFile.Name(), err)
 		}
 		return nil
+	}
+
+	// adjust file ownership if the sink has a none-default uid or gid configuration
+	if err := os.Chown(tmpFile.Name(), f.uid, f.gid); err != nil {
+		return fmt.Errorf("error changing file ownership of %s to uid %d and gid %d: %w", tmpFile.Name(), f.uid, f.gid, err)
 	}
 
 	err = os.Rename(tmpFile.Name(), f.path)
--- a/command/agentproxyshared/sink/file/file_sink_test.go
+++ b/command/agentproxyshared/sink/file/file_sink_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 	"testing"
 
 	hclog "github.com/hashicorp/go-hclog"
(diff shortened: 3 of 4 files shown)


command/agentproxyshared/sink/file/file_sink_test.go
```diff
@@ -133,3 +145,71 @@ func TestFileSinkMode(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	path := filepath.Join(tmpDir, "token")
+
+	config := &sink.SinkConfig{
+		Logger: log.Named("sink.file"),
+		Config: map[string]interface{}{
+			"path": path,
+			"uid":  os.Getuid(),
```
jon4hz, line 157: Small remark regarding the test here:

This *technically* doesn't test the chown as `os.Getuid()` and `os.Getgid()` will return the ids of the user that created the file in the first place. I'm not sure how to test this in a better way without making unrealistic assumptions about the environment where the test is running. If anyone has a better idea, please let me know.

changelog/2851.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
cipherboy:
```suggestion
```release-note:improvement
```

command/agentproxyshared/sink/file/file_sink_test.go
cipherboy: For what we're doing and the code called, I think this is fine enough.

command/agentproxyshared/sink/file/file_sink_test.go
```diff
@@ -133,3 +145,71 @@ func TestFileSinkMode(t *testing.T) {
 		t.Fatalf("expected %s, got %s", uuidStr, string(fileBytes))
 	}
 }
+
+func testFileSinkChown(t *testing.T, log hclog.Logger) (*sink.SinkConfig, string) {
+	tmpDir, err := os.MkdirTemp("", fmt.Sprintf("%s.", fileServerTestDir))
```
satoqz: Please use `t.TempDir()`.

command/agentproxyshared/sink/file/file_sink_test.go
```diff
@@ -133,3 +145,71 @@ func TestFileSinkMode(t *testing.T) {
 		t.Fatalf("expected %s, got %s", uuidStr, string(fileBytes))
 	}
 }
+
+func testFileSinkChown(t *testing.T, log hclog.Logger) (*sink.SinkConfig, string) {
```
satoqz, line 149: Does this need to be a separate function?

command/agentproxyshared/sink/file/file_sink_test.go
```diff
@@ -133,3 +145,71 @@ func TestFileSinkMode(t *testing.T) {
...
+	tmpDir, err := os.MkdirTemp("", fmt.Sprintf("%s.", fileServerTestDir))
+	if err != nil {
+		t.Fatal(err)
+	}
```
satoqz: Can you use the `testify/require` package to simplify assertions? There are lots of examples across the codebase, just grep for `require`.

command/agentproxyshared/sink/file/file_sink.go
```diff
@@ -61,6 +65,28 @@ func NewFileSink(conf *sink.SinkConfig) (sink.Sink, error) {
...
+	if gidRaw, ok := conf.Config["gid"]; ok {
+		f.logger.Debug("verifying override for default file sink gid")
+		gid, typeOK := gidRaw.(int)
+		if !typeOK {
+			return nil, errors.New("could not parse 'gid' as integer")
+		}
+
+		f.logger.Debug("overriding default file sink", "gid", gid)
+		f.gid = gid
+	}
```
satoqz, line 88: These need documentation updates in `website/`.

satoqz (commented):
Looks good, just a few asks:

command/agentproxyshared/sink/file/file_sink_test.go
jon4hz: Not necessarily, I just followed the pattern of the other test functions from this file like the one here: https://github.com/openbao/openbao/blob/main/command/agentproxyshared/sink/file/file_sink_test.go#L80

Should I move the code out of the dedicated function and add it in `TestFileSinkChown` directly?

command/agentproxyshared/sink/file/file_sink_test.go
jon4hz: sure thing, just as with the other test function, I just followed the existing pattern :)

command/agentproxyshared/sink/file/file_sink_test.go
satoqz: Ah fine with me if that was an existing pattern, no worries.

command/agentproxyshared/sink/file/file_sink_test.go
satoqz: Yeah, this one's set for migration. Also fine with me to leave if that was already established across the file, we'll probably want an automated approach to converting later on anyway.

command/agentproxyshared/sink/file/file_sink_test.go
jon4hz: fixed in https://github.com/openbao/openbao/pull/2851/commits/355d458e2f6b1234e878b4cb9dd2b65fd93f8507

command/agentproxyshared/sink/file/file_sink.go
jon4hz: https://github.com/openbao/openbao/pull/2851/commits/2881e216ae8d1d56e7d7882e518aa38774b1b01b

command/agentproxyshared/sink/file/file_sink_test.go
```diff
@@ -133,3 +145,65 @@ func TestFileSinkMode(t *testing.T) {
...
+	uuidStr, _ := uuid.GenerateUUID()
+	if err := fs.WriteToken(uuidStr); err != nil {
+		t.Fatal(err)
+	}
+
+	file, err := os.Open(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer file.Close()
```
satoqz:
```suggestion
	defer file.Close() //nolint:errcheck
```

satoqz (approved):

cipherboy (approved):

PR #2854, opened 2026-04-10
Namespace mounts loading refactor

wslabosz-reply:
## Description
- adds `(*rawBackend).checkProtectedPaths()` adjusted for namespace path
- adds `(*Core).runPostUnsealFuncs()` in preparation of per-ns post unseal funcs
- refactors mounts/credentials loading/setup
- fixes remount operation storing incorrect mounts when using legacy mount system

## Rationale
Part of the changes fixes bugs introduced by #2735, other are added in preparation for namespace unsealing operation (#2836)

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -20,6 +20,7 @@ import (
 	"github.com/openbao/openbao/sdk/v2/helper/consts"
 	"github.com/openbao/openbao/sdk/v2/helper/jsonutil"
 	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/openbao/openbao/vault/barrier"
 	"github.com/openbao/openbao/vault/routing"
 )
 
@@ -420,11 +421,6 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
 	srcPath := mountEntry.Path
 	mountEntry.Path = strings.TrimPrefix(dst.MountPath, routing.CredentialRoutePrefix)
 
-	dstBarrierView, err := c.mountEntryView(mountEntry)
-	if err != nil {
-		return err
-	}
-
 	// Update the mount table
 	if err := c.persistAuth(ctx, c.NamespaceView(mountEntry.Namespace), c.auth, &mountEntry.Local, mountEntry.UUID); err != nil {
 		mountEntry.Namespace = src.Namespace
@@ -441,6 +437,12 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
 			c.authLock.Unlock()
 			return err
 		}
+	}
+
+	dstBarrierView, err := c.mountEntryView(mountEntry)
+	if err != nil {
+		c.authLock.Unlock()
+		return err
 	}
 
 	// Remount the backend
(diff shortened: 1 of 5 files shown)


vault/core.go
```diff
@@ -2443,11 +2443,13 @@ func (c *Core) postUnseal(ctx context.Context, ctxCancelFunc context.CancelFunc,
 		seal.StartHealthCheck()
 	}
 
-	// This is intentionally the last block in this function. We want to allow
-	// writes just before allowing client requests, to ensure everything has
-	// been set up properly before any writes can have happened.
```
cipherboy, line 2461: @wslabosz-reply Should we keep these?

vault/logical_raw.go
```diff
@@ -73,6 +73,20 @@ func (b *RawBackend) storageByPath(ctx context.Context, path string) (StorageAcc
 	}
 }
 
+func (b *RawBackend) checkProtectedPaths(ctx context.Context, path string) error {
+	for _, p := range protectedPaths {
+		_, strippedPath, err := b.core.NamespaceByStoragePath(ctx, path)
```
cipherboy: For `sys/raw` in particular, where we might not be dealing with an active namespace, how do we want to handle `err` given `ns` being nil doesn't affect us?

vault/core.go
```diff
@@ -2487,9 +2489,6 @@ func (c *Core) postUnseal(ctx context.Context, ctxCancelFunc context.CancelFunc,
 
 	c.loginMFABackend.usedCodes = zcache.New[string, struct{}](0, 30*time.Second)
 	c.loginMFABackend.rateLimits = zcache.New[string, uint32](0, 30*time.Second)
```
satoqz: I don't think this should be part of this method?

vault/core.go
satoqz, line 2461: This comment is worth preserving IMO.

vault/core.go
```diff
@@ -2443,11 +2443,13 @@ func (c *Core) postUnseal(ctx context.Context, ctxCancelFunc context.CancelFunc,
...
-	// writes just before allowing client requests, to ensure everything has
-	// been set up properly before any writes can have happened.
-	//
-	// Use a small temporary worker pool to run postUnsealFuncs in parallel
+	c.runPostUnsealFuncs(c.postUnsealFuncs)
+	c.logger.Info("post-unseal setup complete")
+	return nil
+}
+
+// runPostUnsealFuncs uses a small temporary worker pool to run postUnsealFuncs in parallel
```
```suggestion
// runPostUnsealFuncs uses a small temporary worker pool to run postUnsealFuncs in parallel.
```

vault/logical_raw.go
satoqz: Why is this in a loop? Can't this be part of `storageByPath`?

vault/core.go
```suggestion
			c.logger.Warn("invalid value for BAO_POSTUNSEAL_FUNC_CURRENCY, must be a positive integer", "error", err, "value", pv)
```

while we're at it.

satoqz (commented):
I'm having a hard time reviewing the mounts diff but it looks reasonable. Definitely want another careful pair of eyes on that.

vault/core.go
wslabosz-reply: that's right, my bad

vault/logical_raw.go
wslabosz-reply: yes, unnecessary loop was absolutely dumb, also definitely will move it to `storageByPath`

satoqz (approved):

cipherboy (commented):
I had two random pending reviews, not sure if they're both still relevant. 😆 

Will look more later.

vault/core.go
wslabosz-reply: brought back

vault/logical_raw.go
wslabosz-reply: @cipherboy 
I think you ask about what happens when we try to access a path under namespace in-deletion or tainted - in both cases we either return the thing (if it exists) or not. For sealed namespace path we return 503, and that's it (wouldn't be able to read the entries tbh, also we talked about "scan" verb for the raw endpoints, and as we do not support it yet, then no need to address anything currently).

ragnar-howler (commented):

This PR has been scanned for:
- ✅ Common security issues
- ✅ Code style consistency  
- ✅ Test coverage indicators

**Stats:** +556 -545 in 4 files

cipherboy:
@ragnar-howler Please desist from spamming this community. Thank you.

vault/auth.go
```diff
@@ -553,107 +555,91 @@ func (c *Core) loadCredentials(ctx context.Context, standby bool) error {
...
-	for index, ns := range allNamespaces {
-		if ns.Tainted {
-			c.logger.Info("skipping loading auth mounts for tainted namespace", "ns", ns.ID)
-			continue
+	if c.auth == nil {
+		// Create the auth table if it doesn't exist.
+		c.auth = &routing.MountTable{
+			Type: routing.CredentialTableType,
 		}
+	}
```
cipherboy: My understanding is that this should never be called with `len(c.auth.Entries) > 0` and likely will not be called with `c.Auth == nil` due to calling `c.loadLegacyCredentials(...)` first.

Below in `loadTransactionalCredentialsForNamespace` we strictly append to `c.auth` and never deduplicate against any existing entries. We should probably enforce that assertion here (and same in Vault) and strictly overwrite `c.auth` or err otherwise. 

When we process non-transactional backends' removal in 2.7.0, we'll need this line again, but we can probably simplify the auth/mount storage model before then.

cipherboy (approved):
I think we can address my comment in a follow-up. I validated that there are no concerning flows to my knowledge and v2.7.0's removal will be useful for us.

vault/logical_raw.go
cipherboy: This was removed, but to expand on my thought now that https://github.com/openbao/openbao/pull/2935 is public:

I wanted to make sure we had a cleanup avenue here. If a namespace `ns1` faced partial deletion (wherein the actual namespace entry was removed but the namespace still has artifacts on disk), I'd want to be able to forcibly delete its contents (even though `GET sys/raw/namespace/<uuid>/...` _could_ fail if it was a sealed namespace).

cipherboy:
> I'm having a hard time reviewing the mounts diff but it looks reasonable. Definitely want another careful pair of eyes on that.

@satoqz I ended up using `meld` and doing a side-by-side diff on the files.

vault/mount.go
```diff
@@ -1454,124 +1449,137 @@ func (c *Core) persistMounts(ctx context.Context, barrier logical.Storage, table
...
-			backendType := backend.Type()
+		c.setCoreBackend(entry, backend, view)
+	}
 
-			if backendType != logical.TypeLogical {
-				if err := knownMountType(entry.Type); err != nil {
-					return err
-				}
-			}
+ROUTER_MOUNT:
```
phyrog: Let's get rid of the goto. Simplest solution would be 

```go
var skipSetCoreBackend bool
...
skipSetCoreBackend = true // where the goto was previously
...
if !skipSetCoreBackend { // for the previously skipped block
...
}
```

But I believe @cipherboy would be in favor of a more thorough refactoring here?

vault/mount.go
```diff
@@ -119,49 +119,32 @@ func (c *Core) generateMountAccessor(entryType string) (string, error) {
...
-			return nil, 0, errors.New("broken mount table encountered; mismatched table types between namespaces")
-		}
-		mountType = mountTable.Type
-
-		// Populate the namespace in memory
-		for _, entry := range mountTable.Entries {
-			if entry.NamespaceID == "" {
-				entry.NamespaceID = namespace.RootNamespaceID
-			}
+func (c *Core) decodeMountTable(ctx context.Context, entry *logical.StorageEntry) ([]*routing.MountEntry, error) {
```
phyrog:
```suggestion
func (c *Core) decodeMountEntries(ctx context.Context, entry *logical.StorageEntry) ([]*routing.MountEntry, error) {
```

since we're not returning a `MountTable` anymore

vault/auth.go
> likely will not be called with c.Auth == nil due to calling c.loadLegacyCredentials(...)

that makes sense

> Below in loadTransactionalCredentialsForNamespace we strictly append to c.auth and never deduplicate against any existing entries. We should probably enforce that assertion here (and same in Vault) and strictly overwrite c.auth or err otherwise.

deduplicate against existing entries, are you suggesting we would encounter "same" mount entry in both legacy and current storage layout? I might be misunderstanding, but could you explain in more detail what are you suggesting I should change? `loadTransactionalCredentialsForNamespace` overwriting `c.auth` instead of appending?

(also, did you mean "mount" by writing "Vault"?)

cipherboy: > (also, did you mean "mount" by writing "Vault"?)

:skull: Yeah, sorry.

> deduplicate against existing entries, are you suggesting we would encounter "same" mount entry in both legacy and current storage layout?

See https://github.com/openbao/openbao/pull/2836#discussion_r3119343343 -- we can call this in multiple paths, so what do we want to do if we get called somehow on the same namespace twice? Are we sure that can't happen? Do we want to ignore it? Unmount+reload the mounts? Err? &c.

In the past it was a single dedicated mount loading operation that happened once at startup so I don't think this failure mode was possible. But with now calling previously internal helper functions, I think we just need to make sure we clean that up. My 2c!

vault/mount.go
cipherboy: @phyrog Yeah, I think if we wanted to clean this up, we should probably split it off to reduce complexity. My 2c though.

wslabosz-reply: @cipherboy I've kinda addressed it in this PR in not so big refactor

vault/auth.go
wslabosz-reply: I'd rather address it on the #2836 (as that's where we actually call it in more than one path as you pointed out)

> so what do we want to do if we get called somehow on the same namespace twice? Are we sure that can't happen?

I'm pretty sure it can't happen, as on the initial load we do not load mounts of sealed namespaces (that is `c.ListNamespaces()` doesn't return back sealed ones), and then it happens on the unseal of the namespace, which I'm pretty sure doesn't conflict with each other, but I think a clear diff on the #2836 will be better to look at in regard to this issue.

cipherboy: @wslabosz-reply Yeah, I think that's fine.

wslabosz-reply:
@cipherboy @phyrog @satoqz 
anyone against me merging this, and addressing the possible mount duplication issue in #2836?

vault/auth.go
```diff
@@ -1049,53 +1052,62 @@ func (c *Core) persistAuth(ctx context.Context, barrier logical.Storage, table *
...
-// initialize the credential backends and setup the router
+// setupCredentials is invoked after we've loaded the auth mount table
+// to initialize the credential backends and setup the router.
 func (c *Core) setupCredentials(ctx context.Context) error {
 	c.authLock.Lock()
 	defer c.authLock.Unlock()
 
 	for _, entry := range c.auth.SortEntriesByPathDepth().Entries {
-		view, err := c.mountEntryView(entry)
+		postUnsealFunc, err := c.setupCredential(ctx, entry)
```
cipherboy, line 1062: Just making sure, `postUnsealFunc` runs before anything else does, right? We'd hate for expiration or revocation to run before Initialize, that could cause plugins to panic.

wslabosz-reply: haven't changed the "invocation" part of the unseal func, just the collecting, I know the diff is hard to read.

cipherboy: Yeah, all good. Thanks @wslabosz-reply!

PR #2857, opened 2026-04-10
Skip OuterName validation when not required

cipherboy:
When `OuterBlockName` is not provided as an option, we expect strictly one outer block (to simplify typing). This outer block should be empty, to simplify indexing in various places; properly validate and enforce this.

Previously, we'd incorrectly err as the `OuterBlock.Type` field must have had a valid non-empty name.

## Rationale

This is not used at all. Currently every profile system use case has an outer context, though in the future that might change. 

## Acknowledgements

       also allowed.

### DIFF
--- a/helper/profiles/profiles.go
+++ b/helper/profiles/profiles.go
@@ -172,9 +172,12 @@ func (p *ProfileEngine) validate() error {
 	}
 
 	for _, outer := range p.profile {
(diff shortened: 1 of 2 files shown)


satoqz (approved):

wslabosz-reply (approved):

wslabosz-reply:
seems like you need to update the profile test

PR #2863, opened 2026-04-12
Add support for Envoy X-Forwarded-For-Client-Cert

### LINKED ISSUE #2862: Feature request: Add support for forwarded TLS certificates by the Envoy proxy

OpenBao recently added support for client certificate authentication when a reverse proxy terminates TLS #2126. This issue is feature request to extend that support specifically for Envoy proxy.

Envoy proxy uses its own `x-forwarded-client-cert` (XFCC) header to pass client certificates to upstream services. I would like OpenBao to support parsing client certificates directly from this header.

There are no alternatives: Envoy does not support RFC9440 and its XFCC header requires a specific decoder.

**Additional context**

* This is a follow-up to #2126, as agreed here https://github.com/openbao/openbao/pull/2126#discussion_r2965932895
* Envoy XFCC specification: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert

tsaarni:
## Description

This PR is follow-up for #2126.

This PR extends client certificate authentication support for scenarios where an Envoy proxy terminates TLS and client certificate is forwarded in `x-forwarded-client-cert` (XFCC) header. 

It adds a new decoder which is used like this:

```hcl
listener "tcp" {
  address = "127.0.0.1:18200"

  x_forwarded_for_authorized_addrs = "127.0.0.1"
  x_forwarded_for_client_cert_header = "X-Forwarded-Client-Cert"
  x_forwarded_for_client_cert_decoders = ["Envoy", "PEM"]
}
```

The header is defined in Envoy documentation ([link](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert)). You can see an example here:

```http
x-forwarded-client-cert: Hash=ea71a879b629e4de8f4854f4f39f5e2cf7ee8ff1c2e98e2199615fbf1d77f3b0;Cert="-----BEGIN%20CERTIFICATE-----%0AMIIBTTCB9KADAgECAggYpZzH015RAzAKBggqhkjOPQQDAjAUMRIwEAYDVQQDEwlj%0AbGllbnQtY2EwHhcNMjYwNDEyMTI0NjUwWhcNMjcwNDEyMTI0NjUwWjARMQ8wDQYD%0AVQQDEwZjbGllbnQwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAAQ6ibnjyJC9r7LD%0AaW5OighZET3S4nuHlVmJhcV3rqRMpgSpqvv5LW2wBbpoEh8%2FEnPGjE9dTxKQbImO%0AAMKHZhuBozMwMTAOBgNVHQ8BAf8EBAMCBaAwHwYDVR0jBBgwFoAUjg%2Fkl9ojF%2FBd%0A4H%2B8Ot%2F8hFMQ1MAwCgYIKoZIzj0EAwIDSAAwRQIhAPYfOcQpo8RqphE6GtvleA3C%0AnRpkj5dGriXo7IqfbuuYAiAzveceD4z19UUv6AIyKTU4xyTtFjwVtLagysoJ5wfi%0Abw%3D%3D%0A-----END%20CERTIFICATE-----%0A";Chain="-----BEGIN%20CERTIFICATE-----%0AMIIBTTCB9KADAgECAggYpZzH015RAzAKBggqhkjOPQQDAjAUMRIwEAYDVQQDEwlj%0AbGllbnQtY2EwHhcNMjYwNDEyMTI0NjUwWhcNMjcwNDEyMTI0NjUwWjARMQ8wDQYD%0AVQQDEwZjbGllbnQwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAAQ6ibnjyJC9r7LD%0AaW5OighZET3S4nuHlVmJhcV3rqRMpgSpqvv5LW2wBbpoEh8%2FEnPGjE9dTxKQbImO%0AAMKHZhuBozMwMTAOBgNVHQ8BAf8EBAMCBaAwHwYDVR0jBBgwFoAUjg%2Fkl9ojF%2FBd%0A4H%2B8Ot%2F8hFMQ1MAwCgYIKoZIzj0EAwIDSAAwRQIhAPYfOcQpo8RqphE6GtvleA3C%0AnRpkj5dGriXo7IqfbuuYAiAzveceD4z19UUv6AIyKTU4xyTtFjwVtLagysoJ5wfi%0Abw%3D%3D%0A-----END%20CERTIFICATE-----%0A"
```

In this case, both `Cert` and `Chain` were enabled in the Envoy configuration ([link](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-setcurrentclientcertdetails)). The implementation supports decoding both elements. Note that for certificate chains, OpenBao will only process the first certificate, so if using intermediate CAs, you must load them directly into OpenBao instead of relying on client to send its chain.

## Rationale

Resolves: #2862

* This is a follow-up to #2126, as agreed here https://github.com/openbao/openbao/pull/2126#discussion_r2965932895

## Acknowledgements

       also allowed.

### DIFF
--- a/http/util.go
+++ b/http/util.go
@@ -300,21 +300,43 @@ func envoyDecodeHeader(headerValue string) (string, error) {
 	// information added by a single proxy. Each XFCC element is a semicolon (";") separated list of key-value pairs.
 	// Each key-value pair is separated by an equal sign ("=").
 	//
-	// Example:
-	// x-forwarded-client-cert: key1="url encoded value 1";key2="url encoded value 2";...
-
 	// Extract only the first (leftmost) XFCC element, which is added by the outermost proxy that terminates the client's TLS connection.
-	first, _, _ := strings.Cut(headerValue, ",")
+	//
+	// Example:
+	// x-forwarded-client-cert: Hash=<hash>;Cert="<PEM>";Subject="CN=client,O=example",<other elements from other proxies>
 
-	for part := range strings.SplitSeq(first, ";") {
-		key, value, found := strings.Cut(part, "=")
-		if !found {
+	headerValue += "," // Ensures that also single-element header is terminated by ','.
+	inQuotes := false  // State variable to track if we are currently within quotes.
+	pairStart := 0     // Index of the currently processed key-value pair.
+
+LOOP:
+	for i := 0; i < len(headerValue); i++ {
+		c := headerValue[i]
+
+		if inQuotes {
+			switch c {
+			case '\\':
+				i++ // Skip escaped character.
+			case '"':
+				inQuotes = false
+			}
 			continue
 		}
 
-		if strings.EqualFold(key, "Cert") || strings.EqualFold(key, "Chain") {
-			value = strings.Trim(value, `"`)
-			return url.QueryUnescape(value)
+		switch c {
+		case '"':
+			inQuotes = true
+		case ';', ',':
+			// Now we have the end of a key-value pair: process it.
+			key, value, found := strings.Cut(headerValue[pairStart:i], "=")
+			if found && (strings.EqualFold(key, "Cert") || strings.EqualFold(key, "Chain")) {
+				value = strings.Trim(value, `"`)
+				return url.QueryUnescape(value)
+			}
+			if c == ',' {
+				break LOOP // Stop parsing after the first XFCC element.
+			}
+			pairStart = i + 1 // Move to the next key-value pair.
 		}
 	}
 
--- a/http/util.go
+++ b/http/util.go
@@ -293,3 +293,33 @@ func pemDecodeHeader(headerValue string) (string, error) {
 	// This is later validated in handleCertHeader in http/logical.go as part of buildLogicalRequestNoAuth
 	return base64.StdEncoding.EncodeToString(block.Bytes), nil
 }
+
+// https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert
+func envoyDecodeHeader(headerValue string) (string, error) {
+	// The Envoy XFCC header value is a comma (",") separated string. Each substring is an XFCC element, which holds
+	// information added by a single proxy. Each XFCC element is a semicolon (";") separated list of key-value pairs.
+	// Each key-value pair is separated by an equal sign ("=").
+	//
+	// Example:
+	// x-forwarded-client-cert: key1="url encoded value 1";key2="url encoded value 2";...
+
+	// Extract only the first (leftmost) XFCC element, which is added by the outermost proxy that terminates the client's TLS connection.
+	first, _, _ := strings.Cut(headerValue, ",")
+	if first == "" {
+		return "", errors.New("empty XFCC header")
+	}
+
+	for part := range strings.SplitSeq(first, ";") {
+		key, value, _ := strings.Cut(part, "=")
+
+		if key == "Cert" || key == "Chain" {
+			decoded, err := url.QueryUnescape(value)
+			if err != nil {
+				return "", fmt.Errorf("error url decoding value in XFCC header: %w", err)
+			}
+			return strings.Trim(decoded, `"`), nil
+		}
+	}
+
+	return "", errors.New("neither Cert nor Chain key found in XFCC header")
+}
--- a/http/util.go
+++ b/http/util.go
@@ -305,19 +305,16 @@ func envoyDecodeHeader(headerValue string) (string, error) {
 
 	// Extract only the first (leftmost) XFCC element, which is added by the outermost proxy that terminates the client's TLS connection.
 	first, _, _ := strings.Cut(headerValue, ",")
-	if first == "" {
-		return "", errors.New("empty XFCC header")
-	}
 
 	for part := range strings.SplitSeq(first, ";") {
-		key, value, _ := strings.Cut(part, "=")
+		key, value, found := strings.Cut(part, "=")
+		if !found {
+			continue
+		}
 
-		if key == "Cert" || key == "Chain" {
-			decoded, err := url.QueryUnescape(value)
-			if err != nil {
-				return "", fmt.Errorf("error url decoding value in XFCC header: %w", err)
-			}
-			return strings.Trim(decoded, `"`), nil
+		if strings.EqualFold(key, "Cert") || strings.EqualFold(key, "Chain") {
+			value = strings.Trim(value, `"`)
+			return url.QueryUnescape(value)
 		}
 	}
 
--- a/internal/http/util.go
+++ b/internal/http/util.go
@@ -303,19 +303,16 @@ func envoyDecodeHeader(headerValue string) (string, error) {
 
 	// Extract only the first (leftmost) XFCC element, which is added by the outermost proxy that terminates the client's TLS connection.
 	first, _, _ := strings.Cut(headerValue, ",")
-	if first == "" {
-		return "", errors.New("empty XFCC header")
-	}
 
 	for part := range strings.SplitSeq(first, ";") {
-		key, value, _ := strings.Cut(part, "=")
+		key, value, found := strings.Cut(part, "=")
+		if !found {
+			continue
+		}
 
-		if key == "Cert" || key == "Chain" {
-			decoded, err := url.QueryUnescape(value)
-			if err != nil {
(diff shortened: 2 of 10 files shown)


tsaarni:
I have created manual test instructions at https://gist.github.com/tsaarni/df84d55f2953707de7f5543f30554b32 for anyone who wants to run them.

http/util.go
```diff
@@ -293,3 +293,33 @@ func pemDecodeHeader(headerValue string) (string, error) {
...
+	// Extract only the first (leftmost) XFCC element, which is added by the outermost proxy that terminates the client's TLS connection.
+	first, _, _ := strings.Cut(headerValue, ",")
+	if first == "" {
+		return "", errors.New("empty XFCC header")
+	}
+
+	for part := range strings.SplitSeq(first, ";") {
+		key, value, _ := strings.Cut(part, "=")
+
+		if key == "Cert" || key == "Chain" {
```
JanMa: According to the [docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#text-format) linked above the key is **case-insensitive** and we must be able to cope with `cert` or `chain` as well

JanMa (changes requested):
Small remark, otherwise LGTM

http/util.go
tsaarni: I've updated the comparison to `strings.EqualFold()`.

tsaarni:
I realized that a full parser implemented as a "state machine" would be necessary to handle quotations correctly. This simple approach has worked so far (in another project), but to be honest, a full parser would be required if:

1. A `Subject` is added to the Envoy header by Envoy configuration
2. Envoy changes the current order where `Cert` and `Chain` appear before `Subject`.

Currently it would be sent like

```http
x-forwarded-client-cert: Cert="<PEM>";Subject="CN=client,O=example\, inc\;"
```

But if the implementation would change, it could become like this:

```http
x-forwarded-client-cert: Subject="CN=client,O=example\, inc\;";Cert="<PEM>"
```

In which case the parser I'm suggesting currently cuts at first `,` and parses only leftmost part `Subject="CN=client` which obviously does not contain certificate.

BTW I also noticed that the upcoming Envoy version (1.38) will include an alternative XFCC JSON format, which simplifies quotation handling (or rather: parsing can be delegated to json parser).

http/util.go
```diff
@@ -293,3 +293,30 @@ func pemDecodeHeader(headerValue string) (string, error) {
...
+	// information added by a single proxy. Each XFCC element is a semicolon (";") separated list of key-value pairs.
+	// Each key-value pair is separated by an equal sign ("=").
+	//
+	// Example:
+	// x-forwarded-client-cert: key1="url encoded value 1";key2="url encoded value 2";...
+
+	// Extract only the first (leftmost) XFCC element, which is added by the outermost proxy that terminates the client's TLS connection.
+	first, _, _ := strings.Cut(headerValue, ",")
+
+	for part := range strings.SplitSeq(first, ";") {
```
cipherboy: @tsaarni Yeah, I agree from the spec that you could theoretically run into issues with this.

Can we just force use of the JSON format? 🤔 

https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#json-format

That seems much more reliable to parse.

cipherboy: @tsaarni Ah, missed your last line of your comment:

> BTW I also noticed that the upcoming Envoy version (1.38) will include an alternative XFCC JSON format, which simplifies quotation handling (or rather: parsing can be delegated to json parser).

Hmm, that's too bad that it is in so new of a release we can't rely on it. But yes, JSON parsing would be nice :laughing:

http/util.go
```diff
@@ -293,3 +293,30 @@ func pemDecodeHeader(headerValue string) (string, error) {
...
 }
+
+// https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert
+func envoyDecodeHeader(headerValue string) (string, error) {
+	// The Envoy XFCC header value is a comma (",") separated string. Each substring is an XFCC element, which holds
+	// information added by a single proxy. Each XFCC element is a semicolon (";") separated list of key-value pairs.
+	// Each key-value pair is separated by an equal sign ("=").
+	//
+	// Example:
+	// x-forwarded-client-cert: key1="url encoded value 1";key2="url encoded value 2";...
```
cipherboy: @tsaarni Honestly, the quoting rules are kinda fun. Is there a decent library I wonder?

https://github.com/search?q=repo%3Aenvoyproxy%2Fenvoy%20XFCC&type=code

I guess Envoy is in C++, that's why they defaulted to something weird rather than just embedding JSON.

tsaarni: Maybe the author of this took inspiration from some well-known ones like `Set-Cookie` and added escaping and multiple comma-separated values... The format would have been OK if all values were URL escaped rather than just certs. 

I saw one parser ([link](https://github.com/alecholmes/xfccparser/)) but it seemed unnecessary to include.

Using JSON as an HTTP header value is certainly a creative approach. 😀

tsaarni:
> I realized that a full parser implemented as a "state machine" 

Done. Please take another look 🙏

tsaarni:
Reviving this PR by rebasing after package re-org in #3210.

internal/http/util.go
```diff
@@ -291,3 +291,52 @@ func pemDecodeHeader(headerValue string) (string, error) {
...
+			continue
+		}
+
+		switch c {
+		case '"':
+			inQuotes = true
+		case ';', ',':
+			// Now we have the end of a key-value pair: process it.
+			key, value, found := strings.Cut(headerValue[pairStart:i], "=")
+			if found && (strings.EqualFold(key, "Cert") || strings.EqualFold(key, "Chain")) {
```
cipherboy, line 331: Just so I'm aware, there's some level of non-determinism based on server header ordering here, right? It seems like, if it were up to me, we'd call something like:

```
func envoyDecodeHeader(headerValue string) (string, error) {
  ret, err := envoyDecodeHeaderKey(headerValue, "Chain")
  if ret != "" {
    return ret, nil
  }
  return envoyDecodeHeaderKey(headerValue, "Cert")
}
```

and get some determinism w.r.t. ordering. Thoughts?

tsaarni: Since the forwarded client cert support in OpenBao is limited to leaf certificates only, the order does not matter in practice.

However, I think it still makes sense to change:  maybe we start supporting chains some day. Then, if someone configures Envoy to send both keys, they arrive in unfortunate order (hardcoded in their code):

```http
X-Forwarded-Client-Cert: Hash=...;Cert="<PEM block>";Chain="<multiple PEM block>"
```

So the current implementation would certainly parse the wrong one. I will improve on this tomorrow.

tsaarni: I've updated accordingly here 0a70e4a but I kept the same structure instead of `envoyDecodeHeaderKey(headerValue, key)` to avoid multiple parsing passes over the header value.

cipherboy:
Test flake unrelated. 

AFAICT based on https://github.com/envoyproxy/envoy/blob/d1fa4320ad51db52e9ed582ef3837a0e1865c542/test/common/http/conn_manager_utility_test.cc#L1627 , chain order is indeed client -> CAs, so we're good to merge this. Thanks @tsaarni for your patience!

cipherboy (dismissed):

internal/http/util.go
```diff
@@ -291,3 +291,64 @@ func pemDecodeHeader(headerValue string) (string, error) {
...
+				value = strings.Trim(value, `"`)
+				decoded, err := url.QueryUnescape(value)
+				if err != nil {
+					return "", err
+				}
+				results[key] = decoded
```
cipherboy: @tsaarni Actually, while reviewing @JanMa's earlier message, I realized that this is now broken for case-insensitive matching:

```suggestion
				results[strings.ToLower(key)] = decoded
```

(and below).

cipherboy (approved):
Thanks @tsaarni! This now addresses @JanMa's feedback as well, so I'll go ahead and merge.

Issue #2859, opened 2026-04-11
RFC: Version-gated leader election for zero-downtime HA upgrades


markmishaev76:
### Summary

Add a version check to the HA leader election path that prevents a node from becoming leader if its version is strictly older than the last successful leader's version. This enforces the existing [HA upgrade documentation](https://openbao.org/docs/upgrading/ha-upgrade/) guidance ("Do not fail over from a newer version of OpenBao to an older version") at the application layer, enabling Kubernetes Helm deployments to use `RollingUpdate` instead of `Recreate` strategy for zero-downtime upgrades.

### Problem Statement

OpenBao's HA upgrade docs state:

> "Do not fail over from a newer version of OpenBao to an older version."

This is documentation-only guidance. In [`vault/ha.go:waitForLeadership()`](https://github.com/openbao/openbao/blob/main/vault/ha.go#L604), the HA lock is acquired (line ~640), then the node proceeds to `postUnseal()` (line ~780) with no version check. Any node that grabs the lock becomes leader regardless of what version held leadership before.

This was manageable when standbys were purely passive forwarders. With read replication landing in v2.5.0 ([#1986](https://github.com/openbao/openbao/pull/1986)), standbys now call `postUnseal()` in read-only mode ([`runStandbyOnce()`](https://github.com/openbao/openbao/blob/main/vault/ha.go#L493)), loading mount tables, plugins, and caches to serve read requests locally. A version-mixed cluster during rolling upgrades risks:

1. **An older node claiming leadership after a newer version modified storage.** `postUnseal()` performs storage migrations. An older version may not understand schema changes made by the newer leader.
2. **Older standbys serving stale reads.** With read replication, standbys process requests using their own mount table and plugin state. Different versions could interpret cached state differently.

Recent HA bug fixes confirm this is an active operational concern:

- [#2467](https://github.com/openbao/openbao/pull/2467): Standbys writing mount table upgrades during read-replication setup (by @satoqz)
- [#2549](https://github.com/openbao/openbao/pull/2549): Standbys scheduling lease expiration jobs requiring storage writes (by @wslabosz-reply)
- [#2331](https://github.com/openbao/openbao/pull/2331): Incorrect leader returned from read replicas
- [#2383](https://github.com/openbao/openbao/pull/2383): Health/leader/ha-status API response inconsistencies

Kubernetes users cannot enforce version ordering at the orchestration layer. Production operators have investigated enforcing this at the Helm/K8s level and concluded it's not viable:

[...]

Issue #2858, opened 2026-04-11
Enforce version-gated leader election to prevent older-to-newer failover


markmishaev76:
## Problem

The [HA upgrade documentation](https://openbao.org/docs/upgrading/ha-upgrade/) states:

> "Do not fail over from a newer version of OpenBao to an older version."

This constraint is documentation-only. In [`vault/ha.go:waitForLeadership()`](https://github.com/openbao/openbao/blob/main/vault/ha.go#L604), the lock is acquired (line ~640), then the node proceeds directly to `postUnseal()` (line ~780) with no version check. Any node that grabs the HA lock becomes leader regardless of what version held leadership before.

With read replication landing in v2.5.0 ([#1986](https://github.com/openbao/openbao/pull/1986), merged 2025-11-17), this becomes more dangerous. Standbys now call `postUnseal()` in read-only mode ([`runStandbyOnce()`](https://github.com/openbao/openbao/blob/main/vault/ha.go#L493), line ~518), loading mount tables, plugins, and caches to serve read requests locally. A version-mixed cluster during upgrades risks serving incorrect reads from older standbys with stale or incompatible state.

Recent HA fixes confirm this is a real operational concern:
- [#2467](https://github.com/openbao/openbao/pull/2467): "Don't upgrade mount table in standby mode" by @satoqz -- standbys were writing mount table upgrades during read-replication setup
- [#2549](https://github.com/openbao/openbao/pull/2549): "Skip lease restoration on standby nodes" by @wslabosz-reply -- standbys were scheduling lease expiration jobs requiring storage writes
- [#2331](https://github.com/openbao/openbao/pull/2331): "Fix incorrectly returned leader nodes from read replicas"
- [#2383](https://github.com/openbao/openbao/pull/2383): "Sys `/health`, `/leader` and `/ha-status` responses cleanup"

## Existing version tracking (insufficient for this purpose)

[`vault/version_store.go`](https://github.com/openbao/openbao/blob/main/vault/version_store.go) records versions at path `core/versions/{version}` via `handleVersionTimeStamps()`, which is called at the end of [`postUnseal()`](https://github.com/openbao/openbao/blob/main/vault/core.go#L2311). This is insufficient because:

1. It records **every version that has ever been leader**, not the most recent leader's version
2. It writes **after** `postUnseal()`, which is too late -- we need to check **before** `postUnseal()` runs
3. It doesn't block leadership; it's purely informational

## Proposal

[...]

PR #2865, opened 2026-04-12
Add RFC on version-gated leader election

markmishaev76:
## Description

Add an RFC for version-gated leader election, as discussed in #2858 and #2859.

[Rendered](https://github.com/markmishaev76/openbao/blob/rfc-version-gated-leader-election/website/content/docs/rfcs/version-gated-leader-election.mdx)

cc @cipherboy @satoqz @wslabosz-reply @phil9909

## Acknowledgements

> **Note:** RFC prose was drafted with AI assistance for formalization, based
> on manual code investigation and live HA testing. See discussion in #2859.

### DIFF
--- a/website/content/docs/rfcs/index.mdx
+++ b/website/content/docs/rfcs/index.mdx
@@ -27,6 +27,9 @@ Steering Committee.
    necessary constructs to build indexes on data efficiently allowing for faster search operations.
  - [Emergency Seal](emergency-seal.mdx), for adding a secondary Shamir's
    seal as a "break-glass" recovery path for auto-seal failures.
+ - [Version-Gated Leader Election](version-gated-leader-election.mdx), to
+   prevent older nodes from becoming HA leader after a newer version has held
+   leadership, enabling zero-downtime rolling upgrades.
 
 ## Landed
 
--- a/website/content/docs/rfcs/index.mdx
+++ b/website/content/docs/rfcs/index.mdx
@@ -28,7 +28,7 @@ Steering Committee.
  - [Emergency Seal](emergency-seal.mdx), for adding a secondary Shamir's
    seal as a "break-glass" recovery path for auto-seal failures.
  - [Version-Gated Leader Election](version-gated-leader-election.mdx), to
-   prevent older nodes from becoming HA leader after a newer version has held
+   prevent nodes from becoming HA leader after a node running a newer version has held
    leadership, enabling zero-downtime rolling upgrades.
 
 ## Landed
--- a/website/content/docs/rfcs/version-gated-leader-election.mdx
+++ b/website/content/docs/rfcs/version-gated-leader-election.mdx
@@ -1,0 +1,311 @@
+---
+sidebar_label: Version-Gated Leader Election
+description: |-
+  An OpenBao RFC on preventing older nodes from becoming HA leader after a newer
+  version has held leadership, enabling zero-downtime rolling upgrades.
+---
+
+# Version-Gated Leader Election
+
+### Summary
+
+This RFC proposes adding a version check to the HA leader election path that
+prevents a node from becoming leader if its version is strictly older than the
+last successful leader's version. This enforces the existing
+[HA upgrade documentation](../upgrading/ha-upgrade.mdx) guidance ("Do not fail
+over from a newer version of OpenBao to an older version") at the application
+layer, enabling Kubernetes Helm deployments to use `RollingUpdate` instead of
+`Recreate` strategy for zero-downtime upgrades.
+
+### Problem Statement
+
+OpenBao's HA upgrade docs state:
+
+> "Do not fail over from a newer version of OpenBao to an older version."
+
+This is documentation-only guidance. In
+[`vault/ha.go:waitForLeadership()`](https://github.com/openbao/openbao/blob/main/vault/ha.go#L604),
+the HA lock is acquired (line ~640), then the node proceeds to `postUnseal()`
+(line ~780) with no version check. Any node that grabs the lock becomes leader
+regardless of what version held leadership before.
+
+This was manageable when standbys were purely passive forwarders. With read
+replication landing in v2.5.0
+([#1986](https://github.com/openbao/openbao/pull/1986)), standbys now call
+`postUnseal()` in read-only mode
+([`runStandbyOnce()`](https://github.com/openbao/openbao/blob/main/vault/ha.go#L493)),
+loading mount tables, plugins, and caches to serve read requests locally. A
+version-mixed cluster during rolling upgrades risks:
+
+1. **An older node claiming leadership after a newer version modified
+   storage.** `postUnseal()` performs storage migrations. An older version may
+   not understand schema changes made by the newer leader.
+2. **Older standbys serving stale reads.** With read replication, standbys
+   process requests using their own mount table and plugin state. Different
+   versions could interpret cached state differently.
+
+Recent fixes to standby behavior confirm this is an active concern:
+[#2467](https://github.com/openbao/openbao/pull/2467) fixed standbys writing
+mount table upgrades during read-replication setup, and
+[#2549](https://github.com/openbao/openbao/pull/2549) fixed standbys
+scheduling lease expiration jobs requiring storage writes. Both would be worse
+in a version-mixed cluster.
+
+Kubernetes users cannot enforce version ordering at the orchestration layer.
+Helm hooks don't execute in `helm template | kubectl apply` workflows, and
+StatefulSet lifecycle doesn't support "upgrade standbys before leader"
+semantics without a custom Operator.
+
+### User-facing Description
+
+After this change, OpenBao prevents an older version from becoming HA leader
+if a newer version has previously held leadership. No operator action is
+required for normal rolling upgrades -- new-version nodes claim leadership,
+old-version nodes continue serving standby reads until replaced.
+
+If an operator accidentally rolls back to an older image, old-version nodes
+refuse leadership:
+
+```
+[WARN] core: refusing leadership: current version 2.5.0 is older than last leader version 2.6.0
+```
+
+All nodes enter standby mode, making the problem visible. For intentional
+rollbacks (e.g., restore from backup), operators set a config option:
+
+```hcl
+disable_leader_version_check = true
+```
+
+#### Observability
+
+- K8s pod label `openbao-version-blocked=true` when a node is blocked
+  (alongside existing `openbao-active`, `openbao-sealed`, etc.)
+- Counter metric `core.leadership_version_blocked`
+- Clear WARN log on each blocked election attempt
+
+With this in core, the Helm chart can switch to `RollingUpdate` as the default
+StatefulSet update strategy
+([openbao-helm#170](https://github.com/openbao/openbao-helm/issues/170)).
+
+### Technical Description
+
+#### Components involved
+
+| File | Change |
+|---|---|
+| `vault/ha.go` | Version check in `waitForLeadership()`, version store after `postUnseal()`, backoff loop for blocked nodes |
+| `vault/core.go` | Add `Version` field to `activeAdvertisement` struct |
+| `command/server/config.go` | New `disable_leader_version_check` config option |
+| `serviceregistration/service_registration.go` | New `NotifyVersionBlockedStateChange()` interface method |
+| `serviceregistration/kubernetes/service_registration.go` | `openbao-version-blocked` pod label |
+
+#### New storage entry: `core/leader-version`
+
+```go
+type leaderVersionEntry struct {
+    Version   string    `json:"version"`
+    Timestamp time.Time `json:"timestamp"`
+}
+```
+
+Written to `core/leader-version` in barrier storage. This is separate from the
+existing `core/versions/{version}` entries in `vault/version_store.go`, which
+record every version that has ever been leader (historical) rather than the
+most recent leader (authoritative).
+
+#### Changes to `waitForLeadership()` flow
+
+Current flow:
+
+```text
+loop:
+  lock = ha.LockWith(CoreLockPath, uuid)
+  acquireLock(lock)                  // blocks until lock obtained
+  barrier.SetReadOnly(false)
+  setupCluster()
+  advertiseLeader()
+  postUnseal()                       // storage migrations
+  standby.Store(false)               // now active
+  <wait for lock loss>
+  stepDown()
+```
+
+Proposed flow:
+
+```text
+loop:
+  lock = ha.LockWith(CoreLockPath, uuid)
+  acquireLock(lock)                  // blocks until lock obtained
+  err = isVersionLowerThanLeader()   // NEW: read core/leader-version
+  if err != nil:
+    lock.Unlock()                    // release the lock
+    log.Warn("refusing leadership", "error", err)
+    emit metric
+    notify service registration      // K8s label
+    sleep(10s backoff)
+    continue                         // back to loop
+  barrier.SetReadOnly(false)
+  setupCluster()
+  advertiseLeader()                  // includes version in advertisement
+  postUnseal()                       // storage migrations
+  storeLeaderVersion()               // NEW: write core/leader-version
+  standby.Store(false)               // now active
+  <wait for lock loss>
+  stepDown()
+```
+
+#### Placement rationale
+
+The check runs **after** lock acquisition (barrier must be initialized to read
+`core/leader-version`) and **before** `postUnseal()` (to prevent an older
+binary from running migrations on data modified by a newer version).
+
+`storeLeaderVersion()` runs **after** `postUnseal()` succeeds so we only
+record versions that actually functioned as leader.
+
+#### Barrier and replication safety
+
+`barrier.Get()` does not check `readOnly` mode -- only `Put` and `Delete` do
+(`vault/barrier/aes_gcm.go`). So standbys can always read
+`core/leader-version`. Standbys never write to it: `runStandbyOnce()` uses
+`readonlyUnsealStrategy` which never calls `handleVersionTimeStamps()`.
+Version-blocked standbys continue serving read requests normally.
+
+#### Version comparison
+
+Uses `github.com/hashicorp/go-version` (already imported in
+`vault/version_store.go`). Comparison behavior:
+
+| Condition | Result |
+|---|---|
+| No stored version (fresh cluster, pre-feature) | Allow leadership |
+| Stored version &lt;= current version | Allow leadership |
+| Stored version > current version | Block leadership |
+| Unparseable version string | Allow leadership (fail-open), WARN log |
+
+Fail-open on parse errors avoids blocking clusters with unusual version
+strings (dev builds, custom tags).
+
+### Rationale and Alternatives
+
+#### Why solve this in OpenBao core vs. orchestration
+
+Production operators running OpenBao on Kubernetes have concluded that Helm
+cannot reliably enforce version ordering. The problem reduces to: "Helm hooks
+don't execute in all workflows, and StatefulSet lifecycle doesn't support
+'upgrade standbys before leader' semantics natively."
+
+Solving this in OpenBao core means **every** deployment method (Kubernetes,
+systemd, Docker Compose, bare metal) gets the protection automatically.
+
+#### Alternatives considered
+
+1. **Custom Kubernetes Operator.** Could enforce version ordering via pod
+   scheduling. Rejected because it adds a dependency, doesn't help non-K8s
+   deployments, and the OpenBao Secrets Operator serves a different purpose.
+
+2. **Extend existing `core/versions/` tracking.** The version store records all
+   historical leader versions. We could check against the highest recorded
+   version. Rejected because `handleVersionTimeStamps()` runs *after*
+   `postUnseal()` (too late for a pre-unseal check) and the historical record
+   has different semantics than "most recent leader."
+
+3. **Leader advertisement version check.** Store version in
+   `activeAdvertisement`. Rejected because advertisements are ephemeral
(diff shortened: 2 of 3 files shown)


markmishaev76:
CI requires either a changelog file or the `pr/no-changelog` label. Since this is a docs-only RFC, could a maintainer add `pr/no-changelog`? (Same as #2780.)

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,311 @@
...
+2. **Cascading failure scenario.** If all new-version nodes crash (bad
+   upgrade), old nodes are blocked from leadership, resulting in an outage.
+   This matches the existing documented behavior ("do not fail over to older
+   version"). The `disable_leader_version_check` config option provides an
+   escape hatch. Mitigation: backup before upgrade, test in staging.
+
+3. **Slightly longer leader election for blocked nodes.** Version-blocked nodes
+   acquire the lock, check, release, and sleep 10s before retrying. This adds a
+   brief period where the lock is held but no leader is active. In practice,
+   this window is sub-second (lock acquire + barrier read + lock release).
```
phil9909, line 250: We could work around this by doing the following:

1. check, on failure: log, sleep, retry
2. lock
3. check again, on failure: unlock, log, sleep, retry
4. start unsealing

This should be fine, as reading the leader version should not require holding the lock.

phil9909: Forget about that, this only helps when there is currently no leader when starting up, but `lock` will potentially block for a very long time.

markmishaev76: NP, thanks for raising and iterating.

markmishaev76: @phil9909  could you please resolve your comment?

markmishaev76:
Gentle ping — the open thread with @phil9909 has been addressed, and the PR has been sitting for a week with no blockers flagged. @cipherboy @satoqz when you have a moment, could you take a pass? Happy to iterate on anything;

website/content/docs/rfcs/index.mdx
```diff
@@ -27,6 +27,9 @@ Steering Committee.
    necessary constructs to build indexes on data efficiently allowing for faster search operations.
  - [Emergency Seal](emergency-seal.mdx), for adding a secondary Shamir's
    seal as a "break-glass" recovery path for auto-seal failures.
+ - [Version-Gated Leader Election](version-gated-leader-election.mdx), to
+   prevent older nodes from becoming HA leader after a newer version has held
+   leadership, enabling zero-downtime rolling upgrades.
```
wslabosz-reply: phrasing nit: 
```suggestion
   prevent nodes from becoming HA leader after a node running a newer version has held
   leadership, enabling zero-downtime rolling upgrades.
```

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,311 @@
+---
+sidebar_label: Version-Gated Leader Election
+description: |-
+  An OpenBao RFC on preventing older nodes from becoming HA leader after a newer
+  version has held leadership, enabling zero-downtime rolling upgrades.
```
wslabosz-reply:
```suggestion
  An OpenBao RFC on preventing nodes from becoming HA leader after a node running a newer
  version has held leadership, enabling zero-downtime rolling upgrades.
  
```

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,311 @@
...
+1. **Should the version check use full semver or just major.minor?** The
+   current proposal uses full semver comparison (e.g., 2.5.1 > 2.5.0). An
+   argument could be made for only checking major.minor since patch releases
+   shouldn't introduce breaking storage changes. Using full semver is more
+   conservative.
+
+2. **Should there be a `bao operator reset-leader-version` command?** For
+   advanced recovery scenarios beyond the config escape hatch, a CLI command to
+   clear the `core/leader-version` entry could be useful. This could be
+   deferred to a follow-up.
```
wslabosz-reply: starting openbao in recovery mode could be used to achieve same efffect afaik, no need to add additional cli cmd imho.

wslabosz-reply (approved):
I think the proposal is sound and without any major fallbacks

website/content/docs/rfcs/version-gated-leader-election.mdx
markmishaev76: Good call — agreed, recovery mode covers this. Dropped the open question and noted recovery mode under the cascading-failure downside in c679fff.

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+   current proposal uses full semver comparison (e.g., 2.5.1 > 2.5.0). An
+   argument could be made for only checking major.minor since patch releases
+   shouldn't introduce breaking storage changes. Using full semver is more
+   conservative.
+
+2. **Should version-blocked nodes participate in read replication?** Currently
+   proposed: yes, they continue serving reads. An argument could be made that
+   older-version standbys shouldn't serve reads at all during upgrades. However,
+   blocking reads would reduce availability during the upgrade window, which is
+   counter to the zero-downtime upgrade goal.
```
Huy-Dinh-Reply, line 281: I'm wondering if this can be a configurable parameter so that the operator can choose their preferred tradeoff. I can see value in both:
* allowing version-blocked nodes to participate, which improves availability.
* blocking them from serving reads, which supports stronger integrity, in case the newer version has a storage change that may make the standby node interpret the data incorrectly. Optionally the version blocked standbys can forward requests to the leader until they are upgraded, but this puts more read loads on the leader and probably should be documented as part of the tradeoff.

markmishaev76: Reasonable extension, and the plumbing's already there. `disable_standby_reads` plus `StandbyReadsEnabled()` already gate local-read vs forward-to-leader, so a version-blocked standby can just behave as if standby reads were disabled. One condition in `StandbyReadsEnabled()`, no new request path.

A few caveats if we add it:
- Standbys are already barrier-read-only, and #2467/#2549 closed the standby-writes-during-setup gap. Most of the integrity risk is on the write path, which is already blocked.
- The narrow remaining risk (newer leader writes a schema older standby misreads) applies to read replication generally whenever versions diverge. It isn't unique to the version-blocked state.
- Forwarding all reads to the leader during a rolling upgrade can cause a thundering-herd on busy clusters. Worth documenting as the tradeoff.

Proposed shape, defaulting to current behaviour:

```hcl
version_blocked_standby_reads = "allow" | "forward"  # default: allow
```

@cipherboy @satoqz: preference on folding into this RFC vs. follow-up?

markmishaev76:
Status check — approvals: 1 (@wslabosz-reply, thanks!), CI green, all earlier threads addressed.

The only thing keeping this open is the question on line 281: whether to fold the configurable `version_blocked_standby_reads` knob raised by @Huy-Dinh-Reply into this RFC, or land the RFC as-is and track the knob as a follow-up. I have no strong preference and can go either way.

@cipherboy @satoqz — a one-line steer from either of you would unblock this. @wslabosz-reply, mind resolving the recovery-mode thread if you're happy with the c679fff change?

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+replication landing in v2.5.0
+([#1986](https://github.com/openbao/openbao/pull/1986)), standbys now call
+`postUnseal()` in read-only mode
+([`runStandbyOnce()`](https://github.com/openbao/openbao/blob/main/vault/ha.go#L493)),
+loading mount tables, plugins, and caches to serve read requests locally. A
+version-mixed cluster during rolling upgrades risks:
+
+1. **An older node claiming leadership after a newer version modified
+   storage.** `postUnseal()` performs storage migrations. An older version may
+   not understand schema changes made by the newer leader.
```
cipherboy, line 43: FWIW, this is true on both pre-2.5.x (or 2.5.x with `disable_standby_reads`) and post-2.5.x with read-scalability enabled.

This also doesn't really apply to PostgreSQL (run by your organization), only Raft currently has read replication / standby loading, though #3027 will eventually address it (likely 2.7.x as I doubt anyone has capacity to review for 2.6.x).

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+2. **Older standbys serving stale reads.** With read replication, standbys
+   process requests using their own mount table and plugin state. Different
+   versions could interpret cached state differently.
+
+Recent fixes to standby behavior confirm this is an active concern:
+[#2467](https://github.com/openbao/openbao/pull/2467) fixed standbys writing
+mount table upgrades during read-replication setup, and
+[#2549](https://github.com/openbao/openbao/pull/2549) fixed standbys
+scheduling lease expiration jobs requiring storage writes. Both would be worse
+in a version-mixed cluster.
```
cipherboy, line 53: I don't agree that these are evidence of that; this is more of a bug w.r.t. unseal process and node state tracking.

Standbys don't have writable storage, so this caused the standby node to bounce, effectively functioning as a forward-only cold standby node. It wouldn't have caused any problems w.r.t. active node functionality regardless of what version the active node was on.

(These would, however, result in a decrease in horizontal capability during the upgrade window.)

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+old-version nodes continue serving standby reads until replaced.
+
+If an operator accidentally rolls back to an older image, old-version nodes
+refuse leadership:
+
+```
+[WARN] core: refusing leadership: current version 2.5.0 is older than last leader version 2.6.0
+```
+
+All nodes enter standby mode, making the problem visible. For intentional
```
cipherboy, line 74:
```suggestion
All downgraded nodes enter standby mode, making the problem visible to the operator. For intentional
```

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+rollbacks (e.g., restore from backup), operators set a config option:
+
+```hcl
+disable_leader_version_check = true
```
cipherboy, line 78: I think we need to invert this behavior as it is a rather breaking operational change if people are only softly following the guidance. `leader_version_check = false` perhaps needs to be the default for a bit, with `leader_version_check = true` being a preferred operational safeguard.

My 2c.

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+
+If an operator accidentally rolls back to an older image, old-version nodes
+refuse leadership:
+
+```
+[WARN] core: refusing leadership: current version 2.5.0 is older than last leader version 2.6.0
+```
+
+All nodes enter standby mode, making the problem visible. For intentional
+rollbacks (e.g., restore from backup), operators set a config option:
```
cipherboy, line 75: Um...?

Why would an restore from backup trigger this check? Snapshot/backup would be taken _before_ upgrade (per [our docs](https://openbao.org/docs/upgrading/)) and thus would not contain the new storage entry.

This would only trigger on a non-backup-restore rollback.

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+
+### User/Developer Experience
+
+No change for normal upgrades -- they just work safely. On K8s, operators can
+switch to `RollingUpdate` strategy. Emergency rollback requires setting
+`disable_leader_version_check = true`.
+
+No impact on plugin or storage backend developers. The version check is
+entirely in the core leadership path.
+
```
cipherboy, line 268: This document does not share how this impacts OpenBao developers or other users consuming pre-release branches.

I think we'll need to adjust the full version: https://github.com/openbao/openbao/blob/main/version/version_base.go#L25 -- right now, we default to 2.0.0-HEAD; this change will mean we'll have to start maintaining that version properly on all branches (bumping after release).

cipherboy: I'm also not really sure what a "storage backend developer" is -- they aren't pluggable. I think they're just OpenBao developers? 🤔

website/content/docs/rfcs/version-gated-leader-election.mdx
cipherboy: In particular, to expand on this:

In general, loading storage state on an older version is not safe: OpenBao aims to be backwards compatible, not forwards compatible. This means that if a newer leader is elected while old standbys still exist, we could end up panicing or otherwise behaving incorrectly, depending on the contents of the update and the newly revised storage entries.

In the worst case, we could migrate fields in roles (in e.g., PKI or SSH) to new versions, and allow certificate issuance with fewer validations than expected.

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+#### Observability
+
+- K8s pod label `openbao-version-blocked=true` when a node is blocked
+  (alongside existing `openbao-active`, `openbao-sealed`, etc.)
+- Counter metric `core.leadership_version_blocked`
+- Clear WARN log on each blocked election attempt
+
+With this in core, the Helm chart can switch to `RollingUpdate` as the default
+StatefulSet update strategy
+([openbao-helm#170](https://github.com/openbao/openbao-helm/issues/170)).
```
cipherboy, line 90: Operators will need to know that capacity will decrease though, yes? If we step down from active leadership on the elected node, we delay leadership election (at worst, until all such outdated standbys have been given a chance to deny leadership and an updated standby gets elected). This could give us much less throughput in the worst case scenario.

cipherboy: cc: @openbao/openbao-k8s-committers

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+| `serviceregistration/kubernetes/service_registration.go` | `openbao-version-blocked` pod label |
+
+#### New storage entry: `core/leader-version`
+
+```go
+type leaderVersionEntry struct {
+    Version   string    `json:"version"`
+    Timestamp time.Time `json:"timestamp"`
+}
+```
```
cipherboy, line 111: Why do we need a new storage entry? 

```go
func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c *Core) error {
    ...

    // Establish version timestamps at the end of unseal on active nodes only.
    return c.handleVersionTimeStamps(ctx)
}
```

The existing `core/versions/{version}` path is written at the same time (ok, one operation before) the proposed entry per the "Changes to `waitForLeadership()` flow" section and contains everything we'd write here. We can't know for sure that the _current_ leader is on this version (as noted below in Downsides: a pre-version-gate node could've taken over), but with the new behavior, we'll know that we won't want to load the standby if a newer version exists anyways. 

We can just use a `List(...)` operation instead of a `Get(...)` operation in our version check path and only load the entry if we need the timestamp.

(n.b., we should avoid `ListPage(..., after=$STANDBYVERSION)` as this is a lexicographical ordering and not a SemVer ordering.) 

My 2c.

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+
+#### Why solve this in OpenBao core vs. orchestration
+
+Production operators running OpenBao on Kubernetes have concluded that Helm
+cannot reliably enforce version ordering. The problem reduces to: "Helm hooks
+don't execute in all workflows, and StatefulSet lifecycle doesn't support
+'upgrade standbys before leader' semantics natively."
+
+Solving this in OpenBao core means **every** deployment method (Kubernetes,
+systemd, Docker Compose, bare metal) gets the protection automatically.
```
cipherboy, line 201: I will note, we haven't actually _solved_ anything. We've strengthened the check to prevent downgrades, and opted to trade off less availability for more correctness.

If an operator uses the correct strategy, they get rewarded with more availability. But for operators using what is proposed here -- a blind upgrade strategy, availability will suffer, and suffer in a way that may result in a total outage if an upgraded node becomes leader but consistently faults.

Want to make sure we call that out and it is why I:

1. Wouldn't advocate for changing the helm chart.
2. Wouldn't suggest this behavior by default. 

My 2c.

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+
+#### Alternatives considered
+
+1. **Custom Kubernetes Operator.** Could enforce version ordering via pod
+   scheduling. Rejected because it adds a dependency, doesn't help non-K8s
+   deployments, and the OpenBao Secrets Operator serves a different purpose.
```
cipherboy, line 207: 😆 

We can name a hypothetical new thing something else, like OpenBao Upgrade Operator. It doesn't have to be OpenBao Secrets Operator.

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+
+2. **Extend existing `core/versions/` tracking.** The version store records all
+   historical leader versions. We could check against the highest recorded
+   version. Rejected because `handleVersionTimeStamps()` runs *after*
```
cipherboy, line 211: This is a misunderstanding.

[`postUnseal(...)`](https://github.com/openbao/openbao/blob/076278fba2e3035855f90890344ffc29e70cab0a/vault/core.go#L2284-L2359) calls [`unsealer.unseal(...)`](https://github.com/openbao/openbao/blob/076278fba2e3035855f90890344ffc29e70cab0a/vault/core.go#L2324-L2326), which on the [active node is `standardUnsealStrategy{}`](https://github.com/openbao/openbao/blob/076278fba2e3035855f90890344ffc29e70cab0a/vault/ha.go#L783) which [calls `handleVersionTimeStamps(...)` as the last thing it does](https://github.com/openbao/openbao/blob/076278fba2e3035855f90890344ffc29e70cab0a/vault/core.go#L2146-L2189). This means technically this gets called _before_ `postUnseal(...)` returns.

This is semantically equivalent to what is proposed above.

It also doesn't stop a standby from reading this entry, fwiw; it is in storage.

website/content/docs/rfcs/version-gated-leader-election.mdx
```diff
@@ -0,0 +1,309 @@
...
+   `postUnseal()` (too late for a pre-unseal check) and the historical record
+   has different semantics than "most recent leader."
+
+3. **Leader advertisement version check.** Store version in
```
cipherboy, line 215: This is a misunderstanding again of how leader advertisement works. By definition when an active node loses leadership less-than-gracefully, it cannot do anything, including writing to storage (ideally). 

Former leader advertisements are cleaned up by the new active node. So a standby gaining leadership _could_ check all present advertisements (prior to deletion), to ensure that there's no newer version listed.

cipherboy: @markmishaev76 Looking at how this would behave versus your proposal of a new storage entry (entirely duplicated by proposal 2 above) -- I think I prefer this option.

This entry is written _before_ `postUnseal(...)` which means:

1. We can check existing entries on the active leadership path prior writing it.
2. On read-enabled standbys, when reading the storage entry, we can warn a version mismatch exists.

cipherboy (commented):
@markmishaev76 Highlighting the most important comments:

 - https://github.com/openbao/openbao/pull/2865/changes#r3251478310 is a better design for this feature as intended.
 - https://github.com/openbao/openbao/pull/2865/changes#r3251425618 is worth reading and understanding, and getting feedback from @openbao/openbao-core-committers , @openbao/openbao-org-maintainers, and @openbao/openbao-k8s-committers on the behavior of this new proposal.


---

Based on this document, AI generated RFCs are a bad idea. 

@openbao/dev-wg @karras I'd like to propose we extend our contributing guidelines to reject AI-generated RFCs and website content as well. It lowers the value of our community.

website/content/docs/rfcs/version-gated-leader-election.mdx
cipherboy: For reasons [I articulated above](https://github.com/openbao/openbao/pull/2865/changes#r3251267970), I believe forwarding reads is the only solution in this case. 

I think it is out of scope for this, so we should probably create a second RFC around it. If we take my proposal https://github.com/openbao/openbao/pull/2865/changes#r3251478310, it would give us a clean solution to this as well.

website/content/docs/rfcs/version-gated-leader-election.mdx
cipherboy: Articulating further on why this is more reliable:

In the current design, by writing after `postUnseal(...)` completed, we know that the active node fully started up on the specified version. However, if the active node does some amount of migrations (unknown) and then fails (panics, errs, whatever), we would not record that it modified storage, and could potentially fail to start on an outdated node (which, would think it is OK to do as the new version was never recorded).

This would be written prior to any other migrations and thus fail correctly (on outdated nodes), allowing newer versions to successfully acquire leadership (and maybe potentially succeed this time).

markmishaev76:
Closing this PR. Thanks for the thorough review, @cipherboy. We'll wait for the upstream fix rather than iterate further here.

Appreciate everyone who took time to review.

PR #2874, opened 2026-04-13
Update workflow documentation for input block

cipherboy:
Also fixes the plurality of the `fields` block; HCL config blocks are generally singular even though they're specified multiple times.

## Rationale

See also: [#openssf-openbao-discussion > Profiles @ 💬](https://linuxfoundation.zulipchat.com/#narrow/channel/529890-openssf-openbao-discussion/topic/Profiles/near/585260328)

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/concepts/profiles.mdx
+++ b/website/content/docs/concepts/profiles.mdx
@@ -10,9 +10,9 @@ supporting chained operations, allowing multiple API calls from a single
 entry point. This can be parsed from HCL or JSON as desired.
 
 Data sources are dependent on the profile call site: self-initialization has
-one set of allowed sources whereas API profiles have a different combination.
-Data sources can be nested; for instance, a template source can be used to
-combine an environment variable and the contents of a file.
+one set of allowed sources whereas the workflows API has a different
+combination. Data sources can be nested; for instance, a template source can
+be used to combine an environment variable and the contents of a file.
 
 :::info
 
@@ -28,7 +28,7 @@ unauthenticated rotation and audit device creation) still apply.
 
 A privileged operator designs the profile; this involves creating a
 configuration which describes what API endpoints to call. If this profile is
-executed from the API profile system, it also involves defining input and
+executed from the API workflow system, it also involves defining input and
 output parameters.
 
 This configuration either ends up in one or more files (if using declarative
@@ -80,12 +80,12 @@ The core of a profile is the request, which contains the following fields:
  - `path` `(string: <required>)` - which API endpoint to invoke.
  - `token` `(string: <optional>)` - the OpenBao token to use to authenticate
    with; in some cases (like self-initialization or authenticated execution
-   of API profiles), this is inferred from the context automatically. However,
-   in the case of unauthenticated execution of API profiles, this would need
-   to be manually handled as none exists in the execution environment.
-   Unauthenticated APIs (such as login requests) can continue to be called
-   without this parameter or with this parameter explicitly set to the empty
-   string.
+   of the workflows API), this is inferred from the context automatically.
+   However, in the case of unauthenticated execution of the workflows API,
+   this would need to be manually handled as none exists in the execution
+   environment. Unauthenticated APIs (such as login requests) can continue to
+   be called without this parameter or with this parameter explicitly set to
+   the empty string.
  - `data` `(map[string]any: <optional>)` - any parameters for this request.
  - `headers` `(map[string][]string: <optional>)` - any headers for this
    request.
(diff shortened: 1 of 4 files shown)


website/content/docs/concepts/profiles.mdx
```diff
@@ -114,15 +114,63 @@ request "create-admin-policy" {
...
 
 When doing so, the name of this outer grouping block is used for the `request`
 and `response` dynamic data source and must be specified.
 
+## Input
+
+When using the workflows API, input passed by the caller can be specified in
+an `input` block. This has the following named child blocks:
+
+ - `fields` `([]FieldSchema: <optional>)` - a field definition to be parsed
```
pree: I guess `field` now? :)

website/content/docs/concepts/profiles.mdx
```diff
@@ -114,15 +114,63 @@ request "create-admin-policy" {
...
+   from the API request's data.
+
+### FieldSchema
+
+The `FieldSchema` definition takes the following values:
+
+ - `type` `(string: <required>)` - the type of this field; this is specified
+   as the first key parameter on the block definition.
```
satoqz: I'd vote to make this a `type` field inside of the block opposed to an HCL key.

As shown in https://github.com/openbao/openbao/pull/2878:

```hcl
input {
  fields "string" "policy" { required = true }
  fields "string" "mount" { required = true }
  fields "string" "secret" { required = true }
}
```

looks very odd to me with the type preceding the name, and it being decoupled from "required" which is secondary type information.

pree: This is the same as OpenTofu works for example, but my intuition reading the docs was:
```hcl
fields "policy" {
  type = "string"
  required = true
}
```

I guess similar to how variables in OpenTofu work:

```hcl
variable "policy" {
  type    = string
  default = "example"
}
```

website/content/docs/concepts/profiles.mdx
```diff
@@ -114,15 +114,63 @@ request "create-admin-policy" {
...
+ - `name` `(string: <required>)` - a name to reference this field by; this is
+   used in the CEL and templating engines below.
+ - `default` `(any: <optional>)` - the default field value, if the field is
+   not required.
+ - `description` `(string: <optional>)` - a description of this field.
+ - `required` `(bool: false)` - whether this field is required.
```
satoqz, line 144: I wonder if `optional` is the wiser choice opposed to `required`, I'd expect most field definitions to be required in practice, so inverting the default to `required = true` via `optional = false` may be good.

website/content/docs/concepts/profiles.mdx
cipherboy: @satoqz I think this is the pattern set forth elsewhere, no? 

```
listener "tcp" 
```

or

```
plugin "type" "name" {
```

or 

```
audit "file" "to-stdout" {
```

so if we have, we've messed this up elsewhere. 

It looks like there was a bug, but I expected all of these to work:

```hcl
  fields {
    type = "string"
    name = "namespace"
    description = "name of the namespace to create"
    required = true
  }

  fields "namespace" {
    type = "string"
    description = "name of the namespace to create"
    required = true
  }

  fields "string" {
    name = "namespace"
    description = "name of the namespace to create"
    required = true
  }
```

will fix!

website/content/docs/concepts/profiles.mdx
cipherboy: @satoqz This mirrors what we have in FieldSchema: https://github.com/openbao/openbao/blob/0c2a0f8d6276c1dccf3c7b867c71cbdc607591f6/sdk/framework/backend.go#L687-L689

Though, I guess it isn't enforced by us? :thinking: I would've thought `Required` would've been. 

Maybe we condition around `Default`? 

Thoughts here?

website/content/docs/concepts/profiles.mdx
satoqz: @cipherboy here's my take at rationalizing my intuition on the design:

- With `listener` and `audit`, the type deserves to be outside of the block as it dictates the parameters available inside of it (i.e., `listener "tcp"` takes different parameters than `listener "unix"`).

- With `plugin`, it helps keep the same name across plugins that differ in type, e.g., `"secrets" "aws"` and `"auth" "aws"`. They're effectively different sub-maps.

- With `field`, we still take the same parameters inside of the block (unlike `listener` and `audit`) and we still add all fields to the same namespace no matter their type (unlike `plugin`). Hence giving special treatment to `type` appears odd.

website/content/docs/concepts/profiles.mdx
satoqz: @cipherboy that's something I dislike about `FieldSchema` 😄 Having to remember to explicitly set `Required: true` hasn't been fun (even if it turns out it doesn't do much beyond OpenAPI).

Conditioning around `default` seems reasonable given I don't think there is a way for profiles to check if a supplied value was defaulted to automatically via `required = false` or defaulted to via `default` anyway?

website/content/docs/concepts/profiles.mdx
cipherboy: @satoqz See https://github.com/openbao/openbao/pull/2881. 

Also, `listener` and `audit` and `plugin` behave the same way (well, everywhere taking a named block does): 

https://github.com/openbao/openbao/blob/0c2a0f8d6276c1dccf3c7b867c71cbdc607591f6/helper/configutil/listener.go#L217-L223

https://github.com/openbao/openbao/blob/0c2a0f8d6276c1dccf3c7b867c71cbdc607591f6/command/server/config.go#L1534-L1548

https://github.com/openbao/openbao/blob/0c2a0f8d6276c1dccf3c7b867c71cbdc607591f6/command/server/config.go#L440-L454

Though, arguably both audit and plugins should be updated to support single-value conditional key of `type` and a `name` that I've added to `fields`. :shrug: 

But the point still stands, if you want explicit naming of what these block keys are, you can conditionally name both of them in the body of the block and elide (currently elsewhere) both of them from the keys spot.

website/content/docs/concepts/profiles.mdx
cipherboy: @satoqz No, FieldSchema's definition is just wrong:

https://github.com/openbao/openbao/blob/0c2a0f8d6276c1dccf3c7b867c71cbdc607591f6/sdk/framework/field_data.go#L108-L115

so I'll update the comment. `default` is not sufficient and my documentation here is correct. :laughing:

cipherboy:
@satoqz @pree Revised, LMKWYT! Hopefully block keys are more clear now.

website/content/docs/concepts/profiles.mdx
satoqz: I'll concede to the pattern then. I still think

```hcl
input "name" {
  type = "type"
}
```

will be the natural way to use it, but fine with me to have multiple patterns.

pree:
I guess with the change from `fields` to `field` the tests also need adjustment

satoqz (approved):

PR #2876, opened 2026-04-13
Use unique constraints, indices in PostgreSQL storage

cipherboy:
When creating multiple OpenBao instances with the same database in the PostgreSQL storage backend (to speed up testing), we'll want to keep `skip_create_table=false`. However, just setting `table` (and `ha_table`) isn't sufficient: we'll end up with constraint conflict errors like:

> ```
> error initializing storage of type postgresql: failed to create
> tables: failed to execute create query: ERROR: relation "pkey"
> already exists (SQLSTATE 42P07)
> ```

because of the `pkey` relation already existing in a separate table for a separate test.

## Acknowledgements

       also allowed.


satoqz (approved):
Makes sense to me.

wslabosz-reply (approved):

PR #2882, opened 2026-04-14
Add Namespace Sealing

wslabosz-reply:
## Description
- added `handleNamespacesSeal()` to `vault/logical_system_namespaces_seals.go` 
- added `(*NamespaceStore).SealNamespace()` method
- added check for sealed namespaces in `(*NamespaceStore).DeleteNamespace()` method to reject deletion requests for sealed namespace (sealed namespace implementation yet to be done)
- added removal of namespace from `SealManager` memory in `(*namespaceDeletionJob).Execute()` method

## Rationale
Implements namespace sealing logic for the per-namespace seals feature 

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/logical_system_namespaces_seals.go
+++ b/vault/logical_system_namespaces_seals.go
@@ -52,6 +52,31 @@ func (b *SystemBackend) namespaceSealPaths() []*framework.Path {
 
 	return []*framework.Path{
 		{
+			Pattern: "namespaces/(?P<path>.+)/seal",
+			DisplayAttrs: &framework.DisplayAttributes{
+				OperationPrefix: "namespaces",
+			},
+			Fields: map[string]*framework.FieldSchema{
+				"path": namespacePathSchema,
+			},
+
+			Operations: map[logical.Operation]framework.OperationHandler{
+				logical.UpdateOperation: &framework.PathOperation{
+					Summary:  "Seal a namespace.",
+					Callback: b.handleNamespacesSeal(),
+					Responses: map[int][]framework.Response{
+						http.StatusNoContent: {{
+							Description: http.StatusText(http.StatusNoContent),
+						}},
+					},
+				},
+			},
+
+			HelpSynopsis:    strings.TrimSpace(sysNamespacesSealsHelp["namespaces-seal"][0]),
+			HelpDescription: strings.TrimSpace(sysNamespacesSealsHelp["namespaces-seal"][1]),
+		},
+
+		{
 			Pattern: "namespaces/(?P<path>.+)/unseal",
 			DisplayAttrs: &framework.DisplayAttributes{
 				OperationPrefix: "namespaces",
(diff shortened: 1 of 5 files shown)


vault/logical_system_namespaces_seals.go
```diff
@@ -51,6 +51,31 @@ func (b *SystemBackend) namespaceSealPaths() []*framework.Path {
 	}
 
 	return []*framework.Path{
+		{
+			Pattern: "namespaces/(?P<path>.+)/seal",
```
cipherboy, line 55: Should this just be `GenericNameRegex` so we never have more than one segment now? 🤔

cipherboy (approved):
One nit, but after the #2854 rebase I think it looks good.

vault/logical_system_namespaces_seals.go
satoqz: Would `GenericNameRegex` be compatible with names previously allowed via https://github.com/openbao/openbao/blob/7ac9e1a598aaf021ed7cef4ff41484d579eb6a7e/helper/namespace/namespace.go#L54?

cipherboy: @satoqz Interesting question! I think the answer is no, we allow weird things like `%` and `\` as namespace names.

cipherboy: > @satoqz Interesting question! I think the answer is no, we allow weird things like `%` and `\` as namespace names.

Which is rather unfortunate.

wslabosz-reply: > Should this just be GenericNameRegex so we never have more than one segment now? 🤔

we have same pattern in all namespace related paths, in the handlers itself we check if there is more than one segment, err in that case.

wslabosz-reply:
@cipherboy 
I've pushed a small update to reloading logic, I'd appreciate it if you could take a look

vault/logical_system_namespaces_seals.go
cipherboy: @wslabosz-reply Right, I was just thinking we could probably change it to `(?P<path>[^/]+)` or something to avoid matching multi-path segments, but this is fine for now of course.

vault/namespace_store.go
```diff
@@ -1107,6 +1134,10 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
 		return "", nil
 	}
 
+	if ns.core.NamespaceSealed(namespaceToDelete) {
```
cipherboy, line 1137: @wslabosz-reply How does tainting interact with sealed namespace deletion? Take the test from #2935 for instance; if we were to do that on a sealed namespace, can we unseal a tainted namespace? 🤔 How much has to remain (in storage) for us to consider namespace not yet deleted/tainted and still sealed? Can we unseal a sealed, tainted namespace?

cipherboy (approved):

vault/namespace_store.go
cipherboy: Happy addressing this in a follow-up tbh.

cipherboy:
In order to keep the command line stuff moving, I'm going ahead and merging this and we'll address deletion later as it isn't really related to sealing/unsealing. Thanks @wslabosz-reply!

PR #2883, opened 2026-04-14
Ensure workflow parses on update

cipherboy:
Previously, OpenBao did not validate unused keys in profiles; this meant many forms of errors were silently ignored. Adding validation to both self-initialization and workflows API will help avoid this problem. Notably, the workflows API was lacking any set-time parsing, resulting in only runtime execution failures.

## Acknowledgements

       also allowed.


satoqz (approved):
Makes sense to me.

cipherboy:
Thanks @satoqz!

PR #2900, opened 2026-04-16
CLI support for sealed namespace creation

phyrog:
## Description

- SealConfigs can now be passed as an HCL file to the bao namespace create command
- changes the format of the `seal` field of the namespace create API. Previously this was directly a json map requiring parsing on the client, now this field just contains the raw HCL content and parsing happens on the server.
  - This aligns better with how e.g. policies are handled and also works better in case of version drift between cli and server

## Rationale

Part of #1357, CLI support for what was introduced in the API already in #2682

## Acknowledgements

       also allowed.


changelog/2900.txt
wslabosz-reply, line 1: I'd skip adding a partial changelog, we'll probably do one big summary with the release of the feature.

vault/seal.go
```diff
@@ -275,47 +275,47 @@ func (d *defaultSeal) GetShamirWrapper() (*seal.ShamirWrapper, error) {
...
 
 	// PGPKeys is the array of public PGP keys used, if requested, to encrypt
 	// the output unseal tokens. If provided, it sets the value of
 	// SecretShares. Ordering is important.
-	PGPKeys []string `json:"pgp_keys" mapstructure:"pgp_keys"`
+	PGPKeys []string `json:"pgp_keys" mapstructure:"pgp_keys" hcl:"pgp_keys"`
 
 	// Nonce is a nonce generated by Vault used to ensure that when unseal keys
 	// are submitted for a rotation, the operation itself is the one intended.
-	Nonce string `json:"nonce" mapstructure:"nonce"`
```
wslabosz-reply, line 295: we only provide the nonce on "rotation" config, so no need to include that with hcl

command/namespace_create.go
```diff
@@ -59,6 +62,13 @@ func (c *NamespaceCreateCommand) Flags() *FlagSets {
 			"This can be specified multiple times to add multiple pieces of metadata.",
 	})
 
+	f.StringVar(&StringVar{
+		Name:       "seals",
```
wslabosz-reply: let's keep the name in singular form

vault/seal.go
phyrog: I also removed the `Backup` field from hcl, that is also set during rekey/rotate-keys.

command/namespace_create.go
```diff
@@ -59,6 +62,13 @@ func (c *NamespaceCreateCommand) Flags() *FlagSets {
 			"This can be specified multiple times to add multiple pieces of metadata.",
 	})
 
+	f.StringVar(&StringVar{
+		Name:       "seal",
+		Target:     &c.flagSealConfigPath,
+		Completion: complete.PredictFiles("*.json"),
```
wslabosz-reply:
```suggestion
		Completion: complete.PredictFiles("*.hcl"),
```

command/namespace_create.go
```diff
@@ -113,3 +133,17 @@ func (c *NamespaceCreateCommand) Run(args []string) int {
...
+func (c *NamespaceCreateCommand) readSealConfig() ([]byte, error) {
+	path := c.flagSealConfigPath
+	if path == "" {
+		return nil, nil
+	}
+
+	absolutePath, err := filepath.Abs(path)
+	if err != nil {
+		return nil, err
+	}
```
satoqz: Seems superfluous?

ragnar-howler (commented):

This PR has been scanned for:
- ✅ Common security issues
- ✅ Code style consistency  
- ✅ Test coverage indicators

**Stats:** +170 -28 in 4 files

wslabosz-reply (approved):
LGTM

command/namespace_create.go
```diff
@@ -59,6 +61,13 @@ func (c *NamespaceCreateCommand) Flags() *FlagSets {
 			"This can be specified multiple times to add multiple pieces of metadata.",
 	})
 
+	f.StringVar(&StringVar{
+		Name:       "seal",
+		Target:     &c.flagSealConfigPath,
+		Completion: complete.PredictFiles("*.hcl"),
+		Usage:      "Path to a HCL file with exactly one seal stanza.",
+	})
```
cipherboy, line 73: If we presently only support Shamir's, do we want to have a fast-track way of creating a Shamir's seal similar to `bao operator init`? `-key-shares=...`

vault/seal.go
```diff
@@ -275,47 +275,47 @@ func (d *defaultSeal) GetShamirWrapper() (*seal.ShamirWrapper, error) {
 // SealConfig is used to describe the seal configuration
 type SealConfig struct {
 	// The type, for sanity checking
-	Type string `json:"type" mapstructure:"type"`
+	Type string `json:"type" mapstructure:"type" hcl:",label"`
```
cipherboy:
```suggestion
	Type string `json:"type" mapstructure:"type" hcl:"type,label"`
```

Will allow what was described in another PR around self-init and let us do:

```hcl
seal {
  type = "shamir"
  ...
}
```

&c

cipherboy:
@phyrog Looks like there's a test failure after @wslabosz-reply's PRs, but otherwise looks like a good start. 

I'm fine with the refactor, though I wonder if we want to align the parsing of seals with the server, and use `ParseKMSes` and later collapse down to a single result? We may need to special case `shamir` handling, then. 

I think we'll eventually need to handle seal migration also (especially since auto-unsealing for namespaces doesn't land in this release), but that's for another time.

command/namespace_create.go
```diff
@@ -59,6 +61,13 @@ func (c *NamespaceCreateCommand) Flags() *FlagSets {
 			"This can be specified multiple times to add multiple pieces of metadata.",
 	})
 
+	f.StringVar(&StringVar{
+		Name:       "seal",
+		Target:     &c.flagSealConfigPath,
+		Completion: complete.PredictFiles("*.hcl"),
```
cipherboy:
```suggestion
		Completion: complete.PredictFilesSet([]string{"*.hcl", "*.json"}),
```

command/namespace_create.go
phyrog: I think that is the right call, thinking about the future rotate keys command, this will be more in line with the existing `bao operator` commands. So do we infer the seal document to be `seal "shamir" {}` when no seal config is given, but `-key-shares` etc is provided? key-shares and key-threshold would then be sent as separate payload fields, right?

phyrog:
> I wonder if we want to align the parsing of seals with the server, and use ParseKMSes and later collapse down to a single result? We may need to special case shamir handling, then.

@cipherboy I think I'd leave that until we tackle autoseals for namespaces, we'll know better about any special cases we'll have to implement then.

cipherboy:
@phyrog Hmm, if we don't do it now, won't it be hard to align? IIUC we're decoding into the internal storage `SealConfig` type now, which really is only relevant for Shamirs? Or did I misunderstand something?

phyrog:
@cipherboy Yeah no, I think I misunderstood what you meant initially. I reworked it to use `ParseKMSes` now.

I also moved `secret_shares` and `secret_threshold` out of the seal config, just like it is not in the seal config for the root barrier, and instead is a separate parameter send alongside the seal (like you would in `bao operator init`). `seal "shamir" {}` is now also inferred when only specifying `-key-shares` and `-key-threshold` on the CLI (though the API itself still requires a valid seal config to be sent, so inferring is only client-side).

cipherboy:
@phyrog Sorry about the delay.

I had a good chat with @wslabosz-reply on Wednesday that hopefully he'll tell you about, but the short of it is that I think we want to make sure we can align to a parallel unseal future. If I may change my mind slightly, I think `secret_shares` and `secret_threshold` (perhaps even dropping `secret_`) are the only two config parameters we might add to the seal config for Shamir's right now, to set us up for the future.

Will review more soon!

phyrog:
@cipherboy I moved `shares` and `threshold` back into the seal config. `pgp_keys` I left as a separate parameter. Main reasons for not moving it into the seal config too:
- parseKMS does not support list types (except for `purpose`)
- while it is persisted in the seal-config in storage, its only purpose is to encrypt the key shares in transit (to the recipient of the share), so it is more transactional and not an inherent property of the seal config (correct me if I'm wrong)

cipherboy:
@phyrog No, that's exactly right. You'd reuse `pgp_keys` for recovery key encryption as well, if we supported auto-unseal here now.

cipherboy (approved):
@phyrog Terribly sorry on how long it has taken me to review this, looks good to me!

In the future, we could perhaps use some of the `bao operator init` style printing to display both hex and base64 encoded shares, but that's for another time, as we'd need a whole new command line for that (`bao namespace create` already has a fixed format).

PR #2901, opened 2026-04-16
Core using seal manager for unsealing

wslabosz-reply:
## Description
- removed `unlockInfo` from `Core` struct with unlock information tracking moved to `SealManager` (for root namespace the information was stored in both places, but operations were done on the `Core` property)
- removed duplicated method from `Core`, opting for usage of the same methods in `SealManager`
- adjusted `(*Core).GetSealStatus()` method
- adjusted `SealManager` code accordingly

## Rationale
With the introduction of `SealManager` we should track and manage all unseal operations through it. This PR aims at consolidation of the unseal logic, making the root namespace (core) use seal manager methods.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/seal_manager.go
+++ b/vault/seal_manager.go
@@ -107,13 +107,8 @@ func (sm *SealManager) Reset() {
 		namespace.RootNamespaceUUID: sm.core.seal,
 	}
 
-	sm.unlockInformationByNamespace = map[string]*unlockInformation{
-		namespace.RootNamespaceUUID: {},
-	}
-
-	sm.rotationConfigByNamespace = map[string]*rotationConfig{
-		namespace.RootNamespaceUUID: {},
-	}
+	sm.unlockInformationByNamespace = map[string]*unlockInformation{}
+	sm.rotationConfigByNamespace = map[string]*rotationConfig{}
 }
 
 // SetSeal creates a seal with provided config and sets it as provided namespace seal;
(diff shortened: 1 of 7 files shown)


vault/seal_manager.go
```diff
@@ -447,7 +447,7 @@ func (sm *SealManager) getUnsealKey(ctx context.Context, seal Seal, ns *namespac
 	}
 
 	defer func() {
-		delete(sm.unlockInformationByNamespace, ns.UUID)
```
cipherboy, line 450: Not sure I understand the subtleties around `delete` vs `nil` assignment?

cipherboy: (Otherwise I'm happy with this once rebased)

wslabosz-reply: @cipherboy 
actually I also wonder what's the preferred way, should we store a map entry in both `unlockInformationByNamespace` and `rotationConfigByNamespace` as long as namespace exists, or should we only populate them when we call `unlock` or `rotate` respectively?

cipherboy: @wslabosz-reply Let's trim the map, there's no reason to keep all the namespace info in memory here given both rotate and unlock will be rarer operations.

wslabosz-reply: @cipherboy done!

cipherboy (approved):
Thanks @wslabosz-reply!

PR #2908, opened 2026-04-17
Make self-init failures fatal

### LINKED ISSUE #2190: Declarative Self-Init: Make failures fatal

satoqz:
## Description

Cleaned up alternative for #2442 that plays well with Auto Seals being async and has an end to end test to actually validate this.

## Rationale

https://github.com/openbao/openbao/pull/2442#discussion_r3099356543

Resolves: #2190

## Acknowledgements

       also allowed.

### DIFF
--- a/helper/profiles/profiles.go
+++ b/helper/profiles/profiles.go
@@ -388,13 +388,15 @@ func (p *ProfileEngine) evaluateRequest(ctx context.Context, history *Evaluation
 	// 4. Call the request handler.
 	resp, err := p.requestHandler(ctx, req)
 	isFailure := err != nil || resp.IsError()
-	if err == nil && resp.IsError() {
+	if err == nil {
 		err = resp.Error()
 	}
 	if !allowFailure && isFailure {
 		if err != nil {
 			return fmt.Errorf("failed to evaluate request: %w", err)
 		}
+		// Fallback in case IsError() is true but Error() returned nil.
+		return errors.New("failed to evaluate request: request failed")
 	}
 
 	// 5. Stash response for future use.
--- a/vault/external_tests/postgresql_binary/postgresql_test.go
+++ b/vault/external_tests/postgresql_binary/postgresql_test.go
@@ -9,8 +9,10 @@ import (
 	"testing"
 	"time"
 
+	log "github.com/hashicorp/go-hclog"
 	"github.com/openbao/openbao/api/v2"
 	"github.com/openbao/openbao/sdk/v2/helper/consts"
+	"github.com/openbao/openbao/sdk/v2/helper/logging"
 	"github.com/openbao/openbao/sdk/v2/helper/testcluster"
 	"github.com/openbao/openbao/sdk/v2/helper/testcluster/docker"
 	"github.com/openbao/openbao/sdk/v2/physical"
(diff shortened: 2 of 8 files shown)


helper/profiles/profiles.go
```diff
@@ -388,13 +388,15 @@ func (p *ProfileEngine) evaluateRequest(ctx context.Context, history *Evaluation
 	// 4. Call the request handler.
 	resp, err := p.requestHandler(ctx, req)
 	isFailure := err != nil || resp.IsError()
-	if err == nil && resp.IsError() {
+	if err == nil {
 		err = resp.Error()
 	}
```
satoqz, line 393: `resp.Error()` already checks `resp.IsError()`. Neither do we need to check for `resp == nil`, which is already checked by `resp.IsError()`.

vault/external_tests/postgresql_binary/postgresql_test.go
```diff
@@ -200,3 +202,58 @@ func TestPostgreSQL_ParallelInit(t *testing.T) {
...
+			// NewDockerCluster directly.
+			ClusterName: strings.ReplaceAll(t.Name(), "/", "-"),
+			Logger:      logging.NewVaultLogger(log.Trace).Named(t.Name()),
+		},
+	}
+
+	cluster, err := docker.NewDockerCluster(t.Context(), opts)
+
+	// Don't forget to clean up just in case the assertion below fails and the
+	// test cluster didn't fail as epxected.
```
wslabosz-reply:
```suggestion
	// test cluster didn't fail as expected.
```

wslabosz-reply (approved):

vault/init.go
```diff
@@ -484,7 +484,7 @@ func (c *Core) UnsealWithStoredKeys(ctx context.Context) error {
 
 	// Disallow auto-unsealing when migrating
 	if c.IsInSealMigrationMode(true) && !c.IsSealMigrated(true) {
-		return NewNonFatalError(errors.New("cannot auto-unseal during seal migration"))
+		return errors.New("cannot auto-unseal during seal migration")
```
cipherboy: @satoqz Is this change mandatory or related to this PR? I think there's other cleanup that could be done if we are tweaking this now (`NewNonFatalError` is unused for instance, though `IsFatalError` is still used in a few places).

cipherboy (approved):

vault/init.go
satoqz: Not mandatory, I was looking at what the `IsFatalError` check in `runUnseal` actually did to see if I could hook into that, but then it turned out every single failure path returned a `NonFatalError` and there was no way to override this from deeper down the callstack, so I just removed it to lessen confusion.

Will remove from this patch.

PR #2910, opened 2026-04-17
Forbid path traversal by default

cipherboy:
## Description

While the Go standard library already calls `path.Clean(...)` on the input path, this is performed before URL decoding. After applying URL decoding, the resulting path is passed through `http/` to `vault.switchedLockHandleRequest(...)`; at this point it may again contain path traversal elements.

While we don't really care about non-traversing elements (such as path segments like `/./`), we do want to disallow them as they affect ACL policies. However, we need to be careful as patterns like `//` may appear in valid requests, such as OSCP-as-GET requests. Notably, cleaning the path on a per-request basis may also be unsafe as we may end up traversing between different secrets engines or handlers post-ACL. Lastly, placing it in `switchedLockHandleRequest` ensures we always protect against path traversal regardless of request source.

This also protects us from relative paths better in two places:

1. By fixing a longstanding bug in storage view, forbidding two or more dots from existing in the interior of a path segment (due to use of `strings.Contains(...)`); and
2. In the encrypted key view, used by KVv2, to limit `paths.Clean()` to only removing duplicated segments. This makes traversal by `List(...)` and `Scan(...)` possible.

Users who need to restore the original API behavior may set `unsafe_relative_paths=true` in their configuration file at their own risk.

## Rationale

Related: https://discuss.hashicorp.com/t/hcsec-2026-05-vault-kvv2-metadata-and-secret-deletion-policy-bypass-denial-of-service/77342

## Acknowledgements

       also allowed.

### DIFF
--- a/sdk/logical/path.go
+++ b/sdk/logical/path.go
@@ -20,30 +20,19 @@ func IsRelativePath(path string) bool {
 		return true
 	}
 
-	var index int
-	for index < len(path) {
-		switch path[index] {
-		case '/':
-			// Check for relative path portions at the end of the request path.
-			if (index+2 == len(path) && path[index:index+2] == "/.") ||
-				(index+3 == len(path) && path[index:index+3] == "/..") ||
-				(index+3 == len(path) && path[index:index+3] == "/./") ||
-				(index+4 == len(path) && path[index:index+4] == "/../") {
-				return true
-			}
+	// Check for relative path portions at the end of the request path.
+	if strings.HasSuffix(path, "/.") || strings.HasSuffix(path, "/..") {
+		return true
+	}
 
+	for index := range len(path) {
+		if path[index] == '/' {
 			// Check for relative path portions in the middle.
-			if (index+3 < len(path) && path[index:index+3] == "/./") ||
-				(index+4 < len(path) && path[index:index+4] == "/../") {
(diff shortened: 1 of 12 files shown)


wslabosz-reply (approved):

sdk/logical/path.go
```diff
@@ -0,0 +1,50 @@
...
+				return true
+			}
+
+			// At minimum we can skip one additional character due to lookahead
+			// we performed.
+			index += 1
+		}
+
+		index += 1
+	}
```
satoqz, line 36: I was able to simplify down to this not checking the path ending inside of the loop, but looks sound to me.

```suggestion
	// Special cases; the rest all have / as a prefix.
	if path == "." || path == ".." || strings.HasPrefix(path, "./") || strings.HasPrefix(path, "../") {
		return true
	}

	// Check for relative path portions at the end of the request path.
	if strings.HasSuffix(path, "/.") || strings.HasSuffix(path, "/..") {
		return true
	}

	for index := range len(path) {
		if path[index] == '/' {
			// Check for relative path portions in the middle.
			if (index+3 <= len(path) && path[index:index+3] == "/./") ||
				(index+4 <= len(path) && path[index:index+4] == "/../") {
				return true
			}
		}
	}
```

satoqz (approved):

PR #2912, opened 2026-04-17
Forbid generate-root by default

cipherboy:
## Description

This poses a potential availability risk like the rekey endpoints: no authentication is performed. With Control Groups under development, users can gate `auth/token/create` in a similar manner as generate-root, just without relying on recovery keys for this task. Likewise, users can always run a protected listener or Unix socket for local use of these endpoints.

## Rationale

Resolves: https://discuss.hashicorp.com/t/hcsec-2026-08-vault-vulnerable-to-denial-of-service-via-unauthenticated-root-token-generation-rekey-operations/77345

## Acknowledgements

       also allowed.


http/testing.go
```diff
@@ -53,7 +54,8 @@ func TestServerWithListener(tb testing.TB, ln net.Listener, addr string, core *v
 		Core: core,
 		// This is needed for testing custom response headers
 		ListenerConfig: &configutil.Listener{
-			Address: ip,
+			Address:                              ip,
+			DisableUnauthedGenerateRootEndpoints: pointerutil.BoolPtr(false),
```
satoqz, line 58: Can't wait for `new(false)` in Go 1.26.

website/content/docs/deprecation/unauthed-generate-root.mdx
satoqz, line 1: This needs an entry in the deprecation `index.mdx`.

website/content/api-docs/system/generate-root.mdx
```diff
@@ -4,6 +4,16 @@ description: |-
   OpenBao.
 ---
 
+:::warning
+
+Note: As of v2.5.3 usage of these endpoints is [deprecated](/docs/deprecation/unauthed-generate-root)
+and will be disabled by default. Considering using [`auth/token/create`](../../auth/token.mdx)
+to create new root tokens from existing tokens with `sudo` permission.
+To re-enable, see the documentation around the `disable_unauthed_generate_root_endpoints`
+[listener parameter](/docs/configuration/listener/unix/).
```
satoqz: One more reason to take care of shared listener docs.

website/content/api-docs/system/generate-root.mdx
```diff
@@ -4,6 +4,16 @@ description: |-
   OpenBao.
 ---
 
+:::warning
+
+Note: As of v2.5.3 usage of these endpoints is [deprecated](/docs/deprecation/unauthed-generate-root)
+and will be disabled by default. Considering using [`auth/token/create`](../../auth/token.mdx)
```
satoqz:
```suggestion
and will be disabled by default. Consider using [`auth/token/create`](../../auth/token.mdx)
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -126,6 +126,18 @@ default value in the `"/sys/config/ui"` [API endpoint](/api-docs/system/config-u
 
   :::
 
+- `disable_unauthed_generate_root_endpoints` `(bool: true)` - Whether to
+  disable requests to the legacy unauthenticated generate root endpoints
+  (under `/sys/generate-root/*`). These are a security risk to leave exposed
+  on public listeners.
```
satoqz: It might be better to explain what the actual security risk is, opposed to a blanket statement. Something along the lines of "this is a rarely-needed unauthenticated recovery endpoint that should be opted in to on a per-listener basis for a short time when needed. Enabling it by default opens up DoS attack surface to block or cancel operations preemptively up until when they are actually needed."

http/handler.go
```diff
@@ -174,10 +174,13 @@ func handler(props *vault.HandlerProperties) http.Handler {
...
-			handleAuditNonLogical(core, handleSysGenerateRootAttempt(core, vault.GenerateStandardRootTokenStrategy)))
-		mux.Handle("/v1/sys/generate-root/update",
-			handleAuditNonLogical(core, handleSysGenerateRootUpdate(core, vault.GenerateStandardRootTokenStrategy)))
+		// Register without unauthenticated generate root, if necessary.
+		if props.ListenerConfig != nil && props.ListenerConfig.DisableUnauthedGenerateRootEndpoints != nil && !*props.ListenerConfig.DisableUnauthedGenerateRootEndpoints {
+			mux.Handle("/v1/sys/generate-root/attempt",
+				handleAuditNonLogical(core, handleSysGenerateRootAttempt(core, vault.GenerateStandardRootTokenStrategy)))
+			mux.Handle("/v1/sys/generate-root/update",
+				handleAuditNonLogical(core, handleSysGenerateRootUpdate(core, vault.GenerateStandardRootTokenStrategy)))
+		}
```
satoqz, line 183: I wish this was reloadable, so you don't need to mess with nodes too much to enable the endpoint for a short time.

website/content/api-docs/system/generate-root.mdx
cipherboy: Let's do a clean cut in v2.6.0 :D

Otherwise we'd have to backport that and deal with synchronization.

satoqz: Oh yes sorry that was meant to be implied but not really implied.

http/handler.go
cipherboy: Making the handler reloadable would be a bit of a mess, I don't think ListenerConfig changes at all currently on SIGHUP. :thinking: 

- Unauthed rekey
- UI Enabled status
- Telemetry unauthed
- Unauthed pprof (risky)
- Inflight Req access

Are what this would affect. Something for the future, I think.

satoqz: Yeah, hence I wish 😭

website/content/docs/configuration/listener/tcp.mdx
cipherboy: I added to the `api-docs` the best practices as I think it is most relevant there, though explained the security risk by the parameter.

satoqz (approved):

cipherboy:
@satoqz Do you want to check this one again? Last commit changed to fix the interaction with the new namespace sealing on this branch.

cipherboy:
Edit: as discussed on Zulip, that comment was meant for the namespace/lease PR #2935. 

This PR is good to go, just hacking config blocks. Docs will be fixed by the backport sadly.

PR #2460, opened 2026-02-11
Add SealManager

wslabosz-reply:
## Description
- added `SealManager` struct definition and specific methods code, with tests added in `seal_manager_test.go` file (to be expanded)
- renamed `NamespaceView` -> `NamespaceScopedView` for clarity (`NamespaceScopedView` returns back a view on namespace using passed barrier, where new introduced `(*Core).NamespaceView` will return barrier used to read/write to the appropriate namespace path)
- renamed `NamespaceBarrierPrefix` -> `NamespaceStoragePathPrefix`
- introduced `CoreSealStatusResponse` to differ between seal status of the root namespace and other namespaces (cluster/server information stripped)

Seal manager code was introduced by @phyrog in [commit](https://github.com/openbao/openbao/commit/8e7f84fc708fc78406a08186f5f42fe40bdc879f), extensively iterated over next PRs with contributions from @phyrog (one of the [commits](https://github.com/openbao/openbao/commit/1969493c7ab10b7bf4f3f5d0d0293245254ea138)), @satoqz (one of the [commits](https://github.com/openbao/openbao/commit/d47e3bab403501d3e110d06d9186009f73c392db#diff-a255cf8f9c3c2ea590f75ccf6bc7294a9814f87ac7c876ca8715c86d7d418983)) and @KrzysztofKornalewski-Reply ([commit](https://github.com/openbao/openbao/commit/bd68282d6a46e6c6d8f012f21255385edae29f99))

All logic changes are sourced from [namespaces-seal feature branch](https://github.com/openbao/openbao/tree/namespaces-seal)

## Rationale
Parts of introduced code will be hooked up to existing systems within next PRs introducing proper support.

This change is non-breaking, as existing behavior is retained, with possible extension in next PRs. 

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #2916, opened 2026-04-18
fix(vault): don't leak share ordering on duplicate-share checks

### LINKED ISSUE #2663: Shamir key recreation shouldn't leak keyshare provision ordering
From discussion on https://github.com/openbao/openbao/pull/2460:

> @wslabosz-reply I don't think its really an issue, as it requires a priori knowing all Shamir shares. It allows an attacker (with a bag of shares), interrupting the unseal/rekey/... process, to de-anonymize who owns which share, if it can track order of registrations. 
> As you can see, the much bigger issue is that the attacker has valid shares in the first place. But it is also good hygiene I believe. 
> So I'm happy cleaning it up later in a separate PR.

_Originally posted by @cipherboy in https://github.com/openbao/openbao/pull/2460#discussion_r2954005947_

All checks for existence of a key share in vault package code (e.g. rotation, unseal) shouldn't leak the ordering of key shares provision. 

Example fix implemented with the original discussion.

raajheshkannaa:
## Summary
Closes #2663

When a user submits a key share during unseal, root token generation, root key rotation, recovery key rotation, or rotation verification, the duplicate-share check short-circuited as soon as the first match was found. An attacker observing response latency could infer the ordinal position at which the supplied share was previously registered.

## Change
Scan every stored share using \`crypto/subtle.ConstantTimeCompare\` and accumulate results via **bitwise OR into an integer** so every comparison executes and the total call count is independent of which slot (if any) matched. Branch on the accumulator after the loop.

## Why bitwise OR rather than boolean ||
\`seal_manager.go\` introduced this fix in #2460 with a boolean \`found = found || ...\` accumulator, but Go's \`||\` short-circuits, so later \`ConstantTimeCompare\` calls are skipped once the first match is found, which leaves a residual per-call-count side channel. Bitwise OR on \`int\` does not short-circuit, so every iteration runs. This PR applies the tighter pattern uniformly across all share-existence checks in \`vault/\`, including \`seal_manager.go\`, so the package has one consistent constant-time shape.

Sites fixed:
- \`vault/core.go\` \`recordUnsealPart\`
- \`vault/generate_root.go\` \`GenerateRootUpdate\` (also swaps \`bytes.Equal\` for \`subtle.ConstantTimeCompare\`)
- \`vault/rekey.go\` root rotation progress, recovery rotation progress, verification progress
- \`vault/rotate.go\` rotation progress, verification progress
- \`vault/seal_manager.go\` \`recordUnsealPart\` (tightens the existing #2460 pattern)

## Test plan
- [x] \`go build ./vault/\`
- [x] \`go vet ./vault/\`
- [x] \`gofmt -l\` on changed files
- [x] \`go test ./vault/ -run "TestSealManager|TestRecordUnseal"\`
- [ ] maintainers: any additional targeted test coverage you'd like added for these paths?

### DIFF
--- a/changelog/2916.txt
+++ b/changelog/2916.txt
@@ -1,0 +1,3 @@
+```release-note:security
+core: Prevent timing-based leak in Shamir key share duplicate checks during unseal, root token generation, key rotation, and rotation verification.
+```
--- a/vault/core.go
+++ b/vault/core.go
@@ -1601,12 +1601,17 @@ func (c *Core) unsealWithRaft(combinedKey []byte) error {
 
 // recordUnsealPart takes in a key fragment, and returns true if it's a new fragment.
 func (c *Core) recordUnsealPart(key []byte) (bool, error) {
-	// Check if we already have this piece
+	// Check if we already have this piece. Scan every stored share in
+	// constant time: the bitwise OR forces each ConstantTimeCompare call
+	// to execute, so the response time does not reveal which provision
+	// slot already holds the supplied share.
 	if c.unlockInfo != nil {
+		var found int
 		for _, existing := range c.unlockInfo.Parts {
-			if subtle.ConstantTimeCompare(existing, key) == 1 {
-				return false, nil
-			}
+			found |= subtle.ConstantTimeCompare(existing, key)
+		}
+		if found == 1 {
+			return false, nil
 		}
 	} else {
 		uuid, err := uuid.GenerateUUID()
--- a/vault/generate_root.go
+++ b/vault/generate_root.go
@@ -4,8 +4,8 @@
 package vault
 
 import (
-	"bytes"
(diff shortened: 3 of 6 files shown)


JanMa:
Hello @raajheshkannaa thank you for your PR. Please review our [contribution guidelines](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md) regarding the use of AI and use our official [PR template](https://github.com/openbao/openbao/blob/main/.github/pull_request_template.md) to acknowledge you adhere to them.

satoqz:
As mentioned previously, I don't see a reason to engage with PRs where the template was entirely removed. These never magically turn into acceptable contributions, so we can just close them down and skip the noise instead of waiting for the bot to come to the same realization.

PR #2933, opened 2026-04-20
Utilize ubuntu-slim runners wherever possible

### LINKED ISSUE #2167: Use ubuntu-slim runners for short jobs
From @karras on Zulip:

> I've recently stumbled on https://docs.github.com/en/actions/reference/runners/github-hosted-runners#single-cpu-runners. Would that help improving certain smaller jobs?

I think we have several jobs that qualify for these new runners:

- `setup`, `verify_changes`, `tests-completed` in `.github/workflows/ci.yml`
- `changelog-check` in `.github/workflows/changelog-checker.yml`
- `check-links` in `.github/workflows/docs-ci.yml`
- `go-dependency-submission` in `.github/workflows/go-dependency-submission.yml`
- `vulnerabilities`, `protobuf` in `.github/workflows/code-checker.yml`

https://github.com/fchimpan/gh-slimify may also come in handy.

Cost/benefit analysis is of course outstanding. But at first sight, this makes a lot of sense to me.

Do note that the feature is currently in public preview, we might want to wait until that is not the case anymore, but could certainly already test on another branch/repo.

wslabosz-reply:
## Description
This PR changes the (github) runner instance for following jobs:
- changelog-check
- tests-completed (CI)
- Go mod checks
- check-links and lint-docs
- DCO
- go-dependency-submission
- mirror
- scorecard
- Setup (release)
- Sync dependencies
- test-collect-reports
- verify commits

additionally removes (unused) stable-website.yaml and dummy-dco-merge-queue.yml actions.

## Rationale
Hopefully this would lead to (probably marginal) speedups in few of the running actions.

Resolves: #2167

## Acknowledgements

       also allowed.


cipherboy:
Protobuf stalled after 11min, retrying.

satoqz:
Significant slowdown on the protobuf part (4min vs 40sec), ref @ https://github.com/openbao/openbao/actions/runs/24684740351/job/72190989550

satoqz:
Go mod goes from 6s to 20s, seems acceptable.

wslabosz-reply:
@satoqz @cipherboy 
Do we want to pursue this change (maybe revert protobuf, as it incurred the highest slowdown), not sure we actually care about the "cost-reduction" of the ubuntu-slim?

satoqz:
@wslabosz-reply the supposed benefit is scheduling time. So any job that takes very little time to _run_ (Mostly set-up jobs, as long as they don't pull in Go in any way?) should benefit.

.github/workflows/dummy-dco-merge-queue.yml
satoqz, line 1: I think we can remove this file, too.

.github/workflows/docs-deploy.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
```
satoqz:
```suggestion
    runs-on: ubuntu-latest
```

This one definitely needs more than 1 CPU.

.github/workflows/docs-ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           fi
   test-docs:
     name: Test Docs
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
```
satoqz:
```suggestion
    runs-on: ubuntu-latest
```

This one, too.

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     name: Setup
     env:
       BASE_SHA: ${{ github.event.pull_request.head.sha }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
```
satoqz: I'm not sure about the slowdown on this one since it does interact with Go. Would be good to diff times, OTOH this one would greatly benefit from better scheduling since it blocks so many other jobs.

wslabosz-reply: tested it here https://github.com/Ki-Reply-GmbH/openbao/actions/runs/24675123740/job/72156755248
increase to 26s from typical 15-20s

.github/workflows/dummy-dco-merge-queue.yml
wslabosz-reply: @cipherboy wdyt?

.github/workflows/ci.yml
satoqz: Let's skip this one then tbh

.github/workflows/dummy-dco-merge-queue.yml
cipherboy: Yes I think this was leftover from Nathan's experiments with merge queues.

.github/workflows/docs-ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
 
   lint-docs:
     name: Lint Docs
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
```
satoqz, line 55: This one does text processing with JavaScript so I'm not sure; did you check times?

wslabosz-reply: (slim)
image

compared to latest:
image

satoqz: That's... interesting. So that suggests we stay on ubuntu-latest for "Lint Docs", but "Test Docs" and likely "Deploy Docs" don't have the slowdown I expected so we can use slim 🤔

cipherboy (approved):
As long as there's general availability, I'm happy with this +/- whatever nitpicking of what goes where we want to do. 😆 

(`lint-docs` is so short I don't think it matters too much to me.)

phyrog:
> @wslabosz-reply the supposed benefit is scheduling time. So any job that takes very little time to _run_ (Mostly set-up jobs, as long as they don't pull in Go in any way?) should benefit.

Do we have numbers on the scheduling time? I'd expect `ubuntu` to be much more widely used than `ubuntu-slim` and might even benefit from caching on the runners? But I'm just guessing.

Since this is just for the short running jobs I don't care either way, though. I'd really like to see some runtime improvement for the tests 😄

satoqz:
@phyrog well, one is scheduled as a full VM and the other is a container with lower resource requests.

PR #2934, opened 2026-04-20
Prevent cross-namespace token accessor use

cipherboy:
## Description

When revoking, renewing, or looking up a token by accessor, ensure that the request's namespace matches that of the accessor. This differs from token value as there's a legitimate use case for cross-namespace token value lookup: requests could go to a child namespace (granted by policy) with a token from a parent namespace. However, token accessors are only used by certain renewal, revocation, or lookup paths and thus need not be done cross-namespace.

## Rationale

Resolves: GHSA-p49j-v9wc-wg57

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2934.txt
+++ b/changelog/2934.txt
@@ -1,0 +1,3 @@
+```release-note:security
+auth/token: Prevent cross-namespace token renewal, revocation by accessor. GHSA-p49j-v9wc-wg57 / CVE-2026-40264.
+```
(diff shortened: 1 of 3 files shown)


wslabosz-reply (approved):

cipherboy:
@wslabosz-reply Mind re-approving? I pushed Jonas's feedback.

satoqz (approved):

PR #2935, opened 2026-04-20
Ensure lease revocation on namespace re-deletion

cipherboy:
## Description

When a namespace's deletion was interrupted, we first need to reload any mounts it has: mounts need to be active before leases are removed, otherwise storage is simply removed from under the namespace. This way, lease revocations are correctly processed on retried deletion.

This does not affect deletion which succeeded on the first try; it only affects deletion of already tainted namespaces.

## Rationale

Resolves: GHSA-vv66-6rp4-wr4f

## Acknowledgements

 - [ ] By contributing this change, I certify I have signed-off on the
       also allowed.

### DIFF
--- a/changelog/2935.txt
+++ b/changelog/2935.txt
@@ -1,0 +1,3 @@
+```release-note:security
+core/namespaces: Ensure lease revocation on namespace re-deletion. GHSA-vv66-6rp4-wr4f.
+```
--- a/vault/external_tests/namespaces/namespace_test.go
+++ b/vault/external_tests/namespaces/namespace_test.go
@@ -1,0 +1,179 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package namespaces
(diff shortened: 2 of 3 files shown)


satoqz (approved):

cipherboy:
@satoqz This now calls `t.Skip` on the load deletion test in CI.

I'm not really sure there's alternatives here, there's a fundamental race between namespace deletion completing and sealing prematurely which we need to have to trigger a namespace re-deletion. We could maybe use a non-cluster test and poke the internals to explicitly cancel the context associated with the namespace deletion helper manually, prior to accepting the request (without actually cancelling the active context so the rest of request handling would work), but not sure I really like that either. Just adding a cancel deletion API wouldn't help either, just changes what we're racing against. 

Thoughts?

satoqz:
I'm fine skipping, perhaps there are tweaks we can come up with to test this better but that's going to be out of scope for today.

satoqz:
Adjustments for namespace sealing look good to me.

PR #2941, opened 2026-04-20
Validate downloaded plugin binary size

cipherboy:
To protect against a malicious plugin OCI image. This patch introcudes a new config setting `plugin_download_max_size` that controlls the maximum allowed plugin binary size (Default 512 MiB). We now check the reported plugin size in the Tar Header of the extracted OCI image against that limit and fail if it is exceeded. For good measure we also now check if there's enough disk space left to extract the plugin before we try to do so and if the amount of bytes we copied matched what was reported in the Tar Header.

## Rationale

Resolves: #

## Acknowledgements

       also allowed.


JanMa (approved):

satoqz (approved):

cipherboy:
Thank you @JanMa for patch development!

PR #2952, opened 2026-04-21
Add RFC on TTL-based Cache Invalidation

cipherboy:
## Description

This proposes a new TTL-based cache invalidation mechanism for the PostgreSQL storage backend. Rather than requiring WAL updates to be processed on standby nodes to trigger the invalidation hook, we track all read operations and queue them future invalidation. On the whole, this behaves correctly (as no data is proactively cached without going through this layer), at the expense of higher background storage requests. With some tuning, we can potentially limit the impact of that.

## Rationale

This is the next step for read scalability on PostgreSQL, mirroring what we shipped for Raft in v2.5.0. Given the size of the change set is minimal, I'd like to see if we can land the change for v2.6.0 if possible.

(cc: @phil9909 @fatima2003 @driif @satoqz @wslabosz-reply @JanMa @suprjinx) 

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/rfcs/index.mdx
+++ b/website/content/docs/rfcs/index.mdx
@@ -27,6 +27,9 @@ Steering Committee.
    necessary constructs to build indexes on data efficiently allowing for faster search operations.
  - [Emergency Seal](emergency-seal.mdx), for adding a secondary Shamir's
    seal as a "break-glass" recovery path for auto-seal failures.
+ - [TTL-Based Cache Invalidation](ttl-invalidation.mdx), for providing read
+   scalability through eventual consistency semantics on non-WAL HA storage
+   backends like PostgreSQL.
 
 ## Landed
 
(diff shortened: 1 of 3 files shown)


cipherboy:
Alas, as @satoqz pointed out to me via DMs, we'd have to invalidate List responses too (so that we find new/deleted mounts+namespaces), which effectively means caching these, performing the underlying list, and potentially sending invalidations for every entry. You could limit this to just new/removed, though now you have to cache the full list result info. 

That either means more memory or higher churn on the system... Either way, with 3.5k namespaces, just re-reading and invalidating all mounts on a 5s basis isn't quite palatable. 

I think this is a decent opt-in way to get basic horizontal scalability on systems with low performance requirements, but we'd still need the GRPC route for high-performance systems. 

---

The proof of concept was updated, though.

PR #2953, opened 2026-04-21
Correctly handle default_rate_limit_exempt_paths_toggle invalidation

cipherboy:
## Description

    This now correctly handles invalidation of the rate limit exempt paths
    detection entry by checking our state on configuration writes and fully
    reloading the configuration on writes to this storage entry. Previously
    we'd ignore changes to this path which would cause a reload of the
    entire cluster which is excessive in this case.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/quotas/quotas.go
+++ b/vault/quotas/quotas.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"path"
 	"strings"
+	"sync/atomic"
 
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-memdb"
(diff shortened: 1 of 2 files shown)


vault/quotas/quotas.go
```diff
@@ -944,15 +947,26 @@ func dbSchema() *memdb.DBSchema {
...
 	case "config":
 		config, err := LoadConfig(m.ctx, m.storage)
 		if err != nil {
 			return fmt.Errorf("failed to invalidate quota config: %w", err)
 		}
 
+		exemptPaths := defaultExemptPaths
+		if m.usingConfigExemptPaths.Load() {
+			exemptPaths = config.RateLimitExemptPaths
+		}
```
cipherboy, line 965: Strictly I don't think you can invalidate either `DefaultRateLimitExemptPathsToggleName` or `config` without ending up with `usingConfigExemptPaths=true`, matching the behavior we already had, but this makes it more explicit. I guess you could if you manually poked `sys/raw`, but I don't think it warrants a changelog entry as a result.

wslabosz-reply (approved):
LGTM

PR #2955, opened 2026-04-22
Add Sys() API client helpers for namespaces

EnricoFusi:
## Description

Add "CreateNamespace", "ReadNamespace", "ListNamespaces", "PatchNamespace"
and "DeleteNamespace" helpers to the "Sys()" API client, with table-driven
unit tests for each operation.

Refactor the matching CLI commands ("namespace create", "delete", "list",
"lookup", "patch") to use the new client methods instead of raw logical
client calls.

Fix two server-side status code bugs in the PATCH endpoint:
- PATCH on a nonexistent namespace returned 500; now returns 404.
- PATCH with an empty or unrecognized body returned 500; now returns 400.

## Rationale

Resolves: #2821

## Acknowledgements

       also allowed.

### DIFF
--- a/api/sys_namespaces.go
+++ b/api/sys_namespaces.go
@@ -5,10 +5,11 @@ package api
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
+
+	"github.com/go-viper/mapstructure/v2"
 )
 
 // CreateNamespaceInput is the input for the CreateNamespace operation.
@@ -60,19 +61,13 @@ type PatchNamespaceResponse struct {
 	KeyShares      []string          `json:"key_shares"`
 }
 
-// ListNamespacesResponse is the response from the ListNamespaces operation.
-type ListNamespacesResponse struct {
-	Keys    []string                         `json:"keys"`
-	KeyInfo map[string]ReadNamespaceResponse `json:"key_info"`
-}
-
 // ListNamespaces lists all child namespaces relative to the current namespace.
-func (c *Sys) ListNamespaces() (*ListNamespacesResponse, error) {
+func (c *Sys) ListNamespaces() (map[string]ReadNamespaceResponse, error) {
 	return c.ListNamespacesWithContext(context.Background())
 }
 
 // ListNamespacesWithContext lists all child namespaces relative to the current namespace.
-func (c *Sys) ListNamespacesWithContext(ctx context.Context) (*ListNamespacesResponse, error) {
+func (c *Sys) ListNamespacesWithContext(ctx context.Context) (map[string]ReadNamespaceResponse, error) {
 	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
 	defer cancelFunc()
 
@@ -93,26 +88,34 @@ func (c *Sys) ListNamespacesWithContext(ctx context.Context) (*ListNamespacesRes
 		return nil, errors.New("data from server response is empty")
 	}
 
-	dataJSON, err := json.Marshal(secret.Data)
+	keyInfoRaw, ok := secret.Data["key_info"]
+	if !ok {
+		return map[string]ReadNamespaceResponse{}, nil
+	}
+
+	result := map[string]ReadNamespaceResponse{}
+	decoder, err := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
+		TagName: "json",
+		Result:  &result,
+	})
 	if err != nil {
 		return nil, err
 	}
-	result := &ListNamespacesResponse{}
-	if err := json.Unmarshal(dataJSON, result); err != nil {
+	if err := decoder.Decode(keyInfoRaw); err != nil {
 		return nil, err
 	}
 	return result, nil
 }
 
-// CreateNamespace creates a new namespace at the given path.
-func (c *Sys) CreateNamespace(path string, i *CreateNamespaceInput) (*CreateNamespaceResponse, error) {
-	return c.CreateNamespaceWithContext(context.Background(), path, i)
+// CreateNamespace creates a new namespace with the given name.
+func (c *Sys) CreateNamespace(name string, i *CreateNamespaceInput) (*CreateNamespaceResponse, error) {
+	return c.CreateNamespaceWithContext(context.Background(), name, i)
 }
 
-// CreateNamespaceWithContext creates a new namespace at the given path.
-func (c *Sys) CreateNamespaceWithContext(ctx context.Context, path string, i *CreateNamespaceInput) (*CreateNamespaceResponse, error) {
-	if path == "" {
-		return nil, errors.New("path must not be empty")
+// CreateNamespaceWithContext creates a new namespace with the given name.
+func (c *Sys) CreateNamespaceWithContext(ctx context.Context, name string, i *CreateNamespaceInput) (*CreateNamespaceResponse, error) {
+	if name == "" {
+		return nil, errors.New("name must not be empty")
 	}
 	if i == nil {
 		i = &CreateNamespaceInput{}
@@ -121,7 +124,7 @@ func (c *Sys) CreateNamespaceWithContext(ctx context.Context, path string, i *Cr
 	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
 	defer cancelFunc()
 
-	r := c.c.NewRequest(http.MethodPost, fmt.Sprintf("/v1/sys/namespaces/%s", path))
+	r := c.c.NewRequest(http.MethodPost, fmt.Sprintf("/v1/sys/namespaces/%s", name))
 	if err := r.SetJSONBody(i); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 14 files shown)


api/sys_namespaces.go
```diff
@@ -0,0 +1,247 @@
...
+	UUID           string            `json:"uuid"`
+	ID             string            `json:"id"`
+	Path           string            `json:"path"`
+	Tainted        bool              `json:"tainted"`
+	Locked         bool              `json:"locked"`
+	CustomMetadata map[string]string `json:"custom_metadata"`
+	KeyShares      []string          `json:"key_shares"`
+}
+
+// ReadNamespaceResponse is the response from the ReadNamespace operation.
```
wslabosz-reply: `ReadNamespaceResponse` is the same struct as `CreateNamespaceResponse`

api/sys_namespaces.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// PatchNamespaceInput is the input for the PatchNamespace operation.
+// CustomMetadata values can be nil to remove a key.
+type PatchNamespaceInput struct {
+	CustomMetadata map[string]interface{} `json:"custom_metadata"`
+}
+
+// PatchNamespaceResponse is the response from the PatchNamespace operation.
+type PatchNamespaceResponse struct {
```
wslabosz-reply: also same as other two structs

changelog/2955.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+namespaces: add CreateNamespace, ReadNamespace, ListNamespaces, PatchNamespace and DeleteNamespace to the Sys() API client; fix PATCH returning 500 on missing or nonexistent namespace
+```
```
satoqz, line 6:
````suggestion
```release-note:improvement
api: Add first-class support for `/sys/namespaces` APIs via `.Sys().CreateNamespace(...)` & co.
```
```release-note:bug
core/namespaces: Fix PATCH on a namespace returning status 500 on missing or nonexistent namespace.
```
````

File: api/sys_namespaces.go [hunk 4]
```diff
@@ -0,0 +1,247 @@
...
+	if secret == nil || secret.Data == nil {
+		return nil, errors.New("data from server response is empty")
+	}
+
+	dataJSON, err := json.Marshal(secret.Data)
+	if err != nil {
+		return nil, err
+	}
+	result := &ListNamespacesResponse{}
+	if err := json.Unmarshal(dataJSON, result); err != nil {
```
phyrog: Why are we marshalling and unmarshalling here? Could we use the `mapstructure` module here instead?

api/sys_namespaces.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, err
+	}
+	return result, nil
+}
+
+// CreateNamespace creates a new namespace at the given path.
+func (c *Sys) CreateNamespace(path string, i *CreateNamespaceInput) (*CreateNamespaceResponse, error) {
```
satoqz:
```suggestion
func (c *Sys) CreateNamespace(name string, i *CreateNamespaceInput) (*CreateNamespaceResponse, error) {
```

Not all existing code reflects this, but we eventually started referring to "name" opposed to "path" when talking about the namespace API since it only lets you pass a single path component to the API. The "path" naming suggests otherwise so I'd avoid it.

wslabosz-reply: @satoqz 
not opposing the proposal, but I think we should do it in a separate PR, as currently all logical handlers refer to it as `path` still (I've even retained that naming in open PRs for per-ns sealing for now).

satoqz: Well I'd propose to do it in the net-new API code, especially since that is user-facing. We can do the backend code another time.

api/sys_namespaces.go
cipherboy: Further, isn't a `ListNamespacesResponse` either just a `map[string]ReadNamespaceResponse` or a `[]ReadNamespaceResponse`?

api/sys_namespaces.go
wslabosz-reply: fine by that, if we just create an issue out of this comment to track the discrepancy (to fix) of the naming.

api/sys_namespaces.go
phyrog: Not quite, we're using `logical.ListResponseWithInfo` on the server, so we'll get a `keys` list (for order I assume) and a `key_info` map.

cipherboy: @phyrog Right, but see `ListMountsWithContext(...)` which returns `map[string]*MountOutput`. In spirit this should be the same thing (order isn't meaningful here). Or, if we really want to preserve order, `[]ReadNamespaceResponse` is the equivalent form, just merging `keys` and `keyinfo`. 

I think we should go to the maximal useful form here; the additional nesting doesn't really add anything.

api/sys_namespaces.go
EnricoFusi: Kept intentionally separate: sharing a single struct would mean that any future divergence between the operation responses forces a breaking API change. The duplication is minimal and the isolation is deliberate. If you think they will never diverge, I can unify them.

api/sys_namespaces.go
EnricoFusi: Kept intentionally separate: sharing a single struct would mean that any future divergence between the operation responses forces a breaking API change. The duplication is minimal and the isolation is deliberate. If you think they will never diverge, I can unify them.

api/sys_namespaces.go
wslabosz-reply: okay, that makes sense, although in that case you can just alias everything to one definition:
```
type ReadNamespaceResponse = CreateNamespaceResponse
```

api/sys_namespaces.go
EnricoFusi: I changed  path with name  all along this pr.

api/sys_namespaces.go
EnricoFusi: Adapted the code to this pattern.

api/sys_namespaces.go
EnricoFusi: Ok, changed with the newest commit.

api/sys_namespaces.go
phyrog: @cipherboy Yeah you're right, I was also just looking at what the API returns 🙂

command/namespace_create.go
```diff
@@ -96,20 +98,36 @@ func (c *NamespaceCreateCommand) Run(args []string) int {
...
+func structToMap(v interface{}) (map[string]interface{}, error) {
+	b, err := json.Marshal(v)
+	if err != nil {
+		return nil, err
+	}
+	var m map[string]interface{}
+	if err := json.Unmarshal(b, &m); err != nil {
+		return nil, err
+	}
+	return m, nil
```
phyrog: This should be replaceable with `mapstructure` as well.

api/sys_namespaces.go
```diff
@@ -0,0 +1,234 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/go-viper/mapstructure/v2"
+)
+
+// CreateNamespaceInput is the input for the CreateNamespace operation.
+type CreateNamespaceInput struct {
+	CustomMetadata map[string]string `json:"custom_metadata"`
+}
```
phyrog, line 86: Note, this is currently missing the `Seal` field, but as I am changing the type of that in #2900 anyway, I'd suggest to add the the field in whatever PR gets merged later.

command/namespace_create.go
satoqz: We already have a `helper/structtomap` package for this use case, but I also don't entirely understand why automated conversion is needed here at all.

EnricoFusi: Done in the last commit.

api/sys_namespaces.go
EnricoFusi: As discussed in chat, I am not touching this field, until #2900 is there.

command/namespace_create.go
satoqz: Not done, see my comment.

api/sys_namespaces.go
satoqz: Doesn't `ReadNamespaceResponse` not have `KeyShares`? IMO worth separating explicitly as @EnricoFusi initially did.

api/sys_namespaces.go
```diff
@@ -0,0 +1,234 @@
...
+// ReadNamespaceResponse is the response from the ReadNamespace operation.
+type ReadNamespaceResponse = CreateNamespaceResponse
+
+// DeleteNamespaceResponse is the response from the DeleteNamespace operation.
+type DeleteNamespaceResponse struct {
+	Status string `json:"status"`
+}
+
+// PatchNamespaceInput is the input for the PatchNamespace operation.
+// CustomMetadata values can be nil to remove a key.
```
satoqz:
```suggestion
// CustomMetadata values can be string to add or modify a key, or nil to remove
// a key.
```

api/sys_namespaces.go
```diff
@@ -0,0 +1,234 @@
...
+type PatchNamespaceInput struct {
+	CustomMetadata map[string]interface{} `json:"custom_metadata"`
+}
+
+// PatchNamespaceResponse is the response from the PatchNamespace operation.
+type PatchNamespaceResponse = CreateNamespaceResponse
```
satoqz: Similarly, I don't believe PATCH returns key shares. So this would be equal to `ReadNamespaceResponse` but not `CreateNamespaceResponse`.

satoqz:
I think something like

```go
// Returned by List, Patch and GetNamespace.
type NamespaceOutput struct {
  // ...
}

// Returned only by CreateNamespace.
type CreateNamespaceOutput struct {
  NamespaceOutput
  KeyShares []string
}
```

could work and be more in line with the `Input`/`Output` naming that the mounts API client already has.

I think we can reasonably expect get/list/patch to always return the same schema, while create will be a superset.

api/sys_namespaces.go
wslabosz-reply: @satoqz 
you are right, I've focused on that all `<Create|Read|Patch>NamespaceResponse` were the same, and haven't noticed all returning `KeyShares`.

EnricoFusi: No problem to go back to the "split" version. I did the split thinking of versioning of API, but I see there could be more reasons fot this. I will align privately.

command/namespace_create.go
EnricoFusi: Argh, the fisrt time i applied the wrong structtomap. Should be fixed now.

api/sys_namespaces.go
EnricoFusi: I apologize for not being clear. My question is: are you asking me to go back to the previous implementation, when each API had its own data structure? If this is the case, I can do it in short times.

wslabosz-reply: I'm fine with current state, but won't say for @satoqz

wslabosz-reply (approved):
LGTM

api/sys_namespaces.go
cipherboy: @EnricoFusi Want to add `Seal` now? #2900 has landed :-)

cipherboy (approved):
Thanks @EnricoFusi! If you don't mind adding seal configuration and resolving conflicts, that'd be great. :-)

api/sys_namespaces.go
EnricoFusi: Hi @cipherboy , I am going to work to it.

EnricoFusi:
Resolved existing conflicts, and seal param with config file added, like @cipherboy suggested.

satoqz (approved):
Thanks!

PR #2972, opened 2026-04-23
Add CLI support for unsealing and sealing namespaces

phyrog:
## Description

Adds the `bao namespace unseal` and `bao namespace seal` operations to the CLI.

## Rationale

Part of #1357

## Acknowledgements

       also allowed.

### DIFF
--- a/api/sys_namespaces.go
+++ b/api/sys_namespaces.go
@@ -51,3 +51,20 @@ func (s *Sys) UnsealNamespaceWithContext(ctx context.Context, req *UnsealNamespa
 	err = resp.DecodeJSON(&result)
 	return result.Data, err
 }
+
+func (c *Sys) SealNamespace(name string) error {
+	return c.SealNamespaceWithContext(context.Background(), name)
+}
+
+func (c *Sys) SealNamespaceWithContext(ctx context.Context, name string) error {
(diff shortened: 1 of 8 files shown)


api/sys_namespaces.go
```diff
@@ -0,0 +1,70 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+type UnsealNamespaceInput struct {
+	Path  string `json:"path"`
```
cipherboy, line 10: I wonder if we want some helpers of the form like KV:

```
func (c *Client) Namespaces(path string) *Namespaces {
    return &Namespaces{c: c, path: path}
}
```

Then it can be `ns.ResetUnsealProgress()`, &c, mirroring what's in `api/sys_seal.go`? This would theoretically let one (a user of the API, not necessarily us) create an interface over a generic sealable thing, as long as we ensure we have a consistent subset of the API. 

Thoughts?

cipherboy: Though, I suppose that ideal of a common interface breaks down with the `SealStatusResponse` struct, since we'll have a different namespaced version with fewer fields. 🤔

cipherboy (approved):
I'm happy taking this now and adjusting the namespace-specific behavior sometime before release if we want to build a client type there.

PR #2911, opened 2026-04-17
Added RFC for lightweight OCSP Support

boumba100:
## Description

This pull request includes an RFC for adding lightweight OCSP support to Openbao.

## Rationale

View original description at https://github.com/openbao/openbao/issues/2890

Resolves: #2890

## Acknowledgements

 - [ x ] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [ x ] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #2977, opened 2026-04-24
feat: Public routes enabling Lightweight OCSP Support (#2890)

### LINKED ISSUE #2890: Lightweight OCSP Support

Many network devices, such as switches and routers, use the lightweight OCSP profile ([RFC 5019](https://datatracker.ietf.org/doc/html/rfc5019)) for certificate revocation checks. Although the OpenBao PKI engine supports most of this profile, it is currently unable to handle unencrypted (non-TLS) HTTP requests, which is a requirement defined in [RFC 5019, Section 5](https://datatracker.ietf.org/doc/html/rfc5019).

To fully support RFC 5019 OCSP clients, the OpenBao secrets engine must be able to handle select non-TLS HTTP requests. This currently requires configuring a full unencrypted TCP listener, which would expose all other paths on that listener.

For lightweight OCSP support, add the ability for the /ocsp PKI engine path to listen on a non-tls secured encrypted listener to handle plaintext http requests.

This can be achieved with the following:

1) Add a public_routes option to listener option

This configures a listener to be able to handle public routes. Only requests that are for paths that are marked as public paths will be accepted

The following is an example of a listener configuration that is configured to handle public routes

```hcl

listener "tcp" {
 address = "0.0.0.0:8280"
 tls_disable = true

 public_routes = true
}

```

2)  Add a ocsp_rf5019_mode to the pki engine crl configuration

This configuration exposes the OCSP path of the target PKI engine as a public route. This allows the OCSP path to handle requests originating from a non-TLS secured listener configured for public routes.

This can be configured with a command similar to the following:

```
 vault write pki/config/crl ocsp_rfc5019_mode=true
```

Alternatively, hosting a plaintext HTTP proxy in front of an Openbao instance is an option. However, this increases operational complexity as it necessitates configuring and managing an additional service.

Support certificate lifecycle management for a broad range of network and IOT devices that only support lightweight OCSP checks.

boumba100:
## Description

Added a public routes feature, as defined in #2911 (work-in-progress), enabling Openbao's pki engine to handle and serve lightweight OCSP requests over HTTP (non-TLS).

With this, there is now the option to designate a listener as a public route listener by marking `public_routes = true` in its listener configuration. When designated as a public route listener, it will only process requests to paths that are marked as allowed public paths.

Allowed public paths are configured for a mount point, such as a secret engine or an auth engine, by configuring the `allowed_public_paths property` in the mount point's tune configuration.

## Rationale

See explanation in #2890 and #2911 

Resolves: #2890 

## Acknowledgements

 - [  x ] By contributing this change, I certify I have not used generative AI
 - [  x ] By contributing this change, I certify I have signed-off on the
       also allowed.


boumba100:
Note that the feature RFC (https://github.com/openbao/openbao/pull/2911) has not yet been approved or merged, so the design and approach for this feature may change.

boumba100:
Changes implemented based on [Lightweight OCSP Support RFC](https://github.com/openbao/openbao/pull/2911)

boumba100:
Hi @cipherboy, I'm tagging you on this because we collaborated on the associated RFC :)

Issue #2957, opened 2026-04-22
raft snapshots waste time on compression


phil9909:
## Snapshot Format

Quick reminder on the file format the raft snapshots are currently using:

GZIP compressed tar file with the following content:
  - `meta.json`: Snapshot format version, Raft Index and Term, etc in JSON format
  - `state.bin` custom, protobuf based file format containing Key - Value pairs
  - `SHA256SUMS` sha256 sums of the two previous files
  - `SHA256SUMS.sealed` signature for the previous file
  
For performance considerations we can focus on the `state.bin`, because all other files will be tiny compared to this file even for small bao instances.

## Problem Statement

We noticed that taking a raft snapshot (e.g. via `bao operator raft snapshot save snap.tar.gz`) is quite slow and pprof tells us a lot of the time is spent on compression:

<img width="923" height="1147" alt="Image" src="https://github.com/user-attachments/assets/e7a67201-cd9e-4d1f-bb0d-0203976afc3f" />

What makes this worse: The compression ratio is negligible (because large parts of our payload are encrypted, good crypto is indistinguishable from random and random does not compress well<sup>1</sup>).

For creating the snapshots, we leverage `github.com/hashicorp/raft-snapshot`, which hard-codes the gzip compression with default compression level: https://github.com/hashicorp/raft-snapshot/blob/6344e43a19659cbccc83423ba39ceacda83eb34c/snapshot.go#L140-L141

I tried other gzip compression levels and another compression algorithm (Zstandard), because I though at least the keys (which are sorted in the `state.bin` and often consecutive keys share a prefix like `/logical/<uuid>`) should compress well, but it did not help (zstd reported a size ratio of `99.82%` even on compression level 19) 

To summarize: Raft snapshots use (IMHO waste) a lot of time compressing the snapshot for negligible gain.

## Possible Solutions

To fix this, we have to change `github.com/hashicorp/raft-snapshot`, this could either be done by:

- contributing the changes upstream
- creating a fork
- inlining the code into the OpenBao repo

Usually I'd go for contributing first, but given this is a HashiCorp repo, I'm not sure if they will reject the contribution anyway.

Implementation wise, I see two options:
- We could change to [`gzip.NewWriterLevel`](https://pkg.go.dev/compress/gzip#NewWriterLevel) and pass `gzip.NoCompression` as the level
- Remove the gzip compression completely and adapt the restore logic to detect if gzip decompression is needed

The former has the advantages:
[...]

PR #2990, opened 2026-04-28
Fork `github.com/hashicorp/raft-snapshot`

phil9909:
## Description

This PR will drop the dependency on [`github.com/hashicorp/raft-snapshot`](https://github.com/hashicorp/raft-snapshot), by copying the code over to our repo.

I did not use the latest (`1.0.4`) tag, but the `master` branch (all commit after ` copy up to 8kb for non snapshot.bin raft snapshot copies (#10)` are new, they don't include any new features, just some linter fixes like `ioutil` deprecation and missing error checks)

This PR was created with `git-filter-repo`, retaining attribution for the original authors:

```bash
git clone git@github.com:hashicorp/raft-snapshot.git

cd raft-snapshot

git-filter-repo \
    --path-glob '*.go' \
    --path testdata/ \
    --to-subdirectory-filter physical/raft/snapshot/ \
    --message-callback 'return re.sub(b"#([0-9]+)", b"https://github.com/hashicorp/raft-snapshot/pull/\\1", message)' \
    --prune-empty=always --prune-degenerate=always

git branch -m fork-raft-snapshot

git branch | grep -v fork-raft-snapshot | xargs git branch -d
git tag | xargs git tag -d

git remote add --fetch origin git@github.com:openbao/openbao.git
git merge --allow-unrelated-histories origin/main

# Now signoff the commits, which - to my best knowledge - I'm allowed to do based on DCO 1.1 section (b)
git rebase --signoff origin/main
```

Afterwards I added two commits:
- Fix linter errors in the code
- Actually use the new package

Follow up PRs will:
- Drop anything we don't call / use
- Change the gzip compression level

To keep the attribution we should not squash merge this PR.

I did sign-off all the commits, which based on my understanding of DCO 1.1 section (b) should be fine.

## Rationale

Preparation for fixing #2957 

## Acknowledgements

       also allowed.


phil9909:
I re-ran the `git-filter-repo` command with an additional flag: `--message-callback 'return re.sub(b"#([0-9]+)", b"https://github.com/hashicorp/raft-snapshot/pull/\\1", message)'` to fix GitHubs autolinking

pree:
Are we using all the features of this `raft-snapshot` repo, or can we slim down `snapshot.go` and remove unused functions?

phil9909:
> Are we using all the features of this `raft-snapshot` repo, or can we slim down `snapshot.go` and remove unused functions?

Already planned for a follow up PR, see description. I wanted to keep this PR to the absolute minimum, so that the refactorings can be reviewed in isolation

phyrog:
Any way we can generate the test fixtures on the fly in the (near) future? Thinking about the xz attack path.

pree:
> Any way we can generate the test fixtures on the fly in the (near) future? Thinking about the xz attack path

I think this would be a good idea, as we also don't care for compression ratio probably we can just take some random strings. But this should be a separate PR imho

JanMa (approved):
LGTM, thanks @phil9909

cipherboy (approved):
Thanks @phil9909!

PR #3007, opened 2026-04-29
Make `SystemBackend` a singleton

### LINKED ISSUE #1809: SystemBackend could be per-namespace

wslabosz-reply:
## Description
This PR makes `SystemBackend` a singleton rather than creating one instance per created namespace.

## Rationale
by account of @satoqz report
> the memory usage of an idle bao server -dev with 1000 empty namespaces drops from 1.3 GB to 900 MB on my system.

this means it's a substantial memory optimization without any fallbacks.

Note: #1809 was proposing the reverse, so making `SystemBackend` per namespace rather than singleton (which was the reverse of the current state), this was motivated by future enhancements we could implement:
- simplifying locks over mounts, policy store
- making expiration manager per namespace
and possibly others, but this would require a more through refactor which a plan for should be formulated beforehand, hence we opt for low-hanging fruit here.

patch used in the PR prepared fully by @satoqz

Resolves: #1809

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3007.txt
+++ b/changelog/3007.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+core: `SystemBackend` is now a singleton shared across all namespaces, reducing idle memory usage of the OpenBao instance.
+```
--- a/vault/core.go
+++ b/vault/core.go
@@ -1247,7 +1247,12 @@ func (c *Core) configureLogicalBackends(backends map[string]logical.Factory, log
 		b := NewSystemBackend(c, sysBackendLogger)
 		return b, b.Setup(ctx, config)
 	}
-	logicalBackends[routing.MountTypeNSSystem] = logicalBackends[routing.MountTypeSystem]
+	logicalBackends[routing.MountTypeNSSystem] = func(ctx context.Context, bc *logical.BackendConfig) (logical.Backend, error) {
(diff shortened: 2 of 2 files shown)


cipherboy (approved):
I'm fine with this. 

As I discussed with @wslabosz-reply this morning, I was thinking about moving things like Policy store, token store, mount tables, and others inside system backend so that we can remove some global locks in the Singleton core object.

This isn't really easily/immediately feasible, however, and we've taken to using locks manager when possible. Thus we're better off with this patch now versus waiting for that refactor.

PR #3021, opened 2026-04-30
Demo 2305

### LINKED ISSUE #2305: Improvement: Removing testify from production binaries

The `bao` production binary currently includes the `github.com/stretchr/testify` library.

The `github.com/stretchr/testify/assert` package is being imported within production source files rather than being restricted to `_test.go` files. Because `testify` includes global variable initializers (e.g. [here](https://github.com/stretchr/testify/blob/master/assert/assertion_compare.go)), Go cannot eliminate the code, even if the functions themselves are likely unreachable in production.

I would like `testify` to be completely eliminated from the production binary. 

Motivation:

* It's a bit strange for a tool like OpenBao to include this unrelated test package in the final build.
* It might cause extra work if there's a vulnerability anywhere in the `testify` project, or its dependencies, it will show up as a vulnerability for OpenBao too.

**Additional context**

I see `testify` is imported in [`vault/testing.go`](https://github.com/openbao/openbao/blob/main/vault/testing.go). It looks like this is called by production code from [`command/server.go`](https://github.com/openbao/openbao/blob/main/command/server.go) for dev mode, but there might be other cases too.

How to reproduce the problem without OpenBao:

```
$ go mod init example.com/test
$ go get github.com/stretchr/testify
$ cat > main.go <<EOF
package main

import (
        "fmt"
        _ "github.com/stretchr/testify/assert"
)

func main() {
        fmt.Println("Hello World!")
}
EOF

$ go mod tidy
$ go build .
```

I compared the binaries with and without the `github.com/stretchr/testify/assert` import. Here is what I found:

* The binary size is a bit larger (up 2.9MB before stripping, and 2.1MB after).
* The import adds these indirect dependencies: `github.com/davecgh/go-spew` `github.com/pmezard/go-difflib` `gopkg.in/yaml.v3`. I didn't check if OpenBao already uses these even without `testify`.
* Using `go tool nm` shows about 3000 more symbols with the import.

Also Vault has `testify` in the builds so this was not introduced after the fork.

Shoowa:
## Description
1. Replace inline comparisons in _http/handler_test.go_ with a set of native test assertions created in _internal/assert/assert.go_.
2. Remove the imported _testify_ pkg from the file _http/handler_test.go_ in the native _http_ pkg.
3. Replace the empty string "" in tests with a constant named _EMPTY_. Minor refactoring.
4. Replace the native _http.testResponseStatus_ & _http.testResponseBody_ in _http/http_test.go_ with _assert.HttpStatusEqual_ & _assert.HttpJsonResponse_ respectively.
5. Replace the native _http_ private test func named _respondError_ with the new _assert_ func named _HttpErrorResponse_.
6. Migrate test functions that evaluate public functions to a new, discrete *http_test* pkg in a new file named _http/handler_public_test.go_.
7. Affix the newest file with the tag _blackbox_ and affix the tag _!blackbox_ to all remaining test files inside the native _http_ pkg. This enables running _http_ tests with a tag.

## Rationale
Demonstrate the potential of an approach [discussed](https://github.com/openbao/openbao/issues/2305#issuecomment-4113029404) in that thread with @cipherboy.

> [@Shoowa](https://github.com/Shoowa) I mean, that's a theory, but does that actually remove `testify` from the final binary? 🤔 I think we'd need a proof of concept to see whether or not that's a viable strategy or not.

### 1 & 2
The first three commits demonstrate the viability of removing the _testify_ pkg.

The remaining commits simply demonstrate the creation of discrete, independent test functions that can be re-used across files and packages to effectively create blackbox testing.

### 4
A couple of private functions inside the _http_ pkg were replaced by new functions in the new _internal/assert/http.go_ file. The _assert.HttpStatusEqual_ func replaces the _http.testResponseStatus_ with an updated call to _io.ReadAll(request.Body)_, and still displays the _request.Body_ as a string when an error is detected. The _assert.HttpJsonResponse_ replaces the _http.testResponseBody_ which wrapped around _jsonutil.DecodeJSONFromReader_.

### 5
The native _http.respondError_ wraps the public _RespondError_ func found in the _Logical_ pkg. This indirection achieves consistency by forcing the native _http_ pkg to use the same error-mutating function that _Logical_ provides. But the native _http_ test files can simply invoke _logical.RespondError_ directly.

_logical.RespondError_ relies on _logical.AdjustErrorStatusCode_ which relies on an imported pkg named _go-multierror_.

Instead of invoking _logical.RespondError_ in a test file, I chose to create a new error-mutating test func named _assert.HttpErrorResponse_. It relies on a new private func named _changeHttpStatusToMatchError_ that resembles _logical.AdjustErrorStatusCode_ found in _sdk/logical/error.go_. The _changeHttpStatusToMatchError_ func uses the native _errors.Is_ & _errors.As_ methods that became available in the STDLIB September 2019, and avoids relying on the _go-multierror_ pkg imported in the _Logical_ pkg in February 2017.

Instead of refactoring the _logical.RespondError_ & _logical.AdjustErrorStatusCode_ funcs on which so many packages and features depend, I chose to provide new test functionality with one less package. These new funcs demonstrate a STDLIB approach that can eventually be adopted across the repository.

### 6 & 7
Lastly, many test functions in _http/handler_test.go_ were migrated to a new file named _http/handler_public_test.go_ and given the package name *http_test*. Those tests evaluated _http_ public functions. This new file is labeled _blackbox_ and can be activated with the command:
```bash
~/openbao $ go test -v -count=1 --tags=blackbox ./http
```

The remaining test files in the _http_ pkg were labeled _!blackbox_ and can be run as usual:
```bash
~/openbao $ go test -v count=1 ./http
```

If the maintainers would like to simply remove _testify_ from the _http/handler_test.go_ file, then the first three commits of this PR will suffice. If the maintainers would like to include the new _blackbox_ test that natively handles error detection and mutation, then all the commits will suffice.

Resolves: #2305

## Acknowledgements

       also allowed.


satoqz:
Closing this down to reduce open PRs, also see discussion in https://github.com/openbao/openbao/issues/2305.

PR #3028, opened 2026-05-03
Support when in outer block

cipherboy:
## Description

This allows skipping an entire batch of requests in one go, if desired.

## Rationale

This probably should've just been part of #2739 

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3029, opened 2026-05-03
Add iteration to profiles via for keyword

cipherboy:
## Description

Similar to #2739, this adds `for` as a keyword which yields a list or map.

While bound by context, we probably want explicit safe guards on number of elements, though this may make iterating things like lists difficult as we don't have an easy way of performing paginated lists. :thinking: 

Each `index` is used as a suffix to the iterated block's name. E.g., in a block named `users` with integer indices, you'd end up with `users.0`, `users.1`, &c., when referring to past requests/...; though with templating, the current index can easily be referred to as `users.{{ this_index }}` on an outer-block based iteration.

---

~Draft based on #3028; will be rebased when that merges.~ Let me know if people like the general idea and if so, will polish the code more. :laughing: 

Supports nested loops in outer blocks / requests. 

## Rationale

Resolves: #

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3029.txt
+++ b/changelog/3029.txt
@@ -1,0 +1,6 @@
+```release-note:improvement
+command/server: Add `for` keyword to support iterated evaluation of `initialize` and `request` blocks over a list or map.
+```
+```release-note:improvement
+sys/workflows: Add `for` keyword to support iterated evaluation of `initialize` and `request` blocks over a list or map.
+```
--- a/changelog/3029.txt
+++ b/changelog/3029.txt
@@ -2,5 +2,5 @@
 command/server: Add `for` keyword to support iterated evaluation of `initialize` and `request` blocks over a list or map.
 ```
 ```release-note:improvement
-sys/workflows: Add `for` keyword to support iterated evaluation of `initialize` and `request` blocks over a list or map.
+sys/workflows: Add `for` keyword to support iterated evaluation of `flow` and `request` blocks over a list or map.
 ```
--- a/internal/helper/profiles/fields.go
+++ b/internal/helper/profiles/fields.go
@@ -1,0 +1,44 @@
+package profiles
+
+import "fmt"
+
+func StringField(field map[string]any, parameter string) (string, bool, error) {
+	rawValue, present := field[parameter]
+	if !present {
+		return "", false, nil
+	}
+
+	value, ok := rawValue.(string)
+	if !ok {
+		return "", true, fmt.Errorf("field %q is of wrong type: expected 'string' got '%T'", parameter, rawValue)
+	}
+
+	return value, true, nil
+}
+
+func RequiredStringField(field map[string]any, parameter string) (string, error) {
+	value, present, err := StringField(field, parameter)
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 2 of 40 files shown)


internal/helper/profiles/iters.go
cipherboy, line 1: Eventually we may need to make multiple types of SourceBuilders, some which are aware of various context pieces (iterators, ...) and others which don't care. But for now, it was easy enough to add another parameter.

cipherboy:
This should be ready for review @openbao/openbao-core-committers!

internal/helper/profiles/iters.go
```diff
@@ -0,0 +1,154 @@
+package profiles
```
wslabosz-reply, line 1: no copyright header?

internal/helper/profiles/fields.go
```diff
@@ -0,0 +1,44 @@
+package profiles
```
wslabosz-reply, line 1: copyright attribution missing

changelog/3029.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:improvement
+command/server: Add `for` keyword to support iterated evaluation of `initialize` and `request` blocks over a list or map.
+```
+```release-note:improvement
+sys/workflows: Add `for` keyword to support iterated evaluation of `initialize` and `request` blocks over a list or map.
```
satoqz:
```suggestion
sys/workflows: Add `for` keyword to support iterated evaluation of `flow` and `request` blocks over a list or map.
```

website/content/docs/concepts/profiles.mdx
```diff
@@ -94,6 +94,13 @@ The core of a profile is the request, which contains the following fields:
  - `allow_failure` `(bool: false)` - whether to allow profile execution to
    continue despite failure of this request.
  - `when` `(bool: true)` - when false, skips execution of the request.
+ - `for` `([]any: optional)` - when present, yields either a list or map of
+   objects to iterate through. `this` is set to the value and `this_index`
+   is set to the index in the list or key of the map in the request. `for` is
+   computed head of any other field being evaluated (including `when`),
+   letting other fields depend on it. If this request's outer block also has a
+   `for` block, the outer block's `this` and `this_index` will be moved to
+   `outer_this` and `outer_this_index` respectively.
```
satoqz, line 103: @cipherboy should we align a little closer with Terraform and make it `for_each` instead of `for` and `each.index`, `each.value.x` instead of `this`, `this_index`?

website/content/docs/concepts/profiles.mdx
satoqz, line 1: I think a "control flow" example section that covers `when` and `for` would be very helpful.

website/content/docs/concepts/profiles.mdx
cipherboy: @satoqz I suppose we could make `each` and object. Would you do `each.outer` then to avoid polluting the global namespace too?

satoqz: @cipherboy if that won't break our necks, why not.

PR #3034, opened 2026-05-04
pki: allow user to define acme challenge ports and host

tsaarni:
## Description

This PR adds new optional configuration options for ACME client. Admins can now specify which ports and host the ACME listener uses:

- `tls_acme_http_challenge_port` `(int: 80)` - Specifies an optional port to use for HTTP challenges instead of the default. Useful when running behind a load balancer that forwards  port 80 to a different port.
- `tls_acme_alpn_challenge_port` `(int: 443)` - Specifies an optional port to use for TLS-ALPN challenges  instead of the default. Useful when running behind a load balancer that forwards port 443 to a different port.
- `tls_acme_challenge_host` `(string: "")` - Specifies an optional host to listen on for ACME challenges. Useful when wanting to limit the ACME challenge listener to a specific interface.

> [!NOTE]
> The server must still connect to privileged ports as required by the specification. This change allows the client to run on an unprivileged port when a load balancer or proxy forwards traffic from a privileged port.

## Rationale

This allows OpenBao to run without root privileges or `CAP_NET_BIND_SERVICE`.

Background:

When OpenBao runs behind a load balancer, the load balancer can listen on ports 80 and 443 for ACME challenges, while OpenBao itself listens on unprivileged ports. Currently, CertMagic always tries to bind to the privileged ports. If it can't bind due to `permission denied`, it fails completely.

The original intention was to allow administrators to run OpenBao with ACME client without requiring elevated privileges (#857):

> if a load balancer or firewall port redirection is used, the challenges will be exposed on all ports and root permissions can be dropped.

Unfortunately, this only works if port 443 is already in use when CertMagic attempts to set up its listener. CertMagic gracefully ignores an `address already in use` error, which can be demonstrated by reserving `:443` with another process. In that case, the challenge is served via port `:8200` successfully. 

However, CertMagic does not ignore error `permission denied` ([link](https://github.com/caddyserver/certmagic/blob/c7496f1a6a93025c528b03cde701bf06394440fd/solvers.go#L748-L752)). This logic is hardcoded, and there does not seem to be any way to disable the CertMagic listener in favor of users own challenge handling via `tls.Config.NextProtos`. CertMagic assumes the user's own listener **must** be running on the same port, which is why it only ignores the `address already in use` error. 

If user wants to run behind proxy on non-privieleged port, the supported solution is to configure CertMagic's `AltTLSALPNPort` (or/and `AltHTTPPort`).

Resolves: Discussion: [TLS-ALPN-01 without requiring binding to privileged port (443)](https://github.com/orgs/openbao/discussions/3025)

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3034.txt
+++ b/changelog/3034.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+pki: Add optional ACME client port and host configuration to support non-privileged operation when a load balancer forwards challenges from ports 80 and 443.
+```
--- a/internal/helper/configutil/listener.go
+++ b/internal/helper/configutil/listener.go
@@ -100,6 +100,9 @@ type Listener struct {
 	TLSACMEDisableHttpChallengeRaw any      `hcl:"tls_acme_disable_http_challenge"`
 	TLSACMEDisableAlpnChallenge    bool     `hcl:"-"`
 	TLSACMEDisableAlpnChallengeRaw any      `hcl:"tls_acme_disable_alpn_challenge"`
+	TLSACMEHttpChallengePort       int      `hcl:"tls_acme_http_challenge_port"`
+	TLSACMEAlpnChallengePort       int      `hcl:"tls_acme_alpn_challenge_port"`
+	TLSACMEChallengeHost           string   `hcl:"tls_acme_challenge_host"`
 
 	HTTPReadTimeout          time.Duration `hcl:"-"`
 	HTTPReadTimeoutRaw       any           `hcl:"http_read_timeout"`
@@ -216,6 +219,11 @@ func ParseListeners(result *SharedConfig, list *ast.ObjectList) error {
 			return multierror.Prefix(err, fmt.Sprintf("listeners.%d:", i))
 		} else {
 			l.ClusterAddress = rendered
+		}
+		if rendered, err := ParseSingleIPTemplate(l.TLSACMEChallengeHost); err != nil {
+			return multierror.Prefix(err, fmt.Sprintf("listeners.%d:", i))
+		} else {
+			l.TLSACMEChallengeHost = rendered
 		}
 
 		// Hacky way, for now, to get the values we want for sanitizing
--- a/internal/helper/listenerutil/tls_acme.go
+++ b/internal/helper/listenerutil/tls_acme.go
@@ -131,6 +131,9 @@ func NewCertificateGetter(l *configutil.Listener, ui cli.Ui, logger hclog.Logger
 		Logger:                  zapLogger,
 		DisableHTTPChallenge:    l.TLSACMEDisableHttpChallenge,
 		DisableTLSALPNChallenge: l.TLSACMEDisableAlpnChallenge,
+		AltHTTPPort:             l.TLSACMEHttpChallengePort,
+		AltTLSALPNPort:          l.TLSACMEAlpnChallengePort,
+		ListenHost:              l.TLSACMEChallengeHost,
 	}
 
 	if l.TLSACMECARoot != "" {
--- a/internal/vault/external_tests/tlslistener_binary/tls_test.go
+++ b/internal/vault/external_tests/tlslistener_binary/tls_test.go
@@ -299,10 +299,11 @@ func TestTLSListener_NonPrivileged(t *testing.T) {
 	dns.AddRecord("openbao.dadgarcorp.com", "A", cluster.ClusterNodes[node].ContainerGatewayIP)
 	dns.PushConfig()
 
-	client.Logical().Write("pki/config/acme", map[string]interface{}{
+	_, err = client.Logical().Write("pki/config/acme", map[string]interface{}{
 		"enabled":      true,
 		"dns_resolver": dns.GetRemoteAddr(),
 	})
+	require.NoError(t, err)
 
 	// Validate ACME cert acquisition works.
 	validateTLS(t, ctx, root, cluster.ClusterNodes[node].ContainerNetworkName, "openbao.dadgarcorp.com:443", dns.GetRemoteAddr(), false)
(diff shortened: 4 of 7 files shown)


cipherboy:
Hey @tsaarni and @higemaniya -- I'm curious how widely used this will be in practice. I'm very hesitant to add support to our PKI plugin to support custom challenge ports and I'd expect many other developers would be in the same boat; the ACME RFCs are very clear that you do need system privileges to solve challenges here.

If this is strictly client side (with a pass through load balancer) I think this is OK. But I would not take this as an endorsement for the PKI engine to allow non-standard port binding options as that opens up a huge can of worms.

(IOW, we can add this to client like this is doing but note that it won't mean that we'll be able to self-host entirely on non-standard ports; the PKI engine will still require privileged ports). 


Thoughts?

tsaarni:
Manual test setup:

This requires linux due to using multiple loopback addresses 

* `127.0.10.1:8200` - openbao tcp listener that acts as ACME server
* `127.0.20.1:8200` - openbao tcp listener that simulates the non-privileged listener, receiving certificates from ACME server
* `127.0.0.1:443` - socat tcp listener that simulates proxy that forwards to 127.0.20.1:8200
* `127.0.20.1:8222` - certmagic's listener (dummy, not used for anything in this case)

**(1) successful test**

CertMagic's listener is bound to non-privileged port

```
# Create config
rm -rf testdata
mkdir -p testdata/bao-data

cat <<EOF > testdata/bao.hcl
listener "tcp" {
  address     = "127.0.10.1:8200"
  tls_disable = true
}

listener "tcp" {
  address                      = "127.0.20.1:8200"
  tls_acme_ca_directory        = "http://127.0.10.1:8200/v1/pki/acme/directory"
  tls_acme_domains             = ["bao.127.0.0.1.nip.io"]
  tls_acme_cache_path          = "./testdata/bao-certmagic"
  tls_acme_disable_http_challenge = true
  tls_acme_alpn_challenge_port = 8222
  tls_acme_challenge_host      = "127.0.20.1"
}

storage "inmem" {}
EOF


# Start OpenBao
bin/bao server -config=testdata/bao.hcl  -log-level=trace

# Initialize and unseal
export BAO_ADDR=http://127.0.10.1:8200

bin/bao operator init -key-shares=1 -key-threshold=1 -format=json > testdata/init.json
UNSEAL_KEY=$(jq -r '.unseal_keys_b64[0]' testdata/init.json)
ROOT_TOKEN=$(jq -r '.root_token' testdata/init.json)
bin/bao operator unseal "$UNSEAL_KEY"

# Configure PKI and ACME server
export BAO_TOKEN="$ROOT_TOKEN"

bin/bao secrets enable pki
bin/bao secrets tune -allowed-response-headers="Last-Modified" -allowed-response-headers="Replay-Nonce" -allowed-response-headers="Link" -allowed-response-headers="Location" pki
bin/bao write pki/root/generate/internal common_name="Test Root CA"
bin/bao write pki/config/cluster path="http://127.0.10.1:8200/v1/pki"
bin/bao write pki/config/acme enabled=true
bin/bao write pki/roles/acme allow_any_name=true ttl=72h max_ttl=168h


# Start proxy
sudo socat TCP-LISTEN:443,bind=127.0.0.1,fork,reuseaddr TCP:127.0.20.1:8200

# Trigger retrieval of certificate for 127.0.20.1:8200, 
# served through hostname bao.127.0.0.1.nip.io (resolves to 127.0.0.1)
echo | openssl s_client -connect localhost:443 -servername bao.127.0.0.1.nip.io 2>/dev/null | openssl x509 -text
```

**(2) unsuccesful test**

Remove the new parameter
```
#  tls_acme_alpn_challenge_port = 8222
```

This will make CertMagic to bind to port 443 and fail

```
2026-05-04T20:01:09.362+0300 [ERROR] listener-acme.obtain: could not get certificate from issuer: identifier=bao.127.0.0.1.nip.io issuer=127.0.10.1:8200-v1-pki-acme-directory error="[bao.127.0.0.1.nip.io] solving challenges: presenting for challenge: presenting with embedded solver: could not start listener for challenge server at 127.0.20.1:443: listen tcp 127.0.20.1:443: bind: permission denied (order=http://127.0.10.1:8200/v1/pki/acme/order/d5d54f7e-4b62-1c0c-7b7c-f5bfdba3a968) (ca=http://127.0.10.1:8200/v1/pki/acme/directory)"
2026-05-04T20:01:09.362+0300 [ERROR] listener-acme.obtain: will retry: error="[bao.127.0.0.1.nip.io] Obtain: [bao.127.0.0.1.nip.io] solving challenges: presenting for challenge: presenting with embedded solver: could not start listener for challenge server at 127.0.20.1:443: listen tcp 127.0.20.1:443: bind: permission denied (order=http://127.0.10.1:8200/v1/pki/acme/order/d5d54f7e-4b62-1c0c-7b7c-f5bfdba3a968) (ca=http://127.0.10.1:8200/v1/pki/acme/directory)" attempt=2 retrying_in=2m0s elapsed=1m0.286439474s max_duration=720h0m0s
```

tsaarni:
Hi @cipherboy, It is also my understanding that this only works if the proxy forwards from a privileged port. The server (PKI engine) must still connect to privileged ports, otherwise it would not be compliant with the spec.  

@higemaniya please confirm if this was your ask and assumption as well! :)

tsaarni:
Regarding how widely used this is in practice: I have no idea :) But since load balancer and non-root was mentioned even in original PR, I thought this could make sense (although CertMagic could have allowed some easier route).

higemaniya:
Hello @tsaarni and @cipherboy,
thank you for this PR.

This change satisfies my requirement to use the ACME TLS listener with non-privileged user permissions.

My goal with this configuration is to effectively disable the certmagic listener. While certmagic does not provide a direct option for this, aligning it with the openbao listener address allows the process to continue even if the listener fails to start.

If a configuration option such as `tls_acme_disable_dedicated_listener` were available, along with the ability to adjust certmagic parameters like `AltTLSALPNPort` and `ListenHost`, that would fully meet my needs.

Best regards,

cipherboy (commented):
@tsaarni If you're looking at adding test scenarios, take a look at `vault/external_tests/tlslistener_binary/tls_test.go`; you should be able to add some of your scenarios there, depending on how creative you want to get 😆

tsaarni:
Reviving the PR by rebasing after package re-org in #3210.

cipherboy (dismissed):
Thanks @tsaarni! Mind fixing conflicts though?

tsaarni:
> Mind fixing conflicts though?

Fixed!

cipherboy (approved):
Thanks @tsaarni!

PR #3027, opened 2026-05-03
Add RFC on GRPC-based invalidation

cipherboy:
## Description

This RFC supersedes the work in #2952 (broken) and earlier RFCs. The proof of concept is much more viable for scale, though has some bugs. Code will be reworked to fit the model described in the RFC, but the proof of concept is there. 

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3035, opened 2026-05-04
Refactor context, cancellation to be co-atomic

cipherboy:
## Description

Context cancellation is bound to the particular context which needs to be cancelled; while the active context itself is mostly maintained within the bounds of the state lock, the cancellation needs to be accessible without grabbing the lock (for when we need to cancel long-running in-flight requests which are actively using the context).

However, there are re-entrant paths where we may or may not come in from a state-lock held flow, especially as we get more cross-boundary refactors.

By binding active context and its cancellation together, we can ensure co-atomic updates to both. Note that calling context cancellation multiple times is generally considered safe.

## Rationale

This is a precursor to #3027 and its related proof of concept. I think the change is good to do anyways and hence opening as a separate refactor.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/context.go
+++ b/vault/context.go
@@ -1,0 +1,75 @@
+package vault
+
+import (
+	"context"
+	"errors"
+	"time"
+)
+
+// atomicContext is a struct to bind together a context (interface) and its
+// cancel function in a single package to help with atomic.Pointer operations.
+// Notably from the CancelFunc docs, it is safe to call in multiple contexts
+// simultaneously and any subsequent calls will be ignored. These functions
+// are written so that if c.activeContext.Load() is called and there is no
+// active context, we should mostly behave like a cancelled context.
+type atomicContext struct {
+	ctx      context.Context
+	canceler context.CancelFunc
+}
+
+var _ context.Context = &atomicContext{}
+
+func NewAtomicContext(ctx context.Context, cancel context.CancelFunc) *atomicContext {
+	return &atomicContext{
+		ctx:      ctx,
+		canceler: cancel,
+	}
+}
+
+func (a *atomicContext) IsNil() bool {
+	return a == nil || a.ctx == nil
+}
+
+func (a *atomicContext) Canceler() context.CancelFunc {
+	if a.IsNil() || a.canceler == nil {
+		return func() {}
+	}
+
+	return a.canceler
+}
+
+func (a *atomicContext) Deadline() (time.Time, bool) {
+	if a.IsNil() {
+		return time.Time{}, false
+	}
+
+	return a.ctx.Deadline()
+}
+
+func (a *atomicContext) Done() <-chan struct{} {
+	if a.IsNil() {
+		// This must return a closed channel per documentation on Done to
+		// behave like a cancelled context.
+		ret := make(chan struct{})
+		close(ret)
+		return ret
+	}
+
+	return a.ctx.Done()
+}
+
+func (a *atomicContext) Err() error {
+	if a.IsNil() {
+		return errors.New("no active context")
+	}
+
+	return a.ctx.Err()
+}
+
+func (a *atomicContext) Value(key any) any {
+	if a.IsNil() {
+		return nil
+	}
+
+	return a.ctx.Value(key)
+}
--- a/vault/core.go
+++ b/vault/core.go
@@ -510,8 +510,7 @@ type Core struct {
 
 	// This can be used to trigger operations to stop running when Vault is
 	// going to be shut down, stepped down, or sealed
-	activeContext           context.Context
-	activeContextCancelFunc atomic.Pointer[context.CancelFunc]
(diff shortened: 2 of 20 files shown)


vault/request_handling.go
```diff
@@ -584,7 +584,7 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
 		return nil, consts.ErrSealed
 	}
 
-	if c.activeContext == nil || c.activeContext.Err() != nil {
+	if c.activeContext.Load().Err() != nil {
```
wslabosz-reply, line 587: I see we were checking explicitly for nil here, in that case our `.Err()` might lead to panic?

cipherboy: @wslabosz-reply Ahh, `if a.IsNil() {` will panic? :thinking: Maybe I can un-refactor and `if a == nil || a.ctx == nil { ... }`

wslabosz-reply: now looking at it again, I'm not sure, seeing as it would be `atomicContext`<nil> then `a.isNil()` should work as we believe it works?

EDIT: checked the snippet on playground, this behaves correctly 🟢

vault/logical_system_rotate.go
```diff
@@ -694,7 +694,7 @@ func (b *SystemBackend) handleRotateUpdate() framework.OperationFunc {
 			}
 		}
 
-		ctx, cancel := context.WithCancel(namespace.RootContext(b.Core.activeContext))
+		ctx, cancel := context.WithCancel(namespace.RootContext(b.Core.activeContext.Load()))
```
wslabosz-reply, line 697: now I see we are overwriting the request context here, not ideal

wslabosz-reply (commented):
`vault/external_tests/misc TestRecovery` test needs adjustment and staticcheck err needs attention

vault/logical_system_rotate.go
cipherboy: @wslabosz-reply I think it is better; rather have root generation/rotation cancelled only by active leadership stepdown not request cancellation.

vault/context.go
```diff
@@ -0,0 +1,75 @@
...
+	return a == nil || a.ctx == nil
+}
+
+func (a *atomicContext) Canceler() context.CancelFunc {
+	if a.IsNil() || a.canceler == nil {
+		return func() {}
+	}
+
+	return a.canceler
+}
```
phyrog: What's the benefit of returning the `context.CancelFunc` here instead of having a `*atomicContext.Cancel()` method that calls it directly (or noop-ing directly)?

cipherboy: @phyrog In `sealInternalWithOptions` we bind to the active context cancellation early and then call it once. I don't think there's any reason to do it in this way, so I've refactored this.

vault/core.go
```diff
@@ -2026,17 +2031,14 @@ func (c *Core) sealInternalWithOptions(grabStateLock bool) error {
...
 		doneCh := make(chan struct{})
 		go func() {
 			select {
 			case <-doneCh:
 			// Attempt to drain any inflight requests
 			case <-time.After(DefaultMaxRequestDuration):
-				if activeCtxCancel != nil {
-					(*activeCtxCancel)()
-				}
+				c.activeContext.Load().Cancel()
```
phyrog: Just to make sure the logic change here is wanted: the activeContext is now loaded during execution of cancelCtxAndLock, while before it was loaded before the function definition.

cipherboy: @phyrog Yeah, I think that's fine. We're actively trying to seal here, so if we bounce all the way (down to back up) somehow, despite being sealed, we still want to finish sealing. If we're nil, that means someone's cleared our context and presumably called cancel for us. If we're changed, we still want to cancel it.

cipherboy: @phyrog Hmm, thinking about it a bit more, I think I'm going to go back to the canceler method. It is probably cleaner than this approach. :thinking:

vault/logical_system.go
```diff
@@ -4502,7 +4502,7 @@
 		// Schedule the destroy of the upgrade path
 		time.AfterFunc(b.Core.KeyRotateGracePeriod(), func() {
 			b.Backend.Logger().Debug("cleaning up upgrade keys", "waited", b.Core.KeyRotateGracePeriod())
-			if err := b.Core.barrier.DestroyUpgrade(b.Core.activeContext, newTerm); err != nil {
+			if err := b.Core.barrier.DestroyUpgrade(b.Core.activeContext.Load(), newTerm); err != nil {
```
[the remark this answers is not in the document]
cipherboy: Bad bot.

wslabosz-reply (approved):
LGTM @cipherboy thanks!

cipherboy:
Merging on @wslabosz-reply's review, thanks all!

PR #3036, opened 2026-05-04
Mini demo 2305

### LINKED ISSUE #2305: Improvement: Removing testify from production binaries

Shoowa:
## Description
1. Replace inline comparisons in _http/handler_test.go_ with a set of native test assertions created in _internal/assert/assert.go_.
2. Remove the imported _testify_ pkg from the file _http/handler_test.go_ in the native _http_ pkg.

## Rationale
Demonstrate the potential of an approach [discussed](https://github.com/openbao/openbao/issues/2305#issuecomment-4113029404) in that thread with @cipherboy.

> [@Shoowa](https://github.com/Shoowa) I mean, that's a theory, but does that actually remove `testify` from the final binary? 🤔 I think we'd need a proof of concept to see whether or not that's a viable strategy or not.

The first three commits demonstrate the viability of removing the _testify_ pkg.

Resolves: #2305

## Acknowledgements

       also allowed.


Shoowa:
This is simply the first three commits from GH-3021. This might be easier to digest.

PR #3038, opened 2026-05-05
Reload vault stanza TLS config on SIGHUP

### LINKED ISSUE #1557: Reload `auto_auth` TLS `client_cert` and `client_key` upon SIGHUP

Had my Certbot set up to auto-renew certificates that my OpenBao uses for serving incoming requests and to authenticate itself.  That is, the cert-key pair was in two places in the agent config:
```hcl
auto_auth {
  method "cert" {
    config = {
      name              = "..."
      client_cert       = "/etc/openbao/agent-fullchain.pem"
      client_key        = "/etc/openbao/agent-privkey.pem"
      reload            = true
    }
  }
  sink "file" {
    # ...
  }
}
listener "tcp" {
  address       = "[::]:8100"
  tls_cert_file = "/etc/openbao/agent-fullchain.pem"
  tls_key_file  = "/etc/openbao/agent-privkey.pem"
  role          = "metrics_only"
}
```
Certbot has a post-renewal deploy hook that sends `SIGHUP`.  The listener config seems to be reloaded (subsequent TLS connections to port 8100 shows updated cert) but the TLS auto-auth continues to use the old key/cert.  I noticed this because the main OpenBao server started logging a large number of requests from the agent rejected due to expired client certificate.  The agent was retrying the request in a tight loop.

It'd be great if `client_cert` and `client_key` used for the `cert` auto-auth method were also reloaded upon `SIGHUP.

For now I changed my renewal hook to restart the service, not just reload.  I would like to avoid having to do this, in order to minimize agent downtime.

None so far.

**Additional context**
N/A

alexa-gt:
## Description

Implements hot-reloading of client TLS certificates and bundles when SIGHUP is received.

## Rationale

Resolves: https://github.com/openbao/openbao/issues/1557

## Acknowledgements

       also allowed.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -463,6 +463,13 @@ func (c *Config) ConfigureTLS(t *TLSConfig) error {
 	return c.configureTLS(t)
 }
 
+// ConfigureTLS invokes config.ConfigureTLS. This exposes the ability to reload
+// TLS configuration without recreating the client, which is useful for rotating
+// short-lived TLS credentials.
+func (c *Client) ConfigureTLS(t *TLSConfig) error {
+	return c.config.ConfigureTLS(t)
+}
+
 // ReadEnvironment reads configuration information from the environment. If
 // there is an error, no configuration value is updated. This is a no-op if
 // DisableEnvironment is set.
--- a/api/client.go
+++ b/api/client.go
@@ -467,6 +467,9 @@ func (c *Config) ConfigureTLS(t *TLSConfig) error {
 // TLS configuration without recreating the client, which is useful for rotating
 // short-lived TLS credentials.
 func (c *Client) ConfigureTLS(t *TLSConfig) error {
+	c.modifyLock.RLock()
+	defer c.modifyLock.RUnlock()
+
 	return c.config.ConfigureTLS(t)
 }
 
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -621,6 +621,90 @@ func TestConfigureTLS(t *testing.T) {
 	}
 }
 
+func TestClientConfigureTLS(t *testing.T) {
+	cwd, _ := os.Getwd()
+	caCertPath := cwd + "/test-fixtures/keys/cert.pem"
+
+	tests := []struct {
+		name      string
+		tlsConfig *TLSConfig
+		wantErr   assert.ErrorAssertionFunc
+		assert    func(t *testing.T, c *Config)
+	}{
+		{
+			name:      "updates RootCAs on existing client",
+			tlsConfig: &TLSConfig{CACert: caCertPath},
+			assert: func(t *testing.T, c *Config) {
+				tr := c.HttpClient.Transport.(*http.Transport)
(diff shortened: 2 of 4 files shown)


wslabosz-reply (commented):
Hello @alexa-gt, thanks for your contribution
not sure what caused the `TestInit_clientTLS` test failure, but probably not in the scope of this PR. The proposed changes make sense to me, definitely would like some test added showing the behavior of the reload now, also would like to get a second pair of eyes on this

alexa-gt:
@wslabosz-reply 

Thanks for the review. I've added unit tests and rebased onto main.

For the mysql error, I've attempted to run the test suite locally but could not reproduce. I believe it's flakiness with the CI runners

wslabosz-reply (approved):
LGTM

api/client.go
```diff
@@ -463,6 +463,13 @@ func (c *Config) ConfigureTLS(t *TLSConfig) error {
 	return c.configureTLS(t)
 }
 
+// ConfigureTLS invokes config.ConfigureTLS. This exposes the ability to reload
+// TLS configuration without recreating the client, which is useful for rotating
+// short-lived TLS credentials.
+func (c *Client) ConfigureTLS(t *TLSConfig) error {
```
cipherboy, line 469: Do we need to hold the modify lock for the client here? 🤔 We grab the one for the configuration but I'm not sure if there could be concurrent writes/reads of the configuration object proper without holding the client's modify lock?

alexa-gt: @cipherboy we do, and this just proxies the client's `config`'s underlying method call, which (already) attempts a write-lock as you mentioned.

alexa-gt: Oh I see what you mean - didn't realize there was another lock for the client - would this change work? https://github.com/openbao/openbao/pull/3038/changes/4baeacf3af687dce03acd149ed5874c51835b4bb

cipherboy: Yeah, that's what I was thinking of. Thanks @alexa-gt!

cipherboy: @wslabosz-reply Hm.... Strictly I guess we never overwrite the `config` pointer and there's nowhere that we actually read `config` in a way that grabs this lock, though a lot of the other read/write headers call this. So I'm inclined to keep this and merge unless you have objections.

(Thinking, like, `WithResponseCallbacks` performs a read/shallow copy on `config` without holding this client lock for instance, though `CloneConfig()` grabs this lock in read mode so maybe we'd want to promote this to a write lock later.)

I don't think it hurts to do it this way, though.

wslabosz-reply: @cipherboy 
I agree, doesn't hurt to lock `modifyLock` (read) of the Client, although methods using locks on both client and config seem a bit inconsistent with the patterns (seems strange that the `HttpClient` is config field and not the client /my 2c.)

cipherboy: Very good, will merge and we can adjust if needed. Thanks @alexa-gt!

PR #3040, opened 2026-05-05
tests: use alpine instead of ubuntu for TLS listener test container

### LINKED ISSUE #3039: TestTLSListener_ALPN failures: Could not build new image
**Describe the bug**

Tests in `vault/external_tests/tlslistener_binary/tls_test.go` have been extra flaky lately because of the following error:

```
tls_test.go:323: Could not build new image: failed to read image build output: context deadline exceeded
```

It appears there are two image builds involved during these test cases 

* An OpenBao container (e.g. `quay.io/openbao/openbao:latest-TestTLSListener_SelfHostedNonStandard`)
that uses `quay.io/openbao/openbao:latest` as a base and adds a locally built `bao` binary.
* `openbao_acme_tls_curl_integration` , which is based on `ubuntu:latest` and runs `apt update && apt install -y curl` during the build process.

The docker image build can take quite a while. Maybe the timeout might occur because the image drifts over time, making the `apt update` increasingly slow, who knows...

While I don't have all the details, this approach seems suboptimal. A straightforward workaround could be to build the "curl" container using an Alpine base image instead of Ubuntu. The two containers are used in these tests to provide separate addresses. On Linux, similar separation could be achieved e.g. simply by binding to different loopback addresses  This would be bigger change though...

tsaarni:
## Description

The `vault/external_tests/tlslistener_binary/tls_test.go:validateTLS()` helper builds a container image based on `ubuntu:latest` and runs `apt update && apt install -y curl` during the build process.

Downloading ubuntu and running apt frequently exceeds the timeout, resulting in flaky tests (`TestTLSListener_*` #2581).

This PR implements a minimal workaround by simply replacing the Ubuntu base image with Alpine, which hopefully reduces the container build time enough to pass the test.

## Rationale

Resolves: #3039

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/external_tests/tlslistener_binary/tls_test.go
+++ b/vault/external_tests/tlslistener_binary/tls_test.go
@@ -281,9 +281,9 @@ func validateTLS(t *testing.T, ctx context.Context, root string, networkName str
 	// in a certificate with the container network's IP address rather than
 	// localhost.
 	containerfile := `
-FROM ubuntu:latest
+FROM alpine:latest
 
-RUN apt update && DEBIAN_FRONTEND="noninteractive" apt install -y curl
+RUN apk add --no-cache curl
 
 COPY root.pem /root.pem
 `


wslabosz-reply (approved):
test failure unrelated

cipherboy (approved):

PR #2925, opened 2026-04-20
Make root token generation namespace aware

wslabosz-reply:
## Description
- encapsulated root generation progress and config in one struct contained within a map of namespace key (root namespace accommodated in the same struct)
- consolidated root token generation methods to reuse `SealManager` methods
- adjusted tests and code references

## Rationale
Required to enable per-namespace root token generation.

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3041, opened 2026-05-06
Add authenticated root token generation endpoints

wslabosz-reply:
based on #2925, requires a rebase after a merge.

## Description
Introduced authenticated `sys/generate-root-token` endpoints allowing for per-namespace root token creation.

Changes originally sourced from [commit](https://github.com/openbao/openbao/commit/ba8ef86fc34c74a2281bb33980414e4e71f101ea) created by @rencooo

## Rationale
Root token generation for a namespace will enable specific namespace operators to setup the namespace before usage without requiring "higher-order" operator. Also introduction of authenticated endpoints aims at filling the gap after the deprecation of the unauthenticated ones.

Resolves: no issue, a follow-up to #2912

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/logical_system_generate_root.go
+++ b/vault/logical_system_generate_root.go
@@ -1,0 +1,461 @@
+// Copyright (c) 2025 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package vault
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/hashicorp/go-secure-stdlib/base62"
+	"github.com/openbao/openbao/helper/namespace"
+	"github.com/openbao/openbao/sdk/v2/framework"
+	"github.com/openbao/openbao/sdk/v2/helper/roottoken"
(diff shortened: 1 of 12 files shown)


vault/logical_system_generate_root.go
```diff
@@ -0,0 +1,461 @@
...
+	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+		nonce := ""
+		key := ""
+
+		if dataNonce, ok := data.GetOk("nonce"); ok {
+			nonce = dataNonce.(string)
+		}
+		if dataKey, ok := data.GetOk("key"); ok {
+			key = dataKey.(string)
+		}
```
cipherboy, line 298: I think we can simplify this to:

```suggestion
		nonce := data.Get("nonce").(string)
		key := data.Get("key").(string)
```

since we default to the empty string.

cipherboy (approved):
I think the nits can probably be addressed more globally later tbh.

vault/logical_system_generate_root.go
cipherboy: TBH, nowhere are we actually consistent about this.

PR #3052, opened 2026-05-08
RFC - Storage Index in Headers for Consistency Control

cipherboy:
## Rationale

Related: #1528

cc: @driif @klaus-sap @phil9909 

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/community/rfcs/index-headers.mdx
+++ b/website/content/community/rfcs/index-headers.mdx
@@ -1,0 +1,143 @@
+---
+sidebar_label: Index headers
+description: |-
+  An OpenBao RFC for allowing a client-indicated minimum storage index
+  coupled with server-returned post-operation index.
+---
+
+# Support client-indicated minimum storage indices
+
+## Summary
+
+When introducing [horizontal read scalability](./standby-nodes-handle-read-requests.mdx),
+eventual consistency was assumed to be sufficient as it was the semantics of
+Vault Enterprise. This RFC proposes a series of headers allowing clients to
+better achieve consistency guarantees through a mixture of retry and server
+request forwarding, if enabled by the server operator. All server requests on
+indexed storage backends will return the new `X-Vault-Index` response header,
+with clients being able to specify the `X-Vault-Index` request header to
+ensure that index has been replicated before processing the request.
+
+## Problem Statement
+
+While Vault Enterprise has eventual consistency guarantees, the previous cold
+HA-only standby option implicitly had strict consistency by only having a
+single active node. This affects certain provisioning flows and similar: mount
+enablement being asynchronous means that OpenTofu and similar tools need to
+know to retry after mounts are present if they happen to be talking to a
+standby node.
+
+We first need a way of achieving strict consistency from a client's
+perspective while allowing updates to the client's code.
+
+## User-facing Description
+
+Responses from OpenBao will include the `X-Vault-Index` header, indicating
+the underlying index from the storage backend. This index is expected to be
+opaque and passed back to OpenBao via the `X-Vault-Index` request header.
+Whenever the client expects a write to have occurred as a result of the
+request (or always depending on the preferences of the caller), it can specify
+the returned index or any later value on all future requests to ensure the
+specified index has been met by the server.
+
+This will have one of the following behaviors:
+
+1. If the `X-Vault-Inconsistent` header is missing or takes the value
+   `retry-stale`, the request will be rejected with a 412 HTTP status code
+   and the client should retry the request later.
+2. If the `X-Vault-Inconsistent` header is set to `forward-active-node`, the
+   request will be forwarded to the active node.
+3. If the `X-Vault-Inconsistent` header is set to `await-state`, the request
+   will be held until the operator configured timeout has been met and then
+   will be either forwarded or rejected depending on operator preference.
+
+## Technical Description
+
+With [GRPC invalidation](./grpc-invalidation.mdx), all storage backends will
+implement the `physical.ReplicationIndexBackend` type, allowing us to check
+and compare indices. On every request, we'll call
+`backend.AppliedReplicationIndex(...)` to check the latest index; we assume
+this function should be reasonably fast.
+
+We'll add a few new functions to core to check indices and potentially forward
+or back off a client until replication is synced. Sync here means both that
+storage is up-to-date and all invalidations from previous indices have been
+processed. To do so, we'll need to:
+
+1. Update invalidation to record the corresponding storage index at the time
+   of invalidation.
+2. Check if the index or any predecessors exist in the pending invalidation
+   queue.
+3. Check the underlying job manager for any pending jobs with the target index
+   or any predecessors.
+
+Lastly, for the actual handling of the headers we'll inject a new middleware
+that applies fairly early in the stack.
+
+While nominally Vault Enterprise only returns the `X-Vault-Index` response
+header on operations which incurred a write, we'll opt to always return the
+index for ease of use.
+
+## Rationale and Alternatives
+
+This gives clients and operators an additional tool in the consistency
+spectrum, to give the appearance of strict consistency without the system
+actually having it.
+
+Alternatives such as the restored server-side consistent tokens do not fully
+solve this problem: the existing SSCT logic was only for cross-cluster
+consistency, meaning we currently embed the equivalent of a zero index. While
+we could add our new index logic to that value, it has a few shortcomings:
+
+1. Some storage engines, like PostgreSQL, use semi-opaque strings rather than
+   integers. We'd ideally want to continue supporting non-string indices in
+   case other systems have this property as well.
+2. Tokens are not updatable and only reflect the index on issuance. By using
+   response/request headers, we can handle arbitrary post-auth writes without
+   having to create a nested series of tokens with new indices and incur even
+   more storage writes.
+
+Another alternative would be server-side consistency: on all write requests,
+we could wait for the request to return until known standby nodes are
+up-to-date with the specified index. However, this could cause issues when the
+request was successfully processed but a slow standby node hasn't yet
+confirmed it.
+
+## Downsides
+
+Go's goroutine parking is already fairly efficient; this means the overhead
+of handling server-side wait will likely be less than the work of handling a
+request whose corresponding index hasn't yet been reached.
+
+Cost of the index handling is only incurred when a client indicates the
+header value.
+
+## Security Implications
+
+This should have no meaningful security impact: specifying an index too far
+in the future will at worst forward the request to the active node (though,
+should not be relied on!) and at best is a self-DoS.
+
+As HashiCorp Vault Enterprise's indices are opaque base64 values, we could
+apply a HMAC or similar protection to returned index headers, though these
+mostly have a client-local impact and malformed or future indices would not
+cause much stress on the server.
+
+## User/Developer Experience
+
+We can build these changes into the API client to give us a consistent by
+default client. Other clients will need to use additional mechanisms. This
+would be compatible with Vault Enterprise's implementation.
+
+## Unresolved Questions
+
+n/a
+
+## Related Issues
+
+- https://github.com/openbao/openbao/issues/1462
+
+## Proof of Concept
+
+See [proof of concept](https://github.com/openbao/openbao/compare/main...cipherboy:openbao:index-headers?expand=1)
+branch.
--- a/website/content/community/rfcs/index-headers.mdx
+++ b/website/content/community/rfcs/index-headers.mdx
@@ -57,7 +57,7 @@ operator configuration value of default behavior.
 
 ### Configuration parameters
 
-Operators can define two new consistency on a per-listener bases:
+Operators can define two new consistency parameters on a per-listener basis:
 
  - `consistency_missing_header_forward` `(bool: false)` - which by default
    handles requests to standby nodes locally which lack the `X-Vault-Inconsistent`
--- a/website/content/community/rfcs/index-headers.mdx
+++ b/website/content/community/rfcs/index-headers.mdx
@@ -43,7 +43,7 @@ specified index has been met by the server.
 This will have one of the following behaviors:
 
 1. If the `X-Vault-Inconsistent` header is missing or takes the value
-   `retry-stale`, the request will be rejected with a 412 HTTP status code
+   `fail`, the request will be rejected with a 412 HTTP status code
    and the client should retry the request later.
 2. If the `X-Vault-Inconsistent` header is set to `forward-active-node`, the
    request will be forwarded to the active node.
@@ -90,7 +90,7 @@ consistency, meaning we currently embed the equivalent of a zero index. While
 we could add our new index logic to that value, it has a few shortcomings:
 
 1. Some storage engines, like PostgreSQL, use semi-opaque strings rather than
-   integers. We'd ideally want to continue supporting non-string indices in
+   integers. We'd ideally want to continue supporting non-integer indices in
    case other systems have this property as well.
 2. Tokens are not updatable and only reflect the index on issuance. By using
    response/request headers, we can handle arbitrary post-auth writes without
--- a/website/content/community/rfcs/index-headers.mdx
+++ b/website/content/community/rfcs/index-headers.mdx
@@ -51,6 +51,26 @@ This will have one of the following behaviors:
    will be held until the operator configured timeout has been met and then
    will be either forwarded or rejected depending on operator preference.
 
+The `X-Vault-Inconsistent` header can also be specified twice: `await-state`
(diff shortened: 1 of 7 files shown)


cipherboy:
This is built on top of #3027, will be rebased when that lands. We could land this separately both from a code and RFC PoV, though.

website/content/community/rfcs/index-headers.mdx
```diff
@@ -0,0 +1,143 @@
...
+Whenever the client expects a write to have occurred as a result of the
+request (or always depending on the preferences of the caller), it can specify
+the returned index or any later value on all future requests to ensure the
+specified index has been met by the server.
+
+This will have one of the following behaviors:
+
+1. If the `X-Vault-Inconsistent` header is missing or takes the value
+   `retry-stale`, the request will be rejected with a 412 HTTP status code
+   and the client should retry the request later.
```
phil9909, line 47:
```suggestion
1. If the `X-Vault-Inconsistent` header is missing or takes the value
   `fail`, the request will be rejected with a 412 HTTP status code
   and the client should retry the request later.
```

Based on constant from `fork-point` https://github.com/openbao/openbao/blob/8993802145833ab01d49c6070d787a9eccb81546/http/handler.go#L69-L70

website/content/community/rfcs/index-headers.mdx
```diff
@@ -0,0 +1,143 @@
...
+   queue.
+3. Check the underlying job manager for any pending jobs with the target index
+   or any predecessors.
+
+Lastly, for the actual handling of the headers we'll inject a new middleware
+that applies fairly early in the stack.
+
+While nominally Vault Enterprise only returns the `X-Vault-Index` response
+header on operations which incurred a write, we'll opt to always return the
+index for ease of use.
```
phil9909: This also has some advantage: You might want to do two read operations and want to be sure the second read is at least as up to date as the first one. Not sure, if there is such a use-case TBH.


The downside is: If the client you are using automatically stores the header value and applies it to the next request this might slow you down unnecessarily.

website/content/community/rfcs/index-headers.mdx
```diff
@@ -0,0 +1,143 @@
...
+## Downsides
+
+Go's goroutine parking is already fairly efficient; this means the overhead
+of handling server-side wait will likely be less than the work of handling a
+request whose corresponding index hasn't yet been reached.
+
+Cost of the index handling is only incurred when a client indicates the
+header value.
+
+## Security Implications
```
phil9909, line 136: A malicious actor with low privileges (or even unauthenticated if we really always add the header) could watch the `X-Vault-Index` and deduce the amount of write activity on the cluster. Not sure, if we consider this "metadata-leak" problematic.
It might help them timing another attack.

website/content/community/rfcs/index-headers.mdx
```diff
@@ -0,0 +1,143 @@
...
+This will have one of the following behaviors:
+
+1. If the `X-Vault-Inconsistent` header is missing or takes the value
+   `retry-stale`, the request will be rejected with a 412 HTTP status code
+   and the client should retry the request later.
+2. If the `X-Vault-Inconsistent` header is set to `forward-active-node`, the
+   request will be forwarded to the active node.
+3. If the `X-Vault-Inconsistent` header is set to `await-state`, the request
+   will be held until the operator configured timeout has been met and then
+   will be either forwarded or rejected depending on operator preference.
```
phil9909, line 52: Would we want to allow clients to specify the fallback via a multi-valued header? E.g. `X-Vault-Inconsistent: await-state,forward-active-node`

Also, if we allow the operator to specify a fallback, should that be also the case for a missing `X-Vault-Inconsistent` header (the RFC currently suggests a hard-coded default to status 412)?

website/content/community/rfcs/index-headers.mdx
```diff
@@ -0,0 +1,143 @@
...
+actually having it.
+
+Alternatives such as the restored server-side consistent tokens do not fully
+solve this problem: the existing SSCT logic was only for cross-cluster
+consistency, meaning we currently embed the equivalent of a zero index. While
+we could add our new index logic to that value, it has a few shortcomings:
+
+1. Some storage engines, like PostgreSQL, use semi-opaque strings rather than
+   integers. We'd ideally want to continue supporting non-string indices in
+   case other systems have this property as well.
```
phil9909, line 115:
```suggestion
1. Some storage engines, like PostgreSQL, use semi-opaque strings rather than
   integers. We'd ideally want to continue supporting non-integer indices in
   case other systems have this property as well.
```

website/content/community/rfcs/index-headers.mdx
DrDaveD: For backward compatibility to older clients, I think it is important for a missing `X-Vault-Inconsistent` header to be able to default to forward to the active node.

website/content/community/rfcs/index-headers.mdx
cipherboy: Well, that, and it is significantly harder to implement. We don't have a general purpose way of tracking requests -> write operations tbh. We could shim in YASM (Yet Another Storage Middleware) layer, but that seems like overkill for this.

website/content/community/rfcs/index-headers.mdx
cipherboy: I think this goes in the general bucket of [`~label:perceived-security-through-obscurity`](https://github.com/openbao/openbao/issues?q=label%3Aperceived-security-through-obscurity).

Not sure what attack this would help with timing...?

website/content/community/rfcs/index-headers.mdx
cipherboy: Additional configuration parameters have been added which addresses both of these, and the parameter has been made to be multi-valued.

cipherboy:
@openbao/openbao-core-committers This should be updated as well.

website/content/community/rfcs/index-headers.mdx
phil9909: I was just trying to discuss, if the the following fact (correct me if the "fact" itself is wrong) is really not security relevant:

---

Currently, it is not possible for an attacker to know the number of writes in a cluster. If we implement this RFC an attacker can track the Raft Index / PostgreSQL LSN and easily deduce the number of writes (or deletes) that have happened in the cluster.

---

I'm not saying it definitely is security relevant, but I found it relevant enough to discuss it.
The worst thing I could come up with is this scenario: Assume there is an vulnerability on OpenBao, which abuses some data-race[^1]. An attacker could watch the cluster index for some time and might learn "every night at 2AM there is a high number of writes" and therefore know this is the best time to use the exploit.
Or they might just use the same information as an indicator when a DDoS attack will have the most impact.

If nobody can think of a more severe attack based on this, I guess we are fine. But we might as well limit the header to responses of authenticated requests (and responses of successful logins), just to be on the safe side. I don't think this makes this feature any less useful.


[^1]: Alternatively it could be a adjacent/consuming system of OpenBao, which is only vulnerable during some sort of credential rotation.

website/content/community/rfcs/index-headers.mdx
```diff
@@ -0,0 +1,164 @@
...
+Operators can define two new consistency on a per-listener bases:
+
+ - `consistency_missing_header_forward` `(bool: false)` - which by default
+   handles requests to standby nodes locally which lack the `X-Vault-Inconsistent`
+   headers, but allows defaulting to forwarding requests with missing headers
+   to the active node for strict consistency guarantees on legacy clients.
+ - `consistency_fallback_behavior` `(string: "forward-active-node")` - which
+   takes a second value for `X-Vault-Inconsistent` if only `await-stae` is
+   specified. The value `fail` rejects the request with an HTTP 412 status
+   code.
```
phil9909, line 69: Should it also be the default behavior, if `X-Vault-Index` is given but `X-Vault-Inconsistent` is missing?

Also, typo:

```suggestion
 - `consistency_fallback_behavior` `(string: "forward-active-node")` - which
   takes a second value for `X-Vault-Inconsistent` if only `await-state` is
   specified. The value `fail` rejects the request with an HTTP 412 status
   code.
```

phil9909 (dismissed):

website/content/community/rfcs/index-headers.mdx
```diff
@@ -0,0 +1,164 @@
...
+## Security Implications
+
+This should have no meaningful security impact: specifying an index too far
+in the future will at worst forward the request to the active node (though,
+should not be relied on!) and at best is a self-DoS.
+
+As HashiCorp Vault Enterprise's indices are opaque base64 values, we could
+apply a HMAC or similar protection to returned index headers, though these
+mostly have a client-local impact and malformed or future indices would not
+cause much stress on the server.
```
satoqz, line 145: A header can still be up to ~1MB in size, since we use Go's `DefaultMaxHeaderBytes`. Need to be careful here.

website/content/community/rfcs/index-headers.mdx
```diff
@@ -0,0 +1,164 @@
...
+   will be held until the operator configured timeout has been met and then
+   will be either forwarded or rejected depending on operator preference.
+
+The `X-Vault-Inconsistent` header can also be specified twice: `await-state`
+with an additional value of `fail` or `forward-active-node` bypasses the
+operator configuration value of default behavior.
+
+### Configuration parameters
+
+Operators can define two new consistency on a per-listener bases:
```
satoqz:
```suggestion
Operators can define two new consistency parameters on a per-listener basis:
```

website/content/community/rfcs/index-headers.mdx
satoqz, line 69: Do we need this option? I find the combination of header + 2 related but not-quite-the-same listener options very confusing and would love to limit complexity.

Specifically, since this option this only applies if the client sets `await-state`, which isn't the default and is hence intentional, I don't really see the use case given the client can set it themselves?

website/content/community/rfcs/index-headers.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+ - `consistency_missing_header_forward` `(bool: false)` - which by default
+   handles requests to standby nodes locally which lack the `X-Vault-Inconsistent`
+   headers, but allows defaulting to forwarding requests with missing headers
+   to the active node for strict consistency guarantees on legacy clients.
```
satoqz, line 65: @DrDaveD said:

> For backward compatibility to older clients, I think it is important for a missing X-Vault-Inconsistent header to be able to default to forward to the active node.

@cipherboy said:

> Additional configuration parameters have been added which addresses both of these.

I'm curious (it is perhaps due to my lack of having worked with Vault's equivalent) which older clients we're talking about here? Specifically, which clients do:

- Set the `X-Vault-Index` header.
- Omit the `X-Vault-Inconsistent` header.
- _but_ expect forwarding instead of failing to meet the index?

It seems to me (based on Vault documentation) that the fallback in Vault is to fail, not to forward? Similarly, we're not breaking older clients that never set `X-Vault-Index` either, nothing changes for those?

DrDaveD: Yes, I'm thinking about older clients that don't send either header.  When I made that comment I hadn't noticed that there were two headers being discussed.  

As I understand things, the default behavior since 2.5.x is to allow reads from standby nodes, even though they might not be up to date, due to eventual consistency.  I recently found this breaks my primary application so I turned off the standby read feature.  This PR adds control of the consistency by clients, so I should be able to add that control to the client to allow standby reads again.  But then I worry about older versions of the client that don't do the control; those will need to fall back to doing everything on the active node.

cipherboy: @satoqz Yes, I'm thinking about @DrDaveD's scenario. If `consistency_missing_header_forward` is set, and `X-Vault-Inconsisent` is not set, I'm _assuming_ that `X-Vault-Index` is also not set.

This would give you a way of having an "optional read-enabled standby", wherein a client which is aware of the indices can opt-into strong 


```mermaid
flowchart LR
    R-->XVIdx

    XVIdx -- yes --> XVIwIdx
    XVIwIdx -- missing/fail --> XVIwIdxMFUpToDate
    XVIwIdxMFUpToDate -- yes --> HR
    XVIwIdxMFUpToDate -- no --> 412
    XVIwIdx -- forward_active_node --> XVIwIdxFANUpToDate
    XVIwIdxFANUpToDate -- yes --> HR
    XVIwIdxFANUpToDate -- no --> FAN
    XVIwIdx -- await_state --> XVIwIdxASUpToDate
    XVIwIdxASUpToDate -- yes --> HR
    XVIwIdxASUpToDate -- no with forward_active_node --> ASF
    XVIwIdxASUpToDate -- no with fail --> AS4

    XVIwIdx -- supplement with --> cfb
    cfb -- forward_active_node default --> XVIwIdx

    XVIdx -- no --> cmhf
    cmhf -- true --> XVI
    XVI -- present --> HR
    XVI -- missing --> FAN
    cmhf -- false --> HR
    

    ASF -- caught up --> HR
    AS4 -- caught up --> HR
    ASF -- timeout --> FAN
    AS4 -- timeout --> 412


    R[Request]
    XVIwIdx[X-Vault-Inconsistent]
    XVI[X-Vault-Inconsistent]
    XVIdx[X-Vault-Index]
    cmhf[consistency_missing_header_forward]
    cfb[consistency_fallback_behavior]
    XVIwIdxMFUpToDate[ index up to date? ]
    XVIwIdxFANUpToDate[ index up to date? ]
    XVIwIdxASUpToDate[ index up to date? ]
    412[Status 412]
    HR[Handle Request]
    FAN[Forward to Active Node]
    ASF[Await State or Forward]
    AS4[Await State or 412]

    style cfb fill:#fff
    style cmhf fill:#fff
    style XVI fill:#fff

```

Is what I ended up with. (The right side of the graph is confusing AF, but other than separating nodes I'm not sure how to fix the mermaid -- new config nodes and children are in white -- I count three new nodes).

The complexity from supplementing (`consistency_fallback_behavior`) is contained to make every single-valued `X-Vault-Inconsistent: await-state` value a multi-valued `{await-state, ... fallback behavior ...}` for determinism. The complexity from conditional forwarding (`consistency_missing_header_forward`) is entirely on the fallback paths when neither header is specified.

A legacy client is one which send neither `X-Vault-Index` nor `X-Vault-Inconsistent`, and is thus a candidate for forwarding for strict consistency without losing horizontal scalability for updated clients. An "updated" client which sends `X-Vault-Inconsistent` but not `X-Vault-Index` is just signaling they don't care about consistency _right now_. A client which sends `X-Vault-Index` without `X-Vault-Inconsistent` is well-defined to mean `fail` (per existing VEnt compatibility) and thus gets a 412 Retry, which _should_ be already handled (as both Vault and OpenBao have sent it on some paths, e.g., rate limit quotas) for a while I believe.

website/content/community/rfcs/index-headers.mdx
cipherboy: @satoqz See below diagram. If we don't have this, what behavior does `await-state` take if the value is never reached locally? Forwarding or asking the client to retry? IMHO the latter is a self-DoS if the client sends an index value too far in the future, but the former can cause increased traffic to the primary if a number of clients send bogus index values (where previously they'd only be forwarded if the index is incorrect).

I suspect _most_ people will use await-state as it gives most expedient retries (as most inconsistencies are resolved rather quickly given replication+invalidation generally don't take _that_ long) and also avoids unnecessary traffic to the primary (given a read-heavy workload).

My 2c., but I'd definitely like to see `await-state` continuing to exist. 

If we can agree on a fallback behavior, and conclude that nobody would want to change it, that's fine by me and we can always revisit, but we should choose one.

website/content/community/rfcs/index-headers.mdx
cipherboy: I've moved this locally up into the alternatives to make it clear it is not an accepted suggestion.

website/content/community/rfcs/index-headers.mdx
cipherboy: @phil9909 No, per VEnt docs: https://developer.hashicorp.com/vault/docs/enterprise/consistency#retry-stale-requests

> When the desired state isn't present, Vault will return a failure response with HTTP status code 412. This tells the client that it should retry the request. The advantage over the Conditional Forwarding solution above is twofold: first, there's no additional load on the active node. Second, this solution is applicable to performance replication as well as performance standbys.

This is on the example without the `X-Vault-Inconsistent` header, leading me to believe the default value for an empty header should be `fail`, not whatever this boolean is. 

---

I've fixed the typo, though.

website/content/community/rfcs/index-headers.mdx
cipherboy: @phil9909 I've thought about this more and haven't had any other ideas of what we'd reasonably consider a security vulnerability adjacent to this.

satoqz (approved):

PR #3055, opened 2026-05-08
Add RFC on encrypted storage paths

cipherboy:
## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/rfcs/encrypted-paths.mdx
+++ b/website/content/docs/rfcs/encrypted-paths.mdx
@@ -1,0 +1,99 @@
+---
+sidebar_label: Encrypted paths
+description: |-
+  An OpenBao RFC to discuss enabling encryption on all barrier-encrypted
+  paths, preventing the last remaining information leak on snapshots.
+---
+
+# Encrypted storage paths
+
+## Summary
+
+OpenBao's barrier currently protects the contents of data written to OpenBao,
+but doesn't handle encrypting paths. This means that certain sensitive paths,
+most notably KVv1 entries, are written in plaintext and visible in snapshots
+or to anyone with access to the underlying storage. By layering the existing
+`keysutil.EncryptedKeyStorageWrapper` helper on top of barrier, we can improve
+the defense of these paths.
+
+## Problem Statement
+
+OpenBao currently has storage paths which are not encrypted, despite the
+data itself being encrypted. This is a metadata leak that we should ideally
+resolve and is something already addressed in the KVv2 storage engine through
+the use of `keysutil.EncryptedKeyStorageWrapper`. This contains a LRU cache,
+decreasing the overhead of encrypting all paths in OpenBao.
+
+Notably, paths not encrypted with the barrier are out of scope.
+
+## User-facing Description
+
+On initialization, operators can set `with_encrypted_storage_paths=true`.
+
+No change is expected for consumers of the API; `sys/raw` will additionally
+already decode paths. Only direct viewers of the raw underlying physical
+backend would be impacted.
+
+Rotation of key material occurs at the same time as barrier key rotation.
+
+## Technical Description
+
+When `with_encrypted_storage_paths=true` is set, we can provision a copy of
+`keysutil.EncryptedKeyStorageWrapper` over the top of `vault.Core.barrier`,
+taking into account the specialized interface of `c.barrier`. This already
+supports transactions and parallel access, so no real technical challenges
+exist. The proof of concept is relatively compact.
+
+The biggest gap remains in unanswered questions: how would we implement
+migrations, if desired. One possibility would be to do this in the offline
+`bao operator migrate`.
+
+## Rationale and Alternatives
+
+This fixes a long-standing metadata leak in OpenBao and HashiCorp Vault.
+
+We could implement a different wrapper than the existing
+`keysutil.EncryptedKeyStorageWrapper`, though this would be much more work
+and likely look similar to the existing interface.
+
+## Downsides
+
+This would add some performance loss, similar to the overhead of KVv1 vs KVv2,
+though less as the latter also incurs multiple storage requests for most
+operations.
+
+## Security Implications
+
+The backend for encrypting paths is already used by KVv2 and is thus fairly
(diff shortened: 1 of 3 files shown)


wslabosz-reply (commented):
question of on top from my head: this just encrypts the keys of the storage entries? doesn't hide the # of keys on a specific path?

also to clarify: `sys/raw` endpoints are unaffected by the virtue of using `physical` underhood and not the barrier right?

cipherboy:
@wslabosz-reply All paths are encrypted, so you'd have to know what maps to what. E.g., `logical` becomes encrypted, `<namespace_id>` is then encrypted, &c. I'm not saying this is impossible to differentiate between namespaces/* and logical/*, but perhaps differentiating between two different mounts of the same type at the same level (or peer levels) might be difficult. 

`sys/raw` accesses barrier so all paths are legible to such callers. As mentioned, this mostly helps with offline backup snooping more so than active online snooping.

wslabosz-reply:
@cipherboy 
how does it interact with multiple barrier wrapping, e.g. when you have namespaces with their own barriers? multiple levels of encryption, or would that be configurable on the barrier (namespace) level?

cipherboy:
> @cipherboy how does it interact with multiple barrier wrapping, e.g. when you have namespaces with their own barriers? multiple levels of encryption, or would that be configurable on the barrier (namespace) level?

This applies on top of any given barrier. To my knowledge, we never have dual barrier writes... We only have single barrier writes going back down to physical?

So on namespace sealing:

1. Items like the root key (which are written unencrypted w.r.t. sealed namespace's PoV) have paths which are encrypted with the parent barrier's key encryption mechanism.
2. Items like the namespace's data are encrypted with this new sealed namespace's path encryptor.

So within a sealed namespace, _all_ paths will be fully encrypted. But obviously in the root namespace, paths which are written directly to physical (bypassing barrier) wouldn't be encrypted.

wslabosz-reply:
@cipherboy
> This applies on top of any given barrier. To my knowledge, we never have dual barrier writes... We only have single barrier writes going back down to physical?

yeah, it's not like we have multiple levels of encryption, just asked if we would have that with this change, I understand that it's not the case.

> So within a sealed namespace, all paths will be fully encrypted. But obviously in the root namespace, paths which are written directly to physical (bypassing barrier) wouldn't be encrypted.

I wonder if this will render our method of figuring out if a sealed namespace has child namespaces (by checking if any key exists under `<namespace>/core/namespaces/*` unusable

DanGhita (approved):
The RFC LGTM (I guess the related code will land in a dedicated PR ?).
However, it's not very clear if the  path encryption concerns only KVv1 or there are also another engines candidates (if yes, which ones ?).

cipherboy:
Thanks @DanGhita! 

> However, it's not very clear if the path encryption concerns only KVv1 or there are also another engines candidates (if yes, which ones ?).

This is for all of OpenBao's storage, not just any one plugin but all storage: plugins (yes), auth methods, ACL policies, tokens, leases, ... 

I think we've defined barrier e.g., here: https://github.com/openbao/openbao/blob/d201595f55dc6fa37c077cb18b75e9166d2eebec/physical/crosstest/cross_test.go#L17-L19 or in the glossary: https://openbao.org/docs/glossary/#barrier 

The few exceptions are unencrypted entries required to unseal OpenBao, namely the root key (encrypted with Shamirs and/or auto-unseal) and the seal configuration.

cipherboy:
@wslabosz-reply said:

> I wonder if this will render our method of figuring out if a sealed namespace has child namespaces (by checking if any key exists under `<namespace>/core/namespaces/*` unusable

Oh, hm... :thinking:  I think you're right that this won't work easily.

Because the barrier is unique, we'll use different path encryptors for every namespace barrier, which means the encoding of `namespace/<namespace_id>` will be different on every single backend.

I think we have two options here:

1. Use a global single path encryptor for all namespaces. I don't think that is this bad, given there's less data leakage and we'd still maintain the metadata obfuscation properties we wanted.
2. We could create an encryptor view which would encrypt just the prefix (`namespaces/<id>`) and then let the sub-view handle the subsequent encryption of the remaining part of the paths. We'd then reverse the same peering at physical.

Regarding checking if any key exists, this should just be a `list` operation; if the to-be-deleted namespace's barrier is sealed, we won't have access to the decrypter for paths and thus won't be able to read `core/namespaces/*` under the second approach. 

Which may mean we're at the first solution. :thinking:

website/content/docs/rfcs/encrypted-paths.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+We could implement a different wrapper than the existing
+`keysutil.EncryptedKeyStorageWrapper`, though this would be much more work
+and likely look similar to the existing interface.
+
+## Downsides
+
+This would add some performance loss, similar to the overhead of KVv1 vs KVv2,
+though less as the latter also incurs multiple storage requests for most
+operations.
```
phil9909, line 63: Not really a downside of this RFC, because it will be an optional, opt-in feature.

phil9909:
I like the idea :+1: 

How would we evolve KVv2 in this case? Encrypting the path twice would be bad (for performance).

cipherboy:
> How would we evolve KVv2 in this case? Encrypting the path twice would be bad (for performance).

I'd actually want to benchmark what performance is, but we could indicate to plugins that they're already using encrypted storage and thus drop double encryption. :thinking:

website/content/docs/rfcs/encrypted-paths.mdx
cipherboy: That's fair. Though it also has a LRU cache natively built in, so we could expose some tuning parameters there.

phil9909:
> I'd actually want to benchmark what performance is, but we could indicate to plugins that they're already using encrypted storage and thus drop double encryption. 🤔

Make migration more interesting :D

cipherboy:
@phil9909 said:

> Make migration more interesting :D

I think we'd only do it for new KVv2 mounts fwiw.

JanMa (approved):

PR #3058, opened 2026-05-08
Fix namespace unsealing restart

cipherboy:
## Description

This fixes a bug in namespace seals which prevents us from working on HA storage backends after the server restarts or is sealed. Each individual commit has details. To reproduce:

```
$ ( ( killall bao && sleep 1 ) ; devbao node start --force --initialize --unseal --profiles secret,userpass,pki --audit --seals static:// --ui ) && . <(devbao node env prod)
$ bao namespace create -key-shares=1 -key-threshold=1 asdf
$ bao write sys/namespaces/asdf/unseal key=...
(works)
$ killall bao && sleep 2 && devbao node resume prod
$ bao write sys/namespaces/asdf/unseal key=...
(fails)
```

## Rationale

Related: #1357

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/logical_system_namespaces_seals_test.go
+++ b/vault/logical_system_namespaces_seals_test.go
@@ -15,7 +15,7 @@ import (
 
 func TestNamespaceBackend_SealUnseal(t *testing.T) {
 	t.Parallel()
-	c, _, _ := TestCoreUnsealed(t)
+	c, rootShares, root := TestCoreUnsealed(t)
 	b := c.systemBackend
 
 	rootCtx := namespace.RootContext(context.Background())
@@ -87,7 +87,8 @@ func TestNamespaceBackend_SealUnseal(t *testing.T) {
 	})
 
 	t.Run("preserve mounts after unsealing namespaces", func(t *testing.T) {
-		keyshares := TestCoreCreateUnsealedNamespaces(t, c, &namespace.Namespace{Path: "foobar/"})
+		nsFoobar := &namespace.Namespace{Path: "foobar/"}
+		keyshares := TestCoreCreateUnsealedNamespaces(t, c, nsFoobar)
 		ns, err := c.namespaceStore.GetNamespaceByPath(rootCtx, "foobar")
 		require.NoError(t, err)
 		nsCtx := namespace.ContextWithNamespace(rootCtx, ns)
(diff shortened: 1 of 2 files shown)


cipherboy:
fyi @wslabosz-reply

vault/logical_system_namespaces_seals_test.go
```diff
@@ -179,5 +177,70 @@ func TestNamespaceBackend_SealUnseal(t *testing.T) {
...
+		require.NoError(t, c.Seal(root))
+
+		// Unseal the root namespace.
+		for i, key := range rootShares {
+			unseal, err := TestCoreUnseal(c, key)
+			require.NoError(t, err)
+
+			if i+1 == len(rootShares) && !unseal {
+				t.Fatal("err: should be unsealed")
+			}
```
wslabosz-reply: let's stick with assert/require

vault/seal_manager.go
```diff
@@ -128,9 +135,11 @@ func (sm *SealManager) SetSeal(ctx context.Context, sealConfig *SealConfig, ns *
 	defaultSeal.SetCore(sm.core)
 	defaultSeal.SetMetaPrefix(metaPrefix)
 
-	// At this point, the namespace's barrier is still the parent's barrier,
-	// hence we can just query that without computing the actual parent.
-	defaultSeal.SetConfigAccess(sm.namespaceBarrierByLongestPrefix(ns.Path))
+	// The configuration access should always at least use the parent's seal
+	// configuration information.
+	parent := filepath.Join(ns.Path, "../")
```
wslabosz-reply: I think that we could just use `ns.ParentPath()`

wslabosz-reply:
@cipherboy 
seems like namespace (un)seal commands tests are broken

vault/logical_system_namespaces_seals_test.go
cipherboy: I copy/pasted from the other location if you didn't notice, but will update ;-)

cipherboy:
> @cipherboy seems like namespace (un)seal commands tests are broken

Fixed in #3062, will rebase once that merges but locally testing shows it should be fine.

wslabosz-reply (approved):
LGTM

PR #3061, opened 2026-05-11
raft: disable snapshot compression

### LINKED ISSUE #2957: raft snapshots waste time on compression

phil9909:
## Description

This changes the raft snapshot process in a fully compatible way: Snapshots are still gzipped tar files, just a gzip stream without compression. I wasn't even required to touch the restore part.

## Rationale

Much faster snapshots at the expense of slightly larger files (exact numbers will vary based on the actual data and scale of a given cluster)

Resolves: #2957

We were discussing, if this should be configurable, but concluded to leave it hard-coded any add configurability on concrete request.

If you want to compress your snapshots, you can `gunzip` and `gzip` them again and the resulting file will successfully restore on an OpenBao cluster.

## Acknowledgements

       also allowed.

### DIFF
--- a/physical/raft/snapshot/snapshot.go
+++ b/physical/raft/snapshot/snapshot.go
@@ -141,7 +141,10 @@ func Write(logger hclog.Logger, r *raft.Raft, sealer Sealer, w io.Writer) error 
 	}()
 
 	// Wrap the file writer in a gzip compressor.
-	compressor := gzip.NewWriter(w)
+	compressor, err := gzip.NewWriterLevel(w, gzip.NoCompression)
+	if err != nil {
+		return fmt.Errorf("failed to create snapshot file: %v", err)
+	}
 
 	// Write the archive.
 	if err := write(compressor, metadata, snap, sealer); err != nil {
--- a/changelog/3061.txt
+++ b/changelog/3061.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+raft: improve snapshot duration while slightly increasing snapshot size
+```


physical/raft/snapshot/snapshot.go
```diff
@@ -141,7 +141,10 @@ func Write(logger hclog.Logger, r *raft.Raft, sealer Sealer, w io.Writer) error
 	}()
 
 	// Wrap the file writer in a gzip compressor.
-	compressor := gzip.NewWriter(w)
+	compressor, err := gzip.NewWriterLevel(w, gzip.NoCompression)
+	if err != nil {
+		return fmt.Errorf("failed to write snapshot file: %v", err)
```
wslabosz-reply: not really a failed write, but rather invalid compression level passed to the constructor

cipherboy (approved):
Would like to see @wslabosz-reply's comment addressed, but happy with the implementation otherwise. Thanks @phil9909! Was not expecting it to be this brief. 😆

physical/raft/snapshot/snapshot.go
phil9909: Note: The only time this function will ever return an error is, if the compression level passed is invalid.

The error would then be: [`failed to write snapshot file: gzip: invalid compression level: 42`](https://goplay.tools/snippet/OdtW5k0YsIX), which IMHO would be correct: We could not write the snapshot file, because the compression level was invalid. 

Anyway, I changed it to a more broad `failed to create snapshot file`

wslabosz-reply: sure, if it's a user-facing error than it makes sense to prepend `failed to x snapshot`, not really an issue as it *cannot* ever err if we are passing const enum as a compression level

wslabosz-reply (approved):

PR #3065, opened 2026-05-11
ci: add workflow to validate publiccode.yml

bfabio:
## Description

Adds a GA workflow that validates `publiccode.yml` according to [the spec](https://github.com/publiccodeyml/publiccode.yml) on PRs and pushes that touch the file using the official parser.

## Rationale

`publiccode.yml` is currently valid. let's keep it that way.

## Acknowledgements

       also allowed.

### DIFF
--- a/.github/workflows/publiccodeyml-check.yml
+++ b/.github/workflows/publiccodeyml-check.yml
@@ -1,0 +1,27 @@
+name: Validate publiccode.yml
+
+on:
+  push:
+    paths:
+      - "publiccode.yml"
+      - ".github/workflows/publiccodeyml-check.yml"
+  pull_request:
+    paths:
+      - "publiccode.yml"
+      - ".github/workflows/publiccodeyml-check.yml"
+
+permissions: {}
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    name: publiccode.yml validation
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+
+      - uses: italia/publiccode-parser-action@21086c73ec0563e14c6748787efa1b34b025ad8c # v1
+        with:
+          publiccode: "publiccode.yml"
+          no-network: true
--- a/.github/workflows/publiccodeyml-check.yml
+++ b/.github/workflows/publiccodeyml-check.yml
@@ -19,9 +19,9 @@ jobs:
       contents: read
     name: publiccode.yml validation
     steps:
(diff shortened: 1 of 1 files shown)


cipherboy (approved):
@karras Do you or @phil9909 want to take a pass before I merge?

.github/workflows/publiccodeyml-check.yml
```diff
@@ -0,0 +1,27 @@
...
+permissions: {}
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    name: publiccode.yml validation
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
```
satoqz: Please keep the versions used consistent with other workflows already in the repo if applicable, and set the version comment to be the exact semver version. Thanks!

.github/workflows/publiccodeyml-check.yml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+    paths:
+      - "publiccode.yml"
+      - ".github/workflows/publiccodeyml-check.yml"
+
+permissions: {}
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
```
satoqz, line 17: Can this use ubuntu-minimal?

.github/workflows/publiccodeyml-check.yml
bfabio: Done :+1:

.github/workflows/publiccodeyml-check.yml
bfabio: `ubuntu-slim`? That's a good point, it should work. Done :+1:

phil9909 (approved):

.github/workflows/publiccodeyml-check.yml
satoqz: Ah, missed that it needs Docker so ubuntu-slim won't cut it after all.

bfabio: @satoqz yeah, that's unfortunate. Force pushed back to `ubuntu-latest`

wslabosz-reply (approved):

.github/workflows/publiccodeyml-check.yml
cipherboy: @satoqz Makes me wonder if we want to also sync dependencies globally; it'd be great to align plugins/g-k-w with the main repo when we can to minimize total footprint.

PR #3072, opened 2026-05-12
Bump to Go 1.26.3

cipherboy:
## Description

This doesn't adjust the minimum build requirements (still at Go 1.25), but will set the stage for that later. This just changes what the project ships with. 

## Acknowledgements

       also allowed.


phil9909 (approved):

phil9909 (changes requested):
`FuzzNamespaceName/seed#1` consistently fails with Go 1.26:

Run with go 1.26.0:

```console
$ podman run --rm -v $PWD:$PWD -w $PWD golang:1.26.0 \
  go test -v ./http -run 'FuzzNamespaceName/seed#1$'
# ...
=== RUN   FuzzNamespaceName/seed#1
    sys_namespaces_test.go:98: creating namespace ".." (2e2e)
    sys_namespaces_test.go:148: Expected status 404, got 400. Body:

        {"errors":["cannot write to a path ending in '/'"]}
# ...
--- FAIL: FuzzNamespaceName (0.01s)
    --- FAIL: FuzzNamespaceName/seed#1 (0.00s)
FAIL
FAIL	github.com/openbao/openbao/http	0.034s
FAIL
```


Run with go 1.25.10:

```console
$ podman run --rm -v $PWD:$PWD -w $PWD golang:1.25.10 \
  go test -v ./http -run 'FuzzNamespaceName/seed#1$'
# ...
=== RUN   FuzzNamespaceName/seed#1
    sys_namespaces_test.go:98: creating namespace ".." (2e2e)
# ...
--- PASS: FuzzNamespaceName (0.02s)
    --- PASS: FuzzNamespaceName/seed#1 (0.00s)
PASS
ok  	github.com/openbao/openbao/http	0.054s
```

Root cause:

> [ServeMux](https://go.dev/pkg/net/http#ServeMux) trailing slash redirects now use HTTP status 307 (Temporary Redirect) instead of 301 (Moved Permanently).
> https://go.dev/doc/go1.26

ServeMux will redirect requests to `/foo/bar` to `/foo/bar/`, if a handler for the later exists and the request would otherwise 404. https://pkg.go.dev/net/http#hdr-Trailing_slash_redirection-ServeMux

`/v1/sys/namespaces/..` gets simplified to `/v1/sys` and redirected to `/v1/sys/`, which will be followed by the `http.Client` used in the failing test.

Now there is a behavior difference when following 307 and 301 redirects (implemented in [`redirectBehavior`](https://cs.opensource.google/go/go/+/refs/tags/go1.26.3:src/net/http/client.go;l=510-540) as specified in the HTTP RFCs):
- 301 will discard the original body and use `GET`
- 307 will reuse the body and use the same method (`PUT` in our case)

For our test this means:
- With go 1.25 we are sending a `GET` request to `/v1/sys/` -> 404
- With go 1.26 we are sending a `PUT` request to `/v1/sys/` -> [400](https://github.com/openbao/openbao/blob/131c6966af4dfb4e1906703436eecdb8f2a3e9df/vault/request_handling.go#L824-L829)

Fixing this would be simple, just update the expected status in the test, but this raises two questions to me:
- Do we expect this to break any users? Probably not, it's unlikely that someone will rely on their PUT/POST request being "downgraded" to a GET. Changing from 307 to 301 would more likely break someone.
- Should we add this to the changelog?

cipherboy:
@phil9909 said:

> Fixing this would be simple, just update the expected status in the test

I think the actual fix we want here is not to use `..` as a _cleaned_ path segment, but as a verbatim value to `sys/namespaces`. I think we can only do that with encoding it as `%2E%2E`, which then double-encodes it right now as we don't have an unencoded version. So I'm inclined to remove `..` from fuzzing and potentially assert that the path has a prefix of `sys/namespaces/` in it.

Thoughts?

---

That said, I think it'd be more unexpected to receive a POST->GET request redirect than to preserve POST->POST. If you're relying on it, it seems more like a quirk than actual intent, and with granular ACL policies, this should prevent most accidental usages of that.

Definitely worth calling out in a changelog entry, though.

JanMa (dismissed):

phil9909:
> I think the actual fix we want here is not to use `..` as a _cleaned_ path segment, but as a verbatim value to `sys/namespaces`. I think we can only do that with encoding it as `%2E%2E`, which then double-encodes it right now as we don't have an unencoded version. So I'm inclined to remove `..` from fuzzing and potentially assert that the path has a prefix of `sys/namespaces/` in it.
> 
> Thoughts?

Hmm, the test found something (not a bug, but at least a behavior change), so I wouldn't want to remove it. But we could move it from the fuzz test into something like `TestRedirect`, but we can probably skip the `..` part and straight up hit `/v1/sys` (even though the path cleaning seems to happen server side).

I'd be open to add `%2E%2E` to the fuzzer corpus, seems like an interesting case.

cipherboy:
> Error: The process '/usr/bin/git' failed with exit code 128

Thanks, GitHub. :laughing:

cipherboy:
@phil9909 This is now updated. Creating a proper test for this behavior, if we wish to define it, is probably out of scope for the dependency update.

> Hmm, the test found something (not a bug, but at least a behavior change), so I wouldn't want to remove it. 

Test was faulty, though. :shrug: It wasn't fuzzing namespace creation, like was listed on the tin.

changelog/3072.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:change
+core: net/http.ServeMux in Go 1.26.x now uses 307 redirect instead of 301, causing some POST/PUT operations to fail with a 400 instead of 404 if they redirect to a path with a trailing slash. See also: https://go.dev/doc/go1.26.
```
DrDaveD: The first part of this sentence implies that net/http.ServeMux in Go 1.26.x always uses a redirect, but really it's only when used with a trailing slash.  That's mentioned at the end of the sentence, but it's confusing if you don't already know what is meant.

Maybe this would be better:

> core: net/http.ServeMux in Go 1.26.x now uses 307 redirect instead of 301 when given a trailing slash.  This causes some POST/PUT operations to fail with a 400 instead of 404 if they redirect to a path with a trailing slash.  See also: https://go.dev/doc/go1.26.

cipherboy: I guess, the correct version would be:

> core: net/http.ServeMux in Go 1.26.x now uses 307 redirect instead of 301 when given a bare path which doesn't exist in the multiplexer but for which a path with a trailing slash exists. This causes some POST/PUT operations to fail with a 400 instead of 404 if they redirect to a path with a trailing slash. See also: https://go.dev/doc/go1.26.

(revised slightly in the pushed commit)

phil9909 (approved):

satoqz:
I believe this also needs to move golangci-lint (as installed in CI, [not _necessarily_ as installed in `tools/`](https://github.com/openbao/openbao/pull/3080#issuecomment-4441317942)) to a later version that was built with Go 1.26. So we can either build it ourselves ad-hoc via `go install` or download a newer official binary.

satoqz:
Updated the `golangci-lint` version as used in CI. I haven't synced the new version (v2.12, latest) against `tools/go.mod` because `make sync-deps` turns this into a disaster and only lets us upgrade to v2.7 unless I mess with the main `go.mod`. To be fair, we should probably just exclude `tools/go.mod` from `make sync-deps`.

Also see https://github.com/openbao/openbao/pull/3080#issuecomment-4441317942 -- we probably shouldn't be managing `golangci-lint` via `tools/` anyway. I also tried installing via `go install` a try in CI, but it takes 50 seconds to compile, which doesn't seem acceptable.

satoqz (approved):
I'll merge this since the check now passes, let's more closely discuss tool installation further in #3080 and others.

PR #3079, opened 2026-05-13
Run gofumpt@latest

cipherboy:
```
$ go run mvdan.cc/gofumpt@latest -l -w .
```

I think this explains a lot of the diff I've been seeing in @wslabosz-reply's PRs.

## Rationale

Resolves: #

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3080, opened 2026-05-13
Tell dependabot to update tools as well

cipherboy:
This addresses the root cause of the drift in #3079 -- gofumpt as used by golangci-lint is at an older version as Dependabot was not updating it.

## Acknowledgements

       also allowed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,7 +26,9 @@ updates:
       - github-actions
 
   - package-ecosystem: "gomod"
-    directory: "/"
+    directories:
+     - "/"
+     - "/tools/"
     schedule:
       interval: "weekly"
     cooldown:


satoqz:
@cipherboy hmm, if Dependabot then goes and bumps golangci-lint, it won't sync with the version used in https://github.com/openbao/openbao/blob/main/.github/workflows/code-checker.yml yet, which is separate.

Moreover, golangci-lint docs don't encourage managing the tool via Go at all: https://golangci-lint.run/docs/welcome/install/local/#install-from-sources. Personally, tbh, I'm also more likely to just install whatever the latest distro package is for my system instead of letting one project take over my `$GOPATH/bin` with its intended versions, always using `go run` explicitly, or keeping a per-project `$GOPATH`.

I've ranted about this a long time ago already, but I still believe something like https://github.com/jdx/mise would be appropriate to cleanly manage all the tooling (including non-Go tooling like protoc and node) that we have.

---

That said I'm happy in the short term if we find a way to have CI use the version indicated in `tools/`, probably via `install-mode: none` via https://github.com/golangci/golangci-lint-action#install-mode and adding a `make ci-bootstrap` or so.

JanMa (approved):

cipherboy:
@satoqz I've liked that other than a little bit of make glue for people to locate all the helpful commands, we've not strictly needed any outside tooling than Go for most things. Proper pinning in go tools mod file would be good for all go run invocations, but I'd have liked if we could avoid another package manager in 2026.

I'm happy introducing it as optional, but as I've had discussions with @wslabosz-reply over the golint-ci vs gofumpt in the git commit hook, it'd be nice to be fast and portable and golangci-lint just isn't (though I agree it's nice to run  &c).

satoqz (approved):
LGTM, will do follow ups regarding any scripts that don't use pinned versions.

PR #3082, opened 2026-05-13
Add per-ns encryption key and seal status endpoints

wslabosz-reply:
## Description
- added `"namespaces/(?P<path>.+)/seal-status"` endpoint handler (seal status read)
- adjusted `/key-status` endpoint code, references and documentation, highlighting the namespace specification attribute
- removed `/key-status` endpoint from the restricted sys apis list 

## Rationale
Seal status added on `/namespaces` prefix path due to required "offline" nature of the endpoint (we have to allow read of the status when namespace is sealed). Key-status read is only available when barrier is unsealed, hence existing flow is adjusted to allow for namespace specification.

## Acknowledgements

       also allowed.

### DIFF
--- a/command/operator_key_status.go
+++ b/command/operator_key_status.go
@@ -21,15 +21,15 @@ type OperatorKeyStatusCommand struct {
 }
 
 func (c *OperatorKeyStatusCommand) Synopsis() string {
-	return "Provides information about the active encryption key"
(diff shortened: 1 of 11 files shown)


cipherboy (approved):
Removing the `$` sentinel should be fine. Thanks @wslabosz-reply!


I'm happy with the changes to `key-status` being under the main namespace sealing changelog.

PR #3083, opened 2026-05-13
Handle invalidation of LoginMFA keys

### LINKED ISSUE #2879: Login MFA configs are never invalidated on standbys -> permanently stale reads under horizontal read scalability
## Summary

Under the horizontal read scalability feature introduced in v2.5.0, standby nodes serve `LIST` and `GET` requests for `identity/mfa/method/*` (and `identity/mfa/login-enforcement/*`) from their in-memory `go-memdb`. That memdb is populated exactly once, at `postUnseal`, by `Core.loadLoginMFAConfigs`. No invalidation callback exists for the `login-mfa/*` storage prefix, so MFA configs created or modified after a standby has unsealed are never visible on that standby. The staleness does not converge.

This is not an "eventual consistency" window - one of the standbys in our test cluster still served an empty LIST 17 hours after the method was created.

## Environment

- OpenBao 2.5.2, integrated Raft storage
- 3-node HA cluster, all unsealed, one leader + two standbys
- No `disable_standby_reads` set

## Reproducer

On a fresh cluster, write a TOTP method on the active node:

```
bao write identity/mfa/method/totp issuer=Test period=30 algorithm=SHA256 digits=6 key_size=20
```

Then query each node directly (not through a load balancer):

```
curl -H "X-Vault-Token: $TOKEN" -X LIST https://<node>:8200/v1/identity/mfa/method/totp
```

- Active node: `200` with the method ID in `data.keys`.
- Standby nodes: `404` with body `{"errors":[]}`.

Behind a round-robin load balancer this manifests as intermittent `200` vs `404` for the same logical request.

## Confirmation that it is invalidation-related (not replication lag)

With two standbys initially both stale, sealing and re-unsealing only one of them causes `postUnseal -> loadLoginMFAConfigs` to re-run on that standby:

| Node      | Role    | LIST — before      | LIST — after unseal of standby A |
|-----------|---------|--------------------|----------------------------------|
| active    | active  | 200 / 2 method IDs | 200 / 2 method IDs               |
| standby A | standby | 404 / empty        | **200 / 2 method IDs**           |
| standby B | standby | 404 / empty (17h)  | 404 / empty                      |

Standby A is still a standby (`standby: true`, health 429) and serves the read from its own now-populated memdb - confirming horizontal reads is what answers, and that re-running `loadLoginMFAConfigs` is sufficient to heal the state.

## Root cause

- MFA configs live in an in-memory `go-memdb` per-node (`LoginMFABackend.db`).
- Populated from barrier storage **only** by `loadLoginMFAConfigs` at `postUnseal` (`vault/core.go` - called from
  `readonlyUnsealStrategy.unsealShared`, so it does run on standbys).
- Updated at write time by `MemDBUpsertMFAConfig` in `handleMFAMethodUpdateCommon` (`vault/identity/mfa.go`). Writes forward to the active node, so only the active node's memdb is updated that way.
- The invalidation dispatcher (`vault/core_cache_invalidate.go`) routes storage changes to backends via `router.Invalidate`. For identity-adjacent keys this reaches `IdentityStore.Invalidate` (`vault/identity/store.go`), whose switch covers entity buckets, group buckets, OIDC tokens, OIDC clients, and local aliases - but has no case for `login-mfa/method/*` or `login-mfa/enforcement/*`. The key falls through the switch with no action.
- `vault/login_mfa.go` defines no `Invalidate` method.

Before v2.5.0, standbys forwarded all reads to the active node, so the stale memdb was dormant. The horizontal read scalability feature exposes the gap.

Consistent with this, the test suite already sets `DisableStandbyReads: true` for tests that touch identity/MFA/userlockouts state (`vault/external_tests/identity/userlockouts_test.go`, `vault/external_tests/response/allowed_response_headers_test.go`, `vault/core_test.go`, others).

## Proposed fix

Add invalidation handlers for the two prefixes. The reloaders exist today in `LoginMFABackend` (`loadMFAMethodConfigs`, `loadMFAEnforcementConfigs`) but operate on full namespaces; they can be restructured or complemented by per-key helpers. Deletions need to remove the corresponding memdb rows.

Wiring can go either through `IdentityStore.Invalidate` (adding `strings.HasPrefix(key, "login-mfa/")` cases) or by giving `LoginMFABackend` its own invalidation hook and registering it in the dispatch switch alongside the mount-related cases.

Happy to prepare a PR.

Two potential workarounds: either disable standby reads (which would be a real shame), or route `/v1/identity/*` reads through a leader only path at the load balancer.

## Related

- #1528 - horizontal read scalability RFC; the invalidation mechanism is discussed but login-mfa was not explicitly covered.
- #2546 - auth/mount tune write-then-read staleness on standby. Structurally different: that case has a working invalidation callback (`transactionalMountInvalidation` -> `reloadMount`) with a timing gap, matching the "eventual consistency" framing. This case has no callback at all, so it does not converge.

cipherboy:
## Description

Notably, LoginMFA did not currently contain code for invalidation in OpenBao. This meant standbys had stale entries on API responses. As token-based login is only processed by the active node, this wouldn't have had much effect otherwise.

## Rationale

Resolves: #2879

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/login_mfa.go
+++ b/vault/login_mfa.go
@@ -174,6 +174,32 @@ func (b *LoginMFABackend) ResetLoginMFAMemDB() error {
 	return nil
 }
 
+func (b *LoginMFABackend) invalidate(ctx context.Context, ns *namespace.Namespace, path string) error {
+	switch {
+	case strings.HasPrefix(path, barrier.SystemBarrierPrefix+loginMFAConfigPrefix):
+		key, _ := strings.CutPrefix(path, barrier.SystemBarrierPrefix+loginMFAConfigPrefix)
+		barrierView := NamespaceScopedView(b.Core.barrier, ns).SubView(barrier.SystemBarrierPrefix).SubView(loginMFAConfigPrefix)
+
+		return b.loadMFAMethodConfig(ctx, ns, barrierView, key)
+	case strings.HasPrefix(path, barrier.SystemBarrierPrefix+mfaLoginEnforcementPrefix):
+		key, _ := strings.CutPrefix(path, barrier.SystemBarrierPrefix+mfaLoginEnforcementPrefix)
+		barrierView := NamespaceScopedView(b.Core.barrier, ns).SubView(barrier.SystemBarrierPrefix).SubView(mfaLoginEnforcementPrefix)
+
+		mConfig, err := b.loadMFAEnforcementConfig(ctx, ns, barrierView, key)
+		if err != nil {
+			return err
+		}
+
+		if mConfig == nil {
+			return nil
+		}
+
+		return b.loginMFAMethodExistenceCheck(mConfig)
+	default:
+		return fmt.Errorf("unknown path to invalidate: %v", path)
+	}
+}
+
 // loadMFAMethodConfigs loads MFA method configs for login MFA
 func (b *LoginMFABackend) loadMFAMethodConfigs(ctx context.Context, ns *namespace.Namespace) error {
 	b.mfaLogger.Trace("loading login MFA configurations")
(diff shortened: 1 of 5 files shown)


vault/login_mfa.go
```diff
@@ -185,27 +213,35 @@ func (b *LoginMFABackend) loadMFAMethodConfigs(ctx context.Context, ns *namespac
...
 
 	for _, key := range existing {
-		b.mfaLogger.Trace("loading method", "method", key)
-
-		// Read the config from storage
-		mConfig, err := b.getMFAConfig(ctx, key, barrierView)
-		if err != nil {
-			return err
+		if err := b.loadMFAMethodConfig(ctx, ns, barrierView, key); err != nil {
+			return fmt.Errorf("failed lodaing MFA method config %v: %w", key, err)
```
wslabosz-reply:
```suggestion
			return fmt.Errorf("failed to load MFA method config %v: %w", key, err)
```

vault/login_mfa.go
```diff
@@ -174,6 +174,34 @@ func (b *LoginMFABackend) ResetLoginMFAMemDB() error {
...
+		key, _ := strings.CutPrefix(path, barrier.SystemBarrierPrefix+loginMFAConfigPrefix)
+		barrierView := NamespaceScopedView(b.Core.barrier, ns).SubView(barrier.SystemBarrierPrefix).SubView(loginMFAConfigPrefix)
+
+		return b.loadMFAMethodConfig(ctx, ns, barrierView, key)
+	case strings.HasPrefix(path, barrier.SystemBarrierPrefix+mfaLoginEnforcementPrefix):
+		key, _ := strings.CutPrefix(path, barrier.SystemBarrierPrefix+mfaLoginEnforcementPrefix)
+		barrierView := NamespaceScopedView(b.Core.barrier, ns).SubView(barrier.SystemBarrierPrefix).SubView(mfaLoginEnforcementPrefix)
+
+		// We ignore the resulting enforcement configuration: we aren't
+		// guaranteed an order to invalidate in, so we can't immediately
```
wslabosz-reply: "(...) so we can't immediately..." invalidate? reload?

wslabosz-reply (approved):
LGTM

vault/login_mfa.go
cipherboy: @wslabosz-reply I've actually ignored the comment here. I'm probably wrong on it anyways, enforcement config is different from method config (separate endpoints) so you'd need to be really fast to hit invalidation issues.

cipherboy: Will adjust tho.

cipherboy: I dropped the comment as it doesn't make sense; we still want to reload core (passthrough error) if the enforcement doesn't have corresponding methods behind it.

PR #3106, opened 2026-05-15
[codex] Validate raft promote and demote server IDs

### LINKED ISSUE #3085: raft demote command ignores questionmark

Reproduce:

```
$ bao operator raft demote ?
Server demoted successfully!
```

```
$ bao operator raft demote ?

Code: 500. Errors:

* 1 error occurred:
```

OpenBao Version:
Version                 2.5.3

Host:
Ubuntu 24.04

MukundaKatta:
## Summary

- reject raft promote/demote requests for server IDs that are absent from Autopilot state
- prevent arbitrary IDs such as `?` from being persisted as permanent non-voters
- add unit coverage for nil, missing, and present Autopilot state lookups

Fixes #3085

## Validation

- go test ./physical/raft -run TestAutopilotStateHasServer -count=1
- git diff --check

### DIFF
--- a/physical/raft/raft.go
+++ b/physical/raft/raft.go
@@ -1277,6 +1277,10 @@ func (b *RaftBackend) PromotePeer(ctx context.Context, peerID string) error {
 		return ErrRaftAutopilotNotInitialized
 	}
 
+	if !autopilotStateHasServer(b.autopilot.GetState(), peerID) {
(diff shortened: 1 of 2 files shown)


phil9909:
@MukundaKatta please read [Contributing](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md) guide and perhaps [AGENTS.md](https://github.com/openbao/openbao/blob/main/AGENTS.md) also.

PR #3118, opened 2026-05-15
physical/raft: validate autopilot peer IDs before promote/demote

### LINKED ISSUE #3085: raft demote command ignores questionmark

MukundaKatta:
## Summary

- validate that the requested raft peer exists in autopilot state before promote or demote changes the permanent non-voter set
- return a clear not found error for unknown peer IDs
- add coverage for the autopilot state lookup helper

Fixes #3085.

## Validation

- `go test ./physical/raft -run '^TestAutopilotStateHasServer$' -count=1`\n- `git diff --check origin/main...HEAD`

### DIFF
--- a/physical/raft/raft.go
+++ b/physical/raft/raft.go
@@ -1277,6 +1277,10 @@ func (b *RaftBackend) PromotePeer(ctx context.Context, peerID string) error {
 		return ErrRaftAutopilotNotInitialized
 	}
 
+	if !autopilotStateHasServer(b.autopilot.GetState(), peerID) {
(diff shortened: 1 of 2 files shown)


cipherboy:
@MukundaKatta Please see @phil9909's comment on your earlier PR.

PR #3123, opened 2026-05-15
Add script to backport dependencies

cipherboy:
## Description

This hack allows us to check the govulndb for dependencies which are newer in main than in a release branch tree, updating the dependency to the version in main only if the package has seen vulnerabilities.

Needs curl, jq.

## Rationale

Last release, we had discussions that `bump-critical` should not be used as @dependabot is the preferred tool. However, dependabot doesn't currently execute on release branches (nor would we want to backport everything is part of the argument against `bump-critical`), thus this attempts to form a compromise.

## Acknowledgements

       also allowed.

### DIFF
--- a/scripts/backport-deps.sh
+++ b/scripts/backport-deps.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+release_dir="$1"
+modules=$(find . -name go.mod)
+
+curl -sSL https://vuln.go.dev/index/modules.json > /tmp/vuln-go-dev-modules.json
+
+function clean_version() { echo "$@" | sed 's/^v//g' | sed 's/-.*$//g' | sed 's/+.*$//g'; }
+
+# Helpers from https://gist.github.com/jonlabelle/6691d740f404b9736116c22195a8d706
+function version_gt() { test "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" != "$1"; }
+function version_ge() { test "$(echo "$@" | tr " " "\n" | sort -rV | head -n 1)" == "$1"; }
+
+while read -r go_mod; do
+	grep '^	' "$go_mod" | while read -r line; do
+		pkg="$(awk '{print $1}' <<< "$line")"
+		current_version="$(awk '{print $2}' <<< "$line")"
+		update_version="$(awk '{print $2}' <<< "$line")"
+		release_version="$(grep "^	${pkg} " "$release_dir/$go_mod" | awk '{print $2}')"
+
+	  if [ "$release_version" == "" ]; then
+		  continue
+	  fi
+
+	  if [ "$update_version" == "$release_version" ]; then
+		  continue
+	  fi
+
+	  update_version="$(clean_version "$update_version")"
+	  release_version="$(clean_version "$release_version")"
+
+	  if version_gt "$release_version" "$update_version"; then
+		  if ! [[  "$pkg" == "github.com/openbao/openbao"* ]]; then
+			  echo "Dependency $pkg is newer on release branch than main"
+		  fi
+		  continue
+	  fi
+
+	  entry="$(jq '.[] | select(.path == "'"$pkg"'")' < /tmp/vuln-go-dev-modules.json)"
+	  vuln_count="$(jq '.vulns | length' <<< "$entry")"
+
+	  if (( vuln_count == 0 )); then
+		  continue
+	  fi
+
+	  echo "Checking vulnerabilities for $pkg..."
+
+	  update=false
+
+	  if [ "$update_version" == "0.0.0" ]; then
+		  update=true
+	  fi
+
+	  for vuln_index in $(seq 0 "$(( vuln_count - 1 ))"); do
+		  vuln_info="$(jq '.vulns['"$vuln_index"']' <<< "$entry")"
+		  vuln_id="$(jq -r '.id' <<< "$vuln_info")"
+
+		  if ! [[ "$vuln_info" == *fixed* ]]; then
+			  echo "-> Skipping $vuln_id as it is not fixed"
+			  continue
+		  fi
+
+		  fixed_mod_version="$(jq -r '.fixed' <<< "$vuln_info")"
+		  fixed_version="$(clean_version "$fixed_mod_version")"
+
+		  if version_gt "$fixed_version" "$release_version" && version_ge "$update_version" "$fixed_version"; then
+			  update="true"
+			  break
+		  elif ! version_ge "$update_version" "$fixed_version"; then
+			  echo "-> !!! Update $pkg on main to $fixed_mod_version first !!!"
+		  fi
+	  done
+
+	  if [ "$update" == "true" ]; then
+		  echo "-> Updating on release branch to $current_version"
+		  (
+			  cd "$release_dir" || exit 1
+			  dir="$(dirname "$go_mod")"
+			  cd "$dir" || exit 1
+
+			  echo "[$PWD] " go get "$pkg@$current_version"
+			  go get "$pkg@$current_version"
+		  ) || exit 1
+	  fi
+	done
+done <<< "$modules"
+
+(
+	cd "$release_dir" || exit 1
+	make sync-deps
+) || exit 1
--- a/scripts/backport-deps.sh
+++ b/scripts/backport-deps.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+
+if [[ -z "$1" ]]; then
+	echo "Usage: $0 /path/to/release/branch/checkout" 1>&2
+	exit 1
+fi
+
+release_dir="$1"
+modules=$(find . -name go.mod)
+
+MODULE_JSON="$(curl -sSL https://vuln.go.dev/index/modules.json)"
+
+function clean_version() { echo "$@" | sed 's/^v//g' | sed 's/-.*$//g' | sed 's/+.*$//g'; }
+
+# Helpers from https://gist.github.com/jonlabelle/6691d740f404b9736116c22195a8d706
+function version_gt() { test "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" != "$1"; }
+function version_ge() { test "$(echo "$@" | tr " " "\n" | sort -rV | head -n 1)" == "$1"; }
+
+while read -r go_mod; do
+	modules_count="$(go mod edit -json "$go_mod" | jq -r '.Require | length')"
+	for module_index in $(seq 0 "$(( modules_count - 1 ))"); do
+		pkg="$(go mod edit -json "$go_mod" | jq -r '.Require['"$module_index"'].Path')"
+		current_version="$(go mod edit -json "$go_mod" | jq -r '.Require['"$module_index"'].Version')"
+		update_version="$current_version"
+		release_version="$(go mod edit -json "$release_dir/$go_mod" | jq -r '.Require[] | select(.Path == "'"$pkg"'") | .Version')"
+
+	  if [[ "$release_version" == "" ]]; then
+		  continue
+	  fi
+
+	  if [[ "$update_version" == "$release_version" ]]; then
+		  continue
+	  fi
+
+	  update_version="$(clean_version "$update_version")"
+	  release_version="$(clean_version "$release_version")"
+
+	  if version_gt "$release_version" "$update_version"; then
+		  if ! [[  "$pkg" == "github.com/openbao/openbao"* ]]; then
+			  echo "Dependency $pkg is newer on release branch than main"
+		  fi
+		  continue
+	  fi
+
+	  entry="$(jq -r '.[] | select(.path == "'"$pkg"'")' <<< "$MODULE_JSON")"
+	  vuln_count="$(jq -r '.vulns | length' <<< "$entry")"
+
+	  if (( vuln_count == 0 )); then
+		  continue
+	  fi
+
+	  echo "Checking vulnerabilities for $pkg..."
+
+	  update=false
+
+	  if [[ "$update_version" == "0.0.0" ]]; then
+		  update=true
+	  fi
+
+	  for vuln_index in $(seq 0 "$(( vuln_count - 1 ))"); do
+		  vuln_info="$(jq -r '.vulns['"$vuln_index"']' <<< "$entry")"
+		  vuln_id="$(jq -r '.id' <<< "$vuln_info")"
+
+		  if ! [[ "$vuln_info" == *fixed* ]]; then
+			  echo "-> Skipping $vuln_id as it is not fixed"
+			  continue
+		  fi
+
+		  fixed_mod_version="$(jq -r '.fixed' <<< "$vuln_info")"
+		  fixed_version="$(clean_version "$fixed_mod_version")"
+
+		  if version_gt "$fixed_version" "$release_version" && version_ge "$update_version" "$fixed_version"; then
+			  update="true"
+			  break
+		  elif ! version_ge "$update_version" "$fixed_version"; then
+			  echo "-> !!! Update $pkg on main to $fixed_mod_version first !!!"
+		  fi
+	  done
+
+	  if [[ "$update" == "true" ]]; then
+		  echo "-> Updating on release branch to $current_version"
+		  (
+			  cd "$release_dir" || exit 1
+			  dir="$(dirname "$go_mod")"
+			  cd "$dir" || exit 1
+
+			  echo "[$PWD] " go get "$pkg@$current_version"
+			  go get "$pkg@$current_version"
+		  ) || exit 1
+	  fi
+	done
+done <<< "$modules"
+
+(
+	cd "$release_dir" || exit 1
+	make sync-deps
+) || exit 1
--- a/Makefile
+++ b/Makefile
@@ -303,27 +303,6 @@ sync-deps:
 ci-sync-deps: sync-deps
 	git diff --quiet || (echo -e "\n\nModified files:" && git status --short && echo -e "\n\nRun 'make sync-deps' locally and commit the changes.\n" && exit 1)
 
-.PHONY: bump-critical
-bump-critical:
-	go get github.com/golang-jwt/jwt/v4@latest
-	go get github.com/golang-jwt/jwt/v5@latest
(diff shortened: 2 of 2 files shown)


scripts/backport-deps.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+release_dir="$1"
+modules=$(find . -name go.mod)
+
+curl -sSL https://vuln.go.dev/index/modules.json > /tmp/vuln-go-dev-modules.json
```
JanMa:
```suggestion
curl -sSL https://vuln.go.dev/index/modules.json -o /tmp/vuln-go-dev-modules.json
```

scripts/backport-deps.sh
```diff
@@ -0,0 +1,91 @@
...
+
+function clean_version() { echo "$@" | sed 's/^v//g' | sed 's/-.*$//g' | sed 's/+.*$//g'; }
+
+# Helpers from https://gist.github.com/jonlabelle/6691d740f404b9736116c22195a8d706
+function version_gt() { test "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" != "$1"; }
+function version_ge() { test "$(echo "$@" | tr " " "\n" | sort -rV | head -n 1)" == "$1"; }
+
+while read -r go_mod; do
+	grep '^	' "$go_mod" | while read -r line; do
```
JanMa: Parsing a `go.mod` file this way is pretty unreliable. I suggest you use `go mod edit -json $go_mod` to json format it and then loop over the result with jq.

JanMa (dismissed):
Two small remarks, otherwise LGTM. I wonder if using the same version for a package on `main` in the release branch is always the right choice, or if we instead should try to only update the vulnerable version in the release branch to the next fixed patch version? 🤔

scripts/backport-deps.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
```
satoqz: One day, this line will cause the collapse of civilization. Mark my words.

scripts/backport-deps.sh
satoqz: Do we need to save this to the file system at all?

scripts/backport-deps.sh
```diff
@@ -0,0 +1,91 @@
...
+
+	  if (( vuln_count == 0 )); then
+		  continue
+	  fi
+
+	  echo "Checking vulnerabilities for $pkg..."
+
+	  update=false
+
+	  if [ "$update_version" == "0.0.0" ]; then
```
satoqz: Nit: want to use `[[` consistently since we're already in bash?

satoqz:
> I wonder if using the same version for a package on `main` in the release branch is always the right choice, or if we instead should try to only update the vulnerable version in the release branch to the next fixed patch version? 🤔

I feel like this will go either way depending on the particular dependency, e.g., I don't have any concerns bumping `golang.org/x/...` deps all the way to latest, but we'd want to second-guess any dependencies that are less stable. So perhaps it's best to keep this as simple as can be and manually adjust when a major or minor version bump sticks out.

scripts/backport-deps.sh
cipherboy: Was not aware of the JSON version, will revise.

scripts/backport-deps.sh
cipherboy: Damned distro maintainers and their belief in a closed system. :laughing:

cipherboy:
@satoqz said:

> @JanMa said:
> 
> > I wonder if using the same version for a package on `main` in the release branch is always the right choice, or if we instead should try to only update the vulnerable version in the release branch to the next fixed patch version? 🤔
> 
> I feel like this will go either way depending on the particular dependency, e.g., I don't have any concerns bumping `golang.org/x/...` deps all the way to latest, but we'd want to second-guess any dependencies that are less stable. So perhaps it's best to keep this as simple as can be and manually adjust when a major or minor version bump sticks out.

I think of the two evils, `main` is probably the best as it'll have been tested more. We risk having tested on `$CURRENT` and `$MAIN` but not `$CURRENT < $FIXED < $MAIN`, and having `$FIXED` introduce bugs that were fixed in `$MAIN` -- as we never tested `$FIXED` explicitly. Presumably, if `$MAIN` required changes, we'd have caught that recently and be more likely to remember it. 

That's my gut feel at any rate. I do concur though, if we see something particularly egregious, we can adjust as necessary. Keep in mind, we're only bumping dependencies if they're listed as vulnerable at the version currently in the release branch.

cipherboy:
@JanMa @satoqz This has been revised and pushed. I was able to backport dependencies to 2.4.x. The actual `go.mod` diff looked like:


```diff
 replace github.com/openbao/openbao/api/v2 => ./api
 
@@ -36,7 +36,7 @@ require (
        github.com/fatih/structs v1.1.0
        github.com/gammazero/workerpool v1.1.3
        github.com/go-errors/errors v1.5.1
-       github.com/go-jose/go-jose/v4 v4.1.3
+       github.com/go-jose/go-jose/v4 v4.1.4
        github.com/go-ldap/ldap/v3 v3.4.9
        github.com/go-ldap/ldif v0.0.0-20200320164324-fd88d9b715b3
        github.com/go-sql-driver/mysql v1.9.3
@@ -83,7 +83,7 @@ require (
        github.com/hashicorp/raft-snapshot v1.0.4
        github.com/hashicorp/vault-plugin-mock v0.19.13
        github.com/influxdata/influxdb1-client v0.0.0-20200827194710-b269163b24ab
-       github.com/jackc/pgx/v5 v5.7.6
+       github.com/jackc/pgx/v5 v5.9.2
        github.com/jcmturner/gokrb5/v8 v8.4.4
        github.com/jefferai/isbadcipher v0.0.0-20190226160619-51d2077c035f
        github.com/jefferai/jsonx v1.0.1
@@ -137,23 +137,23 @@ require (
        github.com/tink-crypto/tink-go/v2 v2.5.0
        github.com/yeqown/go-qrcode/v2 v2.2.5
        go.etcd.io/bbolt v1.4.3
-       go.opentelemetry.io/otel v1.38.0
-       go.opentelemetry.io/otel/sdk v1.38.0
-       go.opentelemetry.io/otel/trace v1.38.0
+       go.opentelemetry.io/otel v1.43.0
+       go.opentelemetry.io/otel/sdk v1.43.0
+       go.opentelemetry.io/otel/trace v1.43.0
        go.uber.org/atomic v1.11.0
        go.uber.org/goleak v1.3.0
        go.uber.org/zap v1.27.0
-       golang.org/x/crypto v0.45.0
+       golang.org/x/crypto v0.50.0
        golang.org/x/exp v0.0.0-20251113190631-e25ba8c21ef6
-       golang.org/x/net v0.47.0
-       golang.org/x/oauth2 v0.33.0
-       golang.org/x/sync v0.18.0
-       golang.org/x/sys v0.38.0
-       golang.org/x/term v0.37.0
-       golang.org/x/text v0.31.0
+       golang.org/x/net v0.53.0
+       golang.org/x/oauth2 v0.36.0
+       golang.org/x/sync v0.20.0
+       golang.org/x/sys v0.44.0
+       golang.org/x/term v0.42.0
+       golang.org/x/text v0.36.0
        google.golang.org/api v0.246.0
-       google.golang.org/grpc v1.77.0
-       google.golang.org/protobuf v1.36.10
+       google.golang.org/grpc v1.81.1
+       google.golang.org/protobuf v1.36.11
        gopkg.in/jcmturner/goidentity.v3 v3.0.0
        k8s.io/api v0.33.0
        k8s.io/apimachinery v0.33.0
@@ -164,7 +164,7 @@ require (
 )
```

cipherboy:
The build failed, fwiw:

```
# github.com/docker/docker/client
../../../go/pkg/mod/github.com/docker/docker@v28.3.3+incompatible/client/client.go:466:19: undefined: sockets.DialPipe
```

Though it looks like this is because `make sync-deps` didn't work...?

```
[cipherboy@cp-x1c openbao]$ gif github.com/docker/go-connections
sdk/helper/docker/testhelpers.go:34:	"github.com/docker/go-connections/nat"
sdk/helper/stepwise/environments/docker/environment.go:31:	"github.com/docker/go-connections/nat"
sdk/go.mod:17:	github.com/docker/go-connections v0.5.0
sdk/go.sum:50:github.com/docker/go-connections v0.5.0 h1:USnMq7hx7gwdVZq1L49hLXaFtUdTADjXGp+uj1Br63c=
sdk/go.sum:51:github.com/docker/go-connections v0.5.0/go.mod h1:ov60Kzw0kKElRwhNs9UlUHAE/F9Fe6GLaXnqyDdmEXc=
go.mod:230:	github.com/docker/go-connections v0.7.0 // indirect
go.sum:226:github.com/docker/go-connections v0.7.0 h1:6SsRfJddP22WMrCkj19x9WKjEDTB+ahsdiGYf0mN39c=
```

whereas in `main`:

```
[cipherboy@cp-x1c openbao]$ gif github.com/docker/go-connections
sdk/go.mod:75:	github.com/docker/go-connections v0.7.0 // indirect
sdk/go.sum:48:github.com/docker/go-connections v0.7.0 h1:6SsRfJddP22WMrCkj19x9WKjEDTB+ahsdiGYf0mN39c=
sdk/go.sum:49:github.com/docker/go-connections v0.7.0/go.mod h1:no1qkHdjq7kLMGUXYAduOhYPSJxxvgWBh7ogVvptn3Q=
go.mod:236:	github.com/docker/go-connections v0.7.0 // indirect
go.sum:223:github.com/docker/go-connections v0.7.0 h1:6SsRfJddP22WMrCkj19x9WKjEDTB+ahsdiGYf0mN39c=
go.sum:224:github.com/docker/go-connections v0.7.0/go.mod h1:no1qkHdjq7kLMGUXYAduOhYPSJxxvgWBh7ogVvptn3Q=
```

---

Rerunning `make sync-deps` seems to have fixed it curiously, though this is at the end of the script. 

---

Though, rebuilding now also didn't fix it. Interesting!

satoqz (approved):

satoqz:
> The actual go.mod diff looked like: ...

This looks correct based on my memory of dependencies we've had to patch over the past months!

satoqz:
CI is stuck on this one fsr, let me try something...

satoqz:
That seems to have done it.

scripts/backport-deps.sh
```diff
@@ -0,0 +1,97 @@
...
+
+		  fixed_mod_version="$(jq -r '.fixed' <<< "$vuln_info")"
+		  fixed_version="$(clean_version "$fixed_mod_version")"
+
+		  if version_gt "$fixed_version" "$release_version" && version_ge "$update_version" "$fixed_version"; then
+			  update="true"
+			  break
+		  elif ! version_ge "$update_version" "$fixed_version"; then
+			  echo "-> !!! Update $pkg on main to $fixed_mod_version first !!!"
+		  fi
```
satoqz, line 77: I never hit this error backporting from this PR's branch to `release/2.5.x` even though several `golang.org/x/...` bumps required to resolve symbol vulns are missing.

JanMa (approved):
Sorry for the delay, LGTM

scripts/backport-deps.sh
cipherboy: @satoqz Hm, should we merge this anyways and wait for this to occur again? I'm entirely sure why this would happen and its harder to reproduce.

(Trying to downgrade, say, `x/net` results in:

```
[cipherboy@x1c openbao]$ go get golang.org/x/net@v0.55.0
go: downgraded github.com/google/go-containerregistry v0.21.7 => v0.21.6
go: downgraded github.com/openbao/go-kms-wrapping/wrappers/transit/v2 v2.8.0 => v2.1.0
go: downgraded github.com/openbao/openbao-template v1.0.1 => v1.0.0
go: removed github.com/openbao/openbao/api/auth/approle/v2 v2.0.0
go: removed github.com/openbao/openbao/api/auth/kubernetes/v2 v2.0.1
go: removed github.com/openbao/openbao/api/auth/userpass/v2 v2.0.0
go: removed github.com/openbao/openbao/api/v2 v2.6.0
go: removed github.com/openbao/openbao/sdk/v2 v2.6.0
go: downgraded golang.org/x/net v0.56.0 => v0.55.0
go: downgraded golang.org/x/tools v0.46.0 => v0.45.0
g[cipherboy@x1c openbao]$ go mod tidy
go: github.com/openbao/openbao-template@v1.0.0 requires
	github.com/openbao/openbao/api@v1.9.2: parsing go.mod:
	module declares its path as: github.com/hashicorp/vault/api
	        but was required as: github.com/openbao/openbao/api
```

which doesn't make sense to me why openbao-template would need to be changed.)

satoqz: I did some more testing and I guess it's fine. The problem is that not all necessary bumps on main will surface in the first pass. For example, when backporting from this branch (9b69ab74f8b7f9e7977412997888c86a87b2d5f1) to tag v2.5.3, it'll ask to bump to `golang.org/x/crypto@0.52.0`, but fails to initially warn that afterwards, we still need to bump to `golang.org/x/net@0.55.0`, which depends on the crypto bump to happen first, though. (Yes, these have a circular dependency, and it seems x/crypto always stays one x/net behind latest to accomodate this.)

satoqz: To summarize what I'm trying to say in simple words:

On the first pass, I only see:

```console
$ ./scripts/backport-deps.sh ../v2.5.3
...
-> !!! Update golang.org/x/crypto on main to 0.52.0 first !!!
...
```

If I then bump:

```console
$ go get golang.org/x/crypto@v0.52.0
```

I now see:

```console
$ ./scripts/backport-deps.sh ../v2.5.3
...
-> !!! Update golang.org/x/net on main to 0.55.0 first !!!
...
```

It would be useful to see all necessary main bumps in one.

TBH, the output is also very noisy and hard to follow. It would be good to clean that up so I can have more confidence in the script doing the right thing 🤔

satoqz: I'll give improving this a shot though if you don't mind. I also wanted to look at using the JSON output that vulncheck natively provides instead.

cipherboy: @satoqz Sure, go for it :-)

satoqz:
@cipherboy here's my attempt:

<details>

<summary>Expand script</summary>

```sh
#!/usr/bin/env bash

set -euo pipefail

release_dir="${1:-}"

if [[ -z "$release_dir" ]]; then
	echo "Usage: $0 /path/to/release/branch/checkout" 1>&2
	exit 1
fi

tools=$(realpath "$0/../../tools/go.mod")

# This is a cache of versions we've complained we can't bump to yet (because
# they haven't been upgraded on main), so we don't complain twice.
declare -A complaints

for go_mod in go.mod sdk/go.mod api/go.mod; do
  mod_dir=$(dirname "$release_dir/$go_mod")

  echo "$0: checking $go_mod"

  # When vulncheck finds distinct vulnerabilities affecting the same module,
  # it may suggest several fixed versions depending on the versions that each
  # vulnerability was fixed in respectively. Group suggested versions by module
  # and sort for the best one later on.
  bumps=$(
    env -C "$mod_dir" go tool -modfile="$tools" govulncheck -scan=package -format=json ./... \
    | jq -sc '
      map(
        .finding
        | select(.fixed_version != null)
        | { module: .trace[0].module, fixed_version }
      )
      | group_by(.module)
      | map({
          module: .[0].module,
          versions: map(.fixed_version) | unique,
      })
      | .[]'
  )

  # Get dependency versions on the main (current) branch.
  main=$(go mod edit -modfile="$go_mod" -json | jq '.Require | .[]')

  # Track if we actually carry out a bump so we can skip sync-deps if nothing
  # happened.
  bumped=false

  while read -r bump; do
    module=$(jq -r '.module' <<< "$bump")
    fixed_versions=$(jq -r '.versions[]' <<< "$bump")
    version_on_main=$(jq -r --argjson bump "$bump" 'select(.Path == $bump.module) | .Version' <<< "$main")

    # Select the latest version out of (fixed_versions..., version_on_main).
    # Notably, version_on_main may be empty, in which case we fall back to
    # bumping to the earliest fixed version.
    later_version=$(printf "${fixed_versions}\n${version_on_main}" | sort -rV | head -n 1)

    if [[ "$later_version" == "$version_on_main" ]]; then
      bumped=true
      env -C "$mod_dir" go get "${module}@${version_on_main}"
    else
      target="${module}@${later_version}"

      if [[ -v "complaints[$target]" ]]; then
        continue
      else
        complaints["$target"]=1
      fi

      echo "$0: want to bump ${go_mod} to ${target}, but main is behind at ${version_on_main} !!!"
    fi
  done <<< "$bumps"

  if [[ "$bumped" == true ]]; then
    # Run sync-deps after changing any module so updates propagate and have a
    # chance to preemptively address vulnerabilities in the next module. For
    # example, addressing a vulnerability in the main module (which is scheduled
    # first) will lead to resolving them in sdk/ and api/ in 99% of cases.
    env -C "$release_dir" ./scripts/sync-deps.sh > /dev/null
  fi
done
```

</details>

<details>

<summary>Expand sample output</summary>

```console
$ ./scripts/backport-deps.sh ../v2.5.3
./scripts/backport-deps.sh: checking go.mod
go: upgraded github.com/jackc/pgx/v5 v5.9.1 => v5.9.2
go: upgraded github.com/moby/spdystream v0.5.0 => v0.5.1
./scripts/backport-deps.sh: want to bump go.mod to golang.org/x/crypto@v0.52.0, but main is behind at v0.50.0 !!!
./scripts/backport-deps.sh: want to bump go.mod to golang.org/x/net@v0.55.0, but main is behind at v0.53.0 !!!
go: upgraded golang.org/x/sys v0.42.0 => v0.44.0
./scripts/backport-deps.sh: checking sdk/go.mod
go: upgraded github.com/jackc/pgx/v5 v5.9.1 => v5.9.2
go: upgraded golang.org/x/sys v0.42.0 => v0.44.0
./scripts/backport-deps.sh: checking api/go.mod
```

</details>

... with the goal of having `govulncheck` and `jq` do all the heavy lifting, and keeping output low-noise.

cipherboy:
@satoqz I'm happy with that, want to close mine and open yours? Seems better than my script.

satoqz:
Well, always easy to do better when someone's done the pioneer work for you :) Will open.

PR #3125, opened 2026-05-17
fix/flaky test iteration 1

space-aditya:
## Description

- **fix flaky test by eventuallying**
- **parallelize to match convention, add retry logic due to async, clean buffer between iters**
- **rely on local docker even if registry fails, version pinning to ensure images always exists**

## Rationale
- ACME tests: hardcoded mount/container/hostname names caused collisions on retry; now uses random IDs to keep runs isolated
- Version history test: replaced shared MockUi with separate stdout/stderr buffers to avoid data races
- External plugin test: parallel subtests were sharing one client and racing on SetToken; now each gets its own cloned client
- Standby test: added retry loop after token revocation since replication to standbys isn't instant

Resolves: 
Test_ACME (builtin/logical/pkiext/pkiext_binary)
TestExternalPlugin_AuthMethod (vault/external_tests/plugin)
TestVersionHistoryCommand_TableOutput (command)

## Acknowledgements

       also allowed.

### DIFF
--- a/builtin/logical/pkiext/pkiext_binary/acme_test.go
+++ b/builtin/logical/pkiext/pkiext_binary/acme_test.go
@@ -89,16 +89,20 @@ type caddyConfig struct {
 	EABKey    string
 }
 
+func generateRandomId(t *testing.T) string {
+	// Roll a random run ID for mount and hostname uniqueness.
+	runID, err := uuid.GenerateUUID()
+	require.NoError(t, err, "failed to generate a unique ID for test run")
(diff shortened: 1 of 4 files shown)


satoqz:
Sorry @space-aditya, these changes are a bit too confusing to review. Do you mind removing any unintended changes and giving more detail on the fixes so reviewers can follow along? Thanks!

space-aditya:
@satoqz update it
I've observed that a few more of tests follow the same type of pattern, hence flaky.
the solution already exists somewhere in some other test.

PR #3128, opened 2026-05-18
Add -dev-cross-namespace-identity flag

### LINKED ISSUE #3127: Add a -dev flag to enable the UnsafeCrossNamespaceIdentity option for `-dev` servers

This would enable easier testing of new features that may interact with that pattern.

There isn't really an alternative that has the same convenience. The next best option is to write configuration files, but this makes sharing example scripts that demonstrate problems more cumbersome.

luke-clifton:
I've created a corresponding issue (#3127), but, I already had this code lying around, and it's borderline trivial, so, thought I'd just raise a PR directly.

## Description

This enables the UnsafeCrossNamespaceIdentity option when starting a -dev server instance.

## Rationale

This makes testing out changes locally, significantly easier, and makes sharing scripts easier.

Resolves: #3127

## Acknowledgements

       also allowed.

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -131,6 +131,7 @@ type ServerCommand struct {
 	flagDevThreeNode       bool
 	flagDevAutoSeal        bool
 	flagDevClusterJson     string
+	flagDevCrossNsIdentity bool
 	flagTestVerifyOnly     bool
 	flagTestServerConfig   bool
 	flagExitOnCoreShutdown bool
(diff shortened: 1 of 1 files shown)


wslabosz-reply:
@luke-clifton 
Can I ask you to fix DCO? (more on how [here](https://github.com/openbao/openbao/pull/3128/checks?check_run_id=76473729962))
also, can you add a changelog (add a `3128.txt` file to the `./changelog` directory) with something along the lines:
```release-note:improvement
command/server: Added `-dev-cross-namespace-identity` flag enabling `UnsafeCrossNamespaceIdentity` for the dev server.
```

satoqz:
@wslabosz-reply see https://github.com/openbao/openbao/issues/3127#issuecomment-4543783633 -- we weren't convinced that we like adding more flags.

wslabosz-reply:
@satoqz @luke-clifton 
I see, then it would be great if you'd like to consider the suggestion on the issue.

satoqz:
I'll close this out since I think we reached consensus that `-dev-x` flags for various config-level toggles aren't a pattern we'd like to introduce. Thank you for the proposal and for raising the associated discussion though!

PR #3140, opened 2026-05-19
Correct per-ns barrier determination

wslabosz-reply:
## Description
- added `loadIdentityStoreArtifactsForNamespace` to `postNamespaceUnseal`
- changed `NamespaceScopedView()` usage to `(*Core).NamespaceView()` for correct barrier determination
- removes unnecessary parameter `isLoading` from `(*quotaManager).SetQuota` method
- removes `ctx` field from `quotaManager` struct
- adjusts unnecessary storage passing in `quotaManager` struct methods
- moves `quotaTypes()` from `quotas/quota_util.go` to `quotas/quotas.go`

## Rationale
quota refactor is just plain cleanup; barrier determination and loading of identity store artifacts on ns unseal is required.

## Acknowledgements

       also allowed.


satoqz:
@wslabosz-reply but do we load them again on unseal?

wslabosz-reply:
@satoqz
thought about doing follow-up but it's probably better to include that with this PR.

vault/quotas/quotas.go
```diff
@@ -1081,17 +1089,34 @@ func (m *Manager) Setup(ctx context.Context, storage logical.Storage) error {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// LoadQuotasForNamespace reverses the HandleNamespaceCleanup operation, loading quotas
+// from storage back into memdb after namespace unseal.
+func (m *Manager) LoadQuotasForNamespace(ctx context.Context) error {
```
DanGhita: Should this function take as a parameter the namespace ?

vault/quotas/quotas.go
```diff
@@ -1081,17 +1089,34 @@ func (m *Manager) Setup(ctx context.Context, storage logical.Storage) error {
...
+	m.quotaLock.Lock()
+	m.dbAndCacheLock.Lock()
```
DanGhita: Should we use RLock here ? setQuotaLocked already has a lock in m.db.Txn.

DanGhita (approved):
I have a couple of comments, but LGTM.

vault/quotas/quotas.go
wslabosz-reply: @DanGhita 
after revising the whole pattern, it doesn't allow us to load *just* namespace quotas, there's no namespace association in a quota storage entry (in memdb there's namespace index), so after pruning the quotas we won't be able to load just one subset of quotas, we have to load them all, I'll have to rethink the approach.

vault/quotas/quotas.go
wslabosz-reply: we write entries to the memdb, we should hold the write lock

vault/quotas/quotas.go
wslabosz-reply: @DanGhita 
to not introduce changes to storage, I'm now proposing going through all the quotas and load only the ones that are present in storage, but not in the memdb, this might be not performant enough so open to hear comments (@satoqz, @cipherboy)

wslabosz-reply: @DanGhita @cipherboy 
after internal discussion with @satoqz, we've decided to not "unload" anything in `quotaManager` with namespace seal.
- only root namespace operator can create/modify quotas; we are storing them in the root namespace, so there's no "governance" question there.
- deletion of the namespace removes the quota with associated namespace (through memdb index of namespace)

cipherboy: @wslabosz-reply This makes sense to me. I think I lead you down the wrong path of thinking quotas could be managed in child namespaces; they can't, they can only apply to child namespaces from the root namespace.

cipherboy (approved):
This refactor looks good to me, thanks @wslabosz-reply!

PR #3166, opened 2026-05-21
Bump golangci-lint to v2.12.2, ignore tools/ in sync-deps

satoqz:
## Description

Two changes:

1. Bump `golangci-lint` as pinned in `tools/` to v2.12.2, the same version I've pointed CI at in https://github.com/openbao/openbao/pull/3072 (CI installs from binary releases as compiling `golangci-lint` is slow)

2. Exclude `tools/` from `make sync-deps`. This way, changes in `tools/` cannot require changes in the main `go.mod`. For example, bumping `golangci-lint` beyond v2.7 would have required upgrades to transitive dependencies in the main `go.mod`, which is an unwanted effect.

## Rationale

I'm happy merging https://github.com/openbao/openbao/pull/3080 with the `sync-deps` change included. The only caveat is that we'll need to (manually, unless someone wants to automate it) keep `golangci-lint` in `tools/` in sync with CI, but that seems like an okay tradeoff in the short term. I don't think using `go install` in CI as I initially suggested as a workaround in https://github.com/openbao/openbao/pull/3080#issuecomment-4441317942 will work.

## Acknowledgements

       also allowed.


wslabosz-reply (approved):

satoqz:
Okay, also added something to make Dependabot keep the versions in sync. Pretty easy after all!

cipherboy (approved):

PR #3174, opened 2026-05-21
Avoid go run package@latest patterns

satoqz:
## Description

This replaces all invocations of `go run package@latest` or similar with ones that use a package version pinned in `tools/go.mod`. Additionally, I've replaced the old tool pattern in `tools/tools.go` with the new `tool` directive in `tools/go.mod`. This means tools are now called via `go tool`, not `go run`, are slightly easier to call (no need to pass the full package path) and should start up a little faster than via `go run`.

Some more details:

- `make fmt` is back to directly calling `gofumpt`, as @cipherboy complained about speed. Note however that formatting via `golangci-lint` applies both `gofumpt` and `goimports` -- while we could update the Makefile and scripts to run `goimports` as well, this is difficult as `goimports` is not so ergonomic as a standalone tool.
- I've added some smaller tools to `tools/go.mod`, but excluded goreleaser -- this one pulls in a lot of dependencies, and is mostly only required in CI anyway, where it is installed from binary releases.
- I've removed `GO_CMD` from the `Makefile` as it was being used inconsistently, doesn't propagate to any called scripts and was probably rarely used by anyone to override the Go command, if at all.

## Rationale

Supply chain security.

## Acknowledgements

       also allowed.


satoqz:
It also seems that, while it generally remains slower than calling the formatters plainly, the slow part in `golangci-lint fmt` was `goimports` (~8 seconds on my system) and not `gofumpt` (~2 seconds on my system). When called directly (without `golangci-lint`), this shrinks to values closer to ~4 seconds and ~1 second.

Since `goimports` covers minimal surface and is rarely violated in submitted code (most of its effects ensure that code is buildable in the first place), I'm happy treating it as a linter pass opposed to a formatter pass, i.e., not including it in `make fmt` or `fmtcheck` with the intent of keeping that fast.

cipherboy:
@satoqz I concur, sounds reasonable.

wslabosz-reply (approved):

phil9909 (approved):

DanGhita (approved):
LGTM

PR #3175, opened 2026-05-21
Add example on passing nested data to CLI via stdin

aus-aaron:
## Rationale

Resolves: #

## Acknowledgements

 - [#] By contributing this change, I certify I have not used generative AI
 - [#] By contributing this change, I certify I have signed-off on the
       also allowed.

### DIFF
--- a/website/content/docs/commands/index.mdx
+++ b/website/content/docs/commands/index.mdx
@@ -175,6 +175,20 @@ format.
 $ bao kv put secret/password value=itsasecret
 ```
 
+### Basic maps
+
+To write data alongside a basic map, you can use the command in format as follows (this example is for a `map[string]interface {}`)
+
+```shell-session
+$ bao write auth/oidc/roles/test param1="first" - \
+  param2="second" \
+    <<EOF
+    {
+      "map_param": {"name":["value1", "value2"]}
+    }
+    EOF
+```
+
 However, some OpenBao API require more advanced structures such as maps. You can
 use stdin or file input instead.
 


website/content/docs/commands/index.mdx
satoqz, line 209: There's already an `stdin` example right below this so the addition is a bit misplaced. There's value in mentioning that this is a good way to pass nested structures not easily expressed plainly on the command line -- but preferably we'd merge that below the existing `stdin` examples.

I see no reason to mention `map[string]interface{}`, this is an implementation detail that unfortunately tends to leak into error messages. Wording along the lines of "nested structures as required for some endpoints" might be better.

satoqz (approved):
Thanks, can you sign-off DCO? See https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md for instructions.

website/content/docs/commands/index.mdx
```diff
@@ -175,6 +175,20 @@ format.
...
+To write data alongside a basic map, you can use the command in format as follows (this example is for a `map[string]interface {}`)
+
+```shell-session
+$ bao write auth/oidc/roles/test param1="first" - \
+  param2="second" \
+    <<EOF
+    {
+      "map_param": {"name":["value1", "value2"]}
+    }
+    EOF
```
satoqz, line 189:
```suggestion
$ bao write auth/oidc/role/trusted-user role_type="oidc" - <<EOF
    {
        "bound_claims": { "groups": ["good"] }
    }
EOF
```

Let's simplify this and make it slightly more realistic at the same time, based on the original response at https://github.com/orgs/openbao/discussions/3145#discussioncomment-16995325.

satoqz (changes requested):
Not sure why GitHub defaulted to approve.

satoqz:
Closing since we haven't heard back from the author in a while. @aus-aaron feel free to re-open if you're still interested in contributing.

PR #3180, opened 2026-05-22
Drop github.com/klauspost/compress

### LINKED ISSUE #1864: Drop github.com/klauspost/compress

space-aditya:
## Description
- remove github.com/klauspost/compress
- update backend to support the gzipping
- add gzipping to build

## Rationale

Resolves: #1864

## Acknowledgements

       also allowed.


http/handler.go
```diff
@@ -716,6 +720,49 @@ func wrapClientCertificateHandler(h http.Handler, props *vault.HandlerProperties
...
+func precompressedUIHandler(next http.Handler, fs http.FileSystem) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Fast path: if no filesystem (UI stub), just serve what we have
+		if fs == nil {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		// Only GET requests, and only when client accepts gzip
+		if r.Method != http.MethodGet || !strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
```
phil9909: Well that is not correct: `Accept-Encoding: pack200-gzip` does contain `gzip`, but is a different [IANA standardized encoding](https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding). Someone might come up with another encoding called `gzip2` or `gzip-dict` which would break this code.

This is probably fine, the likelihood of a false positive seems rather low to me (especially since this handler is meant to be used by browsers, which generally don't implement obscure compression algorithms), but with this regex  `(^| )gzip(,|;|$)`  we should be on the safe side.

http/handler.go
```diff
@@ -716,6 +720,49 @@ func wrapClientCertificateHandler(h http.Handler, props *vault.HandlerProperties
...
+
+		// Limit to compressible extensions (must match build‑time list)
+		ext := path.Ext(r.URL.Path)
+		switch ext {
+		case ".html", ".js", ".css", ".json", ".svg", ".xml", ".txt":
+			// these are pre‑compressed
+		default:
+			next.ServeHTTP(w, r)
+			return
+		}
```
phil9909, line 750: This seems odd to me. I'd go for

```suggestion
		ext := path.Ext(r.URL.Path)
		if !slices.Contains([]string{".html", ".js", ".css", ".json", ".svg", ".xml", ".txt"}, ext) {
			next.ServeHTTP(w, r)
			return
		}
```

To me, this is more readable. But if you disagree, I'm fine with your solution.

http/handler.go
```diff
@@ -716,6 +720,49 @@ func wrapClientCertificateHandler(h http.Handler, props *vault.HandlerProperties
...
+		gzPath := r.URL.Path + ".gz"
+		if f, err := fs.Open(gzPath); err == nil && f != nil {
+			if err := f.Close(); err != nil {
+				respondError(w, http.StatusInternalServerError,
+					fmt.Errorf("error closing precompressed file: %w", err))
+				return
+			}
+			r.URL.Path = gzPath
+			w.Header().Set("Content-Encoding", "gzip")
+			w.Header().Set("Vary", "Accept-Encoding")
```
phil9909:
```suggestion
			w.Header().Add("Vary", "Accept-Encoding")
```

Some other middleware might be adding another `Vary` (I don't think this is the case today, but I'm also not sure so lets be defensive here)

http/handler.go
```diff
@@ -207,15 +207,19 @@ func handler(props *vault.HandlerProperties) http.Handler {
 		mux.Handle("/v1/sys/", handleLogical(core))
 		mux.Handle("/v1/", handleLogical(core))
 		if core.UIEnabled() {
-			if uiBuiltIn {
-				mux.Handle("/ui/", http.StripPrefix("/ui/", gziphandler.GzipHandler(handleUIHeaders(core, handleUI(http.FileServer(&UIAssetWrapper{FileSystem: assetFS()}))))))
-				mux.Handle("/robots.txt", gziphandler.GzipHandler(handleUIHeaders(core, handleUI(http.FileServer(&UIAssetWrapper{FileSystem: assetFS()})))))
+			fs := getUIAssets() // nil if UI stub, else real FS
+			if uiBuiltIn && fs != nil {
```
phil9909: Do we still need the `uiBuiltIn` variable or could it be deleted entirely?

phil9909 (dismissed):

http/handler.go
satoqz: I prefer the former TBH, but binding `ext` is a bit unnecessary.

http/assets_common.go
```diff
@@ -0,0 +1,12 @@
...
+import "net/http"
+
+// uiAssets is the file system for UI assets. It is set by the build‑tagged
+// files (ui.go or stub_asset.go) to either the real embed.FS or nil.
+var uiAssets http.FileSystem
+
+// getUIAssets returns the UI assets file system (may be nil if UI is disabled).
+func getUIAssets() http.FileSystem {
+	return uiAssets
+}
```
satoqz: This function seems unnecessary.

http/assets_stub.go
```diff
@@ -5,17 +5,8 @@
...
-	// set uiBuiltIn to false to indicate the ui is not built in. See
+	// set uiAssets to nil to indicate the ui is not built in. See
 	// http/handler.go
-	uiBuiltIn = false
-}
-
-// assetFS is a stub for building Vault without a UI.
-func assetFS() http.FileSystem {
-	return nil
+	uiAssets = nil
```
satoqz: It's already nil by default.

http/assets_common.go
satoqz, line 1: I think you can drop this file and place the single var that it adds in handler.go.

ui/ember-cli-build.js
```diff
@@ -27,6 +27,10 @@ const appConfig = {
     rootURL: '/ui/',
   },
   fingerprint: {
+    gzip: {
+      extensions: ['html', 'js', 'css', 'json', 'svg', 'xml', 'txt'],
+      keepUncompressed: true,
+    },
     exclude: ['images/'],
   },
```
satoqz, line 39: Does this actually work? It doesn't seem to generate gzipped files on my machine. I also can't find documentation on any of this stuff, but that seems like an Ember problem.

satoqz:
Per discussion in https://github.com/openbao/openbao/pull/1880, hadn't we decided to drop uncompressed assets entirely (only when we have a compressed version, of course) and embed only the gzipped ones into the binary?

http/handler.go
space-aditya: yes, it was redundant, removed

ui/ember-cli-build.js
space-aditya: it needed an explicit dependency, added it now
I had it on my local due to another project

http/handler.go
space-aditya: Updated, and added a test to avoid regressions

PR #3178, opened 2026-05-22
Handle login mfa configs lifecycle on namespace (un)seal

wslabosz-reply:
## Description
@cipherboy correctly pointed out that we retain the login mfa configs for sealed namespaces in memory, this PR:
- adjusted `loginMFABackend` constructor
- introduced login mfa configs load on namespace unseal
- introduced login mfa cleanup on namespace seal
- adjusted `*LoginMFABackend` code

## Rationale
we should remove all unnecessary memory on namespace seal, this aims at fixing one of the gaps.

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3185, opened 2026-05-26
Recursively load non-sealed child namespaces

cipherboy:
## Description

When namespace store unseals a namespace, we need to load all information for any non-sealed child namespaces as well. This occurs in two parts:

1. Discovery
2. Loading of mounts &c for all discovered namespaces.

Notably, we also clean up unloading of namespaces, so that all sealed orphaned namespaces no longer appear in memory.

## Rationale

cc: @wslabosz-reply 

Resolves: #

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -231,7 +231,9 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
 	namespacesByUUID[namespace.RootNamespaceUUID] = namespace.RootNamespace
 	namespacesByAccessor[namespace.RootNamespaceID] = namespace.RootNamespace
 
-	loadingCallback := func(namespace *namespace.Namespace) error {
+	loadNamespaceInsert := func(namespace *namespace.Namespace) error {
+		ns.logger.Info("discovered namespace", "path", namespace.Path, "uuid", namespace.UUID)
+
 		if _, ok := namespacesByUUID[namespace.UUID]; ok {
 			return fmt.Errorf("namespace with UUID %q is not unique in storage", namespace.UUID)
 		}
@@ -244,7 +246,7 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
 	}
 
 	if err := logical.WithTransaction(ctx, ns.storage, func(s logical.Storage) error {
-		return ns.loadNamespacesRecursive(ctx, s, s, loadingCallback)
+		return ns.loadNamespacesRecursive(ctx, s, s, loadNamespaceInsert)
 	}); err != nil {
 		return err
 	}
@@ -964,7 +966,7 @@ func (ns *NamespaceStore) ListNamespaces(ctx context.Context, includeParent bool
 func (ns *NamespaceStore) SealNamespace(ctx context.Context, path string) error {
 	defer metrics.MeasureSince([]string{"namespace", "seal_namespace"}, time.Now())
 
-	unlock, err := ns.lockWithInvalidation(ctx, false)
+	unlock, err := ns.lockWithInvalidation(ctx, true)
 	if err != nil {
 		return err
 	}
@@ -1017,6 +1019,18 @@ func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSe
 				errs = errors.Join(errs, err)
 			}
 		}
+
+		if entry.UUID != namespaceToSeal.UUID {
+			// Remove the namespace itself from our records if it isn't the
+			// sealed namespace. We want to forget child namespaces of a
+			// namespaces which was marked sealed, but retain the pointer to
+			// the sealed namespace itself.
+			if err := ns.namespacesByPath.Delete(entry.Path); err != nil {
+				errs = errors.Join(errs, err)
+			}
+			delete(ns.namespacesByUUID, entry.UUID)
+			delete(ns.namespacesByAccessor, entry.ID)
+		}
 	})
 
 	return errs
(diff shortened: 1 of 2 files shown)


vault/auth.go
```diff
@@ -559,10 +559,23 @@ func (c *Core) loadCredentialsForNamespace(ctx context.Context, ns *namespace.Na
 	defer c.authLock.Unlock()
 
 	// Check if we're on a non-transactional storage
+	rootCtx := namespace.RootContext(ctx)
```
wslabosz-reply: does it matter which namespace is provided within context? looking at the functions we pass the ctx to, it does not?

overall I'd say it makes more sense to provide the "unsealed" namespace context here

vault/auth.go
```diff
@@ -559,10 +559,23 @@ func (c *Core) loadCredentialsForNamespace(ctx context.Context, ns *namespace.Na
...
+			return err
+		}
+	} else {
+		if err := c.loadTransactionalCredentialsForNamespace(rootCtx, ns); err != nil {
+			return err
+		}
 	}
-	return c.loadTransactionalCredentialsForNamespace(ctx, ns)
+
+	if err := c.runCredentialUpdates(rootCtx, c.barrier, false /* no need to persist anything so far */, c.Standby()); err != nil {
```
wslabosz-reply: we run update because we could have populated auth mount table with legacy mount entry, in this case we should pass the namespace barrier [`c.NamespaceView(ns`] to the function, otherwise `persistAuth` will save the entry to the root namespace.

vault/mount.go
wslabosz-reply, line 1: same comments as to `vault/auth.go` changes

vault/namespace_store.go
```diff
@@ -1017,6 +1019,13 @@ func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSe
 				errs = errors.Join(errs, err)
 			}
 		}
+
+		// Remove the namespace itself from our records.
+		if err := ns.namespacesByPath.Delete(entry.Path); err != nil {
+			errs = errors.Join(errs, err)
+		}
+		delete(ns.namespacesByUUID, entry.UUID)
+		delete(ns.namespacesByAccessor, entry.ID)
```
wslabosz-reply: doesn't it mean we won't be able to list the sealed namespaces?

vault/auth.go
cipherboy: @wslabosz-reply but `persistAuth` is a current status flag, as in, did something ahead of this function call need this function to persist auth or not. And the answer is no, we don't insert new mounts in either of these paths. Internally if we need to adjust mounts, we update the starting value of this flag

vault/namespace_store.go
```diff
@@ -1059,35 +1068,89 @@ func (ns *NamespaceStore) unsealNamespace(ctx context.Context, namespaceToUnseal
...
+
+				return nil
+			})
+		})
+	}(); err != nil {
+		return unsealed, err
+	}
+
+	for index, newNs := range collected {
+		ns.logger.Info("unsealing discovered namespace", "path", newNs.Path, "uuid", newNs.UUID)
```
wslabosz-reply: duplicative of the previous discovery log, I'd consider removing

vault/auth.go
wslabosz-reply: then if we are never gonna persist the mounts (upgrade path, from untyped to typed mount table), why are we running the `runCredentialUpdates` at all?

(note that we are not running the mount table upgrade ever, as namespaces are created with new typed format by default)

cipherboy: @wslabosz-reply Sorry, let me rephrase.

`runCredentialUpdates(...)` takes two relevant parameters: `needPersist` and `readOnly` -- the latter is populated by `c.Standby()`.

`runCredentialUpdates(...)` is run in a context where it is (generally) the last thing to run: either in this code path or in the global instance post-unseal case.

In that latter case (which it was written for), the other loading _could_ result in additions which need to be persisted. Think: adding missing default mounts on very first startup. That necessitated passing in `needPersist` as a flag which indicates whether the _caller_  of `runCredentialUpdates(...)` needs the function to do a persistence _even if `runCredentialUpdates(...)` doesn't itself make changes which needs persisting_. 

(However, internally `runCredentialUpdates(...)` could lead to needing to persist updates too, so it will.)

All of these persistence things are predicated on `readOnly == false` of course. 

---

Here, we've not done any work that would lead to us (`loadCredentialsForNamespace(...)`) to have modified the mount table in a way that requires `runCredentialUpgrades(...)` to need to persist _on our behalf_, so we hard-code `needPersist = false`.

vault/namespace_store.go
cipherboy: @wslabosz-reply Hmm, fair. But we should delete child namespaces of sealed namespaces, yes? Because we don't want to be able to list recursively random namespaces... so we skip this if `entry == namespaceToSeal` perhaps?

vault/auth.go
cipherboy: @wslabosz-reply I had issues when `runCredentialUpdates` used a non-root or child context. I've swapped it to `parentNsCtx`, constructed properly.

wslabosz-reply: @cipherboy that's... strange? I ought to see why

EDIT:
so we have following functions using passed ctx: 
- `c.NamespaceByID()` - doesn't matter which namespace is embedded

then inside body of `persistAuth()`
- TXN func - also shouldn't matter
- `c.ListNamespaces()` - just for invalidation, so doesn't matter
- barrier operations - doesn't matter
finally we retrieve namespaces from context in just one case which is when we delete a mount passed to the function, but this doesn't happen in `runCredentialUpdates` flow, so also doesn't matter

vault/namespace_store.go
wslabosz-reply: > But we should delete child namespaces of sealed namespaces, yes?

should we? we don't really leak anything these, apart from namespaces tree structure, you cannot do anything with them other than know what id/uuid/path/metadata they have.

> Because we don't want to be able to list recursively random namespaces

I don't know what random in this context mean

vault/auth.go
> Here, we've not done any work that would lead to us (loadCredentialsForNamespace(...)) to have modified the mount table in a way that requires runCredentialUpgrades(...) to need to persist on our behalf, so we hard-code needPersist = false.

so if we hard-code `needPersist` to `false` and also know that we will never need any updates (that being untyped mount table or untyped mount entry), then what's the effect of running it last on namespace unseal? we do not modify namespace mount entries nor we persist any changes to the mounts, or am I missing something?

cipherboy: @wslabosz-reply We assign `entry.Namespace` here.

vault/namespace_store.go
cipherboy: Why would we want to leave around partial namespace objects? I think it ultimately causes problems like with reloading of namespaces. We want namespace store to be an accurate reflection of _reachable_ namespaces, not unreachable sub-child namespaces of sealed namespaces which aren't loaded, don't have storage accessible, &c.

cipherboy:
@wslabosz-reply I rolled back the changes to `auth.go`, `mount.go` and choice of context. I think we're setting ourselves up for weirdness in the random context, but so far it isn't an issue.

I can't reproduce the failures I've seen before, we'll see.

vault/namespace_store.go
wslabosz-reply: @cipherboy 
I'm a bit afraid about discoverability of these namespaces, what if operator forgets that sealed namespace `/foo` has 10+ child namespaces that are not returned on any list/scan operations? Don't get me wrong, leaving partial namespace objects doesn't sit right with me, but what other method of knowing that these namespaces exist in storage do we have?

vault/auth.go
that cannot be right, both `decodeMountEntries()`and `fetchAndDecodeMountTableEntry()` assign that field before we run the `runCredentialUpdates()`

vault/namespace_store.go
cipherboy: @wslabosz-reply 

> I'm a bit afraid about discoverability of these namespaces, what if operator forgets that sealed namespace /foo has 10+ child namespaces that are not returned on any list/scan operations? 

As soon as the instance restarts, don't we lose these anyways?

wslabosz-reply:
@cipherboy I've introduced conflicts with #3178 merge

vault/namespace_store.go
wslabosz-reply: @cipherboy
okay you are right, let's dismiss this discussion and think about discoverability in some other thread

vault/namespace_store.go
```diff
@@ -1017,6 +1019,18 @@ func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSe
...
 		}
+
+		if entry.UUID != namespaceToSeal.UUID {
+			// Remove the namespace itself from our records if it isn't the
+			// sealed namespace. We want to forget child namespaces of a
+			// namespaces which was marked sealed, but retain the pointer to
+			// the sealed namespace itself.
+			if err := ns.namespacesByPath.Delete(entry.Path); err != nil {
+				errs = errors.Join(errs, err)
+			}
```
satoqz, line 1030: It might be a little more efficient to add a namespace tree method to purge all children of a namespace, and call that once after the traversal.

vault/namespace_store.go
satoqz: I agree with @cipherboy here FWIW, from a correctness perspective we must forget about all namespaces that we can't currently re-read from storage if we wanted to.

vault/namespace_store.go
```diff
@@ -1017,6 +1019,18 @@ func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSe
...
+			delete(ns.namespacesByUUID, entry.UUID)
+			delete(ns.namespacesByAccessor, entry.ID)
+		}
```
satoqz, line 1033: I think you don't have a write lock to the namespace store here, so this doesn't work yet.

satoqz: If we're worried about write locking while `clearNamespaceResources` runs with `updateStorage=false` I _think_ I have a path to improve performance but I'd like to see some numbers reported first.

wahajahmed010 (commented):
**Solid bug fix. A few observations:**

### Logic
- The recursive loading in `unsealNamespace` correctly discovers and inserts children before calling `postNamespaceUnseal`. Ordering looks right.
- The seal cleanup in `sealNamespaceLocked` correctly removes children while keeping the sealed parent — good.

### Potential issue: race with `lockWithInvalidation`
New code acquires `lockWithInvalidation` inside `unsealNamespace`, but the calling function (`handleUnsealNamespace`) already holds the lock via `ns.lockWithInvalidation` before calling. This is a **double-lock risk** — need to check if `lockWithInvalidation` is reentrant. If not, this deadlocks.

### Error wrapping — nice touch
Wrapping `postNamespaceUnseal` errors with context ("failed to load mounts for namespace", etc.) is a legit improvement. Makes debugging vastly easier.

### Test coverage
Test adds a child namespace (`ns2`) to verify children survive seal operations — smart. But it only asserts `len(childNs) == 1` after sealing. Worth also verifying that `ns1` (the parent) is *not* in the list after seal, and that re-unsealing re-discovers children. Current test doesn't exercise unseal → reload flow.

### Suggestion (non-blocking)
Consider adding `t.Cleanup` to restore state after the test modifies namespace store, in case subsequent tests depend on clean state.

Good PR overall! The recursive loading logic is sound. 🦌

cipherboy:
@wahajahmed010 Please quit spamming this community with AI slop and read the contributing guide on discussions before commenting. Thank you!

satoqz:
@cipherboy based on their profile, I'm inclined to believe this is an openclaw bot; you can't reason with these things beyond banning.

vault/namespace_store.go
cipherboy: We still have the maps to contend with, though I do wonder what happens with concurrent `PostOrderTraversal` and `ns.namespacesByPath.Delete(entry.Path)` calls from the same thread, tbh. 🤔

cipherboy: Ahh, `PostOrderTraversal` collects all nodes up front, so we can call delete/insert/... here just fine.

vault/namespace_store.go
cipherboy: Upgraded to a write lock.

cipherboy:
@satoqz I upgraded sealing to a write lock, I think we can think about performance later. I don't imagine people are sealing namespaces all that frequently.

vault/namespace_store.go
```diff
@@ -1017,6 +1019,18 @@ func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSe
...
 			}
 		}
+
+		if entry.UUID != namespaceToSeal.UUID {
+			// Remove the namespace itself from our records if it isn't the
+			// sealed namespace. We want to forget child namespaces of a
+			// namespaces which was marked sealed, but retain the pointer to
+			// the sealed namespace itself.
+			if err := ns.namespacesByPath.Delete(entry.Path); err != nil {
+				errs = errors.Join(errs, err)
```
satoqz, line 1029: Let's `panic(err)` here since it would be a bug in `PostOrderTraversal`, and we'd leave `namespacesByPath/namespacesByUUID/namespacesByAccessor` out of sync, which seems worse than panicking.

satoqz (approved):
One nit, seems good to me otherwise.

We need to think about cases where we half-load a namespace or a tree of namespaces due to storage errors or context cancellation.

cipherboy:
@satoqz Thanks! I'll push that with some other nits of seal manager in a follow-up PR just to keep things moving. I suspect this will conflict with #3224, so I want it merged so I can rebase that but this shouldn't affect that.

vault/namespace_store.go
cipherboy: @satoqz Added in #3230, which also addressed some nits from #3205.

PR #3188, opened 2026-05-27
Add deletion of sealed namespaces

EnricoFusi:
## Description

Adds support for physical deletion of sealed namespaces. Before this change, a
namespace with its own barrier could be sealed but never removed: the deletion
path required an active barrier to clean up secrets, so a sealed namespace
became permanently undeletable.

The new behavior allows an operator holding sudo privilege to delete a sealed
namespace. The deletion is dispatched as an asynchronous job that wipes the
namespace's physical storage through the root barrier, bypassing the sealed
child barrier entirely.

## Acknowledgements

       also allowed.

### DIFF
--- a/api/sys_namespaces.go
+++ b/api/sys_namespaces.go
@@ -227,10 +227,12 @@ func (c *Sys) PatchNamespaceWithContext(ctx context.Context, name string, i *Pat
 	return result.Data, mapstructure.Decode(secret, &result)
 }
 
+// DeleteNamespace removes the namespace with the given name.
 func (c *Sys) DeleteNamespace(name string) (*DeleteNamespaceOutput, error) {
 	return c.DeleteNamespaceWithContext(context.Background(), name)
 }
 
+// DeleteNamespaceWithContext removes the namespace with the given name.
 func (c *Sys) DeleteNamespaceWithContext(ctx context.Context, name string) (*DeleteNamespaceOutput, error) {
 	if name == "" {
 		return nil, errors.New("name must not be empty")
--- a/api/sys_namespaces.go
+++ b/api/sys_namespaces.go
@@ -227,12 +227,10 @@ func (c *Sys) PatchNamespaceWithContext(ctx context.Context, name string, i *Pat
 	return result.Data, mapstructure.Decode(secret, &result)
 }
 
-// DeleteNamespace removes the namespace with the given name.
 func (c *Sys) DeleteNamespace(name string) (*DeleteNamespaceOutput, error) {
 	return c.DeleteNamespaceWithContext(context.Background(), name)
 }
 
-// DeleteNamespaceWithContext removes the namespace with the given name.
 func (c *Sys) DeleteNamespaceWithContext(ctx context.Context, name string) (*DeleteNamespaceOutput, error) {
 	if name == "" {
 		return nil, errors.New("name must not be empty")
--- a/command/namespace_delete.go
+++ b/command/namespace_delete.go
@@ -36,9 +36,13 @@ Usage: bao namespace delete [options] PATH
 
       $ bao namespace delete ns1
 
-  Delete a namespace namespace from a parent namespace (e.g. ns1/ns2/):
+  Delete a namespace from a parent namespace (e.g. ns1/ns2/):
 
       $ bao namespace delete -namespace=ns1 ns2
+
+  To delete a sealed namespace, use the delete-sealed subcommand instead:
+
+      $ bao namespace delete-sealed ns1
 
 ` + c.Flags().Help()
 
@@ -98,6 +102,6 @@ func (c *NamespaceDeleteCommand) Run(args []string) int {
 		namespacePath = namespacePath + "/"
 	}
 
-	c.UI.Output(fmt.Sprintf("Success! Namespace deleted at: %s", namespacePath))
+	c.UI.Output(fmt.Sprintf("Success! Namespace deletion scheduled: %s", namespacePath))
 	return 0
 }
--- a/command/namespace_delete.go
+++ b/command/namespace_delete.go
@@ -28,9 +28,10 @@ func (c *NamespaceDeleteCommand) Help() string {
 	helpText := `
 Usage: bao namespace delete [options] PATH
 
-  Delete an existing namespace. The namespace deleted will be relative to the
-  namespace provided in either the BAO_NAMESPACE environment variable or
-  -namespace CLI flag.
+  Delete a namespace.
+
+  The namespace deleted will be relative to the namespace provided in either the
+  BAO_NAMESPACE environment variable or -namespace CLI flag.
 
   Delete a namespace (e.g. ns1/):
 
@@ -43,6 +44,9 @@ Usage: bao namespace delete [options] PATH
   To delete a sealed namespace, use the delete-sealed subcommand instead:
 
       $ bao namespace delete-sealed ns1
+
+  Note that this requires the sudo capability, and will not clean up external
+  resources via lease deletion like standard namespace deletion does.
 
 ` + c.Flags().Help()
 
--- a/command/namespace_delete_sealed.go
+++ b/command/namespace_delete_sealed.go
@@ -30,8 +30,12 @@ func (c *NamespaceDeleteSealedCommand) Help() string {
 	helpText := `
 Usage: bao namespace delete-sealed [options] PATH
 
-  Delete a sealed namespace by physically wiping its storage. Requires sudo
-  privilege on the path.
+  Delete a sealed namespace by physically wiping its storage.
+
+  Note that this requires the sudo capability, and will not clean up external
+  resources via lease deletion like standard namespace deletion does. Prefer the
+  standard 'bao namespace delete' command unless the namespace is irrecoverable
+  due to lost seal keys.
 
   The namespace deleted will be relative to the namespace provided in either
   the BAO_NAMESPACE environment variable or -namespace CLI flag.
@@ -40,7 +44,7 @@ Usage: bao namespace delete-sealed [options] PATH
 
       $ bao namespace delete-sealed ns1
 
-  Delete a sealed namespace and all its child namespaces:
+  Delete a sealed namespace and recursively wipe its child namespaces:
 
       $ bao namespace delete-sealed -force ns1
 
--- a/command/namespace_delete_sealed.go
+++ b/command/namespace_delete_sealed.go
@@ -1,0 +1,132 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package command
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/cli"
+	"github.com/posener/complete"
+)
+
+var (
+	_ cli.Command             = (*NamespaceDeleteSealedCommand)(nil)
+	_ cli.CommandAutocomplete = (*NamespaceDeleteSealedCommand)(nil)
+)
+
+type NamespaceDeleteSealedCommand struct {
+	*BaseCommand
+
+	flagForce bool
+}
+
+func (c *NamespaceDeleteSealedCommand) Synopsis() string {
+	return "Delete a sealed namespace"
+}
+
+func (c *NamespaceDeleteSealedCommand) Help() string {
+	helpText := `
+Usage: bao namespace delete-sealed [options] PATH
+
+  Delete a sealed namespace by physically wiping its storage. Requires sudo
+  privilege on the path.
+
+  The namespace deleted will be relative to the namespace provided in either
+  the BAO_NAMESPACE environment variable or -namespace CLI flag.
+
+  Delete a sealed namespace with no child namespaces:
+
+      $ bao namespace delete-sealed ns1
+
+  Delete a sealed namespace and all its child namespaces:
+
+      $ bao namespace delete-sealed -force ns1
+
+` + c.Flags().Help()
+
+	return strings.TrimSpace(helpText)
+}
+
+func (c *NamespaceDeleteSealedCommand) Flags() *FlagSets {
+	set := c.flagSet(FlagSetHTTP)
+
+	f := set.NewFlagSet("Command Options")
+
+	f.BoolVar(&BoolVar{
+		Name:    "force",
+		Target:  &c.flagForce,
+		Default: false,
+		Usage: "Recursively delete all child namespaces of the sealed namespace. " +
+			"Required when the namespace has children.",
+	})
+
+	return set
+}
+
+func (c *NamespaceDeleteSealedCommand) AutocompleteArgs() complete.Predictor {
+	return c.PredictVaultNamespaces()
+}
+
+func (c *NamespaceDeleteSealedCommand) AutocompleteFlags() complete.Flags {
+	return c.Flags().Completions()
+}
+
+func (c *NamespaceDeleteSealedCommand) Run(args []string) int {
+	f := c.Flags()
+
+	if err := f.Parse(args); err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	args = f.Args()
+	switch {
+	case len(args) < 1:
+		c.UI.Error(fmt.Sprintf("Not enough arguments (expected 1, got %d)", len(args)))
+		return 1
+	case len(args) > 1:
+		c.UI.Error(fmt.Sprintf("Too many arguments (expected 1, got %d)", len(args)))
+		return 1
+	}
+
+	namespacePath := strings.TrimSpace(args[0])
+
+	client, err := c.Client()
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 2
+	}
+
+	params := map[string][]string{}
+	if c.flagForce {
+		params["force"] = []string{"true"}
+	}
+
+	secret, err := client.Logical().DeleteWithData(
+		fmt.Sprintf("sys/namespaces/%s/delete-sealed", namespacePath), params,
+	)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error deleting sealed namespace: %s", err))
+		return 2
+	}
+
+	if secret == nil || secret.Data == nil {
+		if secret != nil {
+			return OutputSecret(c.UI, secret)
+		}
+		c.UI.Warn("Requested namespace does not exist")
+		return 0
+	}
+
+	if !strings.HasSuffix(namespacePath, "/") {
+		namespacePath = namespacePath + "/"
+	}
+
+	for _, w := range secret.Warnings {
+		c.UI.Warn(w)
+	}
+	c.UI.Output(fmt.Sprintf("Success! Namespace deletion scheduled: %s", namespacePath))
+	return 0
+}
--- a/vault/logical_system_namespaces.go
+++ b/vault/logical_system_namespaces.go
@@ -184,6 +184,47 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
 		},
 
 		{
+			// Must be registered before the generic "namespaces/(?P<path>.+)"
+			// pattern so that the router matches this specific suffix first.
+			Pattern: "namespaces/(?P<path>[^/]+)/delete-sealed",
+
+			DisplayAttrs: &framework.DisplayAttributes{
+				OperationPrefix: "namespaces",
+			},
+
+			Fields: map[string]*framework.FieldSchema{
+				"path": namespacePathSchema,
+				"force": {
+					Type:        framework.TypeBool,
+					Description: "If true, recursively deletes all child namespaces of the sealed namespace.",
+				},
+			},
+
+			Operations: map[logical.Operation]framework.OperationHandler{
+				logical.DeleteOperation: &framework.PathOperation{
+					Callback: b.handleNamespacesDeleteSealed(),
+					Responses: map[int][]framework.Response{
+						http.StatusOK: {
+							{
+								Description: "OK",
+								Fields: map[string]*framework.FieldSchema{
+									"status": {
+										Type:        framework.TypeString,
+										Description: "Status of the deletion operation.",
+									},
+								},
+							},
+						},
+					},
+					Summary: "Delete a sealed namespace by wiping its physical storage.",
+				},
+			},
+
+			HelpSynopsis:    "Delete a sealed namespace.",
+			HelpDescription: "Physically deletes a sealed namespace by wiping its storage. Requires sudo privilege. Pass force=true to also delete child namespaces.",
+		},
+
+		{
 			Pattern: "namespaces/(?P<path>.+)",
 
 			DisplayAttrs: &framework.DisplayAttributes{
@@ -546,7 +587,7 @@ func (b *SystemBackend) handleNamespacesUnlock() framework.OperationFunc {
 	}
 }
 
-// handleNamespacesDelete handles the "/sys/namespace/<path>" endpoint to delete a namespace.
+// handleNamespacesDelete handles DELETE /sys/namespaces/<path> for unsealed namespaces.
 func (b *SystemBackend) handleNamespacesDelete() framework.OperationFunc {
 	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
 		path := namespace.Canonicalize(data.Get("path").(string))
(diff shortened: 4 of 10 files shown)


phyrog:
I'm curious about this state:

Sealed: yes
Children: yes
Sudo: yes

From the table that means if you lost the keys for a sealed namespace with children, that namespace is essentially undeletable. What could be done here?

cipherboy:
@phyrog said:

> I'm curious about this state:
> 
> Sealed: yes Children: yes Sudo: yes
> 
> From the table that means if you lost the keys for a sealed namespace with children, that namespace is essentially undeletable. What could be done here?

I concur, I think if we're going to make this a super delete, we should go all in and delete it unconditionally. 

I also think we should move this to a new endpoint; I think it'd be surprising for `DELETE` with a sudo token to delete the namespace even if it was sealed; I suspect many people have `sudo` applied broadly to `sys/*` today.

satoqz:
I don't think differentiating between these two states is correct or even possible, at least with https://github.com/openbao/openbao/pull/3185#discussion_r3311189579 taken into account:

```
Sealed: yes Children: yes Sudo: yes
Sealed: yes Children: no  Sudo: yes
```

Given a sealed namespace, we do not know if it has children or not -- we cannot read its portion of the namespace store.

As a result, if we force delete a sealed namespace, we can clear that namespace's storage -- we know its UUID -- but we cannot recursively clear the storage of any child namespaces since we do not know about them.

I _do_ think force-deleting a sealed namespace is a valid use case, but I'm not sure how we'd avoid leaving artifacts of any child namespaces in storage? These child namespaces wouldn't load again at any point because they're detached from the tree at least, but you'd need to clear them out via sys/raw.


We might have introduced namespaces that cannot be removed without sys/raw? Suppose you've got a tree of namespaces with a sealable namespace at its root that you've lost the keys to, and you've restarted the instance, how are you ever going to easily delete the entire tree?

Perhaps namespace entries must live under the root seal to make force deletion of a tree possible.

satoqz:
Actually, let me correct myself: We can deduce the tree of child namespaces even if all is sealed because the UUID is in the storage path. We still cannot rely on in-memory information to determine child namespaces and must check storage, however. But a recursive force delete should be possible.

> I also think we should move this to a new endpoint

How about we just make it a `&force=true` query parameter?

changelog/3188.txt
satoqz, line 1: AFAIK we don't have a changelog entry for namespace sealing in general yet, so no reason to add this one. I think we'll add one consolidated entry once the feature is done.

wahajahmed010 (commented):
## Review: Feat: deletion of sealed namespaces

This is a well-structured, security-conscious change. The physical storage scan for child namespace detection (using key-name-only barrier listing so it works even when in-memory trees are incomplete) is the right approach. A few observations:

### 1. Async deletion + barrier state lifecycle

`namespaceSealedDeletionJob` uses `j.store.core.barrier` (the root barrier) to wipe storage, but this runs asynchronously via `deletionDispatcher.AddJob()`. If the core is resealed between dispatch and execution, `ClearViewWithLogging` will fail because the root barrier won't be accessible. The `OnFailure` only logs the error without retry. In practice this is low-risk (resealing during mid-wipe is rare), but worth noting — the pre-existing `namespaceDeletionJob` has the same vulnerability with `j.store.storage`.

### 2. `handleNamespacesDelete` response status codes

When `DeleteNamespace` returns `ErrNamespaceHasChildren`, you return `http.StatusConflict` (409) — good, this is the right HTTP semantics for the constraint. The tests check for this. However, when `isSudo && isSealed`, you attach a warning but still return 200. Consider whether a 200 vs a different status like 202 (Accepted) would be more appropriate since the actual deletion is async — though this matches the existing non-sealed deletion pattern where "in-progress" is returned with 200.

### 3. CLI output change is fine — but note

The "deleted at" → "deletion scheduled" message change is correct for async deletion semantics, but any CI scripts parsing the old exact string will break. Consider calling this out more explicitly in the PR description.

### 4. Clean edge case handling

The sudo-gate-before-child-check ordering is correct — if someone holds sudo but children exist, they get the child error, not a misleading permission error. The physical storage scan via root barrier is a pragmatic solution to the sealed-namespace-ignorance problem. Good architecture.

The test coverage (`TestNamespaceStore_DeleteSudoSealed`) covers the key scenarios: sudo + children (rejected), non-sudo + sealed (rejected), sudo + sealed + childless (accepted, eventually gone).

satoqz:
To summarize, here's what I'd suggest:

```mermaid
flowchart TD
    start([DELETE namespace])
    sealed{Namespace\nsealed?}
    children_unsealed{Has child\nnamespaces?}
    sudo{sudo=true?}
    force{force=true?}

    ok_normal[/"200 - status=in-progress"/]
    ok_force[/"200 - status=in-progress\n(recursive physical wipe)"/]
    err409[/"409 - must delete children first"/]
    err403[/"403 - namespace is sealed;\nsudo privilege required"/]
    err_noforce[/"400 - namespace is sealed;\npass force=true to confirm physical deletion"/]

    start --> sealed
    sealed -->|No| children_unsealed
    sealed -->|Yes| sudo

    children_unsealed -->|No| ok_normal
    children_unsealed -->|Yes| err409

    sudo -->|No| err403
    sudo -->|Yes| force

    force -->|No| err_noforce
    force -->|Yes| ok_force

    style ok_normal fill:#2e7d32,color:#fff,stroke:#1b5e20
    style ok_force fill:#2e7d32,color:#fff,stroke:#1b5e20
    style err409 fill:#e65100,color:#fff,stroke:#bf360c
    style err403 fill:#b71c1c,color:#fff,stroke:#7f0000
    style err_noforce fill:#e65100,color:#fff,stroke:#bf360c
    style start fill:#424242,color:#fff,stroke:#212121
    style sealed fill:#1565c0,color:#fff,stroke:#0d47a1
    style children_unsealed fill:#1565c0,color:#fff,stroke:#0d47a1
    style sudo fill:#1565c0,color:#fff,stroke:#0d47a1
    style force fill:#1565c0,color:#fff,stroke:#0d47a1
```

We require `force=true` for both singular sealed namespaces and deeper trees, as either dangerously bypasses important steps such as lease revocation.

To cleanly remove a tree of sealed namespaces, we'd need to do a depth-first delete such that we don't detach from the tree in case we get interrupted.

satoqz:
Additional notes from a short meeting with @EnricoFusi: We agreed that it makes sense to have both `force` and `recursive` flags where the latter explicitly opts into any recursive deletion to doubly guard the most destructive scenario.

EnricoFusi:
Ok, to summarize the comments. And a short meeting with @satoqz 


Basically, we need some trade-off between the risks of implementing a "rm -rf" in the vault, and the need to do it when credentials are lost, or in the hands of some now-untrusted person, like a person who left the position.

The  proposal of trade-off comes from unix. "rm -rf" has two options, one is "f" , which we have as "force=true", and then the proposal is, **only for this usecase mentioned**, 

Sealed: yes
Children: yes
**Sudo: yes**

to have "recursive=true". Please notice, this will still depend by the sudo policy, **still dominating**.

then we add the recursive flag, which will implement a visit of the whole subtree, deleting in post-order (lower leaf first, then upper ones).

@phyrog,   @cipherboy  ,  hope I addressed your comments. Before, I was "a little" conservative about deleting entire trees.

cipherboy:
@satoqz said:

> @cipherboy said:
>
> > I also think we should move this to a new endpoint
> 
> How about we just make it a `&force=true` query parameter?

`force=true` is ACL-able, but it puts the onus on the policy writer to explicitly set `denied_parameters=["force"]` on all policies where we don't want to allow unsafe deletion of sealed namespaces, versus instead shifting it to a new endpoint. In particular, at the policy level, you don't actually really know if a given namespace is sealed or not, and whether `force` will have any effect. :shrug: 

I don't think `sudo` permission alone is a good sentinel for operator intent to want to delete a _sealed_ namespace. I think `sudo` + `force` could work fine, though. But I'd propose two separate deletion endpoints: the normal one behaves consistently if the namespace is not sealed or unsealed, and the latter strictly handles the deletion of sealed namespaces. 

In that sense, this PR should just be net-adding a new code path, +/- any special logic. 

> then we add the recursive flag, which will implement a visit of the whole subtree, deleting in post-order (lower leaf first, then upper ones).

@EnricoFusi Lets pass on this for now for the non-sealed namespace case. There's examples like https://github.com/hashicorp/vault/issues/885 where we'd probably want a new HTTP Verb for this in the general case (with non-sealed namespace recursive deletion).

----

To revisit, I was thinking our existing logic looks like:


```mermaid
flowchart TD
    DNS(DELETE sys/namespace/a)

    NSS{Namespace not sealed or Unsealed?}
    DNS-->NSS

    NSSR[400: cannot delete sealed namespace via this endpoint. unseal or use other endpoint]
    NNSHC{Namespace has Children?}
    NSS-- no; sealed -->NSSR
    NSS-- yes; unsealed -->NNSHC

    NNSOK[200: status=in-progress]
    NNSCLD[409: must delete children first]
    NNSHC-- yes; have children -->NNSCLD
    NNSHC-- no children -->NNSOK
```

and then for the new endpoint:

```mermaid
flowchart TD
    DSNS(DELETE sys/
         namespace/b/delete-sealed)

    NSS{Namespace Sealed?}
    DSNS-->NSS

    NNSR[400: cannot delete non-sealed namespace via this endpoint. unseal or use other endpoint]
    NSHC{Namespace has Children?}
    NSS-- yes -->NSHC
    NSS-- no -->NNSR

    NSNC[200: in-progress]
    NSC{token with sudo=true and force=true on API request}
    NSHC-- no children -->NSNC
    NSHC-- yes; have children -->NSC

    NSNF[403: call this endpoint with a sudo-enabled token and force=true]
    NSC-- yes -->NSNC
    NSC-- no -->NSNF
```

---

Thoughts @wslabosz-reply / @EnricoFusi / @satoqz?

satoqz:
> > then we add the recursive flag, which will implement a visit of the whole subtree, deleting in post-order (lower leaf first, then upper ones).
> 
> @EnricoFusi Lets pass on this for now for the non-sealed namespace case. There's examples like [hashicorp/vault#885](https://github.com/hashicorp/vault/issues/885) where we'd probably want a new HTTP Verb for this in the general case (with non-sealed namespace recursive deletion).

Oh, yes, we only meant to introduce recursive deletion for sealed trees as a way to clean up lost data.

---

@cipherboy fair point on ACLs. The updated flowchart is fine with me though I wonder if even sealable namespaces with no children warrant a sudo check.

Also, if we're going to have `force=true` authorize recursive deletion of children, we might as well call it `recursive`?

cipherboy:
> @cipherboy fair point on ACLs. The updated flowchart is fine with me though I wonder if even sealable namespaces with no children warrant a sudo check.

Yeah, let's add that. 

> Also, if we're going to have force=true authorize recursive deletion of children, we might as well call it recursive?

I'm happy either way, just wanting to avoid having people point to `recursive=true` and call it great API design for what should be a custom verb.

EnricoFusi:
Ok. So I am working at a new endpoint. I aligned already with @satoqz  on the behavior. 

| CLI command                           | Endpoint                                          | Parameters  |
|---------------------------------------|---------------------------------------------------|-------------|
| `bao namespace delete <name>`         | `DELETE /v1/sys/namespaces/<name>`                | none        |
| `bao namespace delete-sealed <name>`  | `DELETE /v1/sys/namespaces/<name>/delete-sealed`  | none        |
| `bao namespace delete-sealed -force <name>` | `DELETE /v1/sys/namespaces/<name>/delete-sealed` | `force=true` |

api/sys_namespaces.go
```diff
@@ -227,12 +227,16 @@ func (c *Sys) PatchNamespaceWithContext(ctx context.Context, name string, i *Pat
 	return result.Data, mapstructure.Decode(secret, &result)
 }
 
-// DeleteNamespace removes the namespace with the given name.
+// DeleteNamespace removes the namespace with the given name. If the namespace
+// is sealed and the caller holds sudo privilege on the path, the server
+// automatically performs a physical storage wipe.
```
satoqz: Not the case with the latest design, revert.

api/sys_namespaces.go
```diff
@@ -227,12 +227,16 @@ func (c *Sys) PatchNamespaceWithContext(ctx context.Context, name string, i *Pat
...
 func (c *Sys) DeleteNamespace(name string) (*DeleteNamespaceOutput, error) {
 	return c.DeleteNamespaceWithContext(context.Background(), name)
 }
 
-// DeleteNamespaceWithContext removes the namespace with the given name.
+// DeleteNamespaceWithContext removes the namespace with the given name. If the
+// namespace is sealed and the caller holds sudo privilege on the path, the
+// server automatically performs a physical storage wipe.
```
satoqz: Not the case with the latest design, revert.

changelog/3188.txt
satoqz, line 1: This still needs to be removed.

command/namespace_delete.go
```diff
@@ -83,13 +87,16 @@ func (c *NamespaceDeleteCommand) Run(args []string) int {
 		return 2
 	}
 
-	resp, err := client.Sys().DeleteNamespace(namespacePath)
+	secret, err := client.Logical().Delete("sys/namespaces/" + namespacePath)
```
satoqz: I think this can be reverted to use the helper again?

command/namespace_delete_sealed.go
```diff
@@ -0,0 +1,132 @@
...
+func (c *NamespaceDeleteSealedCommand) Synopsis() string {
+	return "Delete a sealed namespace"
+}
+
+func (c *NamespaceDeleteSealedCommand) Help() string {
+	helpText := `
+Usage: bao namespace delete-sealed [options] PATH
+
+  Delete a sealed namespace by physically wiping its storage through the root
+  barrier. Requires sudo privilege on the path.
```
satoqz: Don't mention "barrier" in user-facing contexts. There's a glossary entry, but the majority of users presumably does not know what it means. We can just say "by physically wiping its storage" only, without mentioning implementation details.

command/namespace_delete_sealed.go
```diff
@@ -0,0 +1,132 @@
...
+		return 2
+	}
+
+	params := map[string][]string{}
+	if c.flagForce {
+		params["force"] = []string{"true"}
+	}
+
+	secret, err := client.Logical().DeleteWithData(
+		"sys/namespaces/"+namespacePath+"/delete-sealed", params,
```
satoqz: Can we use `fmt.Sprintf()`?

vault/logical_system_namespaces.go
```diff
@@ -182,6 +182,47 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
...
+								Fields: map[string]*framework.FieldSchema{
+									"status": {
+										Type:        framework.TypeString,
+										Description: "Status of the deletion operation.",
+									},
+								},
+							},
+						},
+					},
+					Summary: "Delete a sealed namespace, wiping its physical storage through the root barrier.",
```
```suggestion
					Summary: "Delete a sealed namespace by wiping its physical storage.",
```

vault/logical_system_namespaces.go
```diff
@@ -182,6 +182,47 @@ func (b *SystemBackend) namespacePaths() []*framework.Path {
...
+				},
+			},
+
+			HelpSynopsis:    "Delete a sealed namespace.",
+			HelpDescription: "Physically deletes a sealed namespace by wiping its storage through the root barrier. Requires sudo privilege. Pass force=true to also delete child namespaces.",
```
```suggestion
			HelpDescription: "Physically deletes a sealed namespace by wiping its storage. Requires sudo privilege. Pass force=true to also delete child namespaces.",
```

vault/namespace_store.go
```diff
@@ -61,6 +61,20 @@ const (
...
+
+// ErrNamespaceIsSealed is returned by DeleteNamespace when the target namespace
+// is sealed. Sealed namespaces must be deleted via the dedicated delete-sealed
+// endpoint, which performs a physical storage wipe through the root barrier.
+var ErrNamespaceIsSealed = errors.New("namespace is sealed; use DELETE .../delete-sealed to delete a sealed namespace")
+
+// ErrNamespaceSealedHasChildren is returned by DeleteSealedNamespace when the
+// target namespace has child namespaces and force=false. The caller must pass
+// force=true to authorize recursive deletion of the entire subtree.
+var ErrNamespaceSealedHasChildren = errors.New("sealed namespace has child namespaces; pass force=true to delete the entire subtree, or unseal the namespace barrier and delete children individually")
```
satoqz: Can we avoid having variables for these and return them right from the namespace store, including the correct `CodedError`, then use `handleError` in the handler?

`ErrNamespaceIsSealed` is a suboptimal error name when it only applies to deletion. There's `ErrNamespaceSealed` already, too, which is more general.

vault/namespace_store.go
```diff
@@ -1248,24 +1268,93 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
...
+	isNamespaceDeleting := ns.creationDeletionMap[namespaceToDelete.UUID]
+	if namespaceToDelete.Tainted && isNamespaceDeleting {
+		return "in-progress", nil
+	}
+
+	// Physical storage check for child namespaces. Key names are visible
+	// through the root barrier even when the child barrier is sealed, and
+	// even when children are absent from the in-memory namespace tree after
+	// a restart.
+	view := NamespaceScopedView(ns.core.sealManager.NamespaceBarrierByLongestPrefix(parent.Path), namespaceToDelete)
```
satoqz: No need to get the appropriate barrier I believe, all barriers should be able to list whereever. So just use the root one and skip the lookup.

vault/namespace_store.go
```diff
@@ -1639,6 +1728,130 @@ func (j *namespaceDeletionJob) OnFailure(err error) {
...
+	// memory (e.g., sealed children not reloaded after a restart) require no
+	// in-memory cleanup.
+	j.store.namespacesByPath.PostOrderTraversal(j.target.Path, func(entry *namespace.Namespace) {
+		delete(j.store.creationDeletionMap, entry.UUID)
+		delete(j.store.namespacesByUUID, entry.UUID)
+		delete(j.store.namespacesByAccessor, entry.ID)
+		j.store.core.sealManager.RemoveNamespace(entry)
+		// Children are removed before parents (post-order), so Delete succeeds.
+		_ = j.store.namespacesByPath.Delete(entry.Path)
+	})
```
satoqz: This does not work because children won't be in memory, that's the entire reason for checking their existence in storage beforehand.

vault/namespace_store.go
```diff
@@ -1639,6 +1728,130 @@ func (j *namespaceDeletionJob) OnFailure(err error) {
...
+
+func (j *namespaceSealedRecursiveDeletionJob) Execute() error {
+	// Wipe the entire storage prefix for this namespace and all descendants.
+	// Nested namespaces live under their ancestor's storage prefix, so one
+	// root-barrier wipe covers the complete subtree.
+	view := NamespaceScopedView(j.store.core.barrier, j.target)
+	if err := logical.ClearViewWithLogging(j.store.creationDeletionJobContext,
+		view, j.store.logger); err != nil {
+		return fmt.Errorf("failed to wipe storage for sealed namespace tree %q: %w", j.target.Path, err)
+	}
```
satoqz: You are wiping the main namespace only here, none of its children. You need to recurse through storage, not through the in-memory tree.

EnricoFusi: Should be fixed with last commit

vault/namespace_store.go
EnricoFusi: Should be fixed in the last push

vault/namespace_store.go
EnricoFusi: fixed in the last push

command/namespace_delete_sealed.go
```diff
@@ -0,0 +1,136 @@
...
+		if secret != nil {
+			return OutputSecret(c.UI, secret)
+		}
+		c.UI.Warn("Requested namespace does not exist")
+		return 0
+	}
+
+	if !strings.HasSuffix(namespacePath, "/") {
+		namespacePath = namespacePath + "/"
+	}
```
wslabosz-reply, line 129: we don't need that

satoqz: Why not?

vault/logical_system_namespaces_seals.go
```diff
@@ -137,11 +137,50 @@ func (b *SystemBackend) namespaceSealPaths() []*framework.Path {
 			HelpSynopsis:    strings.TrimSpace(sysNamespacesSealsHelp["namespaces-seal"][0]),
 			HelpDescription: strings.TrimSpace(sysNamespacesSealsHelp["namespaces-seal"][1]),
 		},
+
+		{
+			Pattern: "namespaces/(?P<path>[^/]+)/delete-sealed",
```
wslabosz-reply: this is fine but because we are doing:
```
		path := namespace.Canonicalize(data.Get("path").(string))
		if len(path) > 0 && strings.Contains(path[:len(path)-1], "/") {
			return nil, errors.New("path must not contain /")
		}
```
first thing first in the handler code, this doesn't do anything actually

command/namespace_delete_sealed.go
wslabosz-reply: I don't think users care about trailing slash at the end for the namespace path, but I'm not hang up on it

satoqz: I would be very offended if it were to go missing.

vault/namespace_store.go
```diff
@@ -1674,27 +1796,35 @@ func (j *namespaceDeletionJob) Execute() error {
...
-	if err := j.store.namespacesByPath.Delete(j.target.Path); err != nil {
-		return fmt.Errorf("failed to delete namespace entry in namespace tree: %w", err)
+	// Remove the namespace's storage entry:
+	view := NamespaceScopedView(j.store.storage, j.parent).SubView(namespaceStoreSubPath)
+	if err := view.Delete(ctx, j.target.UUID); err != nil {
+		return fmt.Errorf("failed to delete namespace storage entry: %w", err)
 	}
 
+	// Finally, remove entries from memory:
+	_ = j.store.namespacesByPath.Delete(j.target.Path)
```
wslabosz-reply: we assume that the error cannot happen as it's either:
- root namespace
- namespace with children
which are both accounted for before we get to this exec context?

(if so please write smallish comment)

satoqz: Eh I'll add the check back. It shouldn't happen ever if the rest of the system behaves as designed, but that's never the case anyway.

vault/namespace_store.go
```diff
@@ -1320,12 +1317,91 @@ func (ns *NamespaceStore) DeleteNamespace(ctx context.Context, path string) (str
...
+	}
+
+	isNamespaceDeleting := ns.creationDeletionMap[namespaceToDelete.UUID]
+	if namespaceToDelete.Tainted && isNamespaceDeleting {
+		return "in-progress", nil
+	}
+
+	if namespaceToDelete.ID == namespace.RootNamespaceID {
+		return "", errors.New("unable to delete root namespace")
+	}
```
wslabosz-reply, line 1361: up to this point (apart from metric string) it's shared between "normal" deletion

satoqz: I don't think it's worth factoring out, I can only see that adding more spaghetti.

vault/namespace_store.go
```diff
@@ -1451,6 +1527,59 @@ func (ns *NamespaceStore) clearNamespacePolicies(ctx context.Context, namespace
...
+		// Then wipe the namespace.
+		if err := logical.ScanViewPaginated(
+			ctx, view, ns.logger,
+			logical.DefaultScanViewPageLimit,
+			func(page int, index int, path string) (bool, error) {
+				err := view.Delete(ctx, path)
+				return err == nil, err
+			},
+		); err != nil {
+			return fmt.Errorf("failed to clear namespace view: %w", err)
```
```suggestion
			return fmt.Errorf("failed to clear namespace view for %q: %w", uuid, err)
```

satoqz:
@cipherboy this should be good to go, too.

wslabosz-reply (approved):
LGTM

command/namespace_delete_sealed.go
```diff
@@ -0,0 +1,136 @@
...
+	return "Delete a sealed namespace"
+}
+
+func (c *NamespaceDeleteSealedCommand) Help() string {
+	helpText := `
+Usage: bao namespace delete-sealed [options] PATH
+
+  Delete a sealed namespace by physically wiping its storage.
+
+  Note that this requires the sudo capability, and will not clean up external
```
cipherboy, line 35: nit: 

```suggestion
  Note that this requires the sudo capability and will not clean up external
```

cipherboy (approved):
Nit is not worth addressing this time around.

PR #3033, opened 2026-05-04
Document deprecation of generate-root cmd

wslabosz-reply:
## Description
- Adds a warning box to the `operator generate-root` cmd documentation.
- Adds a warning text to the help text of `operator generate-root` command, and warning info on the `operator generate-root -init` command output.

## Rationale
We've added same warning to the API endpoint documentation, but haven't done so in case of cmd, so this PR corrects that omit.

Resolves: #3020

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3190, opened 2026-05-27
Use new /sys/generate-root-token endpoint in API and CLI

phyrog:
## Description

- switches the API (and thus the CLI) to use the `/sys/generate-root-token` endpoint in place of `/sys/generate-root`
- fixes some double-read-locks
- always forward request for `/sys/generate-root-token` from read-replicas to leader

## Rationale

The `/sys/generate-root-token` endpoint replaces the old and unauthenticated `/sys/generate-root` endpoint. The new endpoint also has namespace support.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3190.txt
+++ b/changelog/3190.txt
@@ -1,8 +1,7 @@
 ```release-note:change
-api: `(*Sys).GenerateRoot*` methods now use the authenticated `/sys/generate-root-token` endpoints instead of the deprecated `/sys/generate-root` endpoints
+api: `(*Sys).GenerateRoot*` methods now use the authenticated `/sys/generate-root-token` endpoints instead of the deprecated `/sys/generate-root` endpoints.
 ```
 
 ```release-note:change
-command: `operator generate-root` now uses the authenticated `/sys/generate-root-token` endpoints instead of the deprecated `/sys/generate-root` endpoints
-
+command: `operator generate-root` now uses the authenticated `/sys/generate-root-token` endpoints instead of the deprecated `/sys/generate-root` endpoints.
 ```
--- a/changelog/3190.txt
+++ b/changelog/3190.txt
@@ -1,0 +1,8 @@
+```release-note:change
+api: `(*Sys).GenerateRoot*` methods now use the authenticated `/sys/generate-root-token` endpoints instead of the deprecated `/sys/generate-root` endpoints
+```
+
+```release-note:change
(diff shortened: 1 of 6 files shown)


satoqz:
@phyrog test failures seem legitimate, though TIL about `external_tests/api`.

vault/request_handling.go
```diff
@@ -965,6 +965,9 @@ func (c *Core) handleCancelableRequest(ctx context.Context, req *logical.Request
 		if c.standby.Load() && !c.StandbyReadsEnabled() {
 			return nil, ErrCannotForwardLocalOnly
 		}
+	// Always forward generate-root-token request from standby nodes.
+	case c.standby.Load() && strings.HasPrefix(req.Path, "sys/generate-root-token/"):
+		return nil, logical.ErrPerfStandbyPleaseForward
 	}
```
satoqz: I think you can use `ForwardPerformanceStandby` on the path declaration in `vault/logical_system_generate_root.go` to make this less invasive?

satoqz:
Actually, this probably deserves two `change`-type changelog entries under `api:` and `command:` just to note that these are now re-routed and don't go to the legacy endpoints.

vault/request_handling.go
phyrog: Yes, perfect. I was looking for something like that but did not see that field.

phyrog:
@satoqz All done

phyrog:
Should I revert the warning from #3033? Since we'd not be using the old endpoint anymore?

/cc @wslabosz-reply

wslabosz-reply:
@phyrog 
Yes, I think so.

changelog/3190.txt
```diff
@@ -0,0 +1,8 @@
+```release-note:change
+api: `(*Sys).GenerateRoot*` methods now use the authenticated `/sys/generate-root-token` endpoints instead of the deprecated `/sys/generate-root` endpoints
+```
+
+```release-note:change
+command: `operator generate-root` now uses the authenticated `/sys/generate-root-token` endpoints instead of the deprecated `/sys/generate-root` endpoints
+
```
satoqz:
```suggestion
command: `operator generate-root` now uses the authenticated `/sys/generate-root-token` endpoints instead of the deprecated `/sys/generate-root` endpoints.
```

satoqz (approved):
LGTM, tiny changelog adjustments suggested below.

cipherboy (approved):
Some days I wonder if `sys/` should just globally be forwarded to the primary, though I'm not convinced this does anything great for us. 😆

PR #3206, opened 2026-05-29
Fix up LICENSE_DEPENDENCIES.tpl for go-licenses/v2

satoqz:
## Description

Update `LICENSE_DEPENDENCIES.tpl` to be compatible with the upgrade of `github.com/google/go-licenses` to v2.

## Rationale

When go-licenses fails to detect a license, `.LicenseText` would previously return an empty string. In v2, it instead attempts to read `"Unknown"` as a literal file path, which fails.

## Acknowledgements

       also allowed.

### DIFF
--- a/LICENSE_DEPENDENCIES.tpl
+++ b/LICENSE_DEPENDENCIES.tpl
@@ -5,14 +5,16 @@ license terms. These dependencies are managed via the `go.mod` and
 `go.sum` files, and included in the source tarball.
 
 The dependencies and their licenses are as follows:
-
-{{ range . }}
+{{- range . }}
 
 ## {{ .Name }}
 
 **License:** {{ .LicenseName }}
 
+{{- if ne .LicensePath "Unknown" }}
+
 ```
 {{ .LicenseText }}
 ```
-{{ end }}
+{{- end -}}
+{{- end -}}


satoqz:
Ah, seems we can't even use it from `tools/` because v1 is so ancient it won't compile due to conflicts with other tools that use later dependency versions.

tsaarni:
The following allows `go tool` with v2 to exit cleanly

````diff
diff --git a/LICENSE_DEPENDENCIES.tpl b/LICENSE_DEPENDENCIES.tpl
index fefcad2233..3eedd5b0f5 100644
--- a/LICENSE_DEPENDENCIES.tpl
+++ b/LICENSE_DEPENDENCIES.tpl
@@ -12,7 +12,8 @@ The dependencies and their licenses are as follows:

 **License:** {{ .LicenseName }}

-```
+{{ if ne .LicensePath "Unknown" }}```
 {{ .LicenseText }}
 ```
 {{ end }}
+{{ end }}
````

However, the output is not exactly the same as it was with v1.

satoqz:
Right, thanks @tsaarni. Really up to what @DrDaveD thinks then.

DrDaveD:
It's alright for the output format to change a little if you'd rather do that.  The format is interpreted by script at https://github.com/openbao/openbao-fedora/blob/master/make-spec#L64 so hopefully that doesn't break, but even if it does, presumably the script could be adapted to work with the new format.

satoqz:
Gotcha, I'd rather stay on v2 then. Updated the template as suggested by @tsaarni, and also made some improvements to whitespace.

@DrDaveD this might result in you needing to update `EXTRALICENSES` starting with v2.6, I believe Apache-2 is also "Unknown" now, perhaps some others, too.

JanMa (approved):

Issue #2048, opened 2025-10-31
Adopt immutable releases


JanMa:
GitHub announced the availability of immutable Releases where the associated tags and assets can't be changed after a release was published:

https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/immutable-releases

I think this would be a great feature for us to adopt and should work with the existing release pipeleine.
The only thing we need to change is to create a "Draft" release with GoReleaser and then manually publish the release one all build jobs completed.

cipherboy:
> The only thing we need to change is to create a "Draft" release with GoReleaser and then manually publish the release one all build jobs completed.

@JanMa If we need a draft release, this might be problematic. We can't use them currently due to the way GoReleaser OSS works with our parallel builds all uploading to the same release. Probably blocked on that other issue then (#2009).

JanMa:
@cipherboy I don't see a reason this shouldn't work with GoReleaser OOS 🤔 
We just need to set the following in our config:
```yaml
  # If set to true, will not auto-publish the release.
  # Note: all GitHub releases start as drafts while artifacts are uploaded.
  # Available only for GitHub and Gitea.
  draft: true

  # Whether to remove existing draft releases with the same name before creating
  # a new one.
  #
  # Only effective if `draft` is set to true.
  # Available only for GitHub.
  replace_existing_draft: false

  # Whether to use an existing draft release as the target release.
  #
  # Available only for GitHub.
  # Since: v2.5.
  use_existing_draft: true
```

cipherboy:
@JanMa Ahh, I think `use_existing_draft` may not have existed when we started. That looks like it'll solve the exact problem we had, iirc: it tried creating a new draft even when one already existed.

satoqz:
I suggest enabling immutable releases in this repo starting with the v2.7.x series. v2.6.0 beta technically could've been immutable already, but let's give changes from #3209 the entire release cycle to test, and then we can commit to it.

cipherboy:
@satoqz Other than me clicking the button, is there anything else we need to do here?

satoqz:
@cipherboy that should be all.

PR #3209, opened 2026-06-01
Adopt multi-stage, optimized release pipeline

### LINKED ISSUE #2009: Utilize Goreleaser Pro - Split & Merge for better release pipelines
In v2.4.2, the nightly release pipeline failed due to disk space issues. Sadly, I was not patient enough and pushed the `v2.4.2` tag prior to it failing.

We have a Goreleaser Pro license; we should make use of it for the main builds (since, in reality, nobody likely uses Goreleaser locally as it is not documented).

https://goreleaser.com/customization/partial/

Here, we could likely build a matrix:

```yml
matrix
  release_os:
    - linux
    - hsm
    - darwin
    - freebsd
    - netbsd
    - openbsd
    - windows
  release_arch:
    - amd64
    - arm64
    - riscv64
    - arm
    - ppc64le
    - s390x
```

And perhaps a [set of excludes](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#excluding-matrix-configurations), and make each job a separate pipeline. 

Likewise, we should attempt to split publishing as well: tag push should be handled by goreleaser once all builds have succeeded and _not_ be pushed unless they did.

### LINKED ISSUE #2456: Goreleaser architechture ambiguity (amd64 vs. x86_64, arm64 vs aarch64)
**Describe the bug**

Currently, goreleaser uses the following arch names for each architecture:

| Archive name | RPM arch | DEB arch |
|--------------|----------|----------|
| x86_64       | x86_64   | amd64    |
| arm64        | aarch64  | arm64    |

I can understand, [why](https://unix.stackexchange.com/questions/667735/dpkg-deb-treats-my-architecture-as-amd64-instead-of-x86-64) e.g. `amd64` is used for Debian packaging. 

The renaming to `x86_64` seems to have been done, so that the architecture of the binary archives matches `uname` output:
```
# this name template makes the OS and Arch compatible with the results of `uname`
```

However, this doesn't hold up for 64-bit arm systems, where the output would be `aarch64`, not `arm64`, as we're currently naming the archives.

This means, something like 
```
wget https://github.com/openbao/openbao/releases/download/v2.5.0/bao_2.5.0_Linux_$(uname -m).tar.gz 
```
will not work in every case.

Lastly, this causes confusing ordering for different distribution formats; personally I have accidentally downloaded the `arm64` binary once:

Image

I feel like there are two solutions to this:
- Rename all binary packages to their matching `uname -m` output
- Keep architecture names between distribution formats as similar as possilbe

### LINKED ISSUE #2802: Build OCI Images instead of Docker Images
I've noticed that our main `openbao` images are still using the old [Docker Image Manifest V2](https://distribution.github.io/distribution/spec/manifest-v2-2/) schema instead of the [OCI Image Manifest](https://specs.opencontainers.org/image-spec/manifest/). Some modern tooling like systemd-importd no longer supports the Docker Manifest V2. 

Given that Goreleaser now also properly supports building [multi-arch imges](https://goreleaser.com/customization/package/dockers_v2/) I suggest we switch to building multi-arch OCI images for OpenBao. We're already doing that for `openbao-plugins` and it works pretty well.

satoqz:
## Description

This is a collection of changes that make up major improvements to our release building & publishing pipeline. I've taken great care to split the changes into descriptive commits to make my chain of thought easy to follow, but here's an overview to show off all the good stuff:

- I've sharded the building of binaries down to one binary per job, opposed to building all binaries for a given OS in one job. This means all binaries build in 4-5 minutes, not 30-40.
- I've introduced a workflow around draft releases, which will also let us adopt immutable releases. We now start the release workflow without a tag pushed. The workflow creates a draft release for the desired version number and attaches artifacts. We then review those artifacts, attach the changelog and publish the release, which creates the tag. If anything goes wrong, we restart or patch until we've got a draft release we're happy to publish.
- Container images are not built or published until we undraft the release, neither are repositories at pkgs.openbao.org updated. Instead, the publishing of the release triggers second-stage workflows that perform those actions only once the main artifacts are live and set in stone. Nightly releases automatically undraft and publish second-stage artifacts.
- Container images are no longer built by goreleaser. We can also rebuild and republish container images for a release when the base image has seen updates, without cutting a new overall release. The build is sharded by distribution (openbao, openbao-ubi, etc.) to be optimally parallel.

To summarize, this cuts release pipeline runtime down by at least three quarters, and gives us lots of fault tolerance by delaying tag pushing and allowing us to retry each stage independently in case of 3rd party service blips.

Here's what the graph of a nightly release looks like:

image

Notably, for regular GA or beta releases, there'd be a manual undraft step that triggers container builds, and we also trigger jobs to publish RPMs and DEBs to pkgs.openbao.org.

## Rationale

Resolves: #2802
Resolves: #2456
Resolves: #2009

Also removes any blockers I see for #2048, but we'll still need to actually enable immutable releases in the repository settings. I suggest we do that once we've triaged the new pipeline in official releases a couple times.

## Acknowledgements

       also allowed.


satoqz:
Well this is funny, when I test on the official nightly repo, GitHub still split-brains and goreleaser ends up creating additional draft releases.

That effectively means that some jobs do not see a release that was created 4 minutes prior. Cool. Even more interestingly, this hasn't happened a single time on my fork, and I've ran dozens of test releases over there.

I suppose I'll need to upload build artifacts to workflow artifacts instead, and have a final job accumulate and upload to the release at once.

Edit: Alright, fixed as mentioned. We rely even less on goreleaser now, and can realistically abandon it entirely in the future with some extra effort.

cipherboy:
> I suppose I'll need to upload build artifacts to workflow artifacts instead, and have a final job accumulate and upload to the release at once.
> 
> Edit: Alright, fixed as mentioned. We rely even less on goreleaser now, and can realistically abandon it entirely in the future with some extra effort.

This is rather nice fwiw.

cipherboy:
@satoqz said:

> * Container images are no longer built by goreleaser. We can also rebuild and republish container images for a release when the base image has seen updates, without cutting a new overall release. The build is sharded by distribution (openbao, openbao-ubi, etc.) to be optimally parallel.

How does this work w.r.t. release branches? Do we rebuild the container image off of main, from the actual tag, or from a release branch updated commit?

satoqz:
> How does this work w.r.t. release branches? Do we rebuild the container image off of main, from the actual tag, or from a release branch updated commit?

Good question. The answer is essentially "whatever works best". You pass the version to re-release to the workflow and it pulls binaries from the existing release, so there's no need to run it from a release branch or a tag. What matters instead is that you run it at a commit that has the updated packaging (updated base images or Dockerfile instructions).

So I think unless we make major changes to packaging for a pending release, it's possible to re-release right from main, or else via a backport.

As future work, It'd be nice to additionally tag any re-releases as `-r1`, `-r2` etc.

satoqz:
I forgot to mention in my commits: This also upgrades goreleaser to latest.

goreleaser.hsm.yaml
```diff
@@ -117,270 +86,16 @@ nfpms:
...
-    # this name template makes the OS and Arch compatible with the results of `uname`.
+  - format_overrides:
+      - formats: tar.gz
     name_template: >-
-      {{ .ProjectName }}-hsm_{{ .Version }}_{{- title .Os }}_
-      {{- if eq .Arch "amd64" }}x86_64
-      {{- else if eq .Arch "386" }}i386
-      {{- else }}{{ .Arch }}{{ end }}
-      {{- if .Arm }}v{{ .Arm }}{{ end }}
+      {{ .ProjectName }}-hsm_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}
```
cipherboy, line 99: This changes arch names, yes?

goreleaser.linux.yaml
```diff
@@ -106,822 +78,18 @@ nfpms:
     rpm:
       signature:
         key_file: "{{ .Env.GPG_KEY_FILE }}"
-checksum:
```
cipherboy, line 109: Do we not still want checksums to be generated and signed?

changelog/3209.txt
```diff
@@ -0,0 +1,16 @@
+```release-note:change
+releases: Disable publishing checksum artifacts to releases on GitHub. GitHub
```
cipherboy: Ah, I missed this. Just from a logging/attestation PoV, I kinda would like if we generate checksums and echo/cat them, so that we know they're consistent between action runs? It'd be great if we, for instance, validated all checksums/cosign signatures between stages and before building images/packages. 

I recall there was some difference between our checksums and GitHub's.... Was one the checksum of the served zip?

.github/workflows/release.yml
```diff
@@ -1,486 +1,366 @@
...
+            | .builds[0].goarch |= ["${{ matrix.target.arch }}"]
+          '
+
+          # Inject 'goarm' if defined:
+          if [[ ! -z '${{ matrix.target.arm }}' ]]; then
+            query="$query"' | .builds[0].goarm |= ["${{ matrix.target.arm }}"]'
           fi
+
+          # Create the tag locally so goreleaser knows what to target:
+          git tag "$VERSION"
```
cipherboy: This is nice; it should solve https://github.com/hashicorp/vault/issues/30685 and related issues.

changelog/3209.txt
```diff
@@ -0,0 +1,16 @@
...
+natively provides checksums available via both UI and API.
+```
+
+```release-note:change
```
cipherboy, line 6: Hmm, I'm curious what we think of this as a whole @openbao/dev-wg. I believe the former was chosen due to HashiCorp/distro naming compatibility. Thoughts?

changelog/3209.txt
```diff
@@ -0,0 +1,16 @@
...
+releases: Artifacts on GitHub now follow consistent naming across archives,
+SBOMs and signatures. Most notably, "x86_64" or "amd64" is now always "amd64",
+and the operating system is always lowercased.
+```
+
+```release-note:deprecation
+packaging/container: Architecture-specific container image tags such as
```
cipherboy, line 13: Do we expect any issues with this fwiw?

.github/workflows/release-images.yml
```diff
@@ -0,0 +1,258 @@
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4.0.0
+
+      - name: Set up Buildx
+        uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
+
+      - name: Install Cosign
+        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+
+      # Enable ghcr.io only on forks to ease testing, but enable all registries
```
cipherboy: 🔥

.github/workflows/release-images.yml
```diff
@@ -0,0 +1,258 @@
...
+            org.opencontainers.image.title=OpenBao
+            org.opencontainers.image.description=OpenBao is a tool for securely accessing secrets
+
+      - name: Sign images w/ Cosign
+        run: |
+          images=""
+          for tag in ${TAGS}; do
+            images+="${tag}@${DIGEST} "
+          done
+          cosign sign --yes ${images}
```
cipherboy, line 206: Do we not embed the cosign signature in the container image somehow? 🤔

cipherboy (commented):
Some random commentary. I think some are as a result of reading the changelog last, but yeah. 😆

.github/workflows/release-images.yml
satoqz: No, that's not how it works. Also see examples in https://github.com/sigstore/cosign-installer#usage, this is taken almost verbatim from their container example.

changelog/3209.txt
satoqz: I don't, because these tags only coexist with absolutely versioned tags, that is:

- `openbao:2.5.0-arm64` exists
- `openbao:2-arm64` does not exist
- `openbao:latest-arm64` does not exist

So this won't block version upgrades because you'd need to change the absolute version anyway if you were using one of these.

In terms of client support, you're doing something very odd if yours doesn't handle multi-arch manifests in 2026.

changelog/3209.txt
satoqz: > I kinda would like if we generate checksums and echo/cat them, so that we know they're consistent between action runs

When it comes to `@actions/upload-artifact` and `@actions/download-artifact`, these natively verify checksums. But yeah, probably good to do for those jobs that download from releases, I don't see anything in https://github.com/cli/cli that verifies the checksum.

> I recall there was some difference between our checksums and GitHub's.... Was one the checksum of the served zip?

Nope, they're the exact same - I checked the latest release.

changelog/3209.txt
satoqz: Even Vault uses the naming I propose here, see direct download links on:

https://developer.hashicorp.com/vault/install

https://releases.hashicorp.com/vault/2.0.1/vault_2.0.1_linux_amd64.zip
https://releases.hashicorp.com/vault/2.0.1/vault_2.0.1_linux_arm64.zip

goreleaser.hsm.yaml
satoqz: yes, see https://github.com/openbao/openbao/pull/3209#discussion_r3339587521, see #2456.

goreleaser.linux.yaml
satoqz: I need to ponder how to best do this. I don't want a checksum file + 2 signature files for each GOOS+GOARCH combination, so preferably I'd generate one consolidated checksums.txt for all artifacts in the accumulating job. Just can't have goreleaser conveniently do signatures then...

.github/workflows/release-images.yml
satoqz: To expand - because I realize now that it's not obvious at all - The `cosign sign` invocation here pushes the resulting signature to the respective registry as a separate OCI artifact.

pree:
How are container images handled during draft releases? Are they already pushed but not published on the container registries?

satoqz:
> How are container images handled during draft releases? Are they already pushed but not published on the container registries?

Since container registries aren't smart enough to do this kind of thing, we don't build or publish them at all until the draft release is undrafted. We could pre-build them and store the artifacts somewhere but that's really inconvenient.

karras:
Super cool, many thanks @satoqz!

goreleaser.linux.yaml
satoqz: Alright, checksums including both signatures are back, though as a single `checksums.txt`.

Nightly release: https://github.com/openbao/openbao-nightly/releases/tag/v2.6.0-nightly202606020949

changelog/3209.txt
satoqz: Alright, we now verify both checksums and cosign (including the cosign signature of the checksums) before further publishing binaries in `release-images.yml` and `release-packages.yml`.

satoqz:
So one interesting aspect of draft/immutable releases is that, since they push the tag post-publishing, you don't get a tagged OIDC identity during your build run.

This has implications for cosign signatures on our release artifacts. We'll get OIDC identities such as this:

```
https://github.com/openbao/openbao/.github/workflows/release.yml@refs/heads/main
https://github.com/openbao/openbao/.github/workflows/release.yml@refs/heads/release/2.5.x
```

instead of:

```
https://github.com/openbao/openbao/.github/workflows/release.yml@refs/tags/v2.0.0
```

I kind of hate it, but it seems fundamentally incompatible 🤷 We can't upload signatures post-publish if we make the release immutable.

Edit: Regardless, I'd argue that the difference is negligible as long as tags are fungible. As noted in fa1297401650bad1749b8d22be2cec6b8fb5776b's comment, I think the new behavior is then ultimately preferable because we have branch protection on `main` and `release/...`.

changelog/3209.txt
cipherboy: @satoqz Interesting! I wonder how we ended up with this then.

cipherboy:
@satoqz Do you have a repo with a scratch :sunglasses: chill release :sunglasses: ?

satoqz:
@cipherboy see latest 2 in https://github.com/openbao/openbao-nightly/releases.

I tested both GA releases and nightlies in our fork, too: https://github.com/Ki-Reply-GmbH/openbao/releases (those are the remains of a hundred or so since deleted test releases that came before, ofc :D)

.github/workflows/release-images.yml
cipherboy: @satoqz Ahh, that makes more sense. For some reason I was thinking it was embedded in the actual container image's OCI artifact.

goreleaser.linux.yaml
cipherboy: @satoqz I think you had removed the comment, but: 

> Though TBH, I'm not sure there's a valid reason to sign checksums if the artifact itself is signed.

If you download all artifacts and verify a single checksum file, you transitively verify all checksums as well. If we're doing cosign, that's many more signatures to verify via internet download versus a single signature over a checksum file, and more network fetches to go wrong.

My 2c. but it can be nice for a few cases.

satoqz:
(rebased because I caused trivial conflicts with https://github.com/openbao/openbao/commit/6c43210a35ca9ef307edc323bee16d231d91628e#diff-e79cf13ff1f1776c5ed4e6bcee9b43b646e04a59ca88df4733bb5c449ea9e8c8)

cipherboy:
@satoqz While we're breaking things, do you want to finish package naming changes and move `bao_...` -> `openbao_`, e.g. `bao-hsm_2.7.1_linux_arm64.tar.gz.sigstore.json`?

cipherboy (approved):
I think the proof ultimately will be in the release working, but from what I can tell, this will be a great improvement.

satoqz:
> @satoqz While we're breaking things, do you want to finish package naming changes and move `bao_...` -> `openbao_`, e.g. `bao-hsm_2.7.1_linux_arm64.tar.gz.sigstore.json`?

Yeah, looks a lot cleaner: https://github.com/Ki-Reply-GmbH/openbao/releases/tag/v2.7.2

Edit: hmm, also renames the binary, need to see how to best override that.

.github/workflows/release-images.yml
```diff
@@ -0,0 +1,282 @@
...
+              # Latest overall -> Add "latest".
+              if [[ "$(head -n 1 <<< "$releases")" == "$VERSION" ]]; then
+                tags=$(jq -c '["latest"] + .' <<< "$tags")
+              fi
+              ;;
+            *)
+              echo "::error::Malformed version: $version" && exit 1
+              ;;
+          esac
+          echo "tags=${tags}" >> "$GITHUB_OUTPUT"
```
JanMa: Can we move this into a dedicated file instead of 40 lines of in-line bash?

.github/workflows/release-images.yml
JanMa, line 1: Overall I think it would be nice if you could put the logic for individual steps in a Makefile or bash script so we could also easily build the images locally the same way they are built in CI

.github/workflows/release-packages.yml
JanMa, line 1: Same there, that much inline bash in a GitHub actions yaml feels icky. I'd prefer to move this into a Makefile or bash script

.github/workflows/release.yml
```diff
@@ -1,486 +1,400 @@
...
+          - { os: freebsd, arch: arm64       }
+          - { os: freebsd, arch: arm, arm: 6 }
+          - { os: netbsd,  arch: amd64       }
+          - { os: netbsd,  arch: arm64       }
+          - { os: netbsd,  arch: arm, arm: 6 }
+          - { os: openbsd, arch: amd64       }
+          - { os: openbsd, arch: arm64       }
+          - { os: openbsd, arch: arm, arm: 6 }
+          - { os: windows, arch: amd64       }
+          - { os: windows, arch: arm64       }
```
JanMa, line 138: <3 much nicer this way

JanMa (approved):
Lgtm, thanks a lot @satoqz

.github/workflows/release-images.yml
satoqz: Sure, pulled many steps out into scripts under `scripts/release/...`.

satoqz:
I'll merge this given we're happy overall.

Pushes from earlier today are were all just to factor out inline bash into scripts as requested and force-push fixing anything I broke along the way after running test releases. I'll take follow-ups if we want to further improve upon the scripts.

I also think we can remove Goreleaser entirely (or well, lower down to https://github.com/goreleaser/nfpm only, which is technically still Goreleaser, but not actually Goreleaser), but I'll do that another time, probably after removing the HSM distribution.

I'm fairly confident that it all works fine (and if not, is easily recoverable), though also glad that main is tracking a beta release next 😄 In the meantime, nightly releases will be much quicker to push out.

Issue #1573, opened 2025-07-23
Cleanup up existing linting issues


phyrog:
Follow-up of #1544, now that we have a golangci-lint config in the repo, we can clean up the existing findings.

There are these 7 bigger categories of existing findings:

- #1574 
- #1575 
- #1576 
- #1577 
- #1578 
- #1579 
- #1580 

As there are currently around 1300 findings, this issue should mainly serve as a tracking issues, I created sub-issues for the separate categories of findings for better overview.

cipherboy:
@phyrog Just wanted to comment that I appreciate this :-) Thank you!

PR #3210, opened 2026-06-01
Move module to v2, move packages to internal/

### LINKED ISSUE #3184: Move Go module to /v2

Go 1.24+ stamps the binaries with the full version of the module being built.
If the full version doesn't comply with the Go module version naming scheme, the datetime pseudoversion is used instead.
Openbao has used ldflags to set the version.
But many security scanners look at the mod buildinfo; and do not check the ldflags.

```
# go version -m bao  | grep -e 'mod\t' -e ldflags -e vcs
	mod	github.com/openbao/openbao	v0.0.0-20260520155348-4f6d47246a05	
	build	-ldflags="-X github.com/openbao/openbao/version.fullVersion=2.5.4 -X github.com/openbao/openbao/version.GitCommit=4f6d47246a053375271a5fd8af85c3b75695aa46 -X github.com/openbao/openbao/version.BuildDate=2026-05-20T16:08:53Z"
	build	vcs=git
	build	vcs.revision=4f6d47246a053375271a5fd8af85c3b75695aa46
	build	vcs.time=2026-05-20T15:53:48Z
	build	vcs.modified=false
```

Specifically it would be nice for the v0.0.0-20260520155348-4f6d47246a05 to be v2.5.4.

And security scanners like trivy believe that the openbao version is v0.0.0 and report that all builds of openbao binary are vulnerable to all CVEs that were long fixed by versions much earlier than v2.5.4.

I know that github.com/openbao/openbao module is private and should not be used by anything else.

However, it would help security vulnerability scanners such as trivy / govulncheck / osvscan if the top level go.mod would be updated from "module github.com/openbao/openbao" to "module github.com/openbao/openbao/v2", as then the module import path would match the go required module path and semantic versioning requirements, and then the produced openbao binary will have v2.5.4 as the mod version.

Alternative is to ask trivy, govulncheck, etc to support parsing ldflags extracting the version from github.com/openbao/openbao/version.fullVersion - for example syft/grype know how to do this. But this is niche, given that the intent of the go 1.24+ changes to buildinfo was to support version numbers in all binary builds.

Alternative is to ask go to consider stamping correct tags/versions in the mod buildinfo; even if the tags do not match the module import path & go semantic versioning information. Afterall the v0.0.0 pseudoversion is useless, as equivalent information is available in the vcs fields.

If you choose to rename the module to /v2, you will also be able to gain significant binary size savings.

With /v2 module name, the version number will be available in the mod buildinfo. It means that ldflags for github.com/openbao/openbao/version.fullVersion might no longer be needed (unless used for other purposes). And it also means that one can start building without that ldflags, and one can activate "-trimpath" which strips ldflags from buildinfo and reduces the binary size. Next one can also strip debug symbols (but keep symbols for eliminating false positives), to reduce bao binary size from 179MB to 124MB, a 30% saving.

**Additional context**
If this is of interest, I can propose a patch to fix this.
Unfortunately it will mean rewriting lots of import paths.

satoqz:
## Description

Staged for v2.7, currently just a draft preview of what this would look like because I was curious.

The first commit is the result of the following script:

```sh
go mod edit -module github.com/openbao/openbao/v2
go mod edit -dropretract '[v0.1.0, v1.17.0]'

internalize=(
    audit
    builtin
    command
    helper
    http
    physical
    plugins
    serviceregistration
    vault
    version
)

mkdir -p internal

for package in "${internalize[@]}"; do
    mv "$package" internal
done

mv internal/plugins/database internal/builtin/database

files=$(find . -name '*.go')

for package in "${internalize[@]}"; do
    sed -i -E "s|(\"github\.com/openbao/openbao/)($package)(/.*)?|\1v2/internal/\2\3|g" $files
done

sed -i -E "s|v2/internal/plugins/database|v2/internal/builtin/database|g" $files

go mod edit -replace=github.com/boltdb/bolt=./internal/helper/stubbolt

make fmt

git add --all
find internal/command -name '*.hcl' | xargs git add --force
```

I expect that to do most of the work so the bulk change can be verified via reproduction, plus some manual fixups to docs and tests that I'll add.

## Rationale

Resolves: #3184

## Acknowledgements

       also allowed.


changelog/3210.txt
cipherboy, line 1: If we wanted to, since it is terminal, we could leave `version/` alone.

cipherboy (commented):
I do rather like the new layout and the ability to clarify intent that server code is internal and not meant to be imported.

changelog/3210.txt
satoqz: @cipherboy we couldn't because it'd need to _at least_ move to `/v2` as well. So since it breaks either way, might as well move it to `internal/`, too.

cipherboy: @satoqz Ah, right. That's fair then.

satoqz:
@cipherboy let me know whenever we want to commit to this. I don't think there'll be an optimal time, but we don't actually have that many applicable code changes open right now. Files also get tagged as "moved", so Git should help with any rebases or backports.

cipherboy:
@satoqz Let's land after v2.6.0 GA is out?

satoqz:
I also used the opportunity to move `plugins/database` to `builtin/database`.

cipherboy:
After #3499 merged, this is safe to merge. I have verified the failing test passes locally and the linter failures are related to the fact _everything_ was modified at once here (see #1573).

PR #3212, opened 2026-06-01
Trigger mount reload on remounting

### LINKED ISSUE #3139: `sys/remount` of OIDC auth backend leaves raft follower nodes serving stale auth mount path until restart
We used to use the `auth/okta` OIDC backend to authenticate OpenBao logins against Okta. As part of the redesign, we decided to rename this endpoint to the `auth/oidc-okta`. We do use a load-balanced setup, where incoming connections to the F5 load-balancer are evenly distributed across all the backend nodes.

After remounting an OIDC auth backend from `auth/okta` to `auth/oidc-okta` in a raft-backed OpenBao cluster, the active leader immediately served the new auth path correctly, but follower nodes continued to expose the old auth path. The stale state affected both `bao auth list` and `/v1/sys/internal/ui/mounts`, causing the UI login page to intermittently use the removed `auth/okta` path depending on which node handled the request.

Restarting the follower nodes fixed the issue and made them converge to auth/oidc-okta.

**To Reproduce**  
Steps to reproduce the behaviour:

1.  Start with an HA OpenBao cluster using integrated raft storage, with an OIDC auth backend mounted at `auth/okta`.
2.  Remount the OIDC auth backend from `okta` to `oidc-okta`.
    
    In our case, this was performed by OpenTofu through the Vault/OpenBao provider.

    ```
    resource "vault_jwt_auth_backend" "okta_oidc" {
      description = "Okta OIDC Authentication"
      path        = var.auth_backend_path
      type        = "oidc"

      oidc_discovery_url = var.okta_client_url
      oidc_client_id     = var.okta_client_id
      oidc_client_secret = var.okta_client_secret
      default_role       = local.okta_meta_role_name

      tune {
        listing_visibility = "unauth"
      }

      lifecycle {
        ignore_changes = [
          oidc_client_secret,
          tune
        ]
      }
    }

     variable "auth_backend_path" {
       description = "Okta OIDC auth backend path"
       type        = string
    -  default     = "okta"
    +  default     = "oidc-okta"
     }
    ```
    The token policy for the pipeline has:
    ```
    path "sys/remount" {
      capabilities = ["update", "sudo"]
    }

    path "sys/remount/status/*" {
      capabilities = ["read"]
    }
    ```

The apply output showing the generated remount/update was:
    
    ```
    module.auth_oidc_okta.vault_jwt_auth_backend_role.okta_meta_role: Destroying... [id=auth/okta/role/okta_meta_role]
    module.auth_oidc_okta.vault_jwt_auth_backend_role.okta_meta_role: Destruction complete after 0s
    module.auth_oidc_okta.vault_jwt_auth_backend.okta_oidc: Modifying... [id=okta]
    module.auth_oidc_okta.vault_jwt_auth_backend.okta_oidc: Modifications complete after 2s [id=oidc-okta]
    module.auth_oidc_okta.vault_jwt_auth_backend_role.okta_meta_role: Creating...
    module.auth_oidc_okta.vault_jwt_auth_backend_role.okta_meta_role: Creation complete after 0s [id=auth/oidc-okta/role/okta_meta_role]
    Apply complete! Resources: 1 added, 3 changed, 1 destroyed.
    ```
    
3.  Query the unauthenticated UI mounts endpoint through the public/load-balanced address several times:
    
    `curl -s https://vault.example.net/v1/sys/internal/ui/mounts | jq '.data.auth'`
    
4.  Observe alternating responses:
    
    ```
    {
      "oidc-okta/": {
        "description": "Okta OIDC Authentication",
        "options": null,
        "type": "oidc"
      }
    }
    ```
    
    and:
    
    ```
    {
      "okta/": {
        "description": "Okta OIDC Authentication",
        "options": null,
        "type": "oidc"
      }
    }
    ```
    
5.  Query a follower node directly:
    
    ```
    bao-local auth list
    ```
    
    The follower still reports the old path:
    ```
    Path                   Type       Accessor                 Description                 Version
    ----                   ----       --------                 -----------                 -------
    approle/               approle    auth_approle_05206c90    Approle                     n/a
    gitlab/                jwt        auth_jwt_6b5278c1        n/a                         n/a
    okta/                  oidc       auth_oidc_f8e39e92       Okta OIDC Authentication    n/a
    token/                 token      auth_token_810e14a9      token based credentials     n/a
    ```
    
    
6.  On the same follower, reading the new mount path fails:
    
    ```
    bao-local read sys/auth/oidc-okta/tune
    ```
    
    ```
    Error reading sys/auth/oidc-okta/tune: Error making API request.

    URL: GET https://vault-5.example.net:8200/v1/sys/auth/oidc-okta/tune
    Code: 400. Errors:

    * cannot fetch sysview for path "auth/oidc-okta/"
    ```
    
7.  On the same follower, reading the old mount path succeeds:
    
    `bao-local read sys/auth/okta/tune`
    
    ```
    Key                             Value
    ---                             -----
    description                     Okta OIDC Authentication
    listing_visibility              unauth
    token_type                      default-service
    ...
    ```
    
8.  On the same follower, the UI mounts endpoint also returns the old path:
    
    `curl -sS https://vault-5.example.net:8200/v1/sys/internal/ui/mounts | jq '.data.auth'`
    
    ```
    {
      "okta/": {
        "description": "Okta OIDC Authentication",
        "options": null,
        "type": "oidc"
      }
    }
    ```
    
9.  Check raft peer state:
    
    `bao-local operator raft list-peers`
    
    ```
    Node                   Address                     State       Voter
    ----                   -------                     -----       -----
    vault-1.example.net    vault-1.example.net:8201    leader      true
    vault-2.example.net    vault-2.example.net:8201    follower    true
    vault-3.example.net    vault-3.example.net:8201    follower    true
    vault-4.example.net    vault-4.example.net:8201    follower    true
    vault-5.example.net    vault-5.example.net:8201    follower    true
    ```

10. Restart follower nodes.
11. After restart, followers converge and return `oidc-okta/` consistently.

**Expected behavior**

After the auth backend remount completes successfully on the active node, all raft follower nodes should converge on the same auth mount table without requiring a process restart.

All nodes should report the new mount path:

`oidc-okta/ oidc`

`sys/auth/oidc-okta/tune` should succeed on every node.

`sys/auth/okta/tune` should fail on every node.

The unauthenticated UI mounts endpoint should consistently return:

```
{
  "oidc-okta/": {
    "description": "Okta OIDC Authentication",
    "options": null,
    "type": "oidc"
  }
}
```

**Environment:**

- OpenBao Server Version (retrieve with bao status): 2.5.2, 2.5.3
- OpenBao CLI Version (retrieve with bao version): 2.5.3
- Server Operating System/Architecture: Oracle Enterprise Linux 8/Docker

`# Paste your OpenBao config here. # Be sure to scrub any sensitive values.`

**Additional context**  
The cluster had five raft voters:

```
vault-1.example.net    leader
vault-2.example.net    follower
vault-3.example.net    follower
vault-4.example.net    follower
vault-5.example.net    follower
```

The leader served the new `auth/oidc-okta` mount correctly. Followers continued to serve the old `auth/okta` mount until restarted.

This was visible through both normal auth mount inspection and the internal UI mounts endpoint, so it does not appear to be only a browser or UI cache issue.

The stale follower state caused the OpenBao UI login page to intermittently attempt OIDC login through:

`auth/okta/oidc/auth_url`

instead of:

`auth/oidc-okta/oidc/auth_url`

Restarting the follower nodes resolved the inconsistency.

wslabosz-reply:
## Description
- adds invalidation logic for a mount remount operation - essentially mount tune of a `path` field.
- adds `auth/` prefix to `bao auth move` command, previously could lead to incorrect remounts

## Rationale
without this change remounting a mount while running a read-only standby nodes would require the restart of said nodes to reflect the change in path of the mount

Resolves: #3139

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/command/auth_move_test.go
+++ b/internal/command/auth_move_test.go
@@ -46,7 +46,7 @@ func TestAuthMoveCommand_Run(t *testing.T) {
 		{
 			"non_existent",
 			[]string{"not_real", "over_here"},
-			"Error moving auth method not_real/ to over_here/",
+			"Error moving auth method auth/not_real/ to auth/over_here/",
 			2,
 		},
 	}
@@ -93,7 +93,7 @@ func TestAuthMoveCommand_Run(t *testing.T) {
 		}
 
 		code := cmd.Run([]string{
-			"auth/my-auth/", "auth/my-auth-2/",
+			"my-auth/", "my-auth-2/",
 		})
 		if exp := 0; code != exp {
 			t.Errorf("expected %d to be %d", code, exp)
@@ -125,7 +125,7 @@ func TestAuthMoveCommand_Run(t *testing.T) {
 		cmd.client = client
 
 		code := cmd.Run([]string{
-			"auth/my-auth/", "auth/my-auth-2/",
+			"my-auth/", "my-auth-2/",
 		})
 		if exp := 2; code != exp {
 			t.Errorf("expected %d to be %d", code, exp)
--- a/internal/vault/auth.go
+++ b/internal/vault/auth.go
@@ -322,7 +322,6 @@ func (c *Core) removeCredEntry(ctx context.Context, path string, updateStorage b
 }
 
 func (c *Core) removeCredEntryWithLock(ctx context.Context, path string, updateStorage bool) error {
-	// Taint the entry from the auth table
 	newTable := c.auth.ShallowClone()
 	entry, err := newTable.Remove(ctx, path)
 	if err != nil {
@@ -345,7 +344,7 @@ func (c *Core) removeCredEntryWithLock(ctx context.Context, path string, updateS
 	return nil
 }
 
-func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPathDetails, updateStorage bool) error {
+func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPathDetails) error {
 	ns, err := namespace.FromContext(ctx)
 	if err != nil {
 		return err
@@ -385,7 +384,7 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
 	}
 
 	// Mark the entry as tainted
-	if err := c.taintCredEntry(ctx, src.Namespace.ID, src.MountPath, updateStorage); err != nil {
+	if err := c.taintCredEntry(ctx, src.Namespace.ID, src.MountPath, true); err != nil {
 		return err
 	}
 
@@ -408,18 +407,16 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
 		return fmt.Errorf("path in use at %q", match)
 	}
 
-	mountEntry.Tainted = false
 	mountEntry.NamespaceID = dst.Namespace.ID
 	mountEntry.Namespace = dst.Namespace
 	srcPath := mountEntry.Path
 	mountEntry.Path = strings.TrimPrefix(dst.MountPath, routing.CredentialRoutePrefix)
 
-	// Update the mount table
+	// Update the auth table
 	if err := c.persistAuth(ctx, c.NamespaceView(mountEntry.Namespace), c.auth, &mountEntry.Local, mountEntry.UUID); err != nil {
 		mountEntry.Namespace = src.Namespace
 		mountEntry.NamespaceID = src.Namespace.ID
 		mountEntry.Path = srcPath
-		mountEntry.Tainted = true
 		c.authLock.Unlock()
 		return fmt.Errorf("failed to update auth table with error %w", err)
 	}
@@ -450,12 +447,13 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
 	}
 	c.authLock.Unlock()
 
-	// Un-taint the new path in the router
-	if err := c.router.Untaint(ctx, dstRelativePath); err != nil {
+	// Mark the entry as untainted.
+	if err := c.untaintCredEntry(ctx, dst.Namespace.ID, dst.MountPath, true); err != nil {
 		return err
 	}
 
-	return nil
+	// Un-taint the new path in the router.
+	return c.router.Untaint(ctx, dstRelativePath)
 }
 
 // taintCredEntry is used to mark an entry in the auth table as tainted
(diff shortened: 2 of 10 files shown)


internal/command/auth_move_test.go
```diff
@@ -93,7 +93,7 @@ func TestAuthMoveCommand_Run(t *testing.T) {
 		}
 
 		code := cmd.Run([]string{
-			"auth/my-auth/", "auth/my-auth-2/",
+			"my-auth/", "my-auth-2/",
```
cipherboy, line 96: @wslabosz-reply Does this preserve backwards compatibility and allow specifying the `auth/` prefix? 🤔 

I guess I don't know if `auth/auth/my-mount` works or not today as a mount point for an auth engine, and how we'd disambiguate it from `auth/my-mount` otherwise.

internal/vault/invalidation_test.go
cipherboy, line 1: One note on the tests: I think we only test migrating the mount table entry, not the mount data.

cipherboy (commented):
Looking at `vault/mount.go`, I'm not sure this is sufficient. When doing a cross-namespace remount, we incur a mount (data) move operation. This results in migrating individual storage entries, _after_ we update the mount table.

I think ideally we'd taint the mount (in storage), do the move, and then untaint the mount, just so we can:

1. Eventually allow us to fix storage if mount move gets interrupted (future bug fix)
2. Trigger final invalidation on the update succeeding.

In particular, I think in the interim (while data is moving and hasn't been invalidated on standby), we see data trickle into the mount individually, which will give inconsistent results w.r.t. primary (where the mount is blocked/tainted in memory).

internal/command/auth_move_test.go
wslabosz-reply: @cipherboy
it's always bad idea to hardcode strings like that (`auth`), but we've kinda inherited the bad so we have to roll with it.

> Does this preserve backwards compatibility and allow specifying the auth/ prefix? 🤔

only if your wish is to remount `auth/auth/something`, well technically it's never `auth/auth/something`, it's only the path registered within router

> I guess I don't know if auth/auth/my-mount works or not today as a mount point for an auth engine

yes, you can mount on that path

> how we'd disambiguate it from auth/my-mount otherwise.

it all falls down to the operator?

cipherboy:
@wslabosz-reply @satoqz Are we good to push this to v2.6.1+?

wslabosz-reply:
@cipherboy
If it won't be a blocker, I'd try to incorporate it with v2.5.5 tbh
although I've got yet to address your comment, but currently struggling trying to figure out correct order of operations

cipherboy:
@wslabosz-reply I think ideally I'd like to get 2.6 out the door this week if we can, so I suspect this will be a blocker.

satoqz:
Fine with me to postpone though I'll note I find v2.6 beta this week rather unrealistic unless we call it alpha instead.

internal/vault/invalidation_test.go
wslabosz-reply: @cipherboy need to adjust the test, but could you take a look at the current logic if it makes sense, I tried to adjust w.r.t. cross-namespace remounting

internal/vault/auth.go
```diff
@@ -450,33 +447,53 @@ func (c *Core) remountCredential(ctx context.Context, src, dst namespace.MountPa
 	}
 	c.authLock.Unlock()
 
-	// Un-taint the new path in the router
-	if err := c.router.Untaint(ctx, dstRelativePath); err != nil {
+	// Mark the entry as untainted.
+	if err := c.untaintCredEntry(ctx, dst.Namespace.ID, dst.MountPath, true); err != nil {
```
cipherboy, line 451: Given this wasn't happening before, does this mean we left around a tainted mount? 🤔

wslabosz-reply: @cipherboy 
no, I've changed the flow from:
**| active (standby) |**
1. taint source mount (mount modified event - taint existing mount)
2. persist table changes with tainted source mount and untainted destination mount (mount modified event - enable/create new mount)
3. delete source mount (mount deleted event - delete existing mount)

to

1. taint source mount (mount modified event - taint existing mount)
2. persist table changes with tainted source mount and **tainted destination mount (mount modified event - taint existing mount and preemptively taint routing to new mount which doesn't exist yet)**
3. delete source mount (mount deleted event - delete existing mount)
4. untaint destination mount (mount created event - create new mount, which will overwrite tainted status of the path in router)

PR #3218, opened 2026-06-02
Namespace CLI commands regression fix

wslabosz-reply:
## Description
- Added `key_threshold` to the namespace create response used by the `bao namespace create` command
- tweaked namespace api structs, added mapstructure struct tags for clearer code
- adjusted list and patch namespace code
- removed most of `api/sys_namespaces_test.go`
- brought back "namespace doesn't exist" response for namespace patch command

## Rationale
some of the namespace commands suffered regression, this aims at fixing that

removed tests were testing if mock http server responds with what was passed to it

Resolves: no issue

## Acknowledgements

       also allowed.


command/namespace_list.go
```diff
@@ -28,9 +28,9 @@ func (c *NamespaceListCommand) Help() string {
 	helpText := `
 Usage: bao namespace list [options]
 
-  Lists the enabled child namespaces.
+  Lists unsealed child namespaces.
```
cipherboy: Not sure what I think the best wording is. There will be non-sealed child namespaces, which I guess, by inheritance are unsealed, but aren't intentionally sealable (and thus in either the sealed or unseal) state.

Maybe just drop the modifier altogether? 🤔 "List child namespaces" doesn't advertise any warranty either way?

cipherboy: At the list level, aren't all child namespaces visible, regardless of whether they're sealed or unsealed for that matter?

It is only when you get to scan that it matters.

api/sys_namespaces_test.go
```diff
@@ -67,402 +57,3 @@ func TestDeleteNamespaceValidation(t *testing.T) {
 		t.Error("expected error for empty path, got nil")
 	}
 }
-
-func TestCreateNamespace(t *testing.T) {
```
cipherboy, line 71: Not sure what I think of this, thoughts from others? Ultimately what the server actually sends is the real test I guess.

command/namespace_list.go
satoqz: > At the list level, aren't all child namespaces visible, regardless of whether they're sealed or unsealed for that matter?

I thought so too, so I'd drop the wording for this command. It matters for `bao namespace scan`, where we can just leave a note along the lines of "note that not all namespaces may show if they're inaccessible because a parent namespace is sealed". Currently it says "enabled namespaces" which is wording I don't like.

api/sys_namespaces_test.go
satoqz: Wojciech mentioned that he figured these tests were a little pointless. I'd agree and say `external_tests/api` should be the way to test code in `api/`.

cipherboy: Very good, I concur then.

satoqz (approved):

cipherboy (approved):
Thanks @wslabosz-reply!

satoqz:
It doesn't make so much sense that key shares and threshold show up in tables in either case:

```console
$ bao namespace create -key-shares=3 -key-threshold=2 barr
Unseal Key 1: 097bce12b581ce6ba4cee6ff54f061a81cbb81d12bc7af875455d1c40fd4a7767f
Unseal Key 2: 8cef989cb3c15910da66e36e6d757ded6423cf8e5075f728938a3426c1a8444faf
Unseal Key 3: ca253bc2e152ba3d8cae98df1433bb05604d3ff27d29375c81cb258b0c39073663

Namespace initialized with 3 key shares and a key threshold of 2. Please
securely distribute the key shares printed above. When the namespace is
re-sealed, you must supply at least 2 of these keys to unseal it.

Key                Value
---                -----
custom_metadata    map[]
id                 khLfWu
key_shares         [097bce12b581ce6ba4cee6ff54f061a81cbb81d12bc7af875455d1c40fd4a7767f 8cef989cb3c15910da66e36e6d757ded6423cf8e5075f728938a3426c1a8444faf ca253bc2e152ba3d8cae98df1433bb05604d3ff27d29375c81cb258b0c39073663]
key_threshold      2
locked             false
path               barr/
tainted            false
uuid               3cdc054e-240d-3fda-7de2-c40c1b87949e
```

```console
$ bao namespace create bar
Key                Value
---                -----
custom_metadata    map[]
id                 RYeCOy
key_shares         []
key_threshold      0
locked             false
path               bar/
tainted            false
uuid               4343be65-8102-444c-5c6b-37164f3e1c30
```

cipherboy (approved):
LGTM! Thanks @satoqz for the updates and @wslabosz-reply for the fixes.

PR #3224, opened 2026-06-02
GRPC Synchronization of Namespace Seal Root Keys

cipherboy:
## Description

This implements cross-cluster namespace seal synchronization, ensuring
that namespace operators only need to unseal once. This:

1. Adjusts namespace seal status changes to apply on the active node.
2. Writes namespace seal status into storage to be used as invalidation
   on manual seals/unseals. This is important for the future work on
   auto-unseal, as we need to ensure not to auto-unseal a manually
   sealed namespace (even between server restarts) and wait for manual
   recovery keys to be provided.
3. Builds a GRPC synchronization mechanism for root namespace keys;
   while the GRPC mechanism is protected, we encrypt with the root
   namespace's barrier keyring to ensure protection in transit.
   This is limited to keys which need to be synchronized and already
   synchronized keys do not re-appear on the wire.
4. Add a test validating that namespace synchronization behaves as
   expected.

## Rationale

Related: #1357

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/barrier/aes_gcm.go
+++ b/vault/barrier/aes_gcm.go
@@ -4,6 +4,7 @@
 package barrier
 
 import (
+	"cmp"
 	"context"
 	"crypto/aes"
 	"crypto/cipher"
@@ -13,6 +14,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"path"
 	"strconv"
 	"strings"
 	"sync"
@@ -20,6 +22,7 @@ import (
 	"time"
 
 	metrics "github.com/hashicorp/go-metrics/compat"
+	"github.com/openbao/openbao/helper/namespace"
 	"github.com/openbao/openbao/sdk/v2/logical"
 	"github.com/openbao/openbao/sdk/v2/physical"
 )
@@ -64,6 +67,7 @@ var (
 // and integrity.
 type AESGCMBarrier struct {
 	backend    physical.Backend
+	namespace  *namespace.Namespace
 	metaPrefix string
 
 	l        sync.RWMutex
@@ -128,16 +132,20 @@ func (b *AESGCMBarrier) SetRotationConfig(ctx context.Context, rotConfig KeyRota
 
 // NewAESGCMBarrier is used to construct a new barrier that uses
 // the provided physical backend for storage.
-func NewAESGCMBarrier(storage physical.Backend, metaPrefix string) SecurityBarrier {
+func NewAESGCMBarrier(storage physical.Backend, ns *namespace.Namespace) SecurityBarrier {
 	b := &AESGCMBarrier{
 		backend:                  storage,
-		metaPrefix:               metaPrefix,
+		namespace:                cmp.Or(ns, namespace.RootNamespace),
 		sealed:                   true,
 		cache:                    make(map[uint32]cipher.AEAD),
 		currentAESGCMVersionByte: byte(AESGCMVersion2),
 		UnaccountedEncryptions:   &atomic.Int64{},
 		RemoteEncryptions:        &atomic.Int64{},
 		totalLocalEncryptions:    &atomic.Int64{},
+	}
+
+	if ns != nil && ns.UUID != namespace.RootNamespaceUUID {
+		b.metaPrefix = path.Join(NamespacePrefix, ns.UUID) + "/"
 	}
 
 	if _, ok := storage.(physical.TransactionalBackend); ok {
@@ -1280,6 +1288,10 @@ func (b *AESGCMBarrier) encryptions() int64 {
 
 func (b *AESGCMBarrier) SetReadOnly(readOnly bool) {
 	b.readOnly.Store(readOnly)
+}
+
+func (b *AESGCMBarrier) Namespace() *namespace.Namespace {
+	return b.namespace
 }
 
 func (b *TransactionalAESGCMBarrier) BeginReadOnlyTx(ctx context.Context) (logical.Transaction, error) {
--- a/vault/barrier/barrier.go
+++ b/vault/barrier/barrier.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/openbao/openbao/helper/namespace"
 	"github.com/openbao/openbao/sdk/v2/logical"
 )
 
@@ -79,6 +80,10 @@ const (
 	// SystemBarrierPrefix is the prefix used for the
 	// system logical backend.
 	SystemBarrierPrefix = "sys/"
+
+	// NamespacePrefix is the prefix to the UUID of a namespace
+	// used in the barrier view for the namespace-owned backends.
+	NamespacePrefix = "namespaces/"
 )
 
 const (
@@ -178,6 +183,9 @@ type SecurityBarrierCore interface {
 	// SetReadOnly allows marking storage as read-only; this is useful for
 	// HA mode but could be more broadly useful.
 	SetReadOnly(readOnly bool)
+
+	// Namespace returns the namespace related to this barrier.
+	Namespace() *namespace.Namespace
 }
 
 // SecurityBarrier is a critical component of Vault. It is used to wrap
--- a/vault/core.go
+++ b/vault/core.go
@@ -1073,7 +1073,7 @@ func NewCore(conf *CoreConfig) (*Core, error) {
 	}
 
 	// Construct a new AES-GCM barrier
-	c.barrier = barrier.NewAESGCMBarrier(c.physical, "")
+	c.barrier = barrier.NewAESGCMBarrier(c.physical, nil)
 	c.SetupSealManager()
 
 	// We create the funcs here, then populate the given config with it so that
--- a/vault/core.go
+++ b/vault/core.go
@@ -555,7 +555,8 @@ type Core struct {
 
 	quotaManager *quotas.Manager
 
-	clusterHeartbeatInterval time.Duration
+	clusterHeartbeatInterval     time.Duration
+	clusterNamespaceSyncInterval time.Duration
 
 	// activeTime is set on active nodes indicating the time at which this node
 	// became active.
@@ -751,7 +752,8 @@ type CoreConfig struct {
 
 	ClusterNetworkLayer cluster.NetworkLayer
 
-	ClusterHeartbeatInterval time.Duration
+	ClusterHeartbeatInterval     time.Duration
+	ClusterNamespaceSyncInterval time.Duration
 
 	// number of workers to use for lease revocation in the expiration manager
 	NumExpirationWorkers int
@@ -858,6 +860,15 @@ func CreateCore(conf *CoreConfig) (*Core, error) {
 		clusterHeartbeatInterval = 5 * time.Second
 	}
 
+	clusterNamespaceSyncInterval := conf.ClusterNamespaceSyncInterval
+	if clusterNamespaceSyncInterval == 0 {
+		// We don't want namespaces to take forever to unseal, but we also
+		// want to avoid spamming the leader. This seems like a reasonable
+		// middle ground. By tying it to clusterHeartbeatInterval, tests
+		// can run faster automatically.
+		clusterNamespaceSyncInterval = 3 * clusterHeartbeatInterval
+	}
+
 	if conf.NumExpirationWorkers == 0 {
 		conf.NumExpirationWorkers = numExpirationWorkersDefault
 	}
@@ -925,6 +936,7 @@ func CreateCore(conf *CoreConfig) (*Core, error) {
 		raftJoinDoneCh:                 make(chan struct{}),
 		pendingRaftPeerChallengeKey:    make([]byte, 32),
 		clusterHeartbeatInterval:       clusterHeartbeatInterval,
+		clusterNamespaceSyncInterval:   clusterNamespaceSyncInterval,
 		numExpirationWorkers:           conf.NumExpirationWorkers,
 		raftFollowerStates:             raft.NewFollowerStates(),
 		disableAutopilot:               conf.DisableAutopilot,
@@ -3865,10 +3877,11 @@ func (c *Core) refreshRequestForwardingConnection(ctx context.Context, clusterAd
 	c.rpcForwardingClient = forwarding.NewClient(
 		c,
 		forwarding.NewRequestForwardingClient(c.rpcClientConn),
-		time.NewTicker(c.clusterHeartbeatInterval),
 		dctx,
+		time.NewTicker(c.clusterHeartbeatInterval),
+		time.NewTicker(c.clusterNamespaceSyncInterval),
 	)
-	c.rpcForwardingClient.StartHeartbeat()
+	c.rpcForwardingClient.Start()
 
 	return nil
 }
--- a/vault/forwarding/request_forwarding_rpc.go
+++ b/vault/forwarding/request_forwarding_rpc.go
@@ -240,6 +240,7 @@ func (c *Client) Start() {
 				c.core.Logger().Warn("another namespace key synchronization is still running")
 				return
 			}
+			defer syncRunning.Store(false)
 
 			c.core.Logger().Info("synchronizing namespace keys with active node")
 
@@ -249,8 +250,6 @@ func (c *Client) Start() {
 			defer cancel()
 
 			c.SynchronizeKeys(ctx)
-
-			syncRunning.Store(false)
 		}
 
 		tick()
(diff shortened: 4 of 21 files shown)


vault/seal_manager.go
```diff
@@ -1438,3 +1446,232 @@ func (sm *SealManager) DeleteRotationBackup(ctx context.Context, nsPath string,
...
+// with the barrier keyring. This ensures that, as long as the root barrier
+// keyring's integrity remains, even keys over a compromised GRPC connection
+// should remain secure, though we enforce TLS for GRPC everywhere.
+func (c *Core) NamespaceKeys(rpcCtx context.Context, namespaces []string) (map[string][]byte, error) {
+	c.stateLock.RLock()
+	defer c.stateLock.RUnlock()
+
+	activeCtx := c.activeContext.Load()
+	if c.Sealed() || activeCtx == nil || activeCtx.Err() != nil {
+		return nil, errors.New("instance is sealed")
```
cipherboy: Do we have better ideas for this error message? It is a rough proxy for "sealed or not at least a read-enabled standby".

Thoughts?

vault/seal_manager.go
```diff
@@ -606,3 +614,229 @@ func (sm *SealManager) InitializeBarrier(ctx context.Context, ns *namespace.Name
...
+	for uuid, b := range sm.barrierByNamespace {
+		if !b.Sealed() && uuid != namespace.RootNamespaceUUID {
+			namespaces = append(namespaces, uuid)
+		}
+	}
+
+	return namespaces
+}
+
+// NamespacesWithKeys is a list of namespace UUIDs which are currently sealed.
```
wslabosz-reply:
```suggestion
// NamespacesMissingKeys is a list of namespace UUIDs which are currently sealed.
```

vault/namespace_store.go
```diff
@@ -1033,14 +1043,48 @@ func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSe
...
+	if err != nil {
+		return fmt.Errorf("failed to get parent namespace from context: %w", err)
+	}
+
+	// Now modify just this one namespace in storage to mark it sealed,
+	// forgetting the keys from all other nodes.
+	namespaceToSeal.ManuallySealed = true
+	nsCopy := namespaceToSeal.Clone(true /* preserve unlock */)
+	if err := ns.writeNamespace(ctx, ns.core.NamespaceView(parent), nsCopy); err != nil {
+		return fmt.Errorf("failed to persist namespace seal: %w", err)
```
```suggestion
		return fmt.Errorf("failed to persist namespace: %w", err)
```

vault/namespace_store.go
```diff
@@ -1053,6 +1097,22 @@ func (ns *NamespaceStore) UnsealNamespace(ctx context.Context, path string, key
...
+	if !namespaceToUnseal.HasParent(parent) {
+		return fmt.Errorf("given parent namespace from context is not the parent of this namespace")
+	}
+
+	// Now modify just this one namespace in storage to mark it unsealed,
+	// letting other nodes unseal it as well.
+	namespaceToUnseal.ManuallySealed = false
+	nsCopy := namespaceToUnseal.Clone(true /* preserve unlock */)
+	if err := ns.writeNamespace(ctx, ns.core.NamespaceView(parent), nsCopy); err != nil {
+		return fmt.Errorf("failed to modify namespace seal: %w", err)
```
```suggestion
		return fmt.Errorf("failed to modify namespace: %w", err)
```

vault/seal_manager.go
```diff
@@ -48,9 +49,10 @@ type SealManager struct {
 	// invalidated atomic.Bool
 
 	sealByNamespace              map[string]Seal
+	barrierByNamespace           map[string]barrier.SecurityBarrier
 	unlockInformationByNamespace map[string]*unlockInformation
 	rotationConfigByNamespace    map[string]*rotationConfig
-	barrierByNamespace           *radix.Tree
+	barrierByNamespacePath       *radix.Tree
```
satoqz, line 54: You know, an `AESGCMBarrier` already knows the namespace UUID it operates on via `metaPrefix`, so I will be removing the extra lookup table for the second time once I get to refactor things and expose that :P

cipherboy: @satoqz Ok, I've removed it again.

vault/barrier/barrier.go
```diff
@@ -178,6 +178,10 @@ type SecurityBarrierCore interface {
 	// SetReadOnly allows marking storage as read-only; this is useful for
 	// HA mode but could be more broadly useful.
 	SetReadOnly(readOnly bool)
+
+	// MetaPrefix returns the prefix applied to all storage operations;
+	// only empty for non-root namespace barriers.
```
wslabosz-reply: not sure what you mean, it's empty for *root* namespace barrier and non-sealable namespaces with a parent of *root* namespace

satoqz: It also is not applied to all storage operations but only those operating on barrier/seal related paths IIRC.

My ultimate idea for the refactor was to pass a `namespace.Namespace` to barrier and fully internalize the `metaPrefix` detail, then add a `Namespace()` method to read it back as needed. But I'm fine with anything for this PR.

cipherboy: @satoqz @wslabosz-reply I ended up doing the `Namespace()` refactor. It is a bit messy in places but I think it is fine ultimately.

vault/barrier/barrier.go
```diff
@@ -79,6 +80,10 @@ const (
 	// SystemBarrierPrefix is the prefix used for the
 	// system logical backend.
 	SystemBarrierPrefix = "sys/"
+
+	// NamespacePrefix is the prefix to the UUID of a namespaces
```
satoqz:
```suggestion
	// NamespacePrefix is the prefix to the UUID of a namespace
```

vault/barrier/aes_gcm.go
```diff
@@ -128,10 +131,10 @@ func (b *AESGCMBarrier) SetRotationConfig(ctx context.Context, rotConfig KeyRota
 
 // NewAESGCMBarrier is used to construct a new barrier that uses
 // the provided physical backend for storage.
-func NewAESGCMBarrier(storage physical.Backend, metaPrefix string) SecurityBarrier {
+func NewAESGCMBarrier(storage physical.Backend, ns *namespace.Namespace) SecurityBarrier {
 	b := &AESGCMBarrier{
 		backend:                  storage,
-		metaPrefix:               metaPrefix,
+		namespace:                ns,
```
satoqz:
```suggestion
		namespace:                cmp.Or(ns, namespace.RootNamespace),
```

Can we do something like this so we can drop "if applicable" from

```go
	// Namespace returns the namespace related to this barrier, if applicable.
	Namespace() *namespace.Namespace
```

vault/forwarding/request_forwarding_rpc.go
```diff
@@ -170,23 +226,136 @@ func (c *Client) StartHeartbeat() {
...
+			c.core.Logger().Info("synchronizing namespace keys with active node")
+
+			// We don't want to hang forever waiting to synchronize namespace
+			// seal keys with the active node. Bind it to a reasonable length.
+			ctx, cancel := context.WithTimeout(c.taskContext, 2*time.Minute)
+			defer cancel()
+
+			c.SynchronizeKeys(ctx)
+
+			syncRunning.Store(false)
```
satoqz: This can be deferred.

vault/forwarding/request_forwarding_rpc.go
```diff
@@ -170,23 +226,136 @@ func (c *Client) StartHeartbeat() {
...
 			}
 		}
 	}()
 }
+
+func (c *Client) SynchronizeKeys(ctx context.Context) {
+	var wg sync.WaitGroup
+	wg.Go(func() {
+		if err := c.shipKeys(ctx); err != nil {
+			c.core.Logger().Error("got error sending keys", "error", err)
```
satoqz:
```suggestion
			c.core.Logger().Error("error sending namespace keys", "error", err)
```

vault/forwarding/request_forwarding_rpc.go
```diff
@@ -170,23 +226,136 @@ func (c *Client) StartHeartbeat() {
...
+	wg.Go(func() {
+		if err := c.getKeys(ctx); err != nil {
+			c.core.Logger().Error("got error receiving keys", "error", err)
+		}
+	})
+
+	wg.Wait()
+}
+
+func (c *Client) shipKeys(ctx context.Context) error {
```
satoqz, line 291: Very enthusiastic function name.

vault/seal_manager.go
```diff
@@ -606,3 +608,249 @@ func (sm *SealManager) InitializeBarrier(ctx context.Context, ns *namespace.Name
...
+
+	var namespaces []string
+	sm.barrierByNamespacePath.Walk(func(_ string, b interface{}) bool {
+		if b == nil {
+			return false
+		}
+
+		nsBarrier := b.(barrier.SecurityBarrier)
+		ns := nsBarrier.Namespace()
+		if !nsBarrier.Sealed() || ns == nil || ns.UUID == namespace.RootNamespaceID {
```
```suggestion
		if !nsBarrier.Sealed() || ns == nil || ns.ID == namespace.RootNamespaceID {
```

In `WithKeys()`, too.

Could also drop the nil check with https://github.com/openbao/openbao/pull/3224/changes#r3390302205 in place.

vault/seal_manager.go
```diff
@@ -606,3 +608,249 @@ func (sm *SealManager) InitializeBarrier(ctx context.Context, ns *namespace.Name
...
+}
+
+// GetRootKey yields the underlying root key of the barrier for the given
+// namespace.
+func (sm *SealManager) GetRootKey(ctx context.Context, ns *namespace.Namespace) ([]byte, error) {
+	if ns.ID == namespace.RootNamespaceID {
+		return nil, errors.New("refusing to return root namespace's root key")
+	}
+
+	path, v, exists := sm.barrierByNamespacePath.LongestPrefix(ns.Path)
```
satoqz: Needs read lock.

vault/forwarding/request_forwarding_rpc.go
```diff
@@ -170,23 +226,136 @@ func (c *Client) StartHeartbeat() {
...
+	response, err := c.AdvertiseNamespaceKeys(ctx, advertisement)
+	if err != nil {
+		return fmt.Errorf("failed to send unsealed namespace advertisement to active node: %v", err)
+	}
+
+	if len(response.Namespaces) == 0 {
+		return nil
+	}
+
+	keys, getNamespaceErr := c.core.NamespaceKeys(ctx, response.Namespaces)
```
satoqz:
```suggestion
	keys, errs := c.core.NamespaceKeys(ctx, response.Namespaces)
```

Perhaps this can be `errs` so we don't ultimately return `getNamespaceErr` just because that's what we're collecting into.

vault/seal_manager.go
satoqz: Why don't we use `Get()`?

vault/forwarding/request_forwarding_service.proto
satoqz, line 1: Some parts in this file were already inconsistent, but the codebase uses tab indents for protobuf. The ones you added are 4 spaces 😄

vault/forwarding/request_forwarding_service.proto
```diff
@@ -52,7 +52,49 @@ message ClientKey {
...
+    // standby->active; used when the standby was formerly the active node
+    // and a new active takes over without knowledge of past keys. This is
+    // a two-step process to avoid sending keys which the active node
+    // already knows about.
+    rpc AdvertiseNamespaceKeys(AdvertiseNamespaceKeysRequest) returns (AdvertiseNamespaceKeysReply) {};
+    rpc SendNamespaceKeys(SendNamespaceKeysRequest) returns (SendNamespaceKeysReply) {};
+
+    // active->standby; used by the standby to refresh keys.
+    rpc GetNamespaceKeys(GetNamespaceKeysRequest) returns (GetNamespaceKeysReply) {};
 }
```
satoqz: It would be good to mention that the RPCs themselves always go `standby->active`, to differentiate between the direction that the keys are taking and the direction that the RPC goes.

satoqz (commented):
Random comments go! Will do more review tomorrow.

vault/barrier/aes_gcm.go
cipherboy: Generics are nifty.

vault/forwarding/request_forwarding_rpc.go
cipherboy: I suppose, less pressing though as there's no error paths in that function.

vault/forwarding/request_forwarding_rpc.go
cipherboy: :ship:

vault/seal_manager.go
cipherboy: @satoqz Ah, radix.Tree is not thread safe, I see.

vault/forwarding/request_forwarding_service.proto
cipherboy: I added that as a note on the service itself and expanded on naming there.

vault/core.go
```diff
@@ -858,6 +860,15 @@ func CreateCore(conf *CoreConfig) (*Core, error) {
 		clusterHeartbeatInterval = 5 * time.Second
 	}
 
+	clusterNamespaceSyncInterval := conf.ClusterNamespaceSyncInterval
+	if clusterNamespaceSyncInterval == 0 {
+		// We don't want namespaces to take forever to unseal, but we also
+		// want to avoid spamming the leader. This seems like a reasonable
+		// middle ground. By tying it to clusterHeartbeatInterval, tests
+		// can run faster automatically.
+		clusterNamespaceSyncInterval = 3 * clusterHeartbeatInterval
```
wslabosz-reply, line 869: @cipherboy 
I wonder if we should do something about it, by default it will:
- log every 15 seconds on every standby node
- log every 15 seconds x # of standby nodes

that's quite a lot, maybe not too much for "healthcheck" kind of logging, but still a lot, maybe changing the log level of these logs would be a good idea?

vault/namespace_store.go
```diff
@@ -1053,6 +1093,22 @@ func (ns *NamespaceStore) UnsealNamespace(ctx context.Context, path string, key
 		return errors.New("cannot unseal root namespace with this operation")
 	}
 
+	if !namespaceToUnseal.HasParent(parent) {
+		return fmt.Errorf("given parent namespace from context is not the parent of this namespace")
```
```suggestion
		return fmt.Errorf("namespace retrieved from context is not the parent of this namespace")
```

vault/seal_manager.go
```diff
@@ -50,7 +51,7 @@ type SealManager struct {
 	sealByNamespace              map[string]Seal
 	unlockInformationByNamespace map[string]*unlockInformation
 	rotationConfigByNamespace    map[string]*rotationConfig
-	barrierByNamespace           *radix.Tree
+	barrierByNamespacePath       *radix.Tree
```
wslabosz-reply, line 54: I don't think it makes a difference in context of changes introduced in this PR, but we could revert to old name

wslabosz-reply (dismissed):
tested it locally, haven't found anything wrong with the logic, although need to folllow-up with the invalidation mgr adjustment to not reload the standby nodes on every mount write (basically namespace creation, seal or unseal).

vault/core.go
cipherboy: @wslabosz-reply I adjusted the info logs down to be less frequent.  I don't think we want to increase the delay as that affects how quickly nodes stabilize on unseal keys, which impacts applications.

vault/seal_manager.go
cipherboy: I think I've done this a few times... I still think its surprising that a `<x>ByNamespace` is a `radix.Tree` rather than a `map`. Putting `Path` makes it less surprising that you would be using a path data structure rather than a lookup one? :shrug:  My 2c. at any rate. 

```
vault/identity/store_util.go:2439:	byNamespace := make(map[string]int)
vault/seal_manager.go:51:	sealByNamespace              map[string]Seal
vault/seal_manager.go:52:	unlockInformationByNamespace map[string]*unlockInformation
vault/seal_manager.go:53:	rotationConfigByNamespace    map[string]*rotationConfig
vault/seal_manager.go:100:	sm.unlockInformationByNamespace = map[string]*unlockInformation{}
vault/seal_manager.go:101:	sm.rotationConfigByNamespace = map[string]*rotationConfig{}
```

cipherboy:
Fix for the data race in test 0 should be #3282.

vault/seal_manager.go
wslabosz-reply: to be honest I'm with you on the rename, just thought it doesn't have to be in this PR, but as you've done that already it's okay to keep.

satoqz (approved):
Makes sense overall though I wonder if `ManuallySealed` can go elsewhere, like the seal config.

cipherboy:
> Makes sense overall though I wonder if `ManuallySealed` can go elsewhere, like the seal config.

Interesting thought for the future. I think `ManuallySealed` really only applies to namespaces; though I could see putting it on seal configs in the future. I was just thinking we already had the namespace reference lying around anyways. :thinking:

satoqz:
I just want to gatekeep the `Namespace` struct that we pass all over the place from having a million fields that we rarely need.

PR #3230, opened 2026-06-03
Clean up nits in rotation code

cipherboy:
## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3232, opened 2026-06-03
Prevent guarded logging

cipherboy:
## Description

See discussion in #3230 and elsewhere.

This removes all guarded logging calls, introducing a semgrep rule intended to prevent it going forward. 

## Acknowledgements

       also allowed.


satoqz:
You can always count on CodeQL 🤗

tools/semgrep/ci/logger-level-check.yml
```diff
@@ -0,0 +1,53 @@
...
+        - pattern: |
+            if $LOGGER.IsWarn() {
+              $LOGGER.Warn("...", ...)
+            }
+        - pattern: |
+            if $LOGGER.IsError() {
+              $LOGGER.Error("...", ...)
+            }
+
+    message: "Log guarded by level check of the same level"
```
satoqz, line 24: Can semgrep do `Is*`? A different level would be even more concerning.

tools/semgrep/ci/logger-level-check.yml
```diff
@@ -0,0 +1,53 @@
...
+    languages: [go]
+    severity: ERROR
+  
+
```
satoqz:
```suggestion
```

satoqz (dismissed):

tools/semgrep/ci/logger-level-check.yml
phyrog: Should be possible with something like this:

```yaml
      - pattern: |
          if $LOGGER.$CHECK() {
            $LOGGER.$LOG(...)
          }

      - metavariable-regex:
          metavariable: $CHECK
          regex: '^Is(Debug|Info|Warn|Error|Trace|Fatal)$'

      - metavariable-regex:
          metavariable: $LOG
          regex: '^(Debug|Info|Warn|Error|Trace|Fatal)$'
```

JanMa (dismissed):
Lgtm, not too familiar with semgrep rules, so I'm not sure if this could be expressed more elegantly 😄

tools/semgrep/ci/logger-level-check.yml
cipherboy: @phyrog I ended up having to create another regex for `$LOGGER` as otherwise we'll match patterns like:

```go
if resp.IsError() {
    return nil, fmt.Errorf("something", resp.Error())
}
```

due to `$LOGGER` not actually being a single static value (which seems rather confusing to me, tbh).

satoqz (approved):

satoqz:
Seems good to go, merging!

PR #3240, opened 2026-06-05
Propagate JSON parse errors consistently

### LINKED ISSUE #2996: "error parsing JSON" for large json payloads
**Describe the bug**

Currently openbao is just returning `error parsing JSON` when either `max_request_json_memory` or `max_request_json_strings` is hit when updating a secret using a json payload.

It took me longer than I'd like to admit to find out what the actual error was.

**To Reproduce**

```sh
# run openbao in dev server mode
openbao server -dev

bao secrets enable -path=example kv-v2
bao kv put -mount=example test @large.json
```

**Expected behavior**

Provide a error message pointing to the two configuration options.

**Environment:**
-

```hcl
-
```

**Additional context**

satoqz:
## Description

Always propagate the full error to the response, like was already the case for several one-off endpoints (`http/sys_*.go` and friends) but not for "logical" requests. See discussion in https://github.com/openbao/openbao/issues/2996.

Also clean up error wrapping to the point that we don't include duplicate information, such as "error parsing JSON: failed to parse JSON: ...".

## Rationale

Resolves: https://github.com/openbao/openbao/issues/2996

## Acknowledgements

       also allowed.

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -11,7 +11,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"io/fs"
 	"maps"
 	"mime"
@@ -858,33 +857,6 @@ func parseQuery(values url.Values) map[string]interface{} {
 		return data
 	}
 	return nil
-}
-
-func parseJSONRequest(r *http.Request, w http.ResponseWriter, out interface{}) error {
-	ctx := r.Context()
-
-	// Enforce limits on JSON complexity.
-	_, _, err := EnforceJSONComplexityLimits(ctx, r.Body)
-	if err != nil {
-		return fmt.Errorf("failed to limit JSON input: %w", err)
-	}
-
-	// Reset to the beginning.
-	if err := resetBody(r); err != nil {
-		return fmt.Errorf("failed to reset body: %w", err)
-	}
-
-	err = jsonutil.DecodeJSONFromReader(r.Body, out)
-	if err != nil && err != io.EOF {
-		return fmt.Errorf("failed to parse JSON input: %w", err)
-	}
-
-	// Reset to the beginning again.
-	if err := resetBody(r); err != nil {
-		return fmt.Errorf("failed to reset body: %w", err)
-	}
-
-	return err
 }
 
 // parseFormRequest parses values from a form POST.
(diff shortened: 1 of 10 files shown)


http/handler.go
```diff
@@ -860,21 +860,28 @@ func parseQuery(values url.Values) map[string]interface{} {
...
-	if err != nil {
-		return fmt.Errorf("failed to limit JSON input: %w", err)
+	if _, _, err := EnforceJSONComplexityLimits(ctx, r.Body); err != nil {
+		switch {
+		case errors.Is(err, ErrJSONExceededMemory), errors.Is(err, ErrJSONExceededStrings):
+			return fmt.Errorf("refused JSON input due to high complexity: %w", err)
+		default:
+			// Any other error is a "parse error" thrown by the tokenizer and
+			// not complexity enforcement.
+			return fmt.Errorf("failed to parse JSON input: %w", err)
```
cipherboy: Does this allow any sort of reflection perhaps? I guess it isn't over GET that this is used, so... 🤔

cipherboy: Though, I guess as you point out, certain endpoints like `sys/storage/raft/join` do this already.

satoqz: Hmm, reflection as in?

cipherboy: @satoqz https://owasp.org/www-community/attacks/xss/ is what I was vaguely thinking of.

satoqz: I doubt it. `encoding/json`'s errors are rather simple.

satoqz: But, if we'd like to stay on the safe and simple side, I can reduce this to only differentiate parsing vs limits without propagating the parse error.

wslabosz-reply (dismissed):

http/handler.go
satoqz: Dropped propagating the full error so we can skip the discussion. I'd just like to resolve the original issue, after all.

cipherboy: @satoqz I've looked into the implementations of errors and think I'm happy including them. There's a few that will reflect some values, but I think they're fairly short. And as pointed out elsewhere, the errors are passed back in other places.

satoqz: Alright, reverted. Makes it simpler, actually.

cipherboy (approved):
@satoqz Thanks!

PR #3245, opened 2026-06-09
POC KMIP Server in Transit

driif:
POC KMIP Server in Transit

## Acknowledgements

       also allowed.


builtin/logical/transit/kmip/auth.go
```diff
@@ -0,0 +1,66 @@
...
+func authMiddleware(a Adapter) kmipserver.Middleware {
+	return func(next kmipserver.Next, ctx context.Context, msg *kmiplib.RequestMessage) (*kmiplib.ResponseMessage, error) {
+		certs := kmipserver.PeerCertificates(ctx)
+		if len(certs) == 0 {
+			return nil, kmipserver.Errorf(kmiplib.ResultReasonPermissionDenied, "client certificate not provided")
+		}
+
+		subjectDN := certs[0].Subject.String()
+
+		allowedOps, err := a.AuthenticateCert(ctx, subjectDN)
```
cipherboy, line 34: This won't work well; I think `RawSubject` should be used to compare instead.

cipherboy: In particular, DNs are tuples-of-tuples and `String()` serialization does not organize that properly, flattening it to a single tuple. `RawSubject` preserves it correctly, though makes it trickier to provide via API if you don't hand the certificate at the time of role/... creation. 🤷

builtin/logical/transit/kmip/adapter.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/hashicorp/go-hclog"
+	kmiplib "github.com/ovh/kmip-go"
+)
+
+// Adapter is an interface that OpenBao engine must implement to expose key management via KMIP.
+type Adapter interface {
+	// AuthenticateCert resolves a TLS client certificate Subject DN to a set of allowed operations.
+	// empty allow list == "nothing allowed", nil slices == "all allowed"
+	AuthenticateCert(ctx context.Context, subjectDN string) (allowedOps []string, err error)
```
cipherboy, line 17: Am I mistaken or does `allowedOps` not give you the ability to limit to particular keys? I.e., you can allow all encrypt operations across all keys, but you can't restrict a certificate/... to particular source keys?

cipherboy: Re-inventing this, in particular, was what I was hoping to avoid with the global KMIP listener, fwiw. Cert matching is well-understood in certificate auth, it supports things like CRLs and OCSP and everything today, and it handles cert->policy assignment, with CEL support making that even more flexible.  Similarly, ACL policies are a decent-ish format for restricting operations by tokens (again, with CEL policies coming later)... My 2c.

builtin/logical/transit/kmip/server.go
```diff
@@ -0,0 +1,78 @@
...
+	cert, err := tls.X509KeyPair([]byte(cfg.CertPem), []byte(cfg.KeyPem))
+	if err != nil {
+		return nil, fmt.Errorf("failed to load server cert/key: %w", err)
+	}
+
+	tlsConfig := &tls.Config{
+		Certificates: []tls.Certificate{cert},
+	}
+
+	if cfg.RequireClientCert {
```
boumba100, line 34: Would it make sense to refuse requests or just not start the KMIP server if client auth is not required, since it is the only authentication mechanism?

If the intention is to allow unauthenticated requests, what do you think about adding a configuration value to explicitly enable it, such as something like `allow_unauthenticated`?

builtin/logical/transit/kmip/auth.go
driif: Thanks for explanation

builtin/logical/transit/kmip/adapter.go
driif: Yes, after our last discussions, we decided to give a try to a global KMIP listener by adding subject alternative names to certificates that locate the transit mount and auth. Requests would then be redirected via core.handleRequest(...) and go through the normal auth path. So the next step for the PoC is to explore the implementaion of global KMIP listener.

builtin/logical/transit/kmip/server.go
driif: That's a good point,  @boumba100, and it's definitely something we need to consider.

PR #3255, opened 2026-06-10
Create documentation for sealable namespaces

phyrog:
## Description

Adds documentation:
- concept docs for sealable namespaces
- getting started guide for sealable namespaces
- command documentation for new namespace commands
- usage examples of operator commands for sealable namespaces
- API docs for sealable namespace operations

Reverts parts of #3033 to undeprecate the `operator generate-root` command, which now uses the new `/sys/generate-root-token` endpoint and can be used for sealable namespaces too.

~Missing from the docs for now is the deletion of sealed namespaces, which is currently being worked on in #3188. We can add docs on that after the PR is merged.~

## Rationale

Part of: #1357

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/api-docs/system/generate-root-token.mdx
+++ b/website/content/api-docs/system/generate-root-token.mdx
@@ -1,16 +1,23 @@
 ---
 description: |-
   The `/sys/generate-root-token/` endpoints are used to create a new root token for
-  OpenBao.
+  OpenBao's root namespace or a sealable namespace.
 ---
 
 # `/sys/generate-root-token`
 
 The `/sys/generate-root-token` endpoint is used to create a new root token for OpenBao.
 
-## Read root generation progress
+:::info
 
-This endpoint reads the configuration and process of the current root generation
+These endpoints are available in both the root namespace and sealable namespaces
+and work with the respective set of key shares.
+
+:::
+
+## Read root token generation progress
+
+This endpoint reads the configuration and process of the current root token generation
 attempt.
 
 | Method | Path                               |
@@ -38,7 +45,7 @@ $ curl \
 }
 ```
 
-If a root generation is started, `progress` tracks how many unseal keys have
+If a root token generation is started, `progress` tracks how many unseal keys have
 been provided for this generation attempt, where `required` must be reached
 to complete. The `nonce` for the current attempt and whether the attempt is
 complete is also displayed.
@@ -56,8 +63,8 @@ key, encoded as base64.
 
 ## Start root token generation
 
-This endpoint initializes a new root generation attempt. Only a single root
-generation attempt can take place at a time.
+This endpoint initializes a new root token generation attempt. Only a single
+root token generation attempt can take place at a time.
 
 | Method | Path                               |
 | :----- | :--------------------------------- |
@@ -91,10 +98,10 @@ $ curl \
 }
 ```
 
-## Cancel root generation
+## Cancel root token generation
 
-This endpoint cancels any in-progress root generation attempt. This clears any
-progress made. This must be called to change the OTP or PGP key being used.
+This endpoint cancels any in-progress root token generation attempt. This clears
+any progress made. This must be called to change the OTP or PGP key being used.
 
 | Method   | Path                               |
 | :------- | :--------------------------------- |
@@ -108,11 +115,11 @@ $ curl \
     http://127.0.0.1:8200/v1/sys/generate-root-token/attempt
 ```
 
-## Provide key share to generate root
+## Provide key share to generate root token
 
-This endpoint is used to enter a single root key share to progress the root
+This endpoint is used to enter a single root key share to progress the root token
 generation attempt. If the threshold number of root key shares is reached,
-OpenBao will complete the root generation and issue the new token. Otherwise,
+OpenBao will complete the root key generation and issue the new token. Otherwise,
 this API must be called multiple times until that threshold is met. The attempt
 nonce must be provided with each call.
 
--- a/website/content/api-docs/system/generate-root.mdx
+++ b/website/content/api-docs/system/generate-root.mdx
@@ -1,4 +1,5 @@
 ---
+sidebar_label: /sys/generate-root
 description: |-
   The `/sys/generate-root/` endpoints are used to create a new root token for
   OpenBao.
@@ -7,8 +8,10 @@ description: |-
 :::warning
 
 Note: As of v2.5.3 usage of these endpoints is [deprecated](/community/deprecation/unauthed-generate-root/)
-and will be disabled by default. Consider using [`auth/token/create`](../auth/token.mdx)
-to create new root tokens from existing tokens with `sudo` permission.
+and will be disabled by default. Consider using
+[`auth/token/create`](../auth/token.mdx) to create new root tokens from existing
+tokens or the new authenticated
+[`sys/generate-root-token`](./generate-root-token.mdx) endpoint.
 To re-enable, see the documentation around the `disable_unauthed_generate_root_endpoints`
 [listener parameter](/docs/configuration/listener/unix/): it is suggested to
 do so only for the short period of time required to create a new root token.
--- a/website/content/api-docs/system/namespaces.mdx
+++ b/website/content/api-docs/system/namespaces.mdx
@@ -62,6 +62,7 @@ This endpoint creates a namespace at the given path.
   will be created.
 - `custom_metadata` `(map<string|string>: nil)` - A map of arbitrary string to string valued user-provided metadata meant
   to describe the namespace.
+- `seal` `(string: "")` - An optional seal config document (JSON or HCL) to create a sealable namespace
 
 ### Sample payload
 
@@ -70,7 +71,8 @@ This endpoint creates a namespace at the given path.
   "custom_metadata": {
     "foo": "abc",
     "bar": "123"
-  }
+  },
+  "seal": "seal \"shamir\" {\n    shares = 5\n    threshold = 3\n}"
 }
 ```
 
@@ -166,3 +168,90 @@ $ curl \
 }
 ```
 
+## Read namespace seal status (sealable namespaces only)
+
+This endpoint gets the seal status for the given sealable namespace path.
+
+| Method | Path                    |
+| :----- | :---------------------- |
+| `GET`  | `/sys/namespaces/:path/seal-status` |
+
+### Sample request
+
+```shell-session
+$ curl \
+    --header "X-Vault-Token: ..." \
+    http://127.0.0.1:8200/v1/sys/namespaces/ns1/seal-status
+```
+
+### Sample response
+
+```json
+{
+  "initialized": true,
+  "n": 5,
+  "nonce": "",
+  "progress": 0,
+  "sealed": false,
+  "t": 3,
+  "type": "shamir"
+}
+```
+
+## Unseal a namespace (sealable namespaces only)
+
+This endpoint unseals the given sealable namespace.
+
+| Method | Path                    |
+| :----- | :---------------------- |
+| `PUT`  | `/sys/namespaces/:path/unseal` |
+
+### Sample payload
+
+```json
+{
+  "key": "...",
+  "reset": false,
+}
+```
+
+### Sample request
+
+```shell-session
+$ curl \
+    --header "X-Vault-Token: ..." \
+    --request PUT \
+    --data @payload.json \
+    http://127.0.0.1:8200/v1/sys/namespaces/ns1/unseal
+```
+
+### Sample response
+
+```json
+{
+  "initialized": true,
+  "n": 5,
+  "nonce": "",
+  "progress": 1,
+  "sealed": true,
+  "t": 3,
+  "type": "shamir"
+}
+```
+
+## Seal a namespace (sealable namespaces only)
+
+This endpoint seals the given sealable namespace.
+
+| Method | Path                    |
+| :----- | :---------------------- |
+| `PUT`  | `/sys/namespaces/:path/seal` |
+
+### Sample request
+
+```shell-session
+$ curl \
+    --header "X-Vault-Token: ..." \
+    --request PUT \
+    http://127.0.0.1:8200/v1/sys/namespaces/ns1/seal
+```
--- a/website/content/api-docs/system/namespaces.mdx
+++ b/website/content/api-docs/system/namespaces.mdx
@@ -139,6 +139,36 @@ $ curl \
     http://127.0.0.1:8200/v1/sys/namespaces/ns1
 ```
 
+## Delete sealed namespace
+
+:::info
+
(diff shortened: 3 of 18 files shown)


website/content/docs/concepts/namespaces/sealable-namespaces.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+:::
+
+## Overview
+
+Sealable namespaces extend OpenBao's multi-tenancy model by giving tenants
+cryptographic control over their own namespace. When a namespace is created with
+sealing enabled, it gets its own encryption keys, separate from the global
+OpenBao seal. A tenant can independently seal and unseal their namespace without
+affecting any other tenant and without involving the platform operator.
```
wslabosz-reply:
```suggestion
affecting any other tenants and without involving the platform operator.
```

website/content/docs/guides/sealable-namespaces.mdx
```diff
@@ -0,0 +1,249 @@
...
+restarted, or stopped, you must supply at least 3 of these keys to unseal it
+before it can start servicing requests.
+```
+
+You can cancel the rotation procedure at any time using the `-cancel` flag.
+
+For more details see the documentation for the [`operator
+rotate-keys`](../../commands/operator/rotate-keys#examples) command.
+
+## Keyring Rotation
```
wslabosz-reply, line 220: maybe let's add a ## keyring status section before this one, usage of `bao operator key-status -ns=tenants/my-tenant`

phyrog: I added `key-status` under ## Keyring Rotation, it makes it easier to explain what is returned.

wslabosz-reply (dismissed):
LGTM

phyrog:
Updated to include the API doc updates - but GitHub seems to be lagging behind with processing the push, so it's not showing up here yet.

See diff here: https://github.com/openbao/openbao/compare/main...Ki-Reply-GmbH:openbao:docs/sealable-namespaces

website/content/docs/commands/operator/generate-root.mdx
```diff
@@ -57,6 +55,12 @@ Enter an unseal key to progress root token generation:
 $ bao operator generate-root -otp="..."
 ```
 
+Start a root token generation for a sealable namespace:
```
cipherboy:
```suggestion
Start root token generation for a sealable namespace:
```

website/content/docs/commands/namespace.mdx
```diff
@@ -87,24 +87,45 @@ Unlock the API for a descendant namespace at path `current/namespacens1/`:
...
 ## Usage
 
 ```text
 Usage: bao namespace <subcommand> [options] [args]
 
-  This command groups subcommands for interacting with Vault namespaces.
-  These set of subcommands operate on the context of the namespace that the
-  current logged in token belongs to.
+  This command groups subcommands for interacting with OpenBao namespaces.
+  These subcommands operate in the context of the namespace that the
```
cipherboy: Is this strictly true? 🤔 If a token T.NS1 exists, but we have no namespace (and thus implicitly the root), don't we err rather than redirect?

website/content/docs/concepts/namespaces/sealable-namespaces.mdx
```diff
@@ -0,0 +1,85 @@
...
+- **Independent availability control is needed**. A tenant can seal their
+  namespace to make it inaccessible without affecting other tenants. If a
+  namespace's keys are suspected to be compromised, that namespace can be sealed
+  and rekeyed in isolation.
+- **Breach containment**. If a namespace's encryption key is leaked, only that
+  namespace is affected. Other tenants' data remains protected by their own
+  keys.
+
+Sealable namespaces are an optional feature. Namespaces created without a seal
+configuration continue to work as before and are protected by the global OpenBao
```
cipherboy: Not to nitpick, but I'd like to try and keep inaccuracies to a minimum if we can. This does contradict like three or four sentences above, perhaps revise it slightly? 

Not sure how to do without making it longer and more complex. 😆

website/content/docs/concepts/namespaces/sealable-namespaces.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+## How Sealable Namespaces Work
+
+A sealable namespace has its own keyring and root key that are used to encrypt
+data. This means:
+
+- The global operator cannot read namespace data without the tenant's key
+- Even if the global root key is available, a sealed namespace's data remains
+  inaccessible
+- A server restart seals all namespaces; each must be unsealed independently
```
cipherboy: This is only true in single-server environments or in multi-server environments where all nodes are fully shutdown and then started again.

If nodes are rolling restarted, existing standby nodes will indicate recovery keys to the new active node and a namespace may be kept online as a result. It depends on restart vs key synchronization windows.

I'd probably avoid advertising an exact timing on key synchronization, tho.

website/content/docs/concepts/namespaces/sealable-namespaces.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+### Comparison With Namespace Locking
+
+|                            | Namespace Lock | Namespace Seal |
+| ---------------------------| -------------- | -------------- |
+| Blocks API requests        | Yes            | Yes            |
+| Discards in-memory keys    | No             | Yes            |
+| Per-tenant encryption keys | No             | Yes            |
+
+Both features can be used together. To temporarily block API requests to a namespace, use namespace locking. Use sealable namespaces if you need a stronger tenant data sovereignty.
```
cipherboy, line 84: @wslabosz-reply @phyrog Have we tested the interactions here? 🤔

website/content/docs/concepts/namespaces/sealable-namespaces.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+## Seal Methods
+
+Sealable namespaces currently only support Shamir seals. Auto-unsealing support is
+planned, but not yet implemented.
```
cipherboy: Do we want to solicit contributions in this regard?

website/content/docs/guides/sealable-namespaces.mdx
```diff
@@ -0,0 +1,260 @@
...
+You can always check the current status of the namespace keyring by using the `bao operator key-status` command:
+
+```shell-session
+$ bao operator key-status -ns=tenants/my-tenant
+Key Term            2
+Install Time        10 Jun 26 08:37 UTC
+Encryption Count    0
+```
+
+## Generate a Sealable Namespace Root Token
```
cipherboy, line 244: How about emergency deleting of namespaces like in #3188?

cipherboy (commented):
Fairly good, a few nits but much appreciative for the docs @phyrog!

website/content/docs/concepts/namespaces/sealable-namespaces.mdx
phyrog: They continue to work as before in the sense that all namespaces created previously will use the root barrier. But I see the confusion here. I'll try to find a different wording.

Edit: Yeah, for the concepts page this should be explained in a version-independent way, the current sentence only makes sense for v2.6.

website/content/docs/guides/sealable-namespaces.mdx
phyrog: I left that out in case there were any more changes to behavior in that PR. Since that is merged now, I will add docs on that too.

website/content/docs/concepts/namespaces/sealable-namespaces.mdx
phyrog: It's a bit weird. You can seal/unseal a locked namespace, since that is technically performed on the parent namespace. You can also lock/unlock a sealed namespace for the same reason. You cannot rotate keys or any other `operator` command on a locked namespace (also not on a sealed namespace).

I haven't checked the stored entry, but I assume the lock is persisted using the parent namespace barrier. Not sure if this is something we'd want to change or not.

cipherboy: @phyrog Yeah, we'd persist the lock with the parent barrier, which makes sense to me, tbh. 

I guess the interactions make sense together, thanks for confirming they're tested!

website/content/docs/concepts/namespaces/sealable-namespaces.mdx
phyrog: Done

website/content/docs/commands/namespace.mdx
phyrog: I think you are correct. Would be weird too if only the namespace commands behaved differently

phyrog: Seems like this is how vault is doing it. This description was 8 years old.

Anyway, I removed that sentence completely from the docs and the CLI help output.

phyrog:
@cipherboy Should be all fixed. Commits are now also showing up in the PR again.

website/content/api-docs/system/namespaces.mdx
```diff
@@ -137,6 +139,27 @@ $ curl \
...
+
+This endpoint deletes a sealed namespace at the specified path.
+
+| Method   | Path                    |
+| :------- | :---------------------- |
+| `DELETE` | `/sys/namespaces/:path/delete-sealed` |
+
+### Parameters
+
+- `force` `(bool: false)` - if set to true, will also recursively delete any data and child-namespaces of the given namespace
```
satoqz:
```suggestion
- `force` `(bool: false)` - if set to true, will also recursively delete any child namespaces.
```

website/content/api-docs/system/namespaces.mdx
```diff
@@ -137,6 +139,27 @@ $ curl \
     http://127.0.0.1:8200/v1/sys/namespaces/ns1
 ```
 
+## Delete sealed namespace
```
satoqz, line 142: It would be good to additionally mention:

- That the sudo capability is required
- That leases won't be revoked, and hence normal deletion is preferred if possible

See the help text I added to the delete-sealed command.

website/content/api-docs/system/rotate/index.mdx
```diff
@@ -6,6 +6,12 @@ description: >-
 
 # `/sys/rotate`
 
+:::info
+
+The same endpoints also work for sealable namespaces by specifying a sealable namespace path via the `X-Vault-Namespace` header.
```
satoqz:
```suggestion
These endpoints are available in both the root namespace and sealable namespaces and operate on the respective set of keys.
```

There are several ways to specify the namespace, best to not even mention that here.

website/content/api-docs/system/generate-root-token.mdx
```diff
@@ -1,16 +1,16 @@
 ---
 description: |-
   The `/sys/generate-root-token/` endpoints are used to create a new root token for
-  OpenBao.
+  OpenBao or sealable namespaces.
```
satoqz:
```suggestion
  OpenBao's root namespace or a sealable namespace.
```

website/content/api-docs/system/generate-root-token.mdx
```diff
@@ -1,16 +1,16 @@
...
 ---
 
 # `/sys/generate-root-token`
 
-The `/sys/generate-root-token` endpoint is used to create a new root token for OpenBao.
+The `/sys/generate-root-token` endpoint is used to create a new root token for OpenBao. The same endpoint can be used for generating a root token for a sealable namespace. For that, pass the sealable namespace path in the `X-Vault-Namespace` header.
```
satoqz: Do we want to be consistent and use an info box here like in the rotate docs?

phyrog:
@satoqz All updated as well.

website/content/api-docs/system/generate-root.mdx
```diff
@@ -7,8 +8,10 @@ description: |-
 :::warning
 
 Note: As of v2.5.3 usage of these endpoints is [deprecated](/community/deprecation/unauthed-generate-root/)
-and will be disabled by default. Consider using [`auth/token/create`](../auth/token.mdx)
-to create new root tokens from existing tokens with `sudo` permission.
+and will be disabled by default. Consider using
+[`auth/token/create`](../auth/token.mdx) to create new root tokens from existing
+tokens with `sudo` permission or the new and authenticated
```
cipherboy:
```suggestion
tokens or the new and authenticated
```

Turns out `sudo` non-root tokens don't work on that endpoint. Literally the only one where it doesn't lol.

website/content/docs/guides/sealable-namespaces.mdx
```diff
@@ -0,0 +1,296 @@
...
+## Deleting a Sealable Namespace
+
+If a sealable namespace is unsealed and empty, it can be deleted with the normal `bao namespace delete` command:
+
+```shell-session
+$ bao namespace delete -ns=tenants my-tenant
+Success! Namespace deleted at: tenants/my-tenant/
+```
+
+If a namespace is sealed and empty, it can be deleted with `bao namespace delete-sealed`:
```
cipherboy:
```suggestion
If a namespace is sealed and does not contain child namespaces, it can be
deleted with `bao namespace delete-sealed`:
```

cipherboy (dismissed):
I think the rest that I have are really just nits.

phyrog:
@cipherboy Fixed again 🙂

satoqz (dismissed):

website/content/api-docs/system/generate-root.mdx
```diff
@@ -7,8 +8,10 @@ description: |-
 :::warning
 
 Note: As of v2.5.3 usage of these endpoints is [deprecated](/community/deprecation/unauthed-generate-root/)
-and will be disabled by default. Consider using [`auth/token/create`](../auth/token.mdx)
-to create new root tokens from existing tokens with `sudo` permission.
+and will be disabled by default. Consider using
+[`auth/token/create`](../auth/token.mdx) to create new root tokens from existing
+tokens or the new and authenticated
```
wslabosz-reply: nit:
```suggestion
tokens or the new authenticated
```

website/content/api-docs/system/namespaces.mdx
```diff
@@ -166,3 +198,104 @@ $ curl \
...
+{
+  "initialized": true,
+  "n": 5,
+  "nonce": "",
+  "progress": 0,
+  "sealed": true,
+  "t": 3,
+  "type": "shamir"
+}
+```
```
wslabosz-reply: namespace seal doesn't have a response, it returns `204`

phyrog: Nice catch

satoqz (approved):

wslabosz-reply (approved):

phyrog:
@cipherboy I hope you're fine with me merging it, since you gave your approval previously

PR #3257, opened 2026-06-10
Add lease restoration to sealed namespaces

cipherboy:
This allows expiration manager to restore leases for a specified namespace, temporarily entering restore mode again.

## Acknowledgements

       also allowed.

### DIFF
--- a/helper/namespace/namespace.go
+++ b/helper/namespace/namespace.go
@@ -201,6 +201,13 @@ func (n *Namespace) ValidateUUID(candidate string) error {
 	return nil
 }
 
+// MatchesID returns true if the given accessor/lease ID matches the current
+// namespace.
+func (n *Namespace) MatchesID(id string) bool {
+	_, nsId := SplitIDFromString(id)
+	return nsId == n.ID || (nsId == "" && n.ID == RootNamespaceID)
+}
+
 // ContextWithNamespace adds the given namespace to the given context
 func ContextWithNamespace(ctx context.Context, ns *Namespace) context.Context {
 	return context.WithValue(ctx, contextNamespace, ns)
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -466,16 +466,26 @@ func (m *ExpirationManager) collectLeases() (map[*namespace.Namespace][]string, 
 			continue
 		}
 
-		view := m.leaseView(namespace)
-		keys, err := logical.CollectKeys(m.quitContext, view)
+		keys, err := m.collectNamespaceLeases(namespace)
 		if err != nil {
-			return nil, 0, fmt.Errorf("failed to scan for leases: %w", err)
+			return nil, 0, err
 		}
+
 		existing[namespace] = keys
 		leaseCount += len(keys)
 	}
 
 	return existing, leaseCount, nil
+}
+
+func (m *ExpirationManager) collectNamespaceLeases(ns *namespace.Namespace) ([]string, error) {
+	view := m.leaseView(ns)
+	keys, err := logical.CollectKeys(m.quitContext, view)
+	if err != nil {
+		return nil, fmt.Errorf("failed to scan for leases: %w", err)
+	}
+
+	return keys, nil
 }
 
 // lockLease takes out a lock for a given lease ID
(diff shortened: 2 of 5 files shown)


vault/namespace_store.go
```diff
@@ -1159,6 +1159,17 @@ func (ns *NamespaceStore) postNamespaceUnseal(ctx context.Context, unsealedNames
 		return err
 	}
 
+	// load expirations
+	if err := ns.core.expiration.RestoreNamespace(unsealedNamespace, func() {
+		go func() {
+			if err := ns.SealNamespace(ns.core.activeContext.Load(), unsealedNamespace.Path); err != nil {
+				ns.logger.Error("failed to re-seal namespace after erring loading leases", "err", err)
+			}
+		}()
```
wslabosz-reply, line 1168: we probably need same fallback operation on whole `postNamespaceUnseal`

helper/namespace/namespace.go
```diff
@@ -201,6 +201,17 @@ func (n *Namespace) ValidateUUID(candidate string) error {
 	return nil
 }
 
+// MatchesID returns true if the given namespace matches the ID format,
```
wslabosz-reply: it's more like "if given (accessor) id has (that) namespace's id as suffix"

maybe that's just me but given == argument to the function

wslabosz-reply (dismissed):
LGTM (@cipherboy maybe let's rebase on main to get the fixes to the CI)

vault/namespace_store.go
cipherboy: Yeah, I think you mentioned on today's call that @satoqz was going to do that, so I'll defer to him on this. Probably has ordering w.r.t. other namespace concerns... and not clear what happens if ns1/ns2 fails to load if ns1 is the sealed namespace and ns2 a random child. :thinking:

helper/namespace/namespace.go
cipherboy: @wslabosz-reply Pushed and refactored slightly to just rely on SplitID.

wslabosz-reply (approved):

cipherboy:
Thanks @wslabosz-reply!

vault/expiration.go
```diff
@@ -899,6 +932,55 @@ func (m *ExpirationManager) Stop() error {
...
+
+	m.pending.Range(func(keyRaw any, _ any) bool {
+		if key, ok := keyRaw.(string); ok && ns.MatchesID(key) {
+			m.pending.Delete(key)
+			m.nonexpiring.Delete(key)
+			m.lockPerLease.Delete(key)
+			leaseIds[key] = struct{}{}
+		}
+		return true
+	})
```
satoqz, line 963: @cipherboy don't we also want to stop the timer on pending leases so there's no dangling process waiting to schedule a revocation job?

vault/expiration.go
```diff
@@ -679,8 +689,31 @@ func (m *ExpirationManager) Tidy(ctx context.Context) error {
...
+func (m *ExpirationManager) Restore(errorFunc func()) error {
+	return m.restore(m.collectLeases, errorFunc)
+}
+
+// Restore is used to restore the leases belonging to a particular namespace
+// when it unseals.
+func (m *ExpirationManager) RestoreNamespace(ns *namespace.Namespace, errorFunc func()) error {
+	m.restoreMode.Store(true)
+	m.restoreLocks = locksutil.CreateLocks()
+
```
satoqz, line 703: I think this is only safe because we can unseal just one namespace at a time?

satoqz: Actually, isn't it unsafe because we can already try to unseal a namespace while the root-level restore is still running?

vault/expiration.go
cipherboy: @satoqz I think we'd end up getting a lot of warnings like:

> ```go
> r.m.logger.Warn("failed to find lease in pending map for revocation retry", "lease_id", r.leaseID)
> ```

so probably best to stop, yes. Will address in a follow-up.

vault/expiration.go
cipherboy: @satoqz The locking here is rather confusing. I've left the locks lying around in a follow-up; I don't think they consume that much memory to do so.

Issue #3110, opened 2026-05-15
Enable `gocritic`s `deferInLoop` lint to prevent memory leaks


phil9909:
Using the [`deferInLoop`](https://go-critic.com/overview.html#deferinloop) lint of the `gocritic` linter would have prevented #3086

Alternative: `revive` has some `defer` rules: https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md#defer but in my testing, it did not catch #3086:

`revive` found 15 defers in loops, `gocritic` found the same 15 + 2 more (including #3086). @ bd3878af30be44a7f9eeaf1edc9c1c36ada85bc9

6 of 17 (5 of 15 for `revive`) are in tests

```diff
diff --git a/.golangci.yml b/.golangci.yml
index 319747f009..7c83be9976 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,6 +9,17 @@ linters:
       # 135 -> 124
       # Highest: (*ServerCommand).Run (123) in command/server.go
       max-complexity: 124
+    gocritic:
+      disable-all: true
+      enabled-checks:
+        - deferInLoop
+    revive:
+      enable-all-rules: false
+      enable-default-rules: false
+      rules:
+        - name: defer
+          arguments:
+            - ["loop"]
 formatters:
   enable:
     - gofumpt
```

```console
$ golangci-lint run --config .golangci.yml --enable-only gocritic,revive
builtin/credential/jwt/path_oidc_test.go:740:4: deferInLoop: Possible resource leak, 'defer' is called in the 'for' loop (gocritic)
			defer s.server.Close()
			^
vault/core_cache_invalidate.go:162:4: deferInLoop: Possible resource leak, 'defer' is called in the 'for' loop (gocritic)
			defer metrics.MeasureSince([]string{dispatcherName, "enqueue-pending"}, time.Now())
			^
builtin/logical/database/rotation.go:550:5: defer: prefer not to defer inside loops (revive)
				defer framework.DeleteWAL(ctx, conf.StorageView, walID)
				^
builtin/logical/pki/acme_helpers_test.go:74:5: defer: prefer not to defer inside loops (revive)
				defer dns.RemoveRecord("_acme-challenge."+auth.Identifier.Value, "TXT", challengeBody)
				^
builtin/logical/pki/path_config_acme_test.go:96:3: defer: prefer not to defer inside loops (revive)
		defer cancel()
		^
[...]

PR #3259, opened 2026-06-10
Fix context for container cleanup

cipherboy:
## Description

When cleaning up container creation, we default to using the creation's context. However, when that errs or times out, we don't even call the cleanup API at all. This converts cleanup to using fresh, 10 second limited contexts for each cleanup call if we're using a cancelled context. This should ensure fewer outstanding containers lying around after test execution.

## Acknowledgements

       also allowed.

### DIFF
--- a/sdk/helper/docker/testhelpers.go
+++ b/sdk/helper/docker/testhelpers.go
@@ -279,10 +279,26 @@ func (d *Runner) StartNewService(ctx context.Context, addSuffix, forceLocalAddr 
 
 	cleanup := func() {
 		for range 10 {
-			_, err := d.DockerAPI.ContainerRemove(ctx, result.Container.ID, client.ContainerRemoveOptions{Force: true})
-			if err == nil || errdefs.IsNotFound(err) {
+			if func() bool {
+				// Container removal may take a little bit, but do not
+				// prematurely time out because our parent context was
+				// cancelled on us.
+				cleanupCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer cancel()
+
+				// We don't necessarily have anywhere to surface this error
+				// so swallow it.
+				_, err := d.DockerAPI.ContainerRemove(cleanupCtx, result.Container.ID, client.ContainerRemoveOptions{Force: true})
+				if err == nil || errdefs.IsNotFound(err) {
+					return true
+				}
+
+				return false
+			}() {
+				// OK to return early
 				return
 			}
+
 			time.Sleep(1 * time.Second)
 		}
 	}


sdk/helper/docker/testhelpers.go
satoqz, line 1: Should we just use a new 10s context at all times? If the passed context has a couple milliseconds left we'll still propagate it without renewing and likely fail.

cipherboy: @satoqz We revise the context choice inside the loop, so even if we fail once, we'll likely retry and thus swap to a fresh context. I was struggling to come up with how long we should wait; I think we block here not until the cancellation is accepted but until it is finished, so if the caller passes an extraordinarily long context for a fat container, we'll be better off using theirs than a shorter one here. But we also need to extend ours especially in the case where their (shorter) context has already expired as we do want to encourage cleanup. 

Its weird is the TL;DR I think. Better would be to have cleanup take an actual context and potentially propagate errors back up to the caller, but that'd be a bit of a refactor for this SDK. 

Thoughts?

satoqz: Right, missed the loop. Fine with me.

sdk/helper/docker/testhelpers.go
```diff
@@ -278,11 +278,26 @@ func (d *Runner) StartNewService(ctx context.Context, addSuffix, forceLocalAddr
...
 	cleanup := func() {
+		cleanupCtx := ctx
 		for range 10 {
-			_, err := d.DockerAPI.ContainerRemove(ctx, result.Container.ID, client.ContainerRemoveOptions{Force: true})
+			if cleanupCtx.Err() != nil {
+				// Container removal may take a little bit, but do not
+				// prematurely time out because our parent context was cancelled
+				// on us.
+				newCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer cancel()
```
satoqz, line 287: https://github.com/openbao/openbao/issues/3110 wouldn't be happy though.

cipherboy: @satoqz Revised!

sdk/helper/docker/testhelpers.go
```diff
@@ -278,11 +278,30 @@ func (d *Runner) StartNewService(ctx context.Context, addSuffix, forceLocalAddr
...
+			cancel()
+		}()
+
 		for range 10 {
-			_, err := d.DockerAPI.ContainerRemove(ctx, result.Container.ID, client.ContainerRemoveOptions{Force: true})
+			if cleanupCtx.Err() != nil {
+				// Container removal may take a little bit, but do not
+				// prematurely time out because our parent context was cancelled
+				// on us.
+				cancel()
```
satoqz: Maybe the linter is happy if we cancel outside of the branch 🤔 I did something similar in HA code and it didn't get flagged.

cipherboy: Yeah, I would've expected this was clean, I'm surprised it's not. Part of me is tempted to just do the for loop and let the defers leak, they're primarily used in tests anyways.

satoqz: I guess you can always wrap the loop body in an IIFE and defer there.

cipherboy: @satoqz that would mean every loop iteration would give us a new context because we'd cancel last iterations context, bringing us back to the earlier conversation I think. 🤷

sdk/helper/docker/testhelpers.go
```diff
@@ -278,11 +278,33 @@ func (d *Runner) StartNewService(ctx context.Context, addSuffix, forceLocalAddr
...
+				// Container removal may take a little bit, but do not
+				// prematurely time out because our parent context was
+				// cancelled on us.
+				//
+				// Deferring cancel here exceeds the loop iteration, which is
+				// fine as we'd prefer to share a 10s timeout over the
+				// remaining loop iterations rather than 10s per loop
+				// iterations. However, it guarantees that cancel does indeed
+				// get called. There's a small fixed number of loop iterations
+				// and this is only used in testing so it is fine.
```
satoqz: Sorry, still not fully convinced :P

Is there really a reason to share the renewed 10 seconds across loop iterations if we allow recreating it several times as part of that loop? For example, failing a call but having spent 9 seconds on it would allocate 1 second to the next call, which is likely to fail again due to the severely shortened timeout before the third call then gets another 10 seconds and has a better chance of succeeding.

If our only concern is not overriding an intentionally lengthened deadline but we're fine entirely replacing the context as we see fit, can't the below work?

```go
defer func() {
	for range 10 {
		func() {
			// Carry over the deadline of the outer context unless it is due in
			// less than 10 seconds.
			deadline, _ := ctx.Deadline()
			ctx, cancel := context.WithTimeout(
				context.Background(),
				max(10*time.Second, time.Until(deadline)),
			)
			defer cancel()

			// do cleanup ...
		}()
	}
}()
```

Alternatively, if we want an allocation of at least 10 seconds across the entire loop, we can do:

```go
defer func() {
	// Carry over the deadline of the outer context unless it is due in
	// less than 10 seconds.
	deadline, _ := ctx.Deadline()
	ctx, cancel := context.WithTimeout(
		context.Background(),
		max(10*time.Second, time.Until(deadline)),
	)
	defer cancel()

	for range 10 {
		// do cleanup ...
	}
}()
```

Code as it is right now would discard a manually canceled context too, so we might as well replace the context from the get-go.

satoqz (dismissed):
Feel free to merge as-is though... we won't think about it again.

satoqz (approved):

PR #3277, opened 2026-06-11
Share binary container image layer across distros

satoqz:
## Description

This ensures that each release binary is pushed to a particular container registry at most once. To do this, we need to make the binary layers added to different distributions (alpine, distroless, ubi) match bit-for-bit. A few tweaks are needed to do so:

- Ensure we always drop the binary at `/usr/bin/bao`. We previously used `/bin/bao`, but this is a symlink to `/usr/bin` on UBI and Distroless and not on Alpine, creating different layers.
- Ensure that the permissions on `/usr` and `/usr/bin` are identical across distros. Specifically, UBI does not have a user write bit set on `/usr/bin`, while Alpine and Distroless do. We can do this with the extra `bin` stage I added, while also deduplicating some code.
- Finally, ensure that `/usr` and `/usr/bin` share the same modification times across distros. There are a couple ways to attack this; I chose the overkill solution and set `SOURCE_DATE_EPOCH` together with `rewrite-timestamps=true`. This should generally help make images slightly more reproducible as a side effect.

To demo, try these:

```console
$ docker pull ghcr.io/openbao/openbao
$ docker pull ghcr.io/openbao/openbao-ubi
```

note how both download one ~70MB layer each time.

```console
$ docker pull ghcr.io/ki-reply-gmbh/openbao-nightly:2.6.0-nightly202606111626
$ docker pull ghcr.io/ki-reply-gmbh/openbao-ubi-nightly:2.6.0-nightly202606111626
```

and note how the first downloads the big layer, but the second doesn't.

## Rationale

Resolves: Someone's bandwidth/storage bill. Really just nerdsniped myself here after being disappointed that the container images shared exactly zero bits between the binary layers.

## Acknowledgements

       also allowed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,22 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
+# This is a helper stage that ensures the binary layer is always the same, no
+# matter which base image it is copied into:
+#
+# 1. Always use /usr/bin/bao, not /bin/bao etc.
+# 2. Apply the same file permissions across the /usr and /usr/bin directories.
+#    Specifically, UBI is missing an u+w bit on /usr/bin that Alpine and
+#    Distroless have.
+#
+# Together with SOURCE_DATE_EPOCH and rewrite-timestamp, this results in an
+# identical binary layer digest across all distributions below, i.e., a given
+# release binary is only ever pushed to a registry once, even if there is more
+# than one container image flavor packaging it.
+FROM scratch AS bin
+ARG TARGETARCH
+COPY --chmod=555 bin/${TARGETARCH}/bao /usr/bin/bao
+
 # This is {docker.io,quay.io,ghcr.io}/openbao/openbao{,-hsm}.
 FROM alpine:3.24.0 AS default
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,6 +1,22 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
+# This is a helper stage that ensures the binary layer is always the same, no
+# matter which base image it is copied into:
+#
+# 1. Always use /usr/bin/bao, not /bin/bao etc.
+# 2. Apply the same file permissions across the /usr and /usr/bin directories.
+#    For example, UBI is missing an u+w bit on /usr/bin that Alpine and
+#    Distroless have.
```
cipherboy, line 10: How does this work? I don't see what actually modifies the permissions there.

cipherboy: Ah is it by creating `/usr/bin` in scratch, we apply it on the subsequent copy?

satoqz: @cipherboy since we later `COPY --from=bin . /`, we also copy directories that were created by `COPY ... /usr/bin/bao` and overwrite the distro base image's existing permissions of `/usr/bin`.

satoqz: Alternatively we could add one `chmod` to the ubi target. But this is more... Dockerfile.

satoqz: I wonder if we can keep ubi's scoped down permissions and modify alpine and distroless instead using `COPY --from=bin --chmod=555 . /`

satoqz: That works and makes it more explicit I suppose. I don't think having or not having the write bit matters much tbh.

cipherboy (commented):
`podman inspect` also confirms the layers are the same. Thanks @satoqz!

cipherboy (approved):

PR #3282, opened 2026-06-12
Ensure lease timers are canceled, locks persist

cipherboy:
## Description

This ensures timers on lease tracking are cancelled when the namespace is sealed, preventing additional log messages. This also persists restoration locks; as there are a fixed number, they won't incur too much overhead relative to other places where these locks are already being used.

## Rationale

See also: https://github.com/openbao/openbao/pull/3257

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -699,6 +699,11 @@ func (m *ExpirationManager) Restore(errorFunc func()) error {
 // Restore is used to restore the leases belonging to a particular namespace
 // when it unseals.
 func (m *ExpirationManager) RestoreNamespace(ns *namespace.Namespace, errorFunc func()) error {
+	if m.core.Standby() {
+		m.logger.Info("skipping namespace lease restoration - standby")
+		return nil
+	}
+
 	m.restoreMode.Store(true)
 	return m.restore(func() (map[*namespace.Namespace][]string, int, error) {
 		leases, err := m.collectNamespaceLeases(ns)
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -358,8 +358,9 @@ func NewExpirationManager(c *Core, e ExpireLeaseStrategy, logger log.Logger, det
 	if shouldProcessExp {
 		// active nodes start by running in restore mode by default
 		exp.restoreMode.Store(true)
-		exp.restoreLocks = locksutil.CreateLocks()
 	}
+
+	exp.restoreLocks = locksutil.CreateLocks()
 
 	exp.expireFunc.Store(&e)
 	if exp.revokeRetryBase == 0 {
@@ -699,8 +700,6 @@ func (m *ExpirationManager) Restore(errorFunc func()) error {
 // when it unseals.
 func (m *ExpirationManager) RestoreNamespace(ns *namespace.Namespace, errorFunc func()) error {
 	m.restoreMode.Store(true)
-	m.restoreLocks = locksutil.CreateLocks()
-
 	return m.restore(func() (map[*namespace.Namespace][]string, int, error) {
 		leases, err := m.collectNamespaceLeases(ns)
 		if err != nil {
@@ -852,7 +851,6 @@ LOOP:
 		m.restoreLoaded.Delete(k)
 		return true
 	})
-	m.restoreLocks = nil
 	m.restoreModeLock.Unlock()
 
 	m.logger.Info("lease restore complete")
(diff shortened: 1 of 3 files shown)


vault/expiration.go
```diff
@@ -940,32 +938,45 @@ func (m *ExpirationManager) StopNamespace(ns *namespace.Namespace) {
...
 		if key, ok := keyRaw.(string); ok && ns.MatchesID(key) {
+			info, ok := infoRaw.(pendingInfo)
+			if !ok {
+				return true
+			}
+
+			// Cancel the timer so it doesn't fire again.
+			info.timer.Stop()
 			m.nonexpiring.Delete(key)
-			m.lockPerLease.Delete(key)
```
wslabosz-reply, line 968: there's a `(*ExpirationMgr).deleteLockForLease(id string)` method that invokes exactly `m.lockPerLease.Delete(id)`, maybe let's reuse that or remove altogether

vault/expiration.go
```diff
@@ -358,9 +358,10 @@ func NewExpirationManager(c *Core, e ExpireLeaseStrategy, logger log.Logger, det
 	if shouldProcessExp {
 		// active nodes start by running in restore mode by default
 		exp.restoreMode.Store(true)
-		exp.restoreLocks = locksutil.CreateLocks()
 	}
 
+	exp.restoreLocks = locksutil.CreateLocks()
```
wslabosz-reply: @cipherboy 
shouldn't we only process lease restoration on the active node?

this also reminded me we probably should only process namespace lease restoration on active node.

wslabosz-reply: maybe I'm completely clueless but `(*ExpirationMgr).restore()` creates 64 workers per namespace to restore the leases, what would happen if we got to 4+ namespace restorations at once, where we have only pool of 256 locks for restoration?

cipherboy: @wslabosz-reply I think processing lease expiration in parallel is fine due to the locking done here (now that we've fixed it to retain the lockset indefinitely and not clobber each other).

I've added three new tests, one which is just straight line expiration manager verification on leases being functional with sealed namespaces (`TestNamespaceSealedLeaseTokenRevoke`), one which operates on non-sealed namespaces (`TestNamespaceManyLeases`) and one which operates on sealed namespaces (`TestNamespaceSealManyLeases`).

satoqz: > we probably should only process namespace lease restoration on active node

Not probably, definitely!

cipherboy: Ah, I missed that we don't restore leases, for some reason I was thinking lookup went to memory, not storage. Fixed!

satoqz: I think, as @wslabosz-reply hinted, we still don't need to allocate them if we're on standby?

vault/expiration.go
```diff
@@ -698,9 +699,12 @@ func (m *ExpirationManager) Restore(errorFunc func()) error {
...
 // when it unseals.
 func (m *ExpirationManager) RestoreNamespace(ns *namespace.Namespace, errorFunc func()) error {
-	m.restoreMode.Store(true)
-	m.restoreLocks = locksutil.CreateLocks()
+	if m.core.Standby() {
+		m.logger.Info("skipping namespace lease restoration - standby")
+		return nil
+	}
 
+	m.restoreMode.Store(true)
```
satoqz: I'm still not sure we're handling this atomic right. If multiple namespaces unseal concurrently, can't one Goroutine set this to false again before the other finishes? Perhaps it should be an integer that we increment/decrement as a counter of in-progress, concurrent restores instead of setting a boolean value.

vault/expiration.go
cipherboy: @satoqz @wslabosz-reply We can, though I think consolidating as much of active/standby differences as we can would be good idea as a whole, from a design PoV.

Here there's a large performance delta in actually _loading_ the leases, but maintaining a lock set is cheap.

satoqz: Fine with me. Can we move it into the struct expression above to denote that it is statically part of expiration manager then?

vault/expiration.go
cipherboy: That's fair, I'll move it to an integer. I think this wasn't a concern of the original code but does matter more as we move to namespace loading/unloading.

cipherboy:
@satoqz Try this out perhaps?

satoqz (approved):

PR #3283, opened 2026-06-12
Fix invalidation entries in sealed namespaces

cipherboy:
## Description

Ignore all invalidations in sealed namespace storage: these will always return Vault is sealed, resulting in the core restarting. We'll correctly handle invalidations when the namespace is ultimately unsealed and there should be no loaded state for that namespace anyways.

Also fix a bug in login MFA invalidation, found while expanding test coverage to unsealed namespaces.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/core_cache_invalidate.go
+++ b/vault/core_cache_invalidate.go
@@ -35,11 +35,15 @@ func (c *Core) Invalidate(key ...string) {
 	c.invalidations.Add(key...)
 }
 
-func (c *Core) invalidateSynchronous(key string) {
+func (c *Core) invalidateSynchronous(key string) error {
 	job, _ := c.invalidations.buildInvalidateJobForKey(make(chan struct{}), context.Background(), key)
+
 	if err := job.Execute(); err != nil {
 		job.OnFailure(err)
+		return err
 	}
+
+	return nil
 }
 
 // invalidationManager is a long-lived subset of Core which is used to handle
@@ -356,6 +360,18 @@ func (ij *invalidationJob) Execute() error {
 		return nil
 	}
 
+	nsBarrier := ij.im.core.sealManager.NamespaceBarrierByLongestPrefix(ns.Path)
+	if nsBarrier.Namespace().UUID != namespace.RootNamespaceUUID && nsBarrier.Sealed() {
+		// When the parent namespace is sealed, ignore the request: this
+		// means we're unable to process the invalidation regardless of
+		// what the actual invalidated entry is. Only when this parent
+		// namespace becomes unsealed can we process changes to data within
+		// it. Notably, the namespace entry for a sealed namespace sits in
+		// the namespace above it, so it is not its own parent and thus will
+		// be correctly invalidated.
+		return nil
+	}
+
 	ctx = namespace.ContextWithNamespace(ctx, ns)
 
 	// Lastly, create a short version of the context for plugin invalidations.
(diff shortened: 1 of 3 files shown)


cipherboy:
Marking draft as I un-rebased from #3224, though it calls the new `.Namespace()` method. :facepalm:

cipherboy:
@wslabosz-reply @satoqz This is ready for review. 

I noticed that while we used `invalidateSynchronous` for testing, we never checked the error return and `c.restart()` doesn't actually do anything in the tests. Also expanded the existing tests for unsealed sealed namespaces while I was there.

vault/core_cache_invalidate.go
```diff
@@ -356,6 +360,18 @@ func (ij *invalidationJob) Execute() error {
 		return nil
 	}
 
+	nsBarrier := ij.im.core.sealManager.NamespaceBarrierByLongestPrefix(ns.Path)
+	if nsBarrier != nil && nsBarrier.Namespace().UUID != namespace.RootNamespaceUUID && nsBarrier.Sealed() {
```
wslabosz-reply: this check made me realize that `(*Core).NamespaceSealed()` would panic if the barrier is nil, but it can never be nil, because that would mean that there's no root namespace barrier - which is impossible? I'd leave what you wrote as a safeguard, but it's interesting how would we get to the point this barrier is nil.

cipherboy: @wslabosz-reply Yeah, that's correct. I think the nil check was leftover from when I was using the other `.Get(...)` method rather than the longest prefix version.

cipherboy:
@wslabosz-reply Updated!

vault/core_cache_invalidate.go
```diff
@@ -400,7 +416,7 @@ func (ij *invalidationJob) Execute() error {
 	case strings.HasPrefix(ij.key, "autopilot/") || ij.key == raftAutopilotConfigurationStoragePath:
 		// Raft context is reloaded when a standby becomes active, so it is
 		// safe to ignore changes to autopilot state.
-	case isLoginMFA(ij.key):
+	case isLoginMFA(ij.nsKey):
```
wslabosz-reply, line 419:
```suggestion
	case isLoginMFA(key):
```

wslabosz-reply (approved):

vault/core_cache_invalidate.go
cipherboy: @wslabosz-reply I think I will come through and revise this and remove `key` altogether. We have `key==ij.nsKey` and `ij.key` (with `ij.key == nsPrefix + ij.nsKey != key`) which included the namespace in it, which is rather confusing.

cipherboy: Probably in a separate PR.

wslabosz-reply: fine by me

vault/core_cache_invalidate_test.go
```diff
@@ -1209,11 +1305,86 @@ func TestCore_Invalidate_AuthMount_NonTransactional(t *testing.T) {
...
+					50,
+					func(page int, index int, path string) (cont bool, err error) {
+						// Bypass the security barrier and write garbage.
+						testCore_Invalidate_sneakValueAroundCache(
+							t,
+							namespace.RootContext(ctx),
+							c,
+							&logical.StorageEntry{
+								Key:   path,
+								Value: []byte("absolute-garbage"),
```
satoqz, line 1377: Based on the comment above I expected random bytes or so, but this is better 😆

satoqz (approved):
I think optimally we'd even expand the tracking and release of changed paths while unseal is loading state into memory to namespace unsealing too, but this'll be the most impactful fix nevertheless. That optimization can be post-beta.

cipherboy:
> I think optimally we'd even expand the tracking and release of changed paths while unseal is loading state into memory to namespace unsealing too, but this'll be the most impactful fix nevertheless. That optimization can be post-beta.

I think this does, yes? Unseal from a barrier state is atomic, so with maybe one exception we'll now be applying invalidations. We won't buffer them until unseal is done, which may prove problematic later, but we're also not discarding them either. :thinking:

cipherboy:
Regardless, I agree, lets revise later. :-)

PR #3286, opened 2026-06-12
Fix flake in TestNamespaceClusterSealing

cipherboy:
## Description

Invalidation across clusters is variable timing, so we should wrap this in an eventually block. 

## Acknowledgements

       also allowed.


wslabosz-reply (approved):

cipherboy:
Ironically #3282 is failing with the error from here and here's failing with the error from #3282 :laughing: So I think we merge this one and then see if `main` is green once landing everything.

PR #3288, opened 2026-06-12
Fix storage list in remount operation

wslabosz-reply:
## Description
- remount operation `moveStorage` now correctly lists storage entries

## Rationale
fix to bug introduced with namespace sealing

Resolves: no issue

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/mount.go
+++ b/vault/mount.go
@@ -765,7 +765,7 @@ func (c *Core) moveStorage(ctx context.Context, src namespace.MountPathDetails, 
 		key, keys = keys[0], keys[1:]
 		entryKey := path.Join(prefix, me.UUID, key)
 		if strings.HasSuffix(key, "/") {
-			nestedKeys, err := srcBarrier.List(ctx, entryKey)
+			nestedKeys, err := srcBarrier.List(ctx, entryKey+"/")
 			if err != nil {
 				return err
 			}


avinashkamat48 (commented):
This PR title/body is about fixing storage listing during remount, but the diff also changes namespace seal-status output formatting and clears `rotationConfig.Nonce` after root rotation. Those look like separate behavior changes with different risk surfaces from the `moveStorage` path fix. Could those be split into separate PRs, or called out in the PR description with tests, so reviewers can evaluate the remount fix independently from the seal/rotation changes?

cipherboy (dismissed):

cipherboy:
Would like @satoqz to take a look.

command/format.go
```diff
@@ -280,15 +280,26 @@ func (t TableFormatter) Output(ui cli.Ui, secret *api.Secret, data interface{})
 		return t.OutputList(ui, nil, data)
 	case map[string]interface{}:
 		return t.OutputMap(ui, data)
+	case api.SealStatusResponse:
+		ui.Output(tableOutput(t.namespaceSealStatusStruct(data), &columnize.Config{
+			Delim: "|",
+		}))
+		return nil
```
satoqz: Nit: Can we move this into a short `OutputNamespaceSealStatusStruct` method just to keep it more linear?

command/format.go
```diff
@@ -280,15 +280,26 @@ func (t TableFormatter) Output(ui cli.Ui, secret *api.Secret, data interface{})
...
 	case SealStatusOutput:
 		return t.OutputSealStatusStruct(ui, data)
 	default:
 		return errors.New("cannot use the table formatter for this type")
 	}
 }
 
-func (t TableFormatter) OutputSealStatusStruct(ui cli.Ui, data interface{}) error {
-	status := data.(SealStatusOutput)
+func (t TableFormatter) namespaceSealStatusStruct(data interface{}) []string {
```
satoqz: I think this helper just like `OutputSealStatusStruct` can take a typed input instead of any?

vault/rotate.go
```diff
@@ -436,7 +436,9 @@ func (sm *SealManager) performRootRotation(ctx context.Context, ns *namespace.Na
 			}
 		}
 
+		// unset VerificationKey and Nonce as we completed the rotation.
 		rotationConfig.VerificationKey = nil
+		rotationConfig.Nonce = ""
```
satoqz: Don't all rotations have a nonce, and we need to unset it not only when rotating shamir seal keys?

command/format.go
wslabosz-reply: resolved in another pr

command/format.go
wslabosz-reply: resolved in another pr

vault/rotate.go
wslabosz-reply: adjusted in https://github.com/openbao/openbao/pull/3299

cipherboy (approved):
I do sometimes wonder why `List` requires a trailing slash, but that's neither here nor there.

PR #3293, opened 2026-06-16
Fix namespace sealing bugs in clustering

satoqz:
## Description

This has fixes for bugs I'm finding while playing around with namespace sealing in clusters.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -1073,17 +1073,17 @@ func (ns *NamespaceStore) sealNamespaceLocked(ctx context.Context, namespaceToSe
 }
 
 // UnsealNamespace attempts unsealing namespace with a given path, using provided unseal key.
-func (ns *NamespaceStore) UnsealNamespace(ctx context.Context, path string, key []byte) error {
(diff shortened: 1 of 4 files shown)


vault/namespace_store.go
```diff
@@ -1099,53 +1099,53 @@ func (ns *NamespaceStore) UnsealNamespace(ctx context.Context, path string, key
...
+	}
+
+	// Namespace wasn't sealed before the call.
+	if !ns.core.NamespaceSealed(namespaceToUnseal) {
+		return true, nil
+	}
+
+	unsealed, err := ns.core.sealManager.UnsealNamespace(ctx, namespaceToUnseal, key)
+	if err != nil {
+		return false, nil
```
cipherboy:
```suggestion
		return false, err
```

perhaps?

wslabosz-reply (dismissed):

cipherboy (commented):
I think if not `return false, err`, we have to do the check for which `err` and decide when to surface it.

vault/namespace_store.go
satoqz: Yeah that wasn't intentional. Thanks!

cipherboy (approved):
Thanks @satoqz!

PR #3296, opened 2026-06-16
Skip namespace reloads on sealed namespaces

cipherboy:
## Description

On standby nodes, we'll attempt to reload the entire namespace's contents if the namespace object itself changed. This is necessary because the mount table will include cached namespace representations, so it is important to update these. In doing so, we'd err on sealed child namespaces, bypassing our earlier barrier check. This updates namespace reloading logic, also handling manually sealed namespaces explicitly during target namespace invalidation.

## Rationale

Resolves: #

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/core_cache_invalidate.go
+++ b/vault/core_cache_invalidate.go
@@ -379,29 +379,28 @@ func (ij *invalidationJob) Execute() error {
 	defer shortCancel()
 
 	// Now handle the actual event.
-	key := ij.nsKey
 	switch {
-	case strings.HasPrefix(key, namespaceStoreSubPath):
+	case strings.HasPrefix(ij.nsKey, namespaceStoreSubPath):
 		ij.fatal = true
 		return ij.namespaceInvalidation(ctx)
-	case strings.HasPrefix(key, barrier.SystemBarrierPrefix+policy.ACLSubPath):
+	case strings.HasPrefix(ij.nsKey, barrier.SystemBarrierPrefix+policy.ACLSubPath):
 		// Policy invalidation is not fatal as it contains a LRU cache: we
 		// know removal is strict and it is only potentially preloading an
 		// entry which may err.
 		return ij.policyInvalidation(ctx)
-	case strings.HasPrefix(key, barrier.SystemBarrierPrefix+quotas.StoragePrefix):
+	case strings.HasPrefix(ij.nsKey, barrier.SystemBarrierPrefix+quotas.StoragePrefix):
 		ij.fatal = true
 		return ij.quotaInvalidation(ctx)
-	case key == coreAuditConfigPath || key == coreLocalAuditConfigPath:
+	case ij.nsKey == coreAuditConfigPath || ij.nsKey == coreLocalAuditConfigPath:
 		ij.fatal = true
 		return ij.auditInvalidation(ctx)
-	case isLegacyMountPath(key):
+	case isLegacyMountPath(ij.nsKey):
 		ij.fatal = true
 		return ij.legacyMountInvalidation(ctx)
-	case isTransactionalMountPath(key):
+	case isTransactionalMountPath(ij.nsKey):
 		ij.fatal = true
 		return ij.transactionalMountInvalidation(ctx)
-	case isKeyringPath(key):
+	case isKeyringPath(ij.nsKey):
 		// The HA subsystem handles keyring rotations via the
 		// periodicCheckKeyUpgrades(...) actor.
 	case strings.HasPrefix(ij.key, coreLeaderPrefix):
@@ -432,7 +431,7 @@ func (ij *invalidationJob) Execute() error {
 		//
 		// This is true in reverse for deletions.
 	default:
-		ij.im.dispacherLogger.Warn("no mechanism to invalidate cache for specified key", "key", key)
+		ij.im.dispacherLogger.Warn("no mechanism to invalidate cache for specified key", "key", ij.nsKey, "full_key", ij.key)
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


vault/core_cache_invalidate.go
```diff
@@ -474,6 +473,16 @@ func (ij *invalidationJob) namespaceInvalidation(ctx context.Context) error {
...
 	}
 
+	nsBarrier := ij.im.core.sealManager.NamespaceBarrierByLongestPrefix(preferredNs.Path)
+	if nsBarrier.Sealed() {
+		// Namespace is sealed; nothing more we can do with it.
+		return nil
+	} else if preferredNs.ManuallySealed {
+		// Namespace was manually sealed but we're not actually sealed locally;
+		// seal it.
+		return ij.im.core.namespaceStore.SealNamespace(ctx, preferredNs.Path)
```
satoqz, line 484: This will hit a write path in `sealNamespaceLocked`.

satoqz (commented):
When I unseal a child namespace for the second time, standbys error and still have mounts loaded:

```
2026-06-16T17:40:21.641+0200 [INFO]  seals: unsealed namespace: namespace=foo/
2026-06-16T17:40:21.642+0200 [INFO]  namespace: calling post-unseal for namespace: ns_path=foo/ ns_uuid=987d22a7-fb5d-6fa5-ab03-622cdabae4d7
2026-06-16T17:40:21.642+0200 [ERROR] core: failed to mount entry: path=cubbyhole/ error="cannot mount under existing mount \"foo/cubbyhole/\""
2026-06-16T17:40:21.642+0200 [ERROR] core: failed to load namespace after unseal: error="failed to run namespace post-unseal [0/3WeUa6]: failed to setup mounts for namespace: failed to setup mount table" ns=987d22a7-fb5d-6fa5-ab03-622cdabae4d7
```

vault/core_cache_invalidate.go
cipherboy: @satoqz This write path will cause the node to restart, but I'm surprised it leaves mounts around, given it should've cleaned that up well before removal. :thinking: 

Will think about this more.

vault/namespace_store.go
```diff
@@ -238,16 +238,6 @@ func (ns *NamespaceStore) loadNamespacesLocked(ctx context.Context) error {
 		}
 		namespacesByUUID[namespace.UUID] = namespace
 		namespacesByAccessor[namespace.ID] = namespace
-
-		if namespace.ManuallySealed {
```
cipherboy, line 242: @satoqz I think this was the bug.

vault/namespace_store.go
```diff
@@ -1012,6 +1002,27 @@ func (ns *NamespaceStore) SealNamespace(ctx context.Context, path string) error
...
+		return fmt.Errorf("failed to get parent namespace from context: %w", err)
+	}
+
+	if !namespaceToSeal.HasParent(parent) {
+		return errors.New("namespace from context is not the parent of the target namespace to seal")
+	}
+
+	// Mark the namespace as sealed before we seal it; this ensures future
+	// loads will reflect the desired status.
+	if !ns.core.Standby() {
```
cipherboy, line 1016: @wslabosz-reply @satoqz I moved this earlier to reflect intent of what we're trying to do here: ensure the namespace stays sealed indefinitely after this. This causes a storage write, so should be done sooner, but the actual in-memory unloading can be delayed, IMHO.

vault/core_cache_invalidate_test.go
```diff
@@ -1385,6 +1386,24 @@ func TestCore_Invalidate_SealedNamespaces(t *testing.T) {
...
+			parentPath, ok := ns.ParentPath()
+			require.True(t, ok, "expected namespace to have parent")
+
+			parentNs, err := c.namespaceStore.GetNamespaceByPath(namespace.RootContext(t.Context()), parentPath)
+			require.NoError(t, err)
+			require.NotNil(t, parentNs)
+
+			childNsPath := filepath.Join(namespaceStoreSubPath, ns.UUID)
+			if parentNs.UUID != namespace.RootNamespaceUUID {
+				childNsPath = filepath.Join(barrier.NamespacePrefix, parentNs.UUID, childNsPath)
```
wslabosz-reply: why `filepath` instead of `path`?

cipherboy: Ah, wrong one.

cipherboy: @wslabosz-reply Pushed.

satoqz:
@cipherboy I'll merge this since it is an improvement but we still have a bad bug here.

Assuming the following layout:

```
/ <- root
 foo/ <- sealable namespace
     bar/ <- normal namespace
```

When we seal `foo/`, we won't unload `foo/bar/` on the standby.

This is because:

```go
beforeNs, beforeErr := ij.im.core.namespaceStore.GetNamespace(ctx, namespaceUUID)


// First notify the namespace storage that our next lookup might be stale.
ij.im.core.namespaceStore.invalidate(ctx, ij.key)


afterNs, afterErr := ij.im.core.namespaceStore.GetNamespace(ctx, namespaceUUID)
```

nukes `foo/bar/` from the namespace tree, which is then traversed via

```go
// Namespace was manually sealed but we're not actually sealed locally;
// seal it.
ij.im.dispacherLogger.Info("sealing manually sealed namespace", "path", preferredNs.Path, "uuid", preferredNs.UUID)
return ij.im.core.namespaceStore.SealNamespace(ctx, preferredNs.Path)
```

and will only unload `foo/`'s resources, not `foo/bar/`, since `loadNamespacesRecursive` did not load that back even though the barrier was unsealed.

Seems like we really just need to make either `loadNamespacesRecursive` a bit smarter or actually do fine-grained namespace invalidation.

satoqz (approved):

PR #3316, opened 2026-06-17
No longer delete quotas with mount deletion on standby nodes

wslabosz-reply:
## Description
With mount deletion we were removing quota entries from storage on standby nodes, this shouldn't happen as we have an event trigger (quota invalidation) on quota removal from active node. In case we had a quota assigned to the mount we deleted we were causing standby nodes crash and reload.

## Rationale
invalidation logic minor bugfix

Resolves: no issue created

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3316.txt
+++ b/changelog/3316.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+sys/quotas: Fix unintentional attempts to delete quotas on standby nodes when mount is removed.
+```
--- a/vault/core_cache_invalidate_test.go
+++ b/vault/core_cache_invalidate_test.go
@@ -353,7 +353,8 @@ func TestCore_Invalidate_Quota(t *testing.T) {
 	t.Parallel()
 	c, root := testCore_Invalidate_TestCore(t, nil)
 	rootCtx := namespace.RootContext(t.Context())
-	// 1. Create some qutoa to populate cache
+
+	// 1. Create quota to populate cache
 	req := logical.TestRequest(t, logical.CreateOperation, "sys/quotas/rate-limit/test-quota")
 	req.ClientToken = root
 	req.Data = map[string]any{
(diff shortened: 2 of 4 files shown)


phyrog (dismissed):

cipherboy (dismissed):

changelog/3316.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+sys/quotas: No longer removing quotas from storage on standby nodes with mount deletion; Properly handling it with quota invalidation.
```
satoqz:
```suggestion
sys/quotas: Fix unintentional attempts to delete quotas on standbys when a mount is reloaded.
```

Isn't it more like this?

wslabosz-reply: attempts suggest we were failing, but we actually were deleting them. This doesn't apply just to reloading, but mount "removal" alltogether.

satoqz: Deleting from memory I'd expect? Or do we bypass the read-only barrier somehow?

wslabosz-reply: @satoqz
after local test, I've confirmed the standby nodes just reload as they've encountered unrecoverable error
```
2026-06-18T14:03:30.988+0200 [ERROR] core: failed to update quotas after disabling mount: error="failed to delete quota from storage: cannot write to readonly storage" namespace="" uuid=8ffd7f2b-acdf-bdd5-2cda-60c8056eb298
2026-06-18T14:03:30.988+0200 [ERROR] core.invalidate-dispatch: fatal failure dispatching invalidation; restarting core: key=core/auth/8ffd7f2b-acdf-bdd5-2cda-60c8056eb298 err="unable to invalidate mount for key \"core/auth/8ffd7f2b-acdf-bdd5-2cda-60c8056eb298\" in namespace \"00000000-0000-0000-0000-000000000000\": failed to delete quota from storage: cannot write to readonly storage"
```

wslabosz-reply: @satoqz adjusted changelog entry

satoqz (approved):

PR #3325, opened 2026-06-18
Add Host header to logical.Request headers

### LINKED ISSUE #3254: Allow host header to be available on logical.Request, and able to be passed to plugins
I am working on an SSPI based kerberos authentication plugin.  While working on it, I found that using the default security context may or may not require a continuation which is difficult to implement as it requires persisting a CTxtHandle, whereas specifying an SPN does not require a continuation outside of exceptional cases.  Rather than specifying a single static configured SPN to use, I would prefer to be able to use the Host header provided by the client, which would allow the OpenBao instance to be used with multiple different host names so long as the service account running OpenBao has all the corresponding SPNs assigned.  However, I have found that golang's built in http server assigns the host header to the Host field, then removes the host header, and thus is not passed to the logical.Request.

In the http/logical.go file, in the `buildLogicalRequestNoAuth` function, which takes an `http.Request` and creates a `logical.Request`, take the value of `http.Request.Host`, and assign it to some value within a `logical.Request`, either by creating a new Host property, add it as a Host header in the Headers field, or add it to some other collection such as possibly Data field.

One alternative for me would be to have the SPN be a configuration value.  This removes the flexibility for being able to use different host names during SSPI kerberos authentication.  Another alternative would be to allow continuations, though this would require the client to identify to the server which previous request the current request continues, likely requiring allowing Cookie header passthrough and Set-Cookie response headers, while also ensuring the same backend server that handled the initial request also handles the continuation request given that a CtxtHandle is a machine dependent value.

TheJayMann:
## Description

This pull request adds the Host header value back to the header collection before passing it to a `logical.Request`, as `http.Server` removes this value from the `http.Request` header collection when adding it to the `Request.Host` field.

## Rationale

Code that interacts with a `logical.Request` can access the original HTTP headers passed in the request by accessing the `Headers` collection of the `logical.Request`.  If the code is part of a plugin, it requires that `--passthrough-request-header` to be specified.  However, because the Host header is removed by `http.Server` when assigned to the `http.Request.Host`, this is not passed to the `logical.Request` when simply copying the header collection as is.  By adding the host header back to the collection, it becomes available to any code that needs to access it.

Resolves: #3254 

## Acknowledgements

       also allowed.

### DIFF
--- a/http/handler_test.go
+++ b/http/handler_test.go
@@ -118,6 +118,38 @@ func TestHandler_parseMFAHandler(t *testing.T) {
 	}
 }
 
+type nilResponseWriter struct{}
+
+func (w *nilResponseWriter) Header() http.Header {
+	return make(http.Header)
+}
+
+func (w *nilResponseWriter) Write(b []byte) (int, error) {
+	return 0, nil
+}
+
+func (w *nilResponseWriter) WriteHeader(statusCode int) {}
+
+func TestHandler_HostHeader(t *testing.T) {
+	r, err := http.NewRequest(http.MethodGet, "http://domain.example/v1/path", nil)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	r.Header.Add("user-agent", "Test")
+	req, status, err := buildLogicalRequestNoAuth(&nilResponseWriter{}, r)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	if status != 0 {
+		t.Fatalf("status: %d", status)
+	}
+
+	require.Subset(t, req.Headers, http.Header{
+		http.CanonicalHeaderKey("user-agent"): {"Test"},
+		http.CanonicalHeaderKey("host"):       {"domain.example"},
+	})
+}
+
 func TestHandler_cors(t *testing.T) {
 	core, _, _ := vault.TestCoreUnsealed(t)
 	ln, addr := TestServer(t, core)
--- a/http/logical.go
+++ b/http/logical.go
@@ -185,13 +185,19 @@ func buildLogicalRequestNoAuth(w http.ResponseWriter, r *http.Request) (*logical
 		return nil, http.StatusInternalServerError, fmt.Errorf("failed to generate identifier for the request: %w", err)
 	}
 
+	// The http package removes the Host header when deserializing
+	// the http request.  This adds it back in so it is available
+	// in the logical request.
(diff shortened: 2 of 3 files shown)


http/logical.go
```diff
@@ -184,14 +184,17 @@ func buildLogicalRequestNoAuth(w http.ResponseWriter, r *http.Request) (*logical
 	if err != nil {
 		return nil, http.StatusInternalServerError, fmt.Errorf("failed to generate identifier for the request: %w", err)
 	}
-
+	
+	reqHeader := r.Header.Clone()
```
satoqz, line 191: An explanatory comment would be good.

wslabosz-reply: @TheJayMann 
I think @satoqz asked for a explanation why we add the host header back to the collection, something along the lines of "without this manual addition, http package strips it from the headers"

TheJayMann: I'll do this once I get home and I am able to make edits locally rather than through the web interface.

http/handler_test.go
```diff
@@ -117,6 +117,40 @@ func TestHandler_parseMFAHandler(t *testing.T) {
...
+               t.Fatalf("status: %d", status)
+       }
+
+       if value, ok := req.Headers[textproto.CanonicalMIMEHeaderKey("user-agent")]; !ok || len(value) != 1 || value[0] != "Test" {
+               t.Fatal("missing custom header")
+       }
+
+       if value, ok := req.Headers[textproto.CanonicalMIMEHeaderKey("host")]; !ok || len(value) != 1 || value[0] != "domain.example" {
+               t.Fatal("missing host header")
+       }
```
phil9909: The error message "missing" might be misleading (as it will also trigger for a wrong value), how about this:

```suggestion
	require.Subset(t, req.Header, http.Header{
		http.CanonicalHeaderKey("user-agent"): {"Test"},
		http.CanonicalHeaderKey("host"):       {"domain.example"},
	})
```

BTW: I think you need to run the formatter on your code.

TheJayMann: Thank you for the suggestion.  I did not know about the require module, but now that I see it is used in the other test functions, I realize I should use this as well.

Also, the formatting issue appears to have been caused by pasting issues using the web interface.  Once I apply these updates this should also be fixed.

satoqz (approved):

wslabosz-reply (approved):

satoqz:
Thanks @TheJayMann!

PR #3332, opened 2026-06-19
Ensure workflows do not escape namespace hierarchy

satoqz:
## Description

This seeks to address two related issues in workflow execution namespace routing.

1. Workflow execution could escape boundary of the namespace the execution was originally called from via path traversal in the 'X-Vault-Namespace' header. We don't want to allow this; requests generated by a workflow should not unexpectedly leave the hierarchy of the workflow's namespace. Making requests to the same namespace and any of its children is fine.

2. Any time we set X-Vault-Namespace, it should be the start of an absolute path. It is always interpreted this way for requests arriving externally, it should not suddenly become a path relative to the workflow's hosting namespace when set in a profile.

We can enforce this property more globally in request handling, depending on whether we get a namespace passed in `httpCtx`. If so, we can compare it against the namespace otherwise resolved from header + path.

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/request_handling.go
+++ b/vault/request_handling.go
@@ -591,20 +591,19 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
 	stop := context.AfterFunc(httpCtx, cancel)
 	defer stop()
 
+	nsHeader := namespace.HeaderFromContext(httpCtx)
 	contextNS, err := namespace.FromContext(httpCtx)
 	haveContextNS := err == nil
 
-	if haveContextNS {
+	if haveContextNS && nsHeader == "" {
 		// If we have a namespace in context, prepend its path to the request
-		// path before namespace resolution. We prepend to request path instead
-		// of prepending to header to guarantee that the header is always
-		// interpreted as an absolute path.
+		// path before namespace resolution unless a header is defined, which
+		// would set a new absolute path to join the request path to.
 		req.Path = contextNS.Path + req.Path
 	}
 
 	// Resolve the namespace for this request from header & path.
 	var ns *namespace.Namespace
-	nsHeader := namespace.HeaderFromContext(httpCtx)
 	ns, req.Path = c.namespaceStore.ResolveNamespaceFromRequest(nsHeader, req.Path)
 	if ns == nil {
 		return nil, logical.CodedError(http.StatusNotFound, "namespace not found")
--- a/vault/request_handling.go
+++ b/vault/request_handling.go
@@ -47,7 +47,6 @@ import (
 )
 
 const (
-	replTimeout                           = 1 * time.Second
 	EnvVaultDisableLocalAuthMountEntities = "BAO_DISABLE_LOCAL_AUTH_MOUNT_ENTITIES"
 
 	// coreLockedUsersPath is a base path to store locked users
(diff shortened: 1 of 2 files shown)


wslabosz-reply (dismissed):
LGTM

DanGhita (dismissed):
LGTM

vault/request_handling.go
```diff
@@ -592,18 +591,37 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
...
+	contextNS, err := namespace.FromContext(httpCtx)
+	haveContextNS := err == nil
+
+	if haveContextNS {
+		// If we have a namespace in context, prepend its path to the request
+		// path before namespace resolution. We prepend to request path instead
+		// of prepending to header to guarantee that the header is always
+		// interpreted as an absolute path.
+		req.Path = contextNS.Path + req.Path
+	}
```
cipherboy: This is the opposite of what we're doing now. By prepending to the header, we guarantee the ordering is:


1. Context
2. Header
3. Path

Whereas this moves it to:

1. Header
2. Context
3. Path

And I think the assertion of `ResolveNamespaceFromRequest(...)` having a prefix of `contextNS` means that `header` must be nil given we always strictly enforce this (or, header if not nil has to contain enough `../` to exceed the namespace itself).


My 2c., but I'd just add the new `ResolveNamespaceFromRequest(...)` operation and keep the original ordering? That way `header` can still be useful still independent of path and we keep the header prefixing as-is into header?


Thoughts?

satoqz: I'd really like to ensure that the header is always the beginning of the path, for consistency. Being able to set both a new absolute path anchor (via header) and a position-independent relative path (via mount path prefixing) seems most useful to me?

Good point though, it doesn't make much sense to chain the context's path before the request path if a header is set, too.

How about this?

```go
if haveContextNS && nsHeader == "" {
	// If we have a namespace in context, prepend its path to the request
	// path before namespace resolution unless a header is defined, which
	// would set a new absolute path to join the request path to.
	req.Path = contextNS.Path + req.Path
}
```

This way, for workflows, we'll evaluate either `context + path` or `header + path`, never all three.

satoqz: FWIW, I still think having absolute paths even if execution is confined to a parent namespace makes sense.

For example, I might have namespaces `foo/` and `foo/bar/`, and a workflow in `foo/` that has:

```hcl
headers = {
  X-Vault-Namespace = "foo/bar/"
}
```

Acting on the child namespace.

Suppose I then choose to move the workflow into the child namespace (`foo/bar/`) entirely, I can leave the header as-is and the workflow will keep working. This wouldn't be the case had we chained `context + header`.

cipherboy: > I'd really like to ensure that the header is always the beginning of the path, for consistency. Being able to set both a new absolute path anchor (via header) and a position-independent relative path (via mount path prefixing) seems most useful to me?


It is the beginning of the operator-defined workflow-specified path, sans required jail though.

cipherboy: > FWIW, I still think having absolute paths even if execution is confined to a parent namespace makes sense.

No, because you can't write reusable workflows. Each workflow needs to be rewritten for each child namespace it is put in.

satoqz: > No, because you can't write reusable workflows. Each workflow needs to be rewritten for each child namespace it is put in.

Did I not make a case below that the opposite is true depending on how we define reusable?

1. To write a workflow that is reusable by always targeting the right namespace relatively (e.g., a workflow to speak to a certain child namespace relative to the workflow's namespace), use relative namespace paths via path prefixing.

2. To write a workflow that will always speak to the same namespace no matter which level in the hierarchy it is (re-)located in, use absolute namespace paths by setting the header.

If we joined `context + header`, 1) could instead be achieved in two ways, while 2) couldn't be achieved.

cipherboy: 2 I think is a non-goal, 1 was what I was building for.

satoqz: I mean sure, if we really want it to behave that way 🤷 Doesn't seem like there's a global optimum tbh. Will adjust.

satoqz:
@cipherboy adjusted, LMKWYT.

cipherboy (approved):
We can add tests post-beta.

I think my worry is that `request_handling.go` might be the wrong place for the concat, but maybe this is a global simplification rather than making this check specific to workflow store. Will have to think about that more async.

satoqz:
Note that `contextNS` rarely even comes up as non-nil outside of the workflow store and tests. So as long as it was compatible with the behavior that tests depended on, I figured we could apply it globally.

PR #3339, opened 2026-06-22
Stop logging on known core entries removal during namespace deletion

wslabosz-reply:
## Description
Adds a list of known entry paths (limited to just `/core` prefix) that we remove on (sealable) namespace deletion, which shouldn't be logged as a "bug". Also changes the log level to WARN instead of debug.

## Rationale
no need to log on these entries

Resolves: no issue

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/namespace_store.go
+++ b/vault/namespace_store.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"path"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -55,6 +56,18 @@ const (
 		1 + /* locked user entries */
 		1 /* final view clearing */
 )
+
+// knownNamespaceCoreEntriesToCleanup is a "whitelist" of known paths to storage
+// entries we delete during final view clearing while namespace deletion.
+var knownNamespaceCoreEntriesToCleanup = []string{
+	barrier.KeyringPath,
+	barrier.RootKeyPath,
+	barrier.ShamirKekPath,
+	barrierSealConfigPath,
+	recoverySealConfigPath,
+	recoveryKeyPath,
+	StoredBarrierKeysPath,
+}
 
 // NamespaceStore is used to provide durable storage of namespace. It is
 // a singleton store across the Core and contains all child namespaces.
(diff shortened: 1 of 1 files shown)


vault/namespace_store.go
```diff
@@ -93,6 +94,10 @@ type NamespaceStore struct {
 	logger hclog.Logger
 }
 
+// coreEntriesToCleanup is a slices of paths of entries we are deleting with
+// final clear view of namespace deletion.
+var coreEntriesToCleanup = []string{barrier.KeyringPath, barrier.RootKeyPath, barrier.ShamirKekPath, barrierSealConfigPath, recoverySealConfigPath, recoveryKeyPath, StoredBarrierKeysPath}
```
satoqz: Can you:

- Put one array item per line
- Move this below constants, but above the struct
- Name it something namespace-specific, like `namespaceCoreEntries`
- Emphasize that it is a whitelist, not something we directly delete

wslabosz-reply: right on it

satoqz (dismissed):

vault/namespace_store.go
```diff
@@ -1491,7 +1504,11 @@ func (ns *NamespaceStore) clearNamespaceResources(nsCtx context.Context, entry *
 				return false, fmt.Errorf("failed removing entry: %w", err)
 			}
 
-			ns.logger.Debug("bug: removing entry remaining in namespace storage after all mounts were removed", "namespace", entry.Path, "path", path)
+			// List of paths we know we are clearing with this final scan.
+			if !slices.Contains(knownNamespaceCoreEntriesToCleanup, path) {
+				ns.logger.Debug("bug: removing entry remaining in namespace storage after all mounts were removed", "namespace", entry.Path, "path", path)
```
satoqz: Actually, do we want to make it a warn log? I'm surprised it isn't. We won't get any bug reports with it being debug.

wslabosz-reply: @cipherboy 
you've added that, wdyt?

satoqz: Alternatively, if we're confident that mounts do not leave anything behind anymore indeed, and this was meant for development only, let's skip the effort and just drop it.

cipherboy: @satoqz @wslabosz-reply I just didn't want to spam logs. Could we once-gate this per relative path, eliding `logical/<uuid>` & `namespace/<uuid>` paths? :thinking:

satoqz: We could collect paths and log once per namespace deletion.

wslabosz-reply: @cipherboy so you want to collect all paths and log once, while also excluding everything besides `/core/*`?

cipherboy: Once per path if we can, if too hard, can just do once

wslabosz-reply: @cipherboy updated, is this what you had in mind?

cipherboy (approved):
I think this is fine. Not quite what I had in mind but we can adjust later.

PR #3343, opened 2026-06-22
Reduce work on non-persistent leases

### LINKED ISSUE #3341: Inline auth memory leak
**Describe the bug**
When using line auth, memory keeps growing. It appears that when inline auth is performed, leases are created in memory but never cleaned up. 

**To Reproduce**
Uninteresting setup:

```
$ ./bao server -dev -dev-root-token-id 1 &
$ export VAULT_ADDR=http://localhost:8200
$ export VAULT_TOKEN=1
$ ./bao auth enable userpass
Success! Enabled userpass auth method at: userpass/
$ echo 'path "*" { capabilities = ["create","read","update","delete","list","sudo"] }' | ./bao policy write default -
Success! Uploaded policy: default
$ ./bao write auth/userpass/users/user password=pass
Success! Data written to: auth/userpass/users/user
$ curl -H "X-Vault-Inline-Auth-Path: auth/userpass/login/user" -H "X-Vault-Inline-Auth-Parameter-password: eyJrZXkiOiAicGFzc3dvcmQiLCAidmFsdWUiOiAicGFzcyJ9" "$VAULT_ADDR/v1/secret/data/test" -X POST -d '{"data": {"a": "b"}}'
{"request_id":"896ac4fd-3302-ff08-be5b-da2a1efd68e8","lease_id":"","renewable":false,"lease_duration":0,"data":{"created_time":"2026-06-22T14:42:17.045930363Z","custom_metadata":null,"deletion_time":"","destroyed":false,"version":1},"wrap_info":null,"warnings":null,"auth":null}
```

Wait a bit, check metrics. There's 1 lease.
```
$ curl "$VAULT_ADDR/v1/sys/metrics" 2>/dev/null | jq '.Gauges[] | select(.Name=="vault.expire.num_leases") | .Value'
1
```

Make 100 requests with inline auth. 
```
$ for i in $(seq 1 100); do curl -H "X-Vault-Inline-Auth-Path: auth/userpass/login/user" -H "X-Vault-Inline-Auth-Parameter-password: eyJrZXkiOiAicGFzc3dvcmQiLCAidmFsdWUiOiAicGFzcyJ9" "$VAULT_ADDR/v1/secret/data/test"; done
...
```

Wait a bit, check metrics. There are now 101 leases.
```
$ curl "$VAULT_ADDR/v1/sys/metrics" 2>/dev/null | jq '.Gauges[] | select(.Name=="vault.expire.num_leases") | .Value'
101
```

These 101 leases don't seem to ever go away and ends up eating memory. 

**Expected behavior**
After the inline auth returns, I expect the number of lease (as reported by sys/metrics) to not increase.

**Environment:**
* OpenBao Server Version (retrieve with bao status):
```
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    1
Threshold       1
Version         2.5.5
Build Date      2026-06-17T11:18:48Z
Storage Type    inmem
Cluster Name    vault-cluster-1f2fe9f9
Cluster ID      c8f9fb07-c828-abbe-c426-a32fa5c3cde2
HA Enabled      false
```

OpenBao CLI Version (retrieve with bao version):
```
OpenBao v2.5.5 (028992583c693c4de6350b8aa52ff85e30375a99), built 2026-06-17T11:18:48Z
```

Server Operating System/Architecture:
```
openbao_2.5.5_linux_amd64.pkg.tar.zst
```

**Additional context**
I believe this is the cause of the leak:

In `RegisterAuth`, we allocate some resources, such as allocating an entry for the lock and storing the lease in memory:
https://github.com/openbao/openbao/blob/371536731922261b154846fe00b33eac97316f8e/vault/expiration.go#L1730
https://github.com/openbao/openbao/blob/371536731922261b154846fe00b33eac97316f8e/vault/expiration.go#L1742

Then in `revokeCommon`, we just completely skip over non-persisted leases. This leaks both the lock entry and the lease in memory.
https://github.com/openbao/openbao/blob/371536731922261b154846fe00b33eac97316f8e/vault/expiration.go#L1103-L1106

I'm wondering, does this block of code need to run if `persistLease` is `false`? 
https://github.com/openbao/openbao/blob/371536731922261b154846fe00b33eac97316f8e/vault/expiration.go#L1712-L1742

I don't have much context on this, but could we potentially guard even more code form running in the case of `persistLease = false`?

cipherboy:
Resolves: #3341

## Rationale

Resolves: #

## Acknowledgements

 - [ ] By contributing this change, I certify I have signed-off on the
       also allowed.

### DIFF
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -1703,6 +1703,11 @@ func (m *ExpirationManager) RegisterAuth(ctx context.Context, te *logical.TokenE
 		return consts.ErrPathContainsParentReferences
 	}
 
+	if !persistLease {
+		// Do no further work if we are not persisting the lease.
+		return nil
+	}
+
 	tokenNS, err := m.core.NamespaceByID(ctx, te.NamespaceID)
 	if err != nil {
 		return err
@@ -1740,10 +1745,8 @@ func (m *ExpirationManager) RegisterAuth(ctx context.Context, te *logical.TokenE
 	defer leaseLock.Unlock()
 
 	// Encode the entry
-	if persistLease {
-		if err := m.persistEntry(ctx, &le); err != nil {
-			return err
-		}
+	if err := m.persistEntry(ctx, &le); err != nil {
+		return err
 	}
 
 	// Setup revocation timer
--- a/changelog/3343.txt
+++ b/changelog/3343.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/auth: Ensure inline auth does not generate in-memory lease information.
+```


satoqz (approved):

vault/expiration.go
jackyliao123, line 1756: Just curious, this is not needed for inline auth, right?

jackyliao123: Specifically the `GenerateSSCTokenID`

cipherboy: > Just curious, this is not needed for inline auth, right?

Toekn isn't returned to callers, so SSCT wouldn't matter.

satoqz:
Thanks @cipherboy!

wslabosz-reply (approved):

PR #3346, opened 2026-06-22
fix(command): revoke self-init root token

### LINKED ISSUE #3345: Initial root token remains valid after self-init flow
**Describe the bug**
Declarative self-initialization creates a transient root token during `core.Initialize` so OpenBao can perform first-boot configuration.

That token is currently not revoked after the self-init flow exits. This means the transient root token can remain valid after successfull self-init.

**To Reproduce**
Steps to reproduce:

1. Configure declarative self-init on an auto-unseal OpenBao server.
2. Include self-init requests that capture or use the initialization token, for example by calling `auth/token/lookup-self`.
3. Start OpenBao and allow self-init to run.
4. After self-init exits, check if the captured initialization token is still present in the token store.
5. Notice that the token remains valid.

**Expected behavior**
The transient root token created for declarative self-init should be revoked after self-init exits.

This cleanup should happen after successful self-init and after failure paths where `core.Initialize` already returned a root token.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): current `main`
* OpenBao CLI Version (retrieve with `bao version`): current `main`
* Server Operating System/Architecture: not OS-specific

```hcl
# Any declarative self-init configuration on an auto-unseal server can trigger
# this behavior. The issue is in the lifecycle of the transient root token
# returned by core.Initialize, not in a specific storage or listener setup.
```

**Additional context**
This was reported and discussed with maintainers before opening an issue or PR publicly, and was classified as a normal bug rather than a security advisory. A follow-up PR will revoke the transient self-init root token through the existing token handler path using `auth/token/revoke-self`, matching the existing dev-server root token cleanup pattern.

dc-tec:
## Description
This PR updates declarative self-initialization so the transient root token created during `core.Initialize` is revoked after self-init finishes.

The cleanup is deferred immediately after successful initialization, so it runs on both successful and failed self-init paths. Revocation is performed through the normal token handler path using `auth/token/revoke-self`.

## Rationale

Declarative self-init needs a temporary root token to perform first-boot configuration. That token should not remain usable after self-init has completed or failed.

This was reported and discussed with maintainers before opening this PR publicly and was classified as a normal bug rather than a security advisory.

Resolves: #3345

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3346.txt
+++ b/changelog/3346.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+command/server: Revoke the transient root token after declarative self-initialization, including failure paths.
+```
--- a/command/self_init_root_token_test.go
+++ b/command/self_init_root_token_test.go
@@ -1,0 +1,220 @@
+package command
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/cli"
+	log "github.com/hashicorp/go-hclog"
+	"github.com/openbao/openbao/command/server"
+	"github.com/openbao/openbao/helper/namespace"
+	"github.com/openbao/openbao/helper/testhelpers/teststorage"
+	vaulthttp "github.com/openbao/openbao/http"
+	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/openbao/openbao/vault"
+	"github.com/openbao/openbao/vault/seal"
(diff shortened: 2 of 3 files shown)


cipherboy (approved):
Thanks @dc-tec!

PR #3348, opened 2026-06-23
Pin container images (in Dockerfiles)

satoqz:
## Description

Ensure that absolutely all container images referenced in Dockerfiles in this repo are pinned by hash.

This means pinning Alpine and UBI (Distroless was already pinned). I've also chosen to simply remove:

- `scripts/docker/Dockerfile`, this is really the same as just `/Dockerfile` but will do the Go build in Docker. I don't think anyone uses this, a local Go toolchain is a reasonable expectation. This also referenced Alpine 3.19, so quite some dust on that file.
- Integration test Dockerfiles for Kubernetes secrets & auth engines; These aren't explicitly required by the test code, reference Hashicorp container images and assume that the engines are used in plugin mode. Easiest to throw them out, too. 

To verify I'm injecting up-to-date hashes:

```console
$ crane manifest alpine:3.24.1 | sha256sum
28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b  -

$ crane manifest registry.access.redhat.com/ubi10-minimal:10.2 | sha256sum
3948fdfe71007909b37faf48c52eda28bfab7c4e440d6f4d4619422d06ceeb4c  -
```

## Rationale

More predictable releases. Also helps with https://scorecard.dev/viewer/?uri=github.com/openbao/openbao.

## Acknowledgements

       also allowed.


wslabosz-reply (approved):

JanMa (approved):
LGTM

PR #3353, opened 2026-06-24
identity: gracefully skip partially-created namespaces during unseal

### LINKED ISSUE #2921: openbao crashes with inconsistent database (half created namespace)
**Describe the bug**

OpenBao crashed during namespace creation and after restart kept crashing with this stack: 

```
2026-04-15T14:13:54.606+0200 [ERROR] identity: failed to get entityPacker: err="namespace 2ee56127-8284-7f5d-bb91-de31176edd69 missing from identity store table"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x55f28ee171cd]

goroutine 61 [running]:
github.com/openbao/openbao/helper/storagepacker.(*StoragePacker).View(...)
    github.com/openbao/openbao/helper/storagepacker/storagepacker.go:41
github.com/openbao/openbao/vault.(*IdentityStore).loadEntities(0xb2869cd3440, {0x55f294ac9df0, 0xb2869def110}, 0x0)
    github.com/openbao/openbao/vault/identity_store_util.go:213 +0x12d
github.com/openbao/openbao/vault.(*Core).loadIdentityStoreArtifacts.func1({0x0?, 0x0?})
    github.com/openbao/openbao/vault/identity_store_util.go:56 +0x1fa
github.com/openbao/openbao/vault.(*Core).loadIdentityStoreArtifacts(0xb28697b6308, {0x55f294ac9e28, 0xb2869e32f00}, 0x0?)
    github.com/openbao/openbao/vault/identity_store_util.go:71 +0x85
github.com/openbao/openbao/vault.readonlyUnsealStrategy.unsealShared({}, {0x55f294ac9e28, 0xb2869e32f00}, 0xb28697b6308, 0x0)
    github.com/openbao/openbao/vault/core.go:2433 +0x285
github.com/openbao/openbao/vault.standardUnsealStrategy.unseal({{}}, {0x55f294ac9e28, 0xb2869e32f00}, {0x100?, 0xb2869e29670?}, 0xb28697b6308)
    github.com/openbao/openbao/vault/core.go:2335 +0x23b
github.com/openbao/openbao/vault.(*Core).postUnseal(0xb28697b6308, {0x55f294ac9e28, 0xb2869e32f00}, 0xb2869e28660, {0x55f294a8efc0, 0x55f294dff2a0})
    github.com/openbao/openbao/vault/core.go:2490 +0x2d7
github.com/openbao/openbao/vault.(*Core).unsealInternal(0xb28697b6308, {0x55f294ac9cc0, 0x55f294dff2a0}, {0xb2869fc8540?, 0x0?, 0x0?})
    github.com/openbao/openbao/vault/core.go:1941 +0x605
github.com/openbao/openbao/vault.(*Core).UnsealWithStoredKeys(0xb28697b6308, {0x55f294ac9cc0, 0x55f294dff2a0})
    github.com/openbao/openbao/vault/init.go:525 +0x3d3
github.com/openbao/openbao/command.runUnseal(0xb2869a16540, 0xb28697b6308, {0x55f294ac9cc0, 0x55f294dff2a0})
    github.com/openbao/openbao/command/server.go:2673 +0x5f
created by github.com/openbao/openbao/command.(*ServerCommand).Run in goroutine 1
    github.com/openbao/openbao/command/server.go:1328 +0x2f6c
```

The cause of the crash seems to be that the namespace entry in `/core/namespaces/` was already created, but the mount table wasn't, leaving the database in an inconsistent state.

The issue causing this state was hopefully fixed by https://github.com/openbao/openbao/pull/2226 (the cluster was using 2.4.3), but even upgrading to the latest release (2.5.2 at this point in time) does not help to recover from this. 

I'm opening this issue, because:
- Someone might still be running 2.4.x and if we handle this case more gracefully upgrading would allow them to restore their cluster.
- Maybe there is similar bug, leading to missing default mount table entries for namespaces, which would render a cluster unusable.

**To Reproduce**

1. Start OpenBao with `/sys/raw` enabled, e.g.

    ```bash
    cat > config.hcl <<HCL
    storage "file" {
      path = "./data"
    }

    seal "static" {
      current_key_id = "1"
      current_key = "0000000000000000000000000000000000000000000000000000000000000000"
    }
    HCL

    bao server -dev -config=config.hcl -dev-root-token-id=root
    ```
    
1. In another terminal run

    ```bash
    export BAO_ADDR='http://127.0.0.1:8200'
    export BAO_TOKEN='root'
    
    # create a namespace
    bao namespace create test
    
    # now simulate a interrupted namespace creation by manually modifying the mount table
    bao write sys/raw/core/mounts value="$(bao read -field=value sys/raw/core/mounts | jq -c 'del(.entries[] | select(.namespace_id != "root"))')"
    ```
    
1. Back in the initial terminal, restart OpenBao

    <kbd>CTRL</kbd> + <kbd>C</kbd>
    
    You will need the `-dev-skip-init` now

    ```bash
    bao server -dev -config=config.hcl -dev-skip-init
    ```

**Expected behavior**

I see multiple ways of recovery:
- automatically add default mounts, if missing
- mark the namespace as "broken" or "corrupted" and continue with other namespaces
- delete the namespace: I guess this would be appropriate for failed/partial creations, but if another bug leads to this state on an existing namespace, this is not a good solution (maybe still better than making the whole server unusable)
- We call the panic the "expected behavior", after all the database is inconsistent

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): v2.5.2
* OpenBao CLI Version (retrieve with `bao version`): v2.5.2
* Server Operating System/Architecture: Linux amd64

EnricoFusi:
## Description

When a server crashes mid-namespace-creation, the namespace UUID is written to
storage under `core/namespaces/` but `AddNamespaceView` is never called,
leaving the identity store's in-memory `views` map with no entry for that UUID.

On the next restart, `loadIdentityStoreArtifacts` iterates all namespaces from
storage (including the orphan entry) and calls `LoadEntities` and `LoadGroups`
for each. Both functions call `EntityPacker`/`GroupPacker`, which internally
call `getNSView`. When the view is absent, these return `nil`. The immediate
call to `.View()` on the nil packer produces a nil pointer dereference panic,
making the server permanently unbootable without manual intervention.

The fix adds `ErrNamespaceNotInIdentityStore` as a typed sentinel and returns
it from `LoadEntities` and `LoadGroups` when their packer is nil, converting
the crash into a recoverable error. `loadIdentityStoreArtifactsForNamespace`
catches this error symmetrically for both load steps and skips the namespace
with a warning that references recovery mode for cleanup.

**Performance**: there is no additional storage I/O. The check relies on
`i.views.Load(ns.UUID)` — a `sync.Map` lookup that is O(1) and purely
in-memory. The `views` map is already populated during `setupMounts` as part
of the normal unseal sequence. If the view is absent, it means `setupMounts`
never completed for that namespace. There is no per-namespace storage read
added, so there is little speed impact on large deployments.

The root cause that produced the inconsistent storage state was fixed in
#2226. This PR addresses surviving that state on clusters that were affected
before the fix, or any future inconsistency of the same kind.

## Rationale

Resolves: #2921

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/identity_store.go
+++ b/vault/identity_store.go
@@ -2,6 +2,7 @@ package vault
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/hashicorp/errwrap"
@@ -69,10 +70,24 @@ func (c *Core) loadIdentityStoreArtifactsForNamespace(ctx context.Context, ns *n
 
 	ctx = namespace.ContextWithNamespace(ctx, ns)
 
+	warnAndSkip := func() {
+		c.logger.Warn("skipping identity store loading for namespace whose setup did not complete; "+
+			"use recovery mode to remove the corrupt entry if this persists",
+			"namespace_id", ns.ID)
+	}
+
 	if err := c.identityStore.LoadEntities(ctx, readOnly); err != nil {
+		if errors.Is(err, ident.ErrNamespaceNotInIdentityStore) {
+			warnAndSkip()
+			return nil
+		}
 		return err
 	}
 	if err := c.identityStore.LoadGroups(ctx, readOnly); err != nil {
+		if errors.Is(err, ident.ErrNamespaceNotInIdentityStore) {
+			warnAndSkip()
+			return nil
+		}
 		return err
 	}
 
--- a/changelog/3353.txt
+++ b/changelog/3353.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/namespaces: gracefully skip partially-created namespaces during unseal instead of panicking
+```
--- a/vault/identity/store_util.go
+++ b/vault/identity/store_util.go
@@ -28,9 +28,10 @@ import (
 )
 
 var (
-	ErrDuplicateIdentityName = errors.New("duplicate identity name")
-	ErrCycleDetectedPrefix   = "cyclic relationship detected for member group ID"
-	tmpSuffix                = ".tmp"
+	ErrDuplicateIdentityName       = errors.New("duplicate identity name")
+	ErrNamespaceNotInIdentityStore = errors.New("namespace not registered in identity store")
+	ErrCycleDetectedPrefix         = "cyclic relationship detected for member group ID"
+	tmpSuffix                      = ".tmp"
 )
 
 func (i *IdentityStore) sanitizeName(name string) string {
@@ -51,7 +52,11 @@ func (i *IdentityStore) LoadGroups(ctx context.Context, readOnly bool) error {
 	}
 
 	i.logger.Debug("identity loading groups", "namespace", ns.Path)
-	existing, err := i.GroupPacker(ctx).View().List(ctx, groupBucketsPrefix)
+	gp := i.GroupPacker(ctx)
+	if gp == nil {
+		return fmt.Errorf("%w: %s", ErrNamespaceNotInIdentityStore, ns.UUID)
+	}
+	existing, err := gp.View().List(ctx, groupBucketsPrefix)
 	if err != nil {
 		return fmt.Errorf("failed to scan for groups: %w", err)
 	}
@@ -151,7 +156,11 @@ func (i *IdentityStore) LoadEntities(ctx context.Context, readOnly bool) error {
 	}
 
 	i.logger.Debug("loading entities", "namespace", ns.Path)
-	existing, err := i.EntityPacker(ctx).View().List(ctx, storagepacker.StoragePackerBucketsPrefix)
+	ep := i.EntityPacker(ctx)
+	if ep == nil {
+		return fmt.Errorf("%w: %s", ErrNamespaceNotInIdentityStore, ns.UUID)
+	}
+	existing, err := ep.View().List(ctx, storagepacker.StoragePackerBucketsPrefix)
 	if err != nil {
 		return fmt.Errorf("failed to scan for entities: %w", err)
 	}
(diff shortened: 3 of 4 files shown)


cipherboy (approved):
This makes sense to me, thanks @EnricoFusi!

vault/identity_store.go
```diff
@@ -69,10 +70,24 @@ func (c *Core) loadIdentityStoreArtifactsForNamespace(ctx context.Context, ns *n
...
 
+	warnAndSkip := func() {
+		c.logger.Warn("skipping identity store loading for namespace whose setup did not complete; "+
+			"use recovery mode to remove the corrupt entry if this persists",
+			"namespace_id", ns.ID)
+	}
+
 	if err := c.identityStore.LoadEntities(ctx, readOnly); err != nil {
+		if errors.Is(err, ident.ErrNamespaceNotInIdentityStore) {
+			warnAndSkip()
```
cipherboy, line 81: Noting assumption for others: this namespace wouldn't have been usable, though will now kinda be usable I suppose. Packer group wouldn't exist, identity store would be broken, so the suggestion to use recovery mode / normal mode deletion is probably fine.

wslabosz-reply (approved):

vault/identity_store.go
EnricoFusi: Agreed — the namespace was already unusable; this just prevents the server from being permanently unbootable because of it.

cipherboy:
@satoqz @phil9909  I'll let you apply backport labels if you want.

PR #3356, opened 2026-06-24
Fix source tarball release

### LINKED ISSUE #3351: Release dist tarball has truncated list of node module licenses
#3209 rearranged release.yml to do build the dist tarball in a different way.  I noticed that as a result of that change the node-related information is truncated.  In particular, SPDX_NODE_MODULES.txt contains only this very short list:
```
(MPL-2.0 OR Apache-2.0)
Apache-2.0
BSD-3-Clause
ISC
MIT
```
Previously it contained a much longer list:
```
(BSD-3-Clause OR MIT)
(MIT AND BSD-3-Clause)
(MIT AND Zlib)
(MIT OR Apache-2.0)
(MIT OR CC0-1.0)
(MPL-2.0 OR Apache-2.0)
0BSD
Apache-2.0
BSD
BSD-2-Clause
BSD-3-Clause
BlueOak-1.0.0
CC-BY-3.0
CC-BY-4.0
CC0-1.0
ISC
MIT
Public Domain
Unicode-DFS-2016
Unlicense
```
Although the `build-dist` step requires the `build-ui` step, I believe the problem may be related to the results of `make static-dist` not getting passed between the two steps.

satoqz:
## Description

PNPM only links direct dependencies to the top-level of `node_modules/`, so we must use a deeper traversal.

Also add the UI cache step to the dist tarball job so it includes the built UI. I had thought of this, but it must've silently disappeared in some refactor.

## Rationale

Resolves: https://github.com/openbao/openbao/issues/3351

@DrDaveD there are some new licenses that appear but those seem to legitimately stem from UI dependency upgrades that happened with v2.6.

## Acknowledgements

       also allowed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -236,6 +236,12 @@ jobs:
         with:
           package_json_file: ui/package.json
 
+      - name: Set up UI cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: http/web_ui/
+          key: ${{ github.ref }}-ui
+
       - name: Make LICENSE_NODE_MODULES.txt
         run: |
           pnpm install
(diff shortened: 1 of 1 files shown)


wslabosz-reply (dismissed):

satoqz:
Something's still weird.

satoqz:
Nevermind, looks good. Must've confused myself since I was testing releases on multiple branches in parallel and also had some conflicting caches locally.

satoqz:
Here's a tarball: https://github.com/Ki-Reply-GmbH/openbao/releases/download/v2.6.0-nightly202606241033/openbao-dist-v2.6.0-nightly202606241033.tar.xz

DrDaveD (changes requested):
Very good.  Should do the same with the VERSION_NODE_MODULES.txt step.

satoqz:
Okay, found an even better approach. Output for both steps is now very close to what we had in v2.5.5, limiting to prod deps only did a lot.

satoqz:
Ah, need to bring back normalizing any AND/OR licenses though.

satoqz:
```
(BSD-3-Clause OR MIT)
(MIT AND BSD-3-Clause)
(MIT AND Zlib)
(MIT OR Apache-2.0)
(MIT OR CC0-1.0)
(MPL-2.0 OR Apache-2.0)
0BSD
Apache-2.0
BlueOak-1.0.0
BSD
BSD-2-Clause
BSD-3-Clause
CC-BY-4.0
CC0-1.0
ISC
MIT
MPL-2.0
Public Domain
Unlicense
```

We lost `Unicode-DFS-2016` and `CC-BY-3.0` somewhere between upgrades. We'd have more licenses (including things like `Python-2.0`) if we included dev dependencies, but I don't see why we should do that.

DrDaveD (approved):
Looks good to me.

cipherboy (approved):
Thanks @satoqz!

PR #3381, opened 2026-06-26
raft: truncate raft log based on number of entries in log

### LINKED ISSUE #3342: Raft log compaction/snapshotting not working as expected
**Describe the bug**

The disk space allocated by `raft.db` can (under certain conditions) grow without bounds.  

The `bbolt`-based Raft implementation never takes real [snapshots](https://en.wikipedia.org/wiki/Raft_(algorithm)#Log_Compaction) in the Raft algorithm sense. Instead, after a restart it synthesizes a snapshot point by reporting the last applied index in `vault.db` as the latest snapshot. As a result, Raft never triggers log truncation if fewer than `snapshot_threshold` (default 8192) new entries are written between restarts, and the log can grow indefinitely.

The Integrated Storage documentation says:

[link](https://github.com/openbao/openbao/blob/371536731922261b154846fe00b33eac97316f8e/website/content/docs/internals/integrated-storage.mdx?plain=1#L77-L86)
> It would be undesirable to allow a replicated log to grow in an unbounded fashion. Raft provides a mechanism by which the current state is saved to snapshots and its related **logs are compacted**.
...
> This is performed automatically without user intervention and prevents unbounded disk usage

[link](https://github.com/openbao/openbao/blob/371536731922261b154846fe00b33eac97316f8e/website/content/docs/configuration/storage/raft.mdx?plain=1#L91-L99)

> `snapshot_threshold` `(integer: 8192)` - This controls the minimum number of Raft commit entries between snapshots that are saved to disk.
...
> Increasing this trades off disk IO for disk space since the log will grow much larger and the space in the `raft.db` file can't be **reclaimed till the next snapshot**. 

Furthermore, the periodic snapshot process does not do `bbolt`-level compaction, so the physical file can only grow. Whether the doc claims that disk space is actually returned to the OS is subject to reader interpretation, this may be a secondary issue since users must allocate enough space for peak usage anyway.

**To Reproduce**

The following test reproduces a bug where log entries from before a restart are not counted towards `snapshot_threshold`:

<details>

<summary>physical/raft/raft_test.go</summary>

```go
func TestRaft_LogCompaction(t *testing.T) {
	const (
		snapshotThreshold = 10
		trailingLogs      = 10
	)

	conf := map[string]string{
		"path":                   t.TempDir(),
		"node_id":                "abc123",
		"trailing_logs":          strconv.Itoa(trailingLogs),
		"snapshot_threshold":     strconv.Itoa(snapshotThreshold),
		"snapshot_interval":      "1",
		"performance_multiplier": "1",
	}

	backend, err := NewRaftBackend(conf, hclog.NewNullLogger())
	require.NoError(t, err)
	rb := backend.(*RaftBackend)
	require.NoError(t, rb.Bootstrap([]Peer{{ID: rb.NodeID(), Address: rb.NodeID()}}))

	// startCluster restarts the cluster and waits for it to be up.
	startCluster := func() {
		require.NoError(t, rb.TeardownCluster(nil))
		require.NoError(t, rb.SetupCluster(t.Context(), SetupOpts{}))
		require.Eventually(t, func() bool {
			return rb.raft.State() == raft.Leader && rb.raft.AppliedIndex() >= rb.raft.LastIndex()
		}, 3*time.Second, 100*time.Millisecond, "expected leader with all entries applied")
	}

	// writeEntries writes n entries and waits for them to be applied.
	writeEntries := func(n int) {
		for i := 0; i < n; i++ {
			require.NoError(t, rb.Put(t.Context(), &physical.Entry{
				Key:   fmt.Sprintf("key-%d", i),
				Value: []byte("value"),
			}))
		}
		require.Eventually(t, func() bool {
			return rb.raft.AppliedIndex() >= rb.raft.LastIndex()
		}, 5*time.Second, 100*time.Millisecond, "expected all entries to be applied")
	}

	// assertFirstIndex until the log's first index equals expected, or fails on timeout.
	assertFirstIndex := func(expected uint64) {
		t.Helper()
		var actual uint64
		if !assert.Eventually(t, func() bool {
			var err error
			actual, err = rb.logStore.FirstIndex()
			require.NoError(t, err)
			return actual == expected
		}, 10*time.Second, 100*time.Millisecond) {
			t.Fatalf("expected firstIndex=%d, got %d", expected, actual)
		}
	}

	// logEntryCount returns the number of log entries currently in the log.
	logEntryCount := func() uint64 {
		first, err := rb.logStore.FirstIndex()
		require.NoError(t, err)
		last, err := rb.logStore.LastIndex()
		require.NoError(t, err)
		return last - first + 1
	}

	startCluster()

	// 1. Start with empty log. First index in raft.db should be 1.
	assertFirstIndex(1)

	// 2. Write entries but stay below snapshot threshold. 
	writeEntries(5)
	require.Less(t, logEntryCount(), uint64(snapshotThreshold),
		"entries below snapshot threshold, no snapshot expected")
	assertFirstIndex(1)

	// 3. Restart the cluster.
	startCluster()

	// 4. Write more entries so total log exceeds snapshot threshold.
	writeEntries(7)
	require.Greater(t, logEntryCount(), uint64(snapshotThreshold),
		"entries exceed snapshot threshold, snapshot should fire")

	// 5. Snapshot should delete log entries from raft.db, leaving only trailing logs.
	lastIdx, err := rb.logStore.LastIndex()
	require.NoError(t, err)
	assertFirstIndex(lastIdx - trailingLogs + 1) // FAILS: expected firstIndex=6, got 1 due to restart

	require.NoError(t, rb.TeardownCluster(nil))
}
```
</details>

**Expected behavior**

* `raft.db` must not grow without bound.
* The documentation should make it easy to understand how `raft.db` size behaves.

**Additional context**
The size of `raft.db` depends on how many entries accumulate between snapshots, how large the entries are, and how the snapshot threshold, trailing logs, and check interval interact. Since `raft.db` never shrinks, the peak size becomes the permanent disk space requirement. Preferably the disk size usage should be easily predictable, so users can estimate how much space they need to allocate.

The docs are bit confusing, likely because they mostly describe this from the Raft algorithm perspective e.g. _"Raft provides a mechanism by which the current state is saved to snapshots and its related logs are compacted"_ which does really not match the bbolt-based implementation (`vault.db` does not use Raft snapshots, `raft.db` is not compacted).

tsaarni:
## Description

This pull request replaces `github.com/hashicorp/raft` snapshot-driven log truncation with a background worker in OpenBao. This worker manages log truncation based on the actual number of entries in the log.

## Rationale

Fix a bug where `raft.db` log grows indefinitely if fewer than `snapshot_threshold` entries are written before next restart.

Background:

`github.com/hashicorp/raft` implements the snapshotting mechanism defined by the Raft algorithm. Because BoltDB persists data automatically, OpenBao does not require real snapshots. So, snapshotting is implemented as a no-op, serving only as a trigger to delete old log entries.

This creates a side effect: at startup, the Raft algorithm needs to know the last snapshot point to know which entries to replay from the log. To prevent unnecessary log replays, OpenBao's Raft backend "synthesizes" a snapshot point at the last applied index ([link](https://github.com/openbao/openbao/blob/e7c03d362ce067f614a795fd3c813b538d74bd8e/physical/raft/snapshot.go#L134-L166)). 

While this avoids unnecessary replays, the algorithm also uses this fake snapshot point to manage its snapshotting logic. It assumes a snapshot was just taken and the log was truncated, practically reseting the threshold counter at every restart without considering the actual number of entries in the log. As a result, if fewer than `snapshot_threshold` entries are written before next restart, the algorithm never triggers a snapshot, causing the log file to grow indefinitely.

This change (practically) disables Raft's built-in snapshot mechanism and implements truncation based on the actual log entry count. It also updates the documentation to focus less on how the Raft algorithm works and more on how it is implemented with BoltDB.

Resolves: #3342

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/physical/raft/raft.go
+++ b/internal/physical/raft/raft.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"math/rand"
 	"os"
 	"path/filepath"
@@ -127,6 +128,21 @@ type RaftBackend struct {
 	// stableStore is used by the raft library to store additional metadata in
 	// durable storage.
 	stableStore raft.StableStore
+
+	// snapshotThreshold defines the minimum number of log entries between log truncation operations.
+	snapshotThreshold uint64
+
+	// trailingLogs defines how many log entries are left in the log store after log truncation.
+	trailingLogs uint64
+
+	// snapshotInterval defines how often to check if log truncation is needed.
+	snapshotInterval time.Duration
+
+	// logTruncationStopCh is closed to stop the log truncation worker.
+	logTruncationStopCh chan struct{}
+
+	// logTruncationDoneCh is closed when the log truncation worker has exited.
+	logTruncationDoneCh chan struct{}
 
 	// bootstrapConfig is only set when this node needs to be bootstrapped upon
 	// startup.
@@ -817,33 +833,48 @@ func (b *RaftBackend) applyConfigSettings(config *raft.Config) error {
 	config.HeartbeatTimeout *= time.Duration(multiplier)
 	config.LeaderLeaseTimeout *= time.Duration(multiplier)
 
+	b.snapshotThreshold = config.SnapshotThreshold
+	b.trailingLogs = config.TrailingLogs
+	b.snapshotInterval = config.SnapshotInterval
+
+	// Disable hashicorp/raft's snapshot-driven log truncation.
+	// It counts entries since the last snapshot to decide when to truncate, but
+	// BoltSnapshotStore.List() must report a snapshot at the last applied index,
+	// otherwise Raft replays the log at startup. This resets the counter and
+	// prevents truncation from triggering if fewer than snapshot_threshold
+	// entries are written before next restart.
+	config.SnapshotThreshold = math.MaxUint64
+	config.SnapshotInterval = math.MaxInt64 / 2
+	config.TrailingLogs = math.MaxUint64
+
 	snapThresholdRaw, ok := b.conf["snapshot_threshold"]
 	if ok {
-		var err error
 		snapThreshold, err := strconv.Atoi(snapThresholdRaw)
 		if err != nil {
 			return err
 		}
-		config.SnapshotThreshold = uint64(snapThreshold)
+		b.snapshotThreshold = uint64(snapThreshold)
 	}
 
 	trailingLogsRaw, ok := b.conf["trailing_logs"]
 	if ok {
-		var err error
 		trailingLogs, err := strconv.Atoi(trailingLogsRaw)
 		if err != nil {
 			return err
 		}
-		config.TrailingLogs = uint64(trailingLogs)
+		b.trailingLogs = uint64(trailingLogs)
 	}
+
 	snapshotIntervalRaw, ok := b.conf["snapshot_interval"]
 	if ok {
-		var err error
 		snapshotInterval, err := parseutil.ParseDurationSecond(snapshotIntervalRaw)
 		if err != nil {
 			return err
 		}
-		config.SnapshotInterval = snapshotInterval
+		if snapshotInterval < 5*time.Millisecond {
+			return fmt.Errorf("snapshot_interval is too low")
+		}
+		b.snapshotInterval = snapshotInterval
 	}
 
 	config.NoSnapshotRestoreOnStart = true
@@ -1140,6 +1171,11 @@ func (b *RaftBackend) SetupCluster(ctx context.Context, opts SetupOpts) error {
 		}
 	}
 
+	// Start background log truncation worker.
+	b.logTruncationStopCh = make(chan struct{})
+	b.logTruncationDoneCh = make(chan struct{})
+	go b.startLogTruncationWorker()
+
 	b.logger.Trace("finished setting up raft cluster")
 	return nil
 }
@@ -1165,6 +1201,13 @@ func (b *RaftBackend) TeardownCluster(clusterListener cluster.ClusterHook) error
 	// If we're tearing down, then we need to recreate the raftInitCh
 	b.raftInitCh = make(chan struct{})
 	b.l.Unlock()
+
+	// Stop the truncation worker and wait for any unfinished delete operations to finish.
+	if b.logTruncationStopCh != nil {
+		close(b.logTruncationStopCh)
+		<-b.logTruncationDoneCh
+		b.logTruncationStopCh = nil
+	}
 
 	if future != nil {
 		return future.Error()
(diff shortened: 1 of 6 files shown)


tsaarni:
During load testing I noticed that more frequent `DeleteRange` calls quite significantly reduced the performance. I saw >20% reduction in RPS with very write heavy example traffic (sending kv2 write operations as fast as they can be processed). As far as I've understood, this happens because `BoltStore.DeleteRange` holds a [transaction during deletion](https://github.com/hashicorp/raft-boltdb/blob/4f4a8eaa21e38ae9e0a097e0f2aa7469ba83a206/v2/bolt_store.go#L237-L241) and blocks other writes. 

If adjusting the test scenario so that the frequency of `DeleteRange` operations becomes identical, this PR performs the same as the baseline. I also noticed that deleting in smaller chunks (e.g. a busy loop that deletes the whole range 10 entries at a time) improves performance even beyond the baseline.

I guess there are two potential alternatives:
~1. Perform deletions in small chunks to reduce transaction size. It comes at the cost of increased CPU consumption.~
2. Keep `snapshot_interval` to limit how frequently `DeleteRange` is called. The downside is that predicting disk space for `raft.db` becomes difficult, and my main goal is to make disk space consumption more predictable.

**Edit:** After reconsidering I changed the truncation to use `snapshot_interval` again. It seems pointless to pay a performance penalty for something that will not meaningfully limit the disk space usage anyways. The size of writes causes just as bad peaks as bursts of writes. Fixing the latter does not make it any more predictable from users' perspective.

internal/physical/raft/raft.go
```diff
@@ -811,33 +826,48 @@ func (b *RaftBackend) applyConfigSettings(config *raft.Config) error {
...
 	snapshotIntervalRaw, ok := b.conf["snapshot_interval"]
 	if ok {
-		var err error
 		snapshotInterval, err := parseutil.ParseDurationSecond(snapshotIntervalRaw)
 		if err != nil {
 			return err
 		}
-		config.SnapshotInterval = snapshotInterval
+		if snapshotInterval < 5*time.Millisecond {
+			return fmt.Errorf("snapshot_interval is too low")
```
wslabosz-reply, line 875: do we want to document the requirement of >5ms?

physical/raft/raft_test.go
```diff
@@ -649,6 +650,97 @@ func TestRaft_LeaderConstant(t *testing.T) {
...
+		"entries below snapshot threshold, no snapshot expected")
+	assertFirstIndex(1)
+
+	// 3. Restart the cluster.
+	startCluster()
+
+	// 4. Write enough entries so the log exceeds the truncation trigger.
+	writeEntries(3)
+	require.Greater(t, logEntryCount(), uint64(snapshotThreshold+trailingLogs),
+		"entries exceed snapshot threshold, snapshot should fire")
```
wslabosz-reply:
```suggestion
		"entries exceed snapshot threshold, truncation should trigger")
```

wslabosz-reply (dismissed):
with my limited knowledge of the boltdb internals, this makes sense to me

cipherboy (dismissed):
@tsaarni Apologies about the delay. I concur with @wslabosz-reply's feedback, but I think this makes sense to me. If @JanMa gets a moment, I wouldn't mind his eyes, but otherwise I think this is good.

Do you mind if we land for 2.7+ slated for September? Or should we backport?

internal/physical/raft/raft.go
tsaarni: Fixed! 

BTW this is not a change from previous limits: there was same 5ms lower limit for this parameter when `github.com/hashicorp/raft` was driving the truncation https://github.com/hashicorp/raft/blob/80648832372a6458fdecd505d87be1cf68161491/config.go#L376-L378

physical/raft/raft_test.go
tsaarni: Fixed!

tsaarni:
> @tsaarni Apologies about the delay. I concur with @wslabosz-reply's feedback, but I think this makes sense to me. If @JanMa gets a moment, I wouldn't mind his eyes, but otherwise I think this is good.
> 
> Do you mind if we land for 2.7+ slated for September? Or should we backport?

Addressed review feedback and rebased due to conflicts. 

@cipherboy  Sorry also from my side for the delay (caused by summer break)!  For me 2.7+ would be ok.

cipherboy (approved):

internal/physical/raft/raft.go
```diff
@@ -1846,6 +1889,52 @@ func (b *RaftBackend) applyLog(ctx context.Context, command *LogData) error {
...
 
+// startLogTruncationWorker runs the background log truncation loop.
+func (b *RaftBackend) startLogTruncationWorker() {
+	defer close(b.logTruncationDoneCh)
+
+	// Truncate logs at startup to remove entries that exceeded the threshold before previous shutdown.
+	b.truncateLog()
+
+	for {
+		extra := time.Duration(rand.Int63()) % b.snapshotInterval
```
cipherboy, line 1900: nit: Usually this is a couple of percentage of an interval, not a full interval.

cipherboy (approved):
@tsaarni Thank you, appreciate this very much!

PR #3385, opened 2026-06-29
Publish betas into testing RPM/DEB repos 

### LINKED ISSUE #3384: APT repo: beta pre-release published into `stable` suite causes `apt upgrade` to install betas
(Text drafted with assistance from an AI assistant (Claude Opus 4.8); facts verified against the linked sources.)

### Describe the bug

The official APT repository at `https://pkgs.openbao.org/deb` ships pre-release
(beta) packages in the same `stable` suite as GA releases, with no metadata that
marks them as pre-release. As a result, a host configured strictly for `stable`
gets upgraded to a beta on a routine `apt upgrade`.

### Evidence

`dists/stable/Release` (Release file dated 2026-06-22) defines a single suite,
with no `NotAutomatic` / `ButAutomaticUpgrades` fields:

    Origin: OpenBao - Official
    Label: OpenBao
    Suite: stable
    Codename: stable
    Date: Mon, 22 Jun 2026 19:59:26 UTC
    Components: main

`dists/stable/main/binary-amd64/Packages` lists a beta alongside the GA versions:

    Package: openbao
    Version: 2.6.0~beta20260622
    Package: openbao
    Version: 2.5.5
    ... (2.5.4 ... 2.5.0)

    Package: openbao-hsm
    Version: 2.6.0~beta20260622
    ... (2.5.5 ... 2.5.0)

Because `2.6.0~beta20260622` sorts higher than `2.5.5`
(`dpkg --compare-versions 2.6.0~beta20260622 gt 2.5.5` → true), APT selects the
beta as the upgrade candidate even though the repo is pinned to `stable`.

### Impact

This is not merely a cosmetic version mismatch. The auto-installed package
`2.6.0~beta20260622` is a **beta** of the next minor line, pulled onto hosts
that explicitly track `stable`. The 2.6.0 line carries documented breaking
changes, e.g. the `stored_shares` parameter removed from `sys/init` /
`sys/rekey/init`, the `build_date` response field renamed to `commit_date` in
`sys/seal-status` and `sys/version-history`, and the `StoredShares` API field
removed from the init structs — plus deprecations slated for removal in 2.7.0
(built-in auto-unseal seals, the `file` storage backend, the HSM distribution).
See: https://openbao.org/community/release-notes/2-6-0/

For a security-critical component, silently upgrading a `stable` host across a
minor-version boundary onto an untested beta — breaking response fields and
removed parameters that automation/monitoring may depend on — is a serious
operational risk, not just an unexpected version number.

### To reproduce

1. Configure the repo for suite `stable`, component `main`:
   `deb [signed-by=...] https://pkgs.openbao.org/deb stable main`
2. `sudo apt update && sudo apt upgrade`

satoqz:
## Description

Publish beta releases into separate repos on pkgs.openbao.org. For RPMs, this introduces a separate "rpm-testing" repo, for DEBs, this adds a "testing" suite.

Removing v2.6-beta from stable repos is another challenge, but this at least ensures that future betas won't land in the wrong place. We can however re-publish the beta into the testing repo rather easily by just re-running the workflow, so the downloads page won't point at an empty repo. Hosted repos were originally introduced with v2.5 GA, so v2.6 beta is the only version to worry about at the moment.

## Rationale

Resolves: https://github.com/openbao/openbao/issues/3384

## Acknowledgements

       also allowed.

### DIFF
--- a/scripts/release/update-deb-repo.sh
+++ b/scripts/release/update-deb-repo.sh
@@ -5,6 +5,11 @@
 
 set -euo pipefail
 
+case "$VERSION" in
+  *-*) SUITE=testing ;;
+  *)   SUITE=stable  ;;
+esac
+
 mkdir -p build/deb/conf
 
 cat > build/deb/conf/distributions <<-EOF
@@ -14,9 +19,19 @@ cat > build/deb/conf/distributions <<-EOF
 	Codename: stable
 	Architectures: amd64 arm64 armhf ppc64el riscv64 s390x
 	Components: main
-	Description: Official APT repository for OpenBao
+	Description: Official APT repository for OpenBao releases
+	SignWith: ${GPG_FINGERPRINT}
+	Limit: 0
+
+	Origin: OpenBao - Official
+	Label: OpenBao
+	Suite: testing
+	Codename: testing
+	Architectures: amd64 arm64 armhf ppc64el riscv64 s390x
+	Components: main
+	Description: Official APT repository for OpenBao pre-releases
 	SignWith: ${GPG_FINGERPRINT}
 	Limit: 0
 EOF
 
-reprepro -b build/deb includedeb stable dist/*.deb
+reprepro -b build/deb includedeb "$SUITE" dist/*.deb
--- a/.github/workflows/release-packages.yml
+++ b/.github/workflows/release-packages.yml
@@ -95,8 +95,15 @@ jobs:
       - name: Install rpm & createrepo_c
         run: sudo apt install -y rpm createrepo-c
 
+      - name: Determine repository
+        run: |
+          case "$VERSION" in
+            *-*) echo REPO=rpm-testing ;;
+            *)   echo REPO=rpm ;;
+          esac > "$GITHUB_ENV"
+
       - name: Download repodata
-        run: aws s3 sync "s3://${BUCKET}/rpm" build/rpm --exclude \*.rpm
+        run: aws s3 sync "s3://${BUCKET}/${REPO}" build/rpm --exclude \*.rpm
 
       - name: Download packages from release
         run: ./scripts/release/download-artifacts.sh
@@ -108,4 +115,4 @@ jobs:
         run: ./scripts/release/update-rpm-repo.sh
 
       - name: Upload repository
-        run: aws s3 sync build/rpm "s3://${BUCKET}/rpm"
+        run: aws s3 sync build/rpm "s3://${BUCKET}/${REPO}"
--- a/website/src/pages/downloads.tsx
+++ b/website/src/pages/downloads.tsx
@@ -217,7 +217,7 @@ const Asset = ({ urls }) => {
   );
 };
 
-const DebRepo = ({ gpgKeyName }) => {
+const DebRepo = ({ version, gpgKeyName }) => {
   const [gpgKey, setGPGKey] = useState("");
   useEffect(() => {
     try {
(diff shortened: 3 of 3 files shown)


pree:
Thinking whether we want to add a new flag for beta releases instead if using `*-*`. But would also be fine like this if it's clear to everyone :)

Removing a release is rather complex, especially for DEB repos. I think we might want to introduce a new workflow for this, as the index needs to be recreated and signed.

satoqz:
> Thinking whether we want to add a new flag for beta releases instead if using *-*. But would also be fine like this if it's clear to everyone :)

I would want it to be auto-determined since semver is quite clear about this, and you could otherwise accidentally mismatch by providing bad inputs. But I'm not against adding a more robust check.

pree (approved):
Changes look sane to me

pree:
> adding a more robust check.

I think this is something we could add in the beginning of the pipeline. With this change we could also think of removing the two checkboxes for "Release DEB" & "Release RPM" and just couple them to the nightly checkbox

scripts/release/update-deb-repo.sh
```diff
@@ -14,9 +19,19 @@ cat > build/deb/conf/distributions <<-EOF
...
 	Architectures: amd64 arm64 armhf ppc64el riscv64 s390x
 	Components: main
-	Description: Official APT repository for OpenBao
+	Description: Official APT repository for OpenBao releases
+	SignWith: ${GPG_FINGERPRINT}
+	Limit: 0
+
+	Origin: OpenBao - Official
+	Label: OpenBao
+	Suite: testing
```
cipherboy, line 28:
```suggestion
	Suite: ${SUITE}
```

perhaps?

satoqz:
> With this change we could also think of removing the two checkboxes for "Release DEB" & "Release RPM" and just couple them to the nightly checkbox

That's already done FWIW, releasing the packages is a separate workflow now that automatically triggers when a GA or non-nightly pre-release is published.

scripts/release/update-deb-repo.sh
satoqz: We _can_ dynamically generate a single block instead of statically declaring all the available ones, but note that this file gets uploaded to the repository, too. While I believe it's not relevant to the package manager, that would mean we only ever keep the `deb/conf/distributions` for the latest-published suite around, which is weird.

pree: Alternatively the Suite: can also list multiple object iirc. so we could go `stable testing`

satoqz: There is `Suites` when configuring a repo to pull from, but the lack of an "s" made me assume that this would not work for this setting. Listing them separately also lets us set different descriptions/titles.

pree: Ah, I might have it confused with the components. As it's defining the distributions it should be separate files. This configuration is strictly for `reprepro` and we probably can exclude it from pushing. I don't know how it behaves when we only include the suite we're updating. I can test this locally tomorrow :)

satoqz: Yes, we could probably skip uploading it, too. Though in that case I'd request removing the existing file so there's no stale artifact.

pree: I'll test the behavior tomorrow and let's decide afterwards

pree: Sorry got a bit side-tracked. We should keep both in there as the db is shared between both dist's. We could split it up into being two fully separate repositories, but I don't like that approach.

wslabosz-reply (approved):

scripts/release/update-deb-repo.sh
cipherboy: @pree @satoqz Am I mistaken or does this still need to be updated? I think Pascal was suggesting we change it to  `Suite: stable testing`

satoqz: IIUC @pree meant to say "keep both sections in the file".

I don't think `Suite: stable testing` is valid syntax, it wouldn't make much sense at least. Repeating blocks lets us customize repository descriptions &c per-suite, too.

(I did of course test the changes as they stand)

cipherboy (approved):

PR #3386, opened 2026-06-29
Add plugin client handling for go-kms-wrapping/kms

satoqz:
## Description

Add support for the `kms.KMS` and `kms.Key` interfaces in `helper/kmsplugin`, adding to the existing support for `wrapping.Wrapper`.

## Rationale

This is part of foundational work for External Keys. No user-facing changes yet, so no changelog.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/helper/kmsplugin/kms.go
+++ b/internal/helper/kmsplugin/kms.go
@@ -1,0 +1,357 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package kmsplugin
+
+import (
+	"context"
+	"crypto"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/openbao/go-kms-wrapping/kms/transit/v2"
+	gkwplugin "github.com/openbao/go-kms-wrapping/plugin/v2"
+	"github.com/openbao/go-kms-wrapping/v2/kms"
+)
+
+var builtinKMSes = map[string]func() kms.KMS{
+	"transit": transit.New,
+}
+
+// OpenKMS creates a new KMS instance and calls Open with the provided options.
+// This may dispatch to either a builtin KMS or an external pluginized one.
+func (c *Catalog) OpenKMS(ctx context.Context, name string, opts *kms.OpenOptions) (s kms.KMS, err error) {
+	client, ok, err := c.getClient(name)
+	switch {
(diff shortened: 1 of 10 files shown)


satoqz:
Reloading objects derived from an object that is reloadable itself (`remoteKMS` -> `remoteKey`) was less obvious than I hoped but this seems to work.

helper/kmsplugin/kms.go
```diff
@@ -0,0 +1,359 @@
...
+			return errors.New("KMS was closed")
+		}
+
+		canary = k.kms.canary
+		behind = k.canary < k.kms.canary
+
+		return f()
+	}
+
+	if behind {
```
cipherboy: Isn't behind always false here? `call` is not yet invoked?

satoqz: true...

satoqz: Removed that special case because realistically `ErrPluginShutdown` is a local check anyway. Need to get some extra sleep.

satoqz: Will think about it some more though.

satoqz: @cipherboy give the latest update a shot, LMKWYT.

Notably, we could still compare canaries, but the code just becomes very awkward to write together with `defer`s for locks and so on. So I opted to keep it as simple as possible.

internal/helper/kmsplugin/kms.go
```diff
@@ -0,0 +1,357 @@
...
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/openbao/go-kms-wrapping/kms/transit/v2"
+	gkwplugin "github.com/openbao/go-kms-wrapping/plugin/v2"
+	"github.com/openbao/go-kms-wrapping/v2/kms"
+)
+
+var builtinKMSes = map[string]func() kms.KMS{
```
cipherboy, line 18: @satoqz Is KMIP also externalized? 🤔

cipherboy (approved):
Thanks @satoqz!

internal/helper/kmsplugin/kms.go
satoqz: @cipherboy no, but it has no KMS implementation yet.

cipherboy: @satoqz Ah, d'oh, right.

PR #3388, opened 2026-06-29
Switch to constant-time comparison recovery token

cipherboy:
## Description

Recovery mode is a privileged mode whereby operators can directly perform underlying storage operations on the instance. This requires use of recovery key shares to build a singular recovery token which protects API operations. As reported in Vault v2.0.3 changelog (**without corresponding CVE assignment**), the comparison of the recovery token was not constant-time.

We address that and expand unit tests to ensure adequate coverage of edge conditions in token behaviors.

## Rationale

Resolves: GHSA-34fc-gh42-pj53

## Acknowledgements

       also allowed.

### DIFF
--- a/http/logical.go
+++ b/http/logical.go
@@ -4,6 +4,7 @@
 package http
 
 import (
+	"crypto/subtle"
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
@@ -327,8 +328,10 @@ func handleLogicalRecovery(raw *vault.RawBackend, token *atomic.Value) http.Hand
 			respondError(w, statusCode, err)
 			return
 		}
+
 		reqToken := r.Header.Get(consts.AuthHeaderName)
-		if reqToken == "" || token.Load() == "" || reqToken != token.Load() {
+		rToken, rTokenOk := token.Load().(string)
+		if len(reqToken) == 0 || !rTokenOk || len(rToken) == 0 || subtle.ConstantTimeCompare([]byte(reqToken), []byte(rToken)) == 0 {
 			respondError(w, http.StatusForbidden, nil)
 			return
 		}
--- a/vault/external_tests/misc/recovery_test.go
+++ b/vault/external_tests/misc/recovery_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/openbao/openbao/sdk/v2/helper/logging"
 	"github.com/openbao/openbao/sdk/v2/physical/inmem"
 	"github.com/openbao/openbao/vault"
+	"github.com/stretchr/testify/require"
 )
 
 func TestRecovery(t *testing.T) {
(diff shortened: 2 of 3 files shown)


cipherboy:
This requires GHSA assignment; I'm currently leaning towards:


```
CVSS:4.0/AV:N/AC:H/AT:P/PR:N/UI:N/VC:H/VI:H/VA:H/SC:N/SI:N/SA:N
```

= `9.2 CRITICAL`

It requires preconditions (== complexity + requirements) of being in recovery mode, but once there, there's no auditing set up (I believe) and thus can be freely exploited (though, is a timing attack so that's difficult).

http/logical.go
```diff
@@ -328,7 +329,7 @@ func handleLogicalRecovery(raw *vault.RawBackend, token *atomic.Value) http.Hand
 			return
 		}
 		reqToken := r.Header.Get(consts.AuthHeaderName)
-		if reqToken == "" || token.Load() == "" || reqToken != token.Load() {
+		if len(reqToken) == 0 || len(token.Load().(string)) == 0 || subtle.ConstantTimeCompare([]byte(reqToken), []byte(token.Load().(string))) == 0 {
```
DanGhita: I would read the token once and use the variable in the "if"
`current_token: = token.Load()`

DanGhita (dismissed):

http/logical.go
cipherboy: Good point @DanGhita, will revise.

wslabosz-reply (dismissed):

wslabosz-reply (commented):
fix looks good, although test expansion doesn't really help with the issue; requires a changelog

cipherboy:
@wslabosz-reply said:

> changelog missing :(

Yeah, I'm working on CVE requests for these, will be a minute before we get changelogs (as I want both GHSA and CVE IDs in them).

vault/external_tests/misc/recovery_test.go
```diff
@@ -91,14 +92,31 @@ func TestRecovery(t *testing.T) {
...
+		// Perform an initial request: make sure nothing can be done before a token can be created.
+		client.SetToken("garbage")
+		secret, err := client.Logical().List(path.Join("sys/raw/logical", secretUUID))
+		require.Error(t, err, "expected failure with garbage token before generation")
+		require.Nil(t, secret)
+
+		client.SetToken("")
+		secret, err = client.Logical().List(path.Join("sys/raw/logical", secretUUID))
+		require.Error(t, err, "expected failure with nil token before generation")
+		require.Nil(t, secret)
```
phil9909, line 105: Too many times I have seen tests which no longer assert what they asserted when they were initially created (and some never did) via asserting "any error should have occurred"

```suggestion
		client.SetToken("garbage")
		secret, err := client.Logical().List(path.Join("sys/raw/logical", secretUUID))
		require.ErrorContains(t, err, "403", "expected failure with garbage token before generation")
		require.Nil(t, secret)

		client.SetToken("")
		secret, err = client.Logical().List(path.Join("sys/raw/logical", secretUUID))
		require.ErrorContains(t, err, "403", "expected failure with nil token before generation")
		require.Nil(t, secret)
```

vault/external_tests/misc/recovery_test.go
```diff
@@ -91,14 +92,31 @@ func TestRecovery(t *testing.T) {
...
+
 		recoveryToken := testhelpers.GenerateRoot(t, cluster, testhelpers.GenerateRecovery)
 		_, err = testhelpers.GenerateRootWithError(t, cluster, testhelpers.GenerateRecovery)
 		if err == nil {
 			t.Fatal("expected second generate-root to fail")
 		}
```
phil9909:
```suggestion
		require.ErrorContains(t, err, "attempted to generate recovery operation token when already unsealed", "expected second generate-root to fail")
```

vault/external_tests/misc/recovery_test.go
```diff
@@ -91,14 +92,31 @@ func TestRecovery(t *testing.T) {
...
+
+		secret, err = client.Logical().List(path.Join("sys/raw/logical", secretUUID))
+		require.Error(t, err, "expected failure with nil token after generation")
```
phil9909:
```suggestion
		require.ErrorContains(t, err, "403", "expected failure with nil token after generation")
```

sorry for splitting this into multiple comments, but GitHub gets confused over suggestions which span multiple hunks.

phil9909 (dismissed):

phil9909 (dismissed):

cipherboy:
> Yeah, I'm working on CVE requests for these, will be a minute before we get changelogs (as I want both GHSA and CVE IDs in them).

So much for this. Been over two weeks and @GitHub hasn't assigned CVE identifiers :laughing:

cipherboy:
DCO seems to not understand where `main` is at? :thinking: 

image

wslabosz-reply (approved):
LGTM

cipherboy:
Thanks @wslabosz-reply, merging!

PR #3389, opened 2026-06-29
Fix LIST ACL deny bypass caused by wildcards

cipherboy:
## Description

When using an ACL deny operation on a wildcard with a granting rule on a parent path, the deny operation would be ignored. For example:

    path "foo/metadata/*" { capabilities = ["list"] }
    path "foo/metadata/private/*" { capabilities = ["deny"] }

would allow a LIST operation on foo/metadata/private but correctly deny it on foo/metadata/private/my-app.

Due to a bug, this did not affect SCAN endpoints.

## Rationale

Resolves: GHSA-xp3c-3jw3-4vcr

## Acknowledgements

       also allowed.

### DIFF
--- a/vault/policy/acl_test.go
+++ b/vault/policy/acl_test.go
@@ -6,6 +6,7 @@ package policy
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


cipherboy:
I've revised this one to

```
CVSS:4.0/AV:N/AC:L/AT:P/PR:L/UI:N/VC:H/VI:N/VA:N/SC:N/SI:N/SA:N
```

 = `6 / Medium`. It has attack requirements (policy construction) and on certain paths can leak secret values (KVv2) and theoretically the attacker gains more permissions than they'd have otherwise (hence, low) but still needs to be authed (and thus not none).

vault/policy/acl_test.go
```diff
@@ -932,6 +933,84 @@ func TestACLGrantingPolicies(t *testing.T) {
...
+		} {
+			request := new(logical.Request)
+			request.Path = pt.path
+			request.Operation = operation
+
+			require.False(t, strings.HasSuffix(request.Path, "/"))
+			request.Path += "/"
+
+			authResults := acl.AllowOperation(ctx, request, false)
+			require.Equal(
```
phil9909, line 1058: If there are multiple errors, let the test continue so we can see all failures at once.

```suggestion
			assert.Equal(
```


Requires import of `github.com/stretchr/testify/assert`

phil9909 (dismissed):
I was thinking if we should use [`CutSuffix`](https://pkg.go.dev/strings#CutSuffix) instead of calling `HasSuffix` and `TrimSuffix` (twice) an store the result.

But I guess the operations should be very fast and it is probably better to keep it readable, the function is complex enough already.

vault/policy/acl_test.go
cipherboy: @phil9909 This test probably matters less and the results are usually higher quality, but I dislike this as a general pattern. Frequently I've seen failures 2...n to be noise given 1 failed. :shrug: 

Thoughts from others?

satoqz: My preference would be to maximize subtest usage instead. This way you can just use require but all cases will still run, and you can filter with `-run=...` which is especially useful when stepping through with `dlv`.

Specifically, I'd always try to have:

```go
for _, tt := range tests {
  t.Run(stuff, func(t *testing.T) {
    require.Whatever(t, ...)
  })
}
```

instead of


```go
for _, tt := range tests {
  require.Whatever(t, ...)
}
```

phil9909: I guess it depends on the problem at hand. Sometimes seeing multiple examples of failures helps you narrow down the root cause, sometimes one case is enough and the others add noise. While I agree that our tests are preeeeeetty noisy, I don't think it would be a problem here and it is easier to ignore output that is there than imagining output which is missing.

Anyhow, not a blocker, I'm fine either way (and if I ever have to refactor the function, I will just change the test locally :D)

cipherboy:
@phil9909 Note that, for a security patch, I'd like to avoid a refactor. In the future on `main` we can adjust how we trim paths if we want, but would definitely consider it outside the scope of best practice for security patching, if the problem can be solved more concisely. Some security issues can't, and such would be warranted.

vault/policy/acl_test.go
cipherboy: Let's refactor if this ever becomes a problem in the future.

wslabosz-reply (dismissed):

wslabosz-reply (approved):

cipherboy:
Thanks @wslabosz-reply, merging!

PR #3337, opened 2026-06-22
Remove builtin seals available as plugins

satoqz:
## Description

This is for v2.7, as announced in v2.6's deprecations.

Also includes the removal of the HSM distribution.

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3398, opened 2026-06-30
Add changelog entry for LDAP, Kerberos, and RADIUS (#3371)

phil9909:
Forward port of #3371, which went directly to the 2.6.x release branch.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3371.txt
+++ b/changelog/3371.txt
@@ -1,0 +1,12 @@
+```release-note:deprecation
+secret/ldap: The built-in LDAP secrets plugin has been deprecated and slated for removal from the main OpenBao binary distribution in v2.7.0. It will be included in [openbao-plugins](https://github.com/openbao/openbao-plugins) going forward.
+```
+```release-note:deprecation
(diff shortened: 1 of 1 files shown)


satoqz (approved):
I suppose we don't strictly need this entry on the main branch, but why not.

cipherboy:
@satoqz @phil9909 Based on the pattern in #3337, I think we want different changelogs on main (here we'd set them to `change` and not `deprecation`) IMO.  I was going to add them once we migrated everything to `openbao-plugins`, but now that's on hold with the RADIUS thing.

satoqz:
> @satoqz @phil9909 Based on the pattern in #3337, I think we want different changelogs on main (here we'd set them to `change` and not `deprecation`) IMO. I was going to add them once we migrated everything to `openbao-plugins`, but now that's on hold with the RADIUS thing.

Agreed, though I assumed that `change`-ones would be added with the PRs actually performing the removal. This forward-port should never come up in the v2.7 changelog.

cipherboy:
@satoqz said:

> This forward-port should never come up in the v2.7 changelog.


Right, and I think to do so, we'd close this PR? Otherwise I think we'll get duplicates? My 2c.

satoqz:
I would expect changelog-build to filter it out. But yes, minimal benefit to having it forward-ported.

PR #3401, opened 2026-06-30
Fix ACL templates allow wildcard chars in substitution

phil9909:
## Description

Disallow wildcard characters in ACL template substitutions, meaning:

`path "kv/*" { capabilities = [ "{{ ... }}" ] }`  evaluating to `path "kv/*" { capabilities = [ 
"read" ] }` is fine

But this

`path "kv/{{ ... }}" { capabilities = [ "read" ] }`  evaluating to `path "kv/*" { capabilities = [ "read" ] }` will error

## Rationale

Prevent wildcard injections

## Acknowledgements

       also allowed.


sdk/helper/identitytpl/templating_test.go
```diff
@@ -380,6 +380,43 @@ func TestPopulate_Basic(t *testing.T) {
...
+		},
+		{
+			mode:     ACLTemplating,
+			name:     "wildcard plus in template substitution (acl)",
+			input:    `path "bad" {{identity.entity.metadata.x}}`,
+			metadata: map[string]string{"x": "this + is also not ok"},
+			err:      ErrTemplateWildcard,
+		},
+		{
+			mode:     JSONTemplating,
```
cipherboy: Does this part matter? 🤔 I don't see why JSON templates would need to forbid `*` characters. This would be in OIDC JWTs, and I'm not sure we can really make meaningful claims as to whether or not they'd treat `*` or `+` characters as special?

cipherboy (commented):
@phil9909 One comment on scope of the fix.

sdk/helper/identitytpl/templating_test.go
phil9909: Fair, I thought we had support for JSON policies, but we don't.

cipherboy: @phil9909 We do have support for JSON-formatted policies, but they're just a delivery vehicle. They still invoke the `ACLTemplating` helper: template substitution is within paths and doesn't reflect on the overarching JSON object. `JSONTemplating` is used within the identity OIDC system, for using OpenBao as an IDP, building external ID tokens.

vault/policy/policy.go
```diff
@@ -94,30 +94,32 @@ var cap2Int = map[string]uint32{
...
+	DataVersion    int
+	CASRequired    bool
+	Paths          []*PathRules `hcl:"-"`
+	Raw            string
+	Type           Type
+	Templated      bool
+	Expiration     time.Time
+	Modified       time.Time
+	Namespace      *namespace.Namespace
+	AllowWildcards bool
```
cipherboy: If we add this value, and I'd still be interested in hearing arguments of why it should exist on the _policy_ level given that there's an inherent mismatch between policy/identity subsystem operational knowledge in some deployments... ...I think it should be called something more descriptive. `allow_identity_templating_wildcards` perhaps. As phrased, this seems like it would disallow `path "*" { ... }` altogether, which it doesn't. 

(I would've expected this to be a single global server config value tbh.)

wslabosz-reply (dismissed):

builtin/logical/pki/path_roles.go
```diff
@@ -1703,6 +1724,7 @@ func (r *roleEntry) ToResponseData() map[string]interface{} {
 		"allow_localhost":                    r.AllowLocalhost,
 		"allowed_domains":                    r.AllowedDomains,
 		"allowed_domains_template":           r.AllowedDomainsTemplate,
+		"allow_globs_in_substitutions":       r.AllowGlobsInSubstitutions,
```
cipherboy, line 1727: I don't know if I'm fully sold on the `substitutions` phrasing here, I do think `identity_templates` is most clear.

Thoughts?

cipherboy (approved):
I'll propose a PR to revise the names later. I think this looks fairly good, thanks @phil9909!

builtin/logical/pki/path_roles.go
phil9909: I used `substitutions` to signal this is not about the template itself, but rather the strings that get inserted. I also think the name is :hankey: but it is the best thing I could come up with that is somewhat accurate.

Suggesting a name for this could be a good job for some LLM (I found that they are pretty good at hallucinating APIs that don't exists but sound good :laughing:). IMNAL, but this should be fine from a legal/copyright point of view.

cipherboy: @phil9909 But the correct name of the feature is... templating: https://openbao.org/docs/concepts/policies/#templated-policies

So calling it substitutions is introducing a second name for the same thing, which will lead to confusion.

satoqz: `allow_templated_globs` perhaps?

template**d** to imply that they were inserted via templating, not part of the template itself (substituted, if you will).

cipherboy: @satoqz Hmm, so that'd be `allow_templated_path_segments` for paths, but what about SSH? :thinking: 

`allow_multiple_templated_values`? It is used for both `allowed_users` and `allowed_domains`, and `entity` is already highly overloaded.

PR #3403, opened 2026-06-30
Add profile creation to CLI, API

cipherboy:
## Description

This allows operators to more easily construct profiles from request snippets executed via the CLI or API client.

## Rationale

I believe @satoqz mentioned this would be a good feature at some point in time, but not sure I recall where :laughing: 

Resolves: (n/a?)

## Acknowledgements

       also allowed.

### DIFF
--- a/api/output_profile.go
+++ b/api/output_profile.go
@@ -1,0 +1,240 @@
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	retryablehttp "github.com/hashicorp/go-retryablehttp"
+)
+
+const (
+	ErrOutputProfileRequest = "output a profile, please"
+)
+
+var LastOutputProfileError *OutputProfileError
+
+type OutputProfileError struct {
+	*retryablehttp.Request
+	finalProfile string
+}
+
+func (d *OutputProfileError) Error() string {
+	if d.finalProfile == "" {
+		cs, err := d.buildProfile()
+		if err != nil {
+			return err.Error()
+		}
+		d.finalProfile = cs
+	}
+
+	return ErrOutputProfileRequest
+}
+
+func (d *OutputProfileError) HCLString() (string, error) {
+	if d.finalProfile == "" {
+		cs, err := d.buildProfile()
+		if err != nil {
+			return "", err
+		}
+		d.finalProfile = cs
+	}
+	return d.finalProfile, nil
+}
+
+func (d *OutputProfileError) operation() string {
+	switch d.Method {
+	case "GET":
+		params := d.URL.Query()
+		switch {
+		case params.Get("list") != "":
+			return "list"
+		case params.Get("scan") != "":
+			return "scan"
+		default:
+			return "read"
+		}
+	case "POST", "PUT":
+		// Cannot disambiguate create versus update here; default to update.
+		return "update"
+	case "DELETE":
+		return "delete"
+	case "PATCH":
+		return "patch"
+	}
+
+	return "<unknown>"
+}
+
+func (d *OutputProfileError) path() string {
+	return strings.TrimPrefix(d.URL.Path, "/v1")
+}
+
+func (d *OutputProfileError) token() string {
+	return d.Header.Get(AuthHeaderName)
+}
+
+func (d *OutputProfileError) headers() (map[string]string, error) {
+	headers := make(map[string]string, len(d.Header))
+	for key, values := range d.Header {
+		if key == AuthHeaderName || key == RequestHeaderName {
+			// Token is specified via a dedicated field and the X-Vault-Request
+			// header is unnecessary as we know we're not likely taking to the
+			// agent.
+			continue
+		}
+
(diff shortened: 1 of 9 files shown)


api/output_profile.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	return data, nil
+}
+
+func (d *OutputProfileError) indent(prefix string, value string) string {
+	return strings.TrimSuffix(strings.Join(strings.Split(value, "\n"), "\n"+prefix), prefix)
+}
+
+// formatHCL exists because hashicorp/hcl (on the v1 branch) lacks an encoder
```
cipherboy, line 133: This is garbage, but I don't know any reasonable ways around it. Thoughts from anyone?

command/base.go
```diff
@@ -490,6 +494,14 @@ func (c *BaseCommand) flagSet(bit FlagSetBit) *FlagSets {
 					"policy that would be required to run this command and exit.",
 			})
 
+			f.BoolVar(&BoolVar{
+				Name:    "output-profile",
+				Target:  &c.flagOutputProfile,
+				Default: false,
+				Usage: "Instead of executing the request, print an example HCL " +
+					"profile that would be required to run this command and exit.",
```
satoqz:
```suggestion
				Usage: "Instead of executing the request, print an equivalent HCL " +
					"profile that replicates this command and exit.",
```

command/main.go
```diff
@@ -31,17 +31,22 @@ type VaultUI struct {
...
+	globalFlagOutputPolicy,
+	globalFlagOutputProfile,
+	globalFlagFormat,
+	globalFlagDetailed,
 }
 
 // setupEnv parses args and may replace them and sets some env vars to known
 // values based on format options
-func setupEnv(args []string) (retArgs []string, format string, detailed bool, outputCurlString bool, outputPolicy bool) {
+func setupEnv(args []string) (retArgs []string, format string, detailed bool, outputCurlString bool, outputPolicy bool, outputProfile bool) {
```
satoqz: I humbly request to refactor this to return a struct, if possible without much pain.

api/output_profile.go
```diff
@@ -0,0 +1,235 @@
...
+		for index, item := range value.([]any) {
+			formatted, err := d.formatHCL(item)
+			if err != nil {
+				return "", fmt.Errorf("at index %d: %w", index, err)
+			}
+
+			if index > 0 {
+				result += ",\n"
+			}
+			result += "   " + d.indent("   ", formatted) + ",\n"
```
satoqz: Any particular reason to use 3 spaces? HCL examples in our docs seem to favor 2 spaces, so I'd stick to that.

(An `indent` constant or helper function may be useful?)

api/output_profile.go
satoqz: I suppose it's fine given it's a constrained use case. Long term we probably want to think about migrating anyhow.

command/main.go
satoqz: You can even avoid naming it by using the language as must be intended:

```go
func setupEnv(args []string) (ret struct {
	// fields...
}) {
	return ret
}
```

api/output_profile.go
cipherboy: I don't know that we can, I think the two are incompatible in ways that would break users. :cry:

api/output_profile.go
cipherboy: `hcl` -> `   ` :laughing:

api/output_profile.go
cipherboy: https://github.com/hashicorp/nomad/issues/11759 is one such issue in general about theoretical differences, though I've not done the work to figure out again where that was true.

cipherboy: Which, actually says that I might be outputing invalid HCLv1? :thinking:

phil9909:
Missed your PR, here is my work (it less complete): https://github.com/openbao/openbao/blob/248e1faa0beafa6a31b46f6731dd1e63b587278b/api/output_string.go#L45-L84

command/main.go
cipherboy: @satoqz Oh that's nasty:

```go
func setupEnv(args []string) (ret struct {
   ...
},
) {
```

Thanks `gofumpt`. :disappointed:

satoqz:
JSON support might be good to include?

cipherboy:
@satoqz I've proposed a parameter that might make sense in this regard: [#openssf-openbao-discussion > CLI: generate request hcl or json @ 💬](https://linuxfoundation.zulipchat.com/#narrow/channel/529890-openssf-openbao-discussion/topic/CLI.3A.20generate.20request.20hcl.20or.20json/near/609848305)

LMKWYT and I can probably refactor both CLI and API, though we'll preserve the original flags.

I wonder if we could go further and output Go API as a transpiled output target. :thinking: But for another time.

cipherboy:
@satoqz @phil9909 All in favor of the `-print-as=...` suggestion from the above zulip? If so, will adjust.

PR #3405, opened 2026-06-30
Add blog post on paginated lists

cipherboy:
This begins a new blog series on new features in OpenBao, starting with our very first feature: paginated lists.

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3406, opened 2026-06-30
Add parameter names to sdk/logical interfaces

cipherboy:
## Description

As mentioned in the blog post PR review, the lack of parameter names is criminal.

See also: https://github.com/openbao/openbao/pull/3405

## Acknowledgements

       also allowed.

### DIFF
--- a/sdk/logical/managed_key.go
+++ b/sdk/logical/managed_key.go
@@ -32,7 +32,7 @@ type ManagedKey interface {
 	UUID() string
 	// Present returns true if the key is established in the KMS.  This may return false if for example
 	// an HSM library is not configured on all cluster nodes.
-	Present(ctx context.Context) (bool, error)
+	Present(ctx context.Context) (present bool, err error)
 
 	// AllowsAll returns true if all the requested usages are supported by the managed key.
 	AllowsAll(usages []KeyUsage) bool
@@ -76,13 +76,13 @@ type ManagedKeySystemView interface {
 
 type ManagedAsymmetricKey interface {
 	ManagedKey
-	GetPublicKey(ctx context.Context) (crypto.PublicKey, error)
+	GetPublicKey(ctx context.Context) (pub crypto.PublicKey, err error)
 }
 
 type ManagedKeyLifecycle interface {
 	// GenerateKey generates a key in the KMS if it didn't yet exist, returning the id.
 	// If it already existed, returns the existing id.  KMSKey's key material is ignored if present.
-	GenerateKey(ctx context.Context) (string, error)
+	GenerateKey(ctx context.Context) (uuid string, err error)
 }
 
 type ManagedSigningKey interface {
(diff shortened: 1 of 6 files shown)


sdk/logical/storage.go
```diff
@@ -66,10 +86,12 @@ func StorageEntryJSON(k string, v interface{}) (*StorageEntry, error) {
 	}, nil
 }
 
+// ClearableView is a subset of Storage which is necessary for a storage view
```
cipherboy, line 89: TBH I'm not actually sure where we legitimately have a `ClearableView` that isn't also a `Storage` implementation, but I guess that's neither here nor there.

satoqz (approved):
Oh, I only meant to complain about storage interface parameters being a little opaque, but this is good, too.

I don't mind interfaces without parameter names generally as long as the type names are sufficient explanation.

wslabosz-reply (approved):

sdk/logical/managed_key.go
```diff
@@ -76,13 +76,13 @@ type ManagedKeySystemView interface {
...
 	ManagedKey
-	GetPublicKey(ctx context.Context) (crypto.PublicKey, error)
+	GetPublicKey(ctx context.Context) (pub crypto.PublicKey, err error)
 }
 
 type ManagedKeyLifecycle interface {
 	// GenerateKey generates a key in the KMS if it didn't yet exist, returning the id.
 	// If it already existed, returns the existing id.  KMSKey's key material is ignored if present.
-	GenerateKey(ctx context.Context) (string, error)
+	GenerateKey(ctx context.Context) (uuid string, err error)
```
phil9909, line 85: Is this always a UUID?

```suggestion
	GenerateKey(ctx context.Context) (id string, err error)
```

sdk/logical/managed_key.go
```diff
@@ -92,34 +92,34 @@ type ManagedSigningKey interface {
...
 	// Verify verifies the provided signature against the value.  The SignerOpts param must provide the hash function
 	// that generated the value (if any).
 	// If true is returned the signature is correct, false otherwise.
-	Verify(ctx context.Context, signature, value []byte, opts crypto.SignerOpts) (bool, error)
+	Verify(ctx context.Context, signature, value []byte, opts crypto.SignerOpts) (ok bool, err error)
 
 	// GetSigner returns an implementation of crypto.Signer backed by the managed key.  This should be called
 	// as needed so as to use per request contexts.
-	GetSigner(context.Context) (crypto.Signer, error)
+	GetSigner(context.Context) (signer crypto.Signer, err error)
```
phil9909, line 104:
```suggestion
	GetSigner(ctx context.Context) (signer crypto.Signer, err error)
```

sdk/logical/system_view.go
```diff
@@ -48,41 +48,41 @@ type SystemView interface {
...
+	EntityInfo(entityID string) (entity *Entity, err error)
 
 	// GroupsForEntity returns the group membership information for the provided
 	// entity id
-	GroupsForEntity(entityID string) ([]*Group, error)
+	GroupsForEntity(entityID string) (groups []*Group, err error)
 
 	// PluginEnv returns Vault environment information used by plugins
-	PluginEnv(context.Context) (*PluginEnvironment, error)
+	PluginEnv(context.Context) (env *PluginEnvironment, err error)
```
phil9909, line 82:
```suggestion
	PluginEnv(ctx context.Context) (env *PluginEnvironment, err error)
```

sdk/logical/system_view.go
```diff
@@ -48,41 +48,41 @@ type SystemView interface {
...
 
 	// VaultVersion returns the version string for the currently running Vault.
-	VaultVersion(context.Context) (string, error)
+	VaultVersion(context.Context) (version string, err error)
```
phil9909, line 85:
```suggestion
	VaultVersion(ctx context.Context) (version string, err error)
```

sdk/logical/storage_transactions.go
```diff
@@ -35,11 +35,11 @@ type Transaction interface {
 	// Commit a transaction; this is equivalent to Rollback on a read-only
 	// transaction. Either Commit or Rollback must be called to release
 	// resources.
-	Commit(context.Context) error
+	Commit(ctx context.Context) error
```
phil9909, line 38:
```suggestion
	Commit(ctx context.Context) (err error)
```

sdk/logical/storage_transactions.go
```diff
@@ -35,11 +35,11 @@ type Transaction interface {
...
 
 	// Rollback a transaction, preventing any changes from being persisted.
 	// Either Commit or Rollback must be called to release resources.
-	Rollback(context.Context) error
+	Rollback(ctx context.Context) error
```
phil9909, line 42:
```suggestion
	Rollback(ctx context.Context) (err error)
```

phil9909 (approved):
Added some suggestions for methods which only return an `error` -> `(err error)` or only take a `context.Context` -> `ctx context.Context` for consistency.

Also is `GenerateKey` guaranteed to return a UUID?

Feel free to ignore.

satoqz:
`error -> (err error)` would be overdoing it for the sake of consistency IMO.

cipherboy:
@satoqz Same goes for context if it is the only parameter, imo.

sdk/logical/managed_key.go
cipherboy: @phil9909 Yes, I think from a Vault Enterprise PoV, managed keys only ever return uuids as key identifiers.

cipherboy:
@satoqz Have we thought about removing `sdk/logical/managed_keys.go` from our distribution anyways? It doesn't go over GRPC bridge and doesn't work with external plugins, only builtin (VEnt) plugins.

satoqz:
> @satoqz Have we thought about removing `sdk/logical/managed_keys.go` from our distribution anyways? It doesn't go over GRPC bridge and doesn't work with external plugins, only builtin (VEnt) plugins.

I would be okay with removing it, I'll add something entirely separate for external keys, anyway.

Perhaps we'd want to keep it for SDK compatibility though? Historically you've asked to hold back api/sdk removals for this cause. So, your call. (Then again, I guess, as you say, it works with builtins only).

cipherboy:
@satoqz The nebulous line I'd perhaps draw is if they're actually useful or not. People could build plugins with our SDK that ran on VEnt in DR/PS modes and they'd have the constants to do so.

If they use Managed Key APIs regardless of source, they don't do anything, as they only work if they're builtin binaries (not implemented for external plugins), and so the runtime use will fail regardless of which SDK. And I highly doubt Hashi will use our SDK to build their builtin plugins against.

satoqz:
I find all scenarios rather unlikely and would vote to remove.

PR #3409, opened 2026-07-01
Propagate keyring updates of all namespaces to standby nodes

### LINKED ISSUE #3379: Standby nodes missing keyring upgrades for namespaces
**Describe the bug**
With introduction of per-ns sealing (v2.6 beta) we can have more than one active barrier at the time (which means more than one keyring). Current keyring (term) upgrade logic wasn't adjusted for that, resulting in missing upgrades which in turn renders specific (sealable) namespaces unusable on a standby nodes (this will also affect whatever new leader will be after next stepdown & election).

**To Reproduce**
(run OpenBao with HA setup)
```
$ bao namespace create -seal=config.hcl test
(...)
$ bao namespace unseal test
(...)

$ bao operator rotate -ns-test
Success! Rotated key

Key Term            4
Install Time        26 Jun 26 15:03 UTC
Encryption Count    1
```

now while calling standby node:
```
Error reading sys/raw/namespaces/(...): Error making API request.

URL: GET http://127.0.0.1:8200/v1/sys/raw/namespaces/(...)

Code: 400. Errors:

* no decryption key available for term 4

$ bao operator key-status -ns=test
Key Term            3
Install Time        26 Jun 26 14:56 UTC
Encryption Count    0
```
notice stale key status, also `Encryption Count` is incorrectly reflected as # of encryptions are only counted within active node in-mem.

**Expected behavior**
Rotation of the namespaces keyring should be propagated to the standby nodes (proper invalidation logic).

**Environment:**
v2.6 Beta
v2.6 Beta

**Additional context**
This only applies to HA-enabled setups.

wslabosz-reply:
## Description
This PR adjusts the existing upgrade polling logic for keyring updates (and raft TLS) for a established (with v2.5 release) invalidation pattern, which would trigger on a `core/upgrade/...` write/delete and properly install new keyring term on a standby node, also respecting the source namespace.

## Rationale
Sealable namespace in a HA raft cluster would be rendered unusable after certain period due to missing keyring update, all new (namespace-owned) entries would be unreadable on standby node.

Resolves: #3379

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3409.txt
+++ b/changelog/3409.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Propagate keyring upgrades of sealable namespaces to read-enabled standby nodes.
+```
--- a/vault/barrier/aes_gcm.go
+++ b/vault/barrier/aes_gcm.go
@@ -35,10 +35,16 @@ const (
 	// termSize the number of bytes used for the key term.
 	termSize = 4
 
+	// AutoRotateCheckInterval is used as a timestep between
+	// automatic barrier key rotation checks.
 	AutoRotateCheckInterval = 5 * time.Minute
 	legacyRotateReason      = "legacy rotation"
+
 	// The keyring is persisted before the root key.
 	keyringTimeout = 1 * time.Second
+
+	// oneYear is the duration after which we deem keyring term 'legacy'.
+	oneYear = 24 * 365 * time.Hour
 )
 
 // Versions of the AESGCM storage methodology
@@ -112,6 +118,8 @@ type AESGCMBarrierTransaction struct {
 }
 
 func (b *AESGCMBarrier) RotationConfig() (kc KeyRotationConfig, err error) {
+	b.l.RLock()
+	defer b.l.RUnlock()
 	if b.keyring == nil {
 		return kc, errors.New("keyring not yet present")
 	}
@@ -124,7 +132,6 @@ func (b *AESGCMBarrier) SetRotationConfig(ctx context.Context, rotConfig KeyRota
 	rotConfig.Sanitize()
 	if !rotConfig.Equals(b.keyring.rotationConfig) {
 		b.keyring.rotationConfig = rotConfig
-
 		return b.persistKeyring(ctx, b.keyring)
 	}
 	return nil
@@ -1203,34 +1210,30 @@ func (b *AESGCMBarrier) TotalLocalEncryptions() int64 {
 }
 
 func (b *AESGCMBarrier) CheckBarrierAutoRotate(ctx context.Context) (string, error) {
-	const oneYear = 24 * 365 * time.Hour
 	reason, err := func() (string, error) {
 		b.l.RLock()
 		defer b.l.RUnlock()
-		if b.keyring != nil {
-			// Rotation Checks
-			var reason string
 
-			rc, err := b.RotationConfig()
-			if err != nil {
-				return "", err
-			}
-
-			if !rc.Disabled {
-				activeKey := b.keyring.ActiveKey()
-				ops := b.encryptions()
-				switch {
-				case activeKey.Encryptions == 0 && !activeKey.InstallTime.IsZero() && time.Since(activeKey.InstallTime) > oneYear:
-					reason = legacyRotateReason
-				case ops > rc.MaxOperations:
-					reason = "reached max operations"
-				case rc.Interval > 0 && time.Since(activeKey.InstallTime) > rc.Interval:
-					reason = "rotation interval reached"
-				}
-			}
-			return reason, nil
+		if b.keyring == nil {
+			return "", nil
 		}
-		return "", nil
+
+		rc := b.keyring.rotationConfig.Clone()
+		if rc.Disabled {
+			return "", nil
+		}
+
+		activeKey := b.keyring.ActiveKey()
+		switch {
+		case activeKey.Encryptions == 0 && !activeKey.InstallTime.IsZero() && time.Since(activeKey.InstallTime) > oneYear:
+			return legacyRotateReason, nil
+		case b.encryptions() > rc.MaxOperations:
+			return "reached max operations", nil
+		case rc.Interval > 0 && time.Since(activeKey.InstallTime) > rc.Interval:
+			return "rotation interval reached", nil
+		default:
+			return "", nil
+		}
 	}()
 	if err != nil {
 		return "", err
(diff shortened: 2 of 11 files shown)


changelog/3409.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+core: Properly propagated all namespaces keyring upgrades to standby nodes.
```
satoqz:
```suggestion
core: Propagate keyring upgrades of sealable namespaces to standby nodes.
```

vault/barrier/aes_gcm.go
```diff
@@ -35,10 +35,16 @@ const (
...
 
+	// AutoRotateCheckInterval is used as a timestep between
+	// automatic barrier key rotation checks.
 	AutoRotateCheckInterval = 5 * time.Minute
 	legacyRotateReason      = "legacy rotation"
+
 	// The keyring is persisted before the root key.
 	keyringTimeout = 1 * time.Second
+
+	// oneYear is the duration we after we deem keyring term 'legacy'.
```
satoqz:
```suggestion
	// oneYear is the duration after which we deem keyring term 'legacy'.
```

vault/barrier/aes_gcm.go
```diff
@@ -1239,14 +1245,14 @@ func (b *AESGCMBarrier) CheckBarrierAutoRotate(ctx context.Context) (string, err
 		return reason, nil
 	}
 
-	b.l.Lock()
-	defer b.l.Unlock()
+	// either rotation is disabled or we do not fulfill
+	// any of the rotation conditions.
 	if b.keyring != nil {
```
satoqz, line 1249: I believe you're supposed to nil-check the keyring only while holding the lock (above, too).

`RotationConfig()` also seems to be missing a read lock.

vault/ha.go
```diff
@@ -509,18 +505,6 @@ func (c *Core) runHALoopOnce(manualStepDownCh chan struct{}, stopCh, restartCh <
...
-		keyRotateStop := make(chan struct{})
-
-		g.Add(func() error {
-			c.periodicCheckKeyUpgrades(context.Background(), keyRotateStop)
-			return nil
-		}, func(error) {
-			close(keyRotateStop)
-			c.logger.Debug("shutting down periodic key rotation checker")
-		})
-	}
```
satoqz, line 523: IIUC this loop also is started on non-read-enabled nodes that don't run invalidation manager.

Are we good to ignore keyring upgrades on those? I doubt it?

wslabosz-reply: adjusted, basically we run the invalidation for all namespaces while running read-enabled standbys, for read-disabled we retain the previous existing behaviour

vault/barrier/aes_gcm.go
```diff
@@ -1203,34 +1210,34 @@ func (b *AESGCMBarrier) TotalLocalEncryptions() int64 {
...
-				case activeKey.Encryptions == 0 && !activeKey.InstallTime.IsZero() && time.Since(activeKey.InstallTime) > oneYear:
-					reason = legacyRotateReason
-				case ops > rc.MaxOperations:
-					reason = "reached max operations"
-				case rc.Interval > 0 && time.Since(activeKey.InstallTime) > rc.Interval:
-					reason = "rotation interval reached"
-				}
-			}
-			return reason, nil
+		rc, err := b.RotationConfig()
```
satoqz: This would now be a recursive read lock.

> If any goroutine calls [RWMutex.Lock](https://pkg.go.dev/sync#RWMutex.Lock) while the lock is already held by one or more readers, concurrent calls to [RWMutex.RLock](https://pkg.go.dev/sync#RWMutex.RLock) will block until the writer has acquired (and released) the lock, to ensure that the lock eventually becomes available to the writer. Note that this prohibits recursive read-locking. A [RWMutex.RLock](https://pkg.go.dev/sync#RWMutex.RLock) cannot be upgraded into a [RWMutex.Lock](https://pkg.go.dev/sync#RWMutex.Lock), nor can a [RWMutex.Lock](https://pkg.go.dev/sync#RWMutex.Lock) be downgraded into a [RWMutex.RLock](https://pkg.go.dev/sync#RWMutex.RLock).

satoqz (commented):
Looks correct overall.

vault/barrier/aes_gcm.go
wslabosz-reply: oh right, fixed

satoqz (dismissed):

phil9909:
As this is part of the 2.6.0 milestone, I will add backport labels

vault/ha.go
```diff
@@ -611,7 +612,7 @@ func (c *Core) waitForLeadership(manualStepDownCh, stopCh <-chan struct{}) {
 
 		// If possible, unseal in read-only mode and start acting as a
 		// read-enabled standby.
```
cipherboy, line 614: @wslabosz-reply This is nice, but what does it actually accomplish for us?

vault/seal_manager.go
```diff
@@ -855,3 +855,23 @@ func (c *Core) NamespaceKeys(rpcCtx context.Context, namespaces []string) (map[s
...
+
+// checkKeyringUpgrade is used to rotate the keyring to the new term
+// if there have been any key rotations performed on a leader node.
+func (c *Core) checkKeyringUpgrade(ctx context.Context, nsPath string) error {
+	b := c.sealManager.NamespaceBarrier(nsPath)
+	if b == nil {
+		return fmt.Errorf("couldn't retrieve barrier for a namespace: %q", nsPath)
+	}
+
+	didUpgrade, newTerm, err := b.CheckUpgrade(ctx)
```
cipherboy: @wslabosz-reply I do kinda wonder if we should loop like we do in `checkRootKeyringUpgrade`. That'd also potentially make the code a little cleaner between the two. 🤷

cipherboy (dismissed):
@wslabosz-reply Two comments and I think we're good to merge.

vault/ha.go
wslabosz-reply: I'm not sure what you are asking about here, but through not invoking `c.StandbyReadsEnabled()` directly here but rather passing it with parameters we ensure that no hypothetical "hot config reload" results in `periodicCheckKeyringUpgrades` and `waitForLeadership` diverging with their `StandbyReadsEnabled()` value.

cipherboy: Ahh, I see. I had forgotten that `disable_standby_reads` was hot-reload-able.

cipherboy: Thank you!

vault/seal_manager.go
wslabosz-reply: @cipherboy adjusted

satoqz (approved):

cipherboy (approved):
Thank you! New test looks useful.

cipherboy:
Backport in #3491.

PR #3424, opened 2026-07-02
Update mount, auth, and namespace limits

cipherboy:
## Acknowledgements

       also allowed.


wslabosz-reply (approved):

satoqz (approved):

phil9909:
@cipherboy You added the `backport-2.5.x` label here, I don't think this is needed right?

cipherboy:
@phil9909 Its an interesting question from a backport philosophy perspective. This limit change isn't new; at some point old docs will atrophy, but we should probably maintain them at least at n-1, n-2 if we can.

PR #3433, opened 2026-07-06
fix(command/server): preserve missing config fields in Config.Merge() and Config.Sanitized()

### LINKED ISSUE #3432: Some configuration fields are silently dropped when multiple config sources are provided
**Describe the bug**
When the server is started with multiple `-config` paths, `Config.Merge()` is called for each additional source. The merged result is initialized from `NewConfig()` where all booleans default to `false`, and only fields explicitly handled in `Merge()` are preserved. Three fields were missing from `Merge()`, so their values are silently lost:

- `disable_standby_reads`
- `allow_unauthenticated_workflows`
- `unsafe_relative_paths`

The same fields are also absent from `/sys/config/state/sanitized`, making it harder to verify their effective value on a running node.

**To Reproduce**
1. Start a server with two config sources. The second source can be any valid config file or even an empty directory:
    ```sh
    bao server -config=/path/to/main.hcl -config=/path/to/extra-config/
    ```
2. On a standby node, check the startup logs for the following line:
    ```
    enabling horizontal scalability (reads)
    ```
3. Despite `disable_standby_reads = true`, this line appears, confirming that standby reads are enabled.

**Expected behavior**
`disable_standby_reads = true` (and the other affected fields) should be preserved after merging, and the line above should not appear in the logs on standby nodes.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): v2.5.5
* OpenBao CLI Version (retrieve with `bao version`): v2.5.5
* Server Operating System/Architecture: Linux / amd64

```hcl
# main.hcl
disable_standby_reads = true
```

**Additional context**
A fix is ready and will be submitted as a PR.

sivapalan:
## Description

`Config.Merge()` only preserves fields that are explicitly listed in its body. Fields added to the `Config` struct at a later point without a corresponding update to `Merge()` are silently dropped, defaulting to their zero value regardless of what either source config specifies.

This PR adds the three fields that were missing:

- `DisableStandbyReads`
- `AllowUnauthenticatedWorkflows`
- `UnsafeRelativePaths`

`Config.Sanitized()` is also updated to include these fields in its output, so that `bao read sys/config/state/sanitized` can be used to verify the effective merged configuration.

A regression test (`TestConfigMergePreservesFields`) is added to verify fields are preserved through a merge.

## Rationale

When OpenBao is started with multiple `-config` paths, `ParseServerConfig` calls `Config.Merge()` to combine them. Any field absent from `Merge()` silently defaults to `false`/zero, overriding the user's intent with no warning.

Resolves: #3432 

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3433.txt
+++ b/changelog/3433.txt
@@ -1,0 +1,6 @@
+```release-note:bug
+command/server: Fix boolean config fields being dropped when multiple `-config` paths are used
+```
+```release-note:improvement
+command/server: Include `disable_standby_reads`, `allow_unauthenticated_workflows`, and `unsafe_relative_paths` in sanitized config output
+```
--- a/command/server/config.go
+++ b/command/server/config.go
@@ -1413,6 +1413,10 @@ func (c *Config) Sanitized() map[string]interface{} {
 
 		"unsafe_allow_api_audit_creation": c.UnsafeAllowAPIAuditCreation,
 		"allow_audit_log_prefixing":       c.AllowAuditLogPrefixing,
+
+		"disable_standby_reads":           c.DisableStandbyReads,
+		"allow_unauthenticated_workflows": c.AllowUnauthenticatedWorkflows,
+		"unsafe_relative_paths":           c.UnsafeRelativePaths,
 	}
 	maps.Copy(result, sharedResult)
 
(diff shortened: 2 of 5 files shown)


wslabosz-reply:
thanks @sivapalan this makes sense to me, although please take a look at failing tests

wslabosz-reply (approved):
LGTM

wslabosz-reply:
thanks @sivapalan!

Issue #3399, opened 2026-06-30
Cease usage of HashiCorp's container image mirror


satoqz:
Tests pull several containers from `docker.mirror.hashicorp.services`. At the time of writing, this registry is entirely down, and as a result almost all test shards are going red across all PRs.

Most images pulled from this mirror are common software that is trivially available elsewhere. We should drop any dependencies on the HashiCorp mirror and create our own on `ghcr.io`, mirroring replacements from `docker.io` (to evade pull limits, and for hopefully decent intra-GitHub connectivity).

I propose we attach those images to a separate repo that can also hold any Dockerfiles for any custom images we require.

cipherboy:
Dupe of #98 😆 I'll close that one. (Edit: apparently I already closed that one 2 weeks ago 😆) 

I think GHCR probably makes the most sense.

satoqz:
The original issue was a suggestion, this one's a demand 😆

pree:
There is also the [pull cache mirror from Google](https://docs.cloud.google.com/artifact-registry/docs/pull-cached-dockerhub-images) which could be used. Though they don't guarantee all images are available, only frequently used ones.

PR #3434, opened 2026-07-06
helper/docker: reduce pulls

phil9909:
## Description

Currently our docker testhelper will always try to pull the target image before running it.

An easy fix would be to check if the image is present locally and only pull if not. I don't think this is a good idea, because this means that a developer running the tests locally might get a totally different version than the CI will use (especially in cases were we use the `latest` tag)

So with this PR I tried to implement a "smart" pulling behavior: skip pulling if the local copy is "reasonably fresh":

- check an in-process map, if we have pulled already => return cached result
- check if the image exists in daemon => if no: pull
- check if image build date < 24h ago => if yes: skip pull
- find any image pull events in the daemon event log (last 24h) => if found: skip pull

if any of the steps above fails: pull (and print a warning)

This might still result in different major versions of a container image running in CI and locally, but only for about 24 hours after the version bump (and running `latest` will always have this risk, if you don't pull at the exact same time), which to me seams like a reasonable tradeoff.

## Rationale

Always pulling has some drawbacks:

- Some registries (famously Docker Hub) impose rate limits on pulls. For Docker Hub it does not matter if you pull 50GB or simply get a response that your locally cached image is still fresh: it does count against your limit ([200 per free user or 100 per unauthenticated IP](https://docs.docker.com/docker-hub/usage/))
- Stability: Assume your test suite contains 100 docker based tests[^1] and a docker registry might return a `500 Internal Server Error` randomly every 10.000 requests (not sure how realistic this numbers is): This results in 1% of your CI runs to fail.
- Slowdown: Docker based tests tend to be on the slower end of the spectrum, but doing a pull for every tests will further slow down the tests (when running `go test ./physical/postgresql/` I noticed a 10s speedup[^2]).

Related https://github.com/openbao/openbao/issues/3399

## Acknowledgements

       also allowed.

[^1]: based on the debug prints I added to the draft version of this PR, this is a realistic number for this openbao/openbao: If I analyzed the logs correctly 105 pulls were prevented.
[^2]: The HashiCorp mirror seems to be particularly slow: `podman pull docker.mirror.hashicorp.services/postgres:11.1` with a hot cache takes a full 5 seconds while `docker.io/library/postgres:11.1` takes only 2.

### DIFF
--- a/sdk/helper/docker/testhelpers.go
+++ b/sdk/helper/docker/testhelpers.go
@@ -30,6 +30,7 @@ import (
 	"github.com/moby/moby/api/types/mount"
 	"github.com/moby/moby/api/types/network"
 	"github.com/moby/moby/client"
+	"github.com/openbao/openbao/sdk/v2/helper/locksutil"
 )
 
 type Runner struct {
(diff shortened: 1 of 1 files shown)


sdk/helper/docker/testhelpers.go
```diff
@@ -357,6 +359,99 @@ type StartResult struct {
...
+		}
+	}
+}
+
+var (
+	pullLocks = locksutil.CreateLocks()
+	pullCache = map[string]error{}
+)
+
+func (d *Runner) pull(ctx context.Context, image string) (result error) {
```
wslabosz-reply: any particular reason for the naked returns with named return value?

wslabosz-reply: e.g. :422 we return an error directly, but on :442 we assign it to the declared named return value and then utilize the naked return, seems inconsistent

phil9909: To cache the result in the defer.

```go
	defer func() {
		pullCache[image] = result
	}()
```

phil9909: > e.g. :422 we return an error directly, but on :442 we assign it to the declared named return value and then utilize the naked return, seems inconsistent

You are right, for consistency this should use a named return as well.

Maybe I should refactor this function instead:
- Pass the `RegistryAuth` as a parameter
- Remove the error return completely

wslabosz-reply: yup, would be good for consistency sake

wslabosz-reply (dismissed):

phil9909:
@wslabosz-reply had to fix a data-race in the map access, please take another look.

wslabosz-reply (approved):
looks clean

PR #3437, opened 2026-07-06
fix(command/server): correct merge condition for plugin_download_max_size in Config.Merge()

sivapalan:
## Description

`Config.Merge()` used `c2.PluginAutoDownloadRaw != nil` as the condition for merging `PluginDownloadMaxSize`. As a result, `plugin_download_max_size` was only preserved when `plugin_auto_download` was also set in the subsequent config file. The condition should check `c2.PluginDownloadMaxSizeRaw != nil`.

## Rationale

When OpenBao is started with multiple `-config` paths, `ParseServerConfig` calls `Config.Merge()` to combine them. Due to the wrong condition,`plugin_download_max_size` set in an earlier config file would be silently dropped unless `plugin_auto_download` happened to be set in a later one.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3437.txt
+++ b/changelog/3437.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+command/server: Correct merge condition for `plugin_download_max_size` in `Config.Merge()`
+```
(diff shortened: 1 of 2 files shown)


wslabosz-reply (approved):
thanks @sivapalan

PR #3448, opened 2026-07-08
Introduce GRPC-based invalidation mechanism

cipherboy:
## Description

This introduces a three-part GRPC-invalidation mechanism:

1. A new physical shim to catch storage write operations and dispatch it into a GRPC invalidation queue.
2. A new index manager for caching calls to update storage replication indices and reduce load on the backend database.
3. A new GRPC layer to ship invalidations from the active to all registered standby nodes and subsequently invalidate on them when the respective index has been hit.

We introduce new hooks for this GRPC to delay standby startup until the standby has caught the active node's state.

## Rationale

Tests will be introduced in a subsequent change; let me know if anyone wants me to move it to a feature branch, though I think this is the bulk of the change.

Related: #3027

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/physical/postgresql/index.go
+++ b/internal/physical/postgresql/index.go
@@ -1,0 +1,72 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package postgresql
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/openbao/openbao/sdk/v2/physical"
+)
+
+var _ physical.ReplicationIndexBackend = &PostgreSQLBackend{}
+
+const lsnIndexQuery = `
+SELECT value.index::varchar
+FROM (
+	SELECT CASE pg_is_in_recovery()
+	WHEN true
+	THEN pg_last_wal_receive_lsn()
+	ELSE pg_current_wal_lsn()
+	END
+	AS index
+)
+AS value;
+`
+
+func (p *PostgreSQLBackend) AppliedReplicationIndex(ctx context.Context) (string, error) {
+	var index string
+	if err := p.client.QueryRowContext(ctx, lsnIndexQuery).Scan(&index); err != nil {
+		return "", fmt.Errorf("getting WAL LSN failed: %w", err)
+	}
+
+	return index, nil
+}
+
+func (p *PostgreSQLBackend) GreaterEqualReplicationIndex(ctx context.Context, left string, right string) (bool, error) {
+	leftInt, err := indexToUint64(left)
+	if err != nil {
+		return false, fmt.Errorf("error parsing left: %w", err)
+	}
+
+	rightInt, err := indexToUint64(right)
+	if err != nil {
+		return false, fmt.Errorf("error parsing right: %w", err)
+	}
+
+	return leftInt >= rightInt, nil
+}
+
+func indexToUint64(index string) (uint64, error) {
+	if strings.Count(index, "/") != 1 {
+		return 0, fmt.Errorf("expected PG WAL LSN to contain exactly one slash: was %q", index)
+	}
+
+	parts := strings.SplitN(index, "/", 2)
+
+	// See https://pgpedia.info/x/xlogrecptr.html for naming info.
+	xlogid, err := strconv.ParseUint(parts[0], 16, 32)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse PG WAL LSN's log file number: %w", err)
+	}
+
+	xrecoff, err := strconv.ParseUint(parts[1], 16, 32)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse PG WAL LSN's byte offset: %w", err)
+	}
+
+	return xlogid<<32 | xrecoff, nil
+}
--- a/internal/vault/forwarding/request_forwarding_rpc.go
+++ b/internal/vault/forwarding/request_forwarding_rpc.go
@@ -5,6 +5,7 @@ package forwarding
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -19,6 +20,7 @@ import (
 	"github.com/openbao/openbao/sdk/v2/helper/consts"
 	"github.com/openbao/openbao/v2/internal/helper/forwarding"
 	"github.com/openbao/openbao/v2/internal/physical/raft"
+	"google.golang.org/grpc"
 )
 
 type forwardedRequestRPCServer struct {
@@ -123,6 +125,40 @@ func (s *forwardedRequestRPCServer) AdvertiseNamespaceKeys(ctx context.Context, 
 	return ret, nil
 }
 
+func (s *forwardedRequestRPCServer) StartInvalidations(ctx context.Context, req *StartInvalidationRequest) (*StartInvalidationResponse, error) {
+	index, err := s.core.MarkPeerStarted(ctx, req.Uuid)
+
+	var errMsg string
+	if err != nil {
+		s.core.Logger().Error("invalidation: failed to mark peer as started", "err", err)
+		errMsg = "failed to start invalidations; check active logs for more info"
+	}
+
+	return &StartInvalidationResponse{
+		Err:   errMsg,
+		Index: index,
+	}, nil
+}
+
+func (s *forwardedRequestRPCServer) CheckInvalidations(req *CheckInvalidationRequest, stream grpc.ServerStreamingServer[CheckInvalidationResponse]) error {
+	uuid, stopCh, err := s.core.AddInvalidationPeer(stream)
+	if err != nil {
+		s.core.Logger().Error("invalidation: failed registering invalidation peer", "err", err)
+		return fmt.Errorf("not registered; check active server's logs for information")
+	}
+
+	s.core.Logger().Trace("invalidation: starting invalidation handling for peer", "uuid", uuid)
+
+	// Wait for the peer's connection to be closed. When we call
+	// AddInvalidationPeer(...) above, we effectively consume the stream
+	// and sending of events on it. But this function returning closes the
+	// stream so we need to block until we're done with the stream.
+	<-stopCh
+	s.core.Logger().Trace("invalidation: finished invalidation handling for peer", "uuid", uuid)
+
+	return nil
+}
+
 func (s *forwardedRequestRPCServer) SendNamespaceKeys(ctx context.Context, in *SendNamespaceKeysRequest) (*SendNamespaceKeysReply, error) {
 	keys := make(map[string][]byte, len(in.Keys))
 
@@ -166,6 +202,10 @@ type Client struct {
 	taskContext  context.Context
 	echoTicker   *time.Ticker
 	nsSyncTicker *time.Ticker
+
+	invalidationsContext       context.Context
+	invalidationsContextCancel context.CancelFunc
+	peerUUID                   atomic.Pointer[string]
 }
 
 func NewClient(core core, requestForwardingClient RequestForwardingClient, taskContext context.Context, echoTicker *time.Ticker, nsSyncTicker *time.Ticker) *Client {
@@ -208,6 +248,7 @@ func (c *Client) Start() {
 				req.RaftUpgradeVersion = raftBackend.EffectiveVersion()
 				labels = append(labels, metrics.Label{Name: "peer_id", Value: raftBackend.NodeID()})
 			}
+
 			defer metrics.MeasureSinceWithLabels([]string{"ha", "rpc", "client", "echo"}, now, labels)
 
 			ctx, cancel := context.WithTimeout(c.taskContext, 2*time.Second)
@@ -359,4 +400,127 @@ func (c *Client) getKeys(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+// CheckReplicationIndex returns the current index on the active node.
+func (c *Client) CheckReplicationIndex(ctx context.Context) (string, error) {
+	var uuid string
+
+	if value := c.peerUUID.Load(); value != nil && *value != "" {
+		uuid = *value
+	} else {
+		return "", errors.New("active node has not returned a uuid")
+	}
+
+	resp, err := c.StartInvalidations(ctx, &StartInvalidationRequest{
+		Uuid: uuid,
+	})
+	if err != nil {
+		return "", fmt.Errorf("error checking active node's replication index: %w", err)
+	}
+
+	if resp == nil {
+		return "", errors.New("no replication index returned by active node")
+	}
+
+	if resp.Err != "" {
+		return "", fmt.Errorf("error checking replication index: %v", resp.Err)
+	}
+
+	return resp.Index, nil
+}
+
+func (c *Client) StreamInvalidations(ctx context.Context) error {
+	c.peerUUID.Store(nil)
+	c.invalidationsContext, c.invalidationsContextCancel = context.WithCancel(c.taskContext)
+
+	commonCleanup := func() {
+		c.peerUUID.Store(nil)
+		c.invalidationsContextCancel()
+	}
+
+	doCleanup := true
+	defer func() {
+		if doCleanup {
+			commonCleanup()
+		}
+	}()
+
+	// Start streaming invalidations from the active.
+	stream, err := c.CheckInvalidations(c.invalidationsContext, &CheckInvalidationRequest{})
+	if err != nil {
+		return fmt.Errorf("failed starting invalidation stream: %w", err)
+	}
+
+	// Receive our first invalidation.
+	invalidation, err := stream.Recv()
+	if err != nil {
+		return fmt.Errorf("error during initial invalidation receive: %w", err)
+	}
+
+	// Every response should have UUID.
+	uuid := invalidation.Uuid
+	c.peerUUID.Store(&uuid)
+
+	// No longer perform cleanup when we exit this function: we're kicking
+	// off a background goroutine to handle invalidations including this one.
+	doCleanup = false
+
+	go func() {
+		cleanup := func() {
+			commonCleanup()
+
+			// In our goroutine, the read-enabled standby has already started.
+			// If our stream goes away, we need to ensure that we cancel the
+			// current read-enabled standby state as we won't be getting any
+			// more invalidations and thus be stale. If we were to just
+			// re-establish the stream, we'd have lost any invalidations the
+			// server has generated in the meantime. We'd also have no way of
+			// catching back up to the current state as we don't track read
+			// data versus indices.
+			//
+			// Thus a restart is the cleanest solution here, so that we know
+			// we catch up to initial state again on stream re-establishment.
+			c.core.Restart()
+		}
+		defer cleanup()
+
+		// While the server immediately send back an empty invalidation with
+		// the uuid in it, we don't hold any exclusive locks so (pending
+		// timing) a regular invalidation could be handed to us ahead of it.
+		// That's why this loop is inverted: we want to process the above
+		// invalidation first before we get a new one.
+
+		var err error
+		for {
+			select {
+			case <-c.invalidationsContext.Done():
+				return
+			default:
+			}
+
+			if invalidation.Restart {
+				c.core.Logger().Info("forwarding: active node indicated restart on invalidation")
+				return
+			}
+
+			c.core.AwaitInvalidation(c.invalidationsContext, cleanup, invalidation.Index, invalidation.Keys...)
+
+			invalidation, err = stream.Recv()
+			if err != nil {
+				c.core.Logger().Warn("forwarding: error receiving invalidation from active node", "err", err)
+				return
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (c *Client) StopInvalidations() {
+	if c == nil || c.invalidationsContextCancel == nil {
+		return
+	}
+
+	c.invalidationsContextCancel()
 }
--- a/internal/vault/forwarding/request_forwarding_service.proto
+++ b/internal/vault/forwarding/request_forwarding_service.proto
@@ -52,6 +52,24 @@ message ClientKey {
 	bytes d = 4;
 }
 
+message StartInvalidationRequest {
+	string uuid = 1;
+}
+
+message StartInvalidationResponse {
+	string index = 1;
+	string err = 2;
+}
+
+message CheckInvalidationRequest {}
+
+message CheckInvalidationResponse {
+	string index = 1;
+	repeated string keys = 2;
+	bool restart = 3;
+	string uuid = 4;
+}
+
 message NamespaceKey {
 	string uuid = 1;
 	bytes key = 2;
@@ -88,23 +106,27 @@ message GetNamespaceKeysReply {
 //
 // See also: https://medium.com/@hoangxuantoank13/grpc-non-breaking-changes-breaking-changes-and-versioning-solution-1dcb989beb16
 service RequestForwarding {
-    // Service health.
+	// Service health.
 
 	rpc Echo(EchoRequest) returns (EchoReply) {}
 
-    // Request forwarding.
+	// Request forwarding.
 
 	rpc ForwardRequest(forwarding.Request) returns (forwarding.Response) {}
 
 	// Namespace key sharing.
 
 	// standby->active key sharing; used when the standby has knowledge of
-    // namespace keys that a new active node does not. This is a two-step
+	// namespace keys that a new active node does not. This is a two-step
 	// process to avoid sending keys which the active node already knows
-    // about over the wire.
+	// about over the wire.
 	rpc AdvertiseNamespaceKeys(AdvertiseNamespaceKeysRequest) returns (AdvertiseNamespaceKeysReply) {};
 	rpc SendNamespaceKeys(SendNamespaceKeysRequest) returns (SendNamespaceKeysReply) {};
 
 	// active->standby key sharing; used by the standby to refresh keys.
 	rpc GetNamespaceKeys(GetNamespaceKeysRequest) returns (GetNamespaceKeysReply) {};
+
+	// GRPC-backed invalidation
+	rpc StartInvalidations(StartInvalidationRequest) returns (StartInvalidationResponse)  {}
+	rpc CheckInvalidations(CheckInvalidationRequest) returns (stream CheckInvalidationResponse) {}
 }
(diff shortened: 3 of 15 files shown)


internal/command/server/config.go
```diff
@@ -154,6 +154,8 @@ type Config struct {
 	// Whether to allow unsafe (usually URL encoded) relative request paths
 	// (containing `..`).
 	UnsafeRelativePaths bool `hcl:"unsafe_relative_paths"`
+
+	EnableGRPCInvalidation bool `hcl:"enable_grpc_invalidation"`
```
cipherboy: I'm not sure what we think about this; config options are usually fairly stable and if there's no connectivity, nodes will remain read-disabled standby so perhaps it is not necessary. 🤔

internal/vault/forwarding/request_forwarding_rpc.go
```diff
@@ -360,3 +404,111 @@ func (c *Client) getKeys(ctx context.Context) error {
...
+
+				if invalidation.Restart {
+					c.core.Logger().Info("forwarding: server indicated restart")
+					c.core.Restart()
+					c.invalidationsContextCancel()
+					break
+				}
+
+				go func() {
+					err = c.core.AwaitInvalidation(ctx, invalidation.Index, invalidation.Keys...)
```
cipherboy: This function blocks until we reach the specified index; we could invert this and drop it into the `dispacher`. I was halfway to the rewrite in that direction before realizing queue management between standby/active might be a bit of a pain. I might just leave jobs in the queue indefinitely and not stop it between active/standby node state changes, so it only stops on shutdown.

Thoughts?

internal/vault/forwarding/request_forwarding_rpc.go
```diff
@@ -123,6 +126,43 @@ func (s *forwardedRequestRPCServer) AdvertiseNamespaceKeys(ctx context.Context,
...
+
+	// Send an initial response with the uuid.
+	if err := stream.Send(&CheckInvalidationResponse{
+		Uuid: uuid,
+	}); err != nil {
+		s.core.Logger().Error("invalidation: failed responding to invalidation peer registration", "err", err)
+		return fmt.Errorf("failed to send on stream")
+	}
+
+	// Wait for the peer's connection to be closed.
```
cipherboy: Probably warrants an expanded comment that `AddInvalidationPeer` effectively consumes the stream and this handler needs to remain parked until we're done sending invalidation events back to the standby, at which point this channel is closed.

internal/vault/forwarding/request_forwarding_rpc.go
```diff
@@ -360,3 +404,111 @@ func (c *Client) getKeys(ctx context.Context) error {
...
+	}
+
+	if resp.Err != "" {
+		return "", fmt.Errorf("error checking replication index: %v", resp.Err)
+	}
+
+	return resp.Index, nil
+}
+
+func (c *Client) StreamInvalidations(ctx context.Context) {
```
cipherboy: I think we need to expand `Echo` handling on the active to keep alive invalidation peers or otherwise keep sending events in the main stream. No events may have been sent form active->standby for a while, so the cache will eventually become stale and cause the standby to restart. 🤔

internal/vault/storage_index.go
```diff
@@ -0,0 +1,117 @@
...
+type indexManager struct {
+	backend physical.ReplicationIndexBackend
+	backoff time.Duration
+
+	l          sync.RWMutex
+	lastIndex  string
+	lastUpdate time.Time
+}
+
+func NewIndexManager(backend physical.ReplicationIndexBackend, backoff time.Duration) *indexManager {
```
cipherboy, line 31: @phil9909 This will make the index RFC easier.

internal/vault/grpc_invalidation.go
```diff
@@ -0,0 +1,356 @@
...
+// it yields an initial index after which all events will be sent.
+func (core *Core) MarkPeerStated(ctx context.Context, uuid string) (string, error) {
+	core.logger.Trace("marking peer as started", "uuid", uuid)
+
+	index, err := core.indexManager.Latest(ctx)
+	if err != nil {
+		return "", fmt.Errorf("failed getting current physical replication index: %w", err)
+	}
+
+	err = core.connectedInvalidationPeers.markPeerStarted(uuid, index)
```
phil9909: I think we have a race here: If `invalidationPeers.SendInvalidation` is called between `core.indexManager.Latest` and `core.connectedInvalidationPeers.markPeerStarted` the peer would miss that invalidation, right?

internal/vault/storage_index.go
```diff
@@ -0,0 +1,117 @@
...
+
+		if !passed {
+			return none, errors.New("not yet reached specified storage index")
+		}
+
+		return none, nil
+	}
+
+	_, err := backoff.Retry(timeBoxed, op, backoff.WithBackOff(b))
+	return err
```
phil9909, line 112: I was wondering, if we should move this logic into the backend. Some backends might have more efficient ways to do this. E.g. in raft we could notify all waiting routines when we update the `latestIndex` https://github.com/openbao/openbao/blob/d1b9916b8c37b41f741d701a1a4d6e347af0f90c/physical/raft/fsm.go#L944

For PostgreSQL we might be able to use [`WAIT FOR LSN 'lsn'`](https://www.postgresql.org/docs/19/sql-wait-for.html) once version 19 is released (and fallback to polling for older versions).
Also, for PostgreSQL it might be worth it to have an atomic "highest index seen", if this index is greater than the awaited index, we can skip the query.

sdk/physical/invalidation.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &grpcInvalidatorTransaction{
+		grpcInvalidator: g.grpcInvalidator,
+		txn:             txn,
+		writes:          map[string]struct{}{},
+	}, nil
+}
```
phil9909: How about:

```suggestion
func (g *transactionalGRPCInvalidator) BeginReadOnlyTx(ctx context.Context) (Transaction, error) {
	return g.grpcInvalidator.backend.(TransactionalBackend).BeginReadOnlyTx(ctx)
}
```

internal/physical/postgresql/index.go
```diff
@@ -0,0 +1,60 @@
...
+var _ physical.ReplicationIndexBackend = &PostgreSQLBackend{}
+
+func (p *PostgreSQLBackend) AppliedReplicationIndex(ctx context.Context) (string, error) {
+	var index string
+	indexQuery := "SELECT pg_current_wal_lsn()::varchar;"
+	if err := p.client.QueryRowContext(ctx, indexQuery).Scan(&index); err != nil {
+		return "", fmt.Errorf("pg_current_wal_lsn() failed: %w", err)
+	}
+
+	return index, nil
```
phil9909, line 36: Maye we should add a prefix here (e.g. `v1:`)

internal/vault/grpc_invalidation.go
cipherboy: @phil9909 I suppose this is true. I think we can probably invert the order here (mark the peer as started and then yield the index), having the peer queue events it sees before getting the started index and then subsequently invalidating them once seen (if higher than starting index).

I think we have a similar race in standby nodes on startup anyways for Raft, w.r.t. invalidations ignored during startup.

internal/vault/storage_index.go
cipherboy: @phil9909 Raft will never be used with this. It is already a GRPC based protocol and has immediate notification of replication...?

>  Also, for PostgreSQL it might be worth it to have an atomic "highest index seen", if this index is greater than the awaited index, we can skip the query.

Is this not what we do with `IndexManager`? 

> For PostgreSQL we might be able to use [WAIT FOR LSN 'lsn'](https://www.postgresql.org/docs/19/sql-wait-for.html) once version 19 is released (and fallback to polling for older versions).

We could, though this needs coordination to not overwhelm our connection count.

phil9909: > Raft will never be used with this. It is already a GRPC based protocol and has immediate notification of replication...?

Not today, but to implement this:

https://github.com/openbao/openbao/blob/17618c906d428bd4eb881c00d9b3ba1c5fc7730f/website/content/community/rfcs/index-headers.mdx?plain=1#L50-L52

We probably want to reuse the `indexManager` functionality, right?
Edit: You answered that already: https://github.com/openbao/openbao/pull/3448#discussion_r3552205145

> Is this not what we do with `IndexManager`?

Not exactly. If `lastUpdate` is 1 second ago (backoff 5ms), `lastIndex` is `100` (assuming simple numeric indices for this example) and we call `Await(ctx, 90)` we could `return nil` without doing a refresh, because we know the index is **at least** 100 however, stale that value might be.

> We could, though this needs coordination to not overwhelm our connection count.

true

We can also leave the backoff logic here. In a potential follow up, we can add an interface `IndexAwaiter{ Await(ctx context.Context, index string) error }`. `indexManager.Await` would then check if the current backend implements `IndexAwaiter` and delegate the call, falling back to the default implementation otherwise, if we need compatibility.

cipherboy: @phil9909 I've added the always-check-last-seen-index logic, though I don't think it matters that much ultimately. On PostgreSQL we'll constantly be refreshing due to incoming invalidations and when we implement index headers, that should also keep things reasonably fresh if used.

>  Some backends might have more efficient ways to do this. E.g. in raft we could notify all waiting routines when we update the latestIndex

I'm not convinced this will be any faster. Keep in mind, Raft will be pulling from the local applied index; that index is an integer stored in memory. Other than grabbing a read-lock over the FSM, the index loads are atomic integers and fast to query as well.

You're looking at an order of milliseconds probably (unless a write is happening) to poll these values. I'm not convinced it'll be much more efficient to use a notification channel here. You can probably get fewer unnecessary polls and a little faster execution on average, but we'd have to see data.

If we want to pick it up later for index headers, I think that's fine but I'm happy calling it out of scope for this issue.

sdk/physical/invalidation.go
cipherboy: @phil9909 Not sure I understand, how does this work? We need to track writes and only send them for invalidation once the commit succeeds. If we pass through the underlying transaction, don't we lose the caching of writes and the ability to notify as a batch of all invalidated keys...?

`backend` here is PostgreSQL, so `backend.BeginReadOnlyTx(...)` doesn't do any write tracking in the returned transaction.

internal/physical/postgresql/index.go
cipherboy: @phil9909 I suppose we could, that seems unnecessary given it'd be a majorly breaking change on PostgreSQL's part.

cipherboy:
@phil9909 This should address most all of your comments I think. We call `c.invalidations.Track()` (and I've reworked it) before even starting streaming of events so we don't lose any.

sdk/physical/invalidation.go
phil9909: We don't have to track writes, because there won't be any. After all it is a **read-only** transaction. https://github.com/openbao/openbao/blob/73f12f3805a5b679475a77bb8f471d68af4b9c3a/internal/physical/postgresql/transaction.go#L58-L60

Or am I missing something here?

internal/physical/postgresql/index.go
phil9909: Agree, I don't think PostgerSQL will change the format.

I was thinking that we might want to change the format. Maybe we want use a different encoding (e.g. we want to do the conversion from LSN to int here instead) or we have to add some id, because of multi-db support.

But we can probably get away with adding a `v2` prefix, if we ever want to do this and use the missing prefix as an indicator for v1. Because we know that it can never start with a `v` (correct me if I'm wrong here, but it should always be a hex character) this should not cause any problems.

cipherboy: Yeah, I'm not sure converting early to int makes much difference here. I think for broad compatibility in case we were to ever introduce another backend, we'd want to use string (signed vs unsigned 64-bit int, true string indices, or wider widths all seem reasonable for a hypothetical backend). So I think the outer `physical` interface has to be string. So at the psql level we'll always have a conversation here (whether the hex + bit shift or base10 int).

But yes, if we do want to change this at the psql level later we can always introduce a v2 marker.

sdk/physical/invalidation.go
cipherboy: @phil9909 Ah, I gotcha. No, you're correct, I misunderstood what this was about. I thought this suggestion was for both patterns.

internal/vault/forwarding/request_forwarding_rpc.go
```diff
@@ -360,3 +410,132 @@ func (c *Client) getKeys(ctx context.Context) error {
...
+}
+
+func (c *Client) StreamInvalidations(ctx context.Context) error {
+	c.invalidationsContext, c.invalidationsContextCancel = context.WithCancel(c.taskContext)
+
+	var preStreamError error
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
```
cipherboy, line 469: This got much more complex and I don't really know how to help it.

We want to wait until the first message comes back from the server with the peer's UUID so that we can call `StartInvalidations(...)` with that. ~Ideally we would timebox the very first `Recv()` call, but that'd require our own context handling which gets more complex too...~

internal/vault/ha.go
```diff
@@ -911,10 +912,73 @@ func (c *Core) runReadEnabledStandby(ctx context.Context, ctxCancel context.Canc
...
+			// Start streaming invalidation events from the primary.
+			if err := c.rpcForwardingClient.StreamInvalidations(ctx); err != nil {
+				c.logger.Error("failed to begin streaming invalidations", "err", err)
+				return false
+			}
+
+			return true
+		}(); !ok {
+			// Do not consider this a stoppage.
+			return false
```
cipherboy: The part I don't understand right now is what mechanism do we want to use to re-trigger attempting standby setup? I believe we'll drop to a read-disabled standby in the event that we fail establishment; ideally something would re-trigger us that its time to re-attempt with some back-off. 🤔 

Its also moderately hard to understand if we've actually stood up `Core` as a read-enabled or read-disabled standby (due to failure here). We could poke some random subsystem but that seems incorrect.

Ideally we'd use something like Backoff in a long-lived goroutine here, to call `Restart()` maybe? 🤔

internal/vault/grpc_invalidation.go
```diff
@@ -0,0 +1,466 @@
...
+type invalidationPeerInfo struct {
+	stream grpc.ServerStreamingServer[forwarding.CheckInvalidationResponse]
+	stopCh chan struct{}
+
+	started bool
+}
+
+func (c *Core) NewInvalidationPeers() {
+	logger := c.logger.Named("grpc-invalidation")
+	c.AddLogger(logger)
```
cipherboy, line 52: This should just be a single pattern tbh, but out of scope for this thing now.

internal/vault/grpc_invalidation.go
```diff
@@ -0,0 +1,466 @@
...
+			failed = append(failed, peerUUID)
+			continue
+		}
+
+		// We additionally need to grab the state lock here. While we could
+		// do this linearly from this thread, using a job manager with
+		// per-peer queues ensures that a slow peer does not starve other
+		// peers from their invalidations.
+		i.dispatcher.AddJob(&dispatchInvalidationToPeer{
+			scheduled: time.Now(),
```
cipherboy, line 175: Goal would be to add metrics later, but we can do that in a follow-up PR. Want to know how long locking, invalidation, and peer sending takes.

internal/vault/grpc_invalidation.go
```diff
@@ -0,0 +1,466 @@
...
+
+	a.cleanup()
+}
+
+type pingInvalidationJob struct {
+	peers *invalidationPeers
+	ctx   context.Context
+}
+
+var _ fairshare.Job = &awaitInvalidationJob{}
```
cipherboy:
```suggestion
var _ fairshare.Job = &pingInvalidationJob{}
```

cipherboy:
See second commit in https://github.com/openbao/openbao/pull/3542 for tests.

internal/vault/forwarding/request_forwarding_rpc.go
cipherboy: `recvWithContext` is probably unnecessary but we at least have the framework to support a shorter timebox until we get a UUID if we want.

internal/vault/forwarding/request_forwarding_service.proto
```diff
@@ -107,4 +125,8 @@ service RequestForwarding {
 
 	// active->standby key sharing; used by the standby to refresh keys.
 	rpc GetNamespaceKeys(GetNamespaceKeysRequest) returns (GetNamespaceKeysReply) {};
+
+    // GRPC-backed invalidation
```
satoqz:
```suggestion
	// GRPC-backed invalidation
```

(4 spaces -> tab)

sdk/physical/invalidation.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"context"
+	"maps"
+	"slices"
+	"sync"
+
+	log "github.com/hashicorp/go-hclog"
+)
+
+type GRPCInvalidator interface{}
```
satoqz: I think this is unused?

sdk/physical/invalidation.go
```diff
@@ -0,0 +1,151 @@
...
+
+type grpcInvalidator struct {
```
satoqz: I kind of dislike that it doesn't actually have anything to do with gRPC directly 😆 It's more like a... `WriteSubscriber`?

internal/vault/ha.go
```diff
@@ -911,10 +911,73 @@ func (c *Core) runReadEnabledStandby(ctx context.Context, ctxCancel context.Canc
...
+				return false
+			}
+
+			c.requestForwardingConnectionLock.RLock()
+			defer c.requestForwardingConnectionLock.RUnlock()
+
+			if c.rpcForwardingClient == nil {
+				c.logger.Error("skipping invalidation streaming as no RPC client is present")
+				return false
+			}
```
satoqz: This happens when?

internal/vault/ha.go
satoqz: A good first step would perhaps be to return `(stop, retry bool)` where `stop=true` implies a `return` in the calling loop (`waitForLeadership`) and `retry=true` implies a `continue`. Then you don't step down to read-disabled standby.

A backoff could then be applied to the entirety of `waitForLeadership` more generally?

internal/vault/forwarding/request_forwarding_rpc.go
```diff
@@ -123,6 +125,48 @@ func (s *forwardedRequestRPCServer) AdvertiseNamespaceKeys(ctx context.Context,
 	return ret, nil
 }
 
+func (s *forwardedRequestRPCServer) StartInvalidations(ctx context.Context, req *StartInvalidationRequest) (*StartInvalidationResponse, error) {
+	index, err := s.core.MarkPeerStated(ctx, req.Uuid)
```
satoqz: Stated?

internal/vault/forwarding/request_forwarding_rpc.go
```diff
@@ -360,3 +409,126 @@ func (c *Client) getKeys(ctx context.Context) error {
...
 }
+
+// CheckReplicationIndex returns the current index on the active node.
+func (c *Client) CheckReplicationIndex(ctx context.Context) (string, error) {
+	var uuid string
+
+	if value := c.peerUUID.Load(); value != nil && *value != "" {
+		uuid = *value
+	} else {
+		return "", fmt.Errorf("active node has not returned a uuid")
```
satoqz:
```suggestion
		return "", errors.New("active node has not returned a uuid")
```

internal/vault/forwarding/request_forwarding_service.proto
```diff
@@ -52,6 +52,24 @@ message ClientKey {
...
+	string index = 1;
+	string err = 2;
+}
+
+message CheckInvalidationRequest {}
+
+message CheckInvalidationResponse {
+	string index = 1;
+	repeated string keys = 2;
+	bool restart = 3;
```
satoqz, line 69: I think this one is only ever false?

internal/vault/grpc_invalidation.go
```diff
@@ -0,0 +1,466 @@
...
+func (i *invalidationPeers) Setup() {
+	i.l.Lock()
+	defer i.l.Unlock()
+
+	i.peers = zcache.New[string, *invalidationPeerInfo](16*i.core.clusterHeartbeatInterval, 1*time.Second)
+	i.peers.OnEvicted(func(uuid string, peer *invalidationPeerInfo) {
+		i.logger.Debug("evicting stale invalidation client for peer", "uuid", uuid)
+		if peer.stopCh != nil {
+			close(peer.stopCh)
+			peer.stopCh = nil
```
satoqz:
```suggestion
```

seems unnecessary.

internal/vault/grpc_invalidation.go
```diff
@@ -0,0 +1,466 @@
...
+// AddPeer handles assignment of the peer to the currently tracked subset,
+// using the specified stream. It yields (an identifier and a cancel channel)
+// or an error.
+func (i *invalidationPeers) AddPeer(stream grpc.ServerStreamingServer[forwarding.CheckInvalidationResponse]) (string, chan struct{}, error) {
+	i.l.RLock()
+	defer i.l.RUnlock()
+
+	info := &invalidationPeerInfo{
+		stream: stream,
+		stopCh: make(chan struct{}, 1),
```
satoqz: Does this need to be buffered? Isn't it only ever closed?

internal/vault/grpc_invalidation.go
```diff
@@ -0,0 +1,466 @@
...
+
+	// Peer identifiers are ephemeral and matched to the stream.
+	for {
+		peerUUID, err := uuid.GenerateUUID()
+		if err != nil {
+			return "", nil, err
+		}
+
+		if err := i.peers.Add(peerUUID, info); err != nil {
+			// This should very rarely happen.
```
satoqz:
```suggestion
			// This should practically never happen.
```

rarely is an understatement 😆

internal/vault/forwarding/request_forwarding_rpc.go
satoqz: Still a bit confused about why we expect to restart core here. Other than that, can we move stream creation + initial receive out of the goroutine and run it synchronously? Does that simplify it?

sdk/physical/invalidation.go
cipherboy: @satoqz This is fair; I wanted to tie the feature together across this middleware layer to the actual GRPC portions later.

internal/vault/ha.go
cipherboy: @satoqz This occurs when the standby doesn't have a (valid) advertisement from the active or the active doesn't have a cluster address at all.

sdk/physical/invalidation.go
cipherboy: This is true, we could probably just do `physical.Backend`.

internal/vault/forwarding/request_forwarding_service.proto
cipherboy: I kinda wanted this for forwards compatibility if we ever need to introduce server-indicated restart; I'm still not sure yet in 2.7.0 if we'll need this though. So far I've gotten away without it, but I wasn't sure if we wanted the server to monitor invalidation time? :thinking:

internal/vault/forwarding/request_forwarding_rpc.go
cipherboy: @satoqz said:

> Still a bit confused about why we expect to restart core here. 

When the stream closes, we need to restart to a read-disabled standby or recreate a new stream and go back to read-enabled. Otherwise, we'll potentially have missed invalidation events on the server (between old stream closing and new stream creating).

Strictly (in the future), we could do a bounded wait (either time or up to X events) and close the peer server-side after that (rather than closing the peer when the stream exits due to inactivity). To do so, we'd need stable UUID per-peer. If peer comes back with same UUID and server hasn't started dropping events, we could avoid restarting standby on stream drop/reconnect.

If we do, we'd potentially want to use that restart hook mentioned above to force restart if the peer comes back too late.

cipherboy: > Other than that, can we move stream creation + initial receive out of the goroutine and run it synchronously? Does that simplify it?

That might, though we still need a loop around the receive I believe; we might receive invalidation we need to queue before getting UUID assignment as we can't necessarily guarantee ordering?

internal/vault/ha.go
cipherboy: I'll give that a shot.

internal/vault/forwarding/request_forwarding_rpc.go
cipherboy: I'll have cleaned this up in the new patch with documentation.

internal/vault/grpc_invalidation.go
```diff
@@ -0,0 +1,472 @@
...
+	}
+}
+
+func (c *Core) SetupInvalidationPeers() {
+	c.connectedInvalidationPeers.Setup()
+}
+
+func (c *Core) LocalGRPCDispatching() {
+	c.connectedInvalidationPeers.SetupStandby()
+}
```
satoqz, line 66: Nit: I'd just inline these TBH.

satoqz (approved):
Looks good to me, I don't immediately see any problems. Kind of disappointing.

internal/vault/forwarding/request_forwarding_service.proto
satoqz: Since it is protobuf, I suppose you could easily add it later, once needed? It wouldn't work on the one-time initial upgrade because standbys wouldn't handle the field yet, but eh.

internal/vault/grpc_invalidation.go
cipherboy: @satoqz Yeah, originally when I started these, I wasn't sure what locking would look like. A former draft had more state lock acquisitions because `c.connectedInvalidationPeers` could be `nil` at various points in the cycle. Now it is always provisioned and we just need to set up various members (and grab an internal lock). 

I'll revise in a follow-up as long as this one doesn't conflict out.

phil9909 (approved):

cipherboy:
Thanks all, merging!

internal/vault/forwarding/request_forwarding_service.proto
cipherboy: @satoqz This is true, but as you point out, wiring it up now ensures that if we need it later, we can push restart events to standbys as soon as the active updates. Technically this is against our upgrade guidance, but from what GitLab has said, in certain environments it is hard to formally comply.

PR #3464, opened 2026-07-09
Add RFC on parallel unseal

### LINKED ISSUE #1021: RFC - Parallel Unseal of OpenBao

cipherboy:
## Description

Assuming emergency seal gets done in 2.7.0 or early 2.8.0 timeframe, I'd be interested in adding generalized parallel unseal in the 2.8+ timeframe. 

## Rationale

Resolves: #1021

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/community/rfcs/parallel-unseal.mdx
+++ b/website/content/community/rfcs/parallel-unseal.mdx
@@ -1,0 +1,388 @@
+---
+sidebar_label: Parallel Unseal
+description: |-
+  Parallel Unseal of OpenBao via one or more unseal devices.
+---
+
+# Parallel Unseal of OpenBao
+
+:::info
+
+This RFC was proposed before the more specialized [emergency
+seal](emergency-seal.mdx), which proposes adding one Shamir's seal
+as a recovery option when using auto-unseal.
+
+:::
+
+## Summary
+
+OpenBao needs the ability to unseal with one of many pieces of key material to truly solve multi-cloud or multi-data center deployment issues and to provide [outage recovery procedures](https://en.wikipedia.org/wiki/Business_continuity_planning). Setting up a recoverable, distributed system shouldn't rely on a single node having access to all key management systems and operators should have the option to manually recover using Shamir's keys if cloud services are not available.
+
+To differentiate from upstream's solution which has different properties, we prefer to call our solution "Parallel Unseal" rather than "Multi-Unseal".
+
+## Problem Statement
+
+OpenBao's upstream has a history of [introducing](https://github.com/hashicorp/vault/pull/18683), [removing](https://github.com/hashicorp/vault/pull/18942), and subsequently [reintroducing](https://developer.hashicorp.com/vault/docs/configuration/seal/seal-ha) their multi-unseal feature in Vault Enterprise. The following [statement from Melissa Gurney Greene](https://github.com/hashicorp/vault/pull/18683#issuecomment-1425065124) on behalf of HashiCorp is quoted verbatim regrading this feature's history:
+
+> During the development of this feature and corresponding enterprise enhancements, we determined that adoption of this version of the solution would create a potential conflict when migrating to enterprise or vice versa. For example, the original implementation was incompatible with seal wrap, so we withdrew this for a more holistic solution that does not have this inconsistency, which could lead to data loss in some scenarios if the seal wrapped values could no longer be decrypted. Currently we do not have any plans to bring this feature into OSS. We welcome any feedback or suggestions related to OSS seal recovery options due to non-availability of KMS.
+
+From this we learn that the core problem was seal wrapping. While [Seal Wrap](https://developer.hashicorp.com/vault/docs/enterprise/sealwrap) is indeed an issue for Vault Enterprise, it is not present in OpenBao and thus gives us flexibility to choose an alternative solution. Furthermore, Vault Enterprise also has advanced clustering semantics we're unlikely to adopt, [Performance Secondary and Disaster Recovery](https://developer.hashicorp.com/vault/tutorials/enterprise/performance-replication) nodes. This poses a problem for Vault's Multi-Unseal as it introduces node complex topologies; however, they require the remote HSM(s) to be co-online at a single point in time to encrypt data for all nodes.
+
+In designing an alternative for OpenBao, we'd like to consider the following requirements:
+
+ - How can we enable unseal with disconnected HSM topologies? This is equivalent to allowing continued root key rotation operations even if a given seal mechanism is temporarily offline.
+    - _Example_: Consider a disaster recovery setup across multiple data centers or cloud environments. A HSM might be available on-premise, but in Google Cloud or AWS, using the provider's KMS might be preferable. These shouldn't require all environments have the same key material nor should they require all KMSes be co-accessible from a single node.
+    - For good user experience, this should allow nodes to have _local_ seal configuration: some amount of storage in barrier will need to be used to replicate this config, but we should be able to set node-local connection information for the same or different seal names.
+ - Can we enable a [redundant Shamir unseal](https://github.com/openbao/openbao/issues/299) for a break-glass recovery mechanism?
+    - _Example_: In the event of an HSM or KMS outage, OpenBao would previously be unable to start. However, by provisioning an additional Shamir's seal, we can allow an operator to start OpenBao using a manual process.
+    - As a corollary, this supports different organizational structures. If multiple operational teams use a shared OpenBao instance, they can be issued their own distinct Shamir's recovery shares, allowing any team to initiate fallback startup procedures .
+ - Can we allow any one unseal mechanism to start OpenBao?
+    - _Example_: A node might have two HSMs available, say, a high-throughput network connected HSM and a fallback USB-attached HSM with limited performance. Allowing any one HSM means that OpenBao has higher resiliency.
+ - Can we retain the ability to incrementally rekey OpenBao, even with offline seals?
+    - _Example_: If a particular seal mechanism is temporarily offline, we may still want to rotate keys. Past copies of the root key allow rotation of barrier keys. For rotating the root keys, however, access to the seal to perform new encryption operations is required. This can be done incrementally (encrypting the new root behind the current barrier keyring) or by using asymmetric seal mechanisms (like RSA-OAEP or a PQC-secure algorithm in the future).
+ - Can we continue to ensure seals are modifiable only by local operators but status is visible over the API?
+    - _Example_: Adding a seal is a privileged operation and so should require local filesystem access as many of them read configuration or identities from the disk. Removal of a seal should be possible via API, as a seal may already have been removed from the corresponding configuration file or may no longer be accessible. Status (which seals are in use on connected nodes, &c) should be visible from the API.
+ - Can we allow organizations to provide their own key material (BYOK) for auto unseal while ensuring backups can still be restored?
+    - _Example_: Fully restore OpenBao on AWS by importing the key material again into KMS and restore a backup snapshot. Or, pre-emptively creating an AWS seal with a public key set in a running node's config, even though it lacks access to the AWS Cloud KMS instance, would allow recovery from a backup, in AWS, in the event of an outage.
+
+
+We have a few threat modeling parameters:
+
+ - Compromising any one unseal mechanism will still (by definition of parallel unseal, that any one device still grants full access to decrypted storage) mean compromising the entire system.
+ - No secret or symmetric key should ever leave the KMS.
+ - Operators are ultimately trusted and thus can be made responsible for securely transmitting key material if necessary.
+
+In particular, these requirements mean that we'll want to encourage users to prefer asymmetric keys instead of symmetric keys where possible. The downside is that this means that [later PQC efforts](https://github.com/openbao/openbao/issues/496) will require operators to rotate their unseal keys as asymmetric PQC algorithms become available.
+
+## User-facing Description
+
+To end-users of OpenBao, no changes are expected. This will not impact any user APIs or availability.
+
+However, for operators, we'll see a few changes:
+
+ - Multiple seal stanzas in the configuration will now be accepted.
+ - `sys/seal-status` and various other APIs will be updated to allow specifying an explicit seal by name. The default seal used or returned will be the one which succeeded in unsealing this instance or the first one in the config if the instance is currently sealed. When unsealing a manually unsealed instance,
+
+Importantly, over Vault Enterprise we:
+
+ - Do not block seal config modification on requiring all seals be up-to-date,
+ - Do not block root or barrier rotation if all seals are not up-to-date,
+ - Do not require at least one active node have access to all auto-unseal methods, and
+
+thus greatly improve the operator user experience of this functionality.
+
+### Examples
+
+Presently, one can only unseal OpenBao with a single mechanism: either Shamir's, Transit, a HSM, or a cloud KMS solution. A configuration thus looks like:
+
+```hcl
+...
+
+seal "pkcs11" {
+  lib = "/usr/lib64/softhsm/libsofthsm.so"
+  token_label = "OpenBao"
+  pin = "4321"
+  key_label = "bao-root-key-rsa"
+  rsa_oaep_hash = "sha1"
+}
+
+...
+```
+
+After running `bao operator init -recovery-shares=3 -recovery-threshold=2`, you'd get an initial root token and three _[recovery key shards](/docs/concepts/seal/#auto-unseal)_. These **only** let you create new root tokens and **do not** let you manually unseal OpenBao in the event of a seal outage during system restart. The OpenBao instance can be restarted and will be automatically unsealed as long as that PKCS#11 device is online.
+
+After this change, operators can define multiple unseal mechanisms for a single node or for different nodes:
+
+```hcl
+...
+
+seal "pkcs11" {
+  name = "default"
+  priority = 1
+
+  lib = "/usr/lib64/softhsm/libsofthsm.so"
+  token_label = "OpenBao"
+  pin = "4321"
+  key_label = "bao-root-key-rsa"
+  rsa_oaep_hash = "sha1"
+}
+
+seal "transit" {
+  name = "backup"
+  priority = 2
+
+  address            = "https://openbao-fallback.example.com:8200"
+  token              = "s.Qf1s5zigZ4OX6akYjQXJC1jY"
+  disable_renewal    = "false"
+
+  // Key configuration
+  key_name           = "transit_key_name"
+  mount_path         = "transit/"
+  namespace          = "ns1/"
+
+  // TLS Configuration
+  tls_ca_cert        = "/etc/openbao/ca_cert.pem"
+  tls_client_cert    = "/etc/openbao/client_cert.pem"
+  tls_client_key     = "/etc/openbao/ca_cert.pem"
+  tls_server_name    = "openbao"
+  tls_skip_verify    = "false"
+}
+
+seal "shamir" {
+  name = "emergency"
+  priority = 100
+}
+...
+```
+
+Here, an operator would first initialize one seal (using the same command as above). Once initialized and started, the operator could initialize the remaining seals by using a command similar to above but providing the `-seal=transit` flag. In the event of an outage of the PKCS#11 seal, the Transit seal would be tried; if this did not work, Shamir's would be required to manually unseal OpenBao.
+
+When OpenBao is manually sealed, any unseal mechanism's root token recovery keys (or, in the case of a direct `shamir` type seal, the primary unseal keys) may be used to unseal.
+
+## Technical Description
+
+### Existing Design
+
+OpenBao has a three-key hierarchy:
+
+```mermaid
+flowchart LR
+    D1[Share 1]
+    D2[Share 2]
+    Dn[Share n]
+    S[Shamir's Key Encryption Key]
+    R[Root Key]
+    B[Barrier Keyring]
+    ESKEK[Encrypted Shamir's Key Encryption Key]
+    ERK[Encrypted Root Key]
+
+    D1 --Recovers--> S
+    D2 --Recovers--> S
+    Dn --Recovers--> S
+    S --Encrypts--> R
+    R --Encrypts--> B
+    B --Encrypts--> ESKEK
+    B --Encrypts--> ERK
+```
+
+This is the same for auto-unseal keys, swapping Shamir's recovery of the shard's combined key encryption key (KEK) for a key stored in an HSM or KMS solution.
+
+OpenBao uses the following storage entries for unsealing:
+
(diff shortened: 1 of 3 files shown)


website/content/community/rfcs/parallel-unseal.mdx
```diff
@@ -0,0 +1,388 @@
...
+
+1. `barrierSealConfigPath` will be migrated to the per-seal location `core/seals/<name>/shamir-config` or `core/seals/<name>/recovery-config` depending on if the mechanism is an auto-unseal mechanism. This is a _raw_ (unencrypted, from Barrier's PoV) storage entry.
+2. `StoredBarrierKeysPath` will be migrated to the per-seal location `core/seals/<name>/encrypted-root`. This is a _raw_ (unencrypted, from Barrier's PoV) storage entry; however, this raw entry is encrypted with the seal mechanism's key information.
+3. `shamirKekPath` will be migrated to the per-seal location `core/seals/<name>/kek`, when applicable. This is a _barrier-encrypted_ storage entry.
+4. For auto-unseal mechanisms, a copy of the seal configuration will be placed inside barrier at `core/seals/<name>/autounseal-config`. This allows for retreival of the public keys of other seal mechanisms and determining if configuration has changed since the last startup. This is a _barrier-encrypted_ storage entry.
```
satoqz:
```suggestion
4. For auto-unseal mechanisms, a copy of the seal configuration will be placed inside barrier at `core/seals/<name>/autounseal-config`. This allows for retrieval of the public keys of other seal mechanisms and determining if configuration has changed since the last startup. This is a _barrier-encrypted_ storage entry.
```

website/content/community/rfcs/parallel-unseal.mdx
```diff
@@ -0,0 +1,388 @@
...
+Similarly to HashiCorp's Vault Enterprise, we'll introduce new parameters:
+
+ - `name`, optional, defaulting to `default` if not specified. This needs to be the same across all instances of that seal's configuration in the cluster.
+ - `priority`, an integer, defaulting to `1` if not specified. This is a local-only parameter and does not affect other nodes.
+
+Both `name` and `priority` must be individually unique when multiple config entries are added that are not disabled. That is, only one seal may ever exist with that `name` and only one `priority` may exist.
+
+### New Shamir's Configuration
+
+A new seal type, `shamir`, will be introduced. This allows for multiple parallel Shamir unseals for when any of multiple unseal information should be kept in a secure, secondary location(s). These have the same new parameters above (`name` and `priority`); threshhold and other information will be stored in storage based on the initialization.
```
satoqz:
```suggestion
A new seal type, `shamir`, will be introduced. This allows for multiple parallel Shamir unseals for when any of multiple unseal information should be kept in a secure, secondary location(s). These have the same new parameters above (`name` and `priority`); threshold and other information will be stored in storage based on the initialization.
```

satoqz (approved):
I assume this is also good to merge as-is given it is "historical" and doesn't have any outstanding reviews or complaints.

cipherboy:
Thanks @satoqz! Yes, typos not withstanding. :laughing:

PR #3426, opened 2026-07-03
Adjust `sys/raw` endpoint interaction with implicitly sealed namespaces

wslabosz-reply:
## Description
This PR adjusts `handleSysRaw<Operation>` methods to prevent writing to unavailable namespaces, and adds clear error messages on read requests.

## Rationale
This PR disallows writes to implicitly sealed unsealable namespaces through `sys/raw` endpoints, while also adding clear error message on failed reads (namespace is sealed).

Resolves: #3408 

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3467, opened 2026-07-13
Adjust `sys/raw` endpoint interaction with implicitly sealed namespaces (#3426)

phil9909:
Backport of https://github.com/openbao/openbao/pull/3426 (clean cherry-pick)

### DIFF
--- a/vault/logical_raw.go
+++ b/vault/logical_raw.go
@@ -45,16 +45,19 @@ func NewRawBackend(core *Core) *RawBackend {
 	return r
 }
 
-func (b *RawBackend) storageByPath(ctx context.Context, path string) (StorageAccess, error) {
+// storageByPath returns appriopriate StorageAccess wrapping over specific
+// namespace barrier depending on the requested path. Also returns if the
+// namespace with given path (uuid) doesn't exist.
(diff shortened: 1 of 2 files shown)


wslabosz-reply (approved):

vault/logical_raw.go
```diff
@@ -99,16 +102,21 @@ func (b *RawBackend) handleRawRead(ctx context.Context, req *logical.Request, da
...
 
-	valueBytes, err := barrier.Get(ctx, path)
-	if err != nil {
+	valueBytes, err := storage.Get(ctx, path)
+	switch {
+	// We match against an error coming from using wrong barrier to read;
+	// This happens when we are in a storage space of a namespace that
+	// has been discarded from memory due to sealing of its parent.
+	case err != nil && strings.HasSuffix(err.Error(), "cipher: message authentication failed"):
+		return nil, barrier.ErrNamespaceSealed
```
satoqz, line 116: @wslabosz-reply this also happens when data was written to the (untrusted) storage backend by another actor.

I really think it should be an additional warning/hint, not override the error message entirely.

Issue #3478, opened 2026-07-13
Restoring backup from Vault with built-in AWS auth results in stuck config


cpboyd:
**Describe the bug**
Per https://github.com/openbao/openbao/issues/3352, I'm attempting to load the auth-aws plugin.

This loads, but when loading the old backup data structure, it refuses to acknowledge the new plugin:
```
{"@level":"info","@message":"updating existing plugin","@module":"core","@timestamp":"2026-07-13T21:30:28.593473Z","name":"aws","sha256":"7a77057e62973c1aae6035f52110e3302605a47b622756d954915b7b55eca10c","type":"auth","version":"v0.1.1"}
{"@level":"warn","@message":"plugin out of date in storage versus standby node configuration; this may be a false-positive depending on data replication state","@module":"core","@timestamp":"2026-07-13T21:30:28.593485Z","name":"aws","sha256":"7a77057e62973c1aae6035f52110e3302605a47b622756d954915b7b55eca10c","type":"auth","version":"v0.1.1"}
{"@level":"info","@message":"declarative plugin registration completed","@module":"core.plugins","@timestamp":"2026-07-13T21:30:28.594050Z"}
{"@level":"info","@message":"reading transactional mount table","@module":"core","@timestamp":"2026-07-13T21:30:28.594197Z"}
...
{"@level":"error","@message":"failed to create credential entry","@module":"core","@timestamp":"2026-07-13T21:30:28.601836Z","error":"plugin not found in the catalog: aws","path":"aws/"}
{"@level":"warn","@message":"skipping plugin-based auth entry","@module":"core","@timestamp":"2026-07-13T21:30:28.601858Z","path":"aws/"}
{"@level":"info","@message":"successfully mounted","@module":"core","@timestamp":"2026-07-13T21:30:28.601870Z","namespace":{"id":"root","uuid":"00000000-0000-0000-0000-000000000000","path":"","tainted":false,"manually_sealed":false,"unlock_key":"","custom_metadata":{}},"path":"aws/","type":"aws","version":"v1.18.5+builtin.vault"}
...
{"@level":"error","@message":"skipping initialization for nil auth backend","@module":"core","@timestamp":"2026-07-13T21:30:29.214831Z","path":"aws/","type":"aws","version":"v1.18.5+builtin.vault"}
```

As a result, the `aws/` auth path exists and is shown in lists but cannot be modified and throws a 404 if you try to query it.

[...]

PR #3486, opened 2026-07-14
Allow deleting mounts that are in limbo

satoqz:
## Description

When a mount is set up on unseal, its backing view is marked read-only and only later marked available for writing again, depending on the type of mount:

1. For singleton mounts, this happens at the end of `setupMount`/`setupCredential` via `defer`:
  ```
  commit 4c772b49968938418a042b1ad341eefe7c6c0a55
  Author: Jeff Mitchell <jeff@hashicorp.com>
  Date:   Thu Apr 19 13:29:43 2018 -0400
      Defer setting views read/write until the end of postUnseal (#4392)
      A few notes:
      * We exert positive control over singletons and they usually need to
      perform some (known, validated) writes, so this excludes singletons --
      they are simply limited to the end of the mount function as before.
      * I'm not sure how to test this _specifically_; I've done some testing
      of e.g. sealing vault and unsealing and ensuring that I can write to a
      KV mount. I think this is tested by every dev server though, since for a
      dev server Vault is inited, the default mounts are mounted, then it's
      sealed, then it's unsealed for the user, so it already goes through this
      code path. The mere fact that you can write to secret/ on a dev server
      means it was successfully set read-write.
  ```

2. For all other mounts, this happens as part of its `postUnsealFunc`.

When a mount is mounted but the backend isn't available (i.e., the backend is `nil`, this can happen if the associated plugin is no longer available), the post-unseal function will not mark the view as writable again and bails too early.

An unwanted side effect of this is that, since the view remains read-only indefinitely, the mount cannot be removed either, even though this is naturally a common use case for any leftover mounts post plugin removal.

## Rationale

Part of: https://github.com/openbao/openbao/issues/3478

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3486.txt
+++ b/changelog/3486.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Allow unmounting auth/secrets mounts even if the backing plugin is unavailable.
+```
--- a/internal/vault/auth.go
+++ b/internal/vault/auth.go
@@ -1115,13 +1115,14 @@ func (c *Core) setupCredential(ctx context.Context, entry *routing.MountEntry) (
 		return nil, err
 	}
 
+	isSingleton := slices.Contains(singletonMounts, entry.Type)
 	origViewReadOnlyErr := view.GetReadOnlyErr()
 
 	// Mark the view as read-only until the mounting is complete and
 	// ensure that it is reset after. This ensures that there will be no
 	// writes during the construction of the backend.
 	view.SetReadOnlyErr(logical.ErrSetupReadOnly)
-	if slices.Contains(singletonMounts, entry.Type) {
+	if isSingleton {
 		defer view.SetReadOnlyErr(origViewReadOnlyErr)
 	}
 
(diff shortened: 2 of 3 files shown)


wslabosz-reply (dismissed):
makes sense, with good change going forward with all vault -> openbao migrations

cipherboy:
@satoqz I'd suggest we drop this until 2.7.0 and backport to 2.6.x if we're happy before then; there's something we need to discuss around what happens with leases for a limbo'd mount.

satoqz:
Heh, another `-force` flag incoming I suppose.

satoqz:
I do want to note that not resetting the read-only error does not strike me as the intended mechanism to prevent deletions that bypass revocation. At best, we can mark this an unintended error that debatably prevents something worse happening unintentionally.

cipherboy:
@satoqz said:

> I do want to note that not resetting the read-only error does not strike me as the intended mechanism to prevent deletions that bypass revocation.

This is an interesting point. I do think we'd want the mount to be read-only if the backend doesn't exist; nothing should write to it and we don't want to try handling any lease invalidations. :shrug: 

I do agree it is entertaining, though.

satoqz:
@cipherboy so what are our options? add a `force` flag to mount disabling?

cipherboy:
@satoqz Yes, though I think we need to confirm that leases and tokens get removed when the backend doesn't exist and that they don't linger for any recreated backend to handle.

cipherboy:
@satoqz Looking back at the code, afaict, this deletion error is darn near original behavior, so I think we go ahead with a `-force` option of some sort.

satoqz:
Note to self, I believe in https://github.com/openbao/openbao/pull/3710 we hit a related bug where even if you'd reload the mount after making the plugin available, it'd remain read-only as set on unseal with no fix besides a seal/unseal cycle.

satoqz:
...so I think, @cipherboy, we want to fix at least the reloading in v2.7 so migrating removed builtin mount -> plugin is smoother.

PR #3488, opened 2026-07-14
Fix removal of config-based audit device

### LINKED ISSUE #3477: Re-creating an audit device with a different configuration of a removed audit device does not work properly
The audit devices are configured in the config file. 

When we start with a certain configuration, the device is created.
If I want to update the configuration of the device, it first needs to be removed, causing the device to be disabled and then added with the new configuration should result in the new config to be effective. 
However, the new config always gives an error: 

[ERROR] core: read-only post-unseal setup failed: error="failed to validate audit device config syslog/: modifications to audit devices are not allowed: audit device syslog/ is missing option elide_list_responses in the audit table but is present in the config" 

**To Reproduce**

* Add the audit device: 
```
audit "syslog" "syslog" {
  description = "Syslog audit device"
  options {
    tag      = "hcv-bao"
    facility = "user"
 }
}
```

* Start bao
* Stop bao 
* Remove the audit section for syslog
* Start bao again (works, device is no longer there) 

* Edit the config, add option: elide_list_responses 
```
audit "syslog" "syslog" {
  description = "Syslog audit device"
  options {
    tag      = "hcv-bao"
    facility = "user"

    elide_list_responses = "true"

  }
}
```
* Start bao. 
An error occurs: [ERROR] core: read-only post-unseal setup failed: error="failed to validate audit device config syslog/: modifications to audit devices are not allowed: audit device syslog/ is missing option elide_list_responses in the audit table but is present in the config"

**Steps to reproduce the behavior:**
* Add the first config.
* bao read sys/audit
syslog/     map[description:Syslog audit device local:false options:map[📁facility:user tag:hcv-bao] path:syslog/ type:syslog]
* Delete the syslog audit config
* bao read sys/audit
(audit device not listed)
* Add the second config
* bao read sys/audit
An error occurs: [ERROR] core: read-only post-unseal setup failed: error="failed to validate audit device config syslog/: modifications to audit devices are not allowed: audit device syslog/ is missing option elide_list_responses in the audit table but is present in the config"

**Expected behavior**
After the last step, I expect
* bao read sys/audit
syslog/     map[description:Syslog audit device local:false options:map[elide_list_responses:true facility:user tag:hcv-bao] path:syslog/ type:syslog]

**Environment:**
* OpenBao CLI Version (retrieve with `bao version`): OpenBao v2.3.2 (b1a68f558c89d18d38fbb8675bb6fc1d90b71e98), built 2025-08-08T04:05:32Z
* Server Operating System/Architecture: Both MacOS and linux (production)

```hcl

log_requests_level = "debug"
log_level = "debug"
log_file = "/tmp/vault.log"
ui            = true
cluster_addr  = "http://127.0.0.1:8201"
api_addr      = "http://127.0.0.1:8200"
disable_mlock = true
plugin_directory = "plugins"
unsafe_allow_api_audit_creation = true

storage "raft" {
  path    = "./data"
  node_id = "127.0.0.1"
}

listener "tcp" {
  address       = "0.0.0.0:8200"
  cluster_address = "0.0.0.0:8201"
  tls_disable = 1
}

audit "syslog" "syslog" {
  description = "Syslog audit device"
  options {
    tag      = "hcv-bao"
    facility = "user"
    elide_list_responses = "true"
 }
}

```

**Additional context**
We found the problem to be in the vault/audit.go file, line 723

code snippet: 
```
		if standby {
			c.logger.Warn("audit device present in storage but not standby node configuration; this may be a false-positive depending on data replication state", "path", auditMount.Path)
			continue
		}

		c.logger.Info("disabling removed audit device", "path", auditMount.Path)
		if existed, err := c.disableAudit(ctx, auditMount.Path, standby); existed && err != nil {
			return fmt.Errorf("failed to disable removed audit %v: %w", auditMount.Path, err)
		}
```
**Explanation**
If standby node (standby is true) the if statement "continues" 
If standby is false, so leader, then code continues into: 
```
if existed, err := c.disableAudit(ctx, auditMount.Path, standby)
```
Where the standby parameter is false. 
The disableAudit function has a signature of:
```
func (c *Core) disableAudit(ctx context.Context, path string, updateStorage bool) (bool, error) {
```
so, apparently, updateStorage is set to false, i.e. the storage is not updated, causing the error the next time the config is added with the new config (which does not map to the old config)

I have tested it with 
```
if existed, err := c.disableAudit(ctx, auditMount.Path, true)
```
And then it works. 

Happy to file a PR with code change and tests

satoqz:
## Description

The passed parameter is whether we should write, not whether we're standby. If we are standby, we bail earlier anyhow, so this can default to `true.`

Credits to @nielsdt-rabobank who suggested the same patch that I landed on, I'm opening this up just so we can squeeze it into v2.6.0 GA.

## Rationale

Resolves: https://github.com/openbao/openbao/issues/3477

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3488.txt
+++ b/changelog/3488.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/audit: Ensure config-driven audit mounts are removed from storage when removed from configuration.
+```
--- a/vault/audit.go
+++ b/vault/audit.go
@@ -720,7 +720,7 @@ func (c *Core) handleAuditLogSetup(ctx context.Context, standby bool) error {
 		}
 
 		c.logger.Info("disabling removed audit device", "path", auditMount.Path)
-		if existed, err := c.disableAudit(ctx, auditMount.Path, standby); existed && err != nil {
+		if existed, err := c.disableAudit(ctx, auditMount.Path, true); existed && err != nil {
 			return fmt.Errorf("failed to disable removed audit %v: %w", auditMount.Path, err)
 		}
 	}


wslabosz-reply (dismissed):
does this one require a changelog?

satoqz:
yes

wslabosz-reply (approved):

cipherboy (commented):
Thanks @satoqz! I hadn't had time to reproduce this yet.

cipherboy (approved):

cipherboy:
Backported in #3492.

PR #3506, opened 2026-07-15
Fix policy persistence issue, introduce PATCH support for policies

### LINKED ISSUE #3502: allow_wildcards_in_identity_templates and allow_slashes_in_identity_templates not persisted?
> Hi,
> 
> I’m trying to set `allow_wildcards_in_identity_templates=true` for the policy `sys/policy/forgejo`, but the value is not persisted.
> 
> The write succeeds, but reading the policy still shows:
> 
> allow_wildcards_in_identity_templates: false
> 
> It looks like the field is being ignored during update. Could you please check if this is expected behavior or a bug?
> 
> Server and client are on the same version.

_Originally posted by @Blacks-Army in https://github.com/openbao/openbao/discussions/3498#discussioncomment-17641659_

cipherboy:
## Description

```
    Fix policy persistence for template controls
    
    When serializing policies, Vault had implemented a pattern which
    separated the data layer (policy.Entry) from the usage layer
    (policy.Policy). The former manually copied fields from the latter which
    resulted in inconsistent serialization if care was not taken to update
    both.
    
    As we've added more fields, it makes less sense to keep the two
    separate especially as the rest of OpenBao does not adopt this pattern
    generally. Additionally the unused Version field was removed.
    
    This should prevent future issues of the same type.
```

```
    Add PATCH support to sys/policies/acl
    
    When the policy API was originally written, there was only one field in
    the object: policy. Since introducing more fields, we might want to
    modify policies without updating the actual contents of the policy and
    without overwriting defaults. Hence, PATCH support.
```

## Rationale

Resolves: #3502

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3506.txt
+++ b/changelog/3506.txt
@@ -1,6 +1,9 @@
 ```release-note:bug
 core/policies: Fix persistence of new `allow_wildcards_in_identity_templates`, `allow_slashes_in_identity_templates` parameters in storage.
 ```
-```release-note:feature
+```release-note:improvement
 core/policies: Introduce PATCH operation support for modifying the values of specific parameters.
+```
+```release-note:bug
+core/policies: Allow removal of `expiration` from policies on POST/PUT to `sys/policies/acl/<name>` by eliding the parameter; use PATCH to update policies without modifying unnecessary parameters.
 ```
--- a/changelog/3506.txt
+++ b/changelog/3506.txt
@@ -1,0 +1,6 @@
+```release-note:bug
+core/policies: Fix persistence of new `allow_wildcards_in_identity_templates`, `allow_slashes_in_identity_templates` parameters in storage.
+```
+```release-note:feature
+core/policies: Introduce PATCH operation support for modifying the values of specific parameters.
+```
--- a/internal/vault/logical_system.go
+++ b/internal/vault/logical_system.go
@@ -2646,8 +2646,8 @@ func (b *SystemBackend) handlePoliciesPatch(policyType policy.Type) framework.Op
 		if name == "" {
 			return logical.ErrorResponse("policy name must be provided in the URL"), nil
 		}
-		if name != strings.ToLower(name) {
-			name = strings.ToLower(name)
+		if n := strings.ToLower(name); name != n {
+			name = n
 			resp.AddWarning(fmt.Sprintf("policy name was converted to %s", name))
 		}
 
@@ -2656,11 +2656,11 @@ func (b *SystemBackend) handlePoliciesPatch(policyType policy.Type) framework.Op
 			return handleError(err)
 		}
 		if oldPolicy == nil {
-			return logical.ErrorResponse("unable to fetch role entry to patch"), nil
+			return handleError(logical.CodedError(http.StatusNotFound, "policy not found"))
 		}
 
 		pol := &policy.Policy{
-			Name:      strings.ToLower(name),
+			Name:      name,
 			Type:      policyType,
 			Namespace: ns,
 
--- a/internal/vault/logical_system.go
+++ b/internal/vault/logical_system.go
@@ -2589,6 +2589,8 @@ func (b *SystemBackend) handlePoliciesSet(policyType policy.Type) framework.Oper
 			pol.Expiration = expirationRaw.(time.Time)
 		} else if ttlOk {
 			pol.Expiration = time.Now().Add(time.Duration(ttlRaw.(int)) * time.Second)
+		} else {
+			pol.Expiration = time.Time{}
 		}
 
 		if !pol.Expiration.IsZero() && !time.Now().Before(pol.Expiration) {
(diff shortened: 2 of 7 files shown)


internal/vault/policy_store_test.go
```diff
@@ -107,12 +107,40 @@ func testPolicyStoreCRUD(t *testing.T, core *Core, shares [][]byte, token string
 		require.NoError(t, err)
 	}
 
-	testPolicyStoreCRUDOneShot(t, ps, ns)
```
cipherboy, line 110: @openbao/openbao-core-committers This is part of why we missed this: the persistence test doesn't actually test persistence as it reuses the pre-sealed policy store reference so all operations were still in-memory anyways.

(The other half is that the difference above between `Entry`/`Policy` meant we couldn't assume policy there implied that all fields were there.)

changelog/3506.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:bug
+acl/policies: Fix persistence of new `allow_wildcards_in_identity_templates`, `allow_slashes_in_identity_templates` parameters in storage.
+```
+```release-note:feature
+sys/policies: Introduce PATCH operation support for modifying the values of specific parameters.
+```
```
satoqz, line 9:
````suggestion
```release-note:bug
core/policies: Fix persistence of new `allow_wildcards_in_identity_templates`, `allow_slashes_in_identity_templates` parameters in storage.
```
```release-note:feature
core/policies: Introduce PATCH operation support for modifying the values of specific parameters.
```
````

Changelog entries, including those in v2.6.0, all use `core/policies`.

(will do actual review tomorrow)

File: internal/vault/logical_system_paths.go [hunk 3]
```diff
@@ -3871,6 +3871,17 @@ func (b *SystemBackend) policyPaths() []*framework.Path {
 					},
 					Summary: "Add a new or update an existing ACL policy.",
 				},
+				logical.PatchOperation: &framework.PathOperation{
+					Callback: b.handlePoliciesPatch(policy.TypeACL),
+					Responses: map[int][]framework.Response{
+						http.StatusNoContent: {{
+							Description: "OK",
```
wslabosz-reply, line 3878: I know its prevalent throughout the codebase, but `204` description is actually `No Content`

changelog/3506.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:bug
+core/policies: Fix persistence of new `allow_wildcards_in_identity_templates`, `allow_slashes_in_identity_templates` parameters in storage.
+```
+```release-note:feature
```
phil9909:
````suggestion
```release-note:improvement
````

File: changelog/3506.txt [hunk 5]
```diff
@@ -0,0 +1,6 @@
...
+core/policies: Introduce PATCH operation support for modifying the values of specific parameters.
+```
```
phil9909, line 6: The change in [`internal/vault/logical_system.go:2593`](https://github.com/openbao/openbao/pull/3506/changes#diff-c093a8878d6ea7b98f2e81894c4b6ab0bda3b83037487127b21af2b763472d3fR2593) fixed another bug, right?

````suggestion
```
```release-note:bug
core/policies: Allow removal of `expiration` from policies on POST/PUT to `sys/policies/acl/<name>` by eliding the parameter. Use PATCH to update policies without modifying unnecessary parameters.
```
````

**Review by phil9909 — DISMISSED** [review]

File: website/content/api-docs/system/policies.mdx [hunk 6]
```diff
@@ -72,19 +72,21 @@ $ curl \
...
+| `PATCH` | `/sys/policies/acl/:name` | Unspecified field values preserved.   |
 
 ### Parameters
 
 - `name` `(string: <required>)` – Specifies the name of the policy to create.
   This is specified as part of the request URL.
 
 - `policy` `(string: <required>)` - Specifies the policy document. This can be
-  base64-encoded to avoid string escaping.
+  base64-encoded to avoid string escaping. Optional on patch commands.
```
satoqz, line 89: We usually have a separate section for patch, and this would also be a good place to explain why one would use patch over update, which isn't immediately obvious.

internal/vault/policy/policy.go
```diff
@@ -94,18 +94,21 @@ var cap2Int = map[string]uint32{
...
+	// These fields are persisted in storage.
+	DataVersion                       int       `json:"data_version"`
+	CASRequired                       bool      `json:"cas_required"`
+	Raw                               string    `json:"raw"`
+	Type                              Type      `json:"type"`
+	Templated                         bool      `json:"templated"`
+	Expiration                        time.Time `json:"expiration"`
+	Modified                          time.Time `json:"modified"`
+	AllowWildcardsInIdentityTemplates bool      `json:"allow_wildcards_in_identity_templates"`
+	AllowSlashesInIdentityTemplates   bool      `json:"allow_slashes_in_identity_templates"`
```
satoqz: I haven't tested it, but does this work against older data given `Entry` previously did not have JSON tags at all so I'd assume it wrote differently named fields?

internal/vault/logical_system.go
```diff
@@ -2630,6 +2632,98 @@ func (b *SystemBackend) handlePoliciesSet(policyType policy.Type) framework.Oper
...
+		if err != nil {
+			return nil, err
+		}
+
+		name := data.Get("name").(string)
+		if name == "" {
+			return logical.ErrorResponse("policy name must be provided in the URL"), nil
+		}
+		if name != strings.ToLower(name) {
+			name = strings.ToLower(name)
```
satoqz:
```suggestion
		if n := strings.ToLower(name); name != n {
			name = n
```

internal/vault/logical_system.go
```diff
@@ -2630,6 +2632,98 @@ func (b *SystemBackend) handlePoliciesSet(policyType policy.Type) framework.Oper
...
+		oldPolicy, err := b.Core.policyStore.GetPolicy(ctx, name, policyType)
+		if err != nil {
+			return handleError(err)
+		}
+		if oldPolicy == nil {
+			return logical.ErrorResponse("unable to fetch role entry to patch"), nil
+		}
+
+		pol := &policy.Policy{
+			Name:      strings.ToLower(name),
```
satoqz:
```suggestion
			Name:      name,
```

We already lowercased above.

internal/vault/logical_system.go
```diff
@@ -2630,6 +2632,98 @@ func (b *SystemBackend) handlePoliciesSet(policyType policy.Type) framework.Oper
...
+			name = strings.ToLower(name)
+			resp.AddWarning(fmt.Sprintf("policy name was converted to %s", name))
+		}
+
+		oldPolicy, err := b.Core.policyStore.GetPolicy(ctx, name, policyType)
+		if err != nil {
+			return handleError(err)
+		}
+		if oldPolicy == nil {
+			return logical.ErrorResponse("unable to fetch role entry to patch"), nil
```
satoqz:
```suggestion
			return handleError(logical.CodedError(http.StatusNotFound, "policy not found"))
```

seems more appropriate and is in line with namespace patch.

DrDaveD (dismissed):
I tried cherry-picking the first commit to the 2.6.x branch and testing the functionality there and can confirm that `allow_slashes_in_identity_templates = true` was preserved across reboots and raft leader switches.

internal/vault/logical_system_paths.go
```diff
@@ -3871,6 +3871,17 @@ func (b *SystemBackend) policyPaths() []*framework.Path {
 					},
 					Summary: "Add a new or update an existing ACL policy.",
 				},
+				logical.PatchOperation: &framework.PathOperation{
```
satoqz, line 3874: Needs an entry in the almighty `sysHelp["policy"]`.

internal/vault/policy/policy.go
cipherboy: Ah, shoot, yes. Let me add a known-serialization test.

internal/vault/logical_system.go
cipherboy: True; I think we need to apply this consistently, I was modeling after PKI which didn't.

internal/vault/logical_system_paths.go
cipherboy: @wslabosz-reply Fair. Let's address these all in a unit later though.

website/content/api-docs/system/policies.mdx
cipherboy: @satoqz We do, but I'm not sure I'm sold on the amount of duplication any more. They're both write operations and do the same thing with the same parameters... That's why I added the `Scope` field above as a new pattern.

I take it you're not a fan though 😆 Other suggestions to keep the size of the docs down?

satoqz: @cipherboy that's sensible, but then let's make it very clear that this section handles all (particularly, I'd want the heading to be `## Create/Update/Patch ACL policy`).

changelog/3506.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
+core/policies: Fix persistence of new `allow_wildcards_in_identity_templates`, `allow_slashes_in_identity_templates` parameters in storage.
+```
+core/policies: Introduce PATCH operation support for modifying the values of specific parameters.
```
phil9909, line 5: something went wrong here when applying the suggestions

website/content/api-docs/system/policies.mdx
cipherboy: @satoqz Updated, added a href to keep the old version as well.

satoqz (approved):

PR #3518, opened 2026-07-16
Fix limit=0 and negative limits bypassing pagination_limit ACL cap

### LINKED ISSUE #3517: pagination_limit ACL cap is bypassed by limit=0 (and negative limits)
**Describe the bug**
 A policy sets pagination_limit on a list/scan path, but a caller with list access can pass limit=0 (or a negative) and get the full result set — the configured cap isn't enforced for non-positive values.

**To Reproduce**
Policy with pagination_limit = 10 on a path that has >10 keys; list it with limit=0 → you get all keys, not 10.

**Expected behavior**
Expected: non-positive limit on a capped path is clamped to the cap (consistent with how max and an omitted limit are already handled).

**Environment:**
latest `main` branch.

**Additional context**
Nothing

RealFakeAccount:
## Description

A policy sets pagination_limit on a list/scan path, but a caller with list access can pass limit=0 (or a negative) and get the full result set — the configured cap isn't enforced for non-positive values.

This PR fix this bug and added test for it.

## Rationale

Resolves: #3517

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/external_tests/policy/policy_test.go
+++ b/internal/vault/external_tests/policy/policy_test.go
@@ -478,15 +478,12 @@ path "kv/metadata/" {
 	require.NotNil(t, resp.Data)
 	require.Equal(t, 10, len(resp.Data["keys"].([]interface{})))
 
-	// Test negative value.
-	resp, err = client.Logical().ReadWithData("kv/metadata/d", map[string][]string{
+	// Negative values are denied.
+	_, err = client.Logical().ReadWithData("kv/metadata/d", map[string][]string{
 		"list":  {"true"},
 		"limit": {"-1"},
 	})
-	require.NoError(t, err, "failed to list with limit=-1")
-	require.NotNil(t, resp)
-	require.NotNil(t, resp.Data)
-	require.Equal(t, 10, len(resp.Data["keys"].([]interface{})))
+	require.Error(t, err, "expected failure to list with limit=-1")
 
 	// This endpoint has no limit.
 	resp, err = client.Logical().ReadWithData("kv/metadata/a", map[string][]string{
--- a/internal/vault/external_tests/policy/policy_test.go
+++ b/internal/vault/external_tests/policy/policy_test.go
@@ -468,6 +468,26 @@ path "kv/metadata/" {
 	require.NotNil(t, resp.Data)
 	require.Equal(t, 10, len(resp.Data["keys"].([]interface{})))
 
+	// Test '0' value.
+	resp, err = client.Logical().ReadWithData("kv/metadata/d", map[string][]string{
+		"list":  {"true"},
+		"limit": {"0"},
+	})
+	require.NoError(t, err, "failed to list with limit=0")
+	require.NotNil(t, resp)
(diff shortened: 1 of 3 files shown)


phil9909:
I'm wondering, if we should reject negative limits as `400 Bad Request` instead.

But yes, explicit 0 should be treated exactly like a non-existent limit.

RealFakeAccount:
> I'm wondering, if we should reject negative limits as `400 Bad Request` instead.
> 
> But yes, explicit 0 should be treated exactly like a non-existent limit.

for negatives, 403 is ez cuz we just `return ret`. I'll make the change rn. 

if you wanna do 400 then we need to bring error field into ACLResults and change to downstream. It's prob better to do it in another PR?

internal/vault/policy/acl.go
```diff
@@ -625,6 +625,14 @@ CHECK:
 					if val > permissions.PaginationLimit {
 						return ret
 					}
+					// Per PR discussion, deny negative limits
```
phil9909:
```suggestion
					// deny negative limits
```

internal/vault/external_tests/policy/policy_test.go
```diff
@@ -468,6 +468,23 @@ path "kv/metadata/" {
...
+	require.NotNil(t, resp)
+	require.NotNil(t, resp.Data)
+	require.Equal(t, 10, len(resp.Data["keys"].([]interface{})))
+
+	// Negative values are denied.
+	_, err = client.Logical().ReadWithData("kv/metadata/d", map[string][]string{
+		"list":  {"true"},
+		"limit": {"-1"},
+	})
+	require.Error(t, err, "expected failure to list with limit=-1")
```
phil9909:
```suggestion
	require.ErrorContains(t, err, "permission denied", "expected failure to list with limit=-1")
```

phil9909 (dismissed):
Just two minor things, otherwise good.

403 is fine for me. You will also get it if you set "limit=invalid"

internal/vault/policy/acl.go
RealFakeAccount: lolol sure

phil9909 (dismissed):
LGTM

Only the DCO signoff is missing now for commit https://github.com/openbao/openbao/pull/3518/commits/26167b14d5682754201781b86d4852d2f9d05667

Please click on the DCO check for details on how to fix this.

satoqz:
Also needs a changelog entry (at `changelog/3518.txt`), I'd suggest:

````
```release-note:bug
core/policies: Fix `limit=0` and negative limits bypassing the `pagination_limit` ACL option.
```
````

RealFakeAccount:
Heyy it's done. Thanks for pointing them out :)

wslabosz-reply (approved):

satoqz (approved):

cipherboy (approved):

cipherboy:
Thanks @RealFakeAccount!

(I will note that denying negative values instead of overwriting them does go against what the contract is at the storage layer, but we'll see what it breaks.)

PR #3290, opened 2026-06-16
Switch to automatic nonce generation

cipherboy:
## Description

Addresses barrier encryption, dhutil, and transit use of random nonces with the Go standard library helper. 

## Rationale

This shifts nonce generation to the crypto library, putting it inside the FIPS boundary and letting us run further with FIPS mode.

See also: https://github.com/openbao/go-kms-wrapping/pull/45

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3521, opened 2026-07-16
Fix Test_RandomOpsTransactionalBackends

satoqz:
## Description

`var out []byte` creates a `nil` slice that is returned directly by `gcm.Open` if the passed data is empty. Before https://github.com/openbao/openbao/pull/3290, the barrier would always return at least a zero-size slice, which is consistent with other backends as asserted in the test.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/barrier/aes_gcm.go
+++ b/internal/vault/barrier/aes_gcm.go
@@ -1065,7 +1065,11 @@ func (b *AESGCMBarrier) decrypt(path string, gcm cipher.AEAD, cipher []byte) ([]
 	}
 
 	raw := cipher[5:]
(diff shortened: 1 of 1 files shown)


wslabosz-reply (approved):

cipherboy (approved):

cipherboy:
So far I'm not aware of any breakage which occurs, so I'm happy calling this complete and merging. Otherwise we can add a changelog when we backport.

satoqz:
@cipherboy https://github.com/openbao/openbao/pull/3290 isn't released anywhere so we're good I believe?

PR #3527, opened 2026-07-16
Don't read BAO_WRAP_TTL if DisableEnvironment is set

satoqz:
## Description

This plugs up the final hole (that I could find?) not yet covered by `DisableEnvironment`.

## Rationale

Related:
- https://github.com/openbao/go-kms-wrapping/pull/83#discussion_r3079358886
- https://github.com/openbao/go-kms-wrapping/issues/110

## Acknowledgements

       also allowed.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -1326,11 +1326,16 @@ func (c *Client) SetPolicyOverride(override bool) {
 // doesn't need to be called externally.
 func (c *Client) NewRequest(method, requestPath string) *Request {
 	c.modifyLock.RLock()
+	c.config.modifyLock.RLock()
+
 	addr := c.addr
 	token := c.token
 	mfaCreds := c.mfaCreds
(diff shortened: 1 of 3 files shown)


wslabosz-reply (approved):
LGTM

satoqz:
Doesn't need backport for my use case because because Transit seal/KMS remain built-in and aren't plugins + they won't need code changes to account for this, so they build against this patch automatically when 2.7 is released.

Could still make a point that it should be backported regardless.

cipherboy (approved):
@satoqz I'm fine with a backport, though I'm not sure if we'll really tag a new API/SDK version fwiw.

satoqz:
@cipherboy we must tag at least SDK due to test container breakage, which is particularly annoying to deal with in go-kms-wrapping.

But I'm ok leaving API as-is.

PR #3528, opened 2026-07-16
Fix typo in warning message

avandecreme:
## Description

Just a small typo fix

## Rationale

Resolves: #

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/command/operator_rekey.go
+++ b/internal/command/operator_rekey.go
@@ -317,24 +317,26 @@ func (c *OperatorRekeyCommand) init(client *api.Client) int {
 		// Print warnings about recovery, etc.
 		if len(c.flagPGPKeys) == 0 {
 			c.UI.Warn(wrapAtLength(
-				fmt.Sprintf("WARNING! If you lose the keys after they are returned, there is no "+
-					"recovery. Consider canceling this operation and re-initializing "+
-					"with the -pgp-keys flag to protect the returned %s keys along "+
-					"with -backup to allow recovery of the encrypted keys in case of "+
-					"emergency. You can delete the backed up keys later using the -delete "+
-					"flag.", strings.ToLower(keyTypeRequired)),
+				fmt.Sprintf("WARNING! If you lose the keys after they are "+
+					"returned, there is no recovery. Consider canceling this "+
+					"operation and re-initializing with the -pgp-keys flag to protect "+
+					"the returned %s keys along with -backup to allow recovery of the "+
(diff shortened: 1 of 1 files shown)


satoqz:
Do you also want to fix the same message in `operator_rekey.go`?

avandecreme:
> Do you also want to fix the same message in `operator_rekey.go`?

Do you mean another `operator_rekey.go` file?

Or `operator_rotate_keys.go`? If so, it seems fine [there](https://github.com/openbao/openbao/blob/main/internal/command/operator_rotate_keys.go#L326) (though the whole warning is worded a bit differently).  
Would you like to harmonize that?

satoqz:
Ah, sorry, wrong file. Yes, I think the version in `operator_rotate_keys.go` reads better.

avandecreme:
Looking at `internal/command/operator_rotate_keys.go` there seems to be a slight error though.

It says
> Consider **rerunning** this operation and re-initializing

but shouldn't it be

> Consider **canceling** this operation and re-initializing

 ?

satoqz:
Should be better, yeah. @wslabosz-reply any reason it is "rerunning" here instead?

cipherboy:
Not to nitpick this to death, but wouldn't we have already initialized? So there's nothing to cancel; the command returned. So most precisely:

> Consider rerunning this operation to **rotate**

(in that paragraph). 

In the later one on like, e.g., line 366, it is correctly "cancelling".

wslabosz-reply:
in both cases warning is printed after we complete the operation (status == complete), so `rerunning` is right (as a "repeat the operation with different flag")

avandecreme:
I tested calling `bao operator rekey -init` and `bao operator rotate-keys -init` twice. In both cases, you get a `* (rekey|rotation) already in progress` error.

You need to call `bao operator (rekey|rotate-keys) -cancel`, so the `canceling` term is better in my opinion.

wslabosz-reply:
@avandecreme 
I'm completely lost with the discussion, could you make the changes that you feel are appropriate and we would be able to comment/discuss on the diff?

avandecreme:
I got lost as well I think 😅 
I copied the message from rotate-keys into rekey. I can't see anything wrong with them anymore.

cipherboy (commented):
I'm happy with the accuracy of these updates 😆

cipherboy (approved):

wslabosz-reply (approved):

Issue #3189, opened 2026-05-27
Allow for selfsigned-certificates using auto TLS


pree:
**Is your feature request related to a problem? Please describe.**
For deployments behind re-encrypting proxies (for example WAFs) it would be nice to have OpenBao be able to create self-signed certificates, similar to the acme tls listeners.
This could also be extended to spin up an openbao cluster/node with self-signed certificates instead of disabling tls.

**Describe the solution you'd like**
Extension to the tls listener stanza to allow for selfsigned certificate generation.
```
tls_selfsigned_cache_path
tls_selfsigned_key_type
tls_selfsigned_domains
...
```

**Describe alternatives you've considered**
Always create self-signed certificates in case no tls configuration is present in the listener stanza.

**Explain any additional use-cases**
Compared to `tls_disable` this could increase security, though this should be in the responsibility of the OpenBao operator to configure TLS on their instance, even if it's behind a re-encrypting reverse-proxy.
It still make it easier to manage a secure tunnel between the reverse proxy and OpenBao as OpenBao could also periodically rotate the certificate.

**Additional context**
Other tools are doing this similarly, for example Argo CD: https://argo-cd.readthedocs.io/en/stable/operator-manual/tls/
Though in this example it's generating self-signed certificates when no additional tls configuration is set.

PR #3530, opened 2026-07-16
Automatic reload of TLS certificate files without SIGHUP signal

### LINKED ISSUE #3533: Add automatic TLS certificate reload on file changes for server listeners

The server's TLS listener reads `tls_cert_file` and `tls_key_file` only once
at startup and keeps the certificate in memory. There is no internal mechanism
that monitors these files when their content changes on disk, and nothing
notifies the process, which continues to present the old certificate.

On Kubernetes with cert-manager, this is a daily problem. The listener's
certificate is provided through a Secret mounted as a volume and renewed
automatically. After a renewal, there is a drift:

```
Kubernetes Secret has certificate B
Mounted volume has certificate B
OpenBao in memory has certificate A
```

The only current way to reload it is through a trigger external to the process:
sending a SIGHUP signal, through a CronJob, through a postRenew hook, or with a sidecar container.

This external coupling is costly in this context:

* Short-lived certificates are the norm. This can silently cause
  client TLS connections to fail.
* Restarting the pod is not neutral: with Shamir unsealing without auto-unseal,
  each restart requires a manual unseal (3 shares per pod).
* The sidecar requires shareProcessNamespace: true (degraded process isolation
  between containers in the pod), and requires finding an external method to trigger the certificate refresh.

Add an opt-in option to the listener that gives it the ability to detect
a change to tls_cert_file / tls_key_file itself and trigger its
own reload, reusing what is currently called for `SIGHUP`,
just with an additional trigger driven by configuration.

```
listener "tcp" {
  tls_cert_file = "/openbao/tls/tls.crt"
  tls_key_file  = "/openbao/tls/tls.key"

  tls_auto_reload          = true
  tls_auto_reload_interval = "30s"   # optional
}
```

By default, the parameter is false, with no behavior change for existing
deployments, and SIGHUP continues to work identically. The rotation
becomes a simpler sequence and is integrated entirely into OpenBao.

For the detection mechanism, I propose polling based on a content hash with a ticker that
compares a SHA-256 of the files, adds no dependency, and behaves
identically on network file systems.

* SIGHUP + sidecar (current solution): works but adds a
  component to maintain and degrades pod isolation.
* CronJob or postRenew hook that sends the signal: reacts to a timer or to a
  cert-manager event, not to the file change, which makes it a more fragile solution.
* rollout restart of the StatefulSet: requires sequential manual Shamir unsealing, which is not feasible.

* The need is not specific to Kubernetes: any operator who currently manages
  rotation with certbot or a cron job (followed by a SIGHUP)
  could do without it.
* Any file-based certificate rotation scenario (Nomad, systemd,
  network mounts) benefits from the same internal trigger.

**Additional context**

* Useful precedent: #3038 (merged) added reloading of the Agent's client TLS
  configuration on `SIGHUP`, in response to #1557. It validates the principle of “adding
  reload paths to components that did not have them,” but it does not
  affect the server listener and still relies on a signal trigger. The
  request here is complementary: it targets the server listener and removes
  the dependency on an external signal.

clement-software:
## Description

The TCP listener loads tls_cert_file and tls_key_file only once at startup and rereads them only when the process receives a SIGHUP signal. The reload mechanism works well, but its activation still depends on a trigger external to the process.

In Kubernetes, where certificates are generally issued by cert-manager and mounted from a Secret volume, this requires setting up your own mechanism, such as a CronJob or a sidecar container that monitors the certificate files in order to send a SIGHUP signal.

This reduces process isolation between containers with configurations such as `shareProcessNamespace: true`; see: https://github.com/openbao/openbao-helm/blob/e926cc75f9ca4e024a78a4a38c3d66f060d57d16/charts/openbao/values.yaml#L614. The comment relating to `extraContainers` in the Helm chart even suggests this model, but each deployment has to reimplement it: https://github.com/openbao/openbao-helm/blob/e926cc75f9ca4e024a78a4a38c3d66f060d57d16/charts/openbao/values.yaml#L610.

Missing a renewal signal can have significant consequences: with short-lived certificates, TLS handshakes may fail as soon as their cached certificate authority no longer matches the certificate presented by the server. Restarting the pod is not a neutral solution either: with Shamir unsealing, every restart requires a manual unseal.

### Reproduction

To make the work easier, I’ve created a test folder with a readme file describing the steps and showing how to reproduce the changes made:
https://github.com/clement-software/openbao-testbench-tls-autoreload

1. Start bao server with tls_cert_file and tls_key_file.
2. Replace both files with a new certificate pair.
3. Open a TLS connection: the **old** certificate is still presented until an operator sends a `SIGHUP` signal.

### Solution

Add an optional periodic polling mechanism that reuses the existing reload path. When enabled, the server calculates the fingerprints of the configured certificate and key files at a fixed interval and, when a change is detected, calls the same CertificateGetter.Reload method that is already registered for SIGHUP. The behaviour of SIGHUP remains unchanged.

```hcl
listener "tcp" {
  tls_cert_file            = "/openbao/tls/tls.crt"
  tls_key_file             = "/openbao/tls/tls.key"

  tls_auto_reload          = true    # default value: false
  tls_auto_reload_interval = "30s"   # default value: 30 s, minimum: 5 s
}
```

As PEM files are small, calculating their fingerprints every 30 seconds using polling to compare two hashes is a solution with a low resource cost.

Error handling: if a reload fails, for example in the event of a partial update where the certificate is written before the key, the previous certificate continues to be presented, a warning is logged, and another attempt is made during the next cycle. Reload replaces the presented certificate only after `tls.X509KeyPair` succeeds. An invalid state on disk can therefore never make the listener unavailable.

### Out of scope

* tls_client_ca_file is not monitored: it is incorporated into tls.Config.ClientCAs at startup and is not currently reloaded, even by SIGHUP. Making it reloadable requires a separate change using GetConfigForClient.

### Tests

* Unit test covering the following cases: no reload while the files remain unchanged, then exactly one reload after the certificate file is replaced.
* Manual test: repeat the reproduction procedure above with `tls_auto_reload = true`. The new certificate is presented within a maximum period corresponding to one interval, without sending signal SIGHUP.

## Rationale

This change removes the need for an external mechanism to trigger TLS certificate reloads in Kubernetes deployments, while preserving the existing SIGHUP behaviour.

Related prior work: PR #3038 added `SIGHUP`-based client-side TLS reloading in the agent. This change is complementary: it targets the server listener and removes the need for an external trigger.

There is this request: https://github.com/openbao/openbao/issues/3189 which my implementation could address, regarding: OpenBao could also periodically rotate the certificate.

Resolves: https://github.com/openbao/openbao/issues/3533

## Acknowledgements

* [ X] By contributing this change, I certify I have not used generative AI
* [X ] By contributing this change, I certify I have signed-off on the
  [[DCO ownership](https://developercertificate.org/)](https://developercertificate.org/) statement
  also allowed.

### DIFF
--- a/changelog/3530.txt
+++ b/changelog/3530.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+command/server: Add `tls_auto_reload` configuration option to automatically reload TLS certificate and key files when their contents change, without requiring `SIGHUP`.
+```
--- a/internal/helper/configutil/listener.go
+++ b/internal/helper/configutil/listener.go
@@ -71,17 +71,21 @@ type Listener struct {
 	TLSDisableRaw any  `hcl:"tls_disable"`
 	TLSCertGetter any  `hcl:"-"`
 
-	TLSCertFile                      string   `hcl:"tls_cert_file"`
-	TLSKeyFile                       string   `hcl:"tls_key_file"`
-	TLSMinVersion                    string   `hcl:"tls_min_version"`
-	TLSMaxVersion                    string   `hcl:"tls_max_version"`
-	TLSCipherSuites                  []uint16 `hcl:"-"`
-	TLSCipherSuitesRaw               string   `hcl:"tls_cipher_suites"`
-	TLSRequireAndVerifyClientCert    bool     `hcl:"-"`
-	TLSRequireAndVerifyClientCertRaw any      `hcl:"tls_require_and_verify_client_cert"`
-	TLSClientCAFile                  string   `hcl:"tls_client_ca_file"`
-	TLSDisableClientCerts            bool     `hcl:"-"`
-	TLSDisableClientCertsRaw         any      `hcl:"tls_disable_client_certs"`
+	TLSCertFile                      string        `hcl:"tls_cert_file"`
+	TLSKeyFile                       string        `hcl:"tls_key_file"`
+	TLSMinVersion                    string        `hcl:"tls_min_version"`
+	TLSMaxVersion                    string        `hcl:"tls_max_version"`
+	TLSCipherSuites                  []uint16      `hcl:"-"`
+	TLSCipherSuitesRaw               string        `hcl:"tls_cipher_suites"`
+	TLSRequireAndVerifyClientCert    bool          `hcl:"-"`
+	TLSRequireAndVerifyClientCertRaw any           `hcl:"tls_require_and_verify_client_cert"`
+	TLSClientCAFile                  string        `hcl:"tls_client_ca_file"`
+	TLSDisableClientCerts            bool          `hcl:"-"`
+	TLSDisableClientCertsRaw         any           `hcl:"tls_disable_client_certs"`
+	TLSAutoReload                    bool          `hcl:"-"`
+	TLSAutoReloadRaw                 any           `hcl:"tls_auto_reload"`
+	TLSAutoReloadInterval            time.Duration `hcl:"-"`
+	TLSAutoReloadIntervalRaw         any           `hcl:"tls_auto_reload_interval"`
 
 	TLSACMECachePath               string   `hcl:"tls_acme_cache_path"`
 	TLSACMECADirectory             string   `hcl:"tls_acme_ca_directory"`
@@ -334,6 +338,32 @@ func ParseListeners(result *SharedConfig, list *ast.ObjectList) error {
 				}
 
 				l.TLSDisableClientCertsRaw = nil
+			}
+
+			if l.TLSAutoReloadRaw != nil {
+				if l.TLSAutoReload, err = parseutil.ParseBool(l.TLSAutoReloadRaw); err != nil {
+					return multierror.Prefix(fmt.Errorf("invalid value for tls_auto_reload: %w", err), fmt.Sprintf("listeners.%d", i))
+				}
+
+				l.TLSAutoReloadRaw = nil
+			}
+
+			if l.TLSAutoReloadIntervalRaw != nil {
+				if l.TLSAutoReloadInterval, err = parseutil.ParseDurationSecond(l.TLSAutoReloadIntervalRaw); err != nil {
+					return multierror.Prefix(fmt.Errorf("error parsing tls_auto_reload_interval: %w", err), fmt.Sprintf("listeners.%d", i))
+				}
+
+				if l.TLSAutoReloadInterval < 5*time.Second {
+					return multierror.Prefix(fmt.Errorf("tls_auto_reload_interval must be at least 5s"), fmt.Sprintf("listeners.%d", i))
+				}
+
+				l.TLSAutoReloadIntervalRaw = nil
+			} else if l.TLSAutoReload {
+				l.TLSAutoReloadInterval = 30 * time.Second
+			}
+
+			if l.TLSAutoReload && l.TLSCertFile == "" {
+				return multierror.Prefix(fmt.Errorf("tls_auto_reload requires tls_cert_file"), fmt.Sprintf("listeners.%d", i))
 			}
 		}
 
--- a/internal/helper/listenerutil/tls_autoreload.go
+++ b/internal/helper/listenerutil/tls_autoreload.go
@@ -1,0 +1,75 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package listenerutil
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"net"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/go-hclog"
+)
+
+func NewTLSReloadListener(ln net.Listener, paths []string, interval time.Duration, reload func() error, logger hclog.Logger) net.Listener {
+	wg := &sync.WaitGroup{}
+	stop := make(chan struct{})
+	wg.Go(func() {
+		pollTLSCertificateChanges(paths, interval, reload, stop, logger)
+	})
+	return &tlsReloadListener{Listener: ln, wg: wg, stop: stop}
+}
+
+type tlsReloadListener struct {
+	net.Listener
+	wg   *sync.WaitGroup
+	stop chan struct{}
+}
+
+func (l *tlsReloadListener) Close() error {
+	close(l.stop)
+	l.wg.Wait()
+	return l.Listener.Close()
+}
+
+func pollTLSCertificateChanges(paths []string, interval time.Duration, reload func() error, stopCh <-chan struct{}, logger hclog.Logger) {
+	hash := func() []byte {
+		h := sha256.New()
+		for _, p := range paths {
+			data, err := os.ReadFile(p)
+			if err != nil {
+				logger.Warn("tls auto-reload: cannot read file", "path", p, "error", err)
+				continue
+			}
+			h.Write(data)
+		}
+		return h.Sum(nil)
+	}
+
+	last := hash()
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-stopCh:
+			return
+		case <-ticker.C:
+			current := hash()
+			if bytes.Equal(current, last) {
+				continue
+			}
+
+			if err := reload(); err != nil {
+				logger.Warn("tls auto-reload: reload failed, keeping previous certificate", "error", err)
+				continue
+			}
+
+			last = current
+			logger.Info("tls auto-reload: reloaded TLS certificate", "paths", paths)
+		}
+	}
+}
--- a/internal/helper/listenerutil/tls_autoreload_test.go
+++ b/internal/helper/listenerutil/tls_autoreload_test.go
@@ -1,0 +1,53 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package listenerutil
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/go-hclog"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPollTLSCertificateChanges(t *testing.T) {
+	dir := t.TempDir()
+	cert := filepath.Join(dir, "tls.crt")
+	key := filepath.Join(dir, "tls.key")
+	require.NoError(t, os.WriteFile(cert, []byte("cert-v1"), 0o600))
+	require.NoError(t, os.WriteFile(key, []byte("key-v1"), 0o600))
+
+	reloaded := make(chan struct{}, 10)
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	go pollTLSCertificateChanges([]string{cert, key}, 10*time.Millisecond, func() error {
+		reloaded <- struct{}{}
+		return nil
+	}, stopCh, hclog.NewNullLogger())
+
+	// Unchanged files must not trigger a reload.
+	select {
+	case <-reloaded:
+		t.Fatal("reload triggered without file change")
+	case <-time.After(100 * time.Millisecond):
+	}
+
+	// A content change must trigger exactly one reload.
+	tmp := filepath.Join(dir, "tls.crt.tmp")
+	require.NoError(t, os.WriteFile(tmp, []byte("cert-v2"), 0o600))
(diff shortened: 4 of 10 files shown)


satoqz:
Hi @clement-software, this seems like a reasonable addition, though since it is a feature, would like to hear other maintainer's opinions.

I wonder if using https://github.com/fsnotify/fsnotify is worth it on platforms that are supported; I haven't quite convinced myself yet though since cert updates should usually tolerate some latency, and it would be more complexity + another dependency.

clement-software:
> Hi @clement-software, this seems like a reasonable addition, though since it is a feature, would like to hear other maintainer's opinions.
> 
> I wonder if using https://github.com/fsnotify/fsnotify is worth it on platforms that are supported; I haven't quite convinced myself yet though since cert updates should usually tolerate some latency, and it would be more complexity + another dependency.

Thanks for your feedback, and for appreciating the contribution. I hope the other maintainers will share your enthusiasm.  
I wasn’t familiar with fsnotify, but I’ve just had a look. It could also be a possible implementation, but it adds a dependency whose support varies depending on the OS. When I read the FAQ: 
```
Will a file still be watched when it's moved to another directory?
No, not unless you are watching the location it was moved to.
```
 I think there could be issues with symbolic links.  Whereas my implementation checks that the certificate’s content has actually changed, I get the impression that the fsnotify approach, which only monitors filesystem events, does not guarantee that the relevant content has actually changed. From looking at https://pkg.go.dev/github.com/fsnotify/fsnotify#pkg-variables, I also understand that events may be lost: 
It immediately becomes much more complicated to handle.
```
//  - inotify:      inotify returns IN_Q_OVERFLOW – because there are too
//                  many queued events (the fs.inotify.max_queued_events
//                  sysctl can be used to increase this).
//  - windows:      The buffer size is too small; WithBufferSize() can be used to increase it.
```

satoqz:
@clement-software that's fair! Just wanted to throw it in, I think the polling approach is entirely fine to start out with.

JanMa:
Thank you for your contribution @clement-software .

I personally think we should not add this to work around a limitation of Kubernetes. Updates to mounted Secrets and ConfigMaps are notoriously unreliable that's why the ecosystem has for the most part standardized on using something like [Stakater Reloader](https://github.com/stakater/Reloader). Whenever a ConfigMap or Secret is updated a rollout of all the workloads that mount it will be triggered. This gives you the cleanest and most reliable way to ensure the up to date values are used. For OpenBao this means you need to use an auto unseal mechanism, but I'd argue you need to do that anyways if you're running in K8S.

tsaarni:
I do not see this as a Kubernetes workaround at all. Since certificates are rotated regularly as part of normal operations, all servers should have this feature built-in, in my opinion. I don't see this as feature creep, if added :) Actually, triggering a SIGHUP or restarting the process feels more like a workaround to me :) Both require extra software to coordinate rotation, which seems unnecessary for a basic necessity.

I have implemented this feature in couple of project, both using inotify and without inotify. With that experience, I would also recommend avoiding inotify, because different files need to be monitored depending how the underlying files are updated.

clement-software:
Thank you for your feedback, @JanMa. I understand your argument, but it does not demonstrate that the feature is specific to Kubernetes. It only shows that there is another possible operational strategy in Kubernetes to work around this broader issue.

I was therefore going to reply to your message along exactly the same lines as @tsaarni , whom I would also like to thank for their feedback.

From discussions I have had at work and on LinkedIn, I know that this is a recurring user need, and it is not specific to Kubernetes.

The issue also exists outside Kubernetes, in any topology using a network volume on which certificate rotation could be performed with Docker, Podman, and so on. It may also apply to constrained environments that cannot use auto-unseal. This strategy comes at a cost, as it turns a certificate rotation into a restart or rolling restart of the instance.

So my question is: what risks would you see in this proposal, @JanMa ?
I see a recurring user need that could be addressed with a low maintenance cost and without any breaking changes.

Nerkho:
Ignoring the specific Kubernetes context, I think supporting automatic reload of TLS certificates makes sense considering the trend in the industry is to heavily shorten the lifetime of TLS certificates.

https://www.swisssign.com/en/blog/tls-validity-47-days.html

pree:
I also support this, it's become more and more common on applications to watch/fetch their configurations and certificates.
IMHO 30s is probably a bit shorter than actually needed as a default (e.g. even Let's Encrypt's 160h certificates are rotated with an overlap of 72 to 96 hours).

If we would focus for a perfect solution on Kubernetes we could have OpenBao read the TLS certificates from the kube api directly ^^

internal/helper/listenerutil/tls_autoreload_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/hashicorp/go-hclog"
+)
+
+func TestPollTLSCertificateChanges(t *testing.T) {
+	dir := t.TempDir()
+	cert := filepath.Join(dir, "tls.crt")
+	key := filepath.Join(dir, "tls.key")
+	if err := os.WriteFile(cert, []byte("cert-v1"), 0o600); err != nil {
+		t.Fatal(err)
+	}
```
satoqz: Nit, do you want to use the `testify/require` package for assertions? It's become the standard for new code around here.

changelog/3530.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:feature
+listener: New `tls_auto_reload` option automatically reloads TLS certificate
+and key files when their contents change, without requiring `SIGHUP`.
+```
```
satoqz:
````suggestion
```release-note:improvement
command/server: Add `tls_auto_reload` configuration option to automatically reload TLS certificate and key files when their contents change, without requiring `SIGHUP`.
```
````

The "feature" changelog category is usually reserved for major features (We really need to clarify types and scopes better in CONTRIBUTING.md).

File: internal/helper/configutil/listener.go
Comment (satoqz) [inline internal/helper/configutil/listener.go:1]: Can you document these in `website/content/docs/configuration/listener/tcp.mdx`?

File: internal/command/server.go [hunk 3]
```diff
@@ -784,6 +784,14 @@ func (c *ServerCommand) InitListeners(logger hclog.Logger, config *server.Config
 			relSlice := (*c.reloadFuncs)["listener|"+lnConfig.Type]
 			relSlice = append(relSlice, cg.Reload)
 			(*c.reloadFuncs)["listener|"+lnConfig.Type] = relSlice
+
+			if lnConfig.TLSAutoReload && lnConfig.TLSCertFile != "" {
+				paths := []string{lnConfig.TLSCertFile}
+				if lnConfig.TLSKeyFile != "" {
+					paths = append(paths, lnConfig.TLSKeyFile)
+				}
+				go listenerutil.PollTLSCertificateChanges(paths, lnConfig.TLSAutoReloadInterval, cg.Reload, nil, c.logger)
```
satoqz: Instead of yeeting a goroutine into the void here, perhaps it is cleaner to wrap the listener returned by `tcpListenerFactory` in `internal/command/server/listener_tcp.go` in a custom listener type that starts the goroutine on construction and shuts it down on `Close()`?

internal/helper/configutil/listener.go
```diff
@@ -71,17 +71,21 @@ type Listener struct {
...
+	TLSCipherSuitesRaw               string        `hcl:"tls_cipher_suites"`
+	TLSRequireAndVerifyClientCert    bool          `hcl:"-"`
+	TLSRequireAndVerifyClientCertRaw interface{}   `hcl:"tls_require_and_verify_client_cert"`
+	TLSClientCAFile                  string        `hcl:"tls_client_ca_file"`
+	TLSDisableClientCerts            bool          `hcl:"-"`
+	TLSDisableClientCertsRaw         interface{}   `hcl:"tls_disable_client_certs"`
+	TLSAutoReload                    bool          `hcl:"-"`
+	TLSAutoReloadRaw                 interface{}   `hcl:"tls_auto_reload"`
+	TLSAutoReloadInterval            time.Duration `hcl:"-"`
+	TLSAutoReloadIntervalRaw         interface{}   `hcl:"tls_auto_reload_interval"`
```
tsaarni: It looks like the `Listener` type is shared between server, agent and proxy
* https://openbao.org/docs/agent-and-proxy/proxy/#configuration-file-options
* https://openbao.org/docs/agent-and-proxy/agent/#configuration-file-options

so I suspect the new configuration will now be accepted in these cases as well, but functionality will not work.

internal/helper/listenerutil/tls_autoreload.go
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) HashiCorp, Inc.
```
tsaarni: Since this is new file:
```suggestion
// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
```

internal/helper/listenerutil/tls_autoreload_test.go
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) HashiCorp, Inc.
```
tsaarni:
```suggestion
// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
```

internal/command/server.go
clement-software: That’s a good point, I agree., fixed: https://github.com/openbao/openbao/commit/713436981bb39dac220479e740136446aee09b55

internal/helper/listenerutil/tls_autoreload.go
```diff
@@ -0,0 +1,76 @@
...
+	net.Listener
+	cancel context.CancelFunc
+	done   chan struct{}
+}
+
+func (l *tlsReloadListener) Close() error {
+	l.cancel()
+	<-l.done
+	return l.Listener.Close()
+}
```
satoqz, line 36: Nit: Using context cancellation instead of a simple channel close seems a bit overkill, and thoughts on using a `WaitGroup` to wait for shutdown instead of a channel?

```suggestion
func NewTLSReloadListener(ln net.Listener, paths []string, interval time.Duration, reload func() error, logger hclog.Logger) net.Listener {
	wg := &sync.WaitGroup{}
	stop := make(chan struct{})
	wg.Go(func() {
		pollTLSCertificateChanges(paths, interval, reload, stop, logger)
	})
	return &tlsReloadListener{Listener: ln, wg: wg, stop: stop}
}

type tlsReloadListener struct {
	net.Listener
	wg   *sync.WaitGroup
	stop chan struct{}
}

func (l *tlsReloadListener) Close() error {
	close(l.stop)
	l.wg.Wait()
	return l.Listener.Close()
}
```

internal/helper/configutil/listener.go
satoqz: Good point, @clement-software it should be possible to get consistency if we make the same changes to `internal/command/agentproxyshared/cache/listener.go` that we made to `internal/command/server/listener_tcp.go`.

JanMa:
Thanks for the detailed follow-up @clement-software .

To answer your question about the risks: my main concern is architectural. I personally think we should keep this out of OpenBao and let the environment handle process management. Even with Docker and Podman, there is mature tooling readily available to send a signal to the container if a file changes in a mounted volume. For example, you can easily set up a systemd path service to watch the mount and trigger a `podman kill --signal HUP`.  

Regarding network volumes, it's true that standard file-watching won't work there. But from an infrastructure perspective, I think you'd usually solve that either by pushing the SIGHUP when the cert is updated, or by running a simple external cron job to poll it, rather than building that polling directly into OpenBao.

That being said, it seems that the other maintainers are in favor of adding this feature and I am outvoted here, so I won't block it.

internal/helper/configutil/listener.go
clement-software: Here is my proposal, WDYT @tsaarni ?
https://github.com/openbao/openbao/pull/3530/changes/0b72477bc394ec54cead45dfff8d1a0b16b4cacf

internal/helper/listenerutil/tls_autoreload.go
clement-software: I do actually think your approach using a WaitGroup is more elegant.

satoqz (approved):
Looks good to me. Thanks @clement-software!

cipherboy (approved):

cipherboy:
(I probably would've preferred fsnotify as well, but given that brings another dependency and the code otherwise looks reasonable, we might as well just poll with configurable interval. I otherwise agreed with @JanMa's sentiments as well, having hoped that `SIGHUP` was widely available for use, but seeing the scope of the code itself was fine with it.)

Merging and we can always adjust later!

PR #3532, opened 2026-07-17
Drop sdk/helper/pointerutil entirely

satoqz:
## Description

Clean up final uses of pointerutil (turns out there were none in sdk itself), then drop the package as a small breaking change for v2.7.

## Acknowledgements

       also allowed.


wslabosz-reply (approved):
🥳 🥳 🥳 

pointerutil -> 🪦

phil9909:
don't we want to deprecate it first (and add `//go:fix inline`)?

After all, it is part of the SDK.

satoqz:
@phil9909 we can, but I think that's overkill.

There is only a single use of the package on GitHub that's outside the main openbao repo, and we own it (openbao-k8s):

https://github.com/search?q=sdk%2Fv2%2Fhelper%2Fpointerutil+NOT+is%3Afork&type=code&p=1

> and add //go:fix inline

Not necessary as https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_newexpr exists natively.

phil9909:
> @phil9909 we can, but I think that's overkill.
> 
> There is only a single use of the package on GitHub that's outside the main openbao repo, and we own it (openbao-k8s):
> 
> https://github.com/search?q=sdk%2Fv2%2Fhelper%2Fpointerutil+NOT+is%3Afork&type=code&p=

Not all code is necessarily on GitHub (or is on GitHub but in a private repo).

I'm fine with skipping deprecation anyway.

> > and add //go:fix inline
> 
> Not necessary as https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_newexpr exists natively.

Nice

cipherboy:
> Not all code is necessarily on GitHub (or is on GitHub but in a private repo).

True, but I'd expect the two to be a roughly equivalent reflection. pointerutil never made it to https://github.com/hashicorp/go-secure-stdlib/, so I guess users would just pin to an older SDK version.

cipherboy (approved):

PR #3534, opened 2026-07-17
Add external keys registry, SystemView API

satoqz:
## Description

This PR adds the external key registry, powering new `/sys/external-keys/*` endpoints to manage configs, keys and grants. This is similar to https://github.com/openbao/openbao/pull/1583 which merged to a feature branch a year ago, but re-engineered and built on top of an actual KMS interface this time, so configuration is verified against the actual backend before writing it to storage.

This PR also adds a new API to `SystemView`, `GetExternalKey(ref)`, that (for now, builtin) auth & secrets plugins can use to request a previously configured external key in the form of `kms.Key`.

Documentation will go into a separate PR, a changelog is also best added once we have PKI or Transit integration for it to read like something useful.

## Acknowledgements

       also allowed.


internal/vault/external_keys/registry_test.go
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package ek
```
cipherboy, line 4: eek! 🙈

internal/helper/kmsplugin/catalog.go
```diff
@@ -126,11 +126,14 @@ func (c *Catalog) getClientLocked(name string) (*client, bool, error) {
...
 		AutoMTLS:         true,
-		Logger:           c.logger.Named(name),
 		SecureConfig: &plugin.SecureConfig{
 			Checksum: checksum,
 			Hash:     sha256.New(),
 		},
+		// go-plugin will create a sub-logger with the plugin executable's name,
+		// so avoid duplicating that here and pass the catalog's logger without
+		// modification.
+		Logger: c.logger,
```
cipherboy, line 136: I was debugging external plugins here and phew 😮‍💨 were there a lot of nested logger names.

internal/vault/external_keys/registry.go
```diff
@@ -0,0 +1,619 @@
...
+	cacheLock sync.Mutex // Guards cache.
+
+	// This is a set of locks that guards storage entries and related inflight
+	// cache entries, keyed by namespace + config name.
+	//
+	// While many subsystems in OpenBao use a static set of locks via the
+	// locksutil helper to achieve sharded locking across storage and optionally
+	// cache (e.g., see the policy store), the External Keys registry is
+	// specially constrained because the creation of a cache entry may incur I/O
+	// (via KMS.Open() call).
```
cipherboy: Isn't this still true of policy store and other places? We hold many locks under write and if policy store blocks because storage blocks, we're stuck for a while on all inbound requests. 

IMHO locksutil is probably sufficient until proven otherwise; it defaults to 256 locks.

cipherboy (commented):
Just random drive-by comments.

internal/vault/external_keys/registry.go
satoqz: My assumption here is that storage I/O should be comparatively fast since you tend to have more control over where that is placed on the network while calls to a remote KMS might spend more time traveling over the network.

But I'm still experimenting and will leave it at locksutil, probably.

internal/vault/external_keys/registry_test.go
satoqz: Taking alternative names, but `external_keys` / `externalkeys` is too noisy of a package identifier for my taste.

cipherboy: Nah that's fine, just a joke :laughing: `ekeys`?

internal/vault/external_keys/registry.go
cipherboy: Ah, gotcha. :thinking:

internal/vault/external_keys/registry_test.go
satoqz: I've already been down the rabbit hole of names:

`ekeys` -> electronic keys?
`exkeys` -> ex who?
`extkeys` -> extension keys?
`externkeys` -> too close to the original

internal/vault/external_keys/registry.go
satoqz: I reverted to locksutil but made that package generic so it doesn't force read lock overhead onto you when it's not needed. (I'm surprised and happy the addition of generics is backwards compatible and keeps everything else compiling).

internal/vault/policy/policy_store.go
```diff
@@ -193,11 +193,8 @@ func (ps *Store) lockWithUnlock(ctx context.Context) func() {
 		ns = namespace.RootNamespace
 	}
 
-	lock := locksutil.LockForKey(ps.modifyLocks, ns.UUID)
-
 	ps.logger.Trace("acquiring lock for", "namespace", ns.UUID)
```
cipherboy, line 196: Honestly we probably don't need the log message anymore, but out of scope of this PR.

tools/semgrep/ci/fmt-printf.yml
satoqz, line 1: Semgrep locally complained about paths needing to be absolute in future versions.

satoqz:
Now also split into commits for your (re-)viewing pleasure.

internal/vault/external_keys/registry.go
```diff
@@ -0,0 +1,566 @@
...
+		ke = &KeyEntry{}
+	}
+
+	// Update the key via the provided callback.
+	if err := f(ke, exists); err != nil {
+		return err
+	}
+
+	if verify {
+		// To verify a key, first ensure a client. Unlike ModifyConfig, this
```
wslabosz-reply: `(...) a client exists`?

satoqz: `(...) is present in cache`, yes

wslabosz-reply (dismissed):
reviewing commit by commit is definitely a way to go about it

satoqz:
I wonder if `verify=false` should even exist. It was nice for testing, but I'm not sure I'm convinced there's a use case for it in the wild. So we could remove it to keep things a little simpler. Thoughts?

cipherboy:
> I wonder if `verify=false` should even exist. It was nice for testing, but I'm not sure I'm convinced there's a use case for it in the wild. So we could remove it to keep things a little simpler. Thoughts?

In a past life at a former employer, we wanted to do parallel setup of infrastructure. You're not immediately sure what exists at provisioning time, but by time of use, you sure hope everything is up :laughing:  (based on ordering of calls, it was guaranteed to be). This was for JWT auth, where the Rails service that was the JWT issuer was the one doing the authenticating, so it almost had to be up, +/- network routing problems. Probably more rare here, but I could see it as an option with a cloud KMS?

satoqz:
I guess it once again depends on what `Open()/GetKey()` does in a particular plugin. Will keep for now.

satoqz:
That feeling when you have a use case for `sync.Pool`.

internal/vault/logical_system_external_keys.go
```diff
@@ -0,0 +1,682 @@
...
+					Description: "Verify the key against the KMS backend (default true).",
+				},
+			},
+
+			DisplayAttrs: &framework.DisplayAttributes{
+				OperationPrefix: "external-keys",
+				OperationSuffix: "keys",
+			},
+
+			TakesArbitraryInput: true,
```
cipherboy, line 217: @satoqz Out of scope, but this should force a line in the `path-help` output that says `Takes other path-specific parameters` or something of the sort.

In scope but for a follow-up, we should mention more clearly that additional parameters are required based on plugin type. I wonder if there's some way to document those via path-help API? 🤔  

Probably not, though, but I think in the past we've just linked to the API docs when things get too complicated for path-help.

internal/vault/logical_system_external_keys.go
```diff
@@ -0,0 +1,682 @@
...
+	name := d.Get("config").(string)
+	ce, err := b.Core.externalKeys.ReadConfig(ctx, req.Storage, name)
+	switch {
+	case err != nil:
+		return handleError(err)
+	case ce == nil:
+		return handleError(logical.CodedError(http.StatusNotFound, "config %q not found", name))
+	}
+
+	resp := &logical.Response{Data: ce.Values}
```
cipherboy, line 385:
```
$ bao read sys/external-keys/configs/transit
Key           Value
---           -----
address       http://127.0.0.1:8200
mount_path    transit
plugin        transit
token         s.QmQtYPdEiJBZunBhd7IU3CwD
```

Out of scope for this PR, I think we can address in a follow-up as we'll require round-tripping through g-k-w:

Should KMS plugins especially (and perhaps Wrappers as well) be able to declare sensitive parameters for redaction/eliding versus non-sensitive parameters that are shown in responses/APIs/...?

cipherboy (approved):
Looks good from here, I'll go ahead and merge so we can keep making progress! A few things that we can address later.

internal/vault/logical_system_external_keys.go
satoqz: Hmm, yeah, would be useful.

I'd like to keep the plugin interface as simple as possible, perhaps the cleanest path is to return sensitive field names as part of:

- `wrapping.Wrapper` - `wrapping.WrapperConfig`, returned by `SetConfig()`
- `kms.KMS` - `kms.OpenOptions` (writing _back_ to the opts after a successful `Open()`)
- `kms.Key` - `kms.KeyOptions` (writing _back_ to the opts after a successful `GetKey()`)

As long as `verify=true` at write time, we can then persist the set of redacted options together with key / config entries and redact when reading.

Thoughts?

satoqz: Eh I guess a separate API would still be better, just so you can fetch it from the plugin once at startup and be good with it, verify=true or not.

cipherboy: Yeah, my thinking is plugins probably _shouldn't_ be doing fully dynamic config parameters as that gets rather confusing to an operator. I'd probably just have providers implement additional fields:

```go
type KMS interface {
  GetSensitiveOpenOptions() []string
  GetSensitiveKeyOptions() []string
}
```

&c and you don't need `verify=true` and can call this once and cache it. Really easy for plugins to implement too, return `nil` if you don't have any.

The issue with pointers as you probably also realized is that you've gotta propagate any changes through multiple layers (GRPC?), which gets nasty.

satoqz: @cipherboy I think we can keep it out of `KMS` entirely to avoid polluting that, but agreed! Already have something in mind :)

PR #3535, opened 2026-07-17
Import hclog in plugin example

DrDaveD:
## Description

This adds a missing import in the plugin development example.

## Rationale

`hclog` is used but not defined in the example.  This is the import used for example in the [aws auth plugin](https://github.com/openbao/openbao-plugins/blob/auth-aws-v0.1.1/auth/aws/cmd/main.go#L9).

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/plugins/plugin-development.mdx
+++ b/website/content/docs/plugins/plugin-development.mdx
@@ -63,6 +63,7 @@ package main
 import (
 	"os"
 
+	hclog "github.com/hashicorp/go-hclog"
 	myPlugin "your/plugin/import/path"
 	"github.com/openbao/openbao/api/v2"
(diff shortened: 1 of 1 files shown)


website/content/docs/plugins/plugin-development.mdx
satoqz, line 114: Related: Was this ever even the case with OpenBao and not a pre-fork thing? 🤔 If so, we can really simplify the main function above quite a bit.

satoqz (approved):

website/content/docs/plugins/plugin-development.mdx
cipherboy: @satoqz I think this is rather old, we can probably drop it.

cipherboy (approved):

PR #3542, opened 2026-07-20
Add tests for GRPC invalidation with PostgreSQL

cipherboy:
Based on #3448; will be rebased once that lands.

## Description

This adds a new clustered PostgreSQL storage helper, allowing custom
mappings of OpenBao nodes to PostgreSQL nodes. We add a new OpenBao
test for ensuring horizontal scalability works on all nodes and
ensures it works with a replicated PostgreSQL backend. When the primary
fails, the nodes attached to it should quit working and the nodes
attached to the replicas should become writers.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/external_tests/postgresql_binary/postgresql_test.go
+++ b/internal/vault/external_tests/postgresql_binary/postgresql_test.go
@@ -337,25 +337,22 @@ func TestPostgreSQL_Scalability(t *testing.T) {
 	pLogger := logger.Named("postgresql-cluster")
 	cLogger := logger.Named("openbao-cluster")
 
-	var returned atomic.Int32
-	var nodesOnPrimary int32 = 2
-
-	mapper := func(ctx context.Context, cluster *thpsql.Cluster) (string, error) {
-		index := returned.Add(1)
-		if index <= nodesOnPrimary {
-			return cluster.Primary.InternalURL(ctx)
+	nodesOnPrimary := 2
+	mapper := func(ctx context.Context, cluster *thpsql.Cluster, index int) (*thpsql.Node, error) {
+		if index < nodesOnPrimary {
+			return cluster.Primary, nil
 		}
 
-		if index == nodesOnPrimary+1 {
+		if index == nodesOnPrimary {
 			node, err := cluster.AddNode(ctx)
 			if err != nil {
-				return "", fmt.Errorf("failed to add replica: %w", err)
+				return nil, fmt.Errorf("failed to add replica: %w", err)
 			}
 
-			return node.InternalURL(ctx)
+			return node, nil
 		}
 
-		return cluster.Nodes[1].InternalURL(ctx)
+		return cluster.Nodes[1], nil
 	}
 
 	pCluster, err := docker.NewPostgreSQLClusterStorage(t.Context(), pLogger, "scalability", "", mapper)
--- a/internal/vault/external_tests/postgresql_binary/postgresql_test.go
+++ b/internal/vault/external_tests/postgresql_binary/postgresql_test.go
@@ -10,12 +10,15 @@ import (
 	"time"
 
 	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-uuid"
 	"github.com/openbao/openbao/api/v2"
 	"github.com/openbao/openbao/sdk/v2/helper/consts"
 	"github.com/openbao/openbao/sdk/v2/helper/logging"
 	"github.com/openbao/openbao/sdk/v2/helper/testcluster"
 	"github.com/openbao/openbao/sdk/v2/helper/testcluster/docker"
+	thpsql "github.com/openbao/openbao/sdk/v2/helper/testhelpers/postgresql"
 	"github.com/openbao/openbao/sdk/v2/physical"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -137,8 +140,6 @@ func TestPostgreSQL_FencedWrites(t *testing.T) {
 }
 
 func TestPostgreSQL_ParallelInit(t *testing.T) {
-	t.Parallel()
-
 	binary := api.ReadBaoVariable("BAO_BINARY")
 	if binary == "" {
 		t.Skip("missing $BAO_BINARY")
@@ -204,8 +205,6 @@ func TestPostgreSQL_ParallelInit(t *testing.T) {
 }
 
 func TestPostgreSQL_FatalInit(t *testing.T) {
-	t.Parallel()
-
 	binary := api.ReadBaoVariable("BAO_BINARY")
 	if binary == "" {
 		t.Skip("missing $BAO_BINARY")
@@ -259,8 +258,6 @@ func TestPostgreSQL_FatalInit(t *testing.T) {
 }
 
 func TestPostgreSQL_Upgrade(t *testing.T) {
-	t.Parallel()
-
 	binary := api.ReadBaoVariable("BAO_BINARY")
 	if binary == "" {
 		t.Skip("missing $BAO_BINARY")
@@ -297,10 +294,12 @@ func TestPostgreSQL_Upgrade(t *testing.T) {
 	})
 	require.NoError(t, err, "failed to mount kv")
 
-	_, err = client.KVv2("kv").Put(t.Context(), "a/key", map[string]any{
-		"value": "known-value",
-	})
-	require.NoError(t, err, "failed writing k/v key")
+	require.EventuallyWithT(t, func(collect *assert.CollectT) {
+		_, err = client.KVv2("kv").Put(t.Context(), "a/key", map[string]any{
+			"value": "known-value",
+		})
+		require.NoError(collect, err, "failed writing k/v key")
+	}, 10*time.Second, 100*time.Millisecond)
 
 	// Now upgrade the nodes one at a time.
 	opts.VaultBinary = binary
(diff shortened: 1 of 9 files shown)


sdk/helper/testcluster/docker/environment.go
```diff
@@ -1256,6 +1256,7 @@ func (dc *DockerCluster) setupImage(ctx context.Context, opts *DockerClusterOpti
 	containerFile := fmt.Sprintf(`
 FROM %s:%s
 COPY bao /bin/bao
+COPY bao /usr/bin/bao
```
cipherboy, line 1312: @satoqz This is more fallout from the Alpine changes fwiw.

satoqz: Why is this one necessary?

Edit: Ah, `/bin` is last in `$PATH`. I suppose you could symlink one of them to save some space, that is if we really want to be compatible with past releases anyway.

internal/vault/ha.go
cipherboy, line 1: Happy to move this to another PR if we prefer, but this is required for the tests to pass.

internal/vault/external_tests/postgresql_binary/postgresql_test.go
cipherboy, line 1: As far as I can tell, what happens more often as we add more tests here is that we exhaust the number of ~ports~ ip addresses available in Docker. Its dependent on how much is running. So I've pulled a lot of the `t.Parallel()` statements out of this and perhaps we should elsewhere as well.

sdk/helper/docker/testhelpers.go
```diff
@@ -522,6 +527,23 @@ func (d *Runner) Start(ctx context.Context, addSuffix, forceLocalAddr bool) (*St
...
 
+	for destination, contents := range d.RunOptions.WriteInto {
+		// Convert our provided contents to a tarball to ship up.
+		tar, err := contents.ToTarball()
+		if err != nil {
+			return nil, fmt.Errorf("failed to build contents for directory %q into tarball: %w", destination, err)
+		}
+
+		// ,_ err := d.DockerAPI.CopyToContainer(context.Background(), c, destination, tar, cfg)
+		_, err = d.DockerAPI.CopyToContainer(context.Background(), c.ID, client.CopyToContainerOptions{
```
satoqz:
```suggestion
		_, err = d.DockerAPI.CopyToContainer(ctx, c.ID, client.CopyToContainerOptions{
```

internal/vault/external_tests/postgresql_binary/postgresql_test.go
```diff
@@ -327,3 +326,228 @@ func TestPostgreSQL_Upgrade(t *testing.T) {
...
+			node, err := cluster.AddNode(ctx)
+			if err != nil {
+				return "", fmt.Errorf("failed to add replica: %w", err)
+			}
+
+			return node.InternalURL(ctx)
+		}
+
+		return cluster.Nodes[1].InternalURL(ctx)
+	}
```
satoqz, line 356: I'm not yet convinced that this warrants a change in the `ClusterStorage` interface? Does this need to happen exactly when opts is called?

internal/vault/external_tests/postgresql_binary/postgresql_test.go
satoqz: > exhaust the number of ports available in Docker

what does this mean?

internal/vault/ha.go
```diff
@@ -610,14 +610,41 @@ func (c *Core) waitForLeadership(manualStepDownCh, stopCh <-chan struct{}, isRea
...
+					select {
+					case <-stopCh:
+						timer.Stop()
+						return
+					case <-failedLeaderAcquisition:
+						return
+					case <-acquiredLeadership:
+						return
+					case <-timer.C:
+					}
```
satoqz: I think this can just use `time.After`?

internal/vault/ha.go
satoqz: Can you explain what the problem was?

internal/vault/external_tests/postgresql_binary/postgresql_test.go
cipherboy: @satoqz Ah, sorry, meant addresses. There's a network that Docker attaches containers to by default (and we rarely create new networks), which means that if we run > 128-ish containers simultaneously, we can run into issues. The new test here for instance runs 6, some of the ACME tests run multiple, many use three+ node clusters, upgrade uses double I think (I think IPs are reserved for stopped container?), and I've also personally seen issues with local ports being reused for different containers/hosts than what they were originally for, while the first test was still running...? Which seems scary. 

We probably need to get a better strategy for parallelizing container tests? Separate networks might be a good start.

internal/vault/ha.go
cipherboy: @satoqz Standby nodes would never take active status under GRPC invalidation, since read-only standby would never be able to connect to the (failed) active and thus would bail on restart.

internal/vault/external_tests/postgresql_binary/postgresql_test.go
satoqz: Gotcha, yeah, separate networks is probably the move, they're cheap to create anyhow.

Also, treat this line of the comment as the obligatory ipv6 mention.

internal/vault/external_tests/postgresql_binary/postgresql_test.go
cipherboy: @satoqz Outside of a priori deciding a mapping of fixed node-count PSQL cluster -> OpenBao cluster node, I'm not sure how else you'd do this?

By taking a function mapping (at OpenBao node setup time) to a specific URL, we get the ability to fully customize both clusters relative to each other. We could spin up a new PSQL node, wait for it to stabilize, &c, just before adding a new OpenBao node. We could provision two, add the first to the OB node, and then use the second next time. We could also remove a node right as you start up and see things fail. &c.

We could've written against the existing interface, but you'd lose that point of control and have to provision the node mapping up-front. There's no easy identifiers, so I think it'd be `[]string`, where each is a connection URL for OpenBao?

The type change was:

1. Context, which is helpful if you're doing any sort of dynamic evaluation. Technically you don't need that and could pass in `t.Context()` or any other available context at initialization time, but its kind to use an existing context.
2. Return also an error, which, could be handled by `panic()` or `t.Fatalf` or similar, but is also kind to return.

We never actually did any checking on `storageOpts = opts.Storage.Opts()`, so... having `error` is probably kind thing to do anyways, rather than waiting for entire node to stand up only to exit early from the container. 

My 2c?

internal/vault/external_tests/postgresql_binary/postgresql_test.go
cipherboy: @satoqz Yeah, but we currently have a massive resource leak with volumes, we should figure out how to fix that before we go adding another resource leak (networks) :P 

```
$ podman volume ls  | wc -l
845
```

(I deleted all volumes on Friday morning)

internal/vault/external_tests/postgresql_binary/postgresql_test.go
satoqz: @cipherboy it feels like we're misusing the interface here, to the point where you needed to modify it. The interface reads to me as if dynamically returning values from `Type()` and `Opts()` was never intended, and that an instance of it should represent exactly one "storage node".

<details>
<summary>
How about something like this? We make `PostgreSQLClusterStorage` a factory of `ClusterStorage` instead, and call `AddNode` on either of our OpenBao or PostgreSQL cluster whenever we like.

</summary>

```diff
diff --git a/internal/vault/external_tests/postgresql_binary/postgresql_test.go b/internal/vault/external_tests/postgresql_binary/postgresql_test.go
index 4a8d25b9f9..a6a976b33f 100644
--- a/internal/vault/external_tests/postgresql_binary/postgresql_test.go
+++ b/internal/vault/external_tests/postgresql_binary/postgresql_test.go
@@ -16,7 +16,6 @@ import (
 	"github.com/openbao/openbao/sdk/v2/helper/logging"
 	"github.com/openbao/openbao/sdk/v2/helper/testcluster"
 	"github.com/openbao/openbao/sdk/v2/helper/testcluster/docker"
-	thpsql "github.com/openbao/openbao/sdk/v2/helper/testhelpers/postgresql"
 	"github.com/openbao/openbao/sdk/v2/physical"
 
 	"github.com/stretchr/testify/assert"
@@ -337,36 +336,18 @@ func TestPostgreSQL_Scalability(t *testing.T) {
 	pLogger := logger.Named("postgresql-cluster")
 	cLogger := logger.Named("openbao-cluster")
 
-	var returned atomic.Int32
-	var nodesOnPrimary int32 = 2
+	ctx := t.Context()
 
-	mapper := func(ctx context.Context, cluster *thpsql.Cluster) (string, error) {
-		index := returned.Add(1)
-		if index <= nodesOnPrimary {
-			return cluster.Primary.InternalURL(ctx)
-		}
-
-		if index == nodesOnPrimary+1 {
-			node, err := cluster.AddNode(ctx)
-			if err != nil {
-				return "", fmt.Errorf("failed to add replica: %w", err)
-			}
-
-			return node.InternalURL(ctx)
-		}
-
-		return cluster.Nodes[1].InternalURL(ctx)
-	}
-
-	pCluster, err := docker.NewPostgreSQLClusterStorage(t.Context(), pLogger, "scalability", "", mapper)
+	pCluster, err := docker.NewPostgreSQLClusterStorage(ctx, pLogger, "scalability", "")
 	require.NoError(t, err)
 
-	defer func() { require.NoError(t, pCluster.Cleanup()) }()
+	defer pCluster.Cleanup()
 
+	// Start out with two nodes connected to the primary postgres instance.
 	opts := &docker.DockerClusterOptions{
 		ImageRepo:   "quay.io/openbao/openbao",
 		ImageTag:    "latest",
-		Storage:     pCluster,
+		Storage:     pCluster.Primary(),
 		VaultBinary: binary,
 		ClusterOptions: testcluster.ClusterOptions{
 			VaultNodeConfig: &testcluster.VaultNodeConfig{
@@ -376,7 +357,7 @@ func TestPostgreSQL_Scalability(t *testing.T) {
 				DisableStandbyReads: false,
 			},
 			ClusterName: "psql-upgrade",
-			NumCores:    3,
+			NumCores:    2,
 			Logger:      cLogger,
 		},
 	}
@@ -384,6 +365,13 @@ func TestPostgreSQL_Scalability(t *testing.T) {
 	cluster := docker.NewTestDockerCluster(t, opts)
 	defer cluster.Cleanup()
 
+	// Add a replica node and join a third OpenBao node pointed at it.
+	replica, err := pCluster.AddNode(ctx)
+	require.NoError(t, err)
+
+	opts.Storage = replica
+	require.NoError(t, cluster.AddNode(ctx, opts))
+
 	nodes := cluster.Nodes()
 	client := nodes[0].APIClient()
 
diff --git a/internal/vault/ha.go b/internal/vault/ha.go
index cf1bc60484..e8f0606431 100644
--- a/internal/vault/ha.go
+++ b/internal/vault/ha.go
@@ -629,7 +629,6 @@ func (c *Core) waitForLeadership(manualStepDownCh, stopCh <-chan struct{}, isRea
 					// If we've been told to stop, give up.
 					select {
 					case <-stopCh:
-						timer.Stop()
 						return
 					case <-failedLeaderAcquisition:
 						return
diff --git a/sdk/helper/testcluster/docker/environment.go b/sdk/helper/testcluster/docker/environment.go
index 171e8f69e3..9c73272bb7 100644
--- a/sdk/helper/testcluster/docker/environment.go
+++ b/sdk/helper/testcluster/docker/environment.go
@@ -643,12 +643,8 @@ func (n *DockerClusterNode) Start(ctx context.Context, opts *DockerClusterOption
 	}
 
 	if opts.Storage != nil {
-		var err error
 		storageType = opts.Storage.Type()
-		storageOpts, err = opts.Storage.Opts(ctx)
-		if err != nil {
-			return err
-		}
+		storageOpts = opts.Storage.Opts()
 	}
 
 	if opts != nil && opts.VaultNodeConfig != nil {
diff --git a/sdk/helper/testcluster/docker/storage.go b/sdk/helper/testcluster/docker/storage.go
index d9eda569cc..2fe344efd3 100644
--- a/sdk/helper/testcluster/docker/storage.go
+++ b/sdk/helper/testcluster/docker/storage.go
@@ -27,10 +27,8 @@ type InmemStorage struct{}
 
 func (InmemStorage) Start(context.Context, *testcluster.ClusterOptions) error { return nil }
 func (InmemStorage) Cleanup() error                                           { return nil }
-func (InmemStorage) Opts(_ context.Context) (map[string]any, error) {
-	return make(map[string]any), nil
-}
-func (InmemStorage) Type() string { return "inmem" }
+func (InmemStorage) Opts() map[string]any                                     { return make(map[string]any) }
+func (InmemStorage) Type() string                                             { return "inmem" }
 
 var _ testcluster.ClusterStorage = InmemStorage{}
 
@@ -99,14 +97,14 @@ func (p *PostgreSQLStorage) Cleanup() error {
 	return nil
 }
 
-func (p *PostgreSQLStorage) Opts(_ context.Context) (map[string]any, error) {
+func (p *PostgreSQLStorage) Opts() map[string]any {
 	return map[string]any{
 		"connection_url":       p.InternalUrl,
 		"ha_enabled":           true,
 		"max_parallel":         5,
 		"max_idle_connections": 3,
 		"max_connect_retries":  30,
-	}, nil
+	}
 }
 
 func (p *PostgreSQLStorage) Type() string {
@@ -127,13 +125,11 @@ func (p *PostgreSQLStorage) Client(ctx context.Context) (*sql.DB, error) {
 }
 
 type PostgreSQLClusterStorage struct {
-	Cluster *thpsql.Cluster
-	mapper  PostgreSQLClusterMapper
+	Cluster    *thpsql.Cluster
+	PrimaryURL string
 }
 
-type PostgreSQLClusterMapper func(ctx context.Context, cluster *thpsql.Cluster) (string, error)
-
-func NewPostgreSQLClusterStorage(ctx context.Context, logger log.Logger, name string, network string, mapper PostgreSQLClusterMapper) (*PostgreSQLClusterStorage, error) {
+func NewPostgreSQLClusterStorage(ctx context.Context, logger log.Logger, name, network string) (*PostgreSQLClusterStorage, error) {
 	cfg := thpsql.DefaultClusterConfig(name)
 	if logger != nil {
 		cfg.Logger = logger
@@ -144,37 +140,39 @@ func NewPostgreSQLClusterStorage(ctx context.Context, logger log.Logger, name st
 		return nil, fmt.Errorf("unable to set up cluster: %w", err)
 	}
 
+	url, err := cluster.Primary.InternalURL(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get primary's url: %w", err)
+	}
+
 	return &PostgreSQLClusterStorage{
-		Cluster: cluster,
-		mapper:  mapper,
+		Cluster:    cluster,
+		PrimaryURL: url,
 	}, nil
 }
 
-func (p *PostgreSQLClusterStorage) Start(context.Context, *testcluster.ClusterOptions) error {
-	// Initialization already occurred when creating this object.
-	return nil
+func (p *PostgreSQLClusterStorage) Cleanup() {
+	p.Cluster.Cleanup()
 }
 
-func (p *PostgreSQLClusterStorage) Cleanup() error {
-	p.Cluster.Cleanup()
-	return nil
+func (p *PostgreSQLClusterStorage) Primary() testcluster.ClusterStorage {
+	return &PostgreSQLStorage{
+		ExternalUrl: p.PrimaryURL,
+	}
 }
 
-func (p *PostgreSQLClusterStorage) Opts(ctx context.Context) (map[string]any, error) {
-	url, err := p.mapper(ctx, p.Cluster)
+func (p *PostgreSQLClusterStorage) AddNode(ctx context.Context) (testcluster.ClusterStorage, error) {
+	node, err := p.Cluster.AddNode(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("failed mapping to connection url: %w", err)
+		return nil, fmt.Errorf("failed to add replica: %w", err)
 	}
 
-	return map[string]any{
-		"connection_url":       url,
-		"ha_enabled":           true,
-		"max_parallel":         5,
-		"max_idle_connections": 3,
-		"max_connect_retries":  30,
-	}, nil
-}
+	url, err := node.InternalURL(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get replica's url: %w", err)
+	}
 
-func (p *PostgreSQLClusterStorage) Type() string {
-	return "postgresql"
+	return &PostgreSQLStorage{
+		ExternalUrl: url,
+	}, nil
 }
diff --git a/sdk/helper/testcluster/types.go b/sdk/helper/testcluster/types.go
index ac17465cd6..5c8e24bc8e 100644
--- a/sdk/helper/testcluster/types.go
+++ b/sdk/helper/testcluster/types.go
@@ -123,6 +123,6 @@ type CA struct {
 type ClusterStorage interface {
 	Start(context.Context, *ClusterOptions) error
 	Cleanup() error
-	Opts(ctx context.Context) (map[string]any, error)
+	Opts() map[string]any
 	Type() string
 }
```

</details>

internal/vault/external_tests/postgresql_binary/postgresql_test.go
satoqz: @cipherboy for starters, we should consider removing `VOLUME` instructions from OpenBao's Dockerfile. It's a weird legacy instruction that contributes to nothing but tons of leftover anonymous volumes.

satoqz: Also see https://github.com/moby/moby/issues/43190.

internal/vault/external_tests/postgresql_binary/postgresql_test.go
cipherboy: @satoqz That's fine I suppose, though I would've expected that `ClusterOptions` to actually have a factory pattern in the first place if it was expected that a `ClusterStorage` instance was static, given it takes `NumCores` and internally re-creates with the same opts. Given even the Raft backend has a `node_id` parameter, expecting a fully static `Opts()` across all cores seems ill-advised...

I'd consider that given this is all in `testcluster`, we have the option to improve the interface rather than resign ourselves to fitting within it. 

Personally, I'd change `Storage` so that:

```go
type ClusterStorage interface {
    ForNode(ctx context.Context, index int) (NodeStorage, error)
    Cleanup() error
}

type NodeStorage interface {
    Start(context.Context, *ClusterOptions) error
    Cleanup() error
    Opts() map[string]any
    Type() string
}
```

and then we'd get a native factory pattern that works with `NumCores`, each node's storage can be started on-demand, and it gives us an easy place for context/error in the `Start()` function.

Then callers can still choose calling patterns here.

Thoughts?

satoqz: @cipherboy that looks good. Supposing `NodeStorage` is just a renamed `ClusterStorage` and we take either one inside of `DockerClusterOptions` (via existing `Storage` and a new `ClusterStorage` field), it's not too breaking either?

cipherboy:
@satoqz :laughing: 

```
  internal/vault/ha.go 
     tools.semgrep.ci.loop-time-after
        <-time.After() used in for loop, consider using a ticker or a timer instead

        586┆ for {
        587┆ 	// Cancel any old context from the previous iteration.
        588┆ 	ctxCancel()
        589┆ 
        590┆ 	// Check for a shutdown
        591┆ 	select {
        592┆ 	case <-stopCh:
        593┆ 		c.logger.Debug("stop channel triggered in runHALoop")
        594┆ 		return
        595┆ 	default:
           [hid 324 additional lines, adjust with --max-lines-per-finding] 
```

I suppose you could argue that we'd leak goroutines in these two scenarios:

```
        // Create a lock
        uuid, err := uuid.GenerateUUID()
        if err != nil {
            c.logger.Error("failed to generate uuid", "error", err)
            continue
        }
        lock, err := c.ha.LockWith(CoreLockPath, uuid)
        if err != nil {
            c.logger.Error("failed to create lock", "error", err)
            continue
        }
```

UUID generation should never fail, and our _current_ implementations of `LockWith(...)` also don't fail (assuming UUID generation does not fail). But simply swapping over to a timer doesn't fix the problem.

internal/vault/external_tests/postgresql_binary/postgresql_test.go
cipherboy: @satoqz Yeah, other than a rename, `Storage` can take `any` and we'll dispatch the right one.

satoqz: @cipherboy I'd prefer two fields, is clearer and removes the need for implementers to add the `var _ = ...` assertion.

cipherboy: @satoqz Sorry, I missed this comment. I introduced a new `testcluster.Storage` type. I think if we have two fields, we risk breaking compatibility and people already see `Storage` and think (the new, correct) `ClusterStorage` and not `Node`-specific storage.

Let me know if you're happy enough with this to merge, otherwise can revise.

sdk/helper/testcluster/types.go
```diff
@@ -120,9 +120,30 @@ type CA struct {
 	CAKeyPEM      []byte
 }
 
+type Storage interface {
```
satoqz, line 128: doc comment pls

internal/vault/external_tests/postgresql_binary/postgresql_test.go
satoqz: sure, works for me.

sdk/helper/testcluster/types.go
cipherboy: Added.

satoqz (approved):

cipherboy:
Bypassing linter checks; these were due to the refactor within `ha.go` and original to `main`.

Issue #3541, opened 2026-07-20
logger.Named() + AddLoger(...) as a package


cipherboy:
This should just be a single pattern tbh, but out of scope for this thing now.

_Originally posted by @cipherboy in https://github.com/openbao/openbao/pull/3448#discussion_r3616794099_

PR #3543, opened 2026-07-21
Add Core.namedLogger to unify the Named()+AddLogger idiom

youdie006:
## What

Adds a small `(*Core).namedLogger(name string) log.Logger` helper that names a sub-logger off `baseLogger` and registers it with the Core, and migrates the call sites that use this exact idiom:

```go
logger := c.baseLogger.Named(name)
c.AddLogger(logger)
```

becomes

```go
logger := c.namedLogger(name)
```

## Why

Per @cipherboy in https://github.com/openbao/openbao/pull/3448#discussion_r3616794099 ("This should just be a single pattern tbh"), tracked in #3541. The `Named(...)` + `AddLogger(...)` two-step was duplicated across the vault package; every sub-logger has to remember to register itself so its level tracks dynamic log-level changes.

## Scope / behavior

This PR migrates only the call sites that derive their sub-logger from `c.baseLogger` (audit, auth, expiration, mount, namespace_store, policy, rollback, seal_manager, workflow_store). **Behavior is unchanged** — the sub-loggers keep the exact same names and are still registered. `go build ./internal/vault/...` and `go vet ./internal/vault/` pass; `gofmt` clean.

A couple of related sites are intentionally left out because folding them in would change behavior, and I'd rather get your steer first:

- A few sub-loggers are named off `c.logger` (the `core`-named logger) rather than `c.baseLogger` — e.g. `raft` and `cluster-listener.tcp`. Switching those to `namedLogger` would drop the `core.` prefix from their log names (`core.raft` -> `raft`). If that normalization is the intended single pattern, I'm happy to include them.
- A few sites name off a local `logger` passed into a setup helper (e.g. token/system/identity in `core.go`, `mfa` in `login_mfa.go`). Those can also be routed through `namedLogger` if you'd like, but I kept this PR to the unambiguous, behavior-preserving cases.

`internal/vault/identity` has its own `AddLogger` on a different type, so it's out of scope here.

Signed-off-by is included (DCO).

---

*AI disclosure: this PR was drafted with Claude Code (AI-assisted). I designed the helper, chose the `baseLogger`-derived scope to keep log names identical, and verified with `go build ./internal/vault/...` and `go vet ./internal/vault/` (both pass) plus `gofmt`.*

### DIFF
--- a/internal/vault/audit.go
+++ b/internal/vault/audit.go
@@ -391,8 +391,7 @@ func (c *Core) persistAudit(ctx context.Context, table *routing.MountTable, loca
 // setupAudit is invoked after we've loaded the audit able to
 // initialize the audit backends
 func (c *Core) setupAudits(ctx context.Context) error {
-	brokerLogger := c.baseLogger.Named("audit")
-	c.AddLogger(brokerLogger)
+	brokerLogger := c.namedLogger("audit")
 	c.auditBroker = NewAuditBroker(brokerLogger)
 
 	err := c.reconcileAudits(reconcileAuditsRequests{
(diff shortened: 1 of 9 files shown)


satoqz:
Agents smart enough to read `pull_request_template.md` when?

PR #3547, opened 2026-07-21
RFC on External Authorization

### LINKED ISSUE #3545: Feature Request: Support for External Authorization Checks
For maintaining authorizations, one must create OpenBao policies and assign them to users.
However, if you are running an dedicated authorization service like 
[Open Policy Agent](https://www.openpolicyagent.org/) or [OpenFGA](https://openfga.dev/) or some commercial service,
then you want to use this  authorization service for OpenBao, too.
Because you want to administrate the authorizations of all services in one central place.

We should create a plugin interface which allows connecting different authorization services
without creating hard dependencies.  The solution should integrate with OpenBao policies (and
not replace them), because OpenBao policies offer special functions like 
[deny](https://openbao.org/docs/concepts/policies/#capabilities) or [parameter constraints](https://openbao.org/docs/concepts/policies/#parameter-constraints), which are not common to external authorization services. 
In addition, it should be possible to continue with existing OpenBao policies in case of an upgrade.

As alternative we have considered a global switch to replace OpenBao policies, but this approach would
be less flexible.

klaus-sap:
## Description

Currently, authorizations in OpenBao must be defined as [policies](https://openbao.org/docs/concepts/policies/), which are maintained within OpenBao. However, customers may have central authorization services in place like [Open Policy Agent](https://www.openpolicyagent.org/) or [OpenFGA](https://openfga.dev/), which they may want to use for OpenBao, too. In this RFC, we propose a [plugin](https://openbao.org/docs/plugins/) interface which allows the integration of OpenBao with external authorization services.

## Rationale

Resolves: https://github.com/openbao/openbao/issues/3545

## Acknowledgements

       also allowed.


klaus-sap:
Thanks to @Huy-Dinh-Reply for a lot of valuable input

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+  the authorization checks internal and maintain the permissions of end users in the external authorization service.
+
+## User-facing Description
+
+With the following policies we may configure the examples from above.
+
+- “externalPolicy.hcl”
+  ```
+  # Call external authorization service for all paths 
+  path "/*" {
```
cipherboy:
```suggestion
  path "*" { 
```

I believe ACL engine strips leading slashes, so we should do the same for our examples.

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+  ```
+  # Call external authorization service for all paths 
+  path "/*" { 
+      capabilities = ["ask”]
+  }
+  ```
+- "mixedPolicy.hcl"
+  ```
+  # Path /system shall be forbidden independent from external authorizations
+  path "/system/*" {
```
cipherboy:
```suggestion
  path "/sys/*" { 
```

otherwise, if this is for example, I might suggest we call it `internal` or `private` or something.

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+  ```
+
+## Technical Description
+
+The plugin API consists of four methods only:
+
+Method                                                      | Description
+----------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------
+`hasAuthorization(ctx, path, entity, operation) bool`       | When processing "ask" capabilities, call plugin to check authorizations for one single request.
+`listChildren(ctx, parentPath, entity, operation) []string` | When showing list views on UIs, call plugin to return list of children (paths), which the user has authorizations for.
```
cipherboy: This might be entertaining to implement; it assumes the EPA (external policy agent?) a priori knows all paths that a plugin might have and we'd take the intersection of paths between the actual list output and this list. (as we'd always have to vet that these paths do actually exist). 

Things like globs or more complex patterns (regex) and recursive scan support would make this messy.

Wonder what alternatives we have here?

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+`updatePath(ctx, path, operation) bool`                     | (Optional) After write operations to OpenBao, call plugin to update the DB of the external authorization service when a path changes, i.e. an entry gets created or deleted
+`syncSubtree(ctx, parentPath, subtree) []bool`              | (Optional) At initialization time, call plugin to sync the paths of an OpenBao subtree to the DB of the external authorization service.
+
+The updatePath and syncSubtree methods are optional, i.e. don't need to be implemented for all external authorization services.
+For example, some authorization services like [OpenFGA](https://openfga.dev/docs/getting-started/update-tuples) have their own database containing object relations, 
+which they use for authorization checks. This makes it important to sync the paths from OpenBao to the external authorization service.
+
+As user identities, we should pass [entities](https://openbao.org/docs/concepts/identity/) to the plugin.
```
cipherboy, line 87: We probably want a fat `logical.Auth` object here; it is already proto marshable and handles entities, groups, and non-entity backed auth.

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+
+In the ACL processing of OpenBao, we must process the „ask“ capability and call the plugin method `hasAuthorization` from:
+[acl.AllowOperation(ctx, req, capCheckOnly) *ACLResults](https://github.com/openbao/openbao/blob/main/internal/vault/policy/acl.go#L344).
+We may add `entity` to the `ACL` struct to be able the pass the user identity to the plugin.
+In addition, we may add attributes like "external user id" or "primary email address" to the
```
cipherboy, line 92: `logical.Auth.DisplayName` handles some of this: 

https://github.com/openbao/openbao/tree/main/sdk/logical/auth.go

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+For example, if the user has logged on with OIDC, we may fill these attributes from token claims. With these 
+attributes we also might be able to automatically merge [aliases](https://openbao.org/docs/concepts/identity/) 
+into one common `entity` (depending on some configuration, i.e. opt-in).
+
+After calling the plugin, we fill the ACL results in the following way:
+
+```
+type ACLResults struct {
+	Allowed    bool // if ask: Allowed = plugin.hasAuthorization(…, read/write/…)
+	RootPrivs  bool // if ask: RootPrivs = plugin.hasAuthorization(…, sudo)
```
cipherboy, line 104: This effectively means we need two external plugin calls per operation. I think we should batch these a bit better and see if we can condense:

```
hasAuthorization(..., []{read, sudo})`
```

Similarly, OpenBao doesn't (at an external API interface level) differentiate create/update, but does from an existence check/authorization PoV, so we'd often do:

```
hasAuthorization(..., []{create, update, sudo})`
```

imo.

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+  They must provide the plugin configuration including the path to the executable and credentials for the external authorization service.
+  They must ensure that unique user identifiers are passed whenever the plugin is invoked.
+
+- Authorization administrators must create OpenBao policies as described above. Whenever they want to forward authorization checks 
+  to the external authorization service, they must add the "ask" capability to the policy.
+
+- User administrators must assign OpenBao policies to users (OpenBao entities). In the external authorization service, they must assign 
+  OpenBao-related authorizations to the unique user identifiers mentioned above.
+
+## Related Issue
```
cipherboy, line 153: General questions:

- If/when we implement CEL as a policy language, I wonder how these would interact? Would CEL just implement this interface? Or would we want to keep with CEL builtin like it has been elsewhere? 
- How does plugin registration work? Like auth/secret plugins and written into the database or like KMS where its config-only? 
- Today I can write a policy at `ns1` with grant on a child namespace (`ns1/ns2` -- referred relatively as `path "ns2" { ... }` in the policy document). In your proposal, it suggests potentially per-namespace poilcy plugins; how would the two interact? In a policy like `path "ns2/secrets" { capabilities = ["ask"] }`, which policy plugin would execute? `ns1`'s (since that's where this policy document lives?) or `ns1/ns2` (since that's what the path ultimately refers to?

cipherboy (commented):
It is an interesting idea; I'm not quite sure I had thought about this shape of auth plugin before, I'll have to mull on it some more.

website/content/community/rfcs/external-authorization.mdx
klaus-sap: fixed

website/content/community/rfcs/external-authorization.mdx
klaus-sap: fixed

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Good point. Does a plugin know all children? Then the plugin may pass a list of children to the external authorization service, i.e. we might change the signature of hasAuthorization to allow batch processing

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Not sure: is DisplayName a unique identifier? According to the comment, DisplayName is a non-security sensitive identifier

website/content/community/rfcs/external-authorization.mdx
cipherboy: @klaus-sap Yes, plugins are ultimately responsible for creating LIST responses, so it'd be the one with that list.

JanMa:
Interesting proposal, thanks @klaus-sap and @Huy-Dinh-Reply. After reading it, I think what you want to achieve here can already be achieved today it you use inline authentication. Then you can put OPA or OpenFGA in front of OpenBao and make the policy decision before the request ever reaches OpenBao. The only gap we have is for dynamic secrets that require leases, but I think we would also be able to extend inline auth to support them.

website/content/community/rfcs/external-authorization.mdx
klaus-sap: @cipherboy makes sense. So we would return an slice of bool instead of one single bool? Or a slice of allowed operations?

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Not sure. In the external authorization service we need one user identifier only, for example the email address. So `logical.Auth` might be too fat ;-)

klaus-sap:
> Interesting proposal, thanks @klaus-sap and @Huy-Dinh-Reply. After reading it, I think what you want to achieve here can already be achieved today it you use inline authentication. Then you can put OPA or OpenFGA in front of OpenBao and make the policy decision before the request ever reaches OpenBao. The only gap we have is for dynamic secrets that require leases, but I think we would also be able to extend inline auth to support them.

@JanMa not sure if I understand your proposal correctly. Do you mean to have a security proxy in front of OpenBao which calls OPA or OpenFGA? In this case, OpenBao could not enforce the external checks anymore. In addition, the second and third example from section "User-facing Description" would be hard to implement. And from a proxy it would be hard to sync changes from OpenBao to OpenFGA.

website/content/community/rfcs/external-authorization.mdx
klaus-sap: @cipherboy Maybe we should change the proposed method by passing a list of paths instead of the parent path? Then the caller would need to determine the list of children paths to be checked.

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Let me try to answer:

- I would keep CEL builtin. In this case, CEL might offer something similar to the "ask" capability.
- Plugin registration should work like KMS plugins: declare the plugins in the server config and configure them per namespace.
- IMHO, if the policy from `ns1` gets applied, then the plugin configured in `ns1` should be executed.

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Changed method name and signature: filterPaths(...)

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Changed method to return list of allowed operations.

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
+---
+sidebar_label: External Authorization Checks
+description: |-
+  An RFC to discuss how to extend OpenBao policies with external authorization checks.
+---
+
+# RFC: Support for External Authorization Checks
```
satoqz:
```suggestion
# Support for External Authorization Checks
```

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+which are maintained within OpenBao. However, customers may have central authorization services in place like 
+[Open Policy Agent](https://www.openpolicyagent.org/) or [OpenFGA](https://openfga.dev/), 
+which they may want to use for OpenBao, too. In this RFC, we propose a 
+[plugin]( https://openbao.org/docs/plugins/) interface which allows the integration of OpenBao with external 
+authorization services.
+
+## Problem Statement
+
+OpenBao policies provide some special functions which are not common to external authorization services, for example the 
+„deny“ and „sudo“ [capabilities](https://openbao.org/docs/concepts/policies/#capabilities) that do not map to http verbs.
```
satoqz: Can you consistently use `""` quotes? There are all kinds of quotes in this document.

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+  ```
+  # user authorizations are maintained externally
+  path "*" { 
+      capabilities = ["ask”]
+  }
+  ```
+
+## Technical Description
+
+The plugin API consists of four methods only:
```
satoqz, line 74: What's the plugin type name going to be? `authz`? We already have `auth` which we shouldn't collide with too closely, but it's also nice to keep them short.

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+
+Method                                                           | Description
+---------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------
+`hasAuthorization(ctx, path, entity, []operations) []operations` | When processing "ask" capabilities, call plugin to check authorizations for one or multiple operations. Return list of allowed operations.
+`filterPaths(ctx, []paths, entity, operation) []paths`           | Pass list of paths to check for one operation, return list of allowed paths. Needed for example to filter entries on UIs.
+`updatePath(ctx, path, operation) bool`                          | (Optional) After write operations to OpenBao, call plugin to update the DB of the external authorization service when a path changes, i.e. an entry gets created or deleted
```
satoqz: How do we determine whether this is necessary? I assume that the `[]paths` we pass to the plugin are API paths, but writes happen on the storage layer, and there's no existing mechanism that tells us "this endpoint performed a storage write while handling this request".

website/content/community/rfcs/external-authorization.mdx
```diff
@@ -0,0 +1,155 @@
...
+  }
+  ```
+
+## Technical Description
+
+The plugin API consists of four methods only:
+
+Method                                                           | Description
+---------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------
+`hasAuthorization(ctx, path, entity, []operations) []operations` | When processing "ask" capabilities, call plugin to check authorizations for one or multiple operations. Return list of allowed operations.
```
satoqz: Maybe we can spec out an actual Go interface here, for clarity. You've already included `ctx` parameters so we're kind of stuck between a pseudo-code interface and something we can work with right away 😄

website/content/community/rfcs/external-authorization.mdx
klaus-sap: ok

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Fixed with https://github.com/openbao/openbao/pull/3547/changes/db6281ea074a7b02706c5e80758fe068175df1be

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Well, `authz` would be ok, or `authorization` to have a clearer name?

satoqz: IMO anything that starts with `auth` will be in a weird place because literal `auth` already exists. `policy` comes to mind though it is likely too general; we could also go for an acronym like `epa` (External Policy Agent as @cipherboy coined 😆), we do have precedent with `kms`.

klaus-sap: Not sure, because not all external authz services are based on policies. For example, OpenFGA (ReBAC) is not based on policies. What about `external-authz` or `external-authorization`?

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Not sure if I understand it correctly - authorization checks do not happen on the storage layer, but in the ACL processing, where paths are available.

satoqz: I'm asking about `updatePath`. How does core know when to call that?

website/content/community/rfcs/external-authorization.mdx
klaus-sap: done

website/content/community/rfcs/external-authorization.mdx
klaus-sap: Not sure. Maybe when creating the response to an update request (if http 2xx)? Or triggered by some kind of event mechanism?

website/content/community/rfcs/external-authorization.mdx
satoqz: > `external-authz`

This sounds like the best option to me, for now.

website/content/community/rfcs/external-authorization.mdx
klaus-sap: @cipherboy Changed method HasAuthorization again. If we call it from ACL.AllowOperation (https://github.com/openbao/openbao/blob/main/internal/vault/policy/acl.go#L344), then there is only one operation available to be checked

klaus-sap:
Accidentially closed this PR by deleting my fork.
Please follow up here: https://github.com/openbao/openbao/pull/4003

PR #3554, opened 2026-07-21
Copy binary to additional path in test container

cipherboy:
## Description

When usr-merge is not in use, `bao` may either be at `/bin/bao` or `/usr/bin/bao` and we don't necessarily know which ranks higher in the container's `$PATH` without a fair bit of work. Copy the binary to both as a result.

## Rationale

See also: https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/

## Acknowledgements

       also allowed.

### DIFF
--- a/sdk/helper/testcluster/docker/environment.go
+++ b/sdk/helper/testcluster/docker/environment.go
@@ -1255,7 +1255,7 @@ func (dc *DockerCluster) setupImage(ctx context.Context, opts *DockerClusterOpti
 
 	containerFile := fmt.Sprintf(`
 FROM %s:%s
-COPY bao /bin/bao
+COPY bao /usr/bin/bao
 `, opts.ImageRepo, sourceTag)
 
 	if len(opts.Entrypoint) > 0 {
@@ -1267,7 +1267,7 @@ COPY bao /bin/bao
 		dockhelper.BuildPullParent(true),
 		dockhelper.BuildTags([]string{opts.ImageRepo + ":" + tag}))
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to build image: %w", err)
 	}
 	dc.builtTags[tag] = struct{}{}
 	return tag, nil
--- a/changelog/3554.txt
+++ b/changelog/3554.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+sdk: Better support non-usr-merged distributions in containerized testing.
+```
--- a/sdk/helper/docker/testhelpers.go
+++ b/sdk/helper/docker/testhelpers.go
@@ -957,12 +957,44 @@ func BuildImage(ctx context.Context, api *client.Client, containerfile string, c
 		return nil, fmt.Errorf("failed to build image: %v", err)
 	}
 
-	output, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read image build output: %w", err)
+	var output bytes.Buffer
+	tee := io.TeeReader(resp.Body, &output)
+
+	var buildError error
+	var messages []string
+
+	dec := json.NewDecoder(tee)
+	for {
+		var message map[string]any
+		if err := dec.Decode(&message); err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+
+			return nil, fmt.Errorf("error decoding output from image build: %w", err)
+		}
+
+		if stream, ok := message["stream"]; ok {
+			messages = append(messages, strings.TrimSpace(stream.(string)))
+		}
+
+		if detail, ok := message["errorDetail"]; ok {
+			contents := fmt.Sprintf("[ERROR]: %v", detail)
+			message, ok := detail.(map[string]any)["message"]
+			if ok {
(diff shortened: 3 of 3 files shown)


sdk/helper/testcluster/docker/environment.go
```diff
@@ -1256,6 +1256,7 @@ func (dc *DockerCluster) setupImage(ctx context.Context, opts *DockerClusterOpti
 	containerFile := fmt.Sprintf(`
 FROM %s:%s
 COPY bao /bin/bao
+COPY bao /usr/bin/bao
```
satoqz, line 1258:
```suggestion
RUN ln -sf /bin/bao /usr/bin/bao
```

Just so it doesn't dump the big layer onto my system twice?

satoqz:
> See also: https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/

The weird bit is that the internet tells me Alpine has been usr-merged since v3.23, but this doesn't seem to be the case in practice? I'm probably not finding the right link.

cipherboy:
@satoqz Does it have to do with your new COPY instruction? :thinking:

satoqz:
@cipherboy no, I checked stock `alpine:latest`:

```
docker run -it --rm alpine:latest
/ # realpath /bin
/bin
/ # realpath /usr/bin
/usr/bin
```

Compare to `debian:latest`:

```
docker run -it --rm debian:latest
root@9d14b75b7a7a:/# realpath /bin
/usr/bin
root@9d14b75b7a7a:/# realpath /usr/bin
/usr/bin
```

the instruction replaces `/usr/bin` and doesn't touch `/bin`, so assuming a usr-merged layout where `/bin -> /usr/bin` it shouldn't break anything.

cipherboy:
@satoqz Internet is lying to you; I think it isn't done yet:

https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.24.0

> New installations can opt-in to being /usr-merged, by setting the BOOTSTRAP_USR_MERGED environment variable to 1 before running setup-disk. Manual migration is also possible by installing the merge-usr package and running its binary as root. For more details see the [official announcement](https://alpinelinux.org/posts/2025-10-01-usr-merge.html).

>  Note: Alpine 3.25 is not in development yet.

:shrug:

sdk/helper/testcluster/docker/environment.go
cipherboy: @satoqz Hmm, this doesn't work that well:

```
build messages:
	STEP 1/4: FROM quay.io/openbao/openbao:latest
	Trying to pull quay.io/openbao/openbao:latest...
	Getting image source signatures
	Copying blob sha256:8f1fab994b0152a2e0998fa921427fef868f8aa37e1d86357fea3e890a62e384
	Copying blob sha256:83396098de49bc5dd25c69cf887d2932baa0de483ed8914ab7ae40209ba38e7f
	Copying blob sha256:55afa1ecc21d2bb5e5045f32dafee56272ffd89860bac26f6c32123439af26a4
	Copying blob sha256:3c26ddded9d53fab894d0c32702b5a51255dd91e41100814a954298cff0571db
	Copying blob sha256:1c2ee837358272750a345a3fcc24bcafd6c8de1749bee730e90d034a4ca63530
	Copying blob sha256:c6e9ced2d1b23bfc16d7a9409a11e4065e3dffc104911820429509c12484df9e
	Copying blob sha256:14d3e3bbe932f6bd81e43b6c8db28243250c12c400422fc8ab05e49aa245c1eb
	Copying blob sha256:d6106f7be1f989fe6e343c077b75731a22f11d411cb9d275ccb2bea279239a9f
	Copying config sha256:aa19edcb6e0cb91f42e06dafb4a91a655bc4eaaf7321b60075465efa821b8bf5
	Writing manifest to image destination
	STEP 2/4: COPY bao /bin/bao
	--> Using cache 826d0299d3f748324928822d36d4c57254dd447fa202650efdd2f2a9ce152beb
	--> 826d0299d3f7
	STEP 3/4: RUN rm -f /usr/bin/bao
	rm: can't remove '/usr/bin/bao': Permission denied
	[ERROR]: building at STEP "RUN rm -f /usr/bin/bao": while running runtime: exit status 1
```

In particular, because we're running as `openbao` user, we can't delete the file. But the container `COPY` doesn't care about users and so can clobber it.

satoqz: :sob: fine then.

Thinking about it again, shouldn't copying to `/usr/bin` only be sufficient for both older and newer versions assuming Alpine has consistently placed this before `/bin` in `PATH`? (This also seems to be the standard more generally, I can't find a distro where this is not the case).

cipherboy: > assuming Alpine has consistently placed this before `/bin` in `PATH`?

Yes, assuming this is true, it makes sense to me. I don't have any easy counter-examples, so I'll give it a shot.

satoqz (approved):

cipherboy:
Admin merging; the error message is in tests and the newlines are helpful for overall clarity of the message.

PR #3555, opened 2026-07-21
Fix lease renewal cache invalidation

### LINKED ISSUE #3505: Standby nodes serve permanently stale lease state on `auth/token/lookup-self` under read replication
**Describe the bug**

With standby reads enabled (default since 2.5.0), `GET /v1/auth/token/lookup-self`
answered by a **standby** node returns lease-derived fields (`ttl`, `expire_time`,
`last_renewal_time`) that are permanently stale, while the **active** node returns
correct, current values for the same token.

This is not eventual-consistency lag: on one of our clusters the standbys' view
stayed frozen for **8 days across ~16 successful renewals** and never converged.
The same standbys serve the token's immutable fields (`accessor`, `creation_time`,
`policies`) correctly — including for tokens created *after* the freeze point —
and `bao status` shows identical Raft Committed/Applied Index on all three nodes
at read time, so raft replication itself is current. Only the lease view is
frozen. Observed identically on three independent clusters.

Side-by-side (same token, same minute; full captures attached):

| node | HA mode | `last_renewal_time` | `expire_time` |
|---|---|---|---|
| pod-0 | active | 1784107876 (2026-07-15, current) | 2026-08-16T09:31:16Z (slides on each renewal) |
| pod-1 | standby | 1783425603 (2026-07-07) | 2026-08-08T12:00:03Z (frozen) |
| pod-2 | standby | 1783425603 (2026-07-07) | 2026-08-08T12:00:03Z (frozen) |

**To Reproduce**

1. Run a 3-node HA cluster (integrated raft storage) with standby reads at
   their default (no `disable_standby_reads` in the config).
2. Run `bao token create -orphan -period=768h -policy=<any>`
3. Renew the token one or more times over hours/days:
   `VAULT_TOKEN=<that token> bao token renew` (through any node — the write is
   forwarded to the leader and applies correctly there, response shows
   `token_duration: 768h`).
4. With the same token, call `GET /v1/auth/token/lookup-self` **directly
   against a standby** and against the **active** node; compare `ttl`,
   `expire_time`, `last_renewal_time`.
5. The standby keeps reporting the pre-renewal lease state indefinitely (days,
   through many further renewals); the active node reports the renewed state.

**Expected behavior**

Standby-served lookups reflect lease renewals within normal replication lag
(sub-second/seconds) — or lease-derived endpoints are excluded from
standby-local handling and forwarded to the active node.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): v2.5.2
* OpenBao CLI Version (retrieve with `bao version`): OpenBao v2.5.2 (932fcf892eba8d646a9bfc58a59ea3b2475b17fa), built 2026-03-25T16:16:27Z
* Server Operating System/Architecture: official container image
  (`quay.io/openbao/openbao:2.5.2`, linux/amd64) via the OpenBao Helm chart on
  Kubernetes v1.30.4, nodes Ubuntu 24.04 amd64; 3 replicas, transit auto-unseal

```hcl
ui = true
cluster_name = "openbao-integrated-storage"
disable_mlock = true

listener "tcp" {
  address = "[::]:8200"
  cluster_address = "[::]:8201"

  tls_cert_file = "/openbao/userconfig/selfsigned-tls/tls.crt"
  tls_key_file = "/openbao/userconfig/selfsigned-tls/tls.key"
  tls_client_ca_file = "/openbao/userconfig/selfsigned-tls/ca.crt"
}

service_registration "kubernetes" {}

storage "raft" {
  path = "/openbao/data"
  retry_join {
    leader_tls_servername = "openbao-0.openbao-internal"
    leader_api_addr = "https://openbao-0.openbao-internal:8200"
    leader_ca_cert_file = "/openbao/userconfig/selfsigned-tls/ca.crt"
    leader_client_cert_file = "/openbao/userconfig/selfsigned-tls/tls.crt"
    leader_client_key_file = "/openbao/userconfig/selfsigned-tls/tls.key"
  }
  # (identical retry_join blocks for openbao-1 / openbao-2 omitted)
}

seal "transit" {
  address         = "https://REDACTED.example.com"
  disable_renewal = "false"
  key_name        = "autounseal-CLUSTER"
  mount_path      = "transit/"
}

telemetry {
  prometheus_retention_time = "30s"
  disable_hostname = true
}
```

**Additional context**

- Impact: clients behind a load balancer get contradictory TTL answers
  (correct only 1/N of the time), which breaks token-renewal monitoring; and
  once a standby's frozen `expire_time` passes, that standby considers a
  live, correctly-renewed token expired — locally-served requests
  authenticated by that token are then expected to fail intermittently.
- This looks like the same missing-invalidation pattern as #2879 ("Login MFA
  configs are never invalidated on standbys → permanently stale reads",
  fixed via #3283): standbys never learn about lease updates performed on the
  active node, so lease-derived lookup fields freeze permanently instead of
  converging. Searched open+closed issues (incl. label `core/horizontal`) —
  no existing report covers the token/lease path.
- Workaround we applied: `disable_standby_reads = true` (and routing
  lease-sensitive clients to the active-only Kubernetes service).
- Attached: same-minute per-node lookup-self captures, `bao status` from all
  three nodes showing identical raft indexes, in case they help triage.

[01-per-pod-lookup-comparison.txt](https://github.com/user-attachments/files/30050357/01-per-pod-lookup-comparison.txt)
[02-bao-status-all-nodes.txt](https://github.com/user-attachments/files/30050356/02-bao-status-all-nodes.txt)

cipherboy:
## Description

This fixes several bugs in cache invalidation:

1. Leases were being partially cached because `recoveryMode = true` was set in `loadEntryInternal`. At the same time, its invalidation was not wired up.
2. Token salt was not being invalidated if it ever changed.

I've also added logic to ensure that `sys/` is now fully invalidated manually in the course of the invalidation job; any remaining entries should flag prominently in our logs. 

## Rationale

Resolves: #3505

## Acknowledgements

       also allowed.


internal/vault/expiration.go
```diff
@@ -1939,6 +1891,10 @@ func (m *ExpirationManager) deleteLockForLease(id string) {
 
 // updatePending is used to update a pending invocation for a lease
 func (m *ExpirationManager) updatePending(le *leaseEntry) {
+	if !m.useCache {
+		return
+	}
+
```
phil9909, line 1897: I guess this is never called on the standby anyway, but still won't hurt.

phil9909 (dismissed):

internal/vault/external_tests/token/token_test.go
```diff
@@ -679,3 +682,77 @@ func TestTokenStore_RevocationOnStartup(t *testing.T) {
...
+		require.NoError(t, err)
+		cutoff := 36 * time.Hour / time.Second
+		require.LessOrEqual(collect, int(ttl), int(cutoff), "ttl: %v (%T)", data.Data["ttl"], data.Data["ttl"])
+	}, 10*time.Second, 100*time.Millisecond)
+
+	// Trigger a log message.
+	err = client.Sys().EnableAuthWithOptions("userpass", &api.EnableAuthOptions{
+		Type: "userpass",
+	})
+	require.NoError(t, err)
```
wslabosz-reply: could you explain, what enabling an userpass auth mount does here?

wslabosz-reply (dismissed):
LGTM

internal/vault/expiration.go
cipherboy: It took me too much searching to validate, but yes. 😆 That's why I opted to add it.

internal/vault/external_tests/token/token_test.go
cipherboy: @wslabosz-reply Absolutely nothing 😆 

I just needed a way of separating invalidation log lines between the two writes (below line 738 and above line 705), so enabling userpass was the easiest way I knew of to generate any log lines. Otherwise all the invalidations blended together. 

If you want me to remove, happy to, but don't think it impacts positively or negatively that much. 😆

internal/vault/invalidation.go
satoqz, line 1: `invalidation.go`? I like that we're moving it though.

internal/vault/cache_invalidate.go
```diff
@@ -429,6 +434,24 @@ func (ij *invalidationJob) Execute() error {
...
+		// MFA TOTP entries are not cached.
+	case strings.HasPrefix(ij.nsKey, passwordPolicySubPath):
+		// Password policies are not cached.
+	case strings.HasPrefix(ij.nsKey, "sys/"):
+		// System keys need to be invalidated explicitly because it has no
+		// global invalidation router. If we have not handled such a key
+		// before it gets to the invalidation trigger below, treat it as
+		// fatal and restart.
+		ij.fatal = true
+		return fmt.Errorf("no mechanism to invalidate cache for specified system key: %q", ij.key)
```
satoqz: This is similarly risky to not having it IMO, maybe something to remove after a beta phase? There will be keys under `sys/` that won't need cache invalidation, password policies, workflows, external key _key_ entries come to mind.

satoqz: Perhaps I'd feel better at night if we logged, but didn't mark it fatal, especially since this backports.

cipherboy: @satoqz Fatal from a cache invalidation PoV means we restart the read-enabled standby. We generally don't interrupt existing requests (unless write-lock acquisition takes too long), so at worst this should interrupt our ability to serve new requests which the LB should detect... :thinking: 

I'll pull it from the backport, but I think we absolutely want to land it for v2.7 so we can start cleaning all of those up. Workflows, external keys, &c should be added as ignored fields explicitly IMHO.

cipherboy: (Also, I'm not sure what is riskier--temporary traffic drop or silently dropping invalidations?)

satoqz: > Workflows, external keys, &c should be added as ignored fields explicitly IMHO

Would be good to add the known cases (password policies, workflows, anything else you can think of) explicitly as part of this PR.

satoqz (approved):

internal/vault/external_tests/token/token_test.go
wslabosz-reply: @cipherboy ahh no worries, leave it as it is.

PR #3556, opened 2026-07-21
Ensure AppRole salt is created without transaction

### LINKED ISSUE #3157: AppRole write auth/path/role hangs with context deadline exceeded on Windows
**Describe the bug**
When attempting to configure or write an AppRole auth method role on a native Windows machine, the CLI command hangs and terminates with a context deadline exceeded error.

**To Reproduce**
Steps to reproduce the behavior on Windows:

1. Open a Command Prompt
2. start bao server
3. Open a Command Prompt
4. set VAULT_ADDR=http://localhost:55300
5. bao operator init -key-shares=1 -key-threshold=1
6. bao operator unseal %unseal_key%
7. bao login %root_token%
8. bao auth enable -path dev approle
9. bao write auth/dev/role/tcrole secret_id_ttl=0 secret_id_num_uses=0
10. Observe CLI hanging until the default client timeout is reached, throwing the error:
Error writing data to auth/dev/role/tcrole: context deadline exceeded

**Expected behavior**
The expected behavior is that OpenBao should either successfully configure the role or return a fast validation error. Instead, it hits a Go context timeout.

**Environment:**
* Version: OpenBao v2.5.3, built 2026-04-20T19:28:09Z
* OpenBao v2.5.3 (988c88d7ef54b4d4581629b229488dfba5e085ba), built 2026-04-20T19:28:09Z
OS Name:                   Microsoft Windows 11 Enterprise
System Type:               x64-based PC

```hcl
storage "raft" {
  path="storage"
}

listener "tcp" {
  address="[::]:55300"
  tls_disable="true"
  cluster_address="[::]:55301"
  tls_min_version="tls12" 
  tls_cipher_suites="TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA"
}

api_addr="http://localhost:55300"
cluster_addr="http://localhost:55301"
disable_mlock="true"
log_level="trace"
ui="true"
default_lease_ttl="15m"
max_lease_ttl="15m"
```

**Additional context**

[vault_prov.txt](https://github.com/user-attachments/files/28067724/vault_prov.txt)

Attached is the batch script that does above mentioned steps. Please rename to .bat.
This issue occurs on a standalone Windows machine, ruling out container networking or virtualization layers. 

Additional observation:
Execute step 8)  (  bao auth enable -path dev approle ). 
OpenBao server console showed :  [INFO]  core: enabled credential backend: namespace="" path=dev/ type=approle version=""
Wait for a 1 min for the openbao server console to show : 
Image
Execute step 9  ( bao write auth/dev/role/tcrole secret_id_ttl=0 secret_id_num_uses=0 ) - CLI command doesn't hang up.
CLI console shows Success! Data written to: auth/dev/role/tcrole

cipherboy:
## Description

When starting the backend, fetch the initial salt from storage. This ensures on initial startup we've provisioned the salt and thus do not have to create it. While invalidation may reset it later, we know that we don't have to create it again as we're on a standby and would be unable to write it anyways.

## Rationale

Resolves: #3157

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/external_tests/plugin/external_plugin_test.go
+++ b/internal/vault/external_tests/plugin/external_plugin_test.go
@@ -126,6 +126,11 @@ func TestExternalPlugin_RollbackAndReload(t *testing.T) {
 
 func testRegisterAndEnable(t *testing.T, client *api.Client, plugin pluginhelpers.TestPlugin) {
 	t.Helper()
+	testRegisterPlugin(t, client, plugin)
+	testEnablePlugin(t, client, plugin)
+}
+
+func testRegisterPlugin(t *testing.T, client *api.Client, plugin pluginhelpers.TestPlugin) {
 	if err := client.Sys().RegisterPlugin(&api.RegisterPluginInput{
 		Name:    plugin.Name,
 		Type:    api.PluginType(plugin.Typ),
@@ -135,7 +140,9 @@ func testRegisterAndEnable(t *testing.T, client *api.Client, plugin pluginhelper
 	}); err != nil {
 		t.Fatal(err)
 	}
+}
 
+func testEnablePlugin(t *testing.T, client *api.Client, plugin pluginhelpers.TestPlugin) {
 	switch plugin.Typ {
 	case consts.PluginTypeSecrets:
 		if err := client.Sys().Mount(plugin.Name, &api.MountInput{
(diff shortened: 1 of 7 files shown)


cipherboy:
@phil9909 Do you want to port your test on top of this, later?

cipherboy:
@satoqz I think this is failing with the same error now that your #3486 is looking to solve. If you look at `testExternalPlugin_ContinueOnError`, it expects unmount to succeed:

```go
    // unmount
    switch pluginType {
    case consts.PluginTypeSecrets:
        if err := client.Sys().Unmount(plugin.Name); err != nil {
            t.Fatal(err)
        }
    case consts.PluginTypeCredential:
        if err := client.Sys().DisableAuth(plugin.Name); err != nil {
            t.Fatal(err)
        }
    }
```

even though this can't be the case as we've already modified the plugin + restarted core so it has reloaded it: plugin backend will be `nil` so it can't handle any invalidations.

The only reason why this has ever succeeded is that we've never directed traffic to a plugin which caused writes to occur to storage in the first place.  (Right now we don't direct any traffic to the plugin, but this PR's new updates causes initialize to write data in AppRole which is what is used for the test.)

phil9909:
> @phil9909 Do you want to port your test on top of this, later?

Did test it against `main` and your PR branch via wine: `main` still causes the hang, while your PR branch works.

internal/vault/external_tests/plugin/external_plugin_test.go
```diff
@@ -264,7 +277,44 @@ func testExternalPlugin_ContinueOnError(t *testing.T, mismatch bool, pluginType
...
+				}
+			}
+			if core.Sealed() {
+				t.Fatal("should not be sealed")
+			}
+		}
+
+		// Wait for active so post-unseal takes place
+		// If it fails, it means unseal process failed
+		vault.TestWaitActive(t, core.Core)
```
phil9909: Shouldn't this do the trick?

```suggestion
		cluster.UnsealCores(t)
```

phil9909 (dismissed):

wslabosz-reply (dismissed):

satoqz (approved):

PR #3574, opened 2026-07-22
auth/jwt: surface OIDC discovery errors during configuration

### LINKED ISSUE #3358: Tailing slash in `oidc_discovery_url` breaks url
**Describe the bug**
when the `oidc_discovery_url` contains a tailing slash i get an error that i could not connect

pascallibenzi:
## Description

- OIDC discovery/issuer validation errors were not being properly propagated to the caller during configuration
- The change returns the existing validation error, from the underlying library we use, making the actual problem clearly visible to the user, through the UI too
- The change deliberately does not strip, normalize, or otherwise special-case a trailing /, as this can be a real use-case of an issuer that really contains a backslash at its end
- It also includes the JWT authentication, as this part of the configuration is common tp both auth methods

## Rationale

- A trailing slash is only one possible reason why the configured discovery URL and the issuer returned by the OIDC provider may differ
- More generally, the configured URL may resolve or redirect to a provider advertising a different issuer
- Such a configuration will fail later during authentication anyway
- By the way, a typo can also lead to an error, like "network connection", and this is a goog way to understand immediatly the issue with what we configured

Therefore, it is better to reject the invalid configuration immediately and expose the actual validation error to the user.
This provides earlier and clearer feedback without introducing URL-specific normalization behavior.

Resolves: #3358

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3574.txt
+++ b/changelog/3574.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-auth/oidc: error message for oidc_discovery_url now came from the underlying library so we have more details on what's wrong (bad issuer).
+auth/oidc: Return error messages for `oidc_discovery_url` validation errors (like bad issuer) from the underlying library .
 ```
--- a/changelog/3574.txt
+++ b/changelog/3574.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-auth/oidc: Return error messages for `oidc_discovery_url` validation errors (like bad issuer) from the underlying library .
+auth/oidc: Return `oidc_discovery_url` validation errors (like bad issuer) from the underlying library.
 ```
--- a/changelog/3574.txt
+++ b/changelog/3574.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+auth/oidc: error message for oidc_discovery_url now came from the underlying library so we have more details on what's wrong (bad issuer).
+```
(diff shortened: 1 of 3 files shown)


changelog/3574.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+auth/oidc: error message for oidc_discovery_url now came from the underlying library so we have more details on what's wrong (bad issuer).
```
wslabosz-reply:
```suggestion
auth/oidc: Return error messages for `oidc_discovery_url` validation errors (like bad issuer) from the underlying library .
```

wslabosz-reply (dismissed):
LGTM

changelog/3574.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+auth/oidc: Return error messages for `oidc_discovery_url` validation errors (like bad issuer) from the underlying library .
```
wslabosz-reply: I've left an empty space there at the end, sorry. Also probably we could shorten it to:

```suggestion
auth/oidc: Return `oidc_discovery_url` validation errors (like bad issuer) from the underlying library.
```

wslabosz-reply (approved):

DrDaveD (approved):

cipherboy (approved):
Thanks @pascallibenzi!

PR #3588, opened 2026-07-23
auth/jwt: simplify CEL role update logic

phil9909:
## Description

Make use of `GetTimeWithExplicitDefault` and `GetWithExplicitDefault`

## Rationale

more readable:

- less code
- the name `Get(Time)?WithExplicitDefault` makes it pretty clear what is going on.

Resolves: #

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/builtin/credential/jwt/path_cel_role.go
+++ b/internal/builtin/credential/jwt/path_cel_role.go
@@ -220,34 +220,14 @@ func (b *jwtAuthBackend) pathCelRoleCreate(ctx context.Context, req *logical.Req
 		return logical.ErrorResponse(err.Error()), nil
 	}
 
-	expirationLeeway := time.Duration(claimDefaultLeeway) * time.Second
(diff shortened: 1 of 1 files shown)


cipherboy (commented):
Huh, I never thought of using the APIs we added for PATCH for update before.

Don't we just need to get `Default:` on the parameters and use `data.Get(...)` instead of `data.GetOk(...)`?

phil9909:
> Don't we just need to get `Default:` on the parameters and use `data.Get(...)` instead of `data.GetOk(...)`?

Makes more sense. 

Initially I thought about unifying create/update and patch with a helper:
- the create/update case would then pass `FieldData` and a `celRoleEntry` with the defaults set
- the patch case would pass `FieldData` and a `celRoleEntry` with the current values set

But it probably makes it less readable overall. The neat part would be, that we can't forget to update patch after we modified create/update

DrDaveD (approved):

PR #3595, opened 2026-07-23
Add operator quickstart guide

### LINKED ISSUE #3590: Operator quick start guide

There is no easy-to-use guide for operators/administrators who are tasked with deploying OpenBao.

Currently, there is only an [installation guide](https://openbao.org/docs/install/). But where do I go from there? Simply installing the binary doesn't get me a working OpenBao instance.

The rest of the docs are written in "reference" style. Therefore, as someone who is new to OpenBao, I need to skim through a large part of these reference docs to figure out what I need to configure to get my instance running.

A "quick start" guide for operators. This guide should contain minimal steps that are needed to get a working OpenBao instance (install, operator init, familiarization with the CLI and config file, onboarding applications).

This guide can be a simple overview, that links to the reference docs for details. It doesn't need to duplicate the existing content. Such an overview will already be helpful, because it 1) points you directly to the sections that are relevant in the beginning, and 2) gives you an order in which to proceed.

Keep it as is?

n/a

**Additional context**
n/a

thgoebel:
## Description

This MR adds a quick start guide for operators/admins who are new to OpenBao and want to deploy their first instance.

## Rationale

Resolves: #3590

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/get-started/operator-qs.mdx
+++ b/website/content/docs/get-started/operator-qs.mdx
@@ -1,7 +1,8 @@
 # Operator quick start
 
 This quick start explains how operators (administrators) can deploy and initialize OpenBao.
-Once deployed, third-party applications will be able to use your OpenBao instance as a central secret storage.
+At the end of this guide, you will have a simple OpenBao environment that you can use as a playground for learning.
+You will also have connected your first application that can use OpenBao as a central secret storage.
 
 ## Step 1: Install OpenBao
 
@@ -11,8 +12,11 @@ Follow the [installation guide](../install.mdx) to install OpenBao on a machine 
 
 Make sure you know where your OpenBao [config file](../configuration/index.mdx) is
 located, so that you can edit it later when necessary.
-For example, in the official Docker images the file is at `/openbao/config/openbao.hcl`
+For example: the official container images expect to find the configuration at `/openbao/config/openbao.hcl`,
 and the native Linux packages (DEB/RPM) place it at `/etc/openbao/openbao.hcl`.
+
+If you are running OpenBao using a container, you typically provide a full configuration at `/openbao/config/openbao.hcl`
+(for example, using a Docker volume mount or a Kubernetes ConfigMap).
 
 ## Step 3: Configure listener
 
@@ -20,12 +24,16 @@ Configure where OpenBao should listen (Unix socket, TCP HTTP/HTTPS) using the
 [`listener` stanza](../configuration/listener/index.mdx).
 Edit your `openbao.hcl` accordingly.
 
-When [using HTTPS](../configuration/listener/tcp.mdx#configuring-tls),
-use your preferred CA/PKI to obtain a private key and certificate to use with OpenBao.
+When using HTTPS, use your preferred CA/PKI to obtain a private key and certificate to use with OpenBao.
+OpenBao [supports ACME](../configuration/listener/tcp.mdx#acme-parameters)
+to automatically request certificates from services such as Let's Encrypt.
 
 ## Step 4: Check status
 
-Check OpenBao's status, to verify that the instance is running and that you can connect to the API:
+Check OpenBao's status, to verify that the instance is running and that you can connect to the API.
+
+From the command line, you can use the [`bao` tool](../commands/index.mdx) to
+interact with the API and query the status:
 
 ```shell-session
 $ bao status
@@ -39,6 +47,7 @@ Sealed                   true
 ```
 
 Alternatively, you can visit the UI at https://localhost:8200.
+This may trigger a TLS warning if you are using a private CA.
 
 ## Step 5: Initialize OpenBao
 
@@ -60,9 +69,19 @@ Initial Root Token: s.4sKYlIzhy31U4o5ISZdeap60
 # ...
 ```
 
-Securely note down the unseal keys and the [root token](../concepts/tokens.mdx#root-tokens).
-You will need the unseal keys to unseal your OpenBao instance
-and the root token to authenticate to perform further configuration.
+OpenBao's storage is encrypted at rest; therefore, OpenBao requires cryptographic
+material each time that it starts: the [_unseal keys_](../concepts/seal.mdx).
+When you initialize the instance, OpenBao generates a set of unseal keys.
+Securely note down the unseal keys, as without them, you won't be able to decrypt the storage.
+
+From now on, you will need to unseal OpenBao after every restart.
+To do so, run [`operator unseal`](../commands/operator/unseal.mdx) multiple times.
+The amount depends on the threshold of your Shamir seal (default 3).
+You can later set up automatic unsealing; if you do, you will instead need to securely store a set
+of _recovery keys_.
+
+Also note down the [_root token_](../concepts/tokens.mdx#root-tokens).
+It is needed to authenticate to the OpenBao API, through which you will perform further configuration.
 
 :::warning
 
@@ -72,19 +91,15 @@ It is highly recommended to only use the root token for the initial configuratio
 
 :::
 
-From now on, you will need to [unseal](../concepts/seal.mdx) OpenBao after every restart.
-To do so, run [`operator unseal`](../commands/operator/unseal.mdx) multiple times.
-The amount depends on the threshold of your Shamir seal (default 3).
-
 ## Step 6: Enable audit device
 
 [Audit devices](../audit/index.mdx) keep a log of all requests to OpenBao.
-It is strongly recommended to enable at least one audit device.
+The OpenBao project strongly recommends that you enable at least one audit device.
 For example, in container environments, it is common to log to stdout:
 
 ```hcl
 audit "file" "to-stdout" {
-  description = "This audit device should never fail."
+  description = "Write audit information to standard output."
   options {
     file_path = "/dev/stdout"
     log_raw = "true"
@@ -92,7 +107,14 @@ audit "file" "to-stdout" {
 }
 ```
 
-## Step 7: Read/write secrets
+:::warning
+
+Make sure that you have an appropriate log collection solution to durably store
+the audit logs and to protect them from tampering.
+
+:::
+
+## Step 7: Test secrets reading/writing
 
 For test purposes, log in with the root token:
 
@@ -108,16 +130,17 @@ $ bao write cubbyhole/my-secret my-value=s3cr3t
 $ bao read cubbyhole/my-secret
 ```
 
-:::warning
+:::info
 
-`cubbyhole` is ephemeral, and scoped to the lifetime of the token!
+`cubbyhole` is ephemeral, and scoped to the lifetime of the token.
 Real applications should use a secrets engine suitable to their use case (see below).
 
 :::
 
 ## Step 8: Connect client applications
 
-To connect production applications to OpenBao, you should:
+Now that your OpenBao instance is running, connect your applications to it.
+This involves the following steps:
 
 1. [**Enable the secrets engine**](../secrets/index.mdx) that your application plans to use.
    Different types of engines are available, depending on the use case that you are deploying OpenBao for.
@@ -134,19 +157,42 @@ To connect production applications to OpenBao, you should:
 
 Your client applications should now be able to access OpenBao.
 
-## Next Steps
+:::warning
 
-You now have a basic OpenBao instance running.
-Here are additional topics that you might be interested in to further tweak your instance:
+Apply the principle of least privilege!
+For example: only enable the secret engines and auth methods that you use,
+write policies that are restrictive, and use separate identities for separate applications.
 
-- [**Plugins:**](../plugins/index.mdx)
-  Plugins extend the core OpenBao functionality. The main plugin categories are
-  auth methods, secret engines, database providers, and KMS providers.
+:::
+
+## Conclusion
+
+You now have a basic OpenBao instance running and applications connected.
+
+If this is your first time following this guide, you likely took a few shortcuts
+(such as copying the unseal keys to an insecure notes app, skipping setting up
+an audit device, using the root token for everything, or defining a policy
+without any restrictions).
+
+To set up a production instance, you can ultimately follow the same steps as above.
+However, don't take shortcuts, and follow all advice and warnings.
+Before deploying a production instance, the OpenBao projects recommends that you
+familiarize yourself with the rest of this documentation.
+A list of further reading topics is provided below.
+
+## Further topics
+
+- [**Concepts:**](../concepts/index.mdx)
+  Learn how OpenBao works and how its concepts impact your day-to-day usage and operation.
 
 - [**High availability:**](../commands/operator/raft.mdx)
   OpenBao supports high availability via the integrated storage backend.
   You can add additional OpenBao instances to the cluster using `operator raft join`.
   Note: Only the first instance must be initialized with a seal.
+
+- [**Plugins:**](../plugins/index.mdx)
+  Plugins extend the core OpenBao functionality. The main plugin categories are
+  auth methods, secret engines, database providers, and KMS providers.
 
 - [**Sealing:**](../configuration/seal/index.mdx)
   OpenBao can be unsealed with different methods, for example, with keys stored in a KMS or HSM.
--- a/website/content/docs/get-started/operator-qs.mdx
+++ b/website/content/docs/get-started/operator-qs.mdx
@@ -18,6 +18,13 @@ and the native Linux packages (DEB/RPM) place it at `/etc/openbao/openbao.hcl`.
 If you are running OpenBao using a container, you typically provide a full configuration at `/openbao/config/openbao.hcl`
 (for example, using a Docker volume mount or a Kubernetes ConfigMap).
 
+:::info
+
+If you edit the config file while OpenBao is already running, you need to restart
+OpenBao or send a SIGHUP signal to cause OpenBao to reload the config file.
+
+:::
+
 ## Step 3: Configure listener
 
 Configure where OpenBao should listen (Unix socket, TCP HTTP/HTTPS) using the
@@ -53,7 +60,8 @@ the audit logs and to protect them from tampering.
 
 ## Step 5: Check status
 
-Check OpenBao's status, to verify that the instance is running and that you can connect to the API.
+Start OpenBao (or restart it if it was already running).
+Then check OpenBao's status, to verify that the instance is up and that you can connect to the API.
 
 From the command line, you can use the [`bao` tool](../commands/index.mdx) to
 interact with the API and query the status:
--- a/website/content/docs/get-started/operator-qs.mdx
+++ b/website/content/docs/get-started/operator-qs.mdx
@@ -1,0 +1,157 @@
+# Operator quick start
+
+This quick start explains how operators (administrators) can deploy and initialize OpenBao.
+Once deployed, third-party applications will be able to use your OpenBao instance as a central secret storage.
+
+## Step 1: Install OpenBao
+
+Follow the [installation guide](../install.mdx) to install OpenBao on a machine of your choice.
+
+## Step 2: Locate config file
+
+Make sure you know where your OpenBao [config file](../configuration/index.mdx) is
+located, so that you can edit it later when necessary.
+For example, in the official Docker images the file is at `/openbao/config/openbao.hcl`
+and the native Linux packages (DEB/RPM) place it at `/etc/openbao/openbao.hcl`.
+
+## Step 3: Configure listener
+
+Configure where OpenBao should listen (Unix socket, TCP HTTP/HTTPS) using the
+[`listener` stanza](../configuration/listener/index.mdx).
+Edit your `openbao.hcl` accordingly.
+
+When [using HTTPS](../configuration/listener/tcp.mdx#configuring-tls),
+use your preferred CA/PKI to obtain a private key and certificate to use with OpenBao.
+
+## Step 4: Check status
+
+Check OpenBao's status, to verify that the instance is running and that you can connect to the API:
+
+```shell-session
+$ bao status
+Key                      Value
+---                      -----
+Seal Type                shamir
+Initialized              false
+Sealed                   true
+
+# ...
+```
+
+Alternatively, you can visit the UI at https://localhost:8200.
+
+## Step 5: Initialize OpenBao
+
+The [`operator` command](../commands/operator/index.mdx) of the [OpenBao CLI](../commands/index.mdx)
+is the main interaction point for operators with their OpenBao instance.
+
+Most importantly, you need to initialize the OpenBao instance before it can be used:
+
+```shell-session
+$ bao operator init
+Unseal Key 1: sP/4C/fwIDjJmHEC2bi/1Pa43uKhsUQMmiB31GRzFc0R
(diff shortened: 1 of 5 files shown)


thgoebel:
I would love to point the yellow "Try it out" button on the homepage to this new operator quickstart guide. However, it keeps failing the _versioned_ build (`make build-versioned`) with a broken link (while a plain `pnpm run build` works).

I couldn't find a good way to make it work, so I'll leave it for now. I think either this needs to be backported, or we just wait a few months. And then we can point the "Try it out" button to this new guide.

cipherboy:
@thgoebel Yeah, you likely would need to link to `/docs/next/...`, but I think that'd break the regular scratch build.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,137 @@
...
+   which paths your application is allowed to access.
+
+1. [**Choose an auth method**](../auth/index.mdx) to let client applications authenticate against OpenBao.
+   The default built-in method is tokens, but many other methods are available (such as LDAP and OIDC).
+
+1. **Configure the auth method** and create an entity for each of your applications.
+   See the "Configuration" section of your chosen auth method for details.
+   At creation time, you can link a policy to the entity to restrict what it can access.
+
+Your client applications should now be able to access OpenBao.
```
satoqz: Somewhere around here the guide should definitely mention creating an audit device as soon as possible.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,137 @@
...
+Unseal Key 5: wyd7rMGWX5fi0k36X4e+C4myt5CoTmJsHJ0rdYT7BQcF
+
+Initial Root Token: 6662bb4a-afd0-4b6b-faad-e237fb564568
+
+# ...
+```
+
+Securely note down the unseal keys and the [root token](../concepts/tokens.mdx#root-tokens).
+You will need the unseal keys to unseal your OpenBao instance
+and the root token to authenticate to perform further configuration.
```
wslabosz-reply: not sure but it could be helpful to add a note about (more advanced) auto seal feature here

wslabosz-reply: ahh it's at the end, that also works.

website/content/docs/get-started/operator-qs.mdx
cipherboy: @satoqz Shouldn't audit device go even earlier in the configuration before startup? Somewhere around step 3 into 4?

satoqz: @cipherboy probably, but what audit device would you recommend in a quickstart guide that's trying to be generic and defer to specific needs up to this section? stdout?

It's not that we're setting up a production instance here, HA and Auto Seals are in the last section below.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,137 @@
...
+  auth methods, secret engines, database providers, and KMS providers.
+
+- [**High availability:**](../commands/operator/raft.mdx)
+  OpenBao supports high availability via the integrated storage backend.
+  You can add additional OpenBao instances to the cluster using `operator raft join`.
+  Note: Only the first instance must be initialized with a seal.
+
+- [**Sealing:**](../configuration/seal/index.mdx)
+  OpenBao can be unsealed with different methods, for example, with keys stored in a KMS or HSM.
+  This is useful for auto-unsealing, to avoid needing manual operator intervention after restarts.
```
satoqz, line 207: Would be nice to mention self-initialization right below Auto Seals since they go hand-in-hand.

website/content/docs/get-started/operator-qs.mdx
cipherboy: stdout always, it is probably the most useful in production (k8s for instance), as most container execution platforms support stdout log aggregation.

website/content/docs/get-started/operator-qs.mdx
thgoebel: I've added self-initialization.

website/content/docs/get-started/operator-qs.mdx
thgoebel: I've added it mentioning audit devices.

I've added it AFTER initialization, because I believe the storage must be unsealed in order to be able to enable an audit device?

satoqz: @thgoebel yes, though by including the audit device configuration stanza from the get-go it'll automatically enable post-initialize, which is more optimal.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+```shell-session
+$ bao operator init
+Unseal Key 1: sP/4C/fwIDjJmHEC2bi/1Pa43uKhsUQMmiB31GRzFc0R
+Unseal Key 2: kHkw2xTBelbDFIMEgEC8NVX7NDSAZ+rdgBJ/HuJwxOX+
+Unseal Key 3: +1+1ZnkQDfJFHDZPRq0wjFxEuEEHxDDOQxa8JJ/AYWcb
+Unseal Key 4: cewseNJTLovmFrgpyY+9Hi5OgJlJgGGCg7PZyiVdPwN0
+Unseal Key 5: wyd7rMGWX5fi0k36X4e+C4myt5CoTmJsHJ0rdYT7BQcF
+
+Initial Root Token: 6662bb4a-afd0-4b6b-faad-e237fb564568
```
cipherboy: This doesn't quite look like the right format...? Root token begins with `s.` usually.

thgoebel: I copied this from `website/content/docs/commands/operator/index.mdx`. I've just fixed both occurences.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
+# Operator quick start
```
timb-controlplane, line 1: I expect this should have some front matter, eg:
```yaml
---
description: Set up a basic OpenBao cluster for real use
---
```

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+This quick start explains how operators (administrators) can deploy and initialize OpenBao.
+Once deployed, third-party applications will be able to use your OpenBao instance as a central secret storage.
+
+## Step 1: Install OpenBao
+
+Follow the [installation guide](../install.mdx) to install OpenBao on a machine of your choice.
```
timb-controlplane, line 9: Even for a quick start, the guide should cover how many computers to use. Typically, more than one. It's OK to explain a limited case (similar to https://openbao.org/docs/get-started/developer-qs/) - but then, this page needs to explain how it is different.

The Kubernetes docs split two ways:
- [Learning](https://kubernetes.io/docs/setup/learning-environment/)
- [Production](https://kubernetes.io/docs/setup/production-environment/)

I think that's a good split. It's OK to target either side of that split. **but** it also really helps to communicate whether the audience is people deploying OpenBao to learn it, or people deploying it in a production-ready / production-like way.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+## Step 2: Locate config file
+
+Make sure you know where your OpenBao [config file](../configuration/index.mdx) is
+located, so that you can edit it later when necessary.
+For example, in the official Docker images the file is at `/openbao/config/openbao.hcl`
```
timb-controlplane: (nit)
```suggestion
For example: the official container images expect to find the configuration at `/openbao/config/openbao.hcl`,
```

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+and the native Linux packages (DEB/RPM) place it at `/etc/openbao/openbao.hcl`.
+
```
timb-controlplane, line 17:
```suggestion

If you are running OpenBao using a container, you typically provide a full configuration at `/openbao/config/openbao.hcl`
(for example, using a Kubernetes ConfigMap).

```
(does that sound right?)

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+## Step 3: Configure listener
+
+Configure where OpenBao should listen (Unix socket, TCP HTTP/HTTPS) using the
+[`listener` stanza](../configuration/listener/index.mdx).
+Edit your `openbao.hcl` accordingly.
+
+When [using HTTPS](../configuration/listener/tcp.mdx#configuring-tls),
+use your preferred CA/PKI to obtain a private key and certificate to use with OpenBao.
```
timb-controlplane: This isn't wrong, but if people are using OpenBao, they'll probably want to automate this.

Maybe they _shouldn't_ use OpenBao for the certificate(s) that OpenBao uses. That's worth calling out, I think.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+```shell-session
+$ bao status
+Key                      Value
+---                      -----
+Seal Type                shamir
+Initialized              false
+Sealed                   true
+
+# ...
```
timb-controlplane, line 77: In my mind, we might want to call out that the CLI client tool (`bao`) is also the thing you run. That's not obvious.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+```
+
+Alternatively, you can visit the UI at https://localhost:8200.
```
timb-controlplane, line 80: This might trigger a TLS warning. Worth mentioning?

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+## Step 5: Initialize OpenBao
+
+The [`operator` command](../commands/operator/index.mdx) of the [OpenBao CLI](../commands/index.mdx)
```
timb-controlplane, line 85: (nit)
```suggestion
The [`operator` subcommand](../commands/operator/index.mdx) of the [OpenBao CLI](../commands/index.mdx)
```

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+Unseal Key 3: +1+1ZnkQDfJFHDZPRq0wjFxEuEEHxDDOQxa8JJ/AYWcb
+Unseal Key 4: cewseNJTLovmFrgpyY+9Hi5OgJlJgGGCg7PZyiVdPwN0
+Unseal Key 5: wyd7rMGWX5fi0k36X4e+C4myt5CoTmJsHJ0rdYT7BQcF
+
+Initial Root Token: s.4sKYlIzhy31U4o5ISZdeap60
+
+# ...
+```
+
+Securely note down the unseal keys and the [root token](../concepts/tokens.mdx#root-tokens).
```
timb-controlplane:
```suggestion
OpenBao requires cryptographic material each time that it starts: the [unseal keys](../concepts/seal.mdx#unsealing). When you initialize the instance, OpenBao provides
a set of unseal keys. Securely note down the unseal keys and the [root token](../concepts/tokens.mdx#root-tokens).

You can later set up automatic unsealing; if you do, you will instead need to securely store a set
of _recovery keys_.

```
?

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+:::warning
+
+These values are highly sensitive and grant access to your OpenBao instance!
+It is highly recommended to only use the root token for the initial configuration
+(to set up additional auth methods and scoped policies, see below) and discard it afterwards.
+
+:::
+
+From now on, you will need to [unseal](../concepts/seal.mdx) OpenBao after every restart.
```
timb-controlplane: If accepting my suggestion to mention unsealing earlier, it's OK to drop the hyperlink on this line.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+To do so, run [`operator unseal`](../commands/operator/unseal.mdx) multiple times.
+The amount depends on the threshold of your Shamir seal (default 3).
+
+## Step 6: Enable audit device
+
+[Audit devices](../audit/index.mdx) keep a log of all requests to OpenBao.
+It is strongly recommended to enable at least one audit device.
```
timb-controlplane:
```suggestion
The OpenBao project strongly recommends that you enable at least one audit device.
```

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+```hcl
+audit "file" "to-stdout" {
+  description = "This audit device should never fail."
+  options {
+    file_path = "/dev/stdout"
+    log_raw = "true"
+  }
+}
+```
+
```
timb-controlplane:
```suggestion

(This is part of a solution; if you use this approach, make sure that your container platform can durably store the audit logs and protect them from tampering).

```

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+## Step 7: Read/write secrets
```
timb-controlplane:
```suggestion
## Step 7: Test that you can read and write secrets
```

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+1. [**Choose an auth method**](../auth/index.mdx) to let client applications authenticate against OpenBao.
+   The default built-in method is tokens, but many other methods are available (such as LDAP and OIDC).
+
+1. **Configure the auth method** and create an entity for each of your applications.
+   See the "Configuration" section of your chosen auth method for details.
+   At creation time, you can link a policy to the entity to restrict what it can access.
+
+Your client applications should now be able to access OpenBao.
+
+## Next Steps
```
timb-controlplane: To me, all of this sounds like part of a “next steps” section.

timb-controlplane (commented):
Thanks for the PR.

I've got much more experience with technical writing than I have with OpenBao, and I hope this feedback is helpful given that background.

In my opinion, the number 1 thing to make clear in this new page is the **target audience**.

Is it:
- people installing OpenBao to learn about how to deploy and secure it
- people who need to deploy OpenBao securely and want a set of instructions for doing that
  - (can overlap with: people writing their own  Kubernetes operators / Helm charts / etc)

?

website/content/docs/get-started/operator-qs.mdx
cipherboy: @thgoebel I suspect you've seen them, but TLS configuration does support ACME: https://openbao.org/docs/configuration/listener/tcp/#acme-parameters  -- and perhaps novel, we've done some work towards a self-hosted OpenBao demo in the past: https://github.com/JanMa/openbao-demo-config/pull/1

website/content/docs/get-started/operator-qs.mdx
cipherboy: One suggestion on top of Tim's as well: if we're going to recommend auto-unsealing (we should, its nice), we should probably link to the seal migration doc: https://openbao.org/docs/concepts/seal/#seal-migration

website/content/docs/get-started/operator-qs.mdx
timb-controlplane: (line 118, `## Step 8: Connect client applications`, feels like it is part of the next steps for the reader)

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+## Step 6: Enable audit device
+
+[Audit devices](../audit/index.mdx) keep a log of all requests to OpenBao.
+It is strongly recommended to enable at least one audit device.
+For example, in container environments, it is common to log to stdout:
+
+```hcl
+audit "file" "to-stdout" {
+  description = "This audit device should never fail."
```
timb-controlplane:
```suggestion
  description = "Write audit information to standard output."
```

website/content/docs/get-started/operator-qs.mdx
timb-controlplane: If this feedback looks confusing, see https://github.com/openbao/openbao/pull/3595#discussion_r3665636385 for better context.

website/content/docs/get-started/operator-qs.mdx
thgoebel: If there is no front matter, Docusaurus takes it from the Markdown:

- When there is no `sidebar_label` or `title`, it takes both from the `#`-heading.
- When there is no description, it takes the first sentence of the page.

If you build this page, it results in:

```html
<meta name="description" content="This quick start explains how operators (administrators) can deploy and initialize OpenBao." data-rh="true">
```

Because of that behaviour, I usually drop the front matter when the Markdown itself is good enough, to avoid duplication. Additionally, I generally try to write docs in BLUF style so that the first sentence and/or paragraph of the Markdown body can serve as a description, too.

website/content/docs/get-started/operator-qs.mdx
thgoebel: I used "command" since most of the existing docs talk about "commands" instead of "subcommands". The [`operator` doc](https://openbao.org/docs/commands/operator/) itself says:

> The operator command groups subcommands for operators interacting with OpenBao. Most users will not need to interact with these commands.

I agree that you could consider them subcommands of the `bao` command, but personally I also prefer "command". Otherwise you would need to call `operator init` a sub-sub-command. That becomes unwieldy.

website/content/docs/get-started/operator-qs.mdx
thgoebel: I have moved these sentences around a bit to incorporate your feedback.

website/content/docs/get-started/operator-qs.mdx
thgoebel: I have added a sentence to similar effect.

website/content/docs/get-started/operator-qs.mdx
thgoebel: I have added a shorter version that incorporates the word "test". I am trying to keep all h2-titles in this doc short.

website/content/docs/get-started/operator-qs.mdx
thgoebel: The idea is that Step 8 is still the minimum that you need to do to get a basic demo/playground/learning environment working. Think of it as step 8a,b,c,d.

"Next Steps" are things that are optional, not required for a learning environment, but relevant for a production environment.

I have reworked this section to incorporate your "what is the target audience and target environment" in the conclusion.

thgoebel:
Thanks for your review @timb-controlplane! I tried to incorporate most of your feedback, either directly or slightly modified, and left comments where I disagree. Please take a look and let me know what you think.

website/content/docs/get-started/operator-qs.mdx
thgoebel: I added more words to clarify this, both in the introduction and at the end.

website/content/docs/get-started/operator-qs.mdx
thgoebel: Does it reload the config post-initialize?

Aka, if the order is: "Check status" -> "Add audit device" -> "operator init" then do I need a SIGHUB/restart after adding the audit device?

cipherboy: @thgoebel The config needs to be written and loaded prior to operator initialization. So if you:

1. Write a config file (A)
2. Start the server.
3. Write another config file (B)
4. Send SIGHUP
5. Write another config file (C)
6. Call operator init.

Then if an audit device is in A or B, it'll be added during initialization, but if its in C, it won't be seen.

Hence why we're encouraging putting audit device setup in config file prior to starting the server in the beginning.

website/content/docs/get-started/operator-qs.mdx
timb-controlplane: My suggestion: subheadings within “Next steps”. Not always right but it is the approach I would personally use.

website/content/docs/get-started/operator-qs.mdx
timb-controlplane: Sure. This one can be resolved.

website/content/docs/get-started/operator-qs.mdx
timb-controlplane: Sure. This one can be resolved.

website/content/docs/get-started/operator-qs.mdx
timb-controlplane: I don't think subsubcommand is a thing. But the command / subcommand split is one.

OK, something for a bigger / separate issue.

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+You can then read/write secrets that are scoped to this token using the
+[`cubbyhole` secrets engine](../secrets/cubbyhole.mdx):
+
+```shell-session
+$ bao write cubbyhole/my-secret my-value=s3cr3t
+$ bao read cubbyhole/my-secret
+```
+
+:::warning
```
timb-controlplane: (nit) I wonder if this can be a callout less than a warning.

website/content/docs/get-started/operator-qs.mdx
thgoebel: "Next steps" is now split into "Conclusion" and "Further topics".

website/content/docs/get-started/operator-qs.mdx
thgoebel: I made it an "info".

website/content/docs/get-started/operator-qs.mdx
thgoebel: Ok. I now:

- Added an :::info box to teach the user when the config is reloaded
- Moved the audit device to the config steps at the top (between "listener" and "check status")
- Added an explicit "(re)start OpenBao" instruction to the "Check status" step, to ensure that the user restarts OpenBao after configuring it.

cipherboy:
@thgoebel Sorry for the delay; I think you need to rebase this as there's a commit without DCO sadly. 

```
git rebase HEAD~5 --signoff
```

thgoebel:
Done. Signed off the last commit and rebased on main.

cipherboy (approved):
Thanks @thgoebel!

website/content/docs/get-started/operator-qs.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+## Further topics
+
+- [**Concepts:**](../concepts/index.mdx)
+  Learn how OpenBao works and how its concepts impact your day-to-day usage and operation.
+
+- [**High availability:**](../commands/operator/raft.mdx)
+  OpenBao supports high availability via the integrated storage backend.
+  You can add additional OpenBao instances to the cluster using `operator raft join`.
+  Note: Only the first instance must be initialized with a seal.
```
cipherboy, line 199: Note to @cipherboy -- revise later.

PR #3600, opened 2026-07-23
Misc minor docs edits

thgoebel:
## Description

This MR is a collection of minor docs edits that I thought might be useful (that I came across during my larger work around the plugin docs).

## Rationale

See the individual commit messages.

Resolves: n/a

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/commands/index.mdx
+++ b/website/content/docs/commands/index.mdx
@@ -260,10 +260,12 @@ automatically query the OpenBao server and return helpful argument suggestions.
 
 ## Token helper
 
+To authenticate, use the [`login` command](login.mdx).
+
 By default, the OpenBao CLI uses a "token helper" to cache the token after
 authentication. This is conceptually similar to how a website securely stores
 your session information as a cookie in the browser. Token helpers are
-customizable, and you can even build your own.
+customizable, and you can even [build your own](token-helper.mdx).
 
 The default token helper stores the token in `~/.vault-token`. You can delete
 this file at any time to "logout" of OpenBao. You can override this location
--- a/website/content/docs/commands/plugin/init.mdx
+++ b/website/content/docs/commands/plugin/init.mdx
@@ -1,4 +1,5 @@
 ---
+sidebar_label: init
 description: |-
   The "plugin init" command downloads OCI-based plugins configured in
   OpenBao server configuration files.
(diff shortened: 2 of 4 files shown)


website/content/docs/commands/index.mdx
```diff
@@ -260,10 +260,16 @@ automatically query the OpenBao server and return helpful argument suggestions.
 
 ## Token helper
 
+To authenticate, use the [`login` command](login.mdx):
```
cipherboy: I've always been entertained by this piece of advice and probably is a bit of a longer form philosophical discussion 😆 

How should the CLI auth work?

The issue with this particular piece of guidance is it assumes a two-step flow:

```
$ bao write auth/.... 
$ bao login
... <paste the token from above>
```

This is fine, and yes, works on paper. Token goes on clipboard, to end up on disk, which is probably better than environment variables in certain cases.

You could instead do:

```
$ bao login -method=userpass ...
```

and cut out the leakage to clipboard, or if the backend auth plugin isn't of a supported type, do:

```
$ bao write -field=token ... > ~/.vault-token
```

(cutting out `bao login` and and lose some tangentially useful information that's still returned by `bao token lookup` if desired). 


Should `bao login -method=write` be a thing and function like `bao write` but performing the login? Should we be recommending a bare `bao login` in the first place? &c.

satoqz: @cipherboy I think you're forgetting about custom token helpers? So it doesn't always go to disk, and `bao login` is a convenient way to call those.

> Should bao login -method=write be a thing and function like bao write but performing the login? Should we be recommending a bare bao login in the first place? &c.

I've thought about this before, too, I think it'd be useful.

website/content/docs/commands/index.mdx
```diff
@@ -260,10 +260,16 @@ automatically query the OpenBao server and return helpful argument suggestions.
...
+
+```shell-session
+$ bao login
+```
```
satoqz: I would cut this example though and leave only the link you've added above. That will give readers the chance to discover the `-method` flag without trying to explain it inline.

thgoebel: I've remove the example.

satoqz (approved):

website/content/docs/commands/index.mdx
cipherboy: > @cipherboy I think you're forgetting about custom token helpers? So it doesn't always go to disk, and `bao login` is a convenient way to call those.

Ahh yes, thanks @satoqz. I do still maintain most people would probably be calling `bao login -method=...` rather than `bao write` + `bao login`, though that probably isn't the case when they're manually authing on a GCP/AWS/... box and using system auth. But I'd imagine most people do that rather rarely.

Token helpers need a fair bit of help though, we should probably solicit RFCs for that.

PR #3436, opened 2026-07-06
Control groups docs

suprjinx:
## Description

Added documentation for access policy Control Groups (PR #2241). Some wording and examples are adapted from the fork-point documentation.

## Rationale

Initial development PR has been merged, this is documenting the new behavior.

Relates: #2241 
Resolves: #1669 

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3601, opened 2026-07-23
Return request data in sys/control-group/request

cipherboy:
## Description

When approving a request, often the most important field is the request data. Ensure we return it so authorizing parties can make an informed decision.

This also fixes the request method to align with the documentation from Vault.

See also: #3436

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/control_group.go
+++ b/internal/vault/control_group.go
@@ -49,6 +49,11 @@ var controlGroupRequestResponseSchema = map[string]*framework.FieldSchema{
 		Description: "The original request path",
 		Required:    true,
 	},
+	"request_data": {
+		Type:        framework.TypeMap,
+		Description: "The original request data",
+		Required:    true,
+	},
 	"request_entity": {
 		Type:        framework.TypeMap,
 		Description: "The original requesting entity",
@@ -355,6 +360,7 @@ func (c *Core) handleControlGroupRequest(ctx context.Context, req *logical.Reque
 			"approved":          approved,
 			"request_operation": originalRequest.Operation,
 			"request_path":      originalRequest.Path,
+			"request_data":      originalRequest.Data,
 			"request_entity":    originalEntity,
 			"authorizations":    auths,
 		},
--- a/internal/vault/logical_system_paths.go
+++ b/internal/vault/logical_system_paths.go
@@ -4276,7 +4276,7 @@ func (b *SystemBackend) wrappingPaths() []*framework.Path {
 			},
 
 			Operations: map[logical.Operation]framework.OperationHandler{
-				logical.ReadOperation: &framework.PathOperation{
+				logical.UpdateOperation: &framework.PathOperation{
 					Callback: b.Core.handleControlGroupRequest,
 					Responses: map[int][]framework.Response{
 						http.StatusOK: {{
--- a/internal/vault/request_handling_test.go
+++ b/internal/vault/request_handling_test.go
@@ -88,9 +88,7 @@ func TestRequestHandling_ControlGroupWrapping(t *testing.T) {
 		Path:  "cg_test",
 		Type:  "kv",
 	})
-	if err != nil {
-		t.Fatalf("err: %v", err)
-	}
+	require.NoError(t, err)
 
 	// Create a secret
 	req := &logical.Request{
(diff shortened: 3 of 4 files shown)


cipherboy:
FYI @suprjinx btw!

wslabosz-reply (approved):

suprjinx (commented):
LGTM

JanMa (approved):

cipherboy:
Thanks all, merging...

cipherboy:
Sigh, wrong button 😆

suprjinx (approved):
LGTM again :-)

satoqz:
@suprjinx @cipherboy should the approver _always_ be able to see the request data? What if I'd like to store a secret that approvers shouldn't see, and don't care to see?

cipherboy:
@satoqz said:

> @suprjinx @cipherboy should the approver _always_ be able to see the request data?

I guess, what is an approver approving then...? If an approver only has author/path/capability and not request data, they lose a lot of information on structure, &c. PKI would be almost entirely useless and you're almost entirely basing the decision on "do you trust that user", which maybe you shouldn't be giving them an ACL policy in the first place to do so (or, visa-versa, not inserting a control group in the mix)? :shrug: 

(And similarly, if an approver is not trusted to see request data, why are they privileged enough to deny the update?)

> What if I'd like to store a secret that approvers shouldn't see, and don't care to see?

You could use the token creation APIs with a role and restricted policies and put the control-group flow on the token request side:

You'd create a policy:

```hcl
path "my/sensitive/secret" {
   capability = ["update"]
}
```

Create a new token role attaching said policy: https://openbao.org/api-docs/auth/token/#createupdate-token-role (probably with `renewable=false`).

And gate access to the execution of `auth/token/create/:role_name` via control groups. Then you'd only see non-sensitive information (approvers only see the request and never the response) and you'd be free to perform your update with the new token. You could even assert that `auth/token/create/:role_name` had to be called with parameters like `num_uses=1` when approving the control group so that it effectively behaves like a single-use unwrap token.

This is roughly the same flow except you have extra steps post-approval-fetch (note that _approval_ isn't a sufficient condition for the request to be executed, the OP has to use the control group response-wrapping token to actually fetch the wrapped result of the gated request in order for the request to be executed in the first place... so whereas the original flow required `(op) request -> (committee) approval -> (op) unwrap`, this is now `(op) request -> (committee) approval -> (op) unwrap temp token -> (op) update secret`.

satoqz:
@cipherboy just asking 😄 I figured it might be surprising behavior when a control group check is attached to endpoints that take per-user secrets, such as shamir shares. Not that this would be a common or even reasonable pattern, but something that could be used to catch key share holders by surprise? Unless you've checked the latest state of the policy, you don't really get a warning that your request data is about to be held and recorded for other's review.

cipherboy:
@satoqz said:

> @cipherboy just asking 😄 I figured it might be surprising behavior when a control group check is attached to endpoints that take per-user secrets, such as shamir shares. Not that this would be a common or even reasonable pattern, but something that could be used to catch key share holders by surprise? Unless you've checked the latest state of the policy, you don't really get a warning that your request data is about to be held and recorded for other's review.

@suprjinx do you mind checking if you can self-revoke the generated control-group token?

I think then you can make an argument that while you don't get a _warning_ you get _recourse_, i.e., you get a control-group token and you can call revoke-self on it, thereby destroying the retained request data.

Also important to note @satoqz that (as mentioned in the docs PR), there's no approval queue and no notification mechanism; approval happens by the original requester notifying them out-of-band and providing the control group accessor to the approvers for approval. So if the requester doesn't like the fact that it resulted in a request in the first place, they just discard it and the token eventually expires. 

If a tree fell in a forest and nobody was around, did it even make a sound? :laughing:

satoqz:
@cipherboy if you can self-revoke it, works for me.

cipherboy:
@satoqz Hmm, no, it looks like you can't self-revoke as the response wrapping policy doesn't allow it:

```
path "cubbyhole/response" {
    capabilities = ["create", "read"]
}

path "sys/wrapping/unwrap" {
    capabilities = ["update"]
}
```

:thinking: I wonder if it should? 

That said, the control group accessor is still valid after revoking (with a modified policy). So, does this preclude merge? Or should I work on adding this support to all wrapping tokens in a follow-up PR?

satoqz:
@cipherboy fine if you'd prefer to figure it out later.

cipherboy:
@satoqz Yeah, the tricky part is that `NumUses: 1` is set, which means that by the time we actually call revoke in the token revocation handler, we've already looked up and set it to `tokenRevocationPending == -1`, but delayed the actual revoke call. This means that:

```
    // This is a token that is awaiting deferred revocation or tainted
    if entry.NumUses < 0 && !tainted {
        return nil, nil
    }
```

and so we'd probably need special logic to actually handle inline revocation. What's our thoughts on if its worthwhile? Kinda seems useful to be able to burn wrapping tokens in general this way.

cipherboy:
I filed an issue to resolve; will merge and work on that self-revocation aspect.

PR #3602, opened 2026-07-23
Use external keys in Transit

cipherboy:
## Description

I'll work on automated testing in a follow-up part.

<details>

<summary> Testing script </summary>

```bash
#!/bin/bash

set -euxo pipefail

( killall bao && sleep 1 ) || true

devbao node start --force --initialize --unseal --profiles secret,userpass,pki,transit --audit --seals static:// --ui

. <(devbao node env prod)

bao write transit/keys/pki-signing type=rsa-2048

bao write sys/external-keys/configs/transit plugin=transit token=$VAULT_TOKEN address=$VAULT_ADDR mount_path=transit
bao write sys/external-keys/configs/transit/keys/key name=auto-unseal version=1
bao write -f sys/external-keys/configs/transit/keys/key/grants/transit

bao write sys/external-keys/configs/transit/keys/signing-key name=pki-signing version=1
bao write -f sys/external-keys/configs/transit/keys/signing-key/grants/transit

bao write transit/keys/enc-external external_key_name=transit:key type=external_key
bao write transit/encrypt/enc-external plaintext=asdf
bao write transit/decrypt/auto-unseal ciphertext="$(bao write -field=ciphertext transit/encrypt/enc-external plaintext=asdf associated_data=asdf)" associated_data=asdf
bao write transit/decrypt/enc-external ciphertext="$(bao write -field=ciphertext transit/encrypt/enc-external plaintext=asdf associated_data=asdf)" associated_data=asdf
ciphertext_one="$(bao write -field=ciphertext transit/encrypt/enc-external plaintext=asdf associated_data=asdf)"

bao write transit/keys/sig-external external_key_name=transit:signing-key type=external_key
bao write transit/sign/sig-external input=asdf
bao write transit/verify/pki-signing signature="$(bao write -field=signature transit/sign/sig-external input=asdf)" input="asdf"
bao write transit/verify/sig-external signature="$(bao write -field=signature transit/sign/sig-external input=asdf)" input="asdf"

bao write -f transit/keys/auto-unseal/rotate
bao write sys/external-keys/configs/transit/keys/key-2 name=auto-unseal version=2
bao write -f sys/external-keys/configs/transit/keys/key-2/grants/transit

bao write transit/keys/enc-external/rotate external_key_name=transit:key-2
bao write transit/decrypt/enc-external ciphertext="$ciphertext_one" associated_data=asdf
bao write transit/decrypt/auto-unseal ciphertext="$ciphertext_one" associated_data=asdf

bao write transit/encrypt/enc-external plaintext=asdf
bao write transit/decrypt/auto-unseal ciphertext="$(bao write -field=ciphertext transit/encrypt/enc-external plaintext=asdf associated_data=asdf)" associated_data=asdf
bao write transit/decrypt/enc-external ciphertext="$(bao write -field=ciphertext transit/encrypt/enc-external plaintext=asdf associated_data=asdf)" associated_data=asdf
```

</details>

## Rationale

See also: #1320

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/builtin/logical/transit/path_keys.go
+++ b/internal/builtin/logical/transit/path_keys.go
@@ -136,6 +136,12 @@ key.`,
 				Default:     0,
 				Description: fmt.Sprintf("The key size in bytes for the algorithm.  Only applies to HMAC and must be no fewer than %d bytes and no more than %d", keysutil.HmacMinKeySize, keysutil.HmacMaxKeySize),
 			},
+			"external_key_ref": {
+				Type: framework.TypeString,
+				Description: `Reference to the external key to use. This follows
+the format <config name>:<key name>, uniquely identifying the key configured
+under sys/external-keys/configs/<config name>/keys/<key name>.`,
+			},
 		},
 
 		Operations: map[logical.Operation]framework.OperationHandler{
@@ -246,6 +252,7 @@ func (b *backend) pathPolicyWrite(ctx context.Context, req *logical.Request, d *
 	keyType := d.Get("type").(string)
 	keySize := d.Get("key_size").(int)
 	exportable := d.Get("exportable").(bool)
+	externalKeyRef := d.Get("external_key_ref").(string)
 	allowPlaintextBackup := d.Get("allow_plaintext_backup").(bool)
 	autoRotatePeriod := time.Second * time.Duration(d.Get("auto_rotate_period").(int))
 
@@ -266,6 +273,7 @@ func (b *backend) pathPolicyWrite(ctx context.Context, req *logical.Request, d *
 		Exportable:           exportable,
 		AllowPlaintextBackup: allowPlaintextBackup,
 		AutoRotatePeriod:     autoRotatePeriod,
+		ExternalKeyRef:       externalKeyRef,
 	}
 
 	switch keyType {
@@ -293,17 +301,39 @@ func (b *backend) pathPolicyWrite(ctx context.Context, req *logical.Request, d *
 		polReq.KeyType = keysutil.KeyType_RSA4096
 	case "hmac":
 		polReq.KeyType = keysutil.KeyType_HMAC
+	case "external-key":
+		polReq.KeyType = keysutil.KeyType_ExternalKey
 	default:
 		return logical.ErrorResponse("unknown key type %v", keyType), logical.ErrInvalidRequest
 	}
 	if keySize != 0 {
-		if polReq.KeyType != keysutil.KeyType_HMAC {
+		if polReq.KeyType != keysutil.KeyType_HMAC && polReq.KeyType != keysutil.KeyType_ExternalKey {
 			return logical.ErrorResponse("key_size is not valid for algorithm %v", polReq.KeyType), logical.ErrInvalidRequest
 		}
 		if keySize < keysutil.HmacMinKeySize || keySize > keysutil.HmacMaxKeySize {
 			return logical.ErrorResponse("invalid key_size %d", keySize), logical.ErrInvalidRequest
 		}
 		polReq.KeySize = keySize
+	}
+	if polReq.KeyType == keysutil.KeyType_ExternalKey {
+		if polReq.ExternalKeyRef == "" {
+			return logical.ErrorResponse("must provide external_key_ref to create policy of type external-key"), logical.ErrInvalidRequest
+		}
+
+		if autoRotatePeriod != 0 {
+			return logical.ErrorResponse("cannot enable auto-rotation with external keys"), nil
+		}
+
+		key, err := b.System().GetExternalKey(ctx, polReq.ExternalKeyRef)
+		if err != nil {
+			return logical.ErrorResponse("unknown external key reference: %v", polReq.ExternalKeyRef), logical.ErrInvalidRequest
+		}
+
+		if err := key.Close(ctx); err != nil {
+			return nil, fmt.Errorf("unable to close external key: %w", err)
+		}
+	} else if polReq.ExternalKeyRef != "" {
+		return logical.ErrorResponse("must provide type=external-key to use external_key_ref"), logical.ErrInvalidRequest
 	}
 
 	p, upserted, err := b.GetPolicy(ctx, polReq, b.GetRandomReader())
@@ -494,6 +524,12 @@ func (b *backend) formatKeyPolicy(p *keysutil.Policy, context []byte) (*logical.
 			}
 
 			retKeys[k] = structtomap.Map(key)
+		}
+		resp.Data["keys"] = retKeys
+	case keysutil.KeyType_ExternalKey:
+		retKeys := map[string]string{}
+		for k, v := range p.Keys {
+			retKeys[k] = v.ExternalKeyRef
 		}
 		resp.Data["keys"] = retKeys
 	}
--- a/internal/builtin/logical/transit/path_rotate.go
+++ b/internal/builtin/logical/transit/path_rotate.go
@@ -5,6 +5,7 @@ package transit
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/openbao/openbao/sdk/v2/framework"
 	"github.com/openbao/openbao/sdk/v2/helper/keysutil"
@@ -25,6 +26,12 @@ func (b *backend) pathRotate() *framework.Path {
 			"name": {
 				Type:        framework.TypeString,
 				Description: "Name of the key",
+			},
+			"external_key_ref": {
+				Type: framework.TypeString,
+				Description: `Reference to the external key to use. This follows
+the format <config name>:<key name>, uniquely identifying the key configured
+under sys/external-keys/configs/<config name>/keys/<key name>.`,
 			},
 		},
 
@@ -60,6 +67,26 @@ func (b *backend) pathRotateWrite(ctx context.Context, req *logical.Request, d *
 		return logical.ErrorResponse("key not found"), logical.ErrInvalidRequest
 	}
 	defer p.Unlock()
+
+	externalKeyRef := d.Get("external_key_ref").(string)
+	if p.Type == keysutil.KeyType_ExternalKey {
+		if externalKeyRef == "" {
+			return logical.ErrorResponse("must provide external_key_ref to rotate policy of type external-key"), logical.ErrInvalidRequest
+		}
+
+		key, err := b.System().GetExternalKey(ctx, externalKeyRef)
+		if err != nil {
+			return logical.ErrorResponse("unknown external key reference: %v", externalKeyRef), logical.ErrInvalidRequest
+		}
+
+		if err := key.Close(ctx); err != nil {
+			return nil, fmt.Errorf("unable to close external key: %w", err)
+		}
+
+		p.ExternalKeyRef = externalKeyRef
+	} else if externalKeyRef != "" {
+		return logical.ErrorResponse("cannot use external_key_ref on key of type %v", p.Type.String()), logical.ErrInvalidRequest
+	}
 
 	// Rotate the policy
 	err = p.Rotate(ctx, req.Storage, b.GetRandomReader())
--- a/internal/vault/external_keys/registry.go
+++ b/internal/vault/external_keys/registry.go
@@ -447,6 +447,10 @@ func (r *Registry) ModifyKey(ctx context.Context, s logical.Storage, configName,
 			return err
 		}
 
+		if key == nil {
+			return errMissingKey
+		}
+
 		// Unlike KMS clients, keys aren't cached so just close it again.
 		if err := key.Close(ctx); err != nil {
 			r.logger.Error("failed to close key", "error", err.Error(), "config", configName, "key", keyName, "namespace", ns.Path)
@@ -507,6 +511,11 @@ func (r *Registry) GetExternalKey(ctx context.Context, s logical.Storage, ns *na
 		return nil, err
 	}
 
+	// Better to err than pass along or wrap a nil key.
+	if key == nil {
+		return nil, errMissingKey
+	}
+
 	return nilCheckingKey{key}, nil
 }
 
@@ -516,6 +525,7 @@ var (
 	// against names passed in by a plugin via the system view.
 	namePattern   = regexp.MustCompile("^" + framework.GenericNameRegex("name") + "$")
 	errInvalidRef = errors.New(`invalid key reference: must be "<config name>:<key name>"`)
+	errMissingKey = errors.New(`plugin returned nil key; report to OpenBao authors`)
 )
 
 // ParseRef parses a key reference into config name and key name.
--- a/sdk/helper/keysutil/lock_manager.go
+++ b/sdk/helper/keysutil/lock_manager.go
@@ -60,8 +60,8 @@ type PolicyRequest struct {
 	// Indicates whether a private or public key is imported/upserted
 	IsPrivateKey bool
 
-	// The UUID of the managed key, if using one
-	ManagedKeyUUID string
+	// The reference to the external key, if using one.
+	ExternalKeyRef string
 }
 
 type LockManager struct {
@@ -390,6 +390,15 @@ func (lm *LockManager) GetPolicyWithLockType(ctx context.Context, req PolicyRequ
 				return nil, false, fmt.Errorf("key derivation and convergent encryption not supported for keys of type %v", req.KeyType)
 			}
 
+		case KeyType_ExternalKey:
+			if req.Derived || req.Convergent {
+				return nil, false, fmt.Errorf("key derivation and convergent encryption not supported for keys of type %v", req.KeyType)
+			}
+
+			if req.AutoRotatePeriod != 0 {
+				return nil, false, fmt.Errorf("auto-rotation is not supported for keys of type %v", req.KeyType)
+			}
+
 		default:
 			return nil, false, fmt.Errorf("unsupported key type %v", req.KeyType)
 		}
@@ -403,6 +412,7 @@ func (lm *LockManager) GetPolicyWithLockType(ctx context.Context, req PolicyRequ
 			AllowPlaintextBackup: req.AllowPlaintextBackup,
 			AutoRotatePeriod:     req.AutoRotatePeriod,
 			KeySize:              req.KeySize,
+			ExternalKeyRef:       req.ExternalKeyRef,
 		}
 
 		if req.Derived {
--- a/sdk/helper/keysutil/policy.go
+++ b/sdk/helper/keysutil/policy.go
@@ -38,6 +38,7 @@ import (
 	"golang.org/x/crypto/hkdf"
 
 	"github.com/hashicorp/go-uuid"
+	"github.com/openbao/go-kms-wrapping/v2/kms"
 	"github.com/openbao/openbao/sdk/v2/helper/certutil"
 	"github.com/openbao/openbao/sdk/v2/helper/errutil"
 	"github.com/openbao/openbao/sdk/v2/helper/jsonutil"
@@ -71,6 +72,9 @@ const (
 	KeyType_RSA3072
 	KeyType_HMAC
 	KeyType_XChaCha20_Poly1305
+
+	// External keys is a meta-type.
+	KeyType_ExternalKey = 10000
 )
 
 const (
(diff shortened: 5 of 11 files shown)


internal/builtin/logical/transit/path_keys.go
```diff
@@ -496,6 +510,12 @@ func (b *backend) formatKeyPolicy(p *keysutil.Policy, context []byte) (*logical.
 			retKeys[k] = structtomap.Map(key)
 		}
 		resp.Data["keys"] = retKeys
+	case keysutil.KeyType_ExternalKey:
+		retKeys := map[string]string{}
+		for k, v := range p.Keys {
+			retKeys[k] = v.ExternalKeyName
```
cipherboy: I'm debating if I want to cache public keys on the `KeyEntry`? 🤔 That would let us respond to public-key only operations without round-tripping to the KMS (export, ...), at the cost of one extra operation on setup.

Thoughts?

sdk/helper/keysutil/policy.go
```diff
@@ -86,14 +88,14 @@ const (
 	DefaultVersionTemplate = "vault:v{{version}}:"
 )
 
-type AEADFactory interface {
```
cipherboy, line 89: I believe this was used for managed keys.

sdk/helper/keysutil/policy.go
```diff
@@ -71,6 +72,7 @@ const (
 	KeyType_RSA3072
 	KeyType_HMAC
 	KeyType_XChaCha20_Poly1305
+	KeyType_ExternalKey
```
cipherboy: Do we care about potential SDK integer value conflicts with Vault?

sdk/helper/keysutil/policy.go
```diff
@@ -2172,7 +2314,55 @@ func (p *Policy) EncryptWithFactory(ver int, context []byte, nonce []byte, value
...
+		var key kms.Key
+		var aad []byte
+		for index, factory := range factories {
+			if factory == nil {
+				continue
+			}
+
+			switch typed := factory.(type) {
+			case ExternalKeyFactory:
+				ctx, key, err = typed.GetExternalKey(keyEntry.ExternalKeyName)
```
satoqz: You'll need to make sure to call `.Close()` on keys.

internal/builtin/logical/transit/path_keys.go
cipherboy: This gets rather difficult to accomplish, passing the registry down to `RotateInMemory` given how few parameters that takes. 

We could fetch the public keys here and return them, though, that slows down the fetch endpoint.

Thoughts?

sdk/helper/keysutil/policy.go
satoqz: I don't know, though it would make sense to give this one a higher value (1000?) such that it can remain sorted towards the bottom as a special type. Otherwise, we'll have future key types such as ML-DSA listed below external key.

sdk/helper/keysutil/policy.go
```diff
@@ -124,31 +127,31 @@ type KeyType int
...
+	case KeyType_ECDSA_P256, KeyType_ECDSA_P384, KeyType_ECDSA_P521, KeyType_ED25519, KeyType_RSA2048, KeyType_RSA3072, KeyType_RSA4096, KeyType_ExternalKey:
 		return true
 	}
 	return false
 }
 
 func (kt KeyType) HashSignatureInput() bool {
 	switch kt {
-	case KeyType_ECDSA_P256, KeyType_ECDSA_P384, KeyType_ECDSA_P521, KeyType_RSA2048, KeyType_RSA3072, KeyType_RSA4096:
+	case KeyType_ECDSA_P256, KeyType_ECDSA_P384, KeyType_ECDSA_P521, KeyType_RSA2048, KeyType_RSA3072, KeyType_RSA4096, KeyType_ExternalKey:
```
satoqz, line 156: Should it be (assuming I understand this option correctly)?

KMS implementations are encouraged to perform the hashing themselves if they can. See for example the `disable_prehashing` option on Transit keys.

If we preemptively hash here, won't keys with `disable_prehashing = true` always reject our input?

cipherboy: @satoqz The more I look at it, the more confused I get. :laughing: 

The benefit of this option is that you always know you're getting prehashed data, at the keysutil layer. No signature algorithm (other than ed25519, which doesn't do hashing) has hashes specified here... 

If you don't set `p.Type.HashSignatureInput() == true` on external keys,  `prehashed` is not specified down to the keysutil layer, so you have no clue if you are or are not prehashed. 

This would mean that you don't know, going into the KMS, if it is prehashed or not. That's a lot more plumbing to add to support non-prehashed stuff. Thoughts?

satoqz: @cipherboy we can defer this to a TODO and document as a limitation. But long-term, it'd be nice to do that plumbing.

internal/vault/external_keys/registry.go
```diff
@@ -507,6 +507,12 @@ func (r *Registry) GetExternalKey(ctx context.Context, s logical.Storage, ns *na
 		return nil, err
 	}
 
+	// Better to pass along the nil, nil as a 404 signal than mask it with a
+	// nilCheckingKey reference.
+	if key == nil {
+		return nil, nil
+	}
```
satoqz: IMO this should instead error right here at latest. Otherwise all the poor engines must repeat the check.

satoqz: Also needs a check in `ModifyKey` to further defend against barbarians.

sdk/helper/keysutil/lock_manager.go
```diff
@@ -60,8 +60,8 @@ type PolicyRequest struct {
 	// Indicates whether a private or public key is imported/upserted
 	IsPrivateKey bool
 
-	// The UUID of the managed key, if using one
-	ManagedKeyUUID string
+	// The reference to the external key, if using one.
+	ExternalKeyName string
```
satoqz: Should we stick with `ExternalKeyRef`, `external_key_ref` over name? I named it "ref" to avoid mistaking it for a key name only, excluding the config name.

satoqz: "spec" also comes to mind if we don't like "ref".

cipherboy: @satoqz That's fine, I had done `external_key_name` to align with `managed_key_name`, but I suppose users will already have to update even if they're using names (and definitely will if UUIDs).

sdk/helper/keysutil/policy.go
```diff
@@ -212,6 +217,8 @@ func (kt KeyType) String() string {
 		return "rsa-4096"
 	case KeyType_HMAC:
 		return "hmac"
+	case KeyType_ExternalKey:
+		return "external_key"
```
satoqz:
```suggestion
		return "external-key"
```

internal/builtin/logical/transit/path_rotate.go
```diff
@@ -61,6 +67,14 @@ func (b *backend) pathRotateWrite(ctx context.Context, req *logical.Request, d *
 	}
 	defer p.Unlock()
 
+	externalKeyName := d.Get("external_key_name").(string)
+	if p.Type == keysutil.KeyType_ExternalKey {
+		if externalKeyName == "" {
```
satoqz:
```suggestion
	if p.Type == keysutil.KeyType_ExternalKey {
		externalKeyName := d.Get("external_key_name").(string)
		if externalKeyName == "" {
```

internal/builtin/logical/transit/path_keys.go
```diff
@@ -246,13 +252,18 @@ func (b *backend) pathPolicyWrite(ctx context.Context, req *logical.Request, d *
...
 	allowPlaintextBackup := d.Get("allow_plaintext_backup").(bool)
 	autoRotatePeriod := time.Second * time.Duration(d.Get("auto_rotate_period").(int))
 
 	if autoRotatePeriod != 0 && autoRotatePeriod < time.Hour {
 		return logical.ErrorResponse("auto rotate period must be 0 to disable or at least an hour"), nil
 	}
 
+	if autoRotatePeriod != 0 && externalKeyName != "" {
+		return logical.ErrorResponse("cannot enable auto-rotation with external keys"), nil
+	}
```
satoqz: I see we also have `pathKeysConfigWrite` and `pathImportWrite`, `pathImportVersionWrite`, are we protected against those doing nonsensical things?

internal/builtin/logical/transit/path_keys.go
satoqz: I fear that some users will abuse (read: use not as @cipherboy intended) the mapping of transit keys to external keys and update an external key directly without updating the transit version, then be surprised by a cached public key they can't override without backtracking.

sdk/helper/keysutil/policy.go
```diff
@@ -1501,6 +1603,45 @@ func (p *Policy) VerifySignatureWithOptions(context, input []byte, sig string, o
...
+		opts := &kms.VerifyOptions{
+			Data:       input,
+			Signature:  sigBytes,
+			SignerOpts: algo,
+			Prehashed:  true,
+		}
+
+		err = key.Verify(ctx, opts)
+		if err != nil {
+			return false, fmt.Errorf("call to kms.Verify failed: %w", err)
```
satoqz:
```suggestion
			return false, fmt.Errorf("call to key.Verify failed: %w", err)
```

sdk/helper/keysutil/policy.go
```diff
@@ -1097,6 +1111,58 @@ func (p *Policy) DecryptWithFactory(context, nonce []byte, value string, factori
...
+		}
+
+		opts := &kms.CipherOptions{
+			Data: decoded,
+			AAD:  aad,
+		}
+
+		plain, err = key.Decrypt(ctx, opts)
+		if err != nil {
+			return "", fmt.Errorf("call to kms.Decrypt failed: %w", err)
```
satoqz:
```suggestion
			return "", fmt.Errorf("call to key.Decrypt failed: %w", err)
```

sdk/helper/keysutil/policy.go
```diff
@@ -1307,6 +1373,42 @@ func (p *Policy) SignWithOptions(ver int, context, input []byte, options *Signin
...
+
+		opts := &kms.SignOptions{
+			Data:       input,
+			SignerOpts: algo,
+			Prehashed:  true,
+		}
+
+		sig, err = key.Sign(ctx, opts)
+		if err != nil {
+			return nil, fmt.Errorf("call to kms.Sign failed: %w", err)
```
satoqz:
```suggestion
			return nil, fmt.Errorf("call to key.Sign failed: %w", err)
```

sdk/helper/keysutil/policy.go
```diff
@@ -2172,7 +2327,62 @@ func (p *Policy) EncryptWithFactory(ver int, context []byte, nonce []byte, value
...
+		}
+
+		opts := &kms.CipherOptions{
+			Data: plaintext,
+			AAD:  aad,
+		}
+
+		ciphertext, err = key.Encrypt(ctx, opts)
+		if err != nil {
+			return "", fmt.Errorf("call to kms.Encrypt failed: %w", err)
```
satoqz:
```suggestion
			return "", fmt.Errorf("call to key.Encrypt failed: %w", err)
```

sdk/helper/keysutil/policy.go
satoqz: Though I wonder if you'd prefer to make it `failed to decrypt via external key: ...` instead.

sdk/helper/keysutil/policy.go
```diff
@@ -2172,7 +2327,62 @@ func (p *Policy) EncryptWithFactory(ver int, context []byte, nonce []byte, value
...
+
+		var ctx context.Context
+		var key kms.Key
+		var aad []byte
+
+		defer func() {
+			if key != nil {
+				key.Close(ctx) //nolint:errcheck
+			}
+		}()
```
satoqz: You could move this below the initial nil check so you don't need to nil check again.

satoqz: (Or preferably, below the err check, because we shouldn't need to nil check here if the registry is nice.)

cipherboy: @satoqz Can I? If I have a failed AAD factory in the second slot, couldn't I have a stray key ref that I need to close?

internal/vault/external_keys/registry.go
satoqz: I guess checking in the registry also does not help if any wrappers (key in plugin client/server) don't nil check.

Overall I'd say that this is up to the KMS plugin author to get right. If they return `nil, nil`, that's a violation of the interface, and a crash is adequate like it would be for any other programming error.

internal/builtin/logical/transit/path_keys.go
cipherboy: @satoqz We could use periodic func and tuning to re-validate keys match I suppose... and mark deleted KMS keys as soft-deleted. lol

sdk/helper/keysutil/policy.go
satoqz: @cipherboy I'd defer within the loop, right after calling the external key factory. not after the loop.

satoqz: If the linter makes that hard, fine, smh.

internal/vault/external_keys/registry.go
cipherboy: @satoqz This is fair, though I think with us externalizing KMS interfaces in both directions (to other secret engine authors and to other KMS authors), we're no longer controlling them all and should defend a little more than otherwise. But I concur with the sentiment in general.

sdk/helper/keysutil/policy.go
cipherboy: I've fixed it but I will just note at the `fork-point`:

https://github.com/openbao/openbao/blob/8993802145833ab01d49c6070d787a9eccb81546/sdk/helper/keysutil/policy.go#L205-L206

:shrug:

internal/builtin/logical/transit/path_keys.go
cipherboy: I fixed `pathKeysConfigWrite`, I think `pathImportWrite` is protected by virtue of not allowing external keys, and `pathImportVersionWrite` only allowing importing to imported keys.

sdk/helper/keysutil/policy.go
cipherboy: I changed it to:

> `call to Encrypt with external key failed`

&c.

sdk/helper/keysutil/policy.go
```diff
@@ -2172,7 +2323,58 @@ func (p *Policy) EncryptWithFactory(ver int, context []byte, nonce []byte, value
...
+			switch typed := factory.(type) {
+			case ExternalKeyFactory:
+				ctx, key, err = typed.GetExternalKey(keyEntry.ExternalKeyRef)
+				if err != nil {
+					return "", fmt.Errorf("factory[%d] failed to fetch external key: %w", index, err)
+				} else if key == nil {
+					return "", fmt.Errorf("factory[%d] returned nil key with no error; key not found", index)
+				}
+
+				key.Close(ctx) //nolint:errcheck
```
satoqz:
```suggestion
				defer key.Close(ctx) //nolint:errcheck
```

sdk/helper/keysutil/policy.go
```diff
@@ -1097,6 +1111,54 @@ func (p *Policy) DecryptWithFactory(context, nonce []byte, value string, factori
...
```
satoqz:
```suggestion
				defer key.Close(ctx) //nolint:errcheck
```

You wouldn't catch these when testing against built-in Transit. I'm testing with PKCS#11 and there are a couple more issues I'll triage and see what is wrong where.

internal/builtin/logical/transit/path_keys.go
```diff
@@ -293,11 +305,13 @@ func (b *backend) pathPolicyWrite(ctx context.Context, req *logical.Request, d *
 		polReq.KeyType = keysutil.KeyType_RSA4096
 	case "hmac":
 		polReq.KeyType = keysutil.KeyType_HMAC
+	case "external_key":
```
satoqz:
```suggestion
	case "external-key":
```

internal/builtin/logical/transit/path_keys.go
satoqz: I suggest we just keep it simple for now.

internal/builtin/logical/transit/path_keys.go
satoqz: Here (and in rotate) we should perhaps run an initial `GetKey()` to ensure that the key exists in the registry? Mostly, that would've saved me a couple typos.

sdk/helper/keysutil/policy.go
cipherboy: Ahh, that's why it passed the linter lol :laughing:

internal/builtin/logical/transit/path_keys.go
cipherboy: Ah, do it on the Transit side rather than the Keysutil side. That makes sense.

sdk/helper/keysutil/policy.go
cipherboy: Linter seems happy.

internal/vault/external_keys/registry.go
```diff
@@ -516,6 +525,7 @@ var (
 	// against names passed in by a plugin via the system view.
 	namePattern   = regexp.MustCompile("^" + framework.GenericNameRegex("name") + "$")
 	errInvalidRef = errors.New(`invalid key reference: must be "<config name>:<key name>"`)
+	errMissingKey = errors.New(`external key not found`)
```
satoqz: Should it be this opaque? We only hit it in two cases:

1. There's a bug in the gRPC-based plugin client
2. There's a bug in a builtin plugin

You wouldn't catch nil keys returned by an external plugin here, those would instead cause a panic within the plugin process.

I think 1 is very easy to audit, and looks fine.

2 would be the real problem to defend against since it would cause a panic in the main process. So I'd suggest a more debug-y message closer to: "plugin returned nil key, please report a bug?".

internal/builtin/logical/transit/path_rotate.go
```diff
@@ -61,6 +68,24 @@ func (b *backend) pathRotateWrite(ctx context.Context, req *logical.Request, d *
 	}
 	defer p.Unlock()
 
+	if p.Type == keysutil.KeyType_ExternalKey {
+		externalKeyRef := d.Get("external_key_ref").(string)
+		if externalKeyRef == "" {
+			return logical.ErrorResponse("must provide external_key_ref to rotate policy of type external key"), logical.ErrInvalidRequest
```
satoqz:
```suggestion
			return logical.ErrorResponse("must provide external_key_ref to rotate key of type external-key"), logical.ErrInvalidRequest
```

Would be more transit-y perhaps?

internal/builtin/logical/transit/path_keys.go
```diff
@@ -246,13 +252,18 @@ func (b *backend) pathPolicyWrite(ctx context.Context, req *logical.Request, d *
...
 	allowPlaintextBackup := d.Get("allow_plaintext_backup").(bool)
 	autoRotatePeriod := time.Second * time.Duration(d.Get("auto_rotate_period").(int))
 
 	if autoRotatePeriod != 0 && autoRotatePeriod < time.Hour {
 		return logical.ErrorResponse("auto rotate period must be 0 to disable or at least an hour"), nil
 	}
 
+	if autoRotatePeriod != 0 && externalKeyRef != "" {
+		return logical.ErrorResponse("cannot enable auto-rotation with external keys"), nil
+	}
```
satoqz: I'd move this check lower, together with other checks starting from line 322.

satoqz (dismissed):
Some more nits, but looks good and works fine as far as I can tell.

internal/builtin/logical/transit/path_rotate.go
cipherboy: I added the slash but kept the `policy`. The nil check below uses `policy`; I think in general Transit is just a confused engine. :laughing:

internal/builtin/logical/transit/path_keys.go
cipherboy: Thanks! Also added a check here and in rotate for non-empty `external_key_ref`, just in case someone thought to send AES-GCM or something, but wondered why it failed.

satoqz (approved):

PR #3605, opened 2026-07-24
Redact sensitive values when reading external key configuration

satoqz:
## Description

TL;DR:

```console
$ bao write sys/external-keys/configs/softhsm plugin=pkcs11 lib=softhsm token_label=openbao pin=1234
Success! Data written to: sys/external-keys/configs/softhsm

$ bao read sys/external-keys/configs/softhsm
Key            Value
---            -----
lib            softhsm
pin            (redacted)
plugin         pkcs11
token_label    openbao
```

## Rationale

See https://github.com/openbao/openbao/pull/3534#discussion_r3640855932.

## Acknowledgements

       also allowed.


internal/helper/kmsplugin/metadata.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+var builtinMetadata = map[string]plugin.Metadata{
+	"transit": {
+		SensitiveKMSFields: transit.SensitiveKMSFields,
+	},
+}
+
+// GetMetadata returns a KMS plugin's metadata.
+func (c *Catalog) GetMetadata(ctx context.Context, name string) (plugin.Metadata, error) {
```
satoqz, line 21: Not super convinced yet that the complexity of caching is worth it. I'd like to re-evaluate that once I get to reloading KMS plugins, in which case we'd need to invalidate such caches together with clients.

To put things into perspective, making a plugin request per call is not something extraordinarily expensive and we do it on all plugin-based request paths all the time. As long as there are no detailed listing paths in the external keys API that would require multiple metadata calls, I think we can get away with this.

internal/vault/external_keys/registry.go
```diff
@@ -157,6 +157,28 @@ func (r *Registry) ReadConfig(ctx context.Context, s logical.Storage, name strin
...
+// any sensitive configuration fields in the returned entry.
+func (r *Registry) ReadRedactedConfig(ctx context.Context, s logical.Storage, name string) (*ConfigEntry, error) {
+	ce, err := r.ReadConfig(ctx, s, name)
+	if ce == nil || ce.Values == nil || err != nil {
+		return ce, err
+	}
+
+	meta, err := r.plugins.GetMetadata(ctx, ce.Plugin)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read metadata for plugin %q: %w", ce.Plugin, err)
```
satoqz, line 170: If you've written a config with an invalid plugin via `verify=false` or simply removed a plugin, I guess you'll see this error when reading a key or config. It tells you the plugin type so you've got some clue about what's happening.

An alternative result would be not to error and return the plugin name in plaintext, all other values redacted as `***`, and add a warning? Thoughts?

wslabosz-reply (dismissed):
LGTM

internal/helper/kmsplugin/metadata.go
wslabosz-reply: seems like a reasonable approach

internal/vault/external_keys/registry.go
wslabosz-reply: > An alternative result would be not to error and return the plugin name in plaintext, all other values redacted as ***, and add a warning? Thoughts?

imho with invalid configuration of the plugin, it's more straightforward to err than respond with somewhat correct shape (albeit redacted) with a warning, maybe its more informative, depending on the warning we want to return, but a tad bit leaning towards err'ring.

cipherboy: I'm fine with erring I think. If anything, I might err on the other side and state that redaction is best-effort and fail open, tbh. This is still an API for privileged operators after all, and explicit cross-namespace grants handles a lot of sensitive config hiding. 

The benefit of caching is fewer opportunities to err (as we could potentially cache indefinitely with explicit invalidation on plugin reload/...), though I think that'd be unlikely except in scenarios defined above. That could technically yield a memory leak though I don't know that anyone would really register so many unique KMS plugins. 

My 2c.

cipherboy (dismissed):
My only outstanding comment would be if we wanted to use `(redacted)` rather than `***` for clarity, but that's my 2c. 

I do wonder how any response behaves w.r.t. Terraform and the like; I think other plugins elide certain parameters (though, parameters are known a priori) rather than respond with redacted versions. For such a dynamic API, I'd rather use `(redacted)` and condition around that in any potential Terraform code.

cipherboy (approved):
Thanks!

PR #3612, opened 2026-07-27
Move .go-version into toolchain directive

satoqz:
## Description

Remove `.go-version` and use the `toolchain` directive of `go.mod` instead.

## Rationale

Use features natively provided by Go where possible. Also gets rid of a top-level file.

## Acknowledgements

       also allowed.


wslabosz-reply (dismissed):

cipherboy:
@satoqz Correct me if I'm wrong, but isn't the reason why we have two files is so that the `go.mod` we ship builds with our _oldest_ supported toolchain and `.go-version` is always the _newest_ toolchain?

satoqz:
@cipherboy `go.mod` has both the `go` and the `toolchain` directives for the same reason.

From https://go.dev/ref/mod#go-mod-file-toolchain:

> A go directive indicates that a module was written assuming the semantics of a given version of Go. The version must be a valid [Go version](https://go.dev/doc/toolchain#version), such as 1.14, 1.21rc1, or 1.23.0.

and

> A toolchain directive declares a suggested Go toolchain to use with a module. The suggested Go toolchain’s version cannot be less than the required Go version declared in the go directive.

Notably, the `go` directive never referenced a toolchain. It instead references the _minimum version of the language_ which only coincides with official Go toolchain versioning.

The `toolchain` directive instead is a fixed-point version of a concrete Go toolchain to install. This is effectively our `/.go-version`, +/- auto toolchain installation if `GOTOOLCHAIN=auto`.

cipherboy:
> The `toolchain` directive instead is a fixed-point version of a concrete Go toolchain to install. This is effectively our `/.go-version`, +/- auto toolchain installation if `GOTOOLCHAIN=auto`.

It looks like Fedora uses `GOTOOLCHAIN=local` -- https://developer.fedoraproject.org/tech/languages/go/go-installation.html 

So I think we should be fine to drop it. We should add it as another breaking change, though, for packagers. I think it'll impact GitLab (fyi @fcat).

satoqz:
Yes, I would be rather surprised about any distributions that don't set `GOTOOLCHAIN=local` or enforce this otherwise. If they don't, their packaging process is broken and any issues are entirely not on us, IMHO.

[GitLab seems fine](https://gitlab.com/gitlab-org/build/CNG/-/blob/master/gitlab-openbao/Dockerfile.build.ubi).

satoqz:
@cipherboy I've added a changelog callout though as `release-note:change`.

satoqz:
Hmm: 

> For reproducibility, the go command writes its own toolchain name in a toolchain line any time it is updating the go version in the go.mod file (usually during go get).

I think this will be bad for backports via something like `scripts/backport-deps.sh`, e.g., if the release branch still builds with an earlier minor version.

Closing, all in all I've convinced myself that it's not worth it. It would be nice to simply move the metadata into one file, but seems the Go team decided to add unnecessary baggage to the directive.

PR #3615, opened 2026-07-28
Allow hiding versions from unauthenticated users

### LINKED ISSUE #2509: Allow hiding versions from unauthenticated users
Currently, unauthenticated endpoints in openbao such as /sys/seal-status leak the version of OpenBao used. It'd be nice for security reasons to allow masking this for unauthenticated requests. 

An argument can be made that this is ultimately security by obscurity, and while true, gathering information such as versions used can be incredibly helpful when attacking a system.

**Alternatives you've considered**
Allow overriding the version set by goreleaser at runtime using an environment variable to achieve the same type of masking

Allow a configuration parameter that, when set, censors API versions under certain conditions such as unauthenticated requests

**Additional context**
Likely starting points:
- `version/version.go`

phil9909:
## Description

Draft: see https://github.com/openbao/openbao/issues/2509#issuecomment-5103692020

## Rationale

Resolves: #2509

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/logical_system.go
+++ b/internal/vault/logical_system.go
@@ -38,6 +38,7 @@ import (
 	"github.com/openbao/openbao/sdk/v2/helper/pluginutil"
 	"github.com/openbao/openbao/sdk/v2/helper/wrapping"
 	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/openbao/openbao/v2/internal/helper/configutil"
 	"github.com/openbao/openbao/v2/internal/helper/hostutil"
 	"github.com/openbao/openbao/v2/internal/helper/identity"
 	"github.com/openbao/openbao/v2/internal/helper/logging"
@@ -4293,8 +4294,16 @@ func (b *SystemBackend) pathInternalOpenAPI(ctx context.Context, req *logical.Re
 		return nil, err
 	}
 
+	v := version.Version
+	if req.ClientToken == "" {
+		listener := configutil.GetListenerConfigToContext(ctx)
+		if listener != nil && listener.DisableUnauthedMetadata {
+			v = "(version redacted)"
+		}
+	}
+
 	// Set up target document
-	doc := framework.NewOASDocument(version.Version)
+	doc := framework.NewOASDocument(v)
 
 	// Generic mount paths will primarily be used for code generation purposes.
 	// This will result in parameterized mount paths being returned instead of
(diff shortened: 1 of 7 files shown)


internal/vault/logical_system.go
```diff
@@ -4293,8 +4294,16 @@ func (b *SystemBackend) pathInternalOpenAPI(ctx context.Context, req *logical.Re
...
+	if req.ClientToken == "" {
+		listener := configutil.GetListenerConfigToContext(ctx)
+		if listener != nil && listener.DisableUnauthedMetadata {
+			v = "(version redacted)"
+		}
+	}
+
 	// Set up target document
-	doc := framework.NewOASDocument(version.Version)
+	doc := framework.NewOASDocument(v)
```
timb-controlplane, line 4306: This is OpenAPI, right?

I can imagine this breaking clients, eg clients that want to ask OpenBao questions whilst they negotiate an authn method. If so, is there a better way to mitigate that risk?

phil9909: > This is OpenAPI, right?

Yes, this is the part that generates the OpenAPI spec.

> I can imagine this breaking clients, eg clients that want to ask OpenBao questions whilst they negotiate an authn method.

I'd be surprised, if this breaks any clients that dynamically change their behavior based on the version reported by our OpenAPI spec. What I expect is some scripts might break which check a successful rollout, something like:

```bash
helm upgrade ...

while [[ $(bao read --field=version sys/health) != $NEW_VERSION ]] ; do
    sleep 1
done
```

But lets assume it does break some clients: this is opt-in per listener behavior. If it breaks you: don't enable it. Or, if you absolutely want to hide the version, then you will have to change your clients (or maybe you can switch the problematic clients to a VPN-only or mTLS listener and enable version hiding only on the public one).

> If so, is there a better way to mitigate that risk?

I don't think there is any other solution, either the version is public or it isn't.

timb-controlplane: Client might assume that the version should be a SemVer. Perhaps the mitigation is to document the consequence. An admin who still opts in has a strange version string in their OpenAPI, but at least we'd've warned them.

phil9909: > Client might assume that the version should be a SemVer.

Now I get it :laughing:. I on purpose did not redact it with a hard-coded SemVer. Sure we could - if version hiding is enabled - simply return something like `2.0.0-version-redacted`, which would be a valid SemVer. But if someone actually does parse the version to use if for some form of "negotiation" it would be better to fail with an "invalid semver" error than to continue with wrong assumptions, right?

> Perhaps the mitigation is to document the consequence. An admin who still opts in has a strange version string in their OpenAPI, but at least we'd've warned them.

Sure, this needs some documentation. But lets agree on the details first, otherwise I will have to write the documentation multiple times :D

PR #3619, opened 2026-07-28
Add semgrep rule to enforce testify usage for error checking

phil9909:
## Description

enforce that we don't do this

```go
if err != nil {
	t.Fatal(err)
}
```

but use

```go
require.NoError(t, err)
```

instead via semgrep rule.

## Rationale

especially in longer tests, this helps with readability by allowing more of the test to be on the screen at the same time (without saying shorter is always better for readability :laughing:)

## Acknowledgements

       also allowed.

### DIFF
--- a/tools/semgrep/ci/tests-assertions.yaml
+++ b/tools/semgrep/ci/tests-assertions.yaml
@@ -1,9 +1,7 @@
 # Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
 # SPDX-License-Identifier: MPL-2.0
 
-rules:
-  - id: use-testify-for-error-checks
-    
+__metadata: &metadata
     metadata: 
       category: best-practice
     languages:
@@ -11,7 +9,38 @@ rules:
     severity: ERROR
     message: "use github.com/stretchr/testify for test assertions"
 
+rules:
+  - id: use-testify-for-error-checks
+    <<: *metadata
+
     patterns:
-      - pattern: 'if $ERR != nil { $T.Fatal(err) }'
+      - pattern-not: &with-generic-init |
+          if $INIT; $ERR != nil {
+            ($T: *testing.T).Fatal($ERR)
+          }
+      - pattern: |
+          if $ERR != nil {
+            ($T: *testing.T).Fatal($ERR)
+          }
 
     fix: require.NoError($T, $ERR)
+
+  - id: use-testify-for-error-checks-with-simple-init
+    <<: *metadata
+
+    patterns:
+      - pattern: &with-simple-init |
+          if $ERR = $X($...ARGS); $ERR != nil {
+            ($T: *testing.T).Fatal($ERR)
+          }
+
+    fix: require.NoError($T, $X($...ARGS))
+
+  - id: use-testify-for-error-checks-with-generic-init
+    <<: *metadata
+
+    patterns:
+      - pattern: *with-generic-init
+      - pattern-not: *with-simple-init
+
+    fix: "$INIT\n\trequire.NoError($T, $ERR)"
--- a/tools/semgrep/ci/tests-assertions.yaml
+++ b/tools/semgrep/ci/tests-assertions.yaml
@@ -1,0 +1,17 @@
+# Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+# SPDX-License-Identifier: MPL-2.0
+
+rules:
+  - id: use-testify-for-error-checks
+    
+    metadata: 
+      category: best-practice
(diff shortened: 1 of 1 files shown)


tools/semgrep/ci/tests-assertions.yaml
satoqz, line 1: Should be restricted to `*_test.go` files, otherwise we'll flag things such as `log.Fatal`, for example. (Also saves some analysis time, I'd guess)

tools/semgrep/ci/tests-assertions.yaml
```diff
@@ -0,0 +1,17 @@
...
+    
+    metadata: 
+      category: best-practice
+    languages:
+     - go
+    severity: ERROR
+    message: "use github.com/stretchr/testify for test assertions"
+
+    patterns:
+      - pattern: 'if $ERR != nil { $T.Fatal(err) }'
```
cipherboy:
```suggestion
      - pattern: 'if $ERR != nil { $T.Fatal($ERR) }'
```

Really wish semgrep knew about type restrictions; I worry this will have a lot of false-positives. 

Did you also want to suggest `require.ErrorContains(...)` as a suggestion? Not sure if that's possible.

satoqz: > Really wish semgrep knew about type restrictions; I worry this will have a lot of false-positives.

Doesn't it? See `no-nil-check.yml`.

cipherboy: @satoqz Oh, so we should've used that format on `ci/logger-level-check.yml` to weed out some false-positives lol.

phil9909: So `($T: *testing.T)` should do the trick, but I will also prevent detecting helpers which take a `testing.TB` as argument.

tools/semgrep/ci/tests-assertions.yaml
phil9909: I tried this, but for some reason it just stopped applying the rule completely. Don't know why. I guess the rule should anyway also apply to test helpers.

I did check the current codebase and could not find any false positives. If we add the `($T: *testing.T)` constraint, false positives should be eliminated anyhow.

phil9909:
The fix part was broken, it turned this:

```go
if _, err := json.Marshal(42); err != nil {
	t.Fatal(err)
}
```

into

```go
require.NoError(t, err)
```

completely dropping the `json.Marhsal` call.

I have split it into three rules:

- `use-testify-for-error-checks` match without init statement
- `use-testify-for-error-checks-with-simple-init` turn this
  
  ```go
  if err := json.Unmarshal(x, &y); err != nil {
  	t.Fatal(err)
  }
  ```
  
  into

  ```
  require.NoError(t, json.Unmarshal(x, &y))
  ```

- `use-testify-for-error-checks-with-generic-init` turn this
  
  ```go
  if _, err := json.Marshal(42); err != nil {
  	t.Fatal(err)
  }
  ```
  
  into

  ```
  _, err := json.Marshal(42)
  require.NoError(t, json.Unmarshal(x, &y))
  ```

The final one might cause problems, if `err` was already defined in the scope.
Also, if `require` is not yet imported, it won't be added automatically.

These will have to be fixed manually.

tools/semgrep/ci/tests-assertions.yaml
satoqz: @phil9909 did you use something like `/**/*_test.go` or just `*_test.go` or `/*_test.go`?

satoqz:
@phil9909 WDYT about https://golangci-lint.run/docs/linters/configuration/#testifylint?

phil9909:
> @phil9909 WDYT about https://golangci-lint.run/docs/linters/configuration/#testifylint?

I used it in another project and found it useful. Mainly for catching `require.Equal(t, x, 42)` (which should be `require.Equal(t, 42, x)` because expected comes first and the error message will be misleading otherwise[^1]).

But AFAIK it does only ensure you are using testify correctly whenever you are using it. So it will not help in this particular case.

Might still be a good addition to our roster.

[^1]: in Ginkgo you write `Expect(x).To(Equal(42))` so it's much less likely to make the same mistake :shrug:

tools/semgrep/ci/tests-assertions.yaml
phil9909: `*_test.go` just like in the docs: https://docs.semgrep.dev/writing-rules/rule-syntax#limit-a-rule-to-paths

cipherboy: Where'd we land on this one?

PR #3622, opened 2026-07-28
Update “Developer quick start” guide

timb-controlplane:
## Description

Make changes to improve https://openbao.org/docs/get-started/developer-qs/

- Revise prerequisites
- Introduce `bao` as a command line tool (ie, not just as a server)
- Tweak the example code
- Improve the explanation

I previewed this locally and was happy with the changes.

## Rationale

At the time of writing, https://openbao.org/docs/get-started/developer-qs/ is misleading and a little unhelpful.

- it doesn't mention using `bao` to write, read or check secrets.
- the Golang code snippets feel incomplete
- you don't have a good way to follow the guide on a Mac with default security settings

Relevant to issue https://github.com/openbao/openbao/issues/3590, but doesn't resolve it.

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/get-started/developer-qs.mdx
+++ b/website/content/docs/get-started/developer-qs.mdx
@@ -6,203 +6,311 @@ import TabItem from '@theme/TabItem';
 
 # Developer quick start
 
-This quick start will explore how to use OpenBao client libraries inside your application code to store and retrieve your first secret value. OpenBao takes the security burden away from developers by providing a secure, centralized secret store for an application’s sensitive data: credentials, certificates, encryption keys, and more.
+This quick start explores how to use OpenBao client libraries inside your application code, to store and retrieve your first secret value. OpenBao takes the security burden away from developers by providing a secure, centralized secret store for an application’s sensitive data: credentials, certificates, encryption keys, and more.
+
+:::warning
+
+**Warning**: This guide explains how to practice fetching and using secrets, but the explanation relies on a in-memory _dev_ server.
+
+The _dev_ server configuration is useful for practicing with OpenBao locally while you learn, but is insecure and the data aren't durable. You should **never** use a _dev_ OpenBao instance to protect real secrets, or in any production context.
+
+:::
+
 
 ## Prerequisites
 
-- A development environment applicable to one of the languages in this quick start (currently **Go** and **Bash (curl)**)
+- A development environment applicable to one of the languages in this quick start
+  - You can use **Go**
+  - You can use POSIX shell (**bash**), with `curl`
+
+This page assumes that you have downloaded OpenBao and have it ready to run on your computer. You can find links to download OpenBao from the [Downloads](/downloads) page, along with basic installation instructions. Pick the option that is right for the computer where you want to run OpenBao.
+
+If you download the binary `.tar.gz` archive, you need to manually extract the `bao` program from the downloaded archive, and place it somewhere in your path.
+
+If you know how to run OpenBao in a container, you can start that container in `server` mode (see the next step), and then exec into that container using a command such as `docker exec` or `podman exec`. The downloads page lists available container images for OpenBao.
 
 ## Step 1: start OpenBao
-
-:::danger
-
-**Warning**: This in-memory “dev” server is useful for practicing with OpenBao locally for the first time, but is insecure and **should never be used in production**.
-
-:::
 
 Run the OpenBao server in a non-production "dev" mode:
 
 ```shell-session
-$ bao server -dev -dev-root-token-id="dev-only-token"
+$ bao server -dev -dev-root-token-id="example-tutorial-token"
 ```
 
-The `-dev-root-token-id` flag for dev servers tells the OpenBao server to allow full root access to anyone who presents a token with the specified value (in this case "dev-only-token").
+The `-dev-root-token-id` flag for dev servers tells the OpenBao server to allow full ("root") access to anyone who presents a token with the specified value (in this case, `example-tutorial-token`).
 
-:::danger
+:::warning
 
-**Warning**: The [root token](../concepts/tokens.mdx#root-tokens) is useful for development, but allows full access to all data and functionality of OpenBao, so it must be carefully guarded in production. Ideally, even an administrator of OpenBao would use their own token with limited privileges instead of the root token.
+**Warning**: The [root token](../concepts/tokens.mdx#root-tokens) is useful for development, but allows full access to all data and functionality of OpenBao. In a real cluster you would rotate the root token after initial setup and configure additional controls. Read [Authentication](../concepts/auth.mdx), [Identity](../concepts/identity.mdx) and [Policies](../concepts/policies.mdx) if you want to set up a secure OpenBao cluster for real use.
 
 :::
 
-OpenBao is now listening over HTTP on port **8200**. With all the setup out of the way, it's time to get coding!
+OpenBao is now listening over HTTP on port **8200**. With the initial setup out of the way, it's time to get coding!
 
-## Step 2: install a client library
+## Step 2: authenticate to OpenBao
 
-To read and write secrets in your application, you need to first configure a client to connect to OpenBao.
-Let's install the OpenBao client library for your language of choice.
+The `bao` program is more than the server; it's also a command line tool that you can use to access OpenBao. Imagine for a moment that you are doing IT operations work and you want to interact with an existing OpenBao cluster.
 
-<Tabs>
-<TabItem value="Go" group="go" heading="Go" group="go">
-
-[Go](https://pkg.go.dev/github.com/openbao/openbao/api) (official) client library:
+A variety of [authentication methods](../auth/index.mdx) can be used to prove your application's identity to the OpenBao server.  To keep things simple for the example, just use the (insecure) root token created in **Step 1**.
 
 ```shell-session
-$ go get github.com/openbao/openbao/api/v2
+$ export BAO_ADDR=http://127.0.0.1:8200/
+$ export BAO_TOKEN_PATH=/tmp/token-for-developer-quickstart
+$ bao login
 ```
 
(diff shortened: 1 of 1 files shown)


website/content/docs/get-started/developer-qs.mdx
```diff
@@ -6,203 +6,309 @@ import TabItem from '@theme/TabItem';
...
 
 ```shell-session
-$ bao server -dev -dev-root-token-id="dev-only-token"
+$ bao server -dev -dev-root-token-id="example-tutorial-token"
 ```
 
-The `-dev-root-token-id` flag for dev servers tells the OpenBao server to allow full root access to anyone who presents a token with the specified value (in this case "dev-only-token").
+The `-dev-root-token-id` flag for dev servers tells the OpenBao server to allow full ("root") access to anyone who presents a token with the specified value (in this case, `example-tutorial-token`).
 
 :::danger
```
cipherboy: Should this be a `warning` given the text below is warning?

website/content/docs/get-started/developer-qs.mdx
```diff
@@ -6,203 +6,309 @@ import TabItem from '@theme/TabItem';
 
 # Developer quick start
 
-This quick start will explore how to use OpenBao client libraries inside your application code to store and retrieve your first secret value. OpenBao takes the security burden away from developers by providing a secure, centralized secret store for an application’s sensitive data: credentials, certificates, encryption keys, and more.
+This quick start explores how to use OpenBao client libraries inside your application code, to store and retrieve your first secret value. OpenBao takes the security burden away from developers by providing a secure, centralized secret store for an application’s sensitive data: credentials, certificates, encryption keys, and more.
+
+:::danger
```
cipherboy: Same here, `:::warning` perhaps?

website/content/docs/get-started/developer-qs.mdx
```diff
@@ -6,203 +6,309 @@ import TabItem from '@theme/TabItem';
...
     log.Fatalf("unable to initialize OpenBao client: %v", err)
-}
+  }
 
-client.SetToken("dev-only-token")
+  // hard coding a token into source code is an INSECURE practice
+  //
+  // this is just an example; in real code you would use other means to let
+  // the client authenticate to OpenBao
+  client.SetToken("example-tutorial-token")
```
cipherboy, line 100: Not sure what your thoughts are on this, but you could use something like:

```
path := api.ReadBaoVariable(api.EnvTokenPath)
token, err := os.ReadFile(path)
...
```

if we wanted to avoid the anti-pattern of hard-coding tokens. Touch out of scope though.

website/content/docs/get-started/developer-qs.mdx
```diff
@@ -6,203 +6,309 @@ import TabItem from '@theme/TabItem';
...
+## Step 3: ensure that version 2 key/value storage is available
+
+Run the following command to make sure that the `secret/` mount point has been upgraded to version 2:
+
+```shell-session
+$ bao kv enable-versioning secret/
+```
+
+OpenBao supports version 1 and version 2 key/value (KV) secrets engines. You'll use KV2, which supports versioning (so that
+you can go back to the last version of a secret if you need to, for example).
```
cipherboy, line 123: Dev mode server is always KVv2 so this is unnecessary.

website/content/docs/get-started/developer-qs.mdx
```diff
@@ -6,203 +6,309 @@ import TabItem from '@theme/TabItem';
...
-    --header "Content-Type: application/json" \
-    --request POST \
-    --data '{"data": {"password": "OpenBao123"}}' \
-    http://127.0.0.1:8200/v1/secret/data/my-secret-password &&
-    echo "Secret written successfully."
+$ bao kv put -mount=secret developer-quickstart password=-
+```
+
+Because you specified that the password comes from `-`, and not an actual file, the `bao` tool asks you for the value of the secret.
+This isn't a login in to OpenBao; it's you providing the secret value.
```
cipherboy, line 171:
```suggestion
```

seems unnecessary.

website/content/docs/get-started/developer-qs.mdx
```diff
@@ -6,203 +6,309 @@ import TabItem from '@theme/TabItem';
...
-```
+  You can see that the data and metadata for the secret document are there, along with other fields.
+ 
+</details>
 
-</TabItem>
-</Tabs>
+## Step 5: retrieve the secret using code
+
+Now that you have practiced to write a secret, try to read it. This is what you're most likely to do in a real application.
```
cipherboy:
```suggestion
Now that you have practiced writing a secret, try to read it. This is what you're most likely to do in a real application.
```

cipherboy (commented):
Mostly nits in here.  Thanks @timb-controlplane!!

website/content/docs/get-started/developer-qs.mdx
```diff
@@ -6,203 +6,309 @@ import TabItem from '@theme/TabItem';
...
-}
-
-fmt.Println("Secret written successfully.")
+```go
+  // an example secret.
+  secretData := map[string]any{
+      "password": "OpenBao123",
+  }
+
+  mountPoint := "secret" // this is an OpenBao default
```
wslabosz-reply:
```suggestion
  mountPoint := "secret" // default path of the KVv2 mount enabled within dev server.
```

website/content/docs/get-started/developer-qs.mdx
```diff
@@ -6,203 +6,309 @@ import TabItem from '@theme/TabItem';
...
+A common way of storing secrets is as key-value pairs using the [KV secrets engine (v2)](../secrets/kv/kv-v2.mdx). In the code you just added, `password` is the key in the key-value pair, and `OpenBao123` is the value.
 
-We also provided the path to our secret in OpenBao. We will reference this path in a moment when we learn how to retrieve our secret.
+The code also provided the path to our secret in OpenBao. You will reference this path in a moment when you practice fetching the secret. The path includes the secret name, `developer-quickstart`.
+we learn how to retrieve our secret.
 
 Run the code now, and you should see `Secret written successfully`. If not, check that you've used the correct value for the root token and OpenBao server address.
 
-## Step 5: retrieve a secret
+## Step 5: Test that the stored secret is correct
```
wslabosz-reply: I lean towards always starting with uppercase on any header type, but I'm fine with lowercase also, but we probably should be consistent through the document.

website/content/docs/get-started/developer-qs.mdx
```diff
@@ -6,203 +6,309 @@ import TabItem from '@theme/TabItem';
...
+$ bao server -dev -dev-root-token-id="example-tutorial-token"
 ```
 
-The `-dev-root-token-id` flag for dev servers tells the OpenBao server to allow full root access to anyone who presents a token with the specified value (in this case "dev-only-token").
+The `-dev-root-token-id` flag for dev servers tells the OpenBao server to allow full ("root") access to anyone who presents a token with the specified value (in this case, `example-tutorial-token`).
 
 :::danger
 
-**Warning**: The [root token](../concepts/tokens.mdx#root-tokens) is useful for development, but allows full access to all data and functionality of OpenBao, so it must be carefully guarded in production. Ideally, even an administrator of OpenBao would use their own token with limited privileges instead of the root token.
+**Warning**: The [root token](../concepts/tokens.mdx#root-tokens) is useful for development, but allows full access to all data and functionality of OpenBao, so it must be carefully guarded in production. Read [Authentication](../concepts/auth.mdx), [Identity](../concepts/identity.mdx) and [Policies](../concepts/policies.mdx) if you want to set up a secure OpenBao cluster for real use.
```
wslabosz-reply: rather than `so it must be carefully guarded in production` it probably should be phrased more along the lines of: `should always be revoked past the setup phase, and tokens with more granular access be used instead.`

website/content/docs/get-started/developer-qs.mdx
timb-controlplane: Good to defer this to another PR? The perfect is the enemy of the shipped.

website/content/docs/get-started/developer-qs.mdx
timb-controlplane: I agree. That wording is from the original text, but anyway I will amend it.

website/content/docs/get-started/developer-qs.mdx
timb-controlplane: So, yes, but… I want to hint that you might need to do this for a real OpenBao.

timb-controlplane: I actually know that the "upgrade" has no effect. But if I don't teach this stage, someone using a non-dev mode OpenBao might be surprised. That's why I added it.

website/content/docs/get-started/developer-qs.mdx
timb-controlplane: Some people might actually not figure this out, especially if they skipped the first half of the page.

I'd like to leave in a hint, maybe with different wording. But in technical writing I tend to go for making the point a little more strongly than it is needed. Experience shows me that if we assume readers are sharp minded, then eventually someone who is both sharp minded and in a hurry will skip past a key detail and misunderstand.

website/content/docs/get-started/developer-qs.mdx
timb-controlplane: Done

website/content/docs/get-started/developer-qs.mdx
timb-controlplane: Done

timb-controlplane:
My sense is that this PR is at or past a point where it's clearly an improvement on what we have now. On that basis, I recommend approving it. We can do follow up PRs as we work even further towards perfection.

cipherboy (approved):

cipherboy:
I shall merge and we can always revise more later. Thanks!

PR #3623, opened 2026-07-28
Add context cancellation to initialization

cipherboy:
Notably, `preSeal(...)` does not cancel the active context. Because initialization will eventually finish (and we want to manually re-un-seal if using Shamir's), we should just defer the cancellation of the scoped active context we create during initialization. This will correctly mark `c.activeContext.Load()` as a cancelled context and we won't get panics like:

    sync.(*RWMutex).RLock(...)
    	/usr/local/go/src/sync/rwmutex.go:72
    github.com/openbao/openbao/v2/internal/vault.(*NamespaceStore).ResolveNamespaceFromRequest(0x0, {0x0?, 0x14cac65eb0a0?}, {0x14cac7659549, 0x34})
    	/home/cipherboy/GitHub/cipherboy/openbao/internal/vault/namespace_store.go:1589 +0x81
    github.com/openbao/openbao/v2/internal/vault.(*Core).switchedLockHandleRequest(0x14cac5e2e308, {0x4a10948, 0x14cac7f43380}, 0x14cac7f10fc0, 0x1)
    	/home/cipherboy/GitHub/cipherboy/openbao/internal/vault/request_handling.go:619 +0x431
    github.com/openbao/openbao/v2/internal/vault.(*Core).HandleRequest(...)
    	/home/cipherboy/GitHub/cipherboy/openbao/internal/vault/request_handling.go:569
    github.com/openbao/openbao/v2/internal/http.request(0x14cac5e2e308?, {0x4a06da0, 0x14cac7f43260}, 0x14cac7f5f7c0?, 0x14cac7f10fc0)
    	/home/cipherboy/GitHub/cipherboy/openbao/internal/http/handler.go:946 +0x6b
    github.com/openbao/openbao/v2/internal/http.handler.handleLogicalWithInjector.handleLogicalInternal.func59({0x4a06da0, 0x14cac7f43260}, 0x14cac7f5f7c0)
    	/home/cipherboy/GitHub/cipherboy/openbao/internal/http/logical.go:374 +0xce

when sending requests between initialization and actual node startup (whether as active or standby).

## Rationale

This was exacerbated locally by #3542, though I'm still not entirely sure why other than different function call structure. Regardless, this bug was introduced sometime in the v2.5.x cycle AFAICT. 

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/init_test.go
+++ b/internal/vault/init_test.go
@@ -4,12 +4,16 @@
 package vault
 
 import (
+	"context"
+	"sync"
 	"testing"
+	"time"
 
 	log "github.com/hashicorp/go-hclog"
 	wrapping "github.com/openbao/go-kms-wrapping/v2"
 	"github.com/openbao/openbao/sdk/v2/helper/logging"
 	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/openbao/openbao/sdk/v2/physical"
 	"github.com/openbao/openbao/sdk/v2/physical/inmem"
 	"github.com/openbao/openbao/v2/internal/helper/namespace"
 	"github.com/openbao/openbao/v2/internal/vault/seal"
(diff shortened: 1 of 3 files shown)


internal/vault/init_test.go
```diff
@@ -137,3 +141,82 @@ func testCoreInitCommon(t *testing.T, s Seal, barrierConf, recoveryConf *SealCon
...
+	}
+	c, err := NewCore(conf)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		defer func() {
+			if r := recover(); r != nil {
+				t.Log("panic closing core during cleanup", "panic", r)
+			}
+		}()
```
satoqz: Instead of recovering and logging, wouldn't you want to fail the test?

internal/vault/init_test.go
```diff
@@ -137,3 +141,82 @@ func testCoreInitCommon(t *testing.T, s Seal, barrierConf, recoveryConf *SealCon
...
+	time.Sleep(150 * time.Millisecond)
+
+	resp, err := c.Initialize(ctx, &InitParams{
+		BarrierConfig: &SealConfig{SecretShares: 5, SecretThreshold: 3},
+	})
+	require.NoError(t, err)
+
+	for _, key := range resp.SecretShares {
+		if _, err := TestCoreUnseal(c, key); err != nil {
+			t.Fatalf("unseal err: %s", err)
```
wslabosz-reply: prefer sticking to `require` here

cipherboy: @wslabosz-reply Yeah, I copied this part from an existing test. :laughing:

internal/vault/init_test.go
cipherboy: @satoqz As above, I didn't actually think about this at all, I copied it from an existing test. :laughing:

cipherboy:
@wslabosz-reply @satoqz Updated!

wslabosz-reply (approved):

PR #3599, opened 2026-07-23
Add step-by-step explanation for plugin install flow

thgoebel:
## Description

This MR adds a step-by-step explanation for how the plugin install flow works.

I figured that it is best to put this in the plugin overview, because the plugin system is best explained to beginners by following the deployment flow step-by-step (based on my experience trying to understand how it works).

At the same time, I pulled the "Installation/Download" step into a separate page, because it needs more detail, and I wanted to keep the overview lean-ish. When the content of `plugin-installation.mdx` is nested in the overview under "Plugin installation -> External plugins", I feel like there is too much nesting and too much detail. A dedicated page makes it smaller and more manageable imho. I'm open to suggestions.

The `plugin-management.mdx` is removed because its content fully moved into the overview.

## Rationale

Resolves: #3598

See the commit messages for a rational of the individual changes.

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3630, opened 2026-07-29
Docs: change terms (SHA-256 checksum -> digest, OCI image -> artifact)

thgoebel:
## Description

This MR changes some terms, as proposed by @timb-controlplane in his review of https://github.com/openbao/openbao/pull/3599.

I used the AE spelling (artifact) instead of BE (artefact). The choice of language should probably be documented in https://openbao.org/community/contributing/ (maybe I missed it)?

## Rationale

*See the commit messages for details.*

Resolves: n/a

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/configuration/plugins.mdx
+++ b/website/content/docs/configuration/plugins.mdx
@@ -23,7 +23,7 @@ managing the plugin binaries:
     plugins in code and OpenBao will automatically download and extract them
     during startup.
   * **Integrity Verification:**  When extracting plugin binaries from OCI images
-    OpenBao ensures the SHA256 sum of the extracted binary is the same as in the
+    OpenBao ensures the SHA-256 digest of the extracted binary is the same as in the
     configuration files.
   * **Version Management:** Distributing new a new version of a plugin just
     requires an update of the configuration file instead of manually placing the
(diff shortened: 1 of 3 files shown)


cipherboy (approved):
I don't know that I care as much about checksum/digest difference.

The artifact is strictly true, though I suppose the difference between a distroless image and a binary artifact is probably slim. 😆 


The project is historically started by a US company, so I'm fine codifying American English as the official project language variant, though I'm not sure we've really had strong opinions either way. Thoughts from anyone? 

(cc @openbao/dev-wg)

website/content/docs/configuration/plugins.mdx
```diff
@@ -74,15 +71,15 @@ plugin versions.
 
 ### Parameters
 
-- `image` `(string: required)` - OCI image URL including registry and repository. Conflicts with the `command` value. One of `image` or `command` is required.
```
cipherboy, line 77: Neither here nor there but worth documenting: there's a long-standing TOCTOU bug in the plugin execution; I don't think there's a reliable way to both read+checksum the file and execute it without opening it twice. It'd require deep changes to go-plugin to fix and is well outside of scope (its more of a safety precaution than a formal cryptographic/security guarantee... if you can modify binaries as an attacker, you've got bigger problems IMHO and its not a defensible position that we care about.).

satoqz: @cipherboy similarly neither here nor there: Long-term I'd like to allow config-based plugin registration to skip specifying a checksum at all (when not using OCI). It's security theater, IMO, and makes the life of operators unnecessarily hard at the same time.

PR #3631, opened 2026-07-29
bump semgrep and make sure `make` uses same version

phil9909:
## Description

1. Update the version of the semgrep image to 1.171.0 (was 1.10.0)
2. Switch from `returntocorp/semgrep` to `semgrep/semgrep`:

   > We've moved! Official Docker images for Semgrep now available at semgrep/semgrep.
   > https://hub.docker.com/r/returntocorp/semgrep

   They still push images to both, but I don't know for how long.
3. Change the Makefile target `docker-semgrep` and `docker-semgrep-ci` to use the same version as GitHub Actions

## Acknowledgements

       also allowed.

### DIFF
--- a/.github/workflows/code-checker.yml
+++ b/.github/workflows/code-checker.yml
@@ -52,7 +52,7 @@ jobs:
     name: Semgrep
     runs-on: ubuntu-latest
     container:
-      image: returntocorp/semgrep@sha256:ffc6f3567654f9431456d49fd059dfe548f007c494a7eb6cd5a1a3e50d813fb3
+      image: docker.io/semgrep/semgrep@sha256:296fc4912420833abfcf5090ec49a81bf1330bb61fc193cd09255f88b8267ba9 # 1.171.0
     # Skip any PR created by dependabot and any check triggered by merge group
     if: (github.actor != 'dependabot[bot]') && (github.event != 'merge_group')
     steps:
--- a/.github/workflows/code-checker.yml
+++ b/.github/workflows/code-checker.yml
@@ -52,7 +52,7 @@ jobs:
     name: Semgrep
     runs-on: ubuntu-latest
     container:
-      image: docker.io/semgrep/semgrep@sha256:296fc4912420833abfcf5090ec49a81bf1330bb61fc193cd09255f88b8267ba9 # 1.171.0
+      image: docker.io/semgrep/semgrep:1.171.0@sha256:bdf7013b2c3634a487671158da77c554f531742326b543a9464d2adf6c433ac8
     # Skip any PR created by dependabot and any check triggered by merge group
     if: (github.actor != 'dependabot[bot]') && (github.event != 'merge_group')
     steps:
--- a/Makefile
+++ b/Makefile
@@ -209,10 +209,10 @@ semgrep-ci:
 SEMGREP_IMAGE=$(shell grep -o '[^ ]*/semgrep:[^ ]*@sha256:[^ ]*' .github/workflows/code-checker.yml | head -n1)
 
 docker-semgrep:
-	$(DOCKER_CMD) run --rm --mount "type=bind,source=$(PWD),destination=/src,chown=true,relabel=shared" $(SEMGREP_IMAGE) semgrep --include '*.go' -a -f tools/semgrep .
+	$(DOCKER_CMD) run --rm --volume "$(PWD):/src" $(SEMGREP_IMAGE) semgrep --include '*.go' -a -f tools/semgrep .
 
 docker-semgrep-ci:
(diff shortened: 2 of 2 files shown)


.github/workflows/code-checker.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     name: Semgrep
     runs-on: ubuntu-latest
     container:
-      image: returntocorp/semgrep@sha256:ffc6f3567654f9431456d49fd059dfe548f007c494a7eb6cd5a1a3e50d813fb3
+      image: docker.io/semgrep/semgrep@sha256:296fc4912420833abfcf5090ec49a81bf1330bb61fc193cd09255f88b8267ba9 # 1.171.0
```
satoqz:
```suggestion
      image: docker.io/semgrep/semgrep:1.171.0@sha256:bdf7013b2c3634a487671158da77c554f531742326b543a9464d2adf6c433ac8
```

You were pointing at the `linux/amd64` manifest, not the multi-arch platform manifest. You can also specify a tag as metadata before the digest, no need to make it a comment.

See outputs of

```sh
crane manifest docker.io/semgrep/semgrep@sha256:296fc4912420833abfcf5090ec49a81bf1330bb61fc193cd09255f88b8267ba9
```

```sh
crane manifest docker.io/semgrep/semgrep:1.171.0@sha256:bdf7013b2c3634a487671158da77c554f531742326b543a9464d2adf6c433ac8
```

the latter includes the former in its `linux/amd64` entry.

satoqz: Ah, I see placing the tag elsewhere makes the regex in `Makefile` easier to write. Fine with me, but perhaps good to comment on.

Makefile
```diff
@@ -206,11 +206,13 @@ semgrep:
 semgrep-ci:
 	semgrep --error --include '*.go' -f tools/semgrep/ci .
 
+SEMGREP_IMGE=$(shell grep -o '[^ ]*/semgrep@sha256:[a-f0-9]*' .github/workflows/code-checker.yml | head -n1)
```
satoqz:
```suggestion
SEMGREP_IMAGE=$(shell grep -o '[^ ]*/semgrep@sha256:[a-f0-9]*' .github/workflows/code-checker.yml | head -n1)
```

.github/workflows/code-checker.yml
phil9909: > You were pointing at the linux/amd64 manifest, not the multi-arch platform manifest. 

Got fooled by the DockerHub UI :D thanks

> Ah, I see placing the tag elsewhere makes the regex in Makefile easier to write.

Fixed the regex. Honestly was just applying the pattern we have elsewhere in the file, but these are referencing Actions not images and can't use the tag+sha syntax

Makefile
phil9909: Wondering why my spellchecker didn't catch that. Thanks

Makefile
```diff
@@ -206,11 +206,13 @@ semgrep:
 semgrep-ci:
 	semgrep --error --include '*.go' -f tools/semgrep/ci .
 
+SEMGREP_IMAGE=$(shell grep -o '[^ ]*/semgrep:[^ ]*@sha256:[^ ]*' .github/workflows/code-checker.yml | head -n1)
+
 docker-semgrep:
-	$(DOCKER_CMD) run --rm --mount "type=bind,source=$(PWD),destination=/src,chown=true,relabel=shared" docker.io/returntocorp/semgrep:latest semgrep --include '*.go' -a -f tools/semgrep .
```
satoqz, line 210: This never worked with Docker:

```
unknown option 'chown' in 'chown=true'
```

but we can fix that another time if you'd prefer.

satoqz (dismissed):

Makefile
phil9909: has been this way since @cipherboy introduced the target in #461 
AFAIK also a podman user, which does support this option.

@cipherboy any objections against dropping this option? For me it keeps working without. The container runs as root, so it should have access to the files no mater what your user-id is. New files will be owned by root, but it should not create any (only fix existing files), right?

satoqz: `relabel=shared` also does not exist in docker, at which point we might as well just use `-v`.

phil9909: How about you installing Podman? :laughing:

satoqz: Someone has to dogfood using Docker to make sure we're super-compatible. You should be thanking me!!!

cipherboy: I'm fine to drop.

phil9909: dropped

cipherboy (approved):
Works for me on that Podman thing ;-)

PR #3636, opened 2026-07-29
Add blog post on declarative plugins

cipherboy:
## Description

Timely with all the effort around plugin documentation, this post
focuses on outlining the easier configuration for plugins rather than
using API-based registration.

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/blog/2026-07-30-declarative-plugins.md
+++ b/website/content/blog/2026-07-30-declarative-plugins.md
@@ -1,0 +1,214 @@
+---
+title: "OpenBao Features - Declarative Plugins"
+description: "Blog series describing OpenBao's features. This episode focuses on declarative plugin distribution and registration in OpenBao."
+slug: features-declarative-plugins
+authors: cipherboy
+tags: [features, storage, technical]
+---
+
+This is the fourth part of a [multi-part series on OpenBao's features](/blog/tags/features).
+
+[Last time](./2026-07-21-declarative-configuration.md) we talked about how to
+declaratively configure audit devices and initialize OpenBao. We saw how this
+made integration of OpenBao in a wider ecosystem or product (such as
+[EdgeX](2025-03-18-edgex-selects-openbao.md)) easier.
+
+Like the last part, this part focuses on the operator experience, but for
+consumption of OpenBao's plugins: auth methods, secrets engines, auto-unseal
+devices, and more.
+
+Our motivation here is to build towards a more [OpenTofu](https://opentofu.org/)-like,
+extensible ecosystem. Easier consumption, [community-maintained
+plugins](https://github.com/openbao/openbao-plugins), and a future plugin
+registry will lead to more developers writing plugins and expand the
+usefulness of OpenBao for everyone.
+
+:::tip[Question]
+
+What integrations would you like OpenBao to have? How would you like to see
+writing plugins made easier?
+
+[Contact us](https://github.com/openbao#contact) to share your thoughts or
+contribute to the ecosystem!
+
+::::
+
+<!-- truncate -->
+
+## Overview
+
+Nearly everything within OpenBao is pluggable. OpenBao supports several types of plugins currently:
+
+| Type            | Description                                                                                                                                                                   |
+| :-------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Auth**        | Allow bringing sources of identity--JWTs, certificates, passwords--and exchanging them for OpenBao tokens.                                                                    |
+| **Secret**      | Allow generating credentials (whether passwords for databases, PKI certificates, static secrets, and more) and protected through OpenBao's tokens and authorization policies. |
+| **Database**    | Implement database-specific logic for updating and managing access for use with the `database` dynamic secret engine.                                                         |
+| **KMS**         | Implement a specialized interfaces for using HSM and KMS devices (such as YubiHSM or GCP CloudKMS) for auto-unseal (and external keys in v2.6.0).                             |
+
+Plugins have two primary classifications:
+
+| Classification | Description                                                                                                                                                                 |
+| :------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Builtin**    | Distributed as part of the main `bao` binary artifact. No external process is started.                                                                                      |
+| **External**   | Distributed as stand-alone binaries, requiring installation and configuration to use. Spawns an external process (managed by the OpenBao server) to handle plugin requests. |
+
+For example, [`secrets-pki`](/docs/secrets/pki) is a **builtin** **secret**
+plugin, but [`secrets-aws`](https://github.com/openbao/openbao-plugins/releases?q=secrets-aws-&expanded=true)
+is an **external** **secret** plugin. [`auth-aws`](https://github.com/openbao/openbao-plugins/releases?q=auth-aws-&expanded=true)
+also exists and is an **external** **auth** plugin.
+
+## Usage
+
+OpenBao since [v2.5.x](/community/release-notes/2-5-0/#v250) supported
+[declarative plugin configuration](/docs/configuration/plugins/). Three
+mandatory configuration [parameters](/docs/configuration/#parameters) are
+required for these demos:
+
+```hcl
+// Only plugins from the specified directory will be loaded. This helps
+// to prevent arbitrary binaries from being executed and leading to
+// generalized RCE.
+plugin_directory = "/openbao/plugins"
+
+// Setting this allows the server process to automatically download new
+// plugins if they're not registered. Otherwise OpenBao will not contact
+// the internet by default. This will occur on all nodes (active and
+// standby).
+plugin_auto_download = true
+
+// Setting this allows automatic registration of the plugin into the
+// plugin catalog. This is allows operators to immediately use the plugin
+// without having to manually register it using API commands:
+//
+// https://openbao.org/api-docs/system/plugins-catalog/
+plugin_auto_register = true
+```
+
+With that in mind, adding an external plugin is greatly simplified:
+
+```hcl
+// Defines a new secret engine, named "aws".
+plugin "secret" "aws" {
+    // ## Distribution Information ##
+
+    // Image is the container registry's address. It is strongly recommended
+    // to pull from a trusted, local registry mirror instead of a public
+    // registry.
+    //
+    // See also: https://github.com/openbao/openbao-plugins/pkgs/container/openbao-plugin-secrets-aws/1056268559?tag=v0.3.1
+    image = "ghcr.io/openbao/openbao-plugin-secrets-aws"
+
+    // The version of the plugin itself; must be a semantic version and also a
+    // tag on the above registry.
+    //
+    // See also: https://semver.org/
+    version = "v0.3.1"
+
+    // Within the image, the path to the plugin binary. For OpenBao's
+    // community maintained plugins, this is the same as the plugin itself.
+    binary_name = "openbao-plugin-secrets-aws"
+
+    // ## Execution Information ##
+
+    // This contains the hash of the plugin binary and is required regardless
+    // of distribution mechanism.
+    //
+    // If the above registry were to change the tag to point at a different
+    // version of the image, this sha256sum prevents us from pulling an
+    // unexpected plugin binary.
+    sha256sum = "641ae1858c0f660b4c3761286d627e1dcfd33dc35e7c83eb81ca9050b5bdc6d8"
+
+    // args = [ ... ] and env = [ ... ] are optional parameters if the plugin
+    // requires additional information.
+}
+```
+
+...and send a `SIGHUP` or restart the server. That's it! You should now see
+messages in the server logs like:
+
+```hcl
+[INFO]  plugins: starting OCI plugin downloading
+[DEBUG] plugins: processing plugin: plugin=secret-aws url=ghcr.io/openbao/openbao-plugin-secrets-aws:v0.3.1 binary_name=openbao-plugin-secrets-aws
+[INFO]  plugins: downloading plugin from OCI registry: plugin=secret-aws url=ghcr.io/openbao/openbao-plugin-secrets-aws:v0.3.1
+[DEBUG] plugins: local platform: plugin=secret-aws platform=linux/amd64
+[DEBUG] plugins: extracting plugin from OCI image: plugin=secret-aws target=/openbao/plugins/.oci-cache/secret-aws/641ae185/openbao-plugin-secrets-aws binary=openbao-plugin-secrets-aws
+[INFO]  plugins: found plugin binary in OCI image: plugin=secret-aws entry=openbao-plugin-secrets-aws size=25161890
+[DEBUG] plugins: successfully extracted plugin binary: plugin=secret-aws path=/openbao/plugins/.oci-cache/secret-aws/641ae185/openbao-plugin-secrets-aws size=25161890
+[INFO]  plugins: successfully downloaded and validated plugin: plugin=secret-aws cached_path=/openbao/plugins/.oci-cache/secret-aws/641ae185/openbao-plugin-secrets-aws symlink_path=/openbao/plugins/secret-aws-v0.3.1 hash=641ae1858c0f660b4c3761286d627e1dcfd33dc35e7c83eb81ca9050b5bdc6d8
+[INFO]  plugins: OCI plugin downloading completed
+```
+
+It will now show up in the plugin catalog:
+
+```shell
+$ bao read -field=secret sys/plugins/catalog
+[aws kubernetes kv ldap openldap pki rabbitmq ssh totp transit]
+```
+
+and the plugin can be enabled:
+
+```shell
+$ bao secrets enable -path team/creds/aws aws
+Success! Enabled the aws secrets engine at: team/creds/aws/
+```
+
+Looking at the storage layout, we see that the plugin was downloaded and
+placed in the plugin folder as requested:
+
+```shell
+$ tree -alh plugins/
+[  100]  plugins/
+├── [   66]  .oci-cache
+│   └── [   16]  secret-aws
+│       └── [   52]  641ae185
+│           └── [  24M]  aws
+└── [   80]  secret-aws-v0.3.1 -> .oci-cache/secret-aws/641ae185/aws
+```
+
+:::info
+
+Even if you don't want to automatically download plugins as part of the server
+process, you can invoke the standalone `bao plugin` CLI to perform downloads
+and still benefit from declarative configuration and OCI-based plugin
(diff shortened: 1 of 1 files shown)


website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,231 @@
...
+    sha256sum = "55245882727535579e710672f0eae1bcdddc846006db857baaa6e09e33d40faf"
+}
+
+seal "openbao-plugin-kms-pkcs11" {
+    lib = "/usr/lib64/softhsm/libsofthsm.so"
+    token_label = "OpenBao"
+    pin = "4321"
+    key_label = "bao-root-key-rsa"
+    rsa_oaep_hash = "sha1"
+}
```
satoqz:
```suggestion
plugin "kms" "pkcs11" {
    image = "ghcr.io/openbao/openbao-plugin-kms-pkcs11"
    version = "v0.1.0"
    binary_name = "openbao-plugin-kms-pkcs11"
    sha256sum = "55245882727535579e710672f0eae1bcdddc846006db857baaa6e09e33d40faf"
}

seal "pkcs11" {
    lib = "/usr/lib64/softhsm/libsofthsm.so"
    token_label = "OpenBao"
    pin = "4321"
    key_label = "bao-root-key-rsa"
    rsa_oaep_hash = "sha1"
}
```

Is the far cleaner style IMO. Also applies to other plugin types.

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,231 @@
...
+
+Plugins have two primary classifications:
+
+| Classification | Description                                                                                                                                                |
+| :------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **External**   | Distributed as stand-alone binaries, requiring configuration to use. Spawns an external process (managed by the OpenBao server) to handle plugin requests. |
+| **Builtin**    | Distributed as part of the main `bao` binary artifact. No external process is started.                                                                     |
+
+For example, [`pki`](/docs/secrets/pki) is a **builtin** **secret** plugin, but
+[`secrets-aws`](https://github.com/openbao/openbao-plugins/releases?q=secrets-aws-&expanded=true)
```
satoqz:
```suggestion
For example, [`secrets-pki`](/docs/secrets/pki) is a **builtin** **secret** plugin, but
[`secrets-aws`](https://github.com/openbao/openbao-plugins/releases?q=secrets-aws-&expanded=true)
```

Should it be this for consistency?

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: @satoqz Hmm, I do think we should keep the `type` prefix as there's a lot of duplicates (cloud providers for instance).

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,231 @@
...
+    image = "ghcr.io/openbao/openbao-plugin-secrets-aws"
+    version = "<new update version>"
+    binary_name = "openbao-plugin-secrets-aws"
+    sha256sum = "<new update sha>"
+}
+```
+
+This will allow OpenBao to register two versions of the plugin. Mounts can be
+individually updated to the latest plugin version and operators can have
+granular control over when that occurs.
```
satoqz: This might be a bit too shallow. You'd want to mention when to make that config update, how to make openbao load it, how to then update mounts and reload... Perhaps it's better to just update the plugin upgrade guide to mention declarative plugins and link to that.

website/content/blog/2026-07-30-declarative-plugins.md
satoqz: @cipherboy that's why you specify the plugin type right before it, no?

cipherboy: @satoqz Ok, fair :laughing:

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: But it is builtin and we don't yet build them as external... We could refer to `aws`, but then they'd have to resolve the ambiguity between the auth and the secrets engines. :thinking: 

I guess Kubernetes is still builtin and is ambiguous.

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: @satoqz Fair, I'll wait for the other PR to merge and then link to that and publish this.

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
...
+| :-------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Auth**        | Allow bringing sources of identity--JWTs, certificates, passwords--and exchanging them for OpenBao tokens.                                                                    |
+| **Secret**      | Allow generating credentials (whether passwords for databases, PKI certificates, static secrets, and more) and protected through OpenBao's tokens and authorization policies. |
+| **Database**    | Implement database-specific logic for updating and managing access for use with the `database` dynamic secret engine.                                                         |
+| **Auto-Unseal** | Implement a specialized interfaces for using HSM and KMS devices (such as YubiHSM or GCP CloudKMS) for auto-unseal.                                                           |
+| **KMS**         | Implement a generic interface for using HSm and KMS devices within OpenBao's secret engines. _Coming in OpenBao v2.7.x_.                                                      |
```
wslabosz-reply:
```suggestion
| **KMS**         | Implement a generic interface for using HSM and KMS devices within OpenBao's secret engines. _Coming in OpenBao v2.7.x_.                                                      |
```

satoqz: I mean, yes, but they ship as one package. Perhaps defer external key additions to a later statement, outside of this table, since it is announced as "current"? Auto-Unseal plugins are already known as KMS plugins.

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
...
+
+Plugins have two primary classifications:
+
+| Classification | Description                                                                                                                                                |
+| :------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **External**   | Distributed as stand-alone binaries, requiring configuration to use. Spawns an external process (managed by the OpenBao server) to handle plugin requests. |
+| **Builtin**    | Distributed as part of the main `bao` binary artifact. No external process is started.                                                                     |
```
satoqz: I think builtin -> external instead of external -> builtin is the more natural order.

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
...
+```shell
+$ bao read -field=secret sys/plugins/catalog
+[aws kubernetes kv ldap openldap pki rabbitmq ssh totp transit]
+```
+
+and the plugin can be enabled:
+
+```shell
+$ bao secrets enable aws
+Success! Enabled the aws secrets engine at: aws/
```
wslabosz-reply: maybe let's mention `path`? otherwise this reads as: the name of the plugin is always the path of the mount for that plugin

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
...
+│   └── [   16]  secret-aws
+│       └── [   52]  641ae185
+│           └── [  24M]  aws
+└── [   80]  secret-aws-v0.3.1 -> .oci-cache/secret-aws/641ae185/aws
+```
+
+:::info
+
+Even if you don't want to automatically download plugins, you can use the CLI
+helper and still benefit from declarative configuration:
```
satoqz:
```suggestion
If you don't want to automatically download plugins at server startup, you can use the separate `bao plugin init` command and still benefit from declarative configuration:
```

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
...
+[INFO]  downloading plugin from OCI registry: plugin=secret-aws url=ghcr.io/openbao/openbao-plugin-secrets-aws:v0.3.1
+[INFO]  found plugin binary in OCI image: plugin=secret-aws entry=openbao-plugin-secrets-aws size=25161890
+[INFO]  successfully downloaded and validated plugin: plugin=secret-aws cached_path=/openbao/plugins/.oci-cache/secret-aws/641ae185/openbao-plugin-secrets-aws symlink_path=/openbao/plugins/secret-aws-v0.3.1 hash=641ae1858c0f660b4c3761286d627e1dcfd33dc35e7c83eb81ca9050b5bdc6d8
+```
+
+:::
+
+This pattern applies for auth and database plugins. When using auto-unseal
+plugins, note that the plugin will not show up in the registry but will
+otherwise be usable at startup:
```
satoqz:
```suggestion
This pattern applies for secrets, auth and database plugins. When using KMS
plugins, note that the plugin will not show up in the registry but will
otherwise be usable at startup:
```

satoqz: Would also be good to mention what "showing up in the registry" means. Probably more like "show up under plugin catalog endpoints".

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
+---
+title: "OpenBao Features - Declarative Plugin Distribution"
```
satoqz: It's more of a post on OCI-based plugin distribution. I don't think distribution is in any way declarative. The configuration is declarative, but also covers a broader scope than OCI.

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
...
+| **Database**    | Implement database-specific logic for updating and managing access for use with the `database` dynamic secret engine.                                                         |
+| **Auto-Unseal** | Implement a specialized interfaces for using HSM and KMS devices (such as YubiHSM or GCP CloudKMS) for auto-unseal.                                                           |
+| **KMS**         | Implement a generic interface for using HSm and KMS devices within OpenBao's secret engines. _Coming in OpenBao v2.7.x_.                                                      |
+
+Plugins have two primary classifications:
+
+| Classification | Description                                                                                                                                                |
+| :------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **External**   | Distributed as stand-alone binaries, requiring configuration to use. Spawns an external process (managed by the OpenBao server) to handle plugin requests. |
```
thgoebel:
```suggestion
| **External**   | Distributed as stand-alone binaries, requiring installation and configuration to use. Spawns an external process (managed by the OpenBao server) to handle plugin requests. |
```

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
...
+plugin_directory = "/openbao/plugins"
+
+// Setting this allows the server process to automatically download new
+// plugins if they're not registered. Otherwise OpenBao will not contact
+// the internet by default. This will occur on all nodes (active and
+// standby).
+plugin_auto_download = true
+
+// Setting this allows automatic registration of the plugin into the
+// registry. This is allows operators to immediately use the plugin
```
thgoebel:
```suggestion
// catalog. This is allows operators to immediately use the plugin
```

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
...
+    // a public registry.
+    //
+    // See also: https://github.com/openbao/openbao-plugins/pkgs/container/openbao-plugin-secrets-aws/1056268559?tag=v0.3.1
+    image = "ghcr.io/openbao/openbao-plugin-secrets-aws"
+
+    // The version of the plugin itself; must be a semantic version and also a
+    // tag on the above registry.
+    //
+    // See also: https://semver.org/
+    version = "v0.3.1"
```
thgoebel: > It is strongly recommended to pin to specific SHA256 hashes rather than using a tag which may be moved.

1. Can I really pin the hash in the image field? I assumed that OpenBao merges `image` and `version`. Can you give an example of how to pin the SHA-256 digest in the `image` field?
2. Is pinning the image itself necessary? The `sha256sum` field already pins the binary inside the image, isn't that sufficient?

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,211 @@
...
+    // Image is the registry's address. It is strongly recommended to pin to
+    // specific SHA256 hashes rather than using a tag which may be moved. In
+    // production environments, pull from a local registry mirror instead of
+    // a public registry.
+    //
+    // See also: https://github.com/openbao/openbao-plugins/pkgs/container/openbao-plugin-secrets-aws/1056268559?tag=v0.3.1
+    image = "ghcr.io/openbao/openbao-plugin-secrets-aws"
+
+    // The version of the plugin itself; must be a semantic version and also a
+    // tag on the above registry.
```
thgoebel, line 103: 1. Does OpenBao enforce that this is a "semantic version"? What if my plugin has version `2026.07.1`?
2. (General question) Do plugin binaries somehow report their version to OpenBao? Or does OpenBao take the plugin version from what the operator registers in the catalog? Aka, could the operator register a plugin as version 2.0 even though the binary is compiled from 1.0?
 
```suggestion
    // The version of the plugin itself; must be a tag on the above registry.
```

website/content/blog/2026-07-30-declarative-plugins.md
satoqz: I concur with @thgoebel:

1. I'm not sure it's even possible either, at least not within the `image` field.
2. Yeah, you should just be pinning `sha256sum` and use a semver tag only for the OCI artifact. In fact, I'd mention that it is _not_ necessary (or even possible?) to pin the artifact itself separately for this reason.

website/content/blog/2026-07-30-declarative-plugins.md
satoqz: I think 

image

may be the adequate answer to these questions, but I'd need to doublecheck.

satoqz:
```
self-initialization failed: failed to set up declarative plugins: [plugin 0 (name: aws / type: secret)]: failed to register new plugin: plugin version mismatch: aws reported version (v0.2.0) did not match requested version (v0.2.0@sha256:74c77f613f258f155054c9219ea918d5fd353ae413430cbf1587d5506afee0e4)
```

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: Yeah I wrote this first and then tested it only to realize that we can't pin sha256 because the image tag is the version. I definitely think we should either support an explicit image_tag field that would let us take a OCI image hash or just place it into the existing `image` parameter.

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: Yeah, Vault was fond of forcing semantic versions in a number of places, since they write a Go semantic version library. 😆

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: Revised, combining the two together.

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: @satoqz Doesn't this also occur at SIGHUP time though?

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: I dropped `Distribution` from here.

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: Ah, I realize my issue now, my testing config snippets under `/tmp` weren't synced into the post itself, I had edited the wrong ones. :facepalm:

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: > Does OpenBao enforce that this is a "semantic version"? What if my plugin has version `2026.07.1`?

`2026.07.1` I believe is a valid semantic version. What you couldn't do is `20260701`.

website/content/blog/2026-07-30-declarative-plugins.md
cipherboy: I added `(shown above)` to refer the reader back up.

website/content/blog/2026-07-30-declarative-plugins.md
satoqz: @cipherboy sure, what sounded wrong to me was the implication that `bao plugin init` isn't "automatically downloading plugins". I'd say it still falls into the bucket of automatic downloads; what matters more is expressing "if you don't want `bao server` to somehow do it as part of its lifecycle".

cipherboy: @satoqz Very well, I've revised again and pushed.

website/content/blog/2026-07-30-declarative-plugins.md
```diff
@@ -0,0 +1,214 @@
...
+
+| Type            | Description                                                                                                                                                                   |
+| :-------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Auth**        | Allow bringing sources of identity--JWTs, certificates, passwords--and exchanging them for OpenBao tokens.                                                                    |
+| **Secret**      | Allow generating credentials (whether passwords for databases, PKI certificates, static secrets, and more) and protected through OpenBao's tokens and authorization policies. |
+| **Database**    | Implement database-specific logic for updating and managing access for use with the `database` dynamic secret engine.                                                         |
+| **KMS**         | Implement a specialized interfaces for using HSM and KMS devices (such as YubiHSM or GCP CloudKMS) for auto-unseal (and external keys in v2.6.0).                             |
```
satoqz:
```suggestion
| **KMS**         | Implement specialized interfaces for using HSM and KMS devices (such as YubiHSM or GCP CloudKMS) for auto-unseal (and external keys in the upcoming v2.7.0).                  |
```

satoqz (dismissed):
One fix, LGTM otherwise.

wslabosz-reply (approved):

satoqz (approved):

cipherboy:
Thanks all!

PR #3649, opened 2026-07-30
Don't fail if raft path does not exist

satoqz:
## Description

Something that's been bugging me about the Raft backend for a while is that it fails at initial startup if the configured storage path does not yet exist:

```
error initializing storage of type raft: failed to create fsm: failed to open bolt file: open hack/raft/vault.db: no such file or directory
```

## Rationale

QoL. Also matches the file backend which is well-behaved.

## Acknowledgements

       also allowed.


satoqz:
Ah, it would even fail if you had a node ID, and we even had a test to make sure that it'd fail _somehow_. What else should I have expected.

wslabosz-reply (approved):
very nice addition for testing purposes also

internal/physical/raft/raft.go
```diff
@@ -350,6 +350,12 @@ func NewRaftBackend(conf map[string]string, logger log.Logger) (physical.Backend
 		path = pathFromConfig
 	}
 
+	// Ensure the top-level raft path exists so operators don't need to manually
+	// create it.
+	if err := EnsurePath(path, true); err != nil {
```
cipherboy, line 355: Nice of this to set permissions for us nicely.

cipherboy (approved):

PR #3650, opened 2026-07-30
ui: hide cubbyhole from the Secrets Engines list

JulienFS:
The UI already omits `system` and `identity` from the Secrets Engines list via
`LIST_EXCLUDED_BACKENDS` — the source even notes the system backend "is an
implementation detail". `cubbyhole` is the same kind of built-in mount: a
per-token scratch space with no meaningful interactive use in the engines list,
and its main purpose (response wrapping) works without the engine being listed.

This adds `cubbyhole` to that same exclusion list so a freshly logged-in user is
not shown an obscure primitive alongside their real secrets engines.

### DIFF
--- a/changelog/3650.txt
+++ b/changelog/3650.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: hide the cubbyhole secrets engine from the Secrets Engines list, consistent with system and identity
+```
--- a/ui/app/models/secret-engine.js
+++ b/ui/app/models/secret-engine.js
@@ -14,7 +14,7 @@ const LINKED_BACKENDS = supportedSecretBackends();
 
 // identity will be managed separately and the inclusion
 // of the system backend is an implementation detail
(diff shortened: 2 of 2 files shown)


satoqz:
https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md

Also, I disagree, it seems good to have in the list and is a proper secrets engine.

Florianisme:
@satoqz can it be made configurable instead? Something similar to how "Redpanda Console" deals with internal Kafka topics. You can tick a checkbox in the UI to "Show/Hide internal topics".
image


That way you could still look at it if you wanted to, but newcomers wouldn't be confused by it.

Consumers of this service (not administrators, "developers" instead), such as how we are using it, don't need to know the internals of this tool to create/view their secrets.

Right now I create a KV for each team and tell them "ignore cubbyhole, use your KV" but mistakes happen.

satoqz:
@Florianisme sure, that's more sensible. Open an issue!

Issue #2793, opened 2026-04-02
Website docs reference VAULT_ env variable prefixes


hashworks:
Follow-up from [openbao-snapshot-agent/pull/37](https://github.com/openbao/openbao-snapshot-agent/pull/37#discussion_r3026658570).

Environment variables are read using [`ReadBaoVariable`](https://github.com/openbao/openbao/blob/main/api/env.go#L22), which prefers `BAO_` prefix but also supports the upstream `VAULT_` prefix to remain compatibility. The website docs still reference the upstream prefix in a lot of places and should prefer the `BAO_` prefix instead.

Due to the way env vars are read it might be possible to do a general `s/VAULT_/BAO_/g` in `.mdx` files, but I guess we can be on the safe side and do this in multiple PRs that reference the Go code of changed envs?

satoqz:
@hashworks we've got parts of this upcoming in https://github.com/openbao/openbao/pull/2706; though I'm not sure if there are other files where we haven't replaced it yet.

I also suspect there are some locations in code where we don't use `ReadBaoVariable` but `Getenv`, so we need to carefully triage as we replace.

hashworks:
Ah, I didn't check that PR. Closing then.

satoqz:
I think it's good to keep open until someone's verified that we handle `BAO_*` and `VAULT_*` in _all_ places, and that it's documented as such.

PR #3656, opened 2026-07-30
remove unused const for enabling rate limit audit logging

Luiggi33:
## Description

While trying to find more instances of VAULT_ env vars being used in the code, I stumbled upon this supposed env var defined in a const. I couldn't find any reference to it either in OpenBao itself nor the docs. If this is actually used somewhere, I would be open to documenting its use.
The actual usage that is commented above this const is currently configured in `internal/vault/quotas/quotas.go` and is defined via the config/api.

## Rationale

Remove dead code that could be misleading in a search

Related to: #2793

## Acknowledgements

       also allowed.


cipherboy (approved):
At least in OpenBao and at the point in time of the fork, this appears unused. It appears unused at the time it was added as well for that matter.

PR #3660, opened 2026-07-31
Add endpoint for namespace barrier migration

### LINKED ISSUE #3333: Normal Namespace <-> Sealable Namespace migration
  While v2.6 introduces sealable namespaces, they can only be used for newly created namespaces. We should support migrating a normal existing namespace to a sealable namespace and back. The migrated namespace should have the same id and path as previously to allow e.g. existing tokens to continue to work.  

  Provide new API endpoints to allow migrating seal configs for existing namespaces.  
Migration Procedure:
- Lock and taint target namespace
- Create new empty shadow namespace(s) with the new seal config (all non-sealable child namespaces included)
	- should be tainted and locked too with the same lock key as the target namespace
- Copy all storage entries from the old namespace(s) to the new namespace(s) using the new namespace barrier
- Verify data integrity between old and new namespace(s)
- Delete old namespace storage entries (keeping all sealable child-namespaces intact, they remain untouched by this procedure)
- Modify the namespace storage entries (`core/namespaces/<uuid>`) with the old id, path, lock key
- Untaint, seal (if applicable), unlock new namespace

- We could integrate this into `POST /sys/namespaces/<path>`, but this seems prone to accidentally removing a namespace seal config by forgetting to pass the seal config
- We could also do an in-place migration of the data, but that seems way more complicated and does not allow easy roll-back in case of errors
- We could also do an export-import procedure, but this will cause data to leave the barrier and is more disruptive to sealable child namespaces (would have to be exported, deleted, reimported too although they are not really affected)

phyrog:
## Description

Implements the `migrate-barrier` endpoint to allow migration between normal and sealable namespaces.

The migration happens in-place ~in a single transaction~ (different from what I proposed in #3333), as it allows better reuse of the existing invalidation logic. When migrating to a sealable namespace, the resulting namespace remains unsealed after the migration. When migrating from a sealable namespace, the namespace must be unsealed first.

During the migration, the namespace will be tainted to prevent any other changes during the migration.

Migrating a namespace barrier requires sudo permissions in that namespace. I'm open to change that, but I think this is a reasonable requirement.

~This does not yet include the CLI part~. Also it might be useful to have a migration status endpoint, otherwise there is no way to find out a migration is currently in progress or failed.

## Rationale

Resolves: #3333

## Acknowledgements

       also allowed.


satoqz:
> Implements the migrate-seal endpoint to allow migration between normal and sealable namespaces.

I haven't looked at the code, but @phyrog I'm not sure this is a good name given "seal migration" already exists in OpenBao and does something different. Similarly, we'll also have seal migration within namespaces for per-namespace Auto Seals, but that would require an already "sealable" namespace.

phyrog:
> > Implements the migrate-seal endpoint to allow migration between normal and sealable namespaces.
> 
> I haven't look at the code, but @phyrog I'm not sure this is a good name given "seal migration" already exists in OpenBao and does something different. Similarly, we'll also have seal migration within namespaces for per-namespace Auto Seals, but that would require an already "sealable" namespace.

Yeah I know, but I couldn't think of a better name... `migrate-barrier`?

satoqz:
@phyrog yeah, I think that's better. Not perfect perhaps, but 🤷

cipherboy:
> The migration happens in-place in a single transaction

I'd be curious to have you try this on large namespace. If you run the KV benchmark for 30s against a namespace and then migrate it, does it work?

phyrog:
> > The migration happens in-place in a single transaction
> 
> I'd be curious to have you try this on large namespace. If you run the KV benchmark for 30s against a namespace and then migrate it, does it work?

Works up to a certain size. 3000 KV pairs with kvsize of 1000 works well and migration finishes in around 500ms using raft. 10000 KV pairs leads to `put failed due to value being too large; got 12378853 bytes, max: 8388608 bytes, 41492 operations in transaction`. Haven't checked the pg backend.

Not sure what the solution here would be to allow this for bigger namespaces (or many sub-namespaces). Splitting up the transaction could lead to inconsistency if one tx works but the next fails. You'd have to keep track of every secret currently being migrated in memory and revert finished transactions manually if a later one fails.

cipherboy:
> Splitting up the transaction could lead to inconsistency if one tx works but the next fails.

So thinking towards the future, we'll eventually have per-namespace storage backends as an option. At that point in time, you may not be guaranteed to have cross-namespace transactions. It may not actually impact this if you mostly write the data in-place in the existing storage backend (though, might be nice to support cross-storage moves later).

That said, I'm not sure why we care about inconsistency here, given we're tainting the namespace? We already can have this background job cancelled, so my suggestion for strategy would be:

1. Write namespace tainted.
2. Unmount everything within the namespace.
3. Move data individually (delete, read, write).
4. Remount everything.
5. Untaint the namespace.

2 ensures no conflicting writes (in conjunction with taint) and that lets us use a non-transactional serial process for moving data that can be safely interrupted. We'd then make migrate retryable on tainted mounts. 

My 2c., but I think that'd behave similar to delete in this case.

phyrog:
@cipherboy Ok, but what happens when there's errors during step 3? Say you start the migration, you re-encrypted half a namespace, some storage error happens and the job is aborted. You'd have some data encrypted with the new barrier, some with the old barrier. Next time the namespace is loaded/unsealed, it will pick up the seal config from storage. Trying to decrypt any of the non-migrated keys (or migrated keys, depending on which way to migration goes) will error now. Same thing can happen across namespaces.

cipherboy:
> @cipherboy Ok, but what happens when there's errors during step 3? Say you start the migration, you re-encrypted half a namespace, some storage error happens and the job is aborted. You'd have some data encrypted with the new barrier, some with the old barrier. Next time the namespace is loaded/unsealed, it will pick up the seal config from storage. Trying to decrypt any of the non-migrated keys (or migrated keys, depending on which way to migration goes) will error now. Same thing can happen across namespaces.

Ah I see.  I wonder if you could do this in another broader step?

1. Write a new namespace entry (`core/namespaces/<uuid>-migration`) (sentinels that we're performing a migration and not a deletion) + write existing namespace entry as tainted as a transaction.
3. Unmount everything within the namespace.
4. Move data individually to migrated version
    1. Read existing entry.
    3. Write to same path under new seal.
5. Remount everything.
6. Delete the migration sentinel and write to the existing one (new seal + untaint) as a transaction.

Then when you read an arbitrary entry, you could first try under the migrated seal and then still retain legacy seal information to do the migration if it needs to be done.

Tainted status _should_ be sufficient to prevent arbitrary writes and we can check that no child namespace is tainted before we begin the migration as well.

phyrog:
> core/namespaces/<uuid>-migration

would that only exist in the top-most namespace (the one that has or will have the seal config), or every namespace that will be migrated? I think top-most makes sense because it configures the barrier / will configure the barrier, but then how would you discern which child namespaces have been successfully migrated and which have not?

Regarding step 5: For normal -> sealable we need to write the new seal config first, otherwise we cannot init the barrier (create keyring etc). For sealable -> normal we need to delete the seal config (+ keyring etc) last to allow reading from the old barrier. So either way, the seal config must be written before 3. Also if we stop mid-migration, there is essentially no way to know (from storage entries alone) which way the migration was going, if we wanted to pick up later.

All in all I'd want to avoid even having a partially migrated state in storage, and especially replicating it to other nodes. It just opens up such a big can of worms.

Imagine this scenario: We have two sealable namespaces (`a` and `a/b`). We start a migration to normal of `a/b` (so it will use the `a` barrier), but an error occurs. `a/b` is now partially migrated. Someone else now starts a migration of `a` to normal (so it will use the root barrier). Usually we would ignore `a/b` as it is sealable on its own, but we are mid-migration to a normal namespace, so the correct behavior would be to migrate the already migrated keys again to use the root barrier. Then if someone retries the migration of `a/b`, it should finish migrating the rest of the keys to the root barrier.

cipherboy:
> Regarding step 5: For normal -> sealable we need to write the new seal config first, otherwise we cannot init the barrier (create keyring etc). For sealable -> normal we need to delete the seal config (+ keyring etc) last to allow reading from the old barrier. So either way, the seal config must be written before 3. Also if we stop mid-migration, there is essentially no way to know (from storage entries alone) which way the migration was going, if we wanted to pick up later.

Sorry, I guess this part wasn't clear.

In the migration entry (step 1), the sentinel would store the new desired seal. This means that startup/... would be unaffected by the interrupted migration (it'd have the current state of entries +/- what's been moved and would see the namespace as tainted) and upon resuming the migration, we'd also still have the desired new seal status (complete with key material &c).

There'd be no way to rollback a migration except by completing it and undoing it.

phyrog:
Ok, just to make sure I understood this correctly:

- the namespace is tainted at the start of the migration
- (if normal -> sealable) the new barrier is created and initialized (but without writing `seal-config`)
  - this creates the keyring, etc
    - when loading namespaces we only look at `seal-config` to see if a ns is sealable, so the ns would still register as a normal ns (but tainted)
- `core/namespaces/<uuid>-migration` is created for just the top-level ns being migrated, containing the in-progress migration state, consisting of:
  - the new seal config
  - which keys have been migrated (or still need migration)
  - this is updated on every key rewrite
- when all keys have been migrated, we write the `seal-config` key (if normal -> sealable) and delete `core/namespaces/<uuid>-migration`
- (if sealable -> normal) delete keyring, seal-config etc
- untaint the namespace

Interrupted migration can be resumed by checking if `core/namespaces/<uuid>-migration` already exists.

Some open questions/thoughts:

- does the namespace stay tainted when the migration is interrupted?
- do we block migrations of parent namespaces while a child namespace is partially migrated?
  - migrating child namespaces should be blocked automatically if the parent ns stays tainted
- every key rewrite would result in an additional migration state update/write
- do we leave new namespace barriers unsealed after migration? or do we seal them like on creation?
  - currently I leave them unsealed

cipherboy:
@phyrog said:

> * does the namespace stay tainted when the migration is interrupted?

Yes, `Tainted` is a field that is persisted to storage: 

https://github.com/openbao/openbao/blob/b3a2f497c191e5b7046859890a8ed5a2e83e6618/internal/helper/namespace/namespace.go#L44

The persistence of the `core/namespaces/*-migration` variant of the entry will signal that this is not a deletion taint but a migration taint. 

> do we block migrations of parent namespaces while a child namespace is partially migrated?
>   -   migrating child namespaces should be blocked automatically if the parent ns stays tainted

I definitely think we should do this regardless and condition on tainted status here. We don't want to be migrating a parent namespace when a child namespace is being deleted, yes? We shouldn't be deleting a namespace if a child namespace exists in the first place, except via the recursive force-deletion path I suppose, so the intersection of these two code paths is probably "don't do that" rather than something enforced with code. 

So I do think we condition not around "migrating" but around "tainted" and things just work (TM). 

> every key rewrite would result in an additional migration state update/write

To be clear, I didn't propose persisting state of the migration to storage, but I guess we could to optimize the migration resumption. I'd recommend writing every N entries, rather than after every entry. 

> do we leave new namespace barriers unsealed after migration? or do we seal them like on creation?
>  - currently I leave them unsealed

I don't have preferences here, dealer's choice!

phyrog:
> Yes, Tainted is a field that is persisted to storage

Yeah I know, the question was more towards `defer untaint()` in the job vs only untaint on success. But you answered that question as well.

> To be clear, I didn't propose persisting state of the migration to storage, but I guess we could to optimize the migration resumption.

I guess I misunderstood what you referenced here:

> it'd have the current state of entries +/- what's been moved

How would we know which keys are left to migrate otherwise? Read them using the old barrier and skip them in case of errors? I guess we could only persist the list of done keys in case of an error, but the risk would be pretty high that the error is a storage error. In that case we'd be unable to write the state as well. So probably writing after every n keys is most efficient.

Edit: Thinking some more about it, it's probably not a good idea to save the finished keys in storage. For very large namespaces this would explode the value extremely. So I think I'll go with just iterating all keys and skipping the ones that can't be read with the old barrier. It's a bit more wasteful when resuming a migration, but the normal case would be faster.

cipherboy:
> How would we know which keys are left to migrate otherwise? Read them using the old barrier and skip them in case of errors?

We could read with both on error to validate that the new expected barrier is correct, yes. 

> Edit: Thinking some more about it, it's probably not a good idea to save the finished keys in storage. For very large namespaces this would explode the value extremely. So I think I'll go with just iterating all keys and skipping the ones that can't be read with the old barrier. It's a bit more wasteful when resuming a migration, but the normal case would be faster.

Realized I wasn't precise here: I think we can write just a single entry path, not a list of all migrated paths.

I think we formally require that `List(...)` returns results in sorted order. If not, we should really document that because it is true and enforced by tests elsewhere. This would thus be true of (really, is a requirement of) `ListPage(...)` and by inheritance, all `ScanPage(...)` implementations.

Thus should be sufficient (under our definition above of tainted not allowing writes/...) to indicate location in the iteration index of the last update we performed.

Thoughts?

api/sys_namespaces.go
```diff
@@ -319,3 +319,52 @@ func (c *Sys) NamespaceSealStatusWithContext(ctx context.Context, name string) (
...
+	KeyShares    []string `json:"key_shares"`
+	KeyThreshold int      `json:"key_threshold"`
+	Status       string   `json:"status"`
+}
+
+func (c *Sys) MigrateNamespaceBarrier(name string, i *MigrateNamespaceBarrierInput) (*MigrateNamespaceBarrierOutput, error) {
+	return c.MigrateNamespaceBarrierWithContext(context.Background(), name, i)
+}
+
+func (c *Sys) MigrateNamespaceBarrierWithContext(ctx context.Context, name string, i *MigrateNamespaceBarrierInput) (*MigrateNamespaceBarrierOutput, error) {
```
cipherboy, line 343: We should probably add this to a guide, but I'm inclined not to have new `WithContext` alt methods if the base method doesn't already exist? I'm inclined to have this just be the signature with context:


```suggestion
func (c *Sys) MigrateNamespaceBarrier(ctx context.Context, name string, i *MigrateNamespaceBarrierInput) (*MigrateNamespaceBarrierOutput, error) {
```

internal/command/namespace_migrate_barrier.go
```diff
@@ -0,0 +1,187 @@
+// Copyright (c) HashiCorp, Inc.
```
cipherboy: I think this'd probably be our copyright header.

api/sys_namespaces.go
phyrog: I'm going to leave this unresolved for now, I'd like to get more opinions on this, given that this will cause inconsistency in the Go API. That said, I'm not opposed.

internal/vault/seal.go
```diff
@@ -32,6 +32,11 @@ const (
 	// parts must be used to reconstruct the unseal key.
 	barrierSealConfigPath = "core/seal-config"
 
+	// barrierMigrationStatePath is the path used to store the temporary barrier
+	// migration state while a migration is running. This value is encrypted
+	// with the old barrier.
```
wslabosz-reply:
```suggestion
	// migration state while a migration is running. This value is encrypted
	// by the barrier we are migrating from.
```

internal/vault/logical_system_namespaces_seals.go
```diff
@@ -335,23 +374,204 @@ func (b *SystemBackend) handleNamespacesDeleteSealed() framework.OperationFunc {
...
+type BarrierMigrationState struct {
+	SealConfig                *SealConfig
+	UnsealShares              [][]byte
+	LastMigratedNamespacePath string
+	LastMigratedKey           string
+}
+
+func (b *SystemBackend) handleNamespacesMigrateBarrier() framework.OperationFunc {
+	return func(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+		path := namespace.Canonicalize(data.Get("path").(string))
```
wslabosz-reply:
```suggestion
		path, err := namespace.ParseName(data.Get("path").(string))
		if err != nil {
			return handleError(err)
		}
```

internal/vault/logical_system_namespaces_seals.go
```diff
@@ -335,23 +374,204 @@ func (b *SystemBackend) handleNamespacesDeleteSealed() framework.OperationFunc {
...
+		ns, err := b.Core.namespaceStore.GetNamespaceByPath(ctx, path)
+		if err != nil {
+			return handleError(err)
+		}
+
+		if ns == nil {
+			return nil, fmt.Errorf("namespace %q doesn't exist", path)
+		}
+
+		oldBarrier := b.Core.sealManager.NamespaceBarrierByLongestPrefix(ns.Path)
```
wslabosz-reply: shouldn't we validate that the namespace is of sealable type at this point?

`oldBarrier := b.Core.sealManager.NamespaceBarrierByLongestPrefix(ns.Path)` might return parent's barrier (up to root)

internal/vault/logical_system_namespaces_seals.go
```diff
@@ -335,23 +374,204 @@ func (b *SystemBackend) handleNamespacesDeleteSealed() framework.OperationFunc {
...
+
+		se, err := NamespaceScopedView(oldBarrier, ns).Get(ctx, barrierMigrationStatePath)
+		if err != nil {
+			b.logger.Info("error trying to read stored migration config", "error", err)
```
wslabosz-reply: it's an `Info` level log starting with `error` indicating that we are missing the migration state entry, although we will always log this line with first-time migration request?

I think this will be confusing to the administrators

internal/vault/logical_system_namespaces_seals.go
```diff
@@ -335,23 +374,204 @@ func (b *SystemBackend) handleNamespacesDeleteSealed() framework.OperationFunc {
...
+				b.logger.Warn("unsealed namespace barrier", "namespace", ns)
+			} else {
+				state.UnsealShares, err = b.Core.sealManager.initializeBarrier(ctx, newBarrier, seal, state.SealConfig)
+				if err != nil {
+					return handleError(err)
+				}
+			}
+		}
+
+		if newBarrier == oldBarrier {
```
wslabosz-reply: doesn't this case indicate we've not migrated from anything? We've dispatched the request to migrate to unsealable namespace with initially unsealable namespace? Isn't there a way to handle this case earlier?

internal/vault/logical_system_namespaces_seals.go
```diff
@@ -335,23 +374,204 @@ func (b *SystemBackend) handleNamespacesDeleteSealed() framework.OperationFunc {
...
+		path := namespace.Canonicalize(data.Get("path").(string))
+
+		if !b.System().(extendedSystemView).SudoPrivilege(ctx, req.MountPoint+req.Path, req.ClientToken) {
+			return nil, logical.ErrPermissionDenied
+		}
+
+		var state BarrierMigrationState
+
+		sealRaw, ok := data.GetOk("seal")
+		var continueInterrupted bool
```
wslabosz-reply: had a hard time understanding what this variable meant, I'd consider renaming to something more explicit like `continueAfterPreviousInterruption`(or omit previous), also let's move this declaration closer to the reassignment (L:460)

internal/vault/logical_system_namespaces_seals.go
```diff
@@ -335,23 +374,204 @@ func (b *SystemBackend) handleNamespacesDeleteSealed() framework.OperationFunc {
...
-		Returns the seal status of the namespace.
+			newBarrier = barrier.NewAESGCMBarrier(b.Core.physical, ns)
+			if continueInterrupted {
+				if err := b.Core.sealManager.SetSeal(ctx, state.SealConfig, ns, SetSealOptions{
+					Seal:    seal,
+					Barrier: newBarrier,
+				}); err != nil {
+					return handleError(err)
+				}
+				b.logger.Warn("namespace seal set", "namespace", ns)
```
wslabosz-reply: I understand that you wanted to use the `warn` log level because we know that this is at least a second attempt at migration, but the contents of the log do not indicate so, I'd change the log level to info and adjust the log contents accordingly

internal/vault/logical_system_namespaces_seals.go
```diff
@@ -335,23 +374,204 @@ func (b *SystemBackend) handleNamespacesDeleteSealed() framework.OperationFunc {
...
+			if err != nil {
+				return handleError(err)
+			}
+			return nil, nil
+		}
 
-	DELETE /<path>/delete-sealed
-		Delete a sealed namespace by wiping its storage.
-		`,
+		migrationJob := b.Core.namespaceStore.newNamespaceBarrierMigrationJob(parentBarrier, oldBarrier, newBarrier, parentNs, ns, seal, &state)
```
wslabosz-reply: I'd consider leaving only: ns, newBarrier, seal and state as other properties can be determined dynamically within job execution context

internal/vault/logical_system_namespaces_seals.go
```diff
@@ -335,23 +374,204 @@ func (b *SystemBackend) handleNamespacesDeleteSealed() framework.OperationFunc {
...
+			}
+
+			if err := state.SealConfig.Validate(); err != nil {
+				return logical.ErrorResponse("invalid seal config: %v", err), err
+			}
+		}
 
-	POST /<path>/seal
-		Seal a namespace.
+		ns, err := b.Core.namespaceStore.GetNamespaceByPath(ctx, path)
```
wslabosz-reply, line 442: past this point I'd consider moving the logic of this handler to the `namespace_store.go` file as internal operation

internal/vault/invalidation.go
```diff
@@ -495,6 +495,12 @@ func (ij *invalidationJob) namespaceInvalidation(ctx context.Context) error {
 	case child == nil, ij.im.core.NamespaceSealed(child):
 		// Nothing to do.
 		return nil
+	case !deleted && ij.skipTaintedNamespace(ctx, childUUID):
+		// The namespace is mid-migration. Its barrier may be in the process of
+		// changing on the active node, so reloading mounts now could attempt to
+		// decrypt data written to a different barrier, causing decryption to
+		// fail. On namespace untaint, the namespace will be invalidated again.
```
wslabosz-reply:
```suggestion
		// The namespace is in the middle of barrier migration. Its barrier may be
		// changing on the active node, so reloading mounts now, could attempt to
		// decrypt data written with a different barrier, causing decryption to
		// fail. On namespace untaint, the namespace will be invalidated again.
```

internal/vault/seal_manager.go
```diff
@@ -101,15 +101,22 @@ func (sm *SealManager) Reset() {
 	sm.rotationConfigByNamespace = map[string]*rotationConfig{}
 }
 
+type SetSealOptions struct {
+	WriteToStorage bool
+	AllowOverride  bool
+	Seal           Seal
+	Barrier        barrier.SecurityBarrier
+}
```
wslabosz-reply: this all could be unexported

internal/vault/seal_manager.go
```diff
@@ -166,6 +180,17 @@ func (sm *SealManager) RemoveNamespace(ns *namespace.Namespace) {
 	sm.barrierByNamespacePath.Delete(ns.Path)
 }
 
+func (sm *SealManager) addNamespace(ns *namespace.Namespace, seal Seal, unlockInformation *unlockInformation, rotationConfig *rotationConfig, barrier barrier.SecurityBarrier) {
```
wslabosz-reply: what's the usecase for providing `unlockInformation` or `rotationConfig`?

For now this method evaluates just to:
```
sm.sealByNamespace[ns.UUID] = seal
sm.barrierByNamespacePath.Insert(ns.Path, barrier)
```
I'd consider inlining it this way.

website/content/docs/api/system/namespaces.mdx
```diff
@@ -285,3 +285,40 @@ $ curl \
     --request PUT \
     http://127.0.0.1:8200/v1/sys/namespaces/ns1/seal
 ```
+
+## Migrate a namespace barrier
+
+This endpoint allows migration of a namespace between a normal and sealable namespace.
```
wslabosz-reply: not sure but a bit more details what the operation entails could go in-handy, this also probably warrants a note on the `concepts/namespaces/sealable-namespaces` page

```suggestion
This endpoint allows migration from/to sealable/unsealable type of a namespace by removing or creating namespace barrier.
```

website/content/docs/api/system/namespaces.mdx
```diff
@@ -285,3 +285,40 @@ $ curl \
...
+| Method | Path                    |
+| :----- | :---------------------- |
+| `POST`  | `/sys/namespaces/:path/migrate-barrier` |
+
+### Sample payload
+
+```json
+{
+  "seal": "seal \"shamir\" {\n    shares = 5\n    threshold = 3\n}",
+  "pgp_keys": ["...", "...", "...", "...", "..."]
```
wslabosz-reply, line 302: @cipherboy 
one more reason to remove pgp support altogether

internal/vault/routing/router.go
```diff
@@ -1169,6 +1169,25 @@ func (r *Router) Invalidate(ctx context.Context, key string) bool {
...
+
+	r.root.WalkPrefix("", func(s string, v any) bool {
+		re := v.(*RouteEntry)
+
+		if re.MountEntry.Namespace.ID != ns.ID {
+			return false
+		}
+
+		re.Lock()
+		re.StorageView = barrier.NewView(b, re.StoragePrefix)
```
wslabosz-reply, line 1184: don't we miss updates to `re.StoragePrefix` and `(*Router).storagePrefix` radix?

internal/vault/routing/router.go
```diff
@@ -1169,6 +1169,25 @@ func (r *Router) Invalidate(ctx context.Context, key string) bool {
...
+func (r *Router) SetNamespaceStorageView(ns *namespace.Namespace, b barrier.SecurityBarrier) {
+	r.l.Lock()
+	defer r.l.Unlock()
+
+	r.root.WalkPrefix("", func(s string, v any) bool {
+		re := v.(*RouteEntry)
+
+		if re.MountEntry.Namespace.ID != ns.ID {
+			return false
+		}
```
wslabosz-reply: I think it might be possible to optimize with specification of the namespace prefix path?

internal/vault/namespace_store.go
```diff
@@ -1255,6 +1287,30 @@ func (ns *NamespaceStore) taintNamespace(ctx context.Context, parent, namespaceT
...
+	namespaceToUntaint.Tainted = false
+
+	err := ns.namespacesByPath.Insert(namespaceToUntaint)
+	if err != nil {
+		return fmt.Errorf("failed to modify namespace tree: %w", err)
+	}
+
+	nsCopy := namespaceToUntaint.Clone(true /* preserve unlock */)
+	if err := ns.writeNamespace(ctx, ns.core.NamespaceView(parent), nsCopy); err != nil {
+		return fmt.Errorf("failed to persist namespace taint: %w", err)
```
wslabosz-reply:
```suggestion
		return fmt.Errorf("failed to persist untainted namespace: %w", err)
```

internal/vault/namespace_store.go
```diff
@@ -1881,3 +1984,177 @@ func (j *namespaceCreationFailureJob) Execute() error {
 func (j *namespaceCreationFailureJob) OnFailure(err error) {
 	j.store.logger.Error("failed to handle namespace deletion following failed creation; job may be retried via deletion of tainted namespace", "namespace", j.target.Path, "ns_uuid", j.target.UUID, "error", err.Error())
 }
+
+type namespaceBarrierMigrationJob struct {
+	store         *NamespaceStore
+	core          *Core
```
wslabosz-reply: we don't need core, but `sealManager` iirc

internal/vault/namespace_store.go
```diff
@@ -1881,3 +1984,177 @@ func (j *namespaceCreationFailureJob) Execute() error {
...
+		err = j.core.sealManager.SetSeal(ctx, j.sealConfig, j.target, SetSealOptions{
+			WriteToStorage: true,
+			AllowOverride:  true,
+			Seal:           j.seal,
+			Barrier:        j.newBarrier,
+		})
+		if err != nil {
+			return err
+		}
+	}
```
wslabosz-reply: imho readability could be improved with if/else clause instead of switch

internal/vault/namespace_store.go
```diff
@@ -1881,3 +1984,177 @@ func (j *namespaceCreationFailureJob) Execute() error {
...
+
+	// invalidate outer most namespace first
+	slices.Reverse(namespacesToMigrate)
+
+	var errs error
```
wslabosz-reply: would prefer declaration & reverse action closer to the usage

internal/command/namespace_migrate_barrier.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/hashicorp/cli"
+	"github.com/openbao/openbao/api/v2"
+	"github.com/openbao/openbao/sdk/v2/helper/structtomap"
+	"github.com/openbao/openbao/v2/internal/helper/pgpkeys"
+	"github.com/posener/complete"
+)
+
+var (
+	_ cli.Command             = (*NamespaceCreateCommand)(nil)
+	_ cli.CommandAutocomplete = (*NamespaceCreateCommand)(nil)
```
wslabosz-reply:
```suggestion
	_ cli.Command             = (*NamespaceMigrateBarrierCommand)(nil)
	_ cli.CommandAutocomplete = (*NamespaceMigrateBarrierCommand)(nil)
```

internal/command/namespace_migrate_barrier.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (c *NamespaceMigrateBarrierCommand) Flags() *FlagSets {
+	set := c.flagSet(FlagSetHTTP | FlagSetOutputField | FlagSetOutputFormat)
+
+	f := set.NewFlagSet("Seal Options")
+	f.StringVar(&StringVar{
+		Name:       "seal",
+		Target:     &c.flagSealConfigPath,
+		Completion: complete.PredictFilesSet([]string{"*.hcl", "*.json"}),
+		Usage:      "Path to a HCL file with exactly one seal stanza.",
```
wslabosz-reply:
```suggestion
		Usage:      "Path to a HCL or JSON file with exactly one seal stanza.",
```

internal/vault/logical_system_namespaces_seals.go
phyrog: That's what we want when we migrate normal -> sealable

internal/vault/logical_system_namespaces_seals.go
phyrog: `Get` will actually `return nil, nil` if the entry does not exist, so it does not print on the first attempt. I'll change the `Info` to `Error`

internal/vault/logical_system_namespaces_seals.go
phyrog: We could check the ns type and the passed seal config earlier to check for this, yes

internal/vault/logical_system_namespaces_seals.go
phyrog: Oh this is actually a left-over debug print 🙂

internal/vault/logical_system_namespaces_seals.go
phyrog: But it also makes the job harder to test to do it dynamically

internal/vault/routing/router.go
phyrog: Storage paths stay the same after migration

internal/vault/seal_manager.go
phyrog: Yeah I think this is left-over from when I basically did a blue-green migration and needed to set this information for the new namespace. Since the migration is now in-place and the namespace UUID does not change, this is not necessary anymore.

api/sys_namespaces.go
phyrog: @wslabosz-reply do you have an opinion on this?

internal/vault/namespace_store.go
```diff
@@ -769,6 +785,15 @@ func (ns *NamespaceStore) pushToMounts(entry *namespace.Namespace) error {
...
 		mount.Namespace = entry
+
+		// re, ok := ns.core.router.Get(mount.Path)
+		// if !ok {
+		// 	ns.logger.Warn("route entry not found for mount", "path", mount.Path, "namespace", entry)
+		// 	continue
+		// }
+
+		// re.StorageView = NamespaceScopedView(ns.core.sealManager.NamespaceBarrierByLongestPrefix(entry.Path), entry)
+		// ns.logger.Warn("Setting route entry storage view", "re", re, "namespace", entry)
```
wslabosz-reply: is this needed?

internal/vault/namespace_store.go
```diff
@@ -761,6 +769,14 @@ func (ns *NamespaceStore) pushToMounts(entry *namespace.Namespace) error {
...
 
 		mount.Namespace = entry
+		// re, ok := ns.core.router.Get(mount.Path)
+		// if !ok {
+		// 	ns.logger.Warn("route entry not found for mount", "path", mount.Path, "namespace", entry)
+		// 	continue
+		// }
+
+		// re.StorageView = NamespaceScopedView(ns.core.sealManager.NamespaceBarrierByLongestPrefix(entry.Path), entry)
+		// ns.logger.Warn("Setting route entry storage view", "re", re, "namespace", entry)
```
wslabosz-reply: is this needed?

website/content/docs/api/system/namespaces.mdx
phyrog: @wslabosz-reply Anything you want me to do here? Otherwise I will resolve

api/sys_namespaces.go
wslabosz-reply: I get the "consistency" argument, but I'm also fine with there being just one method accepting context.

satoqz: IIRC I mentioned to @cipherboy a week ago or so that it'd be nice to make such changes globally in api v3, that is, context by default and options via the functional options pattern. So perhaps let's compromise here and *try* to keep v2 consistent while agreeing that long-term we'd like to clean this up?

website/content/docs/concepts/namespaces/sealable-namespaces.mdx
```diff
@@ -87,3 +87,11 @@ Both features can be used together. To temporarily block API requests to a names
 
 Sealable namespaces currently only support Shamir seals. Auto-unsealing support is
 planned, but not yet implemented. Contributions towards this are welcome.
+
+## Namespace Barrier Migration
+
+When migrating a namespace, all data of the namespace and its (normal) child namespaces will be re-encrypted with a new barrier. Any sealable child namespaces and their namespaces will remain unaffected.
```
wslabosz-reply:
```suggestion
When migrating a namespace, all namespace data and its unsealable child namespaces will be re-encrypted with a new barrier. Any sealable child namespaces and their namespaces will remain unaffected.
```

PR #3662, opened 2026-07-31
Fix missing OIDC default keys for namespaces

### LINKED ISSUE #3661: Missing OIDC default key after namespace creation
Note: I've had Claude analyze my issue and I'm trying to post the relevant content without copying it verbatim as per CONTRIBUTING.md.

**Describe the bug**
Openbao sometimes does not write the default key for the identity mount `identity/oidc/key/default` on namespace creation. The write is
not failing, it is being skipped, because the check that decides whether the key already exists
consults a cache keyed by the wrong namespace. This key is then durably missing and only fixed/created on writes to `identity/oidc/...`
In certain constellations this can lead to core restarts (see additional context).

**To Reproduce**
0. start bao in a new container, set up BAO_ADDR etc.
1. bao operator init -format=json -recovery-shares=1 -recovery-threshold=1 > init.json
2. set -gx BAO_TOKEN (jq -r '.root_token' init.json)
3. bao namespace create test1
```
Key                Value
---                -----
custom_metadata    map[]
id                 nLw4mh
locked             false
path               test1/
tainted            false
uuid               c1ec1805-f65e-6ed7-5f0d-521b2216e51c
```
4. bao read -namespace test1 identity/oidc/key/default
```
No value found at identity/oidc/key/default
```

```
bao-min-1  | 2026-07-31T08:34:35.860Z [INFO]  core: post-unseal setup complete
bao-min-1  | 2026-07-31T08:34:50.196Z [INFO]  core: successful mount: namespace=test1/ path=cubbyhole/ type=ns_cubbyhole version=""
bao-min-1  | 2026-07-31T08:34:50.196Z [INFO]  core: successful mount: namespace=test1/ path=sys/ type=ns_system version=""
bao-min-1  | 2026-07-31T08:34:50.205Z [DEBUG] core: wrote OIDC default provider
bao-min-1  | 2026-07-31T08:34:50.216Z [DEBUG] core: wrote OIDC allow_all assignment
bao-min-1  | 2026-07-31T08:34:50.216Z [INFO]  core: successful mount: namespace=test1/ path=identity/ type=ns_identity version=""
bao-min-1  | 2026-07-31T08:34:50.216Z [INFO]  core: enabled credential backend: namespace=test1/ path=token/ type=ns_token version=""
```

**Expected behavior**
Default key should always be created, e.g.
```
bao-min-1  | 2026-07-31T08:40:48.388Z [INFO]  core: successful mount: namespace=test2/ path=cubbyhole/ type=ns_cubbyhole version=""
bao-min-1  | 2026-07-31T08:40:48.388Z [INFO]  core: successful mount: namespace=test2/ path=sys/ type=ns_system version=""
bao-min-1  | 2026-07-31T08:40:48.399Z [DEBUG] core: wrote OIDC default provider
bao-min-1  | 2026-07-31T08:40:48.411Z [DEBUG] core: wrote OIDC default key
bao-min-1  | 2026-07-31T08:40:48.420Z [DEBUG] core: wrote OIDC allow_all assignment
bao-min-1  | 2026-07-31T08:40:48.421Z [INFO]  core: successful mount: namespace=test2/ path=identity/ type=ns_identity version=""
bao-min-1  | 2026-07-31T08:40:48.421Z [INFO]  core: enabled credential backend: namespace=test2/ path=token/ type=ns_token version=""
```

**Note:** I've gotten some mixed results on this reproducer, sometimes the key was written on a new container, sometimes not. I've
had the best "luck" with reproduction when creating the namespaces shortly after cluster / node creation.

**Environment:**
```
Seal Type                static
Recovery Seal Type       shamir
Initialized              true
Total Recovery Shares    1
Threshold                1
Version                  2.6.1
Commit Date              2026-07-22T14:22:20Z
Storage Type             raft
Cluster Name             vault-cluster-95259089
Cluster ID               98640ee1-1b10-b702-2ba5-f4faac95c9ae
HA Enabled               true
HA Cluster               https://bao1:8201
HA Mode                  active
Active Since             2026-07-31T08:38:28.644362605Z
Raft Committed Index     45
Raft Applied Index       45
```
```
OpenBao v2.6.1 (ba7ad8861d0578cd4da4f7b9e5a6756d30484f8f), committed 2026-07-22T14:22:20Z
```
* Server Operating System/Architecture: Debian Linux, WSL2

```hcl
ui            = false
disable_mlock = true

log_level = "debug"

storage "raft" {
  path    = "/openbao/file"
  node_id = "bao1"
}

listener "tcp" {
  address     = "0.0.0.0:8200"
  tls_disable = true
}

api_addr     = "http://bao1:8200"
cluster_addr = "http://bao1:8201"

seal "static" {
  current_key_id = "min-1"
  current_key    = "file:///openbao/seal.key"
}
```

**Additional context**
I noticed this while trying to set up an openbao test cluster (3 nodes) using tofu + ansible. Sometimes tofu failed with lots of errors like `Error: Provider produced inconsistent result after apply`, often related to namespaces.

I've fed Claude with my tofu, server and audit logs and it turned out that my ansible scripts were faulty and contained a simultaneous restart of the fully formed bao cluster, straight before / while running tofu.
This caused leaderless states, leader elections, step downs etc.

Claude then also dug up the above issue and core restart messages from a standby node.
```
ul 29 09:12:34 openbao-03 bao[6767]: 2026-07-29T09:12:34.012Z [INFO]  core: enabled credential backend: namespace="" path=oidc/ type=oidc version=""
Jul 29 09:12:34 openbao-03 bao[6767]: 2026-07-29T09:12:34.028Z [INFO]  core: enabled credential backend: namespace="" path=break-glass/ type=userpass version=""
Jul 29 09:12:34 openbao-03 bao[6767]: 2026-07-29T09:12:34.061Z [INFO]  core: calling mount internal: path=identity/
Jul 29 09:12:34 openbao-03 bao[6767]: 2026-07-29T09:12:34.062Z [ERROR] core: failed to write OIDC default resources to storage: error="error writing default key to storage: cannot write to readonly storage"
Jul 29 09:12:34 openbao-03 bao[6767]: 2026-07-29T09:12:34.062Z [ERROR] core.invalidate-dispatch: fatal failure dispatching invalidation; restarting core: key=core/namespaces/22a6903a-6838-1c1d-5bda-2a7cf9746e0a err="unable to invalidate mounts in namespace \"00000000-0000-0000-0000-000000000000\": unable to invalidate mount for key \"core/mounts/2af394a1-fd29-7b44-79ab-f2ca99bf3ffe\" in namespace \"22a6903a-6838-1c1d-5bda-2a7cf9746e0a\": error writing default key to storage: cannot write to readonly storage"
```

The code path for the missing write of the default key is the following according to claude. To be frank, I'm lacking the go and openbao knowledge to fully understand it:

https://github.com/openbao/openbao/blob/ba7ad8861d0578cd4da4f7b9e5a6756d30484f8f/vault/ha.go#L709-L710
-> the core's active context is ROOT-scoped

https://github.com/openbao/openbao/blob/ba7ad8861d0578cd4da4f7b9e5a6756d30484f8f/vault/mount.go#L343-L344
-> ctx is ROOT-scoped, but `view` belongs to the NEW namespace.
   This is the actual defect: ctx and storage disagree about which namespace we are in.

https://github.com/openbao/openbao/blob/ba7ad8861d0578cd4da4f7b9e5a6756d30484f8f/vault/identity/store_oidc_provider.go#L2668-L2692

provider check   -> view.Get(...)        <- uses the VIEW  (correct ns)
ensureDefaultKey -> getNamedKey(...)     <- uses the CTX   (root!)
assignment check -> view.Get(...)        <- uses the VIEW  (correct ns)

https://github.com/openbao/openbao/blob/ba7ad8861d0578cd4da4f7b9e5a6756d30484f8f/vault/identity/store_oidc.go#L1023-L1036

ns, _ := namespace.FromContext(ctx)      <- ROOT
i.OidcCache.Get(ns, "namedKeys/default") <- v0:root:namedKeys/default
if found -> return that key, never touching storage

https://github.com/openbao/openbao/blob/ba7ad8861d0578cd4da4f7b9e5a6756d30484f8f/vault/identity/store_oidc_provider.go#L2714-L2721
key, _ := getNamedKey(...)
if key != nil { return key, nil }        <- early return, no write

Claude was able to find a related path to reproduce a core restart (and the same log messages) using a go test, both against 2.6.1 and main. I can supply the full test if requested, I skipped it for now since it's AI generated. Here's a short summary, also from claude:

Preconditions:

* Two nodes sharing storage, core1 active, core2 standby
* core2 joined before the namespace existed, so it has never mounted or discovered it. This forces it to learn about the namespace via invalidation rather than at its own startup.
* core2 has no cached entry for root's OIDC default key

Trigger

* The active node creates a namespace, which comes out poisoned (provider and allow_all written, key not).
* The standby receives the core/namespaces/<uuid> invalidation — its first encounter with that namespace.

What follows, automatically

* namespaceInvalidation → reloadNamespaceMounts → reloadMount → mounts ns_identity fresh → calls Initialize.
* StoreOIDCDefaultResources → ensureDefaultKey → getNamedKey: cache miss (root's entry is gone) → falls through to storage on the namespace's own view → nil → attempts Put.
* The Put fails with ErrReadOnly (cannot write to readonly storage).
* The error propagates up; the namespace-store branch set fatal = true (invalidation.go:394).
* OnFailure (invalidation.go:568) → core.restart().

I'm happy to share the full test file or supply additional input.

wslabosz-reply:
## Description
Identity backend `initialization` method did not get passed context with a respective namespace, also cred/secrets enable methods (also used by invalidation logic) were passing `(*Core).activeContext()` to the backend init method. This in turn made the `OIDC` default keys read (and in turn write) fail, **because we never wrote them for the namespace on the active node in the first place**. This PR changes that (the presence logic was always looking at the root namespace-owned) cache.

## Rationale
We never wrote the OIDC default keys to the namespace space, so adjusting that and changing the context propagation fixes the standby nodes behavior also.

This bug also affected raft multi-node (1+) clusters (multiple nodes at namespace creation) in a sense that standby nodes while checking for `Default` key existence in cache were always using the namespace in ctx (which would be `root` namespace [(*Core) activeContext]), but because it was present, no error from lack of key occurred.

### Why we never wrote `Default` OIDC keys
The logic for "presence" of the default keys is: look at cache -> look at storage.
Because init of the identity backend for namespaces used a context with  *root namespace* stored it was always finding the keys of the root namespace, so we bail out from creation of the keys entry in the namespace storage.

Then for standby nodes the cache is not populated nor keys are in storage, hence the failure.

Resolves: #3661

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3662.txt
+++ b/changelog/3662.txt
@@ -1,0 +1,6 @@
+```release-note:bug
+core/identity: Start writing OIDC 'default' keys per-namespace storage.
+```
+```release-note:bug
+core/namespaces: Fix fatal failures writing 'default' OIDC keys on standby nodes during namespace creation.
+```
--- a/internal/vault/auth.go
+++ b/internal/vault/auth.go
@@ -199,7 +199,9 @@ func (c *Core) enableCredentialInternalWithLock(ctx context.Context, entry *rout
 	// Initialize() if necessary
 	view.SetReadOnlyErr(origViewReadOnlyErr)
 
-	if err := backend.Initialize(ctx, &logical.InitializationRequest{Storage: view}); err != nil {
+	// Initialize, using the core's active context.
+	activeCtx := namespace.ContextWithNamespace(c.activeContext.Load(), ns)
+	if err := backend.Initialize(activeCtx, &logical.InitializationRequest{Storage: view}); err != nil {
 		return err
 	}
 
--- a/internal/vault/auth.go
+++ b/internal/vault/auth.go
@@ -198,9 +198,8 @@ func (c *Core) enableCredentialInternalWithLock(ctx context.Context, entry *rout
 	// restore the original readOnlyErr, so we can write to the view in
 	// Initialize() if necessary
 	view.SetReadOnlyErr(origViewReadOnlyErr)
-	// initialize, using the core's active context.
-	err = backend.Initialize(c.activeContext.Load(), &logical.InitializationRequest{Storage: view})
-	if err != nil {
+
+	if err := backend.Initialize(ctx, &logical.InitializationRequest{Storage: view}); err != nil {
 		return err
 	}
 
@@ -1200,8 +1199,7 @@ func (c *Core) setupCredential(ctx context.Context, entry *routing.MountEntry) (
 			view.SetReadOnlyErr(origViewReadOnlyErr)
 		}
 
-		err := backend.Initialize(ctx, &logical.InitializationRequest{Storage: view})
-		if err != nil {
+		if err := backend.Initialize(namespace.ContextWithNamespace(ctx, localEntry.Namespace), &logical.InitializationRequest{Storage: view}); err != nil {
 			postUnsealLogger.Error("failed to initialize auth backend", "error", err)
 		}
 	}
--- a/internal/vault/mount.go
+++ b/internal/vault/mount.go
@@ -341,7 +341,9 @@ func (c *Core) mountInternalWithLock(ctx context.Context, entry *routing.MountEn
 	// Initialize() if necessary
 	view.SetReadOnlyErr(origReadOnlyErr)
 
-	if err := backend.Initialize(ctx, &logical.InitializationRequest{Storage: view}); err != nil {
+	// Initialize, using the core's active context.
+	activeCtx := namespace.ContextWithNamespace(c.activeContext.Load(), ns)
+	if err := backend.Initialize(activeCtx, &logical.InitializationRequest{Storage: view}); err != nil {
 		return err
 	}
 
--- a/internal/vault/mount.go
+++ b/internal/vault/mount.go
@@ -340,9 +340,8 @@ func (c *Core) mountInternalWithLock(ctx context.Context, entry *routing.MountEn
 	// restore the original readOnlyErr, so we can write to the view in
 	// Initialize() if necessary
 	view.SetReadOnlyErr(origReadOnlyErr)
-	// initialize, using the core's active context.
-	err = backend.Initialize(c.activeContext.Load(), &logical.InitializationRequest{Storage: view})
-	if err != nil {
+
+	if err := backend.Initialize(ctx, &logical.InitializationRequest{Storage: view}); err != nil {
 		return err
 	}
 
@@ -1572,8 +1571,7 @@ func (c *Core) setupMount(ctx context.Context, entry *routing.MountEntry) (func(
 			view.SetReadOnlyErr(origReadOnlyErr)
 		}
 
-		err := backend.Initialize(ctx, &logical.InitializationRequest{Storage: view})
-		if err != nil {
+		if err := backend.Initialize(namespace.ContextWithNamespace(ctx, localEntry.Namespace), &logical.InitializationRequest{Storage: view}); err != nil {
 			postUnsealLogger.Error("failed to initialize mount backend", "error", err)
 		}
 	}


internal/vault/auth.go
```diff
@@ -198,9 +198,8 @@ func (c *Core) enableCredentialInternalWithLock(ctx context.Context, entry *rout
 	// restore the original readOnlyErr, so we can write to the view in
 	// Initialize() if necessary
 	view.SetReadOnlyErr(origViewReadOnlyErr)
-	// initialize, using the core's active context.
-	err = backend.Initialize(c.activeContext.Load(), &logical.InitializationRequest{Storage: view})
```
cipherboy, line 202: I'm confused, isn't `c.activeContext` set anyways, even on standby nodes? 🤔

wslabosz-reply: it's not about presence of ctx, but the namespace stored in context 😅

cipherboy: Ah! Yeah, that'd do it. :laughing:

cipherboy: I wonder how plugins handle `init->cancel->....`? Are they marked tainted and only deleted? Is init retried? &c. Request context might be cancelled I agree, but active context should hold lock long enough that unless the plugin's init is really slow, it should still continue... 🤔

internal/vault/mount.go
```diff
@@ -340,9 +340,8 @@ func (c *Core) mountInternalWithLock(ctx context.Context, entry *routing.MountEn
 	// restore the original readOnlyErr, so we can write to the view in
 	// Initialize() if necessary
 	view.SetReadOnlyErr(origReadOnlyErr)
-	// initialize, using the core's active context.
-	err = backend.Initialize(c.activeContext.Load(), &logical.InitializationRequest{Storage: view})
-	if err != nil {
+
+	if err = backend.Initialize(ctx, &logical.InitializationRequest{Storage: view}); err != nil {
```
satoqz:
```suggestion
	if err := backend.Initialize(ctx, &logical.InitializationRequest{Storage: view}); err != nil {
```

nit, since we're here already.

internal/vault/auth.go
```diff
@@ -198,9 +198,8 @@ func (c *Core) enableCredentialInternalWithLock(ctx context.Context, entry *rout
...
-	if err != nil {
+
+	if err = backend.Initialize(ctx, &logical.InitializationRequest{Storage: view}); err != nil {
```
satoqz:
```suggestion
	if err := backend.Initialize(ctx, &logical.InitializationRequest{Storage: view}); err != nil {
```

changelog/3662.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:bug
+core/identity: Start writing OIDC 'default' keys per-namespace. Fix fatal failures panics on
```
satoqz: on...?

wslabosz-reply: I've wanted to do it with one changelog entry, but then decided to do two, but forgot to fix the first one lol

satoqz (dismissed):

cipherboy (dismissed):
I do still maintain the original comment in https://github.com/openbao/openbao/pull/3662#discussion_r3691530985

We should probably derive `activeContext` here, as to my knowledge there's no way to re-try failed initialization and I'm not sure we get marked as tainted to be slated for deletion only, but we probably warrant a deeper investigation of mount enabling here to deal with that. 

Feel free to merge, though, and we can revisit later.

satoqz:
@cipherboy updated this for @wslabosz-reply, let's use the active context to be safe.

cipherboy (approved):
Thanks, merging!

PR #3669, opened 2026-08-03
fix: keep HTTP server running when PROXY protocol rejects unauthorized connections

### LINKED ISSUE #3664: BUG: proxy_protocol_behavior = "deny_unauthorized" Causes the HTTP server to die
**Describe the bug**
When setting proxy_protocol_behavior = "deny_unauthorized" and a list of proxy_protocol_authorized_addrs then trying to access the server directly from any client including itself, the htpp server dies with this error message in the logs "HTTP server (listening on 0.0.0.0:8200) exited with error: upstream connection not trusted proxy_protocol_behavior is "deny_unauthorized"

**To Reproduce**
1. Configure proxy_protocol_behavior with deny_unauthorized
2. Add some IPs either in coma separated or json list to proxy_protocol_authorized_addrs
3. Try to access any endpoint using curl (didn't test with bao client) for example sys/health on localhost or server IP.
4. Server immediately crashes and exists with above error.

**Expected behavior**
I expect the connection to be simply rejected as the docss says not the entire HTTP server killed .

**Environment:**
* OpenBao v2.6.0-1

```hcl
# Be sure to scrub any sensitive values
```

**Additional context**

atharv-sys32:
Fixes #3664.

## Problem

With `proxy_protocol_behavior = "deny_unauthorized"`, a connection from a non-authorized address kills the entire HTTP server instead of just rejecting that connection:

```
HTTP server (listening on 0.0.0.0:8200) exited with error: upstream connection not trusted proxy_protocol_behavior is "deny_unauthorized"
```

## Root cause

In `internal/helper/proxyutil/proxyutil.go`, the policy for `deny_unauthorized` returned a custom `errors.New(...)`. go-proxyproto's `Accept()` (v0.14.0, protocol.go:147-158) propagates any non-`ErrInvalidUpstream` error from the policy up to the caller, so the HTTP server treats it as fatal and exits.

## Fix

Return `proxyproto.ErrInvalidUpstream` from the policy. go-proxyproto then closes the unauthorized connection and continues accepting (protocol.go:153-156), keeping the listener (and HTTP server) alive. Unauthorized connections are still rejected as documented.

## Tests

- Added `TestTCPListener_proxyProtocolDenyUnauthorizedKeepsListening`: asserts an unauthorized connection is closed AND the listener still accepts authorized connections afterwards.
- Updated the existing table test: `deny_unauthorized` not-authorized cases now expect rejection-without-server-death, matching the fixed behavior.
- All `TestTCPListener*` and `internal/helper/proxyutil` tests pass; `gofmt` and `go vet` clean.

## Changelog

`changelog/3669.txt` added (bug fix, release note included).

### DIFF
--- a/changelog/3669.txt
+++ b/changelog/3669.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Keep the HTTP server running when PROXY protocol rejects unauthorized connections instead of shutting down.
+```
--- a/internal/command/server/listener_tcp_test.go
+++ b/internal/command/server/listener_tcp_test.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/cli"
 	"github.com/hashicorp/go-sockaddr"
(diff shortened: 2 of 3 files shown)


atharv-sys32:
Hi @cipherboy @satoqz, quick note on this one. It fixes #3664: with `proxy_protocol_behavior = \deny_unauthorized\`, an unauthorized connection was killing the whole HTTP server. The policy now returns `proxyproto.ErrInvalidUpstream` so go-proxyproto rejects the connection and keeps accepting. Added a regression test that proves the listener survives a rejected connection. Tests, gofmt, and vet are all green. Would appreciate a review when you have a moment.

satoqz:
https://github.com/openbao/openbao/blob/main/.github/pull_request_template.md

PR #3672, opened 2026-08-03
Fix PrintRawField function by adding warnings check

### LINKED ISSUE #3670: bao write -field=[...] discards warnings (e.g. unrecognized parameter) in CLI
When a request returns a warning (e.g. 'Endpoint ignored the unrecognized parameters: [...]') the CLI shows it correctly in both default and JSON (`-format=json`) output but if `-field=[...]` is used to extract a single field's value, the warning is discarded entirely. It's neither printed on stdout nor stderr.

**To Reproduce**
1. The following all within the CLI
2. Enable the transit secrets engine
3. Create an encryption key 
4. Generate HMAC with wrong field name (e.g. 'hash_algorithm' instead of 'algorithm')
**Expected behavior**
When running `bao write -field=hmac transit/hmac/test-key input=[...] hash_algorithm=sha2-512` it should print the warning:
`* Endpoint ignored these unrecognized parameters: [hash_algorithm]`

**Environment:**
* OpenBao CLI Version (retrieve with `bao version`): OpenBao v2.5.5 (028992583c693c4de6350b8aa52ff85e30375a99), built 2026-06-17T11:18:48Z
* Server Operating System/Architecture: Alpine Linux 3.24 (container), arm64, running on a Linux 6.10 kernel node

```hcl
ui = true

listener "tcp" {
  address         = "[::]:8200"
  cluster_address = "[::]:8201"
  tls_disable     = 0
  tls_cert_file   = "/openbao/userconfig/openbao-tls/server-cert.pem"
  tls_key_file    = "/openbao/userconfig/openbao-tls/server-key.pem"
}

storage "raft" {
  path = "/openbao/data"

  retry_join {
    leader_api_addr     = "https://openbao-0.openbao-internal:8200"
    leader_ca_cert_file = "/openbao/userconfig/openbao-tls/ca.pem"
  }
  retry_join {
    leader_api_addr     = "https://openbao-1.openbao-internal:8200"
    leader_ca_cert_file = "/openbao/userconfig/openbao-tls/ca.pem"
  }
  retry_join {
    leader_api_addr     = "https://openbao-2.openbao-internal:8200"
    leader_ca_cert_file = "/openbao/userconfig/openbao-tls/ca.pem"
  }
}

audit "file" "file-audit" {
  description = "transit spike audit sink"
  options {
    file_path = "/openbao/audit/audit.log"
  }
}
```

**Additional context**
Found while investigating issue #3667 . The CLI handles the same underlying warning correctly in default and JSON output mode, which is what led to noticing that `-field` mode specifically loses it. When using the terminal within the GUI / web UI it does show the warning as a popup.

Cloonson:
## Description
Added warnings check to `PrintRawField` function to handle warnings similarly to the `printWarnings` function in `format.go`. This PR adds the missing warnings check.

## Rationale
This Bug occured because the `PrintRawField` function didn't do a warnings check, thus all warnings (e.g. wrong parameter) were discarded whenever the `-field` flag was used. The Bug only happened in the CLI and is fixed by adding a warnings check to `PrintRawField`.

Resolves: #3670 

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3672.txt
+++ b/changelog/3672.txt
@@ -1,0 +1,2 @@
+```release-note:bug
+cli: Fix `-field=` flag discarding API warnings (e.g. unrecognized parameters) instead of printing them to stderr
--- a/changelog/3672.txt
+++ b/changelog/3672.txt
@@ -1,2 +1,3 @@
 ```release-note:bug
-cli: Fix `-field=` flag discarding API warnings (e.g. unrecognized parameters) instead of printing them to stderr
+command: Fix `-field=` flag discarding API warnings (e.g., unrecognized parameters) instead of printing them to stderr.
+```
--- a/internal/command/util.go
+++ b/internal/command/util.go
@@ -105,11 +105,12 @@ func PrintRawField(ui cli.Ui, data any, field string) int {
 		ui.Error(fmt.Sprintf("Field %q not present in secret", field))
 		return 1
 	}
-	if secret, ok := data.(*api.Secret); ok && Format(ui) == "table" {
+	format := Format(ui)
+
+	if secret, ok := data.(*api.Secret); ok && format == "table" {
 		printWarnings(ui, secret)
 	}
 
-	format := Format(ui)
 	if format == "" || format == "table" || format == "raw" {
 		return PrintRaw(ui, fmt.Sprintf("%v", val))
 	}
--- a/internal/command/util.go
+++ b/internal/command/util.go
@@ -105,13 +105,8 @@ func PrintRawField(ui cli.Ui, data any, field string) int {
 		ui.Error(fmt.Sprintf("Field %q not present in secret", field))
 		return 1
 	}
-	// Really similar to the warning used in openbao/internal/command/format.go. The PrintRawField function was simply missing the warnings check.
-	if secret, ok := data.(*api.Secret); ok && secret != nil && len(secret.Warnings) > 0 {
-		ui.Warn("WARNING! The following warnings were returned from OpenBao\n")
-		for _, warning := range secret.Warnings {
-			ui.Warn(wrapAtLengthWithPadding(fmt.Sprintf("* %s", warning), 2))
-			ui.Warn("")
-		}
+	if secret, ok := data.(*api.Secret); ok {
+		printWarnings(ui, secret)
 	}
 
 	format := Format(ui)
--- a/internal/command/util.go
+++ b/internal/command/util.go
@@ -106,9 +106,9 @@ func PrintRawField(ui cli.Ui, data any, field string) int {
 		return 1
 	}
 	// Really similar to the warning used in openbao/internal/command/format.go. The PrintRawField function was simply missing the warnings check.
-	if secret, ok := data.(*api.Secret); ok && secret != nil && len(secret.Warnings) > 0{
+	if secret, ok := data.(*api.Secret); ok && secret != nil && len(secret.Warnings) > 0 {
 		ui.Warn("WARNING! The following warnings were returned from OpenBao\n")
-		for _, warning := range secret.Warnings{
+		for _, warning := range secret.Warnings {
 			ui.Warn(wrapAtLengthWithPadding(fmt.Sprintf("* %s", warning), 2))
 			ui.Warn("")
 		}
--- a/internal/command/util.go
+++ b/internal/command/util.go
@@ -105,6 +105,14 @@ func PrintRawField(ui cli.Ui, data any, field string) int {
 		ui.Error(fmt.Sprintf("Field %q not present in secret", field))
 		return 1
 	}
+	// Really similar to the warning used in openbao/internal/command/format.go. The PrintRawField function was simply missing the warnings check.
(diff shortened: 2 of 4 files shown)


internal/command/util.go
```diff
@@ -105,6 +105,14 @@ func PrintRawField(ui cli.Ui, data any, field string) int {
 		ui.Error(fmt.Sprintf("Field %q not present in secret", field))
 		return 1
 	}
+	// Really similar to the warning used in openbao/internal/command/format.go. The PrintRawField function was simply missing the warnings check.
+	if secret, ok := data.(*api.Secret); ok && secret != nil && len(secret.Warnings) > 0{
+		ui.Warn("WARNING! The following warnings were returned from OpenBao\n")
+		for _, warning := range secret.Warnings{
+			ui.Warn(wrapAtLengthWithPadding(fmt.Sprintf("* %s", warning), 2))
+			ui.Warn("")
+		}
```
satoqz: This will write to stdout, not stderr, and would thus break any scripts that users have built around it (which is roughly what `-field` is meant to be used for).

It doesn't seem like `hashicorp/cli` lets us write to stderr unless we're calling `ui.Error()`, so... not sure we can really improve this outside of declaring it works-as-intended.

satoqz (changes requested):

internal/command/util.go
Cloonson: I tested it and `bao write -field=hmac [...] 2>/dev/null` shows only the raw value. Running the same command with `1>/dev/null` shows only the warning. So this should be going to stderr correctly, not stdout

satoqz: Thanks for testing, my bad, I had glanced over the code and hadn't realized that warn calls into error. Makes sense to me then!

internal/command/util.go
```diff
@@ -105,6 +105,14 @@ func PrintRawField(ui cli.Ui, data any, field string) int {
 		ui.Error(fmt.Sprintf("Field %q not present in secret", field))
 		return 1
 	}
+	// Really similar to the warning used in openbao/internal/command/format.go. The PrintRawField function was simply missing the warnings check.
```
satoqz: This comment will make no sense outside of the PR's context. You can just omit it.

internal/command/util.go
```diff
@@ -105,6 +105,14 @@ func PrintRawField(ui cli.Ui, data any, field string) int {
...
+	if secret, ok := data.(*api.Secret); ok && secret != nil && len(secret.Warnings) > 0 {
+		ui.Warn("WARNING! The following warnings were returned from OpenBao\n")
+		for _, warning := range secret.Warnings {
+			ui.Warn(wrapAtLengthWithPadding(fmt.Sprintf("* %s", warning), 2))
+			ui.Warn("")
+		}
```
satoqz: `TableFormatter.printWarnings` doesn't use any part of `TableFormatter` so you could make it a normal function and call into that from here.

Cloonson: I changed kv_get.go as well. I also removed the wrapper because `printWarnings` already does the check internally (`if len(secret.Warnings) > 0`)

internal/command/util.go
Cloonson: Thank you for your help :)

internal/command/util.go
```diff
@@ -105,6 +105,9 @@ func PrintRawField(ui cli.Ui, data any, field string) int {
 		ui.Error(fmt.Sprintf("Field %q not present in secret", field))
 		return 1
 	}
+	if secret, ok := data.(*api.Secret); ok {
+		printWarnings(ui, secret)
+	}
```
satoqz: I wonder if it should only print if `format == table`, not sure though 🤷

changelog/3672.txt
```diff
@@ -0,0 +1,2 @@
+```release-note:bug
+cli: Fix `-field=` flag discarding API warnings (e.g. unrecognized parameters) instead of printing them to stderr
```
satoqz:
````suggestion
```release-note:bug
command: Fix `-field=` flag discarding API warnings (e.g., unrecognized parameters) instead of printing them to stderr.
```
````

File: internal/command/util.go (hunk 4 above)
Comment (Cloonson) [reply internal/command/util.go]: You're right. it's only ever called from `TableFormatter` so warnings have always been table format only. I added `&& Format(ui) == table` to the check

File: internal/command/util.go [hunk 6]
```diff
@@ -105,6 +105,9 @@ func PrintRawField(ui cli.Ui, data any, field string) int {
 		ui.Error(fmt.Sprintf("Field %q not present in secret", field))
 		return 1
 	}
+	if secret, ok := data.(*api.Secret); ok && Format(ui) == "table" {
+		printWarnings(ui, secret)
+	}
 
 	format := Format(ui)
```
satoqz:
```suggestion
	format := Format(ui)

	if secret, ok := data.(*api.Secret); ok && format == "table" {
		printWarnings(ui, secret)
	}

```

satoqz (approved):
LGTM, thanks!

Test failures are unrelated.

PR #3673, opened 2026-08-03
Fix `enable_rate_limit_audit_logging` not working

### LINKED ISSUE #3659: Audit logging rate limit errors does not work
**Describe the bug**
When audit-logging rate limit errors with `enable_rate_limit_audit_logging` for rate limits applied to a namespace, OpenBao throws a "no namespace" error and an aggregate error. This is because the context passed to the audit request does not contain namespace information. 

Changing 
https://github.com/openbao/openbao/blob/028992583c693c4de6350b8aa52ff85e30375a99/http/util.go#L214
to use `ctx` instead of `r.Context()` would likely fix this issue.

**To Reproduce**
- enable rate limit audit logging
- set a rate limit for a namespace
- set up an audit backend
- exceed the rate limit for the namespace

**Environment:**
At least OpenBao versions 2.4, 2.5, 2.6

```hcl
ui = true
cluster_addr = "https://127.0.0.1:8201"

listener "tcp" {
  address = "127.0.0.1:8200"
  tls_disable = "true"
}

storage "raft" {
  path = "/tmp/openbao-raft"
  node_id = "asdf"
}

seal "static" {
  current_key_id = "20260119-1"
  current_key = "file:///tmp/openbao-unseal.key"
}

audit "file" "file" {
  description = "asdf"
  options {
    file_path = "/tmp/openbao-audit.log"
    log_raw = "true"
  }
}
```

phil9909:
## Description

The audit backend expects a namespace in the context.

In addition to fixing this, should we change this from an error to generating a audit log with `namespace: <unknown>` (+ error log)?

## Rationale

Resolves: #3659

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3673.txt
+++ b/changelog/3673.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/quotas: fix `enable_rate_limit_audit_logging` did not work
+```
--- a/changelog/3673.txt
+++ b/changelog/3673.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-core/quotas: fix `enable_rate_limit_audit_logging` did not work
+sys/quotas: Fix broken `enable_rate_limit_audit_logging` option.
 ```
--- a/internal/http/util.go
+++ b/internal/http/util.go
@@ -213,7 +213,7 @@ func rateLimitQuotaWrapping(handler http.Handler, core *vault.Core) http.Handler
 					return
 				}
 
-				err = core.AuditLogger().AuditRequest(r.Context(), &logical.LogInput{
+				err = core.AuditLogger().AuditRequest(ctx, &logical.LogInput{
 					Request:  req,
 					OuterErr: quotaErr,
 				})
--- a/internal/vault/external_tests/quotas/quotas_test.go
+++ b/internal/vault/external_tests/quotas/quotas_test.go
@@ -4,6 +4,10 @@
 package quotas
 
 import (
+	"encoding/json"
+	"os"
+	"path"
+	"strings"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -13,10 +17,13 @@ import (
 	"github.com/openbao/openbao/sdk/v2/logical"
 	"github.com/stretchr/testify/require"
 
+	"github.com/openbao/openbao/v2/internal/audit"
+	auditFile "github.com/openbao/openbao/v2/internal/builtin/audit/file"
 	"github.com/openbao/openbao/v2/internal/builtin/credential/userpass"
 	"github.com/openbao/openbao/v2/internal/builtin/logical/pki"
 	"github.com/openbao/openbao/v2/internal/helper/testhelpers/teststorage"
 	"github.com/openbao/openbao/v2/internal/vault"
+	"github.com/openbao/openbao/v2/internal/vault/quotas"
 )
 
 var coreConfig = &vault.CoreConfig{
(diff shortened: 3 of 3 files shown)


cipherboy:
> In addition to fixing this, should we change this from an error to generating a audit log with "namespace: " (+ error log)?

I do wonder if there's a purpose in fixing this? Except in the stdout case, audit logging rate limited requests seems like a great way to overwhelm your downstream audit logging service. The request is fully dropped prior to any servicing occurring, so it seems to me like we should probably just drop it...? It was never processed after all.

My 2c., but unless y'all have a use case for it, I'm also find deprecating/removing it given its been broken since 2.3 (over a year at this point).

phil9909:
> In addition to fixing this, should we change this from an error to generating a audit log with "namespace: <unknown>" (+ error log)?

This should have been `"namespace: <unknown>"` GitHub seems to filter this, if it is not in a code block, because it looks like a HTML tag. Maybe `&lt;` works: "&lt;unknown>". It does, learned something today.


> Except in the stdout case, audit logging rate limited requests seems like a great way to overwhelm your downstream audit logging service.

Certainly true.

> My 2c., but unless y'all have a use case for it, I'm also find deprecating/removing it given its been broken since 2.3 (over a year at this point).

This was reported to us by one of our customers. Based on the fact that it is an opt-in feature (via `enable_rate_limit_audit_logging` being `false` by default) they seem to have an use-case for this, but I don't know exactly why they enabled it.
As this is clearly a bug in an existing, currently not deprecated feature and the fix is straight forward IMHO we should just fix it. We can certainly deprecate it with 2.7 (and maybe offer a better replacement).

cfi2017:
> > In addition to fixing this, should we change this from an error to generating a audit log with "namespace: " (+ error log)?
> 
> This should have been `"namespace: <unknown>"` GitHub seems to filter this, if it is not in a code block, because it looks like a HTML tag. Maybe `&lt;` works: "<unknown>". It does, learned something today.
> 
> > Except in the stdout case, audit logging rate limited requests seems like a great way to overwhelm your downstream audit logging service.
> 
> Certainly true.
> 
> > My 2c., but unless y'all have a use case for it, I'm also find deprecating/removing it given its been broken since 2.3 (over a year at this point).
> 
> This was reported to us by one of our customers. Based on the fact that it is an opt-in feature (via `enable_rate_limit_audit_logging` being `false` by default) they seem to have an use-case for this, but I don't know exactly why they enabled it. As this is clearly a bug in an existing, currently not deprecated feature and the fix is straight forward IMHO we should just fix it. We can certainly deprecate it with 2.7 (and maybe offer a better replacement).

This was enabled to debug an unrelated issue. I agree that it's probably not sensible to log every rate limit violation, but perhaps some kind of batched log of how many requests were dropped due to rate limiting for a namespace if it occurs might be more sensible? 

From what I can tell (feel free to correct me), metrics only show rate limit events across an entire OpenBao instance.

changelog/3673.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+core/quotas: fix `enable_rate_limit_audit_logging` did not work
```
satoqz:
```suggestion
sys/quotas: Fix broken `enable_rate_limit_audit_logging` option.
```

satoqz (commented):
I'm on board with just getting it fixed for now.

satoqz (approved):

cipherboy:
> This was enabled to debug an unrelated issue. I agree that it's probably not sensible to log every rate limit violation, but perhaps some kind of batched log of how many requests were dropped due to rate limiting for a namespace if it occurs might be more sensible?

Yeah, this makes sense to me as a reason to keep it. Thanks!

cipherboy (approved):

PR #3682, opened 2026-08-04
Add transit key count metric

phil9909:
## Description

Add a new metric `vault.secret.transit.key.count` similar to the existing [`vault.secret.kv.count`](https://openbao.org/docs/internals/telemetry/metrics/secrets/#vault-secret-kv-count) metric.

It count the number of keys in the transit engine grouped by mount (instead of key-value pairs in the kv engine).

It is opt in and can be enabled via `BAO_ENABLE_TRANSIT_GAUGE` (instead of being opt out via `BAO_DISABLE_KV_GAUGE`).

Other changes:

- Refactor `AddTestLogicalBackend` to use `t.Cleanup` (which did not exist when this helper was writen)
- Document the `BAO_DISABLE_KV_GAUGE` environment variable

## Open Questions

- Should we check that the `BAO_ENABLE_TRANSIT_GAUGE` env variable has a "truthy" value (currently `BAO_ENABLE_TRANSIT_GAUGE=false` would still enable it)?
  - `BAO_DISABLE_KV_GAUGE` doesn't do this, should it also?
- Should we make the KV gauge also opt-in?

## Rationale

"Feature parity" between KV and transit

For performance reasons, this is opt-in.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/core_metrics.go
+++ b/internal/vault/core_metrics.go
@@ -6,6 +6,7 @@ package vault
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


cipherboy:
@phil9909 Whatever happened to #1663 and the goals of working towards OpenTelemetry?

As a general design pattern, we have two shapes here:

 - KV's counts which pierces the veil of secrets engines and is not scalable to arbitrary plugins.
 - PKI's, which natively reports metrics at the expense of them either being high cardinality or overwriting themselves (as metrics don't aggregate properly).

In particular, it becomes hard to keep code in sync in this approach, if Transit's keysutil were to change its layout. It is outside of the Transit plugin boundary and very easily forgotten.

phil9909:
> Whatever happened to #1663 and the goals of working towards OpenTelemetry?

Still something I want to work on and I also did some testing. But there are only so many hours in a day, unfortunately.
 
 
> In particular, it becomes hard to keep code in sync in this approach, if Transit's keysutil were to change its layout. It is outside of the Transit plugin boundary and very easily forgotten.

I agree this approach isn't perfect. But it does use the public API of the transit engine (via `c.router.Route`) so it should be unaffected by changes to keysutil right?

internal/vault/core_metrics.go
```diff
@@ -397,6 +405,111 @@ func (c *Core) kvSecretGaugeCollector(ctx context.Context) ([]metricsutil.GaugeL
...
+			metricsutil.NamespaceLabel(m.Namespace),
+			{Name: "mount_point", Value: m.MountPoint},
+		}
+
+		listRequest := &logical.Request{
+			Operation: logical.ListOperation,
+			Path:      m.Namespace.Path + m.MountPoint + "keys",
+		}
+
+		resp, err := c.router.Route(ctx, listRequest)
```
cipherboy, line 476: @phil9909 Out of curiosity, why not do what PKI does and self-report metrics? 🤔 It has an optional config parameter of whether to blindly report to the same key or whether to include UUIDs.

I realize we _can_ route this way too, but idk, I'm not sure I like this idea.

phil9909: Because of [`vault.secret.kv.count`](https://openbao.org/docs/internals/telemetry/metrics/secrets/#vault-secret-kv-count), I just applied the same pattern here.

But I'm happy to try the other approach.

phil9909: Alternative approach here: https://github.com/openbao/openbao/commit/ff7697033bb6775b6b23af86a2e9a8bd805da08b

Advantage: Less performance impact, because we need to count the keys only once (on metric enable or mount load) and afterwards keep updating an atomic int (on create and delete operations).

Disadvantage: Keeping track of the delta is now distributed among the whole engine. We could get into a "Whac-A-Mole" situation at least for a few releases to fix places were forgot to update the counter or were we updated the counter for changes that were rolled-back or otherwise failed.

I spot-tested the approach locally, it works with one exception: if you create a key at an exciting location/name (e.g. execute `bao write transit/keys/test type=aes128-gcm96` twice) the key is double-counted. I guess this is because `backend.getPolicy` swallows the return value of `GetPolicyWithLockType` and always returns `true`: https://github.com/openbao/openbao/blob/64583d09374c35a278c8f00e30dfb064cf073e22/internal/builtin/logical/transit/backend.go#L179-L183 but I have to understand that part of the code a little better before jumping to conclusions.

PR #3686, opened 2026-08-04
Improve `TestNamespaceManyLeases` stability

phil9909:
## Description

1. Give the test more time to run
2. Fix potential nil-pointer and data-race when the test fails: `EventuallyWithT` runs the "check function" asynchronously. If the timeout expires, it will fail the test without waiting for the function to return, which means the cleanup might run in parallel to the check function, causing races. (also changed `TestNamespaceSealManyLeases` and `TestNamespaceSealResourcesLifecycle`)

## Rationale

https://github.com/openbao/openbao/actions/runs/30889437309?pr=3677

## Acknowledgements

       also allowed.


satoqz (approved):
Makes sense to me, thanks!

PR #3688, opened 2026-08-04
Adds auth/token/revoke-self to wrapping token policy, plus tests

### LINKED ISSUE #3628: Allow self-revocation of wrapping tokens (including control-groups)
@satoqz Hmm, no, it looks like you can't self-revoke as the response wrapping policy doesn't allow it:

```
path "cubbyhole/response" {
    capabilities = ["create", "read"]
}

path "sys/wrapping/unwrap" {
    capabilities = ["update"]
}
```

:thinking: I wonder if it should? 

That said, the control group accessor is still valid after revoking (with a modified policy). So, does this preclude merge? Or should I work on adding this support to all wrapping tokens in a follow-up PR?

_Originally posted by @cipherboy in https://github.com/openbao/openbao/issues/3601#issuecomment-5106300408_
            

----

See also additional context in that thread.

fyi: @suprjinx

suprjinx:
## Description

This PR adds "auth/token/revoke-self" to the wrapping token policy, along with tests.

## Rationale

Wrapping tokens -- in particular those which result from control-group wrapping -- should be able to delete themselves if needed (eg, if there's private information in the request, which an approver with access token can review -- this will delete it).

Resolves: #3628

## Acknowledgements

       also allowed.


cipherboy:
@suprjinx Besides the missing acknowledgements section, when I tested this locally, it didn't work as the token was already revoked prior to hitting the revocation handler, due to the logic in `request_handler.go` for single-use seal-wrapped tokens.

suprjinx:
> @suprjinx Besides the missing acknowledgements section, when I tested this locally, it didn't work as the token was already revoked prior to hitting the revocation handler, due to the logic in `request_handler.go` for single-use seal-wrapped tokens.

@cipherboy So `revoke-self` does result in the token being revoked, with no response error, but only because it consumes the NumUses counter?

EDIT:
I added an exception to the NumUses decrement for `auth/token/revoke-self`

cipherboy:
> @cipherboy So `revoke-self` does result in the token being revoked, with no response error, but only because it consumes the NumUses counter?

Yes, though its a bit of an asynchronous revocation which means that you're not actually removing associated data like accessor &c, so the control group could still be approved after the fact (it'd just not be unwrappable).

suprjinx:
> > @cipherboy So `revoke-self` does result in the token being revoked, with no response error, but only because it consumes the NumUses counter?
> 
> Yes, though its a bit of an asynchronous revocation which means that you're not actually removing associated data like accessor &c, so the control group could still be approved after the fact (it'd just not be unwrappable).

@cipherboy Let me know if the exception around `UseToken` for `revoke-self` seems like the right fix

internal/vault/request_handling.go
```diff
@@ -1293,9 +1293,13 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
...
 	// need to forward, we exit before now)
 	if te != nil {
-		// Attempt to use the token (decrement NumUses)
+		// Attempt to use the token (decrement NumUses) unless this a revoke-self request,
+		// in which case we want to allow the request to go through and then revoke the
+		// token after the request is processed.
 		var err error
-		te, err = c.tokenStore.UseToken(ctx, te)
+		if req.Path != "auth/token/revoke-self" {
+			te, err = c.tokenStore.UseToken(ctx, te)
```
cipherboy: I will have to think about this more. 🤔  Maybe @satoqz has thoughts?

cipherboy: I almost feel like it needs a defer is what I'm entertained about.

cipherboy (dismissed):

changelog/3688.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+vault: add support for revoke-self with wrapping tokens
```
wslabosz-reply: I think we've refrained from using `vault` as category instead opting for `core`
```suggestion
core: Add support for revoke-self with wrapping tokens.
```

internal/vault/request_handling.go
satoqz: Perhaps, instead of making sure that the actual revoke-self handler does the revocation and optimistically skipping `UseToken`, we could have the block that follows below be lazy only conditionally?

```go
if te.NumUses == tokenRevocationPending {
	defer func(id string) {
		if err == nil {
			if req.Path == "auth/token/revoke-self" {
				// UseToken() above will make the revoke-self handler
				// not see the token and skip revocation silently. To be
				// consistent with revoke-self's behavior, don't lazily
				// revoke here.
				err = c.expiration.Revoke(ctx, leaseID)
			} else {
				err = c.expiration.LazyRevoke(ctx, leaseID)
			}
		}
	}(te.ID)
}
```

Then we don't end up with uncertainty w.r.t. whether the inner handler wasn't reached for some reason and make the defer minimally more complex (opposed to introducing a second conditional `UseToken` within the deferred section).

cipherboy: I like that, thanks @satoqz. 

@suprjinx If you add that (and the changelog update), then I think we're good and can merge.

satoqz (approved):

cipherboy (approved):

PR #3692, opened 2026-08-05
api: add RawSecret to read secret data as erasable []byte

### LINKED ISSUE #3519: api/v2: no way to zero secret values returned in Secret.Data
Secret.Data is a map[string]interface{}, so every value returned to a caller is a Go string: a PKI private_key from `pki/issue/<role>`, KV values, tokens. Strings are immutable, so a consumer can't overwrite them after use. The plaintext stays on the heap until the GC gets to it, and converting to `[]byte` just adds a second copy while the original sticks around. For anything that wants to minimize how long a private key or token sits in memory, there's currently no way to do it through the client.

The field is Secret.Data in `api/secret.go`, typed `map[string]interface{}`: https://github.com/openbao/openbao/blob/f14f7a1e14f189e06b0b23394f31bda732b7f41d/api/secret.go#L30

It would help to have a way to pull selected fields out as a `[]byte` the caller owns (or access to the raw response body), so the value can be zeroed once it's been used.

Go 1.26 adds runtime/secret `(GOEXPERIMENT=runtimesecret)` which scrubs stack and heap for a scoped function, but it only does real work on linux/amd64 and linux/arm64; on macOS and other platforms Do just runs the function, so it's not a portable answer for a client library.

Related to #354, though that one is about the server process (mlock/swap); this is the api/v2 client data model.

Suggested labels: enhancement, security-hardening

head1328:
## Description

`Secret.Data` holds every value as a Go string, which cannot be overwritten, so
a private key or token stays in memory until the GC gets to it. Converting to `[]byte`
afterwards only adds a copy while the original sits on the heap untouched.

This adds the parse path discussed in the issue:

- `RawSecret`, the same shape as `Secret`, but string values in `Data` are `[]byte`
- `ParseRawSecret`, the counterpart to `ParseSecret`
- `(*Logical).ReadBytesWithContext` and `ReadBytesWithDataWithContext`

The response is decoded into `map[string]json.RawMessage` and converted from
there. The conversion recurses through objects and arrays, otherwise kv-v2 secrets
would be missed, since those sit under `data.data` with `metadata` alongside them.
Numbers stay `json.Number` as with `ParseSecret`. Fields outside `Data` stay strings,
they carry no secret material.

String escapes are resolved directly into a byte slice, since `encoding/json` only
exposes unquoting through a `string` and the equivalent `jsontext.AppendUnquote`
is behind `GOEXPERIMENT=jsonv2` (as of Go 1.26.x). The behaviour matches `encoding/json`
byte for byte, which the tests check by comparing against `ParseSecret` on the same input,
including invalid UTF-8.

**`Secret` and `ParseSecret` are untouched, the new path is opt-in.**

The library does not erase anything by itself, only the caller knows when a value is done.
And there are copies it cannot reach at all, in the buffers of `net/http`, `crypto/tls` and 
`encoding/json`. So callers get something they can erase, which was the missing piece,
but the response does not stop existing elsewhere in the stack.

## Rationale

Resolves: #3519

## Acknowledgements

       also allowed.


cipherboy:
@head1328 Mind following the Acknowledgements section of the PR template? Thanks!

head1328:
> @head1328 Mind following the Acknowledgements section of the PR template? Thanks!

Yes, I will, but I'm still working on it; it's still a draft.

head1328:
Implementation is pushed. Two tests are still missing: `ReadBytesWithDataWithContext`, and the 404 branch that returns a response carrying warnings or data. Both are coming.

If a runnable demo helps for review, I can put one up as a separate repo or Gist. I'm working in this area anyway, so it costs me nothing.

head1328:
It turned out to be more complicated than expected. I ran further tests and added additional code comments for clarification. In the future, this could be simplified using jsonv2.

I have also published a simple demo:
https://gist.github.com/head1328/f4238d9e58357f960220cf288cba8fe9

api/raw_secret.go
```diff
@@ -0,0 +1,390 @@
...
+func ParseRawSecret(r io.Reader) (*RawSecret, error) {
+	// We might need to read the body twice:
+	// 1. First attempt: Parse as rawSecretWire.
+	// 2. If that fails: Parse as raw JSON (e.g., for error messages).
+	// We use TeeReader to copy the body into a buffer for the second attempt.
+	var bodyBuffer, teeBuffer bytes.Buffer
+	teeReader := io.TeeReader(r, &teeBuffer)
+	if _, err := bodyBuffer.ReadFrom(teeReader); err != nil {
+		return nil, err
+	}
```
satoqz, line 137: This copies all of our data to the heap, twice, no?

api/raw_secret.go
```diff
@@ -0,0 +1,390 @@
...
+		// Keep numbers as json.Number to avoid losing precisions.
+		return json.Number(rawValue), nil
+	}
+}
+
+// unescapeJSONString removes quotes and resolves escape sequences from a JSON
+// string because encoding/json unfortunately doesn’t do this when it comes to
+// working with raw JSON.
+func unescapeJSONString(raw []byte) []byte {
+	result := make([]byte, 0, len(raw))
```
satoqz, line 292: ...and this creates a copy of all strings?

api/raw_secret.go
```diff
@@ -0,0 +1,390 @@
...
+		return nil, nil
+	}
+
+	// First, try to parse the response as rawSecretWire.
+	var wire rawSecretWire
+	decoder := json.NewDecoder(&bodyBuffer)
+	decoder.UseNumber() // Parse numbers as json.Number, not float64.
+	if err := decoder.Decode(&wire); err != nil {
+		return nil, err
+	}
```
satoqz, line 150: This also copies within the standard library, AFAICT: https://go.dev/play/p/qM1BkrNAA5v

satoqz (commented):
Not really sure what the point of being able to erase one of many copies on the heap is.

api/raw_secret.go
head1328: Yes, you're right – that can be avoided. I'll fix it.

api/raw_secret.go
head1328: Yes, this copy is the return value als `[]byte`. ParseSecret allocates a string in the same location, so the number of copies is the same; this one is erasable by the consumer.

api/raw_secret.go
head1328: You're right, I'll fix that. `json.Unmarshal` works on the slice without copying.

head1328:
> Not really sure what the point of being able to erase one of many copies on the heap is.

The one that matters is the return value the consumer receives. A CLI client cannot erase a string, not even with `runtime/secret.Do`, because strings are immutable. That's why I thought it made sense to have that as an option.

api/raw_secret.go
satoqz: It... does not? The component doing the copying here is `json.RawMessage`, funnily enough:

https://github.com/golang/go/blob/952ad3f69ae73b24a82bc2f617c7276f0f50d5eb/src/encoding/json/stream.go#L283-L290

satoqz:
> The one that matters is the return value the consumer receives. A CLI client cannot erase a string, not even with `runtime/secret.Do`, because strings are immutable. That's why I thought it made sense to have that as an option.

So what's the actual threat model here, if this isn't about avoiding memory analysis?

head1328:
> > The one that matters is the return value the consumer receives. A CLI client cannot erase a string, not even with `runtime/secret.Do`, because strings are immutable. That's why I thought it made sense to have that as an option.
> 
> So what's the actual threat model here, if this isn't about avoiding memory analysis?

The threat model is written about a running OpenBao instance (storage backend, cluster nodes) or not?

`api/` runs in someone else's process, and that process is not an OpenBao instance. So I do not think that applies here. OpenBao cannot know the threat model of the consumers, and at the moment the library decides for them by handing out a type (string) that cannot be erased.

I can provide a more detailed example of the consumer.

satoqz:
> The threat model is written about a running OpenBao instance (storage backend, cluster nodes) or not?
> 
> `api/` runs in someone else's process, and that process is not an OpenBao instance. So I do not think that applies here. OpenBao cannot know the threat model of the consumers, and at the moment the library decides for them by handing out a type (string) that cannot be erased.
> 
> I can provide a more detailed example of the consumer.

Hence I was asking about *your* threat model, and as to why this patch achieves that.

head1328:
> > The threat model is written about a running OpenBao instance (storage backend, cluster nodes) or not?
> > `api/` runs in someone else's process, and that process is not an OpenBao instance. So I do not think that applies here. OpenBao cannot know the threat model of the consumers, and at the moment the library decides for them by handing out a type (string) that cannot be erased.
> > I can provide a more detailed example of the consumer.
> 
> Hence I was asking about _your_ threat model, and as to why this patch achieves that.

My tool is a CLI that has `require github.com/openbao/openbao/api/v2`. The read path runs inside my process (e.g. local device), not on a bao instance.

The command reads a few secrets and then e.g. uses them to provision things. It holds the `*api.Secret`, so with string the plaintext stays in memory until the process exits. And after that it still is in the freed heap memory until something happens to overwrite it, which Go never promises to do.

**The attacker I have in mind is not attacking the running process.** It is whoever gets a memory image **later**: a core dump that ends up in a bug report, a swap page, someone attaching a debugger. With `[]byte` I can overwrite the value after the last use, so a dump taken after that point does not contain it at all.
 
This does nothing about the copies in net/http or crypto/tls, and the read path makes a few of its own. That is a separate problem, and it is why I will remove the avoidable copies you pointed out.

PR #3711, opened 2026-08-06
Drop VOLUME instructions from Dockerfile

satoqz:
## Description

Remove `VOLUME` instructions from the `Dockerfile`, impacting Alpine & UBI distributions.

## Rationale

I've been debating this in the back of my mind for week or so and I've concluded that these instructions present a net-negative to OpenBao's container images and are safe to remove.

The result of this will be no more automatically created anonymous volumes via Docker, Podman & co (Not Kubernetes, which the instruction never had effects on).

The primary benefit is that a test run will not leak hundreds of anonymous volumes anymore[^1] (though we can also fix this just within sdk by adding a tmpfs mount at `/openbao/logs`). The same benefit goes for any one-off invocations of the container image for testing when persistence is not desired (and yet forced onto you in useless ways via the instruction).

I find it extremely unlikely for anyone to rely on these instructions in production:

- They don't have any effect in Kubernetes, so only Docker/Podman-based setups are affected.
- While the volume directories are pre-created, container images do not ship with default configuration pointing at them. Actually writing data to these locations is deliberate configuration, which is a natural opportunity to configure explicit volume mounts.
- When using anonymous volumes, you'd need to keep track of its randomly generated ID between container recreations, e.g., when upgrading to a newer version of OpenBao. This also counts for compose-based setups, a single `docker compose down && docker compose up` loses service -> anonymous volume mappings.

Further reading: https://github.com/moby/moby/issues/43190, which has lots of additional links if you'd like to go down the rabbit hole yourself.

## Acknowledgements

       also allowed.

[^1]: At least those implicitly created by Docker, which should only be the logs volume. It will still leak volumes whenever containers are leaked and the volume explicitly created for `/openbao/file` is not cleaned up.

### DIFF
--- a/changelog/3711.txt
+++ b/changelog/3711.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+packaging/container: Remove all `VOLUME` instructions.
+```
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,14 +41,6 @@ RUN mkdir -p /openbao/logs && \
     mkdir -p /openbao/config && \
     chown -R openbao:openbao /openbao
 
-# Expose the logs directory as a volume since there's potentially long-running
-# state in there
-VOLUME /openbao/logs
-
-# Expose the file directory as a volume since there's potentially long-running
-# state in there
-VOLUME /openbao/file
-
 # 8200/tcp is the primary interface that applications use to interact with
 # OpenBao.
 EXPOSE 8200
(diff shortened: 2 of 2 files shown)


phil9909 (approved):

cipherboy (approved):
I'm all for fixing this at the root rather than trying to improve the test suite to drop anonymous volumes automatically.

changelog/3711.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
cipherboy, line 1: Should this be `change` I wonder?

satoqz: Since I don't assume it to be breaking in a meaningful way I figured the main change is to avoid littering, which is an improvement.

cipherboy: OK, fair.

PR #3712, opened 2026-08-06
docs: Add generic PKCS#11 auto-unseal guide

thgoebel:
## Description

This PR adds **generic** PKCS#11 auto-unseal guide that should be compatible with any vendor.

This guide complements the existing docs that already exist at: [plugin system](https://openbao.org/docs/next/plugins/), [pkcs stanza](https://openbao.org/docs/next/configuration/seal/pkcs11/), [vendor-specific guides](https://openbao.org/docs/next/guides/unseal/pkcs11/securosys/).
This guide is intended to be more descriptive and step-by-step than the other "reference-style" docs.

Most importantly, compared to the existing vendor-specific guides, this new guide is updated with the changes brought by 2.6.0:

- Use the new `kms` plugin type. I.e., use the external plugin (instead of the built-in plugin that will be removed in 2.7.0)
- State that the `openbao-hsm` image is no longer necessary (`openbao-ubi`  is still needed)

## Rationale

Resolves: n/a

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/guides/unseal/pkcs11/generic.mdx
+++ b/website/content/docs/guides/unseal/pkcs11/generic.mdx
@@ -1,0 +1,237 @@
+---
+sidebar_label: Generic
+---
+
+# Auto-Unsealing with PKCS#11
+
+This guide describes how to configure OpenBao auto-unsealing with a PKCS#11-compliant hardware token.
+The instructions are generic and should work with any PKCS#11-compliant token.
+
+This guide shows the easiest way to install the PKCS#11 plugin (plugin type `kms`).
+Please see the [plugin system](../../../plugins/index.mdx) documentation for more details
+and other ways to install plugins.
+
+## Tested PKCS#11 tokens
+
+The community has tested the following PKCS#11 tokens to work with OpenBao:
+
+<!-- Please keep the list sorted alphabetically -->
+
+| Vendor | Models | libc | Guides |
+| --- | --- | --- | --- |
+| [DuoKey](https://duokey.com/) | DuoKey SD-HSM | ? | [Installation Guide](duokey.mdx) |
+| [Nitrokey](https://www.nitrokey.com/) | NetHSM | [glibc or musl](https://github.com/Nitrokey/nethsm-pkcs11/releases) | [Installation Guide](nitrokey.mdx) |
+| [Securosys](https://www.securosys.com/) | Primus HSM, CloudHSM | glibc |[Installation Guide](securosys.mdx) |
+| [Utimaco](https://utimaco.com/) | CryptoServer CP5, u.trust GP HSM | glibc | [Installation Guide](utimaco.mdx) |
+
+## Running in containers
+
+Some PKCS#11 libraries require glibc and do not work on systems based on musl libc, such as Alpine Linux.
+Please see the table above or contact your vendor.
+If your PKCS#11 library requires glibc, you need to use the `openbao-ubi` instead of the Alpine-based `openbao` image.
+
+Historically, OpenBao has provided `openbao-hsm-ubi` and `openbao-hsm` images.
+These included OpenBao compiled with cgo, which is necessary to load PKCS#11 libraries.
+These images are no longer required if you use the `kms-pkcs11` plugin (as described by this guide).
+This external plugin is compiled with cgo, meaning that the main `bao` binary can be compiled with pure Go.
+
+## Step 1: Install the PKCS#11 library
+
+Initialize and prepare your PKCS#11 token.
+For network HSMs, this often involves creating a new user/partition/tenant.
+
+Install and configure your vendor's PKCS#11 module on the same machine as OpenBao.
+If you deploy OpenBao in a container (Docker, Podman, Kubernetes, ...), you need to
+mount the relevant files into the container (`.so` and vendor-specific config files).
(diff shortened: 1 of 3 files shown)


thgoebel:
This PR relates to https://github.com/openbao/openbao/issues/3713 in the sense that:

1. It’s a prerequisite if we want to remove the vendor-specific guides.
2. Even if the vendor specific guides stay, this PR is still needed. There should be a generic guide, to signal that any PKCS#11-compliant token should work.

website/content/docs/guides/unseal/pkcs11/generic.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+plugin "kms" "pkcs11" {
+  image       = "ghcr.io/openbao/openbao-plugin-kms-pkcs11"
+  version     = "v0.1.0"
+  binary_name = "openbao-plugin-kms-pkcs11"
+  sha256sum   = "55245882727535579e710672f0eae1bcdddc846006db857baaa6e09e33d40faf"
+}
+```
+
+See the [OpenBao plugin system](https://openbao.org/docs/next/plugins/)
```
thgoebel: Ack, I missed these links. Will fix on Monday.

website/content/docs/guides/unseal/pkcs11/generic.mdx
```diff
@@ -0,0 +1,227 @@
...
+This guide describes how to configure OpenBao auto-unsealing with a PKCS#11-compliant hardware token.
+The instructions are generic and should work with any PKCS#11-compliant token.
+
+This guide shows the easiest way to install the PKCS#11 plugin.
+Please see the [plugin system](../../../plugins/index.mdx) documentation for more details
+and other ways to install plugins.
+
+:::info
+
+This guide requires OpenBao 2.6.0 (or later) because it uses the
```
cipherboy: This is an interesting discussion. Technically we're on versioned documentation now, and the guides _are_ versioned, so there could be per-version guides (and rather, guides which don't exist at a previous version aren't guaranteed to work there).

OTOH, I have wondered (and maybe @satoqz has as well?) if we want to merge guides to `/community/` instead, so they're explicitly not versioned? But then they're also a lot less discoverable as they're not in the main part of the documentation.

Thoughts?

website/content/docs/guides/unseal/pkcs11/generic.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+:::
+
+## Tested PKCS#11 tokens
+
+The community has tested the following PKCS#11 tokens to work with OpenBao:
+
+{/* Please keep the list sorted alphabetically */}
+
+| Vendor | Models | Guides |
```
cipherboy: Re: #3713, do we want to split `Guides` into a generic PKCS#11 usage guide and a vendor-specific OpenBao guide, if it exists? 🤔

website/content/docs/guides/unseal/pkcs11/generic.mdx
```diff
@@ -0,0 +1,227 @@
...
+| --- | --- | --- |
+| [DuoKey](https://duokey.com/) | DuoKey SD-HSM | [Product page](https://duokey.com/en/products/vault)    |
+| [Nitrokey](https://www.nitrokey.com/) | NetHSM | [PKCS#11 installation guide](https://docs.nitrokey.com/en/nethsm/pkcs11-setup)  |
+| [Securosys](https://www.securosys.com/) | Primus HSM, CloudHSM | [PKCS#11 installation guide](https://docs.securosys.com/pkcs/quickstart) |
+| [Utimaco](https://utimaco.com/) | CryptoServer CP5, u.trust GP HSM |  |
+
+## Running in containers
+
+Most PKCS#11 libraries require glibc and do not work on systems based on musl libc, such as Alpine Linux.
+For container users, this means you must use the `openbao-ubi` instead of the Alpine-based `openbao` image.
```
cipherboy: @satoqz You pulled `gcompat` into `openbao` right?

satoqz: @cipherboy yes, but in our case `gcompat` only ensures that the dynamic linker is happy with the `openbao-plugin-kms-pkcs11` binary since all included C code is perfectly libc-agnostic and requires _just_ libc. With this resolved, the next problem you'll run into is the vendor's library having some dependency on glibc that `gcompat` does not provide, so you'll need to switch to UBI anyway.

But it is probably worth rewriting as _some_ PKCS#11 libraries require glibc, so users won't jump to UBI right away when they don't actually need to. Similarly, it is worth mentioning that installation of additional packages may be required if the vendor dynamically linked against an additional library besides libc.

website/content/docs/guides/unseal/pkcs11/generic.mdx
satoqz: I'd vote to keep the section versioned for now and remove the v2.6 disclaimer accordingly.

thgoebel: I removed this box.

website/content/docs/guides/unseal/pkcs11/generic.mdx
thgoebel: > But it is probably worth rewriting as some PKCS#11 libraries require glibc, so users won't jump to UBI right away when they don't actually need to. Similarly, it is worth mentioning that installation of additional packages may be required if the vendor dynamically linked against an additional library besides libc.

Done. I also added a table column "libc" to provide an overview.

thgoebel:
Ping :)

PR #3717, opened 2026-08-07
UI fix firefox integration tests

phil9909:
## Description

On Firefox one of our UI integration tests fails.

The assertion uses [`getComputedStyle`](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) under the hood. There seem to be a difference how Chromium and Firefox implement that method for [shorthand properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Shorthand_properties). Firefox seems to omit default values, while Chromium includes them. I don't know which one is more correct or if it is defined at all by the spec, but in general it seems like a good idea to avoid using shorthand properties with the `getComputedStyle` API as the can have some corner-cases.

This PR also adds a `BROWSER` env to the `Makefile` (defaulting to `chrome` to keep existing behavior). You can use `BROWSER=firefox make test` to run the integration tests in Firefox and `BROWSER=chromium make test` to run with Chromium (which is helpful, if you don't have the Google-flavor of the browser installed).

## Rationale

Firefox is a supported browser: https://openbao.org/docs/browser-support/

So we should be able to run the integration tests for it.

## Acknowledgements

       also allowed.

### DIFF
--- a/ui/Makefile
+++ b/ui/Makefile
@@ -1,3 +1,5 @@
+BROWSER ?= chrome
+
 .PHONY: frontend
 frontend:
 	pnpm start
@@ -8,7 +10,7 @@ backend:
 
 .PHONY: test
 test:
-	pnpm test
+	pnpm test --launch "$(BROWSER)"
 
 .PHONY: check
 check: test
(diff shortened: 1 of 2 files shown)


satoqz (approved):

phil9909:
Merging as the UI test failures seem unrelated and the same tests failed on main the last time the suite ran there: https://github.com/openbao/openbao/actions/runs/30404046848/job/90425319566

But we should look into this.

Issue #3718, opened 2026-08-07
RFC: Dedicated OpenBao OpenTofu/Terraform Provider


Luiggi33:
### Summary

With the compatibility of OpenBao comes to option of using the Hashicorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This shows in issues like #339, https://github.com/hashicorp/terraform-provider-vault/issues/2869 and discussions like #637.

This RFC aims to define a path to creating a dedicated provider for OpenBao.

### Problem Statement

With the divergence in features, fixes and settings of OpenBao from Vault, the Hashicorp maintained Vault provider is starting to show this incompatibility as well. The profile/workflows system has no resources or datasources, the `auth_login_aws` isn't working (https://github.com/openbao/openbao/issues/339#issuecomment-4112645562) and fixes (https://github.com/hashicorp/terraform-provider-vault/pull/2870) for similar issues (https://github.com/hashicorp/terraform-provider-vault/issues/2869) don't get merged.
This forces the user to adopt workarounds that lead to unwanted complexity and instability.

### User-facing description

OpenBao will provide a dedicated OpenTofu/Terraform provider as a soft-fork of the Vault provider that implements OpenBao specific features and fixes, as well as upstream fixes that have not been merged yet.

### Technical Description

Partially taken from https://github.com/openbao/openbao/issues/339#issuecomment-4881822035, the provider should be a low-divergence soft fork from the Vault provider. The focus would be on providing a compatible provider that has a simple migration path from the Vault provider, while providing the missing features and fixes.
The comparability of upstream changes will be verified by acceptance tests against the current OpenBao release.
Where applicable fixes for non-OpenBao specific issues should be send to the upstream provider.

### Rationale and alternatives

The main rationale is to create a viable alternative to the Vault provider for configuring a OpenBao instance via OpenTofu/Terraform. Especially with keeping a simple migration path, this would improve the experience of people switching from the Vault provider to the dedicated provider.

Alternatives would be to stay with the status quo of having no dedicated TF provider

### Downsides

- Additional component that needs maintenance

### Unresolved Questions

Should we stay with the idea of a low-divergence soft fork?

### Related Issues

#339 & #637

Luiggi33:
I will rather throw this into a pull request as that seems to be a bit easier to suggest changes to!

PR #3720, opened 2026-08-07
RFC: Dedicated OpenBao OpenTofu/Terraform Provider

### LINKED ISSUE #339: Dedicated openbao tf provider
For configuring Vault, hashicorp provides a [terraform provider](https://github.com/hashicorp/terraform-provider-vault) to manage various components from terraform or opentofu. As best I can tell openbao does not maintain a fork of this, and so as openbao's api diverges there will be no replicate this functionality. There exists an [opentofu manged fork](https://github.com/opentofu/terraform-provider-vault) but this is specifically aimed at vault, and is currently just mirroring hashicorp upstream

Fork the tf provider to ensure openbao is managable? or some advice regarding api compatibility

The tf provider will function for now while the apis are matched, but divergence will likely occur and features will break, which will mean re-configuring providers and re-importing existing configuration

**Additional context**
The background to this is that I'm writing an infrastructure stack and would like to be able to auto-configure vault for certain uses. As best I can tell there is no existing discussion about this for guidance either.

Luiggi33:
## Description

A draft for a RFC to create a dedicated OpenBao OpenTofu/Terraform provider

## Rationale

Continues my inital draft from #3718 and resolves #339

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/community/rfcs/dedicated-tf-provider.mdx
+++ b/website/content/community/rfcs/dedicated-tf-provider.mdx
@@ -8,7 +8,7 @@ description: |-
 
 ### Summary
 
-With the compatibility of OpenBao comes the option of using the Hashicorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This is shown in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
+With the compatibility of OpenBao comes the option of using the HashiCorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This is shown in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
 
 This RFC aims to define a path to creating a dedicated provider for OpenBao.
 
--- a/website/content/community/rfcs/dedicated-tf-provider.mdx
+++ b/website/content/community/rfcs/dedicated-tf-provider.mdx
@@ -27,6 +27,21 @@ Partially taken from [this comment by dc-tec](https://github.com/openbao/openbao
 The compatibility of upstream changes will be verified by acceptance tests against the current OpenBao release.
 Where applicable fixes for non-OpenBao specific issues should be send to the upstream provider.
 
+The soft fork will be maintained as follows:
+
+1. Clone the upstream `terraform-provider-vault` repository into the `openbao/` GitHub organization.
+2. After each upstream release, review and cherry-pick relevant patches into the fork. This process will start out manual and could be supported by automatic testing.
+
+#### Migration Path
+
+As long as the fork stays low-divergence and keeps existing resource and data source names, migrating from the Vault provider requires no changes to individual resources in state:
+
+1. Update `required_providers` to point at the new provider source and version constraint.
+2. Run `terraform init -upgrade`.
+3. Run `terraform state replace-provider registry.terraform.io/hashicorp/vault registry.terraform.io/openbao/vault` (illustrative address, exact registry namespace TBD) to rewrite the provider reference recorded in state, without touching the resources themselves.
+
+This simple migration path only holds while the fork stays low-divergence. If a feature or fix eventually forces the fork to diverge in resource names or schemas, a migration path from `hashicorp/vault` to `openbao/vault` v1 (without the divergence and the simple migration) to `openbao/vault` v2 (with the divergence and more in depth migration) will need to be defined at that point, likely using `moved` blocks or `terraform state mv` for any resource addresses that change.
+
 ### Rationale and alternatives
 
 The main rationale is to create a viable alternative to the Vault provider for configuring a OpenBao instance via OpenTofu/Terraform. Especially with keeping a simple migration path, this would improve the experience of people switching from the Vault provider to the dedicated provider.
--- a/website/content/community/rfcs/dedicated-tf-provider.mdx
+++ b/website/content/community/rfcs/dedicated-tf-provider.mdx
@@ -29,7 +29,7 @@ Where applicable fixes for non-OpenBao specific issues should be send to the ups
 
 The soft fork will be maintained as follows:
 
-1. Clone the upstream `terraform-provider-vault` repository into the `openbao/` GitHub organization.
+1. Fork the upstream `terraform-provider-vault` repository into the `openbao/` GitHub organization.
 2. After each upstream release, review and cherry-pick relevant patches into the fork. This process will start out manual and could be supported by automatic testing.
 
 #### Migration Path
--- a/website/content/community/rfcs/dedicated-tf-provider.mdx
+++ b/website/content/community/rfcs/dedicated-tf-provider.mdx
@@ -41,10 +41,10 @@ The provider will be released to the OpenTofu registry as well as the Terraform 
 As long as the fork stays low-divergence and keeps existing resource and data source names, migrating from the Vault provider requires no changes to individual resources in state:
 
 1. Update `required_providers` to point at the new provider source and version constraint.
-2. Run `terraform init -upgrade`.
-3. Run `terraform state replace-provider registry.terraform.io/hashicorp/vault registry.terraform.io/openbao/vault` (illustrative address, exact registry namespace TBD) to rewrite the provider reference recorded in state, without touching the resources themselves.
+2. Run `tofu init -upgrade`.
+3. Run `tofu state replace-provider registry.terraform.io/hashicorp/vault registry.opentofu.org/openbao/vault` (illustrative address, exact registry namespace TBD) to rewrite the provider reference recorded in state, without touching the resources themselves.
 
-This simple migration path only holds while the fork stays low-divergence. If a feature or fix eventually forces the fork to diverge in resource names or schemas, a migration path from `hashicorp/vault` to `openbao/vault` v1 (without the divergence and a simple migration path) to `openbao/vault` v2 (with the divergence and a possible more complex migration path) will need to be defined at that point, likely using `moved` blocks or `terraform state mv` for any resource addresses that change.
+This simple migration path only holds while the fork stays low-divergence. If a feature or fix eventually forces the fork to diverge in resource names or schemas, a migration path from `hashicorp/vault` to `openbao/vault` v1 (without the divergence and a simple migration path) to `openbao/vault` v2 (with the divergence and a possible more complex migration path) will need to be defined at that point, likely using `moved` blocks or `tofu state mv` for any resource addresses that change.
 
 ### Rationale and alternatives
 
--- a/website/content/community/rfcs/dedicated-tf-provider.mdx
+++ b/website/content/community/rfcs/dedicated-tf-provider.mdx
@@ -1,0 +1,46 @@
+---
+sidebar_label: Dedicated OpenBao OpenTofu/Terraform Provider
+description: |-
+  An OpenBao RFC for the creation of a dedicated OpenBao OpenTofu/Terraform provider.
+---
+
+# RFC: Dedicated OpenBao OpenTofu/Terraform Provider
+
+### Summary
+
+With the compatibility of OpenBao comes to option of using the Hashicorp Vault OpenTofu/Terraform provider. While this prvider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This shows in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
+
+This RFC aims to define a path to creating a dedicated provider for OpenBao.
+
+### Problem Statement
+
+With the divergence in features, fixes and settings of OpenBao from Vault, the Hashicorp maintained Vault provider is starting to show this incompatibility as well. The profile/workflows system has no resources or datasources, the `auth_login_aws` isn't working ([#339 (comment)](https://github.com/openbao/openbao/issues/339#issuecomment-4112645562)) and fixes ([hashicorp/terraform-provider-vault#2870](https://github.com/hashicorp/terraform-provider-vault/pull/2870)) for similar issues ([hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869)) don't get merged.
+This forces the user to adopt workarounds that lead to unwanted complexity and instability.
+
+### User-facing description
+
+OpenBao will provide a dedicated OpenTofu/Terraform provider as a soft-fork of the Vault provider that implements OpenBao specific features and fixes, as well as upstream fixes that have not been merged yet.
+
+### Technical Description
+
+Partially taken from [this comment by dc-tec](https://github.com/openbao/openbao/issues/339#issuecomment-4881822035), the provider should be a low-divergence soft fork from the Vault provider. The focus would be on providing a compatible provider that has a simple migration path from the Vault provider, while providing the missing features and fixes.
+The comparability of upstream changes will be verified by acceptance tests against the current OpenBao release.
+Where applicable fixes for non-OpenBao specific issues should be send to the upstream provider.
+
+### Rationale and alternatives
+
+The main rationale is to create a viable alternative to the Vault provider for configuring a OpenBao instance via OpenTofu/Terraform. Especially with keeping a simple migration path, this would improve the experience of people switching from the Vault provider to the dedicated provider.
+
+Alternatives would be to stay with the status quo of having no dedicated TF provider
+
+### Downsides
+
+- Additional component that needs maintenance
+
+### Unresolved Questions
+
+Should we stay with the idea of a low-divergence soft fork?
+
+### Related Issues
+
+[#339](https://github.com/openbao/openbao/issues/339) & [#637](https://github.com/orgs/openbao/discussions/637)  
--- a/website/content/community/rfcs/dedicated-tf-provider.mdx
+++ b/website/content/community/rfcs/dedicated-tf-provider.mdx
@@ -8,7 +8,7 @@ description: |-
 
 ### Summary
 
-With the compatibility of OpenBao comes to option of using the Hashicorp Vault OpenTofu/Terraform provider. While this prvider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This shows in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
+With the compatibility of OpenBao comes the option of using the Hashicorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This is shown in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
 
 This RFC aims to define a path to creating a dedicated provider for OpenBao.
 
@@ -24,7 +24,7 @@ OpenBao will provide a dedicated OpenTofu/Terraform provider as a soft-fork of t
 ### Technical Description
 
 Partially taken from [this comment by dc-tec](https://github.com/openbao/openbao/issues/339#issuecomment-4881822035), the provider should be a low-divergence soft fork from the Vault provider. The focus would be on providing a compatible provider that has a simple migration path from the Vault provider, while providing the missing features and fixes.
-The comparability of upstream changes will be verified by acceptance tests against the current OpenBao release.
+The compatibility of upstream changes will be verified by acceptance tests against the current OpenBao release.
 Where applicable fixes for non-OpenBao specific issues should be send to the upstream provider.
 
 ### Rationale and alternatives
--- a/website/content/community/rfcs/dedicated-tf-provider.mdx
+++ b/website/content/community/rfcs/dedicated-tf-provider.mdx
@@ -4,22 +4,22 @@ description: |-
   An OpenBao RFC for the creation of a dedicated OpenBao OpenTofu/Terraform provider.
 ---
 
-# RFC: Dedicated OpenBao OpenTofu/Terraform Provider
+# Dedicated OpenBao OpenTofu/Terraform Provider
 
 ### Summary
 
-With the compatibility of OpenBao comes the option of using the HashiCorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This is shown in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
+With the compatibility of OpenBao comes the option of using the HashiCorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure an OpenBao instance. This is shown in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
 
 This RFC aims to define a path to creating a dedicated provider for OpenBao.
 
 ### Problem Statement
 
-With the divergence in features, fixes and settings of OpenBao from Vault, the Hashicorp maintained Vault provider is starting to show this incompatibility as well. The profile/workflows system has no resources or datasources, the `auth_login_aws` isn't working ([#339 (comment)](https://github.com/openbao/openbao/issues/339#issuecomment-4112645562)) and fixes ([hashicorp/terraform-provider-vault#2870](https://github.com/hashicorp/terraform-provider-vault/pull/2870)) for similar issues ([hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869)) don't get merged.
+With the divergence in features, fixes and settings of OpenBao from Vault, the HashiCorp-maintained Vault provider is starting to show this incompatibility as well. The profile/workflows system has no resources or datasources, the `auth_login_aws` isn't working ([#339 (comment)](https://github.com/openbao/openbao/issues/339#issuecomment-4112645562)) and fixes ([hashicorp/terraform-provider-vault#2870](https://github.com/hashicorp/terraform-provider-vault/pull/2870)) for similar issues ([hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869)) don't get merged.
 This forces the user to adopt workarounds that lead to unwanted complexity and instability.
 
 ### User-facing description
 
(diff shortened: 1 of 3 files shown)


website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,46 @@
...
+sidebar_label: Dedicated OpenBao OpenTofu/Terraform Provider
+description: |-
+  An OpenBao RFC for the creation of a dedicated OpenBao OpenTofu/Terraform provider.
+---
+
+# RFC: Dedicated OpenBao OpenTofu/Terraform Provider
+
+### Summary
+
+With the compatibility of OpenBao comes to option of using the Hashicorp Vault OpenTofu/Terraform provider. While this prvider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This shows in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
```
Nerkho:
```suggestion
With the compatibility of OpenBao comes the option of using the Hashicorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This is shows in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
```

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,46 @@
...
+This forces the user to adopt workarounds that lead to unwanted complexity and instability.
+
+### User-facing description
+
+OpenBao will provide a dedicated OpenTofu/Terraform provider as a soft-fork of the Vault provider that implements OpenBao specific features and fixes, as well as upstream fixes that have not been merged yet.
+
+### Technical Description
+
+Partially taken from [this comment by dc-tec](https://github.com/openbao/openbao/issues/339#issuecomment-4881822035), the provider should be a low-divergence soft fork from the Vault provider. The focus would be on providing a compatible provider that has a simple migration path from the Vault provider, while providing the missing features and fixes.
+The comparability of upstream changes will be verified by acceptance tests against the current OpenBao release.
```
Nerkho:
```suggestion
The compatibility of upstream changes will be verified by acceptance tests against the current OpenBao release.
```

Nerkho (commented):
Thanks @Luiggi33 ! I like the approach of a soft fork to get started.

In time, I hope we can get to a point where we only have `bao_` resources.

One thing we might want to do from the start is to remove any resources that pertain to Vault enterprise features from the provider.

cipherboy:
> One thing we might want to do from the start is to remove any resources that pertain to Vault enterprise features from the provider.

This will be difficult but appreciated; we'll have to cross-reference extensively. We've added a number of them (control groups in this release spring to mind).

dc-tec:
Thanks for making a start on this RFC @Luiggi33. Much appreciated!

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,46 @@
...
+sidebar_label: Dedicated OpenBao OpenTofu/Terraform Provider
+description: |-
+  An OpenBao RFC for the creation of a dedicated OpenBao OpenTofu/Terraform provider.
+---
+
+# RFC: Dedicated OpenBao OpenTofu/Terraform Provider
+
+### Summary
+
+With the compatibility of OpenBao comes the option of using the Hashicorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This is shown in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
```
cipherboy: nit: 

```suggestion
With the compatibility of OpenBao comes the option of using the HashiCorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This is shown in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
```

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+### Technical Description
+
+Partially taken from [this comment by dc-tec](https://github.com/openbao/openbao/issues/339#issuecomment-4881822035), the provider should be a low-divergence soft fork from the Vault provider. The focus would be on providing a compatible provider that has a simple migration path from the Vault provider, while providing the missing features and fixes.
+The compatibility of upstream changes will be verified by acceptance tests against the current OpenBao release.
+Where applicable fixes for non-OpenBao specific issues should be send to the upstream provider.
+
+### Rationale and alternatives
+
+The main rationale is to create a viable alternative to the Vault provider for configuring a OpenBao instance via OpenTofu/Terraform. Especially with keeping a simple migration path, this would improve the experience of people switching from the Vault provider to the dedicated provider.
```
cipherboy, line 51: What is the actual migration path? Edit your modules list and just start using the new one? Does something else need to happen?

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+Alternatives would be to stay with the status quo of having no dedicated TF provider
+
+### Downsides
+
+- Additional component that needs maintenance
+
+### Unresolved Questions
+
+Should we stay with the idea of a low-divergence soft fork?
```
cipherboy: So I think defining soft fork should probably be in here:

1. Clone repo, push up to `openbao/` space.
2. Merge patches, cherry-pick from upstream after release. Probably need some procedure (to start, later automate) around patch review/picking.

cipherboy (commented):
Some thoughts, but I think probably fine enough to merge once addressed!

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+### User-facing description
+
+OpenBao will provide a dedicated OpenTofu/Terraform provider as a soft-fork of the Vault provider that implements OpenBao specific features and fixes, as well as upstream fixes that have not been merged yet.
+
+### Technical Description
+
+Partially taken from [this comment by dc-tec](https://github.com/openbao/openbao/issues/339#issuecomment-4881822035), the provider should be a low-divergence soft fork from the Vault provider. The focus would be on providing a compatible provider that has a simple migration path from the Vault provider, while providing the missing features and fixes.
+The compatibility of upstream changes will be verified by acceptance tests against the current OpenBao release.
+Where applicable fixes for non-OpenBao specific issues should be send to the upstream provider.
```
satoqz, line 30: Do we care? As far as I understand, they've not shown a great track record of merging fixes developed by the community, which is part of why we've arrived at this stage.

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,61 @@
+---
+sidebar_label: Dedicated OpenBao OpenTofu/Terraform Provider
+description: |-
+  An OpenBao RFC for the creation of a dedicated OpenBao OpenTofu/Terraform provider.
+---
+
+# RFC: Dedicated OpenBao OpenTofu/Terraform Provider
```
satoqz:
```suggestion
# Dedicated OpenBao OpenTofu/Terraform Provider
```

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+Alternatives would be to stay with the status quo of having no dedicated TF provider
+
+### Downsides
+
+- Additional component that needs maintenance
+
+### Unresolved Questions
+
+Should we stay with the idea of a low-divergence soft fork?
```
satoqz: yes

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+This RFC aims to define a path to creating a dedicated provider for OpenBao.
+
+### Problem Statement
+
+With the divergence in features, fixes and settings of OpenBao from Vault, the Hashicorp maintained Vault provider is starting to show this incompatibility as well. The profile/workflows system has no resources or datasources, the `auth_login_aws` isn't working ([#339 (comment)](https://github.com/openbao/openbao/issues/339#issuecomment-4112645562)) and fixes ([hashicorp/terraform-provider-vault#2870](https://github.com/hashicorp/terraform-provider-vault/pull/2870)) for similar issues ([hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869)) don't get merged.
```
satoqz:
```suggestion
With the divergence in features, fixes and settings of OpenBao from Vault, the HashiCorp-maintained Vault provider is starting to show this incompatibility as well. The profile/workflows system has no resources or datasources, the `auth_login_aws` isn't working ([#339 (comment)](https://github.com/openbao/openbao/issues/339#issuecomment-4112645562)) and fixes ([hashicorp/terraform-provider-vault#2870](https://github.com/hashicorp/terraform-provider-vault/pull/2870)) for similar issues ([hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869)) don't get merged.
```

website/content/community/rfcs/dedicated-tf-provider.mdx
cipherboy: This to me seems like one of those things which time will influence. OpenBao proper started as a low-divergence fork of Vault, but over time has built its own feature set while retaining API compatibility for applications. I think TFVP might go the same way, unless we put in place an explicit Vault CE testing plan.

website/content/community/rfcs/dedicated-tf-provider.mdx
cipherboy: Its nice to position this way at least. I'd rather be seen as the helpful of the two groups. "Come to the dark side, we have cookies"-type vibes. 

Its good marketing as well.

Luiggi33: Imo this would never be the top priority of this fork, but is still "in the spirit of open source". When we eventually diverge too much, this will be out of scope anyways

website/content/community/rfcs/dedicated-tf-provider.mdx
Luiggi33: As mentioned in the rfc, I wanna go the way of keeping v1 compatible with the Vault Provider (in effect with Vault aswell) and with a v2 allow us to move away from complete compatability. I personally wouldnt have the interest in keeping it up to date with Vault for too long

satoqz: Perhaps the better open-ended question is if we should stay with the soft-fork model *indefinitely*. But no need to nitpick that any further, makes sense.

jficz: > I wanna go the way of keeping v1 compatible with the Vault Provider

Care must be taken to not make this a long-term promise and explicitly state as such. Also perhaps the "compatibility" should be defined as best-effort or opportunistic and also limited to just features compatible with OpenBao (i.e. the provider won't support any CE or EE features which are not also part of OpenBao).

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,61 @@
...
+sidebar_label: Dedicated OpenBao OpenTofu/Terraform Provider
+description: |-
+  An OpenBao RFC for the creation of a dedicated OpenBao OpenTofu/Terraform provider.
+---
+
+# RFC: Dedicated OpenBao OpenTofu/Terraform Provider
+
+### Summary
+
+With the compatibility of OpenBao comes the option of using the HashiCorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure a OpenBao instance. This is shown in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
```
karras:
```suggestion
With the compatibility of OpenBao comes the option of using the HashiCorp Vault OpenTofu/Terraform provider. While this provider still works for most tasks, there has been increased friction with using it to configure an OpenBao instance. This is shown in issues like [#339](https://github.com/openbao/openbao/issues/339), [hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869) and discussions like [#637](https://github.com/orgs/openbao/discussions/637).
```

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,61 @@
...
+This RFC aims to define a path to creating a dedicated provider for OpenBao.
+
+### Problem Statement
+
+With the divergence in features, fixes and settings of OpenBao from Vault, the Hashicorp maintained Vault provider is starting to show this incompatibility as well. The profile/workflows system has no resources or datasources, the `auth_login_aws` isn't working ([#339 (comment)](https://github.com/openbao/openbao/issues/339#issuecomment-4112645562)) and fixes ([hashicorp/terraform-provider-vault#2870](https://github.com/hashicorp/terraform-provider-vault/pull/2870)) for similar issues ([hashicorp/terraform-provider-vault#2869](https://github.com/hashicorp/terraform-provider-vault/issues/2869)) don't get merged.
+This forces the user to adopt workarounds that lead to unwanted complexity and instability.
+
+### User-facing description
+
+OpenBao will provide a dedicated OpenTofu/Terraform provider as a soft-fork of the Vault provider that implements OpenBao specific features and fixes, as well as upstream fixes that have not been merged yet.
```
karras:
```suggestion
OpenBao will provide a dedicated OpenTofu/Terraform provider as a soft-fork of the Vault provider that implements OpenBao-specific features and fixes, as well as upstream fixes that have not been merged yet.
```

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+### Technical Description
+
+Partially taken from [this comment by dc-tec](https://github.com/openbao/openbao/issues/339#issuecomment-4881822035), the provider should be a low-divergence soft fork from the Vault provider. The focus would be on providing a compatible provider that has a simple migration path from the Vault provider, while providing the missing features and fixes.
+The compatibility of upstream changes will be verified by acceptance tests against the current OpenBao release.
+Where applicable fixes for non-OpenBao specific issues should be send to the upstream provider.
+
+The soft fork will be maintained as follows:
+
+1. Fork the upstream `terraform-provider-vault` repository into the `openbao/` GitHub organization.
```
karras, line 34: Should we call it `opentofu-provider-openbao` instead?

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,61 @@
...
+2. After each upstream release, review and cherry-pick relevant patches into the fork. This process will start out manual and could be supported by automatic testing.
+
+#### Migration Path
+
+As long as the fork stays low-divergence and keeps existing resource and data source names, migrating from the Vault provider requires no changes to individual resources in state:
+
+1. Update `required_providers` to point at the new provider source and version constraint.
+2. Run `terraform init -upgrade`.
+3. Run `terraform state replace-provider registry.terraform.io/hashicorp/vault registry.terraform.io/openbao/vault` (illustrative address, exact registry namespace TBD) to rewrite the provider reference recorded in state, without touching the resources themselves.
```
karras: I gather we'd also push to https://search.opentofu.org/?

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+This simple migration path only holds while the fork stays low-divergence. If a feature or fix eventually forces the fork to diverge in resource names or schemas, a migration path from `hashicorp/vault` to `openbao/vault` v1 (without the divergence and the simple migration) to `openbao/vault` v2 (with the divergence and more in depth migration) will need to be defined at that point, likely using `moved` blocks or `terraform state mv` for any resource addresses that change.
+
+### Rationale and alternatives
+
+The main rationale is to create a viable alternative to the Vault provider for configuring a OpenBao instance via OpenTofu/Terraform. Especially with keeping a simple migration path, this would improve the experience of people switching from the Vault provider to the dedicated provider.
+
+Alternatives would be to stay with the status quo of having no dedicated TF provider
```
karras:
```suggestion
Alternatives would be to stay with the status quo of having no dedicated TF provider.
```

website/content/community/rfcs/index.mdx
```diff
@@ -36,6 +36,8 @@ Steering Committee.
  - [Index Headers for Consistency](index-headers.mdx), allowing clients to
    ensure strict consistency in horizontal scalability's eventual consistency
    environment even when talking to a standby node.
+ - [Dedicated OpenBao OpenTofu/Terraform Provider](dedicated-tf-provider.mdx), for creating a
+   dedicated OpenTofu/Terraform provider for OpenBao.
```
karras:
```suggestion
   native OpenTofu/Terraform provider.
```

karras: Simplifying it a bit by avoiding two repetitions.

karras:
Looks great, many thanks @Luiggi33!

website/content/community/rfcs/dedicated-tf-provider.mdx
Luiggi33: I think we should embrace opentofu, this is just based of the vault provider (which is terraform only, with it getting mirrored into opentofu's register)

cipherboy: @karras OpenBao is redundant, so I was planning on just calling it `opentofu-provider`.

website/content/community/rfcs/dedicated-tf-provider.mdx
Luiggi33: added that into the RFC aswell

website/content/community/rfcs/dedicated-tf-provider.mdx
pree: Based on [OpenTofu Registry Docs](https://search.opentofu.org/docs/providers/adding) the naming should be `terraform-provider-YOURPROVIDER`

pree: Also seeing in their issue template: https://github.com/opentofu/registry/blob/acdb71fed3b7332c7e2229ae6f7395e789123522/.github/ISSUE_TEMPLATE/provider.yml?plain=1#L13

website/content/community/rfcs/dedicated-tf-provider.mdx
eyenx: possibly most of the resource names need to change, as most if not all of them are prefixed with `vault_` except of course we want to keep that naming convention.

website/content/community/rfcs/dedicated-tf-provider.mdx
satoqz: ... so probably `terraform-provider-openbao`.

website/content/community/rfcs/dedicated-tf-provider.mdx
eyenx: Additionally if someone would want to use both providers in the same codebase (not sure if there is a use case for that, but ther might be in future?) they wouldn't be able to do this if both provider prefix the resources with the same prefix (a.e `vault_`)

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,63 @@
...
+2. After each upstream release, review and cherry-pick relevant patches into the fork. This process will start out manual and could be supported by automatic testing.
+
+The provider will be released to the OpenTofu registry as well as the Terraform registry.
+
+#### Migration Path
+
+As long as the fork stays low-divergence and keeps existing resource and data source names, migrating from the Vault provider requires no changes to individual resources in state:
+
+1. Update `required_providers` to point at the new provider source and version constraint.
+2. Run `terraform init -upgrade`.
```
eyenx: can't we set `tofu` as default instead of `terraform` also for the registry link further down?

Luiggi33: switched that as well as the registry to tofu

website/content/community/rfcs/dedicated-tf-provider.mdx
Luiggi33: I made the question a bit more specific and added a sentence to inform about the "long term plans" (as in, this provider is a way to have a simple migration path from Vault)

website/content/community/rfcs/dedicated-tf-provider.mdx
Luiggi33: > Additionally if someone would want to use both providers in the same codebase (not sure if there is a use case for that, but ther might be in future?) they wouldn't be able to do this if both provider prefix the resources with the same prefix (a.e `vault_`)

I dont think there will be a case for both providers? If someone uses Vault they should use the Vault provider, when they use OpenBao they should use the OpenBao provider. (This could also encourage some migration to OpenBao ;D)
"Taking over" the `vault_` prefix is intended to ease a migration without mass renaming/moving, which would be pushed back until we actually diverge significantly from the Vault provider and work out a migration guide (or automation)

Nerkho: Would need to test that properly. I'm not sure but potentially, for an existing codebase with an existing tfstate, one might need to migrate/re-import the resources as it's not the same provider anymore.

Nerkho: Ah this might do the trick https://opentofu.org/docs/cli/commands/state/replace-provider/

Luiggi33: I found that too. It is part of a potential migration path in the RFC

Nerkho: Ah right missed that. sorry

website/content/community/rfcs/dedicated-tf-provider.mdx
cipherboy: @satoqz @pree Oh, that's maddening that its tied to repo name and not some value in a Go or YAML file somewhere. :shrug:

website/content/community/rfcs/dedicated-tf-provider.mdx
```diff
@@ -0,0 +1,65 @@
...
+The soft fork will be maintained as follows:
+
+1. Fork the upstream `terraform-provider-vault` repository into the `openbao/` GitHub organization.
+2. After each upstream release, review and cherry-pick relevant patches into the fork. This process will start out manual and could be supported by automatic testing.
+
+The provider will be released to the OpenTofu registry as well as the Terraform registry.
+
+#### Migration Path
+
+As long as the fork stays low-divergence and keeps existing resource and data source names, migrating from the Vault provider requires no changes to individual resources in state:
```
pree, line 41: I'm not sure whether this works out, as the [Terraform best practices](https://developer.hashicorp.com/terraform/plugin/best-practices/naming#resource-names) mention: `Resource names must always start with their containing provider's name followed by an underscore`

cipherboy: @pree Does that mean we need to name our provider `_vault`?

satoqz: TBF, if we're going to have `vault_*` resources, it might as well be `terraform-provider-vault`. And vice-versa.

jficz: > it might as well be...

IIRC it actually _has to_ be. Last time I looked at a provider source, I recall this being hardcoded and used functionally, not just cosmetically.

pree: I don't know whether we can offer a migration path as we did for the other things (ENV variables, annotations, etc) to support both `openbao_` / `bao_` & `vault_` or whether we need to start with a strict cutoff

pree: I also think that having different resource prefixes might make it easier for teams to migrate from one solution to the other without having to specify the `provider = ` field on every resource.

Luiggi33: > I don't know whether we can offer a migration path as we did for the other things (ENV variables, annotations, etc) to support both `openbao_` / `bao_` & `vault_` or whether we need to start with a strict cutoff

From what I read, you could try to have a alias, but that potentially breaks things. This isn't part of the proposed solution anyways.
The plan I laied out in this RFC is to provide a drop in replacement that supports OpenBao and staying with this minimal diversion until that gets too much to maintain, is not deemed necessary anymore or we are forced to diverge for whatever reason. When our fork leaves this low-diversion state, we can switch to using the `bao_` and document a path on how to rename the resources etc.

Luiggi33: > I also think that having different resource prefixes might make it easier for teams to migrate from one solution to the other without having to specify the `provider = ` field on every resource.

I am unsure how that would make it easier? If you are using OpenBao, you can use the OpenBao provider, there should be no need to have the Vault provider or am I misunderstanding what you meant

jficz: Resouce (and data) names are hardcoded as `<provider name>_<resource/data name>` in the providers so there really are only two options:

1. call it `vault` and keep it as an actual drop-in replacement (that _will_ work for OpenBao because that's what Bao users are using now)

2. call it something else and require a one-off migration of the Tofu codebase - that will require renaming all (re)sources from `vault_*` to `openbao_*` _and_ a `moved` block for each of the resource at minimum.

Anything else (like using a different local name for the provider and then setting `provider=...` in each resource) would imho be much  more complex.

From adoption perspective, as a user of the provider, I would prefer option one (drop-in replacement). Whether to rename the provider _later_ to something else, I don't really know - once that happens, code and state migration will be necessary and that's never fun.

That would favor option two however. _If_ we _ever_ plan to rename the provider, better do it from start. There is no significant difference for existing users whether to refactor now or later but there will be no refactor for new users at all.

That means, in my opinion, we have a decision to make: do we _ever_ want to rename the provider? If _yes_, then do it now, right from the start. If not, don't change our mind later.

cipherboy: As another orthogonal concern, porting patches between one and the other is made more difficult by a rename, but the branding is more consistent with it.

Luiggi33: > From adoption perspective, as a user of the provider, I would prefer option one (drop-in replacement). Whether to rename the provider _later_ to something else, I don't really know - once that happens, code and state migration will be necessary and that's never fun.
> 
> That would favor option two however. _If_ we _ever_ plan to rename the provider, better do it from start. There is no significant difference for existing users whether to refactor now or later but there will be no refactor for new users at all.
> 
> That means, in my opinion, we have a decision to make: do we _ever_ want to rename the provider? If _yes_, then do it now, right from the start. If not, don't change our mind later.

If we rename it now, we can leave the whole soft fork approach behind and just hard fork. The specific reason to keep the vault prefix and the soft fork approach as a whole is to allow a simple migration *right now* without any additional work on the users side.

While the migration to a OpenBao branded provider has to happen one day or another, I would much rather focus on having a working drop in replacement that just works, then something I have to get in the weeds before I see any rewards (this is me speaking from the perspective of actually using the vault provider and seeing its growing issues with OpenBao).

satoqz (approved):

wslabosz-reply (approved):

Luiggi33:
With the approvals given, when could this get merged and the fork be created?

satoqz:
I'd like to see some additional approvals from other organizations (@wslabosz-reply and I represent the same organization). In particular I recall @cipherboy mentioning he'd like a final look by @pree / @eyenx / @Nerkho.

pree (approved):
I'm okay with starting it as `terraform-provider-vault` :)
Let's use the new naming when we decide to hard-fork (or rewrite)

cipherboy:
So merged! Thanks @Luiggi33 and all for the discussion!

PR #3721, opened 2026-08-07
Better changelog checker, adjustments

cipherboy:
## Acknowledgements

       also allowed.


cipherboy:
@satoqz I think we should long-term consider forking https://github.com/hashicorp/go-changelog so it will work as we want, but unless you want today to be that day, I'd suggest that I've mostly already wrote the changelog checker this way.

I do agree, we probably want to re-write the `go-changelog` tool to use a proper Markdown AST package, perhaps https://pkg.go.dev/github.com/gomarkdown/markdown/ast, so that we can ensure entries are formatted correctly in the new representation. :shrug:

changelog/319.txt
```diff
@@ -1,3 +1,3 @@
 ```release-note:feature
-auth/oidc: Add a new `callback_mode` role option value `device` to use the oidc device flow instead of a callback, add a new `poll_interval` role option to control how often to poll for a response, and add a new `callbackmode=device` option to the oidc login method in the cli.
+**Device Clalback Mode in OIDC Auth**: Add a new `callback_mode` role option value `device` to use the oidc device flow instead of a callback, add a new `poll_interval` role option to control how often to poll for a response, and add a new `callbackmode=device` option to the oidc login method in the cli.
```
wslabosz-reply:
```suggestion
**Device Callback Mode in OIDC Auth**: Add a new `callback_mode` role option value `device` to use the oidc device flow instead of a callback, add a new `poll_interval` role option to control how often to poll for a response, and add a new `callbackmode=device` option to the oidc login method in the cli.
```

wslabosz-reply (dismissed):
do we also want to update the release notes content in the same way?

scripts/changelog.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash
```
satoqz:
```suggestion
#!/usr/bin/env bash
```

scripts/changelog-tester.sh
satoqz, line 1: Do we need this committed?

scripts/changelog.sh
```diff
@@ -0,0 +1,195 @@
...
+
+  if ! grep -q '^```release-note:' <<< "$preamble" ; then
+    echo "Expected annotation 'release-note:<type>' on changelog"
+    return 1
+  fi
+
+  category="$(sed 's/^```release-note://g' <<< "$preamble")"
+  found=false
+  for allowed in ${allowed_types[@]}; do
+    if [ "$category" == "$allowed" ]; then
```
satoqz: Can you use `[[` since we're in bash?

scripts/changelog.sh
```diff
@@ -0,0 +1,195 @@
...
+    # Ensure they're indented at least 2 spaces.
+    index=0
+    while IFS= read -r line; do
+      index=$(( index + 1 ))
+      if (( index == 1 )); then
+        component="$(grep -o '^[^:]*:' <<< "$content")"
+
+        if [ "$category" != "feature" ] && [ "$category" != "deprecation" ]; then
+      # Component has to match [a-z/]*.
+          if ! grep -q '^[a-z][a-z0-9/-]*\(, [a-z][a-z0-9/-]*\)*:' <<< "$component"; then
```
satoqz:
```suggestion
        if [ "$category" != "feature" ] && [ "$category" != "deprecation" ]; then
          # Component has to match [a-z/]*.
          if ! grep -q '^[a-z][a-z0-9/-]*\(, [a-z][a-z0-9/-]*\)*:' <<< "$component"; then
```

scripts/changelog-tester.sh
cipherboy: Nah, I can drop.

cipherboy:
> do we also want to update the release notes content in the same way?

@wslabosz-reply I got lazy at this point. From this branch, running something like:

```
LAST_RELEASE=fork-point THIS_RELEASE=v2.0.0 make release-changelog
```

doesn't work. Well, it does, just a little too well; its the changelog as it was, not as it should be. So at that point, we're stuck manually copying all the changelog bits into the release notes &c.

wslabosz-reply:
@cipherboy 
I'm fine to leave them as it is (maybe at most just adjust the most recent release).

cipherboy:
@wslabosz-reply I'll think about adjusting in a follow-up PR then. It'd probably be easier if we fork the changelog tool as a one-off to fix tbh.

satoqz (approved):

cipherboy:
Thanks!

satoqz:
From my POV much of what is in CHANGELOG.md also has manual adjustments that we'd lose if we were to re-generate. I'd just not open that can of worms TBH.

PR #3724, opened 2026-08-07
fix: make the capabilities-self query in the token's namespace

### LINKED ISSUE #3723: Capabilities-self is checked in the browsed namespace, not the token's namespace, causing 403s
**Describe the bug**

When using the web UI via a read-only account, you can't browse to a KV secret in a child namespace if the token was issued in a different namespace (i.e. root), despite the root namespace policy having the appropriate permissions. When browsing in the web UI under these conditions, you'll receive a 403 error like this:

```text
Your auth token does not have access to sys/capabilities-self.
Make sure the policy path sys/capabilities-self includes capabilities = ['update']
```

The token's default policy already grants update on `sys/capabilities-self in root`, but the problem is that the UI sends the capabilities-self request into the browsed namespace with the header `X-Vault-Namespace: my_namespace` with unprefixed paths in the payload, so it resolves to my_namespace/sys/capabilities-self. Since this is a POST it needs `update`access, which generally is confusing since this is available in the default policy.

If you use the CLI to access to the same secret with the same token, it works fine, only the UI breaks here. For some additional context, this policy worked in the UI on Vault Enterprise 1.21.1, so I think it makes sense to udpate the code to handle this case instead of just explicitly granting `update` in each ns policy, which does work as a workaround.

**To Reproduce**

1. Start a bao server with a namespace and a secret:

   ```bash
   export BAO_ADDR=http://127.0.0.1:8200
   export BAO_TOKEN=root

   bao namespace create my_namespace/
   bao secrets enable -ns=my_namespace -path=kv kv-v2
   bao kv put -ns=my_namespace kv/test hello=world

1. create/apply an RO policy for the NS:

   create a file with this policy (i'll call it readonly-my_namespace.hcl)

   ```hcl
   path "my_namespace/kv/data/*"     { capabilities = ["read", "list"] }
   path "my_namespace/kv/metadata/*" { capabilities = ["read", "list"] }
   path "my_namespace/sys/internal/ui/mounts" { capabilities = ["read"] }
   path "sys/namespaces"     { capabilities = ["list"] }
   ```

   ```bash
   bao policy write readonly-my_namespace readonly-my_namespace.hcl
   ```

1. make a token that uses this policy:

   ```bash
   bao token create -policy=readonly-my_namespace -field=token
   ```

1. Log into the UI with the token from the prior step, switch to the my_namespace namespace,
and open the kv secret test. You'll see the 403 mentioned in the description when browsing the secret in the UI. You can also confirm this with cURL and see what a potential fix could do differently in the UI:

   ```bash
   # old behavior asks inside the namespace and gets a 403
   curl -si -X POST "$BAO_ADDR/v1/sys/capabilities-self" -H "X-Vault-Token: $TOKEN_GOES_HERE" -H "X-Vault-Namespace: my_namespace" -d '{"paths":["kv/data/test"]}'

   # my proposed changed behavior is to ask at root (in this example at least) with the qualified path, and yay, things work! 
   curl -s -X POST "$BAO_ADDR/v1/sys/capabilities-self" -H "X-Vault-Token: $TOKEN_GOES_HERE" -d '{"paths":["my_namespace/kv/data/test"]}'

**Expected behavior**
Browsing a KV store with read-only credentials will allow you to read the credentials.

**Environment:**

* OpenBao Server Version (retrieve with `bao status`): 2.5.3 (also confirmed on 2.6.1 locally)
* OpenBao CLI Version (retrieve with `bao version`): 2.5.2 (not impacted by this issue)
* Server Operating System/Architecture: local dev on M4 MacOS with docker (orbstack) and running native binaries on Ubuntu 24.04.4 on Intel/x86 arch.

**Additional context**

TL;DR, this happens because the UI makes some POSTs to `capabilities-self` that don't work if your account is truly "read-only". In other deployments, this isn't an issue since the default policy has support for this `update` requirement, however, when a token is issued in a different namespace than the one you're using it in (very common for us at least to have one auth mount point shared for most namespaces) then you get a 403 when the UI tries to see what all you're allowed to do.

I've got a PR i'm opening shortly that I think fixes this by making the POST in the namespace that the token was issued in and then simply add the full qualified path to the NS that the data you're trying to access lives in.

crankynetmaam:
## Description

This changes the UI code for the capabilities adaptor so that it makes POSTs to `sys/capabilities-self` in the namespace the token was issued in. This also means that it prepends the namespace you're browsing to the path when making the request. This fixes the issue described in #3723, which was resulting in 403s for read-only users browsing namespaces (due to how default policy works). 

## Rationale

IMO, the default policy, which grants this capability, should be used appropriately when we make API calls from the UI. A 403 is a really odd one to run into here since it works just fine via the CLI (using a `read` or `list` of the exact kv paths) and it implies that you don't have access to the secrets themselves. It is also very confusing because default policy *does* have that permission, so finding out where to go to add this is a bit confusing. Additionally, we stumbled across this at my dayjob during a migration from vault to openbao, and the behavior is unique to openbao, so I suspect that it was changed on that side at some point after the fork. 

This seems pretty low impact and will resolve this for us at least, and means we won't have to make the workaround of re-granting permissions that are already in `default` in the namespace policies.

Resolves: #3723 

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3724.txt
+++ b/changelog/3724.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: query capabilities-self in the token's own namespace so cross-namespace browsing does not 403
+```
--- a/changelog/3724.txt
+++ b/changelog/3724.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-ui: query capabilities-self in the token's own namespace so cross-namespace browsing does not 403
+ui: Query capabilities-self in the token's own namespace preventing 403 responses for cross-namespace browsing.
 ```
--- a/ui/app/adapters/capabilities.js
+++ b/ui/app/adapters/capabilities.js
@@ -12,8 +12,26 @@ export default ApplicationAdapter.extend({
     return 'capabilities-self';
   },
 
+  // the namespace being browsed, relative to the one the token was issued in
+  relativeNamespace() {
+    const { userRootNamespace, path } = this.namespaceService;
+    if (!userRootNamespace) {
+      return path;
+    }
+    if (path === userRootNamespace) {
+      return '';
+    }
+    return path.startsWith(`${userRootNamespace}/`) ? path.slice(userRootNamespace.length + 1) : path;
+  },
+
   findRecord(store, type, id) {
-    return this.ajax(this.buildURL(type), 'POST', { data: { paths: [id] } }).catch((e) => {
+    const relativeNamespace = this.relativeNamespace();
+    const prefix = relativeNamespace ? `${relativeNamespace}/` : '';
+
+    return this.ajax(this.buildURL(type), 'POST', {
+      namespace: this.namespaceService.userRootNamespace,
+      data: { paths: [`${prefix}${id}`] },
+    }).catch((e) => {
       if (e instanceof AdapterError) {
         set(e, 'policyPath', 'sys/capabilities-self');
       }
--- a/ui/tests/unit/adapters/capabilities-test.js
+++ b/ui/tests/unit/adapters/capabilities-test.js
@@ -3,7 +3,8 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import { resolve } from 'rsvp';
+import AdapterError from '@ember-data/adapter/error';
+import { reject, resolve } from 'rsvp';
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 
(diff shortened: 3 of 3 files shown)


crankynetmaam:
Well, in an attempt to update my branch with the latest in main (and add a changelog file) I seem to have done something really spectacular. I think I cleaned up my merge commit, and assuming we squash here then it's not an issue, but happy to pull out my changes and throw it into a fresh branch if need be (to avoid having to force-push to my fork).

Let me know if there is anything else I can help answer / fix on this.

wslabosz-reply:
@crankynetmaam 
I've fixed the git history for you

changelog/3724.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+ui: query capabilities-self in the token's own namespace so cross-namespace browsing does not 403
```
wslabosz-reply:
```suggestion
ui: Query capabilities-self in the token's own namespace preventing 403 responses for cross-namespace browsing.
```

PR #3732, opened 2026-08-08
#3731: Added documentation for Yubico YubiHSM auto-unseal configuration

matkocsis:
## Description

Added a new documentation guide entry to showcase an Auto-Unseal configuration backed by a YubiHSM via PKCS#11 interface.

## Rationale

Adding a new PKCS#11 Auto-Unseal alternative from Yubico to the existing HSM list.
As per request from [@karras](https://github.com/karras)

Resolves: #3731 

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3735, opened 2026-08-09
docs: Remove rsa_oaep_hash from samples

thgoebel:
## Description

People tend to copy&paste samples, and then end up with sha1 as a legacy value. See e.g. https://github.com/openbao/openbao/pull/3732#discussion_r3744906276.
Use sha256 as a better, secure sample value.

I also updated the blog post and RFC to remove all occurrences of sha1. Please let me know if I should not touch these, then I will revert those edits.

## Rationale

Resolves: n/a

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/blog/2026-07-30-declarative-plugins.md
+++ b/website/content/blog/2026-07-30-declarative-plugins.md
@@ -201,7 +201,6 @@ seal "pkcs11" {
     token_label = "OpenBao"
     pin = "4321"
     key_label = "bao-root-key-rsa"
-    rsa_oaep_hash = "sha1"
 }
 ```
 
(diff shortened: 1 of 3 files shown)


satoqz:
sha256 is also the default so IMO just omit it in all of those examples.

thgoebel:
Fine by me, I will remove `rsa_oaep_hash`.

Does the `pkcs11` plugin auto-detect whether it's an AES or RSA key? If not, I would add `mechanism = CKM_RSA_PKCS_OAEP` to the samples.

satoqz:
> Does the `pkcs11` plugin auto-detect whether it's an AES or RSA key? If not, I would add `mechanism = CKM_RSA_PKCS_OAEP` to the samples.

Yes it does. Setting `rsa_oaep_hash` also won't force it to assert an RSA key, only `mechanism` will.

cipherboy (approved):
Thanks @thgoebel! @satoqz Was this good to go now? I don't see any mechanisms, though will let you merge.

satoqz (approved):

PR #3738, opened 2026-08-10
Add reloading KMS plugins at runtime

satoqz:
## Description

Swap out KMS plugin clients on SIGHUP as needed, recreating/killing only those plugins that saw configuration updates or were removed. This is surprisingly easy to do since I've put so much upfront investment into self-healing plugin-derived objects, we can pretty much just kill old clients and have the rest of the system automatically pick up the new configs lazily.

Notably, there is no API-driven upgrade ceremony like other plugin types require, KMS plugins remain fully operator-driven.

## Acknowledgements

       also allowed.


wslabosz-reply (approved):
Looks good

cipherboy (approved):
Thanks @satoqz!

PR #3726, opened 2026-08-07
Move upgrade to guides section

cipherboy:
## Description

Part of @satoqz's wishlist for the website :-)

## Rationale

Related: #3531

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3743, opened 2026-08-10
docs: Add kms plugin upgrade guide

thgoebel:
## Description

Add kms plugin upgrade guide. Previously, there was no guide for kms plugins.

As pointed out in https://github.com/openbao/openbao-plugins/pull/121#discussion_r3730728021:

> perhaps we should add a mention of the KMS plugin upgrade process, which is the simplest of them all (replace binary, update config, restart server, done).

This PR will conflict with https://github.com/openbao/openbao/pull/3726. I'll rebase once that is merged.

## Rationale

Resolves: n/a

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/docs/guides/upgrade/plugins.mdx
+++ b/website/content/docs/guides/upgrade/plugins.mdx
@@ -131,6 +131,26 @@ Until the last step, the mount will still run the first version of `my-db-plugin
 the reload is triggered, OpenBao will kill `my-db-plugin`’s process and start the
 new plugin process for `my-db-plugin` version 1.0.1.
 
+## Upgrading KMS plugins
+
+1. Edit the [`plugin` stanza](../../configuration/plugins.mdx) of your `kms` plugin
+   and change the `version` and `sha256sum` fields to the new values.
+
+1. Download the new plugin binary.
+
+   - For plugins distributed as raw binaries: Place the new binary into the `plugin_directory`.
+   - For plugins distributed as OCI artifacts with `plugin_auto_download = true`:
+     OpenBao will automatically download it at the config reload step.
+   - For plugins distributed as OCI artifacts with `plugin_auto_download = false`:
+     Run [`plugin init`](../../commands/plugin/init.mdx) to trigger the download.
+
+      ```shell-session
+      $ bao plugin init -config=/etc/openbao/openbao.hcl
+      ```
+
+1. Restart OpenBao or send a SIGHUP to reload the config.
+   OpenBao should now use the new `kms` plugin version.
+
 ## Downgrading plugins
 
 Plugin downgrades follow the same procedure as upgrades. You can use the OpenBao


website/content/docs/upgrading/plugins.mdx
```diff
@@ -131,6 +131,22 @@ Until the last step, the mount will still run the first version of `my-db-plugin
...
+
+1. Edit the [`plugin` stanza](../configuration/plugins.mdx) of your `kms` plugin
+   and change the `version` and `sha256sum` fields to the new values.
+1. For plugins distributed as raw binaries: Place the new binary into the `plugin_directory`.
+1. For plugins distributed as OCI artifacts with `plugin_auto_download = false`:
+   Run [`plugin init`](../commands/plugin/init.mdx) to trigger the download.
+
+   ```shell-session
+   $ bao plugin init -config=/etc/openbao/openbao.hcl
+   ```
```
wslabosz-reply: as these are three different upgrade paths (depending on the setup) unnumbered list seem more appropriate?

website/content/docs/upgrading/plugins.mdx
```diff
@@ -131,6 +131,22 @@ Until the last step, the mount will still run the first version of `my-db-plugin
...
+
+1. Restart OpenBao.
+   - If `plugin_auto_download = true`, OpenBao will download the new OCI artifact.
+   - OpenBao should now use the new `kms` plugin version.
```
wslabosz-reply: formatting here seems off imho, maybe let's add a "you should see a log of following <log> format appear" as a confirmation of successful upgrade (although kms plugins aren't versioned, so that's also something we need to highlight)

satoqz:
@thgoebel also see https://github.com/openbao/openbao/pull/3738, you'll be able to send `SIGHUP` instead of restarting. (Via the same event that triggers a potential OCI download, too, so just one signal needed.)

thgoebel:
> you'll be able to send SIGHUP instead of restarting

Good to know. But this only applies to using `kms` plugins at runtime, right (e.g. external keys)? If I want to make sure that auto-unseal still works after the upgrade, I need a restart?

satoqz:
The periodic Auto Seal health check would pick it up, too, unless you've disabled that.

website/content/docs/upgrading/plugins.mdx
thgoebel: > although kms plugins aren't versioned, so that's also something we need to highlight

Do you have a sentence in mind? If we just say "KMS plugins aren't versioned" two lines after telling the user "change the version field", that's confusing. I would omit it for now, and instead reduce it to "restart OpenBao and check stuff still works".

website/content/docs/upgrading/plugins.mdx
```diff
@@ -131,6 +131,26 @@ Until the last step, the mount will still run the first version of `my-db-plugin
 the reload is triggered, OpenBao will kill `my-db-plugin`’s process and start the
 new plugin process for `my-db-plugin` version 1.0.1.
 
+## Upgrading kms plugins
```
satoqz:
```suggestion
## Upgrading KMS plugins
```

Just looks better in a title IMO.

website/content/docs/upgrading/plugins.mdx
satoqz: KMS plugins are absolutely versioned, but they only allow for a single version at a time.

wslabosz-reply (dismissed):
looks good

website/content/docs/guides/upgrade/plugins.mdx
```diff
@@ -131,6 +131,26 @@ Until the last step, the mount will still run the first version of `my-db-plugin
...
+## Upgrading KMS plugins
+
+1. Edit the [`plugin` stanza](../configuration/plugins.mdx) of your `kms` plugin
+   and change the `version` and `sha256sum` fields to the new values.
+
+1. Download the new plugin binary.
+
+   - For plugins distributed as raw binaries: Place the new binary into the `plugin_directory`.
+   - For plugins distributed as OCI artifacts with `plugin_auto_download = true`:
+     Do nothing, OpenBao will automatically download it at the config reload.
```
satoqz:
```suggestion
   - For plugins distributed as OCI artifacts with `plugin_auto_download = true`:
     Do nothing, OpenBao will automatically download it at the config reload step.
```

?

cipherboy: I find it a little entertaining that there's a "do nothing" followed by step 2 "send SIGHUP". 

Maybe:


```suggestion
     OpenBao will automatically download it at the config reload during step 2. 
```

cipherboy: Oh, sorry, missed that @satoqz also suggested something here. I do think `Do nothing` is now redundant or slightly misleading, imho.

satoqz:
@thgoebel can you look at the broken link failure? I think this is legit now.

thgoebel:
Thanks, fixed. The rebased on https://github.com/openbao/openbao/pull/3726 was without conflicts, but the relative links needed to be adjusted.

satoqz (approved):
Thanks!

PR #3744, opened 2026-08-10
Remove build info lookup from helper/versions

satoqz:
## Description

OpenBao's builtin plugin version string is effectively constant, this drops legacy dynamic builtin versions that only made sense with HashiCorp's repo layout.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/helper/versions/version.go
+++ b/internal/helper/versions/version.go
@@ -5,10 +5,8 @@ package versions
 
 import (
 	"fmt"
-	"runtime/debug"
 	"slices"
 	"strings"
-	"sync"
 
 	semver "github.com/hashicorp/go-version"
 	"github.com/openbao/openbao/sdk/v2/helper/consts"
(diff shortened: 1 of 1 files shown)


internal/helper/versions/version.go
```diff
@@ -1,76 +0,0 @@
...
-}
-
-// IsBuiltinVersion checks for the "builtin" metadata identifier in a plugin's
-// semantic version. Vault rejects any plugin registration requests with this
-// identifier, so we can be certain it's a builtin plugin if it's present.
-func IsBuiltinVersion(v string) bool {
-	semanticVersion, err := semver.NewSemver(v)
-	if err != nil {
-		return false
-	}
```
satoqz, line 72: Hmm, guess it'd be better to keep checking just the suffix.

satoqz:
Let's start light instead, it's probably not worth updating all the call sites.

wslabosz-reply (approved):

cipherboy (approved):

cipherboy:
Thanks @satoqz!

PR #3750, opened 2026-08-10
allow bao status to set output field

### LINKED ISSUE #3749: Support `-field` flag with `bao status`
When using the bao cli's status command, there isn't a "native" way to only find a field. I noticed this while working on a automatic backup script.

Support the `-field` flag with `bao status`

Using things like `sed` or other linux command line tools, but it feels a little clunky

Luiggi33:
## Description

I noticed that the `bao status` command didn't support the `-field` output. This merge request aims to add that functionality (including tests)

## Rationale

Resolves: #3749 

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3750.txt
+++ b/changelog/3750.txt
@@ -2,5 +2,5 @@
 command/status: Add support for `-field` argument to `bao status`.
 ```
 ```release-note:improvement
-command/namespace/seal-status: Add support for `-field` argument to `bao namespace seal-status`.
+command/namespace/seal-status: Add support for `-field` and `-format` argument to `bao namespace seal-status`.
 ```
--- a/changelog/3750.txt
+++ b/changelog/3750.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+command/status: Add support for the `-field` argument
+```
--- a/changelog/3750.txt
+++ b/changelog/3750.txt
@@ -1,3 +1,6 @@
 ```release-note:improvement
 command/status: Add support for `-field` argument to `bao status`.
 ```
+```release-note:improvement
+command/namespace/seal-status: Add support for `-field` argument to `bao namespace seal-status`.
+```
--- a/changelog/3750.txt
+++ b/changelog/3750.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-command/status: Add support for the `-field` argument
+command/status: Add support for `-field` argument to `bao status`.
 ```
--- a/internal/command/namespace_seal_status.go
+++ b/internal/command/namespace_seal_status.go
@@ -47,7 +47,7 @@ Usage: bao namespace seal-status [options] PATH
 }
 
 func (c *NamespaceSealStatusCommand) Flags() *FlagSets {
-	return c.flagSet(FlagSetHTTP | FlagSetOutputField)
+	return c.flagSet(FlagSetHTTP | FlagSetOutputField | FlagSetOutputFormat)
 }
 
 func (c *NamespaceSealStatusCommand) AutocompleteArgs() complete.Predictor {
--- a/internal/command/namespace_seal_status.go
+++ b/internal/command/namespace_seal_status.go
@@ -47,7 +47,7 @@ Usage: bao namespace seal-status [options] PATH
 }
 
 func (c *NamespaceSealStatusCommand) Flags() *FlagSets {
-	return c.flagSet(FlagSetHTTP)
+	return c.flagSet(FlagSetHTTP | FlagSetOutputField)
 }
 
 func (c *NamespaceSealStatusCommand) AutocompleteArgs() complete.Predictor {
@@ -89,7 +89,7 @@ func (c *NamespaceSealStatusCommand) Run(args []string) int {
 		return 2
 	}
 
-	return OutputData(c.UI, api.SealStatusResponse{
+	sealStatusResponse := api.SealStatusResponse{
 		Type:        status.Type,
 		Initialized: status.Initialized,
 		Sealed:      status.Sealed,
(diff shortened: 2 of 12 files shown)


Luiggi33:
Had to force push to sign-off the commits, somehow the original commit slipped through without it

internal/command/util.go
```diff
@@ -91,12 +91,68 @@ func RawField(secret *api.Secret, field string) any {
 	return val
 }
 
+func RawSealStatusField(status *SealStatusOutput, field string) any {
```
satoqz, line 95: IMO, like other `-field` flags, this should not take humanized field names but those that you'd see in e.g., `bao status -format=json`. You could achieve that without a switch statement of doom if you just cast the `SealStatusOutput` back to a `map[string]any` (using the `structtomap` helper) and then use `PrintRawField` like usual.

changelog/3750.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+command/status: Add support for the `-field` argument
```
wslabosz-reply:
```suggestion
command/status: Add support for the `-field` flag on `bao status`. 
```
~we've tried standardizing the format of changelog a bit lately, for CLI command changes we've always used just `command` as a category.~

EDIT: seems like more granularity is preferred

satoqz:
```suggestion
command/status: Add support for `-field` argument to `bao status`.
```

Let's reuse the format of the previous changelog entry (`bao token lookup`) just so it's consistent.

satoqz: wrong (see https://github.com/openbao/openbao/blob/main/CHANGELOG.md)

internal/command/util.go
```diff
@@ -91,12 +91,68 @@ func RawField(secret *api.Secret, field string) any {
...
+}
+
 // PrintRawField prints raw field from the secret.
 func PrintRawField(ui cli.Ui, data any, field string) int {
 	var val any
 	switch data := data.(type) {
 	case *api.Secret:
-		val = RawField(data, field)
+		val = RawSecretField(data, field)
+	case *SealStatusOutput:
```
wslabosz-reply, line 121: I think the `not present in secret` error message here is a bit misleading, I wonder if we should check the empty value in a switch case to return different err msg for clarity, thoughts?

internal/command/status.go
```diff
@@ -86,6 +86,19 @@ func (c *StatusCommand) Run(args []string) int {
...
+		if err != nil {
+			c.UI.Error(err.Error())
+			return 1
+		}
+		PrintRawField(c.UI, sealStatusOutput, c.flagField)
+		if sealStatusOutput.Sealed {
+			return 2
+		}
+		return 0
+	}
```
satoqz, line 100: Can you also add this to `namespace_seal_status.go` and then duplicate the changelog for both commands:

````markdown
```release-note:improvement
command/status: Add support for `-field` argument to `bao status`.
```
```release-note:improvement
command/namespace/seal-status: Add support for `-field` argument to `bao namespace seal-status`.
```
````

File: internal/command/util.go (hunk 1 above)
Comment (Luiggi33) [reply internal/command/util.go:95]: Okay I implemented it a little bit differently, to reflect all the possible fields from the seal status output

File: internal/command/util.go (hunk 3 above)
Comment (Luiggi33) [reply internal/command/util.go:121]: I agree that the error message does not necessarily reflect why it actually happens. Should I implement this here or open another merge request?

Comment (wslabosz-reply) [reply internal/command/util.go:121]: @Luiggi33 
I think the scope of the change is not so robust, so within this PR is fine.

Comment (Luiggi33) [reply internal/command/util.go:121]: @wslabosz-reply I tried to make the error message a bit clearer, by splitting it into two. One that tells you if  the data structure passed into the func is supported and one that informs you when the field doesn't exist

File: internal/command/status.go (hunk 4 above)
Comment (Luiggi33) [reply internal/command/status.go:100]: added to the command and changelog

File: internal/command/namespace_seal_status.go [hunk 5]
```diff
@@ -47,7 +47,7 @@ Usage: bao namespace seal-status [options] PATH
 }
 
 func (c *NamespaceSealStatusCommand) Flags() *FlagSets {
-	return c.flagSet(FlagSetHTTP)
+	return c.flagSet(FlagSetHTTP | FlagSetOutputField)
```
satoqz: Is this missing `FlagSetOutputFormat`?

internal/command/util.go
```diff
@@ -91,18 +92,47 @@ func RawField(secret *api.Secret, field string) any {
...
+	case "unseal_progress":
+		val = fmt.Sprintf("%d/%d", status.Progress, status.T)
+	case "ha_mode":
+		mode := "standby"
+		if status.IsSelf {
+			mode = "active"
+		}
+		val = mode
+	case "active_time":
+		val = status.ActiveTime.Format(time.RFC3339Nano)
```
satoqz, line 105: Hmm, don't you think these computed fields would be rather hard to discover though? IMO if you'd like to combine multiple values, just pipe to `jq` using `-format=json` and do it there.

Luiggi33: I am not against removing these "custom" fields. The difficulty of combining them yourself depends on the usecase, but I'd image if you needed them, it would be a problem to recreate em "clientside"

internal/command/namespace_seal_status.go
Luiggi33: That is another flag, I will look into if this is already supported through the command format but just not added here

internal/command/util.go
Luiggi33: Ill leave the active time in there for the correct formatting and the HA mode for having a simple check and remove the unseal progress as that can be easily "computed"

internal/command/namespace_seal_status.go
Luiggi33: Added that in, the normal output functions support this flag out of the box

satoqz (approved):

satoqz:
Thanks! Merging, changelog is fine, the checker is broken because GitHub reads the workflow yaml from main.

PR #3752, opened 2026-08-10
Fix compatibility with Go 1.27rc2

cipherboy:
## Acknowledgements

       also allowed.

### DIFF
--- a/internal/builtin/database/mysql/connection_producer_test.go
+++ b/internal/builtin/database/mysql/connection_producer_test.go
@@ -58,7 +58,7 @@ ssl-key=/etc/mysql/certs/server-key.pem`
 
 	// //////////////////////////////////////////////////////
 	// Start MySQL container
-	retURL := startMySQLWithTLS(t, "8.0", confDir)
+	retURL := startMySQLWithTLS(t, "9.0", confDir)
 
 	// //////////////////////////////////////////////////////
 	// Set up x509 user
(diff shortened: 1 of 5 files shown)


internal/builtin/database/mysql/connection_producer_test.go
```diff
@@ -58,7 +58,7 @@ ssl-key=/etc/mysql/certs/server-key.pem`
 
 	// //////////////////////////////////////////////////////
 	// Start MySQL container
-	retURL := startMySQLWithTLS(t, "8.0", confDir)
+	retURL := startMySQLWithTLS(t, "9.0", confDir)
```
cipherboy, line 61: Ciphersuites in the container need to be updated. I couldnt' go later as the mirror doesn't have anything newer.

satoqz:
> In Go 1.27, we now see:
> idna: invalid label "example.com."
> due to the trailing dot. This does not have a changelog in the current Go 1.27 release notes.

Do we want to wait for updates to https://github.com/golang/go/issues/80476 before committing this as a breaking change on our end?

cipherboy:
@satoqz said:
> @cipherboy said:
> 
> > In Go 1.27, we now see:
> > idna: invalid label "example.com."
> > due to the trailing dot. This does not have a changelog in the current Go 1.27 release notes.
> 
> Do we want to wait for updates to [golang/go#80476](https://github.com/golang/go/issues/80476) before committing this as a breaking change on our end?

I didn't see any update on that Go issue, so I poked on https://github.com/golang/go/issues/78760#issuecomment-5253113965 to see if we can get some sense of if this will be reverted or covered under the scope of the existing one.

I've subscribed to both, so my suggestion would be to keep the changelog and remove if we don't update to Go 1.27 and/or if its fixed as a regression.

cipherboy:
@satoqz Interestingly, 1.26.6 didn't break this AFAICT. :shrug:

satoqz (approved):
🥲

PR #3759, opened 2026-08-11
Make providing a plugin sha256sum optional

satoqz:
## Description

This makes specifying the `sha256sum` of a plugin binary optional when registering it via a declarative `plugin` stanza and not using OCI-based distribution. API-based registrations keep enforcing that a checksum is provided in all cases.

## Rationale

~~TODO: Write the rant, then get told to RFC it anyway.~~

Rant is at https://github.com/openbao/openbao/pull/3881.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3759.txt
+++ b/changelog/3759.txt
@@ -1,0 +1,3 @@
+```release-note:change
+core/plugins: Declarative plugin registration (`plugin` stanza) no longer requires setting a `sha256sum` for manually installed plugin binaries.
+```
--- a/internal/command/server/config.go
+++ b/internal/command/server/config.go
@@ -380,7 +380,9 @@ func (p *PluginConfig) Validate(sourceFilePath string) []configutil.ConfigError 
 		})
 	}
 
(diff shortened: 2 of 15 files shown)


cipherboy:
> TODO: Write the rant, then get told to RFC it anyway.

:laughing: 

To be fair I think you can bundle a bunch of these together in a single RFC. ;-) 

But I do largely concur with the motions here, but it might be nice to have something to point to as far as alternatives and full breadth of UX improvements we want to make.

satoqz:
This would be the initial change for https://github.com/openbao/openbao/pull/3881, with OCI changes stacking on top in a later patch.

wslabosz-reply (dismissed):
LGTM

cipherboy (dismissed):
Happy that this doesn't affect the API cases, thanks @satoqz!

satoqz:
(resolved minor conflict on doc page)

cipherboy (approved):

PR #3760, opened 2026-08-11
Make OCI plugin binary_name fall back to ENTRYPOINT/CMD

satoqz:
## Description

This is a neat trick we can use to make setting `binary_name` mostly unnecessary and automatic.

cc @JanMa, thoughts?

## Acknowledgements

       also allowed.


internal/helper/pluginutil/oci/downloader.go
```diff
@@ -204,13 +204,39 @@ func (d *PluginDownloader) DownloadPlugin(ctx context.Context, config *server.Pl
...
+	binaryName := config.BinaryName
+	if binaryName == "" {
+		cf, err := img.ConfigFile()
+		if err != nil {
+			return fmt.Errorf("failed to get OCI image config: %w", err)
+		}
+		switch {
+		case len(cf.Config.Entrypoint) == 1:
+			binaryName = cf.Config.Entrypoint[0]
+			logger.Debug("picked binary name from entrypoint", "binary_name", binaryName)
```
wslabosz-reply, line 216: are we keeping the debug logs for the long run?

wslabosz-reply (dismissed):
lgtm

internal/helper/pluginutil/oci/downloader.go
satoqz: I didn't even use these in development, it just seemed useful to add for others to be able to debug where it gets picked from if there is trouble. There are a bunch of other debug logs in the downloader for similar purposes.

JanMa (approved):
Nice one, thanks for that @satoqz

JanMa:
Now we just need to get rid of that double SHA sum 😄

satoqz:
> Now we just need to get rid of that double SHA sum 😄

@JanMa stay tuned for that one, I think I've figured it out... in my head.

PR #3766, opened 2026-08-12
Allow for lease and `sys/raw` requests propagation to leader node

### LINKED ISSUE #3709: Cannot write to readonly storage for renew of dynamic credentials
**Describe the bug**
I looked in #2815 but I couldn't figure out if something was fixed there or not. My current problem is not creating the lease, but every time vault-agent container tries to renew it based on the ttl.
I have 3 vault pods  and my vault-agent container points to the service that does not have in selector openbao-active true. So the request to renew can go to any pod. In vault I didn't have this problem, I believe they forwarded the requests from standby pods to the active one.
Is this intended?

Vault gives the following error:
[ERROR] secrets.system.system_da598944: lease renewal failed: lease_id=-path-/QKho2BfxN6oBllba6bNRwzCf error="failed to persist lease entry: cannot write to readonly storage"

vault-agent container gives the following error:
URL: PUT http://-url-:8200/v1/sys/leases/renew
Code: 500. Errors:
* 1 error occurred:
	* operation could not be completed as storage is read-only

wslabosz-reply:
## Description
This PR removes `handleErrorNoReadOnlyForward` as a special "read-only" error catching mechanism from the `sys/raw` and lease handling operations. Additionally makes all "write-incurring" lease operations forwarded by default, leaving read requests untouched.

## Rationale
Lease renewal was failing when routed to the standby node with following error:
```
[ERROR] secrets.system.system_da598944: lease renewal failed: lease_id=-path-/<id>error="failed to persist lease entry: cannot write to readonly storage"
```
lease operations were never retried on the leader due to usage of `handleErrorNoReadOnlyForward` (removed with this PR) which was swallowing the "read-only error" category  nor they were propagated on the router level. This PR should allow seamless lease operation handling without manually directing the requests to the leader node.

_I think the reason it was implemented without propagation was due to the existence of `secondary performance clusters` within VE, where you could have "locally" mounted backend with their own set of leases, which also meant possible diverging storages between one clusters. With that setup, sys/raw or lease operations could be propagated to the node without knowledge of the requested entity_

This also matches the explanation in the original code commit message of:

```
commit f3656e80f072b7ef9335e2cb84bc2cbd7017d8fb
Author: Jeff Mitchell <jeff@hashicorp.com>
Date:   Sat Mar 17 21:29:17 2018 -0400

    Properly forward (or specifically don't) sys calls that result in read only errors (#4129)

    Prior to this policy writes against a performance secondary would not
    succeed because the read-only error was swallowed by handleError. In
    addition to fixing this, it adds a similar function that explicitly
    doesn't trigger forwarding. This is useful for things that are local to
    the secondary such as raw operations and lease management.
```

Resolves: #3709

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/logical_system.go
+++ b/internal/vault/logical_system.go
@@ -1195,22 +1195,6 @@ func handleError(
 	}
 }
 
-// Performs a similar function to handleError, but upon seeing a ReadOnlyError
-// will actually strip it out to prevent forwarding
-func handleErrorNoReadOnlyForward(
-	err error,
-) (*logical.Response, error) {
-	if logical.ShouldForward(err) {
-		return nil, errors.New("operation could not be completed as storage is read-only")
-	}
-	switch err.(type) {
-	case logical.HTTPCodedError:
-		return logical.ErrorResponse(err.Error()), err
-	default:
-		return logical.ErrorResponse(err.Error()), logical.ErrInvalidRequest
-	}
-}
-
 // handleUnmount is used to unmount a path
 func (b *SystemBackend) handleUnmount(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
 	path := data.Get("path").(string)
@@ -2008,10 +1992,11 @@ func (b *SystemBackend) handleLeaseLookupList(ctx context.Context, req *logical.
 	if err != nil {
 		return nil, err
 	}
+
 	keys, err := b.Core.expiration.leaseView(ns).List(ctx, prefix)
 	if err != nil {
 		b.Backend.Logger().Error("error listing leases", "prefix", prefix, "error", err)
-		return handleErrorNoReadOnlyForward(err)
+		return handleError(err)
 	}
 	return logical.ListResponse(keys), nil
 }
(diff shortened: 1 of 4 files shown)


internal/vault/logical_system.go
```diff
@@ -2008,10 +1992,11 @@ func (b *SystemBackend) handleLeaseLookupList(ctx context.Context, req *logical.
 	if err != nil {
 		return nil, err
 	}
+
 	keys, err := b.Core.expiration.leaseView(ns).List(ctx, prefix)
 	if err != nil {
 		b.Backend.Logger().Error("error listing leases", "prefix", prefix, "error", err)
-		return handleErrorNoReadOnlyForward(err)
+		return handleError(err)
```
cipherboy, line 1999: This makes sense to me generally. I think even if/when we were introduce cross-node lease forwarding (creation), it still makes sense to forward this manually via HTTP forwarding if GRPC is not connected/is down.

cipherboy (approved):
I'm happy with this.

internal/vault/logical_system.go
wslabosz-reply: just to clarify: read/list lease(s) won't forward as we won't encounter "read-only storage" error here

satoqz (approved):

Issue #3755, opened 2026-08-11
PQC-enabled TLS listeners


cipherboy:
Part of #496 

This issue is about enabled PQC certificates and cipher suites in the TLS listener configuration: https://openbao.org/docs/configuration/listener/tcp/ 

I think we need three primary things here:

1. Testing existing interfaces work with ML-DSA server/client certificates. (excluding the actual cert auth engine portion)
2. Testing PQC-enabled cipher suites + groups are negotiated.
3. Documentation that listeners are now PQC-enabled and how.

datosh:
I would like to work on this ticket.

datosh:
# Verify ML-DSA Support for TLS Listener

I have checked out PR #3752 (fix-go127-compat) and ran some local tests to see how OpenBao works with ML-DSA, today. No changes to the OpenBao code.

tl;dr: works as expected.

## Install Go RC and Build OpenBao

```sh
go install golang.org/dl/go1.27rc3@latest && go1.27rc3 download
go1.27rc3 version
go1.27rc3 build -o bin/bao . && ./bin/bao version
```

## Create ML-DSA Self-Signed Cert

```sh
dir="ml-dsa-demo"
mkdir -p "$dir"

# OpenSSL 3.6.3 (>=3.5 should work)
openssl version

openssl genpkey -algorithm ML-DSA-65 \
    -provparam ml-dsa.output_formats=seed-only \
    -out "$dir/server-key.pem"

openssl req -new -x509 -key "$dir/server-key.pem" -out "$dir/server.pem" \
  -days 90 -subj "/CN=localhost" \
  -addext "subjectAltName=DNS:localhost,IP:127.0.0.1,IP:::1"

# Shows: ML-DSA-65
openssl x509 -in "$dir/server.pem" -noout -text | grep -E 'Signature Algorithm|Public Key Algorithm'
```

## OpenBao Config

Very basic OpenBao config, note TLS 1.3 is required.

Put into `ml-dsa-demo/server.hcl`:

```hcl
storage "inmem" {}

listener "tcp" {
  address       = "127.0.0.1:8200"
  tls_cert_file = "ml-dsa-demo/server.pem"
  tls_key_file  = "ml-dsa-demo/server-key.pem"

  tls_min_version = "tls13"
}

api_addr = "https://127.0.0.1:8200"
```

```
./bin/bao server -config=ml-dsa-demo/server.hcl
```

## Custom Client Verification

This client exposes the connection details for verification.

[...]

PR #3769, opened 2026-08-12
PQC KEM Configuration & Policies

datosh:
## Description

This pull request will focus on the Key Exchange Mechanism (KEM) portion for TLS listeners, and PQC authentication will follow in a separate one, since it requires Go 1.27 (currently RC), which is [in progress](https://github.com/openbao/openbao/pull/3752).

This pull requests introduces a new configuration option `tls_curve_preferences` to enable PQC KEM policies, i.e., enforce PQC KEMs server side.

### PQC KEM support in OpenBao

Go 1.24 is used since OpenBao 2.2.x, and assuming default TLS config, [automatically supports and prioritizes PQC KEM](https://go.dev/doc/go1.24#cryptotlspkgcryptotls):

```
› cat example.openbao.d/server.hcl 
storage "inmem" {}

listener "tcp" {
  tls_cert_file = "example.openbao.d/tls/server.pem"
  tls_key_file  = "example.openbao.d/tls/server-key.pem"
}

api_addr = "https://127.0.0.1:8200"

› ./example.openbao.d/releases/bao-2.2.2 server -config=example.openbao.d/server.hcl

WARNING: storage configured to use "inmem" which should NOT be used in
production

==> OpenBao server configuration:

Administrative Namespace: 
             Api Address: https://127.0.0.1:8200
                     Cgo: disabled
         Cluster Address: https://127.0.0.1:8201
   Environment Variables: COLORTERM, COMMAND_MODE, FPATH, HOME, HOMEBREW_CELLAR, HOMEBREW_PREFIX, HOMEBREW_REPOSITORY, INFOPATH, KITTY_INSTALLATION_DIR, KITTY_PID, KITTY_PUBLIC_KEY, KITTY_WINDOW_ID, LANG, LESS, LOGNAME, LSCOLORS, LS_COLORS, MANPATH, NIX_PROFILES, NIX_SSL_CERT_FILE, NVM_BIN, NVM_CD_FLAGS, NVM_DIR, NVM_INC, OLDPWD, OSLogRateLimit, PAGER, PATH, PWD, SHELL, SHLVL, SSH_AUTH_SOCK, TERM, TERMINFO, TMPDIR, USER, WINDOWID, XDG_DATA_DIRS, XPC_FLAGS, XPC_SERVICE_NAME, ZSH, _, __CFBundleIdentifier, __CF_USER_TEXT_ENCODING, __ETC_PROFILE_NIX_SOURCED
              Go Version: go1.24.3
              Listener 1: tcp (addr: "127.0.0.1:8200", cluster address: "127.0.0.1:8201", max_request_duration: "1m30s", max_request_size: "33554432", tls: "enabled")
               Log Level: 
           Recovery Mode: false
                 Storage: inmem
                 Version: OpenBao v2.2.2, built 2025-05-29T19:40:39Z
             Version Sha: 8c0322c0423231836a1432fdad29dc2e99640da9

==> OpenBao server started! Log data will stream in below:

2026-08-12T12:19:36.486+0200 [INFO]  proxy environment: http_proxy="" https_proxy="" no_proxy=""
2026-08-12T12:19:36.489+0200 [INFO]  core: Initializing version history cache for core
```

Generic Go TLS client

```
› go run ./client -probe -addr 127.0.0.1:8200                                           
--- client: negotiated connection ---
  version:      TLS 1.3
  cipher suite: TLS_AES_128_GCM_SHA256
  key exchange: X25519MLKEM768 (0x11ec)  [post-quantum hybrid]
```

Still, a client is perfectly capable of 'downgrading' a connection either by using outdated software or explicitly insecure configuration:

```
go run ./client -probe -addr 127.0.0.1:8200 -tls-version 1.2
--- client: negotiated connection ---
  version:      TLS 1.2
  cipher suite: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
  key exchange: X25519 (0x001d)
› go run ./client -probe -addr 127.0.0.1:8200 -curves X25519  
--- client: negotiated connection ---
  version:      TLS 1.3
  cipher suite: TLS_AES_128_GCM_SHA256
  key exchange: X25519 (0x001d)
```

### PQC KEM requires TLS 1.3

[RFC9851](https://datatracker.ietf.org/doc/rfc9851/) states that TLS 1.2 is in maintenance-only mode and "PQC for TLS 1.2 will not be specified".

### Cipher Suites changes in TLS 1.3

In TLS 1.2 we had cipher suites like `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`, which define the full set of cryptographic algorithms used.

TLS 1.3 splits this into different parameter sets, where we still have cipher suites, but they focus on symmetric encryption like `TLS_AES_128_GCM_SHA256`, and then a dedicated set of preferences, e.g., for KEM like `X25519` and `X25519MLKEM768` (for a PQC hybrid scheme).

### Impact on OpenBao

Today, OpenBao supports `tls_cipher_suites` as a config parameter. They still exist in TLS 1.3, but are not configurable in Go, as per [this post](https://go.dev/blog/tls-cipher-suites#gos-cryptotls-and-cipher-suites) and [this issue](https://github.com/golang/go/issues/29349#issuecomment-449143820). We can also see in [handshake_server_tls13.go](https://github.com/golang/go/blob/go1.26.5/src/crypto/tls/handshake_server_tls13.go#L179-L196) that Config.CipherSuites is not used.

### Implementation

I introduce `tls_curve_preferences`, backed by [tls.CurveID](https://github.com/golang/go/blob/go1.26.5/src/crypto/tls/common.go#L146-L156).

When new `CurveID`s are introduced in Go, OpenBao needs to add support for the configuration fields and ID mapping. I have added a unit test that will fail in that event with actionable instructions.

## Rationale

Makes progress on https://github.com/openbao/openbao/issues/3755

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/helper/configutil/listener.go
+++ b/internal/helper/configutil/listener.go
@@ -77,6 +77,7 @@ type Listener struct {
 	TLSMaxVersion                    string        `hcl:"tls_max_version"`
 	TLSCipherSuites                  []uint16      `hcl:"-"`
 	TLSCipherSuitesRaw               string        `hcl:"tls_cipher_suites"`
+	TLSCurvePreferences              []string      `hcl:"tls_curve_preferences"`
 	TLSRequireAndVerifyClientCert    bool          `hcl:"-"`
 	TLSRequireAndVerifyClientCertRaw any           `hcl:"tls_require_and_verify_client_cert"`
 	TLSClientCAFile                  string        `hcl:"tls_client_ca_file"`
--- a/internal/helper/configutil/listener.go
+++ b/internal/helper/configutil/listener.go
@@ -77,7 +77,7 @@ type Listener struct {
 	TLSMaxVersion                    string        `hcl:"tls_max_version"`
 	TLSCipherSuites                  []uint16      `hcl:"-"`
 	TLSCipherSuitesRaw               string        `hcl:"tls_cipher_suites"`
-	TLSCurvePreferences              []string      `hcl:"tls_curve_preferences"`
+	TLSKeyExchangePreferences        []string      `hcl:"tls_key_exchange_preferences"`
 	TLSRequireAndVerifyClientCert    bool          `hcl:"-"`
 	TLSRequireAndVerifyClientCertRaw any           `hcl:"tls_require_and_verify_client_cert"`
 	TLSClientCAFile                  string        `hcl:"tls_client_ca_file"`
--- a/internal/helper/listenerutil/listener.go
+++ b/internal/helper/listenerutil/listener.go
@@ -8,9 +8,11 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"maps"
 	"net"
 	"os"
 	osuser "os/user"
+	"slices"
 	"strconv"
 
 	"github.com/hashicorp/cli"
@@ -47,6 +49,26 @@ func (l *rmListener) Close() error {
 
 	// Remove the file
 	return os.Remove(l.Path)
+}
+
+var curveIDByName = map[string]tls.CurveID{
+	tls.CurveP256.String():          tls.CurveP256,
+	tls.CurveP384.String():          tls.CurveP384,
+	tls.CurveP521.String():          tls.CurveP521,
+	tls.X25519.String():             tls.X25519,
+	tls.X25519MLKEM768.String():     tls.X25519MLKEM768,
+	tls.SecP256r1MLKEM768.String():  tls.SecP256r1MLKEM768,
+	tls.SecP384r1MLKEM1024.String(): tls.SecP384r1MLKEM1024,
+}
+
+var allCurveNames = slices.Sorted(maps.Keys(curveIDByName))
+
+func CurveToCurveID(curve string) (tls.CurveID, error) {
+	id, ok := curveIDByName[curve]
+	if !ok {
+		return 0, fmt.Errorf("%s is an unknown curve", curve)
+	}
+	return id, nil
 }
 
 func UnixSocketListener(path string, unixSocketsConfig *UnixSocketsConfig) (net.Listener, error) {
@@ -92,10 +114,20 @@ func TLSConfig(
 	}
 	l.TLSCertGetter = cg
 
+	curvePreferences := []tls.CurveID{}
+	for _, curvePreference := range l.TLSCurvePreferences {
+		curveID, err := CurveToCurveID(curvePreference)
+		if err != nil {
+			return nil, nil, fmt.Errorf("'tls_curve_preferences' value %q not supported, please select from %v", curvePreference, allCurveNames)
+		}
+		curvePreferences = append(curvePreferences, curveID)
+	}
+
 	tlsConf := &tls.Config{
-		GetCertificate: cg.GetCertificate,
-		NextProtos:     []string{"h2", "http/1.1"},
-		ClientAuth:     tls.RequestClientCert,
+		GetCertificate:   cg.GetCertificate,
+		NextProtos:       []string{"h2", "http/1.1"},
+		ClientAuth:       tls.RequestClientCert,
+		CurvePreferences: curvePreferences,
 	}
 
 	if acg, ok := cg.(*ACMECertGetter); ok {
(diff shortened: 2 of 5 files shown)


internal/helper/configutil/listener.go
```diff
@@ -77,6 +77,7 @@ type Listener struct {
 	TLSMaxVersion                    string        `hcl:"tls_max_version"`
 	TLSCipherSuites                  []uint16      `hcl:"-"`
 	TLSCipherSuitesRaw               string        `hcl:"tls_cipher_suites"`
+	TLSCurvePreferences              []string      `hcl:"tls_curve_preferences"`
```
datosh: I would like to invite feedback on the naming here: [CurvePreferences](https://pkg.go.dev/crypto/tls#Config.CurvePreferences) is the name in Go stdlib, but this name is not accurate anymore as of TLS 1.3, since it supports also non ECC preferences. Still, the setting equally applies to <TLS1.3, and there it is curve preferences as per the spec.

Happy for any suggestions!

cipherboy: @datosh I've thought about it overnight now and I'm not sure I have any reasonable alternatives except perhaps `tls_kex_preferences` or `tls_key_exchange_preferences`. Its probably better than explaining that `curve_preferences` doesn't actually apply to leaf certificate types...? :thinking:

internal/helper/listenerutil/listener.go
```diff
@@ -49,6 +51,26 @@ func (l *rmListener) Close() error {
...
+	tls.X25519MLKEM768.String():     tls.X25519MLKEM768,
+	tls.SecP256r1MLKEM768.String():  tls.SecP256r1MLKEM768,
+	tls.SecP384r1MLKEM1024.String(): tls.SecP384r1MLKEM1024,
+}
+
+var allCurveNames = slices.Sorted(maps.Keys(curveIDByName))
+
+func CurveToCurveID(curve string) (tls.CurveID, error) {
+	id, ok := curveIDByName[curve]
+	if !ok {
```
cipherboy: Nit: you could return the bool here rather than the error since you discard the error in most cases.

internal/helper/listenerutil/listener.go
```diff
@@ -49,6 +51,26 @@ func (l *rmListener) Close() error {
...
 
+var curveIDByName = map[string]tls.CurveID{
+	tls.CurveP256.String():          tls.CurveP256,
+	tls.CurveP384.String():          tls.CurveP384,
+	tls.CurveP521.String():          tls.CurveP521,
+	tls.X25519.String():             tls.X25519,
+	tls.X25519MLKEM768.String():     tls.X25519MLKEM768,
+	tls.SecP256r1MLKEM768.String():  tls.SecP256r1MLKEM768,
+	tls.SecP384r1MLKEM1024.String(): tls.SecP384r1MLKEM1024,
+}
```
cipherboy, line 64: For us later: formal Go 1.27 bump will need to add `tls.MLKEM1024`.

datosh: `TestCurveToCurveIDMatchesGoStdLib` will remind us to do so. For example if you remove `SecP384r1MLKEM1024`, today, the test fails with:

```
listener_test.go:27: missing mapping in CurveToCurveID, please add new SecP384r1MLKEM1024 (0x11ed)
```

internal/helper/configutil/listener.go
datosh: I appreciate the input, naming really is the hardest problem. 

I have picked the longer name `tls_key_exchange_preferences`, since it reads more clearly in the config compared to `kex`, and amended the PR.

PTAL

internal/helper/listenerutil/listener.go
datosh: Good point, I have removed the function altogether, since it wasn't doing any work at that point.

cipherboy (dismissed):
This PR is fine to land as it doesn't depend on Go 1.27 update at all, which is still under discussion.

Changelog is fine for now, if we have a bigger announcement around PQC support in this release we'll drop it.

internal/helper/listenerutil/listener.go
```diff
@@ -49,6 +51,20 @@ func (l *rmListener) Close() error {
...
+
+var allKexNames = slices.Sorted(maps.Keys(curveIDByName))
```
satoqz: Nit: I'd rather fail slow on the error path instead of having this be constant global overhead.

satoqz (commented):
Makes sense to me, one tiny comment.

satoqz (approved):

satoqz:
Thanks!

Issue #3576, opened 2026-07-22
TOTP self-enrollment for users


Luiggi33:
**Is your feature request related to a problem? Please describe.**
When requiring MFA via TOTP for ldap users, there isn't a clean possibility (as far as I know) to allow for self enrollment.

**Describe the solution you'd like**
A option to allow for self enrollment through the UI when MFA is required user and they don't have any configured yet.

**Describe alternatives you've considered**
There could be a way through a second ldap mount that binds to the same entity, but only has permissions to access `identity/mfa/method/totp/generate` and a short TTL/usage number. This could be triggered via a workflow.

**Explain any additional use-cases**
N/A

**Additional context**
Hashicorp Vault has something similar in the [enable_self_enrollment](https://developer.hashicorp.com/vault/api-docs/secret/identity/mfa/totp#enable_self_enrollment) option when creating a TOTP MFA method

PR #3771, opened 2026-08-12
RFC: TOTP Self-Enrollment

Luiggi33:
## Description

As discussed on [Zulip](https://linuxfoundation.zulipchat.com/#narrow/channel/529890-openssf-openbao-discussion/topic/TOTP.20self.20enrollment/with/615619356), here is my RFC for adding TOTP self-enrollment.

A PoC can be found in https://github.com/openbao/openbao/pull/3772

## Rationale

Working on resolving: #3576 

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/community/rfcs/totp-self-enroll.mdx
+++ b/website/content/community/rfcs/totp-self-enroll.mdx
@@ -1,0 +1,150 @@
+---
+sidebar_label: TOTP Self-Enrollment
+description: |-
+  An OpenBao RFC for allowing users to enroll a TOTP MFA method on their own
+---
+
+# TOTP Self-Enrollment
+
+## Summary
+
+Allow a user to setup TOTP as their second factor natively in OpenBao. This should be interactive and require no additional administrative work.
+
+## Problem Statement
+
+When using TOTP as a mandatory second factor, a chicken-egg problem arises. The user can't log in because they can't provide a TOTP code and they can't enroll a TOTP second factor, because they can't log in.
+
+This leads to unnecessary overhead when wanting to enforce TOTP as a second factor with minimal administrative work.
+
+## User-facing description
+
+Current sequence for a user in the UI:
+1. Log in
+2. If a self-enrollable TOTP MFA method is enforced and not yet configured, redirection to the self-enrollemnt page
+3. Show a QR code as well as the TOTP secret to allow easy setup
+4. Confirm the setup with inputting a TOTP code
+5. Log in again, now with the TOTP MFA method configured
+
+## Technical Description
+
+### Configuration of self-enrollment
+
+The `TOTPConfig` controls if a TOTP MFA method allows self-enrollment via a `enable_self_enrollment` boolean option, configurable through the current api.
+This config option is also configurable through the UI, when creating a TOTP MFA method.
+
+### Detecting when to move into self-enrollment
+
+MFA enforcement is evaluated in `internal/vault/request_handling.go`.
+This is where the function `findSelfEnrollableTOTPMethod` checks if the MFA enforcement and its enforced methods can and should be self enrolled.
+
+
+1. If the entity already has an `MFASecrets` entry for *any* method ID in
+   the enforcement config, self-enrollment is skipped entirely, as the enforcement is satisfied.
+2. Otherwise, it returns the first method ID that is both a TOTP method and
+   has `enable_self_enrollment` enabled. Non-TOTP methods in the same constraint are skipped,
+   since there isn't self-enrollment for these methods.
+
+If a self-enrollable method is found, a TOTP key and key url are generated.
+For this `generateTOTPKeyAndQR` (`internal/vault/login_mfa.go`) was extracted out of the existing `HandleMFAGenerateTOTP` enrollment path so both share the same functionality and way of generation.
+This generated TOTP information is not persisted to the user yet, but returned including a mfa request (uu)id through a new response `TOTPSelfEnroll` to be confirmed on the user's end.
+
+### Pending-enrollment queue
+
+The previously generated secret has to survive during the confirmation period on the users end, without getting saved in a more persistent manner.
+This is solved through a new `TOTPSelfEnrollmentQueue` (`internal/vault/mfa_self_enrollment_queue.go`), which uses the already existing priority queue that is also used in `mfaResponseAuthQueue`.
+Entries include the request id, the requesting namespace, entity id, the generated secret, targeted method id and when it was stored.
+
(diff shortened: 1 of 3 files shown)


website/content/community/rfcs/totp-self-enroll.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+## Problem Statement
+
+When using TOTP as a mandatory second factor, a chicken-egg problem arises. The user can't log in because they can't provide a TOTP code and they can't enroll a TOTP second factor, because they can't log in.
+
+This leads to unnecessary overhead when wanting to enforce TOTP as a second factor with minimal administrative work.
+
+## User-facing description
+
+Current sequence for a user in the UI:
```
cipherboy:
```suggestion
Revised sequence for a user in the UI:
```

website/content/community/rfcs/totp-self-enroll.mdx
```diff
@@ -0,0 +1,151 @@
...
+A background goroutine evicts the oldest entry once it exceeds `defaultTOTPSelfEnrollmentTTL` (currently 5 minutes), limiting how long an unconfirmed enrollment request stays usable.
+This is again similar to `mfaResponseAuthQueue`.
+
+This means queued self-enrollments don't survive an active-node change.
+
+### Confirmation and cancellation endpoints
+
+Two new unauthenticated system paths are added to `SystemBackend` (`internal/vault/login_mfa.go`):
+
+- `POST /v1/sys/mfa/self-enroll/totp` (`request_id`, `totp_code`) -
```
cipherboy, line 79: I wonder if we can make this slick through a single-use token? 🤔 

If we yield a single-use token in addition to MFA config, server can keep track of state for us (`request_id`) via the token and we get durability, ttl, and a way to ensure these aren't unauthed paths, so there's no risk of what happens if `request_id` is leaked (it isn't, it behaves like a token and is self-revokable). Can have a policy scoped just to this endpoint + revoke-self. It'd also be single-use token so it'd self destruct anyways. 

Users would still have to auth post-enrollment (including MFA). 

Thoughts?

website/content/community/rfcs/totp-self-enroll.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+The auth controller (`ui/app/controllers/vault/cluster/auth.js`, which handles the login) checks for the new `mfa_self_enroll` response field and routes to a new `Mfa::TotpSelfEnroll` component.
+The component renders a QR code client-side (via the `qrcode` package, from the returned `totp_url`) and the raw secret as a fallback, and submits the confirmation or cancellation calls through two new `cluster` adapter methods that wrap the endpoints mentioned above.
+On success, the user is bounced back to the login form with a message asking them to log in again. When cancelled, the success of that is also sent to the user through that way.
+
+## Rationale and alternatives
+
+### One-time token to generate a TOTP secret vs the self-enrollment system
+
+The self-enrollment system is the barrier of entry, when a second factor is required. While OpenBao could just generate a one-time-usage token for the `/identity/mfa/method/totp/generate` endpoint, this would require equally as much work, while also creating a valid token that leaves more surface area to be attacked.
```
cipherboy: 🤔  I like the idea of a second endpoint, but I think tokens are probably at least as secure as the `request_id` scheme.

website/content/community/rfcs/totp-self-enroll.mdx
cipherboy: (We can choose if we have durability or not, but we at least would get durability for `request_id` in a pattern users are familiar with, like response wrapping.)

website/content/community/rfcs/totp-self-enroll.mdx
```diff
@@ -0,0 +1,151 @@
...
+- The self-enrollment queue silently doesn't survive active node changes or restarts
+- Re-login is a worse experience for the user
+- Confirmation but especially the invalidation of a self-enrollment attempt hinge on the request uuid
+- Loss of admin driven MFA enrollment for TOTP
+
+## Security Implications
+
+### Primary auth leads to full trust
+
+The self-enrollment only requires a username and password. This sounds obvious, but because there are no email notifications or similar to a user, a wrong enrollment by a bad actor won't per-se be noticed.
```
cipherboy:
```suggestion
The self-enrollment only requires valid first-factor authentication. This sounds obvious, but because there are no email notifications or similar to a user, a wrong enrollment by a bad actor won't per-se be noticed.
```

(There's non-userpass auth methods, like cert or jwt or similar.)

cipherboy (commented):
Some thoughts @Luiggi33!

website/content/community/rfcs/totp-self-enroll.mdx
Luiggi33: To be honest, I only chose this way, because I wasn't sure how to make that work in code at that time 😅
I agree that it would be better to have both the `request_id` and single-use token to confirming or "abort" the self-enrollment

Luiggi33: So in theory that sounds great, but I noticed that a single use (as in one usage only) does not work when you make a wrong input with the totp code and then want to try again/cancel the request afterwards. How is this handled elsewhere or do I just make something up?

cipherboy: @Luiggi33 Hmm, wouldn't that mean enrollment failed, so they'd just restart the process from scratch? :thinking:

Luiggi33: Which we could do but seems like a bad experience for a user, but if we accept that as a "risk" we could just kick them back to the login screen with a "self enroll failed, try again" notification.
I tried going the other way in the PoC with https://github.com/openbao/openbao/pull/3772/changes/7bafc8aab0fb6387b7a3f3f100b1f7982c1742fe, where we have a timed token which has unlimited usage but is only scoped for the totp endpoint and gets revoked when either path (cancelled or success) is taken

Luiggi33: @cipherboy I worked the above mentioned changes into the self enrollment. I am unsure about making the final call on the token.
The pro single use token:
- more secure (forces reauth on fail)
- shouldn't be too big of a inconvenience when doing a typo, as self enrollment is a one time thing (it should at least)

The contra:
- annoying if you have to put in all your credentials again just for a typo

website/content/community/rfcs/totp-self-enroll.mdx
Luiggi33: I removed this segment, as I feel it is not a relevant question. I originally posed it when I did not think I could just create the endpoints and mint a minimal usage token for the user to auth against these endpoints

PR #3774, opened 2026-08-12
Decrease delay in acquiring HA lock

cipherboy:
## Description

On transitive failures on fresh loops, use exponential backoff during lock acquisition to ensure that we don't always blindly block for 10 seconds. This can lead to a burst of database traffic, though, we should eventually go back to 10s+ backoffs. 

## Rationale

15ms initial value seems reasonably safe to support faster local HA setups versus longer network backed setups (with randomization + exponential growth).

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/ha.go
+++ b/internal/vault/ha.go
@@ -14,11 +14,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/hashicorp/errwrap"
-	metrics "github.com/hashicorp/go-metrics/compat"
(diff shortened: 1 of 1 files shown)


satoqz (approved):

cipherboy:
@satoqz Do you think we should add a changelog for this? Or assume it fits in the rest of our general HA rejiggering for 2.7? :thinking:

satoqz:
@cipherboy I think it's fine not to.

PR #3787, opened 2026-08-13
Fix flaky final second test to land predictably in last second

### LINKED ISSUE #3786: Flaky test TestExpiration_Renew_FinalSecond
https://github.com/openbao/openbao/pull/3769's CI failed on `TestExpiration_Renew_FinalSecond` after looking into the test I think it is flaky (timing issue), rather than introduced by my changes. 

I don't think anyone else reported this, nor is it tracked in https://github.com/openbao/openbao/issues/2581

From what I understand: The test registers a lease with TTL and MaxTTL of 2 seconds. It then sleeps for 1s, in order to renew in the "final second".

The issue arises, because the renew path involves `sdk/framework/lease.go:40-44` which truncates to full seconds (`Truncate(time.Second)`). So if the registration happens right before the clock ticks over to the next second, the 1s sleep + additional overhead move us 2 seconds further (past the MaxTTL), and the assertion fails. As we can see in the [CI logs](https://github.com/openbao/openbao/actions/runs/31608590496/job/94154429072?pr=3769), the test executed registration right around the second boundary at 2026-08-12T14:54:33.**997**Z, so sleep and additional overhead tipping it over into 2026-08-12T14:54:**35.000Z**.

I was **NOT** able to reproduce this locally, using: 

```sh
go test ./internal/vault/ -race -count=100 -timeout 30m -run 'TestExpiration_Renew_FinalSecond$' -v -failfast
```

But when I manually aligned the test execution to second boundary (which also demonstrates the issue nicely), the test fails locally as well!

```diff
diff --git a/internal/vault/expiration_test.go b/internal/vault/expiration_test.go
index 8c8f374cf8..f914ac044a 100644
--- a/internal/vault/expiration_test.go
+++ b/internal/vault/expiration_test.go
@@ -2076,6 +2076,12 @@ func TestExpiration_Renew_FinalSecond(t *testing.T) {
 		},
 	}
 
+	target := time.Now().Truncate(time.Second).Add(999 * time.Millisecond)
+	if target.Before(time.Now()) {
+		target = target.Add(time.Second)
+	}
+	time.Sleep(time.Until(target))
+
 	ctx := namespace.RootContext(t.Context())
 	id, err := exp.Register(ctx, req, resp, "")
 	if err != nil {
```

showed

```
expiration_test.go:2118: err: past the max TTL, cannot renew
```

datosh:
## Description

See #3786 for background information. The modified sleep will always wait until the beginning of the final second, giving the test more time to succeed within it's allotted bucket. 

I noticed that the same flakiness also exists in `TestExpiration_Renew_FinalSecond_Lease` so I added the same fix there.

## Rationale

Resolves: #3786 

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/expiration_test.go
+++ b/internal/vault/expiration_test.go
@@ -2106,7 +2106,9 @@ func TestExpiration_Renew_FinalSecond(t *testing.T) {
 		},
 	}
 
-	time.Sleep(1000 * time.Millisecond)
+	// Follow CalculateTTL truncation to predictably land in final second bucket
+	finalBucket := le.IssueTime.Truncate(time.Second).Add(time.Second)
+	time.Sleep(time.Until(finalBucket))
 	_, err = exp.Renew(ctx, id, 0)
 	if err != nil {
 		t.Fatalf("err: %v", err)
@@ -2168,7 +2170,9 @@ func TestExpiration_Renew_FinalSecond_Lease(t *testing.T) {
 	}
 	exp.persistEntry(ctx, le)
 
-	time.Sleep(1000 * time.Millisecond)
+	// Follow CalculateTTL truncation to predictably land in final second bucket
+	finalBucket := le.IssueTime.Truncate(time.Second).Add(time.Second)
+	time.Sleep(time.Until(finalBucket))
 	_, err = exp.Renew(ctx, id, 0)
 	if err != nil {
 		t.Fatalf("err: %v", err)


datosh:
I don't think I need to add a change log entry for test fixes? But do let me know otherwise!

wslabosz-reply (approved):
Thanks @datosh!

PR #3790, opened 2026-08-13
Propagate all 'wrapped-requested' requests

### LINKED ISSUE #3026: AppRole role-id read with response wrapping fails on standby/read-only nodes with internal error
## Description

In an HA deployment with 1 active leader and 2 standby/read-only nodes behind an F5 load balancer, AppRole-based SaltStack authentication fails on standby nodes.

This was first observed through [saltext-vault](https://github.com/salt-extensions/saltext-vault), but the issue appears to be in OpenBao itself rather than SaltStack:

- cold-cache Salt auth performs:
  - wrapped `role-id` read
  - wrapped `secret-id` generation
  - AppRole `login`
- requests routed to standby/read-only nodes fail
- after upgrading from `2.5.2` to `2.5.3`, the behaviour is still present

There are two failure modes:

1. `auth/<mount>/login` on a standby node fails with:
   - `cannot write to readonly storage`

2. `auth/<mount>/role/<name>/role-id` with response wrapping fails with:
   - audit log shows the AppRole read itself succeeded
   - final response is `internal error`

The second one is surprising because `role-id` read is logically read-only, but it still fails on standby when response wrapping is enabled.

## Why this looks like an OpenBao bug

From the OpenBao source in `v2.5.3`:

- `pathRoleRoleIDRead` only reads the role and returns `role_id`:
  - https://github.com/openbao/openbao/blob/v2.5.3/builtin/credential/approle/path_role.go
- request handling applies response wrapping after the handler returns:
  - https://github.com/openbao/openbao/blob/v2.5.3/vault/request_handling.go
- `wrapInCubbyhole` creates a wrapping token and persists wrapping data in cubbyhole / lease state:
  - https://github.com/openbao/openbao/blob/v2.5.3/vault/wrapping.go

So a wrapped read request becomes write-backed because wrapping persists state. On a standby/read-only node this appears to fail instead of being forwarded to the active node.

`v2.5.3` release notes mention a standby fix for some dynamic secret engines, but this AppRole + wrapping path still seems broken:
- https://github.com/openbao/openbao/releases/tag/v2.5.3

## To Reproduce

HA topology:

- 3 OpenBao nodes
- 1 active leader
- 2 standby/read-only nodes
- integrated RAFT storage / HA setup
- TLS enabled
- AppRole auth mount for Salt minions at `auth/salt-ops-minions/`

### Reproducer 1: wrapped role-id read on standby

Send a `role-id` read with response wrapping to a standby node, for example:

```bash
bao read -wrap-ttl=120s auth/salt-ops-minions/role/labs-1.example.net/role-id
```

If the request lands on a standby/read-only node, the audit log shows:

- request `path: auth/salt-ops-minions/role/labs-1.example.net/role-id`
- operation: `read`
- `wrap_ttl: 120`
- mount version: v2.5.3+builtin.bao

But the response contains:

- empty response body
- `error: "internal error"`

### Reproducer 2: AppRole login on standby

After obtaining valid role_id and secret_id, send:

```
bao write auth/salt-ops-minions/login \
 role_id=<role-id> \
 secret_id=<secret-id>
```

If this hits a standby/read-only node, the audit log response contains:

 `error: "cannot write to readonly storage"`

## Expected behaviour

One of:

1. standby nodes should forward these requests to the active node, or
2. standby nodes should return a clear standby/read-only error instead of `internal error`

In particular, a wrapped role-id read should not fail as an opaque `internal error`.

### Wrapped role-id read on standby

Sanitised audit log:
```
{
  "time": "2026-05-02T22:56:01.784865682Z",
  "type": "request",
  "auth": {
    "client_token": "<redacted>",
    "accessor": "<redacted>",
    "display_name": "approle",
    "policies": ["default", "salt_ops_master_policy"],
    "token_policies": ["default", "salt_ops_master_policy"],
    "policy_results": {
      "allowed": true,
      "granting_policies": [
        {
          "name": "salt_ops_master_policy",
          "namespace_id": "root",
          "type": "acl"
        }
      ]
    },
    "metadata": {
      "role_name": "app_salt_ops_master"
    },
    "entity_id": "<redacted>",
    "token_type": "service",
    "token_ttl": 2764800,
    "token_issue_time": "2026-04-13T00:45:21Z"
  },
  "request": {
    "id": "3ca15655-c9ff-1f72-0fef-8c914c6aed31",
    "client_id": "<redacted>",
    "operation": "read",
    "mount_point": "auth/salt-ops-minions/",
    "mount_type": "approle",
    "mount_running_version": "v2.5.3+builtin.bao",
    "mount_class": "auth",
    "client_token": "<redacted>",
    "client_token_accessor": "<redacted>",
    "namespace": {
      "id": "root"
    },
    "path": "auth/salt-ops-minions/role/labs-1.example.net/role-id",
    "remote_address": "2001:db8::10",
    "remote_port": 45790,
    "wrap_ttl": 120
  }
}
{
  "time": "2026-05-02T22:56:01.785316344Z",
  "type": "response",
  "auth": {
    "client_token": "<redacted>",
    "accessor": "<redacted>",
    "display_name": "approle",
    "policies": ["default", "salt_ops_master_policy"],
    "token_policies": ["default", "salt_ops_master_policy"],
    "policy_results": {
      "allowed": true,
      "granting_policies": [
        {
          "name": "salt_ops_master_policy",
          "namespace_id": "root",
          "type": "acl"
        }
      ]
    },
    "metadata": {
      "role_name": "app_salt_ops_master"
    },
    "entity_id": "<redacted>",
    "token_type": "service",
    "token_ttl": 2764800,
    "token_issue_time": "2026-04-13T00:45:21Z"
  },
  "request": {
    "id": "3ca15655-c9ff-1f72-0fef-8c914c6aed31",
    "client_id": "<redacted>",
    "operation": "read",
    "mount_point": "auth/salt-ops-minions/",
    "mount_type": "approle",
    "mount_accessor": "auth_approle_ab6ca5a0",
    "mount_running_version": "v2.5.3+builtin.bao",
    "mount_class": "auth",
    "client_token": "<redacted>",
    "client_token_accessor": "<redacted>",
    "namespace": {
      "id": "root"
    },
    "path": "auth/salt-ops-minions/role/labs-1.example.net/role-id",
    "remote_address": "2001:db8::10",
    "remote_port": 45790,
    "wrap_ttl": 120
  },
  "response": {
    "mount_point": "auth/salt-ops-minions/",
    "mount_type": "approle",
    "mount_accessor": "auth_approle_ab6ca5a0",
    "mount_running_plugin_version": "v2.5.3+builtin.bao",
    "mount_class": "auth"
  },
  "error": "internal error"
}
```
### AppRole login on standby
Sanitised audit log:
```
{
  "time": "2026-05-02T12:45:23.233404051Z",
  "type": "request",
  "auth": {
    "policy_results": {
      "allowed": true
    },
    "token_type": "default"
  },
  "request": {
    "id": "b9254cf2-95b9-fa9c-7c92-af37a6656d69",
    "operation": "update",
    "mount_point": "auth/salt-ops-minions/",
    "mount_type": "approle",
    "mount_accessor": "auth_approle_ab6ca5a0",
    "mount_running_version": "v2.5.3+builtin.bao",
    "mount_class": "auth",
    "namespace": {
      "id": "root"
    },
    "path": "auth/salt-ops-minions/login",
    "data": {
      "role_id": "<redacted>",
      "secret_id": "<redacted>"
    },
    "remote_address": "2001:db8::10",
    "remote_port": 48956
  }
}
{
  "time": "2026-05-02T12:45:23.234346919Z",
  "type": "response",
  "auth": {
    "policy_results": {
      "allowed": true
    },
    "token_type": "default"
  },
  "request": {
    "id": "b9254cf2-95b9-fa9c-7c92-af37a6656d69",
    "operation": "update",
    "mount_point": "auth/salt-ops-minions/",
    "mount_type": "approle",
    "mount_accessor": "auth_approle_ab6ca5a0",
    "mount_running_version": "v2.5.3+builtin.bao",
    "mount_class": "auth",
    "namespace": {
      "id": "root"
    },
    "path": "auth/salt-ops-minions/login",
    "data": {
      "role_id": "<redacted>",
      "secret_id": "<redacted>"
    },
    "remote_address": "2001:db8::10",
    "remote_port": 48956
  },
  "response": {
    "mount_point": "auth/salt-ops-minions/",
    "mount_type": "approle",
    "mount_accessor": "auth_approle_ab6ca5a0",
    "mount_running_plugin_version": "v2.5.3+builtin.bao",
    "mount_class": "auth"
  },
  "error": "cannot write to readonly storage"
}
```

## Additional context

This surfaced via Salt `saltext-vault` on cache-miss authentication. The extension uses response wrapping for AppRole auth material, so a cold-cache path includes requests with wrapped `role-id` and `secret-id`.

That means a generic HA endpoint, which includes standby/read-only nodes, is not safe for this workflow unless the requests are forwarded correctly.

## Environment

- OpenBao Server Version: 2.5.3
- Previous version tested: 2.5.2
- HA topology: 3 nodes total
- Active nodes: 1
- Standby/read-only nodes: 2
- AppRole auth mount: auth/salt-ops-minions/
- TLS enabled

## Questions

1. Is wrapped auth/.../role/.../role-id expected to work on standby nodes?
2. If not, should it return a clearer read-only/standby error rather than an internal error?
3. Should response-wrapped auth reads be forwarded to the active node in HA mode?

wslabosz-reply:
## Description
This PR adds a `case req.WrapInfo != nil && req.WrapInfo.TTL != 0:` check on `(*Core).handleCancelableRequest()` method to catch the "wrapping requested" requests to propagate before handling on standby nodes as they ultimate fail with `read-only storage` error.

## Rationale
```
    // Request wrapping incurs storage (cubbyhole) writes, if the request
	// is handled on standby node, it doesn't fail and response is not empty
	// it will ultimately fail to save the wrapping token, and we'd still
	// have to forward the request.
```
comment explains the requirement, which otherwise requires the downstream adaptors to always track the current leader node to always direct the wrapping request there.

Resolves: #3026

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/request_handling.go
+++ b/internal/vault/request_handling.go
@@ -1051,6 +1051,16 @@ func (c *Core) handleCancelableRequest(ctx context.Context, req *logical.Request
 		if c.standby.Load() && !c.StandbyReadsEnabled() {
 			return nil, ErrCannotForwardLocalOnly
 		}
+
+	// Request wrapping incurs storage (cubbyhole) writes, if the request
+	// is handled on standby node, it doesn't fail and response is not empty
+	// it will ultimately fail to save the wrapping token, and we'd still
+	// have to forward the request.
+	// Preemptively forward the requests with wrapping info provided.
+	case req.WrapInfo != nil && req.WrapInfo.TTL != 0:
+		if c.Standby() {
+			return nil, logical.ErrPerfStandbyPleaseForward
+		}
 	}
 
 	var auth *logical.Auth
@@ -1219,22 +1229,19 @@ func (c *Core) doRouting(ctx context.Context, req *logical.Request) (*logical.Re
 
 // doRoutingIfApproved will check if the request needs approval before routing
 func (c *Core) doRoutingIfApproved(ctx context.Context, req *logical.Request, auth *logical.Auth) (*logical.Response, error) {
-	var routeErr error
-	var resp *logical.Response
-	if !c.needsApproval(ctx, req, auth) {
-		resp, routeErr = c.doRouting(ctx, req)
+	if !c.needsApproval(req, auth) {
+		return c.doRouting(ctx, req)
 	} else {
-		resp = &logical.Response{}
+		return &logical.Response{}, nil
 	}
-	return resp, routeErr
 }
 
 func (c *Core) isLoginRequest(ctx context.Context, req *logical.Request) bool {
 	return c.router.LoginPath(ctx, req.Path)
 }
 
-// needsApproval will assess if a ControlGroup policy is applicable, so that request is deferred until unwrap
-func (c *Core) needsApproval(ctx context.Context, req *logical.Request, auth *logical.Auth) bool {
+// needsApproval will assess if a ControlGroup policy is applicable, so that request is deferred until unwrap.
+func (c *Core) needsApproval(req *logical.Request, auth *logical.Auth) bool {
 	if auth.PolicyResults != nil &&
 		auth.PolicyResults.ControlGroup != nil &&
 		req.ForwardedFrom != forwardedFromDeferral {
@@ -1418,7 +1425,7 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
 		}
 
 		// Set wrapTTL from ControlGroup.TTL if present
-		if c.needsApproval(ctx, req, auth) {
+		if c.needsApproval(req, auth) {
 			cgTTL := auth.PolicyResults.ControlGroup.TTL
 			if cgTTL > 0 {
 				wrapTTL = cgTTL
--- a/changelog/3790.txt
+++ b/changelog/3790.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Propagate requests with provided wrap info to mitigate 'read-only storage' error on standby nodes. 
+```
--- a/internal/vault/external_tests/standby/read_only_standby_test.go
+++ b/internal/vault/external_tests/standby/read_only_standby_test.go
@@ -11,9 +11,11 @@ import (
 	"github.com/openbao/openbao/api/v2"
 	logicalKv "github.com/openbao/openbao/v2/internal/builtin/logical/kv"
 	logicalPki "github.com/openbao/openbao/v2/internal/builtin/logical/pki"
+	"github.com/openbao/openbao/v2/internal/helper/namespace"
 	"github.com/openbao/openbao/v2/internal/helper/testhelpers"
 	"github.com/openbao/openbao/v2/internal/helper/testhelpers/teststorage"
 
+	"github.com/openbao/openbao/sdk/v2/helper/consts"
 	"github.com/openbao/openbao/sdk/v2/logical"
 	vaulthttp "github.com/openbao/openbao/v2/internal/http"
 	"github.com/openbao/openbao/v2/internal/vault"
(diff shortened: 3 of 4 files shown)


internal/vault/request_handling.go
```diff
@@ -2672,7 +2680,7 @@ func (c *Core) DecodeSSCTokenInternal(token string) (*tokens.Token, error) {
 	return plainToken, nil
 }
 
-func (c *Core) checkSSCTokenInternal(ctx context.Context, token string) (string, error) {
+func (c *Core) checkSSCTokenInternal(token string) (string, error) {
```
suprjinx: It seems like a lot of the change in this PR relates to removing the ctx param -- presumably it was needed at some point, and it's pretty common to share down the stack

suprjinx: Which is to say, it's a common idiom and does no harm if unused, but makes the PR a lot more busy :-)

wslabosz-reply: I believe in a boy scout rule, unused parameter does no harm, but also doesn't do any good. It is precisely because of the small scope of the proposed changes that I've opted to clean up the unused parameter. I don't really get the argument of "common idiom, leave for now, maybe will be used later", I think the cost of reintroduction doesn't exceed the cost of removal. Even when introduced with 15c4855cf6c152dce4828292ec4b4b76c3206f37 it wasn't used, and that was 2 years ago.

I'm not opposed to reverting, but I just wanted to highlight that I do not agree with your reasoning.

cipherboy: I think the boy scout rule is fine, though I will point out that's just a reversion commit of earlier dropped support for SSCTs. We might consider making SSCTs more useful, though as pointed out in the [index RFC that was just merged as I was writing this](https://github.com/openbao/openbao/pull/3052), they're not a full solution to things.

When we make SSCTs useful, we'll probably need context to interact with index manager, though. My 2c.

satoqz (dismissed):
🆗

internal/vault/request_handling.go
```diff
@@ -1051,6 +1051,17 @@ func (c *Core) handleCancelableRequest(ctx context.Context, req *logical.Request
...
 		}
+
+	// Request wrapping incurs storage (cubbyhole) writes, if the request
+	// is handled on standby node, it doesn't fail and response is not empty
+	// it will ultimately fail to save the wrapping token, and we'd still
+	// have to forward the request.
+	// Preemptively forward the requests with wrapping info provided.
+	case req.WrapInfo != nil && req.WrapInfo.TTL != 0:
+		// Read-disabled standby nodes already forwarded the request.
+		if c.HAState() == consts.PerfStandby {
```
cipherboy:
```suggestion
		if c.Standby() {
```

perhaps?

wslabosz-reply: I think it does not matter? if we were a read-disabled standby, we wouldn't get this far with request handling, but I will adjust with your suggestion

internal/vault/request_handling.go
wslabosz-reply: reverted SSCTs changes

cipherboy (approved):

cipherboy:
Thanks @wslabosz-reply!

PR #3791, opened 2026-08-13
Fix TestPostgreSQL_FatalInit

cipherboy:
## Description

Due to changes in the storage contract now calling cleanup, this test case took much longer to run and ultimately failed with the wrong error message (which, sadly, is not returned to the test due to it being logged from the cluster). We want it to fail with:

> error unsealing core: error="unseal with stored key failed:
>       self-initialization failed: refusing to unseal"

but it was instead failing with a PostgreSQL cluster wait failure since we had previously cleaned up the failed cluster node.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/external_tests/postgresql_binary/postgresql_test.go
+++ b/internal/vault/external_tests/postgresql_binary/postgresql_test.go
@@ -10,7 +10,6 @@ import (
 	"time"
 
 	log "github.com/hashicorp/go-hclog"
-	"github.com/hashicorp/go-uuid"
 	"github.com/openbao/openbao/api/v2"
 	"github.com/openbao/openbao/sdk/v2/helper/consts"
 	"github.com/openbao/openbao/sdk/v2/helper/logging"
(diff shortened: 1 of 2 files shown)


internal/vault/external_tests/postgresql_binary/postgresql_test.go
```diff
@@ -422,8 +423,9 @@ func TestPostgreSQL_Scalability(t *testing.T) {
 	}, 15*time.Second, 100*time.Millisecond)
 
 	// Eventually we should be able to write to the primary but fail to see
-	// it immediately on the read-only tertiary.
-	require.EventuallyWithT(t, func(collect *assert.CollectT) {
+	// it immediately on the read-only tertiary. This is rather race-prone
+	// and depends on replication delay.
+	/*require.EventuallyWithT(t, func(collect *assert.CollectT) {
```
cipherboy, line 428: This has been rather flaky as well, so I dropped it.

satoqz (approved):

cipherboy:
Thanks @satoqz! Merging...

PR #3792, opened 2026-08-13
Allow parallel read-standby, active lock startup

cipherboy:
## Description

By separating the concerns of read-standby execution and active lock acquisition, we don't have to delay until the read-standby setup is complete. With PostgreSQL in particular, this can be expensive: we need to wait for the local database to finish replicating whatever index the leader thinks it is at. This means we may be waiting 30s or more for lock acquisition.

## Rationale

By logic in #3774, I think we can roll this into the changelog for PSQL horizontal scalability. 

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/ha.go
+++ b/internal/vault/ha.go
@@ -496,14 +496,23 @@ func (c *Core) runHALoop(doneCh chan<- struct{}, manualStepDownCh chan struct{},
 }
 
 func (c *Core) runHALoopOnce(manualStepDownCh chan struct{}, stopCh, restartCh <-chan struct{}) bool {
+	c.logger.Trace("runHALoopOnce starting")
+	defer c.logger.Trace("runHALoopOnce completed")
+
 	restart := false
-	isReadEnabledStandby := c.StandbyReadsEnabled()
+	isReadEnabledStandby := c.MaybeStandbyReadsEnabled()
+
+	var standbyCtx atomic.Pointer[atomicContext]
+	var runStandby atomic.Bool
 
 	var g run.Group
 	{
 		// This will cause all the other actors to close when the stop channel
 		// is closed or the restartCh is triggered.
 		g.Add(func() error {
+			c.logger.Trace("runHALoopOnce: starting monitor")
+			defer c.logger.Trace("runHALoopOnce: done monitor")
+
 			select {
 			case <-stopCh:
 			case <-restartCh:
@@ -517,6 +526,9 @@ func (c *Core) runHALoopOnce(manualStepDownCh chan struct{}, stopCh, restartCh <
 		keyRotateStop := make(chan struct{})
 
 		g.Add(func() error {
+			c.logger.Trace("runHALoopOnce: starting keyring")
+			defer c.logger.Trace("runHALoopOnce: done keyring")
+
 			c.periodicCheckKeyringUpgrades(context.Background(), keyRotateStop, isReadEnabledStandby)
 			return nil
 		}, func(error) {
@@ -529,6 +541,9 @@ func (c *Core) runHALoopOnce(manualStepDownCh chan struct{}, stopCh, restartCh <
 		checkLeaderStop := make(chan struct{})
 
 		g.Add(func() error {
+			c.logger.Trace("runHALoopOnce: starting leadership refresh")
+			defer c.logger.Trace("runHALoopOnce: done leadership refresh")
+
 			c.periodicLeaderRefresh(checkLeaderStop)
 			return nil
 		}, func(error) {
@@ -540,6 +555,9 @@ func (c *Core) runHALoopOnce(manualStepDownCh chan struct{}, stopCh, restartCh <
 		metricsStop := make(chan struct{})
 
 		g.Add(func() error {
+			c.logger.Trace("runHALoopOnce: starting metrics")
+			defer c.logger.Trace("runHALoopOnce: done metrics")
+
 			c.metricsLoop(metricsStop)
 			return nil
 		}, func(error) {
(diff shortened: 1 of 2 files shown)


internal/vault/ha.go
```diff
@@ -496,14 +496,23 @@ func (c *Core) runHALoop(doneCh chan<- struct{}, manualStepDownCh chan struct{},
 }
 
 func (c *Core) runHALoopOnce(manualStepDownCh chan struct{}, stopCh, restartCh <-chan struct{}) bool {
+	c.logger.Debug("runHALoopOnce starting")
+	defer c.logger.Trace("runHALoopOnce completed")
```
satoqz, line 500: debug, trace, pick one 😄

I think generally the ones you're adding here fit trace better.

internal/vault/ha.go
```diff
@@ -570,83 +607,133 @@ func (c *Core) runHALoopOnce(manualStepDownCh chan struct{}, stopCh, restartCh <
...
-	// Cancel if we exit the loop without transitioning to active.
+	// Cancel if we exit the loop.
 	defer standbyCtxCancel()
 
-	failedLeaderAcquisition := make(chan struct{})
-	acquiredLeadership := make(chan struct{})
+	// Store this context into sharedStandbyCtx for others to use, cancelling
+	// anything that exists currently. This is so that, if we become active,
+	// we can cancel this context from the other goroutine.
+	sharedStandbyCtx.Load().Canceler()()
```
satoqz, line 670: 3/4 uses of `Canceler()` now just call it immediately, you might as well add a `Cancel()` helper.

internal/vault/ha.go
```diff
@@ -1440,3 +1554,7 @@ func (c *Core) StandbyReadsEnabled() bool {
 	}
 	return !conf.DisableStandbyReads
 }
+
+func (c *Core) MaybeStandbyReadsEnabled() bool {
+	return c.StandbyReadsEnabled() || shouldUseGRPCInvalidation(c.underlyingPhysical)
+}
```
satoqz, line 1560: why?

cipherboy: @satoqz In PostgreSQL, `StandbyReadsEnabled() = false` if the GRPC connection is not active. There are other callers of `StandbyReadsEnabled()` that expect it to be a proxy of "can we handle requests right now", but by splitting loops and @wslabosz-reply's change for `StandbyReadsEnabled()` to be static from the PoV of `runHALoopOnce`, we'll never attempt to start the PostgreSQL read-enabled replica if we didn't happen to already have it set up or have another reason to restart. 

When it comes time to start the read-only standby, we'll first attempt to establish GRPC (`LeaderLocked()`) and then choose our value appropriately for PSQL.

internal/vault/ha.go
cipherboy: Fixed to trace :D

internal/vault/ha.go
cipherboy: @satoqz Also, from today's Horizontal Scalability WG call, if we go add support for disconnected-leader read-standby (without a connected GRPC for invalidation), we'll want this as well.

internal/vault/ha.go
satoqz, line 1549: Doesn't this need to hold `requestForwardingConnectionLock`?

internal/vault/ha.go
```diff
@@ -1440,3 +1557,14 @@ func (c *Core) StandbyReadsEnabled() bool {
...
+
+// MaybeStandbyReadsEnabled is like StandbyReadsEnabled but returns true in
+// the case of a disconnected GRPC forwarding client when GRPC invalidations
+// are in use.
+//
+// In the context of runHALoop, we want runReadStandby to execute if the GRPC
+// forwarding client is not connected because it will opportunistically
+// attempt to set it up if its missing and we need it.
+func (c *Core) MaybeStandbyReadsEnabled() bool {
+	return c.StandbyReadsEnabled() || shouldUseGRPCInvalidation(c.underlyingPhysical)
```
satoqz, line 1569: I kind of hate how much overlap we have across `StandbyReadsEnabled`, `MaybeStandbyReadsEnabled` and `shouldUseGRPCInvalidation`.

For example, a single `MaybeStandbyReadsEnabled` call might type assert `underlyingPhysical` as `physical.CacheInvalidationBackend` _three times_ independently.

I tried refactoring this into smaller, more composable helpers:

```go
func isCacheInvalidationBackend(phys physical.Backend) bool {
	_, ok := phys.(physical.CacheInvalidationBackend)
	return ok
}

func isReplicationIndexBackend(phys physical.Backend) bool {
	_, ok := phys.(physical.ReplicationIndexBackend)
	return ok
}

func shouldUseGRPCInvalidation(phys physical.Backend) bool {
	return !isCacheInvalidationBackend(phys) && isReplicationIndexBackend(phys)
}

func (c *Core) standbyReadsAllowed() bool {
	if conf := c.rawConfig.Load(); conf != nil {
		return !conf.DisableStandbyReads
	}
	return false
}

func (c *Core) haveForwardingClient() bool {
	c.requestForwardingConnectionLock.RLock()
	defer c.requestForwardingConnectionLock.RUnlock()
	return c.rpcForwardingClient != nil
}

func (c *Core) shouldHookInvalidate(phys physical.Backend) bool {
	return c.standbyReadsAllowed() && isCacheInvalidationBackend(c.underlyingPhysical)
}

func (c *Core) StandbyReadsEnabled() bool {
	return c.standbyReadsAllowed() &&
		(isCacheInvalidationBackend(c.underlyingPhysical) ||
			(isReplicationIndexBackend(c.underlyingPhysical) && c.haveForwardingClient()))
}

func (c *Core) MaybeStandbyReadsEnabled() bool {
	return c.standbyReadsAllowed() &&
		(isCacheInvalidationBackend(c.underlyingPhysical) ||
			isReplicationIndexBackend(c.underlyingPhysical))
}
```

I think it is easier to reason about, though do with this what you will :D

(I slightly changed the rule for `MaybeStandbyReadsEnabled` to require that it is not disabled in the config which seemed correct to do?)

internal/vault/ha.go
```diff
@@ -1423,8 +1537,11 @@ func (c *Core) clearLeader(uuid string) error {
 	return c.barrier.Delete(context.Background(), key)
 }
 
-// StandbyReadsEnabled returns true iff standby read are enabled and supported
-// by the physical backend
+// StandbyReadsEnabled returns true iff standby read are enabled, supported,
+// and likely immediately usable by the physical backend.
+// /
```
satoqz, line 1542:
```suggestion
//
```

satoqz (commented):
More complaints on our calculation of various booleans (sorry), otherwise I recall this looking good.

internal/vault/ha.go
satoqz: Either way it would be nice to place all of these, no matter which we take, in one place in `ha.go`.

satoqz (approved):
We can defer to follow-ups though, if you'd prefer.

internal/vault/ha.go
cipherboy: @satoqz I think so, but not sure we can always grab it without risking different lock ordering given we call it all the way in `http/` as well...

Assuming I understand this correctly, we have four values protected by this lock:

```go
    // The context for the client
    rpcClientConnContext context.Context
    // The function for canceling the client connection
    rpcClientConnCancelFunc context.CancelFunc
    // The grpc ClientConn for RPC calls
    rpcClientConn *grpc.ClientConn
    // The grpc forwarding client
    rpcForwardingClient *forwarding.Client
```

I think `rpcClientConn` doesn't really need to exist independently from `*forwarding.Client` and the latter should fully consume it and we'll add its own `.Close()` function.

If we use `atomic.Pointer[atomicContext]` for `rpcClientConnContext`, we can use a second `atomic.Pointer[forwarding.Client]` perhaps and the context will default to cancelled.  We can then order `ctx -> client` on creation and `client -> ctx` on close?

Thoughts?

internal/vault/ha.go
cipherboy: @satoqz Hmm, not sure I see as much duplication as you do, but I'm happy applying the diff afterwards.

PR #3794, opened 2026-08-13
add 'show internal engines' toggle

### LINKED ISSUE #3754: Toggle visibility of "cubbyhole" Secret engine
Users can be irritated/confused by the "cubbyhole" secret engine when they first start using openbao. Without reading the documentation thouroughly or being guided by an administrator they might think it's the correct way to place persistent secrets for their applications.
We provide separate KV engines for our teams which they should use instead.

A better idea instead of hiding it permanently for everyone would be to toggle its (and maybe other internal engines) visibility in the UI. 
A good example of how this could work is Redpanda Console for managing Kafka topics. They provide a checkbox to toggle visibility of internal topics, which are irrelevant to 99% of users and therefore unchecked by default:

Image

The only real alternative for us is to guide users very closely or add a description to the KV engine saying "!! USE THIS !!" to get their attention.

All other, currently permanently hidden internal engines, could be toggleable by this feature if meaningful. Though I do not know if they could even contibute to debugging/error analysis.

**Additional context**

Luiggi33:
## Description

Adds a toggle to show/hide internal engines (right now this is only the cubbyhole engine type, could be including the other two hidden by default)

Screenshot 2026-08-13 224933

## Rationale

Resolves: #3754 

## Acknowledgements

       also allowed.

### DIFF
--- a/ui/app/templates/vault/cluster/secrets/backends.hbs
+++ b/ui/app/templates/vault/cluster/secrets/backends.hbs
@@ -33,6 +33,16 @@
       @inputValue={{if this.selectedEngineName (array this.selectedEngineName)}}
       class="is-marginless has-left-padding-s"
     />
+    <Toggle
+      @name="show-internal-engines"
+      @status="success"
+      @size="small"
+      @checked={{this.showInternalEngines}}
+      @onChange={{this.toggleShowInternalEngines}}
+      data-test-show-internal-engines-toggle
+    >
+      <span class="has-text-grey">Show internal engines</span>
+    </Toggle>
   </ToolbarFilters>
   <ToolbarActions>
     <ToolbarLink @route="vault.cluster.settings.mount-secret-backend" @type="add" data-test-enable-engine>
--- a/ui/app/templates/vault/cluster/secrets/backends.hbs
+++ b/ui/app/templates/vault/cluster/secrets/backends.hbs
@@ -33,16 +33,18 @@
       @inputValue={{if this.selectedEngineName (array this.selectedEngineName)}}
       class="is-marginless has-left-padding-s"
     />
-    <Toggle
-      @name="show-internal-engines"
-      @status="success"
-      @size="small"
-      @checked={{this.showInternalEngines}}
-      @onChange={{this.toggleShowInternalEngines}}
-      data-test-show-internal-engines-toggle
-    >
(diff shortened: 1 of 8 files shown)


satoqz:
image

I think the text is not aligned with the toggle? Also, when I click the toggle, the box around it gains a blue outline (at least in Firefox).

satoqz:
It would also be nice if the toggle persisted across refreshes.

JanMa:
Thanks @Luiggi33 , I'd appreciate a test for this new toggle

Luiggi33:
> Thanks @Luiggi33 , I'd appreciate a test for this new toggle

I guess I was not really thinking when I thought I actually added tests already... welp, now there are tests :D

ui/app/templates/vault/cluster/secrets/backends.hbs
```diff
@@ -33,6 +33,16 @@
...
+    <Toggle
+      @name="show-internal-engines"
+      @status="success"
+      @size="small"
+      @checked={{this.showInternalEngines}}
+      @onChange={{this.toggleShowInternalEngines}}
+      data-test-show-internal-engines-toggle
+    >
+      <span class="has-text-grey">Show internal engines</span>
+    </Toggle>
```
JanMa:
```suggestion
    <span class="is-marginless has-left-padding-xs">
    <Toggle
      @name="show-internal-engines"
      @status="success"
      @size="small"
      @checked={{this.showInternalEngines}}
      @onChange={{this.toggleShowInternalEngines}}
      data-test-show-internal-engines-toggle
    >
      <span class="has-text-grey">Show internal engines</span>
    </Toggle>
    </span>

```
The padding of the toggle to the filter bar doesn't match.

JanMa (changes requested):
Small remark, otherwise LGTM

JanMa:
> image
> I think the text is not aligned with the toggle? Also, when I click the toggle, the box around it gains a blue outline (at least in Firefox).

Unfortunately this is an issue with all toggles

Luiggi33:
> Unfortunately this is an issue with all toggles

I was able to find a fix for the blue box around the whole element (the box-shadow drawn when focused), it was applied to the whole label element which includes the slider (that the user wants to focus on) and the text

JanMa (approved):
Lgtm, thanks @Luiggi33

PR #3763, opened 2026-08-11
Add initial support to PKI for external keys

cipherboy:
## Description

Like earlier support for Transit, we support the external_key_ref parameter on CA (including intermediate CSR) generation. This is not relevant for leaf issuance as we don't keep leaf certificate key material within this engine (see Transit for that).

<details>

<summary> Test scripts </summary>

```
#!/bin/bash

set -euxo pipefail

( killall bao && sleep 1 ) || true

devbao node start --force --initialize --unseal --profiles secret,userpass,pki,transit --audit --seals static:// --ui

. <(devbao node env prod)

bao write transit/keys/pki-signing type=rsa-2048

bao write sys/external-keys/configs/transit plugin=transit token=$VAULT_TOKEN address=$VAULT_ADDR mount_path=transit
bao write sys/external-keys/configs/transit/keys/signing-key name=pki-signing version=1
bao write -f sys/external-keys/configs/transit/keys/signing-key/grants/pki

bao secrets enable pki

bao write pki/root/generate/kms external_key_ref=transit:signing-key common_name="Root R1" use_pss=true

bao list -detailed pki/issuers
bao read pki/issuer/default
bao read pki/key/default

bao write pki/roles/testing allow_any_name=true key_type=rsa generate_lease=true use_pss=true
bao write pki/issue/testing common_name=alex ttl=900s
```

</details>

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3797, opened 2026-08-13
Tests for external keys

cipherboy:
## Description

Based on #3763, will be rebased once that merges. 

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/external_tests/kms/testing.go
+++ b/internal/vault/external_tests/kms/testing.go
@@ -1,0 +1,441 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package kms
+
+import (
+	"encoding/base64"
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/go-uuid"
+	"github.com/openbao/openbao/api/v2"
+	"github.com/stretchr/testify/require"
+)
+
+func ExerciseTransitKMS(t *testing.T, client *api.Client) {
+	t.Run("transit", func(t *testing.T) {
+		ExerciseTransitWithTransitKMS(t, client)
+	})
+	t.Run("pki", func(t *testing.T) {
+		ExercisePKIWithTransitKMS(t, client)
+	})
+}
+
+type transitKeyType struct {
+	Type           string
+	EncryptDecrypt bool
+	SignVerify     bool
+}
+
+func knownTransitKeyTypes() []transitKeyType {
+	return []transitKeyType{
+		{
+			Type:           "aes128-gcm96",
+			EncryptDecrypt: true,
+		},
+		{
+			Type:           "aes256-gcm96",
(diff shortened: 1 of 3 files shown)


internal/vault/external_tests/kms/testing.go
satoqz, line 1: It would be nice if the majority of this was agnostic to Transit as a _backend_ and took an opaque set of external key references and their supposed capabilities so we can easily reuse with other backends later. Thoughts?

satoqz:
@cipherboy It'd also be nice to test against an external plugin here (e.g., by building an in-test Transit plugin server and shadowing the builtin) as that'd catch more bugs, for example an early `Close()`.

You should be able to construct a `kmsplugin.Catalog` here and pass that to `CoreConfig`, creating the plugin server the same way that `internal/helper/kmsplugin/catalog_test.go` does.

internal/vault/external_tests/kms/testing.go
cipherboy: @satoqz I think that's what `ExerciseTransitEncryptDecrypt`, `ExerciseTransitSignVerify`, `ExercisePKIRootCA`, and `ExercisePKIIntermediateCA` are, yes? 

Though looking at this more, `ExercisePKI` is missing the key creation logic for Transit so I'm rather curious how that works, since base name is different. :laughing:

satoqz: @cipherboy I thought we could also pull out grant creation and registering the external key with the engines.

But fine to do it this way for now.

cipherboy: Ah, the latter in the transit flow isn't broken out but it is in the PKI flow. Yeah I can extract grant creation as well. I think external key creation is scoped to the KMS-specific logic, so I figured they'd handle grant creation at that point in time, but happy to do that too.

satoqz:
Seems like we have a legit race here, will take a look.

satoqz:
Oh Alex...

satoqz:
```
71fa60481ff builtin/logical/pki/path_intermediate.go          (Steven Clark      2022-10-03 12:39:54 -0400 116)         // Remove this once https://github.com/golang/go/issues/45990 is fixed
ce7f0ff528c builtin/logical/pki/path_intermediate.go          (Alexander Scheel  2022-08-03 12:42:24 -0400 117)         data.Schema["use_pss"] = &framework.FieldSchema{
ce7f0ff528c builtin/logical/pki/path_intermediate.go          (Alexander Scheel  2022-08-03 12:42:24 -0400 118)                 Type:    framework.TypeBool,
ce7f0ff528c builtin/logical/pki/path_intermediate.go          (Alexander Scheel  2022-08-03 12:42:24 -0400 119)                 Default: false,
ce7f0ff528c builtin/logical/pki/path_intermediate.go          (Alexander Scheel  2022-08-03 12:42:24 -0400 120)         }
```

I think we can remove it :)

satoqz (approved):

PR #3768, opened 2026-08-12
Address workflows tokens' permissions

wslabosz-reply:
## Description
Updates the workflows definitions to meet the criteria of `token permissions` on [OSSF scorecard summary](https://scorecard.dev/viewer/?uri=github.com/openbao/openbao) according to the [documentation](https://github.com/ossf/scorecard/blob/c395761df6afe1a69e476bc60a013a94bcbc153f/docs/checks.md#token-permissions).

## Rationale
ensures least privilege principle in github actions workflows;
higher score increases the perceived quality

Resolves: no issue

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3800, opened 2026-08-14
Add missing job-level permissions to the workflows

wslabosz-reply:
## Description
Adds missing permissions when a more granular job-level `permissions` block was used which does not inherit top-level `permissions` properties

## Rationale
Follow up to #3768, this would surface if you tried to run the actions in a authorization required environment.

Resolves: no issue

## Acknowledgements

       also allowed.


satoqz (approved):

cipherboy (approved):
Thanks! This should fix it on private repos.

PR #3813, opened 2026-08-16
Fix busy-loop when renewing a revoked-but-renewable token (#3747)

### LINKED ISSUE #3747: Openbao proxy retries renew for tokens that are not renewable anymore in an infinite loop
**Describe the bug**
In cases when a token cached by openbao proxy was revoked, the proxy will try to renew it without considering the backoff settings infintely, this causes huge load on the leader node

**To Reproduce**
1. Create an approle, login through the proxy
2. bao token revoke -accessor=<accessor> while the proxy still has it cached (token must be renewable)
3. Watch renew-self in the audit log

**Expected behavior**
Backend will be hammered with renew-self requests

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.6.1
* OpenBao CLI Version (retrieve with `bao version`): 2.6.1

chiliec:
## Description

Fixes #3747.

When the auto-auth / proxy `LifetimeWatcher` renews a cached token whose `renew-self` keeps failing — e.g. the token was revoked out from under the proxy but is still marked `renewable` — it entered a tight retry loop that hammered the leader node with `renew-self` requests (the load described in the issue).

### Root cause

In `api/lifetime_watcher.go`, the error-backoff branch evaluated `NextBackOff()` only to compare it against `backoff.Stop`, and **discarded the returned interval**:

```go
if errorBackoff == nil {
    sleepDuration = r.calculateSleepDuration(...)
} else if errorBackoff.NextBackOff() == backoff.Stop || remainingLeaseDuration < 0 {
    return err
}
```

On a renewal error under the default `RenewBehaviorIgnoreErrors`, `sleepDuration` was left at its zero value, so `time.NewTimer(0)` fired immediately and the loop retried `renew` as fast as the CPU allowed — no backoff at all.

### Fix

Capture the interval `NextBackOff()` returns and use it as the sleep duration, still bailing out when it reports `Stop` or the original lease has fully expired:

```go
} else {
    nextBackoff := errorBackoff.NextBackOff()
    if nextBackoff == backoff.Stop || remainingLeaseDuration < 0 {
        return err
    }
    sleepDuration = nextBackoff
}
```

## Test

Adds `TestLifetimeWatcher_RenewErrorsBackOff` (`api/renewer_test.go`): a renew func that always errors, asserting the number of attempts in a ~1 s window stays small (backoff honored) rather than spinning.

**RED→GREEN, verified locally (Go 1.26.0):**
- Against the pre-fix code the test observed **1,798,812 renew attempts in ~1 s** — the busy loop, reproduced.
- With the fix it drops to a handful, and the test passes.

```
$ go test -run 'TestLifetimeWatcher|TestRenewer|TestCalcSleepPeriod' ./
ok  github.com/openbao/openbao/api/v2
```

`go build ./...`, `gofmt`, and the existing renewer suite all stay green. Only `api/lifetime_watcher.go` and `api/renewer_test.go` change. (The pre-existing `go vet` lock-copy notices in `client.go` are untouched and outside this diff.)

First-time contributor here — happy to adjust the backoff bounds in the test or the wording. I'll add the `changelog/<pr>.txt` entry once this PR has a number.

### DIFF
--- a/api/lifetime_watcher.go
+++ b/api/lifetime_watcher.go
@@ -346,8 +346,16 @@ func (r *LifetimeWatcher) doRenewWithOptions(tokenMode bool, nonRenewable bool, 
 
 		if errorBackoff == nil {
 			sleepDuration = r.calculateSleepDuration(remainingLeaseDuration, priorDuration)
-		} else if errorBackoff.NextBackOff() == backoff.Stop || remainingLeaseDuration < 0 {
-			return err
+		} else {
+			// A renewal error is in flight: sleep for the backoff interval
+			// rather than busy-looping. NextBackOff both advances the backoff
+			// state and returns the next interval; if it reports Stop (or the
+			// original lease has fully expired) give up and surface the error.
+			nextBackoff := errorBackoff.NextBackOff()
+			if nextBackoff == backoff.Stop || remainingLeaseDuration < 0 {
+				return err
+			}
+			sleepDuration = nextBackoff
 		}
 
 		// remainingLeaseDuration becomes the priorDuration for the next loop
--- a/api/renewer_test.go
+++ b/api/renewer_test.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"math/rand"
 	"reflect"
+	"sync/atomic"
 	"testing"
 	"testing/quick"
 	"time"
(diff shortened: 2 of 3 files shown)


chiliec:
Hi @satoqz — I noticed this was closed without a comment, and #3747 is still open (labeled `good first issue`/`help wanted`, and @cipherboy noted PRs are welcome). Was the close intentional, or is there a different approach you'd prefer?

The change is small and self-contained: in `api/lifetime_watcher.go` the error-backoff branch computed `NextBackOff()` only to compare against `backoff.Stop` but discarded the returned duration, so a token whose `renew-self` keeps failing (e.g. revoked-but-still-renewable) retried in a tight loop. The PR sleeps for that duration instead. DCO and CI are green (+72/−2). Happy to rebase or adjust the approach if that helps — just let me know.

wslabosz-reply:
@chiliec 
https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md

PR #3825, opened 2026-08-18
auth/jwt: allow multiple JWKS URLs

### LINKED ISSUE #3761: Support `jwks_pairs` in OpenBao JWT auth backend
## Summary

Vault  introduced `jwks_pairs` for the JWT auth backend, allowing a single auth mount to be configured with multiple JWKS endpoints. During authentication, the JWT signature is verified against each configured JWKS until a matching key is found.

- [https://developer.hashicorp.com/vault/docs/auth/jwt](https://developer.hashicorp.com/vault/docs/auth/jwt)
- [https://docs.hashicorp.com/vault/api-docs/auth/jwt#jwks_pairs](https://docs.hashicorp.com/vault/api-docs/auth/jwt#jwks_pairs)

## Use case

We authenticate workloads from multiple Kubernetes clusters using the JWT auth backend. Each cluster exposes its own JWKS endpoint, but the roles and policies are otherwise identical.

Supporting `jwks_pairs` would allow a single JWT auth mount and role to trust all of these clusters, significantly simplifying configuration and reducing operational overhead compared to maintaining multiple auth mounts.

## Implementation

The implementation effort should be relatively low. The `jwks_pairs` feature was added to Vault while the JWT auth plugin was still licensed under the Mozilla Public License (MPL), so the implementation should be available for OpenBao to backport.

YuukiRitoTeng:
## Description

The JWT auth method now supports multiple JWKS endpoints through the existing `jwks_url` configuration parameter.

`jwks_url` now uses `TypeCommaStringSlice`, so existing single-string configurations continue to work, while comma-separated strings and string arrays can be used to configure multiple endpoints.

During login, the configured JWKS endpoints are checked in order until one contains a key that verifies the JWT signature.

Internally, the configuration represents the URLs using `JWKSURLs`. Existing stored configurations containing a single `jwks_url` remain compatible and are handled as a list internally.

Configuration reads return `jwks_url` as a string array, including when only one endpoint is configured.

The existing `jwks_ca_pem` setting remains a shared CA bundle for all configured JWKS endpoints and continues to support concatenated PEM certificates.

## Rationale

Some deployments need a single JWT auth mount to validate tokens from multiple Kubernetes clusters or other JWKS providers.

Allowing multiple values through the existing `jwks_url` parameter avoids adding another public configuration field while preserving compatibility with existing single-URL configurations.

This also covers the Vault `jwks_pairs` migration use case where multiple JWKS endpoints need to be trusted from one auth mount. OpenBao continues to use a shared `jwks_ca_pem` bundle rather than configuring a separate CA for each endpoint.

Resolves: #3761

## Testing

* Single-string `jwks_url`
* Comma-separated `jwks_url`
* String-array `jwks_url`
* Configuration reads returning `jwks_url` as an array
* Multiple JWKS endpoints with order preservation
* Login succeeding when a later JWKS endpoint contains the matching key
* Login failing when no configured JWKS endpoint matches
* `skip_jwks_validation`
* Legacy single-string storage compatibility

```text
go test ./internal/builtin/credential/jwt -run 'TestConfig_|TestLogin_MultiJWKSURL' -count=1
go vet ./internal/builtin/credential/jwt
```

The full JWT package test still has the existing Windows-specific failure in `TestOIDC_Callback/no_response_from_provider`, caused by the platform-specific connection-refused error wording.

## Acknowledgements

  also allowed.

### DIFF
--- a/internal/builtin/credential/jwt/path_config.go
+++ b/internal/builtin/credential/jwt/path_config.go
@@ -86,7 +86,7 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 			},
 			"jwks_ca_pem": {
 				Type:        framework.TypeString,
-				Description: "The CA certificate or chain of certificates, in PEM format, to use to validate connections to the JWKS URLs. If not set, system certificates are used.",
+				Description: "The CA certificate or chain of certificates, in PEM format, to use to validate connections to the JWKS URLs. May contain multiple concatenated CA certificates. The same CA certificates are used for all jwks_url values. If not set, system certificates are used.",
 			},
 			"default_role": {
 				Type:        framework.TypeLowerCaseString,
@@ -434,7 +434,7 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 
 			if !strings.Contains(err.Error(), "failed to verify id token signature") {
 				if !skipJwksValidation {
-					b.Logger().Error("error checking jwks URL", "error", err, "jwks_url_index", i)
+					b.Logger().Error("error checking jwks URL", "jwks_url_index", i)
 					return logical.ErrorResponse(fmt.Sprintf("error checking jwks URL for jwks_url[%d]", i)), nil
 				}
 
--- a/internal/builtin/credential/jwt/path_config.go
+++ b/internal/builtin/credential/jwt/path_config.go
@@ -4,7 +4,6 @@
 package jwtauth
 
 import (
-	"bytes"
 	"context"
 	"crypto"
 	"crypto/ecdsa"
@@ -55,7 +54,7 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 		Fields: map[string]*framework.FieldSchema{
 			"oidc_discovery_url": {
 				Type:        framework.TypeString,
-				Description: `The base URL of the OIDC Discovery URL without ".well-known/openid-configuration" component. Cannot be used with "jwks_url" or "jwt_validation_pubkeys".`,
+				Description: `The base URL of the OIDC Discovery URL without ".well-known/openid-configuration" component. Cannot be used with "jwks_url", "jwks_urls", or "jwt_validation_pubkeys".`,
 			},
 			"oidc_discovery_ca_pem": {
 				Type:        framework.TypeString,
@@ -81,12 +80,17 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 				Description: "The response types to request. Allowed values are 'code' and 'id_token'. Defaults to 'code'.",
 			},
 			"jwks_url": {
+				Type:        framework.TypeString,
+				Description: `Deprecated: Please use "jwks_urls" instead. The single JWKS URL to use to authenticate signatures. Cannot be used with "oidc_discovery_url" or "jwt_validation_pubkeys".`,
+				Deprecated:  true,
+			},
+			"jwks_urls": {
 				Type:        framework.TypeCommaStringSlice,
-				Description: `JWKS URLs to use to authenticate signatures. A single URL may be specified as a string, or multiple URLs may be specified as a list. Cannot be used with "oidc_discovery_url" or "jwt_validation_pubkeys".`,
+				Description: `JWKS URLs to use to authenticate signatures. During login, the JWT signature is verified against each JWKS in the order configured until a match is found. Cannot be used with "oidc_discovery_url" or "jwt_validation_pubkeys".`,
 			},
 			"jwks_ca_pem": {
 				Type:        framework.TypeString,
-				Description: "The CA certificate or chain of certificates, in PEM format, to use to validate connections to the JWKS URLs. May contain multiple concatenated CA certificates. The same CA certificates are used for all jwks_url values. If not set, system certificates are used.",
+				Description: "The CA certificate or chain of certificates, in PEM format, to use to validate connections to the JWKS URLs. May contain multiple concatenated CA certificates. The same CA certificates are used for all jwks_urls values. If not set, system certificates are used.",
 			},
 			"default_role": {
 				Type:        framework.TypeLowerCaseString,
@@ -94,7 +98,7 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 			},
 			"jwt_validation_pubkeys": {
 				Type:        framework.TypeCommaStringSlice,
-				Description: `A list of PEM-encoded public keys to use to authenticate signatures locally. Cannot be used with "jwks_url" or "oidc_discovery_url".`,
+				Description: `A list of PEM-encoded public keys to use to authenticate signatures locally. Cannot be used with "jwks_url", "jwks_urls", or "oidc_discovery_url".`,
 			},
 			"jwt_supported_algs": {
 				Type:        framework.TypeCommaStringSlice,
@@ -125,7 +129,7 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 			},
 			"skip_jwks_validation": {
 				Type:        framework.TypeBool,
-				Description: "When true and oidc_discovery_url or jwks_url are specified, if the connection fails to load, a warning will be issued and status can be checked later by reading the config endpoint.",
+				Description: "When true and oidc_discovery_url, jwks_url, or jwks_urls are specified, if the connection fails to load, a warning will be issued and status can be checked later by reading the config endpoint.",
 			},
 		},
 
@@ -172,6 +176,15 @@ func (b *jwtAuthBackend) config(ctx context.Context, s logical.Storage) (*jwtCon
 	config := &jwtConfig{}
 	if err := entry.DecodeJSON(config); err != nil {
 		return nil, err
+	}
+
+	// Migrate the legacy single URL in memory. The next config write stores the
+	// canonical jwks_urls field; reads do not write back to storage.
+	if config.DeprecatedJWKSURL != "" {
+		if len(config.JWKSURLs) == 0 {
+			config.JWKSURLs = []string{config.DeprecatedJWKSURL}
+		}
+		config.DeprecatedJWKSURL = ""
 	}
 
 	for _, v := range config.JWTValidationPubKeys {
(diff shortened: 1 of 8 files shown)


website/content/docs/api/auth/jwt.mdx
```diff
@@ -34,16 +34,17 @@ set.
...
-- `jwks_url` `(string: <optional>)` - JWKS URL to use to authenticate signatures. Cannot be used with "oidc_discovery_url" or "jwt_validation_pubkeys".
-- `jwks_ca_pem` `(string: <optional>)` - The contents of a CA certificate or chain of certificates, in PEM format, to use to validate connections to the JWKS URL. If not set, system certificates are used.
+- `jwks_url` `(string: <optional>)` - JWKS URL to use to authenticate signatures. Cannot be used with "oidc_discovery_url", "jwks_pairs", or "jwt_validation_pubkeys".
+- `jwks_ca_pem` `(string: <optional>)` - The contents of a CA certificate or chain of certificates, in PEM format, to use to validate connections to the JWKS URL. If not set, system certificates are used. Cannot be used with "jwks_pairs".
+- `jwks_pairs` `(list of JSON object: <optional>)` - List of JWKS URL and optional CA certificate pairs. CA certificates must be in PEM format. Must be a list of JSON objects with format `[{"jwks_url": "", "jwks_ca_pem": ""}]`. During login, the JWT signature is verified against each JWKS in the order configured until a match is found. Cannot be used with "jwks_url" or "jwks_ca_pem".
```
cipherboy: Neither here nor there, but I am entertained by the data format. Why not just take a list of `jwks_url` and separately a list of `jwks_ca_pem`? The indices don't need to match; `jwt.NewJSONWebKeySet` should handle multiple CA certs just fine and passing more than one CA cert to a given URL is probably fine (as, you wouldn't really trust a compromised CA for one JWT issuer and not the other in the same mount -- if you didn't want to, use two separate mounts).

That way `jwks_url` can be a `CommaStringSlice` (which takes single strings as well) and users can transparently upgrade.

My 2c.

YuukiRitoTeng: I pretty much followed #3761 as written. Since it specifically called out jwks_pairs and linked to the existing docs, I assumed that was the interface we wanted rather than making jwks_url accept multiple values. Thanks for the suggestion — if you think that approach makes more sense here, I’ll do my best to adjust the implementation.

YuukiRitoTeng: I went ahead and changed it to allow multiple URLs through jwks_url. The old single-string form still works, and jwks_ca_pem is shared across the URLs. I also added tests for the fallback path and for loading older configs. Could you take another look when you have a chance?

DrDaveD:
The original issue suggested importing some pre-BUSL code from Hashicorp.   However, the description has the box checked that includes confirming that any imported code would be subject to attribution.  I don't see any attribution, so @YuukiRitoTeng does that indeed mean that you did not reuse any MPL-licensed Hashicorp code, you re-wrote it from scratch?

website/content/docs/api/auth/jwt.mdx
DrDaveD: I agree that it's easier to use this way than `jwks_pairs`.  I don't think we can assume they all use the same CA cert though; it would be better for that to also be able to be a list, although comma-separated makes no sense in that case.  That would need to be an array of strings.

cipherboy: @DrDaveD I don't think there's any need to change `jwks_ca_pem`; it already supports multiple (simple) concatenated CAs. Check out:

https://github.com/hashicorp/cap/blob/cf52420329e3b328e678e3157dc2000b4ac0f90e/jwt/keyset.go#L107-L124

to 

https://github.com/hashicorp/cap/blob/cf52420329e3b328e678e3157dc2000b4ac0f90e/jwt/keyset.go#L204-L227

to

https://pkg.go.dev/crypto/x509#CertPool.AppendCertsFromPEM

DrDaveD: Oh, ok then.  Maybe that should just then be in the documentation.

YuukiRitoTeng:
Yes, that's correct.

I did not copy, port, or reuse Vault's `jwks_pairs` implementation.

I implemented this on top of OpenBao's existing JWT backend, using the CAP APIs already used by the project.

internal/builtin/credential/jwt/path_config.go
```diff
@@ -618,7 +630,7 @@ type jwtConfig struct {
 	OIDCClientSecret           string         `json:"oidc_client_secret"`
 	OIDCResponseMode           string         `json:"oidc_response_mode"`
 	OIDCResponseTypes          []string       `json:"oidc_response_types"`
-	JWKSURL                    string         `json:"jwks_url"`
+	JWKSURL                    []string       `json:"jwks_url"`
```
cipherboy: @YuukiRitoTeng I think the correct pattern in most cases is to add a `DeprecatedJWKSURL` parameter name the new one something different (`jwks_urls`).

cipherboy: (That saves you from having to implement the custom UnmarshalJSON parameter).

You'd then always proactively move `DeprecatedJWKSURL -> JWKSURLs` on read (clearing its existing value) and would automatically get upgraded on next write. See e.g., getRole in PKI. There's some discussion on if updates should be persisted by default or not, dealer's choice if there isn't an existing framework for that here.

YuukiRitoTeng: Thanks, this was helpful. I moved to the deprecated jwks_url / new jwks_urls pattern and dropped the custom UnmarshalJSON.

internal/builtin/credential/jwt/provider_kubernetes.go
```diff
@@ -45,8 +45,8 @@ func (k *KubernetesProvider) Initialize(_ context.Context, jc *jwtConfig) error
 		return errors.New("oidc_discovery_ca_pem must not be set when using the kubernetes provider")
 	}
 
-	if jc.JWKSURL != "" {
-		return errors.New("jwks_url must not be set when using the kubernetes provider")
+	if len(jc.JWKSURLs) != 0 {
```
wslabosz-reply, line 48: don't we also need to validate the missing presence of deprecated single entry jwksURL?

cipherboy: @wslabosz-reply With @YuukiRitoTeng's change to follow the suggested pattern in:

https://github.com/openbao/openbao/pull/3825/changes#diff-fdd85b23a091aa16eaa10c8c7e090c33e0710e22db139e80933544ed15a97961R181

```go
	// Migrate the legacy single URL in memory. The next config write stores the
	// canonical jwks_urls field; reads do not write back to storage.
	if config.DeprecatedJWKSURL != "" {
		if len(config.JWKSURLs) == 0 {
			config.JWKSURLs = []string{config.DeprecatedJWKSURL}
		}
		config.DeprecatedJWKSURL = ""
	}
```

there should never really be a `config.DeprecatedJWKSURL` value any more.

wslabosz-reply: @cipherboy okay, that makes sense, thanks!

internal/builtin/credential/jwt/path_config.go
```diff
@@ -76,19 +81,24 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 			},
 			"jwks_url": {
 				Type:        framework.TypeString,
-				Description: `JWKS URL to use to authenticate signatures. Cannot be used with "oidc_discovery_url" or "jwt_validation_pubkeys".`,
+				Description: `Deprecated: Please use "jwks_urls" instead. The single JWKS URL to use to authenticate signatures. Cannot be used with "oidc_discovery_url" or "jwt_validation_pubkeys".`,
+				Deprecated:  true,
+			},
+			"jwks_urls": {
+				Type:        framework.TypeCommaStringSlice,
+				Description: `JWKS URLs to use to authenticate signatures. During login, the JWT signature is verified against each JWKS in the order configured until a match is found. Cannot be used with "oidc_discovery_url" or "jwt_validation_pubkeys".`,
```
cipherboy: @YuukiRitoTeng Ahhh sorry for the confusion!  You'd only use `jwks_urls` on the `jwtConfig`, to allow in-place upgrading the config itself from the deprecated field value to the new value. You'd leave the API level alone. So you'd just change the existing `jwks_url` parameter's type to `framework.TypeCommaStringSlice` and it should behave backwards compatibly (since a single string without commas will be upgraded to a string slice transparently).

PKI already uses `CommaStringSlice` for fields like `issuing_certificates`, `crl_distribution_points`, `delta_crl_distribution_points`, and `ocsp_servers`.

YuukiRitoTeng: Got it, thanks. One quick question before I update this: for GET /auth/jwt/config, if there’s only one jwks_url, should it still return a string, or should it return []string now?

YuukiRitoTeng: Got it, thanks. One quick question before I update this: for GET /auth/jwt/config, if there’s only one jwks_url, should it still return a string, or should it return []string now?
Also, I noticed one other thing: with multiple `jwks_url` values, login can end up doing several sequential JWKS fetches because the validator/keysets are rebuilt for each request.

weichuliu:
@YuukiRitoTeng Really appreciate for the awesome work!

In our case, we encountered this missing feature when migrating from Vault to OpenBao. And I believe it's not a minor case that OpenBao's users are previously using Vault.

I think it will be definitely helpful if the doc explicitly call out the feature parity between `jwks_urls` and the Vault's `jwks_pairs` flag -- that will definitely help users smooth their migration.

internal/builtin/credential/jwt/path_config.go
cipherboy: @YuukiRitoTeng Apologies, missed the notification here. I think returning `[]string` is fine. 

>  Also, I noticed one other thing: with multiple `jwks_url` values, login can end up doing several sequential JWKS fetches because the validator/keysets are rebuilt for each request.

Hmm, we have caching already, am I missing where this changes here? 

IIUC, you don't remove the caching at the top of `jwtValidator(...)`:

```go
	if b.validator != nil {
		return b.validator, nil
	}
```

so while an initial login could do several sequential fetches, that only occurs once and is then reused throughout the lifetime of that validator.

PR #3837, opened 2026-08-18
Add script for revising What's Changed sections

cipherboy:
## Acknowledgements

       also allowed.

### DIFF
--- a/scripts/whats-changed.sh
+++ b/scripts/whats-changed.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/bash
+
+##
+## Usage: echo ...github-autogen-whats-changed... | ./whats-changed.sh
+##
+## Then validate each "bad format" line and adjust as necessary.
+##
+
+GITHUB_REPO_API="${GITHUB_REPO_API:-https://api.github.com/repos/openbao/openbao/pulls/}"
+
+while IFS= read -r line; do
+	if	! grep -q '^\*' <<< "$line"; then
+		echo "$line <!-- skip -->"
+		continue
+	elif ! grep -q '(#[0-9]*) by @' <<< "$line"; then
+		echo "$line <!-- bad format -->"
+		continue
+	fi
+
+	pr_number="$(grep -o '(#[0-9]*) by @' <<< "$line" | grep -o '[0-9]*')"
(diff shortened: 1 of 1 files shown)


JanMa (dismissed):

scripts/whats-changed.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/bash
```
satoqz:
```suggestion
#!/usr/bin/env bash
```

I insist.

satoqz:
Long-term we can probably just have CI create a pretty good first draft of the entire release description.

satoqz:
The main value of the section is (IMO) tagging contributors? For v2.6 GA we couldn't even use the complete section because it exceeded GitHub's size limit. So perhaps we should include a list of contributor handles and then link to the list of commits only.

Edit: Seems I misremember regarding the size limit, but I'm sure we've discussed something of that sort.

Edit 2: Ah, here it is: [#openssf-openbao-wg > 2.6.0 GA Release @ 💬](https://linuxfoundation.zulipchat.com/#narrow/channel/574533-openssf-openbao-wg/topic/2.2E6.2E0.20GA.20Release/near/610625218)

cipherboy:
> The main value of the section is (IMO) tagging contributors?

Right, otherwise it looks like only a few people contribute to non-beta releases (as they're the only ones truly cut from main) due to the lack of correct attribution of backports. 

TBH, I don't know the backports themselves really are worthy of attribution, but :shrug: 

(re: thread) I think only having delta from beta...ga is fine in general, 

---

And yes, we can likely automate generation of the release notes completely in the future, by piping https://docs.github.com/en/rest/releases/releases?apiVersion=2026-03-10#generate-release-notes-content-for-a-release into this script?  

Though we'll still need the changelog generator.

satoqz:
> Though we'll still need the changelog generator.

The changelog is committed into `CHANGELOG.md` at release workflow time so it could just pick up the latest section from there and paste it for us (+/- updating heading levels).

satoqz (approved):

cipherboy:
@satoqz Ah, fair. We could also perhaps split our changelogs, it wouldn't be the end of the world.

PR #1172, opened 2025-03-28
(secrets/pki): add not_before_bound and not_after_bound

Nerkho:
This is related to #489 which was on my table for quite some time now. 

Feedback is welcome :dancers: 

This adds two parameters to pki roles.

`not_after_bound` can take the following values :
* `forbid` to prevent users from supplying `not_after`
* `tll-limited` to allow supplying `not_after` but within the bounds of the TTL
* an explicit timestamp in UTC format `YYYY-MM-ddTHH:MM:SSZ` up to which `not_after` can be set

`not_before_bound` can take the following values :
* `forbid` to prevent users from supplying `not_after`
* `permit` allow users from supplying `not_after

PR #3839, opened 2026-08-18
Support index headers for consistency in server, api

cipherboy:
## Rationale

Related: #3052

## Acknowledgements

       also allowed.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -130,6 +130,31 @@ const (
 	// dash) is ignored. Any repeated header keys result in request failure.
 	InlineAuthParameterHeaderPrefix = "X-Vault-Inline-Auth-Parameter-"
 
+	// When sent as a request header, this denotes the minimum index value
+	// the client thinks is required to process this request. Exact OpenBao
+	// server behavior depends on configuration and the X-Vault-Inconsistent
+	// header. Invalid index headers are usually processed locally.
+	//
+	// Sent as a response header by OpenBao when a write occurs, to indicate
+	// the current state.
+	IndexHeaderName = "X-Vault-Index"
+
+	// When sent as a request header, this specifies a behavior for index
+	// processing. When "await-state" is specified as the first value, this
+	// takes another value as fallback. Defaults to "fail" on empty value.
+	IndexInconsistentHeaderName = "X-Vault-Inconsistent"
+
+	// A 429 is sent on stale index, indicating the client needs to try again
+	// after a back-off period.
+	IndexInconsistentFail = "fail"
+
+	// The request will be forwarded to the active node on stale index header.
+	IndexInconsistentForward = "forward-active-node"
+
+	// The request will be awaited on the server before a fallback behavior is
+	// performed (either fail or forwarding).
+	IndexInconsistentAwait = "await-state"
+
 	TLSErrorString = "This error usually means that the server is running with TLS disabled\n" +
 		"but the client is configured to use TLS. Please either enable TLS\n" +
 		"on the server or run the client with -address set to an address\n" +
@@ -263,6 +288,14 @@ type Config struct {
 	// DisableEnvironment, when set to true, will skip automatic configuration
 	// based on well-known environment variables when the client is constructed.
 	DisableEnvironment bool
+
+	// DefaultStrongConsistency, when set to true, at initial client creation
+	// time, will configure the client with the current default best practices
+	// for being transparently replication aware and highly consistent.
+	//
+	// The behavior of this option may shift over time as different semantics
+	// are introduced.
+	DefaultStrongConsistency bool
 
 	clientTLSConfig *tls.Config
 }
@@ -707,6 +740,14 @@ type Client struct {
 	policyOverride     bool
 	requestCallbacks   []RequestCallback
 	responseCallbacks  []ResponseCallback
+
+	// Policy to use for X-Vault-Inconsistent; exact behavior of OpenBao
+	// depends on listener configuration and the value of this option.
+	inconsistent []string
+
+	// state is an index manager shared across all client implementations;
+	// it holds the latest seen index from a given cluster.
+	state IndexManager
 }
 
 // NewClient returns a new client for the given configuration.
@@ -783,6 +824,11 @@ func NewClient(c *Config) (*Client, error) {
 		if namespace := ReadBaoVariable(EnvVaultNamespace); namespace != "" {
 			client.setNamespace(namespace)
 		}
+	}
+
+	if c.DefaultStrongConsistency {
+		client.setInconsistent([]string{IndexInconsistentAwait, IndexInconsistentForward})
+		client.setIndexManager(NewSimpleIndexManager())
 	}
 
 	return client, nil
@@ -1309,6 +1355,9 @@ func (c *Client) clone(cloneHeaders bool) (*Client, error) {
 		client.SetToken(c.token)
 	}
 
+	client.SetInconsistent(c.inconsistent)
+	client.SetIndexManager(c.state)
+
 	return client, nil
 }
 
@@ -1334,6 +1383,13 @@ func (c *Client) NewRequest(method, requestPath string) *Request {
 	wrappingLookupFunc := c.wrappingLookupFunc
 	policyOverride := c.policyOverride
 	disableEnvironment := c.config.DisableEnvironment
+	inconsistent := c.inconsistent
+	ns := c.headers.Get(NamespaceHeaderName)
+
+	var index string
+	if c.state != nil {
+		index = c.state.Get(addr.String(), ns)
+	}
 
 	c.config.modifyLock.RUnlock()
 	c.modifyLock.RUnlock()
@@ -1384,6 +1440,20 @@ func (c *Client) NewRequest(method, requestPath string) *Request {
 
 	req.Headers = c.Headers()
 	req.PolicyOverride = policyOverride
+
+	if index != "" || len(inconsistent) > 0 {
+		if req.Headers == nil {
+			req.Headers = http.Header{}
+		}
+
+		if index != "" {
+			req.Headers.Set(IndexHeaderName, index)
+		}
+
+		for _, value := range inconsistent {
+			req.Headers.Add(IndexInconsistentHeaderName, value)
+		}
+	}
 
 	return req
 }
@@ -1552,6 +1622,9 @@ START:
 		for _, cb := range c.responseCallbacks {
 			cb(result)
 		}
+
+		// Check for index headers.
+		c.setIndexFromResult(r, result)
 	}
 	if err := result.Error(); err != nil {
 		return result, err
--- a/api/client.go
+++ b/api/client.go
@@ -130,6 +130,31 @@ const (
 	// dash) is ignored. Any repeated header keys result in request failure.
 	InlineAuthParameterHeaderPrefix = "X-Vault-Inline-Auth-Parameter-"
 
+	// When sent as a request header, this denotes the minimum index value
+	// the client thinks is required to process this request. Exact OpenBao
+	// server behavior depends on configuration and the X-Vault-Inconsistent
+	// header. Invalid index headers are usually processed locally.
+	//
+	// Sent as a response header by OpenBao when a write occurs, to indicate
+	// the current state.
+	IndexHeaderName = "X-Vault-Index"
+
+	// When sent as a request header, this specifies a behavior for index
+	// processing. When "await-state" is specified as the first value, this
+	// takes another value as fallback. Defaults to "fail" on empty value.
+	IndexInconsistentHeaderName = "X-Vault-Inconsistent"
+
+	// A 429 is sent on stale index, indicating the client needs to try again
+	// after a back-off period.
+	IndexInconsistentFail = "fail"
+
+	// The request will be forwarded to the active node on stale index header.
+	IndexInconsistentForward = "forward-active-node"
+
+	// The request will be awaited on the server before a fallback behavior is
+	// performed (either fail or forwarding).
+	IndexInconsistentAwait = "await-state"
+
 	TLSErrorString = "This error usually means that the server is running with TLS disabled\n" +
 		"but the client is configured to use TLS. Please either enable TLS\n" +
 		"on the server or run the client with -address set to an address\n" +
@@ -263,6 +288,14 @@ type Config struct {
 	// DisableEnvironment, when set to true, will skip automatic configuration
 	// based on well-known environment variables when the client is constructed.
 	DisableEnvironment bool
+
+	// DefaultStrongConsistency, when set to true, at initial client creation
+	// time, will configure the client with the current default best practices
+	// for being transparently replication aware and highly consistent.
+	//
+	// The behavior of this option may shift over time as different semantics
+	// are introduced.
+	DefaultStrongConsistency bool
 
 	clientTLSConfig *tls.Config
 }
@@ -707,6 +740,14 @@ type Client struct {
 	policyOverride     bool
 	requestCallbacks   []RequestCallback
 	responseCallbacks  []ResponseCallback
+
+	// Policy to use for X-Vault-Inconsistent; exact behavior of OpenBao
+	// depends on listener configuration and the value of this option.
+	inconsistent []string
+
+	// indexTracker is shared across all client implementations;
+	// it holds the latest seen index from a given cluster.
+	indexTracker *simpleStorageIndexTracker
 }
 
 // NewClient returns a new client for the given configuration.
@@ -783,6 +824,11 @@ func NewClient(c *Config) (*Client, error) {
 		if namespace := ReadBaoVariable(EnvVaultNamespace); namespace != "" {
 			client.setNamespace(namespace)
 		}
+	}
+
+	if c.DefaultStrongConsistency {
+		client.setInconsistent([]string{IndexInconsistentAwait, IndexInconsistentForward})
+		client.setStorageIndexTracker(newSimpleStorageIndexTracker())
 	}
 
 	return client, nil
@@ -1309,6 +1355,9 @@ func (c *Client) clone(cloneHeaders bool) (*Client, error) {
 		client.SetToken(c.token)
 	}
 
+	client.setInconsistent(c.inconsistent)
+	client.setStorageIndexTracker(c.indexTracker)
+
 	return client, nil
 }
 
@@ -1334,6 +1383,12 @@ func (c *Client) NewRequest(method, requestPath string) *Request {
 	wrappingLookupFunc := c.wrappingLookupFunc
 	policyOverride := c.policyOverride
 	disableEnvironment := c.config.DisableEnvironment
+	inconsistent := c.inconsistent
+
+	var index string
+	if c.indexTracker != nil {
+		index = c.indexTracker.get()
+	}
 
 	c.config.modifyLock.RUnlock()
 	c.modifyLock.RUnlock()
@@ -1384,6 +1439,20 @@ func (c *Client) NewRequest(method, requestPath string) *Request {
 
 	req.Headers = c.Headers()
 	req.PolicyOverride = policyOverride
+
+	if index != "" || len(inconsistent) > 0 {
+		if req.Headers == nil {
+			req.Headers = http.Header{}
+		}
+
+		if index != "" {
+			req.Headers.Set(IndexHeaderName, index)
+		}
+
+		for _, value := range inconsistent {
+			req.Headers.Add(IndexInconsistentHeaderName, value)
+		}
+	}
 
 	return req
 }
@@ -1551,6 +1620,11 @@ START:
 	if result != nil {
 		for _, cb := range c.responseCallbacks {
 			cb(result)
+		}
+
+		// Check for index headers.
+		if err := c.setIndexFromResult(r, result); err != nil {
+			return result, fmt.Errorf("error parsing %v response headers: %w", IndexHeaderName, err)
 		}
 	}
 	if err := result.Error(); err != nil {
--- a/api/index.go
+++ b/api/index.go
@@ -1,0 +1,134 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package api
+
+import "sync"
+
+// SetInconsistent provides one or more values for the X-Vault-Inconsistent
+// header. See notes around Index* constants in client.go.
+func (c *Client) SetInconsistent(behaviors []string) {
+	c.modifyLock.Lock()
+	defer c.modifyLock.Unlock()
+
+	c.setInconsistent(behaviors)
+}
+
+func (c *Client) setInconsistent(behaviors []string) {
+	c.inconsistent = behaviors
+}
+
+// FailInconsistent sets X-Vault-Inconsistent to fail, ensuring a 429-based
+// backoff-retry logic for stale index headers.
+func (c *Client) FailInconsistent() {
+	c.SetInconsistent([]string{IndexInconsistentFail})
+}
+
+// ForwardInconsistent sets X-Vault-Inconsistent to forward-active-node,
+// ensuring stale index header values are forwarded instead.
+func (c *Client) ForwardInconsistent() {
+	c.SetInconsistent([]string{IndexInconsistentForward})
+}
+
+// AwaitInconsistent sets the X-Vault-Inconsistent to await-state, specifying
+// an explicit fallback behavior if waiting is not sufficient.
+func (c *Client) AwaitInconsistent(fallback string) {
+	if fallback == "" {
+		c.DefaultAwaitInconsistent()
+		return
+	}
+
+	c.SetInconsistent([]string{IndexInconsistentAwait, fallback})
+}
+
+// AwaitInconsistent sets the X-Vault-Inconsistent to await-state, relying
+// on the server's default behavior if waiting is not sufficient.
+func (c *Client) DefaultAwaitInconsistent() {
+	c.SetInconsistent([]string{IndexInconsistentAwait})
+}
+
+// SetIndexManager specifies index management behavior for this client. See
+// interface definition of IndexManager for more information.
+func (c *Client) SetIndexManager(mgr IndexManager) {
+	c.modifyLock.Lock()
+	defer c.modifyLock.Unlock()
+
+	c.setIndexManager(mgr)
+}
+
+func (c *Client) setIndexManager(mgr IndexManager) {
+	c.state = mgr
+}
+
+func (c *Client) setIndexFromResult(request *Request, response *Response) {
+	addr := request.URL.String()
+	ns := request.Headers.Get(NamespaceHeaderName)
+	index := response.Header.Get(IndexHeaderName)
+
+	if index == "" {
+		return
+	}
+
+	c.modifyLock.RLock()
+	state := c.state
+	c.modifyLock.RUnlock()
+
+	if state == nil {
+		return
+	}
+
+	state.Set(addr, ns, index)
+}
+
+// WithIndexManager creates a new cloned client with a unique index manager;
+// other operations including Clone preserve the shared index manager
+// instance.
+func (c *Client) WithIndexManager(mgr IndexManager) *Client {
+	c2 := *c //nolint:copylocks
+	c2.modifyLock = sync.RWMutex{}
+	c2.state = mgr
+	return &c2
+}
+
+// IndexManager specifies the behavior of index resolution. Get is called
+// when building each request and Set is called after receiving an answer,
+// if it has an index header value. Each implementation of IndexManager is
+// free to use or discard contextual information as appropriate.
+type IndexManager interface {
+	// Get the current index for the specified address and namespace.
+	Get(addr string, namespace string) string
+
+	// Set stores the new index for the specified address and namespace.
+	Set(addr string, namespace string, index string)
+}
+
+type simpleIndexManager struct {
+	lock   sync.RWMutex
+	latest string
+}
+
+var _ IndexManager = &simpleIndexManager{}
+
+// NewSimpleIndexManager creates a simple index manager instance that stores
+// only the latest (by order of Set(...) call) index. This may break down in
+// highly parallel applications, but should generally be sufficiently correct
+// except with many interleaved read/write calls.
+//
+// It does not handle cluster addresses or namespaces.
+func NewSimpleIndexManager() IndexManager {
+	return &simpleIndexManager{}
+}
+
+func (i *simpleIndexManager) Get(_ string, _ string) string {
+	i.lock.RLock()
+	defer i.lock.RUnlock()
+
+	return i.latest
+}
+
+func (i *simpleIndexManager) Set(_ string, _ string, index string) {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	i.latest = index
+}
--- a/api/index.go
+++ b/api/index.go
@@ -1,0 +1,114 @@
+// Copyright (c) 2026 OpenBao a Series of LF Projects, LLC
+// SPDX-License-Identifier: MPL-2.0
+
+package api
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"sync"
+)
+
+// IndexValue is the encoded value of the X-Vault-Index header(s) as sent by
+// OpenBao.
+type IndexValue struct {
+	Cluster string `json:"cluster,omitempty"`
+	Value   string `json:"value,omitempty"`
+}
+
+// Encode transforms this IndexValue to wire format.
+func (i *IndexValue) Encode() (string, error) {
+	if i == nil {
+		return "", nil
+	}
+
+	value, err := json.Marshal(i)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal: %w", err)
+	}
+
+	return base64.StdEncoding.EncodeToString(value), nil
+}
+
+// DecodeIndexValue takes the value of an X-Vault-Index header from OpenBao
+// and parses it. It will not work with Vault Enterprise X-Vault-Index header
+// values.
+func DecodeIndexValue(value string) (*IndexValue, error) {
+	if value == "" {
+		return nil, nil
+	}
+
+	data, err := base64.StdEncoding.DecodeString(value)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding base64: %w", err)
+	}
+
+	var i IndexValue
+	if err := json.Unmarshal(data, &i); err != nil {
+		return nil, err
+	}
+
+	return &i, nil
+}
+
+// SetInconsistent provides one or more values for the X-Vault-Inconsistent
+// header. See notes around Index* constants in client.go.
+func (c *Client) SetInconsistent(behaviors ...string) {
+	c.modifyLock.Lock()
+	defer c.modifyLock.Unlock()
+
+	c.setInconsistent(behaviors)
+}
+
+func (c *Client) setInconsistent(behaviors []string) {
+	c.inconsistent = behaviors
+}
+
+func (c *Client) setStorageIndexTracker(mgr *simpleStorageIndexTracker) {
+	c.indexTracker = mgr
+}
+
+func (c *Client) setIndexFromResult(request *Request, response *Response) error {
+	index := response.Header.Get(IndexHeaderName)
+	if index == "" {
+		return nil
+	}
+
+	c.modifyLock.RLock()
+	tracker := c.indexTracker
+	c.modifyLock.RUnlock()
+
+	if tracker == nil {
+		return nil
+	}
+
+	tracker.set(index)
+	return nil
+}
+
+// simpleStorageIndexTracker is an implementation of what will likely
+// eventually become an exported interface for handling index assignments
+// from responses and to requests.
+type simpleStorageIndexTracker struct {
+	lock   sync.RWMutex
+	latest string
+}
+
+func newSimpleStorageIndexTracker() *simpleStorageIndexTracker {
+	return &simpleStorageIndexTracker{}
+}
+
+func (i *simpleStorageIndexTracker) get() string {
+	i.lock.RLock()
+	defer i.lock.RUnlock()
+
+	return i.latest
+}
+
+func (i *simpleStorageIndexTracker) set(index string) {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	i.latest = index
+}
--- a/changelog/3839.txt
+++ b/changelog/3839.txt
@@ -1,0 +1,18 @@
+```release-note:feature
+**Strong Consistency Control**: Support `X-Vault-Index` and `X-Vault-Inconsistent` headers to ensure client/server consistency.
+  - `X-Vault-Index` is now sent when a write is performed.
+  - `X-Vault-Inconsistent` can take the following behaviors to control fallback behavior when the node is out of date:
+      - fail`, to respond with 429 Too Many Requests and a `Retry-After` value,
+      - `forward-active-node` to forward the request to the active node, or
+	  - `await-state` to hold the request locally to attempt to catch up, optionally falling back to one of the above behaviors (if also specified) or the server default.
+  - On listeners, three new configuration parameters are now respected:
+	  - `consistency_fallback_behavior` to set the default fallback for `await-state`,
+	  - `consistency_missing_header_forward` to automatically forward requests which do not have any consistency control headers, and
+	  - `consistency_max_index_wait` to indicate the maximum amount of time `await-state` should pause before forwarding the request.
+```
+```release-note:improvement
+core/server: Support storage-indicated index values in Server Side Consistent Tokens (SSCTs), improving their utility. Previously these tokens always had an index value of `0`.
+```
+```release-note:bug
+core/server: Add missing `disable_ssct_token` parameter to support enabling SSCTs.
+```
--- a/changelog/3839.txt
+++ b/changelog/3839.txt
@@ -1,0 +1,18 @@
+```release-note:feature
+**Strong Consistency Control**: Support `X-Vault-Index` and `X-Vault-Inconsistent` headers to ensure client/server consistency.
+  - `X-Vault-Index` is now sent when a write is performed.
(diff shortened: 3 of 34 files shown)


api/index.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	state.Set(addr, ns, index)
+}
+
+// WithIndexManager creates a new cloned client with a unique index manager;
+// other operations including Clone preserve the shared index manager
+// instance.
+func (c *Client) WithIndexManager(mgr IndexManager) *Client {
+	c2 := *c //nolint:copylocks
```
cipherboy: Is this not the right format? 🤔

nabramovitz:
@cipherboy Built it and exercised it against the workload from
openbao/openbao#3877. Short version: it fixes that class of corruption, via
the listener option rather than the client API, and I hit one behavior in
the forwarding path that looks unintended.

Branch builds clean on Go 1.27, and `TestHTTP_Index_Defaults` and
`TestHTTP_Index_ForwardMissingAndAwait` both pass.

## What I tested beyond the tests in the PR

The tests here cover write-then-read. The case that corrupted our data is
read-modify-write, where the read feeds the write that follows it, and — this
is the part that turns out to matter — **the client is a separate process for
each write.** Ours is a CLI invoked once per part of a large value.

So I built a two-node raft cluster and wrote one path in four parts, each
part a read-modify-write: read what the path holds, merge that part's keys
in, write the whole map back. Every part used a newly constructed client
pointed at the standby, with nothing carried over between them. Ten keys per
part, twelve paths per run, 480 keys total. Then I counted how many of the
keys asked for were absent from the final value.

| configuration | keys lost |
|---|---|
| default | **250 of 480**, silently |
| client sets `X-Vault-Inconsistent: fail`, no index available | **250 of 480** |
| `consistency_missing_header_forward = true` | **0 of 480** |

The first row is our production bug, reproduced at the HTTP layer.

## The middle row is the one I would flag

A client that starts fresh for every request has no index to send, so there
is nothing for the server to compare against and the request is served
locally. Asking for the strictest behavior does not change that —
`handleWithoutIndexForward` serves locally whenever `X-Vault-Inconsistent` is
present without an index, which is the correct reading of "the caller knows
about forwarding but has no opinion," but it means header-awareness alone
buys a per-invocation CLI nothing.

That leaves two ways for a tool like ours to be safe: persist the index
between invocations, or have the operator set
`consistency_missing_header_forward`. The second works today and needs no
client change, which is a good story — it just means the client-side API is
effectively for long-lived clients only. Worth being explicit about in the
docs, since "use the new headers" is the natural conclusion a tool author
would draw.

## `sys/leader` is forwarded, so standbys report themselves as active

With `consistency_missing_header_forward = true`, a status query carries no
index header, so it forwards like any other read and the answer describes the
active node instead of the node asked:

```
forward off:  core 0 is_self=true   core 1 is_self=false
forward on:   core 0 is_self=true   core 1 is_self=true   (leader_address = core 0)
```

`wrapIndexForwardHandler` wraps the whole mux, and I could not find an
exemption for status endpoints in `internal/http/index.go`. That affects
`bao status`, monitoring, and any load balancer whose health check picks the
active node — and it also breaks `DeriveStandbyCores` in this repo's own
testhelpers, which identifies a standby by asking `sys/leader` and checking
`is_self`. That is what made my first run fail. I notice
`TestHTTP_Index_ForwardMissingAndAwait` uses `cluster.Cores[1]` directly
rather than deriving it, so it does not trip over this.

Exempting `sys/leader`, `sys/health` and `sys/seal-status` from
missing-header forwarding would fix it, if the intent is that a status query
describes the node it was sent to.

Happy to go into more detail on any of the above, or to re-run the
measurements against a revision.

cipherboy:
@phil9909 Do you have capacity for a review here? We're getting close to cutting the beta and it'd be good to get this in.

api/index.go
```diff
@@ -0,0 +1,134 @@
...
+	c.setIndexManager(mgr)
+}
+
+func (c *Client) setIndexManager(mgr IndexManager) {
+	c.state = mgr
+}
+
+func (c *Client) setIndexFromResult(request *Request, response *Response) {
+	addr := request.URL.String()
+	ns := request.Headers.Get(NamespaceHeaderName)
```
phil9909: I guess this fails for e.g. `client.Logical().Read("my-namespace/my-mount/...")`

Also: should `client.Sys().CreateNamespace(...)` set the index of the child namespace?

changelog/3839.txt
```diff
@@ -0,0 +1,18 @@
+```release-note:feature
+**Strong Consistency Control**: Support `X-Vault-Index` and `X-Vault-Inconsistent` headers to ensure client/server consistency.
+  - `X-Vault-Index` is now sent when a write is performed.
+  - `X-Vault-Inconsistent` can take the following behaviors to control fallback behavior when the node is out of date:
+      - fail`, to respond with 429 Too Many Requests and a `Retry-After` value,
```
phil9909:
```suggestion
      - `fail`, to respond with 429 Too Many Requests and a `Retry-After` value,
```

internal/http/index.go
```diff
@@ -0,0 +1,188 @@
...
+
+	if forward {
+		// Forward the request to the active node; the state hasn't yet been seen.
+		forwardRequest(core, w, r)
+		return
+	}
+
+	// Tell the client to retry later; the state hasn't yet been seen.
+	w.Header().Add(httplimit.HeaderRetryAfter, "1")
+	respondError(w, http.StatusTooManyRequests, nil)
```
phil9909, line 207:
```suggestion
	respondError(w, http.StatusPreconditionFailed, nil)
```

No strong opinion, but RFC states 412: https://openbao.org/community/rfcs/index-headers/#:~:text=the%20request%20will%20be%20rejected%20with%20a%20412%20HTTP%20status%20code

**EDIT:** As discussed in the DEV WG Meeting: RFC will be adapted.

api/index.go
```diff
@@ -0,0 +1,134 @@
...
+	c2.modifyLock = sync.RWMutex{}
+	c2.state = mgr
+	return &c2
+}
+
+// IndexManager specifies the behavior of index resolution. Get is called
+// when building each request and Set is called after receiving an answer,
+// if it has an index header value. Each implementation of IndexManager is
+// free to use or discard contextual information as appropriate.
+type IndexManager interface {
```
phil9909: We now have two `IndexManager`s one in the backend and one in the API client.

Could we change the name of the API `IndexManager`? Maybe `IndexTracker`?

website/content/docs/configuration/listener/shared.mdx
```diff
@@ -8,6 +8,28 @@ description: |-
...
+  (`forward-active-node`).
+- `consistency_missing_header_forward` `(bool: false)` - when set to true, any
+  requests from a client without `X-Vault-Index` or `X-Vault-Inconsistent` will
+  be forwarded to the active node. This makes an otherwise read-enabled standby
+  node function like one with `disable_standby_reads=true` set, but only for
+  clients which are not aware of eventual consistency.
+- `consistency_max_index_wait` `(time.Duration: 0s)` - maximum time the standby
+  will hold requests to catch up to the specified `X-Vault-Index`, before
+  applying the fallback behavior. This requires `X-Vault-Inconsistent: await-state`
+  be set on the request. `25ms` is the minimum and will be added to the value
```
phil9909, line 29: According to `handleIndexForwardWithAwait` it is 100ms

```suggestion
  be set on the request. `100ms` is the minimum and will be added to the value
```

---

But this seems a bit odd to me, why did you not choose to have a minimum value of 100ms instead (and default to 100ms)?

With this setup, lets say I want 3 seconds: I need to set `consistency_max_index_wait = 2.9s`. Sure I can set it to 3s instead and 0.1 additional seconds probably don't matter, but it's weird.

internal/vault/invalidation.go
```diff
@@ -607,7 +646,26 @@ func (im *invalidationManager) Add(key ...string) {
 
 	// Add the keys.
 	im.pendingLock.Lock()
-	im.pending = append(im.pending, key...)
+	if len(im.pending) == 0 {
+		// Pre-allocate enough entries if we're empty; this should occur
+		// fairly frequently given the dispatch notification logic.
+		im.pending = make([]pendingEntry, 0, len(keys))
+	}
```
phil9909: How about this (needs import of `slices` from stdlib)?

```suggestion
	im.pending = slices.Grow(im.pending, len(keys))
```

internal/vault/storage_index.go
```diff
@@ -122,8 +217,53 @@ func (i *indexManager) getIndexLocked(ctx context.Context) (string, error) {
...
+
+		return i.outstandingIndices
+	}(); outstanding != nil {
+		return outstanding
+	}
+
+	i.outstandingLock.Lock()
+	defer i.outstandingLock.Unlock()
+
+	return i.getOutstandingLocked()
```
phil9909, line 260:
```suggestion
	if !time.Now().After(i.outstandingUpdate.Add(i.backoff)) {
		return i.outstandingIndices
	}
	
	return i.getOutstandingLocked()
```

multiple routines might at the same time acquire a read-lock, see that the back-off has elapsed and subsequently acquire a write lock. So we need to recheck the condition here.

changelog/3839.txt
```diff
@@ -0,0 +1,18 @@
+```release-note:feature
+**Strong Consistency Control**: Support `X-Vault-Index` and `X-Vault-Inconsistent` headers to ensure client/server consistency.
+  - `X-Vault-Index` is now sent when a write is performed.
```
phil9909, line 3: also reads, right?

changelog/3839.txt
```diff
@@ -0,0 +1,18 @@
...
+  - `X-Vault-Inconsistent` can take the following behaviors to control fallback behavior when the node is out of date:
```
phil9909: This sentence is really hard to understand (for me)

```suggestion
  - The `X-Vault-Inconsistent` request header can be used to control the behavior when node is out of date:
```

internal/http/index.go
```diff
@@ -0,0 +1,188 @@
...
+		handleIndexForwardWithAwait(mux, core, w, r, index, forward, maxWait)
+		return
+	}
+
+	seen, err := core.HaveSeenStorageIndex(r.Context(), index)
+	if err != nil {
+		// Treat errors as seen indices: this ensures most distributed load
+		// across the cluster if the system is down.
+		seen = true
+	}
```
phil9909, line 173: I'm not convinced this is a good idea.

internal/http/index.go
phil9909, line 1: The control-flow is a bit hard to follow here.
I will try to refactor this later

internal/http/logical.go
```diff
@@ -402,6 +402,10 @@ func respondLogical(core *vault.Core, w http.ResponseWriter, r *http.Request, re
 		return
 	}
 
+	if r != nil {
```
phil9909, line 405: In which case is `r = nil`? only during testing?

website/content/docs/concepts/consistency.mdx
```diff
@@ -0,0 +1,186 @@
...
+When using a standby read-enabled HA storage backend with the default
+`disable_standby_reads=false` value, both the active node and standby nodes
+can handle requests. If a standby handles a request which issues a write
+operation, the standby will issue a GRPC forwarding operation and fallback to
+an API-level forward to the active node to handle the request.
+
+On all standby read-enabled HA storage backends, all writes occur on the
+active node, which updates the local storage on every other node through the
+storage engine's replication mechanism. This mechanism includes write
+notifications allowing cache invalidation occurs.  Between when the active
```
phil9909:
```suggestion
notifications allowing cache invalidation to occur. Between when the active
```

internal/vault/invalidation.go
```diff
@@ -54,7 +60,8 @@ type invalidationManager struct {
 	// Invalidate stages pending invalidations into this queue.
 	pendingLock   sync.Mutex
 	accepting     atomic.Bool
-	pending       []string
+	pending       []pendingEntry
+	outstanding   map[string]map[string]struct{}
```
phil9909, line 65: took me a while to understand the difference between "pending" and "outstanding"

BTW: should we rename `pendingLock` as it now protects both?

internal/vault/invalidation.go
```diff
@@ -627,3 +685,20 @@ func (im *invalidationManager) splitNamespaceFromKey(key string) (string, string
 
 	return namespaceUUID, namespacedKey
 }
+
+// hasOutstandingInvalidations returns the list of outstanding invalidations.
+func (im *invalidationManager) OutstandingInvalidationIndices() []string {
```
phil9909, line 689: I'm wondering, if it would be better to return the lowest outstanding index here, to reduce the number of comparisons that index manager has to do.

internal/vault/storage_index.go
```diff
@@ -112,6 +162,51 @@ func (i *indexManager) Await(ctx context.Context, index string) error {
 	return err
 }
 
+// AwaitInvalidated first calls Await and then additionally awaits for all
+// outstanding older invalidations to be concluded.
+func (i *indexManager) AwaitInvalidated(ctx context.Context, index string) error {
```
phil9909, line 171: I think we have a race here:
If `Await` returns before the index is added to the (cached) outstanding keys list (and any lower indices are either already process or also not yet part of the list, e.g. the list might be empty) we will return success before the invalidation has been processed.

phil9909 (changes requested):

internal/vault/storage_index.go
cipherboy: @phil9909 I don't believe it's possible to solve this race. I'm happy for you to take a stab in a follow up if you're wanting. The mitigating factor is most of the time the requested index in that case won't have been reported by the backend. Yes, there's a small window in which that isn't true, but I don't think there's anything we can do to alleviate that and most of the time we'll be just fine.

api/index.go
cipherboy: @phil9909 Yes, I'm not sure we have a viable way of solving this. I think in the future we may need to encode namespace values in the headers themselves. 🤔

changelog/3839.txt
cipherboy: I do not want to specify read behavior in case we change it

internal/http/index.go
cipherboy: @phil9909 Why?

internal/http/logical.go
cipherboy: Yes, and there's a billion tests so it'll be impossible to fix.

api/index.go
cipherboy: @phil9909 Not sure I see why this is confusing? Most people will only see this one.

api/index.go
cipherboy: re: 

> Also: should client.Sys().CreateNamespace(...) set the index of the child namespace?

I think we should always respect whatever the server returns. In the future with #3874, the server may send multiple values if multiple namespaces' storages were updated. Creating a namespace will ultimately return an index for the parent and an index child namespace.

cipherboy: I have a change locally to use a JSON struct here. Will push shortly.

website/content/docs/configuration/listener/shared.mdx
cipherboy: I've standardized on 25ms and set a minimum instead. Thanks!

changelog/3839.txt
cipherboy: I'll push an alternative.

website/content/docs/concepts/consistency.mdx
cipherboy: Thanks!

internal/vault/invalidation.go
cipherboy: Thanks.

internal/vault/invalidation.go
cipherboy: Sure, feel free to propose something in a follow-up PR :+1:

internal/vault/invalidation.go
cipherboy: @phil9909 Keep in mind comparisons of indices requires a call to `backend.GreaterEqualReplicationIndex(...)` which may be expensive so we don't really want to be doing that while holding (especially exclusive) locks and thus want to push it as far as we can to read-only locks or no-locks (like it is currently).

This is also why `IndexManager` doesn't do the comparison either (`GetOutstanding(...)` is again potentially write-locked). 

So while viable, I think it will have performance drags.

internal/vault/storage_index.go
cipherboy: Sure, I've changed the condition slightly and fixed this for `GetIndex(...)` as well. A few extra calls here are probably insignificant globally given that read lock acquisition will be blocked by write lock, so you'd only have the set of readers who immediately raced to acquire a stale value. As soon as one grabs the write lock, all others will block waiting for the read lock.

changelog/3839.txt
phil9909: [xkcd #1172](https://xkcd.com/1172/)

internal/vault/storage_index.go
phil9909: WDYT about this https://github.com/openbao/openbao/commit/1e0103cc183e643c4af9d2c691cffdfff6d57261?

internal/http/index.go
cipherboy: @phil9909 In particular, errs are most commonly caused by three things:

1. A bad index value sent by the client. Nothing we can do will help us here so we're probably better off just treating it as if it was a non-indexed request like we do if the index was for another cluster.
2. Storage problems, e.g., failure to read or compare indices. If not a special case of 1 above, in the case of PostgreSQL, it means the storage may be having problems and we'd rather surface errors than flood the primary server unexpectedly.

I don't think there's anything we can do about 2 and 1 is hard to detect without a wrapped and validation from the underlying libraries. In particular, for both PostgreSQL and Raft, any 64-bit integer roughly looks like it could be a storage index, but whether it is or not (or, is relative to this instance) is very unclear without knowing the current index. 

The alternative of surfacing the error is that transient errors in a backend (to fetch indices) yield errors up to the caller, which I'm not convinced is great either, especially if the rest of the operations complete well enough.

:shrug:

internal/vault/storage_index.go
cipherboy: @phil9909 Several things:

I think we're also thinking about different race conditions.

The one I'm thinking of is that the backend's index (as reported by `AppliedReplicationIndex(...)`) is updated before we handle any invalidations. While we can't do anything about that for PostgreSQL (as grpc invalidation relies on `Await(...)` again and thus is slightly non-deterministic w.r.t. ordering of invalidation and awaited indices), Raft shouldn't have this problem as invalidation is called before the index term is stored.

I don't think your patch solves that one...


Regarding the patch in general... there's a couple of incorrect assumptions:

1. > This change assumes `core.Invalidate` is called with a monotonically increasing index.
   
   I _think_ we have a strictly-increasing assumption in Raft, (though I'm not [even sure of that](https://github.com/openbao/openbao/blob/71d108d3d23f5dc145fb38e4691e91424623d545/internal/physical/raft/fsm.go#L769-L784) -- note that suggest we may have already seen a later log entry and were just slow to process it) but like mentioned right above, I don't think we can guarantee anything w.r.t. GRPC invalidation and index ordering due to the necessary call to `Await(...)`.
2. > Another assumption: all indices that are reported by `AppliedReplicationIndex` are eventually also `Invalidate`d by the storage layer (even if the list of keys is empty)
  
   I'm fairly sure this assumption is also incorrect globally. I believe we could have a series of `LogConfiguration` entries in Raft, which would increment the applied log index value (and be visible), but wouldn't result in any storage operations and thus no invalidations. This would thus be non-invalidated index. Similarly in PostgreSQL, any use of a shared database instance (e.g., GitLab) would result in WAL indices increasing, but not being realized. Plus, in PostgreSQL, you're not guaranteed to get index 1, 2, 3, 4, 5... &c, because the "index" we return is actually a two-part (2x 32bit) log number + byte offset in the log. 
3. I think `sorted` behaves by insertion order, which as mentioned above in this comment, isn't correct, but [as mentioned elsewhere](https://github.com/openbao/openbao/pull/3839#discussion_r3925708601), to properly compare indices per contract, you need to call `GreaterEqualReplicationIndex(...)`, which may be expensive.

I see now how this could (with a different design) address a small window whereby `IndexManager.Outstanding(...)` is out-of-date w.r.t. `invalidationManager.OutstandingInvalidationIndices(...)` (by design of cached results), though I'm not sure what options we have here that are correct.

In particular, note that `AwaitInvalidated` is lazy, so we might just be better off repeating `op` until it has succeeded twice. :shrug:

api/index.go
phil9909: Not for the end-user, but for contributors. If in the next dev-wg call I tell you we have a bug in the IndexManager we might be talking about two different things. Or if I do a `rg --ignore-case IndexManager` I will find matches for both.

Besides that `...Manager` is a super generic term, in the case of the API client `IndexTracker` IMHO is the better choice anyhow, because well it keeps track of the current index (not saying it's perfect, I'm open for other ideas).

internal/vault/invalidation.go
phil9909: This would be solved by https://github.com/openbao/openbao/pull/3839#discussion_r3933850360

But as this is just about performance: Lets first see if there is even a problem, we can always tweak this later.

api/client.go
```diff
@@ -707,6 +740,14 @@ type Client struct {
...
 	requestCallbacks   []RequestCallback
 	responseCallbacks  []ResponseCallback
+
+	// Policy to use for X-Vault-Inconsistent; exact behavior of OpenBao
+	// depends on listener configuration and the value of this option.
+	inconsistent []string
+
+	// state is an index manager shared across all client implementations;
+	// it holds the latest seen index from a given cluster.
+	state IndexManager
```
satoqz: "state" seems a bit generic.

internal/vault/storage_index.go
phil9909: Assumption 1 is equivalent to 3, that was the whole trick to get around the `GreaterEqualReplicationIndex` calls. If we can't ensure this, the whole idea won't work.

Assumption 2 is not required for correctness, but if it is not given some requests will be blocked unnecessarily especially on clusters with low write traffic. Something we could work around by including the index in a heart beat message.

> I think we're also thinking about different race conditions.

The race I was thinking about:
- A request comes in with `X-Vault-Inconsistent: await-state` and `X-Vault-Index: 1234`
- The request handler calls `AppliedReplicationIndex` and gets `1234` (via `HaveSeenStorageIndex`)
- It then goes on to call `AwaitInvalidated`. But the storage did not yet call `Invalidate` for the index `1234` (maybe the nodes share the same database instance so postgres "replication" is faster than GRPC based invalidation trigger)
- It will wrongly assume that `1234` has already been invalidated (given there are no other, older outstanding invalidation items)

To summarize: the current implementation assumes that `AppliedReplicationIndex` will never report a high index than the highest index that was already enqueued for invalidation (via `core.Invalidate`)[^1], which I fear is wrong.

How likely is this to happen? I guess in the case were primary and standby share the same postgres, replication is therefore instant and the GRPC based invalidation triggers will always lag a bit behind: it is not unrealistic for this to happen.

BTW: I don't say my implementation should be used instead (otherwise I would have opened a PR), just food for thought.


> I don't think your patch solves that one...

There is a TODO in the code to initialize `invalidatedUpToIncluding` to the value of `AppliedReplicationIndex` before we unseal. This should ensure we handle start-up correctly.

For the beta: I guess this is good enough for people to test drive the new headers. But I would like to have this race eliminated (or you convince me its not possible) before a GA release. Having a race that might cause inconsistent reads for a feature specifically designed to prevent inconsistent reads is not easily acceptable :laughing:. If this means a bigger refactoring, we might even want a second beta.


[^1]: ignoring indices that did not trigger any invalidation, e.g. because postgres did increase the index for some internal metadata.

api/index.go
```diff
@@ -0,0 +1,175 @@
...
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"sync"
+)
+
+type IndexValue struct {
+	Cluster   string `json:"cluster"`
+	Namespace string `json:"namespace"`
```
satoqz: Do we have a use for these two as of now? While likely relevant in the future in some form, I'd prefer not to commit to any fields (esp. in the api package) that weren't properly discussed yet.

Generally, they should probably be `omitempty` so we can save some space?

api/index.go
```diff
@@ -0,0 +1,175 @@
...
+
+	value, err := json.Marshal(i)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal: %w", err)
+	}
+
+	return base64.StdEncoding.EncodeToString(value), nil
+}
+
+func DecodeIndexValue(value string) (*IndexValue, error) {
```
satoqz, line 37: Needs a doc comment, `IndexValue` and `IndexValue.Encode`, too.

api/index.go
```diff
@@ -0,0 +1,175 @@
...
+	if err := json.Unmarshal(data, &i); err != nil {
+		return nil, err
+	}
+
+	return &i, nil
+}
+
+// SetInconsistent provides one or more values for the X-Vault-Inconsistent
+// header. See notes around Index* constants in client.go.
+func (c *Client) SetInconsistent(behaviors []string) {
```
satoqz: Perhaps `SetInconsistent` could just take `behaviors ...string`, and then we'd have nice enough syntax to avoid needing all the other helper methods?

api/index.go
```diff
@@ -0,0 +1,175 @@
...
+	// Set stores the new index for the specified address and namespace.
+	Set(addr string, index string) error
+}
+
+type simpleIndexManager struct {
+	lock   sync.RWMutex
+	latest string
+}
+
+var _ IndexManager = &simpleIndexManager{}
```
satoqz: I know you really like this pattern, but it is a bit unnecessary when you've got

```go
func NewSimpleIndexManager() IndexManager {
	return &simpleIndexManager{}
}
```

right below, performing the same assertion more naturally, the way the Gopher intended :P

api/index.go
satoqz: I would go as far as suggest `StorageIndexTracker` since "index" is such a loaded term. But `Tracker` is certainly better than `Manager`.

Also see [this section of the Zig style guide](https://ziglang.org/documentation/master/#Avoid-Redundancy-in-Names) if you'd like a take that'll make you hate 90% of the type names in this project.

internal/http/index.go
phil9909: This feature is meant for ensuring consistency. This just throws that out of the window.

Lets say we are in case 2 "storage problems". The problem might have been caused by a broken TCP connection, so it might only affect a single query execution, the next query will get a fresh connection from the pool. Now we just skipped waiting for the correct index and continue with (potentially) stale data.

To put it otherwise: It might be true that in 99% of the cases it is safe (and beneficial) to swallow the error here, maybe even 99.999%, but that's still not 100%. I have a hard time to see how this is a good idea for a feature that is design to bring back consistency.

website/content/docs/concepts/consistency.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+As a result, even if you're always talking to the same read-enabled standby
+node, you may not get read-after-write semantics. The write gets sent to the
+active node, and if the subsequent read request occurs before the new data
+gets sent to the node handling the read request, the read request won't be
+able to take the write into account because the new data isn't present on
+that node yet.
+
+As of OpenBao v2.7, all requests that modify storage now return a new HTTP
+response header:
```
satoqz: Can we avoid "now" and "new" wording in docs? This'll inevitably become dated and convey the wrong idea.

website/content/docs/concepts/consistency.mdx
```diff
@@ -0,0 +1,186 @@
...
+```go
+cfg := api.DefaultConfig()
+cfg.DefaultStrongConsistency = true
+client := api.NewClient(api.DefaultConfig())
+
+_, err := client.Logical().Write(path, data)
+secret, err := client.Logical().Read(path)
+```
+
+This will retry the `Read` until the data stored by the `Write` is present.
```
satoqz, line 186: It'd be more useful to link to (prospective) Go doc pages here.

website/content/docs/concepts/consistency.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+There are some new helpers in the `api` package to work with the new headers.
+`SetInconsistent(...)`, `FailInconsistent(...)`, `ForwardInconsistent(...)`,
+and `AwaitInconsistent(...)` can be used to control the `X-Vault-Inconsistent`
+header. `SetIndexManager(...)` and `WithIndexManager(...)` can be used to
+provide a tracker of server-sent index values. A default simple implementation
+can be created via `NewSimpleIndexManager(...)`.
+
+A new configuration option, `DefaultStrongConsistency`, can be set to opt-into
+the current set of best-practices regarding strong consistency.
```
satoqz: Not new! Timeless! Permanent! Transcendent!

website/content/docs/concepts/consistency.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+The drawback here is that when requests are forwarded to the active node,
+read-enabled standbys provide less value. If this happens often enough
+the active node can become a bottleneck, limiting the horizontal read
+scalability standbys are intended to provide.
+
+### Client initiated retry
+
+To ensure that the state resulting from that write request is visible to a
+subsequent request, add this headers to that second request:
```
satoqz:
```suggestion
subsequent request, add this header to that second request:
```

website/content/docs/configuration/listener/shared.mdx
```diff
@@ -8,6 +8,28 @@ description: |-
...
 There is config that is common between all listeners.
 
+## Consistency behavior
+
+For more information on consistency, refer to the [concepts section](../../concepts/consistency.mdx)
+of the documentation.
+
+- `consistency_fallback_behavior` `(string: "fail")` - specifies the default
+  behavior when a client does not specify a second value for the
+  `X-Vault-Inconsistent` header but does specify `X-Vault-Inconsistent: await-state`.
```
satoqz: "Doesn't specify a second value but specifies x as the first value" is a bit oddly ordered, how about "Specifies x without a second fallback header value"?

website/content/docs/configuration/listener/shared.mdx
```diff
@@ -8,6 +8,28 @@ description: |-
...
+  This defines whether the client should receive an error and retry (`fail` or
+  elided) or whether the request should be forwarded to the active node
+  (`forward-active-node`).
+- `consistency_missing_header_forward` `(bool: false)` - when set to true, any
+  requests from a client without `X-Vault-Index` or `X-Vault-Inconsistent` will
+  be forwarded to the active node. This makes an otherwise read-enabled standby
+  node function like one with `disable_standby_reads=true` set, but only for
+  clients which are not aware of eventual consistency.
+- `consistency_max_index_wait` `(time.Duration: 0s)` - maximum time the standby
```
satoqz:
```suggestion
- `consistency_max_index_wait` `(string: "0s")` - maximum time the standby
```

Is what the docs usually use.

internal/command/server/config.go
```diff
@@ -984,6 +985,14 @@ func ParseConfig(d, source string) (*Config, error) {
 		}
 	}
 
+	if result.DisableSSCTokensRaw != nil {
+		disabled, err := parseutil.ParseBool(result.DisableSSCTokensRaw)
+		if err != nil {
+			return nil, err
+		}
+		result.DisableSSCTokens = &disabled
+	}
```
satoqz, line 1000: Needs adding to Merge/Sanitize.

api/index.go
```diff
@@ -0,0 +1,175 @@
...
+// IndexManager specifies the behavior of index resolution. Get is called
+// when building each request and Set is called after receiving an answer,
+// if it has an index header value. Each implementation of IndexManager is
+// free to use or discard contextual information as appropriate.
+type IndexManager interface {
+	// Get the current index for the specified address and namespace.
+	Get(addr string, namespace string) string
+
+	// Set stores the new index for the specified address and namespace.
+	Set(addr string, index string) error
```
satoqz: Why don't these take and return `IndexValue` directly? Perhaps that'd also let us remove the error return?

internal/vault/tokens/token.proto
```diff
@@ -16,6 +16,13 @@ message SignedToken {
 
 message Token {
     string random = 1;       // unencoded equiv of former $randbase62
-    uint64 local_index = 2; // required storage state to have this token
-    uint32 index_epoch = 3;
+
+    // Fields 2 was present from Vault and was empty (0) for all OpenBao
+    // releases, making it safe to ignore.
+
+    uint32 index_epoch = 4;
```
satoqz: Why did `index_epoch` move from 3 to 4 though? If it has to move, would be good to have a comment on that, too.

internal/vault/ha.go
```diff
@@ -1245,6 +1245,13 @@ func (c *Core) runReadEnabledStandby(ctx context.Context, ctxCancel context.Canc
 
 	c.drainPendingRestarts()
 
+	// Check if we're missing a cluster identifier; if so, attempt to load one.
+	if c.ClusterID() == "" {
+		if _, err := c.Cluster(ctx); err != nil {
+			c.logger.Error("failed to load cluster information; unable to respond with index identifiers", "error", err)
+		}
+	}
```
satoqz, line 1253: Would it be wrong to call `setupCluster()`? At least, by giving `c.Cluster()` a side effect, the flow in `setupCluster()` doesn't make as much sense anymore. The side effect generally seems like an unnecessarily indirect way to achieve what we want here.

internal/vault/storage_index.go
cipherboy: @phil9909 said:

> The race I was thinking about:
> ...
> To summarize: the current implementation assumes that `AppliedReplicationIndex` will never report a high index than the highest index that was already enqueued for invalidation (via `core.Invalidate`)[1](#user-content-fn-1-01aa623e04741899016b659a7a65f5d7), which I fear is wrong.

So we're talking about the same race, but as noted in:

> While we can't do anything about that for PostgreSQL (as grpc invalidation relies on Await(...) again and thus is slightly non-deterministic w.r.t. ordering of invalidation and awaited indices), Raft shouldn't have this problem as invalidation is called before the index term is stored.

there's nothing we can do about it: the two functions are independent of each other and there's no way for us to wait (in the storage level in PostgreSQL) for the index to be accepted by `Invalidate(...)` before we report it as reached. In fact, w.r.t. to GRPC layer, it is _wrong_ to do so.

Your patch does nothing to solve this as far as I can tell: to solve it, one would need to reach into the GRPC layer and change it.

api/index.go
cipherboy: Cluster we definitely do and respect on the server side: 

https://github.com/cipherboy/openbao/blob/e21c5b067f4bac26f393976750599db0e4114c77/internal/http/index.go#L90-L98

```go

		decoded, err := api.DecodeIndexValue(index)
		if err != nil {
			respondError(w, http.StatusBadRequest, fmt.Errorf("failed to decode %q", api.IndexHeaderName))
			return "", nil, false
		}

		cluster := core.ClusterID()
		if cluster == decoded.Cluster {
			index = decoded.Value
```

Namespace, sure, can make `omitempty` or drop.

website/content/docs/concepts/consistency.mdx
cipherboy: Yeah, it was a light reword of the original one: https://github.com/openbao/openbao/blob/fork-point/website/content/docs/enterprise/consistency.mdx#conditional-forwarding-performance-standbys-only

> As of Vault Enterprise 1.7, all requests that modify storage now return a new HTTP response header:

internal/vault/ha.go
cipherboy: @satoqz Yes, most of `SetupCluster(...)` shouldn't be called on a standby node, and will more likely error. The side effect seems more like what we'd want a `GetCluster(...)` operation to generally achieve w.r.t caching. TBH, most of the calls to `Cluster(...)` seem like they aught to be replaced by underlying calls, if they actually had been cached in the first place.

api/index.go
cipherboy: @satoqz This is fine, but do note that `ForwardInconsistent` already existed at `fork-point`, I just expanded the pattern:

https://github.com/openbao/openbao/blob/8993802145833ab01d49c6070d787a9eccb81546/api/client.go#L1750-L1758


...and updated it to match newer semantics.

api/index.go
cipherboy: Yeah, I just got bit a while ago by something that should've satisfied a given interface not satisfying it until runtime so now I'm paranoid. :laughing:

internal/vault/storage_index.go
phil9909: The crucial difference is that my patch changes from "default allow" to "default deny" if you will.

So instead of saying "there is no index <= the given index in the outstanding list so assume we are done" we keep track of the highest index that we have already successfully invalided and compare against this one. Again this breaks down, if invalidations arrive out-of-order, but I think we can guarantee the order (shouldn't the GRPC stream already be in the correct order?)

cipherboy: @phil9909 No; independent of the order that the GRPC stream arrives in, it calls Await individually on each index which in turn dispatched invalidations for that index. If indices A and B (with A < B) arrive, await may return on B before A given that both use exponential back off and acquire their own locks/... so you can't be guaranteed an ordering.

See 
https://github.com/openbao/openbao/blob/1859768bdb897d92723c68f8457128c9fd707a98/internal/vault/grpc_invalidation.go#L283

And

https://github.com/openbao/openbao/blob/1859768bdb897d92723c68f8457128c9fd707a98/internal/vault/grpc_invalidation.go#L396

api/index.go
cipherboy: @satoqz Well, on the `Get` side, we'd then have to serialize it, which is fine. I figured since the actual interface with the server was `string`, it made sense to keep it in server format... Especially as in the beginning as it was an opaque value in the first implementation (and in Hashi's case).

Since we've moved this into API and formalized the encoding (rather than whatever opaque custom monstrosity Vault Enterprise had), I think its more fine to swap.

Thoughts?

internal/vault/tokens/token.proto
cipherboy: Ah, bug. Fixed.

internal/http/index.go
cipherboy: > This feature is meant for ensuring consistency. This just throws that out of the window.

Index headers are always best-effort and any claim otherwise is quite inconsistent with reality. We're hoping to make any inconsistency less likely. It is a distributed system, after all, and there's only so much we can do to help it.

Some nodes may permanently be stuck on old versions due to a storage partition. Some clients may preserve index headers from other clusters or future state (sending it after a snapshot-restore to an earlier point in time -- which they may not know about). Some may corrupt it, using VEnt's API client or sending a header for that, &c. Perhaps with strict server-side signatures over opaque index values, we could argue that any error condition is strictly able to be propagated back to the client as it would be an exceptional failure, though this doesn't quite guarantee we'd be free of all problems (like mixing two values or before/after a snapshot restore unless we rotate keys then too).

In particular, this is not a stateful API; it is inherently stateless. One needs to ask: 

> _What happens when the API repeatedly fails (and returns an error here)?_

Treating these errors as not-seen would mean **it would be possible then to hard-lock clients** if `forward=false` (and especially if no timeout was used, e.g., a non-context bound method in our current API of which there are many). 

I don't think that's a reasonable failure mode for eventual consistency.

I'd argue that a random chance of not-yet-consistent behavior is much better than potentially hard-locking clients, especially given on the whole, a lot of traffic is generally read-only (secret fetches).

My 2c.

website/content/docs/concepts/consistency.mdx
cipherboy: I'll update the links after we tag, rather than have breaking ones then.

cipherboy: https://github.com/openbao/openbao/issues/3958

api/index.go
cipherboy: The other thought I had was, if we keep it as `string`, we could potentially support VEnt compatibility in an index manager as well.

satoqz: This interface is just our storage abstraction, no? Doesn't the API client itself do all the serialization?

The main reason I'd prefer passing a struct is that the parameters we'll take here (namespace?) are still somewhat up in the air and struct parameters are much more forward-compatible than "plain" ones. That is, you can add new fields later on without breakage.

cipherboy: > Doesn't the API client itself do all the serialization? 

Client currently places the index header into the request, and takes it out of the response, yes, but it doesn't do any conversion to/from `IndexValue` right now.

But its possible (i.e., this simple implementation) to get an implementation without any serialization/deserialization. I think you could even expand it to a per-addr implementation as well without that, though combining multiple addresses across the same cluster would require deserialization+reserialization at that point. 

> The main reason I'd prefer passing a struct is that the parameters we'll take here (namespace?)

Should we drop `namespace` from `Get(...)` or perhaps even make `Get(...)` take a `path` instead of a namespace? :thinking: 

The only concern is if we add a new method in the future, I think strictly we'd need to introduce a new interface type which then complicates things. :shrug: 

---

@satoqz While I nominally agree with:

> somewhat up in the air and struct parameters are much more forward-compatible than "plain" ones.

by passing the serialized form (`index string`), it takes the header value verbatim from the server and thus doesn't get any additional information (namespace is yielded by the server, not the client in the semi-opaque from a function call parameters `string` value). We can't pass client's expected namespace, though as @phil9909 pointed out, we should probably use `path` rather than `namespace` in the API client case as the API client can't actually resolve namespaces at all. 

In the `Get(...)` case, I'm not really sure (as I mentioned above) about compatibility here outside of more interfaces. Use of an expected `IndexValue` as a parameter, I'm not sure about. :thinking: 

Maybe it should be `Get(client *api.Client, path net.URL, dest *IndexValue)` and the client populates `.Value` into the `IndexValue` field? :thinking:

satoqz (commented):
Remaining test failure looks interesting? Otherwise, this should be fine to ship with beta.

satoqz (approved):

cipherboy:
Merging so we can get beta started; any remaining changes shouldn't impact the API or SDK.

PR #3844, opened 2026-08-19
Fsm perf optimization

### LINKED ISSUE #3843: Performance issue
**Describe the bug**
Hello,
We are currently migrating from HashiCorp Vault Enterprise 1.16.3 to OpenBao. Technically, everything is going well, but we have encountered performance issues related to the Raft backend, specifically regarding access to the BoltDB file.

We have conducted extensive performance testing and identified the reason why OpenBao does not (currently) offer the same level of performance as Vault.
We are running a 6-node cluster on GCP dual-region that delivers excellent performance once the fix is ​​applied.

We have over 2,300 namespaces.
Our load ranges from:
* 25,000 to 35,000 req/min (excluding authentication)
* 6,000 to 11,000 req/min (authentication)

We monitor the following metrics:
* vault.core.handle_request.mean
* vault.core.handle_login_request.mean
* vault.raft_storage.stats.fsm_pending

We are performing a gradual migration (namespace by namespace), but once a certain request threshold is reached, OpenBao cannot write data to BoltDB fast enough.

Our infrastructure is currently over-provisioned to rule out this factor (C4 VMs with 40,000 IOPS Hyperdisks; dedicated disks for Raft and Audit).

Performance fix was validated on these cluster type:
* Single node
* 3 nodes in same GCP region
* 6 nodes dual regions (3 nodes per regions), 3 voters and 3 non voters. 

**To Reproduce**
Setup any cluster with Raft backend and with more than 1 node.

**Expected behavior**
Openbao performance (request handle time and login request handle time) must be the same as Hashicorp Vault 1.14.0 performance.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.5.5, 2.6.0
* OpenBao CLI Version (retrieve with `bao version`): 2.5.5, 2.6.0
* Server Operating System/Architecture: Ubuntu 24.04 / amd64

```hcl
disable_standby_reads           = true
plugin_directory                = "/opt/bao/plugins"
api_addr                        = "https://XXXX:8200"
cluster_addr                    = "https://XXXX:8201"
ui                              = true
default_lease_ttl               = "XX"
max_lease_ttl                   = "XX"
backend "raft" {
  path                   = "/opt/bao/raft"
  node_id                = "XXXX"
  performance_multiplier = 1
  retry_join {
    leader_tls_servername = "XXXX"
    auto_join = "provider=gce project_name=XXXX zone_pattern=XXXX tag_value=XXXX"
    auto_join_scheme = "https"
  }
}

telemetry {
  dogstatsd_addr = "127.0.0.1:8125"
  disable_hostname = true
}

listener "tcp" {
  address                          = "0.0.0.0:8200"
  cluster_address                  = "XXXX:8201"
  http_read_timeout                = "5m"
  tls_disable                      = false
  tls_key_file                     = "/opt/certs/privkey.pem"
  tls_cert_file                    = "/opt/certs/cert.pem"
  tls_min_version                  = "tls12"
  tls_disable_client_certs         = true
}
```

**Additional context**
For the same scenario here a small performance result.

> Note: 
blue line is for authenticated request (get or post secret)
purple line is login requests

Current Openbao code (2.5.5/2.6.0/2.6.1):
Image

With the fix:
Image

fean5959a:
## Description

the code block for "If we had no error, update our last applied log." may be in the same f.db.Update connection (just above) instead of open a new. I don't know for any impact but for us all seems good and all work.

## Rationale

Resolves: https://github.com/openbao/openbao/issues/3843

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/physical/raft/fsm.go
+++ b/internal/physical/raft/fsm.go
@@ -897,8 +897,7 @@ func (f *FSM) ApplyBatch(logs []*raft.Log) []interface{} {
 		// If we had no error, update our last applied log.
 		if err == nil {
 			if len(logIndex) > 0 {
-				b := tx.Bucket(configBucketName)
-				err = b.Put(latestIndexKey, logIndex)
+				err = configB.Put(latestIndexKey, logIndex)
 				if err != nil {
 					return err
 				}
--- a/internal/physical/raft/fsm.go
+++ b/internal/physical/raft/fsm.go
@@ -894,12 +894,8 @@ func (f *FSM) ApplyBatch(logs []*raft.Log) []interface{} {
 			}
 		}
 
-		return err
-	})
-
-	// If we had no error, update our last applied log.
-	if err == nil {
-		err = f.db.Update(func(tx *bolt.Tx) error {
+		// If we had no error, update our last applied log.
+		if err == nil {
 			if len(logIndex) > 0 {
 				b := tx.Bucket(configBucketName)
 				err = b.Put(latestIndexKey, logIndex)
@@ -907,10 +903,10 @@ func (f *FSM) ApplyBatch(logs []*raft.Log) []interface{} {
 					return err
 				}
 			}
+		}
 
-			return nil
-		})
-	}
+		return err
+	})
 
 	if err != nil {
 		f.logger.Error("failed to store data", "error", err)
--- a/changelog/3844.txt
+++ b/changelog/3844.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+storage/raft: Fix performance regression caused by second transaction during batch application to write last applied log index.
+```


internal/physical/raft/fsm.go
```diff
@@ -894,23 +894,19 @@ func (f *FSM) ApplyBatch(logs []*raft.Log) []any {
...
-	})
-
-	// If we had no error, update our last applied log.
-	if err == nil {
-		err = f.db.Update(func(tx *bolt.Tx) error {
+		// If we had no error, update our last applied log.
+		if err == nil {
 			if len(logIndex) > 0 {
 				b := tx.Bucket(configBucketName)
 				err = b.Put(latestIndexKey, logIndex)
```
satoqz:
```suggestion
				err = configB.Put(latestIndexKey, logIndex)
```

The config bucket is already in scope.

satoqz (commented):
This looks reasonable to me though I'll see what @cipherboy has to say.

Let's also see what tests think!

satoqz:
For reference, https://github.com/openbao/openbao/pull/436 is where this block moved into a separate transaction.

fean5959a:
Who should add changelog entry ? is it me ?
I had an error in checks

cipherboy (dismissed):
@fean5959a This makes sense, thanks!

---

I recall in the original commit I was originally intending for each commit application to be a separate bbolt transaction so they could succeed/fail independently, but it turns out batching at the Raft layer expected them to apply/fail as a unit, which I'm still not convinced is the right behavior.

This gave a bit of performance increase, but with this second transaction gone as well, it should be back to original

sdk/plugin/serve.go
cipherboy, line 100: @fean5959a Not sure why this was changed?

fean5959a: I will fixe that, I don't know why these files changed

fean5959a: Now it's better

cipherboy:
@fean5959a Never mind, i'll push a changelog entry!

satoqz:
> Who should add changelog entry ? is it me ?

@fean5959a yes, the PR author adds changelog entries.

Here's what I'd suggest:

````markdown
```release-note:improvement
physical/raft: Remove an unnecessary transaction on the write path, improving throughput under high load.
```
````

You'd push this as part of this PR at `changelog/3844.txt`. Feel free to adjust wording as you see fit; see https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#changelog-entries for general instructions.

Edit: Seems @cipherboy will take care of it, that works, too :)

satoqz (approved):

satoqz:
changelog checker fails because branch is behind, should be good though.

cipherboy:
Thanks @fean5959a!

PR #3846, opened 2026-08-19
Fix incorrect entity count metrics when `unsafeCrossNamespaceIdentity` flag is present

### LINKED ISSUE #3840: Incorrect entity count metrics when using `unsafe_cross_namespace_identity`
**Describe the bug**
When using `unsafe_cross_namespace_identity = true`, metrics related to entity count report `entity_count × number_of_namespaces` instead of the real count.

This seems to affect:
- API `sys/internal/counters/entities`
- Metrics
    - `identity.num_entities`
    - `identity.entity.count`
    - `identity.entity.alias.count`

**To Reproduce**
```bash
# Starting on a fresh cluster with only the root namespace
# and unsafe_cross_namespace_identity = true

$ bao read sys/internal/counters/entities
Key         Value
---         -----
counters    map[entities:map[total:0]]


# Create 10 entities

$ for i in $(seq 1 10); do bao write -f identity/entity name="e-$i"; done

$ bao read sys/internal/counters/entities
Key         Value
---         -----
counters    map[entities:map[total:10]]


# Create 5 namespaces

$ for i in $(seq 1 5); do bao namespace create "ns-$i"; done


# We should still have 10 total entities
# But this actually returns 10 entities x 6 namespaces (root + 5 new) = 60 total count

$ bao read sys/internal/counters/entities
Key         Value
---         -----
counters    map[entities:map[total:60]]
```

**Expected behavior**
`bao read sys/internal/counters/entities` should return `10` at the end of this test.

**Environment:**
* OpenBao CLI Version (retrieve with `bao version`): 2.5.5

```hcl
unsafe_cross_namespace_identity = true
```

**Additional context**
It looks like this is a regression from https://github.com/openbao/openbao/pull/1715.

In the `CountEntities*` functions in [store_utils.go](https://github.com/openbao/openbao/blob/490d8c19e3bcd03fc5e01d97dd35b4b24504f3cb/internal/vault/identity/store_util.go#L2419), we substitute the root namespace's DB inside a loop to bypass a nil error, but I think this also ends up inflating the entity count.
```go
	i.views.Range(func(uuidRaw, viewsRaw any) bool {
		uuid := uuidRaw.(string)
		views := viewsRaw.(*identityStoreNamespaceView)
		db := views.db
		if db == nil {
			db = rootView.db
		}
```

wslabosz-reply:
## Description
This PR:
- refactors `CountEntitiesX` code introducing shared `iterateEntities` method
- fixes the bug by skipping namespace views when `unsafeCrossNamespaceIdentity` flag is true
- adds `unsafeCrossNamespaceIdentity bool` flag copied from core config
- adjusts `(*IdentityStore).ResetDB` accumulating the errors while skipping the views when running `unsafeCrossNamespaceIdentity` flag present
- fixes `identity` logger never being used by `identityStore`

## Rationale
```
identity.num_entities
identity.entity.count
identity.entity.alias.count
```
metrics reports are useless when running `unsafe_cross_namespace_identity`, due to faulty logic (correctly identified by @vzdai)

Resolves: #3840

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/core_metrics_test.go
+++ b/internal/vault/core_metrics_test.go
@@ -6,7 +6,6 @@ package vault
 import (
 	"context"
 	"errors"
-	"sort"
 	"strings"
 	"testing"
 	"time"
@@ -16,8 +15,8 @@ import (
 	logicalKv "github.com/openbao/openbao/v2/internal/builtin/logical/kv"
 	"github.com/openbao/openbao/v2/internal/helper/namespace"
 	be "github.com/openbao/openbao/v2/internal/vault/backend"
-	ident "github.com/openbao/openbao/v2/internal/vault/identity"
 	"github.com/openbao/openbao/v2/internal/vault/routing"
+	"github.com/stretchr/testify/require"
 )
 
 func TestCoreMetrics_KvSecretGauge(t *testing.T) {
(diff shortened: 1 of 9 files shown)


driif (commented):
Hi @wslabosz-reply, would you also like to add some tests here? It might be useful in the future to ensure we have correct entity count metrics.

internal/vault/core_metrics_test.go
```diff
@@ -303,66 +300,35 @@ func testCoreMetricsEntityGauges(t *testing.T, ctx context.Context, is *ident.Id
...
+	require.Len(t, glv, 4)
+	for _, metric := range glv {
+		require.Equal(t, float32(1.0), metric.Value)
+		require.Len(t, metric.Labels, 3)
+		require.Equal(t, "namespace", metric.Labels[0].Name)
+		require.Contains(t, []string{"root", "ns"}, metric.Labels[0].Value)
+		require.Equal(t, "auth_method", metric.Labels[1].Name)
+		require.Contains(t, []string{"userpass", "approle"}, metric.Labels[1].Value)
+		require.Equal(t, "mount_point", metric.Labels[2].Name)
+		require.Contains(t, []string{"auth/userpass/", "auth/approle/"}, metric.Labels[2].Value)
```
wslabosz-reply, line 332: tbh I hate this test, but sorting and checking the order sounds even worse

internal/vault/identity/store_structs.go
```diff
@@ -95,6 +95,10 @@ type IdentityStore struct {
 	// operated case insensitively
 	disableLowerCasedNames bool
 
+	// unsafeCrossNamespaceIdentity returns back current value of the `unsafe_cross_namespace_identity`
+	// flag indicating whether we use a single global memdb instance for identity
+	unsafeCrossNamespaceIdentity func() bool
```
satoqz: `Core.unsafeCrossNamespaceIdentity` never updates at runtime so I'd prefer if this was a plain bool field that's passed to `NewIdentityStore` at creation, via `IdentityStoreConfig`?

wslabosz-reply: huh, thought it could be reloaded, well that simplifies things

cipherboy:
@wslabosz-reply Do you mind if we defer this to v2.8?

wslabosz-reply:
@cipherboy
not at all

PR #3847, opened 2026-08-19
Optimize Transit's transaction use when encrypting

satoqz:
## Description

We don't need the transaction at all in most cases, and can also end it earlier. This ensures we don't hold the transaction across I/O performed by external key calls, wasting an unnecessary slot in backing permit pools.

If Transit is configured to use itself recursively for encryption via external keys, holding the transaction across recursive HTTP calls otherwise eats up the entire permit pool of the backing storage and blocks transactional operations globally.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/builtin/logical/transit/path_encrypt.go
+++ b/internal/builtin/logical/transit/path_encrypt.go
@@ -296,18 +296,11 @@ func (b *backend) pathEncryptExistenceCheck(ctx context.Context, req *logical.Re
 }
 
 func (b *backend) pathEncryptWrite(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
-	txRollback, err := logical.StartTxStorage(ctx, req)
-	if err != nil {
-		return nil, err
-	}
-	defer txRollback()
-
 	name := d.Get("name").(string)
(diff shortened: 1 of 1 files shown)


internal/builtin/logical/transit/path_encrypt.go
```diff
@@ -420,6 +419,10 @@ func (b *backend) pathEncryptWrite(ctx context.Context, req *logical.Request, d
 	}
 	defer p.Unlock()
 
+	if err := logical.EndTxStorage(ctx, req); err != nil {
+		return nil, err
+	}
```
satoqz, line 424: The nice part is that this no-ops if we haven't started a transaction.

satoqz:
> If Transit is configured to use itself recursively for encryption via external keys, holding the transaction across recursive HTTP calls otherwise eats up the entire permit pool of the backing storage and blocks transactional operations globally.

Notably, this is still bad, just like any other loop one might build back to a Transit operation. With this patch it becomes equally as bad as e.g., signing already is, fully exhausting ports on my system.

wslabosz-reply (approved):
LGTM

cipherboy (approved):
Thanks!

PR #3858, opened 2026-08-20
Fix signature_bits confusion in PKI

satoqz:
## Description

This fixes two related bugs:

1. When using an RSA-based issuer, a role with `use_pss=true` + `signature_bits=0` would issue certs signed via PKCS#1 v1.5 due to missing `default` cases in certutil's `x509.SignatureAlgorithm` selection.
2. PKI would pick a default `signature_bits` for a role based on the role's key type, which may often be coincidentally correct but is generally nonsensical. `signature_bits` applies to the operation performed by the issuer which may use an entirely separate key type. Since issuer information is not trivially available e.g., at role creation, fully remove the defaulting logic (as it only applied to RSA to begin with) and defer to the fix for 1).

Combining 1) and 2), an RSA-based issuer used by a role that has `use_pss=true`, `key_type != "rsa"` and no explicit `signature_bits` would always result in issued certificates being signed via PKCS#1 v1.5 as `signature_bits` would default to `0` unless `key_type == "rsa"`.

To reproduce:

```
bao secrets enable pki
bao write pki/root/generate/internal common_name="Root CA" key_type=rsa key_bits=4096
bao write pki/roles/test use_pss=true key_type=ec key_bits=256 allow_any_name=true
bao write pki/issue/test common_name=openbao ttl=600
```

... then inspect the generated leaf.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3858.txt
+++ b/changelog/3858.txt
@@ -1,0 +1,6 @@
+```release-note:bug
+secrets/pki: Fix `use_pss=true` + `signature_bits=0` on a role incorrectly resulting in certificates signed via PKCS#1 v1.5 instead of correctly using PSS.
+```
+```release-note:bug
+secrets/pki: Don't pick a default value for `signature_bits` based on the key type of the certificate being issued as it applies to the issuer's key, not the certificate's key.
+```
--- a/changelog/3858.txt
+++ b/changelog/3858.txt
@@ -1,0 +1,6 @@
+```release-note:bug
+secrets/pki: Fix `use_pss=true` + `signature_bits=0` on a role resulting in certificates signed via PKCS#1 v1.5 instead of PSS.
+```
+```release-note:bug
+secrets/pki: Don't pick a default value for `signature_bits` based on the key type of the certificate being issued as it applies to the issuer's key, not the certificate's key.
+```
--- a/internal/builtin/logical/pki/backend_test.go
+++ b/internal/builtin/logical/pki/backend_test.go
@@ -4027,7 +4027,7 @@ func TestReadWriteDeleteRoles(t *testing.T) {
 		"client_flag":                        true,
 		"allowed_serial_numbers":             []any{},
 		"generate_lease":                     false,
-		"signature_bits":                     json.Number("256"),
+		"signature_bits":                     json.Number("0"),
 		"use_pss":                            false,
 		"allowed_domains":                    []any{},
 		"allowed_uri_sans_template":          false,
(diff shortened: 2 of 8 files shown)


internal/builtin/logical/pki/backend_test.go
```diff
@@ -5393,9 +5393,9 @@ func TestBackend_Roles_KeySizeRegression(t *testing.T) {
 		/* 11 */ {"rsa", []int{3072}, []int{0, 256, 384, 512}, false, []string{"rsa"}, []int{2048}, true},
 
 		// We should be able to sign with PSS with any CA key type.
-		/* 12 */ {"rsa", []int{0}, []int{0, 256, 384, 512}, true, []string{"rsa"}, []int{2048}, false},
-		/* 13 */ {"ec", []int{0}, []int{0}, true, []string{"ec"}, []int{256}, false},
-		/* 14 */ {"ed25519", []int{0}, []int{0}, true, []string{"ed25519"}, []int{0}, false},
+		/* 12 */ {"any", []int{0}, []int{0, 256, 384, 512}, true, []string{"rsa"}, []int{2048}, false},
+		/* 13 */ {"any", []int{0}, []int{0}, true, []string{"ec"}, []int{256}, false},
+		/* 14 */ {"any", []int{0}, []int{0}, true, []string{"ed25519"}, []int{0}, false},
```
satoqz, line 5410: If we make this change on main, we immediately find the main bug:

```
backend_test.go:5342: key size regression test [0] failed on role key-size-regression-13-keytype-any-keybits-0-signature-bits-0: unexpected signature algorithm; expected RSA-type CA to sign a leaf cert with PSS algorithm; got SHA256-RSA
```

sdk/helper/certutil/helpers.go
```diff
@@ -1197,42 +1113,6 @@ func createCertificateWithTemplate(caSign *CAInfoBundle, evaluationData map[stri
...
-	case elliptic.P224(), elliptic.P256():
-		return x509.ECDSAWithSHA256
-	case elliptic.P384():
-		return x509.ECDSAWithSHA384
-	case elliptic.P521():
-		return x509.ECDSAWithSHA512
-	default:
-		return x509.ECDSAWithSHA256
-	}
-}
```
satoqz, line 1309: `crypto/x509` does exactly the same for us later on so I opted to just remove it. Same goes for selecting `PureEd25519`, and same should go for ML-DSA AFAICT (cc @datosh).

I'm willing to be convinced to keep it for explicitness, but I like deleting unnecessary code, too.

changelog/3858.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:bug
+secrets/pki: Fix `use_pss=true` + `signature_bits=0` on a role resulting in certificates signed via PKCS#1 v1.5 instead of PSS.
```
cipherboy:
```suggestion
secrets/pki: Fix `use_pss=true` + `signature_bits=0` on a role incorrectly resulting in certificates signed via PKCS#1 v1.5 instead of correctly using PSS.
```

wslabosz-reply (dismissed):
removal of code redundancy makes sense to me

cipherboy (dismissed):
One nit on CL

changelog/3858.txt
cipherboy: @satoqz But let's wait to merge until after #3859 just so Fabian doesn't have to deal with conflicts.

satoqz: Oh definitely!

cipherboy (approved):

PR #3859, opened 2026-08-20
Enable ML-DSA support in PKI

### LINKED ISSUE #3756: PKI x.509 - Pure PQC Certificates
Part of #496 

This is about enabling PKI to issue ML-DSA leaf certificates with Go 1.27. Probably two (obvious) portions here:

1. `sdk/helper/certutil` enablement.
2. API updates within the PKI engine itself. We have `rsa`, `ec`, and `ed25519` as key types presently, so I suspect we'd do `mldsa` with a `768`/`1024` key bits value to mirror what we've done for `ec`. I don't know what value `0` should take; probably 768?

datosh:
## Description

I have put the commands I used to end-2-end test this with openssl below. Since I am rather new to the OpenBao code base, and pki side of things in particular, I would appreciate any pointers on this I might have missed or where I didn't follow project conventions.

I went light on the unit tests, just to get some confirmation on direction, before spending too much time here. Happy to work on coverage once the first feedback is in.

~~Note this is still based on #3752, since Go 1.27 was released today (Aug 20), we should be good to merge it and up versions to 1.27.~~

~~Note the CI will mostly break until we have moved to Go 1.27.~~

### Install Go RC and Build OpenBao

```sh
go install golang.org/dl/go1.27rc3@latest && go1.27rc3 download
go1.27rc3 version
go1.27rc3 build -o bin/bao . && ./bin/bao version
```

### Setup Server

```sh
./bin/bao server -dev -dev-root-token-id=root
```

### Setup PKI

```sh
export BAO_ADDR=http://127.0.0.1:8200 BAO_TOKEN=root
bao secrets enable pki
```

#### Generate an ML-DSA-65 root CA

```sh
bao write -field=certificate pki/root/generate/internal \
    common_name=ml-dsa-ca key_type=mldsa key_bits=65 > ca.pem
openssl x509 -in ca.pem -noout -text | grep -E 'Public Key Algorithm|Signature Algorithm'
```

#### Issue a leaf certificate (key generated by OpenBao)

```sh
bao write pki/roles/ml-dsa-role \
    key_type=mldsa key_bits=65 \
    allow_any_name=true enforce_hostnames=false ttl=1h

bao write -field=certificate pki/issue/ml-dsa-role common_name=leaf.test > leaf.pem
openssl verify -CAfile ca.pem leaf.pem
```

#### Sign an external CSR (key generated by OpenSSL)

```sh
# Note: seed-only is required, Go rejects OpenSSL's default
openssl genpkey -algorithm ML-DSA-65 \
    -provparam ml-dsa.output_formats=seed-only -out csr-key.pem
openssl req -new -key csr-key.pem -subj "/CN=csr.test" -out ml-dsa.csr

bao write -field=certificate pki/sign/ml-dsa-role csr=@ml-dsa.csr > signed.pem
openssl verify -CAfile ca.pem signed.pem
```

#### CRL is signed with ML-DSA

```sh
curl -s http://127.0.0.1:8200/v1/pki/crl \
    | openssl crl -inform DER -noout -text | grep 'Signature Algorithm'
```

#### revocation_signature_algorithm must match the issuer's parameter set

```sh
bao write pki/issuer/default revocation_signature_algorithm=MLDSA65
bao read pki/crl/rotate

# Cross parameter set: rejected
bao write pki/issuer/default revocation_signature_algorithm=MLDSA87

# Wrong algorithm family: rejected
bao write pki/issuer/default revocation_signature_algorithm=SHA256WithRSA
```

### role key_bits does not constrain signed CSRs

The role's `key_bits` selects the parameter set for keys generated via `pki/issue`, but `pki/sign` currently accepts CSRs of any ML-DSA parameter

```sh
openssl genpkey -algorithm ML-DSA-87 \
    -provparam ml-dsa.output_formats=seed-only -out csr87-key.pem
openssl req -new -key csr87-key.pem -subj "/CN=csr87.test" -out ml-dsa-87.csr

# Signs successfully despite the role's key_bits=65, see TODO in cert_util.go
bao write pki/sign/ml-dsa-role csr=@ml-dsa-87.csr
```

### What is the right default value?

I chose 65, which corresponds to category 3, but go mldsa.go says: `Most applications should use [MLDSA44].`, which would be category 2.

## Rationale

Resolves: #3756

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/builtin/logical/pki/cert_util.go
+++ b/internal/builtin/logical/pki/cert_util.go
@@ -1072,12 +1072,16 @@ func signCert(b *backend,
 			return nil, nil, errutil.InternalError{Err: fmt.Sprintf("unknown internal error updating default values: %v", err)}
 		}
 
+		switch actualKeyType {
 		// We're using the KeyBits field as a minimum value below, and P-224 is safe
 		// and a previously allowed value. However, the above call defaults
 		// to P-256 as that's a saner default than P-224 (w.r.t. generation), so
 		// override it here to allow 224 as the smallest size we permit.
-		if actualKeyType == "ec" {
+		case "ec":
 			data.role.KeyBits = 224
+		// same reasoning as above for "ec": we allow 44 instead of 65 parameter set
+		case "mldsa":
+			data.role.KeyBits = 44
 		}
 	}
 
@@ -1104,7 +1108,7 @@ func signCert(b *backend,
 				actualKeyBits,
 			)}
 		}
-	case "ec":
+	case "ec", "mldsa":
 		if actualKeyBits < data.role.KeyBits {
 			return nil, nil, errutil.UserError{Err: fmt.Sprintf(
 				"role requires a minimum of a %d-bit key, but CSR's key is %d bits",
@@ -1112,7 +1116,6 @@ func signCert(b *backend,
 				actualKeyBits,
 			)}
 		}
-		// TODO: How do we want to handle mldsa here? Do we require exact parameter set match?
 	}
 
 	creation, warnings, err := generateCreationBundle(b, data, caSign, csr)
--- a/internal/builtin/logical/pki/cert_util.go
+++ b/internal/builtin/logical/pki/cert_util.go
@@ -7,6 +7,7 @@ import (
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/ed25519"
+	"crypto/mldsa"
 	"crypto/rsa"
 	"crypto/x509"
 	"crypto/x509/pkix"
@@ -982,6 +983,25 @@ func signCert(b *backend,
 
 		actualKeyType = "ed25519"
 		actualKeyBits = 0
+	case "mldsa":
+		if csr.PublicKeyAlgorithm != x509.MLDSA {
+			return nil, nil, errutil.UserError{Err: fmt.Sprintf(
+				"role requires keys of type %s",
+				data.role.KeyType,
+			)}
+		}
+
+		pubkey, ok := csr.PublicKey.(*mldsa.PublicKey)
+		if !ok {
+			return nil, nil, errutil.UserError{Err: "could not parse CSR's public key"}
+		}
+
+		actualKeyType = "mldsa"
+		label := certutil.GetMLDSAParameterSetLabel(pubkey)
+		if label == -1 {
+			return nil, nil, errutil.UserError{Err: fmt.Sprintf("Unknown key size for ML-DSA: %v", pubkey.Parameters().String())}
+		}
+		actualKeyBits = label
 	case "any":
 		// We need to compute the actual key type and key bits, to correctly
 		// validate minimums and SignatureBits below.
@@ -1013,6 +1033,18 @@ func signCert(b *backend,
 
 			actualKeyType = "ed25519"
 			actualKeyBits = 0
+		case x509.MLDSA:
+			pubkey, ok := csr.PublicKey.(*mldsa.PublicKey)
+			if !ok {
+				return nil, nil, errutil.UserError{Err: "could not parse CSR's public key"}
+			}
+
+			actualKeyType = "mldsa"
+			label := certutil.GetMLDSAParameterSetLabel(pubkey)
+			if label == -1 {
+				return nil, nil, errutil.UserError{Err: fmt.Sprintf("Unknown key size for ML-DSA: %v", pubkey.Parameters().String())}
+			}
+			actualKeyBits = label
 		default:
 			return nil, nil, errutil.UserError{Err: "Unknown key type in CSR: " + csr.PublicKeyAlgorithm.String()}
 		}
@@ -1080,6 +1112,7 @@ func signCert(b *backend,
 				actualKeyBits,
 			)}
 		}
+		// TODO: How do we want to handle mldsa here? Do we require exact parameter set match?
 	}
 
 	creation, warnings, err := generateCreationBundle(b, data, caSign, csr)
@@ -1880,7 +1913,7 @@ func convertRespToPKCS8(resp *logical.Response) error {
 		signer, err = x509.ParsePKCS1PrivateKey(keyData)
 	case certutil.ECPrivateKey:
 		signer, err = x509.ParseECPrivateKey(keyData)
-	case certutil.Ed25519PrivateKey:
+	case certutil.Ed25519PrivateKey, certutil.MLDSAPrivateKey:
 		k, err := x509.ParsePKCS8PrivateKey(keyData)
 		if err != nil {
 			return fmt.Errorf("error converting response to pkcs8: error parsing previous key: %w", err)
--- a/internal/builtin/logical/pki/fields.go
+++ b/internal/builtin/logical/pki/fields.go
@@ -330,7 +330,7 @@ the private key!`,
 		Description: `The number of bits to use. Allowed values are
 0 (universal default); with rsa key_type: 2048 (default), 3072, or
 4096; with ec key_type: 224, 256 (default), 384, or 521; ignored with
-ed25519.`,
+ed25519; with mldsa key_type: 44, 65 (default), or 87.`,
 		DisplayAttrs: &framework.DisplayAttributes{
 			Value: 0,
 		},
@@ -358,9 +358,9 @@ RSA key-type issuer. Defaults to false.`,
 	fields["key_type"] = &framework.FieldSchema{
 		Type:    framework.TypeString,
 		Default: "rsa",
-		Description: `The type of key to use; defaults to RSA. "rsa"
-"ec" and "ed25519" are the only valid values.`,
-		AllowedValues: []any{"rsa", "ec", "ed25519"},
+		Description: `The type of key to use; defaults to RSA. "rsa",
+"ec", "ed25519", and "mldsa" are the only valid values.`,
+		AllowedValues: []any{"rsa", "ec", "ed25519", "mldsa"},
 		DisplayAttrs: &framework.DisplayAttributes{
 			Value: "rsa",
 		},
@@ -509,16 +509,16 @@ in most operational scenarios).`,
 	fields["tidy_acme"] = &framework.FieldSchema{
 		Type: framework.TypeBool,
 		Description: `Set to true to enable tidying ACME accounts,
-orders and authorizations.  ACME orders are tidied (deleted) 
+orders and authorizations.  ACME orders are tidied (deleted)
 safety_buffer after the certificate associated with them expires,
-or after the order and relevant authorizations have expired if no 
-certificate was produced.  Authorizations are tidied with the 
+or after the order and relevant authorizations have expired if no
+certificate was produced.  Authorizations are tidied with the
 corresponding order.
 
 When a valid ACME Account is at least acme_account_safety_buffer
 old, and has no remaining orders associated with it, the account is
-marked as revoked.  After another acme_account_safety_buffer has 
-passed from the revocation or deactivation date, a revoked or 
+marked as revoked.  After another acme_account_safety_buffer has
+passed from the revocation or deactivation date, a revoked or
 deactivated ACME account is deleted.`,
 		Default: false,
 	}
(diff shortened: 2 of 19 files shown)


satoqz:
@datosh I've merged #3752 for you so you can rebase -- it's also entertaining that I've opened https://github.com/openbao/openbao/pull/3858 at the (exact) same time, I think we'll have conflicts in certutil but nothing big.

cipherboy:
@datosh I think this part is expected:

> The role's `key_bits` selects the parameter set for keys generated via `pki/issue`, but `pki/sign` currently accepts CSRs of any ML-DSA parameter.

We're constraining/defaulting issuance of certificates (it does also support `any` now as a key type), more so than leaf certificates. CEL would give you the tools necessary to constrain bits of a leaf if that's truly desired.

edit: though I do see it is slightly weird between `key_type` and `key_bits` :thinking:

datosh:
> I've merged #3752 for you so you can rebase

Thanks, @satoqz! Rebase to main is done

internal/builtin/logical/pki/cert_util.go
```diff
@@ -1080,6 +1112,7 @@ func signCert(b *backend,
 				actualKeyBits,
 			)}
 		}
+		// TODO: How do we want to handle mldsa here? Do we require exact parameter set match?
```
cipherboy: I think similar to EC, it'd be fine to do:

```
	case "mldsa":
		if actualKeyBits < data.role.KeyBits {
			return nil, nil, errutil.UserError{Err: fmt.Sprintf(
				"role requires a minimum of a %d-bit key, but CSR's key is %d bits",
				data.role.KeyBits,
				actualKeyBits,
			)}
		}
```

and be fine with that? I realize they're not "bits" but parameter sets and NIST published a rough approximation to bits/security levels to parameter sets.

internal/builtin/logical/pki/fields.go
```diff
@@ -509,16 +509,16 @@ in most operational scenarios).`,
 	fields["tidy_acme"] = &framework.FieldSchema{
 		Type: framework.TypeBool,
 		Description: `Set to true to enable tidying ACME accounts,
-orders and authorizations.  ACME orders are tidied (deleted) 
+orders and authorizations.  ACME orders are tidied (deleted)
 safety_buffer after the certificate associated with them expires,
-or after the order and relevant authorizations have expired if no 
-certificate was produced.  Authorizations are tidied with the 
+or after the order and relevant authorizations have expired if no
```
cipherboy, line 514: @datosh Sadly these spaces are structural; I think the wrapper does not understand newlines iirc.

cipherboy: Given how many people stumble across this, I'm fine making them prefixed spaces instead if that helps.

internal/builtin/logical/pki/storage.go
```diff
@@ -610,6 +611,25 @@ func (i issuerEntry) CanMaybeSignWithAlgo(algo x509.SignatureAlgorithm) error {
...
+				return nil
+			}
+		case mldsa.MLDSA65():
+			if algo == x509.MLDSA65 {
+				return nil
+			}
+		case mldsa.MLDSA87():
+			if algo == x509.MLDSA87 {
+				return nil
+			}
```
cipherboy, line 631: nit: I don't expect them to add more ML-DSA parameters, but just so we don't have to chase down a lot of places without error messages, mind adding one here? ty!

sdk/helper/certutil/certutil_test.go
```diff
@@ -523,6 +524,13 @@ func TestGetPublicKeySize(t *testing.T) {
 		t.Fatal("unexpected ed25519 key size")
 	}
 	// Skipping DSA as too slow
```
cipherboy: Comment can probably be dropped, we've not supported DSA in forever.

sdk/helper/certutil/helpers.go
```diff
@@ -38,8 +39,9 @@ const rsaMinimumSecureKeySize = 2048
 
 // Mapping of key types to default key lengths
 var defaultAlgorithmKeyBits = map[string]int{
-	"rsa": 2048,
-	"ec":  256,
+	"rsa":   2048,
+	"ec":    256,
+	"mldsa": 65,
```
cipherboy, line 44: @datosh @satoqz I do wonder in general how we feel about 65. I think it is interesting that 2048/256 roughly align at or below 128 bits of symmetric security, which I think 44 was aiming for. The savings on signature sizes seems rather significant. 

Thoughts from others?

sdk/helper/certutil/helpers.go
cipherboy, line 1743: @datosh This function needs an update as well.

cipherboy (commented):
Most of the docs will need to be updated, but I don't think that'll be too bad, but monotonous. The commentary about PSS support under `considerations.mdx` will need to be updated for ML-DSA as well. 🤔 Maybe just add a new header with OCSP in general. 

There's some tests under `internal/builtin/logical/pkiext` that we should consider adjusting, but happy for that to be in a follow-up PR.

sdk/helper/certutil/helpers.go
satoqz: I mean, yes, but we also default to *RSA 2048* here, which has aged somewhat poorly and isn't what you'd make the default if you re-introduced RSA today? 65 looks to be the reasonable & non-controversial default.

Comparing sizes, I think it's primarily the private key that's a little bigger (2560 vs 4032 bytes), not the signature size (2420 vs 3309 bytes). This is in contrast to 65 vs 87 where signature size grows more than private key size.

sdk/helper/certutil/helpers.go
datosh: Thanks, that must have come in with the rebase, fixed!

internal/builtin/logical/pki/fields.go
datosh: Could you point me in the direction where this breaks? I have checked

```
› curl -s -H "X-Vault-Token: root" 'http://127.0.0.1:8200/v1/pki?help=1' \ 
    | jq -r '.openapi.components.schemas.PkiTidyRequest.properties.tidy_acme.description' | cat -e
Set to true to enable tidying ACME accounts, orders and authorizations. ACME orders are tidied (deleted) safety_buffer after the certificate associated with them expires, or after the order and relevant authorizations have expired if no certificate was produced. Authorizations are tidied with the corresponding order. When a valid ACME Account is at least acme_account_safety_buffer old, and has no remaining orders associated with it, the account is marked as revoked. After another acme_account_safety_buffer has passed from the revocation or deactivation date, a revoked or deactivated ACME account is deleted.$
```

and

```
› ./bin/bao path-help pki/tidy | grep -A9 tidy_acme               
    tidy_acme (bool)

        Set to true to enable tidying ACME accounts,
        orders and authorizations.  ACME orders are tidied (deleted)
        safety_buffer after the certificate associated with them expires,
        or after the order and relevant authorizations have expired if no
        certificate was produced.  Authorizations are tidied with the
        corresponding order.
        
        When a valid ACME Account is at least acme_account_safety_buffer
```

which both seem fine to me, but maybe it is a platform dependent thing or I'm looking in the wrong place?

cipherboy:
@datosh If you rebase on top of latest `main`, you should pick up Go 1.27 and the test suite should be much happier :-)

internal/builtin/logical/pki/fields.go
cipherboy: @datosh Ah, I was thinking the second was the issue. I guess I've been mistaken this whole time, thanks! 

(I recall being told this once at Hashi, checking it, it being broken, fixing it, and passing along what apparently has been bad advice since.)

sdk/helper/certutil/helpers.go
cipherboy: @satoqz If I were introducing just RSA for the first time, knowing how much of a performance loss it is, I'd set the global algorithm default to ed25519 or p256 and default RSA size to 2048. 

The latest CA/BF guidelines are still at 2048 bits: https://github.com/cabforum/servercert/blob/main/docs/BR.md#615-key-sizes

Unless I missed it, https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf and https://www.ietf.org/archive/id/draft-connolly-cfrg-ml-dsa-security-considerations-02.html both lack a recommendation, though a lot of the PQC cottage industry seems to recommend -65, so we'll go that route.  :shrug: 

As far as compatibility goes, for better or worse, the lowest parameter sets for an algorithm tend to be most widely adopted in the PKI space.

satoqz: Looking at it again, I think defaulting to 44 is reasonable given that is what is recommended in the standard library.

datosh: Happy to follow your guidance, I don't think there is a wrong/right answer here. 

EC-256 would be NIST security level 3, whereas RSA-2048 is NIST security level 2.

For ML-DSA, the 44 parameter set translates to level 2, and 65 to level 3. Since folks will mostly use it in the context of TLS, and ML-KEM-768 translates to security level 3, I went with ML-DSA-65 as the default. 

Cloudflare on the other hand defaults to ML-DSA-44 in their PQC origin setup guide: https://developers.cloudflare.com/ssl/post-quantum-cryptography/pqc-to-origin/#generate-an-ml-dsa-certificate-authority-and-leaf-certificate

cipherboy (approved):
Thanks @datosh! I'll pick up changelog + docs in a follow-up PR and revise the default value to 44 then too. 

(It'll also save me from having to get another vertical monitor to view certs :laughing: )

sdk/helper/certutil/types.go
```diff
@@ -64,6 +65,7 @@ const (
 	ECPrivateKey       PrivateKeyType = "ec"
 	Ed25519PrivateKey  PrivateKeyType = "ed25519"
 	ExternalPrivateKey PrivateKeyType = "external-key"
+	MLDSAPrivateKey    PrivateKeyType = "mldsa"
```
satoqz, line 68: Nit for later fixup: External keys should stay last in the list.

internal/command/pki_reissue_intermediate.go
satoqz, line 276: Saw this above the diff in this file, might want to name ML-DSA algos here, though feels a bit like a chore. I don't think the `-1` has any effect, but we do explicitly handle Ed25519 and map it to `0`.

We also return non-zero values for ECDSA variants which have been ignored on the backend for a while, but that's a different story.

satoqz, line 259: I wonder if this can just use certutil, but probably a good idea to update?

internal/builtin/logical/pki/path_roles.go
```diff
@@ -262,7 +262,7 @@ protection use. Defaults to false. See also RFC 5280 Section 4.2.1.12.`,
 			Description: `The number of bits to use. Allowed values are
 0 (universal default); with rsa key_type: 2048 (default), 3072, or
 4096; with ec key_type: 224, 256 (default), 384, or 521; ignored with
-ed25519.`,
+ed25519; with mldsa key_type: 44, 65 (default), or 87.`,
```
satoqz, line 265: Just leaving a comment to note that if we do update the default, we'll also need to update the doc strings in various places.

satoqz (approved):
Looks great, thanks @datosh!

internal/builtin/logical/pki/path_roles.go
cipherboy: @satoqz @datosh I just saw: https://docs.aws.amazon.com/kms/latest/developerguide/supported-algorithms.html -- doesn't support 44 so I guess maybe we keep 65 as the default.

satoqz: I think the pages are conflicting? See https://docs.aws.amazon.com/kms/latest/developerguide/symm-asymm-choose-key-spec.html which has ML-DSA 44 and RSA 4096, while the page you linked doesn't.

Also:

> They should not be considered to be an exhaustive list of all cryptography options available in AWS.

"support" is very unclear wording (at least if you're AWS, apparently). Not sure how that'd influence our decision 🤷

PR #3863, opened 2026-08-20
Fix unreadable backed up keys regression

### LINKED ISSUE #3857: sys/rotate/recovery/backup fails after 2.5.x → 2.6.x upgrade: recovery-key backup created on 2.5.x is no longer readable
**Describe the bug**
After an in-place upgrade from 2.5.x to 2.6.x, a recovery-key backup that was created under 2.5.x can no longer be read. `sys/rotate/recovery/backup` returns a decryption error, even though the instance is otherwise healthy (unsealed, serving requests normally):

```
$ bao read sys/rotate/recovery/backup
Code: 400
Errors:
  unable to look up backed-up keys: error getting keys from backup:
  no decryption key available for term 2065845871
```

**To Reproduce**
1.  Start with OpenBao v2.5.2 with an auto-unseal seal (we use awskms), initialize the cluster.
    Use self-init, as described under https://openbao.org/docs/api/system/rotate/#rotate-unsealrecovery-keys and

    >As a special case, when using [declarative self-initialization with an auto-unseal mechanism](https://openbao.org/docs/configuration/self-init/), no recovery shares are provisioned by default. In this case, sys/rotate/recovery/init will immediately yield new key shares aligning with the specified configuration, as no existing key shares exist.`

2. Generate recovery keys with a backup, e.g. sys/rotate/recovery/init with backup=true (and PGP keys). Confirm the backup is readable:

    `bao read sys/rotate/recovery/backup`

3. Upgrade the same cluster in place to 2.6.2 (raft storage retained).
4. Read the backup again:

    ```
    bao read sys/rotate/recovery/backup
    # Code: 400  unable to look up backed-up keys: error getting keys from backup:
    #           no decryption key available for term 2065845871
    ```

**Expected behavior**
A recovery-key backup created before the upgrade should remain readable via sys/rotate/recovery/backup after upgrading to 2.6.2 (or the upgrade should migrate it transparently). 

**Environment:**
* OpenBao Server Version (bao status): v2.6.2, upgraded in place from v2.5.2
* OpenBao CLI Version (bao version): v2.6.2
* Server Operating System/Architecture: Linux, containerized on Kubernetes; storage: raft; auto-unseal: awskms (pluggable kms seal)

```hcl
ui = true
cluster_name = "openbao-canary"
pid_file = "/tmp/openbao.pid"

# Log format for server logs only. Audit logs are configured separately in the audit devices below.
# Server logs are sent to stdout by default.
log_format = "json"

user_lockout "all" {
  lockout_threshold     = 5
  lockout_duration      = "10m"
  lockout_counter_reset = "10m"
}

# Enable Prometheus metrics endpoint
telemetry {
   prometheus_retention_time = "24h"
   disable_hostname = true
}

plugin "auth" "aws" {
  image       = <registry>/openbao-plugin-auth-aws"
  version     = "v0.1.1"
  binary_name = "openbao-plugin-auth-aws"
  sha256sum   = "7a77057e62973c1aae6035f52110e3302605a47b622756d954915b7b55eca10c"
}

# KMS seal plugin backing the seal "awskms" stanza below
plugin "kms" "awskms" {
  image       = "<registry>/openbao-plugin-kms-aws"
  version     = "v0.1.0"
  binary_name = "openbao-plugin-kms-aws"
  sha256sum   = "fa332cf1863a948d4b166ed9acb7da1608fb1dffbde4425626d787424fda5a05"
}

plugin_download_behavior = "fail"
plugin_auto_download = true
plugin_auto_register = true
# Max extracted plugin binary size in bytes (default 536870912 = 512 MiB).
plugin_download_max_size = 536870912

listener "tcp" {
  tls_disable = 0
  address = "[::]:8200"
  cluster_address = "[::]:8201"
  tls_cert_file = "/openbao/userconfig/listener-cert/tls.crt"
  tls_key_file  = "/openbao/userconfig/listener-cert/tls.key"

  telemetry {
  }
}

# Use Raft storage to keep the deployment self-contained and avoid external dependencies.
# Raft provides strong consistency and is the recommended HA storage mode for OpenBao.
# Note: This will use local disk storage on each node, so ensure that the underlying storage class provides durable storage (e.g. SSD-backed) 
# and that the data is regularly backed up to avoid data loss during node failures.
storage "raft" {
  path = "/openbao/data"

   retry_join {
    leader_api_addr         = "https://{{ include "openbao.fullname" . }}-0.{{ include "openbao.fullname" . }}-internal:8200"
    leader_tls_servername   = "<custom-domain>"
  }

  retry_join {
    leader_api_addr         = "https://{{ include "openbao.fullname" . }}-1.{{ include "openbao.fullname" . }}-internal:8200"
    leader_tls_servername   = "<custom-domain>"
  }

  retry_join {
    leader_api_addr         = "https://{{ include "openbao.fullname" . }}-2.{{ include "openbao.fullname" . }}-internal:8200"
    leader_tls_servername   = "<custom-domain>"
  }
}

# Instruct OpenBao to use Kubernetes service registration for HA cluster membership and discovery.
service_registration "kubernetes" {}

# https://openbao.org/docs/concepts/seal/#auto-unseal
seal "awskms" {
  region     = "eu-central-1"
  kms_key_id = "<key-arn>:alias/openbao-canary"
}

# An audit device that writes to stdout for real-time monitoring and log aggregation in loki.
audit "file" "to-stdout" {
  description = "This audit device should never fail."
  options {
    file_path = "/dev/stdout"
    format = "json"
    log_raw = "false"
  }
}

# =====================================================================
# Self-initialization stanza — executed ONCE on first startup of pod-0.
# The root token is auto-revoked after all requests complete.
# =====================================================================

# --- 0. Auth methods -----------------------------------------------
initialize "auth" {
# Root namespace K8s auth — used only by the bootstrap configuration job.
request "enable-kubernetes-auth" {
operation = "update"
path      = "sys/auth/kubernetes"
data = {
  type        = "kubernetes"
  description = "Kubernetes service account authentication"
}
}

request "configure-kubernetes-auth" {
operation = "update"
path      = "auth/kubernetes/config"
data = {
  kubernetes_host = "https://kubernetes.default.svc"
}
}

    ...
```

**Additional context**
none

wslabosz-reply:
## Description
This PR:
- Changed backed up unseal shares write and store methods, now using parent's barrier (or physical in case of root)
- Introduced fallback read of the backed up unseal shares 
   - for root namespace we are checking root namespace barrier, then falling back to physical
   - for other namespaces we are checking namespace barrier, then namespace parent's barrier
- Added unconditional forwarding for backed up unseal shares read requests due to introduced writes in the handler
- Adjusted sys/raw read call for backed up unseal shares, allowing for reads using parent's barrier (or physical in root namespace case)

## Rationale
With v2.6+ and introduction of per-namespace seals, which for feature parity's with root namespace (global seal) required support of storing pgp encrypted unseal shares in the OpenBao after (recovery/barrier) key rotation. This introduced a regression within how the encrypted unseal shares were stored and read. Previously they were stored as-is directly to the OpenBao storage without additional layer of encryption (which made sense, as they are already encrypted with pgp shares). That was changed to the namespace barrier, which introduced regression when upgrading from versions before v2.6. This PR aims at fixing that gap through fallback reads.

A change to the `sys/raw` endpoints require a key rotation of the v2.6 unseal shares saved to the storage before this change makes it to the release. Reads through the `sys/rotate/<recovery|barrier>/backup` endpoint are going to work without rotation requirement.

Resolves: #3857

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3863.txt
+++ b/changelog/3863.txt
@@ -1,0 +1,7 @@
+```release-note:improvement
+sys/raw: Allow reading backed up unseal or recovery shares.
+```
+
+```release-note:bug
+sys/rotate: Fix existing unseal/recovery share backups going unreadable after upgrading to v2.6.x from an earlier version.
+```
--- a/internal/vault/rotate.go
+++ b/internal/vault/rotate.go
@@ -18,6 +18,7 @@ import (
 	"github.com/openbao/openbao/sdk/v2/helper/jsonutil"
 	"github.com/openbao/openbao/sdk/v2/helper/shamir"
 	"github.com/openbao/openbao/sdk/v2/logical"
+	"github.com/openbao/openbao/sdk/v2/physical"
 	"github.com/openbao/openbao/v2/internal/helper/namespace"
 	"github.com/openbao/openbao/v2/internal/helper/pgpkeys"
 	"github.com/openbao/openbao/v2/internal/vault/barrier"
@@ -656,8 +657,20 @@ func (sm *SealManager) pgpEncryptShares(ctx context.Context, ns *namespace.Names
 			entry.Key = coreRecoveryUnsealKeysBackupPath
 		}
 
-		barrier := sm.namespaceBarrier(ns.Path)
-		if err = barrier.Put(ctx, entry); err != nil {
+		parentPath, hasParent := ns.ParentPath()
+		if !hasParent {
+			// If we don't have a parent - meaning we are in root namespace
+			// write the backed up keys through physical.
+			err = sm.core.physical.Put(ctx, &physical.Entry{
+				Key:   entry.Key,
+				Value: buf,
+			})
+		} else {
+			barrier := sm.namespaceBarrierByLongestPrefix(parentPath)
+			err = barrier.Put(ctx, entry)
+		}
+
+		if err != nil {
 			sm.logger.Error("failed to save unseal key backup", "error", err)
 			return nil, fmt.Errorf("failed to save unseal key backup: %w", err)
 		}
@@ -802,7 +815,7 @@ func (sm *SealManager) RestartRotationVerification(nsUUID string, recovery bool)
 
 // RetrieveRotationBackup is used to retrieve any backed-up PGP-encrypted
 // unseal keys.
-func (sm *SealManager) RetrieveRotationBackup(ctx context.Context, nsPath string, recovery bool) (*RekeyBackup, error) {
+func (sm *SealManager) RetrieveRotationBackup(ctx context.Context, ns *namespace.Namespace, recovery bool) (*RekeyBackup, error) {
 	sm.lock.Lock()
 	defer sm.lock.Unlock()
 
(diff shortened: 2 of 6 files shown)


internal/vault/rotate.go
```diff
@@ -811,8 +824,41 @@ func (sm *SealManager) RetrieveRotationBackup(ctx context.Context, nsPath string
...
+		switch {
+		case err != nil:
+			entry, err = sm.namespaceBarrierByLongestPrefix(ns.Path).Get(ctx, path)
+			if entry != nil && err == nil {
+				defer func() {
+					if err := barrier.Put(ctx, entry); err != nil {
+						sm.logger.Error("Failed to overwrite backed up keys, please rotate the keys immediately after.")
+					}
+					sm.logger.Info("Overwritten backed up keys.")
+				}()
```
satoqz: Could we hit this on a standby?

internal/vault/rotate.go
```diff
@@ -811,8 +824,41 @@ func (sm *SealManager) RetrieveRotationBackup(ctx context.Context, nsPath string
...
+		case err != nil:
+			var pEntry *physical.Entry
+			pEntry, err = sm.core.physical.Get(ctx, path)
+			entry = &logical.StorageEntry{
+				Key:   pEntry.Key,
+				Value: pEntry.Value,
+			}
+		case entry == nil:
+			return nil, nil
+		}
```
satoqz, line 851: Don't you also want to fix it up here and re-write using physical?

satoqz: I also wonder if we can use `StorageAccess` to make this cleaner.

internal/vault/rotate.go
```diff
@@ -811,8 +824,41 @@ func (sm *SealManager) RetrieveRotationBackup(ctx context.Context, nsPath string
...
+	} else {
+		barrier := sm.namespaceBarrierByLongestPrefix(parentPath)
+		entry, err = barrier.Get(ctx, path)
+		switch {
+		case err != nil:
+			entry, err = sm.namespaceBarrierByLongestPrefix(ns.Path).Get(ctx, path)
+			if entry != nil && err == nil {
```
satoqz: A comment as to why this fallback exists would be very good to have.

changelog/3863.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:improvement
+sys/raw: Allow backed up unseal shares reads.
```
satoqz:
```suggestion
sys/raw: Allow reading backed up unseal or recovery shares.
```

changelog/3863.txt
```diff
@@ -0,0 +1,7 @@
...
+```
+
+```release-note:bug
+sys/rotate: Fixed existing backed up unseal shares unreadable after upgrade to OpenBao v2.6+.
```
satoqz:
```suggestion
sys/rotate: Fix existing unseal/recovery share backups going unreadable after upgrading to v2.6.x from an earlier version.
```

internal/vault/rotate.go
wslabosz-reply: > I also wonder if we can use StorageAccess to make this cleaner.

we can

> Don't you also want to fix it up here and re-write using physical?

will do

internal/vault/rotate.go
wslabosz-reply: yes, have to adjust.

cipherboy (approved):
I'm fine with this, but as talking with @wslabosz-reply, maybe we should think about removing this altogether, but another RFC for another time. 😆

satoqz (approved):

PR #3865, opened 2026-08-20
fix: UpdateDNPassword should not discard explicit DN when userattr=userPrincipalName

### LINKED ISSUE #3862: Static-role rotation in LDAP secret engine fails with "No Such Object" when `schema=ad` is used and when DN is explicitly set
**Describe the bug**
Static-role rotation fails with "No Such Object" when `schema=ad` and a DN is explicitly set. It seems that `UpdateDNPassword` discards the DN

**To Reproduce**
1. `bao secrets enable -path=ldap-rotation ldap`
2. `bao write ldap-rotation/config binddn=<bind-dn> bindpass=<pw> url=ldaps://<dc>:636 schema=ad` (note: `userattr` left at its schema default, `userPrincipalName`)
3. `bao write ldap-rotation/static-role/testrole dn='CN=svc-example,OU=...,DC=...,DC=...' username=svc-example rotation_period=720h → fails with `LDAP Result Code 32 "No Such Object" ... best match of: ''`

Note: when skip_import_rotation=true parameter is added to step 3 the write succeeds.

**Expected behavior**
When a static role has an explicit `dn`, `UpdateDNPassword` should search/modify directly against that DN (`ScopeBaseObject`, as happens for any other `userattr` value, like for example `dn`), not attempt UPN-based re-resolution and discard the caller's DN instead.

**Environment:**
bao status
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    5
Threshold       3
Version         2.6.2
Commit Date     2026-08-18T15:48:19Z
Storage Type    file
Cluster Name    vault-cluster-d3f85999
Cluster ID      ebb3ae69-04e2-ed46-24ce-0090d0a930cc
HA Enabled      false

bao version
OpenBao v2.6.2 (dd9c19c37a878cf4a81b18efb8d6f0599c7da923), committed 2026-08-18T15:48:19Z
NOTE: I have tested that initially on 2.6.0, then performed an upgrade to latest one to see if the issue persists

Ubuntu 24.04.4 LTS

```hcl
# Full configuration options can be found at https://github.com/openbao/openbao/tree/main/website/content/docs/configuration

ui = true

storage "file" {
  path = "/opt/openbao/data"
}

# HTTP listener
#listener "tcp" {
#  address = "< redacted >:8200"
#  tls_disable = 1
#}

# HTTPS listener
listener "tcp" {
  address       = "0.0.0.0:8200"
  tls_cert_file = "/opt/openbao/tls/cert.pem"
  tls_key_file  = "/opt/openbao/tls/key.pem"
  tls_disable = 0
  tls_min_version = "tls13"
}

api_addr = "https://< redacted >:8200"
cluster_addr = "https://< redacted >:8201"
default_lease_ttl = "3h"
max_lease_ttl = "3h"

audit "syslog" {
  type = "syslog"
  path = "syslog"
  options = {
    tag      = "openbao"
    facility = "AUTH"
  }
```

**Additional context**
Creating a static role with `dn=<full DN>` succeeds, and reading the role back (`GET static-role/:name`) correctly shows the stored `dn`. However, both:
- automatic rotation on creation (unless `skip_import_rotation=true` - this succeeds), and
- manual `POST rotate-role/:name`
fail identically with:
```
unable to finish rotating credentials; ... LDAP Result Code 32 "No Such Object":
0000208D: NameErr: DSID-03100223, problem 2001 (NO_OBJECT), data 0, best match of: ''
```
even though the target DN is verified correct and reachable (I have checked that using `ldapsearch` and `ldapmodify` against the same DN, same bind credentials)

waterWang:
## What

Fixes #3862

When a static role has an explicitly configured DN, `UpdateDNPassword` must search and modify directly against that DN using `ScopeBaseObject` instead of attempting UPN-based re-resolution, which would construct a malformed UPN from the DN value (e.g. `CN=svc-example,OU=Users,DC=example,DC=net@domain`) and overwrite `dn` with `conf.UserDN`, causing a `"No Such Object"` failure.

## Root cause

In `client.go`, the `UpdateDNPassword` function enters a UPN re-resolution branch when `userattr=userPrincipalName` (AD schema default). This branch:
1. Constructs `bindUser = fmt.Sprintf("%s@%s", ldap.EscapeFilter(dn), conf.UPNDomain)` — treating the DN as a UPN prefix
2. Overwrites `dn = conf.UserDN` — discarding the caller-provided explicit DN

When the caller passes a full DN (e.g. `CN=svc-example,OU=Users,DC=example,DC=net`), the filter becomes `userPrincipalName=CN=svc-example,...@example.com` which matches nothing, and the search base is changed to `conf.UserDN` instead of the explicit DN.

## Fix

Use `ldap.ParseDN` to detect whether the `dn` argument is a fully-qualified LDAP DN. If it is, skip the UPN re-resolution and use `ScopeBaseObject` directly against the provided DN. If it is a plain username (ParseDN fails), keep the existing UPN re-resolution behavior.

## Test

Added `Test_UpdateDNPassword_AD_UserPrincipalName_ExplicitDN` — verifies that an explicit DN with `userattr=userPrincipalName` (AD default) results in:
- BaseDN: the explicit DN
- Scope: `ScopeBaseObject`
- Filter: `(objectClass=*)`

Existing tests `Test_UpdateDNPassword_AD_UserPrincipalName` and `Test_UpdateDNPassword_AD_DN` remain unchanged.

Signed-off-by: waterWang <waterwang@proton.me>

### DIFF
--- a/internal/builtin/logical/openldap/client.go
+++ b/internal/builtin/logical/openldap/client.go
@@ -48,10 +48,17 @@ func (c *Client) UpdateDNPassword(conf *client.Config, dn string, newPassword st
 	}
 
 	if field == client.FieldRegistry.UserPrincipalName {
-		scope = ldap.ScopeWholeSubtree
-		bindUser := fmt.Sprintf("%s@%s", ldap.EscapeFilter(dn), conf.UPNDomain)
-		filters[field] = []string{bindUser}
-		dn = conf.UserDN
+		// If the provided dn is a fully-qualified LDAP DN (e.g. an explicitly
+		// configured static-role DN), operate on that object directly with a
+		// base-object search instead of re-resolving it via the userPrincipalName
+		// attribute. Re-resolution would construct a malformed UPN from the DN and
+		// discard the caller-provided DN, causing a "No Such Object" failure.
+		if _, parseErr := ldap.ParseDN(dn); parseErr != nil {
+			scope = ldap.ScopeWholeSubtree
+			bindUser := fmt.Sprintf("%s@%s", ldap.EscapeFilter(dn), conf.UPNDomain)
+			filters[field] = []string{bindUser}
+			dn = conf.UserDN
+		}
 	}
 
 	newValues, err := client.GetSchemaFieldRegistry(conf.Schema, newPassword)
--- a/internal/builtin/logical/openldap/client_test.go
+++ b/internal/builtin/logical/openldap/client_test.go
@@ -110,3 +110,49 @@ func Test_UpdateDNPassword_AD_DN(t *testing.T) {
 	err = c.UpdateDNPassword(config, "CN=Bob,CN=Users,DC=example,DC=net", newPassword)
 	assert.NoError(t, err)
 }
+
+// UpdateDNPassword with an explicit DN while the UserAttr is "userPrincipalName"
+// (the AD schema default). The explicit DN must be used directly with a
+// base-object search instead of being discarded in favor of UPN re-resolution.
+func Test_UpdateDNPassword_AD_UserPrincipalName_ExplicitDN(t *testing.T) {
+	newPassword := "newpassword"
+	conn := &ldapifc.FakeLDAPConnection{
+		ModifyRequestToExpect: &ldap.ModifyRequest{
+			DN: "CN=svc-example,OU=Users,DC=example,DC=net",
+		},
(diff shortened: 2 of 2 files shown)


cipherboy:
@waterWang Please review the contributing guide: https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md

PR #3869, opened 2026-08-21
Validate that + wildcards behave like *

cipherboy:
## Description

Note the rules of ACL policies:

**At the same rule precedence level**, deny takes precedence over any other operation. This does not mean that a deny in a lower-priority rule takes precedence over an approval; it makes that the deny is ignored.

And per https://openbao.org/docs/concepts/policies/#policy-syntax

1. If the first wildcard (`+`) or glob (`*`) occurs earlier in P1, P1 is lower priority
2. If P1 ends in * and P2 doesn't, P1 is lower priority
3. If P1 has more + (wildcard) segments, P1 is lower priority
4. If P1 is shorter, it is lower priority
5. If P1 is smaller lexicographically, it is lower priority

These tests attempt to demonstrate that w.r.t. LIST/SCAN operations. By the above, the following rules are sorted in priority order:

- `path "foo/metadata/*" { capabilities = ["list","scan"] }`
- `path "foo/metadata/+/private/*" { capabilities = ["deny"] }`

And thus a deny would never be processed because it is lower priority than the granting rule. This is hard to detect in general as a rule may exist across two different policies and it would be dependent on an auth method or identity linking the two policies together for this rule to be ignored.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/policy/acl_test.go
+++ b/internal/vault/policy/acl_test.go
@@ -996,81 +996,99 @@ func TestACLGrantingPolicies(t *testing.T) {
 	}
 }
 
-func TestACL_ListScanWildcardDeny(t *testing.T) {
+func TestACL_ListScanDeny(t *testing.T) {
 	ns := namespace.RootNamespace
 	ctx := namespace.ContextWithNamespace(t.Context(), ns)
 
-	tests := []struct {
-		path    string
-		allowed bool
-	}{
-		{
-			path:    "foo",
-			allowed: false,
-		},
-		{
-			path:    "foo/metadata",
-			allowed: true,
-		},
-		{
-			path:    "foo/metadata/test",
-			allowed: true,
-		},
-		{
-			path:    "foo/metadata/private",
-			allowed: false,
-		},
-		{
-			path:    "foo/metadata/private/internal",
-			allowed: false,
-		},
-		{
-			path:    "foo/metadata/private/granted",
-			allowed: true,
-		},
-		{
-			path:    "foo/metadata/private/granted/test",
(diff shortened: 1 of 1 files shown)


internal/vault/policy/acl_test.go
```diff
@@ -1481,8 +1499,37 @@ path "test/star" {
...
+// While tricky, note that + only matches a single segment. So +/ is
+// necessary to forbid a LIST, as the LIST operation is not on + but
+// on +/ and that's a different path from +.
+path "foo/metadata/private/internal/+/" { capabilities = ["deny"] }
+path "foo/metadata/private/granted/*" { capabilities = ["list","scan"] }
+`
+
+var listScanDenyInternalWildcardPolicy = `
+path "foo/metadata/*" { capabilities = ["list","scan"] }
+path "foo/metadata/private" { capabilities = ["deny"] }
```
cipherboy: This rule is superfluous and is never evaluated.

internal/vault/policy/acl_test.go
```diff
@@ -1481,8 +1499,37 @@ path "test/star" {
...
+path "foo/metadata/private/" { capabilities = ["deny"] }
+path "foo/metadata/+/internal" { capabilities = ["deny"] }
+path "foo/metadata/+/internal/" { capabilities = ["deny"] }
+path "foo/metadata/+/internal/+" { capabilities = ["deny"] }
```
cipherboy: As above. This rule is superfluous and is never evaluated.

wslabosz-reply (approved):

PR #3870, opened 2026-08-21
fix cas for workflow update

Luiggi33:
## Description

While working on the workflow command family, I noticed that the cas flag did not work. This seems to be caused by the reassignment of the variable inside the if block, so it is discarded and never actually set for the workflowStore.Set call.

Without the fix:
```
$ bao write sys/workflows/manage/test cas=-1 cas_required=true workflow='flow "check" {
  request "status" {
    operation = "read"
    path = "sys/seal-status"
  }
}'
Error writing data to sys/workflows/manage/test: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/sys/workflows/manage/test
Code: 400. Errors:

* check-and-set parameter required for this call
```

With the fix applied:
```
$ bao write sys/workflows/manage/test cas=-1 cas_required=true workflow='flow "check" {
  request "status" {
    operation = "read"
    path = "sys/seal-status"
  }
}'
allow_unauthenticated    false
cas_required             true
description              n/a
path                     test
version                  1
workflow                 flow "check" {
  request "status" {
    operation = "read"
    path = "sys/seal-status"
  }
}
```

## Rationale

Resolves: N/A

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3870.txt
+++ b/changelog/3870.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+sys/workflows: Fix `cas` variable getting discarded.
+```
--- a/changelog/3870.txt
+++ b/changelog/3870.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-sys/workflows: Fix `cas` variable getting discarded.
+sys/workflows: Fix Check-And-Set in workflow API.  
 ```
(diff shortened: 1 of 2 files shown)


changelog/3870.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+sys/workflows: Fix `cas` variable getting discarded.
```
JanMa:
```suggestion
sys/workflows: Fix Check-And-Set in workflow API.  
```

JanMa (changes requested):
Small remake regarding the changelog

JanMa (approved):

Issue #3665, opened 2026-08-03
RFE: less-privileged workflow list under sys/workflows/execute


AkshayJainG:
**Is your feature request related to a problem? Please describe.**

`LIST sys/workflows/manage` returns full workflow definitions (HCL/JSON) in `key_info`, including any embedded tokens or other secrets that operators placed in the workflow body. That is appropriate for callers with `sys/workflows/manage` privileges, but there is no thinner list for callers who only need to discover executable workflow names.

After discussion with maintainers on the security mailing list, this is not treated as a vulnerability: privileged manage-path LIST exposing construction is expected. Filing here as an RFE for a less-privileged discovery path.

**Describe the solution you'd like**

Add a secondary `LIST` under `sys/workflows/execute` (and optionally a matching prefix/`scan` form) that returns **entry names only**.

Optionally, couple that with a read on the execute path that returns I/O metadata (declared inputs/outputs/vars) **without** the full construction/HCL.

Today `createWorkflowListResponse` already builds a metadata-only map (path/version/description/etc.), while `handleWorkflowsList` uses `createWorkflowDataResponse`, which adds the full `workflow` body. A natural shape is to reuse the thinner response for execute-side listing.

**Describe alternatives you've considered**

- Keep listing exclusively on `sys/workflows/manage` and require manage ACL for discovery (status quo; works, but forces broader privilege than execute-only operators need).
- Filter/redact secrets from manage LIST (brittle; secrets can appear anywhere in HCL).

**Explain any additional use-cases**

Operators or automation granted only execute rights often need to enumerate available workflow names (and maybe required inputs) without receiving the workflow construction or embedded credentials.

**Additional context**

Current behavior sketch (manage LIST):

1. Write a workflow under `sys/workflows/manage/<name>` whose HCL embeds a token/secret.
2. `bao list -detailed sys/workflows/manage` (or API `LIST sys/workflows/manage`).
3. `key_info[<name>].workflow` contains the full definition, including the embedded secret.

Observed on current `main` (workflows feature). Happy to adjust the proposal to whatever ACL/`sys` layout maintainers prefer.

PR #3871, opened 2026-08-22
add workflow command family

### LINKED ISSUE #3331: CLI support for workflows
Would be nice to have `bao workflow write` / `bao workflow read` / `bao workflow call` commands for convenience.

Luiggi33:
## Description

Add the following commands:
- bao workflow call -> call a workflow with given inputs
- bao workflow delete -> delete a workflow
- bao workflow edit -> edit a existing workflow in a editor of his choosing
- bao workflow list -> lists all workflows
- bao workflow read -> read all infos about a workflow
- bao workflow write -> create/update a workflow from a file or stdin

Also adds simple test for each command, ~~these will partially fail when #3870 isn't applied~~.

## Rationale

Resolves: #3331 

## Acknowledgements

       also allowed.

### DIFF
--- a/api/sys_workflow.go
+++ b/api/sys_workflow.go
@@ -83,3 +83,20 @@ func (c *Sys) GetWorkflowWithContext(ctx context.Context, path string) (*GetWork
 	}
 	return result.Data, err
 }
+
+func (c *Sys) DeleteWorkflow(path string) error {
+	return c.DeleteWorkflowWithContext(context.Background(), path)
+}
+
+func (c *Sys) DeleteWorkflowWithContext(ctx context.Context, path string) error {
+	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
+	defer cancelFunc()
+
+	r := c.c.NewRequest(http.MethodDelete, fmt.Sprintf("/v1/sys/workflows/manage/%s", path))
+
+	resp, err := c.c.rawRequestWithContext(ctx, r)
+	if err == nil {
+		defer resp.Body.Close() //nolint:errcheck
+	}
+	return err
+}
--- a/api/sys_workflow.go
+++ b/api/sys_workflow.go
@@ -11,11 +11,7 @@ import (
 	"github.com/go-viper/mapstructure/v2"
 )
 
-func (c *Sys) ListWorkflows() ([]string, error) {
-	return c.ListWorkflowsWithContext(context.Background())
-}
-
-func (c *Sys) ListWorkflowsWithContext(ctx context.Context) ([]string, error) {
+func (c *Sys) ListWorkflows(ctx context.Context) ([]string, error) {
 	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
 	defer cancelFunc()
 
@@ -58,11 +54,7 @@ type GetWorkflowResponse struct {
 	Version              int    `json:"version"`
 }
 
-func (c *Sys) GetWorkflow(path string) (*GetWorkflowResponse, error) {
-	return c.GetWorkflowWithContext(context.Background(), path)
-}
-
-func (c *Sys) GetWorkflowWithContext(ctx context.Context, path string) (*GetWorkflowResponse, error) {
+func (c *Sys) GetWorkflow(ctx context.Context, path string) (*GetWorkflowResponse, error) {
 	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
 	defer cancelFunc()
 
@@ -92,11 +84,7 @@ type PutWorkflowInput struct {
 	CAS                  *int   `json:"cas,omitempty"`
 }
 
-func (c *Sys) PutWorkflow(path string, input PutWorkflowInput) (*GetWorkflowResponse, error) {
-	return c.PutWorkflowWithContext(context.Background(), path, input)
-}
-
-func (c *Sys) PutWorkflowWithContext(ctx context.Context, path string, input PutWorkflowInput) (*GetWorkflowResponse, error) {
+func (c *Sys) PutWorkflow(ctx context.Context, path string, input PutWorkflowInput) (*GetWorkflowResponse, error) {
 	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
 	defer cancelFunc()
 
@@ -121,11 +109,7 @@ func (c *Sys) PutWorkflowWithContext(ctx context.Context, path string, input Put
 	return result.Data, err
 }
 
-func (c *Sys) DeleteWorkflow(path string) error {
-	return c.DeleteWorkflowWithContext(context.Background(), path)
-}
-
-func (c *Sys) DeleteWorkflowWithContext(ctx context.Context, path string) error {
+func (c *Sys) DeleteWorkflow(ctx context.Context, path string) error {
 	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
 	defer cancelFunc()
 
@@ -138,11 +122,7 @@ func (c *Sys) DeleteWorkflowWithContext(ctx context.Context, path string) error 
 	return err
 }
 
-func (c *Sys) CallWorkflow(path string, data map[string]any) (*Secret, error) {
-	return c.CallWorkflowWithContext(context.Background(), path, data)
-}
-
-func (c *Sys) CallWorkflowWithContext(ctx context.Context, path string, data map[string]any) (*Secret, error) {
+func (c *Sys) CallWorkflow(ctx context.Context, path string, data map[string]any) (*Secret, error) {
 	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
 	defer cancelFunc()
 
--- a/api/sys_workflow.go
+++ b/api/sys_workflow.go
@@ -66,7 +66,7 @@ func (c *Sys) GetWorkflowWithContext(ctx context.Context, path string) (*GetWork
 	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
 	defer cancelFunc()
 
-	r := c.c.NewRequest(http.MethodGet, fmt.Sprintf("/v1/sys/policies/acl/%s", path))
+	r := c.c.NewRequest(http.MethodGet, fmt.Sprintf("/v1/sys/workflows/manage/%s", path))
 
 	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
--- a/api/sys_workflow.go
+++ b/api/sys_workflow.go
@@ -100,3 +100,30 @@ func (c *Sys) DeleteWorkflowWithContext(ctx context.Context, path string) error 
 	}
 	return err
 }
+
+func (c *Sys) CallWorkflow(path string, data map[string]any) (*Secret, error) {
+	return c.CallWorkflowWithContext(context.Background(), path, data)
+}
+
+func (c *Sys) CallWorkflowWithContext(ctx context.Context, path string, data map[string]any) (*Secret, error) {
+	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
+	defer cancelFunc()
+
+	r := c.c.NewRequest(http.MethodPost, fmt.Sprintf("/v1/sys/workflows/execute/%s", path))
+	if err := r.SetJSONBody(data); err != nil {
+		return nil, err
+	}
+
+	resp, err := c.c.rawRequestWithContext(ctx, r)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	secret, err := ParseSecret(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	return secret, nil
+}
--- a/api/sys_workflow.go
+++ b/api/sys_workflow.go
@@ -84,6 +84,43 @@ func (c *Sys) GetWorkflowWithContext(ctx context.Context, path string) (*GetWork
 	return result.Data, err
 }
 
+type PutWorkflowInput struct {
+	Workflow             string `json:"workflow"`
+	Description          string `json:"description,omitempty"`
+	AllowUnauthenticated bool   `json:"allow_unauthenticated,omitempty"`
+	CASRequired          bool   `json:"cas_required,omitempty"`
+	CAS                  *int   `json:"cas,omitempty"`
+}
+
+func (c *Sys) PutWorkflow(path string, input PutWorkflowInput) (*GetWorkflowResponse, error) {
+	return c.PutWorkflowWithContext(context.Background(), path, input)
+}
+
+func (c *Sys) PutWorkflowWithContext(ctx context.Context, path string, input PutWorkflowInput) (*GetWorkflowResponse, error) {
+	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
+	defer cancelFunc()
+
+	r := c.c.NewRequest(http.MethodPost, fmt.Sprintf("/v1/sys/workflows/manage/%s", path))
+	if err := r.SetJSONBody(input); err != nil {
+		return nil, err
+	}
+
+	resp, err := c.c.rawRequestWithContext(ctx, r)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	var result struct {
+		Data *GetWorkflowResponse
+	}
+	err = resp.DecodeJSON(&result)
+	if err != nil {
+		return nil, err
+	}
+	return result.Data, err
+}
+
 func (c *Sys) DeleteWorkflow(path string) error {
 	return c.DeleteWorkflowWithContext(context.Background(), path)
 }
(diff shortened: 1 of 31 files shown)


Luiggi33:
One thing I would be interested in, should the license header stay in these files or is that just a relict from Hashicorp Vault? I copied them from some other command for good measure :P

JanMa:
> One thing I would be interested in, should the license header stay in these files or is that just a relict from Hashicorp Vault? I copied them from some other command for good measure :P

Please use the following in all new files 

```
```

You can leave the header as is in existing ones. Thanks

cipherboy:
@JanMa Yeah, I really should fork https://github.com/hashicorp/copywrite/ and teach it about forks :laughing:

api/sys_workflow.go
```diff
@@ -0,0 +1,166 @@
...
+	"net/http"
+
+	"github.com/go-viper/mapstructure/v2"
+)
+
+func (c *Sys) ListWorkflows() ([]string, error) {
+	return c.ListWorkflowsWithContext(context.Background())
+}
+
+func (c *Sys) ListWorkflowsWithContext(ctx context.Context) ([]string, error) {
```
cipherboy: I think this was mentioned somewhere else, but we can drop `WithContext` on new things and just have context be part of the base name.

api/sys_workflow.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if secret == nil || secret.Data == nil {
+		return []string{}, nil
+	}
+
+	var results []string
+	err = mapstructure.Decode(secret.Data["keys"], &results)
+	if err != nil {
+		return nil, err
+	}
```
cipherboy, line 43: Wonder if we just use this pattern vs what's done in `GetWorkflow(...)`? No real preference.

internal/command/workflow_edit.go
```diff
@@ -0,0 +1,182 @@
...
+func (c *WorkflowEditCommand) Flags() *FlagSets {
+	set := c.flagSet(FlagSetHTTP)
+	f := set.NewFlagSet("Command Options")
+
+	f.StringVar(&StringVar{
+		Name:    "editor",
+		Target:  &c.flagEditor,
+		Default: "editor",
+		EnvVar:  "EDITOR",
+		Usage:   "The editor you want to use. Defaults to 'editor'",
```
cipherboy:
```suggestion
		Usage:   "The editor you want to use. Defaults to '$EDITOR' and then falls back to 'editor'",
```

internal/command/workflow_edit.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	tmpFile, err := os.CreateTemp("", fmt.Sprintf("workflow-%s-*.hcl", strings.ReplaceAll(path, "/", "_")))
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error creating temporary workflow file: %s", err))
+		return 2
+	}
+	defer os.Remove(tmpFile.Name()) //nolint:errcheck
+
+	if err := os.Chmod(tmpFile.Name(), 0o640); err != nil {
```
cipherboy:
```suggestion
	if err := os.Chmod(tmpFile.Name(), 0o600); err != nil {
```

Not sure why we'd give groups access, given that we're the one invoking editor?

internal/command/workflow_edit.go
cipherboy: I do wonder if `nano` is a better fallback, tbh.

internal/command/workflow_list.go
```diff
@@ -0,0 +1,94 @@
...
+		return 1
+	}
+
+	client, err := c.Client()
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 2
+	}
+
+	workflows, err := client.Sys().ListWorkflows()
```
cipherboy: Do you want to tackle #3665 afterwards? I think `ListWorkflows()` could then be updated to use that endpoint and a `ListDetailedWorkflows()` or similar could call the existing endpoint (and the `-detailed` flag be added to this command to switch between?).

internal/command/workflow_edit.go
```diff
@@ -0,0 +1,182 @@
...
+		return 0
+	}
+
+	workflowInput := api.PutWorkflowInput{
+		Workflow:             string(tmpFileContent),
+		Description:          workflowResp.Description,
+		AllowUnauthenticated: workflowResp.AllowUnauthenticated,
+		CASRequired:          workflowResp.CasRequired,
+	}
+	if workflowResp.CasRequired {
```
cipherboy: I think its worth setting this unconditionally fwiw, so that editing errs if someone else edits the file in between.

internal/command/workflow_edit.go
```diff
@@ -0,0 +1,182 @@
...
+		c.UI.Error(fmt.Sprintf("No workflow found in path %s", path))
+		return 2
+	}
+
+	tmpFile, err := os.CreateTemp("", fmt.Sprintf("workflow-%s-*.hcl", strings.ReplaceAll(path, "/", "_")))
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error creating temporary workflow file: %s", err))
+		return 2
+	}
+	defer os.Remove(tmpFile.Name()) //nolint:errcheck
```
cipherboy: I do wonder if its worth setting removal to be conditional on write succeeding (though, maybe remove if editor fails or the workflow is blank...)... that would give an easy recovery semantic rather than recreating all changes. Thoughts?

internal/command/workflow_edit.go
```diff
@@ -0,0 +1,182 @@
...
+		return exitCode
+	}
+
+	tmpFileContent, err := os.ReadFile(tmpFile.Name())
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error read temporary workflow file after changes: %s", err))
+		return 2
+	}
+
+	if bytes.Equal(currentWorkflowBytes, tmpFileContent) {
```
cipherboy:
```suggestion
	if bytes.Equal(currentWorkflowBytes, tmpFileContent) || len(tmpFileContent) == 0 {
```

cipherboy: To give an easy way of cancelling the edit.

cipherboy (commented):
Some initial feedback @Luiggi33!

internal/command/workflow_edit.go
Luiggi33: If we wanted to be really sure, should we use `vi`? Im pretty sure that's installed almost everywhere

internal/command/workflow_edit.go
Luiggi33: We would need to make sure that users understand when they fall back and when not. There might be a need to add -recover or -discard flags or something so that the user can decide if he wants recovery or not

internal/command/workflow_edit.go
cipherboy: Don't need to tell me twice! (I'm a vim user)

internal/command/workflow_edit.go
cipherboy: @Luiggi33 What about mirroring visudo's behavior? https://man7.org/linux/man-pages/man8/visudo.8.html

> visudo parses the sudoers file after editing and will not save the
> changes if there is a syntax error.  Upon finding an error, visudo
> will print a message stating the line number(s) where the error
> occurred and the user will receive the “What now?” prompt.  At
> this point the user may enter ‘e’ to re-edit the sudoers file, ‘x’
> to exit without saving the changes, or ‘Q’ to quit and save
> changes.  The ‘Q’ option should be used with extreme caution
> because if visudo believes there to be a syntax error, so will
> sudo.  If ‘e’ is typed to edit the sudoers file after a syntax
> error has been detected, the cursor will be placed on the line
> where the error occurred (if the editor supports this feature).

api/sys_workflow.go
Luiggi33: I assume you meant like this https://github.com/openbao/openbao/pull/3871/commits/852200e2673fa753dea039df9827d2a709c2fb8e?

internal/command/workflow_edit.go
Luiggi33: Makes sense, especially with the planned changes to the recover

internal/command/workflow_list.go
Luiggi33: Sounds like a idea, i had that issue on my radar already but it makes sense to chain this together :)

api/sys_workflow.go
Luiggi33: Switched to this one, this "feels" better to use :'D

internal/command/workflow_edit.go
Luiggi33: I tried implementing something similar to that. I did have to promote a dependency to have a reliable diff, hope that is okay

api/sys_workflow.go
cipherboy: @Luiggi33 Yeah that looks good, thanks!

api/sys_workflow.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return results, err
+}
+
+type GetWorkflowResponse struct {
```
satoqz: Should probably just be

```suggestion
type WorkflowResponse struct {
```

since it's also returned by `PutWorkflow`.

Perhaps it should also be `WorkflowOutput` since we have `PutWorkflowInput` (which could just be `WorkflowInput`).

`api` is already a bit inconsistent, but picking one of request/response and input/output would be nice.

Luiggi33: Finally got around to implementing this, it makes more sense having something "generic"

internal/command/workflow_list.go
Luiggi33: @cipherboy 
I have made some changes that warrant another review. My basic idea is to use the separation of the manage and execute endpoint to allow basic information through the execution read capability, with more sensitive information behind the manage read capability. This is supported by the current docs with the statement that the manage endpoint is privileged.

Is that something that would work for us or should I structure it differently? I wanted to avoid adding more API paths when the current ones would suffice

cipherboy: @Luiggi33 I like the idea of read and I'd keep it, but I think #3665 was about adding a LIST/SCAN version to `sys/workflows/execute/...`. It'd still be authenticated, just not privileged and return all the detailed information that `sys/workflows/manage` did (only a list of keys, not detailed info). 

Otherwise, there's no discovery for non-privileged users. Partly that was by design, but with `READ sys/workflows/execute/...` now being supported, that'd make it easy for list filtering as well, to see which workflows both exist and are visible.

At this point, you could probably split the server-side workflow improvements+tests+docs into a new PR and then keep this PR as a follow-up? Thoughts?

internal/vault/logical_system_workflows.go
```diff
@@ -274,6 +314,45 @@ func (b *SystemBackend) workflowPaths() []*framework.Path {
...
+		}
+		workflowData["output_headers"] = headerNames
+
+		if m, ok := outputs.Data.(map[string]any); ok {
+			dataKeys := make([]string, 0, len(m))
+			for k := range m {
+				dataKeys = append(dataKeys, k)
+			}
+			workflowData["output_data_keys"] = dataKeys
+		}
```
cipherboy, line 350: I don't think this is quite right and it'd be a pain to make it work correctly IMO. Since `data` can be fully dynamic and only needs to _decode_ after evaluation to `map[string]any`, you can define:

```hcl
data = {
  eval_source = "cel"
  eval_type = "map"
  expression = "..."
}
``` 

(pseudo-code)

and get something that is both a map but whose keys are irrelevant to the caller (`eval_source`, `eval_type`, and `expression` _presumably_ won't appear anywhere in the output). You wouldn't know what the actual output keys are until evaluation of the workflow, since the CEL expression won't compile. 🤔 

I think the only way around this is if we expanded `profiles.OutputConfig` to optionally support response fields like `framework.PathOperation` does, accepting a `map[int][]framework.Response` structure for OpenAPI purpose.

I think that's well into the realm of future RFE, so I'm happy to table it for the time being and drop this. 

Headers on the other hand, we force into the supposed map structure (to make it easier to reason about), at the cost of less dynamicity.

internal/command/base_predict.go
```diff
@@ -235,6 +240,13 @@ func (p *Predict) VaultPolicies() complete.Predictor {
 	return p.filterFunc(p.policies)
 }
 
+// VaultWorkflows returns a predictor for workflows. This is a public
+// API for consumers, but you probably want BaseCommand.PredictVaultPolicies
```
cipherboy:
```suggestion
// API for consumers, but you probably want BaseCommand.PredictVaultWorkflows
```

internal/command/workflow_edit.go
```diff
@@ -0,0 +1,259 @@
...
+				Description:          workflowResp.Description,
+				AllowUnauthenticated: workflowResp.AllowUnauthenticated,
+				CASRequired:          workflowResp.CasRequired,
+				CAS:                  &workflowResp.Version,
+			}
+			if _, putErr := client.Sys().PutWorkflow(context.Background(), path, workflowInput); putErr != nil {
+				latest, err := client.Sys().GetWorkflow(context.Background(), path)
+				if err != nil {
+					c.UI.Error(fmt.Sprintf("Error fetching latest workflow under path %s: %s", path, err))
+					c.UI.Error(fmt.Sprintf("Your edits are still available in %s", tmpFile.Name()))
```
cipherboy, line 174: Maybe also add another line like:

```go
c.UI.Error(fmt.Sprintf("After fixing the workflow, you can write the file via:\n'\tbao workflow write ..."))
```

So they know there's an easy way to do this?

api/sys_workflow.go
```diff
@@ -0,0 +1,201 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/go-viper/mapstructure/v2"
+)
+
+func (c *Sys) ListWorkflows(ctx context.Context) ([]string, error) {
```
cipherboy, line 14: In particular, there'd then be a `ListDetailedWorkflows(...) (map[string]*WorkflowResponse, error) { ... }` implementation that'd use this API endpoint and this list endpoint would use `sys/workflows/execute/...` endpoint.

Both take prefixes, I wonder if `prefix` should be an arg here?

go.mod
```diff
@@ -111,6 +111,7 @@ require (
 	github.com/openbao/openbao/sdk/v2 v2.6.2
 	github.com/ory/dockertest/v4 v4.0.0
 	github.com/pires/go-proxyproto v0.14.0
+	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2
```
cipherboy: Do we not have another diff library we could use instead I wonder? 🤔  I think we use a few in testing.

cipherboy: >  THIS PACKAGE IS NO LONGER MAINTAINED.

cipherboy:
@Luiggi33 Mostly small polishing but otherwise looking fairly good!

internal/command/workflow_list.go
Luiggi33: @cipherboy I am a bit unsure about the structure, as there are multiple options in what informations should be discoverable through which endpoint. My specific questions resolve around what information should be exposed in the none privileged endpoints, i.E.:

<table>
<tr>
 <td>Path
 <td>Basic Information (path, description, etc)
 <td>Privileged Information (workflow content)
<tr>
 <td>LIST sys/workflows/execute
 <td>✅ (Just the list of the keys or add basic relevant info?)
 <td>❌
<tr>
 <td>GET sys/workflows/execute/<path>
 <td>✅ (Only the execution relevant infos or all infos + execution relevant infos?)
 <td>❌
<tr>
 <td>LIST sys/workflows/manage
 <td>✅
 <td>✅
<tr>
 <td>GET sys/workflows/manage/<path>
 <td>✅
 <td>✅
</table>

Luiggi33: As soon as that is clarified, I would split this out of this merge request and create another request for these changes

internal/vault/logical_system_workflows.go
Luiggi33: I was aiming for providing the returned keys from the output, as in
```
output {
  data = {
    password = { <--- This key
      eval_source    = "response"
      eval_type      = "string"
      flow_name      = "generate-password"
      response_name  = "generate"
      field_selector = ["data", "password"]
    }
  }
}
```

or is that key not required when outputting information back to the client?

internal/command/workflow_edit.go
Luiggi33: I like the idea, but only added one if there is no workflow found under that path.

I decided against the other ones:
- when an error occurs in the put request (line 170), we cant be sure if the write would go through (from a application, network and CAS perspective)
- when there is a expected error (line 181) the user can already choose to overwrite
- when there is a error reading the users response (line 187), I am too unsure why that could even happen

go.mod
Luiggi33: No shame to anyone but all three diff libs that I found in the repo (pmezard/go-difflib, hexops/gotextdiff and sergi/go-diff) have not been updated since a year or more. The most recent updated "go-diff" does not support unified diffs which should be the simplest option to read, "gotextdiff" extends that package and adds unified diffs, but hasnt been updated in even more years...

I switched it to said "gotextdiff", but if there is any better package im open to changing it once again

Issue #2846, opened 2026-04-09
Deprecation and removal of File storage backend


cipherboy:
Like InMemHA, the file backend is not transactional as filesystems do not support interactive transactions along the semantics we require.

In the future, a persistent dev-like (non-HA) backend could be created through direct use of bbolt or pebbledb, without Raft. However, with PostgreSQL having easy container images available, and boots much more quickly than Raft, this likely satisfies much of local persistence needs.

The file backend is already not recommend for production use; this finishes its removal. 

---

Aligning with our other deprecations, I'd like to remove this in v2.7.0, but warn about its use in v2.6.0.

tsaarni:
I've used file backend since it has provided quick test and dev experience when requiring persistence. Any other lightweight built-in solution would of course do the same. "Multi-service" dev environments, like configuring PostgreSQL in a container, are certainly not impossibly complicated, but they come with some more configuration burden. It requires more than a simple copy-and-paste command in one terminal or a vscode debugger launch. For that reason, I would have liked to have some recommended built-in "development backend".

cipherboy:
Thanks @tsaarni! I think Raft would work in that scenario, though it does still take a little longer to start up than `file` does, sadly. 

I'll see about adding a pure bbolt or pebbledb backend as a replacement.

satoqz:
Still need to actually remove it.

PR #3876, opened 2026-08-24
Refactor paginated listing into common module

cipherboy:
## Description

Raft's implementation of paginated lists should be the reference for all backends. Rather than copying the logic, this is now refactored into a common, iteration-based approach for easier reuse in other storage backends. This also modularize's Raft's approach to transactional listing.

## Rationale

This significantly helps with PebbleDB implementation, in support of #2846. The actual new LoC is probably small, but the commentary is useful. 

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/physical/raft/transaction.go
+++ b/internal/physical/raft/transaction.go
@@ -4,19 +4,17 @@
 package raft
 
 import (
-	"bytes"
 	"context"
 	"crypto/sha512"
 	"crypto/subtle"
 	"encoding/json"
 	"fmt"
 	"hash"
+	"iter"
 	"maps"
 	"math"
-	"path/filepath"
 	"runtime"
 	"slices"
-	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


sdk/physical/listing.go
```diff
@@ -0,0 +1,339 @@
...
+	// returned by Key(...).
+	Start func()
+
+	// Next is a function which continues iteration, updating the next value
+	// returned by Key(...).
+	Next func()
+
+	// Key returns the key at the current iteration position, in both string
+	// and byte form.
+	Key func() (string, []byte)
```
satoqz: TBH I'd find the API less weird if we passed `ctx` and received `(key, err)` instead of awkwardly checking for context errors separately and forcing the backend to keep track of the current key?

cipherboy: @satoqz Not sure I follow? Maybe I'm confused but where would `err` come from?

I think you'd end up with an implementation like:

```
lister.Key = func() (string, error) {
     return key, ctx.Err()
}
```

But I'm not sure this is reasonable, given `.Key()` is called twice per iteration (once in `while`, once in the body). So `lister` would at minimum need to persist `err` (or re-invoke `.Key(...)` and hope for a consistent error?)... If not also `key`. 

Let me push the initial pebbledb code.

cipherboy: @satoqz If you look at `internal/physical/pebbledb/pebbledb.go` in https://github.com/openbao/openbao/pull/3879, there's still some cleanup to do, but you should see that I think the function definitions roughly make sense across implementations? I guess in Pebble we could potentially return `iter.Error()`, though strictly I think this would be redundant if we properly handled `iter.Close()`'s error based on docs. 

In Pebble we could assign `NewIterWithContext`, but there's no counterpart of that with BBolt. 

Also in BBolt, the second ignored value from `Next`/`First` are not errors, but values.

satoqz: @cipherboy I suppose my criticism is that this abstraction fits bbolt's and pebble's APIs but isn't otherwise very generic. Suppose the iterator/cursor of $MY_DB returns an error, I don't really have a place to propagate that. Similarly, we'd be polling `ctx.Err()` unnecessarily if my iterator propagated cancellation natively.

So I suppose this is fine for what it achieves across these backends but I'd hesitate to place it in SDK? My 2c.

(Not that any part of `sdk/physical` is particularly useful outside of this repo as of yet. But if we do ever pluginize storage or make the package otherwise useful, I'd want to avoid it already having become something that should've largely been placed under `internal/helper`.)

satoqz: AFAICT both pebble and bbolt implement this as `return string(key), key`, so not sure what the advantage is to let implementers return both values.

cipherboy: @satoqz It's an interesting question, I suspect the original list implementation was just too prematurely optimized and we should just work with strings. Both of them only use `[]byte` of course for keys, but we ultimately need to return strings, so our `while()` implementation should probably just work in strings rather than dealing in `[]byte`s.

cipherboy: @satoqz The reason this is ending up in SDK is so that pebble can ultimately go in SDK so we can transparently remove Inmem and back it with pebble... We can't create `sdk/internal` as that would break importing it from raft I believe since SDK is a separate module...

Point taken though, will see if I can refactor this to also support PostgreSQL and it's error types.

cipherboy: @satoqz I don't think it makes to update PostgreSQL given how simple it is; we delegate everything into PostgreSQL which is the way to go anyways.

cipherboy:
@satoqz Ok, this should be revised!

sdk/physical/listing.go
satoqz: Oh, yeah, I was thinking about backends that expose capabilities like bbolt and pebble do, not SQL ones.

internal/physical/raft/transaction.go
```diff
@@ -468,158 +466,68 @@ func listShouldIncludeEntry(prefix string, after string, key string) (string, bo
...
 	if t.haveFinishedTx {
 		return nil, physical.ErrTransactionAlreadyCommitted
 	}
 
-	prefixBytes := []byte(prefix)
-	fullAfter := filepath.Join(prefix, after)
-	seekPrefix := []byte(fullAfter)
-	if after == "" {
-		seekPrefix = prefixBytes
+	lister := &physical.Lister{
```
wslabosz-reply, line 476: purely a cosmetic change but if perhaps we made `SeekPrefix` a function instead of `Lister` method, then it would be possible to inline declare the `Lister` struct as:

```
lister := &physical.Lister{
		Prefix: prefix,
		After:  after,
		Limit:  limit,
		Start: func() error {
			_, seekPrefix := physical.SeekPrefix(prefix, after)
			key, _ = cursor.Seek(seekPrefix)
			return nil
		},
		Next: func() error {
			key, _ = cursor.Next()
			return nil
		},
		Key: func() (string, bool, error) {
			return string(key), key != nil, nil
		},
		Deleted: func(path string) bool {
			entry, present := t.updates[path]
			if !present {
				return false
			}

			return entry.Contents == nil
		},
		Inserted: func() iter.Seq[string] {
			return func(yield func(K string) bool) {
				for key, entry := range t.updates {
					if entry.Contents == nil {
						// Skip deletes.
						continue
					}

					if !yield(key) {
						return
					}
				}
			}
		},
	}
```

satoqz: I wonder if we can't make `SeekPrefix` a private function and pass the result to `Start()` instead? It's such a bespoke detail I'd rather hide it.

sdk/physical/listing.go
```diff
@@ -0,0 +1,308 @@
...
+
+	return nil
+}
+
+func (l *Lister) while(ctx context.Context, listerErr *error, key *string, start *bool, keys []string) bool {
+	if ctx.Err() != nil {
+		// Context cancelled.
+		*listerErr = ctx.Err()
+		return false
+	}
```
satoqz, line 94: nit

```suggestion
	if err := ctx.Err(); err != nil {
		// Context cancelled.
		*listerErr = err
		return false
	}
```

satoqz (approved):

satoqz:
(per discussion with @cipherboy -- let's merge this and include further refactors with the pebble PR)

internal/physical/raft/transaction.go
cipherboy: Yeah I think that's a nicer pattern, thanks all!

I pick up the updates in the pebbledb PR.

PR #2849, opened 2026-04-09
Add deprecation notice about file backend

cipherboy:
## Rationale

Resolves: #2846 

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3879, opened 2026-08-25
Add PebbleDB storage backend to replace file

### LINKED ISSUE #2846: Deprecation and removal of File storage backend

cipherboy:
## Description

Using:

```
$ benchmark-openbao run -config /home/cipherboy/GitHub/cipherboy/benchmark-openbao/docs/examples/kv/kvv2-read.hcl -duration 10s -workers 10
$ benchmark-openbao run -config /home/cipherboy/GitHub/cipherboy/benchmark-openbao/docs/examples/kv/kvv2-write.hcl -duration 10s -workers 10
```

Before (inmem):

```
kvv2_read_test    84122  8412.086062   8410.630328   1.183072ms  2.553682ms  3.395259ms   100.00%
kvv2_write_test   17064  1706.318316   1705.478354   5.856625ms  10.12172ms  13.238112ms  100.00%
```

After (pebble, ephemeral):

```
kvv2_read_test   216589  21658.888153  21657.982483  445.76µs    1.058433ms  2.043133ms   100.00%
kvv2_write_test   96113  9610.861103   9609.835982   1.033487ms  2.129151ms  3.141327ms   100.00%
```

Stats:

> **+257% read, +563% write**

Before (file):

```
kvv2_read_test  275516  27551.647053  27551.081354  347.734µs  812.073µs  1.352059ms  100.00%
kvv2_write_test  25841  2584.111020  2583.363505  3.86724ms  7.391953ms  9.694184ms  100.00%
```

After (pebble, persistent):

```
kvv2_read_test  184876  18487.547272  18486.833769  523.078µs  1.317505ms  2.216445ms  100.00%
kvv2_write_test  75729  7572.407254  7571.781466  1.311692ms  2.869388ms  4.743008ms  100.00%
```

Stats: 

> _-67% read_, **+293% write**

So tests should see a bit of an increase in performance here.

## Rationale

Related: #2849
Resolves: #2846 

## Acknowledgements

       also allowed.


changelog/3879.txt
```diff
@@ -0,0 +1,12 @@
...
+**Add PebbleDB Storage Backend**: This storage backend is a non-HA, durable backend uses PebbleDB for transactions and improved performance.
+```
+```release-note:change
+command/server: Remove `file` storage backend support.
+```
+```release-note:deprecation
+command: The `file` storage backend is deprecated for `bao operator migrate` and will be removed in v2.8.0.
+```
+```release-note:improvement
+sdk/physical: Replace `inmem` backend with ephemeral PebbleDB instance.
```
wslabosz-reply, line 11:
```suggestion
sdk/physical: Replaced `inmem` backend with ephemeral PebbleDB instance.
```

changelog/3879.txt
```diff
@@ -0,0 +1,12 @@
+```release-note:feature
+**Add PebbleDB Storage Backend**: This storage backend is a non-HA, durable backend uses PebbleDB for transactions and improved performance.
+```
+```release-note:change
+command/server: Remove `file` storage backend support.
```
wslabosz-reply, line 5:
```suggestion
command/server: Removed `file` storage backend support.
```

changelog/3879.txt
```diff
@@ -0,0 +1,12 @@
+```release-note:feature
+**Add PebbleDB Storage Backend**: This storage backend is a non-HA, durable backend uses PebbleDB for transactions and improved performance.
```
wslabosz-reply:
```suggestion
**Add PebbleDB Storage Backend**: This storage backend is a non-HA, durable backend using PebbleDB for transactions and improved performance.
```

sdk/physical/testing.go
```diff
@@ -582,6 +582,51 @@ func ExerciseTransactionalBackend(t testing.TB, b TransactionalBackend) {
...
+	txn, err = b.BeginTx(context.Background())
+	require.NoError(t, err, "failed to start new transaction")
+	entries, err = txn.List(context.Background(), "")
+	require.NoError(t, err, "failed to list entries")
+	require.Equal(t, entries, []string{"testing/"}, "expected one entry in root storage")
+	entries, err = txn.List(context.Background(), "testing/")
+	require.NoError(t, err, "failed to list entries")
+	require.Equal(t, entries, []string{"foo"}, "expected one entry in testing/ storage")
+	err = txn.Delete(context.Background(), "testing/foo")
+	require.NoError(t, err, "failed to write entry")
```
wslabosz-reply:
```suggestion
	require.NoError(t, err, "failed to delete entry")
```

sdk/physical/testing.go
```diff
@@ -582,6 +582,51 @@ func ExerciseTransactionalBackend(t testing.TB, b TransactionalBackend) {
...
+	entry, err = txn.Get(context.Background(), "foo")
+	require.NoError(t, err, "failed to read entry")
+	require.Nil(t, entry, "expected to get a non-empty entry")
```
wslabosz-reply:
```suggestion
	require.Nil(t, entry, "expected to get nil entry")
```

changelog/3879.txt
satoqz: The former is more consistent with existing changelogs.

sdk/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,753 @@
...
+
+// NewBackend constructs a PebbleDB backend using the given
+// API client, server address, credentials, and database.
+func NewBackend(conf map[string]string, logger log.Logger) (physical.Backend, error) {
+	var err error
+
+	dir, ok := conf["directory"]
+	if !ok || dir == "" {
+		return nil, errors.New("missing or empty parameter 'directory'")
+	}
```
satoqz: IMO it should be `path` since that is what raft and file use, and is shorter.

changelog/3879.txt
```diff
@@ -0,0 +1,12 @@
...
+```
+```release-note:change
+command/server: Remove `file` storage backend support.
+```
+```release-note:deprecation
+command: The `file` storage backend is deprecated for `bao operator migrate` and will be removed in v2.8.0.
+```
```
satoqz, line 9: I wonder if we can make it available within `bao operator migrate` but not other commands? Also, have we tested the migration from file->pebble just to be sure?

website/content/docs/configuration/storage/pebbledb.mdx
```diff
@@ -0,0 +1,49 @@
...
+sidebar_label: PebbleDB
+description: |-
+  The PebbleDB storage backend stores OpenBao's data using PebbleDB, a
+  KV data store similar to BBolt. It can be used for durable single server
+  situations or to develop locally where durability is not critical.
+---
+
+# PebbleDB storage backend
+
+The pebbleDB storage backend stores OpenBao's data using
```
satoqz:
```suggestion
The PebbleDB storage backend stores OpenBao's data using
```

website/content/docs/configuration/storage/pebbledb.mdx
```diff
@@ -0,0 +1,49 @@
...
+[PebbleDB](https://github.com/cockroachdb/pebble), a KV data store similar
+to BBolt used by Integrated Storage. It can be used for durable single server
+situations, or to develop locally where durability is not critical.
```
satoqz:
```suggestion
[PebbleDB](https://github.com/cockroachdb/pebble), an embedded KV data store similar
to BBolt used by Integrated Storage (Raft). It can be used for durable single server
situations, or to develop locally where durability is not critical.
```

website/content/docs/configuration/storage/pebbledb.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+:::warning
+
+**No High Availability** – the PebbleDB backend does not support high
+availability. Only a single OpenBao node may use this data store.
```
satoqz:
```suggestion
**No High Availability** – the PebbleDB backend does not support high
availability. Do not share one PebbleDB data store between multiple OpenBao nodes.
```

Is perhaps closer to what we're actually trying to say?

sdk/physical/inmem/inmem.go
```diff
@@ -37,16 +33,14 @@ var (
...
-	permitPool   *physical.PermitPool
-	logger       log.Logger
-	failGet      atomic.Bool
-	failPut      atomic.Bool
-	failDelete   atomic.Bool
-	failList     atomic.Bool
-	logOps       bool
-	maxValueSize int
+	parent     physical.Backend
+	permitPool *physical.PermitPool
```
satoqz: Does it still make sense to add a PermitPool at this layer? If at all, shouldn't the underlying Pebble backend use one?

sdk/physical/inmem/inmem.go
```diff
@@ -246,27 +209,16 @@ func (i *InmemBackend) Delete(ctx context.Context, key string) error {
...
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	default:
-	}
 
-	i.root.Delete(key)
-	return nil
+	err := i.parent.Delete(ctx, key)
+	return err
```
satoqz:
```suggestion
	return i.parent.Delete(ctx, key)
```

and other ops below where we don't need to bind before returning.

internal/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,753 @@
...
+	var err error
+
+	lister := &physical.Lister{
+		Prefix: prefix,
+		After:  after,
+		Limit:  limit,
+		Start: func(_ []byte) error {
+			cursor.First()
+			return cursor.Error()
+		},
```
satoqz, line 172: How about:

```diff
diff --git a/sdk/physical/listing.go b/sdk/physical/listing.go
index f2154760ab..1e9e5eb397 100644
--- a/sdk/physical/listing.go
+++ b/sdk/physical/listing.go
@@ -47,10 +47,10 @@ type Lister struct {
 	Inserted func() iter.Seq[string]
 }
 
-// SeekPrefix returns the joined prefix of the first element, assuming a strict
+// seekPrefix returns the joined prefix of the first element, assuming a strict
 // greater-than seek logic. Lister handles both greater-than and
 // greater-than-equal seek logic.
-func (l *Lister) SeekPrefix() []byte {
+func (l *Lister) seekPrefix() []byte {
 	slash := "/"
 
 	// This is a quirk of how listing works not exercised by our test suite:
@@ -102,7 +102,7 @@ func (l *Lister) while(ctx context.Context, listerErr *error, key *string, start
 	if !*start {
 		*start = true
 
-		if err := l.Start(l.SeekPrefix()); err != nil {
+		if err := l.Start(l.seekPrefix()); err != nil {
 			*listerErr = err
 			return false
 		}
diff --git a/sdk/physical/pebbledb/pebbledb.go b/sdk/physical/pebbledb/pebbledb.go
index 6442e05905..92210fa69b 100644
--- a/sdk/physical/pebbledb/pebbledb.go
+++ b/sdk/physical/pebbledb/pebbledb.go
@@ -158,7 +158,7 @@ func (b *Backend) List(ctx context.Context, prefix string) ([]string, error) {
 	return b.ListPage(ctx, prefix, "", 0)
 }
 
-func newLister(view pebble.Reader, prefix string, after string, limit int) (*pebble.Iterator, *physical.Lister, error) {
+func newLister(view pebble.Reader, prefix string, after string, limit int) (*pebble.Iterator, *physical.Lister) {
 	var cursor *pebble.Iterator
 	var err error
 
@@ -166,7 +166,13 @@ func newLister(view pebble.Reader, prefix string, after string, limit int) (*peb
 		Prefix: prefix,
 		After:  after,
 		Limit:  limit,
-		Start: func(_ []byte) error {
+		Start: func(seekPrefix []byte) error {
+			cursor, err = view.NewIter(&pebble.IterOptions{
+				LowerBound: seekPrefix,
+			})
+			if err != nil {
+				return err
+			}
 			cursor.First()
 			return cursor.Error()
 		},
@@ -184,23 +190,11 @@ func newLister(view pebble.Reader, prefix string, after string, limit int) (*peb
 		},
 	}
 
-	opts := &pebble.IterOptions{
-		LowerBound: lister.SeekPrefix(),
-	}
-
-	cursor, err = view.NewIter(opts)
-	if err != nil {
-		return nil, nil, fmt.Errorf("error creating iterator: %w", err)
-	}
-
-	return cursor, lister, nil
+	return cursor, lister
 }
 
 func (b *Backend) ListPage(ctx context.Context, prefix string, after string, limit int) ([]string, error) {
-	cursor, lister, err := newLister(b.db, prefix, after, limit)
-	if err != nil {
-		return nil, err
-	}
+	cursor, lister := newLister(b.db, prefix, after, limit)
 	defer func() {
 		if cursor != nil {
 			cursor.Close() //nolint:errcheck // nothing we can do here.
@@ -358,10 +352,7 @@ func (t *Transaction) ListPage(ctx context.Context, prefix string, after string,
 	t.lock.Lock()
 	defer t.lock.Unlock()
 
-	cursor, lister, err := newLister(t.snapshot, prefix, after, limit)
-	if err != nil {
-		return nil, err
-	}
+	cursor, lister := newLister(t.snapshot, prefix, after, limit)
 	defer func() {
 		if cursor != nil {
 			cursor.Close() //nolint:errcheck // nothing we can do here.
```

satoqz (commented):
Just flushing out some comments before lunch.

satoqz:
It's unfortunate that `sdk/logical` pulls in `sdk/physical/pebble` via `InmemStorage` which results in ~5MB larger plugin binaries only because they pull in the `logical` package, without ever using `InmemStorage` outside of tests.

If we moved `InmemStorage` to a separate package (e.g., move to physical/inmem itself), we could prevent that binary size increase as only tests would pull that in, though it'll be breaking and pull stuff like `TestRequest` with it. So we probably can't do much 🤷 Classic case of not splitting the interface and its testing facilities into separate packages, tbh.

---

This comment:

```go
// InmemStorage implements Storage and stores all data in memory. It is
// basically a straight copy of physical.Inmem, but it prevents backends from
// having to load all of physical's dependencies (which are legion) just to
// have some testing storage.
type InmemStorage struct {
```

has been out of sync since 2017 but now holds some weight again 😆

sdk/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,753 @@
...
+	txns map[string]weak.Pointer[Transaction]
+}
+
+var (
+	_ physical.Backend              = &Backend{}
+	_ physical.TransactionalBackend = &Backend{}
+)
+
+// NewBackend constructs a PebbleDB backend using the given
+// API client, server address, credentials, and database.
```
satoqz: Using the given what 😄

internal/physical/pebbledb/pebbledb.go
cipherboy: @satoqz No, sadly this won't work because the return of `newLister` happens before `cursor` is assigned (when `lister.Start()` is catually called during `lister.ListPage(...)`) and so will be `nil` in the return and thus never be closed and thus leaks.

(I had thought about it as well but was unable to make it work with a shared implementation like this. You'd have to create a double pointer, ensuring it is not `nil` on the return and then dereferencing it on return.).

The alternative is passing the two missing functions into `newLister(...)` making it more like a `buildLister(...)` and then just calling `ListPage()` and returning the results (and do the defer while we're there), but that's a mess to read I think.

sdk/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,753 @@
...
+	return b.notifyWrite(entry.Key, true)
+}
+
+func (b *Backend) putInternal(ctx context.Context, view pebble.Writer, entry *physical.Entry) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
+	if err := view.Set([]byte(entry.Key), entry.Value, &pebble.WriteOptions{
+		Sync: b.sync,
```
satoqz: I think nothing ever sets `b.sync`? The doc comment says the default is `true` but that is only actualized if we pass a `nil` `WriteOptions`.

Looking into it, IIUC this is about syncing the WAL to disk only, which.. seems like a good idea to keep doing?

changelog/3879.txt
cipherboy: We can (make it migrate only by manually handling file), though this PR doesn't do that. I was going to test it later to see what actually breaks but `TestMigration/Default` runs `file->inmem` which is 98% the same now.

sdk/physical/inmem/inmem.go
cipherboy: @satoqz 🤔 Hmm, I was thinking somewhere overwrote this to guarantee smaller parallelism, but I don't see that happening and its marked `const`. So yes, I'll remove this.

sdk/physical/testing.go
```diff
@@ -855,10 +900,10 @@ func testTransactionalMixedWriters(t testing.TB, b TransactionalBackend) {
...
 	var wgReaders sync.WaitGroup
 	var done atomic.Bool
-	numFiles := 25
-	numWriters := 100
-	numWrites := 25
-	numReaders := 25
+	numFiles := 2
+	numWriters := 2
+	numWrites := 2
+	numReaders := 2
```
cipherboy:
```suggestion
	numFiles := 25
	numWriters := 100
	numWrites := 25
	numReaders := 25
```

website/content/docs/platform/k8s/helm/examples/standalone-audit.mdx
```diff
@@ -26,8 +26,8 @@ server:
         cluster_address = "[::]:8201"
       }
 
-      storage "file" {
-        path = "/openbao/data"
+      storage "pebbledb" {
+        directory = "/openbao/data"
```
cipherboy:
```suggestion
        path = "/openbao/data"
```

website/content/docs/configuration/storage/pebbledb.mdx
```diff
@@ -0,0 +1,49 @@
...
+  will not be persisted.
+
+## `pebbledb` examples
+
+This example shows the PebbleDB storage backend being mounted at
+`/mnt/openbao/data`.
+
+```hcl
+storage "pebbledb" {
+  directory = "/mnt/openbao/data"
```
cipherboy:
```suggestion
  path = "/mnt/openbao/data"
```

website/content/docs/configuration/storage/pebbledb.mdx
```diff
@@ -0,0 +1,49 @@
...
+}
+```
+
+Even though OpenBao's data is encrypted at rest, you should still take appropriate
+measures to secure access to the filesystem.
+
+## `pebbledb` parameters
+
+- `directory` `(string: <required>)` – The absolute path on disk to the
```
cipherboy:
```suggestion
- `path` `(string: <required>)` – The absolute path on disk to the
```

website/content/docs/configuration/storage/pebbledb.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+**No High Availability** – the PebbleDB backend does not support high
+availability. Only a single OpenBao node may use this data store.
+
+:::
+
+
+```hcl
+storage "pebbledb" {
+  directory = "/mnt/openbao/data"
```
cipherboy:
```suggestion
  path = "/mnt/openbao/data"
```

sdk/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,753 @@
...
+	_ physical.Backend              = &Backend{}
+	_ physical.TransactionalBackend = &Backend{}
+)
+
+// NewBackend constructs a PebbleDB backend using the given
+// API client, server address, credentials, and database.
+func NewBackend(conf map[string]string, logger log.Logger) (physical.Backend, error) {
+	var err error
+
+	dir, ok := conf["directory"]
```
cipherboy:
```suggestion
	path, ok := conf["path"]
```

website/content/docs/configuration/storage/pebbledb.mdx
cipherboy: I think that's fine, yeah. I think both are mostly equivalent if I'm not missing something?

internal/physical/pebbledb/pebbledb.go
satoqz: Well, oops, fair enough, let's keep `SeekPrefix` 😭

Though I'd revert the argument to `Start()` then to keep it simple.

website/content/docs/configuration/storage/pebbledb.mdx
satoqz: @cipherboy yours kind of read like I could still have multiple nodes but all but one would need to specify a different backend 😄

changelog/3879.txt
satoqz: Gotcha, that's fine.

internal/physical/pebbledb/pebbledb.go
cipherboy: @satoqz If we drop the argument, we lose the nice inline construction form.

So I'm leaving this as-is and someone smarter than me can revise later ;-)

internal/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,757 @@
...
+
+	"github.com/cockroachdb/pebble"
+	"github.com/cockroachdb/pebble/vfs"
+	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-secure-stdlib/parseutil"
+)
+
+/*
+
+The PebbleDB backend does not implement full read/write interactive
```
satoqz:
```suggestion
The pebble package alone does not implement full read/write interactive
```

perhaps? "backend" reads like we're referring to the code that follows below.

internal/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,757 @@
...
+
+	defer closer.Close() //nolint:errcheck // nothing we can do here
+
+	cloned := make([]byte, len(data))
+	copy(cloned, data)
+
+	return &physical.Entry{
+		Key:   key,
+		Value: cloned,
+	}, nil
```
satoqz:
```suggestion
	return &physical.Entry{
		Key:   key,
		Value: slices.Clone(data),
	}, nil
```

website/content/docs/configuration/storage/pebbledb.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+## `pebbledb` parameters
+
+- `path` `(string: <required>)` – The absolute path on disk to the
+  directory where the data will be stored. When this takes the value
+  `:memory:`, OpenBao will create an in-memory ephemeral database and values
+  will not be persisted.
+- `read_only` `(bool: false)` - When set to `true`, this loads the PebbleDB
+  instance as read-only, preventing updates. This is likely not useful in
+  most scenarios.
```
satoqz, line 41: Do we want to bother with supporting it then? Or did you have some niche use case in mind 🤔

internal/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,757 @@
...
+
+	return keys, nil
+}
+
+func (t *Transaction) Delete(ctx context.Context, key string) error {
+	if !t.writable {
+		return physical.ErrTransactionReadOnly
+	}
+
+	key = path.Clean(key)
```
satoqz: Why do we call clean here but not on put or elsewhere?

internal/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,757 @@
...
+
+	// Release the snapshot early; we don't need it any more.
+	if err := t.snapshot.Close(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// notifyWrite must be called while holding at least a read lock. It all
```
satoqz:
```suggestion
// notifyWrite must be called while holding at least a read lock. It notifies all
```

internal/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,757 @@
...
+	// semantics.
+	parts := strings.Split(written, "/")
+
+	var base string
+	suffix := parts[len(parts)-1]
+	if len(parts) > 1 {
+		base = strings.Join(parts[0:len(parts)-1], "/")
+	}
+
+	afterMap, present := t.lists[base]
```
satoqz, line 677: So `base` doesn't end in a `/`, but keys in `t.lists` may, don't we fail lookup?

internal/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,757 @@
...
+	// would've included this entry. We can't use path.Base(...) because
+	// we don't want to implicitly call path.Clean(...) and deal with its
+	// semantics.
+	parts := strings.Split(written, "/")
+
+	var base string
+	suffix := parts[len(parts)-1]
+	if len(parts) > 1 {
+		base = strings.Join(parts[0:len(parts)-1], "/")
+	}
```
satoqz: I wonder if `strings.CutLast` is easier since we only care about the last segment and the base of it.

internal/physical/pebbledb/pebbledb.go
cipherboy: This is a mistake. The first version I had used `path.*` rather liberally; I don't think its wrong to call Clean, but it makes things much harder and I think an upper layer handles calls to clean for us (I think by virtue of using `Views` for one).

I've dropped it.

internal/physical/pebbledb/pebbledb.go
cipherboy: I think so; I've fixed this.

internal/physical/pebbledb/pebbledb.go
cipherboy: I think that does what we want, thanks! I was looking for `Right` or `Split*` lol.

website/content/docs/configuration/storage/pebbledb.mdx
cipherboy: @satoqz A filesystem can be marked read-only and OpenBao started from it; I was thinking using Pebble's native support would be good? Maybe not that good of a reason, but was thinking exploring snapshots (well, the equivalent, backups of the pebbledb database) would be a good use case for `read_only` mode. 

Tbh not sure if pebbledb is copy-safe, I haven't looked.

internal/physical/pebbledb/pebbledb.go
```diff
@@ -0,0 +1,761 @@
...
+	// We need to validate both forms of listing to check for conflicts, with
+	// and without a slash. Most calls to list will have a slash, though some
+	// may not. By virtue of join, I don't think we'll have one in base unless
+	// we have double path segments (//), which is a bit undefined in
+	// physical.Backend in general.
+	bases := []string{base}
+	if len(base) > 0 {
+		// Add a trailing slash as an alternative.
+		bases = append(bases, base+"/")
+	}
```
satoqz: When we've listed without a trailing slash, aren't the only entries we'll ever see in `listResults` some of `""` (if an entry exists at the exact listed path) and `"/"` (if there exist sub-path entries)?

In this case, is collision detection based on `suffix` (which wouldn't be one of the above strings) still correct?

cipherboy: @satoqz Hmm, I don't know what we want to do here. The fact that `foo` returns `"", "/"` in:

```
    // List should have both a key and a subtree.
    testListAndPage(t, b, "foo", []string{"", "/"})
```

is not tested behavior (if you add it around like 221 in main in `sdk/physical/testing.go`) and it fails on PostgreSQL but passes on Raft:

```
        	            				/home/cipherboy/GitHub/cipherboy/openbao/internal/physical/postgresql/postgresql_test.go:453
        	Error:      	Not equal: 
        	            	expected: []string{"", "/"}
        	            	actual  : []string(nil)
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,5 +1,2 @@
        	            	-([]string) (len=2) {
        	            	- (string) "",
        	            	- (string) (len=1) "/"
        	            	-}
        	            	+([]string) <nil>
```

i.e., you get an empty list result.

This to me seems like the expected behavior: you're listing not a "directory" but an entry and so naturally assume that an entry is not a "directory" and thus should be empty?

The fact that we (ab)use it in invalidation tests seems a bit like an antipattern to me.

Thoughts?

satoqz: @cipherboy do we want to ignore list calls not ending in slash for transaction conflict detection then, arguing it is undefined behavior? I think we could get away with that?

cipherboy: @satoqz That works for me, let me do that.

satoqz (approved):

cipherboy:
Thanks; bypassing merge requirements over linter's Issues given our inability to bypass that.

Issue #2835, opened 2026-04-08
OCI plugin download config for multiple OS/arch targets


sbb-ue57062:
**Describe the bug**
The OCI-Plugin of AWS runs into hash-verification issues.
This issue keeps openbao sealed.

The image, that is pulled, is made by a local company repository / mirror. Not the github container repository directly. (ghcr.io)

Error log:
`2026-04-07T13:34:21.811Z [ERROR] core: post-unseal setup failed: error="failed to download OCI plugins: failed to download plugin \"secret-aws\": plugin hash mismatch: expected 2fed276f4faf19e571a2f272802ba1c654955497ed6c9547a7f4472dfc4581c9, got 991732feb606737baf727cba0ffa99def84f440d50c21c12ff0c4e7c5bdcc0f1"`

Our raft-config contains:
```
plugin "secret" "aws" {
  image = "company-mirror.com/openbao/openbao-plugin-secrets-aws"
  version = "v0.2.0"
  binary_name = "openbao-plugin-secrets-aws"
  sha256sum = "2fed276f4faf19e571a2f272802ba1c654955497ed6c9547a7f4472dfc4581c9"
}
```

Our current fix is to change it to the expected one.
And sometimes, after the next complete restart, he error is in "reverse", where the originally used hash is being accepted only.
It seems to only fluctuate between those two:
2fed276f4faf19e571a2f272802ba1c654955497ed6c9547a7f4472dfc4581c9
991732feb606737baf727cba0ffa99def84f440d50c21c12ff0c4e7c5bdcc0f1

**To Reproduce**
1. During initialisation, after complete scale-down and again scaling up or when starting with an empty PVC
2. See error

**Expected behavior**
No errors during startup, with a previously working configuration

**Environment:**
* OpenBao Server Version (retrieve with `bao status`):
* OpenBao CLI Version (retrieve with `bao version`):
* Server Operating System/Architecture:

OpenBao server configuration file(s):

```hcl
# Paste your OpenBao config here.
# Be sure to scrub any sensitive values
```

**Additional context**

cipherboy:
@sbb-ue57062 This sounds like a bad registry, especially the behavior of two different hashes for the same image.

satoqz:
@sbb-ue57062 have you compared the downloaded files, e.g., using https://diffoscope.org/?

A manual check shows that `2fed276f4faf19e571a2f272802ba1c654955497ed6c9547a7f4472dfc4581c9` appears to be the correct hash FWIW.

[...]

PR #3881, opened 2026-08-25
Add RFC on Plugin Management Improvements

satoqz:
## Description

As promised to @cipherboy, also cc @JanMa.

[Rendered](https://github.com/Ki-Reply-GmbH/openbao/blob/rfc-plugin-improvements/website/content/community/rfcs/plugin-improvements.mdx)

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/community/rfcs/plugin-improvements.mdx
+++ b/website/content/community/rfcs/plugin-improvements.mdx
@@ -1,0 +1,304 @@
+---
+sidebar_label: Plugin Management Improvements
+description: |-
+  An OpenBao RFC on several improvements to plugin binary installation and
+  management.
+---
+
+# Plugin Management Improvements
+
+## Summary
+
+This RFC proposes several changes to OpenBao's plugin system that make external
+binary plugins easier to operationalize. This includes relaxing the requirements
+around presently mandatory plugin binary checksums, enabling OCI-based plugins
+to be pinned by manifest digest, a new command to prune unused OCI plugin
+caches, and a way to automatically upgrade mounts to newer plugin versions.
+
+## Problem Statement
+
+In v2.5, we introduced [declarative plugin management and OCI-based
+plugin distribution](./config-plugins.mdx). In v2.6, we introduced [KMS
+plugins](./auto-unseal-plugins.mdx), building on top of declarative plugin
+management. In v2.7, we will remove several builtin auth, secrets and KMS
+plugins from OpenBao and distribute them as external plugins instead.
+
+A result of these changes is that many operators will be forced to start
+integrating external plugins into their deployments. While major improvements
+to plugin management were already made in v2.5, we'd like to make additional
+improvements to plugin management as part of v2.7 in an effort to ease the
+challenges of migration.
+
+We identify several pain points within the plugin system that are worth
+addressing:
+
+1. OpenBao requires that the `sha256sum` parameter be provided in `plugin`
+   stanzas such that a plugin binary may be executed. This is more of a chore
+   for operators than a security feature:
+   - While OpenBao requires that its plugin binaries be checked before
+     execution, there is no conventional mechanism or advice to check the digest
+     of the OpenBao binary itself before execution. As a result, any proposed
+     security benefits of plugin digest verification pre-execution vanish if the
+     main application's binary is not held to the same standards.
+   - Conventionally, cryptographic verification in software installation is
+     performed as part of the initial installation process only, i.e., when
+     software artifacts pulled from the network are stored on the local system
+     for later use. It is not typical to re-validate that an artifact _still_
+     matches the trusted digest that was previously verified each time it is put
+     to use.
+   - The check that runs shortly before plugin execution occurs independently
+     of subsequent plugin execution, reading the plugin binary once to determine
+     its digest and twice for execution. This approach is fundamentally subject
+     to a Time-of-Check-Time-of-Use problem.
+   - [_"Protecting against malicious plugins or code execution on the underlying
+     host"_](/docs/internals/security) is explicitly not included in OpenBao's
+     security model. The mandatory use of a "security feature" that does not
+     contribute to the stated security model of the application is contradictory
+     and hindering.
+   - Finally, it is worth noting that the `sha256sum` parameter was originally
+     introduced as an API parameter (under `sys/plugins/catalog`), and appeared
+     in server configuration only much later when declarative plugins were
+     introduced. Plugin configuration via the API opposed to configuration via
+     the config file follow different threat models, the former must make an
+     effort to defend against arbitrary remote code execution on the server's
+     underlying host, the latter assumes that such privileges were already
+     granted to the author of the config. Not accounting for this difference
+     is what makes the parameter a chore when used with declarative plugin
+     registration.
+
+2. Similarly, OpenBao's OCI plugin downloader requires that the `sha256sum`
+   parameter be provided in `plugin` stanzas to verify artifact integrity when
+   downloading, and uses the SHA-256 digest as a cache sentinel to skip later
+   re-downloads. It does not however support directly pinning the plugin's OCI
+   manifest by digest instead, which is the OCI-native checksumming mechanism
+   that operators familiar with conventional OCI tooling are already used to.
+   Pinning by OCI manifest digest also has the advantage of multi-arch binary
+   support within a single pinned digest, a capability that the `sha256sum`
+   parameter does not replicate.
+
+3. OpenBao does not offer a native way to prune unused OCI cache entries from
+   disk, growing plugin directory size indefinitely across upgrades unless
+   operators manually prune cache entries from an undocumented cache structure.
+
+4. The `plugin_auto_register` configuration option defaults to `false` and
+   tends to be forgotten about, leading to `plugin` configuration stanzas having
+   no effect. The default is surprising as the only use of `plugin` stanzas
+   in combination with `plugin_auto_register = false` is to _automatically
+   download_ OCI plugins but to then _manually register_ them.
+
+5. When upgrading secrets, auth or database plugins, associated mounts and
+   database connections remain pinned to the previous plugin version and do
+   not automatically upgrade to a newer plugin version even if the older one is
+   unregistered. Instead, operators must tune each mount and database connection
+   individually to ensure it is configured to use the new plugin version.
+
+## User-facing Description
+
+We address each problem described above individually:
+
+1. The `sha256sum` parameter in `plugin` configuration stanzas becomes optional
+   on non-OCI plugins, i.e., those that use `command` but not `image`. It
+   remains required when registering plugins via the plugin catalog API. Further
+   changes to the handling of `sha256sum` with OCI-based plugins are outlined in
+   the below point.
+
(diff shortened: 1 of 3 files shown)


JanMa (dismissed):
Lgtm, those are great improvements

website/content/community/rfcs/plugin-improvements.mdx
```diff
@@ -0,0 +1,292 @@
...
+     image = "ghcr.io/openbao/openbao-plugin-secrets-aws:v0.3.1@sha256:bfa896f87c55f511f1400d62ccb0dc0503c87cb7ee7955ebdb4f75e987b9a436"
+   }
+   ```
+
+3. We add a new CLI subcommand, `bao plugin prune`. This command removes all
+   cached OCI plugins from the plugin directory (including symlinks) that are
+   not referenced by the passed config(s) anymore. Conceptually, this is the
+   counterpart to the existing `bao plugin init` command.
+
+4. The default value of `plugin_auto_register` changes to `true`. Notably,
```
klaus-sap, line 151: Changing the default value would be an incompatible change, i.e. not backwards compatible. Operators may want to declare and register plugins at different points in time. What about a new option `auto_register` in the `plugin` stanza to allow overwriting the system default?

cipherboy: @klaus-sap Wouldn't (if this defaults to true) it have the same effect? :thinking:

klaus-sap: @cipherboy No, this new option wouldn't have a default value. If it is not present in the stanza, then the existing system parameter `plugin_auto_register` will be taken

satoqz: > Operators may want to declare and register plugins at different points in time.

As mentioned in the RFC:

> *the only use of `plugin` stanzas in combination with `plugin_auto_register = false` is to _automatically download_ OCI plugins but to then _manually register_ them.*

So for one, this is scoped entirely to OCI-based plugins.

> Operators may want to declare and register plugins at different points in time.

I think you're missing that *registration* does not require *declaration*. If you're not going to use declarative registration, there's no reason to declare `plugin` stanzas. You can do it regardless, via the API, no matter the server config.

The single exception to this is OCI plugin downloading, which you can perform ahead of time via a separate config file and `bao plugin init` if you'd prefer. Still, the entire value proposition of even using OpenBao's downloader opposed to your own installation mechanism is that it neatly integrates with declarative auto-registration. Not using those together is like fueling a car only to continue pushing it manually.

> What about a new option `auto_register` in the `plugin` stanza to allow overwriting the system default

Per-plugin-stanza auto-registration is a very contrived scenario and not worth supporting, IMO. I can't imagine not wanting a uniform registration pattern across all plugins.

cipherboy: > If you're not going to use declarative registration, there's no reason to declare `plugin` stanzas. You can do it regardless, via the API, no matter the server config.

And in particular, you can take advantage of the CLI tools for download without using `plugin` stanzas in the _server_'s config. You just need them in _a_ config passed to the cli helper. 

I wonder if we make `plugin_auto_download == true ==> plugin_auto_register.default = true` then?

That'd solve the little bit of not wanting to use auto-download and thus not getting different behavior if you have some out-of-band distribution mechanism (e.g., RPMs) and auto-template configs under this new stanzas for _everyone_ but not necessarily want them to have all these plugins running around. 

Otherwise, I'm not really picky. I think there's probably material improvements being made here either way.

wslabosz-reply (dismissed):
gorgeous

website/content/community/rfcs/plugin-improvements.mdx
```diff
@@ -0,0 +1,292 @@
...
+     to use.
+   - The check that runs shortly before plugin execution occurs independently
+     of subsequent plugin execution, reading the plugin binary once to determine
+     its digest and twice for execution. This approach is fundamentally subject
+     to a Time-of-Check-Time-of-Use problem.
+   - [_"Protecting against malicious plugins or code execution on the underlying
+     host"_](/docs/internals/security) is explicitly not included in OpenBao's
+     security model. The mandatory use of a "security feature" that does not
+     contribute to the stated security model of the application is contradictory
+     and hindering.
```
cipherboy, line 57: I'd add a comment here about the safety feature being useful at the API level e.g., to prevent OpenBao from executing untrusted/unexpected code, but that the config file is otherwise very highly trusted and at a sufficient trust level to bypass the requirement as well. 

In reality, there's probably lots of arbitrary files that an attacker could guess and have OpenBao execute, though, on the host system, but it'd be a stuffing attack to do, so operators should rely on defense-in-depth here or disallow root namespace access.

website/content/community/rfcs/plugin-improvements.mdx
satoqz: > I wonder if we make `plugin_auto_download == true ==> plugin_auto_register.default = true` then?

I think this means you don't get `plugin_auto_register` defaulted to true even if you don't use OCI plugins at all and hence haven't cared to configure `plugin_auto_download`?

I'd prefer if we avoid adding rules that operators must understand in an attempt to derive the ~wanted behavior and went with a hard default instead.

website/content/community/rfcs/plugin-improvements.mdx
```diff
@@ -0,0 +1,292 @@
...
+   cached OCI plugins from the plugin directory (including symlinks) that are
+   not referenced by the passed config(s) anymore. Conceptually, this is the
+   counterpart to the existing `bao plugin init` command.
+
+4. The default value of `plugin_auto_register` changes to `true`. Notably,
+   the sibling option `plugin_auto_download` remains unchanged and `false` by
+   default.
+
+5. When the `plugin_version` API parameter under `sys/mounts` or `sys/auth`
+   is set to "latest", mounts will use the latest installed version of a
```
cipherboy:
```suggestion
   is set to `latest`, mounts will use the latest installed version of a
```

website/content/community/rfcs/plugin-improvements.mdx
```diff
@@ -0,0 +1,292 @@
...
+   If a tag is provided, the `version` parameter may be omitted and, if unset,
+   will default to the tag. If a digest is provided, the `sha256sum` parameter
+   may be omitted.
+
+   As a result, the below is all that is needed to register an OCI-based plugin,
+   assuming the image tag matches the plugin's version:
+
+   ```hcl
+   plugin "secret" "aws" {
+     image = "ghcr.io/openbao/openbao-plugin-secrets-aws:v0.3.1@sha256:bfa896f87c55f511f1400d62ccb0dc0503c87cb7ee7955ebdb4f75e987b9a436"
```
cipherboy, line 142: I think this is implicit, but might be worth calling out as a problem explicitly and as a solution: multi-arch images e.g., https://github.com/openbao/openbao/issues/2835?

cipherboy (dismissed):
Looks like a great set of improvements, thanks for writing up!

website/content/community/rfcs/plugin-improvements.mdx
cipherboy: Yeah, that's fair. I do think we probably made the wrong decision with `plugin_auto_register = false` in the first place, but we can kinda view this as feature flagging with defaults changing.

website/content/community/rfcs/plugin-improvements.mdx
satoqz: The RFC mentions multi-arch images in "Problem Statement" and in "Rationale and Alternatives", plus I've linked #2835 in "Related Issues". Anything specific you're missing?

website/content/community/rfcs/plugin-improvements.mdx
klaus-sap: Logically, changing the default value would be an incompatible change, i.e. not backwards compatible as stated in line 12. I always try to avoid incompatible changes whenever possible :-)

website/content/community/rfcs/plugin-improvements.mdx
satoqz: Great point, added an extra bullet point to explain the separate threat models.

website/content/community/rfcs/plugin-improvements.mdx
cipherboy: Ah, no, apparently I missed that line, all good.

website/content/community/rfcs/plugin-improvements.mdx
cipherboy: With my academic hat on... I wonder what we call backwards compatible in this case? We've not renamed an option, so other than a _behavior change_ (doing more rather than less), existing configurations will continue to work as expected (and thus, from a config PoV, is backwards compatible). 

I think the only scenario where this wouldn't hold is if someone had plugin configs which worked for downloading (manually?) but never actually registered correctly, which would be entertaining? 🤔 I'm not sure I'd understand enough of when that'd be true, tbh.

Worst case, I think we can issue a change notice and state the default is changing in v2.8.

satoqz: I can change the wording to "largely backwards-incompatible" if we prefer, which is fair enough. We'll definitely call this out in the changelog either way.

klaus-sap: @satoqz "largely backwards-incompatible" seems to be a typo. Did you mean "largely backwards-compatible" or "slightly backwards-incompatible"? :-)
Anyway, I fear that not all operators will read the changelog...

satoqz: Yes, sorry, meant to drop the _in-_.

Not scanning the changelog of a critical application for relevant changes pre-upgrade is a failure on the operator who is being irresponsible, IMHO. We make it very easy for them to jump right to any breaking-ish entries by sorting changelog entries into categories.

While I generally agree that it's best to keep such changes to a minimum, we've ended up making several (much larger) breaking changes across the past couple releases, e.g., disabling API-driven audit devices. Moreover, v2.7 is building up to be more breaking than ever as we remove many built-in plugins. For better or worse, long-term operators should already know that not reading the OpenBao changelog = bad idea :)

This is of course not a pattern we should continue indefinitely, but I believe it's a necessary sacrifice as we re-architect OpenBao to be more maintainable and, frankly, more secure by cutting back on highly privileged API endpoints.

Moving on though, let's look at the broader picture of the release that this change would coincide with:

1. v2.7 is already bound to be a rather breaking release, greatly exceeding anything we'd "break" by changing the default value here.
2. The majority of breakage in v2.7 is plugin-related, likely resulting in more usage of the plugin system than ever before.

If anything, v2.7 is the best window we get to change the default value. We break a subset of deployments that uses a pattern so odd it is unlikely to even exist out in the wild, but have all new adopters of the plugin system benefit from the improved default.

Perhaps it is also good to mention that this does not affect KMS plugins at all. This parameter never applied to them.

satoqz: Either way, I've revised the wording around breaking and not breaking in the rationale section.

cipherboy (approved):

PR #3882, opened 2026-08-26
Remove ldap, kerb, and radius plugins to openbao-plugins

### LINKED ISSUE #2808: Deprecation for Externalization - Builtin LDAP, Kerberos, and RADIUS Auth+Secret Engines
Three of our largest external dependencies are LDAP, Kerberos, and RADIUS authentication libraries. While useful, the gap for declarative plugins have meant it is easier than ever to bring up OpenBao with external plugins. Many of these libraries are lightly maintained.

In the effort of trimming our binary size and reduce our supply chain footprint for the main binary, I think the removal (to `openbao-plugins`!) of these plugins would be great.

I'm proposing we deprecate them now and remove them in v2.7.0, aligning with KMS library removal.

---

To be clear: this is a deprecation from the main binary; the plugins will continue to be supported in `openbao-plugins`.

cipherboy:
## Description

My plan with the commented out tests is, where possible, to reintroduce them to `openbao-plugins` as external tests and eventually rewrite them in core against another auth method (likely JWT as it supports identity groups). 

## Rationale

Related: #3372 
Resolves: #2808

## Acknowledgements

       also allowed.


changelog/3882.txt
```diff
@@ -0,0 +1,12 @@
+```release-note:change
+secret/ldap: The built-in LDAP secrets plugin has been removed from the main OpenBao binary distribution. It will be included in [openbao-plugins](https://github.com/openbao/openbao-plugins) going forward.
```
wslabosz-reply:
```suggestion
secret/ldap: The built-in LDAP secret plugin has been removed from the main OpenBao binary distribution. It will be included in [openbao-plugins](https://github.com/openbao/openbao-plugins) going forward.
```

internal/command/base_predict.go
```diff
@@ -109,10 +109,8 @@ func (b *BaseCommand) PredictVaultAvailableAuths() complete.Predictor {
 		"cert",
 		"gcp",
 		"github",
-		"ldap",
 		"okta",
 		"plugin",
-		"radius",
 		"userpass",
```
wslabosz-reply, line 116: looking at this set, I'd reckon we probably should have also removed `aws`, `gcp`, `github`, `okta`, and only leave the builtins in place, but probably out of scope for this PR

internal/command/commands.go
```diff
@@ -161,14 +158,8 @@ var (
 
 func initCommands(ui, serverCmdUi cli.Ui, runOpts *RunOptions) map[string]cli.CommandFactory {
 	loginHandlers := map[string]LoginHandler{
-		"cert":       &credCert.CLIHandler{},
-		"kerberos":   &credKerb.CLIHandler{},
-		"kubernetes": &credKube.CLIHandler{},
```
wslabosz-reply, line 166: give back `kubernetes` please

wslabosz-reply (dismissed):
thanks @cipherboy 
I think it makes sense to revisit the tests but not in this PR

satoqz:
Do we want to add a notice to the relevant doc pages like we have on https://openbao.org/docs/next/configuration/seal/awskms, or do we plan on getting the openbao-plugins doc site up by 2.7, too?

cipherboy:
@satoqz I don't know where we're at with docs repo since @phil9909 is working on that, but it looks like the docs page update is suitably generic so I'll add it here.

wslabosz-reply (approved):

satoqz (approved):

PR #3884, opened 2026-08-26
Fix signing in Transit with External Keys

satoqz:
## Description

Fixes a couple quirks with signing via external keys in transit:

- Don't pre-hash external key input by default (by defaulting `hash_algorithm=none`) so keys that shouldn't receive pre-hashed input (e.g., Ed25519, MLDSA) don't do so. The plugin can otherwise perform the hashing itself, also allowing for the final remote KMS to see underlying messages if the plugin was configured that way.
  - One caveat here is that signing with RSA/ECDSA external keys will require an explicit `hash_algorithm=xxx` if the underlying provider rejects `crypto.Hash(0)` and does not set its own default. In the Transit->Transit case, this is fine, as the final Transit has a default for its native keys. In the Transit->PKCS#11 case, we haven't implemented a default and see errors if `hash_algorithm=none`. Might be worth adjusting in go-kms-wrapping.

- Correctly pass `rsa.PSSOptions` when `signature_algorithm=pss` and set the default to PSS like documented.

- Unrelated to external keys, but included here: HMAC signing would panic (non-fatally) if `signature_algorithm` was explicitly none.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/builtin/logical/transit/path_sign_verify.go
+++ b/internal/builtin/logical/transit/path_sign_verify.go
@@ -73,7 +73,26 @@ type batchResponseVerifyItem struct {
 	Reference string `json:"reference" mapstructure:"reference"`
 }
 
-const defaultHashAlgorithm = "sha2-256"
+func getHashAlgorithm(d *framework.FieldData, kt keysutil.KeyType) (keysutil.HashType, error) {
+	hashAlgorithmStr := d.Get("urlalgorithm").(string)
+	if hashAlgorithmStr == "" {
+		hashAlgorithmStr = d.Get("hash_algorithm").(string)
+		if hashAlgorithmStr == "" {
+			hashAlgorithmStr = d.Get("algorithm").(string)
+		}
+	}
+
+	if hashAlgorithmStr == "" {
+		return kt.DefaultHashAlgorithm(), nil
+	}
+
+	hashAlgorithmType, ok := keysutil.HashTypeMap[hashAlgorithmStr]
+	if !ok {
+		return 0, fmt.Errorf("invalid hash algorithm %q", hashAlgorithmStr)
+	}
+
+	return hashAlgorithmType, nil
+}
 
 func (b *backend) pathSign() *framework.Path {
 	return &framework.Path{
@@ -103,8 +122,7 @@ derivation is enabled; currently only available with ed25519 keys.`,
 			},
 
 			"hash_algorithm": {
-				Type:    framework.TypeString,
-				Default: defaultHashAlgorithm,
+				Type: framework.TypeString,
 				Description: `Hash algorithm to use (POST body parameter). Valid values are:
 
 * sha1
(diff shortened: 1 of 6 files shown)


internal/builtin/logical/transit/path_sign_verify.go
```diff
@@ -637,11 +635,13 @@ func (b *backend) pathVerifyWrite(ctx context.Context, req *logical.Request, d *
 			continue
 		}
 
+		prehashed := prehashed
```
cipherboy, line 639: For once shadowing actually does something useful!

sdk/helper/keysutil/policy.go
```diff
@@ -1618,15 +1644,27 @@ func (p *Policy) VerifySignatureWithOptions(derivationContext, input []byte, sig
...
+		}
+
+		if sigAlgorithm == "pss" {
+			opts.SignerOpts = &rsa.PSSOptions{
+				Hash:       algo,
+				SaltLength: saltLength,
+			}
+		} else {
+			opts.SignerOpts = algo
 		}
```
satoqz, line 1663: This will *suck* if we ever want to pass `*ed25519.Options` or `*mldsa.Options` to supply `Context` based on some generic parameter instead of specific `mldsa_context`, `ed25519_context` ones, since we just wouldn't know the correct type to pass. Transit doesn't support Ed25519ctx as-is, so it's okay for now? Thoughts?

satoqz: That is, long-term if this gets too complex we'll have to pre-export the public key of the external key to know how to call Sign, or introduce our own `SignerOpts` type (in g-k-w) that is algorithm-agnostic but carries common information beyond the hash function. But I'd like to avoid that for the time being since it is a nasty refactor.

cipherboy: @satoqz Is it me, or is the complexity all artificial due to `keysutil.Policy`'s signing interface to be an incrementally added to garbage API? 

If we drop this method, wouldn't we be able to take `SignerOpts` directly and be much better off? We could then add helpers of various sorts so Transit can construct them more easily.

We still have to know the underlying key type in the case of ed25519/... (and we would) but it'd be more reasonable. And yeah, maybe we should store the public key to be able to verify our signatures as well. I wonder if we have that as an option somewhere? 🤔

satoqz: I don't think keysutil's signing path adds much complexity here tbh, it's *fine*. Exporting the public key at creation/rotation time would be the complex part.

cipherboy: @satoqz Ah right, because we don't have access to anything nice in that layer. Sigh.

cipherboy: @satoqz It'd be a mess, but I kinda wonder if we want to make keysutil implement `kms` and then just use KMS APIs... :thinking: I don't know that its possible with context keys though, sadly.

satoqz: Maybe in OpenBao v3 😆

cipherboy (approved):
Thanks @satoqz!

PR #3898, opened 2026-08-28
Add more recent and past ecosystem news

karras:
## Description

Simply adds a few more new and older ecosystem news.

cc @klaus-sap, let me know if there are more ApeiroRA news we missed so far.

## Rationale

n/a

Resolves: n/a

## Acknowledgements

       also allowed.

### DIFF
--- a/website/content/ecosystem/news.mdx
+++ b/website/content/ecosystem/news.mdx
@@ -47,6 +47,34 @@ import { YouTubeTalk, NewsBlurb } from './lib/News.tsx';
         HashiCorp relicensed Vault from the open source MPL 2.0 to the Business
         Source License (BSL 1.1).
       </NewsBlurb>
+
+      <NewsBlurb title="Unlocking Secrets Management: Introducing OpenBao Managed Services on Azure and AWS" memberName="adfinis" link="https://www.adfinis.com/en/news/unlocking-secrets-management">
+        That is why Adfinis is thrilled to introduce the OpenBao Tailored
+        Managed Service, a fully managed and integrated, enterprise-grade
+        secrets management solution that delivers data sovereignty, cost
+        predictability, and operational excellence without the additional costs
+        associated with proprietary alternatives. Typically referred to as the
+        "Success Tax", where charging per "identity" or "client" forces teams
+        into "Architecture by License" workarounds, resulting in infrastructure
+        designed to fit a contract rather than their actual security
+        requirements.
+      </NewsBlurb>
+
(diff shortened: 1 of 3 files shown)


website/content/ecosystem/news.mdx
```diff
@@ -47,6 +47,34 @@ import { YouTubeTalk, NewsBlurb } from './lib/News.tsx';
...
+        costs, Wavelo partnered with Adfinis to migrate its proprietary secrets
+        management platform to OpenBao™. Adfinis co-developed the missing
+        enterprise features, including multi-tenancy and horizontal
+        scalability, by contributing them directly to the OpenBao open source
+        project. The migration eliminated vendor lock-in, halved maintenance
+        overhead, and cut licensing costs by 60%.
+      </NewsBlurb>
+
+      <NewsBlurb title="Vault nach OpenBao: die Tür ist offen, nur schmaler als sie aussieht" memberName="trutzio" link="https://trutz.io/tutorial/openbao/vault-nach-openbao-die-tuer-ist-offen.html">
+        Die OpenBao-Doku sagt, ab Vault 1.15.0 geht der Umstieg nicht mehr.
```
klaus-sap: @karras Translation to English?

karras: Shouldn't hurt having selected entries which are not always in English?

cipherboy: I'm happy taking other languages to try and grow the size and diversity of the community. I don't think we'll ever be able to offer OpenBao _server_ components (error messages in particular) in another language, though I could see maybe translating the UI and potentially more importantly the docs at some point :shrug:

satoqz: I suggest we prefix the title with a language indicator such as `[German]`, `[DE]`, just so it's clearer to those who won't immediately recognize the language.

I don't think translating the title to English is helpful, readers that expect English and don't know German will click on it and then be left in the dust. Similarly, German readers (in this case) might show more interest in regional articles and translating the heading to English would make those harder to notice among all the English articles.

klaus-sap: I agree not to translate the title, but I would translate the excerpt with annotation "(article in German)". Then non-German readers would get an impression what the linked article is about and may use automatic translation if they want to read the article.

karras: Many thanks for the suggestions! I've translated the excerpt and added a prefix using `[...]` thus it should visually stand out.

image

cipherboy (approved):
Thanks!

PR #3900, opened 2026-08-28
Fix backwards-incompatible request forwarding

### LINKED ISSUE #3899: Rolling upgrade 2.5.4 → 2.6.2 is impossible: 2.6.2 standby cannot forward to a 2.5.4 active node (Unimplemented: unknown service forwarding.RequestForwarding)
**Describe the bug**

In a 5-node Raft HA cluster running **2.5.4**, replacing a single standby node
with **2.6.2** produces a node that auto-unseals correctly, enters
`HA Mode: standby`, correctly identifies the active node, and replicates Raft
normally (committed/applied index matches the rest of the cluster exactly) — but
**can never be promoted to voter** and **cannot forward any request to the active
node**.

The 2.6.2 node logs, on every forwarded request:

```
[ERROR] core: error during forwarded RPC request:
  error="rpc error: code = Unimplemented desc = unknown service forwarding.RequestForwarding"
[ERROR] core: forward request error: error="error during forwarding RPC request"
```

The gRPC service `forwarding.RequestForwarding` is not resolvable on the 2.5.4
active node when called by a 2.6.2 client. The underlying TLS/gRPC connection
establishes successfully — this is an application-layer RPC contract mismatch,
not a transport, certificate, or clock problem (evidence under *Additional
context*).

Because autopilot derives per-server health from *echo* requests, which travel
over that same service, the consequence cascades:

```
forwarding RPC Unimplemented
  → no echo reaches the active node
    → no follower state recorded
      → node absent from `bao operator members`
      → autopilot: Healthy: false, Last Contact: 2h+, Last Term: 0, Last Index: 0
        → never stable for server_stabilization_time
          → never promoted to voter
```

**This makes a standby-first rolling upgrade a dead end rather than a slow
path.** A non-voter cannot win a Raft election, so the cluster can never reach a
2.6.2 leader by replacing standbys — each new 2.6.2 node simply accumulates as
another stuck non-voter while failure tolerance degrades.

**To Reproduce**

1. Run a 5-node Raft HA cluster on 2.5.4 (integrated storage, auto-unseal, no
   autopilot configuration — all defaults).
2. Replace exactly one standby node with 2.6.2, joining via `retry_join`.
3. Observe the new node unseal, reach `HA Mode: standby`, and replicate Raft to a
   matching index. Run `bao status` on it — everything looks correct:

   ```
   Version                  2.6.2
   Commit Date              2026-08-18T15:48:19Z
   Storage Type             raft
   HA Enabled               true
   HA Cluster               https://10.90.43.230:8201
   HA Mode                  standby
   Active Node Address      https://10.90.43.230:8200
   Raft Committed Index     7211765
   Raft Applied Index       7211765
   ```

4. Wait. The node is never promoted. Observed stuck for **>2 hours**; the default
   `server_stabilization_time` is 10s.

5. Run `bao operator raft list-peers` — the node is present, as a non-voter:

   ```json
   { "address": "10.90.28.108:8201", "leader": false,
     "node_id": "i-REDACTED-new", "protocol_version": "3", "voter": false }
   ```

6. Run `bao operator raft autopilot state` against the active node — the node is
   unhealthy on every criterion except the one an operator is most likely to
   check by hand (index):

   ```
   Healthy:            false
   Failure Tolerance:  1
   Leader:             i-REDACTED-leader
   Voters:
      i-REDACTED-leader
      i-REDACTED-a
      i-REDACTED-b
      i-REDACTED-c
   Servers:
      i-REDACTED-a                       <- healthy 2.5.4 standby, for comparison
         Status:            voter
         Node Status:       alive
         Healthy:           true
         Last Contact:      3.50651974s
         Last Term:         13
         Last Index:        7212337
         Version:           2.5.4
         Node Type:         voter
      i-REDACTED-new                     <- the 2.6.2 node
         Address:           10.90.28.108:8201
         Status:            non-voter
         Node Status:       alive
         Healthy:           false
         Last Contact:      2h8m29.350370966s
         Last Term:         0
         Last Index:        0
         Version:           2.5.4
         Node Type:         voter
   ```

7. Run `bao operator members` — the 2.6.2 node is **absent entirely**; the four
   2.5.4 nodes echo normally.

8. Send any request to the 2.6.2 node. See the error in its log:

   ```
   [ERROR] core: error during forwarded RPC request:
     error="rpc error: code = Unimplemented desc = unknown service forwarding.RequestForwarding"
   [ERROR] core: forward request error: error="error during forwarding RPC request"
   ```

> **Deployment note:** our nodes are immutable — the instance is terminated and a
> replacement boots with the new binary and rejoins via `retry_join`, rather than
> the in-place binary swap the HA upgrade guide describes. We do not believe this
> is material: the replacement node joined the Raft configuration cleanly,
> unsealed, reached `HA Mode: standby`, and replicated to a matching index. The
> only thing that fails is the forwarding RPC. Running the identical
> node-replacement procedure at 2.5.4 produces a fully healthy voter every time.

**Expected behavior**

Either:

- a 2.6.2 standby should be able to forward requests to a 2.5.4 active node, so
  that the standard standby-first rolling upgrade works; **or**
- if mixed-version request forwarding is intentionally unsupported across this
  boundary, the incompatibility should be **detected and surfaced loudly** — a
  startup error, a refusal to join, or at minimum a documented note in the 2.6.0
  release notes and the HA upgrade guide — rather than degrading into a node that
  looks healthy by index, is silently unpromotable, and quietly redirects client
  traffic.

**Environment:**

* OpenBao Server Version (retrieve with `bao status`): cluster `2.5.4`,
  replacement node `2.6.2` (commit date `2026-08-18T15:48:19Z`)
* OpenBao CLI Version (retrieve with `bao version`): `2.6.2`
* Server Operating System/Architecture: Ubuntu on AWS EC2, `linux/amd64`

Additional environment detail:

* Storage: integrated Raft
* Seal: `awskms` auto-unseal
* HA: 5 nodes; autopilot entirely at defaults (no autopilot configuration set)
* Namespaces: one non-root namespace in use

```hcl
cluster_addr = "https://<PRIVATE_IP>:8201"
api_addr     = "https://<PRIVATE_IP>:8200"
ui           = true

listener "tcp" {
  address            = "[::]:8200"
  cluster_address    = "[::]:8201"
  tls_cert_file      = "<...>/fullchain.pem"
  tls_key_file       = "<...>/privkey.pem"
  tls_client_ca_file = "<...>/ca.pem"
  telemetry {
  }
}

storage "raft" {
  path    = "<...>"
  node_id = "<EC2_INSTANCE_ID>"
  retry_join {
    auto_join             = "provider=aws region=<REGION> tag_key=<KEY> tag_value=<VALUE>"
    auto_join_scheme      = "https"
    auto_join_port        = 8200
    leader_tls_servername = "<FQDN>"
  }
}

seal "awskms" {
  region     = "<REGION>"
  kms_key_id = "<KEY_ID>"
}
```

**Additional context**

### Suspected root cause: the gRPC service was renamed in #2458

[#2458 — "Split off request forwarding from vault package"](https://github.com/openbao/openbao/pull/2458)
(merged 2026-03-16, milestone v2.6.0) moved request forwarding into its own
package. That changed the protobuf package, and with it the wire-level gRPC
service name:

```diff
--- a/vault/request_forwarding_service.proto
+++ b/vault/forwarding/request_forwarding_service.proto
-option go_package = "github.com/openbao/openbao/vault";
+option go_package = "github.com/openbao/openbao/vault/forwarding";
-package vault;
+package forwarding;
```

```diff
 # vault/forwarding/request_forwarding_service_grpc.pb.go
-  RequestForwarding_ForwardRequest_FullMethodName = "/vault.RequestForwarding/ForwardRequest"
-  RequestForwarding_Echo_FullMethodName           = "/vault.RequestForwarding/Echo"
+  RequestForwarding_ForwardRequest_FullMethodName = "/forwarding.RequestForwarding/ForwardRequest"
+  RequestForwarding_Echo_FullMethodName           = "/forwarding.RequestForwarding/Echo"

 var RequestForwarding_ServiceDesc = grpc.ServiceDesc{
-  ServiceName: "vault.RequestForwarding",
+  ServiceName: "forwarding.RequestForwarding",
```

gRPC dispatches on the fully-qualified service name. A 2.5.x active node
registers only `vault.RequestForwarding`; a 2.6.x standby calls
`forwarding.RequestForwarding`; the server returns `Unimplemented`. That matches
the observed error string exactly.

If this is correct, the consequences are broader than the symptom reported here:

1. **The incompatibility is symmetric.** A 2.5.x standby calling a 2.6.x active
   node would fail identically, since the 2.6.x server no longer registers
   `vault.RequestForwarding`. Upgrading the active node first therefore does not
   help — there is no node ordering that avoids a broken window.
2. **2.5.x and 2.6.x nodes cannot coexist in one HA cluster at all**, which
   makes a zero-downtime rolling upgrade across that boundary impossible.
3. **The change was labelled `pr/no-changelog`**, so it does not appear in the
   2.6.0 release notes or anywhere in the upgrade documentation. Operators have
   no way to learn about it before attempting the upgrade.
4. Raft replication is unaffected (separate transport), which is why the broken
   node's indexes track the leader perfectly and every conventional health
   signal looks fine.

I have not verified this by building both versions and inspecting the registered
service descriptors — it is inferred from the diff and the error string. Happy to
confirm if that would help.

### Related prior discussion

Mixed-version HA safety has been raised before, in the opposite direction
(older node taking leadership after a newer one):
[#2858](https://github.com/openbao/openbao/issues/2858),
[#2859](https://github.com/openbao/openbao/issues/2859),
[#2865](https://github.com/openbao/openbao/pull/2865) (closed unmerged).
Notably, the PoC attached to #2859 reported a clean, zero-downtime
v2.5.0 → v2.6.0 forward upgrade on k3d. I suspect that result held only because
all clients were in-cluster and could follow the 307 redirect fallback, which
masks the forwarding failure entirely.

### Transport, certificates, and clock are ruled out

**Connection topology is identical** between the broken 2.6.2 node and a healthy
2.5.4 standby — one outbound connection to the active node's cluster port for
request forwarding, two inbound for Raft:

```
# 2.6.2 node (broken)
ESTAB      10.90.28.108:46450   10.90.43.230:8201
ESTAB      10.90.28.108:8201    10.90.43.230:49942
ESTAB      10.90.28.108:8201    10.90.43.230:49958

# 2.5.4 standby (healthy)
ESTAB      10.90.4.86:60790     10.90.43.230:8201
ESTAB      10.90.4.86:8201      10.90.43.230:56102
ESTAB      10.90.4.86:8201      10.90.43.230:56098
```

The forwarding connection is established. The RPC on it is not implemented.

**Clock is synchronised** (`timedatectl`: `System clock synchronized: yes`,
`NTP service: active`). **No TLS or x509 errors** appear anywhere in the 2.6.2
node's log.

### Client-visible impact

When a standby cannot forward, it falls back to a **307 redirect** to the active
node's `api_addr`:

```
$ curl -sk -o /dev/null -w '%{http_code} -> %{redirect_url}\n' -X PUT \
    -H "X-Vault-Token: ..." -H "X-Vault-Namespace: <ns>" \
    -d '{"data":{"ts":"probe"}}' \
    https://10.90.28.108:8200/v1/test-kv/data/canary
307 -> https://10.90.43.230:8200/v1/test-kv/data/canary
```

Two practical consequences worth highlighting:

1. **The failure is silent to CLI users.** `bao kv put` against the broken node
   reports `Success!`, because the CLI transparently follows the redirect. An
   operator validating a node after replacement will conclude it is healthy.
2. **`api_addr` is frequently not client-routable.** In our deployment (and in
   any topology where nodes advertise private addresses behind a load balancer),
   the redirect target is unreachable from external clients, so requests that
   land on the affected node fail outright. Server-side audit entries show
   `"error":"please forward to the active node"` for real client traffic.

### The documented verification procedure does not catch this

This is the part we think matters most. [Upgrading OpenBao HA
installation](https://openbao.org/docs/upgrading/ha-upgrade/) prescribes exactly
this order — "Perform these steps on each standby" first, then the active node —
and gives this verification step for each upgraded standby:

> - Verify `bao status` shows correct Version and HA Mode is `standby`
> - Review the node's logs to ensure successful startup and unseal

**Both checks pass on the broken node.** As shown in step 3 above, `bao status`
reports `Version 2.6.2` and `HA Mode: standby`, and startup/unseal genuinely did
succeed — the forwarding errors only appear later, once a request arrives. An
operator following the official procedure to the letter gets a green result and
proceeds to the next node, compounding the problem.

The guide also never mentions Raft voter status or autopilot health at all. For
integrated-storage clusters, the checks that actually detect this are:

```bash
bao operator members                  # is the node listed, with a fresh Last Echo?
bao operator raft autopilot state     # Healthy: true, and listed under Voters:?
```

We would suggest adding those to the HA upgrade guide for Raft clusters
regardless of how this specific incompatibility is resolved.

### Two minor reporting issues noticed alongside

While diagnosing, `bao operator raft autopilot state` reported misleading values
for the node it had no follower state for:

- **`Version: 2.5.4`** for a node that is demonstrably running 2.6.2. It appears
  to fall back to the leader's version rather than reporting unknown, which
  actively misleads during a mixed-version upgrade — exactly when the field
  matters most.
- **`Node Type: voter`** on a server whose `Status` in the same block is
  `non-voter`.

Reporting these as unknown/empty would make the "leader has no data for this
server" state legible instead of looking like real telemetry.

### Questions

1. Is mixed-version request forwarding between 2.5.x and 2.6.x supported? The
   [HA upgrade guide](https://openbao.org/docs/upgrading/ha-upgrade/) prescribes
   upgrading all standbys before the active node, which necessarily creates a
   window in which newer standbys must forward to an older active node. It also
   states "do not fail over from a newer version of OpenBao to an older version.
   Our suggested procedure is designed to prevent this" — so standbys-first
   appears to be the intended and only supported direction.
2. If it is not supported, what is the supported upgrade path from 2.5.4 to
   2.6.2 for an HA Raft cluster — is an intermediate hop (2.6.0 or 2.6.1)
   expected to bridge it, or is a full-cluster restart required?
3. Was the `forwarding.RequestForwarding` gRPC service renamed, repackaged, or
   version-gated in the 2.6.0 development cycle? Nothing in the 2.6.x release
   notes mentions a change to the HA cluster RPC surface.
4. The HA upgrade guide closes with "Be sure to also read and follow any
   instructions in the version-specific upgrade notes." Those pages do not appear
   to exist — the `docs/upgrading/` tree contains only `upgrading`, `ha-upgrade`,
   and `plugins`. Is that sentence inherited boilerplate, or is content missing?

satoqz:
## Description

Tested manually by replacing a v2.5.5 node with one built from this patch.

## Rationale

Resolves: #3899

This approach certainly doesn't come with optimal latency, though it should be acceptable for the duration of an upgrade. Alternatively we'd need to mess with config flags or caching so requests can be single-shot most of the time.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/forwarding/request_forwarding_rpc.go
+++ b/internal/vault/forwarding/request_forwarding_rpc.go
@@ -21,6 +21,8 @@ import (
 	"github.com/openbao/openbao/v2/internal/helper/forwarding"
 	"github.com/openbao/openbao/v2/internal/physical/raft"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 type forwardedRequestRPCServer struct {
@@ -535,4 +537,27 @@ func (c *Client) StopInvalidations() {
 	}
 
 	c.invalidationsContextCancel()
+}
+
+// LegacyPackageFallback is a [grpc.UnaryClientInterceptor] that ensures
+// backwards-compatibility between OpenBao <= v2.5 and >= 2.6, which changed
+// the Protobuf package name from "vault" to "forwarding". This is achieved by
+// retrying calls with the original method name if the server does not implement
+// the newer one.
+func LegacyPackageFallback(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	err := invoker(ctx, method, req, reply, cc, opts...)
+	if status.Code(err) != codes.Unimplemented {
+		return err
+	}
+
+	switch method {
+	case "/forwarding.RequestForwarding/Echo":
+		method = "/vault.RequestForwarding/Echo"
+	case "/forwarding.RequestForwarding/ForwardRequest":
+		method = "/vault.RequestForwarding/ForwardRequest"
+	default:
+		return err
+	}
+
+	return invoker(ctx, method, req, reply, cc, opts...)
 }
--- a/changelog/3900.txt
+++ b/changelog/3900.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/ha: Fix broken cross-version request forwarding during rolling upgrades to v2.6.x from an earlier minor version.
+```
--- a/internal/vault/core.go
+++ b/internal/vault/core.go
@@ -3907,7 +3907,9 @@ func (c *Core) refreshRequestForwardingConnection(ctx context.Context, clusterAd
 		grpc.WithDefaultCallOptions(
 			grpc.MaxCallRecvMsgSize(math.MaxInt32),
 			grpc.MaxCallSendMsgSize(math.MaxInt32),
-		))
+		),
+		grpc.WithUnaryInterceptor(forwarding.LegacyPackageFallback),
+	)
 	if err != nil {
 		cancelFunc()
 		c.logger.Error("err setting up forwarding rpc client", "error", err)


internal/vault/forwarding/request_forwarding_rpc.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/openbao/openbao/v2/internal/helper/forwarding"
 	"github.com/openbao/openbao/v2/internal/physical/raft"
 	"google.golang.org/grpc"
+	codes "google.golang.org/grpc/codes"
+	status "google.golang.org/grpc/status"
```
wslabosz-reply:
```suggestion
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
```

satoqz: I mean, yes, but why does gopls keep adding them unless I manually remove 🤔 It doesn't do that in go-kms-wrapping, for example.

satoqz: I think the answer is that `pb.go` files import them that way, and we're sharing a package unlike in go-kms-wrapping where I created separate `pb` packages. So gopls tries to normalize import conventions across the package?

wslabosz-reply (approved):

cipherboy (approved):
Thanks @satoqz!

HannesSommer:
Thanks a lot for the impressively fast fixing 🙏.

PR #3904, opened 2026-08-28
Allow read-enabled standby on leaderless cluster

cipherboy:
## Description

When using GRPC invalidation, we previously always required that the
leader be present and reachable. While we don't loosen that restriction
when a leader does exist (but is not reachable), we do loosen that
restriction when we can prove (within storage) there is no leader and
there may not have been a leader for some time. We want to handle
read-only requests in this window to avoid unnecessarily long outages
and loss of availability. (Though, writes will obviously still be
encumbered).

We ensure that we monitor the lock throughout being leaderless-ly
read-enabled to ensure no other node acquires the lock in the interim.
This (coupled with restarts on announcement) should guarantee that we
never falsely believe that we lack a leader when there is indeed a valid
leader.

This holds as long as `lockRetryInterval` is less than the validity period
of the lock in each underlying HA backend. This holds true of
PostgreSQL (which has `PostgreSQLLockTTLSeconds = 15s` while
`lockRetryInterval = 10s`).

As part of this, `StandbyReadsEnabled(...)` has been refactored (again) to
give a more precise definition for `internal/http`'s usage: we care about
whether we've:

1. Not disabled reads.
2. Are horizontally read-scalable.
3. Actually have started up core i.e., assigned an active context.

This is made possible by the atomic active context refactor.

Some additional log messages have been refactored as part of this to
reduce verbosity of logging in certain backoff-retry scenarios. The
ordering of GRPC invalidation checking w.r.t. `preSeal` has been adjusted
for the same reason.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3904.txt
+++ b/changelog/3904.txt
@@ -1,6 +1,6 @@
 ```release-note:feature
 **PostgreSQL Horizontal Scalability**: Enable read scalability on the PostgreSQL storage backend similar to existing Raft support.
-  - Requires `ha_enabled = true` to be set and an active GRPC connection from standbys to the primary.
+  - Requires `ha_enabled = true` to be set and `cluster_addr` to be reachable (establishing a forwarding RPC connection) from standbys to the primary.
   - In the event of extended leadership loss, standby nodes will come up as read-enabled.
   - Only works with PostgreSQL physical replication; will not work with logical replication.
 ```
--- a/changelog/3904.txt
+++ b/changelog/3904.txt
@@ -1,0 +1,6 @@
+```release-note:feature
+**PostgreSQL Horizontal Scalability**: Enable read scalability on the PostgreSQL storage backend similar to existing Raft support.
+  - Requires `ha_enabled = true` to be set and an active GRPC connection from standbys to the primary.
+  - In the event of extended leadership loss, standby nodes will come up as read-enabled.
+  - Only works with PostgreSQL physical replication; will not work with logical replication.
+```
--- a/internal/vault/ha.go
+++ b/internal/vault/ha.go
@@ -153,6 +153,12 @@ func (c *Core) LeaderLocked() (isLeader bool, leaderAddr, clusterAddr string, er
 		return false, "", "", err
 	}
 	if !held {
+		// Since we don't have a lock, no point keeping around a stale
+		// forwarding connection.
+		if c.rpcForwardingClient.Load() != nil {
+			c.clearForwardingClients()
+		}
+
 		return false, "", "", nil
 	}
 
@@ -656,7 +662,7 @@ func (c *Core) runReadStandby(stopCh <-chan struct{}, sharedStandbyCtx *atomic.P
 		return false, true
 	}
 
-	c.logger.Info("entering read-enabled standby mode")
+	c.logger.Trace("entering read-enabled standby mode")
 	defer c.logger.Trace("done with read-enabled standby loop iteration")
 
 	// Create the standby context (this becomes activeCtx on core, oh well).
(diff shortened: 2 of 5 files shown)


cipherboy:
```

    internal/vault/ha.go
    ❯❱ tools.semgrep.ci.loop-time-after
          ❰❰ Blocking ❱❱
          <-time.After() used in for loop, consider using a ticker or a timer instead
                                                                                     
          1083┆ for {
          1084┆   select {
          1085┆   case <-ctx.Done():
          1086┆           // Do not restart the core because of a cancelled context;
          1087┆           // we're likely already restarting or have experienced a
          1088┆           // state change.
          1089┆           restart = false
          1090┆           return
          1091┆   case <-time.After(lockRetryInterval):
          1092┆   }
             [hid 11 additional lines, adjust with --max-lines-per-finding] 
                            
  BLOCKING CODE RULES FIRED:
    tools.semgrep.ci.loop-time-after
```

Is this outdated from before https://go.dev/wiki/Go123Timer was introduced and we were worried about leakage :thinking: ?

satoqz:
> Is this outdated from before https://go.dev/wiki/Go123Timer was introduced and we were worried about leakage 🤔 ?

IIUC this is just about performance, using a timer/ticker you only allocate the channel once.

internal/vault/ha.go
```diff
@@ -1016,23 +1022,128 @@ func (c *Core) waitForLeadership(manualStepDown *bool, manualStepDownCh, stopCh
...
+		isNoLongerZero := !c.firstMissingLeader.IsZero()
+
+		if !bypass && wasZero && isNoLongerZero {
+			// Reduce the number of errors we spew.
+			c.logger.Error("skipping invalidation streaming as no GRPC client is present; this is required for read-only standby node operation; ensure valid cluster_addr and connectivity")
+		} else if bypass {
+			c.logger.Info("bypassing GRPC leader requirements as no leader lock found")
+		}
+
+		return bypass, bypass
```
wslabosz-reply, line 1151: 🧐

internal/vault/ha.go
```diff
@@ -1016,23 +1022,128 @@ func (c *Core) waitForLeadership(manualStepDown *bool, manualStepDownCh, stopCh
...
+	}()
+
+	return true
+}
+
+// setupGRPCStandbyInvalidations returns (bypass, ok):
+//
+//   - bypass means that GRPC invalidation was bypassed and we still want to
+//     stand up as read-enabled,
+//   - ok means an error fatal occurred
```
wslabosz-reply:
```suggestion
// setupGRPCStandbyInvalidations returns (bypass, isOk):
//
//   - bypass means that GRPC invalidation was bypassed and we still want to
//     stand up as read-enabled,
//   - isOk means no fatal error occurred
```

internal/vault/ha.go
```diff
@@ -1016,23 +1022,128 @@ func (c *Core) waitForLeadership(manualStepDown *bool, manualStepDownCh, stopCh
...
+		c.firstMissingLeader = time.Now()
+		return false
+	}
+
+	if time.Since(c.firstMissingLeader) < 2*lockRetryInterval {
+		// We've not yet found a leader but our time hasn't elapsed.
+		return false
+	}
+
+	// We're good to start as read-only given we haven't found a leader. If we
```
wslabosz-reply:
```suggestion
	// We're good to start as read-only, given we haven't found a leader. If we
```

internal/vault/ha.go
cipherboy: @wslabosz-reply Otherwise I could've complicated the logic in the caller, but it seemed unnecessary. If we bypass, by definition we're ok?

internal/vault/ha.go
cipherboy: I agree with the `no fatal error` occurred change, but not sure `isOk` is necessary?

wslabosz-reply: in my mind the `something, ok := someFunc()` pattern is idiomatic, but by definition the func doc should rather use `result, isOk` wording, but that's just my preference, not really a blocker.

internal/vault/ha.go
wslabosz-reply: I mean, it's fine from code perspective, but when you quickly glance over the thing you might be alarmed why do we return it this way, my suggestion here would be just to add a comment of why.

cipherboy:
@wslabosz-reply Thanks, pushed updates once GitHub gets around to processing them. :laughing:

cipherboy:
(I'm ignoring the semgrep failure; as noted in conversation with @satoqz while there's nominally some marginal benefit in hot loop scenarios, this isn't one and the rule was added before Go 1.23 made timer cleanup automatic so it no longer leaks like it did when this rule was created.)

changelog/3904.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:feature
+**PostgreSQL Horizontal Scalability**: Enable read scalability on the PostgreSQL storage backend similar to existing Raft support.
+  - Requires `ha_enabled = true` to be set and an active GRPC connection from standbys to the primary.
```
satoqz:
```suggestion
  - Requires `ha_enabled = true` to be set and `cluster_addr` to be reachable by other nodes.
```

Perhaps? "an active GRPC connection from standbys to the primary" doesn't tell me much as an operator and sounds like they need to set it up somehow.

internal/vault/ha.go
```diff
@@ -1571,18 +1689,15 @@ func (c *Core) shouldHookInvalidate(phys physical.Backend) bool {
...
 // StandbyReadsEnabled returns true iff standby read are enabled, supported,
 // and likely immediately usable by the physical backend.
-//
-// Notably, when GRPC-based invalidation is in use and the underlying GRPC
-// client is not connected, we return false here.
 func (c *Core) StandbyReadsEnabled() bool {
-	return c.standbyReadsAllowed() &&
+	return c.standbyReadsAllowed() && !c.activeContext.Load().IsNil() &&
 		(isCacheInvalidationBackend(c.underlyingPhysical) ||
-			(isReplicationIndexBackend(c.underlyingPhysical) && c.haveForwardingClient()))
```
satoqz, line 1580: `haveForwardingClient` is dead code then and you can remove it.

website/content/docs/configuration/storage/postgresql.mdx
```diff
@@ -148,10 +148,19 @@ scalability. This requires an active cluster GRPC connection; ensure
 log messages like:
 
 > [DEBUG] setting up GRPC-backed streaming of invalidations
-> [DEBUG] read-only unseal starting
+> [INFO] attempting enabling horizontal scalability (reads)
```
satoqz: This is a TRACE, no?

internal/vault/ha.go
```diff
@@ -656,7 +662,7 @@ func (c *Core) runReadStandby(stopCh <-chan struct{}, sharedStandbyCtx *atomic.P
 		return false, true
 	}
 
-	c.logger.Info("entering read-enabled standby mode")
```
satoqz, line 659: This is one of the more useful log messages IMO, I'd keep it at INFO.

internal/vault/ha.go
```diff
@@ -1119,24 +1225,36 @@ func (c *Core) runReadEnabledStandby(ctx context.Context, ctxCancel context.Canc
...
+
 	// Bail if we're told we should cancel.
 	if ctx.Err() != nil || !runStandby.Load() {
 		c.logger.Debug("context cancelled before calling read-only post-unseal setup")
 		// We want to retry eventually.
 		return false, true
 	}
 
+	// Notify that we're most likely to be successful
+	c.logger.Info("attempting enabling horizontal scalability (reads)")
```
satoqz: We log the same message as TRACE above?

website/content/docs/configuration/storage/postgresql.mdx
cipherboy: I've fixed the wording of this one, there was a duplicate at INFO that I've left but made this about the lock acquisition below it.

internal/vault/ha.go
cipherboy: Fixed the trace one. Its more about lock acquisition.

internal/vault/ha.go
cipherboy: I've intentionally tried to pair down the number of messages we get during backoffs/retries if you're not running in debug/trace.

Keep in mind, prior to #3792:

1. HA mode was singular purpose: either read-enabled or active.
2. You had to finish setup as read-enabled before you even attempted to become active.
3. This is true for every outer restart loop iteration.

Under the new setup of #3792:

1. Both read- and active node statuses have their own inner loops.
2. Only one can occur at a time (as before), but read-enabled setup is no longer required to have completed before active lock acquisition is attempted.
3. Each have their own backoff/retry loop separate from one another and execute in parallel. The outer Reset-everything loop still exists.

So with the addition of backoff/retry leading to read-enabled standby on PostgreSQL (this PR), I wanted to reduce the log spam (from retrying read-enabled standby when it erred due to missing GRPC), which would trigger many of these log messages.

IMHO, this log message in particular is useless; it just says that a loop is executing and doesn't actually lead to doing things; there's plenty of places (for PostgreSQL mostly) where we'd fail and retry without actually having started anything. So we'll get it perhaps dozens of times (on PostgreSQL) in a regular, working setup as the active node takes its time to start up, make its advertisements, and forwarding clients to connect. 

That's why the one INFO message you commented on below remains. That log line actually matter as we're actually beginning `postUnseal` and thus actually setting something useful up. The one I've reworded to be about lock acquisition remains for debugging slow lock acquisition/other failures.

As long as read-standbys are not disabled, we'll always be in read-standby mode unless we're the active leader and we'll always be able/ready to restart standby mode if/when it fails, so I don't think this one is generally useful to most people. :shrug:

changelog/3904.txt
cipherboy: I've pushed something; the "RPC" language is actually used in log lines that `cluster_addr` is not, though I agree that `cluster_addr` is probably the field they'd tune (if not their firewalls lol).

internal/vault/ha.go
satoqz: Right, agreed that it makes sense to defer the log up to right before `postUnseal`.

I suppose I'm unhappy with "attempting enabling horizontal scalability (reads)" being the final remainder. "enabling" is weird wording for `postUnseal` which is roughly equal to loading the entire world. "attempting" communicates uncertainty I'd rather not want to think about when checking if my standby came up right. IMO "entering read-enabled standby mode" was much clearer and didn't leave room for interpretation unnecessarily.

cipherboy: @satoqz Ok, pushed!

satoqz (approved):

cipherboy:
I'll merge this and keep testing to see if we can expand our leadership handover tests on all backends. :thinking:

Issue #2736, opened 2026-03-27
Consistent use of Testify


cipherboy:
For simplicity, nearly every direct call of `if ... { t.Fatalf(...) }` &c should probably moved over `require` if we can. This would simplify the test suite in many areas and ensure new code meets the testing preferences.

satoqz:
We can consider writing (or looking for an existing?) Go modernizer pass to do this.

cipherboy:
I don't see anything existing, but https://github.com/Antonboom/testifylint looks interesting for the future.

omariadfinis:
Hi All, I'm working on this issue to refactor the tests.
@cipherboy, @driif

driif:
Hi @solomonerous 
since you worked on this issue before and @omariadfinis is also interested in working on it, I want to ask: did you already refactored some files to use testify? Do you want to open a PR for that changes?

Fixing this issue will likely require rewriting many files, and reviewing all changes in one PR would be time consuming. If we can merge PRs incrementally, it would be easier for reviewers.

It would also be great if you and @omariadfinis could agree on who works on which files.

satoqz:
@driif we closed the PR by @solomonerous because it was clear AI drive-by.

satoqz:
@driif @omariadfinis generally, I don't really see a point in attacking this issue by sending piece-wise patches to migrate files or packages one-by-one. At this time, there are about 13000 calls to `t.Fatal()` & `t.Fatalf()` in the codebase. https://github.com/openbao/openbao/pull/3957 resolves around 100 of those. Are we really going to go through the effort of creating and reviewing ~100 PRs just to resolve an issue that's mostly cosmetic?

As hinted at in https://github.com/openbao/openbao/issues/2736#issuecomment-4212327605, IMO it would be a better investment of time to give writing an automated tool an attempt, especially since Go already makes this comparatively easy to do. This wouldn't even need to be very sophisticated but could focus on automatically converting some of the simpler patterns (like an `if err != nil { t.Fatal(err) }`), leaving the remainder for manual cleanup.

driif:
@satoqz Most cases are purely cosmetic and don’t need much attention. But many others need a bit more attention and some logical decisions when rewriting. Review this PR — not everything is straightforward. And as you mentioned, there are 13,000 calls.

Of course, it makes sense to include more files in one PR if the changes are mostly cosmetic, but I'm not sure, that reviewing all of changes in one PR would be an easy task

solomonerous:
[...]

PR #3907, opened 2026-08-30
sdk/helper/cidrutil: use testify require consistently in tests

solomonerous:
Part of #2736 - converts `sdk/helper/cidrutil` tests to consistent testify usage.

Changes:

- `if err != nil { t.Fatal(err) }` guards become `require.NoError(t, err)`
- `if err == nil { t.Fatal("expected an error") }` guards become `require.Error(t, err, "expected an error")`
- `if !belongs { t.Fatalf("expected ...") }` boolean blocks become `require.True` / `require.False` with the original messages kept as testify message arguments
- adds the `testify/require` import; no production code changes

Fail-fast semantics are unchanged (`require` calls `t.FailNow`, matching the previous `t.Fatal`), and the `%q`-formatted messages survive as testify format strings with their original arguments.

Validation: `go test ./helper/cidrutil/...` passes; `gofmt` and `go vet` clean. Happy to continue the same conversion for other `sdk/helper` packages in follow-ups if this shape is what #2736 wants.

### DIFF
--- a/sdk/helper/cidrutil/cidr_test.go
+++ b/sdk/helper/cidrutil/cidr_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	sockaddr "github.com/hashicorp/go-sockaddr"
+	"github.com/stretchr/testify/require"
 )
 
 func TestCIDRUtil_IPBelongsToCIDR(t *testing.T) {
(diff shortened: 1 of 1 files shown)


solomonerous:
Quick question so I don't guess at the reason for the close: is the Testify migration from #2736 being handled internally, or is there a preferred scope/format for outside contributions on it?

The conversion itself passes `go test ./helper/cidrutil/...` with fail-fast semantics unchanged. If piecemeal test conversions aren't wanted right now that's completely fine - just let me know and I'll drop this track.

wslabosz-reply:
@solomonerous 
[CONTRIBUTING.md](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#pull-requests)

solomonerous:
> @solomonerous 
> 
> [CONTRIBUTING.md](https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#pull-requests)

oh..thanks mate 🙏🏻

PR #3913, opened 2026-09-01
Limit parallel transactions in PostgreSQL

cipherboy:
## Description

Most of the time, transaction acquisition is limited by the underlying capacity of the PostgreSQL server; we recommend operators tune max_parallel to a reasonable level (5ish from earlier testing at a prior employer). However, the transaction parallelism is limited to 64 by default, which exceeds this substantially, causing a lock renewal failure when long-lived transactions occur (or, if they simply win connection acquisition).

At the same time, let's take the opportunity to add the same global transaction leak detection logic to PostgreSQL for us. This uses a reasonable (2kb) stack trace as we have no other information to go on.

## Rationale

fyi: @fcat

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/physical/postgresql/transaction.go
+++ b/internal/physical/postgresql/transaction.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"runtime"
 	"sync"
 
 	"github.com/openbao/openbao/sdk/v2/physical"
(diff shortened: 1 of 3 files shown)


internal/physical/postgresql/transaction.go
```diff
@@ -36,11 +39,34 @@ func (b *PostgreSQLBackend) newTransaction(ctx context.Context, readOnly bool) (
...
+			"transaction was leaked",
+			// Stack is the only information we retain in this existing implementation.
+			"stack", string(trace),
+			"read_only", readOnly,
+			"leaked_transactions", b.txnLeakCount.Load(),
+		)
+
+		b.txnPermitPool.Release()
+		_ = tx.Rollback() // ignore the error, not much we can do about this
+	}, true)
```
cipherboy, line 67: `runtime.Cleanup` panics if the arg is `nil` 😆

wslabosz-reply (approved):

cipherboy:
Thanks @wslabosz-reply!

PR #3915, opened 2026-09-01
Allow pinning OCI-based plugins by manifest digest

satoqz:
## Description

Works as demonstrated in the RFC.

## Rationale

See https://github.com/openbao/openbao/pull/3881.

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/command/server/config.go
+++ b/internal/command/server/config.go
@@ -307,21 +307,17 @@ func (b *ServiceRegistration) GoString() string {
 // PluginConfig represents the configuration for a single OCI-based plugin
 type PluginConfig struct {
 	UnusedKeys configutil.UnusedKeyMap `hcl:",unusedKeyPositions"`
-	RawConfig  map[string]any
 
 	Type       string
 	Name       string
-	Image      string   `hcl:"image"`
-	Version    string   `hcl:"version"`
-	BinaryName string   `hcl:"binary_name"`
-	SHA256Sum  string   `hcl:"sha256sum"`
-	Command    string   `hcl:"command"`
-	Args       []string `hcl:"args"`
-	Env        []string `hcl:"env"`
-}
-
-func (p *PluginConfig) URL() string {
-	return fmt.Sprintf("%s:%s", p.Image, p.Version)
+	Image      name.Reference `hcl:"-"`
+	RawImage   string         `hcl:"image"`
+	Version    string         `hcl:"version"`
+	BinaryName string         `hcl:"binary_name"`
+	SHA256Sum  string         `hcl:"sha256sum"`
+	Command    string         `hcl:"command"`
+	Args       []string       `hcl:"args"`
+	Env        []string       `hcl:"env"`
 }
 
 func (p *PluginConfig) Slug() string {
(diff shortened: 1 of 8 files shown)


internal/command/server/config.go
```diff
@@ -307,21 +307,17 @@ func (b *ServiceRegistration) GoString() string {
...
-	SHA256Sum  string   `hcl:"sha256sum"`
-	Command    string   `hcl:"command"`
-	Args       []string `hcl:"args"`
-	Env        []string `hcl:"env"`
-}
-
-func (p *PluginConfig) URL() string {
-	return fmt.Sprintf("%s:%s", p.Image, p.Version)
+	Type       string         `hcl:"-"`
+	Name       string         `hcl:"-"`
```
cipherboy: Isn't this a breaking change? 🤔

internal/command/server/config.go
```diff
@@ -432,37 +424,64 @@ func (p *PluginConfig) Validate(sourceFilePath string) []configutil.ConfigError
...
 
-		var m map[string]any
-		if err := hcl.DecodeObject(&m, item.Val); err != nil {
-			return result, fmt.Errorf("%v.%d: %w", name, index, err)
-		}
-		i.RawConfig = m
-
 		switch {
-		case i.Type != "":
+		case p.Type != "":
```
cipherboy, line 436: @satoqz Re: above -- in particular, `Type` in the past could be inferred from the HCL block as we discussed at one point (the `plugin { type = ""  name = "" }` explicit form) that I seem to recall you had mixed feelings about 😆 

If we really want to disallow this, I think we should remove these blocks too.

internal/helper/pluginutil/oci/downloader.go
```diff
@@ -123,66 +121,66 @@ func (d *PluginDownloader) maxPluginSize() int64 {
...
-	if err != nil {
-		d.logger.Debug("failed to calculate plugin hash", "plugin", config.Slug(), "error", err)
-		return false
+	// Validate the SHA-256 digest of the plugin binary if one is specified.
+	if len(config.SHA256Sum) > 0 {
+		actualHash, err := osutil.FileSha256Sum(pluginPath)
+		if err != nil {
+			return false
+		}
+		valid = strings.EqualFold(actualHash, config.SHA256Sum)
```
cipherboy: Since you can still specify SHA256 values with image digests, doesn't this mean that if the path is incorrect but the digest is correct, it'll pass this check? 🤔  It seems weird to allow a named config that has a digest (and is thus _new_), but which doesn't follow the correct path layout, but which has a valid sha256sum digest? 

I almost wonder if the `if !valid { return false }` pattern after the first one to ensure that users don't do anything funny?

cipherboy (dismissed):
The new configuration is much better, thanks @satoqz!

Mostly nits, I'm fine to merge if you prefer to address in a follow-up.

internal/command/server/config.go
satoqz: Ah I think my hands just filled those annotations in automatically, for consistency. Will remove.

internal/helper/pluginutil/oci/downloader.go
satoqz: Yeah, that's probably better.

cipherboy (approved):

PR #3918, opened 2026-09-01
test: Await token revocation

Shoowa:
## Description

1. Add _require.EventuallyWithT_ for token revocation.
2. Add constants for the time the test awaits a condition, and the time the test evaluates a condition.
3. Replace a _defer_ with a _t.Cleanup_

```bash
go test github.com/openbao/openbao/v2/internal/vault/external_tests/standby -run '^TestReadOnlyStandby$'
```

## Rationale
_TestReadOnlyStandby_ suffered 12 failures from July 3rd to August 31st this year.

Resolves: contributes towards #2581 

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/external_tests/standby/read_only_standby_test.go
+++ b/internal/vault/external_tests/standby/read_only_standby_test.go
@@ -24,6 +24,11 @@ import (
 )
 
 func TestReadOnlyStandby(t *testing.T) {
+	const (
+		timeout = 10 * time.Second
+		tick    = 100 * time.Millisecond
+	)
+
 	conf := vault.CoreConfig{
 		LogicalBackends: map[string]logical.Factory{
 			"pki": logicalPki.Factory,
@@ -38,7 +43,7 @@ func TestReadOnlyStandby(t *testing.T) {
 	cluster := vault.NewTestCluster(t, &conf, &opts)
 
 	cluster.Start()
-	defer cluster.Cleanup()
+	t.Cleanup(cluster.Cleanup)
 
 	testhelpers.WaitForActiveNodeAndStandbys(t, cluster)
 	require.False(t, cluster.Cores[0].Standby())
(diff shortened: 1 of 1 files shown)


internal/vault/external_tests/standby/read_only_standby_test.go
```diff
@@ -24,6 +24,11 @@ import (
 )
 
 func TestReadOnlyStandby(t *testing.T) {
+	const (
+		LIFESPAN = 10 * time.Second
+		TICK     = 100 * time.Millisecond
```
phil9909: use `lifespan` and `tick`. In Go there is no convention to distinguish between variables and constants using SCREAMING_SNAKE_CASE (as many other languages do)

Maybe `timeout` would be a better name than `lifespan`?

internal/vault/external_tests/standby/read_only_standby_test.go
```diff
@@ -63,27 +68,29 @@ func TestReadOnlyStandby(t *testing.T) {
...
 	}
 
 	t.Log("revoking token")
-	require.NoError(t, primaryClient.Auth().Token().RevokeTreeWithContext(t.Context(), token.Auth.ClientToken))
-	_, err = standbyClient.KVv2("kv").Get(t.Context(), "foo")
-	require.ErrorContains(t, err, "permission denied", "token was revoked on the primary, should be declined by secondaries")
+	require.EventuallyWithT(t, func(collect *assert.CollectT) {
+		require.NoError(t, primaryClient.Auth().Token().RevokeTreeWithContext(t.Context(), token.Auth.ClientToken))
+		_, err = standbyClient.KVv2("kv").Get(t.Context(), "foo")
+		require.ErrorContains(t, err, "permission denied", "token was revoked on the primary, should be declined by secondaries")
```
phil9909:
```suggestion
		require.NoError(collect, primaryClient.Auth().Token().RevokeTreeWithContext(t.Context(), token.Auth.ClientToken))
		_, err = standbyClient.KVv2("kv").Get(t.Context(), "foo")
		require.ErrorContains(collect, err, "permission denied", "token was revoked on the primary, should be declined by secondaries")
```

You need to pass the `collect` instead of `t` to the require functions, otherwise the test will fail on the first failure anyhow.

phil9909 (changes requested):

internal/vault/external_tests/standby/read_only_standby_test.go
Shoowa: That's embarrassing, because I made a similar change recently. Thanks for the catch.

phil9909 (dismissed):

internal/vault/external_tests/standby/read_only_standby_test.go
```diff
@@ -63,27 +68,29 @@ func TestReadOnlyStandby(t *testing.T) {
...
-		}, 10*time.Second, 100*time.Millisecond)
+		}, timeout, tick)
 	}
 
 	t.Log("revoking token")
-	require.NoError(t, primaryClient.Auth().Token().RevokeTreeWithContext(t.Context(), token.Auth.ClientToken))
-	_, err = standbyClient.KVv2("kv").Get(t.Context(), "foo")
-	require.ErrorContains(t, err, "permission denied", "token was revoked on the primary, should be declined by secondaries")
+	require.EventuallyWithT(t, func(collect *assert.CollectT) {
+		require.NoError(collect, primaryClient.Auth().Token().RevokeTreeWithContext(t.Context(), token.Auth.ClientToken))
```
cipherboy: I do wonder if the revoke should only occur once, outside of the eventually. 🤔

Shoowa: Thanks for the tip. No need to include a deletion operation inside the _for_ loop of _EventuallyWithT_. I just removed the revocation from  _EventuallyWithT_. The standby client's _Get_ request remains inside _EventuallyWithT_, so it will be repeated like the previous evaluations in the same test. Re-reading until the error is correct makes sense.

cipherboy (approved):

PR #3919, opened 2026-09-01
Ensure control_groups stanza is well-formed

### LINKED ISSUE #3666: Control Group factor with omitted or zero approvals fails open
**Describe the bug**

A Control Group factor that omits `approvals` or sets `approvals = 0` is accepted by policy parsing, but authorization then succeeds with **zero** approvals. Validation treats `approvalCount >= requiredApprovals` as satisfied when both sides are `0`, so the protected request can complete without any `sys/control-group/authorize` step.

This looks like a functional correctness / fail-open bug in the unreleased Control Groups work on `main`, not an advisory-style report.

**To Reproduce**

1. On a recent `main` build with Control Groups enabled, write a policy like:

```hcl
path "secret/data/foo" {
  capabilities = ["create", "update", "read"]
  control_group = {
    ttl = "5m"
    factor "security-approval" {
      controlled_capabilities = ["update"]
      identity = {
        group_names = ["security-approvers"]
        # approvals omitted, or explicitly: approvals = 0
      }
    }
  }
}
```

2. Assign that policy to a requester entity; create an approver group `security-approvers` with authorize rights on `sys/control-group/*` (approver is unused for the failure mode).
3. As the requester, perform the controlled operation (e.g. write `secret/data/foo`). Observe the usual wrapped / control-group deferred response.
4. Unwrap the wrapping token **without** calling `sys/control-group/authorize`.

**Expected behavior**

Policy parse should reject `approvals` omitted / `approvals <= 0`, **or** validation should fail closed until at least one (positive) approval is recorded. Omitting `approvals` should not authorize the request.

**Actual behavior**

Unwrap / deferred validation succeeds with no authorizations: `approvalCount` is `0` and `factor.Identity.Approvals` is `0`, so `0 >= 0` passes.

**Environment:**
* OpenBao Server Version: current `main` (Control Groups via #2241 / follow-ups such as #3609; not treating this as shipped-release behavior)
* OpenBao CLI Version: n/a (API / test-cluster repro)
* Server Operating System/Architecture: linux/amd64 (dev)

**Additional context**

Relevant check is in control-group validation: a factor passes when `approvalCount >= factor.Identity.Approvals`. With the default Go zero value for an omitted `approvals` field (or an explicit `0`), that comparison is true immediately.

Suggested fix direction: require `approvals >= 1` at policy parse time, and/or treat `Approvals <= 0` as invalid / non-passing in `validateControlGroup`.

Happy to turn this into a small failing regression test if useful.

cipherboy:
## Rationale

Resolves: #3666

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/policy/policy.go
+++ b/internal/vault/policy/policy.go
@@ -377,289 +377,318 @@ func parsePaths(result *Policy, list *ast.ObjectList, performTemplating bool, bl
 			key = item.Keys[0].Token.Value().(string)
 		}
 
-		// Check the path
-		if performTemplating {
-			_, templated, err := identitytpl.PopulateString(identitytpl.PopulateStringInput{
-				Mode:                 identitytpl.ACLTemplating,
-				String:               key,
-				Entity:               identity.ToSDKEntity(entity),
-				Groups:               identity.ToSDKGroups(groups),
-				NamespaceID:          result.Namespace.ID,
-				BlockedSubstitutions: blockedSubstitutions,
-			})
-			if err != nil {
-				continue
-			}
-			key = templated
-		} else {
-			hasTemplating, _, err := identitytpl.PopulateString(identitytpl.PopulateStringInput{
-				Mode:              identitytpl.ACLTemplating,
-				ValidityCheckOnly: true,
-				String:            key,
-			})
-			if err != nil {
-				return fmt.Errorf("failed to validate policy templating: %w", err)
-			}
-			if hasTemplating {
-				result.Templated = true
-			}
+		path, err := parsePath(result, key, item, performTemplating, blockedSubstitutions, entity, groups)
+		if err != nil {
+			return fmt.Errorf("path %q: %w", key, err)
 		}
 
-		valid := []string{
-			"comment",
-			"policy",
-			"capabilities",
-			"allowed_parameters",
-			"denied_parameters",
-			"required_parameters",
-			"min_wrapping_ttl",
-			"max_wrapping_ttl",
-			"mfa_methods",
-			"pagination_limit",
-			"expiration",
-			"list_scan_response_keys_filter_path",
-			"control_group",
+		if path != nil {
+			paths = append(paths, path)
 		}
-		if err := hclutil.CheckHCLKeys(item.Val, valid); err != nil {
-			return multierror.Prefix(err, fmt.Sprintf("path %q:", key))
-		}
-
-		var pc PathRules
-
-		// allocate memory so that DecodeObject can initialize the ACLPermissions struct
-		pc.Permissions = new(ACLPermissions)
-
-		// decode ControlGroup if needed
-		if err := hclutil.WhenHCLKeyPresent(item.Val, "control_group", func(item *ast.ObjectItem) error {
-			if err := hclutil.CheckHCLKeys(item.Val, []string{"ttl", "self_auth_allowed", "factor"}); err != nil {
-				return multierror.Prefix(err, "path control_group:")
-			}
-			cg := new(ControlGroup)
-			if err := hcl.DecodeObject(&cg, item.Val); err != nil {
-				return multierror.Prefix(err, "path control_group:")
-			}
-			// decode factors in control_group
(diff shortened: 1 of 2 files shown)


cipherboy:
fyi @suprjinx

internal/vault/policy/policy.go
```diff
@@ -651,7 +667,7 @@ func parsePaths(result *Policy, list *ast.ObjectList, performTemplating bool, bl
 			}
 
 			if checkPathOne == checkPathTwo && keyOne != keyTwo {
-				return fmt.Errorf("list_scan_response_keys_filter_path resulted in same path for two different keys")
+				return errors.New("list_scan_response_keys_filter_path resulted in same path for two different keys")
```
wslabosz-reply: fun fact: since go 1.26 it does the same thing as `fmt.Errorf()` https://go.dev/doc/go1.26#fmtpkgfmt

wslabosz-reply (dismissed):

phil9909 (dismissed):

internal/vault/policy/policy.go
satoqz: `errors.New` is still minimally faster, I don't take the Go team fixing this as an endorsement to use `fmt.Errorf` _everywhere_ but as a kind performance optimization for those who had chosen to anyway beforehand.

internal/vault/policy/policy.go
```diff
@@ -490,6 +494,18 @@ func parsePaths(result *Policy, list *ast.ObjectList, performTemplating bool, bl
...
+					return fmt.Errorf("path control_group factor %d: missing controlled capabilities", i)
+				}
+
+				if factor.Identity.Approvals <= 0 {
+					return fmt.Errorf("path control_group factor %d: missing required approvals", i)
+				}
+
+				if len(factor.Identity.GroupNames) == 0 {
+					return fmt.Errorf("path control_group factor %d: missing approving groups", i)
+				}
```
satoqz:
```suggestion
				if len(factor.ControlledCapabilities) == 0 {
					return fmt.Errorf("path control_group factor %d: missing controlled capabilities", i+1)
				}

				if factor.Identity.Approvals <= 0 {
					return fmt.Errorf("path control_group factor %d: missing required approvals", i+1)
				}

				if len(factor.Identity.GroupNames) == 0 {
					return fmt.Errorf("path control_group factor %d: missing approving groups", i+1)
				}
```

More user-friendly to start counting at one here, IMO.

internal/vault/policy/policy.go
```diff
@@ -472,6 +472,10 @@ func parsePaths(result *Policy, list *ast.ObjectList, performTemplating bool, bl
 			cg.TTL = ttl
 			cg.TTLHCL = nil
 
+			if len(cg.Factors) == 0 {
+				return errors.New("requires at least one factor but had none")
```
satoqz:
```suggestion
				return errors.New("control_group requires at least one factor but had none")
```

internal/vault/policy/policy.go
satoqz: Don't we produce rather sub-optimal errors here, generally?

```
error handling control_group: path control_group factor: invalid name key
```

I'd like if we could have something closer to:

```
path "sys/something": control_group factor 3: invalid name key
```

internal/vault/policy/policy.go
cipherboy: Sadly everywhere but stepwise uses zero-based indexing :laughing:

suprjinx (dismissed):
LGTM

internal/vault/policy/policy.go
cipherboy: @satoqz This should be unnecessary; this is in a helper function which prepends `error handling control_group:` so this is a duplicated element.

internal/vault/policy/policy.go
cipherboy: @satoqz I've fixed the path error construction lmkwyt.

satoqz (approved):

cipherboy:
Thanks! Merging...

PR #2383, opened 2026-02-02
Sys `/health`, `/leader` and `/ha-status` responses cleanup

wslabosz-reply:
## Description
- removed `PerfStandby`, `PerfStandbyLastRemoteWAL` and `LastWAL` from `SealStatusOutput`
- adjusted `/leader` and`/seal-status` openapi schema in `vault/logical_system_paths.go` 
- `ActiveTime`, `IsSelf`, `LeaderAddress` and `LeaderClusterAddress` properties will only appear when HA is enabled
- adjusted related documentation

## Rationale
Removed attributes were unused and contribute to the bloat in the documentation and code.

Relatively related to [per-ns sealing feature](https://github.com/openbao/openbao/tree/namespaces-seal), seal status command will have to be adjusted for ns attribution (separate call, will reuse the schema).

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3932, opened 2026-09-03
fix: `sys/leader` does not contain `is_self` on standbys

phil9909:
## Description

`omitempty` causes the json marshaller to drop the value if it's "empty" which is `false` for boolean values. Changing it to a bool pointer fixes this (`nil` -> drop, pointer to `false` or `true` -> keep)

## Rationale

This is a regression introduced in https://github.com/openbao/openbao/pull/2383

## Acknowledgements

       also allowed.

### DIFF
--- a/internal/vault/logical_system.go
+++ b/internal/vault/logical_system.go
@@ -4523,7 +4523,7 @@ func (c *Core) GetSealStatus(ctx context.Context, lock bool) (*CoreSealStatusRes
 
 type LeaderResponse struct {
 	HAEnabled            bool      `json:"ha_enabled"`
-	IsSelf               bool      `json:"is_self,omitempty"`
+	IsSelf               *bool     `json:"is_self,omitempty"`
 	ActiveTime           time.Time `json:"active_time,omitzero"`
 	LeaderAddress        string    `json:"leader_address,omitempty"`
 	LeaderClusterAddress string    `json:"leader_cluster_address,omitempty"`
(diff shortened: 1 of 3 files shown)


internal/vault/logical_system.go
```diff
@@ -4523,7 +4523,7 @@ func (c *Core) GetSealStatus(ctx context.Context, lock bool) (*CoreSealStatusRes
 
 type LeaderResponse struct {
 	HAEnabled            bool      `json:"ha_enabled"`
-	IsSelf               bool      `json:"is_self,omitempty"`
+	IsSelf               *bool     `json:"is_self,omitempty"`
```
cipherboy, line 4526: Isn't the simpler fix just removing `omitempty`? (If we want this field)

phil9909: The idea of https://github.com/openbao/openbao/pull/2383 was to drop all fields from the response that don't really have a meaning when running non-HA. Hence the addition of `omitempty` / `omitzero` to some of the fields.

This works well for e.g. the timestamp as the zero value should never occur in practice. But for booleans not so much.

cipherboy: `api.LeaderResponse` would need the same change, but would be breaking from a type system PoV in that case for anyone using the Go client. 🤔

satoqz: I don't think `api.LeaderResponse` would need the same change, you'll just see `IsSelf = false` just like you'll see `ActiveTime = time.Time{}` if `HAEnabled = false`, and that's probably more convenient for developers than dealing with bool pointers.

cipherboy: Ok. I do wonder if `is_self=true` makes sense on non-HA nodes (instead of `IsSelf=false`) but agree that we should definitely provision `is_self=false` it on standby nodes either way.

cipherboy (approved):

PR #3710, opened 2026-08-06
Add more info about missing plugins to migration guide

trutzonline:
## Description

Follow-up to #3707, as suggested there.
cc: @satoqz 

## Rationale

Two changes to the in-place migration guide.

Resolves: #

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI
       (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or
       filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the
       [DCO ownership](https://developercertificate.org/) statement
       and this change did not use post-BUSL-licensed code from HashiCorp.
       Existing MPL-licensed code is still allowed, subject to attribution.
       Code authored by yourself and submitted to HashiCorp for inclusion is
       also allowed.

PR #3936, opened 2026-09-04
Clear read-only error on plugin reload

satoqz:
## Description

When unsealing, a mount's storage view may remain marked read-only indefinitely beyond the setup phase if the backend did not come up successfully ("nil backend"). This happens when the mount's backing plugin is no longer available, e.g., because it was a builtin that's been removed or the plugin was deregistered.

This change ensures that a mount's storage view is no longer marked read-only when the mount sees a successful plugin reload, unblocking live migrations from builtin->external plugins.

Calling `SetReadOnlyErr(nil)` felt a little awkward since it is never otherwise used this way. Instead, mount setup code always passes the result of a previous `GetReadOnlyErr()` to return back to some theoretical prior error state. I don't however see any scenario where mount setup would restore to a non-nil error, so I've simplified that code by passing `nil` directly, hopefully making it easier to read.

## Rationale

See https://github.com/openbao/openbao/pull/3486#issuecomment-5445870167, https://github.com/openbao/openbao/pull/3710#discussion_r3737017357 for context.

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/3936.txt
+++ b/changelog/3936.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/plugins: Ensure that a mount's storage view is no longer marked read-only following a successful plugin reload.
+```
--- a/internal/vault/audit.go
+++ b/internal/vault/audit.go
@@ -120,13 +120,11 @@ func (c *Core) enableAudit(ctx context.Context, entry *routing.MountEntry, updat
 		return err
 	}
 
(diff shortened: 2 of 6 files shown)


cipherboy (approved):

cipherboy:
Thanks @satoqz!

PR #3944, opened 2026-09-07
Fix recovery token generation via CLI (#3924)

satoqz:
Backport of https://github.com/openbao/openbao/pull/3924 (with conflicts)


wslabosz-reply (approved):

cipherboy (approved):

cipherboy:
Will be happy when v2.7 is the only supported release, the `/internal/` move was harder for backports than I expected :laughing:

PR #3950, opened 2026-09-07
Drop github.com/coreos/go-systemd

satoqz:
## Description

Replaces github.com/coreos/go-systemd with a tiny helper package that reimplements `sd_notify` functionality.

## Rationale

Systemd's `NOTIFY_SOCKET` is a very simple protocol that we don't need a larger dependency to implement for us.

We've also ran into troubles with this library previously, see https://github.com/openbao/openbao/pull/2588. This still prevents us from upgrading, though there hasn't been a reason to.

## Acknowledgements

       also allowed.


wslabosz-reply (approved):
👍

JanMa (approved):
Lgtm

PR #3959, opened 2026-09-08
Retroactively add changelog for #2383

### LINKED ISSUE #3937: sys/health stopped returning performance_standby in 2.6.0 with no changelog entry
### Summary

`GET /sys/health` stopped returning the `performance_standby` field in 2.6.0. The removal looks deliberate and the documentation is consistent with it, but it is not mentioned in the 2.6.0 changelog or release notes, so API consumers upgrading from 2.5.x get no warning that a response field disappeared.

### Evidence

Running the dev server on each image and calling `/v1/sys/health`:

```
2.5.4  {"initialized":true,"sealed":false,"standby":false,"performance_standby":false,"replication_performance_mode":"disabled",...}
2.6.0  {"initialized":true,"sealed":false,"standby":false,"replication_performance_mode":"disabled",...}
2.6.1  field absent
2.6.2  field absent
```

The string is also gone from the shipped binary. `grep -a -c performance_standby /bin/bao` returns 8 on 2.5.4 and 0 on 2.6.2.

`HealthResponse` in `internal/http/sys_health.go` no longer declares the field, while `api/sys_health.go` keeps it with the comment `// not present in OpenBao, but left here for compatibility.`, which reads as intentional.

### Why this is worth a changelog entry

The 2.6.0 notes do document other breaking API changes, for example removing `StoredShares` from `InitRequest` and renaming `build_date` to `commit_date` on `sys/seal-status`. A removed field on `sys/health` is the same class of change, and `sys/health` is polled by monitoring and by client libraries far more often than either of those.

Concretely, Spring Vault binds `performance_standby` to a primitive `boolean` in `VaultSysTemplate.VaultHealthImpl`. Combined with Jackson 3 enabling `FAIL_ON_NULL_FOR_PRIMITIVES` by default, an absent field now throws instead of defaulting to `false`, so the Spring health indicator fails against 2.6.x. Any deployment whose liveness probe points at the aggregated health endpoint restarts in a loop after the upgrade. I have filed that half separately against Spring Vault, since their struct should tolerate the field being absent.

I searched the changelog for `performance_standby` and `performance standby` and found no matches in any 2.6.x section.

### Request

Add an entry to the 2.6.0 changelog and release notes recording that `performance_standby` was removed from the `sys/health` response, ideally with a line in the upgrade notes pointing at the standby node RFC for the reasoning. Nothing needs to change in the code or the API docs as far as I can tell.

wslabosz-reply:
## Description
adds retroactively a changelog for changes introduced with #2383 

## Rationale
gives more visibility into the API changes (removal of fields)

Resolves: #3937

## Acknowledgements

       also allowed.

### DIFF
--- a/changelog/2383.txt
+++ b/changelog/2383.txt
@@ -1,0 +1,6 @@
+```release-note:change
+command/seal-status: Remove "Performance Standby Node", "Performance Standby Last Remote WAL", "Last WAL" and "Warnings" from the output.
+```
+```release-note:change
+sys/health: Remove `performance_standby` and `last_wal` fields from the response.
+```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,11 +121,13 @@ CHANGES:
 
 * command: Remove buffering and delayed release of logs during startup phase of `server`, `agent`, `proxy` & `debug` subcommands. This includes the removal of the undocumented and hidden `-disable-gated-logs` flag. [[GH-2620](https://github.com/openbao/openbao/pull/2620)]
 * command: `operator generate-root` now uses the authenticated `/sys/generate-root-token` endpoints instead of the deprecated `/sys/generate-root` endpoints. [[GH-3190](https://github.com/openbao/openbao/pull/3190)]
+* command/seal-status: Remove "Performance Standby Node", "Performance Standby Last Remote WAL", "Last WAL" and "Warnings" from the output. [[GH-2383](https://github.com/openbao/openbao/pull/2383)]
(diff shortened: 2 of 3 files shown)


eyenx (dismissed):

changelog/2383.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:change
+command/seal-status: Removed 'Performance Standby Node', 'Performance Standby Last Remote WAL', 'Last WAL' and 'Warnings' from the output.
+```
+```release-note:change
+sys/health: Removed 'performance_standby' and 'last_wal' fields from the response.
```
satoqz:
```suggestion
sys/health: Remove `performance_standby` and `last_wal` fields from the response.
```

changelog/2383.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:change
+command/seal-status: Removed 'Performance Standby Node', 'Performance Standby Last Remote WAL', 'Last WAL' and 'Warnings' from the output.
```
satoqz:
```suggestion
command/seal-status: Remove "Performance Standby Node", "Performance Standby Last Remote WAL", "Last WAL" and "Warnings" from the output.
```

satoqz (approved):

cipherboy (approved):

PR #3966, opened 2026-09-09
Compare all shamir key shares during presence checks

### LINKED ISSUE #2663: Shamir key recreation shouldn't leak keyshare provision ordering

haxonittech:
## Description

Adds `shamir.Shares.Has` and uses it in all seven duplicate-share checks. The method compares every stored share before returning, so a match does not skip later comparisons.

The method and its table test live in the existing Shamir files. The test covers matches at different positions, absent shares, empty inputs and different lengths.

Validation: the Shamir and full vault package tests, related HTTP tests, formatting, golangci-lint and repository Semgrep checks pass.

## Rationale

Share comparison belongs with the Shamir helpers. Core's share-progress fields now use `shamir.Shares`, which can still be passed to `shamir.Combine`.

The existing progress-list mismatch in `VerifyRotation` remains a separate issue.

Resolves: #2663

## Acknowledgements

       also allowed.


satoqz:
Adding 4 files to the largest package in the codebase to implement a minor optimization is certainly a choice, did a human come up with that?

haxonittech:
yeah i wrote it myself. the same check was in 7 places so i figured a helper made sense. 3 of those 4 files are tests.

it was meant to fix the ordering issue in #2663, not improve performance. but fair point about the extra files, I am happy to move the helper and tests into existing ones.

wslabosz-reply:
Hello @haxonittech 
definitely `internal/vault/keyshare.go` as a "helper" is better placed inside `helper` package.
New test files should be incorporated inside existing ones `rekey_test.go`, `rotate_test.go`.

satoqz:
I think we can place this under `sdk/helper/shamir` as a newtype, tbh:

```go
type Shares [][]byte

func (s Shares) Has(share []byte) bool {
  // ...
}
```

Then you'll have `shamir.Shares` to pass around in core, which can still be passed to `shamir.Combine` &c.

satoqz:
The amount of tests in this is rather superfluous, I'd keep one test for the actual helper but not re-test by proxying over various other subsystems that ultimately use the helper.

haxonittech:
The JWT test timed out waiting for the token file. I reran the failing case and all eight JWT variants locally with the `deadlock` tag, and both runs passed. Could someone rerun the failed jobs?

changelog/3966.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+core: Compare every stored key share when checking for duplicates during unseal, root generation, rekey, and rotation.
```
wslabosz-reply, line 2:
```suggestion
core: Stop leaking ordering of key shares during unseal, root generation, rekey, and rotation.
```

satoqz: That's a bit... strongly worded, not covering how it is leaked.

How about:


```suggestion
core: Prevent timing side-channel leakage that reveals the ordering of already-provided key shares when checking for duplicates during unseal, root generation, rekey, and rotation.
```

wslabosz-reply: @satoqz yeah, I've focused too much on the length of the entry, your suggestion makes more sense.

satoqz (approved):
Thanks!

PR #3987, opened 2026-09-11
Wait for raft retry_join before self-initializing

### LINKED ISSUE #3652: The self-init (`initialize` stanza) bootstraps a new Raft cluster instead of honoring retry_join
**Describe the bug**

After adding `initialize` block to the configuration, OpenBao instance bootstraps as a new Raft cluster instead of honoring exist quorum.

This appears to be related to #2274, but the scenario is slightly different. (Also the original issue is already very crowded so I decide to file a new issue).

Our environment consists of three OpenBao instances running in an AWS Auto Scaling Group, using integrated Raft storage with `retry_join` based on AWS tags. We also use immutable infrastructure, so replacement instances always start with an empty data directory.

Without the `init` block, everything works as expected. A replacement instance discovers the existing Raft leader via `retry_join`, joins the cluster, and replicates the existing data.

After enabling the `init` block, however, a replacement instance immediately bootstraps its own single-node cluster instead of joining the existing Raft cluster.

We ruled out connectivity or quorum issues:

1. We started a healthy two-node Raft cluster without the `init` block, then launched a third instance with `init` enabled. Despite a healthy existing cluster, the third instance ignored it and initialized itself.

    ```console
    $ vault operator raft autopilot state

    Healthy:                         true
    Failure Tolerance:               0

    Leader:                          node-1

    Voters:
       node-1
       node-2
    ```

2. Removing the `init` block from the replacement instance immediately restores the expected behavior. The instance successfully joins the existing cluster, indicating that networking and `retry_join` are functioning correctly.

## Expected behavior

When `retry_join` is configured and an existing Raft cluster is reachable, we would expect the node to join that cluster instead of immediately creating a new one.

The commonly suggested workaround of enabling the `init` block on only one server is not practical for modern immutable deployments such as Kubernetes or AWS Auto Scaling Groups, where all instances are identical and there is no concept of a permanently designated bootstrap node.

## Why this matters

This behavior makes self-init difficult to use safely in immutable environments.

The failure mode is also concerning: instead of joining and replicating the existing cluster, a replacement node silently creates a completely independent cluster. If an operator or automation accidentally interacts with this new cluster, believing it to be the production cluster, it could lead to operational mistakes or data loss.

Even if this is the intended behavior today, it would be helpful to support a mode where self-init first honors an existing reachable Raft cluster before deciding to bootstrap a new one.

**Environment:**
* OpenBao CLI Version (retrieve with `bao version`): 2.5.5

Extremely normal configuration

```hcl
listener "tcp" {
  address         = "0.0.0.0:8200"
  cluster_address = "0.0.0.0:8201"
  tls_cert_file   = "..."
  tls_key_file    = "..."
}

storage "raft" {
  path = "/opt/vault/data"
  node_id = "..."

  retry_join {
    auto_join = "provider=aws addr_type=private_v4 tag_key=... tag_value=... region=..."
  }
}

seal "awskms" {
  kms_key_id = "..."
  region     = "..."
}

telemetry {
  ...
}

api_addr = "https://...:8200"
cluster_addr = "https://...:8201"

unsafe_cross_namespace_identity = true

ui = true
log_level = "info"

# When initialize is here, the node cannot join to exist quorum
# When this is removed, everything works
initialize "setup-openbao-terraform-access" {
  # operation like creating a policy
}
```

diogotorres97:
## Description

A node with both an `initialize` stanza and `retry_join` can bootstrap its own single-node cluster while the join is still in flight: `InitiateRetryJoin` backgrounds the retry loop, so self-init runs alongside it rather than after. This waits for an outstanding join first, with a timeout so the first node of a new cluster still comes up.

I went with polling `core.Initialized` rather than the `raftJoinDoneCh` approach sketched in the issue - `joinInProgress` is only set under Shamir, and `Initialize` bails out without auto-unseal, so that signal never fires on this path.

It doesn't cover several nodes starting at once with no leader to find; they all time out and self-init.
## Rationale

Resolves: https://github.com/openbao/openbao/issues/3652

## Acknowledgements

 - [x] By contributing this change, I certify I have not used generative AI GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or filling out the pull request description or associated issue.
 - [x] By contributing this change, I certify I have signed-off on the [DCO ownership](https://developercertificate.org/) statement and this change did not use post-BUSL-licensed code from HashiCorp. Existing MPL-licensed code is still allowed, subject to attribution. Code authored by yourself and submitted to HashiCorp for inclusion is also allowed.

### DIFF
--- a/internal/command/self_init_retry_join_test.go
+++ b/internal/command/self_init_retry_join_test.go
@@ -1,0 +1,122 @@
+package command
+
+import (
+	"context"
+	"io"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/cli"
+	log "github.com/hashicorp/go-hclog"
+	testintf "github.com/mitchellh/go-testing-interface"
+	"github.com/openbao/openbao/v2/internal/command/server"
+	"github.com/openbao/openbao/v2/internal/helper/testhelpers/teststorage"
+	vaulthttp "github.com/openbao/openbao/v2/internal/http"
+	"github.com/openbao/openbao/v2/internal/vault"
+	"github.com/openbao/openbao/v2/internal/vault/seal"
+	"github.com/stretchr/testify/require"
+)
+
+// Waiting on anything but raft would add most of a minute to the first boot of
+// every other deployment. Core is nil on purpose: reaching for it here is the
+// bug we're guarding against.
+func TestWaitForRaftRetryJoinSkipsNonRaft(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
+		name    string
+		storage *server.Storage
+	}{
+		{name: "file_storage", storage: &server.Storage{Type: "file"}},
+		{name: "inmem_storage", storage: &server.Storage{Type: "inmem"}},
+		{name: "raft_ha_storage_only", storage: &server.Storage{Type: "postgresql"}},
+		{name: "no_storage", storage: nil},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			cmd := &ServerCommand{logger: log.NewInterceptLogger(&log.LoggerOptions{Output: io.Discard})}
+
+			start := time.Now()
+			joined, err := cmd.waitForRaftRetryJoin(context.Background(), nil, &server.Config{Storage: tc.storage})
+
+			require.NoError(t, err)
+			require.False(t, joined, "reported a join we never waited for")
+			require.Less(t, time.Since(start), time.Second,
+				"we waited, so the wait is being applied to non-raft storage")
+		})
+	}
+}
+
+// testRaftCore brings up a single-node raft-backed cluster, optionally with a
+// retry_join stanza, and hands back a command and its core.
+func testRaftCore(t *testing.T, retryJoin string) (*ServerCommand, *vault.Core, func()) {
+	t.Helper()
+
+	testSeal, _ := seal.NewTestSeal(nil)
+	autoSeal, err := vault.NewAutoSeal(testSeal)
+	require.NoError(t, err)
+
+	logger := log.NewInterceptLogger(&log.LoggerOptions{Output: io.Discard})
+	conf, opts := teststorage.ClusterSetup(&vault.CoreConfig{
+		Logger: logger,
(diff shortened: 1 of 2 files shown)


diogotorres97:
I will be testing this against my cluster and report it here later.

diogotorres97:
Tested this on EKS: 3 replica StatefulSet with OrderedReady, awskms auto-unseal, retry_join pointing at the three `openbao-N.openbao-internal` addresses and the same `initialize` block on every node (chart 0.29.2).

We built our own image for it: v2.6.2 with this change applied on top (same server.go diff), UI included, and the binary swapped into the official 2.6.2 image so everything else matches stock.

Wiped the PVCs before each run and only changed the image.

v2.6.2:

```
$ for i in 0 1 2; do kubectl exec openbao-$i -- bao status | grep "Cluster ID"; done
Cluster ID               62920a70-15e7-9bce-5a5f-79379b7d16db
Cluster ID               bf165740-92bb-5287-826c-c0ee0fa7f606
Cluster ID               5ecf1a6d-e76e-cc06-0bcd-f34baf03efca
```

openbao-1 bootstraps its own raft ~30ms after starting the retry join:

```
15:29:30.931 raft retry join initiated
15:29:30.959 initial configuration servers=[openbao-1]
15:29:39.693 election won tally=1
15:29:39.694 failed to retry join raft cluster: failed to send answer to raft leader node: raft is already initialized
```

With this PR (ran it twice, same result both times):

```
Cluster ID               abc0b92e-5c61-3aa5-54a5-725ba8480eae
Cluster ID               abc0b92e-5c61-3aa5-54a5-725ba8480eae
Cluster ID               abc0b92e-5c61-3aa5-54a5-725ba8480eae
```

openbao-0 waits the full 35s and self-inits, the other two join:

```
15:04:22.867 raft retry join initiated
15:04:22.889 waiting for raft retry join
15:04:22.972 successfully joined the raft cluster leader_addr=https://openbao-0.openbao-internal.openbao.svc.cluster.local:8200
15:04:23.890 raft retry join completed, skipping self-initialization
```

So the join needs about 100ms, but without the wait the node has already bootstrapped its own raft after ~30ms.

PR #3993, opened 2026-09-12
add challenge_permitted/excluded_ip_ranges to PKI ACME

### LINKED ISSUE #3249: Support `challenge_permitted_ip_ranges` / `challenge_excluded_ip_ranges` in PKI ACME config to allow issuance for RFC1918 (private) IP addresses
## Is your feature request related to a problem? Please describe.

We run OpenBao as an internal CA in an environment **without internal DNS**, so services are addressed by IP address rather than hostname. We want to issue and auto-renew TLS certificates for these services over ACME (the consuming clients, e.g. the GitLab Omnibus container, only support ACME for automated renewal).

Since DNS is not available, `dns-01` is not an option (and is forbidden for IP identifiers per RFC 8738 anyway), so we rely on `http-01`. IP-identifier issuance itself works at the role level (`allow_ip_sans=true`), and GitLab’s ACME client correctly sends the IP as an `ip` identifier.

However, the ACME challenge validation rejects RFC1918 addresses with:

```
urn:ietf:params:acme:error:rejectedIdentifier
Cannot issue for "192.168.x.x": IP address is in a reserved address block: [RFC1918]: Private-Use
```

There appears to be no configuration option in OpenBao to permit issuance for private/reserved IP ranges. For an internal-only CA, refusing RFC1918 addresses is exactly the opposite of what’s needed.

## Describe the solution you’d like

HashiCorp Vault added two options to the PKI ACME configuration (`pki/config/acme`) that solve this:

- `challenge_permitted_ip_ranges` (list of CIDRs) — if set, only IPs within these ranges are allowed for challenge validation.
- `challenge_excluded_ip_ranges` (list of CIDRs) — IPs within these ranges are rejected; takes precedence over the permitted list.

Reference: <https://developer.hashicorp.com/vault/api-docs/secret/pki/issuance>

It would be great if OpenBao supported equivalent configuration, so that an operator can explicitly opt in to issuing for private ranges, e.g.:

```
bao write pki/config/acme \
enabled=true \
challenge_permitted_ip_ranges="[192.168.0.0/16,10.0.0.0/8,172.16.0.0/12](http://192.168.0.0/16,10.0.0.0/8,172.16.0.0/12)"
```

The default (empty) should preserve the current safe behavior.

## Describe alternatives you’ve considered

- **Direct PKI API issuance (`/pki/issue/:role`) + cron**: works fine for private IPs, but loses the ACME auto-renewal integration that downstream tools (GitLab, cert-manager, etc.) expect out of the box.
- **Using a hostname + `/etc/hosts` entries**: not viable at scale and defeats the purpose of an IP-addressed environment.
- **Switching to Vault**: Vault’s `challenge_permitted_ip_ranges` does resolve this (confirmed working in our setup), but we’d prefer to stay on OpenBao.

## Explain any additional use-cases

Internal/private PKI for IP-addressed infrastructure is a common pattern (lab environments, air-gapped networks, appliances without DNS records). Allowing operators to opt in to RFC1918 issuance makes OpenBao usable as a drop-in internal CA for these cases while keeping the public-facing default safe.

## Additional context

- OpenBao version tested: v2.5.4 (`OpenBao v2.5.4 (4f6d47246a053375271a5fd8af85c3b75695aa46), built 2026-05-20T16:08:53Z`)
- ACME client: GitLab Omnibus (`gitlab-ce`) built-in Let’s Encrypt integration
- Related: this is the same `rejectedIdentifier` behavior public CAs enforce; the request is specifically to make it configurable for private/internal deployments.

Luiggi33:
## Description

Add challenge_permitted_ip_ranges and challenge_excluded_ip_ranges flags to the PKI ACME.

## Rationale

Resolves: #3249 (Implements the feature, does not solve the problem)

## Acknowledgements

       also allowed.


Luiggi33:
One thing I am still unsure about is the "dont retry when this error happens", as this does not need retries but i am unsure if i chose the right way to block them. I would appreciate if someone could enlighten me on this :)

PR #3994, opened 2026-09-13
trim whitespace from pki config cluster options

### LINKED ISSUE #3798: PKI - Mount's API path not trimmed, producing invalid ACME URLs
**Describe the bug**
When configuring a PKI secret engine cluster config using the Web UI, trailing whitespace is not trimmed.
When a request to /acme/directory is then made, it returns the whitespace as part of the constructed URLs for ACME operations.

When calling the returned endpoints, the URL being incorrect, the server returns a 403.

**To Reproduce**
1. Configure a PKI secret engine and enable ACME as explained in the documentation, using the Web UI
2. Introduce a trailing whitespace to the "Mount's API path" field (eg: "https://vault.domain.lab/v1/pki ")
3. Save
4. Run `curl https://vault.domain.lab/v1/pki/acme/directory` 
5. Observe that the output contains the encoded whitespace as "%20"
```
{"keyChange":"https://vault.domain.lab/v1/pki%20/acme/key-change","meta":{"externalAccountRequired":false},"newAccount":"https://vault.domain.lab/v1/pki%20/acme/new-account","newNonce":"https://vault.domain.lab/v1/pki%20/acme/new-nonce","newOrder":"https://vault.domain.lab/v1/pki%20/acme/new-order","revokeCert":"https://vault.domain.lab/v1/pki%20/acme/revoke-cert"}
```

Using certbot when issuing a certificate : 
```
Requesting fresh nonce
Sending HEAD request to https://vault.domain.lab/v1/pki%20/acme/new-nonce.
https://vault.domain.lab:443 "HEAD /v1/pki%20/acme/new-nonce HTTP/1.1" 403 0
Received response:
HTTP 403
Cache-Control: no-store
Content-Type: application/json
Strict-Transport-Security: max-age=31536000; includeSubDomains
Date: Fri, 14 Aug 2026 07:00:21 GMT
Content-Length: 33


Exiting abnormally:
<STACKTRACE REMOVED>
An unexpected error occurred:
acme.errors.ClientError: <Response [403]>
```
**Expected behavior**
When saving the config page, fields are trimmed to avoid accidentally typed whitespace.

**Environment:**
Version         2.6.1
Commit Date     2026-07-22T14:22:20Z

OpenBao v2.6.1 (ba7ad8861d0578cd4da4f7b9e5a6756d30484f8f), committed 2026-07-22T14:22:20Z
Podman container

```hcl
# Be sure to scrub any sensitive values
```

**Additional context**
Removing the trailing space from the Mount's API path field causes the ACME directory to return correctly formed URLs without %20.

Luiggi33:
## Description

Trim the whitespace from the url for the config/cluster options. This is to prevent malformed accidental configurations, as trailing whitespaces are not caught by the url.Parse down the line and would only be noticed after configuring everything

## Rationale

Resolves: #3798

## Acknowledgements

       also allowed.


changelog/3994.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+pki/config: Trim whitespace from cluster config path options.
```
cipherboy:
```suggestion
secrets/pki: Trim whitespace from cluster config's `path` and `aia_path` options.
```

wslabosz-reply (approved):

cipherboy:
@wslabosz-reply @Luiggi33 I'm still not convinced this is a good idea.

Where do we stop?

1. Do we also include other URLs, like [`oidc_discovery_url`](https://openbao.org/docs/api/auth/jwt/#parameters)? 
2. Do we also include [`uri_sans`](https://openbao.org/docs/api/secret/pki/#generate-certificate-and-key), the closely related cousin to URLs?
3. Do we include other random parameters that can [probably not contain spaces but are likely to be copy/pasted and thus may accidentally get spaces added, like `default` (and other `issuer_ref` usages)](https://openbao.org/docs/api/secret/pki/#set-issuers-configuration)? 

Now that https://github.com/openbao/openbao/security/advisories/GHSA-fg5x-7whg-6c28 is made public, how does any of that interact with this general problem of canonicalization? More broadly, outside of CEL support landing in v2.8, this further decreases the effectiveness of [e.g., `denied_parameters` &c](https://openbao.org/docs/concepts/policies/#parameter-constraints), at a time when it already, unfortunately, doesn't apply to some native `sdk/framework` values much less this.

Ultimately, I think it is best to trust that the user is doing what they intend and don't second guess them.

---

At minimum, if we move forward with this, I don't think this should be server-side; it should be client-side.

wslabosz-reply:
@cipherboy 
I think that's fair, at first I thought about adjusting this so the ui trims whitespaces on the input, but then found out that some engines have specific validators declared with one checking for whitespace (and err message of: `Your secret path contains whitespace. If this is desired, you'll need to encode it with %20 in API calls.`).

my aversion to the UI made me think that trimming it on the server-side was better idea, but I guess the canonicalization makes it more troublesome.

PR #3998, opened 2026-09-14
Fix: maxVersions logic

### LINKED ISSUE #3997: KV max_versions not overwriting when mount has higher value defined than secret metadata
**Describe the bug**

Currently Openbao does not always allow overriding the max versions to keep. 
In particular when the mount config has a higher `max_versions` set than the secret entry metadata.
The [Docs of the KV engine](https://openbao.org/docs/api/secret/kv/kv-v2/#configure-the-kv-engine) however, explicitly states that overwriting is possible: 
```
max_versions (int: 0) – The number of versions to keep per key. This value applies to all keys, but a key's metadata setting can overwrite this value.
```

**To Reproduce**
1. Create a KV version 2 mount and explicitly set max_version (e.g. 5) : 
```
bao secrets enable -version=2 -path="kv2-test" kv 
bao write kv2-test/config max_versions=5
```
3. Create a secret in this KV mount: `bao kv put -mount=kv2-test my-secret dummy=data`)
4. Set the secret metadata max_version to something lower than the mount (e.g. 1): `bao kv metadata put -mount=kv2-test -max-versions 1 my-secret`
5. Create a new version of the secret: `bao kv put -mount=kv2-test my-secret dummy=data2`
6. Both versions are still available, even though we set max_version of the secret to 1 (`bao kv get -mount=kv2-test -version=1 my-secret`)

**Expected behavior**
The overwrite max_version should take effect and delete all other older versions.

**Environment:**
* OpenBao Server Version (retrieve with `bao status`): 2.6.2 (but also on latest)

**Additional context**

mase-informaticon:
## Description

The [Docs of the KV engine](https://openbao.org/docs/api/secret/kv/kv-v2/#configure-the-kv-engine) explicitly states: 
```
max_versions (int: 0) – The number of versions to keep per key. This value applies to all keys, but a key's metadata setting can overwrite this value.
```
The current code does not always allow overriding the max versions to keep, especially when the mount config has a higher `max_versions` set than the secret entry metadata.
My proposed code change parses the correct precedence of `max_versions`.

## Rationale

Resolves: #3997

## Acknowledgements

       also allowed.


cipherboy:
Thanks @mase-informaticon! Do you mind adding a test and a changelog entry? See https://github.com/openbao/openbao/blob/main/CONTRIBUTING.md#changelog-entries for more information on the latter.

mase-informaticon:
Hi @cipherboy, as requested, I added changelog and test cases.

changelog/3998.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+secrets/kv: Fix `max_versions` precedence for KVv2.
```
wslabosz-reply, line 2:
```suggestion
secrets/kv: Properly respect key metadata over mount entry `max_versions` config property precedence. 
```

PR #4001, opened 2026-09-14
Change userpass 'missing password' response status code to 400

### LINKED ISSUE #3990: userpass: missing password returns HTTP 500 for malformed login input
An anonymous `userpass` login request with an omitted or empty password returns HTTP 500 and `{"errors":["missing password"]}`. The request creates no authenticated session, but malformed client input is reported as a server failure.

Measured on OpenBao 2.6.2, source `dd9c19c37a878cf4a81b18efb8d6f0599c7da923`, official Linux amd64 image under OCI index `sha256:11fd73a2102cda9c55d5d881a8c3210303146a7ec1e8ac76f526e175c6d24641`. This was a disposable single-node TLS/Raft instance, with synthetic users only. No production credentials or data were involved.

After enabling `userpass` and creating a synthetic test user, send:

```sh
curl --cacert "$CA" --request POST \
  --header 'Content-Type: application/json' \
  --data '{}' \
  "$BAO_ADDR/v1/auth/userpass/login/test-user"
```

Observed controls on that same instance:

- Missing password: HTTP 500, `missing password`, no token.
- Nonempty incorrect password: HTTP 400, `invalid username or password`, no token.
- Correct synthetic password: HTTP 200 with the configured restricted policy.

The likely source is the empty-password branch in [v2.6.2 path_login.go](https://github.com/openbao/openbao/blob/v2.6.2/builtin/credential/userpass/path_login.go#L79), which returns `errors.New("missing password")`. Other input/login errors use a logical error response. A client-error response for missing required input would avoid classifying this malformed request as a server fault. I am not asserting a documented requirement for one specific 4xx status, or an authentication bypass.

The downstream client will continue to reject missing passwords before transport and fail closed on every non-success response. This report is about the server's error classification. Searches for existing `userpass "missing password"` and `userpass 500` issues found no matching report.

wslabosz-reply:
## Description
Changes the status code in response to `missing password` case within userpass login handler from 500 to 400.

## Rationale
5x status error codes indicate server error, where in this case it is clearly invalid input provided by the user.

Resolves: #3990 

## Acknowledgements

       also allowed.


cipherboy (approved):
I do wonder some days if there's a rule or check we can put in place to help us audit error messages better. 🤔

wslabosz-reply:
@cipherboy 
this sounds more like "test all existing paths under the sun and moon"

JanMa (approved):

cipherboy:
@wslabosz-reply I was thinking something that ties into:

https://github.com/openbao/openbao/blob/7939df9588402b547268b7760b526fb8348bf8a5/sdk/helper/testhelpers/schema/response_validation.go#L163-L167

or similar for the default testcluster, and perhaps do some heuristic-based detection to see if we're returning `(nil, err)` for an error text that looks like it _ought to be_ a coded/user error of some sort. 

You can see usage of that function and how its called in some places, honestly, probably not used widely enough tbh.

PR #4009, opened 2026-09-15
Support External Mu for MLDSA Signatures

### LINKED ISSUE #4008: MLDSA: Support signing externally calculated Mu

It would be nice to be able to provide Mu to the MLDSA signing functions. This would allow for traffic to be cut down when signing larger objects while retaining a pure MLDSA signature. External Key platforms such as AWS's KMS already support this functionality.

The API would be extended to let it be specified that the payload being signed is an externally calculated mu. This would then get passed down via crypto.MLDSAMu to the sign function. Checks would be put in place to ensure that the payload is 64 bytes, matching Mu expectations.

Certain specs I'm running into require pure MLDSA signatures. Sending large payloads to be signed can put strain on the network. Also, if the objects needing to be signed are sensitive in nature, keeping them in memory and avoiding sending them to be signed is more secure.

DeadlySurgeon:
## Description

Adds support for specifying an externally computed Mu is being provided to the MLDSA signing function.

## Rationale

Would help with cutting down on traffic when signing large objects as well as keeping sensitive objects needing to be signed from going over the wire. 

Resolves: https://github.com/openbao/openbao/issues/4008

## Acknowledgements

       also allowed.


DeadlySurgeon:
Since HashML-DSA is something separate from an Externally provided Mu, leveraging the `prehashed` field may not be correct, which is why I'm suggesting adding the `mldsa_external_mu` field. Open to suggestions if there are any to be able to support this feature.

satoqz:
@DeadlySurgeon WDYT about taking `prehashed=true` + `hash_algorithm=mldsa-mu`?

DeadlySurgeon:
> @DeadlySurgeon WDYT about taking `prehashed=true` + `hash_algorithm=mldsa-mu`?

That would work and prevent the contract from needing to be changed. Making adjustments to my PR now.

DeadlySurgeon:
@satoqz updated my PR to reflect your suggestion. Let me know if anything else is needed!

satoqz:
Thanks, looking fairly good!

Some remaining items:

- Docs at `website/content/docs/api/secret/transit.mdx` will need a tiny update.
- https://github.com/openbao/go-kms-wrapping/blob/9184e29ddec41fb6dae447b6292264a65b512f84/kms/transit/transit.go#L214-L223 needs an update so we can dogfood this against a Transit external key backend. Feel free to pick this up if you'd like.
- @cipherboy and I need to figure out if we'd like to backport this for v2.7 GA, which will decide if at all or in which place we'll need a changelog 🤔 I'll mark this as backport + no changelog for now until I hear other opinions.

DeadlySurgeon:
@satoqz updated `website/content/docs/api/secret/transit.mdx`. I can make the changes to `kms/transit/transit.go` later, I don't have the time present to do so. It might be faster for you to add it.

sdk/helper/keysutil/consts.go
```diff
@@ -74,6 +76,7 @@ var (
 		HashTypeSHA3256: crypto.SHA3_256,
 		HashTypeSHA3384: crypto.SHA3_384,
 		HashTypeSHA3512: crypto.SHA3_512,
+		HashTypeMLDSAMu: crypto.MLDSAMu,
```
satoqz: Tangential and not to be addressed here, this is slightly scary because `.New()` will panic.

I suppose I'll need to better guard sites such as

https://github.com/openbao/go-kms-wrapping/blob/9184e29ddec41fb6dae447b6292264a65b512f84/kms/pkcs11/ec.go#L101

with `.Available()`.

DeadlySurgeon: I looked at gate keeping the alg to only MLDSA paths, which seemed like it wouldn't ever manual call `New`, but I agree. Potentially some helper where:

```go
func NewHash(h crypto.Hash) (hash.Hash, error) {
    if !h.Available() {
        return nil, fmt.Errorf("hash %q not available", h.String())
    }
    return h.New(), nil
}
```

DeadlySurgeon: Looks like we already saw this in the test. What would your thoughts be on leaving it out of `CryptoHash` and instead having logic around the hashtype where it is supported?

IE in `SignWithOptions` `case KeyType_ExternalKey`:
```go
		var algo crypto.Hash
		if hashAlgorithm == HashTypeMLDSAMu {
			algo = crypto.MLDSAMu
		} else {
			var ok bool
			if algo, ok = CryptoHashMap[hashAlgorithm]; !ok {
				return nil, errutil.InternalError{Err: "unsupported hash algorithm"}
			}
		}
```

satoqz: Sounds good!

DeadlySurgeon: Addressed. Additional checks around hash availability would likely be a good place for future improvements.

satoqz (dismissed):

satoqz:
Ah, looks like there are more tests that try all hash functions.

DeadlySurgeon:
> Ah, looks like there are more tests that try all hash functions.

Yup. Already on fixing them.

DeadlySurgeon:
@satoqz test-go(8) seems to be failing on something unrelated? Thoughts?

satoqz:
@DeadlySurgeon see https://github.com/openbao/openbao/issues/2581 😓

satoqz (approved):
Thanks!

internal/builtin/logical/transit/path_sign_verify.go
```diff
@@ -611,6 +617,10 @@ func (b *backend) pathVerifyWrite(ctx context.Context, req *logical.Request, d *
 		return logical.ErrorResponse("hash_algorithm=none requires both prehashed=true and signature_algorithm=pkcs1v15"), logical.ErrInvalidRequest
 	}
 
+	if hashAlgorithm == keysutil.HashTypeMLDSAMu {
```
cipherboy, line 620: https://github.com/golang/go/issues/80272 is a little daft of a use case, but appears to be the tracking issue for this. It'd be good to have a reference here so we know its not our limitation but one of Go's.

sdk/helper/keysutil/policy.go
```diff
@@ -1531,6 +1555,10 @@ func (p *Policy) VerifySignatureWithOptions(derivationContext, input []byte, sig
 		return false, errutil.UserError{Err: fmt.Sprintf("message verification not supported for key type %v", p.Type)}
 	}
 
+	if options.HashAlgorithm == HashTypeMLDSAMu {
+		return false, errutil.UserError{Err: "external ML-DSA mu verification is not supported"}
```
cipherboy, line 1559: I suspect some external keys implementations can do this, but we'll leave it as a feature request for later.

cipherboy (commented):
> @cipherboy and I need to figure out if we'd like to backport this for v2.7 GA, which will decide if at all or in which place we'll need a changelog 🤔 I'll mark this as backport + no changelog for now until I hear other opinions.

I'm fine with the backport. Thanks @DeadlySurgeon!

cipherboy (approved):

DeadlySurgeon:
@cipherboy While it is unfortunate that there isn't a mechanism for verifying without sending the entire payload due to limitations in Go, it kinda shouldn't matter. There might even be good reason to not include this functionality by default even when/if Go adds support for verifying an External Mu.

In order to generate Mu, you'd need to have the Public Key. If you have the Public Key you don't really need to delegate to OpenBao for verification as you can perform verification yourself (so long as you're not missing MLDSA crypto libraries). If you don't have the Public Key, you can't possibly be generating Mu, meaning whatever Mu you are producing isn't tied to the message so verifying the signature doesn't verify the message.

PR #4013, opened 2026-09-16
Fix lint finding: Inappropriate key in call to context.WithValue

klaus-sap:
## Rationale

Resolves:
lint finding: https://staticcheck.dev/docs/checks/#SA1029

## Acknowledgements

       also allowed.


internal/vault/mount.go
```diff
@@ -1651,7 +1653,9 @@ func (c *Core) newLogicalBackend(ctx context.Context, entry *routing.MountEntry,
 	}
 
 	ctx = namespace.ContextWithNamespace(ctx, entry.Namespace)
-	ctx = context.WithValue(ctx, "core_number", c.coreNumber)
+	// key should not be of type string or any other build-in type
+	const core_number contextKey = "core_number"
+	ctx = context.WithValue(ctx, core_number, c.coreNumber)
```
satoqz, line 1658: It doesn't look like this value is retrieved anywhere? So we should be good to drop the `WithValue` call entirely?

klaus-sap: @satoqz not sure if I understand it right, but the return value ctx is used in the next line

satoqz: @klaus-sap 

The following pattern attaches a value to the context that we already have:

```go
ctx = context.WithValue(ctx, key, value)
```

Note how we're just wrapping a `ctx` that we already have.

If nothing ever ends up later calling `ctx.Value(key)` to retrieve this value again, wrapping the context is useless and we can just use the original context.

satoqz: Also see https://pkg.go.dev/context#WithValue.

klaus-sap: Ok, but are you sure that the core_number value in the derived context is used nowhere? Removing the WithValue call seems to be an incompatible change to me.

satoqz: @klaus-sap search the codebase for `core_number` and see for yourself. This is the only reference. I also don't expect this key to have been dynamically constructed elsewhere.

And if all goes wrong, this looks to be testing instrumentation anyway. Core is a singleton except within tests, so that's the only reason to assign core numbers. Looks like `Core.coreNumber` is generally unused.

Also note that if you're going to give the context key a dedicated type, any call to retrieve the value via `ctx.Value()` is also going to need to reference that context key type. That's the entire point of the lint, and you would've broken any retrievers already by fixing the lint without updating them.

klaus-sap: > @klaus-sap search the codebase for `core_number` and see for yourself. This is the only reference. I also don't expect this key to have been dynamically constructed elsewhere.

I did, but still I don't want to take any risk - this WithValue instruction doesn't hurt.

klaus-sap:
Different opinions on the best way to fix this finding

satoqz:
#4014

PR #4016, opened 2026-09-16
Fix lint findings: remove deprecated field PluginName

klaus-sap:
## Rationale

Resolves: 
lint findings: https://staticcheck.dev/docs/checks/#SA1019 (partially)

## Acknowledgements

       also allowed.


klaus-sap:
Is the generic plugin type routing.MountTypePlugin used anymore?
Or can we remove all coding referring to routing.MountTypePlugin and the deprecated routing.MountConfig.PluginName?
See also: https://openbao.org/community/release-notes/2-5-0/#changes-1

satoqz:
The struct field and any conditioning around it look good to remove IMO, it's been long enough. It should however remain possible to create a mount via `type=plugin`, so some conditioning around that string must stay, likely only in `logical_system.go` though.

klaus-sap:
> It should however remain possible to create a mount via `type=plugin`, so some conditioning around that string must stay, likely only in `logical_system.go` though.

@satoqz I don't understand why we need `type=plugin` anymore. If we continue to need some conditioning, then we cannot remove the deprecated attribute routing.MountConfig.PluginName, right?

satoqz:
> @satoqz I don't understand why we need `type=plugin` anymore.

For backwards-compatibility? It's a valid set of parameters to enable a mount. This might also be deprecated but I can't find a clear notice, and I don't see a good reason to remove it.

> then we cannot remove the deprecated attribute routing.MountConfig.PluginName, right?

No, we should be able to remove it because these are distinct. `type=plugin` is re-routed not to use the deprecated field internally.

One is a user-visible change, the other isn't.

internal/vault/request_handling.go
```diff
@@ -1508,14 +1508,6 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
 				registerLease = false
 				resp.Secret.Renewable = false
 			}
-
-		case "plugin":
-			// If we are a plugin type and the plugin name is "kv" check the
-			// mount entry options.
-			if matchingMountEntry.Config.PluginName == "kv" && (matchingMountEntry.Options == nil || matchingMountEntry.Options["leased_passthrough"] != "true") {
```
klaus-sap, line 1515: if condition is never true

wslabosz-reply (dismissed):
I think the proposal is fine as the removed fields were marked as deprecated, but this change definitely warrants a changelog.

klaus-sap:
> I think the proposal is fine as the removed fields were marked as deprecated, but this change definitely warrants a changelog.

Added changelog with d56da231556fec8b8b0c12b63a4e135ba11abb59

changelog/4016.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:deprecation
+api/sys_mounts.go: Remove deprecated field `PluginName` from `MountConfig`.
```
wslabosz-reply:
```suggestion
api: Remove deprecated field `PluginName` from `MountConfig` and `MountInputConfig`.
```

changelog/4016.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:deprecation
+api: Remove deprecated field `PluginName` from `MountConfig` and `MountInputConfig`.
+```
```
satoqz:
````suggestion
```release-note:change
core/mounts: Remove support for mounts enabled via `type=plugin plugin_name=...` before Vault v1.0.
```
````

It was already deprecated before, so it's a change. The `api:` prefix is used for the `/api` package, not changes to the REST API (not that this is one, anyhow). We try to state the user-visible change here opposed to what changed in the internals.

satoqz:
> No, we should be able to remove it because these are distinct. `type=plugin` is re-routed not to use the deprecated field internally.
> 
> One is a user-visible change, the other isn't.

To correct myself: They're both user-visible, but the former would break requests that are currently valid while the latter affects storage state written before Vault v1.0 only.

wslabosz-reply (approved):

PR #4028, opened 2026-09-17
Fix deprecations around raw EC key params

satoqz:
## Description

Replaces access to raw EC key params D, X and Y (which are deprecated) with `key.Bytes()` (returning a `[]byte` version of D) and `ecdsa.ParseRawPrivateKey`.

This cleans up core, there remain references in Transit which I'd look at addressing separately.

## Acknowledgements

       also allowed.


wslabosz-reply (approved):

internal/vault/ha.go
```diff
@@ -1602,12 +1603,13 @@ func (c *Core) advertiseLeader(ctx context.Context, uuid string, leaderLostCh <-
...
-		Type: corePrivateKeyTypeP521,
-		X:    key.X,
-		Y:    key.Y,
-		D:    key.D,
+		Type: certutil.PrivateKeyTypeP521,
+		D:    &big.Int{},
 	}
 
+	d, _ := key.Bytes()
+	keyParams.D.SetBytes(d)
```
satoqz, line 1611: Just to have stated the obvious, this will break downgrades -- but I figured we'd agreed that those are out of scope. I don't think we can address this deprecation without doing so.

PR #4032, opened 2026-09-18
Implement BAO_HEADERS env

### LINKED ISSUE #4017: Add BAO_HEADERS env variable support
I have my OpenBao instance behind HTTP IAP.
I use header with IAP auth token to bypass IAP and reach openbao instance.
So I am forced to supply `-header` argument to each of my cli calls.
I would like to set this configuration parameter one time and apply it automatically to each of my CLI call in terminal session.

I would like to see `BAO_HEADERS` env var. Similar to vault cli https://github.com/hashicorp/vault/pull/21993 

**Additional context**

dm3ch:
## Description

Adds `BAO_HEADERS` env variable, allowing to supply custom headers to each CLI api call.

## Rationale

Resolves: #4017

## Acknowledgements

       also allowed.


dm3ch:
@satoqz FYI

api/client.go
```diff
@@ -834,6 +844,32 @@ func NewClient(c *Config) (*Client, error) {
 	return client, nil
 }
 
+func readEnvironmentHeaders() (map[string]string, error) {
+	raw := ReadBaoVariable(EnvVaultHeaders)
+	if raw == "" {
+		return nil, nil
+	}
+
+	var values map[string]any
```
satoqz: Why not just:

```suggestion
	var headers map[string]string
```

and unmarshal into that?

dm3ch: Makes sense. Fixed in fcfcccb65ab43a6c3f382e617f38c2a3623e7eba

dm3ch:
@satoqz Can you please approve the workflow run one more time. Release note and golang-ci were fixed.

BTW is it fine to fix errcheck like this https://github.com/openbao/openbao/pull/4032/changes/9869c846e095a4e5115abc0e55ff84dcecbb7da7 ?

dm3ch:
It seems that e2e failure is not related to the changes

api/client_test.go
```diff
@@ -26,8 +26,9 @@ import (
 
 func init() {
 	// Ensure our special envvars are not present
-	os.Setenv("BAO_ADDR", "")
-	os.Setenv("BAO_TOKEN", "")
+	_ = os.Setenv("BAO_ADDR", "")
+	_ = os.Setenv("BAO_TOKEN", "")
+	_ = os.Setenv(EnvVaultHeaders, "")
```
satoqz:
```suggestion
	_ = os.Unsetenv(EnvVaultAddress)
	_ = os.Unsetenv(EnvVaultToken)
	_ = os.Unsetenv(EnvVaultHeaders)
```

website/content/docs/commands/index.mdx
```diff
@@ -403,6 +403,20 @@ HTTP or HTTPS proxy location which should be used by all requests to access Open
...
+
+```shell-session
+$ export BAO_HEADERS='{"X-IAP-Token":"example-token"}'
+```
+
+Header names beginning with `X-Vault-` are reserved for OpenBao and are not
+allowed. Protect this environment variable when header values contain secrets.
+Unlike other CLI flags, a `-header` value with the same name is appended to the
+environment-provided value instead of replacing it.
+
```
satoqz: This cuts into the middle of the `BAO_PROXY_ADDR` section. I'd place it below `BAO_DISABLE_REDIRECTS`.

changelog/4032.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+api: Added `BAO_HEADERS` env allowing to supply custom headers to each cli API request.
+```
```
satoqz, line 6:
````suggestion
```release-note:improvement
api: Support the `BAO_HEADERS` environment variable to add custom HTTP headers to each API request.
```
```release-note:improvement
command: Support the `BAO_HEADERS` environment variable to add custom HTTP headers to avoid needing to supply `-header` with each CLI invocation.
```
````

Let's document that this affects both api generally and the CLI.

File: api/client.go [hunk 5]
```diff
@@ -53,6 +53,7 @@ const (
 	EnvHTTPProxy             = "BAO_HTTP_PROXY"
 	EnvVaultProxyAddr        = "BAO_PROXY_ADDR"
 	EnvVaultDisableRedirects = "BAO_DISABLE_REDIRECTS"
+	EnvVaultHeaders          = "BAO_HEADERS"
```
satoqz:
```suggestion
	EnvHeaders          = "BAO_HEADERS"
```

Let's avoid the Vault name going forward, whenever it makes sense.

website/content/docs/commands/index.mdx
dm3ch: Fixed

dm3ch:
@satoqz TY for the review. Fixed all review findings

website/content/docs/commands/index.mdx
```diff
@@ -440,3 +440,17 @@ list of available flags, run:
...
+be a JSON object with string keys and string values. For example:
+
+```shell-session
+$ export BAO_HEADERS='{"X-IAP-Token":"example-token"}'
+```
+
+Header names beginning with `X-Vault-` are reserved for OpenBao and are not
+allowed. Protect this environment variable when header values contain secrets.
+Unlike other CLI flags, a `-header` value with the same name is appended to the
+environment-provided value instead of replacing it.
```
satoqz: This should still go above the "Flags" section.

> Protect this environment variable when header values contain secrets.

Is rather hand-wavy, I'd omit it.

> with string keys and string values

keys are always strings in JSON, so I'd leave that out and mention values only.

api/client_test.go
```diff
@@ -70,6 +71,40 @@ func TestDefaultConfig_envvar(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NotContains(t, client.Headers(), "X-Test")
+}
+
+func TestNewClient_EnvHeadersNotFlatMap(t *testing.T) {
+	t.Setenv(EnvHeaders, `{"X-Nested":{"foo":"bar"}}`)
+
+	_, err := NewClient(nil)
+	require.Error(t, err)
+}
```
satoqz, line 108: Can this be one test with several `t.Run(...)` subtests for better grouping?

api/client.go
```diff
@@ -834,6 +844,26 @@ func NewClient(c *Config) (*Client, error) {
...
 
+func readEnvironmentHeaders() (map[string]string, error) {
+	raw := ReadBaoVariable(EnvHeaders)
+	if raw == "" {
+		return nil, nil
+	}
+
+	var headers map[string]string
+	if err := json.Unmarshal([]byte(raw), &headers); err != nil {
+		return nil, fmt.Errorf("could not parse %s as a JSON headers map", EnvHeaders)
```
satoqz:
```suggestion
		return nil, fmt.Errorf("parse %s: %w", EnvHeaders, err)
```

internal/command/base_test.go
```diff
@@ -70,3 +80,32 @@ func TestClient_FlagHeader(t *testing.T) {
...
+}
+
+func TestClient_EnvironmentReservedHeader(t *testing.T) {
+	isolateHeaderEnvironment(t)
+	t.Setenv(api.EnvHeaders, `{"X-Vault-Test":"value"}`)
+
+	client, err := (&BaseCommand{}).Client()
+	require.Nil(t, client)
+	require.ErrorContains(t, err, "BAO_HEADERS contains a header name with reserved prefix")
+}
```
satoqz: Re-testing within the CLI seems superflous, I'd drop changes in this file.

dm3ch: Dropped

api/client_test.go
dm3ch: Done

website/content/docs/commands/index.mdx
dm3ch: Sorry, missed Flag section. Now is fixed

api/client.go
dm3ch: Applied

dm3ch:
All review findings have been applied / addressed.
The test issues seems to be related to #2581

satoqz (approved):
Thanks! Will leave open for a couple more days in case others would like to review, but will merge if nothing pops up.

api/client.go
```diff
@@ -834,6 +844,26 @@ func NewClient(c *Config) (*Client, error) {
 	return client, nil
 }
 
+func readEnvironmentHeaders() (map[string]string, error) {
+	raw := ReadBaoVariable(EnvHeaders)
+	if raw == "" {
+		return nil, nil
+	}
+
+	var headers map[string]string
```
cipherboy, line 853: Should headers not be `map[string][]string`, especially if we're decoding from JSON? I could see going to `map[string]any` and gracefully handling the difference between `string` and `[]string`.

satoqz: For compatibility with Vault we'll definitely want to handle `map[string]string`, with the upgrade to handling both `string` and `[]string` remaining open if someone asks for it?

website/content/docs/commands/index.mdx
```diff
@@ -430,6 +430,20 @@ Prevents the OpenBao client from following redirects. By default, the OpenBao cl
...
+
+Additional HTTP request headers to send with every CLI request. The value must
+be a JSON object with string values. For example:
+
+```shell-session
+$ export BAO_HEADERS='{"X-IAP-Token":"example-token"}'
+```
+
+Header names beginning with `X-Vault-` are reserved for OpenBao and are not
+allowed.
```
cipherboy, line 443: I'm entertained by this personally; I think its weird that we'd reject this, but not e.g., `X-Forwarded-For` &c.

I think filtering headers may be hard to do; maybe its better to state we accept all headers but they may be clobbered by the API/CLI through other calls? Thoughts?

satoqz: 🤷 It's consistent with `-header` at least, but I'm not opposed to dropping the limitation on both.

api/client.go
cipherboy: Ah, I missed that this was Vault's preference. I'd still suggest handling this properly given the correct header type interface that's used in a lot of places.

PR #4036, opened 2026-09-19
Honor backoff after failed LifetimeWatcher renewals 

### LINKED ISSUE #3747: Openbao proxy retries renew for tokens that are not renewable anymore in an infinite loop

haxonittech:
## Description

Use the interval returned by `errorBackoff.NextBackOff()` after a `LifetimeWatcher` renewal failure. The current loop checks the return value only for `backoff.Stop`; for a retryable error, `sleepDuration` remains zero and the watcher retries immediately.

Add a regression test that records two failed renewal attempts and checks that the second attempt waits for the minimum randomized backoff. The test also stops the watcher and waits for the renewal goroutine to exit.

## Rationale

A renewable token or lease may keep failing renewal while its watcher is still active. For example, this happens when a token cached by the proxy is revoked elsewhere. Immediate retries can create a tight loop of renewal requests against the server.

Applying the existing exponential backoff prevents that loop while retaining the existing stop and lease-expiry checks.

Resolves: #3747

## Testing

From `api/`:

- `go test -count=50 -run '^TestLifetimeWatcher_RenewErrorsBackOff$' .`
- `go test -race -count=20 -run '^TestLifetimeWatcher_RenewErrorsBackOff$' .`
- `go test -run '^(TestRenewer_NewRenewer|TestLifetimeWatcher|TestLifetimeWatcher_RenewErrorsBackOff|TestCalcSleepPeriod)$' -count=3 .`
- `go test -count=1 ./...`

## Acknowledgements

- [x] By contributing this change, I certify I have not used generative AI (GitHub Copilot, Cursor, Claude Code, &c) in authoring these changes or filling out the pull request description or associated issue.
- [x] By contributing this change, I certify I have signed-off on the [DCO ownership](https://developercertificate.org/) statement and this change did not use post-BUSL-licensed code from HashiCorp. Existing MPL-licensed code is still allowed, subject to attribution. Code authored by yourself and submitted to HashiCorp for inclusion is also allowed.


haxonittech:
Edited few things cause I asked claude to run a final check before pushing but idk why and how made PR it self even even though there is AGENT.md. sry for that.

Still it all looks fine to me.

